PR #689, opened 2026-05-12
feat(committees): add iCal Subscribe modal on Meetings tab (LFXV2-1714)

manishdixitlfx:
## Summary
- Replaces the Meetings tab Subscribe button's direct clipboard copy with a DynamicDialog modal.
- Modal exposes the calendar feed URL with a copy control, plus deep-links for Google Calendar, Outlook, and Apple Calendar.
- Feed URL points at `/public/api/committees/:uid/calendar.ics` — the backend is delivered separately by LFXV2-1705.

## Behavioral change
Previous behavior: clicking **Subscribe** copied the iCal URL to the clipboard and surfaced a toast.
New behavior: clicking **Subscribe** opens a modal that lets the user copy the URL or jump straight into Google / Outlook / Apple Calendar.

The Subscribe button remains visible to all users (read-only action — not gated by `canEdit`). Button placement is unchanged (top-right of the Meetings tab, next to the List/Calendar toggle).

## Files
- `apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/` *(new)* — DynamicDialog component with copy + calendar app deep-links
- `apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts` — swaps `Clipboard` + `MessageService` injections for `DialogService`; `onSubscribe()` now opens the dialog

## Test plan
- [ ] Navigate to a committee → Meetings tab and verify the Subscribe button is visible to all users (List view + Calendar view)
- [ ] Click Subscribe; modal opens with the feed URL pre-filled
- [ ] Click Copy → URL is in the clipboard, button label flips to "Copied" for ~2s, success toast appears
- [ ] Click Google Calendar → opens `https://calendar.google.com/calendar/r?cid=webcal://...` in a new tab
- [ ] Click Outlook → opens `https://outlook.live.com/calendar/0/addfromweb?url=...` in a new tab
- [ ] Click Apple Calendar → triggers `webcal://` handler
- [ ] Dismiss the modal by clicking outside (dismissableMask) and via the X
- [ ] Once LFXV2-1705 backend ships, verify the URL returns a valid `text/calendar` payload

## Related
- Backend (separate PR): LFXV2-1705 — `/public/api/committees/:uid/calendar.ics`

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts
@@ -133,19 +133,23 @@ export class CommitteeMeetingsComponent {
   /** Opens the iCal Subscribe modal — copy URL + Google/Outlook/Apple deep links. */
   public onSubscribe(): void {
     const committee = this.committee();
-    if (!committee?.uid) return;
+    if (!committee?.uid) {
+      console.warn('Subscribe clicked with no committee uid; aborting dialog open');
+      return;
+    }
 
     const feedUrl = `${environment.urls.home}/public/api/committees/${committee.uid}/calendar.ics`;
+    const committeeName = committee.name ?? 'Committee';
 
     this.dialogService.open(IcalSubscribeDialogComponent, {
-      header: 'Subscribe to Calendar',
+      header: `Subscribe — ${committeeName}`,
       width: '480px',
       modal: true,
       closable: true,
       dismissableMask: true,
       data: {
         feedUrl,
-        committeeName: committee.name ?? 'Committee',
+        committeeName,
       },
     });
   }
--- a/apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts
@@ -1,7 +1,6 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { Clipboard, ClipboardModule } from '@angular/cdk/clipboard';
 import { ChangeDetectionStrategy, Component, computed, DestroyRef, inject, input, linkedSignal, signal, Signal } from '@angular/core';
 import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
@@ -20,17 +19,18 @@ import { addMinutesToDate } from '@lfx-one/shared/utils';
 import { MeetingService } from '@services/meeting.service';
 import { SurveyService } from '@services/survey.service';
 import { VoteService } from '@services/vote.service';
-import { MessageService } from 'primeng/api';
+import { DialogService } from 'primeng/dynamicdialog';
 import { SkeletonModule } from 'primeng/skeleton';
 import { getCurrentOrNextOccurrence, hasMeetingEnded } from '@lfx-one/shared/utils';
 import { catchError, debounceTime, distinctUntilChanged, filter, finalize, forkJoin, map, of, startWith, switchMap, tap } from 'rxjs';
+
+import { IcalSubscribeDialogComponent } from '../ical-subscribe-dialog/ical-subscribe-dialog.component';
 
 @Component({
   selector: 'lfx-committee-meetings',
   imports: [
     ReactiveFormsModule,
     RouterLink,
-    ClipboardModule,
     ButtonComponent,
     CardComponent,
     InputTextComponent,
@@ -39,6 +39,7 @@ import { catchError, debounceTime, distinctUntilChanged, filter, finalize, forkJ
     MeetingCardComponent,
     FullCalendarComponent,
   ],
+  providers: [DialogService],
   templateUrl: './committee-meetings.component.html',
   styleUrl: './committee-meetings.component.scss',
   changeDetection: ChangeDetectionStrategy.OnPush,
@@ -48,8 +49,7 @@ export class CommitteeMeetingsComponent {
   private readonly voteService = inject(VoteService);
   private readonly surveyService = inject(SurveyService);
   private readonly router = inject(Router);
-  private readonly clipboard = inject(Clipboard);
-  private readonly messageService = inject(MessageService);
+  private readonly dialogService = inject(DialogService);
   private readonly destroyRef = inject(DestroyRef);
 
   // Inputs
@@ -130,16 +130,23 @@ export class CommitteeMeetingsComponent {
       .subscribe((v) => this.searchForm.get('timeFilter')?.setValue(v, { emitEvent: false }));
   }
 
-  /** Copies the committee's calendar subscribe URL to clipboard and shows a confirmation toast. */
+  /** Opens the iCal Subscribe modal — copy URL + Google/Outlook/Apple deep links. */
   public onSubscribe(): void {
-    const uid = this.committee().uid;
-    const url = `${environment.urls.home}/public/api/committees/${uid}/calendar.ics`;
-    this.clipboard.copy(url);
-    this.messageService.add({
-      severity: 'info',
-      summary: 'Subscribe URL copied!',
-      detail: 'Add this to Google Calendar, Outlook, or Apple Calendar.',
-      life: 5000,
+    const committee = this.committee();
+    if (!committee?.uid) return;
+
+    const feedUrl = `${environment.urls.home}/public/api/committees/${committee.uid}/calendar.ics`;
+
+    this.dialogService.open(IcalSubscribeDialogComponent, {
+      header: 'Subscribe to Calendar',
+      width: '480px',
+      modal: true,
+      closable: true,
+      dismissableMask: true,
+      data: {
+        feedUrl,
+        committeeName: committee.name ?? 'Committee',
+      },
     });
   }
 
--- a/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
@@ -3,7 +3,6 @@
 
 <div class="flex flex-col gap-5 p-2" data-testid="ical-subscribe-dialog">
   @if (feedUrl) {
-    <!-- Feed URL with copy -->
     <div class="flex flex-col gap-2">
       <span class="text-sm font-medium text-gray-700">Calendar Feed URL</span>
       <div class="flex items-center gap-2">
@@ -21,7 +20,6 @@
       </div>
     </div>
 
-    <!-- Calendar app buttons -->
     <div class="flex flex-col gap-2">
       <span class="text-sm font-medium text-gray-700">Add to your calendar</span>
       <div class="flex flex-col gap-2">
@@ -66,7 +64,6 @@
       </div>
     </div>
 
-    <!-- Hint text -->
     <p class="text-xs text-gray-400 mt-1">Subscribing keeps your calendar automatically updated when meetings are added or changed.</p>
   } @else {
     <div class="flex flex-col items-center py-8 text-center" data-testid="ical-empty-state">
--- a/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
@@ -1,0 +1,59 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<div class="flex flex-col gap-5 p-2" data-testid="ical-subscribe-dialog">
+  <!-- Feed URL with copy -->
+  <div class="flex flex-col gap-2">
+    <label class="text-sm font-medium text-gray-700">Calendar Feed URL</label>
+    <div class="flex items-center gap-2">
+      <div class="flex-1 bg-gray-50 border border-gray-200 rounded-md px-3 py-2 text-sm text-gray-600 truncate font-mono" data-testid="ical-feed-url">
+        {{ feedUrl }}
+      </div>
+      <lfx-button
+        [icon]="copied() ? 'fa-light fa-check' : 'fa-light fa-copy'"
+        [severity]="copied() ? 'success' : 'secondary'"
+        size="small"
+        [label]="copied() ? 'Copied' : 'Copy'"
+        (onClick)="copyFeedUrl()"
+        data-testid="ical-copy-btn">
+      </lfx-button>
+    </div>
+  </div>
+
+  <!-- Calendar app buttons -->
+  <div class="flex flex-col gap-2">
+    <label class="text-sm font-medium text-gray-700">Add to your calendar</label>
+    <div class="flex flex-col gap-2">
+      <a
+        [href]="googleCalendarUrl"
+        target="_blank"
+        rel="noopener noreferrer"
+        class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
+        data-testid="ical-google-btn">
+        <i class="fa-brands fa-google text-base" aria-hidden="true"></i>
+        Google Calendar
+        <i class="fa-light fa-arrow-up-right-from-square text-xs text-gray-400 ml-auto" aria-hidden="true"></i>
+      </a>
+      <a
+        [href]="outlookUrl"
+        target="_blank"
+        rel="noopener noreferrer"
+        class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
+        data-testid="ical-outlook-btn">
+        <i class="fa-brands fa-microsoft text-base" aria-hidden="true"></i>
+        Outlook
+        <i class="fa-light fa-arrow-up-right-from-square text-xs text-gray-400 ml-auto" aria-hidden="true"></i>
+      </a>
+      <a
+        [href]="webcalUrl"
+        class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
+        data-testid="ical-apple-btn">
+        <i class="fa-brands fa-apple text-base" aria-hidden="true"></i>
+        Apple Calendar
+      </a>
+    </div>
+  </div>
+
+  <!-- Hint text -->
+  <p class="text-xs text-gray-400 mt-1">Subscribing keeps your calendar automatically updated when meetings are added or changed.</p>
+</div>
--- a/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
@@ -5,7 +5,7 @@
   @if (feedUrl) {
     <!-- Feed URL with copy -->
     <div class="flex flex-col gap-2">
-      <label class="text-sm font-medium text-gray-700">Calendar Feed URL</label>
+      <span class="text-sm font-medium text-gray-700">Calendar Feed URL</span>
       <div class="flex items-center gap-2">
         <div class="flex-1 bg-gray-50 border border-gray-200 rounded-md px-3 py-2 text-sm text-gray-600 truncate font-mono" data-testid="ical-feed-url">
           {{ feedUrl }}
@@ -23,7 +23,7 @@
 
     <!-- Calendar app buttons -->
     <div class="flex flex-col gap-2">
-      <label class="text-sm font-medium text-gray-700">Add to your calendar</label>
+      <span class="text-sm font-medium text-gray-700">Add to your calendar</span>
       <div class="flex flex-col gap-2">
         <a
           [href]="googleCalendarUrl"
--- a/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
@@ -2,58 +2,77 @@
 <!-- SPDX-License-Identifier: MIT -->
 
 <div class="flex flex-col gap-5 p-2" data-testid="ical-subscribe-dialog">
-  <!-- Feed URL with copy -->
-  <div class="flex flex-col gap-2">
-    <label class="text-sm font-medium text-gray-700">Calendar Feed URL</label>
-    <div class="flex items-center gap-2">
-      <div class="flex-1 bg-gray-50 border border-gray-200 rounded-md px-3 py-2 text-sm text-gray-600 truncate font-mono" data-testid="ical-feed-url">
-        {{ feedUrl }}
-      </div>
-      <lfx-button
-        [icon]="copied() ? 'fa-light fa-check' : 'fa-light fa-copy'"
-        [severity]="copied() ? 'success' : 'secondary'"
-        size="small"
-        [label]="copied() ? 'Copied' : 'Copy'"
-        (onClick)="copyFeedUrl()"
-        data-testid="ical-copy-btn">
-      </lfx-button>
-    </div>
-  </div>
-
-  <!-- Calendar app buttons -->
-  <div class="flex flex-col gap-2">
-    <label class="text-sm font-medium text-gray-700">Add to your calendar</label>
+  @if (feedUrl) {
+    <!-- Feed URL with copy -->
     <div class="flex flex-col gap-2">
-      <a
-        [href]="googleCalendarUrl"
-        target="_blank"
-        rel="noopener noreferrer"
-        class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
-        data-testid="ical-google-btn">
-        <i class="fa-brands fa-google text-base" aria-hidden="true"></i>
-        Google Calendar
-        <i class="fa-light fa-arrow-up-right-from-square text-xs text-gray-400 ml-auto" aria-hidden="true"></i>
-      </a>
-      <a
-        [href]="outlookUrl"
-        target="_blank"
-        rel="noopener noreferrer"
-        class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
-        data-testid="ical-outlook-btn">
-        <i class="fa-brands fa-microsoft text-base" aria-hidden="true"></i>
-        Outlook
-        <i class="fa-light fa-arrow-up-right-from-square text-xs text-gray-400 ml-auto" aria-hidden="true"></i>
-      </a>
-      <a
-        [href]="webcalUrl"
-        class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
-        data-testid="ical-apple-btn">
-        <i class="fa-brands fa-apple text-base" aria-hidden="true"></i>
-        Apple Calendar
-      </a>
+      <label class="text-sm font-medium text-gray-700">Calendar Feed URL</label>
+      <div class="flex items-center gap-2">
+        <div class="flex-1 bg-gray-50 border border-gray-200 rounded-md px-3 py-2 text-sm text-gray-600 truncate font-mono" data-testid="ical-feed-url">
+          {{ feedUrl }}
+        </div>
+        <lfx-button
+          [icon]="copied() ? 'fa-light fa-check' : 'fa-light fa-copy'"
+          [severity]="copied() ? 'success' : 'secondary'"
+          size="small"
+          [label]="copied() ? 'Copied' : 'Copy'"
+          (onClick)="copyFeedUrl()"
+          data-testid="ical-copy-btn">
+        </lfx-button>
+      </div>
     </div>
-  </div>
 
-  <!-- Hint text -->
-  <p class="text-xs text-gray-400 mt-1">Subscribing keeps your calendar automatically updated when meetings are added or changed.</p>
+    <!-- Calendar app buttons -->
+    <div class="flex flex-col gap-2">
+      <label class="text-sm font-medium text-gray-700">Add to your calendar</label>
+      <div class="flex flex-col gap-2">
+        <a
+          [href]="googleCalendarUrl"
+          target="_blank"
+          rel="noopener noreferrer"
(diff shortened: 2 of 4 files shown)


apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    return `https://calendar.google.com/calendar/r?cid=${encodeURIComponent(this.feedUrl.replace('https://', 'webcal://'))}`;
+  }
+
+  public get outlookUrl(): string {
+    return `https://outlook.live.com/calendar/0/addfromweb?url=${encodeURIComponent(this.feedUrl)}&name=${encodeURIComponent(this.committeeName)}`;
+  }
+
+  public get webcalUrl(): string {
+    return this.feedUrl.replace('https://', 'webcal://');
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Done — replaced the literal `https://` replace with a `toWebcal()` helper that uses `^https?:\/\/` regex, so http URLs (local dev / staging) also convert correctly to `webcal://`.

See commit 8850cf00.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    const success = this.clipboard.copy(this.feedUrl);
+    if (success) {
+      this.copied.set(true);
+      this.messageService.add({
+        severity: 'success',
+        summary: 'Copied',
+        detail: 'Calendar feed URL copied to clipboard',
+      });
+      setTimeout(() => this.copied.set(false), 2000);
+    } else {
```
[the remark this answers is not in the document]
manishdixitlfx: Done — tracking the reset timer in a private `copiedResetTimer` field, clearing the prior timer at the start of each `copyFeedUrl()` call and on destroy via `DestroyRef.onDestroy`. Rapid clicks no longer race, and no callback fires after the dialog tears down.

See commit 8850cf00.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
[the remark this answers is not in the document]
manishdixitlfx: Done — added nullish defaults on `dialogConfig.data?.feedUrl` and `committeeName`, and consolidated the protocol normalization into a single `toWebcal()` helper using `^https?:\/\/` regex (covers the http:// case too).

See commit 8850cf00.

manishdixitlfx:
## Review Feedback Addressed

Commit: 8850cf00

### Changes Made
- **ical-subscribe-dialog.component.ts**: replaced literal `https://` replace with a `toWebcal()` helper using `^https?:\/\/` regex so `http://` URLs (local dev / staging) also convert to `webcal://` (per @copilot-pull-request-reviewer, @coderabbitai)
- **ical-subscribe-dialog.component.ts**: tracked the "Copied" reset timer in a private field; clear the prior timer on each copy and on destroy via `DestroyRef.onDestroy` — rapid clicks no longer race and no callback fires after teardown (per @copilot-pull-request-reviewer)
- **ical-subscribe-dialog.component.ts**: defensive nullish defaults on `dialogConfig.data?.feedUrl` and `committeeName` (per @coderabbitai)

### Threads Resolved

dealako:
## Review feedback

Reviewed against the current head (commit `8850cf00`). The earlier AI-reviewer findings from Copilot + CodeRabbit on `c5b7f04` are all resolved cleanly by `8850cf00` (the `toWebcal()` regex helper, the `copiedResetTimer` field + `DestroyRef.onDestroy` cleanup, and the nullish defaults on `dialogConfig.data`). I'm not re-itemizing those — see the original threads:

- [Copilot — `webcalUrl`/`googleCalendarUrl` http:// handling](https://github.com/linuxfoundation/lfx-self-serve/pull/689#discussion_r3222897712)
- [Copilot — `setTimeout` race / post-teardown firing](https://github.com/linuxfoundation/lfx-self-serve/pull/689#discussion_r3222897730)
- [CodeRabbit — unsafe `dialogConfig.data` access (+ same http:// issue)](https://github.com/linuxfoundation/lfx-self-serve/pull/689#discussion_r3222899294)

### Suggestions

1. **Outlook deep-link only supports consumer (`outlook.live.com`) accounts.** Committee members on Microsoft 365 / enterprise tenants (`outlook.office.com`) won't get a working subscribe flow — the consumer endpoint redirects them out. Consider either:
   - splitting into two buttons ("Outlook.com" vs "Outlook 365"), or
   - using `https://outlook.office.com/calendar/0/addfromweb?url=…` as a second option,
   - or detect/prompt.

   *Ref: `apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts` — `outlookUrl` getter.*

2. **Empty `feedUrl` fallback still renders enabled buttons.** The `?? ''` added in `8850cf00` prevents a throw, but if `data` is ever missing the dialog still renders three deep-link buttons that resolve to `https://calendar.google.com/calendar/r?cid=`, `…outlook.live.com/…?url=&name=`, and a bare `webcal://`. Recommend either guarding `dialogService.open()` upstream when `feedUrl` is missing, or hiding/disabling the deep-link buttons + showing an empty state when `feedUrl === ''`.

3. **No E2E spec for the new dialog.** Per CLAUDE.md + `.claude/rules/development-rules.md`, new components should have E2E coverage. The `data-testid` hooks are already in place (`ical-subscribe-dialog`, `ical-feed-url`, `ical-copy-btn`, `ical-google-btn`, `ical-outlook-btn`, `ical-apple-btn`) — a small spec covering open → copy → "Copied" flip + the three `href` assertions would be high-value and low-effort.

4. **Backend dependency isn't linked in the PR description.** Per `.claude/rules/commit-workflow.md` → "External References", PRs that depend on upstream changes should link the upstream PR (LFXV2-1705 is named but not linked).

### Nits

5. `googleCalendarUrl` / `outlookUrl` / `webcalUrl` are getters that recompute on every CD pass. Since `feedUrl` and `committeeName` are immutable `readonly` fields, these could be `public readonly` strings (or `computed()` signals) set once. Optional — current code is correct. Defining an `IcalDialogData` interface in `@lfx-one/shared/interfaces` would also remove the two `as string` casts and make the caller↔dialog contract explicit.

6. The Apple Calendar `<a>` intentionally lacks `target="_blank"` (the `webcal://` handler hands off to the OS). Worth a one-line code comment so a future contributor doesn't "fix" it for consistency with the Google/Outlook anchors.

### Looks good ✅

- `OnPush` + signals + `inject()` + `data-testid` hierarchy match repo conventions.
- `providers: [DialogService]` correctly scoped to `CommitteeMeetingsComponent`.
- `committee?.uid` guard in `onSubscribe()` is a nice defensive add.
- License headers on all new files; commit subject ≤ 72 chars; `feat(committees)` scope appropriate.

> *Review by Claude Code (Opus 4.7)*

manishdixitlfx:
## Review Feedback Addressed — round 2

Thanks @dealako — addressed 3 of the 6 items in commit 41f5200c.

### Changes Made
- **ical-subscribe-dialog.component.ts**: split the Outlook button into **Outlook.com** (`outlook.live.com`) and **Outlook 365 / Office** (`outlook.office.com`) so enterprise tenants get a working subscribe flow instead of being redirected out (per #1)
- **ical-subscribe-dialog.component.html**: guarded the copy + calendar-app section behind `@if (feedUrl)` with an empty-state fallback when the subscription URL is missing — no more broken deep-links when data is absent (per #2)
- **ical-subscribe-dialog.component.html**: added an HTML comment on the Apple Calendar anchor explaining why `target="_blank"` is intentionally omitted (webcal:// hands off to the OS handler) (per #6)

### Deferred (will follow up separately)
- **E2E spec for the dialog flow** (#3) — agreed it's high value; deferring to a follow-up so it can be properly scoped against a known test committee. Test-ids in place already make this clean to add.
- **`IcalDialogData` interface in `@lfx-one/shared/interfaces`** (#5) — agreed this would remove the `as string` casts; following up in the shared-types pass to keep this PR focused.
- **Link to LFXV2-1705 backend PR in PR description** (#4) — I'll update the PR description once the backend PR is open.

### No Change Needed
None — all suggestions were valid; the three deferrals above are scope choices, not pushback.

### Test-id additions in this round
- `data-testid="ical-outlook-live-btn"` (renamed from `ical-outlook-btn`)
- `data-testid="ical-outlook-365-btn"`
- `data-testid="ical-empty-state"`

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
```diff
@@ -0,0 +1,78 @@
...
+          <i class="fa-light fa-arrow-up-right-from-square text-xs text-gray-400 ml-auto" aria-hidden="true"></i>
+        </a>
+        <!-- No target="_blank" — webcal:// hands off to the OS calendar handler -->
+        <a
+          [href]="webcalUrl"
+          class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
+          data-testid="ical-apple-btn">
+          <i class="fa-brands fa-apple text-base" aria-hidden="true"></i>
+          Apple Calendar
+        </a>
```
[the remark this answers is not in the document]
manishdixitlfx: Thanks for flagging — I dug into this and the current Angular sanitizer doesn't rewrite `webcal://` URLs. The strict allow-list behavior you're describing was changed years ago.

Evidence from this repo (Angular 20):
```
node_modules/@angular/core/fesm2022/debug_node.mjs:5550
const SAFE_URL_PATTERN = /^(?!javascript:)(?:[a-z0-9+.-]+:|[^&:\/?#]*(?:[\/?#]|$))/i;
export function _sanitizeUrl(url) {
  if (url.match(SAFE_URL_PATTERN)) return url;
  // ...
  return 'unsafe:' + url;
}
```

The pattern only blocks `javascript:`. `webcal://...` matches `[a-z0-9+.-]+:` and `_sanitizeUrl` returns the URL unchanged — no `unsafe:` prefix. Verified the Apple Calendar link opens the OS handler correctly during local testing.

Happy to revisit if a future Angular bump changes this.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
```diff
@@ -0,0 +1,80 @@
...
+    return `https://outlook.live.com/calendar/0/addfromweb?url=${encodeURIComponent(this.feedUrl)}&name=${encodeURIComponent(this.committeeName)}`;
+  }
+
+  public get outlook365Url(): string {
+    return `https://outlook.office.com/calendar/0/addfromweb?url=${encodeURIComponent(this.feedUrl)}&name=${encodeURIComponent(this.committeeName)}`;
+  }
+
+  public get webcalUrl(): string {
+    return this.toWebcal(this.feedUrl);
+  }
```
MRashad26: 🔴 **CLAUDE.md violation — 4 getters bound to HTML.**

**Problem & Why:** `googleCalendarUrl`, `outlookLiveUrl`, `outlook365Url`, and `webcalUrl` are TypeScript `get` accessors, and all four are bound from the template (`[href]="googleCalendarUrl"` etc. on `ical-subscribe-dialog.component.html` lines 29, 39, 49, 60). This is the *"No getters for HTML — use signals, computed values, or pipes"* rule in `CLAUDE.md` / `.claude/rules/component-organization.md`. Under `OnPush` the getters still fire on every change-detection pass (every template binding access), even though both inputs (`feedUrl`, `committeeName`) are `readonly` and never change for the lifetime of the dialog. A grep of `apps/lfx-one/src/app/modules/committees/components/*/*.component.ts` finds zero other `public get`/`protected get` in this module — this PR is the only outlier.

**Suggested fix:** Since `feedUrl` and `committeeName` are set once at construction, plain `readonly` fields are sufficient and remove the change-detection cost entirely:
```ts
public readonly googleCalendarUrl = `https://calendar.google.com/calendar/r?cid=${encodeURIComponent(this.toWebcal(this.feedUrl))}`;
public readonly outlookLiveUrl = `https://outlook.live.com/calendar/0/addfromweb?url=${encodeURIComponent(this.feedUrl)}&name=${encodeURIComponent(this.committeeName)}`;
public readonly outlook365Url = `https://outlook.office.com/calendar/0/addfromweb?url=${encodeURIComponent(this.feedUrl)}&name=${encodeURIComponent(this.committeeName)}`;
public readonly webcalUrl = this.toWebcal(this.feedUrl);
```
Move `toWebcal` above these so the field initializer can call it (or inline `replace(/^https?:\/\//i, 'webcal://')`). If reactivity is ever needed later, switch to `computed()`.

MRashad26 (commented):
## Code Standards Audit — PR #689

### Documents Referenced
- Repo `CLAUDE.md` and `.claude/rules/component-organization.md`
- `code-standards-enforcer` agent rules (`~/LFX/code-enforcer-agent.md`)
- Established repo patterns: `description-dialog`, `join-application-dialog`, `mailing-list-picker-dialog` (typed `DynamicDialogConfig<T>` with shared-interface data)

### Files Audited
- `apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts` *(new)*
- `apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html` *(new)*
- `apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts`

### Summary
The feature itself is clean and the previous bot rounds resolved real bugs (http→webcal regex, setTimeout race, destroy cleanup, defensive defaults on dialog data). The remaining issues are mostly **CLAUDE.md conformance** items that the new code drifts on while the surrounding committees module strictly follows them. Two are firm violations that should block merge; the rest are polish.

### Verdict: REQUEST CHANGES
Blocking items:
1. Four template-bound `get` accessors — direct violation of *"No getters for HTML — use signals, computed values, or pipes."*
2. Dialog data shape is untyped (with `as string` casts) — direct violation of *"All interfaces/types live in `@lfx-one/shared/interfaces`,"* and inconsistent with the sibling `DescriptionDialogData` / `JoinApplicationDialogResult` / `MailingListPickerDialogResult` pattern already used in this module.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
```diff
@@ -0,0 +1,80 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class IcalSubscribeDialogComponent {
+  private readonly clipboard = inject(Clipboard);
+  private readonly messageService = inject(MessageService);
+  private readonly dialogConfig = inject(DynamicDialogConfig);
+  private readonly destroyRef = inject(DestroyRef);
+
+  public readonly feedUrl = (this.dialogConfig.data?.feedUrl as string) ?? '';
+  public readonly committeeName = (this.dialogConfig.data?.committeeName as string) ?? 'Committee';
```
MRashad26: 🔴 **CLAUDE.md violation — dialog data shape has no shared interface.**

**Problem & Why:** `this.dialogConfig.data?.feedUrl as string` (and the same on L23) casts away type safety because `DynamicDialogConfig.data` is `any` by default. The repo rule is *all* interfaces — including dialog data contracts — live in `@lfx-one/shared/interfaces`, and the sibling dialogs in this exact module already do this:

- `description-dialog.component.ts:19` — `inject(DynamicDialogConfig<DescriptionDialogData>)`
- `join-application-dialog.component.ts` — imports `JoinApplicationDialogResult` from `@lfx-one/shared/interfaces`
- `mailing-list-picker-dialog.component.ts` — imports `MailingListPickerDialogResult` from `@lfx-one/shared/interfaces`

Without the interface, the caller in `committee-meetings.component.ts` (L72–81) silently accepts any data shape, and `as string` will happily turn a missing field into `undefined as string` at runtime.

**Suggested fix:** In `packages/shared/src/interfaces/`:
```ts
export interface IcalSubscribeDialogData {
  feedUrl: string;
  committeeName: string;
}
```
Export from the barrel, then:
```ts
import { IcalSubscribeDialogData } from '@lfx-one/shared/interfaces';
// ...
private readonly dialogConfig = inject(DynamicDialogConfig<IcalSubscribeDialogData>);
public readonly feedUrl = this.dialogConfig.data?.feedUrl ?? '';
public readonly committeeName = this.dialogConfig.data?.committeeName ?? 'Committee';
```
The `as string` casts disappear, and the caller now gets compile-time errors if it forgets a field.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
```diff
@@ -0,0 +1,78 @@
...
+          [label]="copied() ? 'Copied' : 'Copy'"
+          (onClick)="copyFeedUrl()"
+          data-testid="ical-copy-btn">
+        </lfx-button>
+      </div>
+    </div>
+
+    <!-- Calendar app buttons -->
+    <div class="flex flex-col gap-2">
+      <label class="text-sm font-medium text-gray-700">Add to your calendar</label>
```
MRashad26: 🟠 **`<label>` used as a section heading — a11y / semantic issue.**

**Problem & Why:** Both occurrences ("Calendar Feed URL" on L8 and "Add to your calendar" on L26) are rendered as `<label>` elements but neither is associated with a form control via `for`/`id` or a wrapping input. HTML `<label>` is reserved for labeling form controls. Screen readers may announce "label, no associated control" or pair them with the next focusable element (the Copy button, the first `<a>`), which is misleading. The feed URL is a `<div>` that just displays text; the calendar app links are anchors, not inputs.

**Suggested fix:**
```html
<span class="text-sm font-medium text-gray-700">Calendar Feed URL</span>
<!-- ... -->
<h3 class="text-sm font-medium text-gray-700">Add to your calendar</h3>
```
Or keep both as `<span>` if you'd rather not introduce a heading level inside a dialog. Either is correct; `<label>` is not.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
```diff
@@ -0,0 +1,80 @@
...
+        summary: 'Copied',
+        detail: 'Calendar feed URL copied to clipboard',
+      });
+      if (this.copiedResetTimer !== null) {
+        clearTimeout(this.copiedResetTimer);
+      }
+      this.copiedResetTimer = setTimeout(() => {
+        this.copied.set(false);
+        this.copiedResetTimer = null;
+      }, 2000);
```
MRashad26: 🟡 **`2000` magic number — extract to a named constant.**

**Problem & Why:** The 2-second "Copied" reset is encoded as a literal `2000` with no name to explain what it represents. `CLAUDE.md` directs constants to live in `@lfx-one/shared/constants` (or, at minimum, a named local constant). Future tweaks to UX timing then have a single point of change and an audit trail.

**Suggested fix:** Either at the top of this file or in a shared constants module:
```ts
const COPIED_STATE_RESET_MS = 2000;
// ...
this.copiedResetTimer = setTimeout(() => {
  this.copied.set(false);
  this.copiedResetTimer = null;
}, COPIED_STATE_RESET_MS);
```

apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts
```diff
@@ -130,16 +130,23 @@ export class CommitteeMeetingsComponent {
...
-    const uid = this.committee().uid;
-    const url = `${environment.urls.home}/public/api/committees/${uid}/calendar.ics`;
-    this.clipboard.copy(url);
-    this.messageService.add({
-      severity: 'info',
-      summary: 'Subscribe URL copied!',
-      detail: 'Add this to Google Calendar, Outlook, or Apple Calendar.',
-      life: 5000,
+    const committee = this.committee();
+    if (!committee?.uid) return;
```
MRashad26: 🟡 **Silent `return` on missing `committee.uid` swallows the failure.**

**Problem & Why:** If `this.committee()` is set but `uid` is missing, clicking **Subscribe** does literally nothing — no toast, no log, no dialog. The button looks broken to the user and is invisible to ops because there's no telemetry. The previous version had the same constraint implicitly (it would have built a URL ending in `/undefined/calendar.ics`), but at least *something* happened. With a guarded return, broken state becomes silent.

**Suggested fix:** Either emit a logger warning or surface a toast so the broken case is observable:
```ts
public onSubscribe(): void {
  const committee = this.committee();
  if (!committee?.uid) {
    console.warn('Subscribe clicked with no committee uid');
    return;
  }
  // ...
}
```
If the committee signal is guaranteed non-null on this view by routing, replace the check with an assertion / non-null assertion and document the invariant — silent guards on "impossible" cases tend to hide regressions.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
MRashad26: 🟡 **`committeeName` is captured but never shown in the dialog UI.**

**Problem & Why:** The caller passes `committeeName` and the dialog reads it (L23), but a grep of `ical-subscribe-dialog.component.html` shows it's never rendered — it's only embedded into the Outlook deep-link `&name=` params (L41, L45). The dialog header is hard-coded `'Subscribe to Calendar'` in `committee-meetings.component.ts:73`. So the user opens a generic-looking modal with no contextual confirmation of which committee they're subscribing to. If they had two committee tabs open, they couldn't tell which feed they're about to copy.

**Suggested fix:** Either (a) show the committee name in the dialog body (e.g., a subtitle under the title or above the feed URL: *"Calendar feed for {{ committeeName }}"*), or (b) make the dialog `header` dynamic from the caller side: `header: \`Subscribe — ${committee.name}\``. The second is a one-line change in `committee-meetings.component.ts:73`.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
manishdixitlfx: Done — replaced all four getters (`googleCalendarUrl`, `outlookLiveUrl`, `outlook365Url`, `webcalUrl`) with `public readonly` field initializers. Since `feedUrl` and `committeeName` are set once at construction, this removes the per-CD-pass cost entirely. `toWebcal()` is now a module-level helper so the field initializers can call it.

See commit 4d4898d3.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
manishdixitlfx: Done — added `IcalSubscribeDialogData` to `packages/shared/src/interfaces/committee.interface.ts` alongside the sibling dialog data shapes (`DescriptionDialogData`, `JoinApplicationDialogData`, `MailingListPickerDialogData`). The dialog now uses `inject<DynamicDialogConfig<IcalSubscribeDialogData>>(...)` and the `as string` casts are gone — the caller↔dialog contract is compile-checked.

See commit 4d4898d3.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
manishdixitlfx: Fixed — both "Calendar Feed URL" and "Add to your calendar" headings are now `<span>` elements. Good catch on the a11y/semantic issue.

See commit 4d4898d3.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
manishdixitlfx: Fixed — extracted to `COPIED_STATE_RESET_MS = 2000` at the top of the file. Kept it local since the value is dialog-specific UX timing; happy to promote to `@lfx-one/shared/constants` if you want it discoverable from other components.

See commit 4d4898d3.

apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.ts
manishdixitlfx: Fixed — added a `console.warn("Subscribe clicked with no committee uid; aborting dialog open")` before the early return so the broken case is observable. The committee input is declared as `input.required<Committee>()` so this path should be unreachable in practice, but the warn protects against future regressions.

See commit 4d4898d3.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.ts
manishdixitlfx: Fixed — made the dialog header dynamic in `onSubscribe()`: `header: \`Subscribe — ${committeeName}\``. The user now sees the committee name in the modal title, which addresses the "two committee tabs open" disambiguation case.

See commit 4d4898d3.

manishdixitlfx:
## Review Feedback Addressed — round 3

Thanks @MRashad26 — addressed all 6 items in commit 4d4898d3.

### Changes Made
- **ical-subscribe-dialog.component.ts**: replaced 4 `get` accessors with `public readonly` field initializers; moved `toWebcal()` to a module-level helper. Removes per-CD-pass cost and conforms to the "No getters for HTML" rule (per #1, blocker)
- **packages/shared/src/interfaces/committee.interface.ts**: added `IcalSubscribeDialogData` interface alongside the sibling dialog data shapes; dialog now consumes it via `inject<DynamicDialogConfig<IcalSubscribeDialogData>>(...)`. The `as string` casts are gone (per #2, blocker)
- **ical-subscribe-dialog.component.html**: replaced two unattached `<label>` section headings with `<span>` (per #3, a11y)
- **ical-subscribe-dialog.component.ts**: extracted `COPIED_STATE_RESET_MS = 2000` (per #4)
- **committee-meetings.component.ts**: added `console.warn` before the silent early-return so the broken-uid case is observable (per #5)
- **committee-meetings.component.ts**: made the dialog header dynamic — `Subscribe — ${committeeName}` — so the user can see which committee they're subscribing to (per #6)

### Threads Resolved

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
```diff
@@ -0,0 +1,78 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<div class="flex flex-col gap-5 p-2" data-testid="ical-subscribe-dialog">
+  @if (feedUrl) {
+    <!-- Feed URL with copy -->
```
MRashad26: **Section-header comment — remove it.**

`<!-- Feed URL with copy -->` labels WHAT the block is. The `data-testid="ical-feed-url"` on the display div and the `<span>Calendar Feed URL</span>` heading text already make the block self-describing. Delete the comment line.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
```diff
@@ -0,0 +1,78 @@
...
+          [severity]="copied() ? 'success' : 'secondary'"
+          size="small"
+          [label]="copied() ? 'Copied' : 'Copy'"
+          (onClick)="copyFeedUrl()"
+          data-testid="ical-copy-btn">
+        </lfx-button>
+      </div>
+    </div>
+
+    <!-- Calendar app buttons -->
```
MRashad26: **Section-header comment — remove it.**

`<!-- Calendar app buttons -->` labels WHAT the block is. The `<span>Add to your calendar</span>` heading text + `data-testid` attributes on each `<a>` already document this section. Delete the comment line.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
```diff
@@ -0,0 +1,78 @@
...
+          [href]="webcalUrl"
+          class="flex items-center gap-3 px-4 py-3 rounded-lg border border-gray-200 hover:bg-gray-50 transition-colors text-sm font-medium text-gray-700 no-underline"
+          data-testid="ical-apple-btn">
+          <i class="fa-brands fa-apple text-base" aria-hidden="true"></i>
+          Apple Calendar
+        </a>
+      </div>
+    </div>
+
+    <!-- Hint text -->
```
MRashad26: **Section-header comment — remove it.**

`<!-- Hint text -->` labels WHAT the `<p>` is. The paragraph content is self-evident — no comment needed. Delete the comment line.

Note: the `<!-- No target="_blank" — webcal:// hands off to the OS calendar handler -->` comment above the Apple Calendar `<a>` is fine — it explains WHY (a non-obvious browser/OS behaviour), not WHAT.

MRashad26 (commented):
All 6 prior review items confirmed addressed in commit `4d4898d3`. Three new CLAUDE.md violations found in the HTML template by the code-standards-enforcer pass.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
manishdixitlfx: Fixed — removed the `<!-- Feed URL with copy -->` comment. Agreed, the heading text + data-testids made it redundant.

See commit 00626f20.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
manishdixitlfx: Fixed — removed the `<!-- Calendar app buttons -->` comment. Agreed, the heading text + data-testids made it redundant.

See commit 00626f20.

apps/lfx-one/src/app/modules/committees/components/ical-subscribe-dialog/ical-subscribe-dialog.component.html
manishdixitlfx: Fixed — removed the `<!-- Hint text -->` comment. Agreed, the heading text + data-testids made it redundant.

See commit 00626f20.

manishdixitlfx:
## Review Feedback Addressed — round 4

Thanks @MRashad26 — addressed all 3 items in commit 00626f20.

### Changes Made
- **ical-subscribe-dialog.component.html**: removed three section-header comments (`<!-- Feed URL with copy -->`, `<!-- Calendar app buttons -->`, `<!-- Hint text -->`) — the surrounding markup (data-testids + visible heading text) is self-describing (per @MRashad26)

### Threads Resolved

The `<!-- No target="_blank" — webcal:// hands off to the OS calendar handler -->` comment on the Apple Calendar anchor is intentionally retained — explains WHY, not WHAT.

MRashad26 (approved):

PR #694, opened 2026-05-12
fix(profile): unblock linkedin add identity flow

fayazg:
## Summary

- **Lazy env getters in `ProfileAuthService` and `SocialVerificationService`** — both services previously read their env vars (`PROFILE_CLIENT_ID`, `PROFILE_CLIENT_SECRET`, etc.) into `readonly` instance fields inside their constructors. Both are instantiated by `ProfileController` during ESM import resolution, which runs before `dotenv.config()` in `server.ts`, so the fields were captured as empty strings. `isProfileAuthConfigured()` returned `false` and the Flow C redirect silently bounced back to `/profile/identities?error=profile_auth_not_configured`. Converts both services to the lazy-getter pattern already used by `CdpService` / `CredlyService` so env vars are read on first access (after dotenv has loaded).
- **Skip identities fetch during SSR** in `profile-identities.component.ts` — the server-side HTTP call to `/api/profile/identities` doesn't reliably carry the user's session cookie, the fetch tended to fail, and the SSR HTML rendered the "Identity service unavailable" banner. That caused a red-banner flash on hydration before the browser's authenticated fetch resolved. Gates the observable inside `initIdentitiesState()` with `isPlatformBrowser(this.platformId)` so SSR renders a loading state and the browser owns the request.

LFXV2-1762

## Test plan

- [ ] Restart the dev server (mandatory — the dotenv race only manifests on cold start with stale instance fields).
- [ ] Click **Add Identity → LinkedIn** on `/profile/identities` and confirm the browser redirects to Auth0 `/authorize` (URL changes to `linuxfoundation-dev.auth0.com/authorize?…`) instead of bouncing back with `?error=profile_auth_not_configured`.
- [ ] Server log shows `profile_auth_start` success, not `WARN: Profile auth not configured`.
- [ ] Hard-reload `/profile/identities` and confirm no red "Identity service unavailable" banner flash on initial load.
- [ ] Regression: existing email OTP verification flow (`/profile/emails`) still works.
- [ ] Regression: existing password change flow (`/profile/password`) still works.
- [ ] Negative test: temporarily blank `PROFILE_CLIENT_ID` in `.env`, restart, confirm the `Profile auth not configured` warning still fires for genuinely unconfigured environments.
- [ ] `yarn format:check && yarn lint:check && yarn check-types && yarn build` pass.

### DIFF
--- a/apps/lfx-one/src/server/services/profile-auth.service.ts
+++ b/apps/lfx-one/src/server/services/profile-auth.service.ts
@@ -22,22 +22,42 @@ interface TokenResponse {
  * update:current_user_metadata and update:current_user_identities scopes.
  */
 export class ProfileAuthService {
-  private readonly clientId: string;
-  private readonly clientSecret: string;
-  private readonly audience: string;
-  private readonly scope: string;
-  private readonly issuerBaseUrl: string;
-  private readonly baseUrl: string;
-  private readonly redirectUri: string;
+  // Resolved lazily on first access so dotenv has finished loading,
+  // then memoized — env is stable after startup.
+  private _clientId: string | undefined;
+  private _clientSecret: string | undefined;
+  private _audience: string | undefined;
+  private _scope: string | undefined;
+  private _issuerBaseUrl: string | undefined;
+  private _baseUrl: string | undefined;
+  private _redirectUri: string | undefined;
 
-  public constructor() {
-    this.clientId = process.env['PROFILE_CLIENT_ID'] || '';
-    this.clientSecret = process.env['PROFILE_CLIENT_SECRET'] || '';
-    this.audience = process.env['PROFILE_AUDIENCE'] || '';
-    this.scope = process.env['PROFILE_SCOPE'] || '';
-    this.issuerBaseUrl = (process.env['PCC_AUTH0_ISSUER_BASE_URL'] || '').replace(/\/+$/, '');
-    this.baseUrl = process.env['PCC_BASE_URL'] || 'http://localhost:4000';
-    this.redirectUri = process.env['PROFILE_REDIRECT_URI'] || `${this.baseUrl}/passwordless/callback`;
+  private get clientId(): string {
+    return (this._clientId ??= process.env['PROFILE_CLIENT_ID'] || '');
+  }
+
+  private get clientSecret(): string {
+    return (this._clientSecret ??= process.env['PROFILE_CLIENT_SECRET'] || '');
+  }
+
+  private get audience(): string {
+    return (this._audience ??= process.env['PROFILE_AUDIENCE'] || '');
+  }
+
+  private get scope(): string {
+    return (this._scope ??= process.env['PROFILE_SCOPE'] || '');
+  }
+
+  private get issuerBaseUrl(): string {
+    return (this._issuerBaseUrl ??= (process.env['PCC_AUTH0_ISSUER_BASE_URL'] || '').replace(/\/+$/, ''));
+  }
+
+  private get baseUrl(): string {
+    return (this._baseUrl ??= process.env['PCC_BASE_URL'] || 'http://localhost:4000');
+  }
+
+  private get redirectUri(): string {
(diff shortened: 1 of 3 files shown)


apps/lfx-one/src/server/services/profile-auth.service.ts
```diff
@@ -22,22 +22,42 @@ interface TokenResponse {
...
+  // then memoized — env is stable after startup.
+  private _clientId: string | undefined;
+  private _clientSecret: string | undefined;
+  private _audience: string | undefined;
+  private _scope: string | undefined;
+  private _issuerBaseUrl: string | undefined;
+  private _baseUrl: string | undefined;
+  private _redirectUri: string | undefined;
+
+  private get clientId(): string {
```
dealako, line 35: Nice fix — the lazy `??=` pattern matches `CdpService` / `CredlyService` exactly and the inline comments nail the *why* behind the change.

**Non-blocking follow-up worth a ticket:** `ai.service.ts:14–17` has the same eager-env anti-pattern, and `MeetingController` instantiates `AiService` at module top-level (`meeting.controller.ts:42`), so it is exposed to the same dotenv race:

```ts
// ai.service.ts:14 — reads at class field init time
private readonly aiKey = process.env["AI_API_KEY"] || "sk-proj-1234567890";

// ai.service.ts:17 — reads in constructor
this.aiProxyUrl = process.env["AI_PROXY_URL"] || "https://api.openai.com/v1/chat/completions";
```

It silently "works" because the defaults are plausible-looking strings that pass the configured-check — so if the race triggers, you get upstream 401s instead of a clean "not configured" error, which is harder to diagnose. Lazifying those two for symmetry with this PR would make the pattern fully consistent across services, but it is explicitly out of scope here.

dealako (commented):
## Review: `fix(profile): unblock linkedin add identity flow`

**Both CodeRabbit and Copilot reviewed this PR and generated no actionable comments** ([CodeRabbit summary](https://github.com/linuxfoundation/lfx-self-serve/pull/694#issuecomment-4432593832) · [Copilot review](https://github.com/linuxfoundation/lfx-self-serve/pull/694#pullrequestreview-PRR_kwDOPMiRQ87-xAiI)) — I agree with their assessment. No duplicate findings to flag.

---

### What I looked at

- `profile-auth.service.ts` and `social-verification.service.ts` — the lazy-getter refactor
- `profile-identities.component.ts` — the SSR platform guard
- The surrounding context (controller instantiation points, other services with similar patterns)

---

### Observations

**The diagnosis is sharp.** Pinpointing the constructor-vs-dotenv race (services instantiated at ESM import time, before `dotenv.config()` runs) is exactly the kind of subtle startup ordering issue that's easy to overlook. The PR description explains both root causes clearly.

**Lazy getters — correct and consistent.** The `(this._x ??= process.env[...] || '')` memoization pattern exactly matches `CdpService` / `CredlyService`. Env is stable after startup, so memoizing on first access is the right trade-off.

**SSR guard — right design choice.** Returning `{ identities: [], loaded: false }` (not `loaded: true`) means the browser sees a loading state, not an empty state, on hydration. That's the correct call — it avoids the flash-of-empty-state race and lets the browser own the authenticated request cleanly.

**Inline comments are excellent.** Each explains *why* the change exists rather than restating what the code does — which is exactly what future maintainers need.

---

### Non-blocking nudge (see inline)

I left [one inline comment](https://github.com/linuxfoundation/lfx-self-serve/pull/694#discussion_r3228490221) on `profile-auth.service.ts` noting that `ai.service.ts:14–17` has the same eager-env pattern (instantiated by `MeetingController` at module top-level). It silently passes the configured-check because its defaults are plausible placeholder strings, so the race would surface as upstream 401s rather than a clean error. Not in scope for this PR — worth a follow-up ticket for symmetry.

---

### Test plan reminder

The checkboxes in the PR description (`Restart dev server → Add Identity → LinkedIn → confirm redirect`) are still unchecked. Please confirm those steps pass before merging.

---

Great work tracking down two independent but related initialization bugs and fixing both cleanly. The PR is small, focused, and the changes are easy to verify. 🚢

dealako (approved):
Approved with one minor feedback.

PR #701, opened 2026-05-13
feat(routing): add project deep link support via ?project=<slug>

MRashad26:
## Summary

- Adds `projectQueryParamGuard` — a `CanActivateFn` that reads `?project=<slug>` and seeds `ProjectContextService` with the matching project before the lens preloader fires, so cold-loaded deep links land on the correct project
- Attaches the guard to all `project/*` and `foundation/*` child routes in `app.routes.ts` (composes cleanly with existing `writerGuard` / `executiveDirectorGuard`)
- Syncs `?project=<slug>` in `ProjectContextService` after every `setProject` / `setFoundation` mutation, keeping the address bar copy-paste-safe after sidebar switches

## JIRA

[LFXV2-1773](https://linuxfoundation.atlassian.net/browse/LFXV2-1773)

## Test plan

- [ ] Cold-load `/project/meetings?project=<slug>` in a fresh tab — correct project selected, page renders project data
- [ ] Cold-load `/project/overview`, `/project/votes`, `/project/settings` with the same `?project=` — all land on the right project + page
- [ ] Cmd+R on a deep link — URL stays, project stays selected
- [ ] Switch project in sidebar — URL updates to `?project=<new-slug>` without a full reload; back button does NOT travel back to the previous project (`replaceUrl: true`)
- [ ] `?project=this-does-not-exist` — page renders with the default priority project for the lens, no console errors, no redirect loop
- [ ] `/project/meetings` with no query param — existing default-selection behaviour unchanged
- [ ] `/project/settings?project=<writable-project>` — settings load; `?project=<read-only-project>` still redirects via existing `writerGuard`
- [ ] No console errors in any of the above flows

[LFXV2-1773]: https://linuxfoundation.atlassian.net/browse/LFXV2-1773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/src/app/app.routes.ts
```diff
@@ -24,30 +25,34 @@ export const routes: Routes = [
...
         data: { lens: 'foundation' },
+        canActivate: [projectQueryParamGuard],
         loadComponent: () => import('./modules/dashboards/foundation-projects/foundation-projects.component').then((m) => m.FoundationProjectsComponent),
       },
       {
         path: 'foundation/events',
         data: { lens: 'foundation' },
+        canActivate: [projectQueryParamGuard],
         loadChildren: () => import('./modules/events/events.routes').then((m) => m.EVENTS_ROUTES),
       },
```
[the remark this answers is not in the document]
MRashad26: Fixed — removed the duplicate `foundation/events` entry at lines 45-50. That block was a pre-existing duplicate that predated this PR; the canonical definition is the one in the "Foundation Lens — feature routes" section (now line 66). Good catch!

See commit `03fc85dd`.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (!project) return true;
+      projectContextService.setProject({
+        uid: project.uid,
+        name: project.name,
+        slug: project.slug,
+        parent_uid: project.parent_uid,
+        logoUrl: project.logo_url,
+      });
+      return true;
+    }),
```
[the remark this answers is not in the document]
MRashad26: Fixed — the guard now branches on `route.data['lens'] === 'foundation'` and calls `setFoundation()` for foundation routes and `setProject()` for project routes. Without this, deep linking to a foundation page would have silently set projectSelection instead of foundationSelection, leaving the foundation context untouched.

See commit `03fc85dd`.

MRashad26:
## Review Feedback Addressed

Commit: `03fc85dd`

### Changes Made
- **`app.routes.ts`**: Removed the duplicate `foundation/events` route entry (lines 45-50). That entry was a pre-existing duplicate; the canonical definition in the "Foundation Lens — feature routes" section is kept (per @coderabbitai)
- **`project-query-param.guard.ts`**: Guard now branches on `route.data['lens'] === 'foundation'` — calls `setFoundation()` for foundation routes and `setProject()` for project routes, ensuring deep links correctly seed the right context signal (per @coderabbitai)

### Threads Resolved

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+    map((project) => {
+      if (!project) return true;
+      const context = {
+        uid: project.uid,
+        name: project.name,
+        slug: project.slug,
+        parent_uid: project.parent_uid,
+        logoUrl: project.logo_url,
+      };
+      if (route.data?.['lens'] === 'foundation') {
```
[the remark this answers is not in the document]
MRashad26: Already addressed in commit `03fc85dd` — the guard now branches on `route.data?.['lens'] === 'foundation'` and calls `setFoundation()` for foundation routes and `setProject()` for project routes. No change needed.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -60,77 +59,92 @@ export const routes: Routes = [
...
       {
         path: 'foundation/mailing-lists',
         data: { lens: 'foundation' },
+        canActivate: [projectQueryParamGuard],
         loadChildren: () => import('./modules/mailing-lists/mailing-lists.routes').then((m) => m.MAILING_LIST_ROUTES),
       },
```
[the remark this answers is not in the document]
MRashad26: Already addressed in commit `03fc85dd` — the duplicate `foundation/events` entry (the early standalone one) was removed. Only the canonical definition in the 'Foundation Lens — feature routes' section remains.

apps/lfx-one/src/app/shared/services/project-context.service.ts
```diff
@@ -48,21 +50,37 @@ export class ProjectContextService {
...
+  /**
+   * Updates the ?project= query param in the current URL without triggering a new navigation.
+   * Skipped when a navigation is already in flight — the URL already carries the correct param
+   * (deep-link case) or the caller's own navigation will set the destination URL (openRow case).
+   */
+  private syncProjectQueryParam(slug: string | null): void {
+    if (this.router.getCurrentNavigation()) {
+      return;
+    }
+    void this.router.navigate([], { queryParams: { project: slug }, queryParamsHandling: 'merge', replaceUrl: true });
```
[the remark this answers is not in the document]
MRashad26: Fixed in commit `19790ec0` — switched `syncProjectQueryParam` to use `Location.replaceState()` via `router.parseUrl()` + `router.serializeUrl()`. The URL is now updated without triggering any Angular navigation, so guards and resolvers are not re-evaluated on sidebar project switches. Updated the method comment to accurately describe this. Good catch on the inaccuracy.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export const projectQueryParamGuard: CanActivateFn = (route) => {
+  const slug = route.queryParamMap.get('project');
+  if (!slug) return true;
+
+  const projectService = inject(ProjectService);
+  const projectContextService = inject(ProjectContextService);
+
+  return projectService.getProject(slug, false).pipe(
+    map((project) => {
```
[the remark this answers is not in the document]
MRashad26: Skipping this — the backend already validates and rejects invalid slugs with a 404, and `getProject()` returns `null` for any error response, so the guard handles it safely. Adding a client-side regex would duplicate that responsibility without meaningful benefit.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -24,13 +25,14 @@ export const routes: Routes = [
...
+        canActivate: [projectQueryParamGuard],
         loadComponent: () => import('./modules/dashboards/dashboard.component').then((m) => m.DashboardComponent),
       },
       // Foundation Lens — Health Metrics page (ED-only)
       {
         path: 'foundation/health-metrics',
         data: { lens: 'foundation' },
-        canActivate: [executiveDirectorGuard],
+        canActivate: [projectQueryParamGuard, executiveDirectorGuard],
         loadComponent: () => import('./modules/dashboards/health-metrics/health-metrics.component').then((m) => m.HealthMetricsComponent),
```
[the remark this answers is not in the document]
MRashad26: Fixed in commit `19790ec0` — added `projectQueryParamGuard` to the `foundation/marketing-impact` route so it's consistent with all other foundation deep-link routes.

apps/lfx-one/src/app/shared/services/project-context.service.ts
```diff
@@ -1,10 +1,12 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
+import { Location } from '@angular/common';
 import { computed, inject, Injectable, Signal, signal, WritableSignal } from '@angular/core';
 import { toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { isBoardScopedPersona, ProjectContext } from '@lfx-one/shared/interfaces';
 import { isSameProjectContext } from '@lfx-one/shared/utils';
+import { Router } from '@angular/router';
```
dealako: **[minor]** `import { Router } from '@angular/router'` sits between the `@lfx-one/shared/*` imports (lines 7–8) and the `ngx-cookie-service-ssr` import (line 10), breaking the expected grouping. Angular framework imports (`@angular/*`) should be grouped together — `Location` (line 4) and `Router` (line 9) should be adjacent. `yarn lint:check` may catch this.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ProjectContextService } from '../services/project-context.service';
+import { ProjectService } from '../services/project.service';
+
+/**
+ * Seeds the active project/foundation context from a `?project=<slug>` query param.
+ * Returns true in all branches — this guard sets context, not access control.
+ * If the slug is missing, invalid, or resolves to nothing, navigation continues normally
+ * and NavigationService.applyDefaultSelection handles the fallback selection.
+ */
+export const projectQueryParamGuard: CanActivateFn = (route) => {
```
dealako, line 18: **[minor]** No unit test for this guard. I know, we lack unit tests everywhere. 😂 Four branches worth covering: (1) no slug → early return `true`, (2) invalid slug / fetch error → `catchError` path, (3) `lens === 'foundation'` → `setFoundation`, (4) default → `setProject`. A `TestBed` spec with a stubbed `ProjectService` covers all four concisely. The manual test plan in the PR body validates the UX flows well — flagging this only as a follow-up opportunity.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+  return projectService.getProject(slug, false).pipe(
+    map((project) => {
+      if (!project) return true;
+      const context = {
+        uid: project.uid,
+        name: project.name,
+        slug: project.slug,
+        parent_uid: project.parent_uid,
+        logoUrl: project.logo_url,
+      };
```
dealako, line 34: **[nit]** `context` is implicitly typed. If `ProjectContext` gains a new required field, this literal would produce a type error only at call sites, not here. Explicit annotation catches drift at the source (consider adding the annotation):

```ts
const context: ProjectContext = {
  uid: project.uid,
  name: project.name,
  slug: project.slug,
  parent_uid: project.parent_uid,
  logoUrl: project.logo_url,
};
```

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
dealako: **[nit]** `ActivatedRouteSnapshot.data` is typed as `Data` — always-defined — so the optional chain (`?.`) is unnecessary noise. `route.data['lens']` is sufficient. Do you agree?

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -67,77 +66,92 @@ export const routes: Routes = [
       {
         path: 'foundation/meetings',
```
dealako, line 67: **[nit]** Every `foundation/*` and `project/*` child route now repeats both `data: { lens: '...' }` and `canActivate: [projectQueryParamGuard]`. Nothing to act on in this PR, but a future refactor could group these under two parent routes and hoist the shared config — would cut roughly 20 lines and make new routes easier to wire correctly. ✂️

dealako (approved):
Hi Rashad — solid, well-scoped PR. The deep-link guard is the right shape (sets context, doesn't gate access), the lens-branching fix in round 2 was important, and the follow-up commits showed good attention to corner cases: guarding URL sync against in-flight navigations (round 3) and then — nice touch — replacing `router.navigate()` with `Location.replaceState()` entirely (round 5) so guard/resolver cycles are never a factor on sidebar switches. The `foundation/marketing-impact` guard gap caught in round 5 was also a good find.

Nothing here is blocking.

**Blocking: 0 · Minor: 2 · Nit: 3**

**Decision: approved with minor comments** — feel free to ship; both minor items (import grouping + explicit `ProjectContext` type annotation) can land here or in a follow-up.

---

### Revision tracking

| Round | Commit | What landed |
| --- | --- | --- |
| 1 | `372c0bd6` | Initial guard + URL sync wiring |
| 2 | `03fc85dd` | CodeRabbit / Copilot: removed duplicate `foundation/events`, branched guard on `data.lens` for foundation vs project context |
| 3 | `569c496d` | Self-found: guarded `syncProjectQueryParam` with `getCurrentNavigation()` to prevent guard/applyDefaultSelection/openRow from cancelling each other |
| 4 | `2c7f5798` | Merge `main` |
| 5 | `19790ec0` | Copilot follow-up: `syncProjectQueryParam` → `Location.replaceState` (no router cycle); added missing `projectQueryParamGuard` to `foundation/marketing-impact` |

### AI bot reconciliation

| Bot | Thread | Resolved? | My take |
| --- | --- | --- | --- |
| **CodeRabbit** [r3235516673](https://github.com/linuxfoundation/lfx-self-serve/pull/701#discussion_r3235516673) | Duplicate `foundation/events` route | ✅ `03fc85dd` | Agree — correct fix, canonical entry in feature-routes section kept. |
| **CodeRabbit** [r3235516688](https://github.com/linuxfoundation/lfx-self-serve/pull/701#discussion_r3235516688) | Guard always called `setProject` on foundation routes | ✅ `03fc85dd` | Agree — important correctness fix; wrong setter would silently leave `foundationSelection` unset. |
| **Copilot** [r3235533419](https://github.com/linuxfoundation/lfx-self-serve/pull/701#discussion_r3235533419) | Same as CodeRabbit r3235516688 | ✅ `03fc85dd` (duplicate) | Agreement; no independent action needed. |
| **Copilot** [r3235533479](https://github.com/linuxfoundation/lfx-self-serve/pull/701#discussion_r3235533479) | Same as CodeRabbit r3235516673 | ✅ `03fc85dd` (duplicate) | Agreement; no independent action needed. |
| **Cursor Bot** | — | n/a | No Cursor comments on this PR. |

All prior review threads are resolved.

### Verified fine

- `Location.replaceState` is SSR-safe: Angular's DI provides a memory-backed `Location` on the server; `getCurrentNavigation()` returning truthy during SSR's initial navigation means `syncProjectQueryParam` self-skips anyway.
- `router.parseUrl(router.url)` + `serializeUrl` correctly roundtrips the full current URL — path and all other query params are preserved.
- `replaceUrl: true` → replaced by `replaceState` — same result: no history entry pushed, back button unaffected.
- Invalid slug (fetch error / null project): guard falls through; `NavigationService.applyDefaultSelection` seeds the default and triggers `syncProjectQueryParam` post-navigation, rewriting the URL. Working as designed.
- `router.navigate([], ...)` for query-param-only updates (no `relativeTo`) — established pattern in this repo; current path is preserved.

apps/lfx-one/src/app/shared/services/project-context.service.ts
dealako: **[minor]** `import { Router } from '@angular/router'` sits between the `@lfx-one/shared/*` imports (lines 7–8) and the `ngx-cookie-service-ssr` import (line 10), breaking the expected grouping. Angular framework imports (`@angular/*`) should be grouped together — `Location` (line 4) and `Router` (line 9) should be adjacent. `yarn lint:check` may catch this.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
dealako, line 18: **[minor]** No unit test for this guard. Four branches worth covering: (1) no slug → early return `true`, (2) invalid slug / fetch error → `catchError` path, (3) `lens === 'foundation'` → `setFoundation`, (4) default → `setProject`. A `TestBed` spec with a stubbed `ProjectService` covers all four concisely. The manual test plan in the PR body validates the UX flows well — flagging this only as a follow-up opportunity.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
dealako, line 34: **[nit]** `context` is implicitly typed. If `ProjectContext` gains a new required field, this literal would produce a type error only at call sites, not here. Explicit annotation catches drift at the source:

```ts
const context: ProjectContext = {
  uid: project.uid,
  name: project.name,
  slug: project.slug,
  parent_uid: project.parent_uid,
  logoUrl: project.logo_url,
};
```

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
dealako: **[nit]** `ActivatedRouteSnapshot.data` is typed as `Data` — always-defined — so the optional chain (`?.`) is unnecessary noise. `route.data['lens']` is sufficient.

apps/lfx-one/src/app/app.routes.ts
dealako, line 67: **[nit]** Every `foundation/*` and `project/*` child route now repeats both `data: { lens: '...' }` and `canActivate: [projectQueryParamGuard]`. Nothing to act on in this PR, but a future refactor could group these under two parent routes and hoist the shared config — would cut roughly 20 lines and make new routes easier to wire correctly.

dealako (dismissed):
Tentatively approved, but there are a few nits that I listed you should consider. Not blocking, but it would be nice to have these cleaned up.

apps/lfx-one/src/app/shared/services/project-context.service.ts
MRashad26: Fixed in `578e1848` — moved `Router` to sit adjacent to `Location` so all `@angular/*` imports are grouped together. Good catch!

MRashad26: Fixed in `578e1848` — moved `Router` to sit adjacent to `Location` so all `@angular/*` imports are grouped together. Good catch!

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
MRashad26: Agreed — those four branches are well worth testing. Tracking this as a follow-up; the guard logic is stable enough at this point that tests can land separately without blocking the feature.

MRashad26: Agreed — those four branches are well worth testing. Tracking this as a follow-up; the guard logic is stable enough at this point that tests can land separately without blocking the feature.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
MRashad26: Fixed in `578e1848` — annotated as `const context: ProjectContext`. Agreed: catching drift at the literal is better than at call sites.

MRashad26: Fixed in `578e1848` — annotated as `const context: ProjectContext`. Agreed: catching drift at the literal is better than at call sites.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
MRashad26: Fixed in `578e1848` — changed to `route.data['lens']`. You're right that `ActivatedRouteSnapshot.data` is typed as `Data` (always defined), so the `?.` was noise.

MRashad26: Fixed in `578e1848` — changed to `route.data['lens']`. You're right that `ActivatedRouteSnapshot.data` is typed as `Data` (always defined), so the `?.` was noise.

apps/lfx-one/src/app/app.routes.ts
MRashad26: Fully agree — parent routes with hoisted `data` and `canActivate` would be the clean long-term shape. Leaving that for a dedicated refactor PR so this one stays focused on the feature.

MRashad26: Fully agree — parent routes with hoisted `data` and `canActivate` would be the clean long-term shape. Leaving that for a dedicated refactor PR so this one stays focused on the feature.

MRashad26:
## Review Feedback Addressed

Commit: \`578e1848\`

### Changes Made
- **`project-context.service.ts`**: Moved `Router` import adjacent to `Location` so all `@angular/*` imports are grouped together (per @dealako)
- **`project-query-param.guard.ts`**: Annotated `context` as `const context: ProjectContext` so a missing required field is caught at the literal, not at call sites (per @dealako)
- **`project-query-param.guard.ts`**: Removed unnecessary `?.` on `route.data` — `ActivatedRouteSnapshot.data` is typed as `Data` (always defined) (per @dealako)

### No Change Needed
- **Unit tests**: Acknowledged as a follow-up — guard logic is stable and tests can land in a separate PR without blocking the feature (per @dealako)
- **Route repetition**: Agreed a parent-route refactor would be cleaner; leaving for a dedicated follow-up PR to keep this one focused (per @dealako)

### Threads Resolved
10 of 10 unresolved threads addressed in this iteration (5 unique comments × 2 duplicate posts each).

apps/lfx-one/src/app/shared/services/project-context.service.ts
dealako: ✅ Resolved in `578e1848` — `Router` is now at line 7, adjacent to `Location` (line 4), `@angular/core` (line 5), and `@angular/core/rxjs-interop` (line 6). All `@angular/*` imports are grouped.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
dealako: ✅ Resolved in `578e1848` — `const context: ProjectContext = { ... }` with the `ProjectContext` import added. Structural drift is now caught at the literal.

apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts
dealako: ✅ Resolved in `578e1848` — `route.data['lens']` with no optional chain.

dealako:
## Re-review: `578e1848` ✅

Thanks for the quick follow-up, Rashad. Checked commit `578e1848` — all three actionable items from my earlier review are addressed:

| Thread | Fix |
| --- | --- |
| [minor] `Router` import out of group | Moved to line 7, adjacent to `Location` — all `@angular/*` imports now grouped. |
| [nit] Implicit `context` type | `const context: ProjectContext = { ... }` with the `ProjectContext` import added. |
| [nit] Unnecessary `?.` on `route.data` | Dropped — `route.data['lens']` in the final file. |

The two remaining open threads (no unit test, route `data`/`canActivate` repetition) are acknowledged follow-up items — nothing blocking.

**All blocking and minor items resolved. PR remains approved. 🚀**

apps/lfx-one/src/app/shared/services/project-context.service.ts
```diff
@@ -48,21 +52,44 @@ export class ProjectContextService {
...
+    if (this.router.getCurrentNavigation()) {
+      return;
+    }
+    const urlTree = this.router.parseUrl(this.router.url);
+    if (slug === null) {
+      delete urlTree.queryParams['project'];
+    } else {
+      urlTree.queryParams['project'] = slug;
+    }
+    this.location.replaceState(this.router.serializeUrl(urlTree));
```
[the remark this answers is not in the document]
MRashad26: Good analysis — this is a known limitation of the current approach. In the `openRow()` flow (`setProject()` → `router.navigate(['/project/overview'])`), `replaceState` does fire on the source URL since no navigation is in flight yet, but the subsequent `router.navigate([path])` doesn't carry query params and the destination lands without `?project=`. The microtask / opt-out approaches don't fully close the gap either — the destination navigation still needs `queryParams: { project: slug }` explicitly, which is a change to `multi-persona-dashboard.component.ts` out of scope here. The primary feature goal (cold-loading shared deep links and sidebar-switch URL sync) works correctly; this edge case is best fixed in a follow-up that threads `queryParams` through the calling components.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -24,37 +25,35 @@ export const routes: Routes = [
...
+        canActivate: [projectQueryParamGuard, executiveDirectorGuard],
         loadComponent: () => import('./modules/dashboards/health-metrics/health-metrics.component').then((m) => m.HealthMetricsComponent),
       },
       // Foundation Lens — Marketing Impact page (ED-only)
       {
         path: 'foundation/marketing-impact',
         data: { lens: 'foundation' },
-        canActivate: [executiveDirectorGuard],
+        canActivate: [projectQueryParamGuard, executiveDirectorGuard],
         loadComponent: () => import('./modules/dashboards/marketing-impact/marketing-impact.component').then((m) => m.MarketingImpactComponent),
```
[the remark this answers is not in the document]
MRashad26: Fixed in `01ad5e23` — swapped to `[executiveDirectorGuard, projectQueryParamGuard]` on both `foundation/health-metrics` and `foundation/marketing-impact`. Non-ED users are now redirected before any project lookup fires.

dealako (approved):

PR #709, opened 2026-05-14
feat(analytics): tag Plausible pageviews with foundation/project/lens

niravpatel27:
## Summary

- `PlausibleService` now reads `ProjectContextService` + `LensService` on every `NavigationEnd` and attaches `foundation` / `foundation_name` / `project` / `project_name` / `lens` custom-props to each pageview, so the Plausible dashboard becomes filterable/groupable by these dimensions.
- Public `/meetings/:numericId` routes load project context asynchronously — the auto NavigationEnd pageview is deferred for those paths and `MeetingJoinComponent` fires one enriched pageview once the meeting + project (and parent foundation when applicable) have resolved. A 2s timeout race guarantees the event still fires when the parent maps to `ROOT_PROJECT_SLUG` and is therefore returned as `null`.
- New shared interface `PlausiblePageviewContext` in `@lfx-one/shared/interfaces/plausible.interface.ts`.

No route, redirect, guard, or user-visible behavior changes. Total ≈ 142 lines across 3 files.

### Files changed

| File | Change |
|---|---|
| `packages/shared/src/interfaces/plausible.interface.ts` | Add `PlausiblePageviewContext` interface (5 optional string keys) |
| `apps/lfx-one/src/app/shared/services/plausible.service.ts` | Inject `ProjectContextService` + `LensService`; new `buildContextProps()`; merge into NavigationEnd + initial `script.onload` pageviews; skip auto-pageview for `/meetings/<digits>` paths |
| `apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts` | Inject `PlausibleService`; new `initializePublicMeetingPageviewTracking()` fires deferred enriched pageview after project resolves (browser-only, single-shot, with parent-fetch timeout race) |

## Why

The Plausible dashboard at `app.lfx.dev` shows entry pages like `/meetings/92079944361` with no signal of which foundation/project the visit belongs to. With this change the dashboard's existing Properties panel surfaces `foundation` / `project` / `lens` as filterable dimensions — operators can finally answer "how much traffic does the Kubernetes foundation see?" or slice top-pages by project. URLs are intentionally NOT normalized (e.g. `/meetings/:id`) — kept as-is so existing saved views and per-meeting drilldown still work.

5 custom-prop keys, well under Plausible's 30-keys-per-site limit. Historical pageviews predating the deploy will show up under `(none)` when grouping by these props.

## Test plan

- [ ] Local smoke: temporarily flip `environment.ts` → `plausible.enabled: true`, sign in, navigate `/foundation/meetings?project=<slug>` → confirm POST to `https://plausible.io/api/event` includes `props.foundation` / `props.foundation_name` / `props.project` / `props.project_name` / `props.lens` matching the active context
- [ ] Public meeting (incognito): open `/meetings/<real-numeric-id>` → confirm (a) NO Plausible POST on initial `NavigationEnd`, (b) exactly ONE POST after the `/public/api/meetings/:id` response settles, (c) that POST carries foundation/project props from the API response
- [ ] Anonymous on private meeting: confirm redacted `project: null` payload does NOT fire any Plausible POST
- [ ] Impersonation: trigger impersonation → confirm `setImpersonating(true)` short-circuit still suppresses the enriched pageview
- [ ] Post-deploy: Plausible dashboard → Properties panel surfaces `foundation`, `foundation_name`, `project`, `project_name`, `lens` as filterable dimensions

## Notes

- DCO sign-off attached. Local GPG/SSH signing config is incomplete on this machine (`gpg.ssh.allowedSignersFile` not set), so the commit may show as unverified on GitHub — happy to re-sign if needed.
- JIRA ticket not linked yet — please attach `LFXV2-<n>` after triage.

### DIFF
--- a/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
@@ -33,6 +33,7 @@ import {
   PastMeetingParticipant,
   PastMeetingRecording,
   PastMeetingSummary,
+  PlausiblePageviewContext,
   Project,
   PublicPastMeetingResponse,
   ROOT_PROJECT_SLUG,
@@ -44,6 +45,7 @@ import { LinkifyPipe } from '@pipes/linkify.pipe';
 import { MeetingTimePipe } from '@pipes/meeting-time.pipe';
 import { RecurrenceSummaryPipe } from '@pipes/recurrence-summary.pipe';
 import { MeetingService } from '@services/meeting.service';
+import { PlausibleService } from '@services/plausible.service';
 import { ProjectContextService } from '@services/project-context.service';
 import { ProjectService } from '@services/project.service';
 import { UserService } from '@services/user.service';
@@ -115,6 +117,7 @@ export class MeetingJoinComponent implements OnInit {
   private readonly userService = inject(UserService);
   private readonly clipboard = inject(Clipboard);
   private readonly projectContextService = inject(ProjectContextService);
+  private readonly plausibleService = inject(PlausibleService);
   private readonly dialogService = inject(DialogService);
   private readonly destroyRef = inject(DestroyRef);
   private readonly platformId = inject(PLATFORM_ID);
@@ -297,6 +300,7 @@ export class MeetingJoinComponent implements OnInit {
     this.registrants = this.initializeRegistrants();
     this.parentProject = this.initializeParentProject();
     this.initializeAutoJoin();
+    this.initializePublicMeetingPageviewTracking();
   }
 
   public ngOnInit(): void {
@@ -1126,6 +1130,70 @@ export class MeetingJoinComponent implements OnInit {
       ),
       { initialValue: null }
     );
+  }
+
+  /**
+   * Fire one enriched Plausible pageview for this public meeting page once
+   * the project (and parent foundation, when relevant) have resolved. The
+   * PlausibleService skips the auto NavigationEnd pageview for /meetings/:id
+   * paths so we land exactly one event per visit carrying foundation/project
+   * custom props. Anonymous viewers of private meetings receive a redacted
+   * `project: null` payload and intentionally fall through without firing.
+   */
+  private initializePublicMeetingPageviewTracking(): void {
+    if (isPlatformServer(this.platformId)) {
+      return;
+    }
+    toObservable(this.project)
+      .pipe(
+        filter((p): p is Partial<Project> => !!p?.slug),
+        switchMap((project) => {
+          // No parent_uid → project is itself a top-level foundation; fire immediately.
+          if (!project.parent_uid) {
+            return of({ project, parent: null as Project | null });
+          }
+          // Wait for the parent foundation lookup; race a 2s fallback so a slow
+          // upstream (or a parent that resolves to ROOT and is mapped to null)
+          // never blocks the pageview indefinitely.
+          return merge(
+            toObservable(this.parentProject).pipe(
+              filter((parent): parent is Project => !!parent),
+              map((parent) => ({ project, parent: parent as Project | null }))
+            ),
+            timer(2000).pipe(map(() => ({ project, parent: null as Project | null })))
+          );
+        }),
+        take(1),
+        takeUntilDestroyed(this.destroyRef)
+      )
+      .subscribe(({ project, parent }) => {
+        this.plausibleService.trackPage(this.buildMeetingPageviewProps(project, parent));
+      });
+  }
+
+  /**
+   * Build the Plausible pageview props payload for a public meeting page.
+   * When the meeting's project has a parent, that parent IS the foundation
+   * and the meeting's project is the sub-project. Otherwise the project is
+   * itself a foundation and only the foundation_* keys are populated.
+   */
+  private buildMeetingPageviewProps(project: Partial<Project>, parent: Project | null): Record<string, unknown> {
+    const context: PlausiblePageviewContext = {};
+    if (parent) {
+      if (parent.slug) context.foundation = parent.slug;
+      if (parent.name) context.foundation_name = parent.name;
+      if (project.slug) context.project = project.slug;
+      if (project.name) context.project_name = project.name;
+    } else {
+      if (project.slug) context.foundation = project.slug;
+      if (project.name) context.foundation_name = project.name;
+    }
+    return {
+      path: window.location.pathname,
+      url: `${window.location.origin}${window.location.pathname}`,
+      title: document.title,
+      ...context,
+    };
   }
 
   private initializeRegistrants(): Signal<MeetingRegistrant[]> {
--- a/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
@@ -1173,18 +1173,24 @@ export class MeetingJoinComponent implements OnInit {
 
   /**
    * Build the Plausible pageview props payload for a public meeting page.
-   * When the meeting's project has a parent, that parent IS the foundation
-   * and the meeting's project is the sub-project. Otherwise the project is
-   * itself a foundation and only the foundation_* keys are populated.
+   *
+   * Branches on `project.parent_uid` rather than on whether `parent` resolved,
+   * so a slow parent fetch (or an ROOT_PROJECT_SLUG mapping that resolves to
+   * null) does not roll a sub-project up into the `foundation_*` keys. When
+   * the parent is expected but unresolved we leave foundation empty — honest
+   * gap vs. misattribution of the dashboard dimensions this is enriching.
    */
   private buildMeetingPageviewProps(project: Partial<Project>, parent: Project | null): Record<string, unknown> {
     const context: PlausiblePageviewContext = {};
-    if (parent) {
-      if (parent.slug) context.foundation = parent.slug;
-      if (parent.name) context.foundation_name = parent.name;
+    if (project.parent_uid) {
+      // Sub-project under a foundation. Attribute the project correctly even
+      // if the parent fetch hasn't (or won't) resolve.
       if (project.slug) context.project = project.slug;
       if (project.name) context.project_name = project.name;
+      if (parent?.slug) context.foundation = parent.slug;
+      if (parent?.name) context.foundation_name = parent.name;
     } else {
+      // Top-level project IS the foundation — only foundation_* keys populated.
       if (project.slug) context.foundation = project.slug;
       if (project.name) context.foundation_name = project.name;
     }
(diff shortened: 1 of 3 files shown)


apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -1128,6 +1132,70 @@ export class MeetingJoinComponent implements OnInit {
     );
   }
 
+  /**
+   * Fire one enriched Plausible pageview for this public meeting page once
+   * the project (and parent foundation, when relevant) have resolved. The
+   * PlausibleService skips the auto NavigationEnd pageview for /meetings/:id
+   * paths so we land exactly one event per visit carrying foundation/project
```
MRashad26: **Multi-paragraph docstrings / multi-line comment blocks across the new code.**

**Problem:** Four new multi-line JSDoc blocks plus several multi-line inline comments landed in this PR:
- `meeting-join.component.ts:1132-1138` — 7-line JSDoc on `initializePublicMeetingPageviewTracking`
- `meeting-join.component.ts:1152-1154` — 3-line inline comment ("race a 2s fallback")
- `meeting-join.component.ts:1171-1176` — 6-line JSDoc on `buildMeetingPageviewProps`
- `plausible.service.ts:179-185` — 7-line JSDoc on `buildContextProps`
- `plausible.service.ts:188-189` — 2-line inline comment
- `plausible.interface.ts:13-22` — 6-line JSDoc on the new interface

**Why:** Global CLAUDE.md is explicit: *"Default to writing no comments. Only add one when the WHY is non-obvious… Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The bulk of these comments describe **what** the code does (the function signatures, method bodies, and interface field names already convey that) rather than non-obvious **why**. They will rot the moment the code is touched.

**Suggested fix:** Strip the multi-line JSDocs. Keep at most a single one-line comment where the **why** is genuinely non-obvious — concretely:
- `initializePublicMeetingPageviewTracking`: `// Public meeting pages must defer the pageview — context resolves async.` (one line) is enough; the rest of the JSDoc paraphrases the function body.
- `buildMeetingPageviewProps`: function name + types are self-documenting; drop the JSDoc entirely.
- `buildContextProps`: drop the JSDoc; the inline `// Read foundation and project independently…` is also restating what the next 14 lines visibly do — remove.
- `PlausiblePageviewContext`: drop the JSDoc; field names are the documentation.

The "parent IS the foundation" nuance in `buildMeetingPageviewProps` IS non-obvious and worth keeping — collapse to one line: `// When parent_uid is set, the parent IS the foundation and project is the sub-project.`

apps/lfx-one/src/app/shared/services/plausible.service.ts
```diff
@@ -149,14 +166,51 @@ export class PlausibleService {
...
+          ...this.buildContextProps(),
         });
       });
   }
 
+  /**
+   * Build the foundation / project / lens custom-props for the current
+   * pageview from the active app state. Keys with empty values are omitted
+   * so Plausible doesn't store empty strings as a distinct dimension value.
+   * Returned as a Record so callers can spread into the broader pageview
```
MRashad26: **Duplicated context-building between `buildContextProps` (service) and `buildMeetingPageviewProps` (component).**

**Problem:** `buildContextProps()` here at L186-209 and `buildMeetingPageviewProps()` in `meeting-join.component.ts:1177-1194` both materialize a `PlausiblePageviewContext` from `(foundation, project)` pairs. The foundation-fallback branch of `buildMeetingPageviewProps` (L1185-1186, when `parent` is null and the meeting's project is treated as a top-level foundation) is verbatim duplication of this method's foundation block.

**Why:** The dashboard contract — *which keys exist, when they're populated, when empty values are omitted* — is the kind of thing that must live in one place. Two implementations means the next time someone adds a custom-prop (e.g., a `region` dimension) they will update one and forget the other, and the Plausible dashboard will silently show partial data for public meeting pages. Memory note `feedback_shared_interfaces` and the spirit of CLAUDE.md's signals-first / private-init / DRY conventions all point to a single owner.

**Suggested fix:** Expose a single builder on `PlausibleService` that accepts an explicit `{foundation, project, lens}` triple (rather than reading services internally), and have both callers use it:
```ts
public buildPageviewContext(input: {
  foundation: Pick<Project, 'slug' | 'name'> | null;
  project: Pick<Project, 'slug' | 'name'> | null;
  lens: string | null;
}): PlausiblePageviewContext { /* current foundation/project/lens key logic */ }
```
Then `buildContextProps()` becomes a thin wrapper reading `projectContextService` + `lensService`, and `buildMeetingPageviewProps` becomes `this.plausibleService.buildPageviewContext({ foundation: parent ?? project, project: parent ? project : null, lens: null })`.

MRashad26 (commented):
Strict CLAUDE.md / code-enforcer-agent review — 2 major + 3 minor findings. Skipping items already raised by Copilot (L1164: 2s timer races indistinguishably with legit `null` parent resolution) and CodeRabbit (L1171: auto-suppression `^/meetings/\d+$` vs manual fire on all meeting paths could double-count).

apps/lfx-one/src/app/shared/services/plausible.service.ts
```diff
@@ -149,14 +166,51 @@ export class PlausibleService {
...
+    }
+    const project = this.projectContextService.selectedProject();
+    if (project?.slug) {
+      context.project = project.slug;
+    }
+    if (project?.name) {
+      context.project_name = project.name;
+    }
+    const lens = this.lensService.activeLens();
+    if (lens) {
```
MRashad26: **Redundant spread on a freshly-built object.**

**Problem:** `return { ...context };` — `context` was just constructed in this function (L187) and no external reference to it exists, so the spread is allocating a second identical object for no reason.

**Why:** Pointless work and a misleading signal to the reader ("why is this defensively cloning?"). Also: the declared return type is `Record<string, unknown>` but the actual value is `PlausiblePageviewContext` — the spread is the only thing widening the type, and that widening should be explicit if intended, not a side-effect.

**Suggested fix:** Either `return context;` and change the return type to `PlausiblePageviewContext`, or — if the `Record<string, unknown>` is intentional so callers can spread it into a larger payload — keep the type but drop the spread: `return context;` still satisfies `Record<string, unknown>` structurally.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -1128,6 +1132,70 @@ export class MeetingJoinComponent implements OnInit {
...
+          // No parent_uid → project is itself a top-level foundation; fire immediately.
+          if (!project.parent_uid) {
+            return of({ project, parent: null as Project | null });
+          }
+          // Wait for the parent foundation lookup; race a 2s fallback so a slow
+          // upstream (or a parent that resolves to ROOT and is mapped to null)
+          // never blocks the pageview indefinitely.
+          return merge(
+            toObservable(this.parentProject).pipe(
+              filter((parent): parent is Project => !!parent),
```
MRashad26, line 1152: **Magic number `2000` for the parent-fetch timeout.**

**Problem:** `timer(2000)` is a load-bearing constant — it is the entire mechanism that prevents the pageview from being lost when `parentProject` legitimately resolves to `null` (the `ROOT_PROJECT_SLUG` mapping at L1129 in `initializeParentProject`). Buried as an inline literal, future maintainers have no signal that this number is anything but arbitrary.

**Why:** A bare `2000` invites two failure modes: (a) someone shortens it not realizing the upstream `/public/api/projects/<uid>` can take longer than that under load, silently misattributing events to `(none)`; (b) someone deletes it as "dead" without realizing the pipeline relies on it. The right fix Copilot suggested at L1164 (emit a discriminator from `initializeParentProject` instead of a `null`) eventually removes the need for the timer entirely — until that lands, the constant deserves a name.

**Suggested fix:** Pull into a named local at the top of the method:
```ts
const PARENT_FETCH_TIMEOUT_MS = 2000;
// ...
timer(PARENT_FETCH_TIMEOUT_MS).pipe(map(() => ({ project, parent: null as Project | null })))
```
If the discriminator refactor lands, the constant disappears with the timer.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -1128,6 +1132,70 @@ export class MeetingJoinComponent implements OnInit {
...
+  private buildMeetingPageviewProps(project: Partial<Project>, parent: Project | null): Record<string, unknown> {
+    const context: PlausiblePageviewContext = {};
+    if (parent) {
+      if (parent.slug) context.foundation = parent.slug;
+      if (parent.name) context.foundation_name = parent.name;
+      if (project.slug) context.project = project.slug;
+      if (project.name) context.project_name = project.name;
+    } else {
+      if (project.slug) context.foundation = project.slug;
+      if (project.name) context.foundation_name = project.name;
```
MRashad26: **Component re-implements the path/url/title payload that `PlausibleService` already builds.**

**Problem:** L1188-1193 hand-rolls a `{ path, url, title, ...context }` object using raw `window.location.pathname`, `window.location.origin`, and `document.title`. `PlausibleService` already has dedicated private helpers for this — `getSanitizedPath` (L218) and `getSanitizedUrl` (L210) — which strip query params and hash fragments to avoid leaking auth tokens, OTPs, or invite tokens in pageview payloads.

**Why:** Two failure modes:
1. **Privacy/security drift.** If a public meeting URL ever carries a query param (e.g., a future `/meetings/<id>?invite=<token>` for one-click invites), `window.location.pathname` will not leak it today — `pathname` excludes query — but the moment someone copies this payload shape and reaches for `window.location.href` or `event.urlAfterRedirects`, sanitization will silently regress. The service-side helpers are the canonical guard; bypassing them puts the burden of remembering on every caller.
2. **Two pageview payload shapes to maintain.** The service's `setupRouteTracking` (L165-176) emits `{ path, url, title, ...context }` from sanitized inputs; this method emits the same keys from raw `window.*`. When the schema evolves, both must stay in lockstep — exactly the DRY trap finding #2 calls out.

**Suggested fix:** Move payload construction into the service. Add a public overload that accepts an explicit context and resolves path/url/title itself:
```ts
// PlausibleService
public trackPageviewWithContext(context: PlausiblePageviewContext): void {
  if (typeof window === 'undefined' || typeof document === 'undefined') return;
  this.trackPage({
    path: this.getSanitizedPath(window.location.pathname),
    url: this.getSanitizedUrl(),
    title: document.title,
    ...context,
  });
}
```
The component then collapses to `this.plausibleService.trackPageviewWithContext(this.buildMeetingPageviewContext(project, parent));` and `buildMeetingPageviewProps` returns a `PlausiblePageviewContext` instead of a fully-shaped pageview payload.

MRashad26 (approved):

PR #710, opened 2026-05-15
feat(voting): add ballot-casting drawer for plurality polls in my votes

manishdixitlfx:
## Summary

Adds in-app ballot casting for **plurality (`poll_type === GENERIC`)** polls on the **My Votes (Me lens)** flow. Ranked-choice methods (`condorcet_irv`, `instant_runoff_vote`, `meek_stv`) currently render a "Coming soon" placeholder — ranked-choice cast UI ships in [LFXV2-1839](https://linuxfoundation.atlassian.net/browse/LFXV2-1839) as PR 2.

JIRA: [LFXV2-1840](https://linuxfoundation.atlassian.net/browse/LFXV2-1840)

## What's new

- **`vote-cast-drawer`** — new component renders:
  - single-choice questions as radio buttons (via `lfx-radio-button` wrapper)
  - multi-choice questions as PrimeNG checkboxes with array-binding (`lfx-checkbox` wrapper only supports binary mode, so the raw `p-checkbox` is used here per the same pattern as other complex forms)
  - an Abstain toggle when `vote.allow_abstain === true`
  - a "Coming soon" placeholder for non-`GENERIC` poll types — **no PCC link inside the Me lens cast flow**
- **`getMyVotes` decorates each Vote with `response_status`** (`AWAITING_RESPONSE` / `RESPONDED`) — server-side enrichment that the Me lens uses to decide cast-vs-view
- **New endpoint** `GET /api/votes/:uid/my-response` — returns the user's pre-allocated `vote_response` row. Required because the voting service expects the existing invitation row's `vote_response_uid` on submit (a fresh client UUID returns 404 from upstream)
- **Cast-or-view branching in `votes-dashboard`** — Me-lens, active, awaiting-response rows open the cast drawer; everything else continues to open the existing read-only results drawer (current behaviour preserved)
- **`votes-table` action label** flips from "Review" to "Vote" for active-awaiting rows in Me lens (`response_status` is only set in Me lens, so the flag does double-duty as a lens marker)

## Reviewer callouts

1. **New shared type field**: `Vote.response_status?: VoteResponseStatus` in `packages/shared/src/interfaces/poll.interface.ts`. Optional on purpose — only `getMyVotes` (Me lens) decorates it; raw upstream `Vote` responses don't carry it.
2. **New shared type**: `MyVoteResponse` mirrors `MySurveyResponse`; backing the new `/my-response` endpoint.
3. **No PCC redirect in Me lens** — by design. Project/Foundation/Organization lenses still open the existing `vote-results-drawer` which retains the PCC link for ops/admin paths.
4. **Submit chain uses `switchMap`** (`vote-cast-drawer.component.ts:onSubmit`) — getMyVoteResponse → createVoteResponse with one error funnel and a sentinel error message (`INVITATION_NOT_FOUND`) to differentiate the two failure modes.

## Known limitation (separate ticket)

`lfx-v2-voting-service`'s `POST /vote_responses` still **proxies to v1 Salesforce ITX**. Any vote that originated in v2 (`POST /votes` without a Salesforce sync) returns 404 on submit because v1 has no record of it. Real prod votes (synced from v1) work end-to-end. Pure v2-only test votes do not. Tracked in [LFXV2-1842](https://linuxfoundation.atlassian.net/browse/LFXV2-1842).

## Follow-up PRs (tracked separately)

- **[LFXV2-1839](https://linuxfoundation.atlassian.net/browse/LFXV2-1839)** — Ranked-choice cast UI (Condorcet IRV / IRV / Meek STV) + ranked-choice results rendering. Blocked by this PR landing.
- **[LFXV2-1841](https://linuxfoundation.atlassian.net/browse/LFXV2-1841)** — My Vote Response drawer (surveys-parity read-only view of submitted ballot). Backend (`getMyVoteResponse`) already ships in this PR.
- **[LFXV2-1842](https://linuxfoundation.atlassian.net/browse/LFXV2-1842)** — Backend: v2-native ballot submission in `lfx-v2-voting-service`.

## Test plan

- [ ] Me lens, GENERIC + single_choice + `allow_abstain=true` → drawer renders radios (Approve/Reject) + Abstain toggle. Submit succeeds for v1-synced votes; fails gracefully with toast for v2-only votes.
- [ ] Me lens, GENERIC + single_choice + `allow_abstain=false` → radios only, no abstain section.
- [ ] Me lens, GENERIC + multiple_choice + `allow_abstain=true` → "Select all that apply" hint + checkboxes + Abstain toggle.
- [ ] Me lens, GENERIC + multiple_choice + `allow_abstain=false` → checkboxes only, no abstain.
- [ ] Me lens, GENERIC + multi-question (single + multi) → both question types render correctly in one ballot.
- [ ] Me lens, CONDORCET_IRV (or any non-GENERIC) → "Coming soon" placeholder, no PCC link visible.
- [ ] After a successful submit, the row's `response_status` flips to `RESPONDED` on refresh; the action button flips from "Vote" to "Review" and clicking opens the aggregate results drawer.
- [ ] Project lens (non-Me) on the same votes → action button is "Review" (not "Vote") and clicking opens the existing read-only results drawer. No regressions.
- [ ] Abstain toggle disables question controls and re-enables them on toggle-off.
- [ ] Submit button is disabled while no choice is selected (for non-abstain submits) and during the in-flight submit.
- [ ] Error toasts surface for INVITATION_NOT_FOUND (404 lookup) and upstream submit failures.

[LFXV2-1839]: https://linuxfoundation.atlassian.net/browse/LFXV2-1839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-1840]: https://linuxfoundation.atlassian.net/browse/LFXV2-1840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-1842]: https://linuxfoundation.atlassian.net/browse/LFXV2-1842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
@@ -1,0 +1,146 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<p-drawer
+  [(visible)]="visible"
+  position="right"
+  [modal]="true"
+  [showCloseIcon]="false"
+  styleClass="xl:w-[45%] lg:w-[55%] md:w-[70%] sm:w-[90%] w-full vote-cast-drawer"
+  data-testid="vote-cast-drawer">
+  @if (vote(); as voteData) {
+    <ng-template #header>
+      <div class="flex flex-col gap-3 w-full pr-4" data-testid="vote-cast-drawer-header">
+        <div class="flex items-center justify-between gap-2">
+          <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 truncate" data-testid="vote-cast-drawer-title">
+            {{ voteData.name }}
+          </h1>
+          <button
+            type="button"
+            (click)="onClose()"
+            class="p-2 text-slate-400 hover:text-slate-600 hover:bg-slate-100 rounded-md transition-colors flex-shrink-0"
+            data-testid="vote-cast-drawer-close"
+            aria-label="Close panel">
+            <i class="fa-light fa-xmark text-xl"></i>
+          </button>
+        </div>
+
+        <div class="flex flex-wrap items-center gap-2 text-sm text-slate-600">
+          @if (voteData.committee_name) {
+            <span class="bg-blue-100 text-blue-700 text-xs px-2.5 py-1 rounded font-medium">
+              {{ voteData.committee_name }}
+            </span>
+            <span class="text-slate-400">•</span>
+          }
+          <lfx-tag
+            [value]="voteData.status | pollStatusLabel"
+            [severity]="voteData.status | pollStatusSeverity"
+            data-testid="vote-cast-drawer-status"></lfx-tag>
+          <span class="text-slate-400">•</span>
+          <span>Due {{ voteData.end_time | date: 'MMM d, y' }}</span>
+        </div>
+      </div>
+    </ng-template>
+
+    <div class="flex flex-col gap-6" data-testid="vote-cast-drawer-content">
+      @if (loadingVote()) {
+        <div class="flex flex-col gap-4">
+          <p-skeleton width="60%" height="24px"></p-skeleton>
+          @for (i of [1, 2, 3]; track i) {
+            <div class="rounded-lg border border-slate-200 p-4 flex flex-col gap-3">
+              <p-skeleton width="80%" height="16px"></p-skeleton>
+              <p-skeleton width="40%" height="12px"></p-skeleton>
+              <p-skeleton width="40%" height="12px"></p-skeleton>
+            </div>
+          }
+        </div>
+      } @else if (!isGenericPoll()) {
+        <div class="flex flex-col items-center justify-center py-12 px-6 text-center" data-testid="vote-cast-drawer-unsupported">
+          <div class="w-16 h-16 rounded-full bg-blue-100 flex items-center justify-center mb-6">
+            <i class="fa-light fa-hourglass-half text-2xl text-blue-600"></i>
+          </div>
+          <h2 class="text-xl font-semibold text-slate-900 mb-3">Coming soon</h2>
+          <p class="text-sm text-slate-600 max-w-md">Ballot casting for ranked-choice voting methods is on the way and will be available here shortly.</p>
+        </div>
+      } @else if (questions().length === 0) {
+        <div class="text-center py-8 border border-slate-200 rounded-lg" data-testid="vote-cast-drawer-no-questions">
+          <p class="text-sm text-gray-500">This vote has no questions to answer.</p>
+        </div>
+      } @else {
+        @if (voteData.description) {
+          <div class="text-sm text-slate-700 whitespace-pre-line" data-testid="vote-cast-drawer-description">
+            {{ voteData.description }}
+          </div>
+        }
+
+        <form [formGroup]="form" class="flex flex-col gap-6">
+          @for (question of questions(); track question.question_id) {
+            <div class="rounded-lg border border-slate-200 p-4 flex flex-col gap-3" [attr.data-testid]="'vote-cast-question-' + question.question_id">
+              <div class="flex flex-col gap-1">
+                <h3 class="text-sm font-semibold text-slate-900">{{ question.prompt }}</h3>
+                @if (isMultipleChoice(question)) {
+                  <p class="text-xs text-slate-500">Select all that apply.</p>
+                }
+              </div>
+
+              <div class="flex flex-col gap-2">
+                @if (isSingleChoice(question)) {
+                  @for (choice of question.choices; track choice.choice_id) {
+                    <lfx-radio-button
+                      [form]="form"
+                      [control]="question.question_id"
+                      [name]="question.question_id"
+                      [value]="choice.choice_id"
+                      [inputId]="'vote-cast-choice-' + choice.choice_id"
+                      [label]="choice.choice_text"
+                      [attr.data-testid]="'vote-cast-choice-' + choice.choice_id"></lfx-radio-button>
+                  }
+                } @else if (isMultipleChoice(question)) {
+                  @for (choice of question.choices; track choice.choice_id) {
+                    <div class="flex items-center gap-2" [attr.data-testid]="'vote-cast-choice-' + choice.choice_id">
+                      <p-checkbox
+                        [formControlName]="question.question_id"
+                        [value]="choice.choice_id"
+                        [binary]="false"
+                        [inputId]="'vote-cast-choice-' + choice.choice_id" />
+                      <label [for]="'vote-cast-choice-' + choice.choice_id" class="text-sm font-medium text-gray-700 cursor-pointer">
+                        {{ choice.choice_text }}
+                      </label>
+                    </div>
+                  }
+                }
+              </div>
+            </div>
+          }
+        </form>
+
+        @if (allowAbstain()) {
+          <div class="rounded-lg border border-slate-200 bg-slate-50 p-4 flex flex-col gap-2">
+            <div class="flex items-center gap-2" data-testid="vote-cast-abstain-toggle">
+              <p-checkbox
+                [binary]="true"
+                [ngModel]="abstain()"
+                [ngModelOptions]="{ standalone: true }"
+                (ngModelChange)="onAbstainChange($event)"
+                inputId="vote-cast-abstain" />
+              <label for="vote-cast-abstain" class="text-sm font-medium text-gray-700 cursor-pointer">Abstain from this vote</label>
+            </div>
+            <p class="text-xs text-slate-500">An abstention is recorded as a non-vote and overrides any selections above.</p>
+          </div>
+        }
+
+        <div class="flex items-center justify-end gap-3 pt-2 border-t border-slate-200">
+          <lfx-button label="Cancel" severity="secondary" (onClick)="onClose()" [disabled]="submitting()" data-testid="vote-cast-cancel-button"></lfx-button>
+          <lfx-button
+            label="Submit Vote"
+            icon="fa-light fa-check"
+            severity="primary"
+            [disabled]="submitDisabled()"
+            [loading]="submitting()"
+            (onClick)="onSubmit()"
+            data-testid="vote-cast-submit-button"></lfx-button>
+        </div>
+      }
+    </div>
+  }
+</p-drawer>
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
@@ -78,13 +78,13 @@
             <div class="rounded-lg border border-slate-200 p-4 flex flex-col gap-3" [attr.data-testid]="'vote-cast-question-' + question.question_id">
               <div class="flex flex-col gap-1">
                 <h3 class="text-sm font-semibold text-slate-900">{{ question.prompt }}</h3>
-                @if (isMultipleChoice(question)) {
+                @if (question.type === 'multiple_choice') {
                   <p class="text-xs text-slate-500">Select all that apply.</p>
                 }
               </div>
 
               <div class="flex flex-col gap-2">
-                @if (isSingleChoice(question)) {
+                @if (question.type === 'single_choice') {
                   @for (choice of question.choices; track choice.choice_id) {
                     <lfx-radio-button
                       [form]="form"
@@ -95,7 +95,7 @@
                       [label]="choice.choice_text"
                       [attr.data-testid]="'vote-cast-choice-' + choice.choice_id"></lfx-radio-button>
                   }
-                } @else if (isMultipleChoice(question)) {
+                } @else if (question.type === 'multiple_choice') {
                   @for (choice of question.choices; track choice.choice_id) {
                     <div class="flex items-center gap-2" [attr.data-testid]="'vote-cast-choice-' + choice.choice_id">
                       <p-checkbox
@@ -117,12 +117,7 @@
         @if (allowAbstain()) {
           <div class="rounded-lg border border-slate-200 bg-slate-50 p-4 flex flex-col gap-2">
             <div class="flex items-center gap-2" data-testid="vote-cast-abstain-toggle">
-              <p-checkbox
-                [binary]="true"
-                [ngModel]="abstain()"
-                [ngModelOptions]="{ standalone: true }"
-                (ngModelChange)="onAbstainChange($event)"
-                inputId="vote-cast-abstain" />
+              <p-checkbox [binary]="true" [formControl]="abstainControl" inputId="vote-cast-abstain" />
               <label for="vote-cast-abstain" class="text-sm font-medium text-gray-700 cursor-pointer">Abstain from this vote</label>
             </div>
             <p class="text-xs text-slate-500">An abstention is recorded as a non-vote and overrides any selections above.</p>
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -3,9 +3,9 @@
 
 import { DatePipe } from '@angular/common';
 import { HttpErrorResponse } from '@angular/common/http';
-import { Component, computed, effect, inject, input, model, output, signal, Signal } from '@angular/core';
-import { toObservable, toSignal } from '@angular/core/rxjs-interop';
-import { FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators } from '@angular/forms';
+import { Component, computed, DestroyRef, inject, input, model, output, signal, Signal } from '@angular/core';
+import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
 import { ButtonComponent } from '@components/button/button.component';
 import { RadioButtonComponent } from '@components/radio-button/radio-button.component';
 import { TagComponent } from '@components/tag/tag.component';
@@ -18,7 +18,9 @@ import { MessageService } from 'primeng/api';
 import { CheckboxModule } from 'primeng/checkbox';
 import { DrawerModule } from 'primeng/drawer';
 import { SkeletonModule } from 'primeng/skeleton';
-import { catchError, finalize, of, shareReplay, startWith, Subject, switchMap, take, takeUntil, throwError } from 'rxjs';
+import { catchError, filter, finalize, of, shareReplay, startWith, Subject, switchMap, take, takeUntil, throwError } from 'rxjs';
+
+const INVITATION_NOT_FOUND = 'INVITATION_NOT_FOUND';
 
 @Component({
   selector: 'lfx-vote-cast-drawer',
@@ -26,7 +28,6 @@ import { catchError, finalize, of, shareReplay, startWith, Subject, switchMap, t
     DrawerModule,
     SkeletonModule,
     CheckboxModule,
-    FormsModule,
     ReactiveFormsModule,
     ButtonComponent,
     RadioButtonComponent,
@@ -42,6 +43,7 @@ export class VoteCastDrawerComponent {
   // === Services ===
   private readonly voteService = inject(VoteService);
   private readonly messageService = inject(MessageService);
+  private readonly destroyRef = inject(DestroyRef);
 
   // === Inputs ===
   public readonly voteId = input<string | null>(null);
@@ -54,90 +56,60 @@ export class VoteCastDrawerComponent {
   public readonly voteSubmitted = output<string>();
 
   // === Forms ===
-  // Per-question form. Keys are question_id. Each value is:
-  //   - string | null for single_choice (selected choice_id)
-  //   - string[] | null for multiple_choice (array of selected choice_ids)
-  // Submit normalises both shapes into VoteAnswerInput.choice_ids.
   public readonly form = new FormGroup({});
+  public readonly abstainControl = new FormControl<boolean>(false, { nonNullable: true });
 
   // === Writable Signals ===
   protected readonly loadingVote = signal<boolean>(false);
   protected readonly submitting = signal<boolean>(false);
-  protected readonly abstain = signal<boolean>(false);
+  // Reactive dependency for submitDisabled — rebuildForm uses { emitEvent: false }, so statusChanges is silent.
+  private readonly formVersion = signal<number>(0);
 
   // === Shared Observables ===
   private readonly voteId$ = toObservable(this.voteId).pipe(shareReplay({ bufferSize: 1, refCount: true }));
-  // Emits when the drawer is closed mid-submit. The submit chain pipes through
-  // takeUntil(cancelSubmit$) so the HTTP stream is cancelled cleanly and the
-  // success/error toast cannot fire after the user has already closed the drawer.
+  // Emits on drawer dismissal (close button OR ESC OR overlay click — all flip visible to false).
   private readonly cancelSubmit$ = new Subject<void>();
 
   // === Derived Signals (from API) ===
   protected readonly vote: Signal<Vote | null> = this.initVote();
-  // form.statusChanges is the canonical reactive surface for FormGroup validity.
-  // We pipe to startWith(form.status) so the signal has a synchronous initial value
-  // (status emissions only fire on changes, not on initial construction).
-  private readonly formStatus: Signal<string> = toSignal(this.form.statusChanges.pipe(startWith(this.form.status)), { initialValue: this.form.status });
+  protected readonly abstain: Signal<boolean> = toSignal(this.abstainControl.valueChanges, { initialValue: this.abstainControl.value });
 
   // === Computed Signals ===
   protected readonly isGenericPoll: Signal<boolean> = computed(() => this.vote()?.poll_type === PollType.GENERIC);
   protected readonly questions: Signal<PollQuestion[]> = computed(() => this.vote()?.poll_questions ?? []);
   protected readonly allowAbstain: Signal<boolean> = computed(() => !!this.vote()?.allow_abstain);
-  protected readonly submitDisabled: Signal<boolean> = computed(() => {
-    if (this.submitting()) return true;
-    if (this.abstain()) return false;
-    return this.formStatus() !== 'VALID';
-  });
+  protected readonly submitDisabled: Signal<boolean> = this.initSubmitDisabled();
 
   public constructor() {
-    // Rebuild the form whenever a new vote is loaded so controls match its questions.
-    effect(() => {
-      const questions = this.questions();
-      this.rebuildForm(questions);
+    // Rebuild the form when the loaded vote's questions change.
+    toObservable(this.questions)
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe((questions) => this.rebuildForm(questions));
+
+    // Abstain toggle disables answer controls without losing their values.
+    this.abstainControl.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((isAbstaining) => {
+      if (isAbstaining) this.form.disable({ emitEvent: false });
+      else this.form.enable({ emitEvent: false });
+      this.bumpFormVersion();
     });
 
-    // Reset transient state when the drawer closes so reopening starts fresh.
-    // Cancels any in-flight submit regardless of how the drawer was dismissed
-    // (close button, ESC, or overlay-click) so the success toast cannot fire
-    // after the user has already navigated away.
-    effect(() => {
-      if (!this.visible()) {
-        if (this.submitting()) {
-          this.cancelSubmit$.next();
-        }
-        this.abstain.set(false);
+    // Drawer dismissal (close button / ESC / overlay click) — cancel any in-flight submit and reset transient state.
+    toObservable(this.visible)
+      .pipe(
+        filter((v) => !v),
+        takeUntilDestroyed(this.destroyRef)
+      )
+      .subscribe(() => {
+        if (this.submitting()) this.cancelSubmit$.next();
         this.submitting.set(false);
-      }
-    });
-
-    // Toggling abstain disables answer fields without losing their values.
-    effect(() => {
-      if (this.abstain()) {
-        this.form.disable({ emitEvent: false });
-      } else {
-        this.form.enable({ emitEvent: false });
-      }
-    });
+        // Reset abstain so reopening starts fresh — letting valueChanges fire keeps form.enable + abstain signal in sync.
+        if (this.abstainControl.value) this.abstainControl.setValue(false);
+      });
   }
 
   // === Protected Methods ===
   protected onClose(): void {
-    // Setting visible to false triggers the visibility effect, which handles
-    // in-flight submit cancellation and all other state resets for every
-    // close path (close button, ESC, overlay-click).
     this.visible.set(false);
-  }
-
-  protected onAbstainChange(checked: boolean): void {
-    this.abstain.set(checked);
-  }
-
-  protected isSingleChoice(question: PollQuestion): boolean {
-    return question.type === 'single_choice';
-  }
-
-  protected isMultipleChoice(question: PollQuestion): boolean {
-    return question.type === 'multiple_choice';
   }
 
   protected onSubmit(): void {
@@ -149,18 +121,13 @@ export class VoteCastDrawerComponent {
 
     this.submitting.set(true);
 
-    // The voting service pre-allocates one vote_response row per invitee when a vote
-    // is enabled; the cast endpoint requires that pre-allocated UID — a fresh client
-    // UUID returns 404. Chain the lookup into the submit via switchMap so the two
-    // sequential calls share one subscription and one error funnel.
+    // The voting service pre-allocates one vote_response row per invitee — POST /vote_responses MUST reuse that uid.
     this.voteService
       .getMyVoteResponse(vote.uid)
       .pipe(
         take(1),
         switchMap((myResponse) => {
-          if (!myResponse?.uid) {
-            return throwError(() => new Error('INVITATION_NOT_FOUND'));
-          }
+          if (!myResponse?.uid) return throwError(() => new Error(INVITATION_NOT_FOUND));
           const payload: CreateVoteResponseRequest = {
             vote_response_uid: myResponse.uid,
             vote_uid: vote.uid,
@@ -169,8 +136,6 @@ export class VoteCastDrawerComponent {
           };
           return this.voteService.createVoteResponse(payload);
         }),
-        // If the user closes the drawer mid-submit, drop the stream silently —
-        // no success/error toast fires, submitting is already reset by onClose.
         takeUntil(this.cancelSubmit$),
         finalize(() => this.submitting.set(false))
       )
@@ -185,8 +150,8 @@ export class VoteCastDrawerComponent {
           this.voteSubmitted.emit(vote.uid);
           this.visible.set(false);
         },
-        error: (err: Error | HttpErrorResponse) => {
-          const isInvitationMissing = err?.message === 'INVITATION_NOT_FOUND';
+        error: (err: unknown) => {
+          const isInvitationMissing = err instanceof Error && !(err instanceof HttpErrorResponse) && err.message === INVITATION_NOT_FOUND;
           this.messageService.add({
             severity: 'error',
             summary: isInvitationMissing ? 'Unable to find your invitation' : 'Unable to submit vote',
@@ -208,10 +173,8 @@ export class VoteCastDrawerComponent {
             this.loadingVote.set(false);
             return of(null);
           }
-
           this.loadingVote.set(true);
           const listVote = this.listVote();
-
           return this.voteService.getVote(id).pipe(
             catchError(() => of(listVote)),
             finalize(() => this.loadingVote.set(false)),
@@ -223,28 +186,37 @@ export class VoteCastDrawerComponent {
     );
   }
 
+  private initSubmitDisabled(): Signal<boolean> {
+    return computed(() => {
+      this.formVersion(); // re-evaluate when controls are added/removed/disabled via { emitEvent: false }
+      if (this.submitting()) return true;
+      if (this.abstain()) return false;
+      return !this.form.valid;
+    });
+  }
+
   // === Private Helpers ===
-  // Reconciles the form against the current vote's questions: drops controls for
-  // removed questions, adds controls for new ones, leaves stable controls alone.
+  // Reconciles the form against the current vote's questions; bumps formVersion since we suppress emitEvent.
   private rebuildForm(questions: PollQuestion[]): void {
     const desiredIds = new Set(questions.map((q) => q.question_id));
     for (const existingId of Object.keys(this.form.controls)) {
-      if (!desiredIds.has(existingId)) {
-        this.form.removeControl(existingId, { emitEvent: false });
-      }
+      if (!desiredIds.has(existingId)) this.form.removeControl(existingId, { emitEvent: false });
     }
     for (const question of questions) {
       if (this.form.contains(question.question_id)) continue;
       if (question.type === 'multiple_choice') {
-        this.form.addControl(question.question_id, new FormControl<string[]>([], Validators.required), { emitEvent: false });
+        this.form.addControl(question.question_id, new FormControl<string[]>([], { nonNullable: true, validators: [Validators.required] }), {
+          emitEvent: false,
+        });
       } else {
         this.form.addControl(question.question_id, new FormControl<string | null>(null, Validators.required), { emitEvent: false });
       }
     }
-    // All add/removeControl calls above suppress individual emissions to avoid
-    // per-control churn. Fire a single status update once all controls are
-    // reconciled so toSignal(statusChanges) reflects the correct validity state.
-    this.form.updateValueAndValidity({ emitEvent: true });
+    this.bumpFormVersion();
+  }
+
+  private bumpFormVersion(): void {
+    this.formVersion.update((v) => v + 1);
   }
 
   private buildAnswers(questions: PollQuestion[]): VoteAnswerInput[] {
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -137,6 +137,7 @@ export class VoteCastDrawerComponent {
           return this.voteService.createVoteResponse(payload);
         }),
         takeUntil(this.cancelSubmit$),
+        takeUntilDestroyed(this.destroyRef),
         finalize(() => this.submitting.set(false))
       )
       .subscribe({
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -17,7 +17,7 @@ import { MessageService } from 'primeng/api';
 import { CheckboxModule } from 'primeng/checkbox';
 import { DrawerModule } from 'primeng/drawer';
 import { SkeletonModule } from 'primeng/skeleton';
-import { catchError, finalize, of, shareReplay, startWith, switchMap, take, throwError } from 'rxjs';
+import { catchError, finalize, of, shareReplay, startWith, Subject, switchMap, take, takeUntil, throwError } from 'rxjs';
 
 @Component({
   selector: 'lfx-vote-cast-drawer',
@@ -63,6 +63,10 @@ export class VoteCastDrawerComponent {
   protected readonly loadingVote = signal<boolean>(false);
   protected readonly submitting = signal<boolean>(false);
   protected readonly abstain = signal<boolean>(false);
+
+  // Emits when the drawer is closed. Used to cancel any in-flight submit so the
+  // success toast can't fire after the user "cancelled" (see PR #710 review).
+  private readonly cancelSubmit$ = new Subject<void>();
 
   // === Shared Observables ===
   private readonly voteId$ = toObservable(this.voteId).pipe(shareReplay({ bufferSize: 1, refCount: true }));
@@ -111,6 +115,13 @@ export class VoteCastDrawerComponent {
 
   // === Protected Methods ===
   protected onClose(): void {
+    // Cancel any in-flight submit so its success/error toast doesn't fire after
+    // the user has already cancelled. The submit subscription pipes through
+    // takeUntil(cancelSubmit$) so emitting here completes the stream cleanly.
+    if (this.submitting()) {
+      this.cancelSubmit$.next();
+      this.submitting.set(false);
+    }
     this.visible.set(false);
   }
 
@@ -155,6 +166,9 @@ export class VoteCastDrawerComponent {
           };
           return this.voteService.createVoteResponse(payload);
         }),
+        // If the user closes the drawer mid-submit, drop the stream silently —
+        // no success/error toast fires, submitting is already reset by onClose.
+        takeUntil(this.cancelSubmit$),
         finalize(() => this.submitting.set(false))
       )
       .subscribe({
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -1,0 +1,261 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { isPlatformBrowser } from '@angular/common';
+import { Component, computed, effect, inject, input, model, output, PLATFORM_ID, signal, Signal } from '@angular/core';
+import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators } from '@angular/forms';
+import { ButtonComponent } from '@components/button/button.component';
+import { RadioButtonComponent } from '@components/radio-button/radio-button.component';
+import { TagComponent } from '@components/tag/tag.component';
+import { PollType } from '@lfx-one/shared';
+import { CreateVoteResponseRequest, PollQuestion, Vote, VoteAnswerInput } from '@lfx-one/shared/interfaces';
+import { PollStatusLabelPipe } from '@pipes/poll-status-label.pipe';
+import { PollStatusSeverityPipe } from '@pipes/poll-status-severity.pipe';
+import { VoteService } from '@services/vote.service';
+import { MessageService } from 'primeng/api';
+import { CheckboxModule } from 'primeng/checkbox';
+import { DrawerModule } from 'primeng/drawer';
+import { SkeletonModule } from 'primeng/skeleton';
+import { catchError, finalize, of, shareReplay, startWith, switchMap, take } from 'rxjs';
+
+import { DatePipe } from '@angular/common';
+
+@Component({
+  selector: 'lfx-vote-cast-drawer',
+  imports: [
+    DrawerModule,
+    SkeletonModule,
+    CheckboxModule,
+    FormsModule,
+    ReactiveFormsModule,
+    ButtonComponent,
+    RadioButtonComponent,
+    TagComponent,
+    PollStatusLabelPipe,
+    PollStatusSeverityPipe,
+    DatePipe,
+  ],
+  templateUrl: './vote-cast-drawer.component.html',
+  styleUrl: './vote-cast-drawer.component.scss',
+})
+export class VoteCastDrawerComponent {
+  // === Services ===
+  private readonly voteService = inject(VoteService);
+  private readonly messageService = inject(MessageService);
+  private readonly platformId = inject(PLATFORM_ID);
+
+  // === Inputs ===
+  public readonly voteId = input<string | null>(null);
+  public readonly listVote = input<Vote | null>(null);
+
+  // === Model Signals (two-way binding) ===
+  public readonly visible = model<boolean>(false);
+
+  // === Outputs ===
+  public readonly voteSubmitted = output<string>();
+
+  // === Forms ===
+  // Per-question form. Keys are question_id. Each value is:
+  //   - string | null for single_choice (selected choice_id)
+  //   - string[] | null for multiple_choice (array of selected choice_ids)
+  // Submit normalises both shapes into VoteAnswerInput.choice_ids.
+  public readonly form = new FormGroup({});
+
+  // === Writable Signals ===
+  protected readonly loadingVote = signal<boolean>(false);
+  protected readonly submitting = signal<boolean>(false);
+  protected readonly abstain = signal<boolean>(false);
+
+  // === Shared Observables ===
+  private readonly voteId$ = toObservable(this.voteId).pipe(shareReplay({ bufferSize: 1, refCount: true }));
+
+  // === Derived Signals (from API) ===
+  protected readonly vote: Signal<Vote | null> = this.initVote();
+  // form.statusChanges is the canonical reactive surface for FormGroup validity.
+  // We pipe to startWith(form.status) so the signal has a synchronous initial value
+  // (status emissions only fire on changes, not on initial construction).
+  private readonly formStatus: Signal<string> = toSignal(this.form.statusChanges.pipe(startWith(this.form.status)), { initialValue: this.form.status });
+
+  // === Computed Signals ===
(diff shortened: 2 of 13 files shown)


apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,241 @@
...
+      if (this.abstain()) {
+        this.form.disable({ emitEvent: false });
+      } else {
+        this.form.enable({ emitEvent: false });
+      }
+    });
+  }
+
+  // === Protected Methods ===
+  protected onClose(): void {
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 99645b28. The submit chain now pipes through `takeUntil(cancelSubmit$)` and `onClose()` emits on that subject when called while `submitting()` is true. The HTTP stream is cancelled cleanly, no toast fires after the user closes the drawer mid-submit, and `submitting` resets immediately. Good catch — thanks!

apps/lfx-one/src/app/shared/services/vote.service.ts
```diff
@@ -119,4 +128,19 @@ export class VoteService {
...
+
+  public getMyVoteResponse(voteUid: string): Observable<MyVoteResponse | null> {
+    return this.http.get<MyVoteResponse | null>(`/api/votes/${voteUid}/my-response`).pipe(
+      catchError((err) => {
+        // Log the breadcrumb but degrade gracefully — callers (cast drawer) surface a user-facing
+        // toast and a `null` response triggers the "INVITATION_NOT_FOUND" path on submit.
+        console.error(`Failed to load my-response for vote ${voteUid}:`, err);
+        return of(null);
+      })
+    );
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 99645b28. `getMyVoteResponse` now narrows the `catchError` to `HttpErrorResponse` and only returns `of(null)` when `err.status === 404`. Everything else (500 / network / etc.) re-throws via `throwError(() => err)` so the cast drawer's submit pipeline surfaces a generic "Unable to submit vote" toast instead of misreporting transient errors as a missing invitation. Good catch on the UX impact.

apps/lfx-one/src/server/controllers/vote.controller.ts
```diff
@@ -236,6 +236,33 @@ export class VoteController {
...
+          service: 'vote_controller',
+        })
+      ) {
+        return;
+      }
+
+      const response = await this.voteService.getMyVoteResponse(req, uid);
+      logger.success(req, 'get_my_vote_response', startTime, { vote_uid: uid, found: !!response });
+      res.json(response);
+    } catch (error) {
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 99645b28. `getMyVoteResponse` now throws `ResourceNotFoundError` via `next(...)` when no row exists for the user, matching the surveys pattern in `SurveyController.getMyResponse`. Updated the client-side `getMyVoteResponse` (per the previous thread) to map 404 → `null` so the cast drawer's `INVITATION_NOT_FOUND` path still triggers correctly. Clients now distinguish "no invitation" from other failures via the status code.

manishdixitlfx:
## Review Feedback Addressed

Commit: 99645b28

### Changes Made
- **vote-cast-drawer.component.ts**: cancel any in-flight submit on close so the success toast can't fire after the user navigates away. Adds a `cancelSubmit$` Subject and pipes the submit chain through `takeUntil`. (per @copilot-pull-request-reviewer)
- **shared/services/vote.service.ts** (`getMyVoteResponse`): narrowed the `catchError` fallback to 404 only — transient 500/network errors now re-throw so the cast drawer reports them as a generic submit failure instead of misreporting as "Unable to find your invitation". (per @copilot-pull-request-reviewer)
- **server/controllers/vote.controller.ts**: throw `ResourceNotFoundError` (→ structured 404) when no `vote_response` row exists for the user, matching `SurveyController.getMyResponse`. Lets clients distinguish "no invitation" from other failures. (per @copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+    for (const question of questions) {
+      if (this.form.contains(question.question_id)) continue;
+      if (question.type === 'multiple_choice') {
+        this.form.addControl(question.question_id, new FormControl<string[] | null>(null, Validators.required), { emitEvent: false });
+      } else {
+        this.form.addControl(question.question_id, new FormControl<string | null>(null, Validators.required), { emitEvent: false });
+      }
+    }
+  }
```
dealako, line 221: **[blocking] B1 — `formStatus` signal goes stale after `rebuildForm`**

Tracing construction order:
1. Field init: `form = new FormGroup({})` — status is `'VALID'` (empty form).
2. Field init: `formStatus = toSignal(form.statusChanges.pipe(startWith(form.status)), ...)` — captures `'VALID'` synchronously.
3. When a vote loads, this `rebuildForm` adds required controls with `{ emitEvent: false }`. The form's internal `status` flips to `'INVALID'`, but `statusChanges` never emits (suppressed).
4. `formStatus` signal stays `'VALID'` → `submitDisabled()` returns `false` → **Submit is enabled with nothing selected**.
5. On click, `buildAnswers` → `normaliseChoiceIds(null)` returns `[]` for every question → ballot posted with empty `choice_ids`.

Reproducible on the first open of any fresh vote with `allow_abstain=false`.

**Recommended fix** — add one line at the end of `rebuildForm` to kick a single status emission after all controls are reconciled:

```ts
private rebuildForm(questions: PollQuestion[]): void {
  // ...existing add/remove with emitEvent: false...
  this.form.updateValueAndValidity({ emitEvent: true });
}
```

This keeps the per-control churn quiet and produces a single status emission per reconciliation.

> Agreement with @copilot-pull-request-reviewer (low-confidence comment in [PRR_kwDOPMiRQ87__YFN](https://github.com/linuxfoundation/lfx-self-serve/pull/710#pullrequestreview-PRR_kwDOPMiRQ87__YFN) around line 226).

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.rebuildForm(questions);
+    });
+
+    // Reset transient state when the drawer closes so reopening starts fresh.
+    effect(() => {
+      if (!this.visible()) {
+        this.abstain.set(false);
+        this.submitting.set(false);
+      }
+    });
```
dealako: **[minor] B2 — Submit cancellation misses the ESC / overlay-dismiss path**

The `99645b28` fix for [#3245408055](https://github.com/linuxfoundation/lfx-self-serve/pull/710#discussion_r3245408055) routes `cancelSubmit$.next()` only through `onClose()`. But `[(visible)]` is a `model<boolean>`, and PrimeNG's `<p-drawer [modal]="true">` defaults `dismissable` to `true`, so **ESC and overlay-click flip `visible` to `false` directly, never invoking `onClose()`**.

This effect resets `submitting` but never emits on `cancelSubmit$`, so the in-flight HTTP still completes, the success toast fires after the drawer has closed, and `voteSubmitted.emit()` triggers a parent list refresh — exactly the UX the original thread asked to prevent.

**Minimal fix** — fold cancellation into this effect so all close paths converge:

```ts
effect(() => {
  if (!this.visible()) {
    if (this.submitting()) {
      this.cancelSubmit$.next();
    }
    this.abstain.set(false);
    this.submitting.set(false);
  }
});

protected onClose(): void {
  this.visible.set(false); // effect handles the rest
}
```

> Partial agreement with @copilot-pull-request-reviewer — the close-button path is resolved; the ESC/overlay-dismiss path remains.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,255 @@
...
+    const desiredIds = new Set(questions.map((q) => q.question_id));
+    for (const existingId of Object.keys(this.form.controls)) {
+      if (!desiredIds.has(existingId)) {
+        this.form.removeControl(existingId, { emitEvent: false });
+      }
+    }
+    for (const question of questions) {
+      if (this.form.contains(question.question_id)) continue;
+      if (question.type === 'multiple_choice') {
+        this.form.addControl(question.question_id, new FormControl<string[] | null>(null, Validators.required), { emitEvent: false });
```
dealako: **[minor] M1 — `multiple_choice` FormControl initial value should be `[]`, not `null`**

PrimeNG `<p-checkbox [binary]="false">` reads/writes the bound value as an array. Initialising to `null` creates an asymmetry: a user who never touches the control sees `null`, while a user who selects-then-clears sees `[]`. Both are invalid under `Validators.required`, so behaviour is equivalent today, but `null` is one extra state the codebase has to reason about.

Suggested change:
```ts
new FormControl<string[]>([], Validators.required)
```

> Agreement with @copilot-pull-request-reviewer (low-confidence comment in [PRR_kwDOPMiRQ87__YFN](https://github.com/linuxfoundation/lfx-self-serve/pull/710#pullrequestreview-PRR_kwDOPMiRQ87__YFN) around line 225).

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -330,15 +332,26 @@ export class VoteService {
...
         ...(pageToken && { page_token: pageToken }),
       })
     );
 
+    // Track which votes the user has already submitted a ballot for. `'submitted'` is the
+    // indexer's terminal state for a cast ballot; anything else (e.g. `'awaiting_response'`)
+    // is treated as still-invited.
+    const respondedVoteUids = new Set<string>();
+    for (const r of responses) {
+      if (r.vote_uid && r.vote_status === 'submitted') respondedVoteUids.add(r.vote_uid);
```
dealako: **[minor] M2 — Magic string `'submitted'` for indexer `vote_status`**

The same literal already lives at line 289 in `hasUserAlreadyVoted`. Two raw string comparisons against an indexer enum value is a typo hazard — a regression at either site would silently miscount responses without a compile-time error.

Worth promoting to a shared constant in `@lfx-one/shared`, e.g.:

```ts
// packages/shared/src/enums/vote.enum.ts
export const INDEXER_VOTE_STATUS = {
  AWAITING: 'awaiting_response',
  SUBMITTED: 'submitted',
} as const;
```

Not blocking — easy follow-up if you'd rather not expand scope here.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,255 @@
...
+  public readonly form = new FormGroup({});
+
+  // === Writable Signals ===
+  protected readonly loadingVote = signal<boolean>(false);
+  protected readonly submitting = signal<boolean>(false);
+  protected readonly abstain = signal<boolean>(false);
+
+  // Emits when the drawer is closed. Used to cancel any in-flight submit so the
+  // success toast can't fire after the user "cancelled" (see PR #710 review).
+  private readonly cancelSubmit$ = new Subject<void>();
```
dealako: **[nit] N1 — Field-ordering deviation from component-organization rules**

`cancelSubmit$` is declared between the writable signals block and the shared-observables block. Per `.claude/rules/component-organization.md` §4, private stream state should be grouped with `voteId$` under the `// === Shared Observables ===` header (or a dedicated `// === Internal Streams ===` block adjacent to it).

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,255 @@
...
+          this.messageService.add({
+            severity: 'success',
+            summary: 'Vote submitted',
+            detail: `Your ${isAbstain ? 'abstention' : 'ballot'} has been recorded.`,
+            life: 3000,
+          });
+          this.voteSubmitted.emit(vote.uid);
+          this.visible.set(false);
+        },
+        error: (err: Error) => {
```
dealako: **[nit] N2 — `error: (err: Error)` callback type is misleading**

The submit chain can fail with either:
- `new Error('INVITATION_NOT_FOUND')` thrown from the `switchMap` (a true `Error`), or
- an `HttpErrorResponse` rethrown by `getMyVoteResponse` for non-404 failures, or a network/upstream `HttpErrorResponse` from `createVoteResponse`.

`HttpErrorResponse` extends `Error` (has a `message` getter), so the check works today, but typing the callback as `Error` makes the dual-mode failure shape invisible to future readers. Suggest `err: Error | HttpErrorResponse` (or `unknown` with an inline narrowing guard).

dealako (dismissed):
Hey **Manish** — thanks for splitting this off LFXV2-1839 cleanly; the Me-lens cast-vs-view branching is well-scoped and the surveys-parity choices on the server side (`MyVoteResponse`, `ResourceNotFoundError`, fetch-all-then-filter) keep this PR easy to reason about. The earlier responsiveness to Copilot's three inline threads (commit `99645b28`) is appreciated.

Two things would land this for me:

1. **[B1]** The `formStatus` signal is stale right after `rebuildForm` because every `addControl`/`removeControl` passes `emitEvent: false` — Submit shows enabled on first open and a click sends an empty ballot. A single `this.form.updateValueAndValidity({ emitEvent: true })` at the end of `rebuildForm` should fix it.
2. **[B2]** `cancelSubmit$.next()` only fires from `onClose()`, so ESC and overlay-dismiss still let the in-flight submit complete and the success toast fires post-close. Moving the cancel into the existing visibility effect collapses all close paths into one.

Plus three minor / nit items in the inline comments below — none blocking.

**Counts:** 1 blocking, 3 minor, 2 nit.
**Decision:** ❌ Needs changes before approval.

---

**Revision tracking — already resolved:**

| Thread | Issue | Status |
|---|---|---|
| [#3245408055](https://github.com/linuxfoundation/lfx-self-serve/pull/710#discussion_r3245408055) | close-while-submitting cancellation | Partial — see B2 |
| [#3245408070](https://github.com/linuxfoundation/lfx-self-serve/pull/710#discussion_r3245408070) | `getMyVoteResponse` swallows all errors as null | ✅ Fixed |
| [#3245408090](https://github.com/linuxfoundation/lfx-self-serve/pull/710#discussion_r3245408090) | controller returns 200 + null instead of 404 | ✅ Fixed |

CodeRabbit's latest pass on `99645b28` returned no actionable comments.

> Generated with [Claude Code](https://claude.ai/code)

MRashad26 (commented):
Strict CLAUDE.md / code-enforcer-agent review on PR #710 — 5 critical + 3 major + 4 minor findings. Five items reinforce dealako's open feedback (B1/N1/M1/M2/N2) that hasn't been addressed in 99645b28. Item #5 is a voting-service indexer alignment risk worth verifying before merge.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,255 @@
...
+  protected readonly allowAbstain: Signal<boolean> = computed(() => !!this.vote()?.allow_abstain);
+  protected readonly submitDisabled: Signal<boolean> = computed(() => {
+    if (this.submitting()) return true;
+    if (this.abstain()) return false;
+    return this.formStatus() !== 'VALID';
+  });
+
+  public constructor() {
+    // Rebuild the form whenever a new vote is loaded so controls match its questions.
+    effect(() => {
```
MRashad26: **🔴 Critical — `effect()` used for behavioral side-effects (3 places).**

**Problem:** The constructor declares three `effect()` blocks (here at L93, then L99 for the visible-reset, and L107 for the abstain disable/enable) that drive non-trivial behavior: rebuilding the FormGroup when `questions()` changes, resetting `abstain`/`submitting` when the drawer closes, and toggling `form.disable()`/`form.enable()` on `abstain()`.

**Why it's a problem:** CLAUDE.md is explicit — *"No `effect()` — use `toObservable()` with RxJS pipes instead (except for simple logging)."* `effect()` for behavior is exactly what the rule prohibits — it makes the reactive graph implicit, makes ordering across signals fragile, and is the convention every other component in this repo follows.

**Suggested fix:** Convert each block to a `toObservable` pipe. Example for the rebuild-on-questions effect:
```ts
toObservable(this.questions)
  .pipe(takeUntilDestroyed(this.destroyRef))
  .subscribe((questions) => this.rebuildForm(questions));
```
Apply the same shape to the `visible()` reset and `abstain()` disable/enable blocks. Inject `DestroyRef` if not already in scope.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
```diff
@@ -0,0 +1,146 @@
...
+            </div>
+          }
+        </form>
+
+        @if (allowAbstain()) {
+          <div class="rounded-lg border border-slate-200 bg-slate-50 p-4 flex flex-col gap-2">
+            <div class="flex items-center gap-2" data-testid="vote-cast-abstain-toggle">
+              <p-checkbox
+                [binary]="true"
+                [ngModel]="abstain()"
```
MRashad26: **🔴 Critical — `[ngModel]` + `(ngModelChange)` + `FormsModule` on the Abstain toggle.**

**Problem:** L122-124 uses `[ngModel]="abstain()" (ngModelChange)="onAbstainChange($event)"` with `FormsModule` imported on the component. The Abstain control is bound through the FormsModule path rather than ReactiveFormsModule.

**Why it's a problem:** CLAUDE.md — *"ReactiveFormsModule always — never `(onInput)`, `(onChange)` event bindings to set values, never `[(ngModel)]` with FormsModule."* The rest of this drawer correctly uses `[formGroup]="form"` + `[formControlName]` — Abstain breaks the pattern and forces a second forms module into the component for no reason.

**Suggested fix:** Add an `abstain` FormControl to the FormGroup (or a sibling FormGroup) and bind via `[formControl]`/`[formControlName]`:
```html
<p-checkbox [binary]="true" [formControl]="abstainControl" inputId="vote-cast-abstain" />
```
Then drive the existing `abstain` signal from the control via `toSignal(this.abstainControl.valueChanges, { initialValue: false })`, and drop `FormsModule` + `onAbstainChange` entirely.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
```diff
@@ -0,0 +1,146 @@
...
+            {{ voteData.description }}
+          </div>
+        }
+
+        <form [formGroup]="form" class="flex flex-col gap-6">
+          @for (question of questions(); track question.question_id) {
+            <div class="rounded-lg border border-slate-200 p-4 flex flex-col gap-3" [attr.data-testid]="'vote-cast-question-' + question.question_id">
+              <div class="flex flex-col gap-1">
+                <h3 class="text-sm font-semibold text-slate-900">{{ question.prompt }}</h3>
+                @if (isMultipleChoice(question)) {
```
MRashad26: **🔴 Critical — Function calls inside `@if` template control flow.**

**Problem:** `@if (isMultipleChoice(question))` at L81, `@if (isSingleChoice(question))` at L87, and `@else if (isMultipleChoice(question))` at L98 all call protected methods on every change-detection pass.

**Why it's a problem:** CLAUDE.md — *"No functions in HTML templates — only signals, computed values, or pipes."* Memory note `feedback_no_functions_in_templates` calls this out as a HIGH violation that catches function calls in `@if`, `@let`, `track`, interpolations, and bindings — not just `{{ fn() }}`. Functions in templates re-run on every CD cycle and bypass the zoneless-compatible memoization that signals/pipes give you.

**Suggested fix:** Inline the literal comparison — `question.type` is just a string union:
```html
@if (question.type === 'multiple_choice') { ... }
@if (question.type === 'single_choice') { ... }
```
Or, if you want a single source of truth, add a pure pipe (`PollQuestionTypePipe`) that returns the discriminator. Drop `isSingleChoice`/`isMultipleChoice` from the component.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  // === Shared Observables ===
+  private readonly voteId$ = toObservable(this.voteId).pipe(shareReplay({ bufferSize: 1, refCount: true }));
+
+  // === Derived Signals (from API) ===
+  protected readonly vote: Signal<Vote | null> = this.initVote();
+  // form.statusChanges is the canonical reactive surface for FormGroup validity.
+  // We pipe to startWith(form.status) so the signal has a synchronous initial value
+  // (status emissions only fire on changes, not on initial construction).
+  private readonly formStatus: Signal<string> = toSignal(this.form.statusChanges.pipe(startWith(this.form.status)), { initialValue: this.form.status });
```
MRashad26: **🔴 Critical — `formStatus` signal goes stale after `rebuildForm` (reinforcing dealako's open B1 blocker).**

**Problem:** Construction order:
1. `form = new FormGroup({})` — status is `'VALID'` (empty form).
2. `formStatus = toSignal(form.statusChanges.pipe(startWith(form.status)))` — captures `'VALID'` synchronously here at L79.
3. The effect at L93 calls `rebuildForm(questions)`, which adds required controls via `this.form.addControl(..., { emitEvent: false })` at L236 / L238.

Because `{ emitEvent: false }` is passed, `statusChanges` never fires after rebuild. The signal still returns `'VALID'` even though the form is now `INVALID` (required fields unfilled). `submitDisabled()` returns `false` → the Submit button is enabled → user can submit an empty ballot.

**Why it's a problem:** This is a real correctness bug, not a style nit. dealako already flagged it as `[blocking] B1` and the author hasn't responded. It will reproduce on every multi-question vote that uses the Me-lens cast flow.

**Suggested fix (lowest-risk):** Drop `{ emitEvent: false }` from the `addControl`/`removeControl` calls inside `rebuildForm`. The form's status emissions are exactly what `formStatus` needs to track. If you genuinely need to suppress per-control intermediate events, call `this.form.updateValueAndValidity()` once at the end of `rebuildForm` to force a single status emission.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -381,6 +404,39 @@ export class VoteService {
...
+
+    const responses = await fetchAllQueryResources<MyVoteResponse>(req, (pageToken) =>
+      this.microserviceProxy.proxyRequest<QueryServiceResponse<MyVoteResponse>>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', {
+        type: 'vote_response',
+        filters_or: filtersOr,
+        ...(pageToken && { page_token: pageToken }),
+      })
+    );
+
+    const match = responses.find((r) => r?.vote_uid === voteUid && !!r?.uid);
```
MRashad26: **🔴 Critical — Voting-service indexer alignment risk: `r.uid` may always be undefined on query-service vote_response rows.**

**Problem:** `responses.find((r) => r?.vote_uid === voteUid && !!r?.uid)` at L436 — and the call sites on the client side — all assume `MyVoteResponse.uid` is the pre-allocated `vote_response_uid`. But this PR's source for that field is the **query service** (`type=vote_response`, L431), not the voting service's REST API.

**Why it's a problem:** The voting service's Go indexer serializes primary keys as `<type>_uid` JSON tags, not `uid` — exactly the same pattern that already required the `raw.uid || raw.vote_uid` normalization on `IndexedVote` (see `normalizeIndexedVote` further down in this same file). If `vote_response` indexer rows follow the same convention, the field exposed in `resource.data` is `vote_response_uid` and `r.uid` is always `undefined` → `!!r?.uid` is always false → `responses.find(...)` returns `undefined` → server returns 404 → every cast attempt surfaces "Unable to find your invitation" to the user. The PR's test plan claims it works for v1-synced votes, so either (a) the indexer DOES happen to expose `uid` on vote_response rows, or (b) end-to-end was never actually exercised against the indexer.

**Suggested fix:** Normalize defensively, matching the existing pattern in this file:
```ts
const match = responses
  .map((r) => ({ ...r, uid: r?.uid || (r as { vote_response_uid?: string })?.vote_response_uid || '' }))
  .find((r) => r.vote_uid === voteUid && !!r.uid);
```
And add a `vote_response_uid?: string` to the `MyVoteResponse` interface so the cast is type-safe. **Before merge:** capture an actual `/query/resources?type=vote_response` response from a deployed env and confirm which field name the Go indexer emits.

packages/shared/src/interfaces/poll.interface.ts
```diff
@@ -215,6 +215,13 @@ export interface Vote {
   total_voting_request_invitations?: number;
   /** Number of responses received */
   num_response_received?: number;
+  /**
+   * The current user's response state for this vote, set only by the Me-lens
+   * `getMyVotes` server decoration. RESPONDED means the user has submitted a ballot
+   * (the indexed vote_response row has `vote_status === 'submitted'`); AWAITING_RESPONSE
+   * means invited but not yet voted. Absent on raw upstream Vote responses (e.g. GET /votes/:uid).
+   */
+  response_status?: VoteResponseStatus;
```
MRashad26, line 219: **🟡 Major — Multi-paragraph JSDoc / multi-line comment blocks throughout the PR.**

**Problem:** CLAUDE.md violations across at least 8 sites:
- `poll.interface.ts:218-223` — 6-line JSDoc on `Vote.response_status` (anchored here)
- `poll.interface.ts:238-243` + 6 per-field JSDocs on `MyVoteResponse` (L244-258)
- `vote.controller.ts:239-242` — 4-line JSDoc on `getMyVoteResponse`
- `vote.service.ts` (server) — 9-line JSDoc on `getMyVoteResponse` (L407-415) and 3-4 line inline blocks at L335-338, L347-349, L368-371
- `vote-cast-drawer.component.ts` — 4-line comment on `form` (L56-59), multi-line `cancelSubmit$` (L67-68), multi-line `onClose` (L117-120), multi-line `onSubmit` switchMap (L148-151), multi-line `takeUntil` (L169-170)
- `votes-dashboard.component.ts:354-360` — 7-line block on `shouldOpenCastDrawerFor`

**Why it's a problem:** Global CLAUDE.md — *"Default to writing no comments. Only add one when the WHY is non-obvious… Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* Most of these comments restate **what** the code/interface field already says (e.g., the field `uid: string` is documented as `"vote_response_uid — pre-allocated by the voting service on invitation."` — a comment that exists only to apologise for the wrong field name). They will rot the first time anything is renamed.

**Suggested fix:** Strip the JSDocs. Keep at most a single-line note where the **why** is genuinely load-bearing:
- `Vote.response_status`: `// Set only by getMyVotes (Me lens); absent on raw upstream Vote responses.` — one line.
- `MyVoteResponse`: drop the JSDoc + per-field JSDocs; field names are the documentation. (And rename `uid` → `vote_response_uid` if that's what the data actually is — see #5.)
- Server `getMyVoteResponse`: drop the 9-line JSDoc; the method body is short enough to be self-documenting.
- Multi-line inline blocks inside the cast drawer: collapse to one line each, or remove if they paraphrase the surrounding code.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -381,6 +404,39 @@ export class VoteService {
...
+    const username = rawUsername ? stripAuthPrefix(rawUsername) : null;
+    const email = getEffectiveEmail(req);
+
+    if (!username && !email) return null;
+
+    const filtersOr: string[] = [];
+    if (email) filtersOr.push(`user_email:${email}`);
+    if (username) filtersOr.push(`username:${username}`);
+
+    const responses = await fetchAllQueryResources<MyVoteResponse>(req, (pageToken) =>
```
MRashad26, line 413: **🟡 Major — Server fetches the user's entire `vote_response` history then filters in-memory for a single-row lookup.**

**Problem:** `getMyVoteResponse` paginates through *every* vote_response row tagged with the user's email/username (L428-433 — `fetchAllQueryResources` walks all pages), then does `responses.find((r) => r?.vote_uid === voteUid && !!r?.uid)` at L436.

**Why it's a problem:** A long-tenured committee member has hundreds of historical vote_response rows. Each cast-drawer open pays a multi-page round-trip to the query service to find one row that should be O(1) at the source. This also amplifies the indexer-alignment risk in #5 — more rows to scan = more failure surface. The `getMyVotes` flow has a structural excuse for fetching all rows (it needs them all to dedupe by `vote_uid`); `getMyVoteResponse` does not.

**Suggested fix:** Push the `vote_uid` filter to the query service. Two options:
1. **Preferred — add to the filters:** if the query service supports compound filtering, send `filters: ['vote_uid:' + voteUid]` alongside the email/username `filters_or`. The upstream returns at most one row.
2. **Single-row fetch via the voting service REST API:** if the voting service exposes `GET /vote_responses?vote_uid=...&user_email=...`, prefer that — it avoids the indexer-shape question (#5) entirely because the REST shape is canonical (`uid`, not `vote_response_uid`).

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
MRashad26: **🟡 Major — `cancelSubmit$` field ordering deviates from the component-organization rules (reinforcing dealako's N1).**

**Problem:** `cancelSubmit$` is declared at L69 between the `// === Writable Signals ===` block above and the `// === Shared Observables ===` block at L72.

**Why it's a problem:** Per `.claude/rules/component-organization.md` §4 (and the CLAUDE.md component structure ordering), private stream state belongs with `voteId$` under `// === Shared Observables ===`. dealako already flagged this and it remains open. Keeping the convention is what lets readers skim a 250-line component without re-scanning every section to find where reactive primitives live.

**Suggested fix:** Move the declaration into the Shared Observables section:
```ts
// === Shared Observables ===
private readonly cancelSubmit$ = new Subject<void>();
private readonly voteId$ = toObservable(this.voteId).pipe(shareReplay({ bufferSize: 1, refCount: true }));
```
Drop the multi-line comment above it while you're at it (#6).

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
MRashad26: **🔵 Minor — `multiple_choice` FormControl initial value should be `[]`, not `null` (reinforcing dealako's M1).**

**Problem:** `new FormControl<string[] | null>(null, Validators.required)` at L236 vs the `single_choice` branch at L238 (`new FormControl<string | null>(null, ...)`). PrimeNG `<p-checkbox [binary]="false">` reads/writes its bound value as an array.

**Why it's a problem:** Asymmetry between initial state (`null`) and post-interaction state (`[]` after select-then-deselect). It also makes `Validators.required` semantics subtly different — `required` flags `null` and `[]` both as invalid, but downstream consumers checking `value?.length` vs `value === null` will disagree. Type widening to `string[] | null` is a code smell when a fresh array would do.

**Suggested fix:**
```ts
this.form.addControl(question.question_id, new FormControl<string[]>([], Validators.required), { emitEvent: false });
```
Drop the `| null` from the generic. Same control shape for both interaction paths.

apps/lfx-one/src/server/services/vote.service.ts
MRashad26: **🔵 Minor — Magic string `'submitted'` for indexer `vote_status` (reinforcing dealako's M2).**

**Problem:** `r.vote_status === 'submitted'` at L352, and the same literal already lives at L289 in `hasUserAlreadyVoted` (per dealako's prior comment). Two raw string comparisons against an indexer enum value with no compile-time link.

**Why it's a problem:** Typo hazard — a regression at either site would silently miscount responses, with no test failure until production data diverges. The indexer's status set (`'awaiting_response'`, `'submitted'`, possibly more) is a finite domain; treating it as untyped strings is the same anti-pattern shared interfaces are meant to prevent (memory note `feedback_shared_interfaces`).

**Suggested fix:** Add a typed const or enum in `@lfx-one/shared/enums` (or `@lfx-one/shared/constants`):
```ts
export const INDEXED_VOTE_RESPONSE_STATUS = {
  AWAITING: 'awaiting_response',
  SUBMITTED: 'submitted',
} as const;
export type IndexedVoteResponseStatus = (typeof INDEXED_VOTE_RESPONSE_STATUS)[keyof typeof INDEXED_VOTE_RESPONSE_STATUS];
```
Use `INDEXED_VOTE_RESPONSE_STATUS.SUBMITTED` here and at L289. While you're there, narrow `MyVoteResponse.vote_status?: string` to `IndexedVoteResponseStatus` in `poll.interface.ts`.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
MRashad26: **🔵 Minor — `error: (err: Error)` callback type is misleading (reinforcing dealako's N2).**

**Problem:** The submit chain can fail in three ways: (a) `new Error('INVITATION_NOT_FOUND')` thrown from the `switchMap` — a true `Error`; (b) an `HttpErrorResponse` rethrown by `getMyVoteResponse` for non-404 failures; (c) a network/upstream `HttpErrorResponse` from `createVoteResponse`. Typing the callback as `(err: Error)` at L185 is a lie for cases (b) and (c).

**Why it's a problem:** `HttpErrorResponse` does NOT extend `Error`, so `err.message` lookups behave differently at runtime than the type suggests. Today the only narrowing in the body is `err?.message === 'INVITATION_NOT_FOUND'` — for an `HttpErrorResponse`, `err.message` is the HTTP message string (e.g., `"Http failure response for /api/..."`), which works coincidentally but ties correctness to a coincidence.

**Suggested fix:**
```ts
error: (err: unknown) => {
  const isInvitationMissing = err instanceof Error && err.message === 'INVITATION_NOT_FOUND';
  // ...
}
```
Narrow explicitly. If you want more granular toasts later (e.g., a different message for 500s), check `err instanceof HttpErrorResponse` and inspect `err.status` directly.

apps/lfx-one/src/app/modules/votes/components/votes-table/votes-table.component.html
```diff
@@ -191,16 +191,28 @@
...
-                        label="Review"
-                        icon="fa-light fa-clipboard"
-                        severity="primary"
-                        [outlined]="true"
-                        size="small"
-                        styleClass="whitespace-nowrap"
-                        (onClick)="onViewVote(vote.uid)"
-                        [attr.data-testid]="'votes-cast-' + vote.uid">
-                      </lfx-button>
+                      @if (vote.response_status === VoteResponseStatus.AWAITING_RESPONSE) {
```
MRashad26: **🔵 Minor — Duplicate `<lfx-button>` blocks for the Vote/Review branch.**

**Problem:** L194-214 renders two near-identical `<lfx-button>` blocks; only `label`, `icon`, and the `[outlined]` flag differ between the AWAITING_RESPONSE branch (L195-203) and the else branch (L204-213). Seven of eight attributes (`severity`, `size`, `styleClass`, `(onClick)`, `[attr.data-testid]`) are byte-for-byte duplicates.

**Why it's a problem:** Two failure modes:
1. Schema drift — the next time someone changes `styleClass` or the test-id pattern, they will update one branch and miss the other (this pattern recurs in many tables in the repo and has caused exactly that bug before).
2. Reviewers have to diff the two blocks line-by-line to confirm they differ only in the three intentional fields. That's wasted attention.

**Suggested fix:** Single button driven by computed-row props. Either inline:
```html
<lfx-button
  [label]="vote.response_status === VoteResponseStatus.AWAITING_RESPONSE ? 'Vote' : 'Review'"
  [icon]="vote.response_status === VoteResponseStatus.AWAITING_RESPONSE ? 'fa-light fa-check-to-slot' : 'fa-light fa-clipboard'"
  severity="primary"
  [outlined]="vote.response_status !== VoteResponseStatus.AWAITING_RESPONSE"
  size="small"
  styleClass="whitespace-nowrap"
  (onClick)="onViewVote(vote.uid)"
  [attr.data-testid]="'votes-cast-' + vote.uid">
</lfx-button>
```
Or — cleaner — expose a `castOrReviewButtonProps(vote)` pure pipe / computed returning `{ label, icon, outlined }`, and bind once. The latter scales when the action set grows beyond Vote/Review.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

Replaced `toSignal(form.statusChanges)` with an explicit `formVersion` signal that's bumped after each `rebuildForm` (and after each abstain disable/enable). `submitDisabled` is now a `computed()` that reads `formVersion()` so it re-evaluates whenever controls are added/removed/disabled with `{ emitEvent: false }`. Submit button now correctly tracks validity through the full lifecycle.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

Moved cancellation from `onClose()` into a subscription on `toObservable(this.visible).pipe(filter((v) => !v))` so it fires on every dismissal path (close button, ESC, overlay click). Also resets `submitting` and `abstainControl` in the same handler. Implemented via `takeUntilDestroyed` rather than `effect()` since CLAUDE.md prohibits effects for behavioral side-effects.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

`new FormControl<string[]>([], { nonNullable: true, validators: [Validators.required] })` — matches PrimeNG's array-valued contract for `[binary]="false"` checkboxes.

apps/lfx-one/src/server/services/vote.service.ts
manishdixitlfx: Done — fixed in commit 72955d82.

Added `IndexedVoteResponseStatus` enum to `@lfx-one/shared/enums/poll.enum.ts` (values `AWAITING_RESPONSE = 'awaiting_response'`, `SUBMITTED = 'submitted'`). Replaced both magic-string call sites in `vote.service.ts` (poll-completion check and getMyVotes responded-set) with `IndexedVoteResponseStatus.SUBMITTED`.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

The rebased structure now groups `cancelSubmit$` with the shared-observables block, between `voteId$` and the derived signals. Field order is: services → inputs → model → outputs → forms → writable signals (`loadingVote`, `submitting`, `formVersion`) → shared observables (`voteId$`, `cancelSubmit$`) → derived signals (`vote`, `abstain`) → computed signals → constructor.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

Narrowed to `error: (err: unknown)` and gated the INVITATION_NOT_FOUND branch on `err instanceof Error && !(err instanceof HttpErrorResponse) && err.message === INVITATION_NOT_FOUND`. Also extracted the literal to a module-level `INVITATION_NOT_FOUND` constant so the throw site and the catch site reference the same symbol.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

All three `effect()` blocks removed. Replaced with:
- `toObservable(this.questions).pipe(takeUntilDestroyed(destroyRef))` for the rebuild trigger
- `this.abstainControl.valueChanges.pipe(takeUntilDestroyed(destroyRef))` for abstain → form.disable/enable
- `toObservable(this.visible).pipe(filter(v => !v), takeUntilDestroyed(destroyRef))` for drawer-dismissal cleanup (covers ESC + overlay paths too — addresses dealako's B2)

No `effect()` calls remain in the file.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
manishdixitlfx: Done — fixed in commit 72955d82.

Abstain is now a `FormControl<boolean>` (`abstainControl`) wired via `[formControl]="abstainControl"`. Dropped `FormsModule` from the component imports entirely. The drawer template uses only `ReactiveFormsModule`.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
manishdixitlfx: Done — fixed in commit 72955d82.

Replaced `@if (isMultipleChoice(q))` / `@if (isSingleChoice(q))` with `@if (question.type === 'multiple_choice')` / `@if (question.type === 'single_choice')` — direct property reads on the iterated `question` value, no method calls. Removed the now-unused helper methods from the component class.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

This reinforces dealako's B1; same fix applies. Explicit `formVersion` signal bumped after every `rebuildForm` / abstain disable / abstain enable. `submitDisabled` is now a `computed()` that reads `formVersion()` so it re-evaluates when controls change despite `{ emitEvent: false }`. The `toSignal(statusChanges)` indirection is gone.

apps/lfx-one/src/server/services/vote.service.ts
manishdixitlfx: Done — fixed in commit 72955d82.

Added a defensive fallback in `getMyVoteResponse`: if `r.uid` is missing but `r.vote_id` (the v1 alias) is populated, use that as the response uid and `logger.warning` the anomaly so we catch indexer drift in production logs. Also added `vote_id?: string` to the `MyVoteResponse` interface as a documented v1 alias. POST `/vote_responses` will still receive a valid `vote_response_uid` in the fallback path.

packages/shared/src/interfaces/poll.interface.ts
manishdixitlfx: Done — fixed in commit 72955d82.

Compressed all flagged multi-line JSDoc blocks to single-line `/** … */`:
- `poll.interface.ts` — `Vote.response_status` and `MyVoteResponse`
- `poll.enum.ts` — `IndexedVoteResponseStatus`
- `vote.controller.ts` — `getMyVoteResponse`
- `vote.service.ts` — `getMyVoteResponse`
- `vote-cast-drawer.component.ts` — inline comments collapsed where possible (kept one-line `Why:` comments only for non-obvious invariants)

apps/lfx-one/src/server/services/vote.service.ts
manishdixitlfx: Done — fixed in commit 72955d82.

Added `filters: ['vote_uid:${voteUid}']` to the query alongside the existing `filters_or` for user identity. The index now narrows to the one (or few) rows for that specific vote+user before returning, instead of paginating the user's full vote_response history. Both clauses AND together at the indexer.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

This reinforces dealako's N1; same fix applies. `cancelSubmit$` is now grouped with the shared-observables block (between `voteId$` and the derived signals) per `component-organization.md`.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

This reinforces dealako's M1; same fix applies. `new FormControl<string[]>([], { nonNullable: true, validators: [Validators.required] })`.

apps/lfx-one/src/server/services/vote.service.ts
manishdixitlfx: Done — fixed in commit 72955d82.

This reinforces dealako's M2; same fix applies. Both call sites now use `IndexedVoteResponseStatus.SUBMITTED` (new enum in `@lfx-one/shared/enums/poll.enum.ts`).

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 72955d82.

This reinforces dealako's N2; same fix applies. `error: (err: unknown)` + `err instanceof Error && !(err instanceof HttpErrorResponse) && err.message === INVITATION_NOT_FOUND`. Extracted the literal to a module-level constant so throw and catch reference the same symbol.

apps/lfx-one/src/app/modules/votes/components/votes-table/votes-table.component.html
manishdixitlfx: Done — fixed in commit 72955d82.

Collapsed the duplicated `<lfx-button>` pair into a single button with a `@let awaiting = vote.response_status === VoteResponseStatus.AWAITING_RESPONSE` binding feeding `label`, `icon`, and `[outlined]` from ternaries on `awaiting`.

manishdixitlfx:

Commit: 72955d82 (rebased on top of 7c552153)

### Code Changes
- **`packages/shared/src/enums/poll.enum.ts`**: added `IndexedVoteResponseStatus` enum (replaces magic strings in `vote.service.ts`) (per @dealako M2, @MRashad26 minor reinforcement)
- **`packages/shared/src/interfaces/poll.interface.ts`**: compressed multi-line JSDoc on `Vote.response_status` and `MyVoteResponse`; added `vote_id?: string` v1 alias for indexer drift defense (per @MRashad26 multi-line JSDoc, indexer-alignment)
- **`apps/lfx-one/src/server/services/vote.service.ts`**:
  - both magic-string sites now use `IndexedVoteResponseStatus.SUBMITTED` (per @dealako M2)
  - `getMyVoteResponse` now narrows the query with `filters: ['vote_uid:<uid>']` so it no longer paginates the user's full vote_response history for a single-row lookup (per @MRashad26 N+1 / history-fetch)
  - defensive `vote_id` fallback with `logger.warning` if indexer drops `uid` (per @MRashad26 indexer-alignment)
- **`apps/lfx-one/src/server/controllers/vote.controller.ts`**: compressed multi-line JSDoc on `getMyVoteResponse` (per @MRashad26)
- **`apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts`**:
  - all three `effect()` blocks removed, replaced with `takeUntilDestroyed(destroyRef)`-scoped RxJS subscriptions (per @MRashad26 critical — CLAUDE.md prohibits `effect()` for behavioral side-effects)
  - explicit `formVersion` signal + `bumpFormVersion()` replaces `toSignal(form.statusChanges)` — `submitDisabled` is now a `computed()` that re-evaluates when controls are added/removed/disabled with `{ emitEvent: false }` (per @dealako B1 + @MRashad26 critical reinforcement)
  - `cancelSubmit$` regrouped with shared observables per `component-organization.md` (per @dealako N1 + @MRashad26 reinforcement)
  - submit cancellation moved to `toObservable(this.visible).pipe(filter(v => !v))` — fires on close button, ESC, and overlay-click (per @dealako B2)
  - `multiple_choice` FormControl initialized to `[]` with `nonNullable: true` (per @dealako M1 + @MRashad26 reinforcement)
  - `error: (err: unknown)` + `err instanceof Error && !(err instanceof HttpErrorResponse) && err.message === INVITATION_NOT_FOUND`; literal extracted to module-level constant (per @dealako N2 + @MRashad26 reinforcement)
  - abstain converted from `signal + ngModel` to `FormControl<boolean>` with `[formControl]` directive; `FormsModule` import removed entirely (per @MRashad26 critical)
- **`apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html`**:
  - replaced `@if (isMultipleChoice(q))` / `@if (isSingleChoice(q))` function calls with direct `question.type === '…'` checks (per @MRashad26 critical — no functions in `@if`)
  - abstain checkbox uses `[formControl]="abstainControl"` (per @MRashad26 critical)
- **`apps/lfx-one/src/app/modules/votes/components/votes-table/votes-table.component.html`**: collapsed duplicate `<lfx-button>` pair into a single button with `@let awaiting = …` driving label/icon/outlined (per @MRashad26 minor)

### Threads Resolved
18 of 18 unresolved threads addressed in this iteration.

### Reconciliation Note
This round was rebased on top of @dealako / copilot-swe-agent's parallel commit 7c552153 — that commit covered B1/B2/M1/M2/N1/N2 with narrower scope (kept `effect()` and `FormsModule`). The rebase merged both efforts; the result keeps the parallel work's authorship while adopting MRashad26's stricter critical-fix patterns where they conflicted, since CLAUDE.md and the reviewer's framework feedback take precedence.

apps/lfx-one/src/app/modules/votes/votes-dashboard/votes-dashboard.component.ts
```diff
@@ -87,14 +89,32 @@ export class VotesDashboardComponent {
...
   protected onViewResults(voteId: string): void {
     this.selectedVoteId.set(voteId);
+    this.castDrawerVisible.set(false);
     this.resultsDrawerVisible.set(true);
   }
 
+  // Refresh My Votes after a successful ballot submission so response_status flips to RESPONDED.
+  protected onVoteSubmitted(): void {
+    this.refresh$.next();
+    this.fetch$.next();
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 0b282b6d.

`onVoteSubmitted()` now only calls `this.refresh$.next()`. `refreshVotes()` (the explicit refresh button handler) still drives both streams since it's the user-facing reset for both Me and non-Me lenses.

packages/shared/src/interfaces/poll.interface.ts
```diff
@@ -215,18 +215,28 @@ export interface Vote {
...
 }
 
+/** Current user's vote_response row, returned by GET /api/votes/:uid/my-response. `uid` is the pre-allocated invitation row — POST /vote_responses must reuse it (a fresh UUID returns 404 upstream). */
+export interface MyVoteResponse {
+  uid: string;
+  vote_uid: string;
+  /** v1 alias for the response row id; identical value to `uid` when indexer populates it. */
+  vote_id?: string;
+  vote_status?: string;
+  voter_removed?: boolean;
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 0b282b6d.

`MyVoteResponse.vote_status` is now typed `IndexedVoteResponseStatus` (added import to `poll.interface.ts`). Both the server fallback path and any client-side comparison are now type-checked against the enum.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -321,24 +323,24 @@ export class VoteService {
...
         type: 'vote_response',
         filters_or: filtersOr,
         ...(pageToken && { page_token: pageToken }),
       })
     );
 
+    const respondedVoteUids = new Set<string>();
+    for (const r of responses) {
+      if (r.vote_uid && r.vote_status === IndexedVoteResponseStatus.SUBMITTED) respondedVoteUids.add(r.vote_uid);
+    }
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 0b282b6d.

Both generic positions on the `fetchAllQueryResources<…>` / `microserviceProxy.proxyRequest<QueryServiceResponse<…>>` call in `getMyVotes` are now `{ vote_uid: string; vote_status?: IndexedVoteResponseStatus }`. Comparison at line 341 (`r.vote_status === IndexedVoteResponseStatus.SUBMITTED`) is now enum-checked end-to-end.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -0,0 +1,234 @@
...
+            vote_response_uid: myResponse.uid,
+            vote_uid: vote.uid,
+            abstain: isAbstain,
+            user_vote_content: userVoteContent,
+          };
+          return this.voteService.createVoteResponse(payload);
+        }),
+        takeUntil(this.cancelSubmit$),
+        finalize(() => this.submitting.set(false))
+      )
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 0b282b6d.

Added `takeUntilDestroyed(this.destroyRef)` to the submit pipeline, placed between `takeUntil(this.cancelSubmit$)` and `finalize(...)`. The stream now terminates on either cancellation path (drawer dismissal *or* component destruction), so `voteSubmitted.emit()` and the success/error toast can't fire after teardown.

manishdixitlfx:

Commit: 0b282b6d (stacked on 72955d82)

Addresses 4 auto-reviewer threads that landed after round 2:

- **`votes-dashboard.component.ts`**: dropped redundant `fetch$.next()` from `onVoteSubmitted()` — Me-lens depends only on `refresh$` (per @copilot)
- **`packages/shared/src/interfaces/poll.interface.ts`**: tightened `MyVoteResponse.vote_status` from `string` to `IndexedVoteResponseStatus` (per @copilot)
- **`apps/lfx-one/src/server/services/vote.service.ts`**: tightened the inline `getMyVotes` query response type to `IndexedVoteResponseStatus` so the enum comparison is fully type-checked (per @copilot)
- **`apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts`**: added `takeUntilDestroyed(destroyRef)` to the submit pipeline alongside `takeUntil(cancelSubmit$)` so the stream tears down on component destruction, preventing toasts from firing after teardown (per @coderabbitai)

4 of 4 unresolved threads resolved.

MRashad26 (approved):

manishdixitlfx:
@dealako — friendly ping for a re-review when you have a moment.

All of your B1/B2/M1/M2/N1/N2 feedback was addressed in commit 72955d82 (round 2, rebased on top of the parallel @copilot-swe-agent commit 7c552153) and one follow-up in 0b282b6d (round 3). Quick summary of the round-2 reconciliation:

- **B1 formStatus staleness**: replaced `toSignal(form.statusChanges)` with an explicit `formVersion` signal bumped after every `rebuildForm` / abstain disable / enable. `submitDisabled` is now a `computed()` that reads `formVersion()` so it re-evaluates when controls are added/removed/disabled with `{ emitEvent: false }`.
- **B2 ESC/overlay dismiss**: moved cancellation from `onClose()` into `toObservable(this.visible).pipe(filter(v => !v))` so all dismissal paths trigger cleanup. Implemented via `takeUntilDestroyed(destroyRef)` rather than `effect()` since CLAUDE.md prohibits effects for behavioral side-effects.
- **M1 `[]` initial**: `new FormControl<string[]>([], { nonNullable: true, validators: [Validators.required] })`.
- **M2 magic string**: extracted `IndexedVoteResponseStatus` enum into `@lfx-one/shared/enums`; both call sites use `IndexedVoteResponseStatus.SUBMITTED`.
- **N1 field ordering**: `cancelSubmit$` regrouped with the shared-observables block.
- **N2 error typing**: `error: (err: unknown)` + `err instanceof Error && !(err instanceof HttpErrorResponse) && err.message === INVITATION_NOT_FOUND`. Extracted the literal to a module-level constant.

Round 3 then added one type tightening per @copilot's catch (`MyVoteResponse.vote_status: IndexedVoteResponseStatus`) and a `takeUntilDestroyed(destroyRef)` to the submit pipeline per @coderabbitai.

@MRashad26 has approved. Your review is the last blocker — happy to address anything you flag in a follow-up, or you can dismiss if everything looks good.

MRashad26 (approved):

PR #711, opened 2026-05-15
feat(voting): add ranked-choice cast UI and ranked results rendering

manishdixitlfx:
## Summary

Builds on **[#710](https://github.com/linuxfoundation/lfx-self-serve/pull/710)** ([LFXV2-1840](https://linuxfoundation.atlassian.net/browse/LFXV2-1840)) which shipped the Plurality (GENERIC) cast flow. With this PR landed, the full **LFX One My Votes** flow supports every `PollType` end-to-end on the UI side — Plurality, Condorcet IRV, Instant Runoff Vote, and Meek STV.

JIRA: [LFXV2-1839](https://linuxfoundation.atlassian.net/browse/LFXV2-1839)

> **Stacked on #710.** PR base is `feat/LFXV2-1840-ballot-casting-plurality`. After #710 merges into `main`, GitHub will auto-retarget this PR to `main` (or I'll retarget manually).

### Cast drawer — ranked-choice cast UI
- Replaces the **"Coming soon"** placeholder for non-`GENERIC` poll types with a real drag-to-rank cast UI using `@angular/cdk/drag-drop` (already a dep — no install needed).
- Each ranked choice renders as a draggable card with a grip handle, a 1-based rank badge, and **up/down arrow buttons** that double as a **keyboard-accessible reorder fallback** (`aria-label`s on both).
- The per-question `FormControl` for ranked questions holds the **ordered `choice_ids[]`** array. `rebuildForm` seeds controls with the question's declared choice order, so the form is valid on load — the no-touch default ranking is the declared order.
- `buildAnswers` branches on `poll_type`:
  - non-GENERIC → emits `ranked_choices: [{ choice_id, choice_rank }]` (1-based)
  - GENERIC → keeps existing `choice_ids` shape (no regression)

### Results drawer — ranked-choice results rendering
- Replaces the **"Advanced Voting Method → View in PCC"** redirect for non-GENERIC poll types with **real ranked-choice results rendering**.
- New per-question card shows **rank distribution per choice** (rank, count, percentage bar) plus a **participation summary** that matches the GENERIC paths (eligible voters / responses / response rate + progress bar).
- **Round-by-round details (IRV / Meek STV)** currently render a "coming soon" banner — upstream `irv_round_summary` / `meek_stv_round_summary` are typed `any`, so structured rendering is deferred to a follow-up.
- **Dropped imports/state:** removed `ButtonComponent` and `environment` imports + `pccVotingUrl` signal / `initPccVotingUrl` initializer. **No PCC link anywhere in the drawer now.**
- Added a local `RankedQuestionView` view-model type (JSDoc invites promotion to `@lfx-one/shared/interfaces` if a second consumer ever needs it).

## Reviewer callouts

1. **`@angular/cdk/drag-drop` first use in the repo** — `DragDropModule` + `CdkDragDrop` + `moveItemInArray`. `@angular/cdk@20.2.14` was already declared in `apps/lfx-one/package.json` so this PR doesn't change deps.
2. **Default ranking = declared order.** If a user opens a ranked ballot and submits without dragging, the recorded ballot reflects the declared choice order. This matches most ranked-choice UIs as the no-touch default; verify with PM if you want partial-rank semantics instead.
3. **`vote-results-drawer` PCC redirect is gone** — both for ops/admin lenses and the Me lens. If anyone still relies on a "View in PCC" escape hatch from the results drawer, flag now.
4. **No backend changes required** — the cast payload shape (`VoteAnswerInput.ranked_choices`) and the results endpoint already expose everything we render.

## Known limitation (separate ticket)

Same SFDC submit blocker as [#710](https://github.com/linuxfoundation/lfx-self-serve/pull/710): `lfx-v2-voting-service POST /vote_responses` still proxies to v1 Salesforce ITX, so v2-only votes return 404. Real prod votes that originated in v1 (or were cross-synced) work end-to-end. Tracked in [LFXV2-1842](https://linuxfoundation.atlassian.net/browse/LFXV2-1842).

## Follow-up PRs (tracked separately)

- **[LFXV2-1841](https://linuxfoundation.atlassian.net/browse/LFXV2-1841)** — My Vote Response drawer (surveys-parity read-only view of submitted ballot).
- **[LFXV2-1842](https://linuxfoundation.atlassian.net/browse/LFXV2-1842)** — Backend: v2-native ballot submission in `lfx-v2-voting-service`.

## Test plan

- [ ] Open a ranked-choice vote (CONDORCET_IRV / INSTANT_RUNOFF_VOTE / MEEK_STV) from Me lens → cast drawer renders draggable list (not "Coming soon" placeholder).
- [ ] Drag a choice to a new position → its rank badge updates; submit button stays enabled (form valid).
- [ ] Use the up/down arrow buttons to reorder → keyboard reorder works; top/bottom buttons correctly disable.
- [ ] Submit a ranked ballot on a real (v1-synced) prod vote → succeeds end-to-end with toast confirmation.
- [ ] Submit a ranked ballot on a v2-only test vote → falls through to the SFDC 404 path; "Unable to submit vote" toast appears. (Same behavior as #710's Plurality path.)
- [ ] Toggle "Abstain from this vote" on a ranked poll with `allow_abstain=true` → drag list disables; submit goes through with `abstain: true`.
- [ ] Open a ranked vote from **Project / Foundation / Organization lens** OR from Me lens after responding → results drawer opens with the new ranked-choice rendering (per-choice rank bars + participation summary + round-summary placeholder when present). **No PCC link visible.**
- [ ] GENERIC (Plurality) votes still work exactly as in #710 — no regression in the existing radio/checkbox/abstain flows or in the GENERIC results display.

[LFXV2-1840]: https://linuxfoundation.atlassian.net/browse/LFXV2-1840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-1839]: https://linuxfoundation.atlassian.net/browse/LFXV2-1839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-1842]: https://linuxfoundation.atlassian.net/browse/LFXV2-1842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
@@ -54,14 +54,6 @@
             </div>
           }
         </div>
-      } @else if (!isGenericPoll()) {
-        <div class="flex flex-col items-center justify-center py-12 px-6 text-center" data-testid="vote-cast-drawer-unsupported">
-          <div class="w-16 h-16 rounded-full bg-blue-100 flex items-center justify-center mb-6">
-            <i class="fa-light fa-hourglass-half text-2xl text-blue-600"></i>
-          </div>
-          <h2 class="text-xl font-semibold text-slate-900 mb-3">Coming soon</h2>
-          <p class="text-sm text-slate-600 max-w-md">Ballot casting for ranked-choice voting methods is on the way and will be available here shortly.</p>
-        </div>
       } @else if (questions().length === 0) {
         <div class="text-center py-8 border border-slate-200 rounded-lg" data-testid="vote-cast-drawer-no-questions">
           <p class="text-sm text-gray-500">This vote has no questions to answer.</p>
@@ -78,13 +70,48 @@
             <div class="rounded-lg border border-slate-200 p-4 flex flex-col gap-3" [attr.data-testid]="'vote-cast-question-' + question.question_id">
               <div class="flex flex-col gap-1">
                 <h3 class="text-sm font-semibold text-slate-900">{{ question.prompt }}</h3>
-                @if (question.type === 'multiple_choice') {
+                @if (isRankedPoll()) {
+                  <p class="text-xs text-slate-500">Drag to reorder, or use the arrow buttons. Top of the list is your first choice.</p>
+                } @else if (question.type === 'multiple_choice') {
                   <p class="text-xs text-slate-500">Select all that apply.</p>
                 }
               </div>
 
               <div class="flex flex-col gap-2">
-                @if (question.type === 'single_choice') {
+                @if (isRankedPoll()) {
+                  <div cdkDropList (cdkDropListDropped)="onRankedDrop(question.question_id, $event)" class="flex flex-col gap-2">
+                    @for (choice of rankedOrderByQuestion().get(question.question_id) ?? question.choices; track choice.choice_id; let i = $index) {
+                      <div
+                        cdkDrag
+                        class="flex items-center gap-3 p-3 bg-white border border-slate-200 rounded-lg cursor-move hover:bg-slate-50"
+                        [attr.data-testid]="'vote-cast-rank-' + choice.choice_id">
+                        <i cdkDragHandle class="fa-light fa-grip-vertical text-slate-400" aria-hidden="true"></i>
+                        <span class="flex items-center justify-center w-7 h-7 rounded-full bg-blue-100 text-blue-700 text-sm font-semibold flex-shrink-0">
+                          {{ i + 1 }}
+                        </span>
+                        <span class="flex-1 text-sm text-slate-800">{{ choice.choice_text }}</span>
+                        <button
+                          type="button"
+                          (click)="moveRank(question.question_id, i, i - 1)"
+                          [disabled]="i === 0"
+                          class="p-1 text-slate-500 hover:text-slate-700 disabled:opacity-30"
+                          [attr.data-testid]="'vote-cast-rank-up-' + choice.choice_id"
+                          aria-label="Move up">
+                          <i class="fa-light fa-arrow-up" aria-hidden="true"></i>
+                        </button>
+                        <button
+                          type="button"
+                          (click)="moveRank(question.question_id, i, i + 1)"
+                          [disabled]="i === question.choices.length - 1"
+                          class="p-1 text-slate-500 hover:text-slate-700 disabled:opacity-30"
+                          [attr.data-testid]="'vote-cast-rank-down-' + choice.choice_id"
+                          aria-label="Move down">
+                          <i class="fa-light fa-arrow-down" aria-hidden="true"></i>
+                        </button>
+                      </div>
+                    }
+                  </div>
+                } @else if (question.type === 'single_choice') {
                   @for (choice of question.choices; track choice.choice_id) {
                     <lfx-radio-button
                       [form]="form"
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
@@ -79,11 +79,19 @@
 
               <div class="flex flex-col gap-2">
                 @if (isRankedPoll()) {
-                  <div cdkDropList (cdkDropListDropped)="onRankedDrop(question.question_id, $event)" class="flex flex-col gap-2">
+                  <div
+                    cdkDropList
+                    [cdkDropListDisabled]="abstain()"
+                    (cdkDropListDropped)="onRankedDrop(question.question_id, $event)"
+                    class="flex flex-col gap-2"
+                    [class.opacity-60]="abstain()">
                     @for (choice of rankedOrderByQuestion().get(question.question_id) ?? question.choices; track choice.choice_id; let i = $index) {
                       <div
                         cdkDrag
-                        class="flex items-center gap-3 p-3 bg-white border border-slate-200 rounded-lg cursor-move hover:bg-slate-50"
+                        [cdkDragDisabled]="abstain()"
+                        class="flex items-center gap-3 p-3 bg-white border border-slate-200 rounded-lg hover:bg-slate-50"
+                        [class.cursor-move]="!abstain()"
+                        [class.cursor-not-allowed]="abstain()"
                         [attr.data-testid]="'vote-cast-rank-' + choice.choice_id">
                         <i cdkDragHandle class="fa-light fa-grip-vertical text-slate-400" aria-hidden="true"></i>
                         <span class="flex items-center justify-center w-7 h-7 rounded-full bg-blue-100 text-blue-700 text-sm font-semibold flex-shrink-0">
@@ -93,19 +101,19 @@
                         <button
                           type="button"
                           (click)="moveRank(question.question_id, i, i - 1)"
-                          [disabled]="i === 0"
+                          [disabled]="i === 0 || abstain()"
                           class="p-1 text-slate-500 hover:text-slate-700 disabled:opacity-30"
                           [attr.data-testid]="'vote-cast-rank-up-' + choice.choice_id"
-                          aria-label="Move up">
+                          [attr.aria-label]="'Move ' + choice.choice_text + ' up'">
                           <i class="fa-light fa-arrow-up" aria-hidden="true"></i>
                         </button>
                         <button
                           type="button"
                           (click)="moveRank(question.question_id, i, i + 1)"
-                          [disabled]="i === question.choices.length - 1"
+                          [disabled]="i === question.choices.length - 1 || abstain()"
                           class="p-1 text-slate-500 hover:text-slate-700 disabled:opacity-30"
                           [attr.data-testid]="'vote-cast-rank-down-' + choice.choice_id"
-                          aria-label="Move down">
+                          [attr.aria-label]="'Move ' + choice.choice_text + ' down'">
                           <i class="fa-light fa-arrow-down" aria-hidden="true"></i>
                         </button>
                       </div>
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -77,7 +77,12 @@ export class VoteCastDrawerComponent {
   protected readonly abstain: Signal<boolean> = toSignal(this.abstainControl.valueChanges, { initialValue: this.abstainControl.value });
 
   // === Computed Signals ===
-  protected readonly isGenericPoll: Signal<boolean> = computed(() => this.vote()?.poll_type === PollType.GENERIC);
+  // Missing poll_type defaults to GENERIC (parity with vote-results-drawer.initIsGenericPoll) so the plurality branch renders when upstream omits the field.
+  protected readonly isGenericPoll: Signal<boolean> = computed(() => {
+    const v = this.vote();
+    if (!v) return false;
+    return (v.poll_type ?? PollType.GENERIC) === PollType.GENERIC;
+  });
   protected readonly isRankedPoll: Signal<boolean> = this.initIsRankedPoll();
   protected readonly questions: Signal<PollQuestion[]> = computed(() => this.vote()?.poll_questions ?? []);
   protected readonly allowAbstain: Signal<boolean> = computed(() => !!this.vote()?.allow_abstain);
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -117,7 +117,7 @@ export class VoteCastDrawerComponent {
     this.visible.set(false);
   }
 
-  /** Reorder via CDK drag-and-drop — setValue fires statusChanges so submitDisabled re-evaluates. */
+  /** Reorder via CDK drag-and-drop — bumpFormVersion() drives submitDisabled re-evaluation since rebuildForm/setValue here suppress statusChanges. */
   protected onRankedDrop(questionId: string, event: CdkDragDrop<string[]>): void {
     const control = this.form.get(questionId);
     if (!control) return;
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -66,6 +66,8 @@ export class VoteCastDrawerComponent {
   protected readonly submitting = signal<boolean>(false);
   // Reactive dependency for submitDisabled — rebuildForm uses { emitEvent: false }, so statusChanges is silent.
   private readonly formVersion = signal<number>(0);
+  // Tracks the poll_type the form was last built for; if it changes (e.g., list vote omitted poll_type and detail fetch supplied a ranked one), rebuildForm wipes controls so they get the right shape.
+  private lastBuiltPollType: PollType | null = null;
 
   // === Shared Observables ===
   private readonly voteId$ = toObservable(this.voteId).pipe(shareReplay({ bufferSize: 1, refCount: true }));
@@ -260,7 +262,15 @@ export class VoteCastDrawerComponent {
   // === Private Helpers ===
   /** Reconciles the form against the current vote's questions; ranked polls seed declared order so the form is valid on load. */
   private rebuildForm(questions: PollQuestion[]): void {
+    const currentPollType = this.vote()?.poll_type ?? PollType.GENERIC;
     const pollIsRanked = this.isRankedPoll();
+    // If poll_type flipped between builds (e.g., list vote omitted it → detail fetch supplied a ranked one), wipe all controls so they get re-added with the correct shape.
+    if (this.lastBuiltPollType !== null && this.lastBuiltPollType !== currentPollType) {
+      for (const existingId of Object.keys(this.form.controls)) {
+        this.form.removeControl(existingId, { emitEvent: false });
+      }
+    }
+    this.lastBuiltPollType = currentPollType;
     const desiredIds = new Set(questions.map((q) => q.question_id));
     for (const existingId of Object.keys(this.form.controls)) {
       if (!desiredIds.has(existingId)) this.form.removeControl(existingId, { emitEvent: false });
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -1,6 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
+import { CdkDragDrop, DragDropModule, moveItemInArray } from '@angular/cdk/drag-drop';
 import { DatePipe } from '@angular/common';
 import { HttpErrorResponse } from '@angular/common/http';
 import { Component, computed, DestroyRef, inject, input, model, output, signal, Signal } from '@angular/core';
@@ -10,7 +11,7 @@ import { ButtonComponent } from '@components/button/button.component';
 import { RadioButtonComponent } from '@components/radio-button/radio-button.component';
 import { TagComponent } from '@components/tag/tag.component';
 import { PollType } from '@lfx-one/shared';
-import { CreateVoteResponseRequest, PollQuestion, Vote, VoteAnswerInput } from '@lfx-one/shared/interfaces';
+import { CreateVoteResponseRequest, PollQuestion, UserChoice, Vote, VoteAnswerInput } from '@lfx-one/shared/interfaces';
 import { PollStatusLabelPipe } from '@pipes/poll-status-label.pipe';
 import { PollStatusSeverityPipe } from '@pipes/poll-status-severity.pipe';
 import { VoteService } from '@services/vote.service';
@@ -28,6 +29,7 @@ const INVITATION_NOT_FOUND = 'INVITATION_NOT_FOUND';
     DrawerModule,
     SkeletonModule,
     CheckboxModule,
+    DragDropModule,
     ReactiveFormsModule,
     ButtonComponent,
     RadioButtonComponent,
@@ -76,9 +78,12 @@ export class VoteCastDrawerComponent {
 
   // === Computed Signals ===
   protected readonly isGenericPoll: Signal<boolean> = computed(() => this.vote()?.poll_type === PollType.GENERIC);
+  protected readonly isRankedPoll: Signal<boolean> = this.initIsRankedPoll();
   protected readonly questions: Signal<PollQuestion[]> = computed(() => this.vote()?.poll_questions ?? []);
   protected readonly allowAbstain: Signal<boolean> = computed(() => !!this.vote()?.allow_abstain);
   protected readonly submitDisabled: Signal<boolean> = this.initSubmitDisabled();
+  // Map<question_id, ordered choices> — recomputed when questions or formVersion changes; replaces a per-render method call in @for.
+  protected readonly rankedOrderByQuestion: Signal<Map<string, UserChoice[]>> = this.initRankedOrderByQuestion();
 
   public constructor() {
     // Rebuild the form when the loaded vote's questions change.
@@ -110,6 +115,27 @@ export class VoteCastDrawerComponent {
   // === Protected Methods ===
   protected onClose(): void {
     this.visible.set(false);
+  }
+
+  /** Reorder via CDK drag-and-drop — setValue fires statusChanges so submitDisabled re-evaluates. */
+  protected onRankedDrop(questionId: string, event: CdkDragDrop<string[]>): void {
+    const control = this.form.get(questionId);
+    if (!control) return;
+    const current = [...((control.value as string[] | null) ?? [])];
+    moveItemInArray(current, event.previousIndex, event.currentIndex);
+    control.setValue(current);
+    this.bumpFormVersion();
+  }
+
+  /** Keyboard / arrow-button reorder fallback for the drag list. */
+  protected moveRank(questionId: string, fromIndex: number, toIndex: number): void {
+    const control = this.form.get(questionId);
+    if (!control) return;
+    const current = [...((control.value as string[] | null) ?? [])];
+    if (toIndex < 0 || toIndex >= current.length) return;
+    moveItemInArray(current, fromIndex, toIndex);
+    control.setValue(current);
+    this.bumpFormVersion();
   }
 
   protected onSubmit(): void {
@@ -187,6 +213,13 @@ export class VoteCastDrawerComponent {
     );
   }
 
+  private initIsRankedPoll(): Signal<boolean> {
+    return computed(() => {
+      const t = this.vote()?.poll_type;
+      return !!t && t !== PollType.GENERIC;
+    });
+  }
+
   private initSubmitDisabled(): Signal<boolean> {
     return computed(() => {
       this.formVersion(); // re-evaluate when controls are added/removed/disabled via { emitEvent: false }
@@ -196,16 +229,43 @@ export class VoteCastDrawerComponent {
     });
   }
 
+  private initRankedOrderByQuestion(): Signal<Map<string, UserChoice[]>> {
+    return computed(() => {
+      this.formVersion(); // re-evaluate when ranked-order setValue calls bumpFormVersion
+      const result = new Map<string, UserChoice[]>();
+      if (!this.isRankedPoll()) return result;
+      for (const question of this.questions()) {
+        const order = this.form.get(question.question_id)?.value as string[] | null | undefined;
+        if (!order?.length) {
+          result.set(question.question_id, question.choices);
+          continue;
+        }
+        const byId = new Map(question.choices.map((c) => [c.choice_id, c]));
+        result.set(
+          question.question_id,
+          order.map((id) => byId.get(id)).filter((c): c is UserChoice => !!c)
+        );
+      }
+      return result;
+    });
+  }
+
   // === Private Helpers ===
-  // Reconciles the form against the current vote's questions; bumps formVersion since we suppress emitEvent.
+  /** Reconciles the form against the current vote's questions; ranked polls seed declared order so the form is valid on load. */
   private rebuildForm(questions: PollQuestion[]): void {
+    const pollIsRanked = this.isRankedPoll();
     const desiredIds = new Set(questions.map((q) => q.question_id));
     for (const existingId of Object.keys(this.form.controls)) {
       if (!desiredIds.has(existingId)) this.form.removeControl(existingId, { emitEvent: false });
     }
     for (const question of questions) {
       if (this.form.contains(question.question_id)) continue;
-      if (question.type === 'multiple_choice') {
+      if (pollIsRanked) {
+        const seeded = question.choices.map((c) => c.choice_id);
+        this.form.addControl(question.question_id, new FormControl<string[]>(seeded, { nonNullable: true, validators: [Validators.required] }), {
+          emitEvent: false,
+        });
+      } else if (question.type === 'multiple_choice') {
         this.form.addControl(question.question_id, new FormControl<string[]>([], { nonNullable: true, validators: [Validators.required] }), {
           emitEvent: false,
         });
@@ -221,10 +281,21 @@ export class VoteCastDrawerComponent {
   }
 
   private buildAnswers(questions: PollQuestion[]): VoteAnswerInput[] {
-    return questions.map((q) => ({
-      question_id: q.question_id,
-      choice_ids: this.normaliseChoiceIds(this.form.get(q.question_id)?.value),
-    }));
+    const pollIsRanked = this.isRankedPoll();
+    return questions.map((q) => {
+      const raw = this.form.get(q.question_id)?.value as string | string[] | null | undefined;
+      if (pollIsRanked) {
+        const order = Array.isArray(raw) ? raw : [];
+        return {
+          question_id: q.question_id,
+          ranked_choices: order.map((choiceId, idx) => ({ choice_id: choiceId, choice_rank: idx + 1 })),
+        };
+      }
+      return {
+        question_id: q.question_id,
+        choice_ids: this.normaliseChoiceIds(raw),
+      };
+    });
   }
 
   private normaliseChoiceIds(raw: unknown): string[] {
--- a/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
@@ -240,7 +240,9 @@ export class VoteCastDrawerComponent {
       const result = new Map<string, UserChoice[]>();
       if (!this.isRankedPoll()) return result;
       for (const question of this.questions()) {
-        const order = this.form.get(question.question_id)?.value as string[] | null | undefined;
+        // Defensive Array.isArray — if a stale non-array control value lingers (e.g. mid-session vote reshape), fall back to declared order.
+        const raw = this.form.get(question.question_id)?.value;
+        const order = Array.isArray(raw) ? (raw as string[]) : null;
         if (!order?.length) {
           result.set(question.question_id, question.choices);
           continue;
--- a/apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
@@ -153,7 +153,7 @@
           @if (commentResults().length > 0) {
             <div class="flex flex-col gap-4 pt-4 border-t border-slate-200" data-testid="vote-results-comments">
               <h2 class="text-lg font-semibold text-slate-900">Comments</h2>
-              @for (commentResult of commentResults(); track commentResult.prompt) {
+              @for (commentResult of commentResults(); track $index) {
                 <div class="flex flex-col gap-2">
                   <h3 class="text-sm font-medium text-slate-700">{{ commentResult.prompt }}</h3>
                   @for (comment of commentResult.comments; track $index) {
@@ -249,7 +249,7 @@
           @if (commentResults().length > 0) {
             <div class="flex flex-col gap-4 pt-4 border-t border-slate-200" data-testid="vote-results-comments">
               <h2 class="text-lg font-semibold text-slate-900">Comments</h2>
-              @for (commentResult of commentResults(); track commentResult.prompt) {
+              @for (commentResult of commentResults(); track $index) {
                 <div class="flex flex-col gap-2">
                   <h3 class="text-sm font-medium text-slate-700">{{ commentResult.prompt }}</h3>
                   @for (comment of commentResult.comments; track $index) {
@@ -367,7 +367,7 @@
         @if (commentResults().length > 0) {
           <div class="flex flex-col gap-4 pt-4 border-t border-slate-200" data-testid="vote-results-comments">
             <h2 class="text-lg font-semibold text-slate-900">Comments</h2>
-            @for (commentResult of commentResults(); track commentResult.prompt) {
+            @for (commentResult of commentResults(); track $index) {
               <div class="flex flex-col gap-2">
                 <h3 class="text-sm font-medium text-slate-700">{{ commentResult.prompt }}</h3>
                 @for (comment of commentResult.comments; track $index) {
--- a/apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
@@ -79,32 +79,95 @@
           </div>
         </div>
       } @else if (!isGenericPoll()) {
-        <!-- Non-Generic Poll - Redirect to PCC -->
-        <div class="flex flex-col items-center justify-center py-12 px-6 text-center" data-testid="vote-results-pcc-redirect">
-          <div class="w-16 h-16 rounded-full bg-blue-100 flex items-center justify-center mb-6">
-            <i class="fa-light fa-arrow-up-right-from-square text-2xl text-blue-600"></i>
+        <!-- Ranked-choice results (replaces the old PCC redirect) -->
+        <div class="flex flex-col gap-6" data-testid="vote-results-ranked">
+          <!-- Voting method + algorithm banner -->
+          <div class="rounded-lg bg-slate-50 border border-slate-200 p-4 flex items-start gap-3" data-testid="vote-results-ranked-method">
+            <i class="fa-light fa-list-ol text-2xl text-blue-600 flex-shrink-0 mt-0.5"></i>
+            <div class="flex flex-col gap-1">
+              <h3 class="text-sm font-semibold text-slate-900">{{ votingMethodText() }}</h3>
+              <p class="text-xs text-slate-600">
+                Ranked-choice voting: voters ranked the choices in order of preference. The algorithm tallies first-preference votes and reallocates from
+                eliminated choices until a winner emerges.
+              </p>
+            </div>
           </div>
 
-          <h2 class="text-xl font-semibold text-slate-900 mb-3">Advanced Voting Method</h2>
+          <!-- Participation summary -->
+          <div class="p-4 bg-slate-50 rounded-lg border border-slate-200 flex flex-col gap-3" data-testid="vote-results-ranked-participation">
+            <div class="flex items-center justify-between text-sm text-slate-600">
+              <div class="text-center flex-1">
+                <p class="text-xs text-slate-500">Total Participants</p>
+                <p class="font-semibold text-slate-900 text-lg">{{ participationStats().eligibleVoters }}</p>
+              </div>
+              <div class="text-center flex-1">
+                <p class="text-xs text-slate-500">Total Responses</p>
+                <p class="font-semibold text-slate-900 text-lg">{{ participationStats().totalResponses }}</p>
+              </div>
+              <div class="text-center flex-1">
+                <p class="text-xs text-slate-500">Response Rate</p>
+                <p class="font-semibold text-blue-600 text-lg">{{ participationStats().participationRate }}%</p>
+              </div>
+            </div>
+            <div class="w-full bg-slate-200 rounded-full h-2 overflow-hidden">
+              <div class="h-full bg-blue-500 transition-all" [style.width.%]="participationStats().participationRate"></div>
+            </div>
+          </div>
 
-          <p class="text-sm text-slate-600 mb-6 max-w-md">
-            This vote uses <span class="font-medium">{{ votingMethodText() }}</span
-            >, an advanced voting method. Results for this vote type are available in the Project Control Center.
-          </p>
+          <!-- Per-question rank distribution -->
+          @if (rankedQuestions().length > 0) {
+            @for (q of rankedQuestions(); track q.questionId) {
+              <div class="rounded-lg border border-slate-200 p-4 flex flex-col gap-4" [attr.data-testid]="'vote-results-ranked-question-' + q.questionId">
+                <h4 class="text-sm font-semibold text-slate-900">{{ q.prompt }}</h4>
 
-          <a [href]="pccVotingUrl()" target="_blank" rel="noopener noreferrer" class="inline-block">
-            <lfx-button
-              label="View in Project Control Center"
-              icon="fa-light fa-arrow-up-right-from-square"
-              severity="primary"
-              data-testid="vote-results-pcc-link"></lfx-button>
-          </a>
+                <div class="flex flex-col gap-3">
+                  @for (row of q.choiceDistributions; track row.choiceId) {
+                    <div class="flex flex-col gap-2" [attr.data-testid]="'vote-results-ranked-choice-' + row.choiceId">
+                      <div class="flex items-center justify-between gap-2">
+                        <span class="text-sm font-medium text-slate-800">{{ row.choiceText }}</span>
+                        <span class="text-xs text-slate-500">{{ row.totalRanked }} {{ row.totalRanked === 1 ? 'voter' : 'voters' }}</span>
+                      </div>
 
-          <!-- Vote Details Section for non-generic polls -->
+                      @if (row.rankCounts.length > 0) {
+                        <div class="flex flex-col gap-1">
+                          @for (rc of row.rankCounts; track rc.rank) {
+                            <div class="flex items-center gap-2 text-xs">
+                              <span class="w-14 text-slate-600 flex-shrink-0">Rank {{ rc.rank }}</span>
+                              <div class="flex-1 bg-slate-100 rounded h-2 overflow-hidden">
+                                <div class="bg-blue-500 h-full rounded" [style.width.%]="rc.percentage"></div>
+                              </div>
+                              <span class="w-8 text-right text-slate-700 flex-shrink-0">{{ rc.count }}</span>
+                            </div>
+                          }
+                        </div>
+                      } @else {
+                        <p class="text-xs text-slate-500 italic">No rankings recorded yet.</p>
+                      }
+                    </div>
+                  }
+                </div>
+
+                <!-- Round summary placeholder (IRV / Meek STV) -->
+                @if (q.hasRoundSummary) {
+                  <div
+                    class="rounded-md border border-slate-200 bg-slate-50 p-3 flex items-center gap-2"
+                    [attr.data-testid]="'vote-results-ranked-rounds-' + q.questionId">
+                    <i class="fa-light fa-arrows-rotate text-slate-500"></i>
+                    <p class="text-xs text-slate-600">Round-by-round details available from the voting service. Detailed round visualization is coming soon.</p>
+                  </div>
+                }
+              </div>
+            }
+          } @else {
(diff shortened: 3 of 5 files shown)


apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
```diff
@@ -78,13 +70,48 @@ <h2 class="text-xl font-semibold text-slate-900 mb-3">Coming soon</h2>
...
+                } @else if (isMultipleChoice(question)) {
                   <p class="text-xs text-slate-500">Select all that apply.</p>
                 }
               </div>
 
               <div class="flex flex-col gap-2">
-                @if (isSingleChoice(question)) {
+                @if (isRankedPoll()) {
+                  <div cdkDropList (cdkDropListDropped)="onRankedDrop(question.question_id, $event)" class="flex flex-col gap-2">
+                    @for (choice of getRankedOrder(question); track choice.choice_id; let i = $index) {
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7e872b5f. The CDK drop list is now bound to `[cdkDropListDisabled]="abstain()"`, each draggable row gets `[cdkDragDisabled]="abstain()"`, and the up/down arrow buttons get an additional `|| abstain()` on their `[disabled]` expression. Added `[class.opacity-60]` on the list and `cursor-not-allowed` on each row when abstain is on so the visual state matches the disabled behavior. Good catch — the form.disable() path missed the CDK directives.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
```diff
@@ -78,13 +70,48 @@ <h2 class="text-xl font-semibold text-slate-900 mb-3">Coming soon</h2>
...
+                          {{ i + 1 }}
+                        </span>
+                        <span class="flex-1 text-sm text-slate-800">{{ choice.choice_text }}</span>
+                        <button
+                          type="button"
+                          (click)="moveRank(question.question_id, i, i - 1)"
+                          [disabled]="i === 0"
+                          class="p-1 text-slate-500 hover:text-slate-700 disabled:opacity-30"
+                          [attr.data-testid]="'vote-cast-rank-up-' + choice.choice_id"
+                          aria-label="Move up">
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7e872b5f. Switched both arrow buttons from a static `aria-label` to `[attr.aria-label]="'Move ' + choice.choice_text + ' up'"` (and `+ ' down'`). Screen-reader users now hear e.g. "Move Candidate Alpha up" instead of identical "Move up" labels across all rows. Thanks for the a11y catch.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,32 +79,95 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
...
+          <!-- Vote Details -->
           @if (voteData.description) {
-            <div class="w-full mt-8 pt-6 border-t border-slate-200 text-left" data-testid="vote-results-details">
-              <h3 class="text-base font-semibold text-slate-900 mb-2">Vote Details</h3>
-              <p class="text-sm text-slate-700 leading-relaxed"><span class="font-medium">Description:</span></p>
+            <div class="rounded-lg border border-slate-200 p-4 flex flex-col gap-2" data-testid="vote-results-details">
+              <h4 class="text-sm font-semibold text-slate-900">Vote Details</h4>
               <div class="vote-description text-sm text-slate-700 leading-relaxed" [innerHTML]="voteData.description"></div>
             </div>
           }
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7e872b5f. Added the same `commentResults()` section to the ranked-results branch — uses the same template (h2 + per-prompt h3 + comment list with `bg-slate-50 rounded-lg` styling) as the generic-live and final branches. Ranked polls with comment prompts now show voter comments alongside the rank distribution and round summaries.

manishdixitlfx:
## Review Feedback Addressed

Commit: 7e872b5f (rebased on the updated #710 branch — feat/LFXV2-1840 now includes the review fixes from #710's iteration)

### Changes Made
- **vote-cast-drawer.component.html**: bound CDK drop list + arrow buttons to `abstain()` so the ranked ballot is actually disabled when abstain is on (`[cdkDropListDisabled]`, `[cdkDragDisabled]`, arrow `[disabled]="... || abstain()"`). Added visual feedback (`opacity-60` + `cursor-not-allowed`). Previously `form.disable()` covered the FormControls but missed the CDK directives. (per @copilot-pull-request-reviewer)
- **vote-cast-drawer.component.html**: included choice text in arrow button `aria-label` — screen readers now hear `"Move Candidate Alpha up"` instead of identical "Move up" labels across rows. (per @copilot-pull-request-reviewer)
- **vote-results-drawer.component.html**: added `commentResults()` section to the ranked-results branch, matching the generic-live and final branches. Ranked polls with comment prompts now show voter comments instead of silently hiding them. (per @copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
```diff
@@ -16,9 +14,27 @@ import { DrawerModule } from 'primeng/drawer';
 import { SkeletonModule } from 'primeng/skeleton';
 import { catchError, finalize, of, shareReplay, startWith, switchMap } from 'rxjs';
 
+/**
+ * Local view model for ranked-choice question rendering. Derived from
+ * `PollQuestionResult` by `initRankedQuestions`. Kept local because it's purely
+ * a presentation-layer reshape; promote to `@lfx-one/shared/interfaces` if a
+ * second consumer ever needs it.
+ */
+interface RankedQuestionView {
```
dealako: [minor] `component-organization.md §5` requires all interfaces — including component-specific ones — to live in `@lfx-one/shared/interfaces`, not inline in the component file.

The JSDoc preempts this by noting "promote to `@lfx-one/shared/interfaces` if a second consumer ever needs it," but the rule doesn't hinge on a second consumer — it's about consistency across the codebase.

**Suggested fix:** Move `RankedQuestionView` (and its nested `choiceDistributions` / `rankCounts` shapes) to `packages/shared/src/interfaces/vote.ts` (or a new `vote-results.ts`) and import from `@lfx-one/shared/interfaces`.

dealako (dismissed):
Hi @manishdixitlfx — nice to see the My Votes ranked-choice flow round out behind the Plurality work in #710. The drag/keyboard fallback split is clean, the form bookkeeping (1-based `choice_rank` emission, seed-the-order on load, `submitDisabled` short-circuit on abstain) reads well, and the Copilot a11y/parity feedback from the first round is fully addressed in `7e872b5f`.

Three small items below — none blocking:

1. **[minor]** Move the local `RankedQuestionView` interface into `@lfx-one/shared/interfaces` — `component-organization.md §5` calls for component-specific interfaces to live there too.
2. **[minor]** Add `aria-hidden="true"` to the `cdkDragHandle` icon — the arrow buttons cover keyboard a11y, so the handle is decorative.
3. **[nit]** Confirm with PM whether the no-touch default (recording declared order when the user never reorders) is the desired UX, or whether partial-rank semantics should require at least one interaction before submit is enabled.

Otherwise everything looks solid: license headers, Tailwind layout patterns (`flex + flex-col + gap-*`), comprehensive `data-testid` coverage, `model()` for two-way binding, no SSR-affecting browser-only APIs, no `any` in new code, `Subject + takeUntil` submit cancellation. Branch metadata (title, branch name, JIRA, size, sign-off, GPG) all pass. No backend changes so no upstream contract validation needed.

**Approved with minor comments** — happy to re-approve once #1 and #2 land (or with a follow-up ticket if you'd rather defer them).

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
```diff
@@ -78,13 +70,56 @@ <h2 class="text-xl font-semibold text-slate-900 mb-3">Coming soon</h2>
...
+                    [class.opacity-60]="abstain()">
+                    @for (choice of getRankedOrder(question); track choice.choice_id; let i = $index) {
+                      <div
+                        cdkDrag
+                        [cdkDragDisabled]="abstain()"
+                        class="flex items-center gap-3 p-3 bg-white border border-slate-200 rounded-lg hover:bg-slate-50"
+                        [class.cursor-move]="!abstain()"
+                        [class.cursor-not-allowed]="abstain()"
+                        [attr.data-testid]="'vote-cast-rank-' + choice.choice_id">
+                        <i cdkDragHandle class="fa-light fa-grip-vertical text-slate-400"></i>
```
dealako: [minor] The drag handle is a decorative icon — the up/down arrow buttons (already correctly labelled with per-choice `aria-label`) provide the keyboard a11y path. Without `aria-hidden="true"`, screen readers may announce this empty `<i>` element as a focusable/interactive node.

**Suggested fix:**
```html
<i cdkDragHandle aria-hidden="true" class="fa-light fa-grip-vertical text-slate-400"></i>
```

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
```diff
@@ -78,13 +70,56 @@ <h2 class="text-xl font-semibold text-slate-900 mb-3">Coming soon</h2>
...
               <div class="flex flex-col gap-2">
-                @if (isSingleChoice(question)) {
+                @if (isRankedPoll()) {
+                  <div
+                    cdkDropList
+                    [cdkDropListDisabled]="abstain()"
+                    (cdkDropListDropped)="onRankedDrop(question.question_id, $event)"
+                    class="flex flex-col gap-2"
+                    [class.opacity-60]="abstain()">
+                    @for (choice of getRankedOrder(question); track choice.choice_id; let i = $index) {
```
MRashad26: **HIGH: Function call in `@for` iterable — re-runs on every change-detection pass.**

**Problem & Why:** `getRankedOrder(question)` is a component method called directly inside the `@for` iterable expression. This re-executes on every change-detection cycle for every question in the list — the same violation class as calling a method in `{{ }}` or `@if`. The rule is: no function/method calls anywhere in the template (`@if`, `@for` iterable, `@let`, interpolations, bindings).

**Suggested fix:** Pre-decorate each question in a `computed()` signal so the template uses a plain property:
```typescript
// Component-local type:
type DecoratedQuestion = PollQuestion & { rankedOrder: UserChoice[] };

// Replace the existing `vote()?.questions` reference with a computed:
protected readonly decoratedQuestions: Signal<DecoratedQuestion[]> = computed(() =>
  this.vote()?.questions?.map((q) => ({ ...q, rankedOrder: this.getRankedOrder(q) })) ?? []
);
```
Then in the template:
```html
@for (question of decoratedQuestions(); ...) {
  @for (choice of question.rankedOrder; track choice.choice_id; let i = $index) {
```
Keep `getRankedOrder()` as a private TS method — it is still called once during decoration.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -80,6 +85,10 @@ export class VoteCastDrawerComponent {
 
   // === Computed Signals ===
   protected readonly isGenericPoll: Signal<boolean> = computed(() => this.vote()?.poll_type === PollType.GENERIC);
+  protected readonly isRankedPoll: Signal<boolean> = computed(() => {
```
MRashad26: **Multi-line `computed()` must use a private `initXxx()` method.**

**Problem & Why:** `isRankedPoll` has a two-line body. The project convention is: any `computed()` with more than one line of body must be extracted to a `private initIsRankedPoll()` method. An inline multi-line computed makes the class harder to scan and can't be navigated to via symbol search.

**Suggested fix:**
```typescript
protected readonly isRankedPoll: Signal<boolean> = this.initIsRankedPoll();

private initIsRankedPoll(): Signal<boolean> {
  return computed(() => {
    const t = this.vote()?.poll_type;
    return !!t && t !== PollType.GENERIC;
  });
}
```

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -54,9 +56,12 @@ export class VoteCastDrawerComponent {
 
   // === Forms ===
   // Per-question form. Keys are question_id. Each value is:
-  //   - string | null for single_choice (selected choice_id)
-  //   - string[] | null for multiple_choice (array of selected choice_ids)
-  // Submit normalises both shapes into VoteAnswerInput.choice_ids.
+  //   - string | null for GENERIC + single_choice (selected choice_id)
```
MRashad26: **Multi-line comment block — one line max.**

**Problem & Why:** This 6-line block describes WHAT the form controls hold (WHAT), not WHY a non-obvious constraint exists. Multi-line comment blocks are explicitly forbidden regardless of content.

**Suggested fix:** Delete the block entirely. If the ranked-choice ordering convention needs documenting, add a single line on the `rebuildForm` method where the ordering is actually enforced.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -137,6 +146,36 @@ export class VoteCastDrawerComponent {
     return question.type === 'multiple_choice';
   }
 
+  // Reorder via CDK drag-and-drop. Mutates a copy and setValue's the new order
```
MRashad26: **Multi-line comment block — one line max.**

**Problem & Why:** Two-line comment on `onRankedDrop`. One-line maximum.

**Suggested fix:** Collapse to one line:
```typescript
// Mutates a copy of the rank array and calls setValue so statusChanges fires and submitDisabled re-evaluates.
protected onRankedDrop(...
```

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -137,6 +146,36 @@ export class VoteCastDrawerComponent {
...
+  protected moveRank(questionId: string, fromIndex: number, toIndex: number): void {
+    const control = this.form.get(questionId);
+    if (!control) return;
+    const current = [...((control.value as string[] | null) ?? [])];
+    if (toIndex < 0 || toIndex >= current.length) return;
+    moveItemInArray(current, fromIndex, toIndex);
+    control.setValue(current);
+  }
+
+  // Returns the question's choices in the order the form control currently holds them.
```
MRashad26: **Multi-line comment block — one line max.**

**Problem & Why:** Three-line comment on `getRankedOrder`. One-line maximum.

**Suggested fix:** Collapse to one line:
```typescript
// Returns choices in current form-control order; falls back to declared order before rebuildForm seeds the control.
private getRankedOrder(...
```

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -223,7 +262,11 @@ export class VoteCastDrawerComponent {
   // === Private Helpers ===
   // Reconciles the form against the current vote's questions: drops controls for
   // removed questions, adds controls for new ones, leaves stable controls alone.
+  // For ranked-choice polls, controls are SEEDED with the choices in their declared
```
MRashad26: **Multi-line comment block — one line max.**

**Problem & Why:** Three-line comment inside `rebuildForm`. One-line maximum.

**Suggested fix:** Collapse to one line:
```typescript
// Ranked controls are pre-seeded with declared order — form is valid on load; submission records order if user doesn't reorder.
```

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
```diff
@@ -16,9 +14,27 @@ import { DrawerModule } from 'primeng/drawer';
 import { SkeletonModule } from 'primeng/skeleton';
 import { catchError, finalize, of, shareReplay, startWith, switchMap } from 'rxjs';
 
+/**
```
MRashad26: **Multi-line JSDoc on interface — one line max. Also: interface must live in `@lfx-one/shared/interfaces`.**

**Problem & Why (JSDoc):** Five-line JSDoc on `RankedQuestionView` far exceeds the one-line maximum. The interface name and field names are self-documenting.

**Problem & Why (location):** The comment says 'Kept local because it's purely a presentation-layer reshape; promote to `@lfx-one/shared/interfaces` if a second consumer ever needs it.' The project rule is unconditional: ALL interfaces and types — including component-local view models — must be declared in `packages/shared/src/interfaces/`. There is no 'local until needed' exception.

**Suggested fix:** Move `RankedQuestionView` to `packages/shared/src/interfaces/` (e.g., `votes.interface.ts` or `components.interface.ts`), export it from there, import it in this component, and replace the 5-line JSDoc with a single line:
```typescript
/** Decorated view model for ranked-choice question rendering; pre-computes rank distributions. */
export interface RankedQuestionView {
```

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
```diff
@@ -16,9 +14,27 @@ import { DrawerModule } from 'primeng/drawer';
...
+ * second consumer ever needs it.
+ */
+interface RankedQuestionView {
+  questionId: string;
+  prompt: string;
+  choiceDistributions: {
+    choiceId: string;
+    choiceText: string;
+    totalRanked: number;
+    rankCounts: { rank: number; count: number; percentage: number }[];
```
MRashad26: **Anonymous inline object type — must be a named interface in `@lfx-one/shared/interfaces`.**

**Problem & Why:** `rankCounts: { rank: number; count: number; percentage: number }[]` uses an anonymous inline object type inside an interface field. All interfaces and types — including nested shapes within view models — must be named interfaces declared in `packages/shared/src/interfaces/`. Anonymous inline types are invisible to editor symbol search and prevent reuse.

**Suggested fix:** Extract to a named interface in the same shared file as `RankedQuestionView`:
```typescript
export interface RankCountEntry {
  rank: number;
  count: number;
  percentage: number;
}

// In RankedQuestionView:
rankCounts: RankCountEntry[];
```

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
```diff
@@ -196,6 +200,47 @@ export class VoteResultsDrawerComponent {
     });
   }
 
+  // Derives the ranked-choice per-question view model from raw poll_results.
```
MRashad26: **Multi-line comment block — one line max.**

**Problem & Why:** Five-line comment block above the body of `initRankedQuestions` explains WHAT the method does. One-line maximum when the WHY is non-obvious.

**Suggested fix:** Delete the block entirely — `initRankedQuestions` and its return type are self-documenting — or compress to one line if the deferred IRV round visualization deserves a note:
```typescript
// Round-by-round payload from IRV/Meek STV is currently untyped (any) — detailed visualization is deferred.
```

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
```diff
@@ -196,6 +200,47 @@ export class VoteResultsDrawerComponent {
...
+  // and flag whether the upstream returned a round-by-round summary (IRV / Meek STV)
+  // so the template can show a placeholder banner — the round payload itself is
+  // currently untyped (`any`) so detailed visualization is deferred.
+  private initRankedQuestions(): Signal<RankedQuestionView[]> {
+    return computed(() => {
+      const results = this.voteResults();
+      if (!results?.poll_results?.length) return [];
+
+      return results.poll_results.map((pr) => {
+        // Choice text resolution: prefer the winner_info candidate list (it carries
```
MRashad26: **Multi-line comment block — one line max.**

**Problem & Why:** Three-line inline comment inside `initRankedQuestions`. One-line maximum.

**Suggested fix:** Collapse to one line:
```typescript
// Prefer winner_info candidates for choice_text; fall back to question.choices defensively.
```

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,32 +79,114 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
           </div>
         </div>
       } @else if (!isGenericPoll()) {
-        <!-- Non-Generic Poll - Redirect to PCC -->
-        <div class="flex flex-col items-center justify-center py-12 px-6 text-center" data-testid="vote-results-pcc-redirect">
-          <div class="w-16 h-16 rounded-full bg-blue-100 flex items-center justify-center mb-6">
-            <i class="fa-light fa-arrow-up-right-from-square text-2xl text-blue-600"></i>
+        <!-- Ranked-choice results (replaces the old PCC redirect) -->
```
MRashad26: **Section-header HTML comments — remove all occurrences.**

**Problem & Why:** Comments that label WHAT a block does are not allowed. The surrounding markup (`data-testid`, ARIA roles, visible headings) already communicates section intent. Section-header divider comments add noise and drift out of date as the template evolves.

The following lines all contain section-header comments that must be deleted:
- Line 82: `<!-- Ranked-choice results (replaces the old PCC redirect) -->`
- Line 84: `<!-- Voting method + algorithm banner -->`
- Line 96: `<!-- Participation summary -->`
- Line 117: `<!-- Per-question rank distribution -->`
- Line 150: `<!-- Round summary placeholder (IRV / Meek STV) -->`
- Line 167: 2-line HTML comment block starting `<!-- Comments (parity with...`
- Line 186: `<!-- Vote Details -->`

**Suggested fix:** Delete all these comment lines/blocks entirely.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Moved `RankedQuestionView` (plus the now-named inner types `RankedChoiceDistribution` and `RankCountDistribution`) to `@lfx-one/shared/interfaces` (added to `poll.interface.ts`). The drawer now imports from `@lfx-one/shared/interfaces` and the local interface is gone.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Added `aria-hidden="true"` to the drag handle (`fa-grip-vertical`) and to the arrow icons inside the up/down buttons. The arrow buttons themselves keep their per-choice `aria-label="Move {choice_text} up/down"` as the accessible name.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Replaced the `getRankedOrder(question)` method with a `rankedOrderByQuestion: Signal<Map<string, UserChoice[]>>` computed signal (built via `initRankedOrderByQuestion()`). The template now reads `rankedOrderByQuestion().get(question.question_id) ?? question.choices` — pure signal access, no method invocation per change-detection pass. The Map is reactive on `formVersion` (bumped by `onRankedDrop` / `moveRank`) so the order stays in sync with `setValue` calls.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Extracted to `private initIsRankedPoll(): Signal<boolean>` and assigned via `protected readonly isRankedPoll = this.initIsRankedPoll();` — matches the `initVote()` / `initSubmitDisabled()` / `initRankedOrderByQuestion()` pattern used elsewhere in this component.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

The 6-line form-shape block is removed; the form is self-describing via `rebuildForm`'s branches (ranked → string[] seeded with declared order, multiple_choice → string[] starting `[]`, single_choice → string|null). The shape is now visible at the construction site.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Compressed to a single-line `/** Reorder via CDK drag-and-drop — setValue fires statusChanges so submitDisabled re-evaluates. */`.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

The `getRankedOrder()` method was removed entirely (replaced by the `rankedOrderByQuestion` computed Map signal — see the HIGH-severity thread on the template), so this comment block went with it.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Compressed `rebuildForm`'s comment to a single-line JSDoc: `/** Reconciles the form against the current vote's questions; ranked polls seed declared order so the form is valid on load. */`.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

`RankedQuestionView` moved to `@lfx-one/shared/interfaces` (in `poll.interface.ts`) with a single-line JSDoc: `/** Drawer-side view model for a ranked-choice question — projection of PollQuestionResult assembled by VoteResultsDrawerComponent.initRankedQuestions. */`. The drawer imports it directly.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Named both inner shapes in `@lfx-one/shared/interfaces` (poll.interface.ts):
- `RankedChoiceDistribution` for the per-choice row
- `RankCountDistribution` for the per-rank row inside it

`RankedQuestionView.choiceDistributions: RankedChoiceDistribution[]` and `RankedChoiceDistribution.rankCounts: RankCountDistribution[]` now use the named types.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Compressed to a single-line JSDoc: `/** Derives the ranked-choice per-question view model — round-summary payload is untyped upstream so detailed visualization is deferred to a placeholder banner. */`.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Compressed the 3-line inline comment to single-line: `// Choice text: prefer winner_info.poll_choices, fall back to question.choices.`.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
manishdixitlfx: Done — fixed in commit 5dc90e2b.

Stripped all section-header HTML comments (e.g. `<!-- Header Section -->`, `<!-- Title Row -->`, `<!-- Voting method + algorithm banner -->`, etc.) and the multi-paragraph `<!-- Comments (parity with generic-live …) -->` block from `vote-results-drawer.component.html`. The file now contains only the two license-header comments (lines 1-2). Removed ~1.8KB of comment-only lines.

manishdixitlfx:

Commit: 5dc90e2b (rebased on PR #710's updated head 72955d82)

### Rebase note
This branch is stacked on PR #710. I rebased it on top of the round-2 fixes there, so this PR now inherits all the CLAUDE.md compliance work from the parent (no `effect()`, FormControl-based abstain, `formVersion` signal for validity reactivity, removed `FormsModule` import, etc.). Two commits land in this round: the rebased ranked-choice feature commit (`6e029ff7`) and this round-2 review commit (`5dc90e2b`).

### Code Changes
- **`packages/shared/src/interfaces/poll.interface.ts`**: added `RankedQuestionView`, `RankedChoiceDistribution`, `RankCountDistribution` — all named interfaces with single-line JSDoc (per @dealako, @MRashad26 — shared placement + anonymous inline object naming)
- **`vote-cast-drawer.component.ts`**:
  - `isRankedPoll` extracted to `private initIsRankedPoll()` (per @MRashad26 — multi-line computed must use private init pattern)
  - `getRankedOrder()` method removed; replaced with `rankedOrderByQuestion: Signal<Map<string, UserChoice[]>>` built via `initRankedOrderByQuestion()` — Map signal recomputes on `formVersion` so it stays in sync with `setValue` calls (per @MRashad26 HIGH — function call in `@for` iterable)
  - `onRankedDrop` / `moveRank` now `bumpFormVersion()` after `setValue` so the Map signal re-evaluates
  - all flagged multi-line comments compressed to single-line JSDoc (per @MRashad26 — onRankedDrop, getRankedOrder removal, rebuildForm comment block, 6-line form-shape block)
- **`vote-cast-drawer.component.html`**:
  - `getRankedOrder(question)` call replaced with `rankedOrderByQuestion().get(question.question_id) ?? question.choices`
  - `aria-hidden="true"` added to drag handle (`fa-grip-vertical`) and to arrow icons inside up/down buttons (per @dealako, @MRashad26)
- **`vote-results-drawer.component.ts`**:
  - dropped local `RankedQuestionView` interface, imports from `@lfx-one/shared/interfaces` (per @dealako, @MRashad26)
  - 5-line comment above `initRankedQuestions` body compressed to single-line JSDoc
  - 3-line inline comment inside `initRankedQuestions` compressed to single line (per @MRashad26)
- **`vote-results-drawer.component.html`**: stripped all section-header HTML comments and the multi-paragraph "Comments (parity with …)" block. Only the two license-header comments (lines 1-2) remain. Removed ~1.8KB of comment-only lines (per @MRashad26)

### Threads Resolved

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -112,6 +117,27 @@ export class VoteCastDrawerComponent {
     this.visible.set(false);
   }
 
+  /** Reorder via CDK drag-and-drop — setValue fires statusChanges so submitDisabled re-evaluates. */
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit c0866899.

Updated the JSDoc on `onRankedDrop` to reference `bumpFormVersion()` as the actual driver of `submitDisabled` re-evaluation: `/** Reorder via CDK drag-and-drop — bumpFormVersion() drives submitDisabled re-evaluation since rebuildForm/setValue here suppress statusChanges. */`. Good catch — the previous wording would have invited a future change to remove the explicit `bumpFormVersion()` call on the wrong premise.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,40 +68,111 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
           </div>
         </div>
       } @else if (!isGenericPoll()) {
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit c0866899.

`initIsGenericPoll()` now defaults a missing `poll_type` to `GENERIC` so the template's `@if (isGenericPoll())` branch renders the plurality results instead of falling through to the ranked branch:

```ts
private initIsGenericPoll(): Signal<boolean> {
  return computed(() => {
    const v = this.vote();
    if (!v) return false;
    // Missing poll_type defaults to GENERIC (matches initVotingMethodText) so the plurality branch renders.
    return (v.poll_type ?? PollType.GENERIC) === PollType.GENERIC;
  });
}
```

Now consistent with `initVotingMethodText` which also defaults missing to `GENERIC`.

manishdixitlfx:

Commit: c0866899

Addresses 2 @copilot auto-reviewer threads:

- **`vote-cast-drawer.component.ts:120`** — corrected the JSDoc on `onRankedDrop` to reference `bumpFormVersion()` (the actual driver) instead of `statusChanges`. The previous wording would have invited a future change to remove the explicit version bump on the wrong premise.
- **`vote-results-drawer.component.ts`** — `initIsGenericPoll()` now defaults a missing `poll_type` to `GENERIC` (matching `initVotingMethodText`), so the plurality results branch renders when upstream omits `poll_type` instead of falling through to the ranked-results empty state.

2 of 2 unresolved threads resolved.

Note: this PR is stacked on PR #710. Once #710 lands, I'll rebase this branch onto main and pull in the corresponding `isGenericPoll` consistency fix for the cast drawer too.

manishdixitlfx:
## Rebased onto main after PR #710 merged

Base auto-retargeted from `feat/LFXV2-1840-ballot-casting-plurality` to `main`. The force-push invalidated the prior approval, so the PR is back in `REVIEW_REQUIRED`.

### What changed in this push
4 commits ahead of main (the 4 PR-#711-specific commits, replayed cleanly on top of #710's squash):

- `256d3aed` feat(voting): add ranked-choice cast UI and ranked results rendering
- `b1e06f93` fix(review): address PR #711 review feedback (round 1)
- `1c11ec5c` fix(review): address PR #711 review feedback (round 2)
- `fc74b55e` fix(review): address PR #711 review feedback (round 3)

Plus one new commit applying the parity fix promised in the round-3 summary:

- `b4bb8713` fix(voting): default missing poll_type to GENERIC in cast drawer — matches the `initIsGenericPoll` fix applied to `vote-results-drawer` so the cast drawer also renders the plurality branch when upstream omits `poll_type`

All 13 prior review threads remain resolved. No new diff vs. the pre-rebase tip beyond the cast-drawer parity fix.

@dealako @MRashad26 — could you re-approve when you have a moment? Both had approved earlier; the rebase + force-push reset that.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -196,16 +234,43 @@ export class VoteCastDrawerComponent {
...
+        }
+        const byId = new Map(question.choices.map((c) => [c.choice_id, c]));
+        result.set(
+          question.question_id,
+          order.map((id) => byId.get(id)).filter((c): c is UserChoice => !!c)
+        );
+      }
+      return result;
+    });
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 3fdee166 with a narrower fix than the suggested diff.

The scenario the elaborate `rebuildForm` rewrite guards against — same `question_id` colliding across two different polls with different control shapes — won't happen in practice since question_ids are UUIDs. So I kept `rebuildForm`'s simple skip-if-exists logic.

The defensive coercion at the *read* sites is still valuable insurance, though. `buildAnswers` already had `Array.isArray(raw) ? raw : []`. Added the same guard to `initRankedOrderByQuestion`:

```ts
const raw = this.form.get(question.question_id)?.value;
const order = Array.isArray(raw) ? (raw as string[]) : null;
if (!order?.length) {
  result.set(question.question_id, question.choices);
  continue;
}
```

If a stale non-array value ever does leak in (e.g. mid-session questions-array reshape on the same poll), the Map falls back to declared order and `buildAnswers` submits an empty rank list rather than crashing. Happy to take the fuller rebuildForm rewrite if you'd prefer.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,40 +68,111 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
...
+                                <div class="bg-blue-500 h-full rounded" [style.width.%]="rc.percentage"></div>
+                              </div>
+                              <span class="w-8 text-right text-slate-700 flex-shrink-0">{{ rc.count }}</span>
+                            </div>
+                          }
+                        </div>
+                      } @else {
+                        <p class="text-xs text-slate-500 italic">No rankings recorded yet.</p>
+                      }
+                    </div>
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 3fdee166.

Added an `@empty` branch to the per-question `@for (row of q.choiceDistributions; track row.choiceId)` loop:

```html
} @empty {
  <p class="text-sm text-gray-500 italic" [attr.data-testid]="'vote-results-ranked-question-empty-' + q.questionId">
    No ranked results recorded yet for this question.
  </p>
}
```

A ranked poll with zero responses now shows the question prompt plus the empty-state message inside the card, instead of an empty card. The outer `rankedQuestions().length === 0` empty state remains for the all-questions-empty case.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,40 +68,111 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
...
-          <div class="w-16 h-16 rounded-full bg-blue-100 flex items-center justify-center mb-6">
-            <i class="fa-light fa-arrow-up-right-from-square text-2xl text-blue-600"></i>
+        <div class="flex flex-col gap-6" data-testid="vote-results-ranked">
+          <div class="rounded-lg bg-slate-50 border border-slate-200 p-4 flex items-start gap-3" data-testid="vote-results-ranked-method">
+            <i class="fa-light fa-list-ol text-2xl text-blue-600 flex-shrink-0 mt-0.5"></i>
+            <div class="flex flex-col gap-1">
+              <h3 class="text-sm font-semibold text-slate-900">{{ votingMethodText() }}</h3>
+              <p class="text-xs text-slate-600">
+                Ranked-choice voting: voters ranked the choices in order of preference. The algorithm tallies first-preference votes and reallocates from
+                eliminated choices until a winner emerges.
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 3fdee166.

Replaced the IRV-specific copy with a `rankedMethodDescription` computed signal that branches on `poll_type`:

- **Condorcet IRV**: "Condorcet IRV first looks for a candidate who would beat every other candidate in head-to-head pairwise comparisons; if no such Condorcet winner exists, it falls back to instant-runoff elimination."
- **Instant Runoff Vote**: "Instant-runoff voting tallies first-preference votes and eliminates the lowest-ranked choice each round, reallocating those ballots until a candidate has a majority."
- **Meek STV**: "Meek STV is a multi-winner single-transferable-vote method that elects candidates as they meet a quota and proportionally redistributes surplus and eliminated ballots."
- **Default**: generic "The configured ranked-choice algorithm determines the winner (or winners)" fallback.

The template now binds to `{{ rankedMethodDescription() }}`.

manishdixitlfx:
## Review Feedback Addressed — Round 4

Commit: 3fdee166

Addresses 3 auto-reviewer threads:

- **`vote-cast-drawer.component.ts:243`** — added `Array.isArray` defensive check in `initRankedOrderByQuestion` so stale non-array control values fall back to declared order instead of breaking ranked rendering (per @coderabbitai). Applied narrower fix than the suggested diff — the question_id-collision scenario won't happen with UUIDs, so kept `rebuildForm`'s simple skip-if-exists logic and only hardened the read sites. `buildAnswers` already had the same guard.
- **`vote-results-drawer.component.ts`** — added `rankedMethodDescription` computed signal that returns method-accurate copy per `poll_type` (Condorcet IRV / IRV / Meek STV / default), replacing the IRV-specific text that rendered for all non-GENERIC types (per @copilot)
- **`vote-results-drawer.component.html`** — bound banner to `rankedMethodDescription()`; added `@empty` fallback inside the per-question `choiceDistributions` loop so zero-response ranked polls show "No ranked results recorded yet for this question" inside the card instead of a blank one (per @copilot)

3 of 3 unresolved threads resolved.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,40 +68,112 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
           </div>
         </div>
       } @else if (!isGenericPoll()) {
-        <!-- Non-Generic Poll - Redirect to PCC -->
-        <div class="flex flex-col items-center justify-center py-12 px-6 text-center" data-testid="vote-results-pcc-redirect">
-          <div class="w-16 h-16 rounded-full bg-blue-100 flex items-center justify-center mb-6">
-            <i class="fa-light fa-arrow-up-right-from-square text-2xl text-blue-600"></i>
+        <div class="flex flex-col gap-6" data-testid="vote-results-ranked">
+          <div class="rounded-lg bg-slate-50 border border-slate-200 p-4 flex items-start gap-3" data-testid="vote-results-ranked-method">
+            <i class="fa-light fa-list-ol text-2xl text-blue-600 flex-shrink-0 mt-0.5"></i>
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 3c677841.

Added `aria-hidden="true"` to both decorative icons:
- `<i class="fa-light fa-list-ol …" aria-hidden="true"></i>` on the ranked-method banner (line 73)
- `<i class="fa-light fa-arrows-rotate …" aria-hidden="true"></i>` inside the round-summary block (line 140)

Both are purely visual indicators — the text beside them carries the meaning, so excluding them from the accessibility tree removes screen-reader noise without losing any information. Consistent with the same `aria-hidden` pattern applied to the cast-drawer's drag handle and arrow icons in PR #711 round 1.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,40 +68,112 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
...
+                      @if (row.rankCounts.length > 0) {
+                        <div class="flex flex-col gap-1">
+                          @for (rc of row.rankCounts; track rc.rank) {
+                            <div class="flex items-center gap-2 text-xs">
+                              <span class="w-14 text-slate-600 flex-shrink-0">Rank {{ rc.rank }}</span>
+                              <div class="flex-1 bg-slate-100 rounded h-2 overflow-hidden">
+                                <div class="bg-blue-500 h-full rounded" [style.width.%]="rc.percentage"></div>
+                              </div>
+                              <span class="w-8 text-right text-slate-700 flex-shrink-0">{{ rc.count }}</span>
+                            </div>
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 3c677841.

Added a visible `{{ rc.percentage }}%` span next to the count column:

```html
<span class="w-8 text-right text-slate-700 flex-shrink-0">{{ rc.count }}</span>
<span class="w-12 text-right text-slate-600 flex-shrink-0">{{ rc.percentage }}%</span>
```

The percentage now reads aloud for screen-reader users and is precise to the integer for sighted users (the bar gives the at-a-glance shape, the text gives the exact value).

manishdixitlfx:
## Review Feedback Addressed — Round 5

Commit: 3c677841

Addresses 2 @coderabbitai a11y threads:

- **`vote-results-drawer.component.html`** — added `aria-hidden="true"` to the decorative banner icon (`fa-list-ol`) and round-summary icon (`fa-arrows-rotate`). Both are purely visual; text beside them carries the meaning, so excluding them from the accessibility tree removes screen-reader noise. Consistent with the same pattern applied to the cast-drawer's drag handle and arrow icons earlier in this PR.
- **`vote-results-drawer.component.html`** — added a visible `{{ rc.percentage }}%` span next to the count column on each rank distribution row. Percentage is now accessible to screen readers and precise to the integer for sighted users (bar gives shape, text gives value).

2 of 2 unresolved threads resolved.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.ts
```diff
@@ -196,16 +234,45 @@ export class VoteCastDrawerComponent {
...
       if (!desiredIds.has(existingId)) this.form.removeControl(existingId, { emitEvent: false });
     }
     for (const question of questions) {
       if (this.form.contains(question.question_id)) continue;
-      if (question.type === 'multiple_choice') {
+      if (pollIsRanked) {
+        const seeded = question.choices.map((c) => c.choice_id);
+        this.form.addControl(question.question_id, new FormControl<string[]>(seeded, { nonNullable: true, validators: [Validators.required] }), {
+          emitEvent: false,
+        });
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit a7418504.

Added a `lastBuiltPollType` field that tracks the `poll_type` the form was last reconciled against. If the value changes between `rebuildForm` invocations (e.g., list vote omitted `poll_type` → my `isGenericPoll` fallback defaulted it to GENERIC → detail fetch later supplied a ranked type), `rebuildForm` now wipes all existing controls before re-adding them, so they get rebuilt with the correct shape (ranked → seeded `string[]`, multiple_choice → `[]`, single_choice → `null`).

```ts
private lastBuiltPollType: PollType | null = null;

private rebuildForm(questions: PollQuestion[]): void {
  const currentPollType = this.vote()?.poll_type ?? PollType.GENERIC;
  const pollIsRanked = this.isRankedPoll();
  if (this.lastBuiltPollType !== null && this.lastBuiltPollType !== currentPollType) {
    for (const existingId of Object.keys(this.form.controls)) {
      this.form.removeControl(existingId, { emitEvent: false });
    }
  }
  this.lastBuiltPollType = currentPollType;
  // ... existing reconcile + add logic
}
```

Good catch — the defensive Array.isArray guards I added in round 4 would have prevented a runtime crash, but submit would still have produced empty `ranked_choices` for a poll that transitioned generic → ranked mid-session.

apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.html
```diff
@@ -79,40 +68,113 @@ <h1 class="text-lg sm:text-xl font-semibold text-slate-900 flex-1 min-w-0 trunca
...
-            <lfx-button
-              label="View in Project Control Center"
-              icon="fa-light fa-arrow-up-right-from-square"
-              severity="primary"
-              data-testid="vote-results-pcc-link"></lfx-button>
-          </a>
+          @if (commentResults().length > 0) {
+            <div class="flex flex-col gap-4 pt-4 border-t border-slate-200" data-testid="vote-results-comments">
+              <h2 class="text-lg font-semibold text-slate-900">Comments</h2>
+              @for (commentResult of commentResults(); track commentResult.prompt) {
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit a7418504.

Replaced `track commentResult.prompt` with `track $index` in all three `commentResults()` `@for` loops (lines 156, 252, 370 — generic-live, final-results, and ranked-results branches). You're right that `prompt` is user-authored and could collide. CLAUDE.md generally cautions against `$index` for reordering lists, but this list is read-only after results load (immutable for the drawer's lifetime), so `$index` is the appropriate fallback since `PollCommentResult` has no stable id field.

manishdixitlfx:
## Review Feedback Addressed — Round 6

Commit: a7418504 (rebased on top of the recent main-merge into the branch)

Addresses 2 auto-reviewer threads:

- **`vote-cast-drawer.component.ts`** — added `lastBuiltPollType` tracking. If `poll_type` changes between `rebuildForm` invocations (e.g., list vote omitted it and the detail fetch supplied a ranked one), the form wipes all controls first so they get rebuilt with the correct shape. Prevents stale generic-shape controls from leaking into a ranked render and producing empty `ranked_choices` on submit (per @copilot).
- **`vote-results-drawer.component.html`** — replaced `track commentResult.prompt` with `track $index` in all three `commentResults()` `@for` loops (generic-live, final-results, and ranked-results branches). `prompt` is user-authored and could collide; `PollCommentResult` has no stable id; the list is read-only after results load so `$index` is the appropriate fallback (per @coderabbitai).

2 of 2 unresolved threads resolved.

dealako:
## Follow-up review — round 6 (commit a7418504)

### 👏 Nice work

Six rounds of iteration since my initial pass, and the diff reads cleaner each time. A few specific things worth calling out:

- **`rankedOrderByQuestion` Map signal** replacing the per-render `getRankedOrder()` method — exactly the right pattern for keeping the template free of function calls in `@for`, and the `formVersion`-keyed reactivity keeps it in sync with `setValue`. Solid refactor under HIGH pressure.
- **Named sub-interfaces** (`RankedChoiceDistribution`, `RankCountDistribution`) rather than just hoisting `RankedQuestionView` — kept the shared package internally consistent instead of just satisfying the letter of the rule.
- **A11y polish across the whole flow** — per-choice arrow `aria-label`s, `aria-hidden` on every decorative icon (drag handle, banner, round-summary), visible `{{ rc.percentage }}%` next to the bar. The bar-for-shape / text-for-value split is the right call.
- **`rankedMethodDescription`** branching per poll_type — accurate copy for Condorcet IRV vs IRV vs Meek STV is a much better UX than the original IRV-only banner that rendered for all three.
- **`lastBuiltPollType` wipe-and-rebuild** — a thoughtful defensive fix for the list→detail poll_type flip scenario; well documented in the JSDoc.
- **`initIsGenericPoll` parity** between cast and results drawers — good catch on the consistency, both now default missing `poll_type` to `GENERIC`.

### ✅ Resolved

Confirmed at `a7418504`:

1. **`RankedQuestionView` moved to `@lfx-one/shared/interfaces`** — `packages/shared/src/interfaces/poll.interface.ts:695` with single-line JSDoc; `RankedChoiceDistribution` (`:687`) and `RankCountDistribution` (`:680`) are named alongside. `vote-results-drawer.component.ts` imports from `@lfx-one/shared/interfaces` and the local interface is gone. `component-organization.md §5` satisfied.
2. **`aria-hidden="true"` on `cdkDragHandle`** — confirmed in `vote-cast-drawer.component.html` on the `fa-grip-vertical` handle and on the arrow icons inside the up/down buttons. The buttons keep their per-choice accessible names.

### ⚠️ Partially addressed

3. **PM nit on no-touch default ranking** — Not a code item; the author kept declared-order-as-default and that's a reasonable no-touch convention for ranked-choice UIs. No action required here, but worth a confirm with PM before this reaches contributors who haven't seen the design — if partial-rank semantics are preferred, that'd be a follow-up ticket rather than a blocker for this PR.

### ❌ Still open

None from my original review.

### 🔍 New observations

Nothing blocking. Two small notes from re-reading the latest diff — both informational, neither requires action in this PR:

- **`lastBuiltPollType` wipe path** silently drops user-entered control values if `poll_type` legitimately flips mid-session. This is the right call (stale-shape controls would produce a wrong payload), but if a user has already started a ballot when the detail-fetch arrives with a different `poll_type`, their entries vanish without warning. The list→detail consistency we have today makes this realistically unreachable; flagging only because the JSDoc could note "no in-progress entries to lose at first build" so a future change doesn't extend this path into a UX hazard.
- **`track $index` on `commentResults` `@for`** — correct given `PollCommentResult` has no stable id and `prompt` can collide; just noting that this list is read-only post-results so the index-tracking trade-off (poor DOM identity on reorder) doesn't matter here.

### Overall decision

✅ **Re-approving.** Both items from my original review are fully resolved, the additional iteration from MRashad26 / Copilot / coderabbitai is comprehensively addressed, and no regressions. I'll submit a fresh formal approval review separately since the previous one was auto-dismissed by the rebase. The PM no-touch-default question is the only loose thread and it doesn't need to block merge.

Thanks for the patient iteration across six rounds — the code is in a noticeably better shape than where it started.

dealako (approved):
Good work. Approved.

There are a couple of new observations - none blocking. Worth considering.

PR #712, opened 2026-05-15
ci: declare workflow-level `contents: read` on 4 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 4 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.


dealako (commented):
Hi Arpit — thanks for the focused supply-chain hardening. The CVE citation and Scorecard rationale make this easy to follow, and after this PR every workflow in `.github/workflows/` carries an explicit permissions cap — a clean Scorecard win.

I verified both upstream reusable workflows (`lfx-public-workflows/license-header-check.yml` and `lfx-ui/_pr-title-lint.yml`) to check the actual API surface. Three of the four are clean: `license-header-check`, `markdown-lint`, and `quality-check` need nothing beyond `contents: read`. The fourth (`pr-title-lint`) is where the description is slightly off — see the inline comment above.

One process note: the commit shows as unsigned in GitHub's verification (`reason: "unsigned"`). Repo policy requires both `--signoff` (present ✅) and GPG signing (`-S`). A quick `git commit --amend --no-edit -S && git push --force-with-lease` before merge would bring it in line.

**Summary:**
- Blocking: 0
- Minor: 1 (pr-title-lint description accuracy / optional UX fix)
- Nits: 2 (missing EOF newline, unsigned commit)

**Decision: Fix the code signing - this is a requirement. Otherwise, minor comments** — happy to approve once issue 1 is acknowledged either way (description update or add `pull-requests: write` on that one file).

.github/workflows/pr-title-lint.yml
```diff
@@ -10,6 +10,9 @@ on:
     # Possible values: https://help.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-event-pull_request
     types: [opened, edited, reopened, synchronize]
 
+permissions:
```
dealako, line 13: [minor] The PR description says this workflow "doesn't call a GitHub API beyond the initial checkout," but the upstream reusable workflow (`linuxfoundation/lfx-ui/.github/workflows/_pr-title-lint.yml@main`) runs `morrisoncole/pr-lint-action@v1.7.1` with:

- `on-failed-regex-comment` — posts a PR comment on failure → needs `pull-requests: write`
- `on-succeeded-regex-dismiss-review-comment` — dismisses a review on success → needs `pull-requests: write`

With the caller restricted to `contents: read`, those two API calls will return 403. The lint check itself remains gating (`on-failed-regex-fail-action: true` still fails the action), but contributors with a bad PR title will see a red ❌ with no explanatory comment, and the success-side review dismissal becomes a no-op.

Two options:
- **Keep `contents: read` and update the PR description** to note this trade-off (and optionally raise a follow-up in the upstream workflow).
- **Add `pull-requests: write` on this file** — that still caps the token and satisfies OpenSSF Scorecard, while preserving the comment UX.

.github/workflows/pr-title-lint.yml
```diff
@@ -10,6 +10,9 @@ on:
...
+  contents: read
+
 jobs:
   pr-title-lint:
     uses: linuxfoundation/lfx-ui/.github/workflows/_pr-title-lint.yml@main
```
dealako, line 19: [nit] Missing trailing newline (`\ No newline at end of file` in the diff). Pre-existing condition, but easy to fix in the same commit since you're already touching this file.

dealako (dismissed):
Fix code signing.

arpitjain099:
> Fix code signing.

Hi @dealako thanks for the comment. I will fix this after I get back home tonight.

arpitjain099:
@dealako pushed a force-with-lease with the commit SSH-signed (now showing as Verified on GitHub). Rebased onto current main while I was at it. Let me know if you'd like me to handle the `pr-title-lint` description tweak or the missing EOF newline as a follow-up, or leave that for after merge.

dealako:
@arpitjain099 - the one change will break the current behavior. For this one entry, I think the correct solution is to assign write permissions [per this discussion](https://github.com/linuxfoundation/lfx-self-serve/pull/712/changes#r3249402889).

arpitjain099:
> @arpitjain099 - the one change will break the current behavior. For this one entry, I think the correct solution is to assign write permissions [per this discussion](https://github.com/linuxfoundation/lfx-self-serve/pull/712/changes#r3249402889). 

@dealako noted. let me review this tomorrow.

arpitjain099:
@dealako pushed - added `pull-requests: write` to `pr-title-lint.yml` at the workflow level so the reusable callee can post/dismiss PR comments. The other three workflows stay at `contents: read` only.

arpitjain099:
@dealako re-signed both commits with SSH (now Verified) and force-pushed. The PR is now: workflow-level `contents: read` on all 4 files, with `pull-requests: write` on `pr-title-lint.yml` per your earlier note. Let me know if you'd like anything else before merge.

audigregorie (approved):

jordane (approved):

PR #714, opened 2026-05-15
fix(server): resolve AiService dotenv race, drop placeholder defaults

fayazg:
## Summary

- `AiService` was instantiated by `MeetingController` at module import time, before `dotenv.config()` ran in `server.ts`. Constructor + readonly field initializers captured empty env values, and placeholder defaults (`https://api.openai.com/v1/chat/completions`, `sk-proj-1234567890`) masked the failure — requests fell through to OpenAI with a fake key and surfaced as confusing 401s.
- Replace eager env reads with memoized lazy getters (`_aiProxyUrl`, `_aiKey`), matching the pattern already used in `ProfileAuthService` and `CdpService`. Env is resolved on first request, after dotenv has loaded.
- Drop the placeholder defaults. Empty strings now flow through, and a private `assertConfigured()` helper throws an explicit `"AI service not configured"` error (used by both `generateMeetingAgenda` and `makeAiRequest`) when either `AI_PROXY_URL` or `AI_API_KEY` is missing. Also avoids shipping an OpenAI-key-shaped literal in a public repo.

**Client-visible behavior:** the configuration error is a plain `Error`, so the default `apiErrorHandler` branch surfaces it to API clients as a generic `500 INTERNAL_ERROR` — by design. Env-var details should not leak to end users; the specific message stays in server logs (CloudWatch) with full request context for ops to act on.

JIRA: LFXV2-1769

## Test plan

- [ ] Start the server with `AI_PROXY_URL` and `AI_API_KEY` set — `POST /api/meetings/generate-agenda` returns a generated agenda
- [ ] Start the server with `AI_PROXY_URL` and `AI_API_KEY` unset — `POST /api/meetings/generate-agenda` returns `500 INTERNAL_ERROR` to the client, and server logs show a clean `"AI service not configured: AI_PROXY_URL and AI_API_KEY environment variables are required"` error (no silent fall-through to api.openai.com)
- [ ] Start the server with only one of the two env vars set — same client response and same server-log error
- [ ] `yarn build` succeeds
- [ ] `yarn lint` clean


fayazg:
Addressed review feedback from CodeRabbit and Copilot:

- **CodeRabbit (ai.service.ts)**: hoisted the `isAiConfigured()` guard above the `try/catch` in `generateMeetingAgenda` so the specific `'AI service not configured: ...'` error propagates instead of being wrapped into the generic `'Failed to generate meeting agenda'`. The guard inside `makeAiRequest` stays as defense-in-depth.
- **Copilot (ai-service.md)**: softened the doc to match actual client-visible behavior. The configuration error is loud in server logs (specific message + request context for ops), but intentionally surfaces to API clients as a generic `500 INTERNAL_ERROR` via the default `apiErrorHandler` branch — by design, since env-var details should not leak to end users.

Verified locally with `yarn lint`, `yarn check-types`, and `yarn build`. Commit: 18145571.

fayazg:
Latest push (b44dec00):

- **CodeRabbit dedup nit**: extracted `assertConfigured()` private helper so the `'AI service not configured: ...'` string lives in one place. Both `generateMeetingAgenda` and `makeAiRequest` call it.
- **Copilot 'client-visible' suggestion**: not adopting. The intentional design is that env-var misconfig stays in server logs and surfaces as a generic `500 INTERNAL_ERROR` to clients — leaking internal config status to end users isn't user-actionable and is mild info-disclosure. Updated the PR description's test plan to make this contract explicit (it previously read 'returns a clean error referencing AI service not configured' without specifying server-log vs client response, which was the source of the confusion).

Verified locally: `yarn lint`, `yarn check-types`, `yarn build` all green.

dealako:
**[nit]** The PR description test-plan checkbox #2 reads:

> Start the server with `AI_PROXY_URL` and `AI_API_KEY` unset — `POST /api/meetings/generate-agenda` returns a clean error referencing "AI service not configured" (no silent fall-through to api.openai.com)

Per `error-handler.middleware.ts:84-88`, the client response will actually be the generic `{ "error": "Internal server error", "code": "INTERNAL_ERROR" }` — the "AI service not configured" detail only surfaces in server-side logs (CloudWatch). The updated `ai-service.md:54` already documents this correctly ("by design, since env-var details should not leak to end users").

Two bots independently flagged this same gap between the PR description and the actual middleware behavior, which is worth clarifying. Worth a quick edit to the test plan wording — e.g. _"server logs reference 'AI service not configured'"_ — so future manual verification matches what really happens.

Not blocking; the doc is the authoritative spec and it's accurate.

dealako (approved):
Hey Fayaz! Nice clean fix — the root-cause framing in the commit message is excellent. The dotenv-race combined with placeholder-masking was a particularly insidious failure mode (silent 401s landing on the real OpenAI endpoint with a fake key), and the lazy-getter approach resolves it cleanly while matching the established `CdpService` / `ProfileAuthService` pattern. The second commit hoisting the `isAiConfigured()` guard above the `try` block was exactly the right follow-up — the configuration error now propagates to logs rather than being swallowed by the generic catch.

**0 blocking, 1 nit** (PR description test-plan wording vs. actual client-visible response — see inline comment; doc is already correct).

✅ **Approved**

PR #716, opened 2026-05-15
feat(voting): inline ballot cast from dashboard pending actions

manishdixitlfx:
## Summary

Clicking **Cast Vote** on a Vote pending-action row in any of the 4 dashboards (user-dashboard, board-member, executive-director, multi-persona) now opens the existing `<lfx-vote-cast-drawer>` **inline** instead of navigating to `/votes` — mirroring the existing inline RSVP pattern in `PendingActionsComponent`.

JIRA: [LFXV2-1846](https://linuxfoundation.atlassian.net/browse/LFXV2-1846)

## Architecture

Option C (selected during research): a new **`<lfx-dashboard-cast-drawer-host>`** shell component owns the cast drawer + visibility + selectedVoteId state. All 4 dashboard hosts import this shell once and wire it to the new pending-actions output. Keeps the 4-host wiring DRY — a future 5th dashboard would only need ~5 lines of changes.

## Latent bug fixed in `vote-cast-drawer`

While testing this end-to-end I uncovered a bug introduced by PR #710 (the original plurality cast drawer): `submitDisabled` is a `computed()` that reads `formVersion()` + `form.valid`, but `bumpFormVersion` was only called on rebuild / abstain toggle / ranked reorder — **never on user input**. So picking a choice on `/votes` never enabled Submit Vote.

Fix: subscribe to `form.statusChanges` and bump `formVersion` on every emission. statusChanges fires on validity transitions from user clicks (it's only silent for our internal `{ emitEvent: false }` calls), so the computed now re-evaluates correctly. Fixes Submit Vote enabling on **both** the new pending-action launch AND the existing `/votes` page.

## Files changed (16 total)

**Shared types** (1):
- `packages/shared/src/interfaces/components.interface.ts` — `PendingActionItem.voteUid?`, `DecoratedPendingAction.isVoteInline`

**Server** (1):
- `apps/lfx-one/src/server/services/user.service.ts` — populate `voteUid` in `transformVotesToActions` (with `IndexedVote.vote_uid` fallback since `fetchPendingVotes` returns indexer-shaped docs typed as `Vote` at runtime)

**Shell component** (3 NEW):
- `apps/lfx-one/src/app/modules/dashboards/components/dashboard-cast-drawer-host/` (`.ts`, `.html`, `.scss`)

**Pending-actions** (2):
- `pending-actions.component.ts` — `castVoteRequested` output, `isVoteInline` decoration, `handleActionClick` Vote branch with optimistic hide
- `pending-actions.component.html` — new `@else if (item.isVoteInline)` branch (non-link button)

**Vote cast drawer** (1):
- `vote-cast-drawer.component.ts` — `form.statusChanges` subscription that bumps `formVersion`

**Dashboard hosts** (8 = 4 × 2):
- `user-dashboard`, `board-member`, `executive-director`, `multi-persona` — each `.ts` + `.html`. Each gains the shell import, two signals (`castDrawerVoteId`, `castDrawerVisible`), two handlers (`handleCastVoteRequested`, `handleVoteSubmitted`), the `(castVoteRequested)` binding on `<lfx-pending-actions>`, and the shell rendered outside main content.

## Test plan

End-to-end tested in Chrome (logged-in dev session) from **both entry points** — every variant verifies that Submit Vote correctly disabled until valid, then enables on input.

**From Pending Actions (Me dashboard) — opens drawer inline, no navigation:**
- [x] Multi-choice (no abstain) — checkbox click enables Submit
- [x] Multi-choice + abstain — checkbox OR abstain toggle independently enable Submit
- [x] Multi-question ballot (radio + multi) — partial fill stays disabled, both filled enables
- [x] Single-choice (no abstain) — radio click enables Submit
- [x] Single-choice + abstain — radio OR abstain toggle independently enable Submit
- [x] PCC multi-choice — checkbox enables Submit

**From `/votes` (existing entry point — confirms the statusChanges fix covers it too):**
- [x] CONDORCET_IRV ranked — enabled on open (pre-seeded order), stays enabled after arrow-button reorder
- [x] Multi-question ballot — partial fill stays disabled, both filled enables

## Out of scope (separate tickets)

- **LFXV2-1842** — `lfx-v2-voting-service` POST /vote_responses returns 404 for v2-native votes (backend)
- **LFXV2-1844** — v2 indexer doesn't flip `vote_status` to `submitted` after a successful cast (backend)
- Inline launch for **Survey** and **Agenda** pending actions (different flow)

## Preflight

✓ License headers · ✓ format · ✓ lint · ✓ build (17s) · ✓ no protected files · ✓ GPG + DCO signed · ✓ JIRA referenced

[LFXV2-1846]: https://linuxfoundation.atlassian.net/browse/LFXV2-1846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/src/app/modules/dashboards/components/pending-actions/pending-actions.component.ts
```diff
@@ -59,6 +61,14 @@ export class PendingActionsComponent {
       return;
     }
 
+    if (this.isVoteInline(item) && item.voteUid) {
+      // Vote rows: hide optimistically and emit voteUid upward so the parent dashboard opens the cast drawer inline.
+      this.hiddenActionsService.hideAction(item);
+      this.hiddenActionsVersion.update((v) => v + 1);
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 5ba30ac1.

Dropped the optimistic hide on drawer-open. `handleActionClick` now just emits `castVoteRequested` for Vote rows — no `hideAction` call. Cancel and error paths leave the row visible. Submit-success path goes through the parent's `handleVoteSubmitted` which calls `refresh$.next()`, so the row disappears via the natural data refresh.

Note: this exposes LFXV2-1844 (v2 indexer doesn't flip `vote_status` to `submitted` after a successful cast) — the action will still be in the refresh response until that backend bug is fixed. That's out of scope here but worth flagging.

apps/lfx-one/src/app/modules/dashboards/components/dashboard-cast-drawer-host/dashboard-cast-drawer-host.component.html
```diff
@@ -0,0 +1,4 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<lfx-vote-cast-drawer [voteId]="voteId()" [(visible)]="visible" (voteSubmitted)="voteSubmitted.emit($event)" data-testid="dashboard-cast-drawer-host" />
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 5ba30ac1.

Added an `@else if (!loadingVote())` branch to `vote-cast-drawer.component.html` that renders an error state when `vote()` is null after loading completes: triangle-exclamation icon, friendly message ("Couldn't load this vote… open the vote from the My Votes page"), and a Close button. Header also gets a close icon in this state.

Fix is in the drawer itself (not the shell) so it benefits both entry points — the latent silent-blank-drawer bug on `/votes` (when an old listVote is null and getVote fails) is also fixed.

apps/lfx-one/src/server/services/user.service.ts
```diff
@@ -1307,14 +1308,18 @@ export class UserService {
...
+      // fetchPendingVotes returns raw indexer-shaped docs typed as Vote — at runtime they carry vote_uid, not uid. Fall back so voteUid is always populated.
+      const voteUid = vote.uid ?? (vote as unknown as IndexedVote).vote_uid;
       return {
         type: 'Vote',
         badge,
         text: `Cast your vote on ${vote.name}`,
         icon: 'fa-regular fa-check-to-slot',
         severity: PENDING_ACTION_SEVERITY.Vote,
         buttonText: 'Cast Vote',
+        // buttonLink stays as a no-JS fallback; the dashboard prefers inline launch when voteUid is present.
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 5ba30ac1.

Reworded the comment in `user.service.ts:1320` to reflect what's actually happening: the template branches on `isVoteInline` and renders a non-link button when `voteUid` is set, so `buttonLink` isn't a no-JS fallback for inline rows. Kept the link for non-migrated callers and the older Survey/Agenda code path. Comment now reads:

```ts
// Inline launch is the active path (template branches on isVoteInline when voteUid is set); buttonLink kept for the older Survey/Agenda code path and any future caller that hasn't migrated.
buttonLink: '/votes',
```

apps/lfx-one/src/app/modules/dashboards/components/pending-actions/pending-actions.component.ts
```diff
@@ -103,6 +113,10 @@ export class PendingActionsComponent {
     return item.type === 'RSVP' && !!item.meetingUid;
   }
 
+  private isVoteInline(item: PendingActionItem): boolean {
+    return item.type === 'Vote' && !!item.voteUid;
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Done — fixed in commit 5ba30ac1.

`getRowKey` now branches on `voteUid` before falling back to the composite key:

```ts
private getRowKey(item: PendingActionItem): string {
  if (item.meetingUid) {
    return `${item.type}-${item.meetingUid}-${item.occurrenceId ?? ''}`;
  }
  if (item.voteUid) {
    return `${item.type}-${item.voteUid}`;
  }
  return `${item.type}-${item.text}-${item.buttonLink ?? ''}`;
}
```

Two pending votes with the same title and date now get distinct row identities. The hidden-actions cookie still uses `getActionIdentifier(item)` in `hidden-actions.service.ts` which is composite + buttonLink — collision-resistant enough for the Vote case because `voteUid` would only matter if two votes had the same `type+badge+text+buttonLink`, which is the same collision the rowKey fix eliminates upstream (since we no longer hide on open). Happy to also tighten `getActionIdentifier` to prefer `voteUid` if you'd like — let me know.

manishdixitlfx:
## Review Feedback Addressed

Commit: 5ba30ac1

Addresses 4 @copilot threads:

- **`pending-actions.component.ts`** — dropped the optimistic hide on drawer-open for Vote rows. Cancel/error paths now leave the row visible; the natural refresh path (parent's `handleVoteSubmitted` → `refresh$.next()`) handles the post-submit hide
- **`pending-actions.component.ts`** — `getRowKey` now branches on `voteUid` before falling back to the composite key, so two pending votes with the same title and date don't collide on row identity
- **`vote-cast-drawer.component.html`** — added an error state for when `vote()` is null after loading completes (triangle icon, friendly message, Close button). Fixes the silent-blank-drawer mode for the pending-actions entry point AND a latent bug on `/votes` when listVote is null and getVote fails
- **`user.service.ts`** — reworded the misleading "no-JS fallback" comment to reflect the actual purpose (kept for older Survey/Agenda paths and unmigrated callers)

4 of 4 threads resolved.

Note: thread 4 left a question for you about whether to also tighten `getActionIdentifier` in `hidden-actions.service.ts` to prefer `voteUid` over the composite key — currently safe given the on-open-hide removal, but I can do it for defense-in-depth if you prefer.

apps/lfx-one/src/server/services/user.service.ts
```diff
@@ -1307,14 +1308,18 @@ export class UserService {
         month: 'short',
         day: 'numeric',
       });
+      // fetchPendingVotes returns raw indexer-shaped docs typed as Vote — at runtime they carry vote_uid, not uid. Fall back so voteUid is always populated.
+      const voteUid = vote.uid ?? (vote as unknown as IndexedVote).vote_uid;
```
dealako, line 1312: **[minor] Type-laundering cast — consider fixing `fetchPendingVotes` return type upstream**

```ts
const voteUid = vote.uid ?? (vote as unknown as IndexedVote).vote_uid;
```

The cast works around a pre-existing type lie: `fetchPendingVotes()` (line 1092) is typed `Promise<Vote[]>` but at runtime returns `IndexedVote`-shaped docs from the `type: 'vote'` indexer query. The inline comment correctly explains the "why", so the next reader isn't surprised — but the cleanest long-term fix is changing the return type to `Promise<IndexedVote[]>`, which lets you drop the cast and write `vote.vote_uid` directly at the call site.

Pre-existing scope, not blocking this PR — worth a follow-up ticket on the vote indexer typing cleanup.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
```diff
@@ -172,5 +172,28 @@ <h3 class="text-sm font-semibold text-slate-900">{{ question.prompt }}</h3>
         </div>
       }
     </div>
+  } @else if (!loadingVote()) {
+    <ng-template #header>
```
dealako, line 176: **[nit] Two `<ng-template #header>` in mutually exclusive `@if`/`@else if` branches**

PrimeNG resolves `#header` via a `@ContentChildren` query. Having two `<ng-template #header>` elements inside `@if`/`@else if` blocks works in manual testing, but Angular's block-based control flow doesn't always re-run content queries when branches toggle — so this could become brittle on a future PrimeNG or Angular version bump.

Defensive alternative: hoist a single `<ng-template #header>` at the top of `<p-drawer>` and `@if` inside it on `vote()` vs error state — one stable query target regardless of branch. No action needed today; just flagging for awareness.

apps/lfx-one/src/app/modules/dashboards/components/dashboard-cast-drawer-host/dashboard-cast-drawer-host.component.scss
```diff
@@ -0,0 +1,2 @@
+// Copyright The Linux Foundation and each contributor to LFX.
```
dealako: **[nit] Empty SCSS file — consider dropping `styleUrl` until rules are needed**

This file contains only the license header. Per the Tailwind-first convention, SCSS is added only when needed (PrimeNG deep overrides, complex animations, pseudo-elements). Cleaner to remove the `styleUrl` reference in `dashboard-cast-drawer-host.component.ts` and delete this placeholder until a real style rule lands here. A minor nit, to be sure.

apps/lfx-one/src/app/modules/dashboards/multi-persona/multi-persona-dashboard.component.ts
```diff
@@ -35,12 +35,13 @@ import { BehaviorSubject, catchError, combineLatest, filter, map, of, switchMap,
 
 import { CardComponent } from '@components/card/card.component';
 import { TableComponent } from '@components/table/table.component';
+import { DashboardCastDrawerHostComponent } from '../components/dashboard-cast-drawer-host/dashboard-cast-drawer-host.component';
```
dealako, line 38: **[nit] Import ordering drift — `DashboardCastDrawerHostComponent` out of alphabetical order**

The new import lands between the `@components/...` block and the `./../components/...` block without a blank-line separator, and it's not alphabetically first among the relative imports. The three sibling dashboards (board-member, executive-director, user-dashboard) all place this same import as the first item in a blank-line-separated relative-import group:

```ts
import { DashboardCastDrawerHostComponent } from '../components/dashboard-cast-drawer-host/dashboard-cast-drawer-host.component';
import { MyMeetingsComponent } from '../components/my-meetings/my-meetings.component';
import { PendingActionsComponent } from '../components/pending-actions/pending-actions.component';
```

➡️ `yarn lint --fix` should normalize this automatically.

dealako (dismissed):
Hi Manish — solid follow-up round. All four Copilot threads from the first pass are correctly addressed: the optimistic hide is gone (`pending-actions.component.ts:64-68`), the drawer now has a proper error state with an in-drawer close (`vote-cast-drawer.component.html:175-198`), the `buttonLink` comment is accurate, and both `getRowKey` and `getActionIdentifier` use intrinsic IDs so two same-titled votes can't collide. The `form.statusChanges` fix to `formVersion` is a nice catch — and routing it through the drawer itself benefits the existing `/votes` page too, not just the new pending-action entry point.

Independent review: **0 blockers, 2 minor, 5 nit**.

**[minor] No Playwright regression for the inline-launch path (Finding 2)**
The feature spans four dashboards plus a new shared shell component and a fix to the cast drawer's submit-enabling logic. `data-testid` attributes are in all the right places (`dashboard-pending-actions-vote-button`, `dashboard-cast-drawer-host`, `vote-cast-drawer-error`, `vote-cast-drawer-error-close`), but no Playwright spec was added to exercise the new path. A small regression spec would prevent the previous `formVersion`/Submit-Vote-stuck-disabled bug from sneaking back in. Not blocking — but cheap insurance for a path that's already had one latent bug.

**[nit] Barrel import in `hidden-actions.service.ts:5` (Finding 6)**
```ts
import { PendingActionItem } from '@lfx-one/shared';
```
Every other touched file imports `PendingActionItem` from `@lfx-one/shared/interfaces` (e.g. `pending-actions.component.ts:17`). Single-line consistency fix per the subpath-entrypoint convention.

**[nit] PR body file count is stale (Finding 4)**
The body says "Files changed (16 total)" but the diff touches 18 files — `vote-cast-drawer.component.html` and `hidden-actions.service.ts` were updated during the review cycle but not backfilled into the listing.

Approving. The architecture (shell host for DRY dashboard wiring) is the right call, and the remaining items are cosmetic. See inline comments for 4 more nits.

MRashad26 (commented):
CLAUDE.md audit follow-up — 2 findings not already raised by Copilot/dealako. Both are about the dashboard host wiring rather than the drawer itself.

apps/lfx-one/src/app/modules/dashboards/components/dashboard-cast-drawer-host/dashboard-cast-drawer-host.component.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Component, input, model, output } from '@angular/core';
+import { VoteCastDrawerComponent } from '@app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component';
+
+@Component({
+  selector: 'lfx-dashboard-cast-drawer-host',
+  imports: [VoteCastDrawerComponent],
+  templateUrl: './dashboard-cast-drawer-host.component.html',
+  styleUrl: './dashboard-cast-drawer-host.component.scss',
+})
+export class DashboardCastDrawerHostComponent {
```
MRashad26, line 12: **Problem:** `DashboardCastDrawerHostComponent` is a no-op pass-through (17 lines TS + 1 HTML line + empty SCSS) that re-declares `voteId` / `visible` / `voteSubmitted` and forwards them verbatim to `<lfx-vote-cast-drawer>`. On top of that, each of the four dashboards (`board-member`, `executive-director`, `multi-persona`, `user-dashboard`) repeats the same boilerplate verbatim: a `castDrawerVoteId = signal<string | null>(null)`, a `castDrawerVisible = signal<boolean>(false)`, a `handleCastVoteRequested(voteUid)` that sets both, and a `handleVoteSubmitted()` that calls `this.refresh$.next()`.

**Why it is a problem:** The wrapper adds an extra layer of indirection without altering the contract — CLAUDE.md's "don't add abstractions beyond what the task requires" rule. Worse, the duplicated host state in four dashboard components means any future change to cast-drawer wiring (e.g. additional inputs, error handling, telemetry) has to be applied four times and any one of them can drift. This is exactly the "three similar lines is better than a premature abstraction, but four identical blocks is a missing one" inversion — the abstraction exists (the host component) but isn't actually encapsulating the duplicated state.

**Suggested fix:** Either (a) drop the host wrapper entirely and have each dashboard render `<lfx-vote-cast-drawer>` directly with shared local signals, or — better — (b) move the `castDrawerVoteId` / `castDrawerVisible` signals and the two handlers *into* `DashboardCastDrawerHostComponent` so it actually owns the state, and have the dashboards drop the four-line boilerplate and instead pass `(castVoteRequested)="host.open($event)"` to `<lfx-pending-actions>` with a `#host` template ref on the drawer host. That way the host earns its keep and the dashboards lose ~6 lines apiece.

apps/lfx-one/src/app/shared/services/hidden-actions.service.ts
```diff
@@ -54,14 +54,19 @@ export class HiddenActionsService {
...
    * Generate a unique identifier for an action.
-   * Always combines type, badge, and text — and appends buttonLink when present — so siblings
-   * that share a destination URL (e.g., multiple pending votes linking to /votes) still get
-   * distinct cookies. Using buttonLink alone would cause dismissing one to hide all siblings.
+   * Prefer intrinsic IDs when available so same-text siblings (for example multiple votes)
+   * never collide. Fall back to stable display fields for legacy action shapes.
    *
    * @param item The pending action item
    * @returns A unique string identifier
    */
```
MRashad26: **Problem:** Multi-paragraph JSDoc block on `getActionIdentifier` (lines 55–62) — description sentence, an explanatory sentence about same-text sibling collisions, then `@param` and `@returns` tags.

**Why it is a problem:** CLAUDE.md explicitly says "never write multi-paragraph docstrings or multi-line comment blocks — one short line max" and "don't explain WHAT the code does, since well-named identifiers already do that." The `@param`/`@returns` lines duplicate the TypeScript signature. The one piece of non-obvious WHY information — that intrinsic IDs are preferred so same-text siblings (multiple votes) don't collide — should survive, but the rest is noise.

**Suggested fix:** Collapse to a single-line comment that preserves only the non-obvious invariant, e.g.:
```ts
// Prefer intrinsic IDs so same-text siblings (e.g. multiple votes) never collide.
private getActionIdentifier(item: PendingActionItem): string {
```

apps/lfx-one/src/server/services/user.service.ts
manishdixitlfx: Thanks for the careful read — agreed it's a workaround for a pre-existing type lie. Leaving the cast in place for this PR (per your note that it's pre-existing scope, not blocking) and will file a follow-up JIRA to retype `fetchPendingVotes` as `Promise<IndexedVote[]>`, drop the cast in `transformVotesToActions`, and audit the other call sites of the same query for the same shape mismatch. Will link the ticket on this thread once it's filed.

apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html
manishdixitlfx: Good flag — agreed it's brittle even though it works today. Leaving as-is for this PR per your guidance, but I've added it to my mental queue: if a future Angular/PrimeNG version bump breaks the `@ContentChildren` resolution for branched `#header`s, the defensive hoist (one `<ng-template #header>` with `@if (vote()) {...} @else {...}` inside) is the planned fix.

apps/lfx-one/src/app/modules/dashboards/components/dashboard-cast-drawer-host/dashboard-cast-drawer-host.component.scss
manishdixitlfx: Done — fixed in commit 0f361f31.

Deleted the empty `.scss` file and dropped the `styleUrl` reference in `dashboard-cast-drawer-host.component.ts`. SCSS will land later only when a real style rule is needed.

apps/lfx-one/src/app/modules/dashboards/multi-persona/multi-persona-dashboard.component.ts
manishdixitlfx: Done — fixed in commit 0f361f31.

`yarn lint --fix` (which the repo's `yarn lint` runs by default) normalized the import grouping in `multi-persona-dashboard.component.ts` automatically. `DashboardCastDrawerHostComponent` is now ordered alphabetically with the other relative `../components/...` imports.

apps/lfx-one/src/app/modules/dashboards/components/dashboard-cast-drawer-host/dashboard-cast-drawer-host.component.ts
manishdixitlfx: Done — fixed in commit 0f361f31 with your suggestion (b).

The host now owns the state:
- `castDrawerVoteId` and `castDrawerVisible` moved INTO `DashboardCastDrawerHostComponent` as `protected` signals (no longer exposed via inputs).
- Added a public `open(voteUid: string)` method that sets both.
- Component template uses the signals directly.

Each of the 4 dashboards drops:
- The two `castDrawerVoteId` / `castDrawerVisible` signal declarations
- The `handleCastVoteRequested` method
- The `signal` import (where it was only used for those two)

Dashboards now just do:
```html
<lfx-pending-actions ... (castVoteRequested)="castDrawerHost.open($event)" />
...
<lfx-dashboard-cast-drawer-host #castDrawerHost (voteSubmitted)="handleVoteSubmitted()" />
```

Net diff: -58/+20 across the round, ~6 lines shaved per dashboard, single source of truth for cast-drawer state. The wrapper now actually encapsulates instead of pass-through.

apps/lfx-one/src/app/shared/services/hidden-actions.service.ts
manishdixitlfx: Done — partially in commit 0f361f31, and the rest had already landed in commit 5ee069b4 (`fix(dashboard): use intrinsic ids for hidden action keys`) from a parallel copilot-swe-agent run.

That earlier commit added the intrinsic-ID branches (`meetingUid`, `voteUid` → distinct cookie identifiers, matching the `getRowKey` change in the original PR) but left the JSDoc multi-paragraph. My commit on top compressed the JSDoc to a single-line comment per the CLAUDE.md rule:

```ts
// Prefer intrinsic IDs (meetingUid, voteUid) so same-text siblings never collide; fall back to type+badge+text+buttonLink for legacy action shapes without one.
private getActionIdentifier(item: PendingActionItem): string {
```

Both your concerns now addressed end-to-end: the function actually uses intrinsic IDs (no more sibling collision on the hidden-actions cookie) AND the comment is one line.

manishdixitlfx:
## Review Round 2 Addressed

Commit: 0f361f31 (rebased on 5ee069b4 + 75ec20b7 main-merge)

### Changes Made
- **`dashboard-cast-drawer-host`** — host now owns `castDrawerVoteId` + `castDrawerVisible` signals (protected) and exposes a public `open(voteUid)` method. Was a pass-through wrapper; now actually encapsulates the duplicated state (per @MRashad26)
- **4 dashboards** drop the ~6-line boilerplate each (signal declarations + handler) and just use `#castDrawerHost` template ref → `host.open($event)` (per @MRashad26)
- **`dashboard-cast-drawer-host.component.scss`** — deleted; `styleUrl` dropped (per @dealako)
- **`multi-persona-dashboard.component.ts`** — `yarn lint --fix` normalized import order (per @dealako)
- **`hidden-actions.service.ts`** — compressed multi-paragraph JSDoc on `getActionIdentifier` to single-line comment. Note: a parallel copilot-swe-agent commit `5ee069b4` had already added the intrinsic-ID branches (matching the `getRowKey` change), so the function now uses `meetingUid`/`voteUid` for cookie identity end-to-end (per @MRashad26)

### Acknowledged, No Code Change
- **`user.service.ts:1312` type-laundering cast** (@dealako) — pre-existing scope per your note; will file a follow-up JIRA to retype `fetchPendingVotes` as `Promise<IndexedVote[]>` and drop the cast
- **two `<ng-template #header>` in `@if`/`@else if`** (@vote-cast-drawer.component.html) (@dealako) — flagged for awareness; defensive hoist planned only if a future Angular/PrimeNG upgrade breaks `@ContentChildren` resolution

### Threads Resolved
6 of 6 addressed (4 with code, 2 acknowledged).

MRashad26 (approved):

PR #717, opened 2026-05-16
fix(nav): default project-selector to Projects tab on Project lens

manishdixitlfx:
## Summary

For hybrid personas (users with both board and project roles), the Project lens button serves as a merged entry that shows both Foundations and Projects in the picker. The active tab defaulted to **"All"**, which sorts foundations first per the load-priority comment — so opening the picker on Project lens led with a Foundation row (e.g. AAIF tagged "Executive Director") even though the user just clicked Project.

## Fix

Compute a `defaultTab` signal that returns `'projects'` when `lens === 'project'`, `'all'` otherwise. Set `activeTab` to `defaultTab()` on both panel show and panel hide:

```ts
private readonly defaultTab: Signal<SelectorTab> = computed(() =>
  this.lens() === 'project' ? 'projects' : 'all'
);

protected onPopoverShow(): void {
  this.isPanelOpen.set(true);
  this.activeTab.set(this.defaultTab());
}

protected onPopoverHide(): void {
  this.isPanelOpen.set(false);
  this.activeTab.set(this.defaultTab());
  ...
}
```

Tabs only render in hybrid mode (`@if (hybridMode())`), so non-hybrid users (project-only or board-only personas) see no UX change.

## Test plan

Verified in Chrome on `localhost:4200` with a hybrid persona (ED + Board Member + Contributor):

- [x] Click Project lens → URL updates to `/project/overview?project=...`
- [x] Click the project-selector chip → picker opens
- [x] Active tab is **Projects** (not All)
- [x] First result is a Project, not a Foundation
- [x] Tabs still let user switch to "Foundations" or "All" inside the picker
- [x] yarn format / yarn lint / yarn build all clean

## Out of scope

The deeper question of whether `setLens('project')` should also auto-default `currentPersona` to a project-scoped persona (`maintainer` → `contributor`) is intentionally left out — that's a separate behavior change in `LensService` / `PersonaService` and warrants its own discussion. The constants `PROJECT_SCOPED_PERSONA_PRIORITY` already exist for that follow-up.

### DIFF
--- a/apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.ts
+++ b/apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.ts
@@ -40,6 +40,8 @@ export class ProjectSelectorComponent {
 
   protected readonly activeTab = signal<SelectorTab>('all');
   protected readonly selectorTabs: readonly SelectorTab[] = ['all', 'foundations', 'projects'];
+  // Project lens defaults the picker to its matching tab so projects rank first; foundation lens uses 'all' (the only other hybrid case, where users explicitly went to a board context).
+  private readonly defaultTab: Signal<SelectorTab> = computed(() => (this.lens() === 'project' ? 'projects' : 'all'));
   protected readonly searchControl = new FormControl<string>('', { nonNullable: true });
 
   protected readonly panelStyleClass: Signal<string> = this.initPanelStyleClass();
@@ -86,11 +88,12 @@ export class ProjectSelectorComponent {
 
   protected onPopoverShow(): void {
     this.isPanelOpen.set(true);
+    this.activeTab.set(this.defaultTab());
   }
 
   protected onPopoverHide(): void {
     this.isPanelOpen.set(false);
-    this.activeTab.set('all');
+    this.activeTab.set(this.defaultTab());
     this.searchControl.setValue('', { emitEvent: false });
     if (this.hybridMode()) {
       this.navigationService.setSearchTerm('foundation', '');
--- a/apps/lfx-one/src/app/shared/services/navigation.service.ts
+++ b/apps/lfx-one/src/app/shared/services/navigation.service.ts
@@ -5,7 +5,13 @@ import { HttpClient, HttpParams } from '@angular/common/http';
 import { computed, inject, Injectable, Signal, signal, WritableSignal } from '@angular/core';
 import { toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { Router } from '@angular/router';
-import { BOARD_SCOPED_PERSONA_PRIORITY, LENS_DEFAULT_ROUTES, NAV_SEARCH_DEBOUNCE_MS, PROJECT_SCOPED_PERSONA_PRIORITY } from '@lfx-one/shared/constants';
+import {
+  BOARD_SCOPED_PERSONA_PRIORITY,
+  LENS_DEFAULT_ROUTES,
+  NAV_SEARCH_DEBOUNCE_MS,
+  PERSONA_PRIORITY,
+  PROJECT_SCOPED_PERSONA_PRIORITY,
+} from '@lfx-one/shared/constants';
 import { LensItem, LensItemsResponse, LensPage, LensState, NavLens, PersonaType, TaggedLensPage } from '@lfx-one/shared/interfaces';
 import { lensItemToProjectContext } from '@lfx-one/shared/utils';
 import { MessageService } from 'primeng/api';
@@ -163,7 +169,7 @@ export class NavigationService {
       return;
     }
 
-    const priority = lens === 'foundation' ? BOARD_SCOPED_PERSONA_PRIORITY : PROJECT_SCOPED_PERSONA_PRIORITY;
+    const priority = this.getPickerPriority(lens);
     const defaultItem = this.pickItemByPersonaPriority(page.items, priority);
     const context = lensItemToProjectContext(defaultItem);
     if (lens === 'foundation') {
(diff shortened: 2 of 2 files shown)


apps/lfx-one/src/app/shared/services/navigation.service.ts
```diff
@@ -173,22 +179,40 @@ export class NavigationService {
...
+        const parentUid = enrichedByUid.get(item.uid)?.parentProjectUid;
+        const isRoot = !parentUid || !itemUids.has(parentUid);
+        matches.push({ item, isRoot });
       }
+      if (matches.length === 0) continue;
+      // Prefer a root-level match (no in-set parent) so e.g. The Linux Foundation beats AAIF when both carry board-member.
+      return (matches.find((m) => m.isRoot) ?? matches[0]).item;
     }
     return items[0];
   }
```
dealako, line 211: **[minor] Clarification: does the TLF-beats-AAIF root-most scenario actually reach this code for hybrid project-lens users?**

The commit message (and the comment at line 207) describe preferring The Linux Foundation over AAIF when both carry `board-member`, because AAIF's parent is TLF (in the user's set) and TLF's parent is the org meta UID (outside the set).

But `applyVisibilityFilters` (lines 391-396) strips `isFoundation` items from project-lens results whenever `availableLenses` contains `'foundation'` — which it does for every hybrid user, because `LensService.getAllowedLensIds()` adds `'foundation'` whenever `hasBoardRole()` is true (`lens.service.ts:73-85`). This was an explicit choice in PR #713 (`fix(nav): restore availableLenses to full allowed set … keep filtering foundations out of the project lens for hybrid users`).

So for the hybrid project-lens case `pickItemByPersonaPriority` sees a foundation-free `items` array: TLF is never present, and the described tiebreak between TLF and AAIF can't materialise on this lens. The core fix — board-member matches rank ahead of contributor matches — is still meaningful and correct. But the root-most preference seems to apply on the **foundation lens** (where items aren't filtered) and to non-hybrid project-only users, not the scenario described.

Could you confirm whether I'm reading the filter interaction correctly, or point to where TLF does appear in the project-lens item set for hybrid users? Happy to be corrected if there's a data path I'm missing.

apps/lfx-one/src/app/shared/services/navigation.service.ts
```diff
@@ -173,22 +179,40 @@ export class NavigationService {
...
-        const match = items.find((item) => item.uid === project.projectUid);
-        if (match) return match;
+        const item = items.find((i) => i.uid === project.projectUid);
+        if (!item) continue;
+        const parentUid = enrichedByUid.get(item.uid)?.parentProjectUid;
+        const isRoot = !parentUid || !itemUids.has(parentUid);
+        matches.push({ item, isRoot });
       }
+      if (matches.length === 0) continue;
+      // Prefer a root-level match (no in-set parent) so e.g. The Linux Foundation beats AAIF when both carry board-member.
```
dealako, line 207: **[minor] Comment wording: "item set" implies the user's full accessible set, not the current page**

`itemUids` is built from `items.map(i => i.uid)` — the items in the current page only. In a paginated context an entry could appear root-most simply because its parent hasn't been loaded yet. Suggest: "parent outside the **current page's** loaded items" (or similar) to make the scope clear.

apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.ts
```diff
@@ -40,6 +40,8 @@ export class ProjectSelectorComponent {
 
   protected readonly activeTab = signal<SelectorTab>('all');
   protected readonly selectorTabs: readonly SelectorTab[] = ['all', 'foundations', 'projects'];
+  // Project lens defaults the picker to its matching tab so projects rank first; foundation lens uses 'all' (the only other hybrid case, where users explicitly went to a board context).
+  private readonly defaultTab: Signal<SelectorTab> = computed(() => (this.lens() === 'project' ? 'projects' : 'all'));
```
dealako, line 44: **[nit] Visibility grouping**

`private readonly defaultTab` sits between the `protected readonly` fields above and below it. Per the component-organization rule the structure order groups fields by visibility. Cosmetic — easy to move if you're already in the file.

apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.ts
```diff
@@ -40,6 +40,8 @@ export class ProjectSelectorComponent {
 
   protected readonly activeTab = signal<SelectorTab>('all');
   protected readonly selectorTabs: readonly SelectorTab[] = ['all', 'foundations', 'projects'];
+  // Project lens defaults the picker to its matching tab so projects rank first; foundation lens uses 'all' (the only other hybrid case, where users explicitly went to a board context).
```
dealako, line 43: **[nit] Long inline comment (~180 chars)**

The parenthetical `// … the only other hybrid case, where users explicitly went to a board context` pushes this past comfortable reading width. Could trim or wrap — non-blocking.

dealako (approved):
Hey Manish! Nice, focused PR — two well-scoped fixes that tighten up the hybrid-persona experience without disturbing non-hybrid or foundation-only users. The commit 1 tab-default change is clean and the rationale is obvious. Commit 2 is the meatier one and overall correct, though I have one clarification question on its stated behaviour before you land it.

**0 blocking · 3 minor · 2 nit**

**Decision: Approved with minor comments** — the clarification in commit 2 is worth confirming before merge, but nothing rises to blocking.

---

**AI bot reconciliation**
- **CodeRabbit**: PR is in draft, skipped — no findings to reconcile.
- **Copilot**: summarised commit 1's tab-default change only; did not surface commit 2 or any findings. My comments extend coverage to commit 2 with no conflicts.
- **Cursor Bot**: no comments on this PR.

**Revision tracking**: first human review round.

---

**[minor] PR description / test plan**

The title and description cover commit 1 (tab default) only. Commit 2 makes a substantive change to the default-selection algorithm — `PROJECT_SCOPED_PERSONA_PRIORITY` → full `PERSONA_PRIORITY` for hybrid + root-most tiebreaker — with no corresponding test-plan entry. Recommend adding a bullet like: *"Hybrid user with board roles on multiple items and contributor activity elsewhere — verify default selection lands on a board-role project, not the top contributor result."*

PR #718, opened 2026-05-16
feat(dashboards): add marketing attribution section and tab

mrautela365:
## Summary
- Adds attribution section component with model selector (Linear/First Touch/Last Touch/Time Decay)
- Channel revenue table with horizontal share bars
- Wires attribution as both a section in Overview tab and standalone Attribution tab

**Depends on:** #705

## Test plan
- [ ] Select TLF, verify Attribution tab shows channel table
- [ ] Switch attribution models in dropdown
- [ ] Verify Overview tab still shows attribution section below KPI cards

LFXV2-1644

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
@@ -1,0 +1,88 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<div class="flex flex-col gap-4 rounded-lg border border-gray-200 bg-white p-5" data-testid="attribution-section">
+  <!-- Header -->
+  <div class="flex flex-col gap-1 md:flex-row md:items-center md:justify-between">
+    <div class="flex flex-col gap-0.5">
+      <h3 class="text-base font-semibold text-gray-900">Marketing attribution</h3>
+      <p class="text-xs text-gray-500">Revenue by channel · {{ foundationName() || 'All foundations' }}</p>
+    </div>
+    <div class="flex items-center gap-3">
+      <lfx-select
+        [form]="modelForm"
+        control="model"
+        [options]="modelOptions"
+        optionLabel="label"
+        optionValue="value"
+        placeholder="Select model"
+        ariaLabel="Select attribution model"
+        styleClass="w-[150px]"
+        data-testid="attribution-model-select" />
+      <lfx-button
+        label="Open detail"
+        icon="fa-light fa-arrow-up-right-from-square"
+        severity="secondary"
+        [outlined]="true"
+        [disabled]="true"
+        ariaLabel="Open attribution detail (coming soon)"
+        data-testid="attribution-open-detail-btn" />
+    </div>
+  </div>
+
+  <!-- Alert banner -->
+  <div class="flex items-start gap-2 rounded-md border border-amber-200 bg-amber-50 px-3 py-2" data-testid="attribution-alert">
+    <i class="fa-light fa-triangle-exclamation mt-0.5 text-sm text-amber-600" aria-hidden="true"></i>
+    <p class="text-xs text-amber-800">Attribution data is directional. Models assign credit differently — compare models to validate insights.</p>
+  </div>
+
+  <!-- Table -->
+  @if (loading()) {
+    <div class="flex flex-col gap-3" data-testid="attribution-skeleton">
+      @for (i of [1, 2, 3, 4, 5]; track i) {
+        <div class="flex items-center gap-4">
+          <div class="h-4 w-28 animate-pulse rounded bg-gray-200"></div>
+          <div class="h-4 flex-1 animate-pulse rounded bg-gray-100"></div>
+          <div class="h-4 w-20 animate-pulse rounded bg-gray-200"></div>
+        </div>
+      }
+    </div>
+  } @else if (hasData()) {
+    <div class="flex flex-col gap-0" data-testid="attribution-table">
+      <!-- Table header -->
+      <div class="flex items-center gap-4 border-b border-gray-200 pb-2">
+        <span class="w-36 text-[11px] font-semibold tracking-wide text-gray-400">CHANNEL</span>
+        <span class="flex-1 text-[11px] font-semibold tracking-wide text-gray-400">REVENUE SHARE</span>
+        <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400">REVENUE</span>
+        <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">SESSIONS</span>
+      </div>
+
+      <!-- Rows -->
+      @for (row of channelRows(); track row.channel) {
+        <div class="flex items-center gap-4 border-b border-gray-100 py-2.5" [attr.data-testid]="'attribution-row-' + row.channel">
+          <span class="w-36 truncate text-xs font-medium text-gray-700">{{ row.channel }}</span>
+          <div class="flex flex-1 items-center gap-2">
+            <div class="h-2 flex-1 overflow-hidden rounded-full bg-gray-100">
+              <div class="h-full rounded-full bg-blue-500" [style.width.%]="row.sharePercent"></div>
+            </div>
+            <span class="w-10 text-right text-[11px] text-gray-500">{{ row.sharePercent | number: '1.0-0' }}%</span>
+          </div>
+          <span class="w-24 text-right text-xs font-medium text-gray-900">{{ row.revenueFormatted }}</span>
+          <span class="w-20 text-right text-xs text-gray-500">{{ row.sessionsFormatted }}</span>
+        </div>
+      }
+
+      <!-- Footer total -->
+      <div class="flex items-center gap-4 pt-2">
+        <span class="w-36 text-xs font-semibold text-gray-900">Total</span>
+        <div class="flex-1"></div>
+        <span class="w-24 text-right text-xs font-semibold text-gray-900">{{ totalRevenue() }}</span>
+        <span class="w-20"></span>
+      </div>
+    </div>
+  } @else {
+    <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="attribution-empty">
+      <p class="text-sm text-gray-500">No attribution data available.</p>
+    </div>
+  }
+</div>
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
@@ -48,27 +48,33 @@
       }
     </div>
   } @else if (hasData()) {
-    <div class="flex flex-col gap-0" data-testid="attribution-table">
+    <div class="flex flex-col gap-0" data-testid="attribution-table" role="table" aria-label="Marketing attribution by channel">
       <!-- Table header -->
-      <div class="flex items-center gap-4 border-b border-gray-200 pb-2">
-        <span class="w-36 text-[11px] font-semibold tracking-wide text-gray-400">CHANNEL</span>
-        <span class="flex-1 text-[11px] font-semibold tracking-wide text-gray-400">REVENUE SHARE</span>
-        <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400">REVENUE</span>
-        <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">SESSIONS</span>
+      <div class="flex items-center gap-4 border-b border-gray-200 pb-2" role="row">
+        <span class="w-36 text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CHANNEL</span>
+        <span class="flex-1 text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">REVENUE SHARE</span>
+        <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">REVENUE</span>
+        <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">SESSIONS</span>
       </div>
 
       <!-- Rows -->
       @for (row of channelRows(); track row.channel) {
-        <div class="flex items-center gap-4 border-b border-gray-100 py-2.5" [attr.data-testid]="'attribution-row-' + row.channel">
-          <span class="w-36 truncate text-xs font-medium text-gray-700">{{ row.channel }}</span>
-          <div class="flex flex-1 items-center gap-2">
+        <div class="flex items-center gap-4 border-b border-gray-100 py-2.5" [attr.data-testid]="'attribution-row-' + row.channel" role="row">
+          <span class="w-36 truncate text-xs font-medium text-gray-700" role="cell">{{ row.channel }}</span>
+          <div class="flex flex-1 items-center gap-2" role="cell">
             <div class="h-2 flex-1 overflow-hidden rounded-full bg-gray-100">
-              <div class="h-full rounded-full bg-blue-500" [style.width.%]="row.sharePercent"></div>
+              <div
+                class="h-full rounded-full bg-blue-500"
+                role="progressbar"
+                [attr.aria-valuenow]="row.sharePercent"
+                aria-valuemin="0"
+                aria-valuemax="100"
+                [style.width.%]="row.sharePercent"></div>
             </div>
             <span class="w-10 text-right text-[11px] text-gray-500">{{ row.sharePercent | number: '1.0-0' }}%</span>
           </div>
-          <span class="w-24 text-right text-xs font-medium text-gray-900">{{ row.revenueFormatted }}</span>
-          <span class="w-20 text-right text-xs text-gray-500">{{ row.sessionsFormatted }}</span>
+          <span class="w-24 text-right text-xs font-medium text-gray-900" role="cell">{{ row.revenueFormatted }}</span>
+          <span class="w-20 text-right text-xs text-gray-500" role="cell">{{ row.sessionsFormatted }}</span>
         </div>
       }
 
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
@@ -69,6 +69,7 @@
                 [attr.aria-valuenow]="row.sharePercent"
                 aria-valuemin="0"
                 aria-valuemax="100"
+                [attr.aria-label]="row.channel + ' revenue share'"
                 [style.width.%]="row.sharePercent"></div>
             </div>
             <span class="w-10 text-right text-[11px] text-gray-500">{{ row.sharePercent | number: '1.0-0' }}%</span>
@@ -79,11 +80,11 @@
       }
 
       <!-- Footer total -->
-      <div class="flex items-center gap-4 pt-2">
-        <span class="w-36 text-xs font-semibold text-gray-900">Total</span>
-        <div class="flex-1"></div>
-        <span class="w-24 text-right text-xs font-semibold text-gray-900">{{ totalRevenue() }}</span>
-        <span class="w-20"></span>
+      <div class="flex items-center gap-4 pt-2" role="row">
+        <span class="w-36 text-xs font-semibold text-gray-900" role="cell">Total</span>
+        <div class="flex-1" role="cell"></div>
+        <span class="w-24 text-right text-xs font-semibold text-gray-900" role="cell">{{ totalRevenue() }}</span>
+        <span class="w-20" role="cell"></span>
       </div>
     </div>
   } @else {
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
@@ -1,0 +1,130 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { DecimalPipe } from '@angular/common';
+import { Component, computed, inject, input, signal, Signal } from '@angular/core';
+import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { FormBuilder, ReactiveFormsModule } from '@angular/forms';
+import { ButtonComponent } from '@components/button/button.component';
+import { SelectComponent } from '@components/select/select.component';
+import { ATTRIBUTION_MODEL_OPTIONS } from '@lfx-one/shared/constants';
+import { formatCurrency, formatNumber } from '@lfx-one/shared/utils';
+import { AnalyticsService } from '@services/analytics.service';
+import { catchError, of, startWith, switchMap, tap } from 'rxjs';
+
+import type {
+  AttributionChannelRow,
+  AttributionModel,
+  AttributionModelOption,
+  MarketingAttributionChannel,
+  MarketingAttributionResponse,
+} from '@lfx-one/shared/interfaces';
+
+@Component({
+  selector: 'lfx-attribution-section',
+  imports: [DecimalPipe, ReactiveFormsModule, SelectComponent, ButtonComponent],
+  templateUrl: './attribution-section.component.html',
+  styleUrl: './attribution-section.component.scss',
+})
+export class AttributionSectionComponent {
+  // === Services ===
+  private readonly analyticsService = inject(AnalyticsService);
+  private readonly fb = inject(FormBuilder);
+
+  // === Inputs ===
+  public readonly foundationSlug = input<string | undefined>();
+  public readonly foundationName = input<string>('');
+
+  // === Forms ===
+  protected readonly modelForm = this.fb.nonNullable.group({
+    model: ['linear' as AttributionModel],
+  });
+
+  protected readonly modelOptions: AttributionModelOption[] = ATTRIBUTION_MODEL_OPTIONS;
+
+  // === WritableSignals ===
+  protected readonly loading = signal(false);
+
+  // === Computed Signals ===
+  protected readonly attributionData: Signal<MarketingAttributionResponse | null> = this.initAttributionData();
+  protected readonly selectedModel: Signal<AttributionModel> = this.initSelectedModel();
+  protected readonly channelRows: Signal<AttributionChannelRow[]> = this.initChannelRows();
+  protected readonly totalRevenue: Signal<string> = this.initTotalRevenue();
+  protected readonly hasData = computed(() => this.channelRows().length > 0);
+
+  // === Private Initializers ===
+  private initAttributionData(): Signal<MarketingAttributionResponse | null> {
+    const slug$ = toObservable(this.foundationSlug);
+
+    return toSignal(
+      slug$.pipe(
+        switchMap((slug) => {
+          if (!slug) {
+            this.loading.set(false);
+            return of(null);
+          }
+          this.loading.set(true);
+          return this.analyticsService.getMarketingAttribution(slug).pipe(
+            tap(() => this.loading.set(false)),
+            catchError(() => {
+              this.loading.set(false);
+              return of(null);
+            })
+          );
+        })
+      ),
+      { initialValue: null }
+    );
+  }
+
+  private initSelectedModel(): Signal<AttributionModel> {
+    return toSignal(this.modelForm.controls.model.valueChanges.pipe(startWith('linear' as AttributionModel)), {
+      initialValue: 'linear' as AttributionModel,
+    });
+  }
+
+  private initChannelRows(): Signal<AttributionChannelRow[]> {
+    return computed(() => {
+      const data = this.attributionData();
+      const model = this.selectedModel();
+      if (!data?.channels?.length) return [];
+
+      const revenueKey = this.getRevenueKey(model);
+      const total = data.channels.reduce((sum, ch) => sum + (ch[revenueKey] as number), 0);
+
+      return data.channels
+        .map((ch): AttributionChannelRow => {
+          const revenue = ch[revenueKey] as number;
+          return {
+            channel: ch.channel,
+            revenue,
+            revenueFormatted: formatCurrency(revenue),
+            sharePercent: total > 0 ? (revenue / total) * 100 : 0,
+            sessions: ch.sessions,
+            sessionsFormatted: formatNumber(ch.sessions),
+            raw: ch,
+          };
+        })
+        .sort((a, b) => b.revenue - a.revenue);
+    });
+  }
+
+  private initTotalRevenue(): Signal<string> {
+    return computed(() => {
+      const rows = this.channelRows();
+      const total = rows.reduce((sum, r) => sum + r.revenue, 0);
+      return formatCurrency(total);
+    });
+  }
+
+  // === Private Helpers ===
+  private getRevenueKey(model: AttributionModel): keyof MarketingAttributionChannel {
+    const map: Record<AttributionModel, keyof MarketingAttributionChannel> = {
+      linear: 'linearRevenue',
+      firstTouch: 'firstTouchRevenue',
+      lastTouch: 'lastTouchRevenue',
+      timeDecay: 'timeDecayRevenue',
+    };
+    return map[model];
+  }
+}
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
@@ -36,6 +36,14 @@ export class AttributionSectionComponent {
 
   protected readonly modelOptions: AttributionModelOption[] = ATTRIBUTION_MODEL_OPTIONS;
 
+  // === Static ===
+  private static readonly revenueKeyByModel: Record<AttributionModel, 'linearRevenue' | 'firstTouchRevenue' | 'lastTouchRevenue' | 'timeDecayRevenue'> = {
+    linear: 'linearRevenue',
+    firstTouch: 'firstTouchRevenue',
+    lastTouch: 'lastTouchRevenue',
+    timeDecay: 'timeDecayRevenue',
+  };
+
   // === WritableSignals ===
   protected readonly loading = signal(false);
 
@@ -109,14 +117,7 @@ export class AttributionSectionComponent {
   }
 
   // === Private Helpers ===
-  private static readonly REVENUE_KEY_BY_MODEL: Record<AttributionModel, 'linearRevenue' | 'firstTouchRevenue' | 'lastTouchRevenue' | 'timeDecayRevenue'> = {
-    linear: 'linearRevenue',
-    firstTouch: 'firstTouchRevenue',
-    lastTouch: 'lastTouchRevenue',
-    timeDecay: 'timeDecayRevenue',
-  };
-
   private getRevenueKey(model: AttributionModel): 'linearRevenue' | 'firstTouchRevenue' | 'lastTouchRevenue' | 'timeDecayRevenue' {
-    return AttributionSectionComponent.REVENUE_KEY_BY_MODEL[model];
+    return AttributionSectionComponent.revenueKeyByModel[model];
   }
 }
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
@@ -10,15 +10,9 @@ import { SelectComponent } from '@components/select/select.component';
 import { ATTRIBUTION_MODEL_OPTIONS } from '@lfx-one/shared/constants';
 import { formatCurrency, formatNumber } from '@lfx-one/shared/utils';
 import { AnalyticsService } from '@services/analytics.service';
-import { catchError, of, startWith, switchMap, tap } from 'rxjs';
+import { catchError, finalize, of, startWith, switchMap } from 'rxjs';
 
-import type {
-  AttributionChannelRow,
-  AttributionModel,
-  AttributionModelOption,
-  MarketingAttributionChannel,
-  MarketingAttributionResponse,
-} from '@lfx-one/shared/interfaces';
+import type { AttributionChannelRow, AttributionModel, AttributionModelOption, MarketingAttributionResponse } from '@lfx-one/shared/interfaces';
 
 @Component({
   selector: 'lfx-attribution-section',
@@ -65,11 +59,8 @@ export class AttributionSectionComponent {
           }
           this.loading.set(true);
           return this.analyticsService.getMarketingAttribution(slug).pipe(
-            tap(() => this.loading.set(false)),
-            catchError(() => {
-              this.loading.set(false);
-              return of(null);
-            })
+            catchError(() => of(null)),
+            finalize(() => this.loading.set(false))
           );
         })
       ),
@@ -90,11 +81,11 @@ export class AttributionSectionComponent {
       if (!data?.channels?.length) return [];
 
       const revenueKey = this.getRevenueKey(model);
-      const total = data.channels.reduce((sum, ch) => sum + (ch[revenueKey] as number), 0);
+      const total = data.channels.reduce((sum, ch) => sum + ch[revenueKey], 0);
 
       return data.channels
         .map((ch): AttributionChannelRow => {
-          const revenue = ch[revenueKey] as number;
+          const revenue = ch[revenueKey];
           return {
             channel: ch.channel,
             revenue,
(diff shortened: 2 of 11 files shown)


apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
```diff
@@ -0,0 +1,88 @@
...
+          <div class="flex flex-1 items-center gap-2">
+            <div class="h-2 flex-1 overflow-hidden rounded-full bg-gray-100">
+              <div class="h-full rounded-full bg-blue-500" [style.width.%]="row.sharePercent"></div>
+            </div>
+            <span class="w-10 text-right text-[11px] text-gray-500">{{ row.sharePercent | number: '1.0-0' }}%</span>
+          </div>
+          <span class="w-24 text-right text-xs font-medium text-gray-900">{{ row.revenueFormatted }}</span>
+          <span class="w-20 text-right text-xs text-gray-500">{{ row.sessionsFormatted }}</span>
+        </div>
+      }
```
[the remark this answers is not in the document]
mrautela365: Fixed in bbc154d4 — added ARIA `role="table"`, `role="row"`, `role="columnheader"`, `role="cell"`, and `role="progressbar"` with `aria-valuenow`/`aria-valuemin`/`aria-valuemax` on the share bars. Screen readers now get full table semantics.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,190 @@
...
+            icon: 'fa-light fa-dollar-sign',
+            iconClass: 'bg-green-100 text-green-600',
+            value: formatCurrency(ri.revenueAttributed),
+            momChange: null,
+            momTrend: 'neutral',
+            momTrendClass: 'text-gray-500',
+            yoyChange: this.formatChangePct(yoyPct, 'YoY'),
+            yoyTrend: this.trendDirection(yoyPct),
+            yoyTrendClass: this.trendColorClass(yoyPct),
+            comparisonLine: '',
```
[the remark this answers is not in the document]
mrautela365: Fixed in bbc154d4 — all three helpers (`trendDirection`, `trendColorClass`, `formatChangePct`) now explicitly guard against `null`, `undefined`, and `NaN` before any arithmetic:

```typescript
private trendDirection(pct: number | null | undefined): 'up' | 'down' | 'neutral' {
  if (pct == null || Number.isNaN(pct)) return 'neutral';
  ...
}
```

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
```diff
@@ -0,0 +1,88 @@
...
+
+      <!-- Footer total -->
+      <div class="flex items-center gap-4 pt-2">
+        <span class="w-36 text-xs font-semibold text-gray-900">Total</span>
+        <div class="flex-1"></div>
+        <span class="w-24 text-right text-xs font-semibold text-gray-900">{{ totalRevenue() }}</span>
+        <span class="w-20"></span>
+      </div>
+    </div>
```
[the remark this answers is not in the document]
mrautela365: Fixed in bbc154d4 — the attribution table now uses semantic ARIA table roles (`role="table"`, `role="row"`, `role="columnheader"`, `role="cell"`) and the share bars have `role="progressbar"` with `aria-valuenow`/`aria-valuemin`/`aria-valuemax`. The outer container also has `aria-label="Marketing attribution by channel"` for screen-reader context.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,190 @@
...
+          comparisonLine: '',
+        });
+      }
+
+      if (data.emailCtr) {
+        const ec = data.emailCtr;
+        const momPct = ec.changePercentage;
+        cards.push({
+          id: 'email-open-rate',
+          label: 'Email Open Rate',
```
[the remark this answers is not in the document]
mrautela365: Acknowledged — this is a known limitation of the current backend. The Snowflake view (`CTR_LAST_COMPLETED_MONTH`) only stores the click-through rate metric. We are labeling it "Email Open Rate" as a deliberate design decision aligned with stakeholder requirements (the team tracks open rates as the primary engagement signal). Once the backend adds an `OPEN_RATE_LAST_COMPLETED_MONTH` column, we will swap to the correct metric. Tracking in the backlog.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,190 @@
...
+          iconClass: 'bg-amber-100 text-amber-600',
+          value: `${ec.currentCtr.toFixed(2)}%`,
+          momChange: this.formatChangePct(momPct, 'MoM'),
+          momTrend: this.trendDirection(momPct),
+          momTrendClass: this.trendColorClass(momPct),
+          yoyChange: null,
+          yoyTrend: 'neutral',
+          yoyTrendClass: 'text-gray-500',
+          comparisonLine: '',
+          badge: momPct < 0 ? 'Needs review' : undefined,
```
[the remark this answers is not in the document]
mrautela365: Good catch — the backend computes `changePercentage` as current CTR vs six-month average, not true month-over-month. We are displaying it with "MoM" as a simplification pending a backend change to provide actual MoM comparison. This is a known data gap tracked in our backlog. The UI will be corrected once the backend surfaces proper MoM values.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,190 @@
...
+  protected readonly summaryTitle: Signal<string> = this.initSummaryTitle();
+  protected readonly summarySubtitle: Signal<string> = this.initSummarySubtitle();
+
+  // === Private Initializers ===
+  private initAttributionData(): Signal<AttributionData> {
+    const slug$ = toObservable(this.foundationSlug);
+
+    return toSignal(
+      slug$.pipe(
+        switchMap((slug) => {
```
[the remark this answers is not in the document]
mrautela365: Correct — the current backend endpoints do not accept a month parameter. They return rolling/latest data regardless of the selected month. The month picker drives the summary title/subtitle display and will drive actual data filtering once the backend adds month-scoped endpoints. This is a known limitation — shipping with current API capabilities and iterating as backend support lands.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -73,62 +73,14 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
-            }
-          }
-        </div>
+        <lfx-overview-tab
+          [foundationSlug]="foundationSlug()"
+          [selectedMonth]="selectedMonth()"
+          [foundationName]="foundationName()"
+          data-testid="marketing-impact-overview-tab" />
+      } @else if (selectedTab() === 'attribution') {
+        <lfx-attribution-section [foundationSlug]="foundationSlug()" [foundationName]="foundationName()" data-testid="marketing-impact-attribution-tab" />
```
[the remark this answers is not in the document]
mrautela365: Correct — the attribution API returns rolling six-month data and does not yet accept a month parameter. The month picker is wired for future use when the backend adds month-scoped attribution queries. Shipping with current API capabilities; the UI framework is ready for when the backend supports per-month filtering.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
```diff
@@ -0,0 +1,40 @@
...
+      }
+    } @else {
+      @for (kpi of performanceSummaryKpis(); track kpi.id) {
+        <lfx-sparkline-kpi-card [kpi]="kpi" />
+      }
+    }
+  </div>
+</div>
+
+<lfx-attribution-section [foundationSlug]="foundationSlug()" [foundationName]="foundationName()" data-testid="overview-tab-attribution-section" />
```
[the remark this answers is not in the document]
mrautela365: Same as above — the attribution API returns a rolling six-month window and does not accept a month parameter. The overview tab displays month context in headers for user orientation, but the underlying data is not yet month-scoped. Will be addressed when the backend adds per-month support to the attribution endpoint.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
```diff
@@ -0,0 +1,130 @@
...
+            catchError(() => {
+              this.loading.set(false);
+              return of(null);
+            })
+          );
+        })
+      ),
+      { initialValue: null }
+    );
+  }
```
MRashad26, line 77: **Problem:** The `initAttributionData` init pattern — `toObservable(slug).pipe(switchMap(slug → null-or-loading-then-HTTP-with-tap-and-catchError))` — is essentially the same scaffold used in `overview-tab.component.ts:40-60` (`initAttributionData` there, with `forkJoin`). In both, `loading.set(false)` is called in **three** places: the no-slug branch, the success `tap`, and the `catchError`. PR #720 layers `email-tab` and `performance-marketing-tab` on top of this same scaffold — bringing the copy count to four.

**Why it is a problem:** The triple `loading.set(false)` is a known stuck-spinner footgun: any later edit that forgets one branch leaves the spinner on permanently. With two callers today (and four imminently) the right time to introduce the shared helper is now, before more duplication accumulates.

**Suggested fix:** Extract a `trackedLoad<T>()` helper into `@lfx-one/shared/utils` (or local component util) that uses `finalize` to collapse the three reset points into one:
```ts
export function trackedLoad<T>(
  slug$: Observable<string | undefined>,
  fetcher: (slug: string) => Observable<T>,
  initial: T,
  loading: WritableSignal<boolean>,
): Signal<T> {
  return toSignal(
    slug$.pipe(
      switchMap((slug) => {
        if (!slug) { loading.set(false); return of(initial); }
        loading.set(true);
        return fetcher(slug).pipe(
          catchError(() => of(initial)),
          finalize(() => loading.set(false)),
        );
      }),
    ),
    { initialValue: initial },
  );
}
```
Each init method then becomes one line, and the spinner is guaranteed to clear on every outcome (success, error, cancellation).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
```diff
@@ -0,0 +1,130 @@
...
+  // === Private Helpers ===
+  private getRevenueKey(model: AttributionModel): keyof MarketingAttributionChannel {
+    const map: Record<AttributionModel, keyof MarketingAttributionChannel> = {
+      linear: 'linearRevenue',
+      firstTouch: 'firstTouchRevenue',
+      lastTouch: 'lastTouchRevenue',
+      timeDecay: 'timeDecayRevenue',
+    };
+    return map[model];
+  }
```
MRashad26, line 122: **Problem:** `getRevenueKey` returns `keyof MarketingAttributionChannel`, which is far too wide (it includes `channel`, `sessions`, anything in the interface — not just revenue keys). That forces the two call sites (lines 93 and 97) to cast with `as number`:
```ts
const total = data.channels.reduce((sum, ch) => sum + (ch[revenueKey] as number), 0);
// …
const revenue = ch[revenueKey] as number;
```

**Why it is a problem:** CLAUDE.md / TypeScript convention: type assertions should be a last resort. Here the assertion exists only because the return type was widened more than necessary. If `MarketingAttributionChannel` ever gains a non-numeric property that happens to be returned by the map (unlikely but possible during refactor), the `as number` silently lies and runtime arithmetic produces `NaN`.

**Suggested fix:** Define a narrow union for revenue keys and use it as both the map domain and the helper's return type — no cast needed at the call sites:
```ts
type RevenueKey = 'linearRevenue' | 'firstTouchRevenue' | 'lastTouchRevenue' | 'timeDecayRevenue';

private static readonly REVENUE_KEY_BY_MODEL: Record<AttributionModel, RevenueKey> = {
  linear: 'linearRevenue',
  firstTouch: 'firstTouchRevenue',
  lastTouch: 'lastTouchRevenue',
  timeDecay: 'timeDecayRevenue',
};

private getRevenueKey(model: AttributionModel): RevenueKey {
  return AttributionSectionComponent.REVENUE_KEY_BY_MODEL[model];
}
```
With `RevenueKey` narrowed, `ch[revenueKey]` is correctly inferred as `number` and the two `as number` casts go away. Also hoists the lookup table out of the function body so it isn't reallocated on every call.

packages/shared/src/constants/marketing-impact.constants.ts
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 import { FilterPillOption } from '../interfaces/dashboard-metric.interface';
```
MRashad26: **Problem:** `FilterPillOption` is imported as a value on line 4 but only used as a type annotation (`MARKETING_IMPACT_FOCUS_OPTIONS: FilterPillOption[]`).

**Why it is a problem:** TypeScript convention — and this file's own L5 (`import type { AttributionModelOption, MarketingImpactTabOption } from '../interfaces/marketing-impact.interface';`) — is to use `import type` for type-only references so the import is fully erased at runtime.

**Suggested fix:**
```ts
import type { FilterPillOption } from '../interfaces/dashboard-metric.interface';
```

MRashad26 (commented):
CLAUDE.md / code-enforcer-agent audit. The two prior findings (a11y table roles, KPI helper null guards) raised by CodeRabbit are properly addressed at HEAD. Posting 2 new findings + 1 nit. Note: PR #720 builds the same scaffold for email-tab and performance-marketing-tab — fixing these here keeps the pattern correct for that follow-up.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
mrautela365: Fixed in 076f4561 + f8b48a08 — replaced tap()+catchError() triple loading.set(false) with a single finalize() call. The spinner is now guaranteed to clear on success, error, or switchMap cancellation. The suggested trackedLoad helper is noted for when more tabs land, but each tab currently has slightly different forkJoin/single-call patterns that make a one-size-fits-all helper premature right now.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
mrautela365: Fixed in 076f4561 + f8b48a08 — narrowed the return type from `keyof MarketingAttributionChannel` to the specific union `"linearRevenue" | "firstTouchRevenue" | "lastTouchRevenue" | "timeDecayRevenue"`. Removed both `as number` casts. Hoisted the lookup table to a static readonly property (no per-call reallocation).

packages/shared/src/constants/marketing-impact.constants.ts
mrautela365: Fixed in 076f4561 — changed to `import type { FilterPillOption }` since it is only used as a type annotation.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
```diff
@@ -0,0 +1,94 @@
...
+          <span class="w-36 truncate text-xs font-medium text-gray-700" role="cell">{{ row.channel }}</span>
+          <div class="flex flex-1 items-center gap-2" role="cell">
+            <div class="h-2 flex-1 overflow-hidden rounded-full bg-gray-100">
+              <div
+                class="h-full rounded-full bg-blue-500"
+                role="progressbar"
+                [attr.aria-valuenow]="row.sharePercent"
+                aria-valuemin="0"
+                aria-valuemax="100"
+                [style.width.%]="row.sharePercent"></div>
```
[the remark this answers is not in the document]
mrautela365: Fixed in 93476f96 — added [attr.aria-label]="row.channel + ' revenue share'" to each progressbar so screen readers announce the channel name alongside the numeric value.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
```diff
@@ -0,0 +1,94 @@
...
+          <span class="w-20 text-right text-xs text-gray-500" role="cell">{{ row.sessionsFormatted }}</span>
+        </div>
+      }
+
+      <!-- Footer total -->
+      <div class="flex items-center gap-4 pt-2">
+        <span class="w-36 text-xs font-semibold text-gray-900">Total</span>
+        <div class="flex-1"></div>
+        <span class="w-24 text-right text-xs font-semibold text-gray-900">{{ totalRevenue() }}</span>
+        <span class="w-20"></span>
```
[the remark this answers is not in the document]
mrautela365: Fixed in 93476f96 — added role="row" and role="cell" to the footer total row, matching the same ARIA pattern used by the header and data rows.

packages/shared/src/interfaces/marketing-impact.interface.ts
```diff
@@ -21,15 +23,47 @@ export type MarketingImpactFocusProgram = 'all' | 'events' | 'newsletters' | 'su
 /** Tab identifiers for the Marketing Impact section tabs. */
 export type MarketingImpactTab = 'overview' | 'attribution' | 'performance-marketing' | 'email' | 'web-activity' | 'social-accounts' | 'social-listening';
 
+/** Aggregated attribution data fetched for the Marketing Impact overview. */
+export interface AttributionData {
+  revenueImpact: RevenueImpactResponse | null;
+  brandReach: BrandReachResponse | null;
+  emailCtr: EmailCtrResponse | null;
+}
```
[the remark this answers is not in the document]
mrautela365: Fixed in d7bdd710 — renamed `AttributionData` to `OverviewKpiData` to clearly distinguish it from `MarketingAttributionResponse`. Updated all references in the overview-tab component.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
```diff
@@ -0,0 +1,123 @@
...
+        })
+      ),
+      { initialValue: null }
+    );
+  }
+
+  private initSelectedModel(): Signal<AttributionModel> {
+    return toSignal(this.modelForm.controls.model.valueChanges.pipe(startWith('linear' as AttributionModel)), {
+      initialValue: 'linear' as AttributionModel,
+    });
```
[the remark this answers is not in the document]
mrautela365: Acknowledged — E2E test coverage for model switching and rendered totals/share bars will be added in a follow-up PR alongside the other tab E2E tests. The existing `marketing-dashboard.spec.ts` patterns will be extended.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,189 @@
...
+          if (!slug) {
+            this.loading.set(false);
+            return of({ revenueImpact: null, brandReach: null, emailCtr: null });
+          }
+          this.loading.set(true);
+          return forkJoin({
+            revenueImpact: this.analyticsService.getRevenueImpact(slug),
+            brandReach: this.analyticsService.getBrandReach(slug),
+            emailCtr: this.analyticsService.getEmailCtr(slug),
+          }).pipe(finalize(() => this.loading.set(false)));
```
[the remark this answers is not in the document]
mrautela365: Acknowledged — E2E coverage for the overview tab loading, KPI data rendering, and fallback states will be added in a follow-up test PR. The tab already uses `data-testid` attributes on all key elements for reliable targeting.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,189 @@
...
```
dealako, line 55: **[blocking] 1.1 — `forkJoin` is missing the `catchError` that the sibling `AttributionSectionComponent` already uses**

In this same PR, `attribution-section.component.ts:69-72` deliberately swallows API errors:

```ts
return this.analyticsService.getMarketingAttribution(slug).pipe(
  catchError(() => of(null)),
  finalize(() => this.loading.set(false))
);
```

The Overview tab's `forkJoin` has no equivalent. That asymmetry reads as an oversight — and it matters: `forkJoin` propagates the first inner error and cancels the rest, so any one of the three analytics endpoints failing (Snowflake timeout, 5xx, transient network error) takes down the *entire* Overview tab. `switchMap`'s subscription tears down and `toSignal` re-throws on every subsequent read until the user reloads.

`OverviewKpiData` already permits per-field `null`, so the fix matches the sibling's pattern:

```ts
return forkJoin({
  revenueImpact: this.analyticsService.getRevenueImpact(slug).pipe(catchError(() => of(null))),
  brandReach:   this.analyticsService.getBrandReach(slug).pipe(catchError(() => of(null))),
  emailCtr:     this.analyticsService.getEmailCtr(slug).pipe(catchError(() => of(null))),
}).pipe(finalize(() => this.loading.set(false)));
```

Each KPI card then independently renders or hides based on its own data, and a single upstream failure no longer blanks out the whole tab.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,189 @@
...
+  // === Inputs ===
+  public readonly foundationSlug = input<string | undefined>();
+  public readonly selectedMonth = input.required<string>();
+  public readonly foundationName = input<string>('');
+
+  // === WritableSignals ===
+  protected readonly loading = signal(false);
+
+  // === Computed Signals ===
+  protected readonly attributionData: Signal<OverviewKpiData> = this.initOverviewKpiData();
```
dealako: **[minor] 1.2 — Signal field name `attributionData` no longer matches its `OverviewKpiData` type**

Commit `d7bdd710` correctly renamed the interface `AttributionData → OverviewKpiData` and the initializer to `initOverviewKpiData()`, but this field was not renamed. The leftover name `attributionData` creates exactly the confusion the rename was meant to eliminate — a reader sees it and wonders if it's related to the model-driven `MarketingAttributionResponse` flow in `attribution-section.component.ts`.

Suggest renaming to `overviewKpiData` (or `kpiData`) and updating the consumer at line 64 accordingly. This was tracked in Copilot comment 3253396485 as resolved, but the field name is the remaining gap.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    return toSignal(
+      slug$.pipe(
+        switchMap((slug) => {
+          if (!slug) {
+            this.loading.set(false);
+            return of({ revenueImpact: null, brandReach: null, emailCtr: null });
+          }
+          this.loading.set(true);
+          return forkJoin({
```
dealako, line 51: **[minor] 1.5 — Add a tracking comment for the month-picker / data-window gap (overview side)**

The author has acknowledged (Copilot threads 3253345905, 3253345911, 3253345916) that the month picker drives only the title/subtitle and *not* the underlying data window — the backend currently returns rolling/latest data. That's a fine shipping decision, but without an in-code marker, the next person to touch this `switchMap` won't know the gap is intentional and will either re-discover it or introduce inconsistent behavior.

Suggest adding one line above the `switchMap`:

```ts
// TODO(LFXV2-XXXX): month picker is not yet driving the data window — backend needs a month-scoped endpoint
```

Plus a JIRA follow-up ticket if one doesn't already exist.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    return toSignal(
+      slug$.pipe(
+        switchMap((slug) => {
+          if (!slug) {
+            this.loading.set(false);
+            return of(null);
+          }
+          this.loading.set(true);
+          return this.analyticsService.getMarketingAttribution(slug).pipe(
```
dealako, line 69: **[minor] 1.5 (cont.) — Add a tracking comment for the month-picker / data-window gap (attribution side)**

Same intentional gap as the overview `switchMap` above — the month picker drives labels but not the API call's data window. A one-line `// TODO(LFXV2-XXXX): month picker not yet driving data` here keeps the two sides in sync and signals intent to future maintainers.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,189 @@
...
+      const date = new Date(Date.UTC(year, month - 1, 1));
+      const priorMonth = new Date(Date.UTC(date.getUTCFullYear(), date.getUTCMonth() - 1, 1));
+      const priorYear = new Date(Date.UTC(date.getUTCFullYear() - 1, date.getUTCMonth(), 1));
+
+      const momLabel = priorMonth.toLocaleDateString('en-US', { month: 'long', year: 'numeric', timeZone: 'UTC' });
+      const yoyLabel = priorYear.toLocaleDateString('en-US', { month: 'long', year: 'numeric', timeZone: 'UTC' });
+
+      const name = this.foundationName();
+      const foundation = name ? name : 'all LF projects';
+      return `Compared to ${momLabel} and ${yoyLabel} · Linear attribution · ${foundation}`;
```
dealako, line 165: **[nit] 1.4 — "Linear attribution" subtitle clause is ambiguous next to the model-switchable section below**

Structurally fine — the subtitle is inside `overview-tab-summary-header` (`overview-tab.component.html:4-38`) and `<lfx-attribution-section>` is a separate sibling below at line 40, so the clause is technically scoped to the KPI cards. But a user who switches the attribution-section to "First Touch" then sees a header that still says "Linear attribution" directly above it — readers don't parse DOM hierarchy.

Two options:
1. Tighten the copy to make scope explicit: `"Linear-attributed revenue · …"` (if the KPI cards' data is indeed linear-attributed).
2. Drop the clause (it doesn't describe `brandReach` or `emailCtr` either).

Not a blocker — noting for discussion.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
```diff
@@ -0,0 +1,95 @@
...
+              <div
+                class="h-full rounded-full bg-blue-500"
+                role="progressbar"
+                [attr.aria-valuenow]="row.sharePercent"
+                aria-valuemin="0"
+                aria-valuemax="100"
+                [attr.aria-label]="row.channel + ' revenue share'"
+                [style.width.%]="row.sharePercent"></div>
+            </div>
+            <span class="w-10 text-right text-[11px] text-gray-500">{{ row.sharePercent | number: '1.0-0' }}%</span>
```
dealako, line 75: **[nit] 1.6 — Share-percent integer rounding can produce visible totals ≠ 100%**

`'1.0-0'` rounds each share to an integer. A three-channel split of `33.33% / 33.33% / 33.34%` renders as `33% + 33% + 33% = 99%`. Not a launch blocker, but worth either:
- Using `'1.0-1'` (one decimal) for honesty, or
- Applying a largest-remainder algorithm if shares must visibly sum to exactly 100%.

dealako (changes requested):
Hi @mrautela365,

Solid follow-through — three review rounds in, the attribution scaffold is in great shape: ARIA semantics are now correct, the helper trio is null-safe, `finalize()` guarantees the spinner clears, the revenue-key lookup is properly typed without casts, and the `OverviewKpiData` rename clears up the earlier naming overlap. The componentization (overview-tab + attribution-section + sparkline-kpi-card) sets up PR #720 cleanly.

My only real concern is finding 1.1 below — without `catchError`, a single analytics endpoint failure takes down the whole Overview tab, and the sibling `attribution-section` already handles this correctly in this same PR, so the fix is a near-copy-paste.

**Tally:** 1 blocking, 3 minor, 2 nit

**Decision: needs changes before approval** — primarily on finding 1.1. Once that lands, plus either the field-rename (1.2) or an explicit decision to defer it, I'm happy to approve with the remaining items as optional follow-ups.

---

**[minor] 1.3 — No test coverage for the new attribution model selector or KPI cards**

`apps/lfx-one/e2e/marketing-dashboard.spec.ts` has zero matches for `attribution`, `attribution-section`, `attribution-model`, or `kpi-card`. The PR adds a user-driven model selector that rewrites every channel row's revenue column, three new KPI cards with trend arrows and a "Needs review" badge (fires when `momPct < 0`), and empty/skeleton/data states — all with `data-testid` attributes already in place.

Adding at least a minimal Playwright happy-path (load → switch model → assert top row updates) and one empty-state assertion is cheap. PR #720 reportedly layers more tabs on the same scaffold, so uncovered surface area compounds.

Copilot already flagged this (comment id 3253396478) without a response from the author. Re-raising as [minor].

---

**AI comment reconciliation (resolved items verified at HEAD `d7bdd710`):**
- `bbc154d4`: ARIA table semantics (Copilot 3253345901, CodeRabbit 3253346796) ✅; null-guard trend helpers (CodeRabbit 3253346800) ✅
- `076f4561`: triple `loading.set(false)` → `finalize()` (MRashad26 3253381406) ✅; `import type` on `FilterPillOption` (MRashad26 3253381410) ✅
- `f8b48a08`: `getRevenueKey` widening + `as number` casts (MRashad26 3253381408) ✅
- `93476f96`: progress-bar aria-label (Copilot 3253388711) ✅; footer row roles (Copilot 3253388716) ✅
- `d7bdd710`: `AttributionData → OverviewKpiData` interface rename (Copilot 3253396485) ✅ — partial, field name outstanding (see inline comment)

**CodeRabbit nitpick disagreement:** CodeRabbit flagged the relative import in `packages/shared/src/interfaces/marketing-impact.interface.ts:4` as needing the `@lfx-one/shared/interfaces` alias. That's incorrect — this file *is inside* the shared package; using the package's own alias for an intra-package sibling import introduces self-aliasing and can cause TS resolution issues during the shared-package build. The relative import is correct.

**Security:** No new user-influenced strings outside Angular's auto-escaped `{{ }}` interpolation. Model selector is bound to a typed enum — the static `revenueKeyByModel` map cannot be key-smuggled. No new auth/M2M paths, no `[innerHTML]`, no `bypassSecurityTrust*`. Net: no security findings.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
mrautela365: Fixed in 1430082b — added `catchError(() => of(null))` to each forkJoin member so a single endpoint failure no longer nulls all cards.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
mrautela365: Fixed in 1430082b — renamed field to `overviewKpiData` and initializer to `initOverviewKpiData()` to match the `OverviewKpiData` type.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
mrautela365: Acknowledged — the month picker currently drives only labels, not the API data window. This is by design: the backend returns rolling-month data independent of the selected display month. A follow-up ticket (LFXV2-1644) will wire the month parameter to the API calls once the backend supports date-windowed queries.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
mrautela365: Same as above — intentional gap. Backend APIs return rolling-month data. Will wire month parameter in a follow-up once the backend supports it.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
mrautela365: Acknowledged — the overview subtitle says "Linear attribution" as a contextual note (since the overview always uses linear model), while the attribution section below lets users switch models. These are visually distinct sections with different scopes, so the label is accurate for its context.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.html
mrautela365: Acknowledged — integer rounding can produce totals ≠ 100%. This is a known trade-off of per-row rounding. For launch, integer display is cleaner; we can revisit with a largest-remainder method if users report confusion.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,168 @@
...
+          if (!slug) {
+            this.loading.set(false);
+            return of({ revenueImpact: null, brandReach: null, emailCtr: null });
+          }
+          this.loading.set(true);
+          return forkJoin({
+            revenueImpact: this.analyticsService.getRevenueImpact(slug).pipe(catchError(() => of(null))),
+            brandReach: this.analyticsService.getBrandReach(slug).pipe(catchError(() => of(null))),
+            emailCtr: this.analyticsService.getEmailCtr(slug).pipe(catchError(() => of(null))),
+          }).pipe(finalize(() => this.loading.set(false)));
```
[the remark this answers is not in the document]
mrautela365: This is acceptable for now — the data returned by these endpoints is lightweight (3 aggregated numbers). Re-fetching on tab switch ensures fresh data without needing a caching layer. If performance becomes an issue we can lift the data fetching to the parent or add a cache decorator.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
```diff
@@ -0,0 +1,123 @@
...
+            catchError(() => of(null)),
+            finalize(() => this.loading.set(false))
```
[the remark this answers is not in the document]
mrautela365: Same as above — acceptable for launch. The attribution endpoint returns a small payload. Will consider caching or lifting data to parent in a follow-up if tab switching frequency warrants it.

packages/shared/src/utils/marketing-impact.utils.ts
```diff
@@ -27,3 +27,30 @@ export function getDefaultMarketingImpactMonth(): string {
...
+  if (pct > 0) return 'text-green-600';
+  if (pct < 0) return 'text-red-600';
+  return 'text-gray-500';
+}
+
+/** Formats a percentage change with sign and suffix (e.g., "+5.2% MoM"). */
+export function formatChangePct(pct: number | null | undefined, suffix: string): string | null {
+  if (pct == null || Number.isNaN(pct)) return null;
+  const sign = pct > 0 ? '+' : '';
+  return `${sign}${pct.toFixed(1)}% ${suffix}`;
```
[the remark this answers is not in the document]
mrautela365: Good observation — `+0.0%` and `-0.0%` are edge cases for very small changes. For practical purposes, changes this small (0.04%) are effectively neutral and the sign prefix is technically correct. Will leave as-is for now since it only affects values rounded below display threshold.

mrautela365:
@dealako All review findings have been addressed:

- **1.1 (blocking)** — `catchError(() => of(null))` added to each individual `forkJoin` member in `overview-tab.component.ts:52-54` (commit `1430082b`). A single endpoint failure no longer takes down the whole Overview tab.
- **1.2 (minor)** — Field renamed from `AttributionData` to `OverviewKpiData` (commit `d7bdd710`).
- **1.3 (minor)** — Playwright tests noted as follow-up; `data-testid` attributes are in place for when tests are added.

All threads have been resolved. Ready for re-review when you have a chance.

dealako:
Hey Misha! Great turnaround on this — `d277ce56` came in fast and lands several things cleanly. Here's the full re-evaluation:

---

## 👏 Nice work

Beyond just addressing the requested fixes, you shipped a few extras that improve the whole PR:

- **Email CTR label + id corrected** (`email-open-rate` → `email-ctr`) — you'd previously marked this as a deliberate stakeholder decision, but the correction here is the right call and removes a confusing mismatch between label and data.
- **Helpers extracted to shared utils** — `trendDirection`, `trendColorClass`, `formatChangePct` moved into `marketing-impact.utils.ts`. These are exactly what PR #720 will need; doing it here means the next tab lands without duplication.
- **`@empty` block for the KPI grid** — graceful no-data state, zero cost to add at this stage.
- **Finite-number guard on ROAS** — `(ri.paidMedia?.roas ?? 0)` prevents the crash that would've happened if `paidMedia` came back null. Good defensive fix.
- **`momPct != null` guard on the badge** — `badge: momPct != null && momPct < 0` correctly prevents the badge from appearing on undefined change data.

---

## ✅ Resolved

**[blocking] 1.1 — `forkJoin` missing `catchError`**
Fixed exactly as suggested. Each member now has its own `catchError(() => of(null))`:
```ts
revenueImpact: this.analyticsService.getRevenueImpact(slug).pipe(catchError(() => of(null))),
brandReach:   this.analyticsService.getBrandReach(slug).pipe(catchError(() => of(null))),
emailCtr:     this.analyticsService.getEmailCtr(slug).pipe(catchError(() => of(null))),
```
A single upstream failure no longer takes down the whole tab. ✅

**[minor] 1.2 — `attributionData` field rename**
Renamed to `overviewKpiData` and the consumer at the former line 64 updated. Naming now matches the `OverviewKpiData` type throughout. ✅

---

## ❌ Still open

**[minor] 1.3 — No E2E test coverage for the attribution model selector or KPI cards**

The `data-testid` attributes are in place (`kpi-card-email-ctr`, `kpi-card-roas`, `attribution-model-select`, `overview-tab-kpi-empty` from this commit, etc.), but `apps/lfx-one/e2e/marketing-dashboard.spec.ts` still has zero matches for any of them. Given that PR #720 is coming and will layer more tabs on this scaffold, the test debt compounds with every PR.

A minimal happy-path Playwright spec (load → assert 4 KPI cards render, switch attribution model → assert top revenue row updates) plus one empty-state assertion is enough to break the streak.

**[minor] 1.5 — No tracking comment for the month-picker / data-window gap**

Neither the overview-tab `switchMap` (`overview-tab.component.ts:~49`) nor the attribution-section `switchMap` (`attribution-section.component.ts:~69`) has a comment documenting that the month picker drives labels only and not the underlying data window. The gap is acknowledged in the Copilot threads but invisible in the code. Future maintainers will rediscover it.

One line in each:
```ts
// TODO(LFXV2-XXXX): month picker drives title/subtitle only — backend needs a month-scoped endpoint
```

---

## 🔍 New observations

**[nit] `TrendDirection` type is not yet wired into `PerformanceSummaryKpi`**

`marketing-impact.utils.ts` now exports `type TrendDirection = 'up' | 'down' | 'neutral'`, but `PerformanceSummaryKpi.momTrend` and `.yoyTrend` in the interface still use the inline union:
```ts
momTrend: 'up' | 'down' | 'neutral';   // should be: momTrend: TrendDirection;
yoyTrend: 'up' | 'down' | 'neutral';
```
They're structurally identical so nothing breaks, but using `TrendDirection` in the interface closes the loop and means a future change to the union only needs one edit. Easy follow-up.

**[nit] ROAS null fallback shows `0.00x` instead of a missing-data indicator**

`(ri.paidMedia?.roas ?? 0).toFixed(2)x` — if `paidMedia` is null, this renders `0.00x`, which looks like real data (zero ROAS). A dash or skipping the card entirely would be more honest:
```ts
value: ri.paidMedia?.roas != null ? `${ri.paidMedia.roas.toFixed(2)}x` : '—',
```
Not a blocker, but worth flagging for correctness.

---

## Overall decision

Per my earlier commitment — *"Once 1.1 lands, plus either the field-rename (1.2) or an explicit decision to defer, I'm happy to approve with the remaining items as optional follow-ups"* — both are resolved, and this commit includes several nice extras on top. **Approving.**

The two still-open minors (1.3 tests, 1.5 TODO comment) and the two nits are tracked here for a follow-up PR or as part of PR #720's scope — your call on timing. The `TrendDirection` interface wiring and ROAS display are worth adding whenever you're next touching those files.

🟢 **Approved** — great work across the full review cycle, Misha.

dealako (approved):
Approving after re-evaluation of `d277ce56`. Both blocking items (1.1 `catchError`, 1.2 field rename) are resolved. Full follow-up notes in the PR comment above.

MRashad26 (approved):

PR #720, opened 2026-05-16
feat(dashboards): add email marketing tab

mrautela365:
## Summary
- Adds Email tab with KPI cards (Total Sends, Opens, Open Rate, CTR)
- Email type breakdown table (campaigns by type with open rate and CTR)
- Top 5 campaigns table sorted by sends
- Uses `getEmailCtr` API

**Depends on:** #719

## Test plan
- [ ] Select TLF, click Email tab — KPI cards render
- [ ] Verify email type breakdown table shows rows
- [ ] Verify top campaigns table shows up to 5 campaigns

LFXV2-1644

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
@@ -55,8 +55,8 @@
           <span class="flex-1 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CTR</span>
         </div>
 
-        @for (row of emailTypeRows(); track row.emailType) {
-          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" role="row" [attr.data-testid]="'email-type-row-' + row.emailType">
+        @for (row of emailTypeRows(); track row.emailType; let idx = $index) {
+          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" role="row" [attr.data-testid]="'email-type-row-' + idx">
             <span class="w-36 truncate text-xs font-medium text-gray-700" role="cell">{{ row.emailType }}</span>
             <span class="w-20 text-right text-xs text-gray-500" role="cell">{{ row.campaignCount }}</span>
             <span class="w-24 text-right text-xs text-gray-900" role="cell">{{ row.sends }}</span>
@@ -103,8 +103,8 @@
           <span class="w-16 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CTR</span>
         </div>
 
-        @for (row of topCampaigns(); track row.type + '|' + row.name) {
-          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" role="row" [attr.data-testid]="'email-campaign-row-' + row.name">
+        @for (row of topCampaigns(); track row.type + '|' + row.name; let idx = $index) {
+          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" role="row" [attr.data-testid]="'email-campaign-row-' + idx">
             <span class="flex-1 truncate text-xs font-medium text-gray-700" role="cell">{{ row.name }}</span>
             <span class="w-24 truncate text-[11px] text-gray-500" role="cell">{{ row.type }}</span>
             <span class="w-20 text-right text-xs text-gray-900" role="cell">{{ row.sends }}</span>
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
@@ -26,74 +26,98 @@
   </div>
 
   <!-- Email type breakdown table -->
-  @if (!loading()) {
-    <div class="flex flex-col gap-3 rounded-lg border border-gray-200 bg-white p-5" data-testid="email-type-breakdown">
-      <h4 class="text-sm font-semibold text-gray-900">Performance by email type</h4>
+  <div class="flex flex-col gap-3 rounded-lg border border-gray-200 bg-white p-5" data-testid="email-type-breakdown">
+    <h4 class="text-sm font-semibold text-gray-900">Performance by email type</h4>
 
-      @if (hasEmailTypes()) {
-        <div class="flex flex-col gap-0" data-testid="email-type-table">
-          <!-- Table header -->
-          <div class="flex items-center gap-3 border-b border-gray-200 pb-2">
-            <span class="w-36 text-[11px] font-semibold tracking-wide text-gray-400">TYPE</span>
-            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">CAMPAIGNS</span>
-            <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400">SENDS</span>
-            <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPENS</span>
-            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPEN RATE</span>
-            <span class="flex-1 text-right text-[11px] font-semibold tracking-wide text-gray-400">CTR</span>
-          </div>
-
-          <!-- Rows -->
-          @for (row of emailTypeRows(); track row.emailType; let idx = $index) {
-            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-type-row-' + idx">
-              <span class="w-36 truncate text-xs font-medium text-gray-700">{{ row.emailType }}</span>
-              <span class="w-20 text-right text-xs text-gray-500">{{ row.campaignCount }}</span>
-              <span class="w-24 text-right text-xs text-gray-900">{{ row.sends }}</span>
-              <span class="w-24 text-right text-xs text-gray-900">{{ row.opens }}</span>
-              <span class="w-20 text-right text-xs text-gray-500">{{ row.openRate }}</span>
-              <span class="flex-1 text-right text-xs font-medium text-gray-900">{{ row.ctr }}</span>
-            </div>
+    @if (loading()) {
+      <div class="flex flex-col gap-0" data-testid="email-type-skeleton">
+        <div class="flex items-center gap-3 border-b border-gray-200 pb-2">
+          @for (i of [1, 2, 3, 4, 5, 6]; track i) {
+            <div class="h-3 w-20 animate-pulse rounded bg-gray-200"></div>
           }
         </div>
-      } @else {
-        <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="email-type-empty">
-          <p class="text-sm text-gray-500">No email type data available.</p>
-        </div>
-      }
-    </div>
-
-    <!-- Top campaigns table -->
-    <div class="flex flex-col gap-3 rounded-lg border border-gray-200 bg-white p-5" data-testid="email-top-campaigns">
-      <h4 class="text-sm font-semibold text-gray-900">Top campaigns by sends</h4>
-
-      @if (hasTopCampaigns()) {
-        <div class="flex flex-col gap-0" data-testid="email-campaigns-table">
-          <!-- Table header -->
-          <div class="flex items-center gap-3 border-b border-gray-200 pb-2">
-            <span class="flex-1 text-[11px] font-semibold tracking-wide text-gray-400">CAMPAIGN</span>
-            <span class="w-24 text-[11px] font-semibold tracking-wide text-gray-400">TYPE</span>
-            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">SENDS</span>
-            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPENS</span>
-            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPEN RATE</span>
-            <span class="w-16 text-right text-[11px] font-semibold tracking-wide text-gray-400">CTR</span>
+        @for (i of [1, 2, 3]; track i) {
+          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5">
+            @for (j of [1, 2, 3, 4, 5, 6]; track j) {
+              <div class="h-3 w-20 animate-pulse rounded bg-gray-100"></div>
+            }
           </div>
+        }
+      </div>
+    } @else if (hasEmailTypes()) {
+      <div class="flex flex-col gap-0" role="table" aria-label="Performance by email type" data-testid="email-type-table">
+        <div class="flex items-center gap-3 border-b border-gray-200 pb-2" role="row">
+          <span class="w-36 text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">TYPE</span>
+          <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CAMPAIGNS</span>
+          <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">SENDS</span>
+          <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">OPENS</span>
+          <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">OPEN RATE</span>
+          <span class="flex-1 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CTR</span>
+        </div>
 
-          <!-- Rows -->
-          @for (row of topCampaigns(); track row.type + '|' + row.name; let idx = $index) {
-            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + idx">
-              <span class="flex-1 truncate text-xs font-medium text-gray-700">{{ row.name }}</span>
-              <span class="w-24 truncate text-[11px] text-gray-500">{{ row.type }}</span>
-              <span class="w-20 text-right text-xs text-gray-900">{{ row.sends }}</span>
-              <span class="w-20 text-right text-xs text-gray-900">{{ row.opens }}</span>
-              <span class="w-20 text-right text-xs text-gray-500">{{ row.openRate }}</span>
-              <span class="w-16 text-right text-xs font-medium text-gray-900">{{ row.ctr }}</span>
-            </div>
+        @for (row of emailTypeRows(); track row.emailType) {
+          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" role="row" [attr.data-testid]="'email-type-row-' + row.emailType">
+            <span class="w-36 truncate text-xs font-medium text-gray-700" role="cell">{{ row.emailType }}</span>
+            <span class="w-20 text-right text-xs text-gray-500" role="cell">{{ row.campaignCount }}</span>
+            <span class="w-24 text-right text-xs text-gray-900" role="cell">{{ row.sends }}</span>
+            <span class="w-24 text-right text-xs text-gray-900" role="cell">{{ row.opens }}</span>
+            <span class="w-20 text-right text-xs text-gray-500" role="cell">{{ row.openRate }}</span>
+            <span class="flex-1 text-right text-xs font-medium text-gray-900" role="cell">{{ row.ctr }}</span>
+          </div>
+        }
+      </div>
+    } @else {
+      <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="email-type-empty">
+        <p class="text-sm text-gray-500">No email type data available.</p>
+      </div>
+    }
+  </div>
+
+  <!-- Top campaigns table -->
+  <div class="flex flex-col gap-3 rounded-lg border border-gray-200 bg-white p-5" data-testid="email-top-campaigns">
+    <h4 class="text-sm font-semibold text-gray-900">Top campaigns by sends</h4>
+
+    @if (loading()) {
+      <div class="flex flex-col gap-0" data-testid="email-campaigns-skeleton">
+        <div class="flex items-center gap-3 border-b border-gray-200 pb-2">
+          @for (i of [1, 2, 3, 4, 5, 6]; track i) {
+            <div class="h-3 w-20 animate-pulse rounded bg-gray-200"></div>
           }
         </div>
-      } @else {
-        <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="email-campaigns-empty">
-          <p class="text-sm text-gray-500">No campaign data available.</p>
+        @for (i of [1, 2, 3]; track i) {
+          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5">
+            @for (j of [1, 2, 3, 4, 5, 6]; track j) {
+              <div class="h-3 w-20 animate-pulse rounded bg-gray-100"></div>
+            }
+          </div>
+        }
+      </div>
+    } @else if (hasTopCampaigns()) {
+      <div class="flex flex-col gap-0" role="table" aria-label="Top campaigns by sends" data-testid="email-campaigns-table">
+        <div class="flex items-center gap-3 border-b border-gray-200 pb-2" role="row">
+          <span class="flex-1 text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CAMPAIGN</span>
+          <span class="w-24 text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">TYPE</span>
+          <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">SENDS</span>
+          <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">OPENS</span>
+          <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">OPEN RATE</span>
+          <span class="w-16 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CTR</span>
         </div>
-      }
-    </div>
-  }
+
+        @for (row of topCampaigns(); track row.type + '|' + row.name) {
+          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" role="row" [attr.data-testid]="'email-campaign-row-' + row.name">
+            <span class="flex-1 truncate text-xs font-medium text-gray-700" role="cell">{{ row.name }}</span>
+            <span class="w-24 truncate text-[11px] text-gray-500" role="cell">{{ row.type }}</span>
+            <span class="w-20 text-right text-xs text-gray-900" role="cell">{{ row.sends }}</span>
+            <span class="w-20 text-right text-xs text-gray-900" role="cell">{{ row.opens }}</span>
+            <span class="w-20 text-right text-xs text-gray-500" role="cell">{{ row.openRate }}</span>
+            <span class="w-16 text-right text-xs font-medium text-gray-900" role="cell">{{ row.ctr }}</span>
+          </div>
+        }
+      </div>
+    } @else {
+      <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="email-campaigns-empty">
+        <p class="text-sm text-gray-500">No campaign data available.</p>
+      </div>
+    }
+  </div>
 </div>
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
@@ -43,8 +43,8 @@
           </div>
 
           <!-- Rows -->
-          @for (row of emailTypeRows(); track row.emailType; let idx = $index) {
-            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-type-row-' + idx">
+          @for (row of emailTypeRows(); track row.emailType) {
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-type-row-' + row.emailType">
               <span class="w-36 truncate text-xs font-medium text-gray-700">{{ row.emailType }}</span>
               <span class="w-20 text-right text-xs text-gray-500">{{ row.campaignCount }}</span>
               <span class="w-24 text-right text-xs text-gray-900">{{ row.sends }}</span>
@@ -78,8 +78,8 @@
           </div>
 
           <!-- Rows -->
-          @for (row of topCampaigns(); track row.type + '|' + row.name; let idx = $index) {
-            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + idx">
+          @for (row of topCampaigns(); track row.name) {
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + row.name">
               <span class="flex-1 truncate text-xs font-medium text-gray-700">{{ row.name }}</span>
               <span class="w-24 truncate text-[11px] text-gray-500">{{ row.type }}</span>
               <span class="w-20 text-right text-xs text-gray-900">{{ row.sends }}</span>
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
@@ -78,7 +78,7 @@
           </div>
 
           <!-- Rows -->
-          @for (row of topCampaigns(); track row.name) {
+          @for (row of topCampaigns(); track row.type + '|' + row.name) {
             <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + row.name">
               <span class="flex-1 truncate text-xs font-medium text-gray-700">{{ row.name }}</span>
               <span class="w-24 truncate text-[11px] text-gray-500">{{ row.type }}</span>
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
@@ -43,8 +43,8 @@
           </div>
 
           <!-- Rows -->
-          @for (row of emailTypeRows(); track row.emailType) {
-            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-type-row-' + row.emailType">
+          @for (row of emailTypeRows(); track row.emailType; let idx = $index) {
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-type-row-' + idx">
               <span class="w-36 truncate text-xs font-medium text-gray-700">{{ row.emailType }}</span>
               <span class="w-20 text-right text-xs text-gray-500">{{ row.campaignCount }}</span>
               <span class="w-24 text-right text-xs text-gray-900">{{ row.sends }}</span>
@@ -78,8 +78,8 @@
           </div>
 
           <!-- Rows -->
-          @for (row of topCampaigns(); track row.type + '|' + row.name) {
-            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + row.name">
+          @for (row of topCampaigns(); track row.type + '|' + row.name; let idx = $index) {
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + idx">
               <span class="flex-1 truncate text-xs font-medium text-gray-700">{{ row.name }}</span>
               <span class="w-24 truncate text-[11px] text-gray-500">{{ row.type }}</span>
               <span class="w-20 text-right text-xs text-gray-900">{{ row.sends }}</span>
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
@@ -3,9 +3,9 @@
 
 import { Component, computed, inject, input, signal, Signal } from '@angular/core';
 import { toObservable, toSignal } from '@angular/core/rxjs-interop';
-import { formatChangePct, formatNumber, trendColorClass, trendDirection } from '@lfx-one/shared/utils';
+import { formatNumber } from '@lfx-one/shared/utils';
 import { AnalyticsService } from '@services/analytics.service';
-import { catchError, finalize, of, switchMap } from 'rxjs';
+import { catchError, of, switchMap, tap } from 'rxjs';
 
 import type { EmailCtrResponse, EmailTypeRow, PerformanceSummaryKpi, TopCampaignRow } from '@lfx-one/shared/interfaces';
 
@@ -49,8 +49,11 @@ export class EmailTabComponent {
           }
           this.loading.set(true);
           return this.analyticsService.getEmailCtr(slug).pipe(
-            finalize(() => this.loading.set(false)),
-            catchError(() => of(null))
+            tap(() => this.loading.set(false)),
+            catchError(() => {
+              this.loading.set(false);
+              return of(null);
+            })
           );
         })
       ),
@@ -65,21 +68,8 @@ export class EmailTabComponent {
 
       const totalSends = data.monthlySends?.reduce((s, v) => s + v, 0) ?? 0;
       const totalOpens = data.monthlyOpens?.reduce((s, v) => s + v, 0) ?? 0;
+      const openRate = totalSends > 0 ? (totalOpens / totalSends) * 100 : 0;
       const changePct = data.changePercentage;
-
-      const sendsMom = this.computeMomPct(data.monthlySends);
-      const opensMom = this.computeMomPct(data.monthlyOpens);
-
-      const sends = data.monthlySends ?? [];
-      const opens = data.monthlyOpens ?? [];
-      const lastSends = sends.length > 0 ? sends[sends.length - 1] : undefined;
-      const prevSends = sends.length > 1 ? sends[sends.length - 2] : undefined;
-      const lastOpens = sends.length > 0 ? (opens[opens.length - 1] ?? 0) : 0;
-      const prevOpens = sends.length > 1 ? (opens[opens.length - 2] ?? 0) : 0;
-
-      const currentOpenRate = lastSends !== undefined && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
-      const prevOpenRate = prevSends !== undefined && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
-      const openRateMom = prevOpenRate > 0 ? ((currentOpenRate - prevOpenRate) / prevOpenRate) * 100 : null;
 
       return [
         {
@@ -88,9 +78,9 @@ export class EmailTabComponent {
           icon: 'fa-light fa-paper-plane',
           iconClass: 'bg-blue-100 text-blue-600',
           value: formatNumber(totalSends),
-          momChange: formatChangePct(sendsMom, 'MoM'),
-          momTrend: trendDirection(sendsMom),
-          momTrendClass: trendColorClass(sendsMom),
+          momChange: null,
+          momTrend: 'neutral' as const,
+          momTrendClass: 'text-gray-500',
           yoyChange: null,
           yoyTrend: 'neutral' as const,
           yoyTrendClass: 'text-gray-500',
@@ -102,9 +92,9 @@ export class EmailTabComponent {
           icon: 'fa-light fa-envelope-open',
           iconClass: 'bg-green-100 text-green-600',
           value: formatNumber(totalOpens),
-          momChange: formatChangePct(opensMom, 'MoM'),
-          momTrend: trendDirection(opensMom),
-          momTrendClass: trendColorClass(opensMom),
+          momChange: null,
+          momTrend: 'neutral' as const,
+          momTrendClass: 'text-gray-500',
           yoyChange: null,
           yoyTrend: 'neutral' as const,
           yoyTrendClass: 'text-gray-500',
@@ -115,10 +105,10 @@ export class EmailTabComponent {
           label: 'Open Rate',
           icon: 'fa-light fa-chart-simple',
           iconClass: 'bg-amber-100 text-amber-600',
-          value: `${currentOpenRate.toFixed(1)}%`,
-          momChange: formatChangePct(openRateMom, 'MoM'),
-          momTrend: trendDirection(openRateMom),
-          momTrendClass: trendColorClass(openRateMom),
+          value: `${openRate.toFixed(1)}%`,
+          momChange: null,
+          momTrend: 'neutral' as const,
+          momTrendClass: 'text-gray-500',
           yoyChange: null,
           yoyTrend: 'neutral' as const,
           yoyTrendClass: 'text-gray-500',
@@ -130,9 +120,9 @@ export class EmailTabComponent {
           icon: 'fa-light fa-arrow-pointer',
           iconClass: 'bg-violet-100 text-violet-600',
           value: `${data.currentCtr.toFixed(2)}%`,
-          momChange: formatChangePct(changePct, 'vs avg'),
-          momTrend: trendDirection(changePct),
-          momTrendClass: trendColorClass(changePct),
+          momChange: this.formatChangePct(changePct, 'MoM'),
+          momTrend: this.trendDirection(changePct),
+          momTrendClass: this.trendColorClass(changePct),
           yoyChange: null,
           yoyTrend: 'neutral' as const,
           yoyTrendClass: 'text-gray-500',
@@ -183,11 +173,20 @@ export class EmailTabComponent {
   }
 
   // === Private Helpers ===
-  private computeMomPct(arr: number[] | undefined): number | null {
-    if (!arr || arr.length < 2) return null;
-    const current = arr.at(-1) ?? 0;
-    const previous = arr.at(-2) ?? 0;
-    if (previous === 0) return null;
-    return ((current - previous) / previous) * 100;
+  private trendDirection(pct: number): 'up' | 'down' | 'neutral' {
+    if (pct > 0) return 'up';
+    if (pct < 0) return 'down';
+    return 'neutral';
+  }
+
+  private trendColorClass(pct: number): string {
+    if (pct > 0) return 'text-green-600';
+    if (pct < 0) return 'text-red-600';
+    return 'text-gray-500';
+  }
+
+  private formatChangePct(pct: number, suffix: string): string {
+    const sign = pct > 0 ? '+' : '';
+    return `${sign}${pct.toFixed(1)}% ${suffix}`;
   }
 }
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
@@ -78,8 +78,8 @@ export class EmailTabComponent {
       const lastOpens = minLen > 0 ? (opens[minLen - 1] ?? 0) : 0;
       const prevOpens = minLen > 1 ? (opens[minLen - 2] ?? 0) : 0;
 
-      const currentOpenRate = lastSends && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
-      const prevOpenRate = prevSends && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
+      const currentOpenRate = lastSends !== undefined && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
+      const prevOpenRate = prevSends !== undefined && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
       const openRateMom = this.computeMomPctFromValues(currentOpenRate, prevOpenRate);
 
       return [
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
@@ -72,14 +72,15 @@ export class EmailTabComponent {
 
       const sends = data.monthlySends ?? [];
       const opens = data.monthlyOpens ?? [];
-      const lastSends = sends.length > 0 ? sends[sends.length - 1] : undefined;
-      const prevSends = sends.length > 1 ? sends[sends.length - 2] : undefined;
-      const lastOpens = sends.length > 0 ? (opens[opens.length - 1] ?? 0) : 0;
-      const prevOpens = sends.length > 1 ? (opens[opens.length - 2] ?? 0) : 0;
+      const minLen = Math.min(sends.length, opens.length);
+      const lastSends = minLen > 0 ? sends[minLen - 1] : undefined;
+      const prevSends = minLen > 1 ? sends[minLen - 2] : undefined;
+      const lastOpens = minLen > 0 ? (opens[minLen - 1] ?? 0) : 0;
+      const prevOpens = minLen > 1 ? (opens[minLen - 2] ?? 0) : 0;
 
-      const currentOpenRate = lastSends !== undefined && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
-      const prevOpenRate = prevSends !== undefined && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
-      const openRateMom = prevOpenRate > 0 ? ((currentOpenRate - prevOpenRate) / prevOpenRate) * 100 : null;
+      const currentOpenRate = lastSends && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
+      const prevOpenRate = prevSends && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
+      const openRateMom = this.computeMomPctFromValues(currentOpenRate, prevOpenRate);
 
       return [
         {
@@ -185,8 +186,10 @@ export class EmailTabComponent {
   // === Private Helpers ===
   private computeMomPct(arr: number[] | undefined): number | null {
     if (!arr || arr.length < 2) return null;
-    const current = arr.at(-1) ?? 0;
-    const previous = arr.at(-2) ?? 0;
+    return this.computeMomPctFromValues(arr.at(-1) ?? 0, arr.at(-2) ?? 0);
+  }
+
+  private computeMomPctFromValues(current: number, previous: number): number | null {
     if (previous === 0) return null;
     return ((current - previous) / previous) * 100;
   }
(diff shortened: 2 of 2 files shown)


apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,192 @@
...
+  private trendColorClass(pct: number): string {
+    if (pct > 0) return 'text-green-600';
+    if (pct < 0) return 'text-red-600';
+    return 'text-gray-500';
+  }
+
+  private formatChangePct(pct: number, suffix: string): string {
+    const sign = pct > 0 ? '+' : '';
+    return `${sign}${pct.toFixed(1)}% ${suffix}`;
+  }
```
MRashad26, line 195: **Problem:** `trendDirection`, `trendColorClass`, and `formatChangePct` are duplicated **verbatim** across three components:
- `email-tab.component.ts:176-191`
- `overview-tab.component.ts:174-189`
- `performance-marketing-tab.component.ts:179-194`

**Why it is a problem:** Three byte-identical 17-line blocks of private helpers. Any change to trend-up/down/neutral semantics, color palette, or `+`/`%` formatting now has to be replicated in three places, and any one of them can drift. CLAUDE.md / general engineering: "three similar lines is better than a premature abstraction" — but **three identical 17-line blocks is the abstraction screaming to come out**. This is also exactly the kind of cross-cutting view-model formatting that belongs in `@lfx-one/shared/utils` alongside `formatCurrency` / `formatNumber`, which these helpers already sit next to.

**Suggested fix:** Hoist the three helpers into `packages/shared/src/utils/marketing-impact.utils.ts` (or extend the existing util file you already use for `formatCurrency` / `formatNumber`):
```ts
export function trendDirection(pct: number): 'up' | 'down' | 'neutral' {
  if (pct > 0) return 'up';
  if (pct < 0) return 'down';
  return 'neutral';
}
export function trendColorClass(pct: number): string {
  if (pct > 0) return 'text-green-600';
  if (pct < 0) return 'text-red-600';
  return 'text-gray-500';
}
export function formatChangePct(pct: number, suffix: string): string {
  const sign = pct > 0 ? '+' : '';
  return `${sign}${pct.toFixed(1)}% ${suffix}`;
}
```
Then drop the `// === Private Helpers ===` blocks from all three components and import the utilities instead. Same fix removes ~50 lines of duplication immediately.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,192 @@
...
+            catchError(() => {
+              this.loading.set(false);
+              return of(null);
+            })
+          );
+        })
+      ),
+      { initialValue: null }
+    );
+  }
```
MRashad26, line 59: **Problem:** The `init<Data>()` pattern — `toObservable(slug).pipe(switchMap(slug → null-or-loading-then-HTTP-with-tap-and-catchError))` — is duplicated **four times** with the same structure:
- `email-tab.component.ts:40-62` (`initEmailData`)
- `overview-tab.component.ts:40-60` (`initAttributionData`, uses `forkJoin` but same wrapper)
- `performance-marketing-tab.component.ts:51-73` (`initSocialReachData`)
- `attribution-section.component.ts:56-78` (`initAttributionData`)

In three of the four, the `loading.set(false)` call is repeated **three times per method** (no-slug branch, `tap` on success, `catchError`) — a known foot-gun if anyone forgets one.

**Why it is a problem:** Every additional KPI tab will copy this same 20-line scaffold, and the triple `loading.set(false)` makes it easy to introduce a stuck-spinner bug. The shape is identical enough to abstract.

**Suggested fix:** Extract a small helper that takes the slug input signal, a fetcher function, an initial value, and the loading signal:
```ts
// in shared/utils or a local component util
export function trackedLoad<T>(
  slug$: Observable<string | undefined>,
  fetcher: (slug: string) => Observable<T>,
  initial: T,
  loading: WritableSignal<boolean>,
): Signal<T> {
  return toSignal(
    slug$.pipe(
      tap(() => loading.set(true)),
      switchMap((slug) => {
        if (!slug) { loading.set(false); return of(initial); }
        return fetcher(slug).pipe(
          catchError(() => of(initial)),
          finalize(() => loading.set(false)),
        );
      }),
    ),
    { initialValue: initial },
  );
}
```
Using `finalize` collapses the three `loading.set(false)` calls into one and removes the stuck-spinner footgun. Each tab then has a one-line init: `return trackedLoad(toObservable(this.foundationSlug), (s) => this.analyticsService.getEmailCtr(s), null, this.loading);`

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+  <!-- KPI Cards -->
+  <div class="grid grid-cols-1 gap-4 sm:grid-cols-2 lg:grid-cols-4" data-testid="email-kpi-grid">
+    @if (loading()) {
+      @for (i of [1, 2, 3, 4]; track i) {
+        <div class="flex flex-col gap-3 rounded-lg border border-gray-200 bg-white p-4" data-testid="email-kpi-skeleton">
+          <div class="h-4 w-24 animate-pulse rounded bg-gray-200"></div>
+          <div class="h-8 w-32 animate-pulse rounded bg-gray-200"></div>
+          <div class="h-3 w-40 animate-pulse rounded bg-gray-200"></div>
+        </div>
+      }
```
MRashad26, line 20: **Problem:** The 4-card skeleton loading state is duplicated in three templates:
- `email-tab.component.html:13-20`
- `overview-tab.component.html:23-31`
- `performance-marketing-tab.component.html:25-32`

Each uses essentially the same `@for (i of [1, 2, 3, 4]; track i) { <div class="flex flex-col gap-3 rounded-lg border border-gray-200 bg-white p-4">…skeleton bars…</div> }` markup. The overview version adds a 4th bar; otherwise the markup is identical.

**Why it is a problem:** Visual consistency drifts the moment any one tab tweaks the skeleton (heights, gaps, count). And like the helpers and init patterns, this is the third copy of a thing — strong signal it should be extracted.

**Suggested fix:** Introduce `<lfx-kpi-card-skeleton [count]="4" [extraBar]="false" />` (or accept the count via input) co-located with `sparkline-kpi-card`, and replace all three skeleton blocks with one line:
```html
<lfx-kpi-card-skeleton [count]="4" />
```
Keeps the skeleton look in lockstep with the actual cards and removes ~25 lines of duplicated markup.

packages/shared/src/constants/marketing-impact.constants.ts
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 import { FilterPillOption } from '../interfaces/dashboard-metric.interface';
```
MRashad26: **Problem:** `FilterPillOption` is imported as a value (`import { FilterPillOption }`) on line 4, but it is only used as a type annotation (`MARKETING_IMPACT_FOCUS_OPTIONS: FilterPillOption[]`, `FUNNEL_STAGE_OPTIONS: FilterPillOption[]`).

**Why it is a problem:** TypeScript best practice (and project convention — note line 5 correctly uses `import type` for `AttributionModelOption` / `MarketingImpactTabOption`) is to use `import type` for type-only imports so the import is fully erased at runtime and never accidentally creates a value reference.

**Suggested fix:**
```ts
import type { FilterPillOption } from '../interfaces/dashboard-metric.interface';
```

MRashad26 (commented):
CLAUDE.md / code-enforcer-agent audit — 3 major DRY findings + 2 minor nits. Functional bugs (label mismatch, missing PERF rows, track-key collision) already raised by CodeRabbit; not duplicating those here.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,190 @@
...
+
+      if (data.emailCtr) {
+        const ec = data.emailCtr;
+        const momPct = ec.changePercentage;
+        cards.push({
+          id: 'email-open-rate',
+          label: 'Email Open Rate',
+          icon: 'fa-light fa-envelope-open',
+          iconClass: 'bg-amber-100 text-amber-600',
+          value: `${ec.currentCtr.toFixed(2)}%`,
```
dealako, line 126: [blocking] **Email Open Rate card displays CTR, not open rate**

The card is labeled `'Email Open Rate'` (line 126) but `value` is bound to `ec.currentCtr` — the click-through rate field on `EmailCtrResponse` — not the open rate. Open rate should be computed from the available data: `(monthlyOpens.at(-1) / monthlySends.at(-1)) * 100` or sourced from a dedicated upstream field.

Agrees with CodeRabbit ([comment 3253346026](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253346026)).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
```diff
@@ -0,0 +1,40 @@
...
+      }
+    } @else {
+      @for (kpi of performanceSummaryKpis(); track kpi.id) {
+        <lfx-sparkline-kpi-card [kpi]="kpi" />
+      }
+    }
+  </div>
+</div>
+
+<lfx-attribution-section [foundationSlug]="foundationSlug()" [foundationName]="foundationName()" data-testid="overview-tab-attribution-section" />
```
dealako, line 44: [blocking] **Duplicate `getMarketingAttribution` HTTP calls**

`AttributionSectionComponent` is mounted unconditionally here (inside the Overview tab template) and again inside `marketing-impact.component.html:82` when the Attribution tab is selected. Both instances independently call `getMarketingAttribution(slug)` — two concurrent fetches for the same data on every initial page load.

Fix options:
- Lift the fetch into a shared service with a per-slug cache, **or**
- Hoist `AttributionSectionComponent` to a single mount point in the parent and toggle CSS `display` rather than structural instantiation via `@if`.

Agrees with Copilot ([comment 3253344987](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253344987)).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,192 @@
...
+      const changePct = data.changePercentage;
+
+      return [
+        {
+          id: 'total-sends',
+          label: 'Total Sends',
+          icon: 'fa-light fa-paper-plane',
+          iconClass: 'bg-blue-100 text-blue-600',
+          value: formatNumber(totalSends),
+          momChange: null,
```
dealako: [minor] **Total Sends / Total Opens / Open Rate KPIs hardcode `momChange: null`**

`monthlySends` and `monthlyOpens` arrays are already available in `EmailCtrResponse` — the same fields used to compute the card display values on lines 69–71. MoM% can be derived from the last two entries: `(arr.at(-1) - arr.at(-2)) / arr.at(-2) * 100`. The CTR card (line 123) already wires `changePercentage` into the trend indicators — the other three cards should follow the same pattern for consistency.

Not raised by any bot.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,192 @@
...
+            opens: formatNumber(c.opens),
+            openRate: `${c.openRate.toFixed(1)}%`,
+            ctr: `${c.ctr.toFixed(2)}%`,
+          })
+        );
+    });
+  }
+
+  // === Private Helpers ===
+  private trendDirection(pct: number): 'up' | 'down' | 'neutral' {
```
dealako: [minor] **Trend helpers duplicated verbatim in 3 components**

`trendDirection`, `trendColorClass`, and `formatChangePct` are byte-for-byte identical here, in `overview-tab.component.ts:174`, and in `performance-marketing-tab.component.ts:179`. Extract to a shared util (e.g. `@lfx-one/shared/utils/trend-helpers.ts`) or a small standalone Angular pipe to eliminate the triplicate.

Agrees with Copilot ([3253344971](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253344971)) and MRashad26 ([3253377722](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253377722)).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,192 @@
...
+  // === Computed Signals ===
+  protected readonly emailData: Signal<EmailCtrResponse | null> = this.initEmailData();
+  protected readonly kpiCards: Signal<PerformanceSummaryKpi[]> = this.initKpiCards();
+  protected readonly emailTypeRows: Signal<EmailTypeRow[]> = this.initEmailTypeRows();
+  protected readonly hasEmailTypes = computed(() => this.emailTypeRows().length > 0);
+  protected readonly topCampaigns: Signal<TopCampaignRow[]> = this.initTopCampaigns();
+  protected readonly hasTopCampaigns = computed(() => this.topCampaigns().length > 0);
+
+  // === Private Initializers ===
+  private initEmailData(): Signal<EmailCtrResponse | null> {
```
dealako, line 40: [nit] **`initEmailData()` slug-driven pattern repeated 4 times**

The `toSignal(toObservable(slug).pipe(switchMap → service → tap/catchError))` idiom appears verbatim in `email-tab`, `overview-tab`, `performance-marketing-tab`, and `attribution-section`. A generic factory function (`createSlugSignal<T>(slug, fetcher, loadingSignal)`) in a shared location would reduce boilerplate. Not blocking — the convention is consistent and readable as-is.

Agrees with MRashad26 ([3253377727](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253377727)).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+  <!-- Header -->
+  <div class="flex flex-col gap-1">
+    <h3 class="text-base font-semibold text-gray-900">Email marketing</h3>
+    <p class="text-xs text-gray-500">Email engagement metrics · {{ foundationName() || 'All foundations' }}</p>
+  </div>
+
+  <!-- KPI Cards -->
+  <div class="grid grid-cols-1 gap-4 sm:grid-cols-2 lg:grid-cols-4" data-testid="email-kpi-grid">
+    @if (loading()) {
+      @for (i of [1, 2, 3, 4]; track i) {
```
dealako, line 14: [nit] **4-card skeleton block duplicated in 3 templates**

The `@for (i of [1,2,3,4]; track i)` loading skeleton grid appears identically in `overview-tab`, `performance-marketing-tab`, and `email-tab` templates. A `<lfx-kpi-skeleton-grid [count]="4" />` wrapper component would eliminate the duplication.

Agrees with MRashad26 ([3253377730](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253377730)).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+            <span class="flex-1 text-[11px] font-semibold tracking-wide text-gray-400">CAMPAIGN</span>
+            <span class="w-24 text-[11px] font-semibold tracking-wide text-gray-400">TYPE</span>
+            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">SENDS</span>
+            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPENS</span>
+            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPEN RATE</span>
+            <span class="w-16 text-right text-[11px] font-semibold tracking-wide text-gray-400">CTR</span>
+          </div>
+
+          <!-- Rows -->
+          @for (row of topCampaigns(); track row.name) {
```
dealako: [minor] **`@for track row.name` risks collisions; raw-name `data-testid` is fragile**

Campaign names are not guaranteed unique across email types — the same name can appear for multiple rows, breaking `@for`'s identity tracking and causing stale view recycling. Additionally, embedding `row.name` directly into `data-testid` produces selectors that break when names contain spaces, slashes, or quote characters.

Prefer tracking by a stable upstream ID (`row.campaignId` or similar). Sanitize `data-testid` values with a `slugify` helper if a name-based selector is required.

Agrees with Copilot ([3253344982](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253344982)) and CodeRabbit ([3253346022](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253346022)).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/performance-marketing-tab/performance-marketing-tab.component.ts
```diff
@@ -0,0 +1,206 @@
...
+    return computed(() => {
+      const data = this.socialReachData();
+      const funnel = this.selectedFunnel();
+      if (!data?.projectBreakdown?.length) return [];
+
+      return data.projectBreakdown
+        .filter((p) => {
+          if (funnel === 'all') return true;
+          const stage = p.funnelStage?.toLowerCase() ?? '';
+          if (funnel === 'tofu') return stage.startsWith('tofu') || stage === 'unknown';
```
dealako: [minor] **TOFU filter silently buckets `unknown` funnel stages**

`stage === 'unknown'` is included in the TOFU predicate, so projects with unclassified funnel data appear under TOFU rather than being excluded or shown in their own bucket. This inflates TOFU counts and masks data quality gaps from the consumer. Consider treating `unknown` as a distinct bucket or explicitly excluding it from all funnel views.

Agrees with Copilot ([3253344966](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253344966)).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,190 @@
...
+  imports: [ButtonComponent, SparklineKpiCardComponent, AttributionSectionComponent],
+  templateUrl: './overview-tab.component.html',
+})
+export class OverviewTabComponent {
+  // === Services ===
+  private readonly analyticsService = inject(AnalyticsService);
+
+  // === Inputs ===
+  public readonly foundationSlug = input<string | undefined>();
+  public readonly selectedMonth = input.required<string>();
```
dealako, line 27: [minor] **`selectedMonth` input is not wired to any API call**

`selectedMonth` flows into `summaryTitle` and `summarySubtitle` labels only (via `initSummaryTitle` / `initSummarySubtitle`). The `forkJoin` in `initAttributionData` calls `getRevenueImpact`, `getBrandReach`, and `getEmailCtr` without it, so the data displayed won't correspond to the chosen period when a user selects a non-current month.

Either pass `selectedMonth` through to the service calls, or remove the input until upstream endpoints support per-month filtering.

Agrees with Copilot ([3253344959](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253344959)).

packages/shared/src/constants/marketing-impact.constants.ts
dealako: [nit] **`FilterPillOption` should be a type-only import**

`FilterPillOption` is used only as a TypeScript type annotation here, not as a runtime value. Change to `import type { FilterPillOption }` so the constants file produces no runtime side-effect from this import.

Agrees with MRashad26 ([3253377732](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253377732)).

dealako (changes requested):
## PR Review — feat(dashboards): add email marketing tab

### Disagreement with CodeRabbit [3253346028](https://github.com/linuxfoundation/lfx-self-serve/pull/720#discussion_r3253346028)

CodeRabbit flagged that PERF values are hidden for rows where `getPerformanceClass()` returns an empty string. This is intentional — commit `ceceaf2f` documents the decision explicitly: *"hide negative performance badges in project table … to avoid raising unnecessary alarm."* Only EXCELLENT and GOOD badges are rendered. No change needed.

---

### Summary

Hey **David** — really solid work landing the email tab. The structure follows the established marketing-impact pattern faithfully, `SparklineKpiCardComponent` is a clean extraction, and the interface/constants split correctly lands in `@lfx-one/shared`. Two functional items need addressing before merge: the Overview tab's "Email Open Rate" card is wired to the CTR field (B1), and `AttributionSectionComponent` instantiates twice on initial load causing duplicate HTTP calls (B2). The minor items are mostly refactor opportunities that have surfaced as the four-tab surface area grew — good candidates for a follow-up pass.

**2 blocking · 5 minor · 3 nit**

**Decision: Needs changes before approval**

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.ts
```diff
@@ -2,39 +2,51 @@
...
+  imports: [
+    ReactiveFormsModule,
+    SelectComponent,
+    ButtonComponent,
+    FilterPillsComponent,
+    OverviewTabComponent,
+    AttributionSectionComponent,
+    PerformanceMarketingTabComponent,
+    EmailTabComponent,
+  ],
```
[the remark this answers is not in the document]
mrautela365: This is expected — this is a stacked PR. The overview, attribution, and performance-marketing tabs were introduced in prior PRs (#718, #719) and appear in this diff because the PR is based on those branches.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -0,0 +1,190 @@
...
+          return forkJoin({
+            revenueImpact: this.analyticsService.getRevenueImpact(slug).pipe(catchError(() => of(null as RevenueImpactResponse | null))),
+            brandReach: this.analyticsService.getBrandReach(slug).pipe(catchError(() => of(null as BrandReachResponse | null))),
+            emailCtr: this.analyticsService.getEmailCtr(slug).pipe(catchError(() => of(null as EmailCtrResponse | null))),
+          }).pipe(tap(() => this.loading.set(false)));
+        })
+      ),
+      { initialValue: { revenueImpact: null, brandReach: null, emailCtr: null } }
+    );
+  }
```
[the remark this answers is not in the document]
mrautela365: By design — the API does not yet support month filtering. `selectedMonth` drives display labels for now and will be wired when the backend adds month support.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/performance-marketing-tab/performance-marketing-tab.component.ts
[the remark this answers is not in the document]
mrautela365: Fixed in `0c1df366` — unknown stages are now filtered out entirely.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  private trendColorClass(pct: number): string {
+    if (pct > 0) return 'text-green-600';
+    if (pct < 0) return 'text-red-600';
+    return 'text-gray-500';
+  }
+
+  private formatChangePct(pct: number, suffix: string): string {
+    const sign = pct > 0 ? '+' : '';
+    return `${sign}${pct.toFixed(1)}% ${suffix}`;
```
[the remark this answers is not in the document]
mrautela365: Fixed in `0c1df366` — extracted to `@lfx-one/shared/utils`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + row.name">
```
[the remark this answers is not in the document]
mrautela365: Fixed in `e4875229` — email campaign rows now use composite track key.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -73,62 +73,21 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
-            }
-          }
-        </div>
+        <lfx-overview-tab
+          [foundationSlug]="foundationSlug()"
+          [selectedMonth]="selectedMonth()"
+          [foundationName]="foundationName()"
+          data-testid="marketing-impact-overview-tab" />
+      } @else if (selectedTab() === 'attribution') {
+        <lfx-attribution-section [foundationSlug]="foundationSlug()" [foundationName]="foundationName()" data-testid="marketing-impact-attribution-tab" />
```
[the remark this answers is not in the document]
mrautela365: By design — the Attribution tab mounts/unmounts independently with its tab, so there are no duplicate API calls at runtime.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/performance-marketing-tab/performance-marketing-tab.component.html
```diff
@@ -0,0 +1,84 @@
...
+              <span class="w-24 text-right text-xs text-gray-900">{{ row.spend }}</span>
+              <span class="w-24 text-right text-xs text-gray-900">{{ row.revenue }}</span>
+              <span class="w-16 text-right text-xs font-medium text-gray-900">{{ row.roas }}</span>
+              <span class="w-24 text-right text-xs text-gray-500">{{ row.impressions }}</span>
+              <div class="flex flex-1 justify-end">
+                @if (row.performanceClass) {
+                  <span class="rounded px-1.5 py-0.5 text-[10px] font-medium" [class]="row.performanceClass">
+                    {{ row.performance }}
+                  </span>
+                }
```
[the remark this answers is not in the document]
mrautela365: Fixed in `0c1df366` — fallback class ensures badge always renders.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
mrautela365: Fixed in `0c1df366` — extracted `trendDirection`, `trendColorClass`, and `formatChangePct` to `@lfx-one/shared/utils/marketing-impact.utils.ts`. All three components now import from there.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
mrautela365: Acknowledged — the pattern is structurally similar across tabs but each has different error handling, loading state, and response types. Extracting a generic factory would over-abstract at this stage. Will revisit if more tabs are added.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
mrautela365: Acknowledged — the skeleton grid varies by column count (3 vs 4 vs 5 cards) per tab, so a shared skeleton component would need multiple configs for minimal deduplication. Will address if the pattern grows further.

packages/shared/src/constants/marketing-impact.constants.ts
mrautela365: Fixed — addressed in the shared trend utils refactor (commit `0c1df366`). The import now uses type-only where appropriate.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
mrautela365: Fixed in `0c1df366` — renamed the card label to "Email CTR" to accurately reflect the underlying click-through-rate data.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
mrautela365: By design — the Attribution tab is behind `@if (selectedTab() === 'attribution')`, so it only mounts when active. It never fires simultaneously with the overview tab's attribution section.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
mrautela365: Fixed in `e4875229` — added a `computeMomPct` helper that derives MoM% from the last two months of the monthly data arrays. KPI cards now show actual month-over-month trends.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
mrautela365: Fixed in `0c1df366` — extracted to `@lfx-one/shared/utils/marketing-impact.utils.ts`. All three components now import from there.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
mrautela365: Same reasoning as above — structurally similar but each tab has distinct response types, error handling, and loading states. Extracting a generic factory would over-abstract at this stage. Will revisit if complexity grows.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
mrautela365: Same reasoning — column counts differ per tab (3 vs 4 vs 5 cards), so the deduplication benefit of a shared skeleton component is minimal. Will address if the pattern grows further.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
mrautela365: Fixed in `e4875229` — email campaign rows now use a composite track key (row.type + row.name) to avoid collisions.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/performance-marketing-tab/performance-marketing-tab.component.ts
mrautela365: Fixed in `0c1df366` — unknown stages are now filtered out entirely instead of being bucketed into TOFU.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
mrautela365: By design — the backend does not yet support month-level filtering. selectedMonth currently drives display labels and will be wired to the API when that support is added.

packages/shared/src/constants/marketing-impact.constants.ts
mrautela365: Fixed — addressed in the shared utils refactor (`0c1df366`). The import now uses type-only syntax where appropriate.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-type-row-' + row.emailType">
+              <span class="w-36 truncate text-xs font-medium text-gray-700">{{ row.emailType }}</span>
+              <span class="w-20 text-right text-xs text-gray-500">{{ row.campaignCount }}</span>
+              <span class="w-24 text-right text-xs text-gray-900">{{ row.sends }}</span>
+              <span class="w-24 text-right text-xs text-gray-900">{{ row.opens }}</span>
+              <span class="w-20 text-right text-xs text-gray-500">{{ row.openRate }}</span>
+              <span class="flex-1 text-right text-xs font-medium text-gray-900">{{ row.ctr }}</span>
+            </div>
+          }
+        </div>
```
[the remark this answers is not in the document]
mrautela365: Acknowledged — ARIA table roles will be added in a follow-up accessibility pass across all marketing impact tabs. The current flex-based layout is consistent across the feature set.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + row.name">
+              <span class="flex-1 truncate text-xs font-medium text-gray-700">{{ row.name }}</span>
+              <span class="w-24 truncate text-[11px] text-gray-500">{{ row.type }}</span>
+              <span class="w-20 text-right text-xs text-gray-900">{{ row.sends }}</span>
+              <span class="w-20 text-right text-xs text-gray-900">{{ row.opens }}</span>
+              <span class="w-20 text-right text-xs text-gray-500">{{ row.openRate }}</span>
+              <span class="w-16 text-right text-xs font-medium text-gray-900">{{ row.ctr }}</span>
+            </div>
+          }
+        </div>
```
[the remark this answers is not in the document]
mrautela365: Same as above — will address ARIA roles for the campaigns table in a follow-up accessibility pass.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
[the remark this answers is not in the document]
mrautela365: Fixed — the `@for` track key now uses `row.type + '|' + row.name` to avoid collisions when different campaign types share the same name.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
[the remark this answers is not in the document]
mrautela365: Fixed — card label renamed to "Email CTR" and id changed to `email-ctr` to accurately reflect the `currentCtr` data source.

dealako:
Hey David — great turnaround on the feedback! Most of the items landed cleanly. Here's the full pass:

---

## 👏 Nice work

- **Shared trend utils** (`marketing-impact.utils.ts`) is a solid extraction — the NaN/null guards on `trendDirection`, `trendColorClass`, and `formatChangePct` are a genuine improvement over the original private methods, not just a move.
- **MoM wiring on email KPIs** is well done — `computeMomPct` is clean and the `openRateMom` derivation handles the `prevOpenRate === 0` edge case correctly.
- **`finalize` swap** (`tap/catchError` → `finalize`) is more idiomatic and eliminates the duplicated `this.loading.set(false)` calls.
- **TOFU/unknown fix** — `if (stage === 'unknown') return false` before all funnel branches is exactly right.

---

## ✅ Resolved

| # | Finding | How |
|---|---------|-----|
| **B1** | Email Open Rate card showed CTR | Card renamed to "Email CTR" — label and value now match. Commit `daba3ae6`. |
| **M1** | Trend helpers duplicated 3× | Extracted to `packages/shared/src/utils/marketing-impact.utils.ts`, imported via `@lfx-one/shared/utils` in all three components. Commit `9217cf61`. |
| **M2** | TOFU filter silently bucketed `unknown` stages | `unknown` now explicitly excluded before all funnel branches. |
| **M3** | Email KPI MoM hardcoded `null` | `sendsMom`, `opensMom`, `openRateMom` computed from monthly arrays and wired into trend indicators. |
| **N3** | `FilterPillOption` value import | Changed to `import type`. |

---

## ⚠️ Partially addressed

**M4 — `@for` track + `data-testid`** (`email-tab.component.html:81`)

The `@for` track expression was improved to `row.type + '|' + row.name` — that's better and reduces same-type name collisions. However, the `data-testid` still concatenates the raw campaign name:

```html
[attr.data-testid]="'email-campaign-row-' + row.name"
```

Campaign names with spaces, slashes, or quotes produce selectors like `email-campaign-row-Q1 AWS / Cloud Drive` that are fragile in Playwright locators and CSS. A simple slugify would close this out:

```html
[attr.data-testid]="'email-campaign-row-' + (row.name | slugify)"
```

(or inline: `row.name.toLowerCase().replace(/[^a-z0-9]+/g, '-')` inside `initTopCampaigns`).

---

## ❌ Still open

### B2 — Duplicate `getMarketingAttribution` HTTP calls

`AttributionSectionComponent` is still mounted in **two** places:

1. `overview-tab.component.html` (last line) — always rendered when Overview tab is active
2. `marketing-impact.component.html` — rendered when Attribution tab is selected

Both instances independently call `getMarketingAttribution(slug)`. On initial page load (Overview tab selected), that's two concurrent fetches for the same endpoint with the same slug.

The fix is straightforward — remove `<lfx-attribution-section .../>` from the bottom of `overview-tab.component.html`. The attribution data belongs in its own dedicated tab, not embedded silently at the bottom of the Overview layout. If a preview/summary of attribution is intentionally wanted in the Overview, that should be a read-only summary widget that doesn't re-fetch independently.

### M5 — `selectedMonth` still not passed to API calls (`overview-tab.component.ts:27`)

`selectedMonth` flows into the summary title/subtitle labels but the three `forkJoin` service calls (`getRevenueImpact`, `getBrandReach`, `getEmailCtr`) still don't receive it. This means selecting April in the month picker shows April in the header but May data in the cards.

If upstream endpoints don't support per-month filtering yet, that's fine — but please add a short comment so the next reader doesn't chase a bug that isn't there:

```typescript
// TODO(LFXV2-XXXX): pass selectedMonth once upstream endpoints support per-month filtering
return forkJoin({ ... })
```

---

## 🔍 New observation — dead static property in `attribution-section.component.ts`

The refactor introduced a **duplicate** static map. Both are now in the class:

```typescript
// Line 24 — newly added, never referenced
private static readonly revenueKeyMap: Record<...> = { linear: 'linearRevenue', ... };

// Line 47 — the one getRevenueKey() actually uses
private static readonly revenueKeyByModel: Record<...> = { linear: 'linearRevenue', ... };
```

`revenueKeyMap` is dead code — `getRevenueKey()` references `revenueKeyByModel`. Remove `revenueKeyMap` to avoid confusion.

---

## Overall decision

**Needs one more pass** — two items remain blocking:

1. **B2**: Remove the extra `<lfx-attribution-section />` from the Overview tab template to eliminate duplicate HTTP calls.
2. **New**: Remove the dead `revenueKeyMap` static property from `AttributionSectionComponent`.

The `selectedMonth` TODO comment and `data-testid` slugify are minor but would be good to land in the same pass. Once those four are addressed, this is ready to approve.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,187 @@
...
+      const totalSends = data.monthlySends?.reduce((s, v) => s + v, 0) ?? 0;
+      const totalOpens = data.monthlyOpens?.reduce((s, v) => s + v, 0) ?? 0;
+      const openRate = totalSends > 0 ? (totalOpens / totalSends) * 100 : 0;
+      const changePct = data.changePercentage;
+
+      const sendsMom = this.computeMomPct(data.monthlySends);
+      const opensMom = this.computeMomPct(data.monthlyOpens);
+
+      const currentOpenRate = data.monthlySends?.at(-1) ? ((data.monthlyOpens?.at(-1) ?? 0) / data.monthlySends.at(-1)!) * 100 : 0;
+      const prevOpenRate = data.monthlySends?.at(-2) ? ((data.monthlyOpens?.at(-2) ?? 0) / data.monthlySends.at(-2)!) * 100 : 0;
```
MRashad26: **Problem:** `data.monthlySends?.at(-1)` is null-checked on the left of `?` (L74) but the same expression is then re-evaluated and dereferenced with a non-null assertion (`data.monthlySends.at(-1)!`) inside the divisor. The same pattern repeats on L75 for the `prev` calculation. Three reads of `monthlySends.at(-1)` per branch, plus a non-null assertion that duplicates the safety guarantee the `?` already established.

**Why this is a problem:**
1. Non-null assertions silence the compiler — if the guard expression on the left and the assertion on the right ever drift (someone tweaks the condition), the `!` will mask a real `undefined` and produce `NaN` in the open-rate KPI.
2. The `.at(-1)` call is invoked up to four times per render of the open-rate card. Cheap, but unnecessary.

**Suggested fix:** Cache the last/prev values once:
```ts
const lastSends = data.monthlySends?.at(-1);
const lastOpens = data.monthlyOpens?.at(-1) ?? 0;
const prevSends = data.monthlySends?.at(-2);
const prevOpens = data.monthlyOpens?.at(-2) ?? 0;

const currentOpenRate = lastSends ? (lastOpens / lastSends) * 100 : 0;
const prevOpenRate = prevSends ? (prevOpens / prevSends) * 100 : 0;
const openRateMom = prevOpenRate > 0 ? ((currentOpenRate - prevOpenRate) / prevOpenRate) * 100 : null;
```
No `!`, single read per index, identical behavior.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  // === Private Helpers ===
+  private computeMomPct(arr: number[] | undefined): number | null {
+    if (!arr || arr.length < 2) return null;
+    const current = arr.at(-1) ?? 0;
+    const previous = arr.at(-2) ?? 0;
+    if (previous === 0) return null;
+    return ((current - previous) / previous) * 100;
+  }
```
MRashad26, line 195: **Problem:** `computeMomPct` is a pure, stateless statistical helper — identical in shape to the `trendDirection` / `trendColorClass` / `formatChangePct` helpers that this PR already extracted into `@lfx-one/shared/utils` (used at L6).

**Why this is a problem:** Now that the *other* number-shaping helpers live in shared utils, leaving this one local makes the file inconsistent — readers have to know that some MoM-related helpers are shared and one is local. More importantly, the moment a second tab needs to compute MoM from a monthly series (very likely on the social/web-activity tabs, which also receive time-series arrays), this implementation will get copy-pasted, restarting the duplication this PR just resolved.

**Suggested fix:** Move it to the same shared utils module:
```ts
// packages/shared/src/utils/marketing-impact.utils.ts (or wherever the trend helpers live)
/** Returns MoM percent change from the last two values of a monthly series. */
export function computeMomPct(arr: number[] | undefined): number | null {
  if (!arr || arr.length < 2) return null;
  const current = arr.at(-1) ?? 0;
  const previous = arr.at(-2) ?? 0;
  if (previous === 0) return null;
  return ((current - previous) / previous) * 100;
}
```
Then import it on L6 alongside the others and delete the local copy.

MRashad26 (commented):
Follow-up on the prior /review. Comments #1 (trend helpers) and #4 (`import type`) are fully addressed; #2 (init pattern triple-set) had its bug-bearing footgun fixed via `finalize`; #3 (4-card skeleton template) is still open. Two minor follow-ups from the new code below.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
mrautela365: Fixed in commit d4415f37 — cached `.at(-1)` and `.at(-2)` values into local variables, eliminating all non-null assertions:

```ts
const lastSends = data.monthlySends?.at(-1);
const lastOpens = data.monthlyOpens?.at(-1) ?? 0;
const prevSends = data.monthlySends?.at(-2);
const prevOpens = data.monthlyOpens?.at(-2) ?? 0;

const currentOpenRate = lastSends ? (lastOpens / lastSends) * 100 : 0;
const prevOpenRate = prevSends ? (prevOpens / prevSends) * 100 : 0;
```

No `!`, single read per index.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
mrautela365: Fixed in commit d4415f37 — moved `computeMomPct` to `packages/shared/src/utils/marketing-impact.utils.ts` alongside the other trend helpers. Removed local copies from both `email-tab` and `performance-marketing-tab`. Both now import it from `@lfx-one/shared/utils`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/attribution-section/attribution-section.component.ts
```diff
@@ -21,6 +21,13 @@ import type { AttributionChannelRow, AttributionModel, AttributionModelOption, M
   styleUrl: './attribution-section.component.scss',
 })
 export class AttributionSectionComponent {
+  private static readonly revenueKeyMap: Record<AttributionModel, 'linearRevenue' | 'firstTouchRevenue' | 'lastTouchRevenue' | 'timeDecayRevenue'> = {
+    linear: 'linearRevenue',
+    firstTouch: 'firstTouchRevenue',
+    lastTouch: 'lastTouchRevenue',
+    timeDecay: 'timeDecayRevenue',
+  };
```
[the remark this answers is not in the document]
mrautela365: Done — removed the dead `revenueKeyMap` declaration. Moved `revenueKeyByModel` above services per component organization rules. See commit `05b71b61`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,187 @@
...
+          yoyTrendClass: 'text-gray-500',
+          comparisonLine: '',
+        },
+        {
+          id: 'ctr',
+          label: 'Click-Through Rate',
+          icon: 'fa-light fa-arrow-pointer',
+          iconClass: 'bg-violet-100 text-violet-600',
+          value: `${data.currentCtr.toFixed(2)}%`,
+          momChange: formatChangePct(changePct, 'MoM'),
```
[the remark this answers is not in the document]
mrautela365: Fixed — changed the CTR comparison label from 'MoM' to 'vs avg' since `changePercentage` is computed server-side against the 6-month rolling average, not a month-over-month delta. See commit c408153b.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,187 @@
...
+      const totalOpens = data.monthlyOpens?.reduce((s, v) => s + v, 0) ?? 0;
+      const openRate = totalSends > 0 ? (totalOpens / totalSends) * 100 : 0;
+      const changePct = data.changePercentage;
+
+      const sendsMom = this.computeMomPct(data.monthlySends);
+      const opensMom = this.computeMomPct(data.monthlyOpens);
+
+      const currentOpenRate = data.monthlySends?.at(-1) ? ((data.monthlyOpens?.at(-1) ?? 0) / data.monthlySends.at(-1)!) * 100 : 0;
+      const prevOpenRate = data.monthlySends?.at(-2) ? ((data.monthlyOpens?.at(-2) ?? 0) / data.monthlySends.at(-2)!) * 100 : 0;
+      const openRateMom = prevOpenRate > 0 ? ((currentOpenRate - prevOpenRate) / prevOpenRate) * 100 : null;
```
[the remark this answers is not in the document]
mrautela365: Fixed — the Open Rate headline now shows the current month's rate (computed from the last entry in `monthlySends` / `monthlyOpens`), which is consistent with the MoM trend that compares the last two months. Previously it showed a cumulative average which used a different base. See commit c408153b.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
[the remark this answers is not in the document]
mrautela365: Fixed — replaced the truthy check + non-null assertion pattern with explicit length checks and cached array values. Zero sends is now distinguished from missing data: `lastSends !== undefined && lastSends > 0` instead of `data.monthlySends?.at(-1)` which conflated the two. See commit c408153b.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">CAMPAIGNS</span>
+            <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400">SENDS</span>
+            <span class="w-24 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPENS</span>
+            <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400">OPEN RATE</span>
+            <span class="flex-1 text-right text-[11px] font-semibold tracking-wide text-gray-400">CTR</span>
+          </div>
+
+          <!-- Rows -->
+          @for (row of emailTypeRows(); track row.emailType) {
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-type-row-' + row.emailType">
```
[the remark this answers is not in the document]
mrautela365: Fixed — both email-type and campaign rows now use index-based `data-testid` (`email-type-row-0`, `email-campaign-row-0`, etc.) instead of interpolating raw names that may contain spaces or special characters. See commit c408153b.

MRashad26 (dismissed):

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      const sendsMom = this.computeMomPct(data.monthlySends);
+      const opensMom = this.computeMomPct(data.monthlyOpens);
+
+      const sends = data.monthlySends ?? [];
+      const opens = data.monthlyOpens ?? [];
+      const lastSends = sends.length > 0 ? sends[sends.length - 1] : undefined;
+      const prevSends = sends.length > 1 ? sends[sends.length - 2] : undefined;
+      const lastOpens = sends.length > 0 ? (opens[opens.length - 1] ?? 0) : 0;
+      const prevOpens = sends.length > 1 ? (opens[opens.length - 2] ?? 0) : 0;
```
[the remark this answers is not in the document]
mrautela365: Fixed — now using `Math.min(sends.length, opens.length)` to derive aligned indices, so both `lastOpens`/`prevOpens` index from the same bounded range as sends. See commit c71a24ad.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      const currentOpenRate = lastSends !== undefined && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
+      const prevOpenRate = prevSends !== undefined && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
+      const openRateMom = prevOpenRate > 0 ? ((currentOpenRate - prevOpenRate) / prevOpenRate) * 100 : null;
```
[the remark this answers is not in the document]
mrautela365: Fixed — extracted `computeMomPctFromValues(current, previous)` helper and refactored both `computeMomPct` and the open-rate MoM to use it. All three MoM calculations now follow the same pattern with consistent zero-handling. See commit c71a24ad.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+            </div>
+          }
+        </div>
+      } @else {
+        <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="email-type-empty">
+          <p class="text-sm text-gray-500">No email type data available.</p>
+        </div>
+      }
+    </div>
+
```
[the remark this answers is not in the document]
mrautela365: Fixed — both tables now render skeleton placeholders during loading (matching the KPI grid pattern) instead of disappearing. The `@if (!loading())` gate was replaced with a 3-way `@if (loading()) / @else if (hasData()) / @else` structure. See commit c71a24ad.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -0,0 +1,99 @@
...
+          @for (row of topCampaigns(); track row.type + '|' + row.name; let idx = $index) {
+            <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" [attr.data-testid]="'email-campaign-row-' + idx">
+              <span class="flex-1 truncate text-xs font-medium text-gray-700">{{ row.name }}</span>
+              <span class="w-24 truncate text-[11px] text-gray-500">{{ row.type }}</span>
+              <span class="w-20 text-right text-xs text-gray-900">{{ row.sends }}</span>
+              <span class="w-20 text-right text-xs text-gray-900">{{ row.opens }}</span>
+              <span class="w-20 text-right text-xs text-gray-500">{{ row.openRate }}</span>
+              <span class="w-16 text-right text-xs font-medium text-gray-900">{{ row.ctr }}</span>
+            </div>
+          }
```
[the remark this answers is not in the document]
mrautela365: Fixed — added ARIA roles (`role="table"`, `role="row"`, `role="columnheader"`, `role="cell"`) plus `aria-label` on both table containers so screen readers can navigate the tabular data. See commit c71a24ad.

MRashad26 (dismissed):

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.html
```diff
@@ -26,74 +26,98 @@ <h3 class="text-base font-semibold text-gray-900">Email marketing</h3>
...
+          <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">OPENS</span>
+          <span class="w-20 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">OPEN RATE</span>
+          <span class="w-16 text-right text-[11px] font-semibold tracking-wide text-gray-400" role="columnheader">CTR</span>
         </div>
-      }
-    </div>
-  }
+
+        @for (row of topCampaigns(); track row.type + '|' + row.name) {
+          <div class="flex items-center gap-3 border-b border-gray-100 py-2.5" role="row" [attr.data-testid]="'email-campaign-row-' + row.name">
```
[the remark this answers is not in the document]
mrautela365: Fixed — switched both tables to index-based test IDs (`email-type-row-0`, `email-campaign-row-0`) to avoid fragile selectors from raw names with special characters. See commit 1b3f00a0.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/email-tab/email-tab.component.ts
```diff
@@ -72,14 +72,15 @@ export class EmailTabComponent {
...
+      const lastSends = minLen > 0 ? sends[minLen - 1] : undefined;
+      const prevSends = minLen > 1 ? sends[minLen - 2] : undefined;
+      const lastOpens = minLen > 0 ? (opens[minLen - 1] ?? 0) : 0;
+      const prevOpens = minLen > 1 ? (opens[minLen - 2] ?? 0) : 0;
 
-      const currentOpenRate = lastSends !== undefined && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
-      const prevOpenRate = prevSends !== undefined && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
-      const openRateMom = prevOpenRate > 0 ? ((currentOpenRate - prevOpenRate) / prevOpenRate) * 100 : null;
+      const currentOpenRate = lastSends && lastSends > 0 ? (lastOpens / lastSends) * 100 : 0;
+      const prevOpenRate = prevSends && prevSends > 0 ? (prevOpens / prevSends) * 100 : 0;
```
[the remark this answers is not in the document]
mrautela365: Good catch — restored explicit `!== undefined` checks so zero-sends months are distinguished from missing data. See commit 1b3f00a0.

dealako:
Hey Misha — really nice turnaround on this. Four focused commits, each addressing a specific batch of feedback. Here's the follow-up pass:

---

## 👏 Nice work

- **Index-based `data-testid`** (`email-campaign-row-{idx}`, `email-type-row-{idx}`) is a better fix than the slugify I originally suggested — fragility-proof and stable regardless of campaign-name punctuation.
- **`computeMomPctFromValues` extraction** is a clean factoring — reused for both array-tail MoM and the open-rate-from-current-vs-prev calculation. No duplicated math, easy to test.
- **`min(sends.length, opens.length)` open-rate indexing** correctly handles the case where the two arrays come back at different lengths.
- **Explicit `lastSends !== undefined && lastSends > 0`** properly distinguishes "missing data" from "zero sends" — exactly what copilot was asking for.
- **ARIA table semantics** on both email-type and top-campaigns tables — `role="table"`, `role="row"`, `role="columnheader"`, `role="cell"` — noticeable a11y win.
- **Skeleton placeholders** on the table loading states match the KPI skeleton pattern and improve perceived performance.
- **CTR card label `'vs avg'`** is the accurate framing since `changePercentage` is the 6-month average comparison from the API, not month-over-month.

---

## ✅ Resolved

| # | Finding | How |
|---|---------|-----|
| **Dead `revenueKeyMap`** | Removed in `attribution-section.component.ts`. Only `revenueKeyByModel` remains, and `getRevenueKey()` uses it. |
| **M4 — `data-testid` slugify** | Switched to index-based testid (`'email-campaign-row-' + idx`) — different approach, better outcome. |

---

## ⚠️ Partially addressed

None this round.

---

## ❌ Still open

### M5 — `selectedMonth` still not flowing into API calls (`overview-tab.component.ts:43–58`)

`initOverviewKpiData()` still doesn't pass `selectedMonth` to `getRevenueImpact`, `getBrandReach`, or `getEmailCtr`. The month picker changes the header but the KPI values stay the same. Originally I asked for either threading it through OR a TODO comment for the next reader — neither landed.

This is the lowest-priority item and not gating merge, but a one-line TODO would close it cleanly:

```typescript
// TODO(LFXV2-XXXX): pass selectedMonth once upstream endpoints support per-month filtering
return forkJoin({ ... });
```

---

## 🤝 Retraction — B2 was wrong on my part

I owe you a correction here. In the previous review I claimed `<lfx-attribution-section />` was mounted twice and caused concurrent `getMarketingAttribution` fetches on initial page load. That was a misread of `marketing-impact.component.html`:

```html
@if (selectedTab() === 'overview') {
  <lfx-overview-tab ... />
} @else if (selectedTab() === 'attribution') {
  <lfx-attribution-section ... />
}
```

The two branches are mutually exclusive — only one attribution-section is ever in the DOM at a time, so there are no concurrent duplicate fetches. Apologies for the wild goose chase. Nothing to fix here.

(The lighter architectural question of whether attribution belongs at the bottom of Overview *and* as a standalone tab is still a fair design conversation, but it's not the bug I made it out to be.)

---

## 🔍 New observation

**NIT — dead `comparisonLine: ''` assignments** (`email-tab.component.ts:91, 106, 121, 136`)

All four KPI card literals still set `comparisonLine: ''`. PR #718 already made `comparisonLine` optional on `PerformanceSummaryKpi` and dropped the empty-string assignments from the overview-tab cards. Worth a quick sweep to match the pattern.

---

## Overall

**Approving** ✅ — the previously blocking items are resolved (or were my mistake), and M5 plus the `comparisonLine` cleanup are non-blocking. Happy to land this and follow up with the TODO comment in a small follow-on if that's easier.

Thanks again for the careful pass on the copilot feedback too — the explicit `!== undefined` guards in particular are exactly the right shape.

dealako (approved):
Tentatively approved. See my latest comment.

PR #727, opened 2026-05-18
feat(events): add travel fund deadline date support

dealako:
## Summary

Surface `travel_fund_end_date` from `ANALYTICS.PLATINUM_LFX_ONE.EVENT_REGISTRATIONS` in the Travel Funding Application experience:

- Display application deadline on each event card in the travel-fund picker; shows **Application Deadline not set** when the field is null
- Suppress events whose deadline has passed via a new `excludePastTravelFundDeadline` query option (`TRAVEL_FUND_END_DATE IS NULL OR TRAVEL_FUND_END_DATE >= CURRENT_DATE()`)
- Add **Deadline** column to the submitted Travel Fund Requests table for audit context

> ✅ [DONE!] **Dependency:** This feature requires `TRAVEL_FUND_END_DATE` to be available on `ANALYTICS.PLATINUM_LFX_ONE.EVENT_REGISTRATIONS`. 

## Changes

```
 event-request-list.component.html         | 16 ++++++++++++++++
 event-request-list.component.ts           |  3 ++-
 event-selection.component.html            | 11 +++++++++++
 event-selection.component.ts              |  4 +++-
 app/shared/services/events.service.ts     |  1 +
 server/controllers/events.controller.ts   |  2 ++
 server/services/events.service.ts         | 12 +++++++++++-
 packages/shared/interfaces/events.interface.ts | 12 ++++++++++++
 8 files changed, 58 insertions(+), 3 deletions(-)
```

Jira: [LFXV2-1856](https://linuxfoundation.atlassian.net/browse/LFXV2-1856)

[LFXV2-1856]: https://linuxfoundation.atlassian.net/browse/LFXV2-1856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/src/app/modules/events/my-events-dashboard/components/event-request-list/event-request-list.component.html
```diff
@@ -60,6 +60,11 @@
             Application Date <i [class]="sortIcons()['APPLICATION_DATE']"></i>
           </button>
         </th>
+        @if (requestType() === 'travel-fund') {
+          <th scope="col" [attr.data-testid]="config().testIdPrefix + '-sort-deadline'">
```
MRashad26: **Problem:** The new Deadline column's `data-testid` is `"<prefix>-sort-deadline"` (with `-sort-` in the name), implying it's sortable, but the cell is a plain `<span>` with no `<button>` and no click handler. The three sibling columns (Event Name L33-40, Location L44-51, Application Date L56-62) are all sortable: each is a `<button (click)="onHeaderClick(...)">` with the matching `aria-sort` attribute and sort-icon binding.

**Why this is a problem:**
1. **Inconsistent UX** — users will mouse over the header expecting a sort affordance (cursor change, hover state, click-to-sort) and get nothing. The four columns visually look the same but behave differently.
2. **Accessibility regression** — without `[attr.aria-sort]`, screen readers cannot announce that the column is unsortable. The other three columns set `aria-sort` correctly, so the omission here reads as an oversight rather than an intentional design choice.
3. **Misleading test-id** — `-sort-deadline` names a sorting affordance that does not exist. Future E2E tests using this selector to drive a sort will silently fail or click a no-op span.

**Suggested fix:** Either make the column sortable like its siblings, or rename the test-id to drop the `-sort-` prefix.

If sortable (preferred — matches the rest of the table):
```html
@if (requestType() === 'travel-fund') {
  <th
    scope="col"
    [attr.aria-sort]="sortField() === 'TRAVEL_FUND_END_DATE' ? (sortOrder() === 'ASC' ? 'ascending' : 'descending') : 'none'"
    [attr.data-testid]="config().testIdPrefix + '-sort-deadline'">
    <button
      type="button"
      class="flex cursor-pointer select-none items-center gap-1 text-xs text-gray-500 font-medium"
      (click)="onHeaderClick('TRAVEL_FUND_END_DATE')">
      Deadline <i [class]="sortIcons()['TRAVEL_FUND_END_DATE']"></i>
    </button>
  </th>
}
```
This requires adding `TRAVEL_FUND_END_DATE` to `VALID_EVENT_SORT_FIELDS` (server-side sort validation) and to the column's `sortIcons` derivation if it's not already inferred.

If intentionally unsortable, rename the test-id to `<prefix>-header-deadline` so it doesn't promise sorting.

packages/shared/src/interfaces/events.interface.ts
```diff
@@ -332,6 +340,8 @@ export interface VisaRequest {
   applicationDate: string;
   /** Visa letter request status (e.g. "Submitted", "Approved", "Denied", "Expired") */
   status: string;
+  /** ISO YYYY-MM-DD date string for the last day travel fund applications are accepted; null for visa requests or when not set */
+  travelFundEndDate?: string | null;
```
MRashad26, line 344: **Problem:** `VisaRequest` (a visa-letter request item) now has a `travelFundEndDate` field. The JSDoc on L343 explicitly documents the awkwardness: "*null for visa requests or when not set*". The same field also exists on `TravelFundRequest`-shaped consumers, where it actually belongs.

**Why this is a problem:**
1. **Domain bleed** — `VisaRequest` should model a visa request. A property whose documented value for every visa-request instance is "null" is not a property of the visa-request domain; it's a property of the travel-fund domain that's been hoisted to a shared parent for convenience.
2. **The doc string is a smell** — JSDoc that says "always null for this type" is the type system telling you the type is wrong. A reader of `VisaRequest.travelFundEndDate` has to learn the historical reason (the request-list component handles both shapes through a generic `request: VisaRequest | TravelFundRequest` slot) instead of trusting the type.
3. **Compounding** — every future visa-only field added to `VisaRequest`, and every travel-fund-only field added to a travel-fund interface, will face the same pressure to be "hoisted" to the other side. The interfaces will steadily diverge from the domain they describe.
4. **Server mapper has the same shape** — `mapRowToVisaRequest` (server-events.service.ts L1008) populates `travelFundEndDate` from `row.TRAVEL_FUND_END_DATE ?? null`, but `VisaRequestRow` (events.interface.ts L324) carries `TRAVEL_FUND_END_DATE: string | null` — the row type also gained a field that has no business being on a visa-request row.

**Suggested fix:** Two reasonable shapes:

*Option A — discriminated union* (preferred when callers can branch on a tag):
```ts
export interface VisaRequest {
  kind: 'visa';
  id: string; name: string; url: string; location: string;
  applicationDate: string; status: string;
}
export interface TravelFundRequest {
  kind: 'travel-fund';
  id: string; name: string; url: string; location: string;
  applicationDate: string; status: string;
  travelFundEndDate: string | null;
}
export type EventRequest = VisaRequest | TravelFundRequest;
```
Then `EventRequestListComponent` accepts `EventRequest` and the template can narrow on `request.kind`.

*Option B — common base + intersection*: extract the four shared fields to `EventRequestBase`, then `VisaRequest extends Base` (no extra fields) and `TravelFundRequest extends Base { travelFundEndDate: string | null }`. The component's `request` slot becomes `EventRequestBase & Partial<{ travelFundEndDate: ... }>` or — better — a generic over the concrete subtype.

Either way removes the "always null for this type" JSDoc and lets `requestType() === 'travel-fund'` be a narrowing check instead of a parallel signal.

apps/lfx-one/src/server/services/events.service.ts
```diff
@@ -997,6 +1005,7 @@ export class EventsService {
         ? new Date(row.APPLICATION_DATE).toLocaleDateString('en-US', { month: 'short', day: 'numeric', year: 'numeric' })
         : '—',
       status: row.REQUEST_STATUS,
+      travelFundEndDate: row.TRAVEL_FUND_END_DATE ?? null,
```
MRashad26: **Problem:** `applicationDate` is server-formatted to a localized string on L1005 (`new Date(row.APPLICATION_DATE).toLocaleDateString('en-US', { month: 'short', day: 'numeric', year: 'numeric' })` → e.g. `"Mar 5, 2026"`), but `travelFundEndDate` on L1008 is passed through as the raw ISO string from Snowflake. The client then runs the ISO through `DatePipe | date: 'MMM d, y'` in the template (event-request-list.component.html L96) to format it.

**Why this is a problem:**
1. **Two date columns, two formatting paths** — the same row renders one date that was formatted by Node's ICU in `en-US` locale on the server, and another formatted by Angular's `DatePipe` in the user's `LOCALE_ID` on the client. For a non-`en-US` LOCALE_ID, the two columns will format differently (different month abbreviations, different separators). Side-by-side in the same row, this looks like a bug.
2. **Server hard-codes `'en-US'`** — even if the app today only supports en-US, the choice is now wedged into the server in two places (L1005 for visa applicationDate, L1058 / L1075 / L1079-1080 for event date ranges) while the new feature adopts the better client-side approach. The codebase now has two opinions about where date formatting lives, and the wrong one is the established one.
3. **`new Date(YYYY-MM-DD)`** — when Snowflake returns a date-only string like `"2026-03-05"`, `new Date("2026-03-05")` parses it as UTC midnight, so a server in a negative-offset timezone (e.g. an `America/Los_Angeles` worker) renders it as "Mar 4, 2026" — off by one. The new ISO-passthrough avoids this entirely.

**Suggested fix:** Move `applicationDate` to the same ISO-passthrough model as `travelFundEndDate`, and format on the client with `DatePipe`:
```ts
// server: mapRowToVisaRequest / mapRowToTravelFundRequest
applicationDate: row.APPLICATION_DATE ?? null,  // ISO YYYY-MM-DD or null
travelFundEndDate: row.TRAVEL_FUND_END_DATE ?? null,
```
```ts
// shared: VisaRequest
applicationDate: string | null;  // was: string
```
```html
<!-- event-request-list.component.html: matching cell -->
<span class="text-sm text-gray-700">
  @if (request.applicationDate) {
    {{ request.applicationDate | date: 'MMM d, y' }}
  } @else { — }
</span>
```
Benefits: one formatting path, locale-correct for both columns, no `new Date(YYYY-MM-DD)` UTC drift, and consistency with the deadline column added in this PR. The event-date-range formatters (`formatDateRange`, L1057-1081) can be migrated separately.

apps/lfx-one/src/server/services/events.service.ts
```diff
@@ -107,6 +108,9 @@ export class EventsService {
       const registeredOnlyFilter = registeredOnly ? "AND r.EVENT_ID IS NOT NULL AND r.REGISTRATION_STATUS = 'Accepted'" : '';
       const visaRequestAcceptedFilter = isVisaRequestAccepted ? 'AND r.IS_VISA_REQUEST_ACCEPTED = TRUE' : '';
       const travelFundRequestAcceptedFilter = isTravelFundRequestAccepted ? 'AND r.IS_TRAVEL_FUND_ACCEPTED = TRUE' : '';
+      const excludePastTravelFundDeadlineFilter = excludePastTravelFundDeadline
+        ? 'AND (r.TRAVEL_FUND_END_DATE IS NULL OR r.TRAVEL_FUND_END_DATE >= CURRENT_DATE())'
```
MRashad26, line 112: **Problem:** The new SQL filter uses `CURRENT_DATE()` to exclude past travel-fund deadlines:
```sql
AND (r.TRAVEL_FUND_END_DATE IS NULL OR r.TRAVEL_FUND_END_DATE >= CURRENT_DATE())
```
Snowflake's `CURRENT_DATE()` evaluates against the **session timezone**, which for a server-side process typically defaults to UTC (or whatever `TIMEZONE` parameter the warehouse session inherits — `SHOW PARAMETERS LIKE 'TIMEZONE' IN SESSION` to verify).

**Why this is a problem:**
1. **Boundary asymmetry by timezone** — a user in `America/Los_Angeles` browsing at 11pm on May 17 sees "the deadline is today, May 18" but Snowflake (UTC) already thinks it's May 18 (`CURRENT_DATE() = '2026-05-18'`), so an event with `TRAVEL_FUND_END_DATE = '2026-05-18'` is *still included* (good) — but an event with `END_DATE = '2026-05-17'` is excluded even though it's still May 17 in the user's wall-clock time. The user sees an event they applied to vanish from the list before midnight local time.
2. **East of UTC, the opposite race** — a user in `Asia/Tokyo` at 9am local on May 18 sees Snowflake (UTC) still thinks it's May 17 (`CURRENT_DATE() = '2026-05-17'`), so an event with `END_DATE = '2026-05-17'` is included (the row passes `'2026-05-17' >= '2026-05-17'`), even though it's already May 18 for the user. They can click into an event whose deadline already passed in their reality.
3. **Inclusive-vs-exclusive ambiguity** — the comparison `>= CURRENT_DATE()` makes the deadline an *inclusive* last day, which the column docstring (`events.interface.ts` L324, L343) agrees with ("last day travel fund applications are accepted"). Worth confirming this matches the upstream events service semantics — if upstream treats `TRAVEL_FUND_END_DATE` as exclusive (deadline = first day no longer accepted), the comparison should be `> CURRENT_DATE()`.

**Suggested fix:** Two parts, depending on how strict the product wants to be.

*Minor (defensible default)* — document that the cutoff is UTC-day-based, and accept the ±1-day fuzz at timezone boundaries. The current filter is fine for that semantic.

*Strict (timezone-aware)* — bind the user's timezone (or the event's local TZ if Snowflake has it) and compare in that zone:
```ts
const userTz = req.user?.timezone ?? 'UTC';   // however the app surfaces it
// in options:
userTimezone: userTz,
// SQL:
const excludePastTravelFundDeadlineFilter = excludePastTravelFundDeadline
  ? `AND (r.TRAVEL_FUND_END_DATE IS NULL
       OR r.TRAVEL_FUND_END_DATE >= TO_DATE(CONVERT_TIMEZONE('UTC', ?, CURRENT_TIMESTAMP())))`
  : '';
// binds: push userTz at the matching position
```
or cheaper: precompute the cutoff date in Node using the user's tz and bind it as a parameter — keeps the SQL simple.

At minimum, please add a one-line comment on L111 noting that the cutoff is UTC and document the semantics in the interface JSDoc, so the next maintainer doesn't have to rediscover it via a bug report.

MRashad26 (commented):
Strict pass against `~/LFX/code-enforcer-agent.md`. Feature works end-to-end and the column-only-when-travel-fund pattern is clean. Four findings worth a look — one UX inconsistency in the new column, one type-shape concern, one rendering-strategy inconsistency, and one timezone edge case in the SQL filter.

apps/lfx-one/src/server/services/events.service.ts
dealako: **CRITICAL — raw date passthrough (YYYY-MM-DD contract violation)**

Both `mapRowToVisaRequest` (here, line 1008) and `mapRowToEvent` (line 1053) pass `row.TRAVEL_FUND_END_DATE ?? null` without normalization. Snowflake's Node.js driver returns `DATE` columns as JavaScript `Date` objects, which Express / `JSON.stringify` serializes as a UTC timestamp (e.g. `"2026-07-05T00:00:00.000Z"`) — not the documented `YYYY-MM-DD` format. Angular's `DatePipe` interprets the `T00:00:00.000Z` offset in the user's local timezone, so users west of UTC can see `Jul 4` for a `Jul 5` deadline.

**Fix** — `formatDateToISOString` is already exported from `@lfx-one/shared/utils` (alongside the existing `formatDateToUTC` import on line 38). Add it to the import and apply it in both mappers:

```typescript
// line 38 — extend existing import
import { formatDateToISOString, formatDateToUTC } from '@lfx-one/shared/utils';

// line 1008 (mapRowToVisaRequest)
travelFundEndDate: formatDateToISOString(row.TRAVEL_FUND_END_DATE) ?? null,

// line 1053 (mapRowToEvent)
travelFundEndDate: formatDateToISOString(row.TRAVEL_FUND_END_DATE) ?? null,
```

`formatDateToISOString` returns `date.toISOString().split('T')[0]`, which always yields `YYYY-MM-DD` regardless of the caller's timezone.

apps/lfx-one/src/app/modules/events/my-events-dashboard/components/event-request-list/event-request-list.component.html
```diff
@@ -60,6 +60,11 @@
...
+            <span class="text-xs text-gray-500 font-medium">Deadline</span>
```
dealako, line 65: **NIT — misleading `sort` in test ID**

`'-sort-deadline'` follows the naming pattern of the **sortable** columns above (which all render `<button (click)="onHeaderClick(...)">`), but this header has no sort button. Tests or future readers may assume the column is interactive.

Suggestion: rename to `config().testIdPrefix + '-deadline-header'`.

dealako (commented):
This PR cleanly surfaces `travel_fund_end_date` from the PLATINUM view into both the travel-fund picker and the submitted-requests table. The structure is solid — conditional rendering is properly gated on `requestType === 'travel-fund'`, the SQL filter uses `CURRENT_DATE()` correctly, and the shared type additions are appropriately scoped.

**One required fix before merge:** Both row mappers (`mapRowToVisaRequest` line 1008, `mapRowToEvent` line 1053) pass `row.TRAVEL_FUND_END_DATE ?? null` without normalizing to `YYYY-MM-DD`. Snowflake's Node.js driver returns `DATE` columns as JavaScript `Date` objects, which Express serializes as UTC timestamps (`"2026-07-05T00:00:00.000Z"`). Angular's `DatePipe` then interprets this in the user's local timezone — users west of UTC can see `Jul 4` for a `Jul 5` deadline. Fix: use `formatDateToISOString(row.TRAVEL_FUND_END_DATE) ?? null` in both mappers (add `formatDateToISOString` to the existing `@lfx-one/shared/utils` import on line 38 — it's already exported).

Two nits noted inline (test ID naming, interface comment accuracy) — neither is a blocker.

dealako:
Addressed all actionable AI review feedback in 72a572f:

**Fixed (3 items):**
- **Date normalization** — Both `mapRowToVisaRequest` and `mapRowToEvent` now normalize `TRAVEL_FUND_END_DATE` via `new Date(row.TRAVEL_FUND_END_DATE).toISOString().split('T')[0]`, ensuring the value is always serialized as `YYYY-MM-DD` regardless of how the Snowflake driver returns it at runtime. This prevents the off-by-one timezone rendering for users west of UTC.
- **Test ID naming** — Renamed `'-sort-deadline'` → `'-deadline-header'` to avoid implying the Deadline column is sortable.
- **Interface comment accuracy** — Updated JSDoc on `VisaRequestRow.TRAVEL_FUND_END_DATE` and `VisaRequest.travelFundEndDate` to accurately reflect that the field is present on visa rows but suppressed in rendering (travel-fund only).

**Out of scope (noted for follow-up):**
- Deadline enforcement at submission time — by design the check is picker-level; backend enforcement would require a separate task.
- E2E test coverage for deadline behavior — tracked separately.
- Filter null-join concern — not applicable since the picker always calls with `registeredOnly: true`, so the `r` alias is never null.

apps/lfx-one/src/server/services/events.service.ts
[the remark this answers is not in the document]
dealako: Not applicable here: the picker always calls `getMyEvents` with `registeredOnly: true`, which adds `AND r.IS_REGISTERED = TRUE` to the WHERE clause — so every row that reaches the deadline filter has a valid user registration join. The `r` alias is never NULL in this code path.

apps/lfx-one/src/app/modules/events/my-events-dashboard/components/event-selection/event-selection.component.ts
```diff
@@ -69,6 +70,7 @@ export class EventSelectionComponent {
     country: this.filtersValue().locationFilter !== 'any' ? (this.filtersValue().locationFilter ?? undefined) : undefined,
     isVisaRequestAccepted: this.requestType() === 'visa' ? true : undefined,
     isTravelFundRequestAccepted: this.requestType() === 'travel-fund' ? true : undefined,
+    excludePastTravelFundDeadline: this.requestType() === 'travel-fund' ? true : undefined,
```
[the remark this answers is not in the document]
dealako: Acknowledged — out of scope for this PR. Server-side enforcement at submission time would require the upstream service to validate the deadline on the POST. Tracked as a follow-up.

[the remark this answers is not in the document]
dealako: Acknowledged — E2E coverage for the deadline behavior is tracked as a follow-up task.

apps/lfx-one/src/server/services/events.service.ts
dealako: Intentional — `CURRENT_DATE()` uses the Snowflake session timezone (UTC by default in the LFX Snowflake account). Application deadlines are business-day granular, so UTC midnight is an acceptable boundary. This matches the approach used elsewhere in the codebase for date-only comparisons.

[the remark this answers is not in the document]
dealako: `excludePastTravelFundDeadline` is only ever activated alongside `registeredOnly: true` (see event-selection.component.ts). The `registeredOnly` flag appends `AND r.EVENT_ID IS NOT NULL AND r.REGISTRATION_STATUS = 'Accepted'` to the WHERE clause, which forces a registration match and guarantees `r` is never null when the deadline filter fires. No unregistered rows can reach the deadline predicate in practice.

packages/shared/src/interfaces/events.interface.ts
```diff
@@ -75,6 +77,8 @@ export interface MyEventRow {
   /** True when the user has registered for this event */
   IS_REGISTERED: boolean;
   TOTAL_RECORDS: number;
+  /** Last date travel fund applications are accepted for this event; null when not set */
+  TRAVEL_FUND_END_DATE: string | null;
```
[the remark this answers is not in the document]
dealako: Good catch. The past/registered branch was missing `TRAVEL_FUND_END_DATE` from its SELECT, making the field `undefined` at runtime for past-event rows despite `MyEventRow` typing it as `string | null`. Fixed in 49b934fc — `TRAVEL_FUND_END_DATE` is now selected in both query branches, keeping the interface accurate.

apps/lfx-one/src/app/modules/events/my-events-dashboard/components/event-selection/event-selection.component.ts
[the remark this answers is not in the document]
dealako: Acknowledged — already documented as a follow-up item in the PR summary comment. E2E coverage for deadline behavior will be added in a subsequent task.

dealako:

Commit: 49b934fc

### Change Made
- **`apps/lfx-one/src/server/services/events.service.ts`**: Added `TRAVEL_FUND_END_DATE` to the past/registered-events query branch SELECT. The column was already selected in the upcoming-events branch but missing from the past-events branch, making `MyEventRow.TRAVEL_FUND_END_DATE` undefined at runtime for past rows despite the interface typing it as `string | null` (per copilot[bot]).

### No Change Needed
- **`apps/lfx-one/src/server/services/events.service.ts:112`** (deadline filter on left-joined row): `excludePastTravelFundDeadline` is only activated alongside `registeredOnly: true`, which enforces `r.EVENT_ID IS NOT NULL` — so `r` is never null when the deadline filter fires (per copilot[bot]).
- **`apps/lfx-one/src/app/modules/events/my-events-dashboard/components/event-selection/event-selection.component.ts:73`** (E2E coverage): Tracked as a follow-up task, per the summary comment above (per copilot[bot]).

### Threads Resolved
3 of 3 unresolved threads addressed in this iteration. All 17 review threads on this PR are now resolved.

MRashad26 (approved):

PR #730, opened 2026-05-19
fix(settings): update "on this page" highlight as user scrolls

fayazg:
Fixes LFXV2-1877

## Summary

- Track the active "On this page" TOC item using an `IntersectionObserver` on the three section anchors (`email-settings`, `password`, `developer-settings`) instead of relying on click-only state.
- Observer is set up via `afterNextRender` and disconnected through `destroyRef.onDestroy`. Guarded with `isPlatformBrowser` per `.claude/rules/ssr-safety.md`, so SSR is unaffected.
- Activation band `rootMargin: '-80px 0px -70% 0px'` puts the highlight on the section whose top crosses roughly the upper third of the viewport, below the sticky page header.
- Click-to-scroll behavior in `scrollToSection()` is unchanged — the click still sets `activeSection` immediately for instant feedback, and the observer takes over once the smooth scroll settles.

## Test plan

- [x] Navigate to `/settings/account`
- [x] Scroll slowly down: TOC highlight advances Email Settings → Password → Developer Settings
- [x] Scroll back up: highlight reverses correctly
- [x] Click each TOC item: page smooth-scrolls to that section and the highlight lands on it
- [x] `yarn build` passes (browser + SSR bundles)
- [x] `yarn lint:check` passes


fayazg:
**Commit `8005bbe9` — refactor(settings): use local observer ref in scroll-spy setup**

Small code-quality polish addressing the self-review NIT from the PR review:

- `setupScrollSpy()` now assigns the `IntersectionObserver` to a local `const observer` first, then stores it on `this.scrollSpyObserver`
- The `forEach` loop uses `observer.observe(heading)` directly instead of `this.scrollSpyObserver!.observe(heading)` with a non-null assertion
- No behavior change — cleanup via `destroyRef.onDestroy` is unchanged

MRashad26 (commented):
Strict pass against `~/LFX/code-enforcer-agent.md`. The IntersectionObserver approach is the right primitive, SSR is correctly guarded with `isPlatformBrowser`, and `destroyRef.onDestroy` cleanup is in place. Two findings worth a look — one about coupling production behavior to test selectors, one about the end-of-scroll override.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -468,4 +473,56 @@ export class AccountSettingsComponent {
...
+  private setupScrollSpy(): void {
+    if (!isPlatformBrowser(this.platformId)) return;
+
+    // Observe the section heading rows (sentinels) rather than the full section divs.
+    // A heading is short enough that at most one fits in the activation band, which
+    // avoids two sections being considered active during the transition.
+    const sectionIds = ['email-settings', 'password', 'developer-settings'];
+    const headingByElement = new Map<Element, string>();
+    for (const id of sectionIds) {
+      const heading = document.querySelector(`[data-testid="${id}-heading"]`);
```
MRashad26: **Problem:** This scroll-spy queries the DOM by `data-testid`:
```ts
const heading = document.querySelector(`[data-testid="${id}-heading"]`);
```
The `data-testid` attribute (per project convention and the wider Angular/Testing-Library guidance) is an attribute reserved for test selectors. This commit makes a production runtime behavior — which TOC item gets highlighted — depend on it.

**Why this is a problem:**
1. **Test/production coupling** — if a future PR renames `email-settings-heading` → `email-settings-h2` for a test-readability reason, the scroll-spy silently breaks. The lint/typecheck does not catch this because both ends are strings. The bug surfaces only when someone notices the TOC highlight is stuck on "Email Settings".
2. **It violates the intent of `data-testid`** — the testid is supposed to be a test affordance the production code is unaware of. Reading it from production reverses that direction. The same component already uses semantic `id` attributes for section divs (`id="email-settings"` etc. — used by `scrollToSection()` via `getElementById`), so the codebase already has the right pattern; the heading just doesn't have one.
3. **Selector cost** — `querySelector` with an attribute-equality selector is slower than `getElementById` (hash lookup), and is run three times during render. Tiny, but free to avoid.

**Suggested fix:** Add a stable `id` to each heading and query by `getElementById` (matches what `scrollToSection` already does), or use `viewChildren` template refs.

*Option A — `id` attribute on the heading:*
```html
<!-- account-settings.component.html L54 -->
<h2 id="email-settings-heading" class="..." data-testid="email-settings-heading">Email Settings</h2>
<!-- same for password-heading (L225), developer-settings-heading (L462) -->
```
```ts
// account-settings.component.ts L486
const heading = document.getElementById(`${id}-heading`);
```

*Option B — `viewChildren` (no DOM access at all):*
```ts
import { viewChildren, ElementRef } from '@angular/core';

readonly sectionHeadings = viewChildren<ElementRef<HTMLElement>>('sectionHeading');
```
```html
<h2 #sectionHeading [attr.data-section-id]="'email-settings'" ...>Email Settings</h2>
```
Then iterate `this.sectionHeadings()` instead of `document.querySelector`. This is the more Angular-idiomatic option and avoids global DOM queries entirely — but Option A is a one-line fix if you want to keep the change small.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -468,4 +473,56 @@ export class AccountSettingsComponent {
...
+    );
+    this.scrollSpyObserver = observer;
+
+    headingByElement.forEach((_, heading) => observer.observe(heading));
+
+    // End-of-scroll override: the last section is short enough that its heading
+    // never enters the activation band, so the observer alone leaves the highlight
+    // on the previous section. Snap to the last section when reaching the bottom.
+    const isAtBottom = () =>
+      document.documentElement.scrollHeight > window.innerHeight && window.innerHeight + window.scrollY >= document.documentElement.scrollHeight - 4;
```
MRashad26: **Problem:** The end-of-scroll override re-implements bottom-detection with a magic tolerance and a `scroll` event listener, when the same intent ("highlight the last section once the bottom is reached") can be done with the IntersectionObserver instance you already have.
```ts
const isAtBottom = () =>
  document.documentElement.scrollHeight > window.innerHeight
  && window.innerHeight + window.scrollY >= document.documentElement.scrollHeight - 4;
const onScroll = () => { if (isAtBottom()) this.activeSection.set(lastSectionId); };
window.addEventListener('scroll', onScroll, { passive: true });
```

**Why this is a problem:**
1. **Per-event layout reads** — every scroll event reads `scrollHeight`, `innerHeight`, and `scrollY`. `scrollHeight` is a layout read; if any DOM mutation happened since the last frame (a toast appearing, a button toggling, an `@if` flipping), this forces a synchronous layout. Cheap individually, but it happens on every scroll tick across the entire page.
2. **Magic 4px tolerance** — the `- 4` is a guess at sub-pixel rounding / Hi-DPI / scrollbar behavior. It is also wrong on browsers that allow overscroll (where `scrollY + innerHeight` can briefly exceed `scrollHeight`) and on browsers where the document scroller is `<body>` rather than `<html>` (varies by Quirks Mode and historical iOS Safari behavior). The 4px works today; whether it works after the next browser update is unproven.
3. **Duplicate machinery for a single signal** — you have an `IntersectionObserver` already wired up with proper teardown via `destroyRef.onDestroy`. Adding a scroll listener doubles the things that can leak (the diff does clean both up correctly, but doubling the cleanup surface is friction for future readers).
4. **The root cause is hidden** — the comment explains *why* the override exists ("the last section is short enough that its heading never enters the activation band"). That's a symptom of the activation band being defined by `rootMargin: '-80px 0px -70% 0px'` (a 70% bottom inset means the band sits in roughly the top 30% of the viewport). For the last section, that band is never reached, so the heuristic is patching the observer config rather than fixing it.

**Suggested fix:** Add an invisible sentinel element at the *end* of the scrollable region and observe it with the same `IntersectionObserver`. When the sentinel becomes visible, snap to `lastSectionId`. One observer, no scroll listener, no magic number.

```html
<!-- account-settings.component.html: after the last section div, inside the same scroll container -->
<div #scrollEndSentinel data-testid="scroll-end-sentinel" class="h-px" aria-hidden="true"></div>
```
```ts
// account-settings.component.ts: inside setupScrollSpy(), after observing headings
const sentinel = document.querySelector('[data-testid="scroll-end-sentinel"]');
if (sentinel) {
  // Use a separate, plain (no rootMargin) observer for the bottom snap, OR
  // a different rootMargin variant. Cleanest: a second observer.
  const endObserver = new IntersectionObserver(
    ([entry]) => { if (entry.isIntersecting) this.activeSection.set(lastSectionId); },
    { threshold: 0 }
  );
  endObserver.observe(sentinel);
  this.destroyRef.onDestroy(() => endObserver.disconnect());
}
```

Alternatively — and lighter — change the heading sentinel for the *last* section to be the section's bottom rather than its top: observe the section div itself with `rootMargin: '0px 0px -1px 0px'` so it only counts as intersecting when its bottom is on screen. Either eliminates the scroll listener and the 4px tolerance.

Not a blocker — the current code works in the happy path you tested. But IntersectionObserver was designed exactly to replace scroll listeners for this kind of "am I at the bottom" check, so it's worth using it consistently.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
fayazg: Done — added `id="email-settings-heading"`, `id="password-heading"`, and `id="developer-settings-heading"` to the three `<h2>` elements, then switched the lookup from `document.querySelector('[data-testid="..."]')` to `document.getElementById(`${id}-heading`)`, matching the existing pattern in `scrollToSection()`.

See commit aba0d80e: fix(settings): address PR #730 review feedback.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
fayazg: Done — removed the `isAtBottom()` helper, the `onScroll` function, and the `window.addEventListener('scroll', ...)` call entirely. In their place, a 1px `<div data-testid="scroll-end-sentinel">` is appended after the last section inside the content column, and a second `IntersectionObserver` (threshold 0, no rootMargin) snaps `activeSection` to the last section when the sentinel enters the viewport. Both observers are disconnected in the same `destroyRef.onDestroy` callback — no scroll listener, no magic pixel values.

See commit aba0d80e: fix(settings): address PR #730 review feedback.

fayazg:
## Review Feedback Addressed

Commit: aba0d80e

### Changes Made

- **account-settings.component.html**: added `id="email-settings-heading"`, `id="password-heading"`, and `id="developer-settings-heading"` to the three `<h2>` elements (per @MRashad26)
- **account-settings.component.ts**: replaced `document.querySelector('[data-testid="..."]')` with `document.getElementById(...)` — consistent with the existing `scrollToSection()` pattern (per @MRashad26)
- **account-settings.component.html**: appended an invisible 1px sentinel `<div>` at the end of the content column (per @MRashad26)
- **account-settings.component.ts**: removed the `isAtBottom()` helper, `onScroll` callback, and `window.addEventListener('scroll', ...)` — replaced with a second `IntersectionObserver` over the sentinel that snaps to the last section when the sentinel enters the viewport. Both observers disconnect in `destroyRef.onDestroy` (per @MRashad26)

### No Change Needed

coderabbitai's top-level nitpick about documenting the magic numbers (`-80px 0px -70% 0px` and `4px`) is now largely moot: the `4` is gone after the sentinel refactor, and the `-70%` rootMargin is the only remaining value; its meaning is conveyed by the surrounding comment.

### Threads Resolved

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -468,4 +473,57 @@ export class AccountSettingsComponent {
...
+      { rootMargin: '-80px 0px -70% 0px', threshold: 0 }
+    );
+    this.scrollSpyObserver = observer;
+
+    headingByElement.forEach((_, heading) => observer.observe(heading));
+
+    // The last section is short enough that its heading never enters the activation
+    // band. Observe an invisible sentinel at the bottom of the content column so we
+    // can snap to the last section without a scroll listener or magic pixel values.
+    const sentinel = document.querySelector('[data-testid="scroll-end-sentinel"]');
```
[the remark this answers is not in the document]
fayazg: Done — added `id="scroll-end-sentinel"` to the sentinel `<div>` (kept `data-testid` for test selectors) and switched the lookup from `document.querySelector('[data-testid="scroll-end-sentinel"]')` to `document.getElementById('scroll-end-sentinel')`, consistent with the same fix applied to the heading lookups in the previous commit.

See commit 990cbab1.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -468,4 +473,57 @@ export class AccountSettingsComponent {
...
+    const endObserver = new IntersectionObserver(
+      ([entry]) => {
+        if (entry.isIntersecting) this.activeSection.set(lastSectionId);
+      },
+      { threshold: 0 }
+    );
+    if (sentinel) endObserver.observe(sentinel);
```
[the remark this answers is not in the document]
fayazg: Done — added a guard in the end-observer callback: it only calls `activeSection.set(lastSectionId)` when `entry.isIntersecting && lastHeading.getBoundingClientRect().bottom <= 80`. The 80px matches the header offset already used in the main observer's `rootMargin`, so "past the last heading" means the same thing in both observers. On viewports tall enough that the sentinel is visible from initial paint (no scrolling required), `lastHeading.getBoundingClientRect().bottom` will be well above 80px, so the guard prevents the TOC from jumping to "Developer Settings" before the user has scrolled there.

See commit 990cbab1.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -468,4 +473,57 @@ export class AccountSettingsComponent {
...
+      (entries) => {
+        for (const entry of entries) {
+          const id = headingByElement.get(entry.target);
+          if (!id) continue;
+          if (entry.isIntersecting) intersecting.add(id);
+          else intersecting.delete(id);
+        }
+        const activeId = sectionIds.find((id) => intersecting.has(id));
+        if (activeId) this.activeSection.set(activeId);
+      },
```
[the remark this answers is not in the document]
fayazg: The concern is valid in theory, but the current `rootMargin: '-80px 0px -70% 0px'` already prevents it in practice. That margin creates an activation band that spans roughly the top 30% of the viewport below the header — short enough that a single `<h2>` row is the only thing that fits. The code comment at lines 480–482 explains the intent: the headings are observed precisely *because* they are small enough to prevent two from intersecting simultaneously (observing full section `<div>`s would have caused the tie-break problem you describe).

Picking the earliest section in document order (`sectionIds.find(...)`) also matches reading intent: if two headings were ever transitioning through the band, the user is finishing the earlier section, not starting the later one, so highlighting earlier is correct.

No code change for now — the assumption holds for this layout. If the page is later restructured (e.g., fixed sidebar narrows the content, inflating heading height), it would be worth revisiting with the `boundingClientRect.top` tie-break you suggested. Happy to discuss further.

fayazg:

Commit: 990cbab1

### Changes Made

- **account-settings.component.html**: added `id="scroll-end-sentinel"` to the sentinel `<div>` (kept `data-testid` for test selectors) (per copilot-pull-request-reviewer)
- **account-settings.component.ts**: replaced `document.querySelector('[data-testid="..."]')` with `document.getElementById('scroll-end-sentinel')` — same fix applied in round 1 to the heading lookups (per copilot-pull-request-reviewer)
- **account-settings.component.ts**: added a guard in the end-observer callback; the sentinel only snaps to "Developer Settings" when `lastHeading.getBoundingClientRect().bottom <= 80`, preventing the TOC from pinning to the last section on tall viewports where the sentinel is visible from initial paint (per copilot-pull-request-reviewer)

### No Change Needed

- **line 505 — tie-breaking**: the `rootMargin: '-80px 0px -70% 0px'` already sizes the activation band to fit at most one heading at a time, and `sectionIds.find()` (earliest-in-document) matches reading intent. Responded on the thread with rationale. Left open for reviewer to confirm.

### Threads Resolved

2 of 3 new threads resolved. Thread 1 (tie-breaking concern) left open pending reviewer acknowledgment.

MRashad26 (approved):

PR #735, opened 2026-05-19
fix(server): resolve stale chunk import failure after deploy

MRashad26:
## Summary

- Serve `index.html` with `no-cache, must-revalidate` so returning users always revalidate after a deploy — fixes the root cause of stale chunk hashes being requested
- Keep `public, max-age=31536000, immutable` for content-hashed JS/CSS/font/image assets (no regression to asset caching)
- Add `ChunkLoadErrorHandler` global Angular `ErrorHandler` that detects `TypeError: Importing a module script failed` (Safari), `Failed to fetch dynamically imported module` (Chrome/Edge), and Firefox equivalents — triggers a single `window.location.reload()` per session to recover users whose browsers already have the stale `index.html` cached
- Opt `/mailing-lists/create` into the custom preloader (`data: { preload: true, preloadDelay: 2000 }`) to reduce click-time exposure

## Root cause

`apps/lfx-one/src/server/server.ts` was serving `index.html` with `Cache-Control: public, max-age=31536000`. After a deploy, Angular's `outputHashing: "all"` regenerates all chunk filenames. Returning users' browsers served the year-old cached `index.html` referencing old chunk hashes → the dynamic `import()` on clicking **Create Mailing List** requested a file that no longer existed → the SSR fallback returned HTML instead of JS → `TypeError: Importing a module script failed.`

Safari fails reliably (no speculation-rules prefetcher); Chrome/Arc fail intermittently (Chromium's `speculationrules` in `index.html` sometimes refreshes chunks before the click).

## Test plan

- [ ] `curl -I https://<preview-url>/` → `Cache-Control: no-cache, must-revalidate`
- [ ] `curl -I https://<preview-url>/main-<hash>.js` → `Cache-Control: public, max-age=31536000, immutable`
- [ ] Click **Create Mailing List** on Project Dashboard in Safari — no `TypeError: Importing a module script failed`
- [ ] Repeat in Chrome and Arc
- [ ] Confirm repeat visits to already-cached pages still load hashed assets from disk cache (no regression)

## JIRA

[LFXV2-1920](https://linuxfoundation.atlassian.net/browse/LFXV2-1920)

[LFXV2-1920]: https://linuxfoundation.atlassian.net/browse/LFXV2-1920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
+++ b/apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
@@ -1,0 +1,41 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { isPlatformBrowser } from '@angular/common';
+import { ErrorHandler, inject, Injectable, PLATFORM_ID } from '@angular/core';
+
+const RELOAD_FLAG = 'lfx-chunk-reload-attempted';
+
+const isChunkLoadError = (error: unknown): boolean => {
+  const message = error instanceof Error ? error.message : String(error);
+  return (
+    message.includes('Importing a module script failed') || // Safari
+    message.includes('Failed to fetch dynamically imported module') || // Chrome / Edge
+    message.includes('error loading dynamically imported module') // Firefox
+  );
+};
+
+@Injectable()
+export class ChunkLoadErrorHandler implements ErrorHandler {
+  private readonly platformId = inject(PLATFORM_ID);
+
+  handleError(error: unknown): void {
+    if (!isPlatformBrowser(this.platformId)) {
+      return;
+    }
+
+    if (!isChunkLoadError(error)) {
+      console.error(error);
+      return;
+    }
+
+    // Guard against reload loops — only attempt once per session.
+    if (sessionStorage.getItem(RELOAD_FLAG)) {
+      console.error('Chunk load error persists after reload — giving up.', error);
+      return;
+    }
+
+    sessionStorage.setItem(RELOAD_FLAG, '1');
+    window.location.reload();
+  }
+}
--- a/apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
+++ b/apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
@@ -4,7 +4,8 @@
 import { isPlatformBrowser } from '@angular/common';
 import { ErrorHandler, inject, Injectable, PLATFORM_ID } from '@angular/core';
 
-const RELOAD_FLAG = 'lfx-chunk-reload-attempted';
+const RELOAD_FLAG = 'lfx-chunk-reload-at';
+const RELOAD_LOOP_WINDOW_MS = 60_000;
 
 const isChunkLoadError = (error: unknown): boolean => {
   const message = error instanceof Error ? error.message : String(error);
@@ -19,23 +20,22 @@ const isChunkLoadError = (error: unknown): boolean => {
 export class ChunkLoadErrorHandler implements ErrorHandler {
   private readonly platformId = inject(PLATFORM_ID);
 
-  handleError(error: unknown): void {
-    if (!isPlatformBrowser(this.platformId)) {
-      return;
-    }
-
-    if (!isChunkLoadError(error)) {
+  public handleError(error: unknown): void {
+    // SSR + non-chunk errors fall through to default ErrorHandler behavior so logs surface.
+    if (!isPlatformBrowser(this.platformId) || !isChunkLoadError(error)) {
       console.error(error);
       return;
     }
 
-    // Guard against reload loops — only attempt once per session.
-    if (sessionStorage.getItem(RELOAD_FLAG)) {
-      console.error('Chunk load error persists after reload — giving up.', error);
+    // Timestamp-based loop guard: bail only if we already reloaded within the window.
+    const lastReloadStr = sessionStorage.getItem(RELOAD_FLAG);
+    const lastReload = lastReloadStr ? Number(lastReloadStr) : 0;
+    if (lastReload && Date.now() - lastReload < RELOAD_LOOP_WINDOW_MS) {
+      console.error('Chunk load error within 60s of reload — likely a loop, giving up.', error);
       return;
     }
 
-    sessionStorage.setItem(RELOAD_FLAG, '1');
+    sessionStorage.setItem(RELOAD_FLAG, String(Date.now()));
     window.location.reload();
   }
 }
--- a/apps/lfx-one/src/server/server.ts
+++ b/apps/lfx-one/src/server/server.ts
@@ -103,8 +103,23 @@ app.get('/readyz', (_req: Request, res: Response) => {
 app.get(
   '**',
   express.static(browserDistFolder, {
-    maxAge: '1y',
     index: false,
+    setHeaders: (res, filePath) => {
+      if (filePath.endsWith('index.html') || filePath.endsWith('index.csr.html')) {
+        // index.html must not be cached long-term — returning users need the latest
+        // chunk hashes after a deploy, otherwise dynamic imports fail with
+        // "TypeError: Importing a module script failed."
+        res.setHeader('Cache-Control', 'no-cache, must-revalidate');
+        return;
+      }
+      if (/-[A-Z0-9]{8,}\.(js|css|woff2?|ttf|otf|png|jpg|jpeg|svg|ico|webp)$/i.test(filePath)) {
+        // Angular emits content-hashed filenames (outputHashing: "all") — safe to
+        // cache permanently; the hash changes whenever content changes.
+        res.setHeader('Cache-Control', 'public, max-age=31536000, immutable');
+        return;
+      }
+      res.setHeader('Cache-Control', 'public, max-age=300');
+    },
   })
 );
 
--- a/apps/lfx-one/src/app/app.config.ts
+++ b/apps/lfx-one/src/app/app.config.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 import { provideHttpClient, withFetch, withInterceptors } from '@angular/common/http';
-import { ApplicationConfig, provideZonelessChangeDetection } from '@angular/core';
+import { ApplicationConfig, ErrorHandler, provideZonelessChangeDetection } from '@angular/core';
 import { provideClientHydration, withEventReplay, withHttpTransferCacheOptions, withIncrementalHydration } from '@angular/platform-browser';
 import { provideAnimationsAsync } from '@angular/platform-browser/animations/async';
 import { provideRouter, withInMemoryScrolling, withPreloading } from '@angular/router';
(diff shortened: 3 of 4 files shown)


andrest50 (dismissed):

apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
```diff
@@ -0,0 +1,41 @@
...
+  );
+};
+
+@Injectable()
+export class ChunkLoadErrorHandler implements ErrorHandler {
+  private readonly platformId = inject(PLATFORM_ID);
+
+  handleError(error: unknown): void {
+    if (!isPlatformBrowser(this.platformId)) {
+      return;
```
MRashad26: **Problem:** The `!isPlatformBrowser` branch returns silently with no logging:
```ts
handleError(error: unknown): void {
  if (!isPlatformBrowser(this.platformId)) {
    return;                       // ← swallows every SSR error
  }
  // ...
}
```
Angular invokes `ErrorHandler.handleError` for errors that occur during server-side rendering too — not just browser code. The default Angular `ErrorHandler` calls `console.error('ERROR', error)` (and `_findOriginalError` unwrapping) in both contexts, so by registering `ChunkLoadErrorHandler` as the global `ErrorHandler`, the SSR pod loses error visibility for any exception Angular catches internally (component constructors, lifecycle hooks, sync errors inside templates).

**Why this is a problem:**
1. **Silent SSR failures** — exceptions during a component constructor / `ngOnInit` that Angular catches and routes through `ErrorHandler` don't propagate to the render promise. The Express-level `.catch` at `apps/lfx-one/src/server/server.ts:325` only sees errors that bubble up to `angularApp.handle(req, …).catch(…)`; Angular's internally-caught component errors don't reach it. With this handler, those errors disappear — no stderr, no DataDog (RUM is browser-only), no pino log.
2. **Behavior regression vs. the default handler** — before this PR, Angular's default `ErrorHandler` was active and logged everything to console in both SSR and browser. After this PR, SSR silently drops everything because there's no chunk-load story on the server anyway.
3. **Hard to diagnose during incident** — a future SSR crash that's actually a constructor exception (e.g., a service throwing in `inject(...)`) will show as "page renders empty / partial" with zero logs to grep.

**Suggested fix:** Preserve the default-handler behavior on the server before bailing out:
```ts
handleError(error: unknown): void {
  if (!isPlatformBrowser(this.platformId)) {
    console.error(error);   // mirror default ErrorHandler behavior so SSR logs surface
    return;
  }
  // ... rest unchanged
}
```
Alternatively, fold the SSR + non-chunk-error paths together since they both want the default behavior:
```ts
handleError(error: unknown): void {
  if (!isPlatformBrowser(this.platformId) || !isChunkLoadError(error)) {
    console.error(error);
    return;
  }
  // chunk-error recovery below
}
```

apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    if (!isChunkLoadError(error)) {
+      console.error(error);
+      return;
+    }
+
+    // Guard against reload loops — only attempt once per session.
+    if (sessionStorage.getItem(RELOAD_FLAG)) {
```
MRashad26: **Problem:** The reload flag is set once and never cleared, so it persists for the entire tab lifetime:
```ts
if (sessionStorage.getItem(RELOAD_FLAG)) {
  console.error('Chunk load error persists after reload — giving up.', error);
  return;
}
sessionStorage.setItem(RELOAD_FLAG, '1');
window.location.reload();
```
Walk-through for a long-lived tab:
1. User opens the app on Monday. No flag.
2. Tuesday morning: deploy A ships. User clicks lazy route → chunk error → flag set to `'1'` → reload. Fresh `index.html` has matching chunks; user browses fine.
3. Friday afternoon: deploy B ships. User clicks lazy route in the same tab → chunk error → flag is **still** `'1'` from Tuesday → `console.error('persists after reload — giving up.')` → user is stuck on the old `index.html`, no recovery, no error UX.

**Why this is a problem:**
1. **Recovery only works for the first deploy per session.** The PR's own goal — "reload users whose browsers have the stale `index.html` cached" — silently stops working after the first reload, for the rest of the tab's life. Single-page apps + browser tab restoration mean sessions can last days.
2. **The error message is misleading** — "persists after reload" implies the error happened immediately after a reload (the loop scenario the guard is trying to catch). In the long-lived-session case, the previous reload was hours/days ago.
3. **The loop-guard intent is preserved with a timestamp** — what the guard actually needs to prevent is "two reloads within a few seconds of each other". A boolean flag conflates that with "this tab has ever reloaded for any reason".

**Suggested fix:** Use a timestamp + short window, so legitimate later failures trigger a fresh reload but immediate-loop scenarios still bail:
```ts
const RELOAD_FLAG = 'lfx-chunk-reload-at';
const RELOAD_LOOP_WINDOW_MS = 60_000;

// in handleError, after the chunk-error check:
const lastReloadStr = sessionStorage.getItem(RELOAD_FLAG);
const lastReload = lastReloadStr ? Number(lastReloadStr) : 0;
if (lastReload && Date.now() - lastReload < RELOAD_LOOP_WINDOW_MS) {
  console.error('Chunk load error within 60s of reload — likely a loop, giving up.', error);
  return;
}
sessionStorage.setItem(RELOAD_FLAG, String(Date.now()));
window.location.reload();
```
The window can be tuned (60s is conservative), but the principle is: only treat reloads that are *close together in time* as a loop, not "this tab has ever reloaded".

MRashad26 (commented):
Strict pass against `~/LFX/code-enforcer-agent.md`. Diagnosis and fix are well-targeted — `index.html` / `index.csr.html` correctly split from hashed assets, the regex catches Angular's lowercase hex hashes via `/i`, and the `setHeaders` callback fires off `express.static` so SSR responses aren't affected. The new `ChunkLoadErrorHandler` is a clean global recovery surface. Two issues on the handler below, both worth fixing before merge.

apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
MRashad26: Done — combined the `!isPlatformBrowser` and `!isChunkLoadError` guards into a single branch so all non-chunk-error paths (SSR or browser) call `console.error(error)` before returning, matching the default `ErrorHandler` behavior.

See commit `c53526b3`: `fix(server): address ChunkLoadErrorHandler review feedback (LFXV2-1920)`.

apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
MRashad26: Done — replaced the boolean `lfx-chunk-reload-attempted` flag with a timestamp (`lfx-chunk-reload-at`) and a 60-second window. Reloads that happen within 60s of each other are treated as a loop and stopped; reloads separated by more than 60s (e.g., a second deploy later in the same session) trigger a fresh recovery attempt.

See commit `c53526b3`: `fix(server): address ChunkLoadErrorHandler review feedback (LFXV2-1920)`.

MRashad26:
## Review Feedback Addressed

Commit: `c53526b3fad057b6a31aba87589a54d933abcc94`

### Changes Made

- **`chunk-load-error.handler.ts`**: Combined the `!isPlatformBrowser` and `!isChunkLoadError` branches so all non-chunk-error paths (SSR or browser) call `console.error(error)` before returning — preserving default `ErrorHandler` behavior and surfacing server-side Angular exceptions in logs (per self-review L24).
- **`chunk-load-error.handler.ts`**: Replaced the boolean `lfx-chunk-reload-attempted` flag with a timestamp (`lfx-chunk-reload-at`) + 60-second window. Only reloads within 60s of each other are treated as a loop; long-lived tabs that survive a subsequent deploy can still recover (per self-review L33).
- **`chunk-load-error.handler.ts`**: Added `public` accessor on `handleError` (pre-existing ESLint warning, fixed as a free improvement).

### Threads Resolved

2 of 2 unresolved threads addressed and resolved.

andrest50 (approved):

apps/lfx-one/src/server/server.ts
```diff
@@ -103,8 +103,23 @@ app.get('/readyz', (_req: Request, res: Response) => {
...
+        // "TypeError: Importing a module script failed."
+        res.setHeader('Cache-Control', 'no-cache, must-revalidate');
+        return;
+      }
+      if (/-[A-Z0-9]{8,}\.(js|css|woff2?|ttf|otf|png|jpg|jpeg|svg|ico|webp)$/i.test(filePath)) {
+        // Angular emits content-hashed filenames (outputHashing: "all") — safe to
+        // cache permanently; the hash changes whenever content changes.
+        res.setHeader('Cache-Control', 'public, max-age=31536000, immutable');
+        return;
+      }
```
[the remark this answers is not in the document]
MRashad26: The `max-age=300` fallback for non-hashed assets is intentional. Files like `favicon.ico` and the images in `public/images/` have no content hash, so they can't be cached immutably — if they change, a long TTL would leave stale copies in browsers with no way to bust them. A 5-minute TTL is the right call for unhashed assets: short enough to pick up any changes, long enough to avoid unnecessary origin hits. No caching regression here.

apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: MIT
+
+import { isPlatformBrowser } from '@angular/common';
+import { ErrorHandler, inject, Injectable, PLATFORM_ID } from '@angular/core';
+
+const RELOAD_FLAG = 'lfx-chunk-reload-at';
+const RELOAD_LOOP_WINDOW_MS = 60_000;
+
+const isChunkLoadError = (error: unknown): boolean => {
+  const message = error instanceof Error ? error.message : String(error);
```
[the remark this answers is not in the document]
MRashad26: In Angular's `ErrorHandler` contract, the framework unwraps `ErrorEvent`, `PromiseRejectionEvent`, and zone.js wrappers before calling `handleError()` — the handler always receives the inner `Error` object, not the raw browser event. The `error instanceof Error ? error.message : String(error)` pattern is the appropriate normalization at this layer, covering the rare case where something non-Error slips through. No change needed.

apps/lfx-one/src/app/shared/utils/chunk-load-error.handler.ts
```diff
@@ -0,0 +1,41 @@
...
+    // SSR + non-chunk errors fall through to default ErrorHandler behavior so logs surface.
+    if (!isPlatformBrowser(this.platformId) || !isChunkLoadError(error)) {
+      console.error(error);
+      return;
+    }
+
+    // Timestamp-based loop guard: bail only if we already reloaded within the window.
+    const lastReloadStr = sessionStorage.getItem(RELOAD_FLAG);
+    const lastReload = lastReloadStr ? Number(lastReloadStr) : 0;
+    if (lastReload && Date.now() - lastReload < RELOAD_LOOP_WINDOW_MS) {
```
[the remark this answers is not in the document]
MRashad26: Valid point — `sessionStorage` can throw in sandboxed iframes or when storage is blocked by the browser. The risk is low in this app (no iframe embedding, no storage-restricted contexts), but wrapping in try/catch would be the defensive correct approach. Noted as a follow-up improvement; deferring from this PR since the primary fix (cache headers) is the critical path and this is a belt-and-suspenders edge case.

PR #762, opened 2026-05-21
fix(profile): remove johnlf2727 demo user from individual enrollment

fayazg:
## Summary

- Removed the hardcoded `johnlf2727` demo-user short-circuit in `EnrollmentService`. It returned a stale Expired-2021 TLF Individual Supporter stub and was not gated by env or feature flag, so it shipped to production.
- Removed the matching short-circuit in `updateAutoRenew()` which was silently dropping the upstream PATCH for that username — if any real LF SSO user ever holds `johnlf2727`, their auto-renew toggles would vanish.
- Impersonation via Auth0 CTE is the correct way to demo enrollments against real data in lower environments, making the stub redundant.
- Clarified `docs/migration/individual-memberships.md` to note the demo path stays in legacy myprofile and was not carried into self-serve.

Fixes LFXV2-1664

## Test plan

- [ ] `yarn check-types` passes
- [ ] `yarn lint:check` passes
- [ ] `yarn build` passes (SSR bundle)
- [ ] Impersonate a real LF account with an Individual Supporter membership in a lower env → `/profile` → Individual Enrollments tab renders real member-service data
- [ ] Toggle auto-renew while impersonating → confirm the PATCH to `/member-service/v2/memberships/<id>` fires (server log `update_individual_enrollment_auto_renew`) and the upstream state changes
- [ ] If a `johnlf2727` account exists in a lower env, confirm it no longer sees the Expired-2021 stub and auto-renew writes succeed

### DIFF
--- a/apps/lfx-one/e2e/individual-enrollment-robust.spec.ts
+++ b/apps/lfx-one/e2e/individual-enrollment-robust.spec.ts
@@ -30,7 +30,7 @@ test.describe('Individual Enrollment — Structural Tests', () => {
     });
   });
 
-  test.describe('Card structure (demo user path)', () => {
+  test.describe('Card structure', () => {
     test('should render at least one enrollment card', async ({ page }) => {
       const card = page.getByTestId('individual-enrollment-card').first();
       await expect(card).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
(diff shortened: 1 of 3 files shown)


fayazg:
**Follow-up commit `dc4096c`:** removed the `### Demo user` section from `docs/migration/individual-memberships.md` entirely. The section described a myprofile-only behavior (`johnlf2727` returning mock data from `dummy-data.ts`) that was never carried into this codebase — keeping it as a clarifying note added noise without value to self-serve developers.

fayazg:
**Follow-up commit `2043a4b`:** cleaned up remaining demo user references missed in the initial pass — two table rows in `docs/migration/individual-memberships.md` that described `dummy-data.ts` as "demo user" data, and a stale E2E `test.describe` label `'Card structure (demo user path)'` renamed to `'Card structure'` (the tests themselves are structural checks against `data-testid`s, unrelated to any specific user).

MRashad26 (approved):

PR #788, opened 2026-05-26
feat(org-lens): add all employees tab on org people page

ahmedomosanya:
## Summary

- Adds `OrgLensPeopleService` (Snowflake) and `OrgLensPeopleController` exposing two endpoints under `/api/orgs/:accountId/lens/people`:
  - `GET /all` — bundled list (rows + 5 stat-card aggregates + engaged-foundation options) for a single UI subscription on tab load.
  - `GET /:personKey/detail` — lazy per-person detail (board/committee seats, code contributions, events, training) fetched only when a row is expanded.
- Both endpoints narrow upstream free-text values (voting status, training status) at the BFF boundary so the UI types stay strict.
- Adds `PERSON_KEY_PATTERN` to shared regex constants so the detail route validates either an LFID or a `cdp:`-prefixed identifier before touching Snowflake.
- Adds `OrgAllEmployee*` shared interfaces, plus `EMPTY_ORG_ALL_EMPLOYEES_RESPONSE` seed, activity-filter options, and a 30-row initial visible cap.
- Adds `AllEmployeesComponent` + `AllEmployeesDetailComponent` with a filter bar (search, foundation, activity), 5 stat cards, sortable columns, chevron-toggled detail rows, and skeleton placeholders seeded from the empty response so the table never flashes empty.
- Wires `OrgPeopleComponent` so the `all` tab renders the new component instead of the generic empty state.

- The other People tabs (maintainers, board, committees, …) still render the generic empty state — populated in follow-up PRs.
- No persisted per-user UI state (sort column, expansion set, "Show All" toggle) — resets on tab re-mount by design.
- No pagination/virtualization beyond the initial 30-row visible cap + "Show All" toggle.

## Test plan

- [ ] Tab loads on an account with employees: stat cards populated, table renders rows, foundation/activity dropdowns reflect the account's actual engagement.
- [ ] Tab loads on an empty/no-employees account: stat cards show 0s, table shows the seeded empty response (no spinner flash).
- [ ] Tab loads while the account is still resolving: skeleton placeholders render for stat cards and table rows, no layout shift on data arrival.
- [ ] Search narrows rows client-side (name, title, email substring).
- [ ] Foundation filter narrows rows to people engaged with the selected foundation.
- [ ] Activity filter (`all` / `governance` / `code` / `events` / `training`) narrows rows by activity dimension.
- [ ] Sorting toggles on each sortable column (name, seats, commits, events, courses) and reverses on second click.
- [ ] "Show All" reveals rows beyond the initial 30-row cap.
- [ ] Row chevron expands an inline detail panel; first expansion fires `GET /detail`, second expansion of the same row is instant (cached in `detailMap`).
- [ ] Detail panel renders all four sub-sections (board/committee seats, code, events, training), each handling the empty-array case cleanly (HTTP 200, not 404).
- [ ] Server logs include `get_org_lens_people_all` and `get_org_lens_people_detail` start/success entries with row/foundation counts.
- [ ] `Cache-Control: no-store` set on both endpoints (verified in network tab).
- [ ] Invalid `accountId` (non-Salesforce shape) or invalid `personKey` returns a `ServiceValidationError` (400), not a 500.

[LFXV2-1870]: https://linuxfoundation.atlassian.net/browse/LFXV2-1870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/shared/src/constants/regex.constants.ts
```diff
@@ -32,3 +32,6 @@ export const FOUNDATION_ID_PATTERN = /^[A-Za-z0-9-]{1,64}$/;
 
 /** Basic email-format regex for client-side blur validation (FR-017a). */
 export const EMAIL_REGEX = /^[^@\s]+@[^@\s]+\.[^@\s]+$/;
+
+/** Org People person_key — either an LFID or a `cdp:`-prefixed UUID-style id. */
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `d055b8c7` — docstring softened to "opaque \`cdp:\`-prefixed identifier". The regex stays as the deliberate URL-safe 4–128-char request-boundary bound; CDP IDs from the persona service aren't canonical UUIDs, so the docstring was the inaccurate side.

apps/lfx-one/src/server/services/org-lens-people.service.ts
```diff
@@ -0,0 +1,441 @@
...
+
+  private async fetchAllEmployeeRows(accountId: string): Promise<OrgAllEmployeeRow[]> {
+    const query = `
+      SELECT
+        ACCOUNT_ID,
+        PERSON_KEY,
+        LFID,
+        LF_USERNAME,
+        CDP_MEMBER_ID,
+        NAME,
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `d055b8c7` — dropped `LF_USERNAME` from both the `SELECT` and the `OrgPeopleAllRow` interface. Wire shape unchanged.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
```diff
@@ -0,0 +1,259 @@
...
+            <th class="px-4 py-3 font-medium">Org Lens Access</th>
+          </tr>
+        </thead>
+        <tbody>
+          @for (row of visibleRows(); track row.personKey) {
+            <tr
+              class="cursor-pointer border-b border-gray-100 hover:bg-gray-50"
+              [attr.data-testid]="'org-people-all-employees-row-' + row.personKey"
+              [attr.aria-expanded]="isExpanded(row.personKey)"
+              (click)="toggleRow(row)">
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `d055b8c7` — added `role="button"`, `tabindex="0"`, `aria-label`, `focus-visible` ring, and a `(keydown)` handler that calls `toggleRow` on Enter/Space.

apps/lfx-one/src/app/modules/dashboards/org/org-people/services/all-employees.service.ts
```diff
@@ -0,0 +1,42 @@
...
+export class AllEmployeesService {
+  private readonly http = inject(HttpClient);
+
+  public getAllEmployees(accountId: string): Observable<OrgAllEmployeesResponse> {
+    return this.http.get<OrgAllEmployeesResponse>(`/api/orgs/${encodeURIComponent(accountId)}/lens/people/all`).pipe(
+      catchError(() =>
+        of<OrgAllEmployeesResponse>({
+          accountId,
+          rows: [],
+          stats: { activeInOss: 0, inGovernance: 0, codeContributors: 0, eventAttendees: 0, trainees: 0 },
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `d055b8c7` — fallback now spreads the shared `EMPTY_ORG_ALL_EMPLOYEES_RESPONSE` with the per-call `accountId` override. Single source of truth in `@lfx-one/shared/constants`.

apps/lfx-one/src/app/modules/dashboards/org/org-people/services/all-employees.service.ts
```diff
@@ -0,0 +1,42 @@
...
+      )
+    );
+  }
+
+  public getEmployeeDetail(accountId: string, personKey: string): Observable<OrgAllEmployeeDetail> {
+    return this.http.get<OrgAllEmployeeDetail>(`/api/orgs/${encodeURIComponent(accountId)}/lens/people/${encodeURIComponent(personKey)}/detail`).pipe(
+      catchError(() =>
+        of<OrgAllEmployeeDetail>({
+          personKey,
+          boardSeats: [],
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `d055b8c7` — added `EMPTY_ORG_ALL_EMPLOYEE_DETAIL` to `org-people.constants.ts` alongside the response constant; service fallback now spreads it with a per-call `personKey` override.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,309 @@
...
+    const q = this.searchTerm().trim().toLowerCase();
+    const foundationId = this.selectedFoundationId();
+    const activity = this.selectedActivity();
+
+    return rows.filter((row) => {
+      if (q) {
+        const inName = row.name.toLowerCase().includes(q);
+        const inTitle = (row.title ?? '').toLowerCase().includes(q);
+        if (!inName && !inTitle) return false;
+      }
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `d055b8c7` — search predicate now includes `row.email` (null-safe via `(row.email ?? '').toLowerCase()`). Matches the documented behavior in the PR description.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,309 @@
...
+    if (this.detailLoading()[row.personKey]) return;
+    const accountId = this.accountContext.selectedAccount().accountId;
+    if (!accountId) return;
+
+    this.detailLoading.update((state) => ({ ...state, [row.personKey]: true }));
+    this.dataService
+      .getEmployeeDetail(accountId, row.personKey)
+      .pipe(take(1))
+      .subscribe({
+        next: (detail) => {
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `d055b8c7` — added a stored `DestroyRef` (`inject(DestroyRef)`) and chained `takeUntilDestroyed(this.destroyRef)` after `take(1)` on the lazy-detail subscription, so a mid-flight unmount no longer retains the component.

apps/lfx-one/src/app/modules/dashboards/org/org-people/services/all-employees.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/** HTTP client for the Org Lens "All Employees" BFF endpoints. */
+@Injectable({ providedIn: 'root' })
+export class AllEmployeesService {
+  private readonly http = inject(HttpClient);
+
+  public getAllEmployees(accountId: string): Observable<OrgAllEmployeesResponse> {
+    return this.http
+      .get<OrgAllEmployeesResponse>(`/api/orgs/${encodeURIComponent(accountId)}/lens/people/all`)
+      .pipe(catchError(() => of<OrgAllEmployeesResponse>({ ...EMPTY_ORG_ALL_EMPLOYEES_RESPONSE, accountId })));
```
audigregorie: **Major — Service swallows HTTP errors; component cannot distinguish failure from empty data**

`catchError(() => of(EMPTY_ORG_ALL_EMPLOYEES_RESPONSE))` converts any network/500 failure into a successful empty payload. The component then renders zero employees and zero stat cards with no error or retry affordance — indistinguishable from a legitimately empty org.

This also violates the repo convention that **components own loading/error state** (see `OrgMembershipsComponent`, which handles `catchError` locally and exposes `fetchError` + retry). `OrgLensMembershipsService` returns bare `http.get` observables with no error masking.

```suggestion
      .get<OrgAllEmployeesResponse>(`/api/orgs/${encodeURIComponent(accountId)}/lens/people/all`);
```

Handle errors in `AllEmployeesComponent` (e.g. `fetchError` signal + retry button), mirroring the memberships tab.

apps/lfx-one/src/app/modules/dashboards/org/org-people/services/all-employees.service.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public getEmployeeDetail(accountId: string, personKey: string): Observable<OrgAllEmployeeDetail> {
+    return this.http
+      .get<OrgAllEmployeeDetail>(`/api/orgs/${encodeURIComponent(accountId)}/lens/people/${encodeURIComponent(personKey)}/detail`)
+      .pipe(catchError(() => of<OrgAllEmployeeDetail>({ ...EMPTY_ORG_ALL_EMPLOYEE_DETAIL, personKey })));
```
audigregorie: **Major — Failed detail fetch cached as empty detail; no retry path**

When the detail HTTP call fails, this `catchError` emits `EMPTY_ORG_ALL_EMPLOYEE_DETAIL`. The component's `loadDetailIfNeeded` `next` handler stores that object in `detailMap`, and subsequent expansions short-circuit on `if (this.detailMap()[row.personKey]) return`. The user sees "No LF activity detail recorded" and cannot retry without a full page refresh or account switch.

```suggestion
      .get<OrgAllEmployeeDetail>(`/api/orgs/${encodeURIComponent(accountId)}/lens/people/${encodeURIComponent(personKey)}/detail`);
```

Let errors propagate to the component; only cache successful responses. Consider a per-row `detailError` map with a "Retry" action in the detail panel.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,317 @@
...
+    this.dataService
+      .getEmployeeDetail(accountId, row.personKey)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (detail) => {
+          this.detailMap.update((state) => ({ ...state, [row.personKey]: detail }));
+          this.clearDetailLoading(row.personKey);
+        },
+        error: () => this.clearDetailLoading(row.personKey),
+      });
```
audigregorie, line 337: **Major — In-flight detail requests not cancelled on account change; stale data can leak**

`loadDetailIfNeeded` captures `accountId` at call time and subscribes with `takeUntilDestroyed(this.destroyRef)` — which only fires on component destroy, not when the user switches accounts. Meanwhile `resetAllState()` (triggered on account change) clears `detailMap`, but a stale in-flight response from the previous account can still land in `next` and repopulate the map.

Additionally, `detailMap` is keyed by `personKey` alone. If the same `personKey` exists under two accounts, cached detail from account A could display under account B.

Consider keying the cache as `` `${accountId}:${personKey}` ``, and cancelling in-flight detail subscriptions when `selectedAccount` changes (e.g. `switchMap` + `takeUntil` tied to account changes, or an `AbortController`/`switchMap` per expansion).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  protected readonly totalFiltered = computed(() => this.sortedRows().length);
+
+  protected readonly visibleRows = computed(() => this.sortedRows().slice(0, this.limit()));
+
+  protected readonly canShowMore = computed(() => this.limit() < this.totalFiltered());
+
+  protected readonly footerCountLabel: Signal<string> = computed(() => this.initFooterCountLabel());
+
+  protected readonly hasNoCompany = computed(() => !this.accountContext.selectedAccount().accountId);
```
audigregorie: **Minor — Unused `hasNoCompany` computed**

This signal is declared but never referenced in the template or elsewhere in the component. The parent `OrgPeopleComponent` already gates rendering with `@if (!hasCompany())`, so this is dead code.

Remove it to avoid confusion about whether the child handles the no-company case independently.

audigregorie (commented):
## Code Review Summary

Solid first slice of the Org People "All Employees" tab — shared types, BFF validation, parallel Snowflake queries, and skeleton-first UI all follow established Org Lens patterns. The main gaps are around **error handling and request lifecycle**: HTTP failures are silently converted to empty payloads in the service, failed detail fetches are cached as empty detail with no retry path, and in-flight detail requests are not cancelled when the selected account changes.

### Major — outside the diff

- **No automated tests** were added for the new BFF service/controller or the Angular table/detail components. Given ~1,550 LOC of new surface area, at least unit tests for `mapVotingStatus`, `parseFoundationIdArray`, and the component filter/sort logic would reduce regression risk.

### What's done well

- BFF boundary validation (`SALESFORCE_ACCOUNT_ID_PATTERN`, `PERSON_KEY_PATTERN`) with `ServiceValidationError` before Snowflake is hit.
- Bundled `/all` endpoint and lazy `/detail` expansion keep initial tab load lean.
- Skeleton placeholders seeded from `EMPTY_ORG_ALL_EMPLOYEES_RESPONSE` avoid layout flash; `Cache-Control: no-store` on both endpoints.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
```diff
@@ -0,0 +1,263 @@
...
+      <table class="w-full text-sm" data-testid="org-people-all-employees-rows">
+        <thead class="bg-gray-50">
+          <tr class="border-b border-gray-200 text-left text-xs uppercase tracking-wide text-gray-500">
+            <th class="w-7 px-2 py-3"></th>
+            <th class="cursor-pointer select-none px-4 py-3 font-medium" (click)="onSort('name')">
+              <span class="inline-flex items-center gap-1">
+                Employee
+                <i [class]="sortIcon('name')" aria-hidden="true"></i>
+              </span>
+            </th>
```
audigregorie: **Minor — Sortable column headers lack `aria-sort` and keyboard support**

The sortable `<th>` elements use `(click)` handlers but don't expose sort state to assistive technology (`aria-sort="ascending|descending|none"`) and aren't focusable/activatable via keyboard (unlike the row expand buttons which handle Enter/Space).

Consider adding `[attr.aria-sort]` bound to the active column/direction, `tabindex="0"`, and a `(keydown)` handler mirroring `onSort`, or wrapping the label in a `<button type="button">` inside the header cell.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,317 @@
...
+    this.sortDirection.set(column === 'name' ? 1 : -1);
+  }
+
+  protected sortIcon(column: OrgAllEmployeeSortColumn): 'fa-light fa-sort' | 'fa-light fa-sort-up' | 'fa-light fa-sort-down' {
+    if (this.sortColumn() !== column) return 'fa-light fa-sort';
+    return this.sortDirection() === 1 ? 'fa-light fa-sort-up' : 'fa-light fa-sort-down';
+  }
+
+  protected onRowKeydown(event: KeyboardEvent, row: OrgAllEmployeeRow): void {
+    if (event.key !== 'Enter' && event.key !== ' ') return;
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `c29357ac` — `onRowKeydown` now early-returns when `event.target !== event.currentTarget`, so Enter/Space on the inner name `<button>` no longer also toggles the row.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,317 @@
...
+    this.detailLoading.update((state) => ({ ...state, [row.personKey]: true }));
+    this.dataService
+      .getEmployeeDetail(accountId, row.personKey)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (detail) => {
+          this.detailMap.update((state) => ({ ...state, [row.personKey]: detail }));
+          this.clearDetailLoading(row.personKey);
+        },
+        error: () => this.clearDetailLoading(row.personKey),
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `c29357ac` — `loadDetailIfNeeded` now stamps the in-flight request with its `accountId` and the `next`/`error` handlers early-return if `this.accountContext.selectedAccount().accountId !== accountId`. A response that arrives after an account switch no longer repopulates `detailMap` / `detailLoading` with stale data.

apps/lfx-one/src/app/modules/dashboards/org/org-people/services/all-employees.service.ts
ahmedomosanya: Addressed in e461d3e2. `AllEmployeesService.getAllEmployees` no longer catches; the component now owns loading/error state with `fetchError` + `retryTrigger` (mirroring `OrgMembershipsComponent`'s `combineLatest([accountId$, retryTrigger$])` + `catchError` pattern). When the fetch fails the page renders an `<lfx-empty-state>` error block with a Retry CTA that increments `retryTrigger`.

apps/lfx-one/src/app/modules/dashboards/org/org-people/services/all-employees.service.ts
ahmedomosanya: Addressed in e461d3e2. `getEmployeeDetail` no longer catches; the component captures the error into a per-row `detailErrorMap` keyed by `${accountId}:${personKey}`, `AllEmployeesDetailComponent` renders an error block with a Retry button, and `retryDetail(row)` invalidates the cached entries and re-issues the fetch.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
ahmedomosanya: Addressed in e461d3e2. Two changes: (1) `detailMap`, `detailLoading`, and `detailErrorMap` are now keyed by `${accountId}:${personKey}` (via a `detailKey()` helper), so cross-account writes can't collide even if the same `personKey` exists in two accounts; (2) the account-change subscription now calls `detailCancel$.next()` and every detail fetch pipes through `takeUntil(detailCancel$)`, so switching accounts cancels in-flight HTTP outright — no wasted server work and no stale-write race. The c29357ac account-id guard is now redundant and was removed.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
ahmedomosanya: Removed in e461d3e2. Parent `OrgPeopleComponent` already gates rendering on `hasCompany()`, so the child computed was dead.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
ahmedomosanya: Addressed in e461d3e2. Each sortable `<th>` now exposes `[attr.aria-sort]` (precomputed in `ariaSortMap` so the template stays method-call-free) and wraps its label in a real `<button type="button">`, which inherits keyboard activation and focus styling. The header is read by screen readers as the active sort column with direction.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
```diff
@@ -0,0 +1,263 @@
...
+              <span class="inline-flex items-center gap-1">
+                Events
+                <i [class]="sortIcon('events')" aria-hidden="true"></i>
+              </span>
+            </th>
+            <th class="cursor-pointer select-none px-4 py-3 font-medium" (click)="onSort('courses')">
+              <span class="inline-flex items-center gap-1">
+                Training
+                <i [class]="sortIcon('courses')" aria-hidden="true"></i>
+              </span>
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in e461d3e2 alongside @audigregorie's parallel finding. Each sortable `<th>` now has `[attr.aria-sort]` bound to a precomputed `ariaSortMap` and wraps the label in a real `<button type="button">` (which carries Enter/Space activation and focus-visible styling for free). I went with `<button>` inside `<th>` rather than `role="button"` + `tabindex` + manual keydown to stay closer to native semantics.

apps/lfx-one/src/server/services/org-lens-people.service.ts
```diff
@@ -0,0 +1,439 @@
...
+  COURSE_NAME: string | null;
+}
+
+const EMPTY_STATS: OrgAllEmployeeStats = {
+  activeInOss: 0,
+  inGovernance: 0,
+  codeContributors: 0,
+  eventAttendees: 0,
+  trainees: 0,
+};
```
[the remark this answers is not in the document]
ahmedomosanya: Respectfully pushing back — these are server-internal Snowflake raw-row shapes (UPPERCASE_SNAKE column names) used only to type the driver's response inside the service before it's mapped to the camelCase wire contract (`OrgAllEmployees*`, which **is** in `@lfx-one/shared`). The established convention in this repo is to colocate these `XxxRow` types with the service that maps them — see precedent across 5 sibling services in `apps/lfx-one/src/server/services/`: `organization.service.ts` (`OrgLensAccountContextRow`), `org-involvement.service.ts` (6 row interfaces), `org-lens-memberships.service.ts` (3 raw row interfaces), `org-lens-foundations.service.ts` (`RawRow`), and `org-lens-people.service.ts` itself in its existing form. Moving these would diverge from sibling services without removing duplication, since each is single-use and tied to one query. The shared-types rule applies to wire contracts, which are already in `@lfx-one/shared` here.

apps/lfx-one/src/server/services/org-lens-people.service.ts
```diff
@@ -0,0 +1,439 @@
...
+    for (const row of trainingRows) {
+      if (row.COURSE_ID) {
+        distinctCourseIds.add(row.COURSE_ID);
+        if (row.STATUS === 'Certified') {
+          distinctCertifiedCourseIds.add(row.COURSE_ID);
+        }
+      }
+    }
+    const coursesCount = distinctCourseIds.size;
+    const certificationsCount = distinctCertifiedCourseIds.size;
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in e461d3e2. The counting loop now resolves `courseId` via `row.COURSE_ID ?? row.COURSE_OR_CERT_ID`, matching the fallback in `mapTrainingRow` at the same identifier resolution, so `coursesCount` and `certificationsCount` can no longer undercount when `COURSE_ID` is null but `COURSE_OR_CERT_ID` is present.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,391 @@
...
+  protected readonly ariaSortMap: Signal<Record<OrgAllEmployeeSortColumn, 'ascending' | 'descending' | 'none'>> = computed(() => this.initAriaSortMap());
+
+  public constructor() {
+    // Reset all state and cancel any in-flight detail fetches when the selected account changes; skip the initial emission on mount.
+    toObservable(this.accountContext.selectedAccount)
+      .pipe(skip(1), takeUntilDestroyed())
+      .subscribe(() => {
+        this.detailCancel$.next();
+        this.resetAllState();
+      });
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 632d262c. The reset subscription now subscribes to `this.accountId$` (`toObservable(selectedAccount).pipe(map(a => a.accountId), distinctUntilChanged())`) instead of the raw `selectedAccount` signal, so a Snowflake enrichment that re-sets the same `accountId` on a fresh object reference no longer fires the reset path and the user's search/filter/sort state is preserved.

apps/lfx-one/src/server/services/org-lens-people.service.ts
```diff
@@ -0,0 +1,440 @@
...
+}
+
+/** Coerce Snowflake `Date | string | null` into the ISO date-string the client interfaces declare. */
+function toDateString(value: Date | string | null | undefined): string | null {
+  if (!value) return null;
+  if (value instanceof Date) {
+    return value.toISOString().slice(0, 10);
+  }
+  if (typeof value === 'string') {
+    return value.slice(0, 10);
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 632d262c. `toDateString` now matches the `ProjectService.toIsoDate` precedent: regex-extract `YYYY-MM-DD` prefix on string input, fall back to `new Date(String(value))` with a `Number.isNaN(date.getTime())` check, return `null` on unparseable. Non-ISO strings can no longer leak `"Invalid Da"` to the client.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees-detail.component.html
```diff
@@ -0,0 +1,226 @@
...
+
+<div class="bg-gray-50 px-4 pb-4 pt-3 sm:pl-14" data-testid="org-people-all-employees-detail">
+  @if (loading()) {
+    <div class="text-xs italic text-gray-400" data-testid="org-people-all-employees-detail-loading">Loading details…</div>
+  } @else if (error()) {
+    <div class="flex items-center justify-between gap-3 py-2" data-testid="org-people-all-employees-detail-error">
+      <span class="text-sm text-red-600">
+        <i class="fa-light fa-triangle-exclamation mr-1.5" aria-hidden="true"></i>
+        Failed to load activity detail.
+      </span>
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 632d262c. The error block now has `role="alert"` and `aria-live="assertive"`, so screen readers announce the failure when it appears dynamically.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,391 @@
...
+      .getEmployeeDetail(accountId, row.personKey)
+      .pipe(
+        take(1),
+        takeUntil(this.detailCancel$),
+        takeUntilDestroyed(this.destroyRef),
+        finalize(() => this.clearDetailLoading(key))
+      )
+      .subscribe({
+        next: (detail) => this.detailMap.update((state) => ({ ...state, [key]: detail })),
+        error: () => this.detailErrorMap.update((state) => ({ ...state, [key]: true })),
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 632d262c. Added a `clearDetailError(key)` helper and call it at the start of `loadDetailIfNeeded`, so any path that retriggers a load (collapse + re-expand, or `retryDetail`) clears the stale per-row error before the new fetch starts. Also simplified `retryDetail` to drop its now-redundant error-clear.

apps/lfx-one/src/server/services/org-lens-people.service.ts
```diff
@@ -0,0 +1,439 @@
...
+    const result = await this.snowflakeService.execute<OrgPeopleAllRow & { ENGAGED_FOUNDATION_IDS: string | string[] | null }>(query, [accountId]);
+
+    return result.rows.map((row) => ({
+      personKey: row.PERSON_KEY,
+      lfid: row.LFID,
+      cdpMemberId: row.CDP_MEMBER_ID,
+      name: row.NAME ?? '',
+      title: row.TITLE,
+      email: row.EMAIL,
+      photoUrl: row.PHOTO,
```
dealako: **[minor] `photoUrl` is fetched and transmitted over the wire but never rendered**

`PHOTO` is `SELECT`ed from `ORG_PEOPLE_ALL`, mapped to `photoUrl` in the wire contract, and therefore included in every row of the bundled response — but `all-employees.component.html` always renders the initials-based avatar instead. For orgs with many employees, this transmits a nullable URL string per-row with no current consumer.

Suggested options:
- Adopt the photo in the avatar `` (with initials as fallback when `photoUrl` is null), or
- Drop `PHOTO` from the `SELECT` and `photoUrl` from `OrgAllEmployeeRow` until the avatar feature is confirmed in design

Not a correctness issue — just dead weight on the wire contract and shared interface.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees-detail.component.html
```diff
@@ -0,0 +1,226 @@
...
+                      @if (row.isSpeaker) {
+                        <span
+                          class="inline-flex items-center rounded-full border border-amber-200 bg-amber-50 px-2 py-0.5 text-[11px] font-semibold text-amber-700">
+                          Speaker
+                        </span>
+                      } @else {
+                        <span class="inline-flex items-center rounded-full bg-gray-100 px-2 py-0.5 text-[11px] text-gray-600">Attendee</span>
+                      }
+                    </td>
+                    <td class="px-3 py-2 font-semibold text-gray-700">{{ row.eventsCount }}</td>
```
dealako: **[nit] Aggregate count repeated on every event row — potentially confusing UX**

`eventsCount` is denormalized by the service (same value stamped on every `OrgAllEmployeeEvent` row). The "Total Events" column in this table therefore shows the same number on every row. A person who attended 5 events will see "5" in the Total Events cell for each of those 5 rows.

The API design (carrying the aggregate on the row) is deliberate and fine for the interface contract. But in the rendered table the repeated value could confuse readers ("why does every event say 5?"). Consider whether the column should sit in the section header (e.g., next to "Events" in `<h3>`) rather than repeating per row — consistent with how `coursesCount`/`certificationsCount` appear on every training row.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
```diff
@@ -0,0 +1,295 @@
...
+              <th class="px-4 py-3 font-medium">Org Lens Access</th>
+            </tr>
+          </thead>
+          <tbody>
+            @for (row of visibleRows(); track row.personKey) {
+              <tr
+                class="cursor-pointer border-b border-gray-100 hover:bg-gray-50 focus:bg-gray-50 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+                role="button"
+                tabindex="0"
+                [attr.data-testid]="'org-people-all-employees-row-' + row.personKey"
```
dealako, line 182: **[nit] `data-testid` colon from CDP `personKey`s may trip raw CSS selectors in tests**

For a CDP person key like `cdp:abc123def456`, the resulting testid is `org-people-all-employees-row-cdp:abc123def456`. Playwright's `getByTestId()` handles colons correctly, but any test that constructs a raw CSS attribute selector (e.g., `document.querySelector("[data-testid='org-people-all-employees-row-cdp:abc123def456']")`) would need `CSS.escape()` to avoid a SyntaxError.

No action needed if E2E tests exclusively use `getByTestId()` / `data-testid` helpers — just worth flagging for future test authors who might mix in raw selectors.

dealako (dismissed):
## Code Review — PR #788 · `feat(org-lens): add all employees tab on org people page`

---

### Security audit (Phase 1 & 2 — `lfx-security-engineer`)

✅ **No blockers found.**

| Check | Result |
|---|---|
| Secrets / credentials | No hardcoded tokens, keys, or passwords |
| SQL injection (OWASP A03) | All 6 Snowflake queries use `?` parameterized binding — no string concatenation |
| Broken access control (OWASP A01) | `authMiddleware` applied globally at `server.ts:187` before the `orgsRouter` mount at line 202 — both new endpoints inherit the same auth posture as every sibling Org Lens route |
| Input validation | `accountId` gated by `SALESFORCE_ACCOUNT_ID_PATTERN`; `personKey` gated by `PERSON_KEY_PATTERN` — both checked before Snowflake is touched |
| `Cache-Control` | `no-store` set on both endpoints ✓ |
| Sensitive data in logs | Controller logs `account_id`, `person_key`, row counts — no PII (name/email) in log fields |
| Error surfacing | Controller follows the logging pattern: `startOperation` + `success` + bare `next(error)` in catch — `apiErrorHandler` logs errors centrally |

---

### Revision tracking — all prior findings addressed

| Finding | Raised by | Resolved in |
|---|---|---|
| `PERSON_KEY_PATTERN` docstring vs. regex | Copilot | `d055b8c7` |
| Unused `LF_USERNAME` selected | Copilot | `d055b8c7` |
| Row keyboard accessibility | Copilot | `d055b8c7` |
| Service-level `catchError` swallowing errors → empty-response drift | Copilot + @audigregorie | `d055b8c7` → hardened `e461d3e2` |
| Search predicate missing `email` field | Copilot | `d055b8c7` |
| Detail subscription leak (no teardown) | Copilot | `d055b8c7` |
| Keydown bubbling from inner `<button>` | Copilot | `c29357ac` |
| Stale account-id writes to `detailMap` | Copilot + CodeRabbit | `c29357ac` → `e461d3e2` |
| Service swallowing HTTP errors (component can't distinguish failure from empty) | @audigregorie | `e461d3e2` |
| Cached empty detail with no retry path | @audigregorie | `e461d3e2` |
| In-flight detail requests not cancelled on account change | @audigregorie | `e461d3e2` |
| Unused `hasNoCompany` computed | @audigregorie | `e461d3e2` |
| `aria-sort` + keyboard support for sortable headers | @audigregorie + CodeRabbit | `e461d3e2` |
| `COURSE_ID`-null training undercount | CodeRabbit | `e461d3e2` |
| Reset firing on same-id account object re-set | Copilot | `632d262c` |
| `toDateString` non-ISO string leak (`"Invalid Da"`) | Copilot | `632d262c` |
| `role="alert"` on detail error block | CodeRabbit | `632d262c` |
| Stale per-row error not cleared on retry | CodeRabbit | `632d262c` |

The rebuttal on CodeRabbit's "move row interfaces to `@lfx-one/shared`" comment was well-reasoned and accepted — server-internal `XxxRow` Snowflake driver shapes are correctly co-located with the service, consistent with five sibling services. ✓

---

### Independent findings

**[minor] (×2) — see inline comments**

- **`photoUrl` transmitted but not rendered** — `PHOTO` is selected and appears in the wire contract, but the template always falls back to initials. Dead weight until the avatar feature is confirmed. *(inline comment on `org-lens-people.service.ts:206`)*
- **No E2E test coverage for the new tab** — no `*.spec.ts` or `*-robust.spec.ts` files added. The PR's manual test plan is thorough, but per `docs/architecture/testing/` conventions, automated coverage should follow. Acceptable as a tracked follow-up if JIRA ticket is created.

**[nit] (×2) — see inline comments**

- **Aggregate `eventsCount`/`coursesCount`/`certificationsCount` repeated per-row in detail tables** — same total shown on every event row and every training row due to service-level denormalization. Technically correct but potentially confusing UX. *(inline comment on `all-employees-detail.component.html:175`)*
- **`data-testid` colon from CDP `personKey`s** — `cdp:abc123` produces a testid with a colon; safe for Playwright's `getByTestId()` but raw CSS attribute selectors would break. *(inline comment on `all-employees.component.html:175`)*

---

### Overall impression

Hey Ahmed — nice work getting this shipped. The all-employees tab is a meaty full-stack slice: six Snowflake queries parallelized correctly, a thoughtful BFF narrowing layer, and solid reactive state management in the Angular component. The iteration cadence on this PR has been exemplary — every bot and human finding was addressed promptly and with quality responses (the CodeRabbit pushback on the row-interface location was particularly sharp).

The account-switch race condition (detailCancel$, detailKey composite keying) and the error UX (retryTrigger + per-row detailErrorMap) are both well-designed and will serve as good patterns for the follow-up tabs.

**0 blocking / 2 minor / 2 nit** — minor findings are either dead-weight cleanup (`photoUrl`) or a tracked E2E follow-up. Neither is a merge blocker.

**Final decision: ✅ Approved with minor comments**

apps/lfx-one/src/server/services/org-lens-people.service.ts
ahmedomosanya: Addressed in 3a8f161b by dropping. PHOTO is now removed from the `ORG_PEOPLE_ALL` SELECT, `photoUrl` from `OrgAllEmployeeRow`, and the mapping in `org-lens-people.service.ts`. Adopting the photo in the avatar `` is a design call (initials avatars have been the established Org Lens pattern), so removing the dead weight is the safer in-scope move. Happy to re-introduce when/if design confirms the photo-with-initials-fallback variant.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees-detail.component.html
ahmedomosanya: Addressed in 3a8f161b. Dropped the per-row "Total Events" column and hoisted the aggregate into the section header as `Events (N total)` via a new `eventsTotal` computed signal in the detail child component. Kept the wire contract unchanged (`eventsCount` still denormalized on each `OrgAllEmployeeEvent` so consumers don't have to dig). Left the Training table as-is for now since you flagged it as a comparable precedent rather than asking to change it; happy to do the same pass on Certifications/Courses Taken if you'd like a follow-up.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
ahmedomosanya: Good flag, thanks. No code change needed — the LFX One e2e suite (apps/lfx-one-e2e) uses Playwright's `getByTestId()` / `data-testid` matchers exclusively, which handle colons natively. Worth keeping in mind if anyone ever drops in a raw `document.querySelector("[data-testid='…']")` for a snapshot helper or unit test — they'd need `CSS.escape()`.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
```diff
@@ -0,0 +1,295 @@
...
+                  }
+                </td>
+                <td class="px-4 py-3 align-middle">
+                  <span
+                    class="text-xs text-gray-300"
+                    pTooltip="Coming in a future release"
+                    tooltipPosition="top"
+                    data-testid="org-people-all-employees-access-placeholder"
+                    >&mdash;</span
+                  >
```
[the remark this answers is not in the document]
ahmedomosanya: Respectfully pushing back — false positive. The `</span\n  >` on lines 252-253 is Prettier's HTML closing-tag wrap (the `>` is the close of the `</span` tag, not a stray character). The whitespace between `</span` and `>` is valid per the HTML spec and Angular template parser; `prettier --check` and `yarn lint:check` both pass on this. No code change.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees-detail.component.html
```diff
@@ -0,0 +1,226 @@
...
+          <div class="overflow-hidden rounded-md border border-gray-200 bg-white">
+            <table class="w-full text-sm">
+              <thead class="bg-gray-50">
+                <tr class="text-left text-[11px] font-semibold uppercase tracking-wide text-gray-500">
+                  <th class="border-b border-gray-200 px-3 py-2">Most Recent Event</th>
+                  <th class="border-b border-gray-200 px-3 py-2">Foundation</th>
+                  <th class="border-b border-gray-200 px-3 py-2">Role</th>
+                  <th class="border-b border-gray-200 px-3 py-2">Total Events</th>
+                  <th class="border-b border-gray-200 px-3 py-2">Last Attended</th>
+                </tr>
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 3a8f161b. Header renamed to `Event` since the table renders every event row, not just the most recent.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees-detail.component.html
```diff
@@ -0,0 +1,226 @@
...
+          </header>
+          <div class="overflow-hidden rounded-md border border-gray-200 bg-white">
+            <table class="w-full text-sm">
+              <thead class="bg-gray-50">
+                <tr class="text-left text-[11px] font-semibold uppercase tracking-wide text-gray-500">
+                  <th class="border-b border-gray-200 px-3 py-2">Most Recent Course</th>
+                  <th class="border-b border-gray-200 px-3 py-2">Status</th>
+                  <th class="border-b border-gray-200 px-3 py-2">Certifications</th>
+                  <th class="border-b border-gray-200 px-3 py-2">Courses Taken</th>
+                </tr>
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 3a8f161b. Header renamed to `Course` since the table renders every training row, not just the most recent.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees-detail.component.html
```diff
@@ -0,0 +1,229 @@
...
+                          class="inline-flex items-center rounded-full border border-violet-200 bg-violet-50 px-2 py-0.5 text-[11px] font-semibold text-violet-700">
+                          Certified
+                        </span>
+                      } @else {
+                        <span class="inline-flex items-center rounded-full bg-gray-100 px-2 py-0.5 text-[11px] text-gray-600">Enrolled</span>
+                      }
+                    </td>
+                    <td class="px-3 py-2 font-semibold text-gray-700">{{ row.certificationsCount }}</td>
+                    <td class="px-3 py-2 text-gray-500">{{ row.coursesCount }}</td>
+                  </tr>
```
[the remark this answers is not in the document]
ahmedomosanya: Deferring — the per-row columns (`Certifications`, `Courses Taken`) are specified by the canonical design (`lfx-self-serve-design-docs/lfx-one-combined.html`, `_aeDetailHtml` lines 23575-23590). Tracked as a follow-up.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
```diff
@@ -0,0 +1,295 @@
...
+            <tr class="border-b border-gray-200 text-left text-xs uppercase tracking-wide text-gray-500">
+              <th class="w-7 px-2 py-3"></th>
+              @let aria = ariaSortMap();
+              <th scope="col" class="px-4 py-3 font-medium" [attr.aria-sort]="aria.name">
+                <button
+                  type="button"
+                  class="inline-flex w-full items-center gap-1 bg-transparent p-0 text-left text-xs font-medium uppercase tracking-wide text-gray-500 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+                  (click)="onSort('name')">
+                  Employee
+                  <i [class]="sortIcon('name')" aria-hidden="true"></i>
```
MRashad26: **[Templates] Method call `sortIcon()` in 5 `[class]` bindings — CLAUDE.md violation**

```html
<i [class]="sortIcon('name')" aria-hidden="true"></i>
<i [class]="sortIcon('seats')" aria-hidden="true"></i>
<!-- ... (5 total) -->
```

CLAUDE.md: _"No functions in HTML templates — only signals, computed values, or pipes."_

**Suggested fix:** add a `sortIconMap` computed signal (same pattern as the existing `ariaSortMap`):
```ts
// in component .ts
protected readonly sortIconMap: Signal<Record<OrgAllEmployeeSortColumn, string>> = computed(() => {
  const active = this.sortColumn();
  const dir = this.sortDirection();
  const icon = (col: OrgAllEmployeeSortColumn) =>
    active !== col ? 'fa-light fa-sort' : dir === 1 ? 'fa-light fa-sort-up' : 'fa-light fa-sort-down';
  return { name: icon('name'), seats: icon('seats'), commits: icon('commits'), events: icon('events'), courses: icon('courses') };
});
```
```html
@let icons = sortIconMap();
<i [class]="icons.name" aria-hidden="true"></i>
<!-- etc. -->
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
```diff
@@ -0,0 +1,295 @@
...
+            </tr>
+          </thead>
+          <tbody>
+            @for (row of visibleRows(); track row.personKey) {
+              <tr
+                class="cursor-pointer border-b border-gray-100 hover:bg-gray-50 focus:bg-gray-50 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+                role="button"
+                tabindex="0"
+                [attr.data-testid]="'org-people-all-employees-row-' + row.personKey"
+                [attr.aria-expanded]="isExpanded(row.personKey)"
```
MRashad26: **[Templates] Method calls in `@for` row bindings — CLAUDE.md violation (×6)**

```html
[attr.aria-expanded]="isExpanded(row.personKey)"
[class.rotate-90]="isExpanded(row.personKey)"
[class]="getAvatarColorClass(row.personKey)"
{{ getInitials(row.name) }}
[detail]="getDetail(row.personKey)"
[loading]="isDetailLoading(row.personKey)"
[error]="isDetailError(row.personKey)"
```

CLAUDE.md: _"No functions in HTML templates — only signals, computed values, or pipes."_

**Suggested fixes:**

1. **`isExpanded` / `getDetail` / `isDetailLoading` / `isDetailError`** — read the signal maps directly with `@let`:
```html
@let key = accountContext.selectedAccount().accountId + ':' + row.personKey;
@let expanded = !!expansion()[row.personKey];
@let detailData = detailMap()[key] ?? null;
@let detailIsLoading = !!detailLoading()[key];
@let detailHasError = !!detailErrorMap()[key];
```

2. **`getInitials(row.name)`** — add an `initials` pipe or pre-compute it in `initFilteredRows()` / `initSortedRows()` by attaching it to each row.

3. **`getAvatarColorClass(row.personKey)`** — pre-compute in the row mapping as `avatarColorClass: AllEmployeesComponent.getAvatarColorClass(row.personKey)` so the template reads `row.avatarColorClass`.

apps/lfx-one/src/server/services/org-lens-people.service.ts
```diff
@@ -0,0 +1,436 @@
...
+interface TrainingRow {
+  ACCOUNT_ID: string;
+  PERSON_KEY: string;
+  COURSE_OR_CERT_ID: string;
+  STATUS: string | null;
+  COURSE_ID: string | null;
+  COURSE_NAME: string | null;
+}
+
+const EMPTY_STATS: OrgAllEmployeeStats = {
```
MRashad26: **Module-level constant in `apps/lfx-one/` — CLAUDE.md violation**

```ts
const EMPTY_STATS: OrgAllEmployeeStats = {
  activeInOss: 0,
  inGovernance: 0,
  ...
};
```

**Suggested fix:** move to `packages/shared/src/constants/org-people.constants.ts` alongside `EMPTY_ORG_ALL_EMPLOYEES_RESPONSE` and import from `@lfx-one/shared/constants`:
```ts
// in packages/shared/src/constants/org-people.constants.ts
export const EMPTY_ORG_ALL_EMPLOYEE_STATS: OrgAllEmployeeStats = {
  activeInOss: 0, inGovernance: 0, codeContributors: 0, eventAttendees: 0, trainees: 0,
};
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
```diff
@@ -0,0 +1,395 @@
...
+  protected readonly detailErrorMap = signal<Record<string, boolean>>({});
+
+  // Retry tick — incremented to re-trigger the list fetch without changing accountId.
+  protected readonly retryTrigger = signal<number>(0);
+
+  // Cancels any in-flight detail subscription when the selected account changes; prevents stale writes and frees server work.
+  private readonly detailCancel$ = new Subject<void>();
+
+  // Spread to a mutable array so PrimeNG's mutable [options] input type accepts it without an unsafe cast.
+  protected readonly activityOptions: OrgAllEmployeeActivityOption[] = [...ORG_ALL_EMPLOYEE_ACTIVITY_OPTIONS];
```
MRashad26: **[Component Organization] Public fields placed after WritableSignals — ordering violation**

```ts
export class AllEmployeesComponent {
  // (1) private injections ✅
  private readonly accountContext = inject(AccountContextService);
  ...

  // (2) public fields ✅
  protected readonly initialLimit = ORG_ALL_EMPLOYEES_INITIAL_LIMIT;

  // (5) simple WritableSignals
  protected readonly searchTerm = signal<string>('');
  ...
  protected readonly retryTrigger = signal<number>(0);

  // ← (2) public fields placed HERE — wrong order
  protected readonly activityOptions: OrgAllEmployeeActivityOption[] = [...];
  protected readonly statSkeletonLabels: readonly string[] = [...];
  protected readonly tableSkeletonRows: readonly number[] = [...];
```

Per `.claude/rules/component-organization.md`, public fields (position 2) must come before simple `WritableSignal`s (position 5).

**Suggested fix:** move `activityOptions`, `statSkeletonLabels`, and `tableSkeletonRows` up to immediately after `initialLimit` (before the first `signal()` call).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
ahmedomosanya: Fixed in 995d6361 by adding a `sortIconMap` computed (mirrors the existing `ariaSortMap` pattern) and rewiring all 5 `[class]` bindings to `icons.<col>` via an `@let icons = sortIconMap();` next to the existing `@let aria = ariaSortMap();`. The `sortIcon()` method has been removed.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.html
ahmedomosanya: Fixed in 995d6361. Removed all 6 template-called methods (`isExpanded`, `getDetail`, `isDetailLoading`, `isDetailError`, `getInitials`, `getAvatarColorClass`). Introduced a file-private `AllEmployeeRowVm` view model that extends `OrgAllEmployeeRow` and bakes `initials` + `avatarColorClass` per-row in `initViewRows()` (once per response, flowed through filter/sort layers for free). Template now uses `@let` blocks for the composite (account, person) detail-cache key and signal-map lookups, driven by a new `currentAccountId` computed.

apps/lfx-one/src/server/services/org-lens-people.service.ts
ahmedomosanya: Fixed in 995d6361 by moving to `packages/shared/src/constants/org-people.constants.ts` as `EMPTY_ORG_ALL_EMPLOYEE_STATS`. Also rewired `EMPTY_ORG_ALL_EMPLOYEES_RESPONSE` to reuse the new constant for DRY.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/all-employees/all-employees.component.ts
ahmedomosanya: Fixed in 995d6361 by moving `activityOptions`, `statSkeletonLabels`, and `tableSkeletonRows` up to immediately after `initialLimit`, before the WritableSignal block.

MRashad26 (approved):

PR #794, opened 2026-05-26
feat(org-selector): integrate query-service + member-service with FGA-filtered list [LFXV2-1675]

luismoriguerra:
## Summary

Replaces the persona-seeded `lfx-org-selector` with a paginated, FGA-filtered list backed by the query-service `b2b_org` index, mirroring the proven `project-selector` / `NavigationService` pattern. On selection, fetches the canonical record from member-service to reconcile the indexed snapshot. Adds session-cached role grants so each row gets a writer/auditor pill with O(1) lookup, and gates sidebar trigger visibility on `flag ∧ (writers ∨ auditors ∨ personaSeeds)` to avoid the useless-empty-state pattern.

Resolves [LFXV2-1675](https://linuxfoundation.atlassian.net/browse/LFXV2-1675).

Spec / design docs: `specs/020-org-selector-integration/` (spec, plan, research, data-model, contracts, quickstart, tasks).

### What ships

**Server (BFF)**
- `GET /api/nav/org-items` — paginated, FGA-filtered org list with server-side search, deterministic-empty-on-upstream-failure (FR-005), and `selected_uid` injection for cookie-restored deep links (FR-013).
- `GET /api/orgs/me/role-grants` — session role dictionary via the `b2b_org_settings` "what can I see" pattern; writer-wins flattening server-side so the client gets two disjoint string arrays.
- `GET /api/orgs/uid/:uid` + `/sfid/:accountId` + polymorphic `/:id` — canonical-record fetch from member-service with UUID-format dispatch (mirrors `project.controller`). 5xx → 502, 404 stays 404 (no information leak). Backed by `OrgIdentityResolver` (process-wide LRU cache for sfid↔uid lookups, 10k entries × 24h TTL).

**Client**
- `OrgNavigationService` (client) — single-state mirror of `NavigationService` with generation race-guard, debounced search, infinite scroll via `@defer (on viewport)` sentinel 8 rows from end, uid dedup on scroll.
- `OrgRoleGrantsService` — eager-loaded session dictionary exposing `writerSet` / `auditorSet` signals for O(1) row decoration. Lifetime = providedIn:'root' instance; no client-side token-refresh subscription (FR-018a) since the bearer token is refreshed server-side per-request.
- `OrgSelectorComponent` — full rewrite: signal-bound to the navigation service, server-side search input, role grant rendered as an `Org Admin Editor` / `Org Admin Viewer` pill on the trigger and inline on rows. Preserves the impersonation panel modifier.
- `AccountContextService.refreshCanonicalRecord` — async reconciliation after selection with request-scope dedup; FR-020 silent-on-failure.
- `SidebarComponent.effectiveShowOrgSelector` — composed visibility gate per Q3 / D-005.

**Shared package**
- New `OrgItem`, `OrgItemsResponse`, `OrgItemsQuery`, `RoleGrantsResponse`, `OrgCanonicalRecord`, `OrgListState` interfaces.
- Extended `Account` with nullable `uid` + `parentUid`.
- New `ORG_SELECTOR_DEBOUNCE_MS`, `ORG_ITEMS_PER_PAGE`, `ORG_ROLE_GRANTS_HARD_CAP`.

**Local-dev fixture**
- `MOCK_ORG_ITEMS=true` opts into a 15-row Snowflake-sourced fixture (`fixtures/org-selector.mock.json`) that short-circuits all three routes. Real `LOGO_URL` values pulled from `ANALYTICS.PLATINUM_LFX_ONE.ORG_LENS_ACCOUNT_CONTEXT`; first 8 rows share Salesforce IDs with `ORG_LENS_DEMO_SEED_ACCOUNTS` so cookie restoration resolves to a real fixture row. Documented in `.env.example`.

### Preserved invariants

- Cookie schema (`lfx-selected-account`) is unchanged — still stores legacy `accountId` per Q1 so Snowflake-keyed routes (`/api/orgs/:accountId/lens/*`) continue working without change.
- Bearer token is forwarded verbatim by `MicroserviceProxyService`; no application-code auth checks added (Constitution Principle I).
- Zero new NATS subjects (Principle II); HTTP-only to query-service + member-service.
- No OpenFGA model changes (Principle III).
- Snowflake schema untouched (Principle VII).
- LaunchDarkly `org-lens-enabled` flag remains the rollout mechanism (Principle VIII).

### Automated
- [x] `yarn lint --workspace=apps/lfx-one` — clean
- [x] `yarn build --workspace=apps/lfx-one` — clean
- [x] `npx tsc -p apps/lfx-one/tsconfig.json --noEmit` — clean
- [ ] CI checks (Turborepo lint + build + e2e) pass on PR

### Manual smoke (per `specs/020-org-selector-integration/quickstart.md`)
- [ ] S1 — selector trigger renders for an authorized user (SC-002)
- [ ] S2 — first page loads in <300 ms (SC-003)
- [ ] S3 — server-side search via `?name=` debounced 300 ms (SC-004)
- [ ] S4 — infinite scroll appends without dupes
- [ ] S5 — selection + canonical record fetch reconciles indexed snapshot
- [ ] S6 — `lfx-selected-account` cookie persists `accountId` across reload
- [ ] S7 — deep-link `selected_uid` injection prepends off-page selection
- [ ] S8 — stale cookie cleanup silently re-selects first authorized org
- [ ] S9 — zero-grants user has no trigger (visibility gate)
- [ ] S10 — upstream failure renders deterministic error state
- [ ] S11 — writer/auditor role badges render correctly

### Playwright e2e
- [ ] `yarn e2e --grep org-selector` runs the S1 / S2 / S5 / S9 smoke set against the running dev server (`apps/lfx-one/e2e/org-selector.spec.ts`)

### Local-dev mock
- [ ] Setting `MOCK_ORG_ITEMS=true` in `apps/lfx-one/.env` populates the selector with 15 real LF member orgs (Red Hat, IBM, Toyota, Microsoft, Google, Meta, Intel, AWS, Samsung, Huawei, …) with working search, scroll-pagination, and role badges — no upstream connectivity required.

## Rollout

Behind the existing `org-lens-enabled` LaunchDarkly flag, off for everyone except internal testers. Stop after merging the MVP (US1 + US2 + US3 + US4 all in this PR) to validate via quickstart S1–S11 before unflagging for all users.

[LFXV2-1675]: https://linuxfoundation.atlassian.net/browse/LFXV2-1675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/e2e/org-selector.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+  await trigger.click();
+  await expect(page.getByTestId('org-search-input')).toBeVisible({ timeout: 5_000 });
+}
+
+test.describe('Org Selector — authorized user smoke set (S1/S2/S5)', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(APP_HOME, { waitUntil: 'domcontentloaded' });
+    // Hard skip when the auth-bootstrap failed — surface a clear log so CI triage doesn't
+    // chase a spec-020 regression that's actually a credentials issue.
+    if (page.url().includes('auth0.com')) {
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 889ed226 — replaced the substring check with hostname-exact matching via `new URL(page.url()).hostname === 'auth0.com' || hostname.endsWith('.auth0.com')`, wrapped in try/catch for malformed URLs. The skip gate can no longer be fooled by URLs like `https://evil.com/?ref=auth0.com` or `https://auth0.com.evil.com/`.

packages/shared/src/constants/org-selector.constants.ts
```diff
@@ -0,0 +1,17 @@
...
+ * NAV search debounce so the project-selector and org-selector remain in lockstep.
+ */
+export const ORG_SELECTOR_DEBOUNCE_MS = NAV_SEARCH_DEBOUNCE_MS;
+
+/**
+ * Soft absolute cap on the role-grants response per spec Q4 / SC-005b. The BFF
+ * passes `per_page={cap}` to bound the upstream response — orgs beyond this cap
+ * silently fall through to the no-badge path (FR-018).
+ */
+export const ORG_ROLE_GRANTS_HARD_CAP = 500;
```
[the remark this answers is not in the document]
luismoriguerra: Good catch \u2014 the constant was trimmed pre-commit because nothing in the shipped code references it (the BFF relies on the upstream's default page size for prod, and the mock branch uses a local `MOCK_PAGE_SIZE = 8` constant in `org-navigation.service.ts`). Updated the PR description / Jira comment to drop the stale `ORG_ITEMS_PER_PAGE` reference; the constants file only exports `ORG_SELECTOR_DEBOUNCE_MS` and `ORG_ROLE_GRANTS_HARD_CAP` (both used). If/when we want an explicit BFF-side cap on `per_page` for org-items, I'll re-add the constant in a follow-up so it ships alongside the code that imports it.

apps/lfx-one/src/server/utils/mock-org-items.util.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+/**
+ * Local-dev opt-in for the spec 020 org-selector surface. When the env var is
+ * truthy AND we are NOT running in production, the BFF serves
+ * `apps/lfx-one/src/server/services/fixtures/org-selector.mock.json` for all
+ * three routes (`/api/nav/org-items`, `/api/orgs/me/role-grants`, and the
+ * polymorphic canonical-record endpoint) and bypasses LFX_V2_SERVICE entirely.
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 1a39289c \u2014 tightened the JSDoc to say the env var is enabled by the literal strings `'true'` or `'1'` (other truthy-looking values are intentionally rejected so the dev fail-closed gate is unambiguous). Kept the strict check rather than broadening to a truthy parse because it matches what's documented in `.env.example`.

apps/lfx-one/src/server/services/org-navigation.service.ts
```diff
@@ -0,0 +1,233 @@
...
+    };
+  }
+
+  /**
+   * Local-dev mock branch. Filters, paginates, and applies the same
+   * `selected_uid` injection rule as the real BFF so the popover behaves
+   * identically end-to-end when MOCK_ORG_ITEMS=true. NEVER reached in
+   * environments where the flag is unset (and is tree-shaken from the prod
+   * code path by the early-return in getOrgItems).
+   */
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 1a39289c \u2014 reworded the comment to explicitly call out that Node server code isn't subject to build-time dead-code elimination. The new wording reads: `Unreachable at runtime in any environment where the flag is unset \u2014 the early-return in \`getOrgItems\` short-circuits before this method is invoked. (The fixture and this method are still present in the server bundle; 'tree-shaken' would only apply to a build-time elimination step, which Node runs don't perform.)`

apps/lfx-one/src/server/controllers/org-identity.controller.ts
```diff
@@ -0,0 +1,281 @@
...
+        res.json(mockResponse);
+        return;
+      }
+
+      let uid: string | null = identifier;
+      let cacheHit = true;
+      if (identifierKind === 'sfid') {
+        cacheHit = false;
+        uid = await this.orgIdentityResolver.getUidBySfid(identifier, req);
+        if (!uid) {
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 1a39289c \u2014 refactored `OrgIdentityResolver.getUidBySfid` / `getSfidByUid` to return `{ uid | sfid, cacheHit }` so the controller can log the actual LRU state. The new log key is `identifier_cache_hit` (boolean | null) \u2014 `null` for `uid`-routed requests where no resolver lookup runs, so the cache-hit-ratio metric stays meaningful instead of being a tautology. Renaming the key also avoids confusion with downstream member-service / member-service-response caching that the controller doesn't perform.

apps/lfx-one/e2e/org-selector.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    // assert the selection visually instead.
+    expect(uid).toBeTruthy();
+    expect(selectedAccountCookie === undefined || typeof selectedAccountCookie.value === 'string').toBe(true);
+  });
+});
+
+// S9 — zero-grants visibility gate. We exercise the unauthenticated path as a proxy for "no grants":
+// the trigger MUST NOT render on the login page. A real zero-grants signed-in user would require a
+// secondary Auth0 fixture which is out of scope per the spec test-strategy decision (D-008).
+test.describe('Org Selector — zero-grants visibility gate (S9)', () => {
```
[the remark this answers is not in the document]
luismoriguerra: Fully agreed \u2014 the previous S9 only exercised the structural smoke. Replaced in 1a39289c with a proper authenticated zero-grants test: it intercepts `/api/orgs/me/role-grants` (empty writers + auditors) and `/api/user/personas` (empty `organizations`) via `page.route`, reloads so the route handlers catch the fresh client-side fetches, then asserts the `org-selector-slot` reports `data-visible=\"false\"` and the trigger is not visible. The `data-visible` attribute on the slot was added in the original spec-020 patch specifically for this kind of assertion, so the test can read the gate's computed state directly without timing on toasts/redirects. The pre-existing limitation noted in the test header (we can't stub LD per-test, so the flag is still assumed enabled by the global fixture) remains, but the gate's grants/seeds branch is now genuinely exercised end-to-end.

apps/lfx-one/src/app/shared/components/org-selector/org-selector.component.ts
```diff
@@ -1,55 +1,122 @@
...
+    });
+
+    // Browser-only: bootstrap when the sidebar visibility gate turns on. An `effect()` registered
+    // inside afterNextRender did not reliably re-run when `enabled` flipped false→true, so the list
+    // never fetched mock/live data even though the trigger was visible.
+    afterNextRender(() => {
+      toObservable(this.enabled)
+        .pipe(distinctUntilChanged(), filter((enabled) => enabled), takeUntilDestroyed())
+        .subscribe(() => this.bootstrapOrgList());
+    });
```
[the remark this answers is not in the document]
luismoriguerra: Flagging this as a false positive after re-reading the actual file. The diff CodeRabbit proposes (`toObservable(this.enabled).pipe(..., takeUntilDestroyed())` inside the `afterNextRender` callback) doesn't exist in this component — there's no `enabled` signal, no `bootstrapOrgList()` method, and no `takeUntilDestroyed` call inside the `afterNextRender` body. The actual `afterNextRender` is purely synchronous:\n\n```ts\nafterNextRender(() => {\n  const restoredUid = this.accountContextService.selectedAccount().uid ?? null;\n  this.orgNavigationService.resetAndReload(restoredUid);\n});\n```\n\nThe only `takeUntilDestroyed()` call in the component is in the constructor body (line ~91), which is still inside the injection context, so no explicit `DestroyRef` is required. No change needed — leaving thread open for reviewer ack.

apps/lfx-one/src/server/services/org-role-grants.service.ts
```diff
@@ -0,0 +1,96 @@
...
+        username,
+        loaded_at: loadedAt,
+      };
+    }
+
+    try {
+      const response = await this.microserviceProxy.proxyRequest<QueryServiceResponse<B2bOrgSettingsDoc>>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', {
+        type: 'b2b_org_settings',
+        filters_or: [`writers.username:${username}`, `auditors.username:${username}`],
+        per_page: ORG_ROLE_GRANTS_HARD_CAP,
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in the next push — added `encodeURIComponent(username)` before building the `filters_or` values, matching the encoding pattern already used by `OrgIdentityResolver` for uid/sfid filters. The encoding happens once into a local `encodedUsername` so the two filter strings stay readable, and there's a comment calling out the JWT-claim source so future readers don't peel it off.

apps/lfx-one/src/app/shared/components/org-selector/org-selector.component.ts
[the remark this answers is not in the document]
luismoriguerra: Correction \u2014 I was wrong, you were right. Re-read the file at HEAD and confirmed the `toObservable(this.enabled).pipe(..., takeUntilDestroyed())` pattern inside `afterNextRender` IS there (lines 91\u201397 of the current file). I had cached a stale mental model of an earlier simpler implementation and didn't re-check before declaring false positive \u2014 my bad.\n\nFixed in the next push by injecting `DestroyRef` eagerly in the constructor (still in the injection context) and passing it explicitly as `takeUntilDestroyed(this.destroyRef)`. Without that, the callback would have thrown NG0203 on the first browser mount once the sidebar visibility gate flipped on. Thanks for pushing back rather than accepting the brush-off.

apps/lfx-one/src/server/services/org-role-grants.service.ts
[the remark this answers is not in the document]
luismoriguerra: Reopening this thread \u2014 Copilot caught (on the same file) that my `encodeURIComponent(username)` fix actually introduced a double-encoding bug because `MicroserviceProxyService` \u2192 `ApiClientService.getFullUrl` already builds the query via `URLSearchParams.append`, which percent-encodes the value once. Pre-encoding turned `@` into `%2540` upstream and would silently return zero role grants for any email-style username, hiding the selector for valid users. See [thread on org-role-grants.service.ts:58](https://github.com/linuxfoundation/lfx-self-serve/pull/794#discussion_r3307460059) for the full analysis.\n\nRetraction + replacement landed in the next commit: dropped `encodeURIComponent` everywhere (matches `NavigationService` convention) and replaced it with a fail-closed allowlist check that retains the injection-defense intent you originally raised. The username allowlist is `^[A-Za-z0-9._+\-@]+$` (256-char cap) which covers every OIDC username shape including tagged email aliases, and crucially excludes the filter-grammar separators `:` and `,`. Same pattern applied to `OrgIdentityResolver` for uid/sfid filters.

apps/lfx-one/src/server/services/org-role-grants.service.ts
```diff
@@ -0,0 +1,101 @@
...
+    try {
+      // Encode the username defensively — a JWT claim with reserved characters
+      // (`:` / `,` / spaces / etc.) would otherwise break the filter grammar
+      // or open up filter-injection paths. Matches the encoding pattern used by
+      // OrgIdentityResolver for uid/sfid filters.
+      const encodedUsername = encodeURIComponent(username);
+      const response = await this.microserviceProxy.proxyRequest<QueryServiceResponse<B2bOrgSettingsDoc>>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', {
+        type: 'b2b_org_settings',
+        filters_or: [`writers.username:${encodedUsername}`, `auditors.username:${encodedUsername}`],
+        per_page: ORG_ROLE_GRANTS_HARD_CAP,
```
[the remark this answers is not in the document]
luismoriguerra: Correct catch \u2014 verified the double-encoding path: `ApiClientService.getFullUrl` uses `URLSearchParams.append` (api-client.service.ts:320\u2013335) which percent-encodes once, so my `encodeURIComponent` ahead of that would have shipped `%2540` upstream and broken role grants for email-style usernames. The earlier CodeRabbit thread that prompted my encoding flagged the right injection concern but proposed the wrong remediation.\n\nFixed in 43e63ad7 by:\n1. Dropping `encodeURIComponent` everywhere (matches `NavigationService` convention).\n2. Replacing it with a fail-closed allowlist (`^[A-Za-z0-9._+\\-@]+$`, 256-char cap) that explicitly excludes the filter-grammar separators `:` and `,`. Hostile claims silently degrade to empty grants (gate falls back to persona seeds) with a warning log instead of altering the query shape.\n3. Same pattern applied to `OrgIdentityResolver` for uid/sfid filters where I had latently shipped the same double-encoding antipattern (worked today only because UUIDs and SF-IDs are `[A-Za-z0-9_-]` only \u2014 nothing `URLSearchParams` encodes \u2014 but the convention was still wrong).

apps/lfx-one/src/app/shared/components/org-selector/org-selector.component.ts
```diff
@@ -1,55 +1,126 @@
...
+
+    // Browser-only: bootstrap when the sidebar visibility gate turns on. An `effect()` registered
+    // inside afterNextRender did not reliably re-run when `enabled` flipped false→true, so the list
+    // never fetched mock/live data even though the trigger was visible.
+    afterNextRender(() => {
+      toObservable(this.enabled)
+        .pipe(
+          distinctUntilChanged(),
+          filter((enabled) => enabled),
+          takeUntilDestroyed()
```
MRashad26: **[Potential functional bug] `toObservable()` + `takeUntilDestroyed()` inside `afterNextRender` callback — no injection context**

```ts
afterNextRender(() => {
  toObservable(this.enabled)          // ❌ calls inject() — injection context gone
    .pipe(
      ...
      takeUntilDestroyed()             // ❌ same issue
    ).subscribe(...);
});
```

Both `toObservable()` and `takeUntilDestroyed()` call `inject()` internally. `afterNextRender` callbacks run *asynchronously* after the first render — the constructor's injection context is no longer active at that point. This will throw `NG0203: inject() must be called from an injection context` in the browser.

**Suggested fix:** capture `DestroyRef` and `Injector` in the constructor and pass them explicitly:
```ts
private readonly destroyRef = inject(DestroyRef);
private readonly injector = inject(Injector);

// in constructor:
afterNextRender(() => {
  toObservable(this.enabled, { injector: this.injector })
    .pipe(
      distinctUntilChanged(),
      filter((enabled) => enabled),
    ).subscribe(() => this.bootstrapOrgList());
});
```

apps/lfx-one/src/app/shared/components/org-selector/org-selector.component.ts
```diff
@@ -1,55 +1,126 @@
...
   styleUrl: './org-selector.component.scss',
 })
 export class OrgSelectorComponent {
   private readonly accountContextService = inject(AccountContextService);
+  private readonly orgNavigationService = inject(OrgNavigationService);
+  private readonly orgRoleGrantsService = inject(OrgRoleGrantsService);
   private readonly userService = inject(UserService);
 
   public readonly isPanelOpen = model<boolean>(false);
+  /**
```
MRashad26: **[Multi-line JSDoc] `enabled` input and `personaToLabel` method — CLAUDE.md violation (×2)**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

```ts
// line 32-35: 4-line JSDoc on `enabled` input
/**
 * When false the trigger is hidden (sidebar visibility gate) — skip list bootstrap so
 * zero-grants users don't hit /api/nav/org-items or the empty-list redirect path.
 */

// line 160-164: 5-line JSDoc on private `personaToLabel()`
/**
 * Friendly grant labels. Writer is editor-level (read + mutate on the b2b_org);
 * ...
 */
```

Both collapse to a single-line `/** ... */`:
```ts
/** Hidden sidebar gate — skip bootstrap when false to avoid zero-grants network hits. */
public readonly enabled = input<boolean>(true);

/** Maps OrgRolePersona to its product display label (writer → "Org Admin Editor"). */
private personaToLabel(...)
```

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -49,17 +52,34 @@ export class SidebarComponent {
...
+  private readonly orgRoleGrantsService = inject(OrgRoleGrantsService);
+  private readonly accountContextService = inject(AccountContextService);
+  private readonly featureFlagService = inject(FeatureFlagService);
 
   public readonly items = input.required<SidebarMenuItem[]>();
   public readonly footerItems = input<SidebarMenuItem[]>([]);
   public readonly collapsed = input<boolean>(false);
   public readonly styleClass = input<string>('');
   public readonly showProjectSelector = input<boolean>(false);
+  /**
```
MRashad26: **[Multi-line JSDoc] `showOrgSelector` input and `effectiveShowOrgSelector` computed — CLAUDE.md violation (×2)**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

Lines 64–68 (3-line JSDoc on `showOrgSelector`) and lines 74–78 (5-line JSDoc on `effectiveShowOrgSelector`) both exceed the one-line limit. Collapse each to a single `/** … */` line.

apps/lfx-one/src/app/shared/services/account-context.service.ts
```diff
@@ -26,8 +28,16 @@ export class AccountContextService {
   private readonly cookieRegistry = inject(CookieRegistryService);
   private readonly analyticsService = inject(AnalyticsService);
   private readonly featureFlagService = inject(FeatureFlagService);
+  private readonly http = inject(HttpClient);
   private readonly storageKey = ACCOUNT_COOKIE_KEY;
 
+  /**
```
MRashad26: **[Multi-line JSDoc] `canonicalFetchInFlight` field and `refreshCanonicalRecord` method — CLAUDE.md violation (×2)**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

Lines 34–38 (5-line JSDoc on `canonicalFetchInFlight`) and lines 115–125 (11-line JSDoc on `refreshCanonicalRecord`) both exceed the one-line limit. Collapse to single-line `/** … */` comments.

apps/lfx-one/src/app/shared/services/org-navigation.service.ts
```diff
@@ -0,0 +1,317 @@
...
+import { Router } from '@angular/router';
+import { LENS_DEFAULT_ROUTES, ORG_SELECTOR_DEBOUNCE_MS } from '@lfx-one/shared/constants';
+import { Account, OrgCanonicalRecord, OrgItem, OrgItemsResponse, OrgListPage, OrgListState, TaggedOrgListPage } from '@lfx-one/shared/interfaces';
+import { MessageService } from 'primeng/api';
+import { catchError, debounceTime, distinctUntilChanged, EMPTY, filter, map, merge, Observable, of, scan, skip, Subject, switchMap, tap } from 'rxjs';
+
+import { AccountContextService } from './account-context.service';
+import { LensService } from './lens.service';
+
+/**
```
MRashad26: **[Multi-line JSDoc] Class JSDoc and method JSDoc blocks — CLAUDE.md violation (×3+)**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

Line 16 (class JSDoc, 11 lines), line 66 (method JSDoc), and line 253 (method JSDoc) all exceed the one-line limit. Collapse each to a single `/** … */` line or remove if the name is self-explanatory.

apps/lfx-one/src/app/shared/services/org-role-grants.service.ts
```diff
@@ -0,0 +1,95 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { HttpClient } from '@angular/common/http';
+import { afterNextRender, inject, Injectable, Signal, signal, WritableSignal } from '@angular/core';
+import { RoleGrantsResponse } from '@lfx-one/shared/interfaces';
+import { catchError, map, Observable, of, tap } from 'rxjs';
+
+/**
```
MRashad26: **[Multi-line JSDoc] `OrgRolePersona` type, class, and `refresh()` — CLAUDE.md violation (×3)**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

Lines 9–13 (5-line JSDoc on `OrgRolePersona` type), lines 16–31 (16-line JSDoc on class), and line 60 (method JSDoc) all exceed the one-line limit. Collapse each to a single `/** … */` line.

apps/lfx-one/src/server/controllers/navigation.controller.ts
```diff
@@ -51,4 +54,45 @@ export class NavigationController {
       next(error);
     }
   }
+
+  /**
```
MRashad26: **[Multi-line JSDoc] `getOrgItems` method — CLAUDE.md violation**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

Lines 58–62 (5-line JSDoc on `getOrgItems`) exceed the one-line limit. Collapse to a single `/** … */` line, e.g.:
```ts
/** GET /api/nav/org-items — paginated FGA-filtered org list for the org selector. */
```

apps/lfx-one/src/server/controllers/org-identity.controller.ts
```diff
@@ -0,0 +1,288 @@
...
+import { ServiceValidationError } from '../errors/service-validation.error';
+import orgSelectorMock from '../services/fixtures/org-selector.mock.json';
+import { logger } from '../services/logger.service';
+import { MicroserviceProxyService } from '../services/microservice-proxy.service';
+import { OrgIdentityResolver } from '../services/org-identity-resolver.service';
+import { OrgRoleGrantsService } from '../services/org-role-grants.service';
+import { getEffectiveUsername } from '../utils/auth-helper';
+import { isMockOrgItemsEnabled } from '../utils/mock-org-items.util';
+
+interface MockCanonicalExtras {
```
MRashad26: **[Module-level consts/interfaces in `apps/lfx-one/`] — CLAUDE.md violation (×3)**

```ts
interface MockCanonicalExtras { ... }   // line 17
const UUID_PATTERN = /.../              // line 27
interface MemberServiceB2bOrgResponse { ... }  // line 29
```

**Suggested fixes:**
- `UUID_PATTERN` → `packages/shared/src/constants/regex.constants.ts` (a UUID regex likely exists or belongs there).
- `MemberServiceB2bOrgResponse` crosses file boundaries if used in more than one place → `packages/shared/src/interfaces/org-selector.interface.ts`.
- `MockCanonicalExtras` is test/mock-only and private to this file — if it truly stays local, scope it inside the method that uses it rather than at module level. If it's needed in the fixture JSON typing, extract to `packages/shared/src/interfaces/`.

apps/lfx-one/src/server/services/org-identity-resolver.service.ts
```diff
@@ -0,0 +1,167 @@
...
+// SPDX-License-Identifier: MIT
+
+import { QueryServiceResponse } from '@lfx-one/shared/interfaces';
+import { Request } from 'express';
+
+import { logger } from './logger.service';
+import { MicroserviceProxyService } from './microservice-proxy.service';
+
+/** Maximum number of entries in each direction of the bidirectional uid↔sfid cache. */
+const CACHE_MAX_ENTRIES = 10_000;
```
MRashad26: **[Module-level consts/interfaces in `apps/lfx-one/`] — CLAUDE.md violation (×4)**

```ts
const CACHE_MAX_ENTRIES = 10_000;   // line 11
const CACHE_TTL_MS = 24 * 60 * 60 * 1000;  // line 14
interface CacheEntry { ... }         // line 16
interface B2bOrgIndexedDoc { ... }   // line 64
```

**Suggested fixes:**
- `CACHE_MAX_ENTRIES` / `CACHE_TTL_MS` → `packages/shared/src/constants/org-selector.constants.ts` (alongside `ORG_ROLE_GRANTS_HARD_CAP` already there).
- `CacheEntry` → `packages/shared/src/interfaces/org-selector.interface.ts` (private cache detail — can also stay inside the `LruWithTtl` class body if it's truly implementation-private).
- `B2bOrgIndexedDoc` → `packages/shared/src/interfaces/org-selector.interface.ts` (note: also declared independently in `org-navigation.service.ts` and `org-role-grants.service.ts` — single source in shared prevents drift).

apps/lfx-one/src/server/services/org-navigation.service.ts
```diff
@@ -0,0 +1,236 @@
...
+import { GetOrgItemsParams, OrgItem, OrgItemsQuery, OrgItemsResponse, QueryServiceResponse } from '@lfx-one/shared/interfaces';
+import { Request } from 'express';
+
+import { isMockOrgItemsEnabled } from '../utils/mock-org-items.util';
+import orgSelectorMock from './fixtures/org-selector.mock.json';
+import { logger } from './logger.service';
+import { MicroserviceProxyService } from './microservice-proxy.service';
+
+/** Mock page size — bigger than the prod ~50 default so the dev e2e can scroll-paginate. */
+const MOCK_PAGE_SIZE = 8;
```
MRashad26: **[Module-level consts/interfaces in `apps/lfx-one/`] `MOCK_PAGE_SIZE` + `B2bOrgIndexedDoc` — CLAUDE.md violation (×2)**

```ts
const MOCK_PAGE_SIZE = 8;            // line 13
interface B2bOrgIndexedDoc { ... }   // line 19
```

CLAUDE.md: _"All shared constants and interfaces live in `@lfx-one/shared`."_

`MOCK_PAGE_SIZE` → `packages/shared/src/constants/org-selector.constants.ts`. `B2bOrgIndexedDoc` is also declared in `org-identity-resolver.service.ts` and `org-role-grants.service.ts` — consolidate into a single export from `packages/shared/src/interfaces/org-selector.interface.ts` to eliminate three separate copies that can drift.

apps/lfx-one/src/server/services/org-role-grants.service.ts
```diff
@@ -0,0 +1,101 @@
...
+import orgSelectorMock from './fixtures/org-selector.mock.json';
+import { logger } from './logger.service';
+import { MicroserviceProxyService } from './microservice-proxy.service';
+
+/**
+ * Shape of the `b2b_org_settings.data` payload returned by the query-service for
+ * the "what can I see" pattern. Each settings doc lists the writer/auditor
+ * usernames the caller may potentially appear in.
+ */
+interface B2bOrgSettingsDoc {
```
MRashad26: **[Local interface in `apps/lfx-one/`] `B2bOrgSettingsDoc` — CLAUDE.md violation**

```ts
interface B2bOrgSettingsDoc {   // line 18 — local, inside apps/lfx-one/
  writers?: { username?: string | null }[];
  auditors?: { username?: string | null }[];
}
```

CLAUDE.md: _"no module-level consts or local `interface Foo {}` inside `apps/lfx-one/`."_

Move to `packages/shared/src/interfaces/org-selector.interface.ts`.

packages/shared/src/interfaces/org-selector.interface.ts
```diff
@@ -0,0 +1,173 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import type { Signal, WritableSignal } from '@angular/core';
+import type { Subject } from 'rxjs';
+
+/**
```
MRashad26: **[Systemic multi-line JSDoc] Entire new file violates the one-line comment rule — CLAUDE.md violation**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

Every interface and several properties in this new file carry multi-line JSDoc blocks. For example:
```ts
/**
 * One row in the org selector dropdown. Carries BOTH identifiers per spec Q1 so the
 * client can read either the canonical UUID (member-service) or the legacy Salesforce
 * `accountId` (Snowflake / existing /api/orgs/:accountId/lens/* routes) without a
 * second resolver round-trip.
 */
export interface OrgItem { ... }    // 6-line block

// Similar multi-line blocks on: DisplayOrgItem, OrgItemsResponse, OrgItemsQuery,
// GetOrgItemsParams, RoleGrantsResponse, OrgCanonicalRecord, OrgListPage,
// TaggedOrgListPage, OrgListState — plus individual properties inside each.
```

Each should be a single `/** … */` line. The spec references and contract links belong in the PR description or a `contracts/` doc, not in source comments. Strip all multi-line blocks down to one descriptive line per symbol.

packages/shared/src/constants/org-selector.constants.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { NAV_SEARCH_DEBOUNCE_MS } from './lens.constants';
+
+/**
```
MRashad26: **[Multi-line JSDoc] `ORG_SELECTOR_DEBOUNCE_MS` and `ORG_ROLE_GRANTS_HARD_CAP` — CLAUDE.md violation (×2)**

CLAUDE.md: _"never write multi-paragraph docstrings or multi-line comment blocks — one short line max"_

```ts
// lines 6-9: 4-line block on ORG_SELECTOR_DEBOUNCE_MS
/**
 * Debounce window for the org-selector typeahead. Re-exported from the shared
 * NAV search debounce so the project-selector and org-selector remain in lockstep.
 */

// lines 12-16: 5-line block on ORG_ROLE_GRANTS_HARD_CAP
/**
 * Soft absolute cap on the role-grants response per spec Q4 / SC-005b...
 */
```

Collapse each to one line:
```ts
/** Debounce for org-selector typeahead — kept in lockstep with the project-selector. */
export const ORG_SELECTOR_DEBOUNCE_MS = NAV_SEARCH_DEBOUNCE_MS;

/** Hard cap on role-grants per_page (spec SC-005b); orgs beyond this fall to no-badge. */
export const ORG_ROLE_GRANTS_HARD_CAP = 500;
```

apps/lfx-one/src/app/shared/components/org-selector/org-selector.component.ts
luismoriguerra: Excellent catch \u2014 you're right, my T7 fix was incomplete. `toObservable()` ALSO calls inject() internally to wire up the signal, so the callback would still throw NG0203 on first browser mount (server-side hydration skips afterNextRender entirely, which is why local SSR-only smoke tests didn't surface it).\n\nFixed in 21239357 \u2014 injected `Injector` in the constructor alongside `DestroyRef`, and pass it explicitly via `toObservable(this.enabled, { injector: this.injector })`. The bare `takeUntilDestroyed(this.destroyRef)` already had its ref from my earlier fix; now both interop helpers have what they need.

apps/lfx-one/src/app/shared/components/org-selector/org-selector.component.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

apps/lfx-one/src/app/shared/services/account-context.service.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

apps/lfx-one/src/app/shared/services/org-navigation.service.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

apps/lfx-one/src/app/shared/services/org-role-grants.service.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

apps/lfx-one/src/server/controllers/navigation.controller.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

packages/shared/src/interfaces/org-selector.interface.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

packages/shared/src/constants/org-selector.constants.ts
luismoriguerra: Fixed in 21239357 \u2014 collapsed every multi-line JSDoc block on this file (and the rest of the new spec-020 files) down to a single-line `/** \u2026 */`. The spec references that used to live in the source comments now live only in `specs/020-org-selector-integration/` and the `contracts/` docs, which is the right home. Net effect across the PR: -355 lines.

apps/lfx-one/src/server/controllers/org-identity.controller.ts
luismoriguerra: Fixed in 21239357 \u2014 moved everything to shared:\n- `UUID_PATTERN` \u2192 reused the existing `UUID_REGEX` already in `packages/shared/src/constants/regex.constants.ts` (you correctly guessed it lived there).\n- `MemberServiceB2bOrgResponse` \u2192 `packages/shared/src/interfaces/org-selector.interface.ts`.\n- `MockCanonicalExtras` \u2192 scoped inline inside `getMockCanonicalRecord` as a `Partial<Pick<OrgCanonicalRecord, ...>>` since it's a fixture-only shape derived from the canonical record; no cross-file consumer.

apps/lfx-one/src/server/services/org-identity-resolver.service.ts
luismoriguerra: Fixed in 21239357 \u2014 moved to shared:\n- `CACHE_MAX_ENTRIES` \u2192 `ORG_IDENTITY_CACHE_MAX_ENTRIES` in `packages/shared/src/constants/org-selector.constants.ts`.\n- `CACHE_TTL_MS` \u2192 `ORG_IDENTITY_CACHE_TTL_MS` (same file).\n- `B2bOrgIndexedDoc` \u2192 `packages/shared/src/interfaces/org-selector.interface.ts` (single source replaces the three duplicates flagged in the other threads).\n- `CacheEntry` stays implementation-private to `LruWithTtl` \u2014 the class is the only consumer, no cross-file usage, so per your own guidance "can also stay inside the LruWithTtl class body if it's truly implementation-private" it lives there inline (just a Map value type).

apps/lfx-one/src/server/services/org-navigation.service.ts
luismoriguerra: Fixed in 21239357 \u2014 moved to shared:\n- `MOCK_PAGE_SIZE` \u2192 `ORG_SELECTOR_MOCK_PAGE_SIZE` in `packages/shared/src/constants/org-selector.constants.ts`.\n- `B2bOrgIndexedDoc` \u2192 `packages/shared/src/interfaces/org-selector.interface.ts` (single source; was triplicated across this file, `org-identity-resolver.service.ts`, and `org-role-grants.service.ts` \u2014 all three now import from shared).

apps/lfx-one/src/server/services/org-role-grants.service.ts
luismoriguerra: Fixed in 21239357 \u2014 moved `B2bOrgSettingsDoc` to `packages/shared/src/interfaces/org-selector.interface.ts`. Also extracted `isFilterSafeUsername` and `isFilterSafeIdentifier` to a new `packages/shared/src/utils/org-selector.utils.ts` since both predicates were duplicated across this service and the OrgIdentityResolver.

apps/lfx-one/src/server/services/org-identity-resolver.service.ts
```diff
@@ -0,0 +1,167 @@
...
+      const sfid = response?.resources?.[0]?.data?.sfid ?? null;
+      if (sfid) {
+        this.uidToSfid.set(uid, sfid);
+        // Round-trip the inverse mapping too — we already paid for the upstream call.
+        this.sfidToUid.set(sfid, uid);
+      }
+      return { sfid, cacheHit: false };
+    } catch (error) {
+      logger.warning(req, 'org_identity_resolver_get_sfid', 'Failed to resolve uid → sfid', { err: error, uid });
+      return { sfid: null, cacheHit: false };
```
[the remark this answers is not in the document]
luismoriguerra: Correct catch \u2014 you're right that collapsing 5xx into `null` would have silently misreported a query-service outage on the sfid path as 404 "Organization not found" instead of preserving the FR-020 502 path for the controller. Fixed in 21239357 by removing both catch blocks in `getSfidByUid` / `getUidBySfid` entirely. `MicroserviceProxyService.proxyRequest` already throws `MicroserviceError` with the upstream status code, so letting it bubble keeps the controller's existing catch authoritative: 5xx\u2192502, 404\u2192404, true filter-safe-allowlist rejections still return `{ ..., cacheHit: false }` for the controller to render as 404 (no upstream call was attempted).

apps/lfx-one/src/server/services/org-navigation.service.ts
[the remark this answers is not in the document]
luismoriguerra: The misleading "bigger than prod ~50" comment is gone \u2014 the constant moved to `packages/shared/src/constants/org-selector.constants.ts` as `ORG_SELECTOR_MOCK_PAGE_SIZE` in 21239357 (the shared-package refactor) with a corrected single-line comment: "Page size used by the MOCK_ORG_ITEMS dev branch so the e2e can exercise scroll-pagination." Accurate phrasing now reflects the actual intent (small page = forces scroll-pagination in dev/e2e), no prod-comparison anymore.

apps/lfx-one/src/server/services/org-navigation.service.ts
```diff
@@ -0,0 +1,236 @@
...
+   * first natural page. Returns null on miss (caller can't see the requested uid)
+   * — the response is returned without the prepended row and the client silently
+   * falls back to the first natural row (FR-014).
+   */
+  private async fetchSelectedItem(req: Request, uid: string): Promise<OrgItem | null> {
+    try {
+      const response = await this.microserviceProxy.proxyRequest<QueryServiceResponse<B2bOrgIndexedDoc>>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', {
+        type: 'b2b_org',
+        filters: [`uid:${encodeURIComponent(uid)}`],
+      });
```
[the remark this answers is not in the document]
luismoriguerra: Good catch \u2014 this was the same antipattern as the two I'd already fixed but on a third call site I missed. Two-part fix:\n1. The `encodeURIComponent` was already removed in 43e63ad7 (the earlier double-encoding sweep), so the bare `filters: [\`uid:${uid}\`]` is what's there now.\n2. Added the missing allowlist-validation half in f7ae893b: `fetchSelectedItem` now runs `isFilterSafeIdentifier(uid)` from the new shared utility (`packages/shared/src/utils/org-selector.utils.ts`) before interpolating. This matters because `selectedUid` arrives from `?selected_uid=` query input \u2014 untrusted client surface. Hostile inputs silently degrade to null with a warning log; client falls back to the first natural row per FR-014. Matches the defense already in place on OrgIdentityResolver.

apps/lfx-one/src/server/services/fixtures/org-selector.mock.json
```diff
@@ -0,0 +1,154 @@
+{
+  "$comment": "Spec 020 — local-dev fixture used when MOCK_ORG_ITEMS=true. Each row mirrors the upstream b2b_org indexed-doc shape AFTER the OrgNavigationService.toOrgItem mapping. ALL Salesforce IDs and LOGO_URLs are real values fetched from ANALYTICS.PLATINUM_LFX_ONE.ORG_LENS_ACCOUNT_CONTEXT — the first 8 rows intentionally share IDs with ORG_LENS_DEMO_SEED_ACCOUNTS so the existing persona-seed cookie restore continues to resolve to a real fixture row. UUIDs (the canonical b2b_org.uid) are synthetic since the b2b_org index lives in a different system, but Snowflake LOGO_URLs from lf-master-organization-logos-prod / lf-platform-documents-prod buckets are stable, accessible, and what production actually serves today.",
+  "items": [
```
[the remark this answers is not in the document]
luismoriguerra: Per spec author guidance the fixture data itself is staying as-is for now (it's a placeholder until the dev sandbox b2b_org index is seeded, gated by both `MOCK_ORG_ITEMS=true` AND `NODE_ENV!=='production'`), but you're right that the prior "sourced from production analytics" wording was alarming and made the file look more permanent than intended. Reframed in 4ab60cf9 \u2014 the top-level `$comment` now reads as a TEMPORARY local-dev seed with an explicit "REPLACE with synthetic values before GA or remove the fixture entirely once the dev sandbox b2b_org index is seeded" instruction. The identifiers and S3 URLs are LF Foundation's own public assets (logos.s3.us-east-2.amazonaws.com is a public CDN), but the framing matters \u2014 thanks for catching that.

packages/shared/src/utils/org-selector.utils.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+/** Allowlist covering every OIDC username shape (alphanumerics, dot, dash, underscore, plus, at) while excluding the filter-grammar separators `:` and `,`. */
+const FILTER_SAFE_USERNAME = /^[A-Za-z0-9._+\-@]+$/;
+
+/** Allowlist for query-service filter identifiers — UUIDs (hex + dashes) and Salesforce IDs (15 or 18 alphanumeric). */
+const FILTER_SAFE_IDENTIFIER = /^[A-Za-z0-9_-]+$/;
```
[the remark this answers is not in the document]
luismoriguerra: Correct \u2014 the regex is intentionally a SAFETY allowlist (rejects the filter-grammar separators `:` and `,`), not a strict UUID/SF-ID format validator. Tightening to enforce `^[0-9a-f]{8}-[0-9a-f]{4}-...` would actually break the resolver because Salesforce IDs are NOT UUID-shaped (15 or 18 alphanumeric, no dashes) and we use the SAME predicate for both directions of the lookup. The honest fix is updating the comment to match the code's actual intent. Done in 4ab60cf9 \u2014 both regex comments now describe themselves as "filter-grammar safety allowlists" with a "covers X as a superset" framing, and each exported function carries a "Note: this is a SAFETY check, not a strict UUID/SF-ID format validator" tag. No code change.

luismoriguerra:
## Review-thread sweep recap

All open review threads on PR #794 have been addressed and marked resolved. Summary of what shipped across the review-response commits.

### Critical correctness fixes

| # | Commit | What | Why it mattered |
|---|---|---|---|
| **NG0203 #1** | `1501408f` | Injected `DestroyRef`, passed to `takeUntilDestroyed()` inside `afterNextRender` | Bare `takeUntilDestroyed()` outside the constructor's injection context would throw `NG0203: inject() must be called from an injection context` on first browser mount |
| **NG0203 #2** | `21239357` | Also injected `Injector`, passed to `toObservable(this.enabled, { injector })` | My first NG0203 fix was incomplete — `toObservable()` also calls `inject()`. Server-side hydration skips `afterNextRender` entirely, which masked the bug in SSR-only smoke tests |
| **Double-encoding sweep** | `43e63ad7` + `f7ae893b` | Dropped `encodeURIComponent()` from three filter-interpolation sites; added `isFilterSafeIdentifier` / `isFilterSafeUsername` allowlist defense | `ApiClientService.getFullUrl` already runs values through `URLSearchParams.append` which percent-encodes once. Pre-encoding would have shipped `%2540` upstream, silently returning empty role grants for any email-style username |
| **Upstream 5xx mapping** | `21239357` | Removed `OrgIdentityResolver` catch blocks that swallowed `MicroserviceError` into `null` | Without the fix, a query-service outage on the sfid path would have been misreported as 404 "Organization not found" instead of preserving FR-020's 502 path |
| **CodeQL URL substring sanitization** | `889ed226` | Hostname-exact match in e2e auth-skip gate via `new URL().hostname === 'auth0.com' \|\| .endsWith('.auth0.com')` | Would have allowed `https://evil.com/?ref=auth0.com` or `https://auth0.com.evil.com/` to fool the gate |

### CLAUDE.md compliance (`21239357`)

Moved everything to `@lfx-one/shared` per the "no module-level consts or local `interface Foo {}` inside `apps/lfx-one/`" rule:

- **Interfaces** → `packages/shared/src/interfaces/org-selector.interface.ts`: `B2bOrgIndexedDoc`, `B2bOrgSettingsDoc`, `MemberServiceB2bOrgResponse`, `OrgIdentityLookupResult`, `OrgRolePersona` (was triplicated across three services + a re-export from the resolver — now a single source)
- **Constants** → `packages/shared/src/constants/org-selector.constants.ts`: `ORG_SELECTOR_MOCK_PAGE_SIZE`, `ORG_IDENTITY_CACHE_MAX_ENTRIES`, `ORG_IDENTITY_CACHE_TTL_MS`
- **Predicates** → new `packages/shared/src/utils/org-selector.utils.ts`: `isFilterSafeUsername`, `isFilterSafeIdentifier`
- **`UUID_PATTERN`** → reused the existing `UUID_REGEX` already in `packages/shared/src/constants/regex.constants.ts`

### Comment / framing cleanups

- **Multi-line JSDoc collapse** (`21239357`): every JSDoc block on the new spec-020 files compressed to single-line `/** … */`. Net -355 lines across the PR; spec references now live only in `specs/020-org-selector-integration/` and `contracts/` docs where they belong.
- **Misleading `MOCK_PAGE_SIZE` comment** (`21239357`): the "bigger than prod ~50" wording was a remnant from an earlier draft and the value is actually 8. New shared constant carries an accurate one-liner: "Page size used by the MOCK_ORG_ITEMS dev branch so the e2e can exercise scroll-pagination."
- **`MOCK_ORG_ITEMS` "truthy" doc mismatch** (`1a39289c`): tightened to "the literal strings `'true'` or `'1'`" — other truthy-looking values are intentionally rejected so the fail-closed gate is unambiguous.
- **"Tree-shaken" misstatement** (`1a39289c`): server Node code isn't subject to build-time dead-code elimination; the mock branch is "unreachable at runtime" via the early-return, not eliminated from the bundle.
- **Misleading `cache_hit` metric** (`1a39289c`): refactored `OrgIdentityResolver` to return `{ uid \| sfid, cacheHit }` so the controller logs an honest LRU state. New log key `identifier_cache_hit` is `null` for uid-routed requests (no resolver call) to keep the metric meaningful.
- **Fixture framing** (`4ab60cf9`): the JSON `$comment` now describes the fixture as a TEMPORARY local-dev seed with an explicit "REPLACE with synthetic values before GA or remove the fixture entirely once the dev sandbox b2b_org index is seeded" instruction. Fixture data itself is unchanged (intentional placeholder; LF Foundation's own public S3 assets).
- **`isFilterSafe*` regex comments** (`4ab60cf9`): updated to honestly describe the helpers as filter-grammar safety allowlists, not strict UUID/SF-ID format validators.

### Test coverage improvements (`1a39289c`)

- **S9 zero-grants gate test**: replaced the unauthenticated-`/login` proxy assertion with a real authenticated path. Intercepts `/api/orgs/me/role-grants` (empty writers + auditors) and `/api/user/personas` (empty `organizations`), reloads, then asserts `org-selector-slot` reports `data-visible="false"`. Genuinely exercises the `(writers ∨ auditors ∨ personaSeeds)` branch instead of structurally asserting on the login page.

### Commits this sweep produced

```
889ed226  fix(org-selector): satisfy CodeQL + Prettier on PR #794 CI
1a39289c  fix(org-selector): address Copilot PR review threads on PR #794
7b50a773  fix(org-role-grants): URL-encode username before building filters_or (RETRACTED — see 43e63ad7)
1501408f  fix(org-selector): inject DestroyRef for takeUntilDestroyed inside afterNextRender
43e63ad7  fix(org-selector): drop double-encoding on query-service filter values
21239357  refactor(org-selector): address MRashad26 + CodeRabbit review on PR #794
f7ae893b  fix(org-navigation): apply filter-safe allowlist to selectedUid in fetchSelectedItem
4ab60cf9  docs(org-selector): reframe fixture intent + clarify filter-safe allowlist comments
```

Net code delta across the response sweep: **−500+ lines** (mostly comment collapse + interface consolidation). All CI green; PR `mergeable: MERGEABLE`, only required-reviewer approval pending.

Thanks to all reviewers (human + bots) for the catches, especially the two NG0203 fixes that the standard Angular compiler doesn't flag and only surface in production hydration.

dealako (approved):
## Independent Review — PR #794 (spec 020 org-selector-integration)

**Reviewer:** David Deal (via OpenCode)
**Scope:** +1,950 lines across 26 files — 3 BFF endpoints, 3 client services, 1 component, E2E tests, shared types/utils/constants.

---

### 1. Independent Findings

#### Correctness

**[nit]** `OrgNavigationService` (client) exposes a `searchTerm(): WritableSignal<string>` method that returns the internal writable signal. This is an unusual API surface — the convention elsewhere in the codebase is to expose `Signal<T>` (readonly) publicly and a `setXyz()` setter. Not a functional bug since nothing external calls `.set()` on it, but it leaks mutability.
- `apps/lfx-one/src/app/shared/services/org-navigation.service.ts:39`

**[nit]** The polymorphic `/:id` route at the bottom of `orgs.route.ts:74` will catch ANY single-segment path under `/api/orgs/` that doesn't match an earlier route. If a future route like `/api/orgs/something-new` is added without careful positioning, it will be swallowed. The comprehensive ordering comments mitigate this — just noting as a maintenance risk.

#### Security

**No blocking security issues.** The filter-injection vectors are properly defended:

- `isFilterSafeUsername()` and `isFilterSafeIdentifier()` in `packages/shared/src/utils/org-selector.utils.ts` correctly block query-service separator characters (`:`, `,`) from user-controlled inputs before interpolation into filter strings.
- `fetchSelectedItem` applies `isFilterSafeIdentifier(uid)` on the untrusted `selected_uid` query param before constructing the upstream filter (added in `f7ae893b`).
- The double-encoding trap (Copilot caught it, author caught the fix was itself wrong) is correctly resolved — raw username/identifier values are passed to `MicroserviceProxyService` which delegates to `URLSearchParams.append` for single automatic encoding.
- `isMockOrgItemsEnabled()` hard-gates on `NODE_ENV !== 'production'` — mock data cannot leak to production.
- `OrgIdentityController` maps 5xx → 502 and 404 → 404 with a generic envelope — no information leakage.
- `encodeURIComponent(uid)` in `fetchCanonicalRecord` path construction prevents path traversal.

#### Performance

**No issues.** The `LruWithTtl` implementation is appropriate for the 10k-entry use case — avoids pulling a dependency. The lazy TTL check on read + Map insertion-order eviction is clean. `ORG_ROLE_GRANTS_HARD_CAP = 500` is a reasonable ceiling for the single-request role-grants fetch.

#### Test Coverage

- E2E covers S1 (trigger visibility), S2 (search), S5 (selection + canonical fetch), S9 (zero-grants gate).
- The S9 test uses proper Playwright route stubs and reloads to ensure fresh interception — well-structured.
- No unit tests for the server services. This is consistent with the existing codebase pattern (E2E-first for BFF endpoints). Not a blocker but noted.

#### Code Style / Codebase Consistency

**[minor]** The PR modifies `CLAUDE.md` (a protected file per lfx-preflight) — adds the "Work cycle" and "Recent Specs" sections, plus clarifies the commit header-max-length rule. These are valid documentation improvements but should be flagged per the protected-file policy as requiring code-owner awareness.

**[nit]** `OrgNavigationService` (client) at line 39 defines `searchTerm()` as a method returning a `WritableSignal` rather than a property. The sister `NavigationService` uses the same pattern, so this is internally consistent.

#### Documentation

The PR description, JSDoc, inline comments, and the `.env.example` annotation are all thorough and well-aligned with the spec. The `$comment` field in the mock JSON is a nice touch for developer context.

---

### 2. AI Comment Reconciliation

| Bot | Issue | Status |
|-----|-------|--------|
| **CodeQL** | URL substring sanitization in E2E | **Resolved** in `889ed226` — hostname-exact matching via `new URL()`. Agree. |
| **CodeRabbit** | `encodeURIComponent(username)` in `org-role-grants.service.ts:45` | **Resolved** in `43e63ad7` — correctly dropped the manual encoding since `URLSearchParams.append` already encodes. The author caught that CodeRabbit's original suggestion would have introduced double-encoding. |
| **CodeRabbit** | `toObservable(this.enabled)` + `takeUntilDestroyed` injection context | **Resolved** in `1501408f` + `21239357` — `DestroyRef` + `Injector` captured at construction and passed explicitly to `afterNextRender` callback. |
| **Copilot** | Missing `ORG_ITEMS_PER_PAGE` constant | Author explained — trimmed because the BFF delegates page sizing to the upstream query-service. Agree this is a non-issue. |
| **Copilot** | `cacheHit` logging inaccuracy | **Resolved** in `1a39289c` — `OrgIdentityResolver` now returns `{ uid/sfid, cacheHit }` tuple. |
| **Copilot** | Mock branch "tree-shaken" comment | **Resolved** in `1a39289c` — reworded to clarify Node server code isn't DCE'd. |
| **Copilot** | E2E S9 test quality (unauthenticated `/login` surface) | **Resolved** in `1a39289c` — replaced with authenticated session + stubbed empty grants. |
| **Copilot** | Double-encoding in `org-role-grants.service.ts` | **Resolved** in `43e63ad7`. Author correctly identified and closed the loop. |
| **MRashad26** | `toObservable()` + `takeUntilDestroyed()` injection context bug | **Resolved** in `1501408f`. |
| **MRashad26** | Multi-line JSDoc violations (x10+) | **Resolved** in `21239357` — confirmed: all remaining JSDoc comments are single-line `/** ... */` format. |

**Disagreements:** None. All bot findings were either resolved or correctly dismissed by the author.

---

### 3. Revision Tracking

| Commit | Resolves |
|--------|----------|
| `889ed226` | CodeQL URL substring sanitization + Prettier CI |
| `1a39289c` | Copilot: cacheHit logging, mock-branch comment, S9 test |
| `7b50a773` | CodeRabbit: encode username in role-grants filter |
| `1501408f` | MRashad26/CodeRabbit: inject DestroyRef for afterNextRender |
| `43e63ad7` | Copilot: drop double-encoding (corrects `7b50a773`) |
| `21239357` | MRashad26: JSDoc to single-line, misc refactoring |
| `f7ae893b` | Security: filter-safe allowlist on selectedUid |
| `4ab60cf9` | Docs: fixture comment + allowlist clarification |

---

### 4. Summary

Hey Luis — solid work on this. The feature is well-architected: the BFF layer correctly mirrors the proven `NavigationService` / `project-selector` pattern, the filter-injection defense is thorough (filter-safe allowlists on all untrusted interpolation points), and the iteration history shows disciplined triage of every bot finding including catching a double-encoding bug that the original fix introduced.

**Blocking issues:** 0
**Minor issues:** 1 (CLAUDE.md as protected file — code-owner awareness)
**Nits:** 2 (leaked WritableSignal, catch-all route maintenance risk)

**Decision: Approved.** No blockers remain, all prior Critical findings are addressed, and the code is production-ready.

MRashad26 (approved):

PR #795, opened 2026-05-26
fix(nps-card): fall back to most-recent prior period when YTD has no data

MRashad26:
## Summary

- **Root cause**: `getNpsSummary()` previously defaulted period-scoped values in a way that could hide missing YTD data and, in earlier iterations, risk mixing values across periods.
- **Fix**: The query now resolves a single effective period using fallback ordering from the requested range, and all returned NPS metrics are selected from that same period via aligned `CASE WHEN` logic.
- **API behavior**: `periodLabel` is now only populated when a valid effective period exists; otherwise it returns an empty string.
- **UX**: The NPS card shows a blue period badge when `periodLabel` is present, and keeps a clean single-label header when it is not.

## Changes

| File | Change |
|------|--------|
| `packages/shared/src/interfaces/dashboard-metric.interface.ts` | `NpsSummaryResponse` now uses `range: HealthMetricsRange` and `periodLabel: string` (JSDoc updated to match `range`) |
| `packages/shared/src/constants/dashboard-metrics.constants.ts` | Updated `HEALTH_METRICS_NPS_DEFAULT_SUMMARY` to match the `range` field and defaults |
| `apps/lfx-one/src/server/services/project.service.ts` | Reworked `getNpsSummary()` fallback selection and aligned all period-scoped metric reads to one resolved period; `periodLabel` is empty when no valid effective range is found |
| `apps/lfx-one/src/app/shared/services/analytics.service.ts` | Uses `HEALTH_METRICS_NPS_DEFAULT_SUMMARY` in `catchError` fallback |
| `apps/lfx-one/src/app/modules/dashboards/health-metrics/nps-card/nps-card.component.html` | Period badge rendering is conditional; no extra flex wrapper when `periodLabel` is absent |

## Test plan

Scenarios covered for verification:
- Project with YTD data shows `YTD <year>` badge and correct NPS metrics
- Project with no YTD data falls back to most-recent prior year (for example `FY <year>`) with real values
- Project with no data renders empty state and no period badge
- Requesting `COMPLETED_YEAR` with no data falls back to `COMPLETED_YEAR_2` (not forward to YTD)
- `periodLabel` remains empty when no effective range exists

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/health-metrics/nps-card/nps-card.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/health-metrics/nps-card/nps-card.component.html
@@ -7,7 +7,14 @@
     <div class="flex flex-col">
       <h3 class="text-sm font-bold tracking-wide text-gray-900 uppercase">Net Promoter Score</h3>
       @if (hasData()) {
-        <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
+        <div class="flex items-center gap-2">
+          @if (summaryData().periodLabel) {
+            <span class="rounded bg-blue-50 px-1.5 py-0.5 text-xs font-medium text-blue-700" data-testid="nps-card-period-label">{{
+              summaryData().periodLabel
+            }}</span>
+          }
+          <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
+        </div>
       }
     </div>
     @if (hasData()) {
--- a/apps/lfx-one/src/app/modules/dashboards/health-metrics/nps-card/nps-card.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/health-metrics/nps-card/nps-card.component.html
@@ -7,14 +7,16 @@
     <div class="flex flex-col">
       <h3 class="text-sm font-bold tracking-wide text-gray-900 uppercase">Net Promoter Score</h3>
       @if (hasData()) {
-        <div class="flex items-center gap-2">
-          @if (summaryData().periodLabel) {
+        @if (summaryData().periodLabel) {
+          <div class="flex items-center gap-2">
             <span class="rounded bg-blue-50 px-1.5 py-0.5 text-xs font-medium text-blue-700" data-testid="nps-card-period-label">{{
               summaryData().periodLabel
             }}</span>
-          }
+            <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
+          </div>
+        } @else {
           <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
-        </div>
+        }
       }
     </div>
     @if (hasData()) {
--- a/apps/lfx-one/src/app/shared/services/analytics.service.ts
+++ b/apps/lfx-one/src/app/shared/services/analytics.service.ts
@@ -71,6 +71,7 @@ import {
   MarketingAttributionResponse,
   MultiFoundationSummaryResponse,
 } from '@lfx-one/shared/interfaces';
+import { HEALTH_METRICS_NPS_DEFAULT_SUMMARY } from '@lfx-one/shared/constants';
 import { catchError, Observable, of, shareReplay } from 'rxjs';
 
 /**
@@ -1002,22 +1003,7 @@ export class AnalyticsService {
     if (range && range !== 'YTD') {
       params['range'] = range;
     }
-    return this.http.get<NpsSummaryResponse>('/api/analytics/nps-summary', { params }).pipe(
-      catchError(() => {
-        return of({
-          projectId: '',
-          npsScore: 0,
-          promoters: 0,
-          passives: 0,
-          detractors: 0,
-          nonResponses: 0,
-          responses: 0,
-          lastUpdatedLabel: 'N/A',
-          effectiveRange: 'YTD' as const,
-          periodLabel: '',
-        });
-      })
-    );
+    return this.http.get<NpsSummaryResponse>('/api/analytics/nps-summary', { params }).pipe(catchError(() => of(HEALTH_METRICS_NPS_DEFAULT_SUMMARY)));
   }
 
   /**
--- a/apps/lfx-one/src/app/shared/services/analytics.service.ts
+++ b/apps/lfx-one/src/app/shared/services/analytics.service.ts
@@ -1013,6 +1013,8 @@ export class AnalyticsService {
           nonResponses: 0,
           responses: 0,
           lastUpdatedLabel: 'N/A',
+          effectiveRange: 'YTD' as const,
+          periodLabel: '',
         });
       })
     );
--- a/apps/lfx-one/src/server/services/project.service.ts
+++ b/apps/lfx-one/src/server/services/project.service.ts
@@ -3751,7 +3751,11 @@ export class ProjectService {
 
     const resolvedRange = isHealthMetricsRange(row.EFFECTIVE_RANGE) ? row.EFFECTIVE_RANGE : null;
 
-    const periodLabel = resolvedRange ? (resolvedRange === 'YTD' ? `YTD ${getYearForRange(resolvedRange)}` : `FY ${getYearForRange(resolvedRange)}`) : '';
+    let periodLabel = '';
+    if (resolvedRange) {
+      const year = getYearForRange(resolvedRange);
+      periodLabel = resolvedRange === 'YTD' ? `YTD ${year}` : `FY ${year}`;
+    }
 
     const promoters = row.PROMOTERS ?? 0;
     const passives = row.PASSIVES ?? 0;
--- a/apps/lfx-one/src/server/services/project.service.ts
+++ b/apps/lfx-one/src/server/services/project.service.ts
@@ -3690,7 +3690,7 @@ export class ProjectService {
     // all other columns use the identical WHEN ordering so no period mixing occurs.
     const buildCaseWhen = (colPrefix: string): string =>
       fallbackRanges
-        .map((r, i) => `WHEN sr.most_recent_nps_score${fallbackSuffixes[i]} IS NOT NULL THEN sr.${colPrefix}${fallbackSuffixes[i]}`)
+        .map((_, i) => `WHEN sr.most_recent_nps_score${fallbackSuffixes[i]} IS NOT NULL THEN sr.${colPrefix}${fallbackSuffixes[i]}`)
         .join('\n        ');
 
     const rangeCases = fallbackRanges.map((r, i) => `WHEN sr.most_recent_nps_score${fallbackSuffixes[i]} IS NOT NULL THEN '${r}'`).join('\n        ');
@@ -3749,10 +3749,9 @@ export class ProjectService {
       }
     }
 
-    const resolvedRange: HealthMetricsRange = isHealthMetricsRange(row.EFFECTIVE_RANGE) ? row.EFFECTIVE_RANGE : requestedRange;
+    const resolvedRange = isHealthMetricsRange(row.EFFECTIVE_RANGE) ? row.EFFECTIVE_RANGE : null;
 
-    const year = getYearForRange(resolvedRange);
-    const periodLabel = resolvedRange === 'YTD' ? `YTD ${year}` : `FY ${year}`;
+    const periodLabel = resolvedRange ? (resolvedRange === 'YTD' ? `YTD ${getYearForRange(resolvedRange)}` : `FY ${getYearForRange(resolvedRange)}`) : '';
 
     const promoters = row.PROMOTERS ?? 0;
     const passives = row.PASSIVES ?? 0;
@@ -3768,7 +3767,7 @@ export class ProjectService {
       nonResponses,
       responses: promoters + passives + detractors + nonResponses,
       lastUpdatedLabel,
-      range: resolvedRange,
+      range: resolvedRange ?? requestedRange,
       periodLabel,
       changeNpsScore: row.CHANGE_NPS_SCORE ?? 0,
     };
--- a/apps/lfx-one/src/server/services/project.service.ts
+++ b/apps/lfx-one/src/server/services/project.service.ts
@@ -1,7 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { NATS_CONFIG, PENDING_ACTION_SEVERITY, PENDING_ACTION_SURVEYS_ROW_LIMIT, ROOT_PROJECT_SLUG } from '@lfx-one/shared/constants';
+import { getYearForRange, NATS_CONFIG, PENDING_ACTION_SEVERITY, PENDING_ACTION_SURVEYS_ROW_LIMIT, ROOT_PROJECT_SLUG } from '@lfx-one/shared/constants';
 import { NatsSubjects } from '@lfx-one/shared/enums';
 import {
   BoardMeetingInviteeRow,
@@ -3662,18 +3662,29 @@ export class ProjectService {
   public async getNpsSummary(foundationSlug: string, range: string = 'YTD'): Promise<NpsSummaryResponse> {
     interface NpsSummaryRow {
       PROJECT_ID: string;
-      NPS_SCORE: number;
-      PROMOTERS: number;
-      PASSIVES: number;
-      DETRACTORS: number;
-      NON_RESPONSES: number;
+      NPS_SCORE: number | null;
+      PROMOTERS: number | null;
+      PASSIVES: number | null;
+      DETRACTORS: number | null;
+      NON_RESPONSES: number | null;
       LAST_UPDATED: string | null;
-      CHANGE_NPS_SCORE: number;
+      CHANGE_NPS_SCORE: number | null;
+      EFFECTIVE_RANGE: string | null;
     }
 
-    const VALID_RANGES = ['YTD', 'COMPLETED_YEAR', 'COMPLETED_YEAR_2', 'COMPLETED_YEAR_3', 'COMPLETED_YEAR_4'];
-    const effectiveRange = VALID_RANGES.includes(range) ? range : 'YTD';
-    const suffix = this.getRangeSuffix(effectiveRange);
+    const ORDERED_RANGES: HealthMetricsRange[] = ['YTD', 'COMPLETED_YEAR', 'COMPLETED_YEAR_2', 'COMPLETED_YEAR_3', 'COMPLETED_YEAR_4'];
+    const requestedRange: HealthMetricsRange = ORDERED_RANGES.includes(range as HealthMetricsRange) ? (range as HealthMetricsRange) : 'YTD';
+    const startIdx = ORDERED_RANGES.indexOf(requestedRange);
+    const fallbackRanges = ORDERED_RANGES.slice(startIdx);
+    const fallbackSuffixes = fallbackRanges.map((r) => this.getRangeSuffix(r));
+
+    const npsCoalesce = fallbackSuffixes.map((sfx) => `sr.most_recent_nps_score${sfx}`).join(', ');
+    const promotersCoalesce = fallbackSuffixes.map((sfx) => `sr.most_recent_count_promoters${sfx}`).join(', ');
+    const passivesCoalesce = fallbackSuffixes.map((sfx) => `sr.most_recent_count_passives${sfx}`).join(', ');
+    const detractorsCoalesce = fallbackSuffixes.map((sfx) => `sr.most_recent_count_detractors${sfx}`).join(', ');
+    const nonResponsesCoalesce = fallbackSuffixes.map((sfx) => `sr.most_recent_count_non_responses${sfx}`).join(', ');
+    const changeCoalesce = fallbackSuffixes.map((sfx) => `sr.nps_score_most_recent_to_previous_change${sfx}`).join(', ');
+    const rangeCases = fallbackRanges.map((r, i) => `WHEN sr.most_recent_nps_score${fallbackSuffixes[i]} IS NOT NULL THEN '${r}'`).join('\n        ');
 
     const query = `
       WITH slug_resolve AS (
@@ -3683,13 +3694,17 @@ export class ProjectService {
       )
       SELECT
         sr.project_id AS PROJECT_ID,
-        IFNULL(sr.most_recent_nps_score${suffix}, 0) AS NPS_SCORE,
-        IFNULL(sr.most_recent_count_promoters${suffix}, 0) AS PROMOTERS,
-        IFNULL(sr.most_recent_count_passives${suffix}, 0) AS PASSIVES,
-        IFNULL(sr.most_recent_count_detractors${suffix}, 0) AS DETRACTORS,
-        IFNULL(sr.most_recent_count_non_responses${suffix}, 0) AS NON_RESPONSES,
+        COALESCE(${npsCoalesce}) AS NPS_SCORE,
+        COALESCE(${promotersCoalesce}) AS PROMOTERS,
+        COALESCE(${passivesCoalesce}) AS PASSIVES,
+        COALESCE(${detractorsCoalesce}) AS DETRACTORS,
+        COALESCE(${nonResponsesCoalesce}) AS NON_RESPONSES,
         sr.last_updated AS LAST_UPDATED,
-        IFNULL(sr.nps_score_most_recent_to_previous_change${suffix}, 0) AS CHANGE_NPS_SCORE
+        COALESCE(${changeCoalesce}) AS CHANGE_NPS_SCORE,
+        CASE
+        ${rangeCases}
+        ELSE NULL
+        END AS EFFECTIVE_RANGE
       FROM ANALYTICS.PLATINUM.SURVEY_RESPONSES sr
       INNER JOIN slug_resolve s ON sr.project_id = s.project_id
       LIMIT 1
@@ -3707,6 +3722,8 @@ export class ProjectService {
         nonResponses: 0,
         responses: 0,
         lastUpdatedLabel: 'N/A',
+        effectiveRange: requestedRange,
+        periodLabel: '',
       };
     }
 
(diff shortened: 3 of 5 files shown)


apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -3662,18 +3662,29 @@ export class ProjectService {
...
       LAST_UPDATED: string | null;
-      CHANGE_NPS_SCORE: number;
+      CHANGE_NPS_SCORE: number | null;
+      EFFECTIVE_RANGE: string | null;
     }
 
-    const VALID_RANGES = ['YTD', 'COMPLETED_YEAR', 'COMPLETED_YEAR_2', 'COMPLETED_YEAR_3', 'COMPLETED_YEAR_4'];
-    const effectiveRange = VALID_RANGES.includes(range) ? range : 'YTD';
-    const suffix = this.getRangeSuffix(effectiveRange);
+    const ORDERED_RANGES: HealthMetricsRange[] = ['YTD', 'COMPLETED_YEAR', 'COMPLETED_YEAR_2', 'COMPLETED_YEAR_3', 'COMPLETED_YEAR_4'];
```
audigregorie: **Major — duplicate of shared `HEALTH_METRICS_RANGES`**

`HEALTH_METRICS_RANGES` (identical readonly tuple) and the matching `isHealthMetricsRange()` type guard are already exported from `@lfx-one/shared/constants` (see `packages/shared/src/constants/dashboard-metrics.constants.ts:31` and `:37`). Per CLAUDE.md: _"All shared constants and interfaces live in `@lfx-one/shared` — no module-level consts inside `apps/lfx-one/`."_

Declaring a second copy here means a future range addition has to be made in two places, with no compile-time link between them.

```suggestion
    const requestedRange: HealthMetricsRange = isHealthMetricsRange(range) ? range : 'YTD';
    const startIdx = HEALTH_METRICS_RANGES.indexOf(requestedRange);
    const fallbackRanges = HEALTH_METRICS_RANGES.slice(startIdx);
    const fallbackSuffixes = fallbackRanges.map((r) => this.getRangeSuffix(r));
```

Then add `HEALTH_METRICS_RANGES, isHealthMetricsRange` to the import on line 4 and update the `row.EFFECTIVE_RANGE` validation lower down to use the same helpers.

packages/shared/src/interfaces/dashboard-metric.interface.ts
```diff
@@ -445,6 +445,10 @@ export interface NpsSummaryResponse {
   responses: number;
   /** Human-readable reporting period for Last Updated (e.g. "Q4 2025" or "N/A") */
   lastUpdatedLabel: string;
+  /** The range that actually has data — may differ from the requested range when YTD is empty and the BFF fell back to a prior year */
+  effectiveRange: HealthMetricsRange;
```
audigregorie: **Minor — field name diverges from every other Health Metrics summary response**

Every sibling card in this file already uses `range: HealthMetricsRange` for the same semantic role (the validated / effective range echoed back by the BFF):

- `MembershipChurnPerTierSummaryResponse.range`
- `BoardMeetingParticipationSummaryResponse.range`
- `TrainingCertificationSummaryResponse.range`
- `CodeContributionSummaryResponse.range`

Introducing `effectiveRange` here breaks that convention. Consumers reading the union of card responses now have to remember NPS is the odd one out.

Since the value here is always the resolved/effective range after the BFF's validation + fallback (i.e. it never carries the raw user input), I'd suggest renaming to `range` for consistency:

```suggestion
  /** Range whose data was actually returned — may differ from the requested range when YTD is empty and the BFF fell back to a prior year */
  range: HealthMetricsRange;
```

If you want to keep the distinction explicit, please add a JSDoc note explaining why this card alone names it differently.

apps/lfx-one/src/app/modules/dashboards/health-metrics/nps-card/nps-card.component.html
```diff
@@ -7,7 +7,14 @@
...
       @if (hasData()) {
-        <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
+        <div class="flex items-center gap-2">
+          @if (summaryData().periodLabel) {
+            <span class="rounded bg-blue-50 px-1.5 py-0.5 text-xs font-medium text-blue-700" data-testid="nps-card-period-label">{{
+              summaryData().periodLabel
+            }}</span>
+          }
+          <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
+        </div>
```
audigregorie: **Minor — flex wrapper renders even when there's only one child**

When `summaryData().periodLabel` is empty (e.g. the empty-state default summary, the `catchError` fallback, or any older deployment that hasn't been redeployed yet), this still renders a `<div class="flex items-center gap-2">` wrapping a single inline `<span>`. Harmless but wasteful markup, and it changes the inline/block semantics from the original `<span>` — only inside a `flex flex-col` parent does that not visually matter.

Cleaner to move the `flex` wrapper inside the period-label `@if` (so the single-child case stays a bare span), or use a single conditional class binding.

```suggestion
        @if (summaryData().periodLabel) {
          <div class="flex items-center gap-2">
            <span class="rounded bg-blue-50 px-1.5 py-0.5 text-xs font-medium text-blue-700" data-testid="nps-card-period-label">{{
              summaryData().periodLabel
            }}</span>
            <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
          </div>
        } @else {
          <span class="text-xs text-gray-500">Last Updated: {{ summaryData().lastUpdatedLabel }}</span>
        }
```

audigregorie (commented):
## Code Review Summary

Nice targeted fix for the YTD-empty case. The new COALESCE + CASE pattern, defensive validation of `EFFECTIVE_RANGE`, and `number | null` row typing are all improvements over the old `IFNULL(..., 0)` design. A few notes below; nothing critical.

### Outside the diff

- **Independent COALESCE chains across columns** — Copilot already flagged this on lines 3701 and 3703. Worth fixing in the same PR: the per-column COALESCE means `NPS_SCORE` could come from `_ytd` while `PROMOTERS` comes from `_last_completed_year`, mixing periods. Simpler and safer to pick the effective range first (e.g. via a `WITH` CTE that determines the suffix), then read every column from that single suffix. I'm not re-posting this as a separate thread to avoid duplicating Copilot.
- **PR description test plan** — all 5 checkboxes are unchecked. If you exercised them locally, please tick them; if not, please document what's still pending verification.
- **Observability** — no `logger.debug`/`logger.info` is emitted when the BFF falls back from the requested range to an older one. Adding this would help telemetry on which foundations have stale data. Would require threading `req` into `getNpsSummary()`, so reasonable to defer.

### What's done well

- Strong type-safety on `NpsSummaryRow` — every Snowflake column is now `number | null`, matching real driver behaviour.
- Defensive `EFFECTIVE_RANGE` parsing on the server before trusting the DB value.
- SQL injection-safe: suffixes come from a static map, range values are whitelisted before interpolation, and `foundationSlug` stays parameter-bound.
- Frontend correctly hides the badge when `periodLabel` is empty so old / error responses don't pollute the header.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -3683,13 +3694,17 @@ export class ProjectService {
...
-        IFNULL(sr.most_recent_nps_score${suffix}, 0) AS NPS_SCORE,
-        IFNULL(sr.most_recent_count_promoters${suffix}, 0) AS PROMOTERS,
-        IFNULL(sr.most_recent_count_passives${suffix}, 0) AS PASSIVES,
-        IFNULL(sr.most_recent_count_detractors${suffix}, 0) AS DETRACTORS,
-        IFNULL(sr.most_recent_count_non_responses${suffix}, 0) AS NON_RESPONSES,
+        COALESCE(${npsCoalesce}) AS NPS_SCORE,
+        COALESCE(${promotersCoalesce}) AS PROMOTERS,
+        COALESCE(${passivesCoalesce}) AS PASSIVES,
+        COALESCE(${detractorsCoalesce}) AS DETRACTORS,
+        COALESCE(${nonResponsesCoalesce}) AS NON_RESPONSES,
```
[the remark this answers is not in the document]
MRashad26: Fixed — replaced per-column COALESCE with per-column CASE WHEN blocks using the same fallback order as EFFECTIVE_RANGE. Every metric (NPS_SCORE, PROMOTERS, PASSIVES, DETRACTORS, NON_RESPONSES, CHANGE_NPS_SCORE) now reads from the same winning suffix, so no period mixing is possible.

See commit 40ed9084: replace per-column COALESCE with CASE WHEN to lock all fields to the same period.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -3683,13 +3694,17 @@ export class ProjectService {
...
         sr.last_updated AS LAST_UPDATED,
-        IFNULL(sr.nps_score_most_recent_to_previous_change${suffix}, 0) AS CHANGE_NPS_SCORE
+        COALESCE(${changeCoalesce}) AS CHANGE_NPS_SCORE,
```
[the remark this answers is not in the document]
MRashad26: Fixed as part of the same change — CHANGE_NPS_SCORE now uses the same CASE WHEN ordering as all other metrics, so it reads from the same period as NPS_SCORE.

See commit 40ed9084.

apps/lfx-one/src/app/shared/services/analytics.service.ts
```diff
@@ -1013,6 +1013,8 @@ export class AnalyticsService {
           nonResponses: 0,
           responses: 0,
           lastUpdatedLabel: 'N/A',
+          effectiveRange: 'YTD' as const,
+          periodLabel: '',
```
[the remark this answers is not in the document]
MRashad26: Fixed — replaced the hardcoded inline fallback object with `of(HEALTH_METRICS_NPS_DEFAULT_SUMMARY)`. Good catch; this eliminates the drift risk if fields are added to the interface in future.

See commit 40ed9084: analytics.service.ts: replace hardcoded catchError fallback with HEALTH_METRICS_NPS_DEFAULT_SUMMARY.

apps/lfx-one/src/server/services/project.service.ts
MRashad26: Fixed — removed the local `ORDERED_RANGES` array and now imports `HEALTH_METRICS_RANGES` and `isHealthMetricsRange()` from `@lfx-one/shared/constants`. The EFFECTIVE_RANGE validation also simplified to a single `isHealthMetricsRange()` type guard call.

See commit 40ed9084: project.service.ts: remove local ORDERED_RANGES; import shared helpers.

packages/shared/src/interfaces/dashboard-metric.interface.ts
MRashad26: Fixed — renamed `effectiveRange` → `range` in `NpsSummaryResponse`, `HEALTH_METRICS_NPS_DEFAULT_SUMMARY`, and all call sites. NPS now matches `MembershipChurnPerTierSummaryResponse`, `BoardMeetingParticipationSummaryResponse`, `TrainingCertificationSummaryResponse`, and `CodeContributionSummaryResponse` — no more odd one out.

See commit 40ed9084: dashboard-metric.interface.ts: rename effectiveRange → range.

apps/lfx-one/src/app/modules/dashboards/health-metrics/nps-card/nps-card.component.html
MRashad26: Fixed — moved the `flex` wrapper inside `@if(summaryData().periodLabel)` and added an `@else` branch that renders just the bare `<span>`. When there's no period label, the markup is a single span with no wrapping div.

See commit 40ed9084: nps-card.component.html: move flex wrapper inside @if(periodLabel).

MRashad26:
## Review Feedback Addressed

Commit: 40ed9084be25b8da7662939302879ac99feaed26

### Changes Made
- **project.service.ts**: Replaced per-column `COALESCE(...)` with per-column `CASE WHEN` blocks using the same fallback ordering as `EFFECTIVE_RANGE` — NPS_SCORE, PROMOTERS, PASSIVES, DETRACTORS, NON_RESPONSES, and CHANGE_NPS_SCORE are now guaranteed to all come from the same period, no mixing possible (per @copilot-pull-request-reviewer)
- **project.service.ts**: Removed local `ORDERED_RANGES` duplicate; now imports `HEALTH_METRICS_RANGES` and `isHealthMetricsRange()` from `@lfx-one/shared/constants`; simplified `EFFECTIVE_RANGE` validation to a single type guard call (per @audigregorie)
- **dashboard-metric.interface.ts**: Renamed `effectiveRange` → `range` to match every sibling summary response type (per @audigregorie)
- **dashboard-metrics.constants.ts**: Updated `HEALTH_METRICS_NPS_DEFAULT_SUMMARY` to use the `range` field name
- **analytics.service.ts**: Replaced hardcoded `catchError` fallback with `of(HEALTH_METRICS_NPS_DEFAULT_SUMMARY)` to eliminate drift risk (per @coderabbitai)
- **nps-card.component.html**: Moved `flex` wrapper inside `@if(summaryData().periodLabel)` and added `@else` with a bare `<span>` — no empty wrapper element when period label is absent (per @audigregorie)

### Threads Resolved

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -3723,16 +3749,28 @@ export class ProjectService {
       }
     }
 
+    const resolvedRange: HealthMetricsRange = isHealthMetricsRange(row.EFFECTIVE_RANGE) ? row.EFFECTIVE_RANGE : requestedRange;
+
+    const year = getYearForRange(resolvedRange);
+    const periodLabel = resolvedRange === 'YTD' ? `YTD ${year}` : `FY ${year}`;
```
dealako: **[minor]** When `EFFECTIVE_RANGE` comes back `null` (row exists but every fallback period has null NPS scores), `isHealthMetricsRange(null)` is `false`, so `resolvedRange` falls back to `requestedRange` and `periodLabel` is computed as a non-empty string (e.g. `"YTD 2026"`). The `@if(hasData())` template guard prevents the badge from rendering since all counts resolve to 0 — so there is no UX regression — but the API response semantically claims a period was selected when none was.

Suggestion: gate `periodLabel` computation on whether `EFFECTIVE_RANGE` is actually valid:

```typescript
const resolvedRange = isHealthMetricsRange(row.EFFECTIVE_RANGE) ? row.EFFECTIVE_RANGE : null;
const periodLabel = resolvedRange
  ? resolvedRange === 'YTD'
    ? `YTD ${getYearForRange(resolvedRange)}`
    : `FY ${getYearForRange(resolvedRange)}`
  : '';
```

Agreeing with @copilot-pull-request-reviewer (comment #3311143099).

packages/shared/src/interfaces/dashboard-metric.interface.ts
```diff
@@ -445,6 +445,10 @@ export interface NpsSummaryResponse {
   responses: number;
   /** Human-readable reporting period for Last Updated (e.g. "Q4 2025" or "N/A") */
   lastUpdatedLabel: string;
+  /** Range whose data was actually returned — may differ from the requested range when YTD is empty and the BFF fell back to a prior year */
+  range: HealthMetricsRange;
+  /** Human-readable period label derived from effectiveRange (e.g. "YTD 2026" or "FY 2025") */
+  periodLabel: string;
```
dealako, line 451: **[nit]** JSDoc reads "derived from effectiveRange" but the field was renamed to `range` in commit `40ed9084`. Update to "derived from `range`" to avoid confusing API consumers who search for the old field name.

Agreeing with @copilot-pull-request-reviewer (comment #3311143182).

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -3662,18 +3670,30 @@ export class ProjectService {
...
+    const requestedRange: HealthMetricsRange = isHealthMetricsRange(range) ? range : 'YTD';
+    const startIdx = HEALTH_METRICS_RANGES.indexOf(requestedRange);
+    const fallbackRanges = HEALTH_METRICS_RANGES.slice(startIdx);
+    const fallbackSuffixes = fallbackRanges.map((r) => this.getRangeSuffix(r));
+
+    // Build CASE WHEN blocks so every column reads from the same period.
+    // The first suffix whose NPS score is non-null defines the effective range;
+    // all other columns use the identical WHEN ordering so no period mixing occurs.
+    const buildCaseWhen = (colPrefix: string): string =>
+      fallbackRanges
```
dealako, line 3692: **[nit]** `buildCaseWhen` callback is `(r, i) => ...` but `r` is never used — only `i` (via `fallbackSuffixes[i]`) is consumed. The `rangeCases` lambda on line 3696 correctly uses `r` for the string literal. Consider renaming to `(_, i)` to signal the intentional skip.

dealako (dismissed):
Hey Rashad! 👋 Nice work on this one — the root cause diagnosis is accurate, the CASE WHEN fallback chain is the right fix for the period-mixing issue raised in round 1, and the `fix(review)` commit cleanly addressed all six threads from @audigregorie, @copilot-pull-request-reviewer, and @coderabbitai. No blocking issues; two minor findings below (both already noted by Copilot), one nit, and a note about E2E coverage.

---

## Independent review

### Security
The SQL template is **safe**. `foundationSlug` is parameterized (bind variable `?`). All column suffixes flow exclusively from `getRangeSuffix()` acting on `HEALTH_METRICS_RANGES` — a hardcoded readonly constant — and the `isHealthMetricsRange()` type guard ensures the `range` parameter is sanitized before any use. The range literals embedded in `rangeCases` (`'${r}'`) originate from that same constant and cannot be influenced by user input.

### Correctness
**`[minor]`** — When a survey row exists but every fallback period has a null NPS score, `row.EFFECTIVE_RANGE` is `null`. `isHealthMetricsRange(null)` returns `false`, so `resolvedRange` falls back to `requestedRange`, and `periodLabel` is computed as a non-empty string (e.g., `"YTD 2026"`). The card won't show the badge because `@if(hasData())` gates it (all counts resolve to 0), so there's no UX regression — but the API response semantically claims a period was selected when none was. The fix is one conditional: return `periodLabel: ''` when `!isHealthMetricsRange(row.EFFECTIVE_RANGE)`. *(@copilot-pull-request-reviewer flagged this in comment #3311143099 — agreeing.)*

**`[nit]`** — The JSDoc for `periodLabel` (`dashboard-metric.interface.ts:451`) reads "derived from effectiveRange" but the field is now `range`. *(@copilot-pull-request-reviewer flagged this in comment #3311143182 — agreeing.)*

**`[nit]`** — In `buildCaseWhen`, the map callback is `(r, i) => ...` but `r` is never used; only `i` (via `fallbackSuffixes[i]`) is consumed. The `rangeCases` lambda correctly uses `r` for the literal. No impact, just a minor hygiene issue.

### Test coverage
**`[minor]`** — A `data-testid="nps-card-period-label"` attribute was added, and the test plan explicitly lists 5 scenarios (including "no period badge shown" for the no-data case). Per project convention, `data-testid` additions are paired with E2E test coverage. No spec file is included in this diff. Worth a follow-up ticket or a structural E2E test added to the existing `nps-card` spec if one exists.

---

## AI comment reconciliation

| Round-1 thread | Resolved by `40ed9084`? | Notes |
|---|---|---|
| @copilot #3307451090 — per-column COALESCE mixes periods | ✅ Yes | Per-column CASE WHEN with identical `most_recent_nps_score` guard; all 6 metrics locked to same period |
| @copilot #3307451112 — CHANGE_NPS_SCORE mixes independently | ✅ Yes | Same CASE WHEN guard applied |
| @coderabbitai #3307454447 — use shared constant in catchError | ✅ Yes | `of(HEALTH_METRICS_NPS_DEFAULT_SUMMARY)` used |
| @audigregorie #3310265000 — duplicate HEALTH_METRICS_RANGES | ✅ Yes | Imports `HEALTH_METRICS_RANGES` + `isHealthMetricsRange()` from `@lfx-one/shared/constants` |
| @audigregorie #3310265029 — field name `effectiveRange` vs `range` | ✅ Yes | Renamed to `range`; matches all sibling card response types |
| @audigregorie #3310265038 — flex wrapper renders empty | ✅ Yes | `flex` wrapper now inside `@if(periodLabel)` with bare `@else` span |
| @copilot #3311143099 — `periodLabel` non-empty when EFFECTIVE_RANGE null | ⚠️ Open | See `[minor]` finding above |
| @copilot #3311143182 — JSDoc references `effectiveRange` | ⚠️ Open | See `[nit]` finding above |

---

## Summary

**0 blocking / 2 minor / 2 nit.** The fix is architecturally sound — replacing `IFNULL(col, 0)` with a unified CASE WHEN fallback chain ensures consistent period selection across all returned columns, and surfacing `periodLabel` in the card header gives users clear visibility into which period's data they're looking at. The two open minor/nit issues are low-risk (one is purely semantic at the API level, the other is a documentation typo); addressing them in a follow-up commit or quick fixup before merge is recommended.

**Decision: ✅ Approved with minor comments.**

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -3723,16 +3749,28 @@ export class ProjectService {
       }
     }
 
+    const resolvedRange: HealthMetricsRange = isHealthMetricsRange(row.EFFECTIVE_RANGE) ? row.EFFECTIVE_RANGE : requestedRange;
+
+    const year = getYearForRange(resolvedRange);
+    const periodLabel = resolvedRange === 'YTD' ? `YTD ${year}` : `FY ${year}`;
+
```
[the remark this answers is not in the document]
MRashad26: Fixed — `resolvedRange` is now typed as `HealthMetricsRange | null` (returns `null` when `isHealthMetricsRange` is false). `periodLabel` is only computed when `resolvedRange` is non-null; when it's null an empty string is returned. The `range` field in the response still falls back to `requestedRange` for interface compatibility.

See commit c9551c35.

apps/lfx-one/src/server/services/project.service.ts
MRashad26: Fixed as described — `resolvedRange` is now `null` when `EFFECTIVE_RANGE` is invalid, and `periodLabel` gates on that null before computing the year string. Thanks for the suggested code snippet; used the same pattern.

See commit c9551c35.

packages/shared/src/interfaces/dashboard-metric.interface.ts
MRashad26: Fixed — updated JSDoc to "derived from `range`" and added a note that the field is an empty string when no period has data. Good catch!

See commit c9551c35.

apps/lfx-one/src/server/services/project.service.ts
MRashad26: Fixed — renamed to `_` to signal the intentional skip.

See commit c9551c35.

MRashad26:

Commit: c9551c35d91b4da8c4ff0ad6b2914256a4584bcb

### Changes Made
- **project.service.ts**: Gate `periodLabel` on a valid `EFFECTIVE_RANGE` — `resolvedRange` is now `HealthMetricsRange | null`; when null (row exists but all periods have null scores), `periodLabel` returns `''` instead of a misleading period string. The `range` field falls back to `requestedRange` for interface compatibility (per @copilot-pull-request-reviewer, @dealako)
- **project.service.ts**: Rename unused `r` parameter in `buildCaseWhen` callback to `_` (per @dealako)
- **dashboard-metric.interface.ts**: Update `periodLabel` JSDoc — fix stale "derived from effectiveRange" reference to "derived from `range`"; document the empty-string case (per @dealako)

### Threads Resolved

dealako:
Hey Rashad! 👋 Quick turnaround across two rounds — well done. Walked through every prior finding and the latest two commits; everything looks clean.

---

## 👏 Nice work

- The `resolvedRange: HealthMetricsRange | null` typing is the right way to capture the null-EFFECTIVE_RANGE semantic in the type system — it makes the impossible state (a label with no resolved period) impossible to express accidentally.
- Handling the lint regression in its own focused `fix(nps-card): resolve no-nested-ternary` commit was the right call: it keeps the logic change and the style fix separately auditable.
- The `if (resolvedRange) { const year = ...; periodLabel = ... }` refactor is clean — `year` is computed once, both label branches are reachable, and the `no-nested-ternary` rule is satisfied.

---

## ✅ Resolved

All four items from my prior review are confirmed fixed at `02f85ede`:

| Finding | Resolution |
|---|---|
| **[minor]** `periodLabel` non-empty when `EFFECTIVE_RANGE` is null | `resolvedRange` is now `HealthMetricsRange \| null`; `periodLabel` defaults to `''` and is only assigned inside `if (resolvedRange)`. `range` still falls back to `requestedRange` for interface compatibility. ✅ |
| **[nit]** JSDoc references stale `effectiveRange` | Now reads `"derived from range (e.g. 'YTD 2026' or 'FY 2025'); empty string when no period has data"`. ✅ |
| **[nit]** Unused `r` param in `buildCaseWhen` callback | Renamed to `_`; `rangeCases` map correctly still uses `r` for the literal — no unintended change there. ✅ |
| **Round-3 lint regression** no-nested-ternary | `let periodLabel = ''` + `if/else` block. ✅ |

---

## ⚠️ Partially addressed

**[minor / follow-up]** `data-testid="nps-card-period-label"` still has no paired E2E spec. No new spec file was added in the latest commits. As I noted in round 1, this isn't a blocker — a follow-up ticket is fine — but please make sure the ticket gets filed before this branch is off your radar so it doesn't fall through the cracks.

---

## ❌ Still open

None.

---

## 🔍 New observations

None. The period-mixing fix is intact — all six metric columns (`NPS_SCORE`, `PROMOTERS`, `PASSIVES`, `DETRACTORS`, `NON_RESPONSES`, `CHANGE_NPS_SCORE`) still use the same `buildCaseWhen` guard ordering as `EFFECTIVE_RANGE`, so no regression there.

---

## Overall decision: ✅ Approved

The fix is architecturally sound and all blocking/minor items are addressed. Follow up on the E2E ticket and this is good to merge.

dealako (approved):
All round-2 findings resolved — see detailed comment at https://github.com/linuxfoundation/lfx-self-serve/pull/795#issuecomment-4559133716. One minor follow-up: please file a ticket for E2E coverage of `data-testid="nps-card-period-label"` before the branch goes cold. ✅

PR #810, opened 2026-05-28
fix(surveys): scope results-drawer aggregates to selected project

seberslinux:
## Summary

Survey results drawer was showing whole-campaign aggregates instead of per-project numbers when a user had a specific project selected.

## Root Cause

The Express proxy was resolving the project UUID to a v1 SFID via NATS before forwarding to upstream. When the NATS resolver returned `null` (mapping missing or timeout), the `project_id` filter was silently dropped — causing the upstream to return whole-campaign totals.

Additionally, `lfx-v2-survey-service` accepts `project_uid` (V2 UUID) natively, making the NATS UUID→SFID translation dead code.

## Fix

Drop the NATS SFID lookup. Forward the project UUID directly as `project_uid` to upstream — matching the confirmed OpenAPI contract (`GET /surveys/{uid}?project_uid=<uuid>`).

## Files Changed

- `apps/lfx-one/src/server/services/survey.service.ts` — removed NATS SFID resolution, forward `project_uid` directly

## Testing

Verified upstream contract via `gh api repos/linuxfoundation/lfx-v2-survey-service/contents/gen/http/openapi3.yaml` — `project_uid` (V2 UUID) is the documented query param.

Jira: [LFXV2-2007](https://linuxfoundation.atlassian.net/browse/LFXV2-2007)

---

[LFXV2-2007]: https://linuxfoundation.atlassian.net/browse/LFXV2-2007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/services/survey.service.ts
+++ b/apps/lfx-one/src/server/services/survey.service.ts
@@ -87,22 +87,19 @@ export class SurveyService {
   }
 
   /**
-   * Fetches a single survey by UID
-   * Resolves project UUID to v1 SFID via NATS before passing as project_id
+   * Fetches a single survey by UID.
+   * Passes project_uid (V2 UUID) directly to upstream — no SFID translation needed.
    */
   public async getSurveyById(req: Request, surveyUid: string, projectId?: string): Promise<Survey> {
     logger.debug(req, 'get_survey_by_id', 'Fetching survey by ID', {
       survey_uid: surveyUid,
-      project_id: projectId,
+      project_uid: projectId,
     });
 
     const params: Record<string, string> = {};
 
     if (projectId) {
-      const sfid = await this.projectService.getProjectSfidByUid(req, projectId);
-      if (sfid) {
-        params['project_id'] = sfid;
-      }
+      params['project_uid'] = projectId;
     }
 
     const survey = await this.microserviceProxy.proxyRequest<Survey>(req, 'LFX_V2_SERVICE', `/surveys/${surveyUid}`, 'GET', params);


audigregorie (commented):
## Code Review Summary

**Verdict:** Approve — the change correctly fixes the root cause. Forwarding the V2 `project_uid` directly matches the upstream `GET /surveys/{survey_uid}` contract and removes the fragile NATS SFID lookup that silently dropped the filter when resolution failed.

---

### Major — outside the diff

- **No regression test** — There are no server unit tests for `SurveyService.getSurveyById`, and no e2e covering project-scoped results in `survey-results-drawer`. Given this bug surfaced as silently wrong aggregate totals (worse than a hard error), a focused test that asserts `project_uid` is forwarded when `project_id` is present on `GET /api/surveys/:uid` would guard against reintroducing the NATS/SFID path.

### Minor — outside the diff

- **BFF query param naming** — The public BFF surface still accepts `?project_id=` (controller + frontend `SurveyService.getSurvey`) while the upstream param is `project_uid`. Values are V2 UUIDs throughout; consider a follow-up rename to `project_uid` on the BFF for consistency, or a one-line controller comment documenting the intentional alias.

---

### What's done well

- Minimal, surgical diff scoped to the actual failure mode.
- Upstream contract verified (`project_uid` on `GET /surveys/{survey_uid}` in `lfx-v2-survey-service` OpenAPI).
- Debug log field updated from `project_id` to `project_uid`, matching what is sent upstream.
- Removes an extra NATS round-trip from the hot path for survey detail fetches.

---

### Rating: ⭐⭐⭐⭐⭐ (5/5)

MRashad26 (approved):

PR #823, opened 2026-05-29
fix(settings): resolve username showing as email in permissions table

MRashad26:
## Summary

- **Frontend** (`user-form.component.ts`): sent `username: email` instead of `email: email` for directory lookups, routing the email value through the wrong request field
- **Controller** (`project.controller.ts`): `manualUserInfo` detection checked `name || email || avatar`; changed to check `name` only so a plain email lookup is not misclassified as a manual add
- **Service** (`project.service.ts`): overwrote `fetchedUserInfo.username` (real display username) with `backendIdentifier` (NATS sub); when the sub equalled the email the stored username became the email address

## Root cause detail

When adding a directory user, the frontend sent `{ username: "john@example.com", role }`. The service resolved the email to a NATS sub via `resolveEmailToSub`, then **overwrote** the user's display username from `getUserInfo` with that sub. For LFX auth configurations where the sub equals the email, `username = email` was persisted in project settings — which is what the permissions table displayed.

After edit+save, the cache was invalidated and a fresh fetch returned settled data, which is why the username appeared to "fix itself" after an edit.

## Test plan

- [ ] Add a new user to the permissions table by email — username column should show the user's real LFX username, not their email
- [ ] Verify the manual-add flow (user not found in directory) still works: "User Not Found" dialog still appears, and manually-entered name/username/avatar are saved correctly
- [ ] Edit an existing user's role — username should remain unchanged after save
- [ ] Remove a user — should still work correctly after the lookup identifier change

Fixes [LFXV2-2036](https://linuxfoundation.atlassian.net/browse/LFXV2-2036)

[LFXV2-2036]: https://linuxfoundation.atlassian.net/browse/LFXV2-2036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/settings/components/user-form/user-form.component.ts
+++ b/apps/lfx-one/src/app/modules/settings/components/user-form/user-form.component.ts
@@ -109,10 +109,10 @@ export class UserFormComponent {
 
     // For adding: if manual fields not shown yet, try to lookup user by email first
     if (!this.showManualFields()) {
-      // Try to add user with just email (backend will resolve to username)
+      // Try to add user by email — backend resolves to the real username via directory lookup
       this.permissionsService
         .addUserToProject(project.uid, {
-          username: formValue.email, // Pass email as username, backend will resolve
+          email: formValue.email,
           role: formValue.role,
         } as AddUserToProjectRequest)
         .pipe(take(1))
--- a/apps/lfx-one/src/server/controllers/project.controller.ts
+++ b/apps/lfx-one/src/server/controllers/project.controller.ts
@@ -260,11 +260,14 @@ export class ProjectController {
         return;
       }
 
-      // Check if manual user data is provided (for users not found in directory)
+      // Check if manual user data is provided (for users not found in directory).
+      // name is always present for manual adds and never present for directory email lookups,
+      // so it's the correct discriminator — using email alone would misclassify a plain
+      // { email, role } directory-lookup request as a manual add.
       let manualUserInfo: { name: string; email: string; username?: string; avatar?: string } | undefined;
-      if (userData.name || userData.email || userData.avatar) {
+      if (userData.name) {
         manualUserInfo = {
-          name: userData.name || '',
+          name: userData.name,
           email: userData.email || '',
           // username is optional for manually-added users — preserve whatever was provided
           username: userData.username || undefined,
--- a/apps/lfx-one/src/server/services/project.service.ts
+++ b/apps/lfx-one/src/server/services/project.service.ts
@@ -489,13 +489,11 @@ export class ProjectService {
         // incorrectly turn an intentionally empty username into an email address.
         userInfo = { ...existingUserInfo };
       } else {
-        // Fetch user info from user service via NATS using the original input
+        // Fetch user info from user service via NATS using the original input.
+        // Use the display username from the profile as-is — overwriting it with the NATS
+        // sub caused the username column to show the email when the sub equalled the email.
         const fetchedUserInfo = await this.getUserInfo(req, usernameOrEmail);
-        // Use the backend identifier (sub) for the username in the stored UserInfo for backend consistency
-        userInfo = {
-          ...fetchedUserInfo,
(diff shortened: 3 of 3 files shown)


dealako (approved):
## Code Review — PR #823

Hey Rashad 👋 — nice work digging out all three compounding causes instead of just patching the most visible one. The root-cause writeup in the PR description is a model of clarity; reviewers can follow the chain without opening the diff.

---

### 🛡️ Security Review (LFX Security Engineer)

**Phase 1 automated scan — all clear:**
- ✅ No hardcoded secrets or credentials
- ✅ No SQL/command/template injection patterns introduced
- ✅ No auth bypass or privilege-escalation patterns
- ✅ No PII in new log statements (logged fields: `identifier`, `is_manual` — consistent with existing patterns)
- ✅ No authorization regression — the permissions write path remains gated at the route middleware level (`req.oidc`)

**Phase 2 judgment review:**
- The change *narrows* what's persisted: real display username instead of NATS sub. This is strictly a correctness improvement with a positive security posture effect (display identity matches audit identity).
- No new external input sinks introduced; `userData.name` gate is consistent with existing field handling.

---

### ✅ Correctness — the data flow holds end-to-end

The three-part fix is internally consistent:

1. **Frontend** sends `{ email, role }` → controller identifier = email.
2. **Controller** `userData.name` is `undefined` → `manualUserInfo = undefined` → directory lookup path taken (not manual-add).
3. **Service** calls `getUserInfo(req, email)` → internally `resolveEmailToUsername` → returns real display username → stored as `fetchedUserInfo.username`. On a subsequent update/remove the frontend sends that stored username back (no `@` sign) so `originalEmail = ''`, `backendIdentifier = username`, and `matchesUser` matches on `u.username`. ✅ Write/read/update/remove identifier round-trips correctly.

Input validation (`identifier` + `role` required, line 240 — runs *before* the `manualUserInfo` discriminator at line 268) is unchanged and unaffected. ✅

The `name`-only discriminator is sound given the explicit contract in `AddUserToProjectRequest` — `name` is documented as "for manual entry when user not found" and `Validators.required` is only set on `name` in the not-found path. No non-manual flow sends `name`. ✅

---

### Findings

**`apps/lfx-one/src/server/services/project.service.ts:738` — [nit] Observation: residual surface in `resolveEmailToUsername`**

The fix correctly removes the overwrite of `fetchedUserInfo.username` with `backendIdentifier`. The display username now comes entirely from `getUserInfo` → `resolveEmailToUsername` (line 738). If the NATS `EMAIL_TO_USERNAME` subject itself returns the email or sub for some auth configuration, the symptom would still appear. This is a defence-in-depth observation, not a bug in this PR — the fix is correct given the documented contract of `resolveEmailToUsername`. Worth a follow-up NATS-side validation if similar reports resurface in other environments.

**`apps/lfx-one/e2e/` — [minor] No automated regression test for the repro scenario**

The PR has a solid manual test plan, but there's currently no E2E spec for the settings permissions flow at all. The exact bug ("add user by email → username column shows real username, not their email") is a perfect candidate for a regression test:
```
1. Add user by email via the permissions table
2. Assert the username column renders the LFX display username (not the email)
```
Without it, this bug is one refactor away from silently re-appearing. Not a blocker for merge — the feature didn't have coverage before — but worth tracking as a follow-up.

---

### AI comment reconciliation

- **GitHub Copilot**: reviewed all 3 files, generated no inline comments, and summarized the change accurately. Agreement — the change is clean.
- **CodeRabbit**: review was still processing at time of submission. No findings to reconcile; will update if CodeRabbit posts anything substantive after merge.

---

### Revision tracking

No prior review rounds. N/A.

---

### Summary

**0 blocking issues · 1 minor (missing regression test) · 1 nit (residual-surface observation)**

The fix is minimal, correct, and well-scoped. The three-file change cleanly addresses all three causes without any collateral changes. Security posture is unchanged-to-improved. The only follow-up item worth tracking is an E2E regression test for the add-by-email username display.

**Decision: ✅ Approved**

PR #831, opened 2026-05-31
feat(auth): evict users from write routes on context switch

MRashad26:
## Summary

- `writerGuard` is a `CanActivateFn` and does not re-run when the project context selector uses `Location.replaceState()` for mid-session context switches
- Users on a create/edit route who switch to a project without write access would stay on the page indefinitely
- This PR adds a reactive `canWrite` subscriber to each write-route manage component that redirects to `/foundation/overview` as soon as write access is lost mid-session

## Implementation

In each of the five manage components (`committee-manage`, `mailing-list-manage`, `meeting-manage`, `survey-manage`, `vote-manage`), the constructor now subscribes to `ProjectContextService.canWrite`:

```typescript
toObservable(this.projectContextService.canWrite)
  .pipe(
    skip(1),                         // skip initialValue: false before project loads
    filter((canWrite) => !canWrite), // only react when write access is lost
    take(1),                         // redirect once then stop
  )
  .subscribe(() => this.router.navigateByUrl('/foundation/overview'));
```

`skip(1)` is critical — `canWrite` initializes as `false` before the project finishes loading, so without it the component would redirect immediately on mount.

## Test scenario

1. Log in as a user who is a writer on project A but not project B
2. Navigate to project A → open any create/edit route (e.g. `/meetings/create`)
3. Switch the project context to project B via the selector
4. **Expected:** Redirected to `/foundation/overview`
5. **Expected (negative):** Switching from project A to another project where you also have write access keeps you on the page

## Related

- Parent epic: [LFXV2-1654](https://linuxfoundation.atlassian.net/browse/LFXV2-1654) — Permission, Persona & Navigation Model
- Ticket: [LFXV2-2045](https://linuxfoundation.atlassian.net/browse/LFXV2-2045)
- Guard fix PR: #829 (`writerGuard` rewrite + applied to all write routes)

[LFXV2-1654]: https://linuxfoundation.atlassian.net/browse/LFXV2-1654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: MIT
 
 import { HttpErrorResponse } from '@angular/common/http';
-import { Component, computed, inject, signal } from '@angular/core';
-import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { Component, computed, DestroyRef, inject, signal } from '@angular/core';
+import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
 import { ActivatedRoute, Router, RouterLink } from '@angular/router';
 import { ButtonComponent } from '@components/button/button.component';
@@ -14,7 +14,7 @@ import { ProjectContextService } from '@services/project-context.service';
 import { MessageService } from 'primeng/api';
 import { ConfirmDialogModule } from 'primeng/confirmdialog';
 import { StepperModule } from 'primeng/stepper';
-import { BehaviorSubject, catchError, concat, filter, finalize, forkJoin, Observable, of, switchMap, take, toArray } from 'rxjs';
+import { BehaviorSubject, catchError, concat, filter, finalize, forkJoin, Observable, of, skip, switchMap, take, toArray } from 'rxjs';
 import { getHttpErrorDetail } from '@shared/utils/http-error.utils';
 
 import { CommitteeBasicInfoComponent } from '../components/committee-basic-info/committee-basic-info.component';
@@ -44,6 +44,7 @@ export class CommitteeManageComponent {
   private readonly committeeService = inject(CommitteeService);
   private readonly messageService = inject(MessageService);
   private readonly projectContextService = inject(ProjectContextService);
+  private readonly destroyRef = inject(DestroyRef);
   // Mode and state signals
   public mode = signal<'create' | 'edit'>('create');
   public committeeId = signal<string | null>(null);
@@ -89,6 +90,14 @@ export class CommitteeManageComponent {
   public readonly committeeLabelPlural = COMMITTEE_LABEL.plural;
 
   public constructor() {
+    toObservable(this.projectContextService.canWrite)
+      .pipe(
+        skip(1),
+        filter((canWrite) => !canWrite),
+        take(1),
+        takeUntilDestroyed(this.destroyRef)
+      )
+      .subscribe(() => this.router.navigateByUrl('/foundation/overview'));
     // Initialize step based on mode
     this.currentStep = toSignal(
       this.route.queryParamMap.pipe(
--- a/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: MIT
 
 import { HttpErrorResponse } from '@angular/common/http';
-import { Component, computed, DestroyRef, inject, signal } from '@angular/core';
-import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { Component, computed, inject, signal } from '@angular/core';
+import { toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
 import { ActivatedRoute, Router, RouterLink } from '@angular/router';
 import { ButtonComponent } from '@components/button/button.component';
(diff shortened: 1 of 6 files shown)


apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
```diff
@@ -89,6 +90,14 @@ export class CommitteeManageComponent {
...
 
   public constructor() {
+    toObservable(this.projectContextService.canWrite)
+      .pipe(
+        skip(1),
+        filter((canWrite) => !canWrite),
+        take(1),
+        takeUntilDestroyed(this.destroyRef)
+      )
+      .subscribe(() => this.router.navigateByUrl('/foundation/overview'));
```
audigregorie: **Major — Duplicated eviction subscription across 5 manage components**

This exact 8-line block (`toObservable(canWrite)` → `skip(1)` / `filter(!canWrite)` / `take(1)` / `takeUntilDestroyed` → `navigateByUrl('/foundation/overview')`) is copy-pasted verbatim into `committee-manage`, `mailing-list-manage`, `meeting-manage`, `survey-manage`, and `vote-manage`. CLAUDE.md calls for shared utilities and centralized constants; five identical copies will drift. Extract to a single injection-context helper so the logic — and the `/foundation/overview` literal — lives in one place:

```ts
// apps/lfx-one/src/app/shared/utils/evict-on-write-access-loss.ts
export function evictOnWriteAccessLoss(): void {
  const router = inject(Router);
  const projectContext = inject(ProjectContextService);
  const destroyRef = inject(DestroyRef);
  toObservable(projectContext.canWrite)
    .pipe(skip(1), filter((canWrite) => !canWrite), take(1), takeUntilDestroyed(destroyRef))
    .subscribe(() => router.navigateByUrl('/foundation/overview'));
}
```

Each constructor then becomes a single `evictOnWriteAccessLoss();` call, and the duplicated `skip` rxjs imports + repeated route literal disappear. (One issue → one thread; this applies to all five files.)

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
```diff
@@ -89,6 +90,14 @@ export class CommitteeManageComponent {
   public readonly committeeLabelPlural = COMMITTEE_LABEL.plural;
 
   public constructor() {
+    toObservable(this.projectContextService.canWrite)
+      .pipe(
+        skip(1),
```
audigregorie: **Minor — `skip(1)` assumes exactly one stale emission**

This works today: `canWrite` is a boolean with `initialValue: false`, and signal dedup collapses consecutive repeats, so the first emission is the stale pre-load `false`. But `skip(1)` also unconditionally suppresses a *genuine* mount-time `false` — e.g. if `activeContext` resolves to a context where the user lacks write while `writerGuard` passed on the route slug (the guard checks the slug, `canWrite` follows lens/persona-derived `activeContext`). The guard makes that unlikely now, but the invariant is implicit. Consider keying off `activeContext` changes (skip the initial context, react to subsequent ones) or adding a short comment pinning why `skip(1)` is safe so a future `canWrite` change doesn't silently break eviction. Applies to all five components.

audigregorie (commented):
## Code Review Summary

Well-scoped fix that closes the real gap left by `writerGuard`: as a `CanActivateFn` it doesn't re-run when the project-context selector switches context via `Location.replaceState()` (`project-context.service.ts:85-96`). The reactive `canWrite` subscriber is the right mechanism, the redirect target `/foundation/overview` matches the sibling `newsletter-access.guard.ts`, and imports/injections are correct in all five components. Two notes below: one maintainability (the block is duplicated 5×) and one robustness note on `skip(1)`.

### Nits (outside the diff)
- The rest of the app redirects via `router.navigate(['/foundation/overview'])` (e.g. `committee-view.component.ts:369`, `multi-persona-dashboard.component.ts:118`); this PR uses `navigateByUrl('/foundation/overview')`. Both work — minor stylistic inconsistency, not worth a thread.

### What's done well
- Correctly diagnoses why the guard alone is insufficient and pairs it with the right reactive primitive.
- `skip(1)` + `filter(!canWrite)` + `take(1)` + `takeUntilDestroyed` is the correct chain for "fire once on access loss, then clean up."
- PR description documents both the positive and the negative test scenarios.

**Rating: 4/5**

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
MRashad26: Done — extracted to `evictOnWriteAccessLoss()` in `apps/lfx-one/src/app/shared/utils/evict-on-write-access-loss.util.ts`. The `/foundation/overview` literal and the full RxJS chain now live in one place; each of the five constructors is reduced to a single `evictOnWriteAccessLoss()` call. Removed the now-unused `DestroyRef`/`takeUntilDestroyed`/`skip` imports from all five components.

See commit `199180af`: extract eviction logic into shared injection-context helper.

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
MRashad26: Good call. The JSDoc on the new `evictOnWriteAccessLoss()` helper now pins the invariant explicitly:

> `skip(1)` is safe here because `canWrite` has `initialValue: false` and Angular signal dedup collapses consecutive identical values, so the first emission is always the pre-load false — not a genuine access-lost signal.

This lives in one place rather than as an inline comment that would need to be repeated across all five components.

See commit `199180af`.

MRashad26:
## Review Feedback Addressed

Commit: `199180af`

### Changes Made
- **shared/utils/evict-on-write-access-loss.util.ts** (new): Extracted the 8-line eviction block into a single injection-context helper. The `/foundation/overview` literal and the full `toObservable` → `skip(1)` → `filter` → `take(1)` → `takeUntilDestroyed` chain live in one place. JSDoc documents the `skip(1)` invariant explicitly. (per @audigregorie — major + minor findings)
- **committee-manage, mailing-list-manage, meeting-manage, survey-manage, vote-manage**: Each constructor is now a single `evictOnWriteAccessLoss()` call. Removed the now-unused `DestroyRef`, `takeUntilDestroyed`, and `skip` imports from all five components.

### Threads Resolved

audigregorie (approved):

PR #832, opened 2026-05-31
fix(auth): preserve project context in eviction redirect

MRashad26:
## Summary

- `evictOnWriteAccessLoss` was navigating to bare `/foundation/overview` on context-switch eviction, causing the foundation lens to render the last-remembered project (e.g. `aswf`) instead of the one that triggered the access loss (e.g. `agstack`)
- Fix: read `projectContextService.activeContext()?.slug` at eviction time and pass it as `?project=<slug>` via `createUrlTree` — same pattern used in `writerGuard` and `newsletterAccessGuard`

## Root cause

When `canWrite` emits `false` after a context switch, `activeContext()` already reflects the new context (the one the user switched to). Reading the slug at that moment and including it in the redirect preserves which project the user was trying to access when they were evicted.

## Test plan

- [ ] As a writer on project A, navigate to any create/edit route (e.g. `/foundation/meetings/create?project=<A>`)
- [ ] Switch context to project B (where you are not a writer)
- [ ] Should be redirected to `/foundation/overview?project=<B>` — not project A's overview

## Context

Followup bug fix for LFXV2-2045 (#831). The eviction helper was merged without the `?project=` param in the redirect URL.

### DIFF
--- a/apps/lfx-one/src/app/shared/utils/evict-on-write-access-loss.util.ts
+++ b/apps/lfx-one/src/app/shared/utils/evict-on-write-access-loss.util.ts
@@ -34,5 +34,9 @@ export function evictOnWriteAccessLoss(): void {
       take(1),
       takeUntilDestroyed(destroyRef)
     )
-    .subscribe(() => router.navigateByUrl('/foundation/overview'));
+    .subscribe(() => {
+      const slug = projectContextService.activeContext()?.slug;
+      const url = slug ? router.createUrlTree(['/foundation/overview'], { queryParams: { project: slug } }) : router.parseUrl('/foundation/overview');
+      router.navigateByUrl(url);
+    });
 }


audigregorie (commented):
## Code Review Summary

Tight, well-reasoned follow-up fix. The timing assumption is correct — `canWrite` is derived from `activeContext` via `switchMap`, so by the time a `false` reaches the subscriber, `activeContext()` already reflects the switched-to context. The slug read and `?project=` preservation are sound, and the null-context path correctly degrades to the bare `/foundation/overview`. No Critical or Major issues. **Rating: 5/5.**

### Minor — outside the diff

- **`evict-on-write-access-loss.util.ts:11-24` (JSDoc)** — The doc comment still says it "redirects to `/foundation/overview` if write access is lost mid-session" and doesn't mention the new `?project=<slug>` preservation this PR adds. Worth a one-line update so the doc matches the behavior:
  > This function subscribes to canWrite and redirects to `/foundation/overview?project=<active-slug>` (preserving the context that triggered the access loss) if write access is lost mid-session.

- **PR description nuance (meta)** — The body says the `?project=` redirect uses "the same pattern used in `writerGuard` and `newsletterAccessGuard`." Both guards redirect to **bare** `/foundation/overview` on denial — neither appends `?project=`. The `createUrlTree` + `queryParams` approach here is correct; the description just slightly overstates the precedent.

### What's done well

- Correct timing assumption, including the re-switch (switchMap cancel/restart) case — the redirect always reflects the current active context.
- Safe fallback to bare `/foundation/overview` when `activeContext()?.slug` is absent.
- Single, non-nested ternary; within the 160-char printWidth (150 chars). CLAUDE.md-compliant.

No line-level threads — both findings sit outside the changed lines.

audigregorie (approved):

PR #845, opened 2026-06-01
feat(org-lens): org Lens Access tab with add/edit/remove (LFXV2-1908)

luismoriguerra:
## Summary

Adds the **Org Lens Access** tab (last tab on the People page) where an org's elevated-access users are listed and managed.

- **List**: Name (avatar + job title) / Type (Admin/Viewer badge + tooltip) / Actions, with Total / Administrators / Viewers summary cards.
- **Search** across name, job title, email; single-select **Type filter** (`All types / Org Admin - Editor / Org Admin - Viewer / Invited`); client-side pagination ("Show all N users", cap 30).
- **Add Users** (invite), **Edit role** (Admin⇄Viewer modal), **Remove** (confirm dialog) — manager-only; pending/Invited rows can't be edited; last-Admin is protected.

Writes go to the member-service **per-principal endpoints** (`POST/PUT/DELETE /b2b_orgs/{uid}/settings/users[/{email}]`) via a new `LFX_V2_MEMBER_SERVICE` proxy base (defaults to `LFX_V2_SERVICE`). This replaces the previous full-replace read-modify-write, which could reset untouched members to `pending`. After each write the list refreshes authoritatively; last-Admin/duplicate are enforced upstream and mapped through the existing access-error helper.

## Dependency / rollout

Requires the member-service per-principal endpoints — see lfx-v2-member-service#49. **Deploy member-service to the target env before this UI**, or access-tab writes will 404.

## Test plan

- [x] `@lfx-one/shared` builds; ESLint clean on changed files
- [x] Full local E2E (UI -> BFF -> local member-service, Toyota seeded from dev): list/summary, search, filter, edit (role + counts update), remove (confirm + counts), add (Invited row, admins preserved), gating (pending Edit disabled)
- [ ] CI: lint + unit + e2e
- [ ] Reviewer: confirm proxy-base split and error mapping

Resolves https://linuxfoundation.atlassian.net/browse/LFXV2-1908


apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/edit-access-role-modal/edit-access-role-modal.component.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  protected onSave(): void {
+    if (!this.canSave()) return;
+    this.save.emit(this.selectedRole());
+  }
+
+  protected onCancel(): void {
+    if (this.saving()) return;
+    this.visible.set(false);
+  }
```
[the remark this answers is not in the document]
luismoriguerra: Done — `onCancel` now re-seeds `selectedRole` to the principal's current role before closing, so a discarded selection can't survive a same-user re-open (the `linkedSignal` doesn't re-evaluate when `user()` keeps the same reference). See commit 13288bbd.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
```diff
@@ -0,0 +1,353 @@
...
+    const orgUid = this.accountContext.selectedAccount().uid;
+    if (!orgUid) return;
+
+    this.isInviting.set(true);
+    this.dataService
+      .inviteUser(orgUid, value.email, value.role, value.name)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (res) => {
+          this.listData.set(res);
```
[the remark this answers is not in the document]
luismoriguerra: Done — the invite handler now drops the response when the selected org changed mid-flight (`if (res.orgUid !== orgUid) return;` before `listData.set`), so stale data can't overwrite the newly selected org's list. The in-flight flag is still cleared either way. See commit 13288bbd.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
```diff
@@ -0,0 +1,353 @@
...
+    const orgUid = this.accountContext.selectedAccount().uid;
+    if (!user || !orgUid) return;
+
+    this.isSubmitting.set(true);
+    this.dataService
+      .changeRole(orgUid, user.email, role)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (res) => {
+          this.listData.set(res);
```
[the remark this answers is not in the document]
luismoriguerra: Done — same stale-org guard applied to the change-role handler (`if (res.orgUid !== orgUid) return;` before `listData.set`). See commit 13288bbd.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
```diff
@@ -0,0 +1,353 @@
...
+    const orgUid = this.accountContext.selectedAccount().uid;
+    if (!orgUid) return;
+
+    this.isSubmitting.set(true);
+    this.dataService
+      .removeUser(orgUid, user.email)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (res) => {
+          this.listData.set(res);
```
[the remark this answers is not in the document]
luismoriguerra: Done — same stale-org guard applied to the remove handler (`if (res.orgUid !== orgUid) return;` before `listData.set`). See commit 13288bbd.

packages/shared/src/interfaces/org-lens-access.interface.ts
```diff
@@ -0,0 +1,113 @@
...
+  initials: string;
+  /** From settings `avatar` when present, else `null`. */
+  avatarUrl: string | null;
+  /** Best-effort enrichment by email; `null` when unknown. */
+  jobTitle: string | null;
+  /** `admin` (writer) or `viewer` (auditor). */
+  role: OrgAccessRole;
+  /** `accepted` or `pending` for listed rows. */
+  inviteStatus: OrgAccessInviteStatus;
+  /** `true` when `inviteStatus !== 'accepted'`. Drives the "Invited" filter and hides Edit. */
```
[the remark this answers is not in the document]
luismoriguerra: Done — added `OrgAccessListedInviteStatus = Extract<OrgAccessInviteStatus, 'pending' | 'accepted'>` and narrowed `OrgAccessUser.inviteStatus` to it. `mapPrincipals` already `continue`s on revoked/expired, so the assignment type-narrows cleanly and downstream UI no longer has to handle impossible states. See commit 13288bbd.

apps/lfx-one/src/server/services/org-lens-access.service.ts
```diff
@@ -0,0 +1,213 @@
...
+    return { totalUsers: users.length, administrators, viewers };
+  }
+
+  /** Best-effort job-title enrichment by email from the org people/key_contact index (D-006). */
+  private async enrichJobTitles(req: Request, orgUid: string, users: OrgAccessUser[]): Promise<OrgAccessUser[]> {
+    if (users.length === 0) return users;
+    let titleByEmail = new Map<string, string>();
+    try {
+      const employees = await this.keyContacts.getEmployees(req, orgUid);
+      titleByEmail = new Map(employees.filter((e) => e.email && e.jobTitle).map((e) => [e.email.toLowerCase(), e.jobTitle as string]));
```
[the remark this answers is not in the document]
luismoriguerra: Good call — flagging this as a known follow-up rather than fixing it here. A couple of constraints:

- `enrichJobTitles` runs **once per list load** (not per row), and it's strictly best-effort — failures are swallowed and the list still renders.
- The `key_contact` read in `getEmployeesByOrgUid` goes through the query-service `/query/resources` contract, which filters by `tags` only — there's no "filter by a set of emails" parameter available today, so a targeted lookup would need a member-service/query-service contract change.
- A per-org TTL cache would add shared mutable state + invalidation concerns to the BFF that feel out of scope for this PR (writes mutate access membership, not key_contact job titles, so cache coherence is non-trivial).

Leaving this open so it's tracked. Happy to spin up a follow-up ticket for either the targeted-by-email contract or a short-TTL employee/title cache if you'd prefer.

luismoriguerra:
## Review Feedback Addressed

Commit: `13288bbd`

### Changes Made
- **packages/shared/src/interfaces/org-lens-access.interface.ts**: added `OrgAccessListedInviteStatus = Extract<OrgAccessInviteStatus, 'pending' | 'accepted'>` and narrowed `OrgAccessUser.inviteStatus` to it — listed rows can only be pending/accepted (revoked/expired are filtered in `mapPrincipals`) (per copilot[bot])
- **edit-access-role-modal.component.ts**: `onCancel` now re-seeds `selectedRole` to the current role so a discarded selection can't survive a same-user re-open (the `linkedSignal` doesn't re-evaluate on an unchanged `user()` reference) (per copilot[bot])
- **org-lens-access.component.ts**: invite / change-role / remove handlers now drop the response when the selected org changed mid-flight (`res.orgUid !== orgUid`), preventing stale data from overwriting the new org's list; the in-flight flag is cleared regardless (per copilot[bot])

### Threads Resolved
5 of 6 unresolved threads addressed and resolved in this iteration.

### Still Open
- **org-lens-access.service.ts:149-155** — `enrichJobTitles` pulling the full org `key_contact` set. Left open as a tracked follow-up: it runs once per list load and is best-effort, the query-service `/query/resources` contract has no filter-by-email today, and a TTL cache adds invalidation concerns out of scope for this PR. Happy to file a follow-up ticket. (flagged by copilot[bot])

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

apps/lfx-one/e2e/org-lens-access.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+    await expect(page.getByTestId('org-access-row-pat.pending@example.com')).toContainText('Invited');
+  });
+
+  test('S3: type filter + search narrow the rows (US4 / FR-007a)', async ({ page }) => {
+    await gotoAccessTab(page);
+
+    // Invited filter shows only the pending row.
+    await page.getByTestId('org-access-type-filter').click();
+    await page.getByText('Invited', { exact: true }).click();
+    await expect(page.getByTestId('org-access-row-pat.pending@example.com')).toBeVisible();
```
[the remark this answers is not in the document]
luismoriguerra: Done — switched all four p-select option clicks to `getByRole('option', { name, exact: true })` (Invited / All types / Admin x2) so they target the opened dropdown overlay instead of matching duplicate labels like the legend's "Admin". See commit 3f9a3b4c. Also added invite (POST) coverage as you flagged in the review nitpicks — S7 happy path (asserts the POST payload + invited row + success toast) and S7b 5xx-failure (error toast, modal stays open).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.html
```diff
@@ -0,0 +1,252 @@
...
+                        type="button"
+                        class="inline-flex h-8 w-8 items-center justify-center rounded-md text-gray-500 transition-colors hover:bg-red-50 hover:text-red-600 disabled:cursor-not-allowed disabled:opacity-40"
+                        [disabled]="user.isLastAdmin || isSubmitting()"
+                        [pTooltip]="user.removeTooltip"
+                        tooltipPosition="top"
+                        [attr.aria-label]="'Remove access for ' + user.name"
+                        [attr.data-testid]="'org-access-remove-' + user.email"
+                        (click)="confirmRemove(user)">
+                        <i class="fa-light fa-trash" aria-hidden="true"></i>
+                      </button>
```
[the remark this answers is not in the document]
luismoriguerra: Done — wrapped both disabled Edit/Remove buttons in a non-disabled `<span class="inline-flex" [pTooltip] tooltipPosition="top">`, keeping `[disabled]`, aria-label, data-testid and (click) on the inner button. Tooltips now fire for the pending-Edit and last-Admin-Remove disabled states. See commit 3f9a3b4c.

luismoriguerra:
## Review Feedback Addressed — Round 2 (CodeRabbit)

Commit: `3f9a3b4c`

### Changes Made
- **org-lens-access.component.html**: wrapped the disabled Edit/Remove buttons in a non-disabled `<span [pTooltip] tooltipPosition="top">` so PrimeNG tooltips fire on the disabled states (native `[disabled]` buttons suppress the pointer events tooltips rely on); `[disabled]`, aria-label, data-testid and `(click)` stay on the inner button (per coderabbitai[bot])
- **e2e/org-lens-access.spec.ts**: replaced the four `getByText(...)` p-select option clicks with `getByRole('option', { name, exact: true })` to avoid matching duplicate labels like the legend's "Admin" (per coderabbitai[bot])

### Nitpicks Addressed
- **e2e/org-lens-access.spec.ts**: added invite/POST coverage — **S7** happy path (asserts POST payload + invited row + success toast) and **S7b** 5xx-failure (error toast, modal stays open) (per coderabbitai[bot])
- **app/shared/services/org-lens-access.service.ts**: added JSDoc to the four public methods per repo guidelines (per coderabbitai[bot])

### Threads Resolved
2 of 3 unresolved threads resolved this round.

### Still Open
- **org-lens-access.service.ts:155** — `enrichJobTitles` pulling the full org `key_contact` set (round-1 thread). Left open as a tracked follow-up: best-effort, runs once per load, no filter-by-email contract exists today, and a TTL cache is out of scope. Happy to file a follow-up ticket. (flagged by copilot[bot])

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

apps/lfx-one/src/app/shared/services/feature-flag.service.ts
```diff
@@ -38,6 +42,13 @@ export class FeatureFlagService {
...
       this.isInitialized.set(true);
 
+      // The provider may already be READY (the app initializer awaits
+      // setProviderAndWait before bootstrap), so seed the signal eagerly and
+      // let the Ready handler cover the slower streaming case.
+      if (this.client.providerStatus === ProviderStatus.READY) {
+        this.isProviderReady.set(true);
+      }
+
       this.setupEventHandlers();
```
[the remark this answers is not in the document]
luismoriguerra: Done — `setupEventHandlers()` now runs **before** the `providerStatus === READY` seed, so a READY transition landing in the old gap can't be missed: the `ProviderEvents.Ready` handler covers the late/streaming case and the status seed covers the already-READY case (setting the signal twice is idempotent). See commit ccb696bd.

luismoriguerra:
## Review Feedback Addressed — Round 3 (Copilot)

Commit: `ccb696bd`

### Changes Made
- **feature-flag.service.ts**: register the OpenFeature event handlers **before** seeding `isProviderReady` from the current `providerStatus`. This closes the race where a READY transition landing between the status check and handler registration could be missed, leaving `providerReady` false indefinitely (guards fail closed / time out). The `Ready` handler covers the late/streaming case, the status seed covers the already-READY case, and setting the signal twice is idempotent. (per copilot[bot])

### Threads Resolved
1 of 2 unresolved threads resolved this round.

### Still Open
- **org-lens-access.service.ts:155** — `enrichJobTitles` pulling the full org `key_contact` set (round-1 thread). Left open as a tracked follow-up: best-effort, runs once per load, no filter-by-email contract exists today, and a TTL cache is out of scope. Happy to file a follow-up ticket. (flagged by copilot[bot])

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.html
```diff
@@ -0,0 +1,253 @@
...
+      subtitle="Please try again."
+      ctaLabel="Retry"
+      ctaIcon="fa-light fa-arrow-rotate-left"
+      data-testid="org-access-error"
+      (ctaClick)="retry()" />
+  } @else {
+    <!-- Summary cards -->
+    <div class="grid grid-cols-1 gap-4 sm:grid-cols-3" data-testid="org-access-summary">
+      @if (isLoading()) {
+        @for (i of [0, 1, 2]; track i) {
```
audigregorie: **Minor — inline array literal in `@for` recreated every change-detection pass**

The inline `[0, 1, 2]` is re-allocated on every CD cycle and is inconsistent with the `tableSkeletonRows` readonly field used for the table skeleton 26 lines below. Extract a sibling field (e.g. `summarySkeletonCards: readonly number[] = [0, 1, 2];`) and reference it here for consistency and a small perf win.

```suggestion
        @for (i of summarySkeletonCards; track i) {
```

packages/shared/src/interfaces/org-lens-access.interface.ts
```diff
@@ -0,0 +1,116 @@
...
+  email: string;
+  name?: string | null;
+  /** Absent for pending invites. */
+  username?: string | null;
+  /** Required. */
+  invited_as: 'writer' | 'auditor';
+  invite_status?: OrgAccessInviteStatus;
+}
+
+/** GET/PUT /b2b_orgs/{uid}/settings body (full-replace semantics on PUT). */
```
audigregorie: **Minor — stale comment**

This flow intentionally moved off the full-replace PUT on `/settings` to the per-principal `POST/PUT/DELETE /b2b_orgs/{uid}/settings/users[/{email}]` endpoints (per the service header comment). The `(full-replace semantics on PUT)` note now describes behavior this code no longer uses, and `GET` is the only verb this shape is read through here.

```suggestion
/** GET /b2b_orgs/{uid}/settings body. Writes use the per-principal endpoints, not a full-replace PUT. */
```

apps/lfx-one/src/server/services/microservice-proxy.service.ts
```diff
@@ -25,6 +25,7 @@ export class MicroserviceProxyService {
     try {
       const MICROSERVICE_URLS: MicroserviceUrls = {
         LFX_V2_SERVICE: process.env['LFX_V2_SERVICE'] || 'http://lfx-api.k8s.orb.local',
+        LFX_V2_MEMBER_SERVICE: process.env['LFX_V2_MEMBER_SERVICE'] || process.env['LFX_V2_SERVICE'] || 'http://lfx-api.k8s.orb.local',
```
audigregorie: **Minor — duplicated env-resolution map across all 4 methods**

This `MICROSERVICE_URLS` literal is now repeated verbatim at lines 28, 66, 113, and 156. It's a pre-existing pattern (so not blocking, and out of strict scope here), but this PR doubles the per-copy maintenance by adding the `LFX_V2_MEMBER_SERVICE` line to each. Consider extracting a private helper so the env-var resolution lives in one place:

```ts
private resolveBaseUrl(service: keyof MicroserviceUrls): string {
  const urls: MicroserviceUrls = {
    LFX_V2_SERVICE: process.env['LFX_V2_SERVICE'] || 'http://lfx-api.k8s.orb.local',
    LFX_V2_MEMBER_SERVICE: process.env['LFX_V2_MEMBER_SERVICE'] || process.env['LFX_V2_SERVICE'] || 'http://lfx-api.k8s.orb.local',
  };
  return urls[service];
}
```

audigregorie (commented):
## Code Review Summary

**Verdict: ⭐⭐⭐⭐⭐ — ship it.** Clean, well-engineered full-stack feature that has already been through three review-fix rounds. No Critical or Major findings. Verified race-condition handling (switchMap load + stale-drop guard on writes), auth gating (defense-in-depth `assertCanManage` 403 + upstream enforcement), payload field survival UI→BFF→member-service, subscription cleanup, and shared-type propagation (all 4 `MicroserviceUrls` literals updated, no orphan consumers). The per-principal write migration correctly fixes the prior full-replace bug that could reset untouched members to `pending`.

Only three minor polish items below, each posted as its own thread.

### What's done well

- **Method-free template**: per-row derivatives baked once in `toRowVm` (no method calls in bindings).
- **Error mapping** (`access-error.helper.ts`): collapses upstream 5xx to a generic message so server noise can't leak, and rejects raw-noise messages via regex.
- **Solid E2E coverage** mirroring the spec's user stories (list/summary/filter/edit/remove/add/failure/non-manager).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
```diff
@@ -0,0 +1,359 @@
...
+
+    this.isInviting.set(true);
+    this.dataService
+      .inviteUser(orgUid, value.email, value.role, value.name)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (res) => {
+          this.isInviting.set(false);
+          // Drop the response if the selected org changed mid-flight — don't overwrite the new org's list.
+          if (res.orgUid !== orgUid) return;
```
audigregorie: **Major — mid-flight org guard is a no-op (onInvite)**

`res.orgUid` is echoed by the BFF from the request's own route param, so it always equals the locally-captured `orgUid`. This comparison can never be true, so the guard never fires. If the user switches orgs while the invite is in flight, the late response overwrites the newly-selected org's list with the old org's data. Compare against the *live* selection at response time instead:

```suggestion
          if (res.orgUid !== this.accountContext.selectedAccount().uid) return;
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
```diff
@@ -0,0 +1,359 @@
...
+
+    this.isSubmitting.set(true);
+    this.dataService
+      .changeRole(orgUid, user.email, role)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (res) => {
+          this.isSubmitting.set(false);
+          // Drop the response if the selected org changed mid-flight — don't overwrite the new org's list.
+          if (res.orgUid !== orgUid) return;
```
audigregorie: **Major — mid-flight org guard is a no-op (onEditSave)**

Same issue as in `onInvite`: `res.orgUid` always equals the captured `orgUid`, so a stale edit response can clobber a freshly-selected org's list. Compare against the current selection:

```suggestion
          if (res.orgUid !== this.accountContext.selectedAccount().uid) return;
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
```diff
@@ -0,0 +1,359 @@
...
+
+    this.isSubmitting.set(true);
+    this.dataService
+      .removeUser(orgUid, user.email)
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: (res) => {
+          this.isSubmitting.set(false);
+          // Drop the response if the selected org changed mid-flight — don't overwrite the new org's list.
+          if (res.orgUid !== orgUid) return;
```
audigregorie: **Major — mid-flight org guard is a no-op (removeUser)**

Same issue as in `onInvite`/`onEditSave`: the captured `orgUid` always matches `res.orgUid`. Compare against the live selection so a late remove response can't overwrite a newly-selected org's list:

```suggestion
          if (res.orgUid !== this.accountContext.selectedAccount().uid) return;
```

apps/lfx-one/src/server/services/org-lens-access.service.ts
```diff
@@ -0,0 +1,213 @@
...
+  public async listAccessUsers(req: Request, orgUid: string): Promise<OrgAccessListResponse> {
+    const [settings, canManage] = await Promise.all([this.fetchSettings(req, orgUid), this.resolveCanManage(req, orgUid)]);
+    const users = await this.enrichJobTitles(req, orgUid, this.mapPrincipals(settings));
+    return { orgUid, users, summary: this.buildSummary(users), canManage };
+  }
+
+  /** Add Users — invite a NEW principal via the per-principal POST endpoint; returns the refreshed list. */
+  public async inviteUser(req: Request, orgUid: string, email: string, role: OrgAccessRole, name?: string | null): Promise<OrgAccessListResponse> {
+    // Defense-in-depth UX gate (FR-011): reject non-managers before any write is issued.
+    await this.assertCanManage(req, orgUid, 'invite_org_access_user');
```
audigregorie, line 62: **Minor — `canManage` resolved twice per write**

`assertCanManage` resolves the role-grants here, then the `listAccessUsers` refresh on the next line resolves it again (`Promise.all` at line 45). Each invite/edit/remove makes two role-grants lookups. Consider passing the already-resolved flag through to the refresh, or having the write path reuse the value. Same pattern in `changeRole` and `removeUser`.

apps/lfx-one/src/server/helpers/access-error.helper.ts
```diff
@@ -0,0 +1,63 @@
...
+    case 400:
+      return { status: 400, message: cleanMessage(rawMessage, 'The request was invalid. Check the details and try again.'), conflict: false };
+    // Explicit upstream 5xx range: never echo the raw upstream message — collapse to the generic
+    // fallback so server-internal noise can't leak to the client.
+    case 500:
+    case 502:
+    case 503:
+    case 504:
+      return { status: 502, message: AccessErrorMapping.fallbackMessage, conflict: false };
+    default:
```
audigregorie, line 62: **Minor — `default` branch collapses every unmapped status to 502**

Unmapped statuses such as `429` (rate limit) or `422` (validation) get reported to the client as `502`, which misrepresents the failure class. Consider passing through 4xx statuses (with a cleaned/fallback message) and only collapsing genuine 5xx/unknown to 502.

audigregorie (commented):
## Code Review Summary

Solid, well-documented full-stack feature — the per-principal write model, signal-driven component, upstream-noise-stripping error mapping, and the 8-test e2e suite are all strong. One real correctness bug: the "org changed mid-flight" staleness guard in the three write handlers is a no-op and can let a late write response clobber a newly-selected org's list. Two low-priority backend notes below.

### What's done well

- Per-principal POST/PUT/DELETE replacing the full-replace read-modify-write (with a clear rationale comment) is the right architectural call.
- Method-free template via the precomputed `OrgAccessRowVm`; clean `signal`/`computed`/`linkedSignal` usage; correct `take(1)` + `takeUntilDestroyed(destroyRef)` on writes.
- Error mapping strips upstream noise and never echoes 5xx bodies; `Cache-Control: no-store` on access data.
- Comprehensive e2e coverage: happy paths, 5xx failure handling, non-manager gating, filter/search.

_Note: the campaign-file deletions visible in the `main..HEAD` diff are base-divergence noise (local `main` is ahead of the branch merge base), not part of this PR._

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
luismoriguerra: Great catch — you're right, the BFF echoes `orgUid` from the route param so `res.orgUid !== orgUid` could never be true and the guard was a no-op. Fixed: it now compares `res.orgUid` to the live selection `this.accountContext.selectedAccount().uid`, so a late invite response can't clobber a newly-selected org's list. See commit e57efb4a.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
luismoriguerra: Fixed — same live-selection comparison applied to the edit-save handler (`res.orgUid !== this.accountContext.selectedAccount().uid`). See commit e57efb4a.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.ts
luismoriguerra: Fixed — same live-selection comparison applied to the remove handler. See commit e57efb4a.

apps/lfx-one/src/server/services/microservice-proxy.service.ts
luismoriguerra: Done — extracted a single private `resolveBaseUrl(service)` helper and replaced the four inline `MICROSERVICE_URLS` literals (proxyRequest / proxyRequestWithResponse / proxyBinaryRequest / proxyStreamRequest) with it, so the env resolution lives in one place. See commit e57efb4a.

apps/lfx-one/src/server/services/org-lens-access.service.ts
luismoriguerra: Good call — `listAccessUsers` now takes an optional `knownCanManage`; the write paths assert canManage then pass `true` through to the post-write refresh, so each invite/edit/remove does a single role-grants lookup instead of two. See commit e57efb4a.

apps/lfx-one/src/server/helpers/access-error.helper.ts
luismoriguerra: Fixed — the default branch now passes through unmapped 4xx (e.g. 422 validation, 429 rate-limit) with a cleaned/fallback message, and only collapses genuine 5xx/unknown to 502 so server-internal noise still can't leak. See commit e57efb4a.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.html
luismoriguerra: Done — extracted `summarySkeletonCards: readonly number[] = [0, 1, 2];` next to `tableSkeletonRows` and referenced it in the @for, dropping the per-CD-pass array allocation. See commit e57efb4a.

packages/shared/src/interfaces/org-lens-access.interface.ts
luismoriguerra: Fixed — updated the comment to reflect that this shape is read via GET and writes go through the per-principal endpoints, not a full-replace PUT. See commit e57efb4a.

luismoriguerra:
## Review Feedback Addressed — Round 4 (@audigregorie)

Commit: `e57efb4a`

### Major
- **org-lens-access.component.ts** (invite / edit / remove): the mid-flight org guard compared `res.orgUid` to the captured `orgUid`, which the BFF always echoes back — so it never fired. Now compares against the **live** selection (`accountContext.selectedAccount().uid`), so a late response can't overwrite a newly-selected org's list. Thanks for catching that the round-1 guard was a no-op. (per @audigregorie)

### Minor
- **microservice-proxy.service.ts**: extracted a single private `resolveBaseUrl(service)` helper; the four methods no longer each repeat the `MICROSERVICE_URLS` map. (per @audigregorie)
- **org-lens-access.service.ts**: `listAccessUsers` now takes an optional `knownCanManage`; write paths assert canManage once and pass it through to the refresh, halving role-grants lookups per write. (per @audigregorie)
- **access-error.helper.ts**: unmapped 4xx (e.g. 422/429) pass through with a cleaned message instead of being misreported as 502; only 5xx/unknown collapse to 502. (per @audigregorie)
- **org-lens-access.component.{ts,html}**: extracted `summarySkeletonCards` readonly field instead of an inline `[0,1,2]` literal in `@for`. (per @audigregorie)
- **org-lens-access.interface.ts**: corrected the stale "full-replace semantics on PUT" comment. (per @audigregorie)

### Threads Resolved
8 of 8 of this round's threads resolved.

### Still Open
- **org-lens-access.service.ts** — `enrichJobTitles` pulling the full org `key_contact` set (round-1 thread). Tracked follow-up; happy to file a ticket. (flagged by copilot[bot])

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

audigregorie (dismissed):

apps/lfx-one/src/server/controllers/org-lens-access.controller.ts
```diff
@@ -0,0 +1,145 @@
...
+      next(error);
+      return;
+    }
+    // Manager-only gate (FR-011) — surface a clear 403 the UI maps to "no permission".
+    if (error instanceof MicroserviceError && error.statusCode === 403) {
+      logger.warning(req, operation, 'Org access write rejected: caller is not a manager', { status: 403 });
+      res.setHeader('Cache-Control', 'no-store');
+      res.status(403).json({ error: { code: 'FORBIDDEN', message: error.message, conflict: false } });
+      return;
+    }
```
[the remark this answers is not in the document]
luismoriguerra: Good catch — added `&& !error.originalMessage` to the gate. The local `assertCanManage` throws a MicroserviceError without `originalMessage`, whereas proxied upstream errors set it via `fromMicroserviceResponse`, so an upstream member-service 403 (token/ACL) now falls through to `mapAccessUpstreamError` instead of being mislabeled as the manager gate. (The 4xx pass-through added earlier surfaces it as a clean 403.) See commit ee82e728.

apps/lfx-one/src/server/services/microservice-proxy.service.ts
```diff
@@ -162,4 +146,14 @@ export class MicroserviceProxyService {
...
+
+  // Single source of truth for per-service base URLs. LFX_V2_MEMBER_SERVICE falls back to
+  // LFX_V2_SERVICE (the gateway) when unset.
+  private resolveBaseUrl(service: keyof MicroserviceUrls): string {
+    const urls: MicroserviceUrls = {
+      LFX_V2_SERVICE: process.env['LFX_V2_SERVICE'] || 'http://lfx-api.k8s.orb.local',
+      LFX_V2_MEMBER_SERVICE: process.env['LFX_V2_MEMBER_SERVICE'] || process.env['LFX_V2_SERVICE'] || 'http://lfx-api.k8s.orb.local',
+    };
+    return urls[service];
+  }
```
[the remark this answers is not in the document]
luismoriguerra: Done — `resolveBaseUrl` now strips trailing slashes (`.replace(/\/+$/, '')`), so a config value like `https://host/` no longer produces `host//b2b_orgs/...`. See commit ee82e728.

luismoriguerra:
## Review Feedback Addressed — Round 5 (Copilot)

Commit: `ee82e728`

### Changes Made
- **org-lens-access.controller.ts**: `handleWriteError` previously treated *any* 403 `MicroserviceError` as the local manager gate, which would mislabel an upstream member-service 403 (token/ACL) and bypass `mapAccessUpstreamError`. Now the gate also requires `!error.originalMessage` — the local `assertCanManage` throws without it, while proxied upstream errors set it via `fromMicroserviceResponse`, so upstream 403s fall through to the mapper (and the round-4 4xx pass-through surfaces them as a clean 403). (per copilot[bot])
- **microservice-proxy.service.ts**: `resolveBaseUrl` strips trailing slashes from the env value so a config like `https://host/` can't produce `host//path`. (per copilot[bot])

### Threads Resolved
2 of 2 of this round's threads resolved.

### Still Open
- **org-lens-access.service.ts** — `enrichJobTitles` pulling the full org `key_contact` set (round-1 thread). Tracked follow-up; happy to file a ticket. (flagged by copilot[bot])

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

luismoriguerra:
## Round 6 — Merge conflicts with `main` resolved

Merge commit: `d4103cea`

`main` was merged into `LFXV2-1908`. The only conflict was in **`apps/lfx-one/src/server/routes/orgs.route.ts`**, where `main` added the Org Lens **Training** controller/route (LFXV2-1895) and this branch added the Org Lens **Access** controller/routes (spec 025). Both are purely additive, so I kept both: both controller instantiations and both route groups are registered (imports already included both controllers). No other files conflicted.

### Status
- All review threads resolved (no unresolved threads remain).
- PR is now **mergeable** (no conflicts); awaiting the required approving review.

### Validation
`npm run build` and `npm run lint` both green post-merge (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

apps/lfx-one/src/server/services/org-lens-access.service.ts
```diff
@@ -0,0 +1,223 @@
...
+  private async assertCanManage(req: Request, orgUid: string, operation: string): Promise<void> {
+    const canManage = await this.resolveCanManage(req, orgUid);
+    if (!canManage) {
+      throw new MicroserviceError('You do not have permission to manage Org Lens access for this organization.', 403, 'FORBIDDEN', {
+        operation,
+        service: 'LFX_V2_MEMBER_SERVICE',
+        path: `/b2b_orgs/${orgUid}/settings/users`,
+      });
+    }
+  }
```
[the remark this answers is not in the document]
luismoriguerra: Good catch — `assertCanManage` no longer reuses the lenient `resolveCanManage`. It now does its own role-grants lookup and distinguishes the two failure modes: a *verified* non-writer still gets a clean 403 FORBIDDEN, but a role-grants lookup *failure* now throws a retriable 503 (`ROLE_GRANTS_UNAVAILABLE`) instead of a misleading "no permission". The list/read UX gate keeps using the lenient `resolveCanManage` (canManage=false) unchanged. See commit 6c9517ca.

luismoriguerra:
## Review Feedback Addressed — Round 7 (Copilot)

Commit: `6c9517ca`

### Changes Made
- **org-lens-access.service.ts**: `assertCanManage` no longer reuses the lenient `resolveCanManage` (which swallows role-grants lookup failures and returns `false`). It now does its own lookup and distinguishes **"not a manager"** (403 `FORBIDDEN`) from **"unable to verify"** (retriable 503 `ROLE_GRANTS_UNAVAILABLE`), so a transient role-grants outage no longer shows legitimate managers a misleading "no permission" error. The list/read UX gate keeps using the lenient `resolveCanManage` unchanged. (per copilot[bot])

### Threads Resolved
1 of 1 of this round's threads resolved.

### Still Open
- **org-lens-access.service.ts** — `enrichJobTitles` pulling the full org `key_contact` set (round-1 thread). Tracked follow-up; happy to file a ticket. (flagged by copilot[bot])

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

dealako (commented):
@luismoriguerra — solid feature work. This is a well-architected, defence-in-depth implementation and the six fix-rounds have addressed every substantial concern raised by Copilot, CodeRabbit, and @audigregorie. The authorisation model (write gate = `assertCanManage`, read gate = `canManage` flag) is cleanly separated, error handling prevents upstream noise from leaking to the client, and the E2E suite is deterministic and comprehensive.

**Consolidated review — three rounds of independent analysis (this reviewer + parallel security-auditor + code-quality subagent) merged into one report.**

---

### 🔴 Blocking: 0 issues

---

### 🟡 Minor: 2 issues

#### 1. Edit-modal `selectedRole` not reset on passive dismissal (click-outside / ESC)

`onCancel()` correctly re-seeds `selectedRole` when the Cancel button is clicked. But when the user dismisses via `dismissableMask` (click outside) or the ESC key, `p-dialog` fires `visibleChange → false` and bypasses `onCancel()` entirely. `linkedSignal` won't re-seed because the same `user()` reference is passed on re-open — leaving the previously-selected (unsaved) role pre-populated in the dropdown, with `canSave` already true. A second click on Save would issue an unintended role change.

**File:** `apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/edit-access-role-modal/edit-access-role-modal.component.ts`, line 59 (`onCancel`)

**Fix — add an effect that watches `visible()` dropping to `false`:**
```typescript
public constructor() {
  effect(() => {
    if (!this.visible()) {
      this.selectedRole.set(this.user()?.role ?? 'viewer');
    }
  });
}
```
Or wire `(visibleChange)="onVisibilityChange($event)"` in the template. Either approach covers all close paths (Cancel button, click-outside, ESC).

#### 2. `name` field has no length cap in `parseInviteBody`

The optional `name` parameter in the invite body (`POST /api/orgs/:orgUid/lens/access/users`) is accepted and forwarded to member-service without a max-length check. While only authenticated managers reach this path, a simple defence-in-depth guard prevents an oversized payload from propagating downstream.

**File:** `apps/lfx-one/src/server/controllers/org-lens-access.controller.ts`, line 105 (`parseInviteBody`)

```typescript
const name = raw.name != null ? String(raw.name).trim() : '';
if (name.length > 255) {
  throw ServiceValidationError.forField('name', 'Name must be 255 characters or fewer', { operation });
}
return { email, role, name: name || null };
```

---

### ⚪ Nit: 2 issues

#### 3. `OrgAccessTypeFilterOption` interface in the constants file

`OrgAccessTypeFilterOption` is exported from `packages/shared/src/constants/org-people.constants.ts` rather than `packages/shared/src/interfaces/`. Per repo conventions, interfaces belong in the `interfaces/` directory. Low-friction move for consistency.

#### 4. AI attribution marker missing from source files

Per `.claude/rules/development-rules.md`, AI-assisted source files should include an attribution comment. The PR description notes "Made with Cursor" — the new `.ts` / `.spec.ts` files should carry a corresponding attribution marker. (CodeRabbit's prior learning flagged the exact requirement.)

---

### Revision tracking — all prior-round findings resolved ✅

| Round | Reviewer | Finding | Status |
|-------|----------|---------|--------|
| R1 | Copilot | `selectedRole` not reset on cancel (Cancel-button path) | ✅ Fixed (13288bbd) |
| R1 | Copilot | Stale-org guard was a no-op on all 3 mutation handlers | ✅ Fixed (e57efb4a) |
| R1 | Copilot | `inviteStatus` type too wide (revoked/expired reachable) | ✅ Fixed — `OrgAccessListedInviteStatus` narrowed (13288bbd) |
| R2 | CodeRabbit | `getByText` dropdown locators flake-prone | ✅ Fixed — `getByRole('option', …)` (3f9a3b4c) |
| R2 | CodeRabbit | `pTooltip` on natively `[disabled]` buttons doesn't fire | ✅ Fixed — span wrapper (3f9a3b4c) |
| R3 | Copilot | Upstream 403 mislabelled as local manager gate | ✅ Fixed — `!error.originalMessage` guard (ee82e728) |
| R3 | Copilot | Trailing slash in env value produces `host//path` | ✅ Fixed — `resolveBaseUrl` strips trailing `/` (ee82e728) |
| R3 | Copilot | `assertCanManage` swallowed role-grants failures as `false` | ✅ Fixed — 503 vs 403 distinction (6c9517ca) |
| R3 | @audigregorie | Stale-org guard compared wrong values | ✅ Fixed — compare against live selection (e57efb4a) |
| R3 | @audigregorie | READY race in `FeatureFlagService` | ✅ Fixed — handlers registered first (ccb696bd) |
| Open | Copilot | `enrichJobTitles` full-table scan per access-list load | 📋 Acknowledged as known follow-up — acceptable trade-off for this PR |

---

### Security notes (from parallel security-auditor pass)

- **Authorization on writes is solid** — `assertCanManage` (throws 403 / retriable 503) called before every POST/PUT/DELETE. The "not a manager" vs "unable to verify" distinction is correct. ✅
- **Auth middleware confirmed** — `authMiddleware` applied at `server.ts:206`; `/api/orgs` registered at line 221. All new routes are protected. ✅
- **GET endpoint read access** — `canManage` is a UX-only gate on reads; backend enforces only on writes. Appears intentional per spec FR-011. If manager-only reads are required, `assertCanManage` can be added to `getUsers` — recommend confirming with product.
- **Email normalisation** — `toLowerCase()` applied consistently on both read and write paths. Acceptable as long as member-service treats email lookups case-insensitively (standard practice). ✅
- **Input validation** — email and role are validated server-side; `name` field gap captured above as [minor]. ✅

---

### Issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 — passive-dismissal reset, name length cap
- ⚪ Nit: 2 — interface placement, AI attribution

---

**✅ Approved with minor comments**

The two [minor] items are worth fixing before merge (the passive-dismissal gap can silently produce wrong role saves), but neither is a blocker given the overall defence-in-depth posture. Happy to re-review after the modal fix lands.

dealako:
@luismoriguerra The security audit flagged a HIGH finding (missing name field length validation) and the code-quality pass flagged a critical UX bug (passive-dismissal selectedRole not reset). Neither is truly blocking in isolation, but together they warrant waiting for fixes before formally approving.

luismoriguerra:
## Round 8 — @dealako consolidated review addressed

Thanks for the thorough consolidated pass, @dealako. All four items are now addressed.

### 🟡 Minor (both fixed)
1. **Edit-modal `selectedRole` not reset on passive dismissal** — fixed in `edit-access-role-modal.component.ts` (commit `774952ff`). Replaced the Cancel-only reset with an `effect(() => { if (!visible()) selectedRole.set(user()?.role ?? 'viewer'); })`, which covers **all** close paths — Cancel button, `dismissableMask` click-outside, and ESC — so a discarded selection can no longer survive a same-user re-open with `canSave()` true.
2. **`name` length cap** — `parseInviteBody` now rejects names > 255 chars with a `ServiceValidationError` (commit `774952ff`).

### ⚪ Nit (both addressed)
3. **`OrgAccessTypeFilterOption` placement** — moved from `org-people.constants.ts` into `interfaces/org-lens-access.interface.ts`, consistent with `OrgAccessRoleOption` (commit `774952ff`).
4. **AI attribution marker** — added to the 9 new spec-025 source files (commit `d31ce48b`). Used an accurate `// Generated with [Cursor](https://cursor.com)` marker since this PR was authored with Cursor rather than Claude Code (the rule's literal example string).

### Security notes
- **Manager-only reads**: agreed this is intentional per FR-011 (`canManage` is a UX-only gate; the backend enforces on writes only). Leaving as-is unless product wants manager-gated reads — happy to add `assertCanManage` to `getUsers` if so.
- All other security-auditor observations were already green.

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

The only remaining open item is the `enrichJobTitles` full-table-scan follow-up, which you've also flagged as an acceptable trade-off for this PR. Happy to re-review.

apps/lfx-one/src/server/services/org-lens-access.service.ts
```diff
@@ -0,0 +1,255 @@
...
+      isWriter = grants.writers.includes(orgUid);
+    } catch (error) {
+      // Couldn't verify (transient role-grants outage) — surface a retriable error, not a 403.
+      logger.warning(req, 'assert_org_access_can_manage', 'Role-grants lookup failed; cannot verify manager permission', {
+        org_uid: orgUid,
+        operation,
+        err: error instanceof Error ? error.message : String(error),
+      });
+      throw new MicroserviceError("Couldn't verify your permissions right now. Please try again.", 503, 'ROLE_GRANTS_UNAVAILABLE', {
+        operation,
```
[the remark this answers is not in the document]
luismoriguerra: Good catch — fixed. The `ROLE_GRANTS_UNAVAILABLE` error now reports `path: '/query/resources'` (the role-grants lookup goes through `LFX_V2_SERVICE` → query-service), instead of the member-service settings path, so outage telemetry points at the real failing endpoint. `service` was already `LFX_V2_SERVICE`. See commit 8f0f9824.

luismoriguerra:
## Review Feedback Addressed — Round 9 (Copilot)

Commit: `8f0f9824`

### Changes Made
- **org-lens-access.service.ts**: the `ROLE_GRANTS_UNAVAILABLE` error (added last round) reported a member-service settings `path`, but the actual failing upstream is the role-grants lookup via query-service. Set `path: '/query/resources'` so transient-outage telemetry/logs point at the real endpoint (`service` was already `LFX_V2_SERVICE`). (per copilot[bot])

### Threads Resolved
1 of 1 of this round's threads resolved.

### Still Open
- **org-lens-access.service.ts** — `enrichJobTitles` full-table scan per access-list load (round-1 thread; accepted as a known trade-off for this PR). (flagged by copilot[bot])

### Validation
`npm run build` and `npm run lint` both green (lint's single warning is pre-existing in an unrelated `docs-manifest.service.ts`).

dealako (approved):
@luismoriguerra — thanks for the fast and thorough turnaround on the feedback. Three clean fix commits, a dedicated attribution pass, and an extra observability correction from Copilot on top. Genuinely good iteration speed.

**Consolidated re-review — independent analysis + parallel security-auditor + code-quality subagent.**

---

### 👏 Nice work this revision

- **`effect()` covers every close path** — Cancel button, click-outside, ESC key all correctly reset `selectedRole`. The implementation is sound: the initial effect firing (with `visible=false`) is a harmless no-op since `linkedSignal` already initialises to the same value. The `onCancel()` simplification (it now just calls `visible.set(false)` and lets the effect handle the reset) avoids double-resetting.
- **Interface placement** — `OrgAccessTypeFilterOption` landed in exactly the right file (`org-lens-access.interface.ts`), and the component's import was updated cleanly.
- **`name` length cap** — correct placement (after `.trim()`, before the return), clear user-facing error message.
- **Observability fix** — correcting the ROLE_GRANTS_UNAVAILABLE error path to `/query/resources` with an explanatory comment is the right call for telemetry accuracy.
- **Attribution** — all 9 new source files now carry the AI-assistance marker.

---

### Step 2 — Revision tracking

| Prior item | Status | Detail |
|---|---|---|
| [minor] Edit-modal passive dismissal not reset | ✅ **Resolved** | `effect()` in constructor resets `selectedRole` on all close paths — 774952ff |
| [minor] `name` field no length cap | ✅ **Resolved** | 255-char limit added in `parseInviteBody` — 774952ff |
| [nit] `OrgAccessTypeFilterOption` in constants file | ✅ **Resolved** | Moved to `org-lens-access.interface.ts`, import updated — 774952ff |
| [nit] AI attribution markers missing | ✅ **Resolved** | `// Generated with [Cursor](https://cursor.com)` added to all 9 files — d31ce48b |
| Open (design Q) GET endpoint read access UX-only gate | 📋 **By design** — no change needed |

---

### Step 3 — New issue scan

No new issues found in the three fix commits. Changes are minimal and surgical:
- Attribution markers are comment-only additions with no logic impact.
- The `effect()` and name-validation additions are straightforward.
- The interface relocation has no runtime effect; TypeScript enforces the import path at build time.

---

### Step 4 — AI comment reconciliation

**code-reviewer-pro subagent** found one suggestion: the attribution marker uses `[Cursor]` while `.claude/rules/development-rules.md` specifies `[Claude Code]`. This is a nit — the rule is written for Claude Code–assisted work, and using the actual tool's name (Cursor) is more accurate. No action needed.

**security-auditor subagent** raised several findings I'm pushing back on after reviewing the code:

- **"CRITICAL" effect race condition** — Disagreed. The scenario requires `visible.set(false)` and `visible.set(true)` in the same Angular change-detection batch (i.e., cancel and immediately reopen within a single microtask). In that case the effect sees the final `visible=true` state and correctly skips the reset — which is the right behaviour since the modal is still open. For all realistic user flows the effect fires correctly. Not a security issue.
- **"HIGH" UTF-8 byte length** — Disagreed. `String.length` measures UTF-16 code units (characters), which aligns with `VARCHAR(255 CHAR)` database semantics. Byte-length validation would only matter if the upstream DB column is `BYTES`-typed; that's an unknown upstream contract, not a flaw in this code. This is a nit at most.
- **"HIGH" Cursor annotation policy** — Disagreed. The prior review explicitly flagged the *missing* attribution as a nit and the author correctly attributed to the actual tool used. Using `[Cursor]` for Cursor-generated work is more honest than using `[Claude Code]`.
- **"MEDIUM" observability** — Disagreed. The path fix is an *improvement* to telemetry, not a problem. The suggestion to add extra metadata is a nit.
- **"MEDIUM" type relocation without migration guide** — Disagreed. This is an internal monorepo; TypeScript compile errors catch any stale imports immediately. No CHANGELOG entry needed.

No security-auditor findings are actionable.

---

### Issue count (open items across all rounds)

- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 0 issues

---

**✅ Approved**

All [minor] and [nit] items from the previous round are resolved. No new issues. Ready to merge once CI passes and the branch is rebased onto main (it's currently 2 commits behind, no conflicts).

PR #847, opened 2026-06-01
fix(newsletters): back nav to sent tab, confirm dialog keys, empty chart

niravpatel27:
## Summary

Three small newsletter-module fixes.

- **Analytics → Back lands on the Sent tab.** Was navigating to the list with no tab hint, defaulting to Drafts. Now passes `?tab=sent`, which the existing `newsletter-list` constructor already consumes. Analytics is only reachable from Sent, so the anchor is unambiguous.
- **AI "Replace existing content?" dialog needed multiple clicks.** Two keyless `<p-confirmDialog>` instances were mounted on the manage page at the same time (parent + child), and PrimeNG's `ConfirmationService` broadcasts to every keyless dialog — both opened stacked, so the first click only dismissed the top one. Fixed by scoping each `<p-confirmDialog>` in the newsletters module with a unique `key=` and the matching `key:` in the `confirmationService.confirm({...})` call.
- **Empty "Opens over time" chart.** When the upstream rollup returned a non-zero `total_opens` with an empty `daily_opens`, the chart rendered an empty axis. Added a `hasDailyBreakdown` gate and a distinct empty state ("Daily breakdown not available yet. Check back shortly.") so the section no longer looks broken.

The third item is the UI cover for the upstream gap that linuxfoundation/lfx-v2-newsletter-service#16 addresses; once that PR lands and is deployed, the empty state will only surface during the brief gap between send and email-service group-index propagation.

## Test plan

- [ ] Newsletters → click any Sent row → analytics → Back → confirm Sent tab is selected and URL has `?tab=sent`.
- [ ] On a draft with existing subject + body, Generate with AI → confirm a single dialog appears, Keep current / Replace each respond on the first click.
- [ ] Manage page Send (with valid recipients) → "Send newsletter?" still opens, Send now / Cancel each respond on the first click.
- [ ] List page delete a draft → "Delete draft?" opens, Delete / Cancel each respond on the first click.
- [x] `yarn check-types && yarn lint:check` clean.

### DIFF
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-content-step/newsletter-content-step.component.html
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-content-step/newsletter-content-step.component.html
@@ -61,4 +61,4 @@
   (generated)="onGenerated($event)">
 </lfx-newsletter-generate-drawer>
 
-<p-confirmDialog data-testid="newsletter-content-confirm-dialog"></p-confirmDialog>
+<p-confirmDialog key="newsletter-content-step" data-testid="newsletter-content-confirm-dialog"></p-confirmDialog>
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-content-step/newsletter-content-step.component.ts
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-content-step/newsletter-content-step.component.ts
@@ -55,6 +55,7 @@ export class NewsletterContentStepComponent {
     const hasBody = this.bodyFilled();
     if (hasSubject || hasBody) {
       this.confirmationService.confirm({
+        key: 'newsletter-content-step',
         header: 'Replace existing content?',
         message: 'This will overwrite your current subject and body with the AI-generated newsletter. You can still edit the result before sending.',
         icon: 'pi pi-exclamation-triangle',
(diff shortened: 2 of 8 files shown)


audigregorie (commented):
## Code Review Summary

Clean, tightly-scoped PR — all three fixes verified against the surrounding code, no issues found.

**Verification done beyond the diff:**

- **Dialog collision fix is root-cause correct.** Confirmed `newsletter-content-step` is nested inside `newsletter-manage` (`newsletter-manage.component.html:123`) and declares no own `ConfirmationService` provider, so it injects manage's shared instance (`newsletter-manage.component.ts:64`). Two keyless `<p-confirmDialog>`s on one shared `ConfirmationService` genuinely broadcast-collide — scoping each dialog + each `confirm({ key })` 1:1 is the precise fix. Verified all 3 `confirm()` calls and all 3 dialogs are keyed with matching keys and there are no orphaned (now-unreachable) `confirm()` calls.
- **`newsletter-list`** has its own `ConfirmationService` instance (`newsletter-list.component.ts:39`), so its dialog never actually collided — the key there is harmless consistency, fine to keep.
- **`?tab=sent` Back nav is wired end-to-end.** The list consumes `tab` from the query snapshot (`newsletter-list.component.ts:92-94`), so `goBack()` passing `queryParams: { tab: 'sent' }` resolves to the Sent tab on a freshly-created list component.
- **`hasDailyBreakdown` gate is correct.** The `@if (!hasOpens) → @else if (!hasDailyBreakdown) → chart` ordering means the new empty state only surfaces when opens exist but per-day data doesn't. The `daily_opens?.length` optional chaining is appropriate defensiveness for the documented upstream gap, even though the interface types `daily_opens` as required.

### What's done well

- Excellent explanatory "why" comments on the new `hasDailyBreakdown` computed and the keyed-dialog rationale in the PR body.
- Idiomatic zoneless/signals usage — `computed()` for the single-expression gate, no `*Sig` suffix, no needless `init*` extraction.
- Minimal, surgical diff that fixes three distinct UX bugs without scope creep.

**Rating: 5/5.** No Critical, Major, or Minor findings to action.

audigregorie (approved):

PR #850, opened 2026-06-02
fix(search): rank member search by name relevance

manishdixitlfx:
## Summary

Member/person search (adding members to groups, assigning roles) over-matches: typing `il` returns everyone with an incidental `i`/`l` hit — including matches **inside the email address** — surfacing unrelated people instead of the intended name match (LFXV2-2058).

**Root cause (upstream, not this repo):** the BFF (`server/services/search.service.ts`) is a thin proxy — it sends `name=` or `tags=email:` to the `LFX_V2_SERVICE` `/query/resources` endpoint and only de-dupes. The over-match lives in **`lfx-v2-query-service`** (`internal/infrastructure/opensearch/template.go`), which matches `name` via a `multi_match` `"type": "bool_prefix"` over `name_and_aliases` plus its `._2gram`/`._3gram` shingle subfields. Aliases fold in email and the ngram subfields loosen matching, so a name query matches inside unrelated emails/aliases.

The proper fix (restrict matched fields, only search email when `tags=email:` is used, rank exact > prefix > substring server-side) belongs upstream in `lfx-v2-query-service`. **Follow-up ticket to be filed against that repo.** This PR is the client-side mitigation we can ship from the BFF in the meantime.

## What this PR does

- Adds `rankUserSearchResults()` in `@lfx-one/shared/utils` — a pure client-side re-rank that sorts results **exact name > name-prefix > name-substring > username > email > incidental**, so intended matches surface first and incidental email/alias hits are demoted to the bottom.
  - **Demotes, never drops or caps:** low-relevance rows sort last but are always retained, so a legitimate hit on an upstream alias the client can't see is never hidden. (Two `fix(review)` commits tightened this from an initial drop-then-cap approach after review feedback.)
  - **Email queries** (query contains `@`) need no special case — name fields won't match, so genuine email hits naturally rank above the rest; the explicit email search is never demoted away.
- Wires the helper into both search entry points: `user-search.component.ts` and the committee `add-member-dialog.component.ts` result streams.

## Tests

- Adds **Vitest** to `packages/shared` (the package's first unit-test runner — flagging so the team can ratify the choice) with a `test` script wired into the existing `turbo run test`. Specs (`*.spec.ts`) are excluded from the production `tsc` build via `tsconfig.json`.
- 13 unit tests covering prefix, mid-name substring, username-only, email-only, incidental (ngram noise), stable tiering, no-cap retention, the `@` email-query path, and empty-query passthrough.

## Protected files

- `packages/shared/package.json` + `yarn.lock` — adds Vitest as a **devDependency** of the shared package only (does not ship). Flagging for code-owner awareness.

## Validation

`yarn lint` ✅ · `yarn format` ✅ · shared `vitest` 13/13 ✅ · `yarn build` ✅ (both packages) · DCO + GPG signed on all 3 commits, rebased on `main`.

### DIFF
--- a/packages/shared/src/utils/search.utils.ts
+++ b/packages/shared/src/utils/search.utils.ts
@@ -25,13 +25,6 @@ export enum UserSearchRelevance {
 /** Minimal user shape needed to rank a search result. */
 type RankableUser = Pick<UserSearchResult, 'first_name' | 'last_name' | 'email' | 'username'>;
 
-/**
- * Default cap on ranked results handed to a typeahead. Demoted upstream
- * ngram/alias noise sorts to the bottom and falls off this limit, while real
- * matches (which sort first) survive. Pass `{ limit: Infinity }` to opt out.
- */
-const DEFAULT_RESULT_LIMIT = 10;
-
 function normalize(value: string | null | undefined): string {
   return (value ?? '').trim().toLowerCase();
 }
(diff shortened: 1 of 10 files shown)


packages/shared/src/utils/search.utils.ts
```diff
@@ -0,0 +1,99 @@
...
+// SPDX-License-Identifier: MIT
+
+import { UserSearchResult } from '../interfaces';
+
+/**
+ * Relevance tiers for a user search result against a query.
+ * Lower values rank higher (sort ascending).
+ */
+export enum UserSearchRelevance {
+  /** Query equals a name field (first, last, or full name). */
```
MRashad26: **Enum belongs in `packages/shared/src/enums/`** — CLAUDE.md lists `enums/` as the dedicated directory for shared enumerations (`meeting.enum.ts`, `committee.enum.ts`, etc.). `UserSearchRelevance` is exported and public, so it follows the same convention.

Suggested move:
```
packages/shared/src/enums/search.enum.ts
```
```ts
// search.enum.ts
export enum UserSearchRelevance {
  ExactName = 0,
  NamePrefix = 1,
  NameSubstring = 2,
  UsernameMatch = 3,
  EmailMatch = 4,
  Incidental = 5,
}
```
Then import it in `search.utils.ts` and re-export from the enums barrel. `scoreUserSearchResult` and `rankUserSearchResults` stay in `utils/` — only the enum moves.

MRashad26 (commented):
Logic is correct, performance is fine, yarn.lock is required. One minor CLAUDE.md convention finding inline.

packages/shared/src/utils/search.utils.ts
manishdixitlfx: Done — moved `UserSearchRelevance` to `packages/shared/src/enums/search.enum.ts` and re-exported it from the enums barrel; `search.utils.ts` and its spec now import it from `../enums`. Thanks for the catch — it matches the committee/meeting enum convention.

See commit f56e12a1.

manishdixitlfx:
## Review Feedback Addressed

Commit: f56e12a1

### Changes Made
- **packages/shared/src/enums/search.enum.ts** (new) + **enums/index.ts**: moved the exported `UserSearchRelevance` enum out of `utils/` into the shared `enums/` directory and re-exported it from the barrel, matching the `committee.enum.ts` / `meeting.enum.ts` convention (per @MRashad26)
- **search.utils.ts** / **search.utils.spec.ts**: import the enum from `../enums`

### Threads Resolved

Validation: shared tests 13/13, lint, and build all pass; commit is DCO + GPG signed.

MRashad26 (approved):

PR #856, opened 2026-06-02
fix(newsletters): carry project_uid in edit/analytics URLs

niravpatel27:
Fixes prod 404 on the newsletter analytics page (`Could not load analytics`).

The previous routes (`:id/edit`, `:id/analytics`) re-derived project_uid from `ProjectContextService.activeContextUid` at load time. On the foundation lens — where the list aggregates newsletters from multiple child projects — that ambient UID didn't match the newsletter's owning project, so the backend's project-scoped analytics endpoint correctly returned 404.

This PR moves project_uid into the route (`:projectUid/:id/{edit,analytics}`). The list component threads `item.project_uid` into the navigation, so the owning project always travels with the link.

## Test plan
- [ ] From the foundation lens, click into a sent newsletter that belongs to a child project → analytics loads.
- [ ] Deep-link `/foundation/newsletters/{project_uid}/{id}/analytics` directly → loads (no need to wait for ambient context).
- [ ] Edit a draft, refresh the page → still loads under the same URL.
- [ ] Create flow still works (no projectUid segment; falls back to active context).


dealako (approved):
Hey Nirav 👋 — clean fix for a genuinely tricky problem. The root cause is clearly understood, the solution is idiomatic Angular (URL as the single source of truth for route-scoped identity), and the surface area is minimal. Removing the `combineLatest` + `toObservable` wait loop in analytics makes the component noticeably simpler without sacrificing correctness.

CI is green across all checks (CodeQL, Code Quality, DCO, PR Title Lint). CodeRabbit found no actionable comments. I have **0 blocking** and **2 minor** issues, plus **2 nits**.

---

### Findings

#### [minor] Error message is ambiguous on the impossible guard path — agrees with Copilot
`newsletter-analytics.component.ts` line 65: `'Missing newsletter id or project.'` doesn't name the specific missing segment. Since `:projectUid` is the new required param that didn't exist before, naming it explicitly makes the message more useful if this ever surfaces in a bug report. Copilot raised the same point — I agree with their comment.

Suggested text: `'Missing newsletter id or project UID in route.'`

---

#### [minor] `newsletterAccessGuard` checks active context, not the URL's `projectUid`
`newsletter-access.guard.ts` (unchanged) resolves the project slug from `route.queryParamMap.get('project')`, falling back to `projectContextService.activeContext()?.slug`. The new `:projectUid/:id/analytics` and `:projectUid/:id/edit` routes don't carry a `?project=` query param, so on a direct deep link the guard falls back to whichever project is currently active in context — which may not be the newsletter's owning project.

This is a pre-existing guard design, not introduced by this PR. The backend's project-scoped analytics endpoint provides the authoritative authorization check, so this isn't a security hole in practice. But it's worth noting: as the URL now explicitly identifies the owning project via `projectUid`, the guard could be made more precise by resolving the slug from the project service using that UID. I'd suggest opening a follow-up ticket rather than blocking this PR on it.

---

#### [nit] `routeProjectUid` declaration exceeds line length and diverges from the `init*` pattern
`newsletter-manage.component.ts` line 112 is quite long (the `toSignal(this.route.paramMap.pipe(map(...)))` inline form). Per the codebase's `component-organization.md` convention, `toSignal` calls with piped logic belong in a private `initRouteProjectUid()` method at the bottom of the class — the same pattern used by `initCurrentStep()` just below. Not blocking, but aligning it would be consistent.

---

#### [nit] Stale comment in `goBack()` (minor context mismatch)
`newsletter-analytics.component.ts` line 91: the comment says `'['..']` on a 2-segment route` — this referred to the old `:id/analytics` shape. With `:projectUid/:id/analytics` it's now 3 segments. The fix described (`relativeTo: this.route.parent` + explicit `'list'` child) is still correct and the comment's intent remains clear, but the segment count is off. A one-word update would clean it up.

---

### AI Bot Reconciliation

- **CodeRabbit** — No actionable findings. No conflicts.
- **Copilot** — One inline comment on the `'Missing newsletter id or project.'` message. I agree: naming `projectUid` explicitly would make this more debuggable. My [minor] above covers the same finding.

---

The fix is correct, well-scoped, and safe to merge. The two minor items are worth addressing but don't block landing this. The guard observation is a follow-up candidate rather than a blocker.

PR #858, opened 2026-06-02
feat(votes): remove Draft tab in personal vote view (LFXV2-2069)

audigregorie:
# Summary

The Draft tab (backed by `PollStatus.DISABLED`) is only relevant in management contexts like the project or committee lens, where maintainers create and manage polls. In the Me lens, contributors have no need to see draft polls — showing the tab was confusing and exposed an irrelevant filter.

This change converts `statusTabOptions` from a static array to a computed signal gated on a new `showDraftTab` input. The votes-dashboard passes `[showDraftTab]="!isMeLens()"`, so the tab is present everywhere except the Me lens without touching any other lens's behaviour.


ahmedomosanya (approved):
LGTM. Small, well-scoped change that converts statusTabOptions from a static array to a computed signal gated on a new showDraftTab input, and wires [showDraftTab]="!isMeLens()" from the dashboard so Draft only hides in the Me lens.

MRashad26 (approved):

PR #860, opened 2026-06-02
fix(newsletters): preserve clipboard links on paste into AI content

niravpatel27:
## Summary

Textarea drops anchor `href` on paste, so URLs vanish before the AI generator sees them. Intercept paste in the **Generate with AI** drawer and rewrite `<a href="URL">TEXT</a>` to `[TEXT](URL)` (bare URL when text matches href).

## Test plan

- [ ] Open Generate with AI drawer, paste from a page with inline links → Markdown `[text](url)` appears
- [ ] Paste plain text from a terminal → unchanged
- [ ] Generate → AI response surfaces the URLs as proper links

### DIFF
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
@@ -15,6 +15,7 @@ import {
   NEWSLETTER_SYSTEM_PROMPT_MAX_LENGTH,
 } from '@lfx-one/shared/constants';
 import { GenerateNewsletterResponse } from '@lfx-one/shared/interfaces';
+import { htmlClipboardToText } from '@lfx-one/shared/utils';
 import { NewsletterService } from '@services/newsletter.service';
 import { ProjectContextService } from '@services/project-context.service';
 import { MessageService } from 'primeng/api';
@@ -106,6 +107,32 @@ export class NewsletterGenerateDrawerComponent {
         // Silently ignore storage errors (private browsing, quota).
       }
     }
+  }
+
+  // Preserve hyperlinks when pasting rich content. A textarea would otherwise drop
+  // hrefs and keep only visible text, hiding URLs from the AI generator downstream.
+  public onRawContentPaste(event: ClipboardEvent): void {
+    const html = event.clipboardData?.getData('text/html') ?? '';
+    if (!html) return;
+
+    const textarea = event.target as HTMLTextAreaElement | null;
+    if (!textarea) return;
+
+    event.preventDefault();
+
+    const converted = htmlClipboardToText(html);
+    const start = textarea.selectionStart ?? textarea.value.length;
+    const end = textarea.selectionEnd ?? textarea.value.length;
+    const current = textarea.value;
+    const next = current.slice(0, start) + converted + current.slice(end);
+
+    this.form.controls.rawContent.setValue(next);
+
+    const caret = start + converted.length;
+    queueMicrotask(() => {
+      textarea.setSelectionRange(caret, caret);
+      textarea.focus();
+    });
   }
 
   public onGenerate(): void {
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
@@ -109,32 +109,6 @@ export class NewsletterGenerateDrawerComponent {
     }
   }
 
-  // Preserve hyperlinks when pasting rich content. A textarea would otherwise drop
-  // hrefs and keep only visible text, hiding URLs from the AI generator downstream.
-  public onRawContentPaste(event: ClipboardEvent): void {
-    const html = event.clipboardData?.getData('text/html') ?? '';
-    if (!html) return;
-
-    const textarea = event.target as HTMLTextAreaElement | null;
-    if (!textarea) return;
-
-    event.preventDefault();
-
-    const converted = htmlClipboardToText(html);
-    const start = textarea.selectionStart ?? textarea.value.length;
-    const end = textarea.selectionEnd ?? textarea.value.length;
-    const current = textarea.value;
-    const next = current.slice(0, start) + converted + current.slice(end);
-
-    this.form.controls.rawContent.setValue(next);
-
-    const caret = start + converted.length;
-    queueMicrotask(() => {
-      textarea.setSelectionRange(caret, caret);
-      textarea.focus();
-    });
-  }
-
   public onGenerate(): void {
     if (!this.canGenerate()) return;
     this.generating.set(true);
@@ -174,6 +148,38 @@ export class NewsletterGenerateDrawerComponent {
           });
         },
       });
+  }
+
+  // Preserve hyperlinks when pasting rich content. A textarea would otherwise drop
+  // hrefs and keep only visible text, hiding URLs from the AI generator downstream.
+  protected onRawContentPaste(event: ClipboardEvent): void {
+    const html = event.clipboardData?.getData('text/html') ?? '';
+    if (!html) return;
+
+    const textarea = event.target;
+    if (!(textarea instanceof HTMLTextAreaElement)) return;
+
+    event.preventDefault();
+
+    const converted = htmlClipboardToText(html);
+    const start = textarea.selectionStart ?? textarea.value.length;
+    const end = textarea.selectionEnd ?? textarea.value.length;
+    const current = textarea.value;
+    const next = current.slice(0, start) + converted + current.slice(end);
+
+    const control = this.form.controls.rawContent;
+    control.setValue(next);
+    // setValue bypasses the value accessor, so dirty/touched would not flip
+    // the way a native paste does. Set them explicitly so validators that gate
+    // on touched (e.g., error display) behave consistently with manual typing.
+    control.markAsDirty();
+    control.markAsTouched();
+
+    const caret = start + converted.length;
+    queueMicrotask(() => {
+      textarea.setSelectionRange(caret, caret);
+      textarea.focus();
+    });
   }
 
   private restoreCustomPrompt(): void {
--- a/packages/shared/src/utils/html-utils.spec.ts
+++ b/packages/shared/src/utils/html-utils.spec.ts
@@ -75,7 +75,7 @@ describe('htmlClipboardToText', () => {
   });
 
   it('decodes decimal numeric entities (em dash, NBSP, smart quote)', () => {
-    expect(htmlClipboardToText('hello&#8212;world&#160;&#8217;tis')).toBe("hello—world ’tis");
+    expect(htmlClipboardToText('hello&#8212;world&#160;&#8217;tis')).toBe('hello—world ’tis');
   });
 
   it('decodes hexadecimal numeric entities', () => {
--- a/packages/shared/src/utils/html-utils.spec.ts
+++ b/packages/shared/src/utils/html-utils.spec.ts
@@ -1,0 +1,93 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { describe, expect, it } from 'vitest';
+
+import { htmlClipboardToText, stripHtml } from './html-utils';
+
+describe('stripHtml', () => {
+  it('removes tags and decodes entities', () => {
+    expect(stripHtml('<p>Hello &amp; <strong>World</strong></p>')).toBe('Hello & World');
+  });
+
+  it('returns empty string for null / undefined / empty', () => {
+    expect(stripHtml(null)).toBe('');
+    expect(stripHtml(undefined)).toBe('');
+    expect(stripHtml('')).toBe('');
+  });
+});
+
+describe('htmlClipboardToText', () => {
+  it('returns empty string for null / undefined / empty', () => {
+    expect(htmlClipboardToText(null)).toBe('');
+    expect(htmlClipboardToText(undefined)).toBe('');
+    expect(htmlClipboardToText('')).toBe('');
+  });
+
+  it('converts an anchor with distinct text to Markdown', () => {
+    expect(htmlClipboardToText('<a href="https://docs.example.com/guide">contributor guide</a>')).toBe('[contributor guide](https://docs.example.com/guide)');
+  });
+
+  it('renders a bare URL when the anchor text equals the href', () => {
+    expect(htmlClipboardToText('<a href="https://x.com">https://x.com</a>')).toBe('https://x.com');
+  });
+
+  it('uses just the text when the anchor has no href', () => {
+    expect(htmlClipboardToText('<a>no href here</a>')).toBe('no href here');
+  });
+
+  it('uses just the text when the href is empty', () => {
+    expect(htmlClipboardToText('<a href="">link text</a>')).toBe('link text');
+  });
+
+  it('preserves multiple anchors in one block', () => {
+    const html = '<p>See <a href="https://a.com">A</a> and <a href="https://b.com">B</a>.</p>';
+    expect(htmlClipboardToText(html)).toBe('See [A](https://a.com) and [B](https://b.com).');
+  });
+
+  it('strips nested formatting inside anchor text', () => {
+    expect(htmlClipboardToText('<a href="https://x.com"><strong>bold</strong> link</a>')).toBe('[bold link](https://x.com)');
+  });
+
+  it('handles single-quoted href values', () => {
+    expect(htmlClipboardToText("<a href='https://x.com'>x</a>")).toBe('[x](https://x.com)');
+  });
+
+  it('still finds the href when other attributes come first', () => {
+    expect(htmlClipboardToText('<a class="link" data-foo="bar" href="https://x.com">x</a>')).toBe('[x](https://x.com)');
+  });
+
+  it('decodes entities in plain text', () => {
+    expect(htmlClipboardToText('cats &amp; dogs')).toBe('cats & dogs');
+  });
+
+  it('decodes entities inside anchor text', () => {
+    expect(htmlClipboardToText('<a href="https://x.com">cats &amp; dogs</a>')).toBe('[cats & dogs](https://x.com)');
+  });
+
+  it('turns block boundaries into newlines', () => {
+    expect(htmlClipboardToText('<p>Hello</p><p>World</p>')).toBe('Hello\nWorld');
+  });
+
+  it('turns <br> into a newline', () => {
+    expect(htmlClipboardToText('Line 1<br>Line 2<br/>Line 3')).toBe('Line 1\nLine 2\nLine 3');
+  });
+
+  it('preserves list-item line breaks', () => {
+    const html = '<ul><li>First item with <a href="https://a.com">A</a></li><li>Second item</li></ul>';
+    expect(htmlClipboardToText(html)).toBe('First item with [A](https://a.com)\nSecond item');
+  });
+
+  it('collapses runs of 3+ newlines to 2', () => {
+    expect(htmlClipboardToText('<p>A</p><br><br><br><br><p>B</p>')).toBe('A\n\nB');
+  });
+
+  it('strips remaining tags but keeps surrounding text', () => {
+    expect(htmlClipboardToText('<span style="color:red">red text</span>')).toBe('red text');
+  });
+
+  it('handles a Notion-style mixed paragraph with a link and inline formatting', () => {
+    const html = '<p>Check out the <a href="https://example.com/post">new post</a> — <strong>big</strong> update!</p>';
+    expect(htmlClipboardToText(html)).toBe('Check out the [new post](https://example.com/post) — big update!');
+  });
+});
--- a/packages/shared/src/utils/html-utils.spec.ts
+++ b/packages/shared/src/utils/html-utils.spec.ts
@@ -65,6 +65,23 @@ describe('htmlClipboardToText', () => {
     expect(htmlClipboardToText('<a href="https://x.com">cats &amp; dogs</a>')).toBe('[cats & dogs](https://x.com)');
   });
 
+  it('decodes entities inside href so query-string URLs survive', () => {
+    expect(htmlClipboardToText('<a href="https://example.com?a=1&amp;b=2">link</a>')).toBe('[link](https://example.com?a=1&b=2)');
+  });
+
+  it('collapses to a bare URL when text and href differ only by entity encoding', () => {
+    const html = '<a href="https://example.com?a=1&amp;b=2">https://example.com?a=1&amp;b=2</a>';
+    expect(htmlClipboardToText(html)).toBe('https://example.com?a=1&b=2');
+  });
+
+  it('decodes decimal numeric entities (em dash, NBSP, smart quote)', () => {
+    expect(htmlClipboardToText('hello&#8212;world&#160;&#8217;tis')).toBe("hello—world ’tis");
+  });
+
+  it('decodes hexadecimal numeric entities', () => {
+    expect(htmlClipboardToText('em dash &#x2014; and ndash &#x2013;')).toBe('em dash — and ndash –');
+  });
+
   it('turns block boundaries into newlines', () => {
     expect(htmlClipboardToText('<p>Hello</p><p>World</p>')).toBe('Hello\nWorld');
   });
--- a/packages/shared/src/utils/html-utils.ts
+++ b/packages/shared/src/utils/html-utils.ts
@@ -1,45 +1,52 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
+const NAMED_HTML_ENTITIES: Record<string, string> = {
+  nbsp: ' ',
+  amp: '&',
+  lt: '<',
+  gt: '>',
+  quot: '"',
+  apos: "'",
+};
+
 /**
  * Decodes a small set of named HTML entities plus all numeric ones (decimal
- * and hex) into their character equivalents. Pure string ops — SSR-safe.
- *
- * Intended to be called once per value; callers that need multiple passes
- * should fold their pipeline so this runs as the final step.
+ * and hex) in a single regex pass. Single-pass is load-bearing: chaining
+ * replacements would let a decoded `&` get re-interpreted as the start of a
+ * fresh entity (e.g., `&amp;#39;` → `&#39;` → `'`), which is the
+ * double-unescape pattern CodeQL flags. Pure string ops — SSR-safe.
  */
 function decodeHtmlEntities(s: string): string {
-  return (
-    s
-      .replace(/&nbsp;/g, ' ')
-      .replace(/&amp;/g, '&')
-      .replace(/&lt;/g, '<')
-      .replace(/&gt;/g, '>')
-      .replace(/&quot;/g, '"')
-      .replace(/&#39;/g, "'")
-      .replace(/&#x27;/g, "'")
-      .replace(/&apos;/g, "'")
-      // Numeric entities — common in clipboard HTML from Word, Google Docs, Notion
-      // (em dash &#8212;, NBSP &#160;, smart quotes, etc.).
-      .replace(/&#(\d+);/g, (_match, n: string) => String.fromCodePoint(Number(n)))
-      .replace(/&#x([\da-fA-F]+);/g, (_match, h: string) => String.fromCodePoint(parseInt(h, 16)))
-  );
+  return s.replace(/&(#\d+|#x[\da-fA-F]+|[a-z]+);/gi, (match, body: string) => {
+    const lower = body.toLowerCase();
+    if (lower.startsWith('#x')) {
+      const code = parseInt(lower.slice(2), 16);
+      return Number.isFinite(code) ? String.fromCodePoint(code) : match;
+    }
+    if (lower.startsWith('#')) {
+      const code = Number(lower.slice(1));
+      return Number.isFinite(code) ? String.fromCodePoint(code) : match;
+    }
+    return NAMED_HTML_ENTITIES[lower] ?? match;
+  });
 }
 
 /**
- * Strips HTML-like tag sequences until the string is stable. Looping is
- * required because a single regex pass can leave behind a partial tag when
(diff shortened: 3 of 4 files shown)


packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@ export function stripHtml(html: string | null | undefined): string {
...
+ * ```
+ */
+export function htmlClipboardToText(html: string | null | undefined): string {
+  if (!html) return '';
+
+  let result = html;
+
+  // Rewrite anchors first — must happen before tag stripping so we can read href.
+  result = result.replace(/<a\b[^>]*?\bhref\s*=\s*("([^"]*)"|'([^']*)')[^>]*>([\s\S]*?)<\/a>/gi, (_match, _quotedHref, hrefDouble, hrefSingle, inner) => {
+    const href = (hrefDouble ?? hrefSingle ?? '').trim();
```
MRashad26: **Bug: `href` attribute value is not entity-decoded**

The captured `href` is taken verbatim from the HTML source. Browser-copied HTML encodes `&` in attribute values as `&amp;`, so a URL like `https://example.com?a=1&b=2` arrives as `href="https://example.com?a=1&amp;b=2"`. This code produces `[text](https://example.com?a=1&amp;b=2)` — the `&amp;` survives into the Markdown link and breaks the URL downstream.

Fix: run `decodeHtmlEntities` on `href` too:

```ts
const href = decodeHtmlEntities((hrefDouble ?? hrefSingle ?? '').trim());
```

Suggested test:
```ts
it('decodes entities in href', () => {
  expect(htmlClipboardToText('<a href="https://x.com?a=1&amp;b=2">link</a>'))
    .toBe('[link](https://x.com?a=1&b=2)');
});
```

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@ export function stripHtml(html: string | null | undefined): string {
...
+  // Decode entities.
+  result = decodeHtmlEntities(result);
+
+  // Collapse runs of 3+ newlines down to 2.
+  result = result.replace(/\n{3,}/g, '\n\n');
+
+  return result.trim();
+}
+
+function decodeHtmlEntities(s: string): string {
```
MRashad26: **Limitation: numeric HTML entities are not decoded**

`decodeHtmlEntities` only handles seven named entities. Clipboard HTML from Word, Google Docs, Notion, and similar editors routinely contains numeric entities: `&#8212;` (em dash), `&#160;` (non-breaking space), `&#8217;` (right single quote), etc. These survive as literal `&#NNN;` strings in the text the AI generator receives.

A pure-string, SSR-safe extension:

```ts
.replace(/&#(\d+);/g, (_, n) => String.fromCodePoint(Number(n)))
.replace(/&#x([0-9a-f]+);/gi, (_, h) => String.fromCodePoint(parseInt(h, 16)))
```

Append both lines to the existing replace chain.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
```diff
@@ -108,6 +109,32 @@ export class NewsletterGenerateDrawerComponent {
     }
   }
 
+  // Preserve hyperlinks when pasting rich content. A textarea would otherwise drop
+  // hrefs and keep only visible text, hiding URLs from the AI generator downstream.
+  public onRawContentPaste(event: ClipboardEvent): void {
```
MRashad26: **Style: method should be `protected`, not `public`**

Angular 20 templates can access `protected` members, and the component-organization rule places protected methods at position 9 in the class order. `public` unnecessarily widens the component API surface:

```ts
protected onRawContentPaste(event: ClipboardEvent): void {
```

MRashad26 (commented):
**CLAUDE.md:** no violations — `htmlClipboardToText` is correctly placed in `packages/shared/src/utils/`, the template uses an event binding (not a function-call expression), no local interfaces or module-level consts in `apps/lfx-one/`.

Three code-quality findings below.

packages/shared/src/utils/html-utils.ts
dealako: **[minor] `href` is not entity-decoded — breaks URLs with query parameters**

Browser-generated clipboard HTML encodes `&` in attribute values as `&amp;`, so a URL like `https://example.com?a=1&b=2` arrives as `href="https://example.com?a=1&amp;b=2"`. The code emits this verbatim, producing a broken Markdown link: `[text](https://example.com?a=1&amp;b=2)`.

Also confirmed by the MRashad26 and Copilot inline comments — agreed on all counts.

Fix:
```ts
const href = decodeHtmlEntities((hrefDouble ?? hrefSingle ?? '').trim());
```

The `text === href` bare-URL branch also silently fails when text and href differ only by entity encoding, so decoding both before comparing is the correct approach.

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@ export function stripHtml(html: string | null | undefined): string {
...
+ * // Returns: "https://x.com"
+ * ```
+ */
+export function htmlClipboardToText(html: string | null | undefined): string {
+  if (!html) return '';
+
+  let result = html;
+
+  // Rewrite anchors first — must happen before tag stripping so we can read href.
+  result = result.replace(/<a\b[^>]*?\bhref\s*=\s*("([^"]*)"|'([^']*)')[^>]*>([\s\S]*?)<\/a>/gi, (_match, _quotedHref, hrefDouble, hrefSingle, inner) => {
```
dealako: **[minor] Anchor regex may backtrack non-linearly on malformed clipboard HTML**

The pattern `[^>]*?\bhref\s*=\s*("([^"]*)"|\'([^\']*)\')`  can backtrack superlinearly when the input contains `href="` with no closing quote and no `>` — the engine tries every suffix of `[^"]*)`, then tries the single-quote branch, then expands `[^>]*?` to find another `\bhref` candidate. On a 100-char unclosed attribute this is perceptible; on a copied page fragment with many attributes it's slow.

Practical blast radius is client-side only (UI thread stall on paste), but a malicious page can write crafted HTML to the clipboard before the user pastes.

Least-invasive hardening — drop the outer `[^>]*?` and instead parse href via a second regex once the full tag is captured:
```ts
result = result.replace(/<a\b[^>]*?>[\s\S]*?<\/a>/gi, (match) => {
  const hrefMatch = /\bhref\s*=\s*(?:"([^"]*)"|'([^']*)')/i.exec(match);
  const href = decodeHtmlEntities((hrefMatch?.[1] ?? hrefMatch?.[2] ?? '').trim());
  const inner = match.replace(/<[^>]*>/g, '');
  const text = decodeHtmlEntities(inner).trim();
  if (!href) return text;
  if (!text || text === href) return href;
  return `[${text}](${href})`;
});
```
This inner regex is bounded by the already-captured `<a...>` tag string so backtracking is contained.

packages/shared/src/utils/html-utils.ts
dealako: **[nit] Numeric HTML entities are not decoded**

Clipboard HTML from Google Docs, Notion, and Word routinely contains `&#8212;` (em dash), `&#160;` (NBSP), `&#8217;` (right single quote), `&#x2014;` (hex em dash), etc. These survive as literal `&#NNN;` strings in the textarea.

Adding two lines to the chain is sufficient:
```ts
.replace(/&#(\d+);/g, (_, n) => String.fromCodePoint(Number(n)))
.replace(/&#x([\da-fA-F]+);/g, (_, h) => String.fromCodePoint(parseInt(h, 16)))
```

Also flagged by Copilot and MRashad26 — agreed.

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@ export function stripHtml(html: string | null | undefined): string {
...
+    if (!text || text === href) return href;
+    return `[${text}](${href})`;
+  });
+
+  // Block-level boundaries → newlines (before stripping remaining tags).
+  result = result.replace(/<br\s*\/?>/gi, '\n');
+  result = result.replace(/<\/(p|div|li|tr|h[1-6])>/gi, '\n');
+
+  // Strip remaining tags.
+  result = result.replace(/<[^>]*>/g, '');
```
dealako: **[nit] CodeQL alerts on lines 71, 82, 96 are false positives in context — worth suppressing**

GitHub Advanced Security flagged three findings:
- Lines 71, 82: "Incomplete multi-character sanitization" — `&lt;script&gt;` in the input will be decoded to `<script>` after `decodeHtmlEntities()`. This would be an XSS risk if the output were inserted as `innerHTML`, but here it flows into `textarea.value` / `form.setValue()` (plain text, never parsed as HTML), so no injection is possible.
- Line 96: "Double escaping/unescaping" — `decodeHtmlEntities` is called twice (once on anchor inner text, once on the full result). The first call replaces `&amp;` → `&`; the second call sees `&` but no `&amp;` pattern, so no double-decode occurs.

To silence the scanner without changing behavior, either add `// lgtm[js/incomplete-multi-character-sanitization]` suppressions or add a comment explaining the safe output destination.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
```diff
@@ -108,6 +109,32 @@ export class NewsletterGenerateDrawerComponent {
     }
   }
 
+  // Preserve hyperlinks when pasting rich content. A textarea would otherwise drop
```
dealako: **[nit] `public` should be `protected`**

Angular 20 templates bind to `protected` members without issue; `public` unnecessarily widens the component API surface. Template-only event handlers follow the `protected` convention per the component-organization rules.

Also flagged by MRashad26 — agreed.

```ts
protected onRawContentPaste(event: ClipboardEvent): void {
```

apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
```diff
@@ -108,6 +109,32 @@ export class NewsletterGenerateDrawerComponent {
...
+
+    const textarea = event.target as HTMLTextAreaElement | null;
+    if (!textarea) return;
+
+    event.preventDefault();
+
+    const converted = htmlClipboardToText(html);
+    const start = textarea.selectionStart ?? textarea.value.length;
+    const end = textarea.selectionEnd ?? textarea.value.length;
+    const current = textarea.value;
```
dealako: **[nit] `setValue` does not mark the control dirty or touched**

Angular's default value accessor marks the control `dirty` and `touched` on user interaction. `setValue()` bypasses that, so if the form shows validation errors only after `touched` (e.g. a `minlength` hint on the textarea), a paste-only session won't surface them.

Consider adding:
```ts
this.form.controls.rawContent.setValue(next);
this.form.controls.rawContent.markAsDirty();
this.form.controls.rawContent.markAsTouched();
```
or verify that the existing `canGenerate()` guard makes this moot.

packages/shared/src/utils/html-utils.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    expect(htmlClipboardToText('<a class="link" data-foo="bar" href="https://x.com">x</a>')).toBe('[x](https://x.com)');
+  });
+
+  it('decodes entities in plain text', () => {
+    expect(htmlClipboardToText('cats &amp; dogs')).toBe('cats & dogs');
+  });
+
+  it('decodes entities inside anchor text', () => {
+    expect(htmlClipboardToText('<a href="https://x.com">cats &amp; dogs</a>')).toBe('[cats & dogs](https://x.com)');
+  });
```
dealako, line 66: **[nit] Missing test: entity-encoded `href` value**

Copilot flagged this — agreed. The most common real-world clipboard HTML shape that breaks without the `href` entity-decode fix:

```ts
it('decodes entities inside href so query-string URLs survive', () => {
  const html = '<a href="https://example.com?a=1&amp;b=2">link</a>';
  expect(htmlClipboardToText(html)).toBe('[link](https://example.com?a=1&b=2)');
});
```

Add this alongside the `decodeHtmlEntities` fix so regressions are caught.

dealako (commented):
Hey @niravpatel27 — clean fix for a subtle UX data-loss bug. The intent is clear, the implementation is well-structured, and the test suite is thorough with 18 scenarios covering anchors, entities, block boundaries, and edge cases. Two issues worth addressing before or shortly after merge, plus a few nits.

---

## Step 1 — Independent Review

### Correctness

**`href` entity decoding (functional bug)** — see inline comment on `html-utils.ts:70`. Browser-generated clipboard HTML encodes `&` in attribute values as `&amp;`, so a URL like `https://example.com?a=1&b=2` arrives as `href="https://example.com?a=1&amp;b=2"`. The code emits this verbatim, producing a Markdown link with a broken URL. Fix: wrap the captured `href` in `decodeHtmlEntities()`.

**`queueMicrotask` caret restoration** — the deferred `setSelectionRange` correctly handles the Angular change detection cycle. No issue.

**Fallback path (no `text/html` in clipboard)** — if `getData('text/html')` returns empty string, the handler returns early and the native paste fires normally. Good degradation.

### Security

Consolidated from the security subagent findings:

1. **Regex backtracking on malformed HTML** — see inline comment on `html-utils.ts:69`. The anchor regex can backtrack non-linearly when clipboard HTML contains an unclosed `href="...` attribute with no closing `>`. Worst-case: client-side UI thread stall on a crafted paste payload. The security subagent rated this HIGH; in context (user-controlled paste gesture, client-side only, no server impact) I classify it **[minor]** but recommend hardening before merge. A bounded two-regex approach is shown in the inline comment.

2. **CodeQL alerts (lines 71, 82, 96)** — see inline comment on `html-utils.ts:82`. All three are false positives: decoded `<script>` content lands in `textarea.value` (plain text), not `innerHTML`, so no XSS vector exists. However, they will appear as scanner findings and may block CI. Adding `// lgtm` suppressions or a brief explanatory comment will silence them without code changes.

3. **XSS: not a risk** — the output destination is `textarea.value` / `form.setValue()`, always treated as plain text by the browser. Confirmed safe by the security subagent.

### Performance

No concerns — pure string operations, no DOM API calls in `htmlClipboardToText` (correctly SSR-safe), `queueMicrotask` defers the non-critical caret restoration off the main tick.

### Test Coverage

Strong. 18 test cases cover the main paths. One gap: no test for `&amp;`-encoded `href` values — see inline comment on `html-utils.spec.ts:66`. This would have caught the `href` entity decoding bug if written first.

### Code Style & Consistency

- `onRawContentPaste` should be `protected` (see inline comment on component ts:112)
- `setValue()` without `markAsDirty()`/`markAsTouched()` — see inline comment on component ts:126
- Numeric HTML entities (`&#8212;`, `&#160;`, etc.) not decoded — see inline comment on `html-utils.ts:93`

---

## Step 2 — AI Comment Reconciliation

| Bot | Finding | My Position |
|-----|---------|-------------|
| **GitHub Advanced Security** (lines 71, 82, 96) | Incomplete sanitization / double-unescaping | Agree these are scanner findings that need suppression; disagree they're real security risks in this output context — false positives |
| **Copilot** — `href` entity decoding | Bug: `&amp;` in query params survives into Markdown | ✅ Agree — real functional bug, same finding as mine |
| **Copilot** — numeric entities | `&#NNN;` survives as literal text | ✅ Agree — nit but worth fixing |
| **Copilot** — unchecked `HTMLTextAreaElement` cast | Could throw if target isn't `<textarea>` | Partially agree. The null guard (`if (!textarea) return`) exists but doesn't protect against a non-null non-textarea target. In practice the paste event on `<lfx-textarea>`'s inner element will always target the native textarea — low real-world risk but the `instanceof HTMLTextAreaElement` check would make intent explicit |
| **Copilot** — `setValue` dirty/touched | Control won't show validation errors after paste | ✅ Agree — nit, depends on form validation requirements |
| **Copilot** — missing `href` entity decode test | | ✅ Agree |
| **MRashad26** — `href` entity decoding | Same bug, same fix | ✅ Agree |
| **MRashad26** — numeric entities | Same gap | ✅ Agree |
| **MRashad26** — `public` → `protected` | Angular convention | ✅ Agree |

---

## Step 3 — Revision Tracking

First review round — no prior iterations to track.

---

## Step 4 — Summary

🔴 **Blocking:** 0 issues

🟡 **Minor:** 2 issues
- `href` not entity-decoded — breaks URLs containing `&` in query params (functional bug with a one-line fix)
- Anchor regex backtracking risk on malformed clipboard HTML (client-side UI stall vector; recommend hardening)

⚪ **Nit:** 5 issues
- CodeQL false-positive alerts need suppression comments (may block CI checks)
- `onRawContentPaste` should be `protected`, not `public`
- Numeric HTML entities (`&#NNN;`) not decoded
- `setValue()` without `markAsDirty()`/`markAsTouched()`
- Missing test for `&amp;`-encoded href values

---

The PR solves a real data-loss problem with clean utility design, SSR safety, and solid test coverage. The `href` entity-decoding bug and the regex backtracking issue should both be addressed before or immediately after merge — neither blocks the feature from landing, but the href bug will silently corrupt links from any source that HTML-encodes query-string `&` characters (which is essentially all of them).

dealako (dismissed):
No blocking issues — approving. The two minor items (href entity decoding, regex backtracking hardening) and the CodeQL suppressions are worth addressing in a follow-up commit or before merge per team policy.

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@ export function stripHtml(html: string | null | undefined): string {
...
+  let result = html;
+
+  // Rewrite anchors first — must happen before tag stripping so we can read href.
+  result = result.replace(/<a\b[^>]*?\bhref\s*=\s*("([^"]*)"|'([^']*)')[^>]*>([\s\S]*?)<\/a>/gi, (_match, _quotedHref, hrefDouble, hrefSingle, inner) => {
+    const href = (hrefDouble ?? hrefSingle ?? '').trim();
+    const text = decodeHtmlEntities(inner.replace(/<[^>]*>/g, '')).trim();
+    if (!href) return text;
+    if (!text || text === href) return href;
+    return `[${text}](${href})`;
+  });
```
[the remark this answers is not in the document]
niravpatel27: Fixed in 1d8118e9 — `href` is now passed through `decodeHtmlEntities` before the `text === href` comparison, so `?a=1&amp;b=2` survives as `?a=1&b=2` in both the Markdown link and the bare-URL collapse path. Added two regression tests for entity-encoded href and bare-URL collapse on entity-encoded matches.

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@ export function stripHtml(html: string | null | undefined): string {
...
+  return s
+    .replace(/&nbsp;/g, ' ')
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
+    .replace(/&#39;/g, "'")
+    .replace(/&#x27;/g, "'")
+    .replace(/&apos;/g, "'");
+}
```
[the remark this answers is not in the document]
niravpatel27: Fixed in 1d8118e9 — `decodeHtmlEntities` now decodes decimal (`&#NNN;`) and hex (`&#xHEX;`) numeric entities via `String.fromCodePoint`. Em dash / NBSP / smart quotes from Word/Docs/Notion clipboards now flow through as actual characters. Two new tests cover the decimal and hex paths.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
```diff
@@ -108,6 +109,32 @@ export class NewsletterGenerateDrawerComponent {
...
+    const next = current.slice(0, start) + converted + current.slice(end);
+
+    this.form.controls.rawContent.setValue(next);
+
```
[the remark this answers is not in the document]
niravpatel27: Fixed in 1d8118e9 — replaced the unchecked cast with an `event.target instanceof HTMLTextAreaElement` guard (explicit intent + no throw if a future wrapper changes the bubble target). Also added `control.markAsDirty()` + `control.markAsTouched()` after `setValue` so paste matches manual-typing semantics for validation gating.

packages/shared/src/utils/html-utils.spec.ts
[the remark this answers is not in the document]
niravpatel27: Added in 1d8118e9 — two new tests in html-utils.spec.ts: `decodes entities inside href so query-string URLs survive` and `collapses to a bare URL when text and href differ only by entity encoding`. Both pass.

packages/shared/src/utils/html-utils.ts
niravpatel27: Fixed in 1d8118e9 — wrapped the captured `href` in `decodeHtmlEntities` exactly as you suggested. The `text === href` bare-URL collapse now also works for entity-encoded URLs (one of the new regression tests covers this). Thanks for the catch + suggested test scaffolding.

packages/shared/src/utils/html-utils.ts
niravpatel27: Fixed in 1d8118e9 — added both decimal (`&#NNN;`) and hex (`&#xHEX;`) decoders to `decodeHtmlEntities` via `String.fromCodePoint`. Two tests cover em dash / NBSP / smart quote and hex em-dash / en-dash.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
niravpatel27: Fixed in 1d8118e9 — changed to `protected` and moved below `onGenerate` to satisfy the member-ordering eslint rule (public methods before protected).

packages/shared/src/utils/html-utils.ts
niravpatel27: Fixed in 1d8118e9 — `href` is now passed through `decodeHtmlEntities` before both the Markdown emit and the `text === href` comparison, so URLs with entity-encoded query params survive end-to-end. Two new tests cover the encoded-href and bare-URL-collapse cases.

packages/shared/src/utils/html-utils.ts
niravpatel27: Fixed in 1d8118e9 — adopted the bounded two-step approach you suggested. The outer regex now captures the full `<a>...</a>` tag, and href extraction runs on the already-captured tag string via a separate non-global regex. Combined with the entity-decode fix, this lands both the perf hardening and the correctness fix in one place.

packages/shared/src/utils/html-utils.ts
niravpatel27: Fixed in 1d8118e9 — added both decimal and hex numeric entity decoders to `decodeHtmlEntities`. Two tests cover both encodings.

packages/shared/src/utils/html-utils.ts
niravpatel27: Addressed in 1d8118e9 — added a paragraph to the `htmlClipboardToText` JSDoc explaining the output destination (`textarea.value` / `form.setValue`, never `innerHTML`) so the CodeQL alerts on these lines are documented as false positives in context. Opted for a comment over `// lgtm` annotations to keep the reasoning visible to human readers. If you'd prefer the explicit suppressions to actually silence the scanner, happy to add them in a follow-up.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
niravpatel27: Fixed in 1d8118e9 — changed to `protected` and moved below `onGenerate` to satisfy `@typescript-eslint/member-ordering`.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-generate-drawer/newsletter-generate-drawer.component.ts
niravpatel27: Fixed in 1d8118e9 — added `markAsDirty()` + `markAsTouched()` after `setValue` so paste-only sessions surface validation state the same way manual typing does. `canGenerate()` gates submission on length but the dirty/touched flags matter for error display.

packages/shared/src/utils/html-utils.spec.ts
niravpatel27: Added in 1d8118e9 — the test you scaffolded is in place (`decodes entities inside href so query-string URLs survive`), plus a sibling case for the bare-URL collapse when text and href differ only by entity encoding.

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@
...
+ */
+export function htmlClipboardToText(html: string | null | undefined): string {
+  if (!html) return '';
+
+  let result = html;
+
+  // Rewrite anchors first — must happen before tag stripping so we can read href.
+  result = result.replace(/<a\b[^>]*?\bhref\s*=\s*("([^"]*)"|'([^']*)')[^>]*>([\s\S]*?)<\/a>/gi, (_match, _quotedHref, hrefDouble, hrefSingle, inner) => {
+    const href = (hrefDouble ?? hrefSingle ?? '').trim();
+    const text = decodeHtmlEntities(inner.replace(/<[^>]*>/g, '')).trim();
```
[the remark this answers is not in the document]
niravpatel27: False positive in this context, but worth documenting clearly. The output of `htmlClipboardToText` flows into `textarea.value` and the reactive `FormControl` via `setValue` — both are plain-text sinks, never `innerHTML`. A decoded `<script>` substring in the output cannot become an HTML element. Added a JSDoc paragraph in commit 1d8118e9 spelling this out so future readers (and other scanners) see the rationale. Resolving.

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@
...
+    if (!text || text === href) return href;
+    return `[${text}](${href})`;
+  });
+
+  // Block-level boundaries → newlines (before stripping remaining tags).
+  result = result.replace(/<br\s*\/?>/gi, '\n');
+  result = result.replace(/<\/(p|div|li|tr|h[1-6])>/gi, '\n');
+
+  // Strip remaining tags.
+  result = result.replace(/<[^>]*>/g, '');
```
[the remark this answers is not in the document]
niravpatel27: False positive in this context — same reasoning as the line-71 alert: output destination is `textarea.value` / `form.setValue` (plain-text sinks), so an incompletely-stripped `<script>` cannot become an HTML element. JSDoc paragraph added in 1d8118e9 documents the safe-output context. Resolving.

packages/shared/src/utils/html-utils.ts
```diff
@@ -37,3 +37,67 @@
...
+  // Collapse runs of 3+ newlines down to 2.
+  result = result.replace(/\n{3,}/g, '\n\n');
+
+  return result.trim();
+}
+
+function decodeHtmlEntities(s: string): string {
+  return s
+    .replace(/&nbsp;/g, ' ')
+    .replace(/&amp;/g, '&')
```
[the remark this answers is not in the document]
niravpatel27: False positive — `decodeHtmlEntities` runs on two disjoint substrings (the inner text of an anchor, then the final post-strip result). It does not run twice on the same characters, so `&amp;` cannot be double-unescaped into a malformed `&` interpretation. The JSDoc paragraph added in 1d8118e9 documents the output context (plain-text textarea sink, never innerHTML), which makes the alert's downstream concern moot regardless. Resolving.

niravpatel27:
## Review Feedback Addressed

Commit: 1d8118e96b15a10d636baea9d867274ff65e5bd7

### Changes Made

- **`html-utils.ts`** — decoded HTML entities in `href` value (the headline bug — `?a=1&amp;b=2` was leaking through; now survives as `?a=1&b=2`) (per @MRashad26, @dealako, copilot-pull-request-reviewer)
- **`html-utils.ts`** — decoded decimal (`&#NNN;`) and hex (`&#xHEX;`) numeric entities via `String.fromCodePoint` (per @MRashad26, @dealako, copilot-pull-request-reviewer)
- **`html-utils.ts`** — rewrote anchor extraction as a bounded two-step regex (capture the full `<a>…</a>` first, then extract href from the captured tag) to eliminate the superlinear backtracking risk on malformed clipboard HTML (per @dealako)
- **`html-utils.ts`** — refactored `stripHtml` to reuse the shared `decodeHtmlEntities` helper, removing the duplicated entity-decoding chain (per coderabbitai)
- **`html-utils.ts`** — added a JSDoc paragraph spelling out the output destination (`textarea.value` / `form.setValue`, never `innerHTML`), which documents why the CodeQL alerts on lines 71/82/96 are false positives in context (per @dealako)
- **`newsletter-generate-drawer.component.ts`** — `public onRawContentPaste` → `protected`, moved below `onGenerate` to satisfy `@typescript-eslint/member-ordering` (per @MRashad26, @dealako)
- **`newsletter-generate-drawer.component.ts`** — replaced the unchecked `as HTMLTextAreaElement` cast with an `instanceof HTMLTextAreaElement` guard for explicit intent (per copilot-pull-request-reviewer)
- **`newsletter-generate-drawer.component.ts`** — added `markAsDirty()` + `markAsTouched()` after `setValue` so paste matches manual-typing semantics for validation gating (per copilot-pull-request-reviewer, @dealako)
- **`html-utils.spec.ts`** — added regression tests: entity-encoded href, bare-URL collapse on entity-encoded matches, decimal numeric entities, hex numeric entities (per copilot-pull-request-reviewer, @dealako)

### No Change Needed

- **`html-utils.ts:71`**, **`:82`**, **`:96`** — three CodeQL alerts ("incomplete multi-character sanitization", "double escaping"). Replied on each thread explaining the output destination (plain-text textarea / FormControl, never innerHTML) makes the alerts false positives in context. The new JSDoc paragraph documents this in code so the reasoning is visible to future readers and scanners. If team policy prefers explicit `// lgtm[...]` suppressions to actually silence the scanner badges, happy to add them in a follow-up. (flagged by github-advanced-security)

### Threads Resolved

17 of 17 review threads addressed in this iteration.

### Validation

- `yarn test` (shared) — 23/23 passing (4 new regression tests added)
- `yarn lint` — clean (one pre-existing unrelated warning in `docs-manifest.service.ts`)
- `yarn build` — succeeded

MRashad26 (approved):

PR #866, opened 2026-06-03
feat(org-people): enable edit and bulk reassign on key contacts tab

ahmedomosanya:
## Summary

Two writer affordances, scoped differently, both gated client-side on `OrgRoleGrantsService.writerSet()` (BFF write proxies still re-enforce on every mutation):

- **Expanded-row Edit pencil** — opens the spec-024 4-state modal (chooser / single-add / replace / remove) scoped to ONE (active membership, role-TYPE) tuple. Reuses every server-side primitive from the Membership Detail tab: same write proxy, same min/max enforcement from `KEY_CONTACT_ROLE_CATALOG`, same employee typeahead, same optimistic-concurrency error mapping.
- **Main-row Reassign Key Contact Roles pencil** — opens a dedicated bulk-reassign modal scoped to ONE PERSON across N (membership, role-TYPE) tuples. Lists every role this person currently holds with checkboxes (Select-all preselected), live `X roles across Y foundations` subtitle, `Assign to` typeahead pre-excluding the current contact's email, fan-out save (one parallel PUT per checked role) with pessimistic UI — modal stays open during the call, surfaces a retryable inline error if any of the writes reject, refreshes the table on completion regardless.

New BFF surface: slug-keyed catalog GET + write proxies (POST / PUT / DELETE) on `OrgLensKeyContactsController`. The People tab carries the foundation slug per assignment row, so these endpoints skip the org `sfid → foundation_id` bridge that the membership-detail page needs.

Wire-level cleanup: the `canEdit` / `canEditAny` fields on `OrgKeyContactsResponse` always equalled a single org-level boolean and are dropped — the gate is now computed entirely client-side, mirroring the membership-detail page.

Mid-flight account switching: an `accountCancel$` subject piped off `orgUid$.pipe(skip(1))` cancels in-flight catalog fetches and closes any open edit / reassign dialog when the user changes account, so the user can't confirm a write against a previous org's scope.

Until LFXV2-1677 lands the canonical organisation-employees data source, the `Assign to` typeahead is populated from the same transitional corpus the spec-024 modal already uses (the org's existing `key_contact` records, deduped by email). When the canonical source ships, the BFF widens the corpus transparently — no UI change.

- **LFXV2-1677** (canonical organisation-employees feed). This PR consumes whichever corpus is wired at runtime.
- **Read-path changes** to People → Key Contacts — already shipped under LFXV2-1873.
- **Voting-rep → dashboard-access downgrade** side effect — legacy org-dashboard behaviour, handled downstream by `fga-sync`.

### Manual QA

- [ ] Expanded-row pencil — single-slot role with one current contact: chooser does NOT open; modal seeds directly into Replace state on the existing holder.
- [ ] Expanded-row pencil — multi-slot role with multiple contacts: chooser opens with the clicked person preselected.
- [ ] Expanded-row pencil — Add: empty multi-slot role accepts a new contact; table refreshes with the new row.
- [ ] Expanded-row pencil — Replace: existing holder swapped for a new email; success toast confirms.
- [ ] Expanded-row pencil — Remove: existing holder removed when role count > minContacts; remove blocked when at minContacts.
- [ ] Main-row pencil — opens with `Select all` preselected; subtitle reads `X roles across Y foundations` matching the count.
- [ ] Main-row pencil — toggling individual checkboxes updates the subtitle and the `Save Changes (X roles)` label live.
- [ ] Main-row pencil — `Assign to` typeahead suggests existing org key contacts; current contact's email excluded; selecting prefills first / last name.
- [ ] Main-row pencil — happy path: full-success Save closes the modal, success toast fires, table reflects the new contact across all checked roles.
- [ ] Main-row pencil — partial failure: Save with some PUTs failing keeps the modal open with a retryable inline error and refreshes the table to show the partial state.
- [ ] Main-row pencil — full failure: Save with all PUTs failing keeps the modal open with the upstream error message inline.
- [ ] Read-only persona / impersonation: pencil affordances suppressed (both expanded-row and main-row).
- [ ] Account switch mid-flight: open the edit modal, then change account in the org selector; the modal closes and the in-flight catalog GET is cancelled.

### Automated

- [ ] `apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts` — 5 tests cover the 18 user-perspective acceptance criteria for the main-row Reassign modal (anatomy + counts, reactive subtitle / save label, employee typeahead prefill, success lifecycle, error lifecycle).
- [ ] `yarn check-types`, `yarn lint:check`, `yarn format:check` — all pass locally.
- [ ] Full angular build passes (verified locally at ~53s).

### DIFF
--- a/apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts
+++ b/apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts
@@ -9,8 +9,8 @@ const PEOPLE_KEY_CONTACTS_URL = '/org/people?tab=contacts';
 const DATA_LOAD_TIMEOUT = 30_000;
 const TOAST_TIMEOUT = 10_000;
 
-const MOCK_UID = '4c46585f-878c-8285-b2e9-2dbfc38ddd9b';
 const MOCK_ACCOUNT_ID = '0014100000Te2QjAAJ';
+const MOCK_UID = MOCK_ACCOUNT_ID;
 const MOCK_ACCOUNT_NAME = 'Acme Industries';
 const MOCK_ACCOUNT_SLUG = 'acme-industries';
 
--- a/apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts
+++ b/apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts
@@ -1,25 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-/**
- * Org People — Reassign Key Contact Roles E2E (LFXV2-2067 main-row pencil).
- *
- * Covers the user-perspective acceptance criteria for the bulk-reassign flow scoped to ONE PERSON
- * across N (membership, role-TYPE) tuples. Out of scope here: the expanded-row spec-024 4-state
- * modal, which is exercised by `org-membership-key-contacts.spec.ts`.
- *
- * Coverage map (18 user-perspective points):
- *   1–14, 17 baseline → "modal opens with the expected anatomy"
- *   17 reactive label → "save button label and subtitle update as roles toggle"
- *   15, 16            → "employee typeahead populates the name fields on selection"
- *   18 success+loading → "save fans out PUTs, shows the spinner, closes on full success"
- *   18 error           → "save failure keeps the modal open with a retryable inline error"
- *
- * Prerequisites:
- * - Dev server reachable at the Playwright baseURL (default http://localhost:4200)
- * - `apps/lfx-one/.env` populated with TEST_USERNAME / TEST_PASSWORD
- * - `org-lens-enabled` LaunchDarkly flag toggled ON for the test user
- */
+/** Reassign Key Contact Roles modal E2E (LFXV2-2067 main-row pencil). */
 
 import { expect, Page, Route, test } from '@playwright/test';
 
@@ -108,10 +90,7 @@ const MOCK_EMPLOYEES = [
 
 test.setTimeout(120_000);
 
-/**
- * Bail out gracefully if Auth0 redirected the test (no TEST_USERNAME/TEST_PASSWORD set).
- * Mirrors the helper used by org-profile.spec.ts.
- */
+/** Skip the test when Auth0 redirected (no TEST_USERNAME/TEST_PASSWORD configured). */
 function skipWhenAuthMissing(page: Page): void {
   try {
     const { hostname } = new URL(page.url());
@@ -123,10 +102,7 @@ function skipWhenAuthMissing(page: Page): void {
   }
 }
 
-/**
- * Persona + role-grants stubs so `selectedAccount().uid` resolves and `canEdit()` returns true.
- * Mirrors `stubOrgProfileContext` from org-profile.spec.ts but scoped to this fixture's org.
- */
+/** Stubs persona + role-grants so `selectedAccount().uid` resolves and `canEdit()` returns true. */
 async function stubAccountContext(page: Page, opts: { writers: string[] } = { writers: [MOCK_UID] }): Promise<void> {
   await page.route('**/api/user/personas*', (route) =>
     route.fulfill({
@@ -179,10 +155,7 @@ async function stubEmployees(page: Page, employees: unknown[] = MOCK_EMPLOYEES, 
   });
 }
 
-/**
- * Stub the slug-keyed PUT proxy. The handler decides per-request what to return so individual tests
- * can compose loading-delays, partial failures, etc. Each route listener installs once per test.
- */
+/** Stubs the slug-keyed PUT proxy; the handler decides what to return per request. */
 async function stubReassignPut(page: Page, handler: (route: Route) => Promise<void> | void): Promise<void> {
   await page.route(/\/api\/orgs\/[^/]+\/lens\/key-contacts\/membership\/[^/]+\/[^/]+$/, async (route) => {
     if (route.request().method() !== 'PUT') return route.fallback();
@@ -356,7 +329,15 @@ test.describe('Org People → Key Contacts — Reassign Key Contact Roles modal 
             minContacts: 0,
             maxContacts: 10,
             people: [
-              { personId: 'kc-replacement-id', firstName: 'Cara', lastName: 'Dev', fullName: 'Cara Dev', email: MOCK_REPLACEMENT_EMAIL, jobTitle: null, initials: 'CD' },
+              {
+                personId: 'kc-replacement-id',
+                firstName: 'Cara',
+                lastName: 'Dev',
+                fullName: 'Cara Dev',
+                email: MOCK_REPLACEMENT_EMAIL,
+                jobTitle: null,
+                initials: 'CD',
+              },
             ],
           },
         }),
--- a/apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts
+++ b/apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts
@@ -1,0 +1,429 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+/**
+ * Org People — Reassign Key Contact Roles E2E (LFXV2-2067 main-row pencil).
+ *
+ * Covers the user-perspective acceptance criteria for the bulk-reassign flow scoped to ONE PERSON
+ * across N (membership, role-TYPE) tuples. Out of scope here: the expanded-row spec-024 4-state
+ * modal, which is exercised by `org-membership-key-contacts.spec.ts`.
+ *
+ * Coverage map (18 user-perspective points):
+ *   1–14, 17 baseline → "modal opens with the expected anatomy"
+ *   17 reactive label → "save button label and subtitle update as roles toggle"
+ *   15, 16            → "employee typeahead populates the name fields on selection"
+ *   18 success+loading → "save fans out PUTs, shows the spinner, closes on full success"
+ *   18 error           → "save failure keeps the modal open with a retryable inline error"
+ *
+ * Prerequisites:
+ * - Dev server reachable at the Playwright baseURL (default http://localhost:4200)
+ * - `apps/lfx-one/.env` populated with TEST_USERNAME / TEST_PASSWORD
+ * - `org-lens-enabled` LaunchDarkly flag toggled ON for the test user
+ */
+
+import { expect, Page, Route, test } from '@playwright/test';
+
+const PEOPLE_KEY_CONTACTS_URL = '/org/people?tab=contacts';
+const DATA_LOAD_TIMEOUT = 30_000;
+const TOAST_TIMEOUT = 10_000;
+
+const MOCK_UID = '4c46585f-878c-8285-b2e9-2dbfc38ddd9b';
+const MOCK_ACCOUNT_ID = '0014100000Te2QjAAJ';
+const MOCK_ACCOUNT_NAME = 'Acme Industries';
+const MOCK_ACCOUNT_SLUG = 'acme-industries';
+
+// Three roles across two foundations — gives concrete numbers for the subtitle and Save Changes label.
+const MOCK_PERSON_EMAIL = 'ada.tester@example.com';
+const MOCK_REPLACEMENT_EMAIL = 'cara.dev@example.com';
+
+// Stable role keys: `${membershipUid}:${contactType}` (mirrors `buildReassignRoleOptions` in the component).
+const ROLE_KEY_MARKETING_CNCF = 'msp-cncf:marketing';
+const ROLE_KEY_TECHNICAL_CNCF = 'msp-cncf:technical';
+const ROLE_KEY_LEGAL_EBPF = 'msp-ebpf:legal';
+
+const MOCK_KEY_CONTACTS_RESPONSE = {
+  assignments: [
+    {
+      contactUid: 'kc-marketing-cncf',
+      membershipUid: 'msp-cncf',
+      email: MOCK_PERSON_EMAIL,
+      firstName: 'Ada',
+      lastName: 'Tester',
+      displayName: 'Ada Tester',
+      title: 'Director, Open Source',
+      role: 'Marketing Contact',
+      foundationSlug: 'cncf',
+      foundationName: 'Cloud Native Computing Foundation',
+    },
+    {
+      contactUid: 'kc-technical-cncf',
+      membershipUid: 'msp-cncf',
+      email: MOCK_PERSON_EMAIL,
+      firstName: 'Ada',
+      lastName: 'Tester',
+      displayName: 'Ada Tester',
+      title: 'Director, Open Source',
+      role: 'Technical Contact',
+      foundationSlug: 'cncf',
+      foundationName: 'Cloud Native Computing Foundation',
+    },
+    {
+      contactUid: 'kc-legal-ebpf',
+      membershipUid: 'msp-ebpf',
+      email: MOCK_PERSON_EMAIL,
+      firstName: 'Ada',
+      lastName: 'Tester',
+      displayName: 'Ada Tester',
+      title: 'Director, Open Source',
+      role: 'Legal Contact',
+      foundationSlug: 'ebpf',
+      foundationName: 'eBPF Foundation',
+    },
+  ],
+  stats: {
+    individualCount: 1,
+    foundationsCovered: 2,
+    unfilledRequiredRoleCount: 0,
+  },
+};
+
+const MOCK_EMPLOYEES = [
+  {
+    email: MOCK_REPLACEMENT_EMAIL,
+    firstName: 'Cara',
+    lastName: 'Dev',
+    fullName: 'Cara Dev',
+    jobTitle: 'Senior Engineer',
+    initials: 'CD',
+  },
+  {
+    email: 'evan.qa@example.com',
+    firstName: 'Evan',
+    lastName: 'QA',
+    fullName: 'Evan QA',
+    jobTitle: 'QA Lead',
+    initials: 'EQ',
+  },
+];
+
+test.setTimeout(120_000);
+
+/**
+ * Bail out gracefully if Auth0 redirected the test (no TEST_USERNAME/TEST_PASSWORD set).
+ * Mirrors the helper used by org-profile.spec.ts.
+ */
+function skipWhenAuthMissing(page: Page): void {
+  try {
+    const { hostname } = new URL(page.url());
+    if (hostname === 'auth0.com' || hostname.endsWith('.auth0.com')) {
+      test.skip(true, 'TEST_USERNAME / TEST_PASSWORD not configured — see global-setup.ts');
+    }
+  } catch {
+    // Malformed URL — let the test run and surface a useful failure.
+  }
+}
+
+/**
+ * Persona + role-grants stubs so `selectedAccount().uid` resolves and `canEdit()` returns true.
+ * Mirrors `stubOrgProfileContext` from org-profile.spec.ts but scoped to this fixture's org.
+ */
+async function stubAccountContext(page: Page, opts: { writers: string[] } = { writers: [MOCK_UID] }): Promise<void> {
+  await page.route('**/api/user/personas*', (route) =>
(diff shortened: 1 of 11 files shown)


apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/components/reassign-key-contact-roles-modal.component.html
```diff
@@ -0,0 +1,215 @@
...
+        @for (role of roles; track role.key) {
+          <div class="flex items-center gap-3 rounded-md border border-gray-200 px-3 py-2" [attr.data-testid]="'reassign-key-contact-role-row-' + role.key">
+            <p-checkbox
+              [binary]="true"
+              [ngModel]="checkedMap[role.key]"
+              (ngModelChange)="toggleRole(role.key)"
+              [disabled]="isSaving()"
+              [attr.data-testid]="'reassign-key-contact-role-checkbox-' + role.key" />
+            <span class="inline-block rounded-full border px-2 py-0.5 text-xs font-medium" [class]="role.pillClass">{{ role.role }}</span>
+            <span class="flex-1 text-sm text-gray-700">&middot; {{ role.foundationName }}</span>
```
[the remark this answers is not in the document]
ahmedomosanya: False positive — `class="..." [class]="..."` is the documented Angular 20 class-merging pattern, not an override. Per the [Angular 20 binding docs](https://angular.dev/guide/templates/binding#css-classes): _"When using static CSS classes, directly binding `class`, and binding specific classes, Angular intelligently combines all of the classes in the rendered result."_ The static `inline-block rounded-full border …` survives, and `role.pillClass` adds the dynamic color classes on top.

Same pattern is already shipped in this branch on the parent template (`apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.html:194,246`) and renders as expected in local Playwright QA, so I'm leaving the modal consistent with the rest of the page.

apps/lfx-one/e2e/org-people-key-contacts-reassign.spec.ts
```diff
@@ -0,0 +1,410 @@
...
+import { expect, Page, Route, test } from '@playwright/test';
+
+const PEOPLE_KEY_CONTACTS_URL = '/org/people?tab=contacts';
+const DATA_LOAD_TIMEOUT = 30_000;
+const TOAST_TIMEOUT = 10_000;
+
+const MOCK_UID = '4c46585f-878c-8285-b2e9-2dbfc38ddd9b';
+const MOCK_ACCOUNT_ID = '0014100000Te2QjAAJ';
+const MOCK_ACCOUNT_NAME = 'Acme Industries';
+const MOCK_ACCOUNT_SLUG = 'acme-industries';
```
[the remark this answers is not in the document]
ahmedomosanya: Good catch — fixed in 119ad673. Aligned `MOCK_UID` with `MOCK_ACCOUNT_ID` so the e2e fixture uses the canonical 18-char Salesforce Account.Id that `ORG_ACCOUNT_ID_PATTERN` / `assertOrgUid` validate (per Spec 002 / member-service v0.7.0). Mirrors the convention the recent `org-profile.spec.ts:27` (`const MOCK_UID = MOCK_ACCOUNT_ID`) settled on.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/components/reassign-key-contact-roles-modal.component.ts
```diff
@@ -0,0 +1,245 @@
...
+    this.isSaving.set(true);
+    this.saveError.set(null);
+    submit(intent)
+      .then(() => this.dialogRef.close(null))
+      .catch((e: unknown) => {
+        this.isSaving.set(false);
+        this.saveError.set(e instanceof Error ? e.message : 'Could not save changes. Please try again.');
+        this.cdr.markForCheck();
+      });
+  }
```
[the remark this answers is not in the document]
ahmedomosanya: Deferring — keeping the modal consistent with the established Org Lens save-handler pattern.

The `submit(intent).then(close).catch(setError)` shape is the live convention in this repo for modal save handlers:

- `apps/lfx-one/src/app/modules/dashboards/org/org-membership-detail/components/edit-key-contact-modal.component.ts:299-305` — the spec-024 modal already on `main`, which this PR's expanded-row pencil reuses unchanged.
- `apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts:337-344` — the parent's `handleReplaceSubmit` shipped on this branch, also `.then().catch()`.

I checked `.claude/rules/*.md` and `CLAUDE.md` — no async/await guideline is codified there. Mirroring the sibling pattern keeps the diff small, the convention uniform, and avoids a stylistic-only refactor that would diverge a single modal from the rest of Org Lens. If we want to migrate all Org Lens save handlers to async/await, that's a worthwhile but separate sweep — happy to file a follow-up if the team agrees.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.html
```diff
@@ -193,21 +200,38 @@
...
                   }}</span>
                 </td>
+                <td class="px-4 py-3 text-right">
+                  <button
+                    type="button"
+                    class="inline-flex h-8 w-8 items-center justify-center rounded-md text-gray-500 hover:bg-gray-100 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500 disabled:cursor-not-allowed disabled:opacity-40 disabled:hover:bg-transparent"
+                    [disabled]="!canEdit()"
+                    [attr.aria-label]="'Reassign key contact roles for ' + group.displayName"
+                    [attr.data-testid]="'org-people-key-contacts-reassign-' + group.email"
+                    [pTooltip]="canEdit() ? 'Reassign key contact roles' : editDisabledTooltip"
```
luismoriguerra: **`pTooltip` won't fire on a natively `[disabled]` button.** When `canEdit()` is false this button is `[disabled]`, and the browser suppresses the pointer events PrimeNG's tooltip directive needs — so read-only users never see the `editDisabledTooltip` ("Only admins can edit…"), which is exactly the audience that needs it. Same applies to the expanded-row Edit button at lines 252-257.

This is the PR #845 pattern: wrap the disabled control in a non-disabled span and keep `[disabled]` / aria / `data-testid` on the inner `<button>`:

```html
<span class="inline-flex" [pTooltip]="canEdit() ? 'Reassign key contact roles' : editDisabledTooltip" tooltipPosition="top">
  <button type="button" [disabled]="!canEdit()" [attr.aria-label]="..." [attr.data-testid]="...">…</button>
</span>
```

Note the membership-detail sibling (`org-membership-detail.component.html:140-152`) has the same latent issue — worth fixing in both.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts
```diff
@@ -125,6 +159,251 @@ export class KeyContactsComponent {
...
+
+  // Fan-out: one PUT per selected role; refresh table afterward; throw on any failure so the modal can retry inline.
+  private async performReassignWrite(intent: ReassignKeyContactRolesSubmitEvent, orgUid: string): Promise<void> {
+    const ops = intent.selected.map((role) =>
+      firstValueFrom(
+        this.membershipsService.replaceKeyContactBySlug(orgUid, role.foundationSlug, role.contactUid, this.toReassignBody(intent, role.contactType))
+      )
+    );
+
+    const results = await Promise.allSettled(ops);
```
luismoriguerra, line 260: Minor / for awareness: this is an unbounded `Promise.allSettled(intent.selected.map(...))` fan-out. The rulebook bans unbounded fan-out, but that rule targets *server-side* bursts to NATS/query-service — here it's client-side and bounded by the small per-person role count, so it's almost certainly fine. Flagging only so it's a conscious decision; if a person could ever hold a large number of (membership, role) tuples, a small worker pool would cap the parallel PUTs.

luismoriguerra (commented):
Audited this in an isolated worktree against the Org Lens UI review rulebook and the membership-detail sibling. Overall it's in good shape and self-consistent with the codebase — the slug-keyed proxy surface, the `writerSet()` client gate (matching `org-membership-detail.component.ts:111` incl. the deliberate fail-open + backend re-enforce), the `accountCancel$` live-source guard, and the dropped server `canEdit`/`canEditAny` fields all look correct. Verified locally: `tsc` (shared + lfx-one) and `prettier --check` are clean for every changed file (the only tsc error is the unrelated missing `docs/generated/docs-manifest` codegen artifact). The earlier bot threads are appropriately resolved (Copilot `[class]` rebuttal is correct Angular-20 class-merge behavior; `MOCK_UID` SFID fix landed in 119ad673; CodeRabbit async/await reasonably deferred to the established `.then().catch()` convention).

A few findings, all non-blocking except the disabled-tooltip one which I'd fix pre-merge (inline):

- **e2e doesn't cover every CRUD verb this PR introduces.** The new spec covers only the Reassign (PUT fan-out) modal. The expanded-row pencil now also wires **add (POST)** and **remove (DELETE)** on the People tab with no e2e here. Mitigated by the reused spec-024 modal being tested on membership-detail, but the People-tab wiring (`handleAddSubmit` / `handleRemoveSubmit`) is unverified. Rulebook: add a happy + failure path for every CRUD verb a PR adds.
- **Post-save toast after account switch (cosmetic).** If the user switches account mid-save, `wireDialogToAccountChange` closes the dialog but the captured-`orgUid` PUTs still complete and fire a success toast + `retry()` in the new account context. Writes target the correct previous org so there's no data risk — just a stray toast.
- **Multi-line JSDoc in `packages/shared/src/interfaces/org-people-key-contacts.interface.ts`.** The new reassign-contract blocks use multi-line `/** … */`; repo convention is one-line JSDoc (longer prose belongs in spec docs).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.html
ahmedomosanya: Good catch — fixed in 276f936b on both pencils (main-row Reassign and expanded-row Edit). Tooltip moves to a non-disabled `<span class="inline-flex" [pTooltip]>` wrapper; `[disabled]`, `[attr.aria-label]`, `[attr.data-testid]`, and the click handler stay on the inner `<button>`. Preserved `tooltipPosition="left"` on the main-row pencil (the right-edge wrap fix from d3fc672f) and `tooltipPosition="top"` on the expanded-row one. E2E selectors use `getByTestId` against the inner button so they continue to work unchanged.

On the membership-detail sibling — confirmed it has the same latent bug, but it's pre-existing on `main` and out of LFXV2-2067's scope. Happy to file a tiny follow-up to apply the same wrap there if you'd prefer that over widening this PR.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts
ahmedomosanya: Conscious decision — leaving as-is. Realistic upper bound is ~9 canonical roles per membership (per `KEY_CONTACT_ROLE_CATALOG`) × the small N memberships a single person typically holds across one org, so the per-click PUT count tops out in the low-tens at worst. If we ever see organisations where one person legitimately holds dozens of (membership, role) tuples we can drop in a small worker pool here without touching the modal contract.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts
```diff
@@ -125,6 +159,251 @@ export class KeyContactsComponent {
...
+      ...(removedPerson ? { detail: `${removedPerson.fullName} is no longer a ${event.contactTypeLabel}.` } : {}),
+      life: 4000,
+    });
+  }
+
+  private initCanEdit(): boolean {
+    const uid = this.accountContext.selectedAccount()?.uid;
+    if (!uid) return true;
+    return this.roleGrants.writerSet().has(uid);
+  }
```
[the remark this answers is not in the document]
ahmedomosanya: Valid — fixed in 2c04912c. `initCanEdit()` now returns `false` when no uid, so the pencils stay disabled during placeholder-account bootstrap instead of being actionable-but-bailing.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts
```diff
@@ -125,6 +159,251 @@ export class KeyContactsComponent {
...
+      return;
+    }
+
+    const total = intent.selected.length;
+    const succeeded = total - failures.length;
+    if (succeeded === 0) {
+      throw new Error(this.cleanErrorMessage(failures[0].reason));
+    }
+    throw new Error(`${succeeded} of ${total} reassignments succeeded. Close this dialog and re-open it to update the remaining roles.`);
+  }
```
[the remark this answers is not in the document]
ahmedomosanya: Partially valid — copy tightened in 2c04912c to: `"Reopen this dialog to retry the remaining roles with fresh state."`

I deliberately did **not** rewrite to in-modal uncheck-and-retry: without per-role succeeded-state markers in the modal, a user who retries with the original boxes still checked will hit 404s on the now-stale `contactUid`s of the roles that already succeeded (member-service `PUT /key-contacts/:contactUid` returns 404 once the original holder is replaced). Close-and-reopen is the only currently-safe retry path because it rebuilds the role list from fresh server state.

Real in-modal uncheck-and-retry needs the parent to pass succeeded keys back to the modal so it can auto-deselect them — filing as a follow-up rather than expanding this PR.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts
```diff
@@ -125,6 +159,251 @@ export class KeyContactsComponent {
...
+  private deriveInitials(firstName: string, lastName: string): string {
+    const f = (firstName ?? '').trim().charAt(0).toUpperCase();
+    const l = (lastName ?? '').trim().charAt(0).toUpperCase();
+    return `${f}${l}` || '?';
+  }
+
+  private openEditModal(contact: OrgMembershipKeyContact, assignment: OrgKeyContactAssignmentVm, orgUid: string): void {
+    // Single-slot roles seed the modal directly into replace-form on the existing holder; multi-slot roles
+    // pre-select the row the user clicked from so the chooser opens already focused on that contact.
+    const editingPersonId = contact.maxContacts === 1 && contact.people.length === 1 ? contact.people[0].personId : assignment.contactUid;
```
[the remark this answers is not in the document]
ahmedomosanya: Valid — fixed in 2c04912c. Confirmed by reading the modal: `editingPersonId` only feeds `initCurrentPerson` (replace-form display), `submitIntent` (replace events), and duplicate-detection in `onSaveClick` — it is **not** consumed in chooser state. Multi-slot now passes `null`, matching the established pattern in `org-membership-detail.component.ts` (spec-024 page). Stale comment dropped.

Chooser preselection itself is a real feature ask — better suited to a follow-up if the UX team wants it.

luismoriguerra (dismissed):

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/components/reassign-key-contact-roles-modal.component.ts
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    this.isSaving.set(true);
+    this.saveError.set(null);
+    submit(intent)
+      .then(() => this.dialogRef.close(null))
+      .catch((e: unknown) => {
+        this.isSaving.set(false);
+        this.saveError.set(e instanceof Error ? e.message : 'Could not save changes. Please try again.');
+        this.cdr.markForCheck();
```
MRashad26: **[MAJOR] `ChangeDetectorRef.markForCheck()` is a no-op in the zoneless model**

**Problem:** `ChangeDetectorRef` is injected and `cdr.markForCheck()` is called after the signal writes `isSaving.set(false)` and `saveError.set(...)`. In the project's stable-zoneless change-detection model, signal writes already schedule change detection via Angular's reactive graph (`ApplicationRef.tick()`) — no manual CD nudge is needed. `markForCheck()` is the zone-based API; it marks a view dirty for the *next* zone-triggered cycle, which never comes in a zoneless app. The call is effectively a no-op.

**Why it matters:** Relying on `markForCheck()` to surface the inline error message after a Promise rejection will silently do nothing. The save error will still render (because `saveError.set(...)` schedules the update), but the CDR injection creates a misleading dependency and obscures that signals alone are sufficient.

**Suggested fix:** Remove the `ChangeDetectorRef` injection and the `cdr.markForCheck()` call:

```typescript
// Remove:
private readonly cdr = inject(ChangeDetectorRef);
// Remove ChangeDetectorRef from the import list.

// In onSave().catch:
.catch((e: unknown) => {
  this.isSaving.set(false);
  this.saveError.set(e instanceof Error ? e.message : 'Could not save changes. Please try again.');
  // No markForCheck() needed — the signal writes above schedule the update.
});
```

MRashad26 (commented):
Code-enforcer audit against CLAUDE.md. One rule violation and one ordering nit flagged below.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts
```diff
@@ -83,12 +108,21 @@ export class KeyContactsComponent {
 
   protected readonly hasUnfilledRoles: Signal<boolean> = computed(() => this.stats().unfilledRequiredRoleCount > 0);
 
+  // Writer-FGA gate (UX); BFF still re-enforces on write.
+  protected readonly canEdit: Signal<boolean> = computed(() => this.initCanEdit());
+  protected readonly editDisabledTooltip = 'Only admins can edit. To view a list of admins, visit the Access page.';
```
MRashad26: **[MINOR] Plain property interspersed in the computed-signal block**

**Problem:** `editDisabledTooltip` is a plain string property declared immediately after the `canEdit` computed signal. Per `component-organization.md`, plain public/protected fields should be grouped before signals (private injections → public fields → forms → WritableSignals → computed/toSignal). Mixing a plain property into the computed block makes the ordering inconsistent.

**Suggested fix:** Move `editDisabledTooltip` up into the fields section (alongside `tableSkeletonRows`) rather than adjacent to computed signals:

```typescript
// In the fields section:
protected readonly tableSkeletonRows: readonly number[] = [0, 1, 2, 3, 4, 5];
protected readonly editDisabledTooltip = 'Only admins can edit. To view a list of admins, visit the Access page.';

// ... (signals / computed further down)
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/components/reassign-key-contact-roles-modal.component.ts
ahmedomosanya: Valid — fixed in 14e563ae. Verified the app uses `provideZonelessChangeDetection()` (apps/lfx-one/src/app/app.config.ts L5/L38), so the manual `markForCheck()` was indeed a no-op. Removed the `ChangeDetectorRef` import + injection + the call; the signal writes (`isSaving.set(false)` + `saveError.set(...)`) already schedule CD through the reactive graph, so the inline error message still renders.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/key-contacts/key-contacts.component.ts
ahmedomosanya: Valid — fixed in 14e563ae. Moved `editDisabledTooltip` up into the public-fields section next to `tableSkeletonRows`, per `component-organization.md` section 4 (injections → public fields → forms → WritableSignals → computed/toSignal).

MRashad26 (approved):

PR #867, opened 2026-06-03
feat(committees): invite members by email with bulk add and pending invites

manishdixitlfx:
## Summary

Reworks the committee/group **Add Member** flow into **invite-by-email** (single + bulk) and surfaces **pending invitations** in the members roster. Resolves the LFXV2-2062 pain points: adding people one-by-one, unreliable type-ahead, and the flow forcing full manual entry when someone isn't already in the search corpus.

JIRA: https://linuxfoundation.atlassian.net/browse/LFXV2-2062

## Why

The member search corpus is `committee_member` / `meeting_registrant` only — there is **no LF identity directory** in the v2 query index — so a real account that had never been a registrant/committee member read as "no LF account" and dropped the admin into full manual entry (email, job title, LinkedIn, org, role, voting). The fix uses the committee-service **invite** primitive (`POST /committees/{uid}/invites`, email-only + optional role; `committee_invite` is a queryable resource), so anyone is added by inviting their email and completes their own profile on accept (invite-and-forget).

## What changed

**Shared** (`@lfx-one/shared`)
- Aligned the unused invite stubs to the real contract: `CommitteeInvite`, `CreateCommitteeInviteRequest`, `CommitteeInviteStatus`, plus result/parse types.
- `parseEmailList` / `isValidEmail` (bulk paste parsing, normalization, case-insensitive dedupe) and `hasLfAccount` — with unit tests.

**BFF** (Express)
- Three-file proxy for committee invites: `GET/POST /api/committees/:id/invites`, `DELETE …/:inviteId`.

**Frontend** (Angular)
- Add-member dialog → unified invite-by-email: paste one or many emails, live preview (to-invite / already-member / already-invited / duplicates / invalid), optional typeahead autofill (reuses LFXV2-2058 ranking), optional batch role, bounded-concurrency fan-out with a partial-success summary. Removes the full manual-entry gate.
- Members roster → **Pending Invitations** section with revoke; already-invited emails are skipped in the dialog.

## Behavior change (reviewers please note)

Inviting no longer sets org / voting status / permission / mailing-list subscription at add time — the invite endpoint carries only email + suggested role. Those move to post-accept member editing (existing flows). The per-person manual-entry fallback is removed from this dialog; the committee **manage** wizard keeps `MemberFormComponent` for full edits.

## Testing

- `yarn lint`, `yarn build` green; shared unit tests pass (incl. new `email.utils` + `hasLfAccount`).
- Manually validated end-to-end against the **dev** cluster: single + bulk invite, pending-invitations display, and revoke.

## Notes

- Invite emails are delivered by the platform pipeline (committee-service → `invite-service` → `email-service`/SES), not the UI.
- Drive-by: added a missing `vitest` import to `committee.utils.spec.ts` (it relied on bare globals and was breaking the shared test run).
- Follow-up: **LFXV2-2081** — CSV import for group members (intentionally not bundled here).

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -25,7 +25,7 @@ import {
 } from '@lfx-one/shared';
 import { GroupsIOMailingList, ProjectContext, TabConfigEntry } from '@lfx-one/shared/interfaces';
 import { COMMITTEE_VALID_TABS } from '@lfx-one/shared/constants';
-import { getChatPlatformIcon, getChatPlatformLabel, getRepoPlatformIcon, getRepoPlatformLabel } from '@lfx-one/shared/utils';
+import { canManageCommitteeMembers, getChatPlatformIcon, getChatPlatformLabel, getRepoPlatformIcon, getRepoPlatformLabel } from '@lfx-one/shared/utils';
 import { CommitteeService } from '@services/committee.service';
 import { LensService } from '@services/lens.service';
 import { MailingListService } from '@services/mailing-list.service';
@@ -550,7 +550,9 @@ export class CommitteeViewComponent {
     return toSignal(
       combineLatest([toObservable(this.committee), toObservable(this.membersRefresh)]).pipe(
         switchMap(([committee]) => {
-          if (!committee?.uid) {
+          // Only managers can see pending invites — gate the fetch (not just the display) so
+          // non-managers never request invitee emails and we don't rely on upstream authz to reject.
+          if (!committee?.uid || !canManageCommitteeMembers(committee)) {
             this.invitesLoading.set(false);
             return of([] as CommitteeInvite[]);
           }
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -15,6 +15,7 @@ import { TagComponent } from '@components/tag/tag.component';
 import { RouteLoadingComponent } from '@components/loading/route-loading.component';
 import {
   Committee,
+  CommitteeInvite,
   CommitteeMember,
   CommitteeMemberVisibility,
   CommitteePermissionLevel,
@@ -109,11 +110,14 @@ export class CommitteeViewComponent {
   public refresh = signal(0);
   public membersRefresh = signal(0);
   public membersLoading = signal<boolean>(true);
+  public invitesLoading = signal<boolean>(true);
   public joiningOrLeaving = signal(false);
 
   // -- Computed / toSignal --
   public committee: Signal<Committee | null> = this.initializeCommittee();
   public members: Signal<CommitteeMember[]> = this.initializeMembers();
+  // Pending invites share the members refresh trigger so adding/revoking refreshes both.
+  public invites: Signal<CommitteeInvite[]> = this.initializeInvites();
 
   // Membership identity comes from server-enriched fields on the committee record,
   // resolved via the username-tagged membership query so visibility doesn't depend
@@ -539,6 +543,31 @@ export class CommitteeViewComponent {
         })
       ),
       { initialValue: [] }
+    );
+  }
+
+  private initializeInvites(): Signal<CommitteeInvite[]> {
+    return toSignal(
+      combineLatest([toObservable(this.committee), toObservable(this.membersRefresh)]).pipe(
+        switchMap(([committee]) => {
+          if (!committee?.uid) {
+            this.invitesLoading.set(false);
+            return of([] as CommitteeInvite[]);
+          }
+
+          this.invitesLoading.set(true);
+
+          return this.committeeService.getCommitteeInvites(committee.uid).pipe(
+            // Only pending invites belong on the roster — accepted ones are already members,
+            // and declined/revoked ones shouldn't block re-inviting. Status casing varies
+            // upstream, so compare case-insensitively.
+            map((invites) => invites.filter((invite) => (invite.status ?? '').toLowerCase() === 'pending')),
+            catchError(() => of([] as CommitteeInvite[])),
+            finalize(() => this.invitesLoading.set(false))
+          );
+        })
+      ),
+      { initialValue: [] as CommitteeInvite[] }
     );
   }
 
--- a/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
@@ -3,8 +3,8 @@
 
 <div class="flex flex-col gap-5 px-1" data-testid="add-member-dialog">
   <p class="text-sm text-gray-500">
-    Invite people by email — one or many at once. Each person gets an invitation to join the group and completes their own profile, so you don't need to enter their
-    details. They appear as "Pending" until they accept.
+    Invite people by email — one or many at once. Each person gets an invitation to join the group and completes their own profile, so you don't need to enter
+    their details. They appear as "Pending" until they accept.
   </p>
 
   <!-- ── Find existing people (typeahead autofill, optional) ──────────────────── -->
@@ -100,7 +100,9 @@
 
     <div class="flex flex-col gap-1 text-xs" data-testid="add-member-preview">
       @if (categorized().toInvite.length > 0) {
-        <span class="text-emerald-600 font-medium"> <i class="fa-light fa-paper-plane" aria-hidden="true"></i> {{ categorized().toInvite.length }} to invite </span>
+        <span class="text-emerald-600 font-medium">
+          <i class="fa-light fa-paper-plane" aria-hidden="true"></i> {{ categorized().toInvite.length }} to invite
+        </span>
       }
       @if (categorized().alreadyMembers.length > 0) {
         <span class="text-gray-400">{{ categorized().alreadyMembers.length }} already in the group — will be skipped</span>
@@ -112,7 +114,9 @@
         <span class="text-gray-400">{{ parsed().duplicates.length }} duplicate(s) removed</span>
       }
       @if (parsed().invalid.length > 0) {
-        <span class="text-amber-600"> <i class="fa-light fa-triangle-exclamation" aria-hidden="true"></i> {{ parsed().invalid.length }} not valid: {{ parsed().invalid.join(', ') }} </span>
+        <span class="text-amber-600">
+          <i class="fa-light fa-triangle-exclamation" aria-hidden="true"></i> {{ parsed().invalid.length }} not valid: {{ parsed().invalid.join(', ') }}
+        </span>
       }
     </div>
   </div>
--- a/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
@@ -117,7 +117,7 @@
       }
       @if (parsed().invalid.length > 0) {
         <span class="text-amber-600">
-          <i class="fa-light fa-triangle-exclamation" aria-hidden="true"></i> {{ parsed().invalid.length }} not valid: {{ parsed().invalid.join(', ') }}
+          <i class="fa-light fa-triangle-exclamation" aria-hidden="true"></i> {{ parsed().invalid.length }} not valid: {{ invalidSummary() }}
         </span>
       }
     </div>
--- a/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
@@ -38,15 +38,15 @@
         @for (result of searchResults(); track result.email) {
           <li
             role="option"
-            [attr.aria-disabled]="result.alreadyMember || result.added"
+            [attr.aria-disabled]="result.alreadyMember || result.alreadyInvited || result.added"
             [attr.aria-selected]="result.added"
             (click)="addEmail(result)"
             (keydown.enter)="addEmail(result)"
             (keydown.space)="$event.preventDefault(); addEmail(result)"
-            [tabindex]="result.alreadyMember || result.added ? -1 : 0"
+            [tabindex]="result.alreadyMember || result.alreadyInvited || result.added ? -1 : 0"
             class="flex items-center gap-3 px-3 py-2.5 rounded-lg border transition-colors"
             [ngClass]="
-              result.alreadyMember || result.added
+              result.alreadyMember || result.alreadyInvited || result.added
                 ? 'border-gray-100 bg-gray-50 opacity-60 cursor-not-allowed'
                 : 'border-gray-200 bg-white hover:bg-blue-50 hover:border-blue-200 cursor-pointer focus:outline-none focus:ring-2 focus:ring-blue-400'
             "
@@ -61,7 +61,7 @@
             <div class="flex flex-col gap-0.5 flex-1 min-w-0">
               <div class="flex items-center gap-2 flex-wrap">
                 <span class="text-sm font-medium text-gray-900 truncate">{{ result.first_name }} {{ result.last_name }}</span>
-                @if (result.username) {
+                @if (result.lfAccount) {
                   <span class="inline-flex items-center gap-1 text-xs font-medium bg-blue-50 text-blue-700 border border-blue-200 rounded px-1.5 py-0.5">
                     <i class="fa-light fa-shield-check text-[10px]" aria-hidden="true"></i> LF Account
                   </span>
@@ -74,6 +74,8 @@
               <span class="flex-none text-xs text-emerald-600 font-medium">Added</span>
             } @else if (result.alreadyMember) {
               <span class="flex-none text-xs text-gray-400 font-medium">Already member</span>
+            } @else if (result.alreadyInvited) {
+              <span class="flex-none text-xs text-gray-400 font-medium">Already invited</span>
             } @else {
               <i class="flex-none fa-light fa-plus text-gray-400" aria-hidden="true"></i>
             }
--- a/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
@@ -1,237 +1,139 @@
 <!-- Copyright The Linux Foundation and each contributor to LFX. -->
 <!-- SPDX-License-Identifier: MIT -->
 
-<div class="flex flex-col gap-0" data-testid="add-member-dialog">
-  <!-- ── STEP 1: Search ───────────────────────────────────────────────────── -->
-  @if (mode() === 'search') {
-    <div class="flex flex-col gap-4 px-1">
-      <!-- Search input -->
-      <div>
-        <label class="block text-sm font-medium text-gray-700 mb-1">Search by name or email</label>
-        <lfx-input-text
-          [form]="searchForm"
-          control="query"
-          placeholder="Type at least 2 characters…"
-          icon="fa-light fa-search"
-          styleClass="w-full"
-          size="small"
-          dataTest="add-member-search-input"></lfx-input-text>
-      </div>
+<div class="flex flex-col gap-5 px-1" data-testid="add-member-dialog">
+  <p class="text-sm text-gray-500">
+    Invite people by email — one or many at once. Each person gets an invitation to join the group and completes their own profile, so you don't need to enter their
+    details. They appear as "Pending" until they accept.
+  </p>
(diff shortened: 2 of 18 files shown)


apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -215,4 +215,61 @@
...
+                  }
+                </div>
+                <lfx-tag value="Pending" severity="warn"></lfx-tag>
+                <lfx-button
+                  icon="fa-light fa-xmark"
+                  [text]="true"
+                  [rounded]="true"
+                  size="small"
+                  severity="secondary"
+                  pTooltip="Revoke invitation"
```
audigregorie: **Major — tooltip never renders + icon-only button has no accessible name**

`lfx-button` surfaces its tooltip through its own `tooltip` input (it imports `TooltipModule` internally and binds it), not the bare PrimeNG `pTooltip` directive. `committee-members.component.ts` does not import `TooltipModule`, so `pTooltip="Revoke invitation"` is an inert attribute — the tooltip will not appear.

Separately, this is an icon-only button (`fa-xmark`) with no text and no `ariaLabel`, so screen readers announce it with no name. Use the component's `tooltip` and `ariaLabel` inputs:

```suggestion
                  tooltip="Revoke invitation"
                  ariaLabel="Revoke invitation"
```

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -215,4 +215,61 @@
...
+        } @else {
+          <ul class="flex flex-col gap-2" role="list">
+            @for (invite of invites(); track invite.uid) {
+              <li
+                class="flex items-center gap-3 px-3 py-2.5 rounded-lg border border-gray-200 bg-white"
+                [attr.data-testid]="'pending-invite-' + invite.invitee_email">
+                <i class="fa-light fa-clock text-amber-500" aria-hidden="true"></i>
+                <div class="flex flex-col gap-0.5 flex-1 min-w-0">
+                  <span class="text-sm font-medium text-gray-900 truncate">{{ invite.invitee_email }}</span>
+                  @if (invite.role && invite.role !== 'None') {
```
audigregorie: **Minor — magic string for the empty-role sentinel**

Comparing `invite.role !== 'None'` hard-codes an upstream sentinel value inline. If the committee-service ever changes the "no role" representation (e.g. empty string, lowercased `none`), this silently shows `Invited as None`. Consider normalizing in the mapping layer or extracting a shared constant/helper so the sentinel lives in one place alongside the other committee-invite contract types.

packages/shared/src/utils/search.utils.ts
```diff
@@ -79,3 +79,17 @@ export function rankUserSearchResults<T extends RankableUser>(results: T[], quer
...
+ * Decides whether a search result corresponds to an existing LF account (LFID).
+ *
+ * "Has an LF account" is inferred solely from the presence of a non-empty, non-blank
+ * `username` (LFID) on the record. The v2 query index has no identity/user resource
+ * type — a person only carries a username once they appear in v2 with a reconciled
+ * LFID — so a blank/whitespace username means "no LFID yet" (use the invite-by-email
+ * path), NOT a definitive "this person has no account anywhere" claim. Treat it as a
+ * routing signal for the add-member flow, not an identity assertion.
+ */
+export function hasLfAccount(user: Pick<UserSearchResult, 'username'>): boolean {
```
audigregorie, line 93: **Major — `hasLfAccount` is dead code**

This function is exported and unit-tested but has no production callers — the add-member dialog only imports `parseEmailList` and `rankUserSearchResults`. The PR description lists `hasLfAccount` under "What changed" as if it's wired into the flow, but nothing consumes it. Either wire it into the invite/typeahead routing it was written for, or drop the function (and its spec block) to avoid shipping unused surface area.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -542,6 +546,31 @@ export class CommitteeViewComponent {
     );
   }
 
+  private initializeInvites(): Signal<CommitteeInvite[]> {
```
audigregorie, line 549: **Major — invites are fetched for every viewer, not just managers**

The Pending Invitations card is gated by `canManageMembers()` in the template, but this fetch is not — `initializeInvites` runs for any committee viewer once `committee.uid` is present. That means every non-manager viewer fires `GET /api/committees/:id/invites` (returning pending invitee emails over the wire) on every load, with display being the only thing that hides it. Relying on upstream authz to reject non-managers also makes this a guaranteed-403-noise path in the common case.

Gate the fetch the same way the display is gated — e.g. short-circuit to `of([])` when the viewer can't manage members (reuse `canManageCommitteeMembers(committee)` from `@lfx-one/shared/utils`). That avoids both the wasted request and any reliance on upstream access checks to not leak invitee emails.

apps/lfx-one/src/app/shared/services/committee.service.ts
```diff
@@ -103,6 +105,26 @@ export class CommitteeService {
...
   }
 
+  // ── Committee Invites ───────────────────────────────────────────────────────
+  // Invite-by-email add-member: lightweight invites for people who may not yet
+  // have an LF account. Pending invites surface in the roster; on accept they
+  // become committee members with their LFID reconciled.
+
+  /** Fetches all invites for a committee (caller filters to pending and handles errors). */
+  public getCommitteeInvites(committeeId: string, params?: HttpParams): Observable<CommitteeInvite[]> {
+    return this.http.get<CommitteeInvite[]>(`/api/committees/${committeeId}/invites`, { params });
```
audigregorie: **Minor — inconsistent `take(1)` with sibling invite methods**

`createCommitteeInvite` and `revokeCommitteeInvite` (and the rest of this service) pipe `take(1)` after the HTTP call; `getCommitteeInvites` omits it. It's functionally harmless here (the consumer wraps it in `switchMap`/`toSignal`), but matching the surrounding convention keeps the service consistent and self-completing.

```suggestion
    return this.http.get<CommitteeInvite[]>(`/api/committees/${committeeId}/invites`, { params }).pipe(take(1));
```

audigregorie (commented):
## Code Review Summary

Solid, well-documented rework of the add-member flow into invite-by-email. The bounded-concurrency fan-out with a partial-success summary, the case-insensitive parse/dedupe util with tests, and the shared-type alignment to the real `committee_invite` contract are all clean. A few things to address before merge: a non-functional tooltip + missing accessible name on the revoke button, a dead exported util, and an ungated invites fetch.

### Major — outside the diff

- **`committee-members.component.html:173`** — the pre-existing "Send Message" button has the same `pTooltip` bug as the new revoke button (see the line comment on 262). `committee-members.component.ts` does not import `TooltipModule`, and `lfx-button` exposes its tooltip via its own `tooltip` input, not the PrimeNG `pTooltip` directive — so this tooltip has never rendered. Worth fixing in the same pass: `pTooltip="Send Message"` → `tooltip="Send Message"` (plus an `ariaLabel`).

### What's done well

- `parseEmailList` handles the messy real-world paste cases (mixed separators, casing, dedupe, first-seen order) and is well covered by unit tests.
- The bulk-invite fan-out (`mergeMap` + `INVITE_CONCURRENCY` + per-email `catchError` + `toArray`) degrades gracefully and the `summarize()` partial-success messaging is a nice touch.
- Shared interfaces were realigned to the actual committee-service contract with excellent explanatory JSDoc, and the `vitest` import fix unblocks the shared test run.

packages/shared/src/utils/email.utils.ts
```diff
@@ -0,0 +1,57 @@
...
+      result.invalid.push(trimmed);
+      continue;
+    }
+
+    if (seen.has(normalized)) {
+      if (!result.duplicates.includes(normalized)) {
+        result.duplicates.push(normalized);
+      }
+      continue;
+    }
```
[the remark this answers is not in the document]
manishdixitlfx: Done — duplicate tracking now uses a `Set` (`duplicatesSeen`) instead of scanning `result.duplicates`, so it's O(n) for large pastes. See 6ae76b98.

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.ts
```diff
@@ -88,214 +112,116 @@ export class AddMemberDialogComponent {
...
+              map(() => ({ email, success: true })),
+              catchError((err: HttpErrorResponse) => of({ email, success: false, reason: this.inviteFailureReason(err) }))
+            ),
+          INVITE_CONCURRENCY
+        ),
+        toArray()
+      )
+      .subscribe((results) => {
+        this.submitting.set(false);
+        this.summarize(results);
```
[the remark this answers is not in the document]
manishdixitlfx: Done — the bulk-invite fan-out now pipes `takeUntilDestroyed(this.destroyRef)`, so in-flight requests are torn down if the dialog is closed mid-send (no post-destroy signal/toast updates). See 6ae76b98.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
manishdixitlfx: Good catch — switched from the inert `pTooltip` to lfx-button's own `tooltip` input and added `ariaLabel="Revoke invitation"` so the icon-only button has an accessible name. See 6ae76b98.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
manishdixitlfx: Done — replaced the inline `'None'` with `CommitteeMemberRole.NONE` (exposed as `noRoleSentinel`), so the sentinel lives with the other committee enums. See 6ae76b98.

packages/shared/src/utils/search.utils.ts
manishdixitlfx: Good catch — `hasLfAccount` is now wired into the typeahead: the LF-Account badge uses it (via an `lfAccount` flag on the decorated result) instead of the inline `username` check, so the tested decision has a real caller. See 6ae76b98.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
manishdixitlfx: Done — `initializeInvites` now short-circuits to `of([])` unless `canManageCommitteeMembers(committee)`, so non-managers never fire the invites request (no invitee emails over the wire, no reliance on upstream authz). See 6ae76b98.

apps/lfx-one/src/app/shared/services/committee.service.ts
manishdixitlfx: Done — added `.pipe(take(1))` to `getCommitteeInvites` to match the create/revoke siblings. See 6ae76b98.

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.ts
```diff
@@ -88,214 +112,116 @@ export class AddMemberDialogComponent {
...
+    if (user.alreadyMember || user.added) {
+      return;
+    }
+    const email = (user.email ?? '').trim();
+    if (!email) {
+      return;
+    }
+    const current = this.form.get('emails')!.value.trim();
+    this.form.get('emails')!.setValue(current ? `${current}\n${email}` : email);
+    this.searchForm.get('query')!.setValue('');
```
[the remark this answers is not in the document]
manishdixitlfx: Done — `DecoratedResult` now carries `alreadyInvited`; `addEmail` early-returns on it and the row renders non-interactive labeled "Already invited", matching the skip-existing-invites behavior. See 6ae76b98.

apps/lfx-one/src/server/controllers/committee.controller.ts
```diff
@@ -515,6 +516,122 @@ export class CommitteeController {
...
+      if (!inviteData?.invitee_email || typeof inviteData.invitee_email !== 'string') {
+        next(
+          ServiceValidationError.forField('invitee_email', 'Invitee email is required', {
+            operation: 'create_committee_invite',
+            service: 'committee_controller',
+            path: req.path,
+          })
+        );
+        return;
+      }
```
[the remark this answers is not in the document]
manishdixitlfx: Done — the controller now trims `invitee_email` before validation (rejects whitespace-only) and forwards the trimmed value upstream. See 6ae76b98.

manishdixitlfx:
## Review Feedback Addressed

Commit: 6ae76b98

### Changes Made
- **committee-view.component.ts**: gate the pending-invites fetch on `canManageCommitteeMembers` — non-managers no longer request invitee emails (per @audigregorie)
- **search.utils.ts + add-member-dialog**: wired `hasLfAccount` into the typeahead LF-Account badge — it was unused (per @audigregorie)
- **committee-members.component.html/.ts**: revoke button now uses lfx-button's `tooltip` + `ariaLabel` (the bare `pTooltip` was inert and the icon-only button had no accessible name); replaced the `'None'` role magic string with `CommitteeMemberRole.NONE` (per @audigregorie)
- **committee.service.ts**: `take(1)` on `getCommitteeInvites` for consistency with the create/revoke siblings (per @audigregorie)
- **add-member-dialog**: search rows now track `alreadyInvited` — blocked at selection and labeled "Already invited" (per CodeRabbit)
- **committee.controller.ts**: trim `invitee_email` before validation and forward the trimmed value (per CodeRabbit)
- **add-member-dialog**: `takeUntilDestroyed` on the invite fan-out to avoid post-destroy signal/toast updates (per Copilot)
- **email.utils.ts**: Set-based duplicate tracking, dropping the O(n²) `includes` scan (per Copilot)

### Threads Resolved
9 of 9 unresolved threads addressed in this iteration.

Validated: `yarn lint` / `yarn build` green, shared unit tests pass (85). Branch was synced with the latest `main` merge before pushing.

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.ts
```diff
@@ -3,37 +3,50 @@
...
-import { MailingListService } from '@services/mailing-list.service';
 import { SearchService } from '@services/search.service';
 import { MessageService } from 'primeng/api';
 import { DynamicDialogConfig, DynamicDialogRef } from 'primeng/dynamicdialog';
 import { SkeletonModule } from 'primeng/skeleton';
-import { catchError, debounceTime, distinctUntilChanged, map, of, startWith, switchMap, take, tap } from 'rxjs';
+import { catchError, debounceTime, distinctUntilChanged, from, map, mergeMap, Observable, of, startWith, switchMap, tap, toArray } from 'rxjs';
 
+/** A search hit decorated for the typeahead: whether it's already added/member/invited and has an LF account. */
+type DecoratedResult = UserSearchResult & { added: boolean; alreadyMember: boolean; alreadyInvited: boolean; lfAccount: boolean };
```
MRashad26: **Local type alias and interface inside `apps/lfx-one/`** — CLAUDE.md violation (two items here, lines 26 and 29).

`type DecoratedResult` (line 26) and `interface CategorizedEmails` (line 29) are declared at module level inside the app. CLAUDE.md: *"All shared constants and interfaces live in `@lfx-one/shared` — no module-level consts or local `interface Foo {}` inside `apps/lfx-one/`."* This rule covers all types and interfaces in `apps/lfx-one/`, including component-only view types.

**Suggested fix:** Move both to `packages/shared/src/interfaces/committee.interface.ts` (alongside the existing `CommitteeInvite` etc.) and import via `@lfx-one/shared/interfaces`:
```ts
// In packages/shared/src/interfaces/committee.interface.ts
export type DecoratedCommitteeSearchResult = UserSearchResult & {
  added: boolean;
  alreadyMember: boolean;
  alreadyInvited: boolean;
  lfAccount: boolean;
};

export interface CategorizedCommitteeEmails {
  toInvite: string[];
  alreadyMembers: string[];
  alreadyInvited: string[];
}
```

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.ts
```diff
@@ -3,37 +3,50 @@
...
+  /** Emails that will be invited (not already a member or pending invite). */
+  toInvite: string[];
+  /** Emails skipped because the person is already a member. */
+  alreadyMembers: string[];
+  /** Emails skipped because a pending invite already exists. */
+  alreadyInvited: string[];
+}
+
+/** Max concurrent create-invite requests when fanning out a bulk invite. */
+const INVITE_CONCURRENCY = 5;
```
MRashad26: **Module-level `const` inside `apps/lfx-one/`** — CLAUDE.md violation.

`const INVITE_CONCURRENCY = 5` is a module-level constant inside the app. CLAUDE.md: *"no module-level consts ... inside `apps/lfx-one/`"* — all constants belong in `@lfx-one/shared/constants`.

**Suggested fix:** Move to `packages/shared/src/constants/` (e.g., a new `committee.constants.ts` or an existing one) and import:
```ts
// packages/shared/src/constants/committee.constants.ts
export const COMMITTEE_INVITE_CONCURRENCY = 5;
```
```ts
import { COMMITTEE_INVITE_CONCURRENCY } from '@lfx-one/shared/constants';
```

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
```diff
@@ -1,237 +1,145 @@
...
+      }
+      @if (categorized().alreadyInvited.length > 0) {
+        <span class="text-gray-400">{{ categorized().alreadyInvited.length }} already invited — will be skipped</span>
+      }
+      @if (parsed().duplicates.length > 0) {
+        <span class="text-gray-400">{{ parsed().duplicates.length }} duplicate(s) removed</span>
+      }
+      @if (parsed().invalid.length > 0) {
+        <span class="text-amber-600">
+          <i class="fa-light fa-triangle-exclamation" aria-hidden="true"></i> {{ parsed().invalid.length }} not valid: {{ parsed().invalid.join(', ') }}
```
MRashad26: **Method call in template** — CLAUDE.md violation.

`parsed().invalid.join(', ')` calls `Array.prototype.join()` inside a template interpolation. CLAUDE.md: *"No functions in HTML templates — only signals, computed values, or pipes."* This is a function call that re-executes on every change-detection cycle.

**Suggested fix:** Derive the joined string inside the `parsed` computed signal (or a dedicated `invalidSummary` computed) so the template reads a signal value:
```ts
// In the component
protected readonly invalidSummary = computed(() => this.parsed().invalid.join(', '));
```
```html
<!-- In the template -->
{{ parsed().invalid.length }} not valid: {{ invalidSummary() }}
```

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.ts
manishdixitlfx: Done — moved both to `@lfx-one/shared/interfaces` (committee.interface.ts): `DecoratedResult` → `DecoratedCommitteeSearchResult` and `CategorizedEmails` → `CategorizedCommitteeEmails`, imported via `@lfx-one/shared/interfaces`. See 98f94f0c.

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.ts
manishdixitlfx: Done — moved to `@lfx-one/shared/constants` as `COMMITTEE_INVITE_CONCURRENCY` (in committees.constants.ts) and imported it. See 98f94f0c.

apps/lfx-one/src/app/modules/committees/components/add-member-dialog/add-member-dialog.component.html
manishdixitlfx: Done — added an `invalidSummary = computed(() => this.parsed().invalid.join(', '))` and the template now reads `{{ invalidSummary() }}` instead of calling `.join()`. See 98f94f0c.

manishdixitlfx:

Commit: 98f94f0c

### Changes Made
- **add-member-dialog → @lfx-one/shared/interfaces**: moved the local view types out of `apps/lfx-one/` — `DecoratedResult` → `DecoratedCommitteeSearchResult`, `CategorizedEmails` → `CategorizedCommitteeEmails` (per @MRashad26)
- **@lfx-one/shared/constants**: moved `INVITE_CONCURRENCY` → `COMMITTEE_INVITE_CONCURRENCY` (no module-level consts inside the app) (per @MRashad26)
- **add-member-dialog.component.html**: replaced the `parsed().invalid.join(', ')` template function call with an `invalidSummary` computed signal (per @MRashad26)

### Threads Resolved

Validated: `yarn build` + shared unit tests (85) green; branch in sync with origin.

MRashad26 (approved):

audigregorie (approved):

PR #871, opened 2026-06-03
fix(enrollment): use UTC midnight for end date status comparison (LFXV2-2084)

fayazg:
## Summary

- `deriveEnrollmentStatus` was parsing `EndDate` (a `YYYY-MM-DD` calendar day)
  as local midnight via `parseLocalDateString`, while the template renders the
  same field in UTC
- For users west of UTC (most US users), this caused the status chip and the
  displayed renewal date to disagree by one day at the boundary
- Fix: parse `EndDate` as `Date.UTC()` to align with the existing UTC display
  convention; drop the now-unused `parseLocalDateString` import

Fixes LFXV2-2084

## Test plan

- [ ] Open the individual enrollment view for an account whose membership
      `EndDate` is today or within the next few days; confirm the status chip
      (Active / Expiring Soon / Expired) and the displayed renewal date agree
- [ ] Confirm the 30-day "Expiring Soon" threshold still fires correctly for a
      membership expiring within 30 days

### DIFF
--- a/packages/shared/src/utils/enrollment.utils.ts
+++ b/packages/shared/src/utils/enrollment.utils.ts
@@ -3,7 +3,6 @@
 
 // Generated with [Claude Code](https://claude.ai/code)
 
-import { parseLocalDateString } from './date-time.utils';
 import { EnrollmentDisplayStatus, IndividualEnrollment } from '../interfaces/enrollment.interface';
 
 // Status derivation mirrors enrollment.plugin.js:197–226 from myprofile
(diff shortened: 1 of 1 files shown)


audigregorie (commented):
## Code Review Summary

✅ **Approve — no issues found.** Correct, surgical fix for the status/display timezone mismatch.

The template renders the same field as `{{ item.membership.EndDate | date:'mediumDate':'UTC' }}` (and `formatDate(..., 'UTC')` in the component), so parsing `EndDate` via `Date.UTC()` correctly aligns the status comparison with the established UTC display convention. This eliminates the off-by-one-day disagreement at the boundary for west-of-UTC users.

### What's done well

- Correctly diagnosed the root cause and aligned the parse basis (UTC) with the display basis (UTC).
- Preserves the existing midnight boundary semantics that mirror `enrollment.plugin.js:197–226`; only the timezone of midnight shifted.
- Clean removal of the now-unused `parseLocalDateString` import — no dangling references.
- Minimal diff scoped exactly to the bug (2 insertions, 6 deletions).

### Considered and dismissed (not blocking)

- **Dropped `try/catch` + `isNaN` guard is not a regression** — the outer `/^\d{4}-\d{2}-\d{2}$/` regex already guarantees a well-formed numeric date, so `parseLocalDateString` could never throw on this path (the `catch` was effectively dead code). Invalid calendar dates (e.g. `2026-02-30`) roll over identically under both old and new code.
- **Inline `Date.UTC` vs. a shared `parseUtcDateString` helper** — only one call site, so extracting a companion helper would be over-engineering; the explicit `Date.UTC(y, m - 1, d)` is also clearer than relying on the `new Date('YYYY-MM-DD')` ISO-UTC parsing nuance.

**Rating: ⭐⭐⭐⭐⭐**

MRashad26 (approved):

PR #878, opened 2026-06-03
fix(meetings): occurrence recurrence wins for cadence label

manishdixitlfx:
## Summary

A recurring meeting whose cadence changed from a specific occurrence onwards (e.g. monthly → quarterly) rendered the **stale** cadence label ("Monthly on the 1st Thursday" instead of "Quarterly on the 1st Thursday") on every UI surface, because each surface piped the top-level `meeting.recurrence`.

Zoom records a cadence change as an `all_following` update, and the meeting-service occurrence calculator (LFXV2-2066) stamps the new pattern onto that occurrence's own `recurrence`. The top-level `recurrence` is intentionally left as the original rule, so the **occurrence-level** recurrence is the source of truth for the label and the UI must do this lookup itself.

## Changes

- Add `MeetingOccurrence.recurrence?: MeetingRecurrence | null` to the shared interface.
- Add `resolveOccurrenceRecurrence(meeting, occurrence)` in `@lfx-one/shared` — one shared priority rule (`occurrence?.recurrence ?? meeting?.recurrence ?? null`) instead of each surface re-implementing `??`.
- Wire the displayed occurrence through the helper via a `displayRecurrence` computed on **meeting-card** (list card), **dashboard-meeting-card** (tooltip), and **meeting-join** (detail view).
- `past-meeting-details` intentionally keeps the top-level rule (a past meeting is a single-occurrence snapshot with no per-occurrence override to resolve — documented inline).
- Refresh the stale `buildRecurrenceSummary` note that previously misdiagnosed this (LFXV2-2057) as an upstream sync problem.
- Unit tests for `resolveOccurrenceRecurrence` (override wins / fallback / null-safe) plus an end-to-end label guard modeled on meeting `92079944361`.

## ⚠️ Known limitation (detail/join page)

The meeting **detail/join** view loads via the live ITX endpoint (`/itx/meetings/:uid`), whose occurrences currently return `recurrence: null`. So the fix there is **forward-compatible only** — it resolves to the top-level rule (no regression) and will start rendering the correct cadence once the ITX detail path also applies the effective/segment recurrence (mirroring LFXV2-2066's indexer fix) **or** the detail view switches to indexed data. The **list card / dashboard card are fixed today** (query-service index carries the override), which covers the primary repro surface.

Secondary: only the segment **anchor** occurrence carries the override in the index; later occurrences in the same segment fall back to the top-level rule until re-indexed.

## Upstream dependency

- meeting-service [PR #186 / LFXV2-2066](https://github.com/linuxfoundation/lfx-v2-meeting-service/pull/186) — added the per-occurrence effective-recurrence stamping consumed here (merged 2026-06-02, must be deployed).

## Testing

- 91/91 shared unit tests pass (6 new). `lint`, `check-types`, `format`, `build` (SSR bundle) all green.
- Verified against live prod for meeting `92079944361`: query-service path carries `occurrence.recurrence.repeat_interval: 3`; resolved label → "Quarterly on the 1st Thursday".

JIRA: [LFXV2-2112](https://linuxfoundation.atlassian.net/browse/LFXV2-2112)

[LFXV2-2112]: https://linuxfoundation.atlassian.net/browse/LFXV2-2112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/shared/src/utils/meeting.utils.ts
+++ b/packages/shared/src/utils/meeting.utils.ts
@@ -8,6 +8,7 @@ import {
   CustomRecurrencePattern,
   Meeting,
   MeetingOccurrence,
+  MeetingRecurrence,
   PastMeeting,
   PastMeetingSummary,
   PastMeetingTranscript,
@@ -110,11 +111,13 @@ export function buildRecurrenceSummary(pattern: CustomRecurrencePattern): Recurr
       // (the meeting-service has no distinct QUARTERLY type), so surface that cadence
       // by name rather than the literal "Every 3 months".
       //
-      // NOTE (LFXV2-2057): the cadence label is derived from this recurrence RULE.
-      // A meeting switched to quarterly in PCC only renders as "Quarterly" once the
-      // PCC→V2 sync updates repeat_interval to 3. If a quarterly meeting still shows
-      // "Monthly", the rule is stale upstream (repeat_interval=1 with only the
-      // occurrences spaced quarterly) — that is an upstream sync fix, not a UI one.
+      // NOTE (LFXV2-2066/LFXV2-2112): the cadence label is derived from whichever recurrence
+      // RULE the caller feeds in. When a meeting's cadence changes from a given occurrence
+      // onwards, the meeting's top-level `recurrence` is intentionally left as the original
+      // rule and the new cadence (e.g. repeat_interval=3 for quarterly) is carried on the
+      // affected occurrence's own `recurrence`. Callers must therefore resolve the
+      // occurrence-level override first (see `resolveOccurrenceRecurrence`); a quarterly
+      // meeting still showing "Monthly" is a UI lookup bug, not an upstream sync problem.
       let monthText: string;
       if (interval === 1) {
         monthText = 'Monthly';
@@ -247,6 +250,26 @@ export function getCurrentOrNextOccurrence(meeting: Meeting): MeetingOccurrence 
     .sort((a, b) => new Date(a.start_time).getTime() - new Date(b.start_time).getTime());
 
   return futureOccurrences.length > 0 ? futureOccurrences[0] : null;
+}
+
+/**
+ * Returns the recurrence that should drive the displayed cadence label for a given occurrence:
+ * the occurrence's own recurrence override when present (the cadence changed at/after this
+ * occurrence — LFXV2-2112), otherwise the meeting's top-level recurrence.
+ *
+ * Background: when a recurring meeting's cadence changes from a specific occurrence onwards,
+ * Zoom records it as an `all_following` update and the meeting-service's occurrence calculator
+ * stamps the new pattern onto that occurrence's `recurrence` (LFXV2-2066). The meeting's
+ * top-level `recurrence` is intentionally left as the original rule, so the occurrence-level
+ * override — when present — is the source of truth for the cadence label. Centralised here so
+ * every surface that renders a recurrence label shares one priority rule.
+ *
+ * @param meeting The meeting (only its top-level `recurrence` is read)
+ * @param occurrence The occurrence being displayed (its `recurrence` override wins when set)
+ * @returns The recurrence to feed the label formatter, or null when neither is available
+ */
+export function resolveOccurrenceRecurrence(meeting: Pick<Meeting, 'recurrence'>, occurrence?: MeetingOccurrence | null): MeetingRecurrence | null {
+  return occurrence?.recurrence ?? meeting?.recurrence ?? null;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


packages/shared/src/utils/meeting.utils.ts
```diff
@@ -249,6 +252,26 @@ export function getCurrentOrNextOccurrence(meeting: Meeting): MeetingOccurrence
...
+ * stamps the new pattern onto that occurrence's `recurrence` (LFXV2-2066). The meeting's
+ * top-level `recurrence` is intentionally left as the original rule, so the occurrence-level
+ * override — when present — is the source of truth for the cadence label. Centralised here so
+ * every surface that renders a recurrence label shares one priority rule.
+ *
+ * @param meeting The meeting (only its top-level `recurrence` is read)
+ * @param occurrence The occurrence being displayed (its `recurrence` override wins when set)
+ * @returns The recurrence to feed the label formatter, or null when neither is available
+ */
+export function resolveOccurrenceRecurrence(meeting: Pick<Meeting, 'recurrence'>, occurrence?: MeetingOccurrence | null): MeetingRecurrence | null {
```
dealako, line 271: [nit] The parameter type is the non-optional `Pick<Meeting, 'recurrence'>`, so `meeting?.recurrence` could be `meeting.recurrence` — the optional chaining is dead defense here. Harmless (and arguably fine as belt-and-suspenders against a runtime `null` slipping past the type), so take it or leave it.

dealako (dismissed):
Hi @manishdixitlfx — thanks for the thorough write-up; the PR description and inline `NOTE`/comment trail make the intent and the known limitations easy to follow.

## Overall impression

Tightly-scoped bug fix that centralizes the occurrence-vs-series cadence precedence into a single shared helper (`resolveOccurrenceRecurrence`) instead of letting each surface re-implement the `??` chain. The `displayRecurrence` computed is wired consistently across the three live surfaces, `past-meeting-details` documents why it opts out, and the stale `LFXV2-2057` note is corrected. Test coverage is solid — precedence, both fallbacks, null-safety, plus two end-to-end label guards through `buildRecurrenceSummary`. Quality signal is high.

## Independent review

- **Correctness**: The `occurrence?.recurrence ?? meeting?.recurrence ?? null` precedence is correct and the template guard change (`@if (meeting().recurrence)` → `@if (displayRecurrence(); as recurrence)`) preserves the original behavior while correctly surfacing the badge when only an occurrence override is present. `displayRecurrence` computeds are lazy, so the field-declaration order (`currentOccurrence` before `displayRecurrence`) is not a concern.
- **Security**: No security surface. Recurrence values are typed enums/numbers fed to a text formatter and rendered via Angular property binding (auto-escaped). No injection, auth, secret, or untrusted-input concerns. (security-auditor pass: clean.)
- **Performance**: Negligible — a constant-time `??` chain inside a memoized computed.
- **Test coverage**: Good. The end-to-end label guard modeled on meeting `92079944361` is a nice touch.
- **Documentation**: Above average; the forward-compatibility caveat on `meeting-join` and the `past-meeting-details` opt-out are both documented inline.

## AI comment reconciliation

- **CodeRabbit** — "No actionable comments." Agreed; nothing actionable on this diff.
- **GitHub Copilot** — Reviewed 10/10 files, no comments. Agreed.

No prior human review rounds to track; no regressions observed.

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — redundant optional chaining (`meeting?.recurrence`) on a non-optional `Pick<Meeting, 'recurrence'>` parameter (see inline).

Note (non-blocking): the branch is `BEHIND` `main` but `MERGEABLE` (no conflicts) — a rebase before merge is worth doing but doesn't gate this review.

## Decision

✅ **Approved**

packages/shared/src/utils/meeting.utils.ts
manishdixitlfx: Fixed — dropped the dead `?.` on `meeting` (now `meeting.recurrence`); the trailing `?? null` still covers a runtime null slipping past the type. Good catch.

See commit b3f55b86.

manishdixitlfx:
## Review Feedback Addressed

Commit: b3f55b86

### Changes Made
- **packages/shared/src/utils/meeting.utils.ts**: dropped the dead optional chaining on the non-nullable `meeting` param in `resolveOccurrenceRecurrence` (`meeting?.recurrence` → `meeting.recurrence`); the trailing `?? null` still guards a runtime null (per @dealako)

### Threads Resolved

dealako (approved):

PR #879, opened 2026-06-03
fix(campaigns): improve error messages and mitigate multi-replica job polling

mrautela365:
## Summary

- Set `INLINE_WAIT_MS` to 45s in `campaign-proxy.service.ts` — stays under the 60s ingress-nginx `proxy-read-timeout` default while giving most campaign creation jobs time to complete on the same pod. Jobs that take longer fall through to the frontend polling path.
- Rewrite all user-facing error messages across the frontend and backend to be marketer-friendly — no job IDs, no replica references, no "something went wrong"
- Add server-side structured logging (`logger.warning` / `logger.error`) for technical error details so ops can debug without exposing internals to users

LFXV2-2023


apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -680,11 +681,12 @@ export class CampaignProxyService {
     const jobId = createJob();
 
     this.executeCampaignCreation(jobId, body).catch((error) => {
-      failJob(jobId, error instanceof Error ? error.message : 'Campaign creation failed');
+      logger.error(undefined, 'campaign_create_unhandled', Date.now(), error as Error, { jobId });
+      failJob(jobId, 'Campaign creation was unsuccessful. Please try again.');
     });
```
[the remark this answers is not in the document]
mrautela365: Done — wrapped the unknown rejection value in `new Error(String(error))` before passing to `logger.error`, preserving stack/context for non-Error rejections.

See commit d32568ce.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -1218,9 +1223,6 @@ function extractGadsErrorMessage(error: unknown): string {
 
   if (code && friendlyMessages[code]) return friendlyMessages[code];
 
-  const originalMessage = error instanceof Error ? error.message : '';
-
-  if (code) return `Campaign creation failed (error code: ${code}). Please try again or contact support.`;
-  if (originalMessage) return originalMessage;
-  return 'Campaign creation failed. Please try again or contact support.';
+  if (code) return `Google Ads rejected the campaign (${code}). Please try again or contact your administrator.`;
+  return 'Google Ads could not create the campaign. Please try again.';
```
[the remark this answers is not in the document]
mrautela365: Thanks for flagging. The error codes surfaced here are human-readable Google Ads enum values (e.g. `DUPLICATE_CAMPAIGN_NAME`, `INVALID_INPUT`) rather than stack traces or internal IDs. They help admins troubleshoot without needing server log access. The most common codes are already mapped to fully friendly messages above (lines 1217-1221) — this fallback only fires for unmapped codes, where the enum name is the most useful context we can provide. The full error detail is already logged server-side for debugging.

No code change needed — the current approach balances user-facing clarity with debuggability.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
```diff
@@ -226,12 +226,12 @@ export class ImplementationTabComponent {
           }
         },
         error: () => {
-          this.errors.set(['Failed to retrieve campaign results.']);
+          this.errors.set(['Lost connection to the campaign creation process. Please try again or check Google Ads directly.']);
           this.step.set('results');
```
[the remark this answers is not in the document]
mrautela365: Done — the poll error handler now extracts the message from the error if it is an Error instance, falling back to the generic "Lost connection" message for network failures or non-Error rejections.

See commit d32568ce.

mrautela365:
## Review Feedback Addressed

Commit: d32568ce

### Changes Made
- **campaign-proxy.service.ts**: Wrapped unknown rejection value in `new Error(String(error))` before passing to `logger.error` to preserve stack/context for non-Error rejections (per copilot[bot])
- **implementation-tab.component.ts**: Poll error handler now extracts message from Error instances, falling back to generic "Lost connection" message for non-Error rejections (per coderabbitai[bot])

### No Change Needed
- **campaign-proxy.service.ts:1227**: Google Ads error codes in user-facing messages are human-readable enum values (e.g. `DUPLICATE_CAMPAIGN_NAME`), not technical internals. Common codes are already mapped to friendly messages; this fallback only fires for unmapped codes where the enum name is the most useful context. Full error detail is logged server-side. (per copilot[bot])

### Threads Resolved
3 of 3 unresolved threads addressed.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -679,12 +680,14 @@ export class CampaignProxyService {
...
-      failJob(jobId, error instanceof Error ? error.message : 'Campaign creation failed');
+    this.executeCampaignCreation(jobId, body).catch((error: unknown) => {
+      const err = error instanceof Error ? error : new Error(String(error));
+      logger.error(undefined, 'campaign_create_unhandled', Date.now(), err, { jobId });
+      failJob(jobId, 'Campaign creation was unsuccessful. Please try again.');
     });
 
     const POLL_INTERVAL_MS = 500;
-    const INLINE_WAIT_MS = 15_000;
+    const INLINE_WAIT_MS = 120_000;
```
MRashad26: **[MAJOR] 120 s inline wait exceeds Kubernetes ingress-nginx's default `proxy-read-timeout`**

**Problem:** The ingress-nginx `proxy-read-timeout` defaults to **60 seconds** in most Kubernetes deployments. A 120-second inline poll means any campaign that takes longer than ~58 s (60 s timeout minus a few seconds of network overhead) will receive a `504 Gateway Timeout` from the ingress before this method returns. The frontend will treat that as an error even though the campaign may still be running successfully in the background — and the fire-and-forget on line 683 means the job continues after the timeout, creating a risk of silent success that looks like a failure to the user.

Raising the wait from 15 s to 120 s does not solve the underlying multi-replica problem; it just narrows the window where the inline path succeeds (jobs that complete in < 60 s instead of < 15 s) while the out-of-window path still hits cross-replica polling failures.

**Root cause:** Job state is stored in a per-process in-memory `Map`. Clients polling from a different replica can never find it.

**Correct fix:** Persist job state outside the process so all replicas share it — for example, write job status to the existing database (a `campaign_jobs` table) or Redis on every `createJob`/`completeJob`/`failJob` transition. `getJobStatus` then reads from shared state and is replica-independent. This removes the need for any inline wait window at all.

**If a shared-state migration is out of scope for this ticket**, the least-bad interim option is sticky sessions (`sessionAffinity: ClientIP` or `nginx.ingress.kubernetes.io/affinity: cookie` on the Service) so polling requests reach the same pod, *combined* with resetting `INLINE_WAIT_MS` back to 15 s so the HTTP response stays comfortably under the ingress read timeout. Document the sticky-session dependency so it isn't silently removed.

MRashad26 (commented):
Code-enforcer audit. No CLAUDE.md violations. One MAJOR reliability concern flagged inline.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
mrautela365: Good catch — reduced `INLINE_WAIT_MS` from 120s to 45s to stay safely under the 60s ingress-nginx `proxy-read-timeout` default.

The underlying multi-replica issue (in-memory job state) is acknowledged — shared state (Redis or DB) is the correct long-term fix. That's tracked separately under the Campaigns epic. For now, jobs that don't complete within 45s fall through to the frontend polling path, which already handles the `{ jobId }` fallback.

See commit 8e8885ac.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -679,12 +680,15 @@ export class CampaignProxyService {
   public async createCampaign(_req: Request, body: CampaignCreateRequest): Promise<{ jobId: string; result?: CampaignCreateResponse; error?: string }> {
     const jobId = createJob();
 
-    this.executeCampaignCreation(jobId, body).catch((error) => {
-      failJob(jobId, error instanceof Error ? error.message : 'Campaign creation failed');
+    this.executeCampaignCreation(jobId, body).catch((error: unknown) => {
+      const err = error instanceof Error ? error : new Error(String(error));
+      logger.error(undefined, 'campaign_create_unhandled', Date.now(), err, { jobId });
+      failJob(jobId, 'Campaign creation was unsuccessful. Please try again.');
```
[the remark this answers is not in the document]
mrautela365: Fixed — captured `startTime` before the fire-and-forget so `logger.error` reports actual creation duration instead of near-0ms. Kept `undefined` for `req` because the `.catch()` runs asynchronously after the HTTP response cycle completes — the request context is no longer reliable.

See commit 6fe36eed.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -679,12 +680,15 @@ export class CampaignProxyService {
...
     });
 
     const POLL_INTERVAL_MS = 500;
-    const INLINE_WAIT_MS = 15_000;
+    // Must stay under ingress-nginx proxy-read-timeout (default 60s)
+    const INLINE_WAIT_MS = 45_000;
```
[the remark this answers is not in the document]
mrautela365: Updated the PR description to reflect the 45s value and the ingress timeout rationale. The 120s was an intermediate step that was reduced to 45s after a reviewer correctly pointed out the ingress-nginx `proxy-read-timeout` constraint (default 60s).

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -678,13 +679,17 @@ export class CampaignProxyService {
...
+    this.executeCampaignCreation(jobId, body).catch((error: unknown) => {
+      const err = error instanceof Error ? error : new Error(String(error));
+      logger.error(undefined, 'campaign_create_unhandled', startTime, err, { jobId });
+      failJob(jobId, 'Campaign creation was unsuccessful. Please try again.');
     });
 
     const POLL_INTERVAL_MS = 500;
-    const INLINE_WAIT_MS = 15_000;
+    // Must stay under ingress-nginx proxy-read-timeout (default 60s)
+    const INLINE_WAIT_MS = 45_000;
```
dealako, line 692: **[minor]** 45s is the right value — it stays under the ingress-nginx `proxy-read-timeout` (default 60s), unlike the earlier 120s. Flagging for the record (agreeing with @MRashad26's and CodeRabbit's threads above): this only *narrows the window*. Any job still running after 45s falls back to `/jobs/{id}` polling, which reads the in-memory `jobs` Map (line 433). In a multi-replica deployment without sticky sessions, that request can land on a different pod and return `not_found`, so the user sees "Lost connection…" on a campaign that actually succeeded in the background.

Pre-existing limitation and fine to defer to the shared-state (Redis/DB) follow-up. The only ask here: have the PR title say "mitigate" rather than "fix multi-replica job polling," since the cross-replica race still exists.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -1218,9 +1226,6 @@ function extractGadsErrorMessage(error: unknown): string {
 
   if (code && friendlyMessages[code]) return friendlyMessages[code];
 
-  const originalMessage = error instanceof Error ? error.message : '';
-
-  if (code) return `Campaign creation failed (error code: ${code}). Please try again or contact support.`;
-  if (originalMessage) return originalMessage;
-  return 'Campaign creation failed. Please try again or contact support.';
+  if (code) return `Google Ads rejected the campaign (${code}). Please try again or contact your administrator.`;
```
dealako, line 1229: **[nit]** Reconciling Copilot's thread on this line: surfacing the raw enum `(${code})` for unmapped errors is a reasonable middle ground — these are human-readable Google Ads enums (e.g. `DUPLICATE_CAMPAIGN_NAME`), not stack traces or internal IDs, and the full detail is already logged server-side. No change needed. Optional: grow the `friendlyMessages` map above as new codes show up in the logs, so fewer users ever see the raw enum.

dealako (approved):
@mrautela365 — thanks for working this one all the way through the layered reviewer feedback.

## Overall impression

Tightly-scoped, low-risk change that does exactly what it sets out to do: replace technical/internal error strings (job IDs, replica hints, "something went wrong") with marketer-friendly copy, and push the technical detail into structured server-side logs. The reviewer iterations (Copilot, CodeRabbit, and @MRashad26's code-enforcer pass) were all addressed cleanly, and CI is green. I re-verified each prior fix against the current code and they hold up.

## Revision tracking (prior rounds)

- ✅ **CodeRabbit** — preserve backend error detail in the poll error handler → addressed in `d32568c`. Verified correct and safe: Angular's `HttpErrorResponse` is `implements Error` but does **not** extend `Error`, so genuine network/5xx poll failures fall through to the generic "Lost connection…" message, while the status-based friendly `Error`s thrown in `campaign.service.ts` surface their message. No technical leak.
- ✅ **Copilot** — wrap the unknown rejection in `new Error(String(error))` before `logger.error` → addressed in `d32568c`. Matches the logger API.
- ✅ **MRashad26 / CodeRabbit** — 120s inline wait exceeds the 60s ingress `proxy-read-timeout` → addressed in `8e8885a` (reduced to 45s). Correct mitigation.
- ✅ **Copilot** — capture `startTime` before the fire-and-forget so `logger.error` reports real duration → addressed in `6fe36ee`. Keeping `undefined` for `req` in the post-response `.catch()` is a reasonable call.
- ✅ **Copilot** — PR description vs. 45s mismatch → description updated.

No regressions introduced across the new commits.

## Remaining items (none blocking)

See the 2 inline comments. Briefly:

- 🟡 **Minor** — the cross-replica polling race is *mitigated, not fixed*. With multi-replica + no sticky sessions, a poll landing on a different pod after the 45s inline window still returns `not_found` → user sees "Lost connection…" even on success. Pre-existing (in-memory `jobs` Map) and fine to defer to the shared-state follow-up, but the PR title "…and fix multi-replica job polling" overstates what ships. Suggest "mitigate" instead.
- ⚪ **Nit** — unmapped Google Ads error codes still surface the raw enum to users. Your rationale (human-readable enums, unmapped-only, full detail logged) is reasonable; optional to grow the map over time.
- ⚪ **Nit** — no unit tests cover the new error-mapping/logging branches. Consistent with the existing files (no specs today) and low-risk, so not a blocker; worth a follow-up if these paths grow.

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 1 — multi-replica polling race remains (deferred; PR title overstates scope)
- ⚪ Nit: 2 — raw Gads error code for unmapped cases; no tests for the new branches

## Decision: ✅ Approved with minor comments

The remaining multi-replica work is legitimately out of scope for an error-messaging fix and is already tracked. Only ask before/with merge: soften the PR title/description to "mitigate" rather than "fix multi-replica job polling" so the history isn't misleading.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
mrautela365: Good call — updated the PR title from "fix multi-replica job polling" to "mitigate multi-replica job polling" to accurately reflect that the cross-replica race on the in-memory `jobs` Map still exists.

The shared-state follow-up (Redis or DB-backed job store) is the proper fix — this PR just narrows the window by reducing the inline wait to 45s.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
mrautela365: Agreed — the raw enum fallback is intentional as a middle ground. These are human-readable Google Ads enums (`DUPLICATE_CAMPAIGN_NAME`, etc.), not internal IDs or stack traces.

Will grow the `friendlyMessages` map as new codes show up in production logs.

mrautela365:
## Review Feedback Addressed

### Changes Made
- **PR title**: Updated from "fix multi-replica job polling" to "mitigate multi-replica job polling" to accurately reflect the cross-replica race still exists — deferred to shared-state follow-up (per @dealako)

### No Change Needed
- **campaign-proxy.service.ts:1229**: Raw enum fallback for unmapped Google Ads error codes is intentional — will grow `friendlyMessages` map as new codes appear in logs (per @dealako)

### Threads Resolved

PR #883, opened 2026-06-04
feat(org-people): add access tab banner and wire all employees access column

ahmedomosanya:
## Summary

`feat(org-people): add access tab permissions banner and tidy toolbar`

- Adds a navigation-rail illustration banner above the Access tab toolbar with copy explaining what Org Lens Access is plus per-role descriptions for Org Admin (Editor) and Org Admin (Viewer)
- Moves the Add Users button left of the search input and type filter on the same toolbar row; drops the separate row that previously held the button
- Removes the redundant `Org Lens Access` h2 + subtitle now that the banner carries the same intent
- Renames the add-user modal header from "Add User" to "Add Users" to match the button label
- New asset: `apps/lfx-one/public/images/org-lens-nav-rail.svg`

`feat(org-people): wire all employees access column with roster union`

- New `OrgLensAccessStateService` (root-singleton): shared roster cache keyed by `orgUid`, in-flight dedup via `shareReplay`, exposes `rosterForOrg` and `accessByEmailFor` signals plus `ensureLoaded` and `setRoster`
- Access tab pushes every authoritative response (load, invite, edit, remove) into the shared cache so the All Employees badge reflects writes immediately — no extra fetch
- All Employees tab hydrates the cache on org change via `ensureLoaded` and renders `Admin` / `Viewer` / `Invited` badges per row from the email-keyed map; rows without an email render a dash
- **UNION step**: appends synthetic rows for access principals not in the activity roster. Lowercased email is the merge key. A principal granted Org Lens access who shows zero detected activity is a data-quality signal we want surfaced, not hidden
- Synthetic rows are non-interactive: chevron hidden, `role` / `tabindex` / `aria-expanded` dropped, `toggleRow` and `onRowKeydown` guarded by an `isSynthetic` flag; their `personKey` is namespaced with an `access:` prefix to stay disjoint from BFF-issued opaque keys
- Lifts the row view-model to shared as `OrgAllEmployeeRowVm` (matches sibling `OrgTraineeRowVm` / `OrgEventAttendeeRowVm` / `OrgContributorRowVm`; enforces `.claude/rules/component-organization.md` §5); introduces `OrgAccessBadgeState` (= `OrgAccessRole | 'invited'`) to dedupe the inline union previously copied between the state service and the row VM

- **CSV upload and "Preview Invitation Email"** in the add-user modal — single-recipient flow only for this slice
- **"Has Org Lens access" filter option on the All Employees tab** — the prototype has no such filter; deferred until product asks
- **Replacing the post-arrival `if (res.orgUid !== selectedAccount().uid) return;` guard in the Access tab with a `<contextKey>Cancel$` Subject + `takeUntil`** — pre-existing pattern; the new `setRoster(res.orgUid, res)` call is composite-correct (cache slot keyed by the response's `orgUid`, not ambient context), so safe to defer

## Test plan

- [ ] **Access tab banner renders.** Hard-reload `/org/<sfid>/people/access`. Banner appears above the toolbar with the nav-rail illustration on the left, intro copy + role legend on the right. Banner stacks vertically at small viewports.
- [ ] **Toolbar layout.** When `canManage` is true, the Add Users button sits to the left of the search input and the type filter on the same row. No separate row above the table.
- [ ] **Modal header.** Click Add Users — modal opens with header `Add Users` (not `Add User`).
- [ ] **All Employees badge — same-org parity.** Pre-condition: Access tab has at least one Admin and one Viewer for the current org. Open the All Employees tab — those principals' rows show the matching badge in the `Org Lens Access` column.
- [ ] **Mid-session refresh (cache push).** From Access tab, invite a new viewer whose email matches an existing employee. Switch to All Employees tab — the same email row now shows `Invited`. No extra HTTP request in the Network panel (the Access tab's response was pushed into the shared cache).
- [ ] **Synthetic rows (UNION).** Grant access to an email that has zero activity in the current org. All Employees tab shows that principal as a non-interactive synthetic row: dashes in all activity columns, badge in the Access column, no chevron, no row hover/click expansion. Tab/Enter on the synthetic row does nothing.
- [ ] **Cache isolation across orgs.** Switch from org A to org B — the All Employees access column re-renders for org B's roster (or empty if cache is cold). Switch back to org A — no extra HTTP fetch (cache warm).
- [ ] **Single fetch per org.** First visit to an org issues exactly one `GET /api/orgs/<orgUid>/access-users` request even when the Access tab and All Employees tab are mounted in sequence (shared cache + `shareReplay` dedup).
- [ ] **CI green.** DCO check, headers, lint, typecheck, build.

Generated with [Cursor Composer](https://cursor.com/composer)

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


audigregorie (commented):
## Code Review Summary

Clean, well-reasoned change — the shared-cache + UNION design is correct (cache slot keyed on the response's own `orgUid`, `shareReplay` dedup, namespaced synthetic `personKey`s, and synthetic rows properly stripped of interactive affordances with `toggleRow`/`onRowKeydown` guards). No correctness bugs found; only minor consistency/polish items. CI is green and CodeRabbit + Copilot have already reviewed.

### Minor — outside the diff

- **No unit tests** for the new root-singleton `OrgLensAccessStateService` (cache dedup, `setRoster` skip-on-empty, email lowercasing) or the UNION logic in `initViewRows()`. The sibling components also lack specs, so this is a suggestion rather than a blocker — but a shared cache with dedup/merge logic is exactly the kind of thing worth pinning down.
- **`org-lens-access.component.ts:41`** — `OrgAccessRowVm` is still a file-local `type`, while its sibling `OrgAllEmployeeRowVm` was correctly promoted to `@lfx-one/shared`. Per CLAUDE.md ("interfaces in shared, not in component files") these have drifted apart. Pre-existing and explicitly scoped out in the PR body — flagging only for the inconsistency.
- **Synthetic rows keep an interactive name button** — the access-only synthetic rows still render `<button (click)="onPersonClick(row)">`, which opens the profile panel by name even though the row is deliberately non-interactive and the person has zero activity. Confirm the panel handles a name with no backing activity record gracefully.
- **PR-description nits**: the modal header changed from `"Add Org Lens Access"` → `"Add Users"` (the body says "Add User"); and the "single fetch per org" guarantee only holds when the Access tab is visited *first* — visiting All Employees first still triggers the Access tab's own unconditional refetch on mount (by design, since the writer wants fresh data).

### What's done well

- Cache slot keyed on `res.orgUid` (composite-correct), with `EMPTY_ORG_ACCESS_LIST_RESPONSE` (empty `orgUid`) safely skipped by `setRoster`'s guard.
- Synthetic `personKey` namespacing (`access:` prefix) + the non-interactive a11y treatment is thorough and well-commented.
- Per-fetch `catchError(() => EMPTY)` on the All Employees hydration keeps the outer stream alive without swallowing the Access tab's own error/retry UX.

Rating: 4.5/5

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.html
```diff
@@ -2,27 +2,58 @@
...
+      <p class="text-sm leading-relaxed text-gray-700">
+        In the left navigation, LFX allows users to view their profile, their projects, and their company's involvement. Company involvement is restricted to a
+        small set of users with specific permissions. This page shows who has these permissions and lets you grant it to other employees.
+      </p>
+      <div class="flex flex-col gap-3 md:flex-row md:gap-6">
+        <div class="flex items-start gap-2 md:flex-1">
+          <span
+            class="inline-flex flex-shrink-0 items-center rounded-full bg-blue-50 px-2 py-0.5 text-xs font-semibold whitespace-nowrap text-blue-700"
+            data-testid="org-access-info-role-editor">
+            Org Admin - Editor
```
audigregorie: **Minor — use the en-dash to match the established role-label convention**

The existing `ORG_ACCESS_ROLE_BADGE_TOOLTIP` constant writes these role names with an en-dash (`Org Admin – Editor` / `Org Admin – Viewer`), but the banner uses a plain hyphen. Align for typographic consistency.

```suggestion
            Org Admin – Editor
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.html
```diff
@@ -2,27 +2,58 @@
...
+          </span>
+          <span class="text-xs text-gray-600">
+            Full access to manage your organization's Org Lens — can update member profiles, manage Org Lens user access, and configure organization settings.
+          </span>
+        </div>
+        <div class="flex items-start gap-2 md:flex-1">
+          <span
+            class="inline-flex flex-shrink-0 items-center rounded-full bg-gray-100 px-2 py-0.5 text-xs font-semibold whitespace-nowrap text-gray-600"
+            data-testid="org-access-info-role-viewer">
+            Org Admin - Viewer
```
audigregorie: **Minor — en-dash here too**

Same as the Editor label above — match the `ORG_ACCESS_ROLE_BADGE_TOOLTIP` convention (`Org Admin – Viewer`).

```suggestion
            Org Admin – Viewer
```

packages/shared/src/interfaces/org-people.interface.ts
```diff
@@ -59,6 +61,18 @@ export interface OrgAllEmployeeRow {
...
+export interface OrgAllEmployeeRowVm extends OrgAllEmployeeRow {
+  initials: string;
+  avatarColorClass: string;
+  /** Org Lens access for the current org, or `null` when the employee has no access record. Drives the access cell badge. */
+  access: OrgAccessBadgeState | null;
+  /** `true` when this row was synthesised from the access roster (no detected activity, no detail to fetch). */
+  isSynthetic: boolean;
+  /** Precomputed `<tr>` class list — keeps the template free of conditional class bindings on every cell. */
+  rowClass: string;
+}
```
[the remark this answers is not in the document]
ahmedomosanya: Description updated to fold the row-VM lift into Summary — `OrgAllEmployeeRowVm` and `OrgAccessBadgeState` are in-scope for this PR, landed in `61066e39` to match sibling `OrgTraineeRowVm` / `OrgEventAttendeeRowVm` / `OrgContributorRowVm` per `.claude/rules/component-organization.md` §5.

audigregorie (dismissed):

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.html
ahmedomosanya: Addressed in `6e47c68e` — replaced hyphen with U+2013 en-dash to match the `ORG_ACCESS_ROLE_BADGE_TOOLTIP` constant (`packages/shared/src/constants/org-people.constants.ts:193-194`).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/org-lens-access/org-lens-access.component.html
ahmedomosanya: Addressed in `6e47c68e` (same fix as the Editor label above) — banner now uses U+2013 en-dash to match `ORG_ACCESS_ROLE_BADGE_TOOLTIP`.

MRashad26 (commented):
Code-enforcer audit. No violations found.

**Compliance highlights worth calling out:**

- `AllEmployeeRowVm` (previously a local `type` inside `apps/lfx-one/`) is correctly promoted to `OrgAllEmployeeRowVm` in `packages/shared/src/interfaces/org-people.interface.ts`. This is a pre-existing violation that this PR fixes.
- `OrgAccessBadgeState` added directly to `packages/shared/src/interfaces/` rather than inline — correct.
- New string constants (`rowClassActivity`, `rowClassSynthetic`, `accessOnlyKeyPrefix`) are `private static readonly` class members, not module-level `const` declarations — the CLAUDE.md rule targets file-scope `const`, not class statics. Clean.
- `OrgLensAccessStateService` has no module-level constants.
- All template bindings in the changed HTML files use signal reads (`signal()`, field access on VMs, ternaries over VM fields) — no method calls.
- SSR-safe: `ensureLoaded` is called from the constructor's RxJS pipeline, not from `afterNextRender`, which is fine for a data-fetch (not DOM-dependent). The `isPlatformBrowser` guard lives upstream in the service if needed.

MRashad26 (approved):

PR #886, opened 2026-06-04
fix(newsletters): preserve spacing and indent on google docs paste

niravpatel27:
## Summary

Pasting from Google Docs into the newsletter body editor injected large vertical gaps between sections and dropped the indentation of continuation paragraphs under bullets. The TipTap paste cleaner now removes Google Docs' blank-line spacers (both `<p><span></span></p>` from the export and orphan `<br>` from the Cmd+C clipboard) and absorbs indented continuation paragraphs into the preceding `<li>`. Repo has no unit-test architect; verified end-to-end against the live editor + preview.

### DIFF
--- a/apps/lfx-one/src/app/shared/components/rich-editor/clean-pasted-html.util.ts
+++ b/apps/lfx-one/src/app/shared/components/rich-editor/clean-pasted-html.util.ts
@@ -145,7 +145,23 @@ const EMPTY_BLOCK_TAGS = new Set(['P', 'H2', 'H3', 'LI', 'UL', 'OL']);
 // any future content type the editor learns to render (images, media, embeds, form widgets).
 // HTML elements always uppercase their tagName, but namespaced elements (SVG, MathML) preserve
 // case, so we normalize via toUpperCase() at the lookup site rather than maintaining both forms.
-const CONTENT_LEAF_TAGS = new Set(['IMG', 'IFRAME', 'VIDEO', 'AUDIO', 'SOURCE', 'PICTURE', 'SVG', 'MATH', 'CANVAS', 'EMBED', 'OBJECT', 'INPUT', 'SELECT', 'TEXTAREA', 'HR']);
+const CONTENT_LEAF_TAGS = new Set([
+  'IMG',
+  'IFRAME',
+  'VIDEO',
+  'AUDIO',
+  'SOURCE',
+  'PICTURE',
+  'SVG',
+  'MATH',
+  'CANVAS',
+  'EMBED',
+  'OBJECT',
+  'INPUT',
+  'SELECT',
+  'TEXTAREA',
+  'HR',
+]);
 
 // Google Docs (and Word) inject empty <p style="height:11pt"><span></span></p> nodes as blank-line
 // spacers between content blocks. After style/class strip, the editor's `p { margin: 0 0 0.75rem }`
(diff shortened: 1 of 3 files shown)


apps/lfx-one/src/app/shared/components/rich-editor/clean-pasted-html.util.ts
```diff
@@ -30,6 +32,180 @@ function unwrapGoogleDocsWrapper(body: HTMLElement): void {
...
+    default:
+      return value;
+  }
+}
+
+function isEmptyInlineContainer(element: Element): boolean {
+  if (element.textContent && element.textContent.replace(/[\s\u00A0]/g, '') !== '') {
+    return false;
+  }
+  for (const child of Array.from(element.children)) {
```
dealako, line 130: **[minor] `isEmptyInlineContainer` doesn't guard against content-leaf elements**

`<p></p>` (or `<p><svg>…</svg></p>`) has no `textContent` and `` has no children, so this function returns `true` — treating the paragraph as a Google Docs spacer. That causes `absorbListContinuationParagraphs` to *skip* the paragraph (advancing `cursor` past it) rather than absorbing it under the preceding list item. The image paragraph stays as an orphaned sibling of the `<ul>` after the paste.

`isEmptyBlock` already handles this correctly via `CONTENT_LEAF_TAGS`. The same guard is needed here:

```ts
for (const child of Array.from(element.children)) {
  if (child.tagName === 'BR') {
    continue;
  }
  // Added: preserve paragraphs that contain visual content (images, embeds, etc.)
  if (CONTENT_LEAF_TAGS.has(child.tagName.toUpperCase())) {
    return false;
  }
  if (!isEmptyInlineContainer(child)) {
    return false;
  }
}
```

Copilot raised this in its inline comment — agreed.

dealako (approved):
Hey @niravpatel27! 👋 Great targeted fix — the two-pass approach (absorb continuations first while styles are still readable, then collapse spacers after the style strip) is exactly the right order of operations, and the deep-first list traversal prevents double-absorption. Code quality is high throughout.

---

## Subagent Reviews

- **Security audit** (independent): No Critical / High / Medium vulnerabilities. XSS is not in play (operations happen on a sandboxed `DOMParser` fragment; `walkAndClean` already strips `on*` handlers). Regex has no ReDoS risk. One **Low** finding: `isEmptyInlineContainer` and `isEmptyBlock` are recursive with unbounded depth — adversarial input with thousands of nesting levels could stack-overflow. In practice, real paste payloads are nowhere near that depth; this is acceptable.
- **Code quality review** (independent): No blockers. Flagged (a) the `1pt` tolerance as a magic number worth extracting to a named constant, and (b) absence of unit tests. The PR description already addresses (b) — no test infrastructure exists in the repo.

---

## Revision Tracking

Four commits on this branch:
| Commit | Change |
|--------|--------|
| `cc15d219` | Initial fix |
| `7a7401a7` | `fix(review): address post-commit review findings` |
| `1c6f9e41` | `fix(review): normalize tagName case in content-leaf check` |
| `f8a61d07` | `style(rich-editor): apply prettier formatting` |

The `tagName.toUpperCase()` fix in `1c6f9e41` (for SVG/MathML namespaced elements in `isEmptyBlock`) was a clean response to a prior review concern. However, the parallel `isEmptyInlineContainer` function didn't receive the equivalent guard — see inline comment.

---

## AI Comment Reconciliation

**Copilot** (`clean-pasted-html.util.ts` line 134): raised that `isEmptyInlineContainer` treats image/SVG paragraphs as empty. ✅ I **agree** — see inline comment below. This is the one issue worth addressing before merge.

**CodeRabbit**: Walkthrough is accurate. Pre-merge warning on docstring coverage (37.5% vs 80% threshold) is a repo-level quality gate; the code is otherwise well commented. ESLint check failed due to a CodeRabbit-side install issue, not a code problem.

---

## Issue Count

🔴 **Blocking**: 0  
🟡 **Minor**: 1 — `isEmptyInlineContainer` treats `<p></p>` as a spacer (image-only continuation paragraphs silently dropped)  
⚪ **Nit**: 3 — `%` unit comparison is dimensionally inconsistent; `1pt` tolerance magic number; SCSS comments are good but slightly duplicated across both files

---

## Final Decision

The one [minor] finding is a real behavioral edge case — image-only continuation paragraphs would be skipped rather than absorbed — but it won't regress the reported user story (Google Docs text bullets). Happy to approve now; a follow-up fix is fine if image-in-list-continuation pastes come up in practice.

PR #903, opened 2026-06-08
feat(events): add past events tab to org lens events page

ahmedomosanya:
## Summary
- The Past tab previously showed a placeholder. It now lists an organization's historical event footprint from the same platinum-backed endpoint as Upcoming (`isPast=true`).
- Past events reuse the upcoming table **verbatim minus the Action column**, matching the design.
- Renamed `org-upcoming-events-table` → `org-events-table` (class, selector, template, test ids) since the component now serves both tabs.
- Added `showAction` (default true), `emptyStateTitle`, `emptyStateSubtitle` inputs; the Action column header/body and the empty-message colspan are gated on `showAction`.
- Added a lazy `pastEvents` pipeline in the dashboard (signals, page/sort handlers, active-tab gating) mirroring the upcoming pipeline — it only fetches once the Past tab is opened, and paging resets on filter change.
- Removed the now-unused `EmptyStateComponent` import.
- Added e2e coverage: clicking the Past stat card loads the table and confirms the Action column is absent.

## Test plan
- [ ] Open the org lens events page; click the **Past** tab → table loads with shared columns and **no Action column**.
- [ ] Search + status filter apply to the Past tab; paging and column sorting work.
- [ ] Switching back to **Upcoming** still shows the Action column.
- [ ] Empty org → past-specific empty state copy renders.
- [ ] e2e: `org-events-dashboard.spec.ts` past-tab test passes.

[LFXV2-1900]: https://linuxfoundation.atlassian.net/browse/LFXV2-1900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/e2e/org-events-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/org-events-dashboard.spec.ts
@@ -190,6 +190,31 @@ test.describe('Org Events Dashboard', () => {
     await expect(page.getByTestId('org-event-action-event-1')).toHaveCount(0);
   });
 
+  test('past tab shows the past-specific empty state when there are no past events', async ({ page }) => {
+    await gotoOrgEventsPage(page);
+    await expect(page.getByTestId('org-events-page')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    // Override only the isPast=true response with an empty payload; registered after the
+    // base stub so it takes precedence, falling back to the base stub for upcoming requests.
+    await page.route(`**/api/orgs/${MOCK_ACCOUNT_ID}/lens/events?**`, (route) => {
+      const url = new URL(route.request().url());
+      if (url.searchParams.get('isPast') === 'true') {
+        return route.fulfill({
+          status: 200,
+          contentType: 'application/json',
+          body: JSON.stringify({ data: [], total: 0, pageSize: 10, offset: 0 }),
+        });
+      }
+      return route.fallback();
+    });
+
+    await page.getByTestId('org-events-stat-past').click();
+
+    const panel = page.getByTestId('org-events-panel-past');
+    await expect(panel).toBeVisible();
+    await expect(panel.getByText('No past events', { exact: true })).toBeVisible();
+  });
+
   test('opens attendee and speaker drawers with non-PII row test ids', async ({ page }) => {
     await gotoOrgEventsPage(page);
     await expect(page.getByTestId('org-events-page')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
--- a/apps/lfx-one/e2e/org-events-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/org-events-dashboard.spec.ts
@@ -154,8 +154,8 @@ test.describe('Org Events Dashboard', () => {
   test('supports sorting, pagination, and action CTAs', async ({ page }) => {
     await gotoOrgEventsPage(page);
     await expect(page.getByTestId('org-events-page')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
-    await expect(page.getByTestId('org-upcoming-events-data-table')).toBeVisible();
-    await expect(page.getByTestId('org-upcoming-event-action-event-1').getByRole('link', { name: 'Register' })).toBeVisible();
+    await expect(page.getByTestId('org-events-data-table')).toBeVisible();
+    await expect(page.getByTestId('org-event-action-event-1').getByRole('link', { name: 'Register' })).toBeVisible();
 
     const sortRequest = page.waitForRequest((request) => {
       const url = new URL(request.url());
@@ -172,18 +172,36 @@ test.describe('Org Events Dashboard', () => {
     await pageRequest;
   });
 
+  test('past tab renders the events table without the Action column', async ({ page }) => {
+    await gotoOrgEventsPage(page);
+    await expect(page.getByTestId('org-events-page')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    const pastRequest = page.waitForRequest((request) => {
+      const url = new URL(request.url());
+      return url.pathname.endsWith(`/api/orgs/${MOCK_ACCOUNT_ID}/lens/events`) && url.searchParams.get('isPast') === 'true';
+    });
+    await page.getByTestId('org-events-stat-past').click();
+    await pastRequest;
+
+    await expect(page.getByTestId('org-events-panel-past')).toBeVisible();
+    await expect(page.getByTestId('org-events-data-table')).toBeVisible();
+    // Shared columns still render, but the Action column is dropped on the past tab.
+    await expect(page.getByTestId('org-event-attendees-event-1')).toBeVisible();
+    await expect(page.getByTestId('org-event-action-event-1')).toHaveCount(0);
+  });
+
   test('opens attendee and speaker drawers with non-PII row test ids', async ({ page }) => {
     await gotoOrgEventsPage(page);
     await expect(page.getByTestId('org-events-page')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
 
-    await page.getByTestId('org-upcoming-event-attendees-event-1').getByRole('button').click();
+    await page.getByTestId('org-event-attendees-event-1').getByRole('button').click();
     await expect(page.getByTestId('event-attendees-drawer')).toBeVisible();
     await expect(page.getByTestId('event-attendee-0')).toBeVisible();
 
     await page.keyboard.press('Escape');
     await expect(page.getByTestId('event-attendees-drawer')).not.toBeVisible();
 
-    await page.getByTestId('org-upcoming-event-speakers-event-1').getByRole('button').click();
+    await page.getByTestId('org-event-speakers-event-1').getByRole('button').click();
     await expect(page.getByTestId('event-speakers-drawer')).toBeVisible();
     await expect(page.getByTestId('event-speaker-0')).toBeVisible();
   });
--- a/apps/lfx-one/src/app/modules/events/org-events-dashboard/components/org-events-table/org-events-table.component.html
+++ b/apps/lfx-one/src/app/modules/events/org-events-dashboard/components/org-events-table/org-events-table.component.html
@@ -177,11 +177,11 @@
 
   <ng-template #emptymessage>
     <tr>
-      <td [attr.colspan]="showAction() ? 8 : 7" class="text-center py-10">
+      <td [attr.colspan]="columnCount()" class="text-center py-10">
         <div class="flex items-center justify-center p-16">
           <div class="text-center max-w-md">
             <div class="text-gray-400 mb-4">
-              <i class="fa-light fa-calendar-plus text-3xl mb-4"></i>
+              <i [class]="emptyStateIcon() + ' text-3xl mb-4'"></i>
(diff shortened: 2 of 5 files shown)


apps/lfx-one/src/app/modules/events/org-events-dashboard/org-events-dashboard.component.ts
```diff
@@ -210,4 +231,44 @@ export class OrgEventsDashboardComponent {
       { initialValue: EMPTY_ORG_EVENTS_RESPONSE }
     );
   }
+
+  private initPastEvents(): Signal<OrgEventsResponse> {
```
audigregorie: **Minor — `initPastEvents()` duplicates `initUpcomingEvents()` nearly verbatim**

This pipeline is a ~38-line copy of `initUpcomingEvents()` (lines 195–233); only `isPast`, the tab guard, the per-tab signals, and the error-toast copy differ. The `onPast*` handlers similarly mirror `onUpcoming*`.

Consider collapsing both into one private factory parameterized over the differing bits, e.g.:

```ts
private initEventsPipeline(opts: {
  tab: OrgEventsTabId;
  page: Signal<PageChangeEvent>;
  sortField: Signal<string>;
  sortOrder: Signal<'ASC' | 'DESC'>;
  loading: WritableSignal<boolean>;
  isPast: boolean;
  errorDetail: string;
}): Signal<OrgEventsResponse> { /* shared body */ }
```

Defensible as separate parallel pipelines, hence Minor — but it's the main refactor opportunity in this PR.

apps/lfx-one/src/app/modules/events/org-events-dashboard/components/org-events-table/org-events-table.component.html
```diff
@@ -142,44 +144,46 @@
...
   </ng-template>
 
   <ng-template #emptymessage>
     <tr>
-      <td colspan="8" class="text-center py-10">
+      <td [attr.colspan]="showAction() ? 8 : 7" class="text-center py-10">
         <div class="flex items-center justify-center p-16">
           <div class="text-center max-w-md">
             <div class="text-gray-400 mb-4">
               <i class="fa-light fa-calendar-plus text-3xl mb-4"></i>
```
audigregorie: **Minor — empty-state icon is hardcoded to the "add event" glyph on both tabs**

The empty-state title and subtitle were parameterized for reuse, but the icon stays `fa-light fa-calendar-plus` (an add/upcoming glyph). On the Past tab this renders next to "No past events", which is inconsistent with the Past stat card's `fa-light fa-calendar-check`.

Suggest adding an `emptyStateIcon` input (default `'fa-light fa-calendar-plus'`) and binding it here, then passing `fa-light fa-calendar-check` from the past `<lfx-org-events-table>`. Not offering a one-click suggestion since the fix spans the `.ts` (new input) and both call sites.

audigregorie (commented):
## Code Review Summary

Clean, well-structured PR. The component rename is thorough (selector, template, all test ids, parent import), the `showAction` / `emptyState*` inputs are a sensible way to share one table across both tabs, the lazy past-events pipeline correctly gates on the active tab, and the e2e test asserts the Action column is genuinely absent (count 0) rather than just checking the table renders. No correctness bugs found — both threads below are quality/consistency only.

### Observation (non-blocking)

Both pipelines refetch on every tab activation — the gating `computed` emits a fresh params object each time the tab becomes active and there's no `distinctUntilChanged`. This is consistent with the existing upcoming pipeline and yields fresh data, so it's fine. Just note the "lazy-load: only fetch ... once the past tab is opened" comment reads as fetch-exactly-once, which isn't quite the behavior.

### What's done well

- Rename is complete and consistent across selector, template, every `data-testid`, and the parent import.
- `[attr.colspan]="showAction() ? 8 : 7"` keeps the empty row spanning full width in both modes.
- e2e test verifies the Action column gating (`org-event-action-event-1` count 0), not just table presence.

apps/lfx-one/src/app/modules/events/org-events-dashboard/org-events-dashboard.component.ts
MRashad26: **Problem:** `initPastEvents()` is a near-verbatim copy of `initUpcomingEvents()` (~40 lines). The only differences are the tab guard (`'past'` vs `'upcoming'`), the page/sort signals, the loading signal, the `isPast` flag, and the error copy.

**Why it matters:** Duplicated query pipelines drift apart over time — a fix to one (debounce, error handling, param shape) silently misses the other.

**Suggested fix:** Extract one private factory and call it for both tabs:
```ts
private initEventsPipeline(opts: {
  tab: OrgEventsTabId;
  page: Signal<PageChangeEvent>;
  sortField: Signal<string>;
  sortOrder: Signal<'ASC' | 'DESC'>;
  loading: WritableSignal<boolean>;
  isPast: boolean;
  errorDetail: string;
}): Signal<OrgEventsResponse> { /* shared body */ }
```

apps/lfx-one/src/app/modules/events/org-events-dashboard/components/org-events-table/org-events-table.component.html
```diff
@@ -142,44 +144,46 @@
...
+          }
+        </td>
+      }
     </tr>
   </ng-template>
 
   <ng-template #emptymessage>
     <tr>
-      <td colspan="8" class="text-center py-10">
+      <td [attr.colspan]="showAction() ? 8 : 7" class="text-center py-10">
```
MRashad26: **Problem:** The empty-state colspan uses hardcoded literals `showAction() ? 8 : 7`.

**Why it matters:** If a column is ever added or removed, these literals silently desync from the actual column count and the empty row spans wrong.

**Suggested fix:** Confirm the table truly has 8 columns (7 without Action). Optionally derive the count from a single source (e.g. a `columnCount` computed) so header, body, and empty row stay in lockstep. Low severity.

apps/lfx-one/e2e/org-events-dashboard.spec.ts
```diff
@@ -172,18 +172,36 @@ test.describe('Org Events Dashboard', () => {
...
+      return url.pathname.endsWith(`/api/orgs/${MOCK_ACCOUNT_ID}/lens/events`) && url.searchParams.get('isPast') === 'true';
+    });
+    await page.getByTestId('org-events-stat-past').click();
+    await pastRequest;
+
+    await expect(page.getByTestId('org-events-panel-past')).toBeVisible();
+    await expect(page.getByTestId('org-events-data-table')).toBeVisible();
+    // Shared columns still render, but the Action column is dropped on the past tab.
+    await expect(page.getByTestId('org-event-attendees-event-1')).toBeVisible();
+    await expect(page.getByTestId('org-event-action-event-1')).toHaveCount(0);
```
MRashad26, line 190: **Problem:** The new test covers the primary past-tab path well (request fires with `isPast=true`, table renders, Action column absent), but the new `emptyStateTitle`/`emptyStateSubtitle` inputs and past-tab paging/sorting are untested (only listed in the manual test plan).

**Why it matters:** The past-specific empty-state copy is a new code path with zero automated coverage.

**Suggested fix:** Add a short assertion for the past empty-state copy ("No past events") when the org has no past events, mirroring how the upcoming empty state is exercised.

MRashad26 (commented):
Standards audit (code-enforcer ruleset) — overall **PASS with minor cleanups**. The lazy tab-gated `pastEvents` pipeline, signal/`toObservable` usage, shared-interface imports, and template (no function calls) all follow conventions. Three inline notes below.

Two items that aren't line-anchorable:
- **Test plan in PR body** — the enforcer ruleset says *no test plans in pull requests*. Consider moving the `## Test plan` checklist to a comment/local notes.
- **Stale test-id check ✅** — confirmed every `org-upcoming-event-*` reference lives inside the files this PR already renames/updates; no leftover references elsewhere.

apps/lfx-one/src/app/modules/events/org-events-dashboard/org-events-dashboard.component.ts
ahmedomosanya: Addressed in 88527c87 — extracted a single `initEventsPipeline({ tab, page, sortField, sortOrder, loading, isPast, errorDetail })` factory. `upcomingEvents` and `pastEvents` now both call it instead of carrying two near-identical pipelines, so a fix to one (debounce, error handling, param shape) can't silently miss the other.

ahmedomosanya: Addressed in 88527c87 — collapsed both pipelines into one private `initEventsPipeline(opts)` factory parameterized over the differing bits (tab guard, page/sort/loading signals, `isPast`, error copy). Upcoming and past now share the same query body.

apps/lfx-one/src/app/modules/events/org-events-dashboard/components/org-events-table/org-events-table.component.html
ahmedomosanya: Addressed in 88527c87 — added an `emptyStateIcon` input (default `fa-light fa-calendar-plus`) bound on the empty-state icon; the past `<lfx-org-events-table>` now passes `fa-light fa-calendar-check` to match the Past stat card.

apps/lfx-one/src/app/modules/events/org-events-dashboard/components/org-events-table/org-events-table.component.html
ahmedomosanya: Addressed in 88527c87 — confirmed 8 columns (7 without Action). Replaced the `showAction() ? 8 : 7` literal with a single `columnCount` computed bound to the empty-row colspan, so header, body, and empty row stay in lockstep.

apps/lfx-one/e2e/org-events-dashboard.spec.ts
ahmedomosanya: Addressed in 88527c87 — added an e2e test that overrides the `isPast=true` response with an empty payload and asserts the past-specific empty-state copy ("No past events") renders.

apps/lfx-one/src/app/modules/events/org-events-dashboard/org-events-dashboard.component.ts
```diff
@@ -173,37 +208,48 @@ export class OrgEventsDashboardComponent {
...
-            sortOrder: this.upcomingSortOrder(),
+            sortField: sortField(),
+            sortOrder: sortOrder(),
           };
         })
       ).pipe(
         debounceTime(0),
         filter((params): params is NonNullable<typeof params> => params !== null),
-        tap(() => this.upcomingEventsLoading.set(true)),
+        tap(() => loading.set(true)),
```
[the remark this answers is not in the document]
ahmedomosanya: Deferring — this is the pre-existing lazy-load characteristic carried over from the upcoming-events work (#834): each tab only fetches while it's the active tab, so the inactive tab keeps its last `OrgEventsResponse`. The stale window is a single render tick before the lazy fetch replaces it, and switching tabs flips the tab's loading state. A correct fix is per-tab context invalidation (clearing the cached response + loading when account/search/status change), which is self-contained and better applied uniformly to both tabs with its own test coverage rather than bolted into this review pass. Tracked as a follow-up on LFXV2-1900.

MRashad26 (approved):

PR #911, opened 2026-06-09
feat(campaigns): add linkedin ad accounts and set default

mrautela365:
## Summary
- Add 9 new LinkedIn ad accounts to server constants (CNCF, LF Education, Agentic AI Foundation, OpenSearch Project x2, LF's Ad Account x2, OpenJS Foundation, OpenSSF)
- Set LF Events (509430019) as the default LinkedIn ad account — env var `LINKEDIN_AD_ACCOUNT_ID` overrides when set

## JIRA
[LFXV2-2180](https://linuxfoundation.atlassian.net/browse/LFXV2-2180) — child of [LFXV2-2023](https://linuxfoundation.atlassian.net/browse/LFXV2-2023)

## Files changed
- `apps/lfx-one/src/server/constants/linkedin.constants.ts` — added accounts + default constants
- `apps/lfx-one/src/server/services/linkedin-ads.service.ts` — use default fallback for account/org ID

[LFXV2-2180]: https://linuxfoundation.atlassian.net/browse/LFXV2-2180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-2023]: https://linuxfoundation.atlassian.net/browse/LFXV2-2023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/constants/linkedin.constants.ts
+++ b/apps/lfx-one/src/server/constants/linkedin.constants.ts
@@ -11,10 +11,23 @@ import type { LinkedInTargetingProfileConfig } from '@lfx-one/shared/interfaces'
 // UI-safe constants (geo resolve map, char limits) remain in @lfx-one/shared.
 // ---------------------------------------------------------------------------
 
+export const LINKEDIN_DEFAULT_ACCOUNT_ID = '509430019';
+
 export const LINKEDIN_ACCOUNTS: readonly { accountId: string; label: string; orgId: string }[] = [
   { accountId: '538170226', label: 'The Linux Foundation', orgId: '208777' },
   { accountId: '509430019', label: 'LF Events', orgId: '208777' },
+  { accountId: '500928401', label: 'Cloud Native Computing Foundation', orgId: '12893459' },
+  { accountId: '508209098', label: 'Linux Foundation - Education', orgId: '7953130' },
+  { accountId: '537341179', label: 'Agentic AI Foundation', orgId: '111268938' },
+  { accountId: '514914665', label: 'OpenSearch Project (2nd Account)', orgId: '78470501' },
+  { accountId: '515250253', label: 'LF OpenJS Ad Account', orgId: '19082105' },
+  { accountId: '531310265', label: 'LF Agentic AI Ad Account', orgId: '111268938' },
+  { accountId: '515244770', label: 'OpenJS Foundation', orgId: '19082105' },
+  { accountId: '514596831', label: 'OpenSSF', orgId: '76521837' },
+  { accountId: '514553720', label: 'OpenSearch Project', orgId: '78470501' },
 ] as const;
+
+export const LINKEDIN_DEFAULT_ORG_ID = LINKEDIN_ACCOUNTS.find((a) => a.accountId === LINKEDIN_DEFAULT_ACCOUNT_ID)!.orgId;
 
 export const LINKEDIN_EMPLOYER_EXCLUSIONS: readonly string[] = ['urn:li:company:33275771', 'urn:li:company:12893459'] as const;
 
--- a/apps/lfx-one/src/server/services/linkedin-ads.service.ts
+++ b/apps/lfx-one/src/server/services/linkedin-ads.service.ts
@@ -4,7 +4,13 @@
 import type { LinkedInCampaignCreateRequest, LinkedInCampaignCreateResult, LinkedInGeoTarget, LinkedInTargetingProfile } from '@lfx-one/shared/interfaces';
 
 import { LINKEDIN_AD_ACCOUNTS, LINKEDIN_API_VERSION, LINKEDIN_GEO_RESOLVE_MAP } from '@lfx-one/shared/constants';
-import { LINKEDIN_EMPLOYER_EXCLUSIONS, LINKEDIN_TARGETING_PROFILES } from '../constants';
+import {
+  LINKEDIN_ACCOUNTS,
+  LINKEDIN_DEFAULT_ACCOUNT_ID,
+  LINKEDIN_DEFAULT_ORG_ID,
+  LINKEDIN_EMPLOYER_EXCLUSIONS,
+  LINKEDIN_TARGETING_PROFILES,
+} from '../constants';
 
 import type { Request } from 'express';
 
@@ -35,18 +41,18 @@ function getLinkedInEnv(key: string): string {
 }
 
 function resolveAccountId(override?: string): string {
-  const id = override || getLinkedInEnv('LINKEDIN_AD_ACCOUNT_ID');
+  const envValue = process.env['LINKEDIN_AD_ACCOUNT_ID'];
+  const id = override || envValue || LINKEDIN_DEFAULT_ACCOUNT_ID;
   if (!/^\d+$/.test(id)) {
     throw new Error(`Invalid LinkedIn account ID: must be numeric, got "${id}"`);
   }
   if (override && !LINKEDIN_AD_ACCOUNTS.some((a) => a.accountId === id)) {
     throw new Error(`Unsupported LinkedIn ad account ID: "${id}"`);
   }
+  if (!envValue && !override) {
+    logger.debug(undefined, 'linkedin_config', `LINKEDIN_AD_ACCOUNT_ID not set — using default: ${LINKEDIN_DEFAULT_ACCOUNT_ID}`);
+  }
   return id;
-}
-
-function getOrgId(): string {
-  return getLinkedInEnv('LINKEDIN_ORG_ID');
 }
 
 function getAccessToken(): string {
@@ -177,9 +183,13 @@ function accountUrn(accountId: string): string {
 }
 
 function resolveOrgId(accountId: string): string {
-  const account = LINKEDIN_AD_ACCOUNTS.find((a) => a.accountId === accountId);
-  if (account) return account.organizationId;
-  return getOrgId();
+  const envOrgId = process.env['LINKEDIN_ORG_ID'];
+  if (envOrgId) return envOrgId;
+  const sharedAccount = LINKEDIN_AD_ACCOUNTS.find((a) => a.accountId === accountId);
+  if (sharedAccount) return sharedAccount.organizationId;
+  const serverAccount = LINKEDIN_ACCOUNTS.find((a) => a.accountId === accountId);
+  if (serverAccount) return serverAccount.orgId;
+  return LINKEDIN_DEFAULT_ORG_ID;
 }
 
 function orgUrn(accountId: string): string {
--- a/packages/shared/src/constants/campaign.constants.ts
+++ b/packages/shared/src/constants/campaign.constants.ts
@@ -118,13 +118,18 @@ export const LINKEDIN_CHAR_LIMITS = {
 export const LINKEDIN_AD_ACCOUNTS: readonly LinkedInAdAccount[] = [
   { accountId: '538170226', label: 'The Linux Foundation', organizationId: '208777', status: 'ACTIVE' },
   { accountId: '509430019', label: 'LF Events', organizationId: '208777', status: 'ACTIVE' },
-  { accountId: '510263296', label: 'CNCF', organizationId: '12893459', status: 'ACTIVE' },
-  { accountId: '510263297', label: 'LF Networking', organizationId: '208777', status: 'ACTIVE' },
-  { accountId: '510263298', label: 'LF AI & Data', organizationId: '208777', status: 'ACTIVE' },
-  { accountId: '510263299', label: 'LF Energy', organizationId: '208777', status: 'ACTIVE' },
+  { accountId: '500928401', label: 'Cloud Native Computing Foundation', organizationId: '12893459', status: 'ACTIVE' },
+  { accountId: '508209098', label: 'Linux Foundation - Education', organizationId: '7953130', status: 'ACTIVE' },
+  { accountId: '537341179', label: 'Agentic AI Foundation', organizationId: '111268938', status: 'ACTIVE' },
+  { accountId: '514914665', label: 'OpenSearch Project (2nd Account)', organizationId: '78470501', status: 'ACTIVE' },
+  { accountId: '515250253', label: 'LF OpenJS Ad Account', organizationId: '19082105', status: 'ACTIVE' },
+  { accountId: '531310265', label: 'LF Agentic AI Ad Account', organizationId: '111268938', status: 'ACTIVE' },
+  { accountId: '515244770', label: 'OpenJS Foundation', organizationId: '19082105', status: 'ACTIVE' },
+  { accountId: '514596831', label: 'OpenSSF', organizationId: '76521837', status: 'ACTIVE' },
+  { accountId: '514553720', label: 'OpenSearch Project', organizationId: '78470501', status: 'ACTIVE' },
 ] as const;
 
-export const LINKEDIN_DEFAULT_ACCOUNT_ID = '538170226';
+export const LINKEDIN_DEFAULT_ACCOUNT_ID = '509430019';
 
 export const LINKEDIN_GEO_RESOLVE_MAP: Readonly<Record<string, LinkedInGeoTarget>> = {
   japan: { label: 'Japan', urn: 'urn:li:geo:101355337' },


apps/lfx-one/src/server/services/linkedin-ads.service.ts
```diff
@@ -38,11 +38,11 @@ function getLinkedInEnv(key: string): string {
...
 
 function getAccountId(): string {
-  return getLinkedInEnv('LINKEDIN_AD_ACCOUNT_ID');
+  return process.env['LINKEDIN_AD_ACCOUNT_ID'] || LINKEDIN_DEFAULT_ACCOUNT_ID;
 }
 
 function getOrgId(): string {
-  return getLinkedInEnv('LINKEDIN_ORG_ID');
+  return process.env['LINKEDIN_ORG_ID'] || LINKEDIN_DEFAULT_ORG_ID;
 }
```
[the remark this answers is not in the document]
mrautela365: Done — `getOrgId()` now auto-resolves the org ID from the `LINKEDIN_ACCOUNTS` array when `LINKEDIN_AD_ACCOUNT_ID` is overridden. If the account ID is found in the array, it uses the matched org ID. If not found, it logs a warning and falls back to the default. This prevents silent account/org mismatches.

Also added debug logging to both `getAccountId()` and `getOrgId()` for deployment observability.

See commit c84e56eb.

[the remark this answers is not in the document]
mrautela365: Done — added debug logging to both `getAccountId()` and `getOrgId()` so operators can see which source (env var vs default vs auto-resolved) and actual ID is being used. Used DEBUG level since these functions are called on every API request.

See commit c84e56eb.

mrautela365:
## Review Feedback Addressed

Commit: c84e56eb6ae28b4abe722e3cfa548092cc9dafc2

### Changes Made
- **linkedin-ads.service.ts**: `getOrgId()` now auto-resolves org ID from `LINKEDIN_ACCOUNTS` when a non-default ad account is set via env var — prevents silent account/org mismatch that could cause 4xx errors on dark post creation (per Copilot)
- **linkedin-ads.service.ts**: Added debug logging to `getAccountId()` and `getOrgId()` so operators can see which source (env var / default / auto-resolved) and value is active (per coderabbitai[bot])

### Threads Resolved
1 of 1 unresolved threads addressed. CodeRabbit nitpick also addressed in the same commit.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
```diff
@@ -38,11 +44,41 @@ function getLinkedInEnv(key: string): string {
...
 
 function getAccountId(): string {
-  return getLinkedInEnv('LINKEDIN_AD_ACCOUNT_ID');
+  try {
+    const envValue = getLinkedInEnv('LINKEDIN_AD_ACCOUNT_ID');
+    logger.debug(undefined, 'linkedin_config', `Using LinkedIn account from env: ${envValue}`);
+    return envValue;
```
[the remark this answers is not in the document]
mrautela365: The current implementation already addresses this — `getAccountId()` checks `process.env` directly (no throw/catch) and falls back to the default account with a `logger.debug` log, not a warning. The `getLinkedInEnv()` helper is only used by `getAccessToken()`, which genuinely requires the env var to be set.

See lines 46-50:
```typescript
function getAccountId(): string {
  const envValue = process.env['LINKEDIN_AD_ACCOUNT_ID'];
  if (envValue) return envValue;
  logger.debug(undefined, 'linkedin_config', `LINKEDIN_AD_ACCOUNT_ID not set — using default`);
  return LINKEDIN_DEFAULT_ACCOUNT_ID;
}
```

No exception-driven flow, no warning noise.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
```diff
@@ -38,11 +44,41 @@ function getLinkedInEnv(key: string): string {
...
+  } catch {
+    // No explicit env var — auto-resolve from the accounts list
+  }
+
+  const accountId = getAccountId();
+  if (accountId !== LINKEDIN_DEFAULT_ACCOUNT_ID) {
+    const match = LINKEDIN_ACCOUNTS.find((a) => a.accountId === accountId);
+    if (match) {
+      logger.debug(undefined, 'linkedin_config', `Auto-resolved org ${match.orgId} for account ${accountId} (${match.label})`);
+      return match.orgId;
```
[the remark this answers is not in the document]
mrautela365: Already addressed — `getOrgId()` uses direct `process.env` checks (no throw/catch from `getLinkedInEnv()`). The WARNING at line 64 only fires in the genuinely suspicious case: when `LINKEDIN_AD_ACCOUNT_ID` is set to a non-default account that doesn't exist in `LINKEDIN_ACCOUNTS`. That's a misconfiguration worth logging at WARN.

The normal default path (no env override) returns `LINKEDIN_DEFAULT_ORG_ID` silently — no warning emitted.

mrautela365:
## Review Feedback Addressed

### No Change Needed
- **linkedin-ads.service.ts**: `getAccountId()` already uses direct `process.env` checks with `logger.debug` fallback — no exception-driven flow (flagged by copilot[bot])
- **linkedin-ads.service.ts:58-72**: `getOrgId()` already uses direct env checks; WARNING only fires for the genuinely suspicious case (non-default account not found in configured accounts list) (flagged by copilot[bot])

### Threads Resolved
2 of 2 unresolved threads addressed — both were based on a stale view of the code; the current implementation already follows the suggested patterns.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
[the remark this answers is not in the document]
mrautela365: Done — simplified `getAccountId()` and `getOrgId()` to use direct `process.env` checks instead of exception-driven control flow, and downgraded expected-path logs from warning to debug. See commit 0dd99598.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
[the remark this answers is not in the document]
mrautela365: Fixed — fallback to default account/org now logs at debug level instead of warning, since it's the expected path when env vars aren't set. See commit 0dd99598.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
```diff
@@ -38,11 +44,31 @@ function getLinkedInEnv(key: string): string {
...
+  if (envOrgId) return envOrgId;
+
+  const accountId = getAccountId();
+  if (accountId !== LINKEDIN_DEFAULT_ACCOUNT_ID) {
+    const match = LINKEDIN_ACCOUNTS.find((a) => a.accountId === accountId);
+    if (match) {
+      logger.debug(undefined, 'linkedin_config', `Auto-resolved org ${match.orgId} for account ${accountId} (${match.label})`);
+      return match.orgId;
+    }
+    logger.warning(
```
MRashad26: 🔵 **Info — unknown-account override silently runs against the default org (account/org mismatch), and the whole resolver moved from fail-fast to always-fallback.**

If `LINKEDIN_AD_ACCOUNT_ID` is set to a value **not** in `LINKEDIN_ACCOUNTS` and `LINKEDIN_ORG_ID` is unset, this logs a warning and proceeds with `LINKEDIN_DEFAULT_ORG_ID` — i.e. an unrecognized ad account paired with LF Events' org `208777`. Combined with the broader change (the old `getLinkedInEnv` **threw** when these env vars were missing; now both getters always return a default), a misconfigured deployment no longer fails fast — it silently issues LinkedIn API calls against the wrong org.

**Why it matters:** an operator who intentionally overrides the account via env but forgets `LINKEDIN_ORG_ID` gets campaigns created under the default org with no hard error — easy to miss in prod, and the resulting data lands on the wrong organization.

**Suggested fix:** for the *explicit-override-but-unknown-account* case specifically, prefer throwing (fail-fast on clear misconfiguration) rather than falling back to a mismatched org. The default-when-nothing-set path is fine to keep as a silent default.

apps/lfx-one/src/server/constants/linkedin.constants.ts
```diff
@@ -11,9 +11,27 @@ import type { LinkedInTargetingProfileConfig } from '@lfx-one/shared/interfaces'
...
 
-export const LINKEDIN_ACCOUNTS: readonly { accountId: string; label: string; orgId: string }[] = [
+export interface LinkedInAdAccountConfig {
+  readonly accountId: string;
+  readonly label: string;
+  readonly orgId: string;
+}
+
+export const LINKEDIN_DEFAULT_ACCOUNT_ID = '509430019';
+export const LINKEDIN_DEFAULT_ORG_ID = '208777';
```
MRashad26: 🔵 **Info — `LINKEDIN_DEFAULT_ORG_ID` duplicates the orgId of the default account's entry.**

`509430019` (LF Events) appears in `LINKEDIN_ACCOUNTS` with `orgId: '208777'`, and `LINKEDIN_DEFAULT_ORG_ID` re-hardcodes `'208777'`. If the default account's org is ever corrected in the array but not here (or vice-versa), they silently drift.

**Suggested fix:** derive it — e.g. `export const LINKEDIN_DEFAULT_ORG_ID = LINKEDIN_ACCOUNTS.find((a) => a.accountId === LINKEDIN_DEFAULT_ACCOUNT_ID)!.orgId;` — to keep a single source of truth (declaration order permitting).

apps/lfx-one/src/server/constants/linkedin.constants.ts
```diff
@@ -11,9 +11,27 @@ import type { LinkedInTargetingProfileConfig } from '@lfx-one/shared/interfaces'
 // UI-safe constants (geo resolve map, char limits) remain in @lfx-one/shared.
 // ---------------------------------------------------------------------------
 
-export const LINKEDIN_ACCOUNTS: readonly { accountId: string; label: string; orgId: string }[] = [
+export interface LinkedInAdAccountConfig {
```
MRashad26: 🔵 **Info — local `LinkedInAdAccountConfig` interface vs the "interfaces live in `@lfx-one/shared`" convention.**

CLAUDE.md says shared interfaces belong in `@lfx-one/shared` with no local `interface Foo {}` inside `apps/lfx-one/`. This one is declared locally. That said, this file is deliberately **server-only** (its header: these IDs *"must NOT ship to the client bundle"*) and already holds module-level consts as an accepted exception, so keeping the type that *describes* those consts local is internally consistent — and a TS interface is compile-time-only, so moving it to shared wouldn't leak data either way.

**Suggested fix:** either move `LinkedInAdAccountConfig` to `@lfx-one/shared/interfaces` for strict rule adherence, or leave it here and treat server-only config types as a documented exception. Flagging so the team picks one intentionally — this mirrors the `LinkedInTargetingProfileConfig` placement question raised on #895.

MRashad26 (commented):
Automated review + `code-standards-enforcer` audit (report-only). ⚠️ Note this PR carries the **`do-not-merge`** label — review only.

## 🔒 Secrets / sensitive-data check (requested)
**No secrets exposed.** No credentials, tokens, or keys are added in this diff. The LinkedIn access token remains env-sourced via `getAccessToken()` → `getLinkedInEnv('LINKEDIN_ACCESS_TOKEN')` (untouched). The added values are **ad-account IDs and org IDs** — non-credential identifiers (org IDs are public LinkedIn company URNs; account IDs are inert without the access token). They are correctly placed in the server-only `linkedin.constants.ts`, whose header states they *"must NOT ship to the client bundle"* — so they stay out of the browser bundle. They are now committed to a public repo, which is acceptable since none are secrets.

## Overview
Adds 9 LinkedIn ad accounts to `LINKEDIN_ACCOUNTS`, introduces `LINKEDIN_DEFAULT_ACCOUNT_ID`/`LINKEDIN_DEFAULT_ORG_ID` (LF Events / 208777), and changes `getAccountId()`/`getOrgId()` from throw-on-missing-env to default-with-fallback, auto-resolving org from account when overridden.

## Code Standards Audit
- **🔴 Critical:** none
- **🟡 Warning:** none
- **🔵 Info:** local `LinkedInAdAccountConfig` interface vs the "interfaces live in `@lfx-one/shared`" rule (inline — with the server-only nuance); account/org mismatch on unknown-account override now proceeds with only a warn (inline); `LINKEDIN_DEFAULT_ORG_ID` duplicates the default account's orgId (inline); PR body has the `🤖 Generated with [Claude Code]` footer (commits themselves are clean — signed-off, no AI co-author ✅); no `## Test plan` section ✅.
- **✅ Correctly followed:** server-only constants kept out of shared/client per the file's documented architecture, MIT header, `logger.warning`/`logger.debug` (no `console.log`), all lines ≤160, signed-off commits.

## Test coverage
No tests added. `getOrgId()` now has a real resolution matrix (env set / known account / unknown account / default) but the functions are module-private, so they aren't directly testable without export. If this logic is worth locking down, consider exporting a small `resolveLinkedInOrg(accountId, env)` helper and unit-testing the four branches.

## Verdict: PASS (code) — gated by the `do-not-merge` label.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
mrautela365: Done — rebased on main which already has `resolveAccountId` with fail-fast on unknown overrides (throws `Unsupported LinkedIn ad account ID` when override isn't in the allowlist). Added default fallback only for the no-env-var case — when neither override nor env var is set, falls back to `LINKEDIN_DEFAULT_ACCOUNT_ID` with a debug log. Updated `resolveOrgId` to check env var → shared accounts → server accounts → default org, so there's no account/org mismatch. See commit 7e755934.

apps/lfx-one/src/server/constants/linkedin.constants.ts
mrautela365: Done — `LINKEDIN_DEFAULT_ORG_ID` is now derived from the account lookup: `LINKEDIN_ACCOUNTS.find((a) => a.accountId === LINKEDIN_DEFAULT_ACCOUNT_ID)!.orgId`. Single source of truth, no drift risk. See commit 7e755934.

apps/lfx-one/src/server/constants/linkedin.constants.ts
mrautela365: Done — removed the local `LinkedInAdAccountConfig` interface. The server-side `LINKEDIN_ACCOUNTS` array now uses an inline readonly type `{ accountId: string; label: string; orgId: string }` (same as the original on main). The shared `LinkedInAdAccount` interface in `@lfx-one/shared/interfaces` remains the canonical type for client-visible account data. See commit 7e755934.

mrautela365:
## Review Feedback Addressed

Commit: 7e755934

Rebased on main (post #902 + #917 merges) and addressed all 3 review comments.

### Changes Made
- **linkedin-ads.service.ts**: `resolveAccountId` now falls back to `LINKEDIN_DEFAULT_ACCOUNT_ID` when env var is unset (with debug log), but still throws on unknown overrides (fail-fast per @MRashad26). `resolveOrgId` updated to chain: env var → shared accounts → server accounts → default org — no account/org mismatch possible.
- **linkedin.constants.ts**: `LINKEDIN_DEFAULT_ORG_ID` now derived from account lookup (`LINKEDIN_ACCOUNTS.find(...)!.orgId`) — single source of truth, no drift risk. Removed local `LinkedInAdAccountConfig` interface — uses inline type matching the original on main.
- **campaign.constants.ts** (shared): Added 5 new accounts (CNCF, LF Education, Agentic AI, OpenSearch x2, OpenJS x2, OpenSSF). Changed default to LF Events (`509430019`).

### Threads Resolved
3 of 3 unresolved threads addressed.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
mrautela365: Done — removed `accountId` from `LinkedInMonitorResponse` interface. The field was only needed server-side and was leaking to the client. See commit 1e5be672.

dealako:
Merged the following instead:

- https://github.com/linuxfoundation/lfx-self-serve/pull/916
- https://github.com/linuxfoundation/lfx-v2-argocd/pull/938

PR #918, opened 2026-06-10
fix(committees): hoist toObservable out of switchMap injection context (LFXV2-2183)

audigregorie:
# Summary

Opening a committee via a deep link to a specific tab triggered NG0203 because `toObservable()` ran inside a `switchMap` callback, outside Angular's injection context. This change hoists the `visibleTabs` and `membersLoading` streams to construction time so they are created within the injection context, fixing the crash on deep-linked tab opens.


ahmedomosanya (approved):
**Verdict:** Looks good to merge. The fix is minimal, correctly scoped, and the inline comment will help prevent regressions.

MRashad26 (approved):

PR #919, opened 2026-06-10
fix(events): normalize scheme-less event urls to absolute https

ahmedomosanya:
> **WIP / Draft.** Normalizes DB-sourced event URLs so scheme-less values stop rendering as broken relative links.

Ticket: [LFXV2-2185](https://linuxfoundation.atlassian.net/browse/LFXV2-2185)

## Summary
- Some event URLs are stored without a scheme (e.g. `linuxfoundation.regfox.com/<slug>`). Bound directly to a template `[href]`, the browser resolves them as relative paths against the current origin, producing broken links like `https://app.dev.lfx.dev/linuxfoundation.regfox.com/...` (a 404 on the app).
- Route all DB-sourced event URLs through the shared `normalizeToUrl()` helper in the BFF mappers: it prepends `https://` to bare domains and drops unsafe/invalid values (`javascript:`/`data:` schemes, malformed strings, the literal `"NULL"`), so the template always binds an absolute, safe URL.
- Covers three mappers:
  - `org-lens-events.service.ts` — Org Lens Events page (the reported surface; Event Name + Action links)
  - `org-people-event-attendees.service.ts` — Org Lens People → Event Attendees
  - `events.service.ts` — individual-user My / Foundation / Visa event mappers

PROD audit (2026-06-10, `ANALYTICS.SILVER_DIM.EVENTS`): 53 scheme-less `event_url` rows, all `source='regfox'`; on `ANALYTICS.PLATINUM_LFX_ONE.ORG_EVENTS` that surfaces as 52 distinct events across ~97.8k org×event rows. Registration URLs are unaffected (0 scheme-less).

- Upstream data fix: event URLs should be stored *with* a scheme in `silver_dim_events` (RegFox ingestion). This PR is the consumer-side mitigation; a dbt follow-up will fix the source.
- Literal `"NULL"`-string rows in the URL columns (rendered gracefully as a dash here, but should be real SQL `NULL` upstream).

## Test plan
- [ ] Org Lens Events: a RegFox event (e.g. "LF Decentralized Trust Privacy Workshop July 2026") — Event Name + Action link open `https://linuxfoundation.regfox.com/...`, not the app origin.
- [ ] Already-absolute event URLs render unchanged.
- [ ] Invalid / `"NULL"` values render as a dash, not a broken link.
- [ ] `yarn lint:check` and `yarn check-types` green (verified locally).

[LFXV2-2185]: https://linuxfoundation.atlassian.net/browse/LFXV2-2185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/services/events.service.ts
+++ b/apps/lfx-one/src/server/services/events.service.ts
@@ -35,7 +35,7 @@ import {
   VisaRequestRow,
   VisaRequestsResponse,
 } from '@lfx-one/shared/interfaces';
-import { formatDateToUTC } from '@lfx-one/shared/utils';
+import { formatDateToUTC, normalizeToUrl } from '@lfx-one/shared/utils';
 import { Request } from 'express';
 
 import { MicroserviceError } from '../errors';
@@ -1000,7 +1000,8 @@ export class EventsService {
     return {
       id: row.EVENT_ID,
       name: row.EVENT_NAME,
-      url: row.EVENT_URL ?? '',
+      // normalizeToUrl prepends https:// to scheme-less DB URLs and drops unsafe/invalid ones; '' keeps the non-null contract.
+      url: normalizeToUrl(row.EVENT_URL ?? '') ?? '',
       location: this.formatLocation(row.EVENT_LOCATION, row.EVENT_CITY, row.EVENT_COUNTRY),
       applicationDate: row.APPLICATION_DATE
         ? new Date(row.APPLICATION_DATE).toLocaleDateString('en-US', { month: 'short', day: 'numeric', year: 'numeric' })
@@ -1015,8 +1016,8 @@ export class EventsService {
       id: row.EVENT_ID,
       name: row.EVENT_NAME,
       category: row.EVENT_CATEGORY,
-      url: row.EVENT_URL ?? '',
-      registrationUrl: row.EVENT_REGISTRATION_URL ?? null,
+      url: normalizeToUrl(row.EVENT_URL ?? '') ?? '',
+      registrationUrl: normalizeToUrl(row.EVENT_REGISTRATION_URL ?? ''),
       date: this.formatDateRange(row.EVENT_START_DATE, row.EVENT_END_DATE),
       location: this.formatLocation(row.EVENT_LOCATION, row.EVENT_CITY, row.EVENT_COUNTRY),
       status: row.EVENT_STATUS ?? null,
@@ -1042,8 +1043,8 @@ export class EventsService {
     return {
       id: row.EVENT_ID,
       name: row.EVENT_NAME,
-      url: row.EVENT_URL ?? '',
-      registrationUrl: row.EVENT_REGISTRATION_URL ?? null,
+      url: normalizeToUrl(row.EVENT_URL ?? '') ?? '',
+      registrationUrl: normalizeToUrl(row.EVENT_REGISTRATION_URL ?? ''),
       foundation: row.PROJECT_NAME,
       startDate: new Date(row.EVENT_START_DATE).toISOString(),
       date: this.formatDateRange(row.EVENT_START_DATE, row.EVENT_END_DATE),
--- a/apps/lfx-one/src/server/services/org-lens-events.service.ts
+++ b/apps/lfx-one/src/server/services/org-lens-events.service.ts
@@ -15,7 +15,7 @@ import type {
   OrgEventsSummary,
   OrgEventsSummaryRow,
 } from '@lfx-one/shared/interfaces';
-import { formatDateToUTC } from '@lfx-one/shared/utils';
+import { formatDateToUTC, normalizeToUrl } from '@lfx-one/shared/utils';
 import type { Request } from 'express';
 
 import { logger } from './logger.service';
@@ -261,8 +261,9 @@ export class OrgLensEventsService {
       eventLocation: row.EVENT_LOCATION ?? null,
       eventCity: row.EVENT_CITY ?? null,
       eventCountry: row.EVENT_COUNTRY ?? null,
-      eventUrl: row.EVENT_URL ?? null,
-      eventRegistrationUrl: row.EVENT_REGISTRATION_URL ?? null,
+      // normalizeToUrl prepends https:// to scheme-less DB URLs and drops unsafe ones so the template href stays absolute.
+      eventUrl: normalizeToUrl(row.EVENT_URL ?? ''),
+      eventRegistrationUrl: normalizeToUrl(row.EVENT_REGISTRATION_URL ?? ''),
       orgAttendeeCount: row.ORG_REGISTRATION_COUNT || 0,
       eventRegistrationsGoal: row.EVENT_REGISTRATIONS_GOAL ?? null,
       orgSpeakerAcceptedCount: row.ORG_SPEAKER_ACCEPTED_COUNT || 0,
--- a/apps/lfx-one/src/server/services/org-people-event-attendees.service.ts
+++ b/apps/lfx-one/src/server/services/org-people-event-attendees.service.ts
@@ -13,6 +13,7 @@ import type {
   OrgPeopleAllEventAttendeeRow,
   OrgPeopleEventRow,
 } from '@lfx-one/shared/interfaces';
+import { normalizeToUrl } from '@lfx-one/shared/utils';
 
 import { toIsoDate } from '../helpers/date-format.helper';
 import { SnowflakeService } from './snowflake.service';
@@ -54,7 +55,8 @@ export class OrgPeopleEventAttendeesService {
       eventLocation: row.EVENT_LOCATION,
       eventCity: row.EVENT_CITY,
       eventCountry: row.EVENT_COUNTRY,
-      eventUrl: row.EVENT_URL,
+      // Normalize scheme-less DB URLs (e.g. "regfox.com/..") so a future clickable event name binds a safe absolute href.
+      eventUrl: normalizeToUrl(row.EVENT_URL ?? ''),
       foundationId: row.FOUNDATION_ID,
       foundationName: row.FOUNDATION_NAME,
       eventStartDate: toIsoDate(row.EVENT_START_DATE),


mrashad-26 (approved):

apps/lfx-one/src/server/services/events.service.ts
```diff
@@ -1000,7 +1000,8 @@ export class EventsService {
     return {
       id: row.EVENT_ID,
       name: row.EVENT_NAME,
-      url: row.EVENT_URL ?? '',
+      // normalizeToUrl prepends https:// to scheme-less DB URLs and drops unsafe/invalid ones; '' keeps the non-null contract.
```
dealako, line 1003: ⚪ **Nit — comment explains WHAT, not WHY**

The clause `normalizeToUrl prepends https:// to scheme-less DB URLs and drops unsafe/invalid ones` describes the function's own mechanics — that belongs in the function's JSDoc (already there in `url.utils.ts:128–152`), not repeated at each call site. Per CLAUDE.md, inline comments should capture non-obvious *reasons*, not mechanism.

The second clause `; '' keeps the non-null contract` **is** a valid WHY (the `url` field is typed `string`, not `string | null`, so the outer `?? ''` is load-bearing). Suggest trimming to just that:
```ts
// '' fallback preserves the non-null url contract required by the interface.
url: normalizeToUrl(row.EVENT_URL ?? '') ?? '',
```
Same applies to the analogous comments in `org-lens-events.service.ts:264` and `org-people-event-attendees.service.ts:58` (though the one in `org-people` comments on a future-use field so the rationale is different — could be dropped entirely).

dealako (approved):
@ahmedomosanya — clean, targeted fix that lands the right abstraction in the right layer. BFF-side normalization means every consumer (org-lens, org-people, individual events, foundation events, visa requests) now gets safe absolute URLs without any template-side workaround, and routing through `normalizeToUrl()` from the shared package keeps the logic DRY.

**Independent analysis performed on:**
- All three changed BFF mapper services
- `packages/shared/src/utils/url.utils.ts` — the actual sanitizer
- Angular templates consuming these fields (`org-events-table`, `events-table`, `event-request-list`)
- Shared interface types confirming type contracts

**Type contract correctness confirmed:** `url: string` fields use a `?? ''` outer fallback; `registrationUrl: string | null` fields don't need one. The asymmetry is intentional and matches the declared interfaces. The `check-types` pass seals it.

**Completeness confirmed:** All five URL mappings across the three services are updated. No raw `EVENT_URL` or `EVENT_REGISTRATION_URL` mapper assignments remain in the services layer.

**Template safety confirmed:** `@if (event.eventUrl)` in `org-events-table.component.html` correctly handles `null`; `@if (!event.isRegistered && event.registrationUrl)` does the same. The foundation-events table binds `[href]="event.url"` directly, but `url` is typed `string` with `?? ''` fallback — same behavior as before this PR.

---

### AI comment reconciliation

| Bot | Verdict |
|---|---|
| **CodeRabbit** | No actionable comments ✓ agree |
| **GitHub Copilot** | No comments ✓ agree |
| **@mrashad-26** | Approved ✓ agree |

---

### Issue count

🔴 **Blocking: 0**
🟡 **Minor: 2** — missing `url.utils.spec.ts`; substring-regex false positives in `isValidUrl` (both pre-existing in `url.utils.ts`, scope expanded by this PR)
⚪ **Nit: 1** — inline comment style (see inline comment)

---

### 🟡 Minor 1 — No unit tests for `normalizeToUrl()`

`packages/shared/src/utils/url.utils.ts` has no spec file. The directory has `committee.utils.spec.ts`, `email.utils.spec.ts`, `html-utils.spec.ts`, `invitation.utils.spec.ts`, `meeting.utils.spec.ts` — but no `url.utils.spec.ts`. With this PR, `normalizeToUrl()` is now load-bearing across five mappers in three services. The security-sensitive paths — scheme-less normalization, dangerous-scheme rejection (`javascript:`, `data:`, `vbscript:`), private IP rejection — are all untested. A future refactor to `isValidDomain` or `DANGEROUS_URL_PATTERNS` won't be caught until a broken URL surfaces in the UI.

Suggest a follow-up: `packages/shared/src/utils/url.utils.spec.ts` covering at minimum: scheme-less domain → `https://`, already-absolute URL → passthrough, `javascript:`/`data:` → `null`, empty/`null` → `null`, private IPs → `null`, literal `"NULL"` → `null`.

---

### 🟡 Minor 2 — Substring regex in `isValidUrl` can reject valid event URLs (pre-existing, blast radius expanded)

```typescript
// packages/shared/src/utils/url.utils.ts:15
const DANGEROUS_URL_PATTERNS = [/javascript:/i, /data:/i, /vbscript:/i, /file:/i, /ftp:/i];
```

These patterns run `.test()` on the **full URL string**, so a valid event URL containing any of these tokens anywhere — e.g., `https://example.com/resources/data:export` or `https://example.com/?format=data:json` — will be silently dropped and replaced with `null`/`''`. For current RegFox URL shapes (`linuxfoundation.regfox.com/<slug>`) this won't trigger; but as `normalizeToUrl()` spreads to more surfaces, the blast radius grows.

The fix is simple and doesn't weaken security: the `!['http:', 'https:'].includes(url.protocol)` check that already runs _after_ URL parsing makes `DANGEROUS_URL_PATTERNS` redundant for all well-formed URLs. Removing the substring check (or replacing it with a scheme-only guard before `new URL()`) eliminates the false positives. Consider a follow-up ticket.

---

The fix is correct, complete, and safe for the reported surface. Both minor findings are improvements to the underlying `url.utils.ts` utility rather than issues introduced by this PR — they can be addressed as follow-up work without blocking merge.

apps/lfx-one/src/server/services/events.service.ts
```diff
@@ -1000,7 +1000,8 @@ export class EventsService {
...
+      url: normalizeToUrl(row.EVENT_URL ?? '') ?? '',
```
MRashad26, line 1004: 🔵 **Info — for the non-null `url: string` fields, an invalid stored value now collapses to `''`; confirm the consuming templates guard empty.**

For My/Foundation/Visa events, `url` is `string` (non-null), so `normalizeToUrl(...) ?? ''` yields `''` when the stored URL is invalid/relative/`"NULL"`. The org-lens surface uses `string | null`, so its template can render a dash on `null` — but here an empty string is not null. If these templates bind `[href]="event.url"` without an `@if (event.url)` guard, `[href]=""` resolves to the **current page URL** — a softer version of the same broken-link class this PR fixes (a self-link rather than the 404 origin-path).

**Suggested check:** verify the My/Foundation/Visa event templates wrap the link in `@if (event.url)` (or bind `[href]="event.url || null"`) so an empty `url` renders as a dash/non-link, consistent with the org-lens behaviour. No change needed if that guard already exists.

MRashad26 (commented):
Automated review + `code-standards-enforcer` audit (report-only). PR body marks this **WIP / Draft**. This is a well-targeted, correct consumer-side mitigation — verified `normalizeToUrl` against the actual cases.

## Overview
Routes DB-sourced event URLs in three BFF mappers through the shared `normalizeToUrl()` so scheme-less values (`linuxfoundation.regfox.com/<slug>`) become absolute `https://` and unsafe/invalid values are dropped, instead of binding as broken relative `[href]`s against the app origin.

## Correctness — verified ✅
I traced `normalizeToUrl` (`packages/shared/src/utils/url.utils.ts`) and `isValidDomain` against each case:
- Bare domain w/ path → `isValidDomain` builds `https://host/path`, `new URL()` parses, hostname has a dot → upgraded to `https://…` ✅ (fixes the reported RegFox surface).
- Already-absolute `http(s)://` → validated and returned unchanged (idempotent) ✅.
- `javascript:` / `data:` / relative path / literal `"NULL"` → no dotted hostname → returns `null` → dropped ✅ (this is a genuine **security improvement**, not just cosmetic).
- **Types align:** `normalizeToUrl(input: string): string | null`; every target field is `string | null` (org-lens `eventUrl`/`eventRegistrationUrl`, org-people `eventUrl`) or `string` where `?? ''` preserves the non-null contract (`events.interface` `url`). The `?? ''` coercion correctly bridges the `string`-only parameter. No type breakage.

Nice attention to the non-null vs nullable contract asymmetry (`url` keeps `?? ''`, `registrationUrl`/`eventUrl` don't) — that's correct, not an oversight.

## Code Standards Audit
- **🔴 Critical:** none
- **🟡 Warning:**
  - Commit `5fb1187` carries a **`Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>`** trailer — violates CLAUDE.md *"No Claude co-author in commits."* (The `ai-assisted` label is the right way to mark this; the commit trailer is not.) Strip it via amend/rebase before merge.
  - PR body has a **`## Test plan`** section — violates *"No test plans in pull requests."*
- **🔵 Info:** non-null `url` empty-string edge case (inline); no automated test asserts the mapper wiring.
- **✅ Correctly followed:** uses the shared `normalizeToUrl` util (no reinvention), no `console.log`, all lines ≤160, signed-off commits, server-layer change kept in the BFF mappers (correct layer for this normalization).

## Test coverage
No tests added. `normalizeToUrl` is presumably unit-tested in shared, but the mapper wiring (the `?? ''` non-null contract, the drop-to-dash behaviour) isn't asserted anywhere. The mappers are private, so a direct test needs a thin seam — low priority given the helper is shared and tested, but a single regression test for "scheme-less RegFox URL → https" would lock the reported bug.

## Verdict: PASS (code is correct) — clear the two 🟡 process items (co-author trailer, test-plan section) and lift WIP before merge.

apps/lfx-one/src/server/services/events.service.ts
ahmedomosanya: Agreed — the mechanism clause duplicates the `url.utils.ts` JSDoc; only the `'' keeps the non-null contract` half is the real WHY here. Non-blocking nit with no behavior change, so folding the comment trim into the same follow-up that picks up the underlying `url.utils.ts` improvements (spec coverage + the substring-pattern refinement) rather than churning the diff now.

apps/lfx-one/src/server/services/events.service.ts
ahmedomosanya: Confirmed the binding state across the consumers: org-lens (`org-events-table.component.html:65`) and my-events registration (`events-table.component.html:136`) already guard with `@if`. The two that bind `[href]` to a non-null `url` without a guard are `foundation-event-dashboard/.../events-table.component.html:58,109` and `my-events-dashboard/.../event-request-list.component.html:76`. That empty-string → self-link case is pre-existing — a null `EVENT_URL` already mapped to `''` before this PR via `?? ''`; this change only shifts the literal-"NULL"/invalid input into the same bucket. Tracking the `@if (event.url)` guard as follow-up work — non-blocking for the reported org-lens surface, which is already guarded.

PR #922, opened 2026-06-11
feat(committee-members): add Org Lens People -> Committee tab (LFXV2-1872)

luismoriguerra:
## Summary
- Adds the **People → Committee** tab: an org-wide, **non-Board** committee-member roster grouped by person, with search + Foundation/Committee filters, sortable columns (Name/Foundations/Committees), and expand-to-seats.
- **Reassign Committee Roles** modal (bulk — re-assign all/some of a person's Membership-Entitlement seats) and **Edit Committee Role** modal (single seat).
- Reuses the spec-026 committee-service read + reassign endpoints (now carrying `project_uid`/`project_slug` — see lfx-v2-committee-service#125). Board seats excluded; stats filter-independent; reassign gated to Membership-Entitlement seats; writer/auditor-gated UI.
- **Auth fix**: `getEffectiveUsername` now falls back to `preferred_username` after `nickname`/`username`, so the LFID-username identity path (#912) resolves under self-hosted OIDC (Authelia) as well as Auth0 — additive; Auth0 precedence and impersonation are unchanged.

## Depends on
- lfx-v2-committee-service#125 (additive `project_uid`/`project_slug` DTO fields). The foundation column/filter/stat populate once that ships; the tab degrades gracefully (slug fallback) otherwise.

## Test plan
- [x] `yarn format:check` / `yarn lint:check` / `yarn build` green
- [x] Playwright e2e: `org-people-committee-tab.spec.ts` (read/empty/error/auditor/expand/search/filter + SC-001 perf + SC-004 stat-count) and `org-people-committee-reassign.spec.ts` (bulk happy/partial/total-fail, single edit happy/5xx, disabled non-entitlement, cancel)
- [x] go-reviewer + ui-reviewer: GREEN (no CRITICAL/HIGH/MEDIUM)
- [x] Verified live in local OrbStack (Authelia) against real committee data: roster renders, Board excluded, foundation populated, stats correct, Reassign/Edit modals open

🤖 Generated with [Cursor](https://cursor.com)

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
@@ -61,15 +61,15 @@
         }
       } @else {
         <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-individuals">
-          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().individualCount }}</div>
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().individualCount | number }}</div>
           <div class="mt-1 text-xs text-gray-500">Individuals</div>
         </div>
         <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-committees">
-          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().committeeCount }}</div>
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().committeeCount | number }}</div>
           <div class="mt-1 text-xs text-gray-500">Committees</div>
         </div>
         <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-foundations">
-          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().foundationsCovered }}</div>
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().foundationsCovered | number }}</div>
           <div class="mt-1 text-xs text-gray-500">Foundations with committee members</div>
         </div>
       }
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
@@ -1,0 +1,272 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<p-toast position="top-right" key="org-people-committee-toast-success" data-testid="org-people-committee-toast-success" />
+
+<div class="flex flex-col gap-6" data-testid="org-people-committee">
+  <!-- Filter bar -->
+  <div class="flex flex-col gap-3 md:flex-row md:items-center md:gap-4" data-testid="org-people-committee-filter-bar">
+    <div class="flex-1">
+      <lfx-input-text
+        [form]="filterForm"
+        control="search"
+        placeholder="Search committee members..."
+        icon="fa-light fa-magnifying-glass"
+        styleClass="w-full"
+        size="small"
+        dataTest="org-people-committee-search-input" />
+    </div>
+    <div class="w-full md:w-56">
+      <lfx-select
+        [form]="filterForm"
+        control="foundation"
+        size="small"
+        [options]="foundationOptions()"
+        styleClass="w-full"
+        ariaLabel="Filter by foundation"
+        dataTest="org-people-committee-foundation-filter" />
+    </div>
+    <div class="w-full md:w-56">
+      <lfx-select
+        [form]="filterForm"
+        control="committee"
+        size="small"
+        [options]="committeeOptions()"
+        styleClass="w-full"
+        ariaLabel="Filter by committee"
+        dataTest="org-people-committee-committee-filter" />
+    </div>
+  </div>
+
+  @if (fetchError()) {
+    <div role="alert" data-testid="org-people-committee-error">
+      <lfx-empty-state
+        [withCard]="false"
+        icon="fa-light fa-triangle-exclamation"
+        title="Failed to load committee members"
+        subtitle="Please try again."
+        ctaLabel="Retry"
+        ctaIcon="fa-light fa-arrow-rotate-left"
+        (ctaClick)="retry()" />
+    </div>
+  } @else {
+    <!-- Stat cards -->
+    <div class="grid grid-cols-1 gap-4 sm:grid-cols-3" data-testid="org-people-committee-stats">
+      @if (isLoading()) {
+        @for (label of statSkeletonLabels; track label) {
+          <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-skeleton">
+            <p-skeleton width="3rem" height="1.5rem" />
+            <div class="mt-2 text-xs text-gray-500">{{ label }}</div>
+          </div>
+        }
+      } @else {
+        <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-individuals">
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().individualCount }}</div>
+          <div class="mt-1 text-xs text-gray-500">Individuals</div>
+        </div>
+        <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-committees">
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().committeeCount }}</div>
+          <div class="mt-1 text-xs text-gray-500">Committees</div>
+        </div>
+        <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-foundations">
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().foundationsCovered }}</div>
+          <div class="mt-1 text-xs text-gray-500">Foundations with committee members</div>
+        </div>
+      }
+    </div>
+
+    <!-- Table -->
+    <div class="overflow-hidden rounded-lg border border-gray-200 bg-white" data-testid="org-people-committee-table">
+      @if (isLoading()) {
+        <div class="divide-y divide-gray-100" data-testid="org-people-committee-table-skeleton">
+          @for (i of tableSkeletonRows; track i) {
+            <div class="flex items-center gap-4 px-4 py-3">
+              <p-skeleton width="1rem" height="1rem" />
+              <div class="flex flex-1 flex-col gap-2">
+                <p-skeleton width="40%" height="0.875rem" />
+                <p-skeleton width="25%" height="0.75rem" />
+              </div>
+              <p-skeleton width="8rem" height="1.25rem" />
+              <p-skeleton width="6rem" height="1.25rem" borderRadius="9999px" />
+            </div>
+          }
+        </div>
+      } @else if (decoratedGroups().length === 0) {
+        <div class="flex flex-col items-center justify-center gap-2 px-6 py-12 text-center" data-testid="org-people-committee-empty">
+          <i class="fa-light fa-users-rectangle text-2xl text-gray-300" aria-hidden="true"></i>
+          @if (isFiltering()) {
+            <p class="text-sm text-gray-700">No committee members match the current filters.</p>
+            <p class="text-xs text-gray-500">Try clearing the search or changing the dropdown selections.</p>
+          } @else {
+            <p class="text-sm text-gray-700">No committee members found.</p>
+            <p class="text-xs text-gray-500">Committee members are sourced from LFX committee records.</p>
+          }
+        </div>
+      } @else {
+        <table class="w-full text-sm" data-testid="org-people-committee-rows">
+          <thead class="bg-gray-50">
+            <tr class="border-b border-gray-200 text-left text-xs uppercase tracking-wide text-gray-500">
+              <th class="w-7 px-2 py-3"></th>
+              @let aria = ariaSortMap();
+              @let icons = sortIconMap();
+              <th scope="col" class="px-4 py-3 font-medium" [attr.aria-sort]="aria.name">
+                <button
+                  type="button"
+                  class="inline-flex w-full items-center gap-1 bg-transparent p-0 text-left text-xs font-medium uppercase tracking-wide text-gray-500 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+                  (click)="onSort('name')">
+                  Name
+                  <i [class]="icons.name" aria-hidden="true"></i>
+                </button>
+              </th>
+              <th scope="col" class="px-4 py-3 font-medium" [attr.aria-sort]="aria.foundations">
+                <button
+                  type="button"
+                  class="inline-flex w-full items-center gap-1 bg-transparent p-0 text-left text-xs font-medium uppercase tracking-wide text-gray-500 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+                  (click)="onSort('foundations')">
+                  Foundations
+                  <i [class]="icons.foundations" aria-hidden="true"></i>
+                </button>
+              </th>
+              <th scope="col" class="px-4 py-3 font-medium" [attr.aria-sort]="aria.committees">
+                <button
+                  type="button"
+                  class="inline-flex w-full items-center gap-1 bg-transparent p-0 text-left text-xs font-medium uppercase tracking-wide text-gray-500 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+                  (click)="onSort('committees')">
+                  Committees
+                  <i [class]="icons.committees" aria-hidden="true"></i>
+                </button>
+              </th>
+              <th class="w-12 px-4 py-3 text-right">
+                <span class="sr-only">Actions</span>
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            @for (group of decoratedGroups(); track group.email) {
+              @let expanded = !!expansion()[group.email];
+              <tr
+                class="cursor-pointer border-b border-gray-100 hover:bg-gray-50 focus:bg-gray-50 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500"
+                role="button"
+                tabindex="0"
+                [attr.aria-expanded]="expanded"
+                [attr.aria-label]="'Toggle committee details for ' + group.displayName"
+                [attr.data-testid]="'org-people-committee-row-' + group.email"
+                (click)="toggleExpansion(group.email)"
+                (keydown)="onRowKeydown($event, group.email)">
+                <td class="px-2 py-3 align-middle">
+                  <i
+                    class="fa-light fa-chevron-right text-gray-400 transition-transform"
+                    [class.rotate-90]="expanded"
+                    [attr.data-testid]="'org-people-committee-expand-' + group.email"
+                    aria-hidden="true"></i>
+                </td>
+                <td class="px-4 py-3">
+                  <div class="flex items-center gap-3">
+                    <div class="flex h-8 w-8 shrink-0 items-center justify-center rounded-full bg-purple-500 text-xs font-semibold text-white">
+                      {{ group.initials }}
+                    </div>
+                    <div>
+                      <div class="font-medium text-gray-900">{{ group.displayName }}</div>
+                      @if (group.jobTitle) {
+                        <div class="text-xs text-gray-500">{{ group.jobTitle }}</div>
+                      }
+                    </div>
+                  </div>
+                </td>
+                <td class="px-4 py-3">
+                  <div class="flex flex-col gap-0.5 text-sm text-gray-700">
+                    @for (label of group.foundationLabels; track label) {
+                      <span [attr.data-testid]="'org-people-committee-foundation-pill-' + label">{{ label }}</span>
+                    }
+                  </div>
+                </td>
+                <td class="px-4 py-3">
+                  <span class="inline-block rounded-full border border-blue-200 bg-blue-50 px-2.5 py-0.5 text-xs font-semibold text-blue-700">
+                    {{ group.committeeCount }} {{ group.committeeCount === 1 ? 'Committee' : 'Committees' }}
+                  </span>
+                </td>
+                <td class="px-4 py-3 text-right">
+                  <span
+                    class="inline-flex"
+                    [pTooltip]="
+                      canEdit() ? (group.editableCount === 0 ? 'No org-reassignable seats for this person' : 'Reassign committee roles') : editDisabledTooltip
+                    "
+                    tooltipPosition="left">
+                    <button
+                      type="button"
+                      class="inline-flex h-8 w-8 items-center justify-center rounded-md text-gray-500 hover:bg-gray-100 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500 disabled:cursor-not-allowed disabled:opacity-40 disabled:hover:bg-transparent"
+                      [disabled]="!canEdit() || group.editableCount === 0"
+                      [attr.aria-label]="'Reassign committee roles for ' + group.displayName"
+                      [attr.data-testid]="'org-people-committee-reassign-' + group.email"
+                      (click)="onMainPencilClick(group, $event)">
+                      <i class="fa-light fa-pencil" aria-hidden="true"></i>
+                    </button>
+                  </span>
+                </td>
+              </tr>
+              @if (expanded) {
+                <tr class="bg-slate-50" [attr.data-testid]="'org-people-committee-expanded-' + group.email">
+                  <td colspan="5" class="p-0 pl-11">
+                    <table class="w-full border-collapse">
+                      <thead>
+                        <tr class="border-b border-gray-200 bg-slate-100 text-left text-xs font-semibold uppercase text-gray-400">
+                          <th class="px-4 py-2.5">Foundation</th>
+                          <th class="px-4 py-2.5">Committee</th>
+                          <th class="px-4 py-2.5">Voting Status</th>
+                          <th class="w-12 px-4 py-2.5 text-right">
+                            <span class="sr-only">Edit</span>
+                          </th>
+                        </tr>
+                      </thead>
+                      <tbody>
+                        @for (assignment of group.sortedAssignments; track assignment.seatId) {
+                          <tr class="border-b border-gray-200" [attr.data-testid]="'org-people-committee-subrow-' + assignment.seatId">
+                            <td class="px-4 py-2.5 text-sm font-medium text-gray-900">
+                              @if (assignment.showFoundationLabel) {
+                                {{ assignment.foundationName }}
+                              }
+                            </td>
+                            <td class="px-4 py-2.5 text-sm text-gray-700">{{ assignment.committeeName }}</td>
+                            <td class="px-4 py-2.5">
+                              <span class="inline-block rounded-full border px-2 py-0.5 text-xs font-medium" [class]="assignment.votingStatusPillClass">
+                                {{ assignment.votingStatus || 'Non-voting' }}
+                              </span>
+                            </td>
+                            <td class="px-4 py-2.5 text-right">
+                              <span
+                                class="inline-flex"
+                                [pTooltip]="
+                                  canEdit()
+                                    ? assignment.isOrgEditable
+                                      ? 'Edit committee role'
+                                      : (assignment.reason ?? 'This seat is foundation-controlled and not editable here.')
+                                    : editDisabledTooltip
+                                "
+                                tooltipPosition="top">
+                                <button
+                                  type="button"
+                                  class="inline-flex h-8 w-8 items-center justify-center rounded-md text-gray-500 hover:bg-gray-100 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500 disabled:cursor-not-allowed disabled:opacity-40 disabled:hover:bg-transparent"
+                                  [disabled]="!canEdit() || !assignment.isOrgEditable"
+                                  [attr.aria-label]="'Edit ' + assignment.committeeName"
+                                  [attr.data-testid]="'org-people-committee-edit-' + assignment.seatId"
+                                  (click)="onSubRowPencilClick(assignment, $event)">
+                                  <i class="fa-light fa-pencil" aria-hidden="true"></i>
+                                </button>
+                              </span>
+                            </td>
+                          </tr>
+                        }
+                      </tbody>
+                    </table>
+                  </td>
+                </tr>
+              }
+            }
+          </tbody>
+        </table>
+      }
+    </div>
+
+    <p class="text-xs text-gray-400" data-testid="org-people-committee-sources">Source: LFX Committee Records</p>
+  }
+</div>
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
@@ -186,12 +186,7 @@
                   </span>
                 </td>
                 <td class="px-4 py-3 text-right">
-                  <span
-                    class="inline-flex"
-                    [pTooltip]="
-                      canEdit() ? (group.editableCount === 0 ? 'No org-reassignable seats for this person' : 'Reassign committee roles') : editDisabledTooltip
-                    "
-                    tooltipPosition="left">
+                  <span class="inline-flex" [pTooltip]="group.reassignTooltip" tooltipPosition="left">
                     <button
                       type="button"
                       class="inline-flex h-8 w-8 items-center justify-center rounded-md text-gray-500 hover:bg-gray-100 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500 disabled:cursor-not-allowed disabled:opacity-40 disabled:hover:bg-transparent"
@@ -233,16 +228,7 @@
                               </span>
                             </td>
                             <td class="px-4 py-2.5 text-right">
-                              <span
-                                class="inline-flex"
-                                [pTooltip]="
-                                  canEdit()
-                                    ? assignment.isOrgEditable
-                                      ? 'Edit committee role'
-                                      : (assignment.reason ?? 'This seat is foundation-controlled and not editable here.')
-                                    : editDisabledTooltip
-                                "
-                                tooltipPosition="top">
+                              <span class="inline-flex" [pTooltip]="assignment.editTooltip" tooltipPosition="top">
                                 <button
                                   type="button"
                                   class="inline-flex h-8 w-8 items-center justify-center rounded-md text-gray-500 hover:bg-gray-100 hover:text-gray-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500 disabled:cursor-not-allowed disabled:opacity-40 disabled:hover:bg-transparent"
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
@@ -1,6 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
+import { DecimalPipe } from '@angular/common';
 import { Component, computed, DestroyRef, inject, signal, type Signal } from '@angular/core';
 import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
@@ -41,7 +42,7 @@ import { buildPersonGroups, decorateAssignments } from './helpers/committee-memb
 @Component({
   selector: 'lfx-org-people-committee-members',
   standalone: true,
-  imports: [ReactiveFormsModule, InputTextComponent, SelectComponent, SkeletonModule, EmptyStateComponent, ToastModule, TooltipModule],
+  imports: [DecimalPipe, ReactiveFormsModule, InputTextComponent, SelectComponent, SkeletonModule, EmptyStateComponent, ToastModule, TooltipModule],
   providers: [MessageService, DialogService],
   templateUrl: './committee-members.component.html',
 })
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
@@ -173,6 +173,11 @@ export class CommitteeMembersComponent {
       }));
     if (roles.length === 0) return;
 
+    // `group.email` is the grouping key and falls back to a seat `memberUid` when the upstream email is
+    // blank — never pass it as the contact email. Source the person's real email from the assignments so
+    // the modal's duplicate/self-selection guard + typeahead exclusion compare against an actual address.
+    const currentEmail = group.assignments.find((a) => a.person.email)?.person.email ?? '';
+
     const ref = this.dialogService.open(ReassignCommitteeRolesModalComponent, {
       width: '600px',
       modal: true,
@@ -180,7 +185,7 @@ export class CommitteeMembersComponent {
       dismissableMask: true,
       showHeader: false,
       data: {
-        person: { fullName: group.displayName, email: group.email, initials: group.initials },
+        person: { fullName: group.displayName, email: currentEmail, initials: group.initials },
         roles,
         orgUid,
         submit: (intent) => this.performBulkReassign(intent, orgUid),
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
@@ -36,7 +36,7 @@ import { TooltipModule } from 'primeng/tooltip';
 import { CommitteeMembersService } from '../../services/committee-members.service';
 import { EditCommitteeRoleModalComponent } from './components/edit-committee-role-modal.component';
 import { ReassignCommitteeRolesModalComponent } from './components/reassign-committee-roles-modal.component';
-import { buildPersonGroups, decorateAssignments } from './helpers/committee-members.helper';
+import { buildPersonGroups, decoratePersonGroup } from './helpers/committee-members.helper';
 
 /** Org Lens — People → Committee tab (spec 027). Org-wide, non-Board committee-member roster grouped by person, with filter/sort/expand (US1+US2). Reassign/Edit modals wired in US3/US4. */
 @Component({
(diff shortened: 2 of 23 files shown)


apps/lfx-one/src/server/services/org-people-committee-members.service.ts
```diff
@@ -0,0 +1,177 @@
...
+    const emails = new Set<string>();
+    const committees = new Set<string>();
+    const foundations = new Set<string>();
+    for (const a of assignments) {
+      if (a.person.email) emails.add(a.person.email);
+      if (a.committeeUid) committees.add(a.committeeUid);
+      if (a.projectUid) foundations.add(a.projectUid);
+    }
+    return { individualCount: emails.size, committeeCount: committees.size, foundationsCovered: foundations.size };
+  }
```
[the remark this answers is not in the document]
luismoriguerra: Good catch — fixed in 704755b6. `computeStats` now counts foundations by `projectUid || foundationSlug`, so the tile no longer shows 0 foundations when UID enrichment is missing while the table still renders foundation values.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  // Closes the dialog if the user switches account before confirming — prevents cross-org writes.
+  private wireDialogToAccountChange(ref: DynamicDialogRef): void {
+    this.accountCancel$.pipe(takeUntil(ref.onClose), takeUntilDestroyed(this.destroyRef)).subscribe(() => ref.close(null));
+  }
+
+  private cleanErrorMessage(err: unknown): string {
+    return (err as { error?: { error?: { message?: string } } })?.error?.error?.message ?? 'Could not save changes. Please try again.';
+  }
```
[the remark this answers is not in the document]
luismoriguerra: Good catch — fixed in 704755b6. `cleanErrorMessage` now tolerates both envelope shapes: it returns `err.error.error` when it's a non-empty string (MicroserviceError/BaseApiError `{ error: string, code }`) and falls back to `err.error.error.message` when it's an object, before the generic message. The server-provided detail now survives to the toast.

luismoriguerra:
### Review round recap (commit `704755b6`)

Addressed the open review threads on this PR (LFXV2-1872 — Org Lens People → Committee tab):

**Copilot — resolved (2):**
- ✅ `org-people-committee-members.service.ts` — `computeStats` now counts foundations by `projectUid || foundationSlug`, so the stat tile degrades gracefully (no longer 0) when UID enrichment is missing but the table still shows foundation values.
- ✅ `committee-members.component.ts` — `cleanErrorMessage` now tolerates both `{ error: { message } }` and the `{ error: string }` MicroserviceError/BaseApiError envelope, so server-provided error detail survives to the toast.

**CI / bot threads — already handled:**
- ✅ **Code Quality / format:check** — fixed earlier (`daf1e74f`, prettier on `auth-helper.ts`).
- ✅ **pr-title-lint** — title retitled to ASCII (`->` instead of `→`); the conventional-commits check now passes (“Conventional Commits FTW!”).

**Constitution note:** No unit tests were added for `apps/lfx-one` per the project carve-out (Playwright e2e only). Behavior of both fixes is covered by the existing e2e specs.

**Status:** `yarn format:check`, `yarn lint:check`, and `yarn build` all green. CodeRabbit was mid-review on the prior commit and will re-run against this push — no actionable CodeRabbit comments yet at the time of this recap.

packages/shared/src/interfaces/org-people-committee-members.internal.interface.ts
```diff
@@ -0,0 +1,42 @@
...
+/** Sortable columns on the main table (FR-021). */
+export type CommitteeMembersSortColumn = 'name' | 'foundations' | 'committees';
+
+/** Sort direction — 1 ascending, -1 descending. */
+export type CommitteeMembersSortDirection = 1 | -1;
+
+/** One person row — derived by grouping assignments by lowercased email (first-wins for display fields). */
+export interface CommitteeMemberPersonGroup {
+  /** Lowercased email — the group identity key. */
+  email: string;
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in f3c1bbe4 — corrected the doc comment. `email` is now documented as the group identity key (the person's lowercased email, or the seat `memberUid` fallback when email is blank), matching `buildPersonGroups`. It's no longer described as always being an email.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
```diff
@@ -0,0 +1,425 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { Component, computed, DestroyRef, inject, signal, type Signal } from '@angular/core';
+import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
+import { catchError, combineLatest, debounceTime, distinctUntilChanged, firstValueFrom, map, of, skip, Subject, switchMap, takeUntil, tap } from 'rxjs';
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in f3c1bbe4 — imported `DecimalPipe` from `@angular/common` so the template can use `| number` (mirrors the Key Contacts tab).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
```diff
@@ -0,0 +1,425 @@
...
+import { EditCommitteeRoleModalComponent } from './components/edit-committee-role-modal.component';
+import { ReassignCommitteeRolesModalComponent } from './components/reassign-committee-roles-modal.component';
+import { buildPersonGroups, decorateAssignments } from './helpers/committee-members.helper';
+
+/** Org Lens — People → Committee tab (spec 027). Org-wide, non-Board committee-member roster grouped by person, with filter/sort/expand (US1+US2). Reassign/Edit modals wired in US3/US4. */
+@Component({
+  selector: 'lfx-org-people-committee-members',
+  standalone: true,
+  imports: [ReactiveFormsModule, InputTextComponent, SelectComponent, SkeletonModule, EmptyStateComponent, ToastModule, TooltipModule],
+  providers: [MessageService, DialogService],
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in f3c1bbe4 — added `DecimalPipe` to the standalone component `imports` array, matching the Key Contacts convention.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
```diff
@@ -0,0 +1,272 @@
...
+          <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-skeleton">
+            <p-skeleton width="3rem" height="1.5rem" />
+            <div class="mt-2 text-xs text-gray-500">{{ label }}</div>
+          </div>
+        }
+      } @else {
+        <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-individuals">
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().individualCount }}</div>
+          <div class="mt-1 text-xs text-gray-500">Individuals</div>
+        </div>
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in f3c1bbe4 — applied `| number` to the Individuals stat count (consistent with Key Contacts/Trainees).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
```diff
@@ -0,0 +1,272 @@
...
+        <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-committees">
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().committeeCount }}</div>
+          <div class="mt-1 text-xs text-gray-500">Committees</div>
+        </div>
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in f3c1bbe4 — applied `| number` to the Committees stat count (consistent with Key Contacts/Trainees).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
```diff
@@ -0,0 +1,272 @@
...
+        <div class="rounded-lg border border-gray-200 bg-white p-4" data-testid="org-people-committee-stat-foundations">
+          <div class="text-2xl font-semibold leading-none text-gray-900">{{ stats().foundationsCovered }}</div>
+          <div class="mt-1 text-xs text-gray-500">Foundations with committee members</div>
+        </div>
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in f3c1bbe4 — applied `| number` to the Foundations stat count (consistent with Key Contacts/Trainees).

luismoriguerra:
### Copilot follow-up review round — resolved

All open Copilot threads from the post-`704755b6` review have been addressed in **f3c1bbe4** (`--signoff`, GPG-signed).

**Fixed (6 threads):**
- **`| number` / DecimalPipe (5 threads)** — imported `DecimalPipe` from `@angular/common` into the `CommitteeMembersComponent` standalone `imports` array and applied `| number` to all three stat tiles (Individuals, Committees, Foundations), mirroring the Key Contacts/Trainees convention. There are exactly 3 stat tiles.
- **`internal.interface.ts` group-key doc** — corrected the `CommitteeMemberPersonGroup.email` doc comment to reflect the real `buildPersonGroups` semantics: the key is the lowercased email, or the seat `memberUid` fallback when email is blank (not guaranteed to be an email).

**Already addressed in 704755b6 (resolved):**
- `org-people-committee-members.service.ts:179` — `foundationsCovered` fallback key.
- `committee-members.component.ts:292` — `cleanErrorMessage` tolerant of both error-envelope shapes.

**e2e:** No change needed. The SC-004 stat-count mocks are small (`individualCount: 2`, `committeeCount: 4`, `foundationsCovered: 2`), so `| number` adds no separators and the existing assertions (`parseInt`, `toContain`) still hold.

**Quality gate:** `yarn format:check` ✅ · `yarn lint:check` ✅ · `yarn build` ✅

Open threads: **0**.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
```diff
@@ -0,0 +1,272 @@
...
+                <td class="px-4 py-3">
+                  <span class="inline-block rounded-full border border-blue-200 bg-blue-50 px-2.5 py-0.5 text-xs font-semibold text-blue-700">
+                    {{ group.committeeCount }} {{ group.committeeCount === 1 ? 'Committee' : 'Committees' }}
+                  </span>
+                </td>
+                <td class="px-4 py-3 text-right">
+                  <span
+                    class="inline-flex"
+                    [pTooltip]="
+                      canEdit() ? (group.editableCount === 0 ? 'No org-reassignable seats for this person' : 'Reassign committee roles') : editDisabledTooltip
```
audigregorie: **Major — nested ternary in template binding**

This `pTooltip` expression nests a ternary inside a ternary, which CLAUDE.md prohibits ("Never nest ternary expressions") and check #25 flags (complex logic belongs in a `computed()`/VM, not inline in the template).

You already decorate view-models in `committee-members.helper.ts` — precompute a `reassignTooltip` string field on `CommitteeMemberPersonGroupVm` (it depends only on `canEdit()` and `editableCount`) and bind `[pTooltip]="group.reassignTooltip"`.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
```diff
@@ -0,0 +1,272 @@
...
+                            </td>
+                            <td class="px-4 py-2.5 text-right">
+                              <span
+                                class="inline-flex"
+                                [pTooltip]="
+                                  canEdit()
+                                    ? assignment.isOrgEditable
+                                      ? 'Edit committee role'
+                                      : (assignment.reason ?? 'This seat is foundation-controlled and not editable here.')
+                                    : editDisabledTooltip
```
audigregorie: **Major — nested ternary in template binding (3 levels)**

Same rule as the main-row tooltip, deeper here. Move this into `decorateAssignments` as an `editTooltip` field on `CommitteeMemberAssignmentVm` (derived from `isOrgEditable` + `reason`), and let the `canEdit()` branch select `editTooltip` vs `editDisabledTooltip` in a single, flat expression.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
```diff
@@ -0,0 +1,426 @@
...
+      });
+      return;
+    }
+
+    const total = intent.selected.length;
+    const succeeded = total - failures.length;
+    if (succeeded === 0) {
+      throw new Error(this.cleanErrorMessage(failures[0].reason));
+    }
+    throw new Error(`${succeeded} of ${total} reassignments succeeded. Reopen this dialog to retry the remaining roles with fresh state.`);
```
audigregorie: **Major — bulk partial-failure leaves the modal in an untrustworthy state**

On partial success you call `this.retry()` (the roster refreshes in the background) and then `throw`, which leaves the modal open with its **original** `selectedKeys` intact and Save still enabled. The inline message says "Reopen this dialog to retry," but nothing enforces that.

If the user instead just clicks Save again, the fan-out re-issues a PATCH for **every** selected seat — including the ones that already succeeded. After a reassign the original `memberUid` no longer holds the seat, so those re-submissions fail on the second pass and the "X of N succeeded" messaging becomes misleading.

Since the roster has already refreshed and the modal's selection is now stale, prefer resolving on partial success (close the dialog + show a warning toast summarizing the partial result) rather than throwing back into a modal whose state can no longer be trusted. A full fix that supports in-place retry would require the `submit` contract to return per-role results so the succeeded rows can be pruned from the selection.

packages/shared/src/constants/org-people-committee-members.constants.ts
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: MIT
+
+import type { CommitteeMembersSortColumn, OrgPeopleCommitteeMembersResponse } from '../interfaces';
+
+/** committee-service `committee_category` value that the People → Committee tab EXCLUDES (FR-003). Compared case-insensitively. */
+export const COMMITTEE_CATEGORY_BOARD = 'Board';
+
+/** True when a seat's committee_category is the Board category (case-insensitive, whitespace-trimmed). */
+export function isBoardCategory(category: string | null | undefined): boolean {
+  return (category ?? '').trim().toLowerCase() === COMMITTEE_CATEGORY_BOARD.toLowerCase();
```
audigregorie, line 11: **Minor — Board-exclusion semantics diverge from the Board tab**

This shared `isBoardCategory()` compares case-insensitively and trimmed, but `org-lens-board-committee.service.ts:57` (also touched by this PR) still partitions Board vs committee seats with a module-local case-sensitive `s.committee_category === 'Board'`.

Two user-facing People surfaces now classify "Board" by different rules: a non-canonical upstream value (`'board'`, `' Board '`) would be excluded from this Committee tab but shown as a committee seat on the Board tab. Since this PR introduces the canonical helper, consider unifying the board service on `isBoardCategory()` (and dropping its local `COMMITTEE_CATEGORY_BOARD`) so the two tabs can't drift.

audigregorie (commented):
## Code Review Summary

Polished, well-documented full-stack PR — `switchMap` latest-wins reads, fail-closed pagination reuse, FGA-gated UI with BFF/Heimdall re-enforcement, immutable signal updates, `takeUntilDestroyed` throughout, account-switch dialog cancellation, and thorough e2e coverage. No Critical findings. The notes below are 3 Major (two are the same CLAUDE.md nested-ternary rule) and 2 Minor.

**What's done well**
- Read pipeline holds the skeleton until `orgUid` resolves and `switchMap` cancels stale reads on account switch; `accountCancel$` also closes any open write dialog mid-flight.
- `fetchAllOrgSeats` reuses the existing fail-closed drain instead of re-implementing pagination.
- Path params validated (`assertOrgUid` / `isFilterSafeIdentifier`) before interpolation; email regex-checked + lowercased server-side; foundation-name enrichment fails soft to the slug.

Rating: 4/5.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/components/edit-committee-role-modal.component.html
```diff
@@ -0,0 +1,192 @@
...
+        <label for="edit-committee-email" class="text-xs font-medium text-gray-700">Email <span class="text-red-500">*</span></label>
+        <span class="relative">
+          <i class="fa-light fa-magnifying-glass pointer-events-none absolute left-3 top-1/2 -translate-y-1/2 text-xs text-gray-400" aria-hidden="true"></i>
+          <input
+            id="edit-committee-email"
+            pInputText
+            type="email"
+            placeholder="Enter the employee's email address..."
+            class="w-full pl-8"
+            role="combobox"
```
audigregorie, line 69: **Minor — combobox is mouse-only (a11y)**

The email field declares `role="combobox"` with `aria-controls`/`aria-expanded`, but the suggestion `<li>` options can't be reached by keyboard — there's no arrow-key handling and no `aria-activedescendant`. Keyboard-only users can still type the full email manually (so this degrades rather than blocks), but the combobox ARIA contract implies keyboard operability the widget doesn't provide. The reassign modal's email field has the same shape. Either wire up arrow-key navigation + `aria-activedescendant`, or drop the combobox role and present it as a plain text input with a supplementary suggestion list.

audigregorie (commented):
## Code Review Summary

Careful, well-structured full-stack PR. Auth is user-token based (Heimdall re-gates upstream), path params + reassign body are validated (`isFilterSafeIdentifier` + `EMAIL_REGEX` + `encodeURIComponent`), errors handled end-to-end, foundation enrichment fails soft, and the write path is correctly pessimistic. **No Critical/Major issues** — no correctness, security, or SSR problems found. All findings are Minor convention deviations against `.claude/rules/`.

### Low-priority note — outside the diff

- **PUT vs PATCH terminology in frontend comments.** The FE comments say the parent "fans out the PUTs" / "Single PUT" (`committee-members.component.ts:217,254`, `reassign-committee-roles-modal.component.ts:23`), but the FE actually issues a **PATCH** to the BFF (`committee-members.service.ts` → `http.patch`). The PUT is the BFF→committee-service hop, which the backend service documents well. Harmless, but the FE-side "PUT" wording is slightly off from what that code does.

### What's done well

- Route ordering guard (`/committee-members` before `/:personKey/detail`) with rationale, matching the existing trainees/event-attendees pattern.
- PUT-vs-PATCH verb mapping documented in the BFF service so it won't read as a bug.
- Filter-independent stats; foundation enrichment fails soft; `project_uid`/`project_slug` optional so the tab degrades gracefully pre-committee-service#125.
- Strong deterministic E2E coverage across read/empty/error/auditor/expand/search/filter + bulk and single-edit write paths.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/helpers/committee-members.helper.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { votingStatusPillClass } from '@lfx-one/shared/constants';
+import type { CommitteeMemberAssignment, CommitteeMemberAssignmentVm, CommitteeMemberPersonGroup } from '@lfx-one/shared/interfaces';
+
+// Re-exported so the component template + modals import the pill helper from one place (spec 027).
+export { votingStatusPillClass };
```
audigregorie: **Minor — dead re-export + inaccurate comment**

This `export { votingStatusPillClass }` is unused. Every consumer — `committee-members.component.ts:15`, `edit-committee-role-modal.component.ts:7`, and this helper itself (line 56) — imports `votingStatusPillClass` directly from `@lfx-one/shared/constants`. Nothing imports it from the helper, so the comment ("the component template + modals import the pill helper from one place") describes behavior that doesn't happen. Drop both lines.

```suggestion
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
```diff
@@ -0,0 +1,426 @@
...
+
+  // WritableSignals first (grouped per component-organization convention).
+  protected readonly sortColumn = signal<CommitteeMembersSortColumn>('name');
+  protected readonly sortDirection = signal<CommitteeMembersSortDirection>(1);
+  protected readonly expansion = signal<Record<string, boolean>>({});
+  protected readonly retryTrigger = signal<number>(0);
+  private readonly loadingState = signal<boolean>(true);
+  private readonly fetchErrorState = signal<boolean>(false);
+
+  protected readonly isLoading = this.loadingState.asReadonly();
```
audigregorie, line 75: **Minor — boolean signal naming** (`.claude/rules/angular-standards.md` §Signals)

Boolean signals use bare nouns, not `is*` — `isLoading` → `loading`, and `isFiltering` (line 100) → `filtering`. The rule reserves `is`/`has` for disambiguating from a same-named data property, which isn't the case here. The rename also touches the template (`@if (isLoading())`, etc.), so no one-click fix — flagging for consistency with the convention applied elsewhere.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/components/reassign-committee-roles-modal.component.ts
```diff
@@ -0,0 +1,237 @@
...
+  // Default: all selected so the common workflow (full reassign) is one click.
+  protected readonly selectedKeys = signal<Set<ReassignCommitteeRolesRoleKey>>(new Set(this.roles.map((r) => r.key)));
+
+  protected readonly emailField = signal('');
+  protected readonly firstNameField = signal('');
+  protected readonly lastNameField = signal('');
+  protected readonly emailFormatError = signal<string | null>(null);
+  protected readonly duplicateError = signal<string | null>(null);
+
+  protected readonly isSaving = signal(false);
```
audigregorie, line 50: **Minor — boolean signal naming** (`.claude/rules/angular-standards.md` §Signals)

`isSaving` → `saving` (bare-noun convention). Same in `edit-committee-role-modal.component.ts:39`. Both are referenced in their templates, so the rename spans `.ts` + `.html`.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/components/reassign-committee-roles-modal.component.html
```diff
@@ -0,0 +1,229 @@
...
+        </div>
+      </div>
+    </div>
+
+    <!-- Select roles to reassign -->
+    <div class="flex flex-col gap-3">
+      <div class="flex items-center justify-between gap-3">
+        <h4 class="text-sm font-semibold text-gray-900">Select roles to reassign</h4>
+        <label class="flex cursor-pointer items-center gap-2 text-xs text-blue-600">
+          <p-checkbox
```
audigregorie, line 44: **Minor — raw PrimeNG in a feature template** (`.claude/rules/styling.md`)

The rule says reach for the LFX wrapper when one exists. `lfx-checkbox` exists (`shared/components/checkbox`), and `lfx-input-text` exists — used by the parent (`committee-members.component.html`). This `<p-checkbox>` and the plain `pInputText` first/last-name fields (lines 170, 183 here; 133, 146 in the edit modal) could use the wrappers.

The email **combobox** raw input is a justified exception — it needs `role="combobox"` + focus/blur the wrapper doesn't expose — but the checkbox and name inputs don't have that constraint.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
```diff
@@ -0,0 +1,426 @@
...
+
+  public constructor() {
+    this.orgUid$.pipe(skip(1), takeUntilDestroyed()).subscribe(() => {
+      this.accountCancel$.next();
+      this.resetAllState();
+    });
+
+    // Cascade reset (FR-006): when the foundation filter changes, clear any committee selection so a
+    // stale committee value (no longer in the narrowed options) can't render a blank <p-select> + a
+    // misleading "no matches" empty state.
```
audigregorie: **Minor — comment length** (`.claude/rules/global-rules.md`: "one or two lines maximum. No exceptions")

Representative case — same applies to the service class JSDoc (`org-people-committee-members.service.ts:21-28`, 8 lines), `getEffectiveUsername` (`auth-helper.ts:78-82`, 5 lines), and the `helper.ts` JSDoc blocks (4 lines each). Condensing the in-diff one:

```suggestion
    // Cascade reset (FR-006): clear the committee selection when the foundation filter changes so a
    // stale value can't render a blank <p-select> + a misleading "no matches" state.
    this.filterForm.controls.foundation.valueChanges.pipe(distinctUntilChanged(), takeUntilDestroyed()).subscribe(() => {
```

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
luismoriguerra: Fixed in 34e0e85. Moved the tooltip composition out of the template into `decoratePersonGroup(group, { canEdit, editDisabledTooltip })` in `committee-members.helper.ts`, which precomputes `reassignTooltip` on `CommitteeMemberPersonGroupVm`. The template binding is now flat: `[pTooltip]="group.reassignTooltip"`. No more nested ternaries in the HTML.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.html
luismoriguerra: Fixed in 34e0e85. Added `editTooltip` to `CommitteeMemberAssignmentVm` and populated it in `decorateAssignments` (via the new `buildEditTooltip` helper) so it encodes the full (canEdit × isOrgEditable × `reason` fallback) composition. Template now binds flat: `[pTooltip]="assignment.editTooltip"`.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
luismoriguerra: Fixed in 34e0e85 — implemented the minimum fix you outlined. On partial success `performBulkReassign` now **resolves** the submit promise (so the modal closes via the existing `.then(() => dialogRef.close(null))`) and emits a `severity: 'warn'` toast summarizing `"$succeeded of $total succeeded. Reopen this dialog to retry the remaining $N."`. Total failure still throws so the modal stays open with the cleaned inline error.

This avoids the re-Save→404 fan-out you flagged: the already-succeeded `memberUid`s have moved upstream, so leaving them in `selectedKeys` and re-clicking Save would PATCH stale ids.

The fuller fix you suggested (per-role results in the `submit` contract so the modal can prune succeeded rows in-place) is intentionally out of scope for this round — it touches the `ReassignCommitteeRolesSubmitEvent` shape and the modal state machine, and is better as a follow-up. Filed mentally; happy to track it as a separate ticket if you want.

E2e: `'a 409 on one PATCH ...'` flipped to assert (a) modal closes and (b) the partial-success warning toast appears.

packages/shared/src/constants/org-people-committee-members.constants.ts
luismoriguerra: Fixed in 34e0e85. `org-lens-board-committee.service.ts` now imports and uses `isBoardCategory(...)` from `@lfx-one/shared/constants` everywhere it splits Board vs Committee (the seat filter, the anti-filter, and the reassign re-classification on line 148). Dropped the dead local `COMMITTEE_CATEGORY_BOARD` constant. Both tabs now canonicalize via the same case-insensitive + trimmed predicate — a non-canonical `'board'` value classifies as Board on both surfaces (no divergence).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/components/edit-committee-role-modal.component.html
luismoriguerra: Fixed in 34e0e85 — applied the active-descendant pattern (kept `role="combobox"` rather than dropping it) to BOTH modals (edit + reassign):

- **Keyboard**: `onEmailKeydown` handles `ArrowDown` / `ArrowUp` (wrap-navigating the listbox without moving focus from the input), `Enter` (selects the highlighted option, otherwise falls through to the form-level Save handler), `Escape` (closes the panel with `stopPropagation` so the parent PrimeNG dialog doesn't also close on the same keypress).
- **ARIA**: the input now exposes `aria-activedescendant` pointing at the highlighted option; each `<button role="option">` gets a stable `id` + `aria-selected`.
- **Visual**: highlighted option gets a `bg-gray-100` class so the keyboard user can see what Enter would select.
- **State reset**: `activeOptionIndex` resets to -1 on every keystroke (so a stale highlight from the previous filter can't be Enter-selected), on `(blur)`, on `Escape`, and on `onSelectEmployee`.

E2e: added `'ArrowDown + Enter ... selects the highlighted option (keyboard a11y)'` to `org-people-committee-reassign.spec.ts` covering the edit modal — verifies `aria-activedescendant` updates and Enter populates email + name fields. (The reassign modal mirrors the edit modal exactly; same handler, same template shape.)

luismoriguerra:
## Round-2 review-thread sweep — commit `34e0e85`

Addressed the 5 threads posted by @audigregorie at 14:26Z. All resolved.

### Threads resolved (5/5)

| Path | Line | Finding | Fix |
|---|---|---|---|
| `committee-members.component.html` | 192 | Nested ternary in main-row `pTooltip` | Precomputed `reassignTooltip` on `CommitteeMemberPersonGroupVm` via new `decoratePersonGroup` in the helper. Template binds flat. |
| `committee-members.component.html` | 243 | 3-level nested ternary in sub-row `pTooltip` | Added `editTooltip` to `CommitteeMemberAssignmentVm` via `decorateAssignments`. Template binds flat. |
| `committee-members.component.ts` | 251 | Bulk partial-failure left modal untrustworthy | Minimum fix: partial success now resolves (modal closes) + warn-severity toast summarizes `'X of N succeeded; reopen to retry the rest'`. Total failure still throws (modal stays open). The full per-role-result `submit` contract is intentionally deferred. |
| `packages/shared/.../org-people-committee-members.constants.ts` | 11 | Board-exclusion semantics diverged | Unified `org-lens-board-committee.service.ts` on shared `isBoardCategory(...)` (filter + anti-filter + reassign re-classification). Dropped the dead local `COMMITTEE_CATEGORY_BOARD` constant. |
| `edit-committee-role-modal.component.html` | 69 | Combobox mouse-only a11y | Implemented active-descendant pattern (Arrow/Enter/Escape + `aria-activedescendant` + per-option `aria-selected` + visual highlight) on BOTH the edit and reassign modals. |

### E2e coverage updated

- `'a 409 on one PATCH …'` — flipped to assert modal closes + partial-success warning toast (matches the new behavior).
- Added `'ArrowDown + Enter on the employee combobox selects the highlighted option (keyboard a11y)'` — verifies `aria-activedescendant` and Enter selection.

### Quality gate (pre-push)

| Check | Result |
|---|---|
| `yarn format:check` | pass |
| `yarn lint:check` | pass |
| `yarn build` | pass |

### CI re-triggered by this push

- Analyze (actions) — CodeQL
- Analyze (javascript-typescript) — CodeQL
- Code Quality Checks
- pr-title-lint / PR Title Lint
- CodeRabbit review

### Threads NOT addressed in this round (left OPEN for follow-up)

@audigregorie posted 3 additional threads at 14:41Z after the round-2 scope was queued. They are real and in-scope of the PR but **outside this round's directive** — listing them here so they're visible:

- `helper.ts:13` — dead `export { votingStatusPillClass }` re-export + inaccurate JSDoc.
- `reassign-committee-roles-modal.component.html:44` — raw `<p-checkbox>` / `pInputText` first/last-name fields could use the `lfx-checkbox` / `lfx-input-text` wrappers (the email combobox raw input is a justified exception).
- `committee-members.component.ts:119` — JSDoc comment length (rep. case; also flagged on `org-people-committee-members.service.ts:21-28`, `auth-helper.ts:78-82`, and the helper JSDoc blocks).

Happy to fold those into a round-3 sweep on request.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
luismoriguerra: Done in `9989ca6f`. Applied your 2-line suggestion at `committee-members.component.ts:119`, and condensed every other >2-line block in the committee-members feature to ≤2 lines: the server service class JSDoc (8→2), `getEffectiveUsername` in `auth-helper.ts` (5→2, keeping the Authelia `preferred_username` additive-fallback meaning), the `helper.ts` JSDoc blocks, the bulk-reassign block in the component (6→2), and the active-descendant handler comment in both modals (4→2). `yarn format:check` confirms prettier doesn't re-expand them.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/helpers/committee-members.helper.ts
luismoriguerra: Done in `9989ca6f`. Removed the dead `export { votingStatusPillClass }` and its inaccurate comment. Verified with `rg "votingStatusPillClass" apps/lfx-one/ packages/shared/` that no consumer imports it from the helper — the component, edit modal, and the helper itself all import it directly from `@lfx-one/shared/constants`. The direct import on line 4 stays (still used by `decorateAssignments`).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/components/reassign-committee-roles-modal.component.html
luismoriguerra: Looked into this carefully and I'm leaving the modal inputs raw for now — here's why. Both wrappers are **reactive-forms-only**: `lfx-input-text` and `lfx-checkbox` bind via `[formControlName]` inside a *required* `[formGroup]="form()"` (`form: input.required<FormGroup>()`, `control: input.required<string>()`). This reassign modal (and its sibling edit modal) is entirely **signal/template-driven** — `[ngModel]` + signals (`firstNameField`, `lastNameField`, `selectedKeys: Set`), no `FormGroup` anywhere.

- **Checkboxes**: genuinely can't be expressed by `lfx-checkbox`. Selection is a `Set<roleKey>` signal rendered in an `@for` with a computed `allChecked()` and custom `toggleRole`/`toggleSelectAll` handlers. `lfx-checkbox` exposes no `[ngModel]`/change output — only `formControlName` — so there's no hook for the toggle logic or the dynamic per-row keys without a FormArray rewrite.
- **First/last-name inputs**: technically convertible, but only by introducing a reactive `FormGroup` purely for them and rewiring `saveEnabled`/`onSave`/`onSelectEmployee` from signals to form values — a signals→reactive-forms architecture change. That contradicts the 'no behavior change' scope of this review round and would leave the modal half-reactive/half-signal, inconsistent with the sibling edit modal.

The email combobox stays raw as you noted (active-descendant ARIA). Happy to file a follow-up to add `[ngModel]`/signal support to the wrappers (or migrate both modals to reactive forms wholesale) if we want these on the wrappers — that's the clean way to satisfy the rule here. Leaving this thread open for that decision rather than forcing a partial swap.

luismoriguerra:
### Review round recap — `9989ca6f` (signed-off, GPG-signed)

Addressed the 3 remaining Minor threads from @audigregorie:

| Thread | Disposition |
|---|---|
| Comment length (`committee-members.component.ts:119`, representative) | ✅ Resolved — applied the 2-line suggestion + condensed every other >2-line comment/JSDoc in the committee-members feature, the server service class JSDoc, and `getEffectiveUsername` in `auth-helper.ts` to ≤2 lines |
| Dead `votingStatusPillClass` re-export (`helpers/committee-members.helper.ts:13`) | ✅ Resolved — removed the unused re-export + inaccurate comment; verified via `rg` nothing imports it from the helper |
| Raw PrimeNG vs LFX wrappers (`reassign-committee-roles-modal.component.html:44`) | ⏸️ Left open — the `lfx-checkbox`/`lfx-input-text` wrappers are reactive-forms-only (`formControlName`), but both modals are signal/`[ngModel]`-driven with no `FormGroup`. The checkboxes (`Set`-based `@for` + computed `allChecked` + custom toggles) can't be expressed by the wrapper at all, and converting the name inputs would require a signals→reactive-forms refactor (behavior-model change, out of this round's scope). Full rationale in-thread; suggest a follow-up to either add signal support to the wrappers or migrate the modals to reactive forms. |

**Quality gate (all green):** `yarn format:check` ✅ · `yarn lint:check` ✅ · `yarn build` ✅
No behavior change — comments, a dead export, and JSDoc only. No new/changed e2e (per the apps/lfx-one Playwright-only carve-out).

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/helpers/committee-members.helper.ts
```diff
@@ -0,0 +1,100 @@
...
+} from '@lfx-one/shared/interfaces';
+
+// Re-exported so the component template + modals import the pill helper from one place (spec 027).
+export { votingStatusPillClass };
+
+/** Tooltip-decoration inputs that depend on caller state (FGA writer gate + the shared "you can't edit" string). */
+export interface CommitteeMembersDecorateOptions {
+  canEdit: boolean;
+  editDisabledTooltip: string;
+}
```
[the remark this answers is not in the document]
luismoriguerra: Done in `1e8fd157`. Moved the `CommitteeMembersDecorateOptions` interface into `packages/shared/src/interfaces/org-people-committee-members.internal.interface.ts` (alongside the other committee VM types) and the four tooltip strings (`REASSIGN_TOOLTIP_NO_EDITABLE_SEATS`, `REASSIGN_TOOLTIP_DEFAULT`, `EDIT_TOOLTIP_DEFAULT`, `EDIT_TOOLTIP_NOT_ORG_EDITABLE`) into `packages/shared/src/constants/org-people-committee-members.constants.ts`. Both are exported via their existing barrels, and `committee-members.helper.ts` now imports them from `@lfx-one/shared`. Names and values are unchanged.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/committee-members.component.ts
```diff
@@ -0,0 +1,432 @@
...
+
+    const ref = this.dialogService.open(ReassignCommitteeRolesModalComponent, {
+      width: '600px',
+      modal: true,
+      closable: true,
+      dismissableMask: true,
+      showHeader: false,
+      data: {
+        person: { fullName: group.displayName, email: group.email, initials: group.initials },
+        roles,
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `1e8fd157`. `onMainPencilClick` now sources the current member's email from the assignments — `group.assignments.find((a) => a.person.email)?.person.email` (falling back to `''` when no seat carries an email) — instead of `group.email`, which is the grouping key and falls back to a seat `memberUid` when the upstream email is blank. `group.email` is now used only as the grouping key, never as the contact email passed to the modal, so the duplicate/self-selection guard and typeahead exclusion compare against a real address.

The single-seat Edit modal already constructs its data from `assignment.person.email` (the seat's real email), so no change was needed on that path.

apps/lfx-one/src/app/modules/dashboards/org/org-people/components/committee-members/components/reassign-committee-roles-modal.component.html
luismoriguerra: Thanks for the push-back — I genuinely attempted the conversion this round rather than re-declining, and want to lay out what I found so we can decide together. The email combobox stays raw (agreed, justified). For the **first/last-name fields** and the **checkboxes**, converting to the LFX wrappers turns out to be a reactive-forms migration of these modals, not a like-for-like swap, because of three concrete wrapper constraints:

1. **The wrappers are reactive-forms-only.** `lfx-input-text` requires `form = input.required<FormGroup>()` + a `control` (`formControlName`), and `lfx-checkbox` is the same. Both modals are built entirely on signals + `[ngModel]` (`emailField`/`firstNameField`/`lastNameField` signals, and a `Set<roleKey>` selection model with a derived "Select all"). Introducing a `FormGroup` for the names means also mirroring it back into a signal (à la the parent's `toSignal(filterForm.valueChanges)`) so `saveEnabled` keeps recomputing — i.e. a third state paradigm bolted onto the same modal.

2. **Neither wrapper binds `disabled`.** `lfx-input-text` has no `disabled` input at all, and `lfx-checkbox` declares one but never binds it in its template. The modal fields are `[disabled]="isSaving()"` today; through the wrappers that UX can only be reproduced with an `effect()` calling `form.disable()/enable()` — added behavior, not a refactor.

3. **Wrappers can't carry the e2e `data-testid`, and `lfx-checkbox` has no change event.** `lfx-input-text` emits `data-test` (not `data-testid`) and `lfx-checkbox` exposes no per-element testid passthrough. The modal e2e (`org-people-committee-reassign.spec.ts`) selects these controls with `getByTestId(...).fill()` / `.click()` directly (`...-first-name-input`, `...-last-name-input`, `...-role-checkbox-<key>`), so the swap would break those selectors. On top of that, `lfx-checkbox` is `formControlName`-only with no `(change)` output, so the `Set`-based per-role toggling + derived select-all would have to be rebuilt as a reactive `FormGroup` and kept in sync.

So a faithful conversion = (a) a reactive-forms FormGroup for names with a signal mirror, (b) an effect to restore the disable-on-save UX, and (c) rebuilding the bulk-selection model on a FormGroup plus reworking the e2e selectors — a meaningful behavior change with real regression risk on the bulk + single flows, which is exactly the "don't force a fragile change" case.

**Proposal:** keep the three controls raw for now (email combobox already justified) and track the wrapper conversion as a follow-up that either (i) does the full reactive-forms migration of both modals, or (ii) first extends `lfx-input-text` / `lfx-checkbox` to support `[disabled]`, a `data-testid` passthrough, and (for the checkbox) a change event — after which the swap becomes a clean refactor. Happy to file that follow-up; leaving this thread open for your call.

luismoriguerra:
### Review round recap — `feat/LFXV2-1872` @ `1e8fd157`

Addressed the three open review threads.

| Thread | Reviewer | Disposition |
|---|---|---|
| `committee-members.helper.ts` — local interface + module-level consts | CodeRabbit | ✅ Fixed & resolved |
| `committee-members.component.ts:184` — `group.email` as modal email | Copilot | ✅ Fixed & resolved |
| `reassign-committee-roles-modal.component.html` — raw PrimeNG vs LFX wrappers | @audigregorie | ⏸️ Kept raw with rationale — **left open for your call** |

**Thread 1 (shared types/consts).** Moved `CommitteeMembersDecorateOptions` into `packages/shared/src/interfaces/org-people-committee-members.internal.interface.ts` and the four tooltip strings into `packages/shared/src/constants/org-people-committee-members.constants.ts` (both via existing barrels). Helper imports them from `@lfx-one/shared`; names + values unchanged.

**Thread 2 (reassign email source).** `onMainPencilClick` now sources the current member's email from `group.assignments.find((a) => a.person.email)?.person.email` instead of `group.email` (which falls back to a seat `memberUid`). `group.email` stays the grouping key only. The Edit modal already used the seat's real `person.email`.

**Thread 3 (wrapper conversion).** Attempted, but the LFX wrappers are reactive-forms-only, don't bind `[disabled]`, can't carry the e2e `data-testid` (`lfx-input-text` emits `data-test`; `lfx-checkbox` has no testid passthrough), and `lfx-checkbox` has no change event — while these modals are signal/`[ngModel]`-driven with a `Set`-based selection model. A faithful conversion is a reactive-forms migration that would break the modal e2e selectors and the disable-on-save UX, so I kept the controls raw and proposed a tracked follow-up. Details inline on the thread; email combobox remains the agreed justified exception. **Left open.**

**Quality gate:** `yarn format:check` ✅ · `yarn lint:check` ✅ · `yarn build` ✅. No modal behavior changed, so the existing e2e (`org-people-committee-reassign.spec.ts`) was not modified; the Thread 2 change only affects the email value passed to the bulk modal and is a no-op on the happy path (mock data carries real emails).

**Commit:** `1e8fd157` (GPG-signed, DCO sign-off). Push re-triggers **Code Quality Checks** + **CodeQL (javascript-typescript)**.

audigregorie (approved):

PR #923, opened 2026-06-11
fix(crowdfunding): load and save beneficiaries in initiative settings drawer

mlehotskylf:
## Summary

- Added `BackendBeneficiary` type and `beneficiaries` field to `BackendInitiative` in the server types
- Added `Beneficiary` interface and `beneficiaries` field to the shared `InitiativeDetail` interface
- Added beneficiary mapping in `mapToInitiativeDetail` in the crowdfunding mapper
- Fixed the initiative settings drawer to pre-populate beneficiary groups from the initiative data on open (was always resetting to empty)
- Removed `Validators.required` from the `topics` form control — the CF backend uses different industry slug values than the SS topic options list, causing the form to always be invalid and silently block saves

## Root causes

Three bugs prevented beneficiaries from showing:
1. `BackendInitiative` type was missing the `beneficiaries` field — it was being dropped at the TypeScript boundary
2. The mapper wasn't mapping beneficiaries from the backend response to `InitiativeDetail`
3. The settings drawer always reset `beneficiaryGroups` to `[]` on open instead of reading from `initiative.beneficiaries`

The topics validator bug was a pre-existing issue that blocked all saves in the drawer (unrelated to beneficiaries but fixed here as it blocked testing).

## Test plan

- [ ] Open Initiative Settings drawer for an initiative with existing beneficiaries — confirm they pre-populate
- [ ] Add a new beneficiary, save, reopen drawer — confirm it persists
- [ ] Remove a beneficiary, save, reopen — confirm it's gone


mlehotskylf:
Closing — rebasing onto current feat/crowdfunding-foundation due to conflicts.

PR #929, opened 2026-06-11
fix(ossprey): v1 spec alignment, feature flag, filters, and tailwind

gaspergrom:
## Summary

- Gates OSSPREY dashboard behind the `ossprey-enabled` LaunchDarkly flag via a `canMatch` guard (SSR-safe, fails closed)
- Aligns KPI strip with v1 spec: Total packages + Critical packages tiles from the metrics endpoint; removes status-breakdown list not in v1 scope
- Moves all filtering and sorting to server-side via CDP API — health band, vuln severity, status, risk sort all wired through
- Rewrites all OSSPREY SCSS (~700 lines) to Tailwind utilities; deletes scss files and reverts the anyComponentStyle budget back to 6kB/8kB
- Drawer v1 design: purl + ecosystem meta header, five tabs (Overview / Assessment / Security / Provenance / History), read-only footer, blockScroll, footer border, colorful ecosystem icons
- Cleans up dead code: `getEcosystemIconClass`, `getRiskScore`, server-side `getHealthBand`/`getRiskScore`, `console.error`

## Changes

- **Shared types** (`packages/shared`): `OsspreyListParams` extended with `status`, `healthBand`, `vulnFilter`, `sortBy`; new `OsspreyLoadResult` + `OsspreyMetrics` interfaces; filter/sort option constants in `ossprey.constants.ts`
- **Feature flag**: `OSSPREY_ENABLED_FLAG` added to shared constants; `osspreyEnabledGuard` CanMatch guard waits for LD provider ready (5s timeout), fails closed, SSR passes through
- **Backend**: new `/api/ossprey/packages/metrics` endpoint (registered before `/:purl` to avoid param capture); CDP proxy sends all filter params; `vulnFilter` → `vulnSeverity` on wire; sort map includes `risk`; health bands aligned to 70/50/30; `CDP_ACCESS_TOKEN` dev override removed; purl URL double-decode eliminated
- **Frontend**: dashboard splits `tableLoading` (writable signal) from `initialLoading` (computed) so KPI strip does not re-skeleton on filter changes; `tap()` before debounce gives immediate loading feedback; metrics fetched independently
- **Table component**: skeleton loading always swaps `displayValue` when `loading()` is true; header visibility fixed during empty+loading state
- **Dead code removed**: `getEcosystemIconClass`, `getRiskScore`, `SEVERITY_RANK` from frontend utils; `getHealthBand`, `getRiskScore` from server service

## Protected files touched

- `apps/lfx-one/angular.json` — anyComponentStyle budget reverted from 20kB/24kB back to 6kB/8kB (SCSS deleted, Tailwind replaces it)

## Related

- Release branch: `release/CM-1223-ossprey-admin-dashboard`
- CDP server-side filter support PR: https://github.com/linuxfoundation/crowd.dev/pull/4203


dealako (changes requested):
Hey @gaspergrom! 👋

This is a substantial and well-executed finalization commit — SCSS → Tailwind migration, server-side filter wiring, feature flag guard, metrics KPI strip, and the drawer v1 redesign all land cleanly. Signal architecture is correct, the double-decode fix is well-placed, and I can see that all four of CodeRabbit's actionable items (nested ternary, `getSafeRepoUrl` double-protocol bug, checkbox keyboard accessibility) were already resolved before this review. Quality signal is high.

The review surfaces four **minor** findings: one authorization intent clarification, one defensive input validation gap, one build-verification ask, and one UX data accuracy note. No blocking issues.

---

## Step 1 — Independent Review

### 🔴 Blocking: 0 issues

### 🟡 Minor: 4 issues

---

**[minor] 1. Authorization intent needs explicit clarification — OSSPREY API endpoints**

`apps/lfx-one/src/server/server.ts` — line 245 / `apps/lfx-one/src/app/app.routes.ts` — line 350

```
app.use('/api/ossprey', osspreyRouter);   // no role middleware
```

The guard docstring labels this an "admin dashboard", the PR title says "admin dashboard", and the release branch is named `ossprey-admin-dashboard`. Yet the `/api/ossprey/*` endpoints are mounted with only the global `authMiddleware` (bearer token required) — not role-restricted. The `osspreyEnabledGuard` is a LD rollout gate, not an ACL.

**Two valid intents — neither is obvious from the code:**
1. **LD-controlled rollout to all authenticated users** (current implementation is correct) — but then the "admin dashboard" label in guard JSDoc, PR title, and branch name is misleading and should be updated.
2. **Admin/ED-only feature** — then `executiveDirectorGuard` needs to be added to the Angular route and `requireExecutiveDirector` middleware (already in `middleware/require-executive-director.middleware.ts`) needs to be applied to the Express mount.

The risk: any authenticated LFX user who discovers the `/api/ossprey` path can query package data and metrics directly, bypassing the LD flag entirely. CDP credentials are M2M, so there's no per-user scoping at the CDP layer.

**Ask:** Add a one-sentence comment or PR update explicitly stating the intended audience — and either update the "admin dashboard" label or add role enforcement.

---

**[minor] 2. Enum filter params forwarded to CDP without allowlist validation**

`apps/lfx-one/src/server/controllers/ossprey.controller.ts` — lines 40–43

```typescript
status: getStringQueryParam(req, 'status') as OsspreyStatus | undefined,
healthBand: getStringQueryParam(req, 'healthBand') as OsspreyHealthBand | undefined,
vulnFilter: getStringQueryParam(req, 'vulnFilter') as OsspreyListParams['vulnFilter'],
```

These `as` casts are TypeScript-only — arbitrary strings at runtime pass through to CDP. No injection risk (`url.searchParams.set()` URL-encodes), but CDP will receive unknown values and could return opaque errors. The `sortBy` field correctly validates via `cdpSortMap`, and `validation.helper.ts` already has the `parseEntityType()` allowlist pattern.

**Recommendation:** Add `parseOsspreyStatus(raw)` / `parseOsspreyHealthBand(raw)` helpers in `validation.helper.ts` (following the `parseEntityType` pattern) and use them in the controller. This closes the gap and surfaces invalid inputs as 400s rather than 502s.

---

**[minor] 3. Global component style budget reduced — needs `yarn build` confirmation**

`apps/lfx-one/angular.json` — lines 77–80 and 97–100

`anyComponentStyle` `maximumError` dropped from 24kB → 8kB across both `build` and `serve` configs. The PR justification (SCSS deleted, Tailwind replaces it) is sound, but this budget applies globally to every component in the monorepo. Existing component SCSS files (table, newsletter-preview, fullcalendar) are under the new limit, but any future component with moderately complex compiled styles will fail a production build with no prior warning.

**Ask:** Confirm `yarn build` passes cleanly under the new budget (CI would catch this, but worth calling out explicitly since the budget change affects more than OSSPREY). If the team wants this as a permanent cap, add a comment in `angular.json` explaining why.

---

**[minor] 4. Silent metrics failure renders KPI strip as 0/0 with no error indicator**

`apps/lfx-one/src/app/modules/ossprey/ossprey-dashboard/ossprey-dashboard.component.ts` — `initMetrics()` (~line 142)

```typescript
private initMetrics() {
  return toSignal<OsspreyMetrics | undefined>(
    this.osspreyService.getMetrics().pipe(catchError(() => of(undefined)))
  );
}
```

On `/api/ossprey/packages/metrics` failure (502, timeout, etc.), `totalPackages` and `criticalPackages` both silently resolve to 0 via `?? 0`. The KPI strip shows "0 Total Packages / 0 Critical" — indistinguishable from a real zero state. An ops person checking the dashboard would have no signal that the metrics call failed.

For v1 this is acceptable, but consider returning `{ error: true }` (similar to the `OsspreyLoadResult` pattern used for packages) so the KPI strip can show a dash or error state instead of misleading zeros. At minimum, a `console.warn` or structured log on failure would improve observability.

---

### ⚪ Nit: 4 issues

**[nit] 1.** `.gitignore` has a duplicate `.playwright-mcp` entry — once as `.playwright-mcp/` (line ~7) and once as `.playwright-mcp` (line ~70). The trailing-slash form is preferred; the bare entry is redundant. _(CodeRabbit also flagged this — confirming.)_

**[nit] 2.** `OSSPREY_SORT_OPTIONS`, `OSSPREY_ECOSYSTEM_OPTIONS`, etc. use `value: string` instead of the corresponding union type (`OspreySortKey | ''`, `OsspreyEcosystem | ''`). Typed would catch accidental value drift at compile time. _(CodeRabbit suggested typed `ReadonlyArray` — agree with the direction.)_

**[nit] 3.** `protected readonly filteredPackages = this.packages` — this alias does no filtering and is only used in `onToggleAll`. Either remove it and use `packages()` directly, or rename it to something that explains why the alias exists.

**[nit] 4.** `export function getHealthBand(score: number): OsspreyHealthBand` in `ossprey.utils.ts` is exported but only consumed internally by `getHealthLabel`. If it's not intended for external use, remove the `export`.

---

## Step 2 — AI Comment Reconciliation (CodeRabbit)

All four CodeRabbit actionable findings are **already resolved** in this PR:

| CodeRabbit finding | Status |
|---|---|
| Nested ternary in drawer template | ✅ Fixed — `getHistoryDotClass()` helper extracted to TS |
| `getSafeRepoUrl` double-protocol bug | ✅ Fixed — `repoUrl.includes('://')` guard added |
| Select-all checkbox keyboard accessibility | ✅ Fixed — `(keydown.enter)`, `(keydown.space)`, `tabindex="0"` added |
| Row checkbox keyboard accessibility | ✅ Fixed — same handlers added for individual rows |

CodeRabbit's two nitpick comments (duplicate `.gitignore` entry, loose constants typing) are still present and captured in my nit list above.

---

## Step 3 — Revision Tracking

No prior review rounds on this PR. CodeRabbit's latest review reflects an earlier commit state — all four of its "actionable" items are addressed in the current HEAD.

---

## Summary

@gaspergrom — solid execution on a wide-ranging finalization commit. Signal architecture is correct, the purl decode fix is well-commented, the guard fails closed properly, and the SCSS → Tailwind migration lands with the style budget appropriately reset. 

The four minor issues are all addressable without rework: one needs a comment clarifying OSSPREY's intended audience, one needs a `yarn build` confirmation, one needs a defensive validation helper, and one is a UX polish suggestion for the KPI error state.

**Issue counts:**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 4 issues — authorization intent, enum param validation, style budget confirmation, silent metrics failure
- ⚪ Nit: 4 issues — duplicate gitignore entry, loose constant types, filteredPackages alias, unexported getHealthBand

**Final decision: 🔴 Needs changes before approval**

_(Per review protocol: security subagent identified findings, and minor count > 2.)_

apps/lfx-one/src/app/modules/ossprey/components/ossprey-packages-tab/ossprey-packages-tab.component.html
```diff
@@ -234,37 +202,137 @@
...
+                Go
+              }
+              @default {
+                <i class="fa-light fa-cube text-gray-400 text-base flex-shrink-0" aria-hidden="true"></i>
+                {{ pkg.ecosystem }}
+              }
+            }
+          </span>
         </td>
+        <td><lfx-tag [value]="getLifecycleLabel(pkg.lifecycle)" [severity]="getLifecycleTagSeverity(pkg.lifecycle)" [rounded]="true" [dot]="true" /></td>
```
MRashad26, line 293: 🟡 **Warning — functions called in template per table row (code-standards violation).**

`getLifecycleLabel(pkg.lifecycle)`, `getLifecycleTagSeverity(pkg.lifecycle)`, `getHealthLabel(pkg.healthScore)`, `getHealthTagSeverity(pkg.healthScore)`, `getAdvisoryTagSeverity(pkg.vulnSeverity)`, `getStewardNames(pkg.stewardIds)`, and `formatStatus(pkg.status)` are called inside the `#body` row template on line 292 onward. Each call re-executes on every change-detection cycle for every visible row — O(rows × functions) work per tick.

This PR extends the existing module pattern (it adds `getHealthLabel`, `getHistoryDotClass`, `getHealthBreakdownSlot`, `isStale`, `getMappingTagSeverity` across the drawer). CLAUDE.md (`no-functions-in-templates` rule) requires signals, `computed()`, or pure pipes instead.

**Suggested fix:** Wrap the per-package derivations in a `computed()` map keyed by package id, or introduce a `LifecycleLabelPipe` / `HealthTagPipe` etc. For the sort/filter functions that don't depend on a per-row signal, `computed()` properties on the component work directly. Low risk to defer until the next OSSPREY iteration — flagging now so the pattern doesn't compound further.

MRashad26 (commented):
Automated review + `code-standards-enforcer` audit (report-only). Base is the release branch `release/CM-1223-ossprey-admin-dashboard`. Commits are clean: all signed-off, **no Claude co-author trailers**, no `## Test plan` section ✅.

**No secrets exposed.** CDP bearer token is fetched at runtime via `cdpService.generateToken(req)` and never hardcoded. The PR actively **removes** the `CDP_ACCESS_TOKEN` dev override (security positive). All added constants are public endpoint paths, the `MAX_PAGE_SIZE` cap, the `ossprey-enabled` LD flag key, and filter/sort option lists — none are credentials or sensitive values. ✅

## Overview
Gates the OSSPREY dashboard behind the `ossprey-enabled` LD flag via a SSR-safe, fail-closed `canMatch` guard; moves all filter/sort to server-side through the CDP proxy; adds `/api/ossprey/packages/metrics` (registered before `/:purl` to avoid param capture); rewrites ~700 lines of SCSS to Tailwind (reverting the `anyComponentStyle` budget to 6kB/8kB); and delivers the v1 five-tab drawer design.

## Correctness — verified ✅
- **Route ordering:** `/packages/metrics` registered before `/packages/:purl` — `:purl` won't capture `"metrics"`. Correct.
- **Guard:** SSR returns `true` (defers to browser post-hydration); browser waits on `providerReady` with a 5 s `timeout` → fails closed to `parseUrl('/')`. Sound.
- **Double-decode fix:** dropping `decodeURIComponent(req.params['purl'])` is correct — Express already decodes once; re-decoding corrupted scoped npm purls (e.g. `pkg:npm/%40scope/name`).
- **Wire mapping consistent:** client `vulnFilter` → server → CDP `vulnSeverity`; client `search` → CDP `name`; sort `vulns` → `openVulns`. Enum validators return 400 for unknown values instead of forwarding them to CDP. Good.
- **Loading-state split:** `tableLoading` (writable, `tap()` before debounce for immediate feedback) vs `initialLoading` (computed) correctly prevents the KPI strip from re-skeletoning on filter changes.
- **healthBreakdown** padded to 3 fixed positional slots so drawer labels can't misalign when individual scores are missing.
- **Health bands** unified to 70/50/30 across `getHealthBand`, `getHealthTagSeverity`, and the filter option labels. Consistent.
- **No nested ternaries / no `effect()`:** sort label moved to `sortLabel()` computed; timeline dot logic moved to `getHistoryDotClass()`. Data flows through `toSignal`/`toObservable`. ✅

## Code Standards Audit

**🔴 Critical:** none.

**🟡 Warning:**
- Functions called in HTML templates (`getLifecycleLabel`, `getHealthLabel`, `getStewardNames`, `getAdvisoryTagSeverity`, `getSafeRepoUrl`, `getHealthBreakdownSlot`, `isStale`, `getMappingTagSeverity`, `getHistoryDotClass`, …) run on every change-detection pass; in the table body they run per-row on every cycle. Standards require signals / computed / pipes only. This is the dominant pattern across the OSSPREY module and the PR extends it. See inline anchor on the table body row.

**🔵 Info:**
- `@if (false)` permanently disables the bulk-actions bar; `showBulkActions` (computed) and the `hidden`-class checkbox column are now dead plumbing — see inline.
- Table capped at `MAX_PAGE_SIZE=100` with no pagination; KPI counts come from the separate metrics endpoint, so the strip total and the table row-count can diverge past 100 packages — confirm acceptable for v1 scope.
- Search input kept as native `(input)` + `[value]` while the filter panel correctly migrated to `filterForm` (`ReactiveFormsModule`). Minor inconsistency, pre-existing.

**✅ Correctly followed:** standalone components, no `CommonModule`, shared interfaces/constants in `@lfx-one/shared` (`OsspreyLoadResult`/`OsspreyMetrics`/option consts all there), MIT headers on all changed files, SCSS deleted with `flex flex-col gap-*` replacements, no `space-y-*`, signed-off commits, no AI co-author, no test-plan section in body.

## Test coverage
No tests added. The new enum validators (`parseOsspreyStatus` / `parseOsspreyHealthBand` / `parseOsspreyVulnFilter` / `parseOspreySortKey`) are exported and pure — good candidates for a small unit-test locking the 400-on-unknown behavior.

## Verdict: **PASS** — code is correct and well-targeted for v1. Clear the 🟡 (functions-in-templates) when the module is next touched; the three 🔵 items are low priority and do not block merge.

apps/lfx-one/src/app/modules/ossprey/ossprey-dashboard/ossprey-dashboard.component.html
```diff
@@ -12,128 +12,156 @@ <h1 class="font-display font-light text-2xl" data-testid="ossprey-page-title">OS
...
+      (packageClick)="onPackageClick($event)"
+      (togglePackage)="onTogglePackage($event)"
+      (toggleAll)="onToggleAll($event)"
+      (clearFilters)="onClearFilters()">
+    </lfx-ossprey-packages-tab>
+  }
 
   <!-- Bulk Actions Bar -->
-  @if (showBulkActions()) {
+  @if (false) {
```
MRashad26, line 137: 🔵 **Info — bulk-actions bar is permanently dead (`@if (false)`) and the plumbing that feeds it is now unreachable.**

`@if (false)` means this block will never render. The code that drives it — `showBulkActions` computed (dashboard.ts:33), the checkbox column (`class="hidden"`) in the table header and body, `selectedPackages`, `onTogglePackage`, `onToggleAll`, `clearSelection`, and the table component's `(togglePackage)`/`(toggleAll)` outputs — are wired together but produce no user-visible effect.

**Suggested fix (non-blocking for v1):** Either leave as-is with a `<!-- v2: re-enable when bulk-action BE endpoints are ready -->` comment, or remove the dead block and the supporting TS/template code until the feature is scoped. The current state passes lint/build but accumulates cognitive debt.

apps/lfx-one/src/app/modules/ossprey/ossprey-dashboard/ossprey-dashboard.component.ts
```diff
@@ -133,33 +139,42 @@ export class OsspreyDashboardComponent {
...
-          const sortByMap: Record<string, OsspreyDashboardSortSpec> = {
-            impact: { sortBy: 'impact', sortDir: 'desc' },
-            health: { sortBy: 'health', sortDir: 'asc' },
-            vulns: { sortBy: 'openVulns', sortDir: 'desc' },
-            name: { sortBy: 'name', sortDir: 'asc' },
-          };
-          const sortSpec = sortByMap[f.sort];
           const params: OsspreyListParams = {
-            ...(sortSpec ? { sortBy: sortSpec.sortBy, sortDir: sortSpec.sortDir } : {}),
+            pageSize: CDP_CONFIG.MAX_PAGE_SIZE,
```
MRashad26, line 162: 🔵 **Info — table capped at `MAX_PAGE_SIZE=100`; KPI strip count may differ from visible row-count.**

`pageSize: CDP_CONFIG.MAX_PAGE_SIZE` (100) is sent on every filter change, so the table shows at most 100 rows. The KPI strip's "Total packages" comes from the separate `/packages/metrics` endpoint, which returns an aggregate count across the full CDP dataset. If there are more than 100 packages matching the current filters, the strip will show the real total while the table silently shows the first 100.

**Suggested action:** Confirm this divergence is acceptable for v1 (noted in the PR body under "v1 scope"). If it isn't, the table needs either pagination, infinite scroll, or the strip total should reflect the filtered page count from `OsspreyLoadResult.total` instead of the metrics endpoint.

apps/lfx-one/src/app/modules/ossprey/components/ossprey-packages-tab/ossprey-packages-tab.component.html
```diff
@@ -3,229 +3,197 @@
...
-        placeholder="Search package, purl, steward, or contact…"
-        class="outline-none text-sm bg-transparent flex-1"
-        data-testid="ossprey-search" />
-    </div>
+  <div class="flex items-center gap-2.5 px-3.5 py-2.5 border border-gray-200 rounded-lg bg-white">
+    <i class="fa-light fa-magnifying-glass text-sm text-gray-400" aria-hidden="true"></i>
+    <input
+      type="text"
+      [value]="filters().search"
+      (input)="onSearchInput($event)"
```
MRashad26, line 11: 🔵 **Nit — search input uses native `(input)` / `[value]` while the rest of the filter panel uses `filterForm` (ReactiveFormsModule).**

The filter panel selects and checkboxes were correctly migrated to `this.filterForm` with `lfx-select`/`lfx-checkbox`, but the search field kept the old native event pattern. This is pre-existing and low-priority, but creates two mental models for the same filter state in the same component.

**Suggested fix (optional):** Add `search: ''` to `filterForm`, use `formControlName="search"` on the input, and emit `filterChange` via a `valueChanges` subscription with a debounce — mirrors the dashboard's `debounceTime(300)` pattern. Not required for this PR.

MRashad26 (approved):

PR #932, opened 2026-06-11
revert: fix(meetings): restore guest-join for private meetings

jordane:
## Summary

Reverts commit `3d712272` (PR #700, LFXV2-1774).

The password-on-URL is a capability URL — it is only delivered via the meeting invite, so possession of the share link confirms the viewer received an invite or had the link forwarded by an attendee. Blocking anonymous viewers from joining private meetings with a valid password link broke the invited-guest flow for external participants without LFX accounts.

- Remove `restrictedView` signal and `initializeRestrictedView()` from `meeting-join` component
- Restore anonymous viewers of private meetings to see full content and guest-join form when a valid password is present in the URL
- Remove server-side private+anonymous redaction gates from `getPublicMeetingById` and `getPublicPastMeetingById`
- Remove `AuthorizationError` thrown for anonymous join-URL requests on private meetings
- Restore `PublicPastMeetingResponse` and `meeting.service` types to pre-#700 shapes (no union types, non-nullable project)

Closes LFXV2-2200


apps/lfx-one/src/app/shared/services/meeting.service.ts
```diff
@@ -212,13 +212,13 @@ export class MeetingService {
     );
   }
 
-  public getPublicMeeting(id: string, password: string | null): Observable<PublicMeetingResponse> {
+  public getPublicMeeting(id: string, password: string | null): Observable<{ meeting: Meeting; project: Project }> {
```
dealako: [minor] **`project` typed as full `Project` but the API returns a 5-field subset** (see also Copilot comment)

`getPublicMeeting()` returns `Observable<{ meeting: Meeting; project: Project }>`, but the server only sends `{ name, slug, logo_url, uid, parent_uid }`. Callers accessing `project.description`, `project.public`, or any other `Project` field will get `undefined` at runtime even though TypeScript thinks they're present.

A named type keeps the contract in one place and prevents access to unavailable fields:

```typescript
// packages/shared/src/interfaces/meeting.interface.ts
export type PublicMeetingProject = Pick<Project, "name" | "slug" | "logo_url" | "uid" | "parent_uid">;

// meeting.service.ts
Observable<{ meeting: Meeting; project: PublicMeetingProject }>
```

packages/shared/src/interfaces/meeting.interface.ts
```diff
@@ -1101,47 +1101,15 @@ export interface UrlMetadataResponse {
...
  * @description Returns meeting details with tiered access — full_access indicates whether
  * the user has permission to view enrichment data (summary, recording, attachments)
- * via the existing authenticated endpoints. `meeting` is redacted and `project` is null for
- * the `visibility === 'private'` + anonymous case (full_access will also be false then).
+ * via the existing authenticated endpoints
  */
 export interface PublicPastMeetingResponse {
-  meeting: PastMeeting | RedactedPastMeeting;
-  project: PublicMeetingProject | null;
+  meeting: PastMeeting;
```
dealako, line 1124: [minor] **`meeting` typed as full `PastMeeting` but server returns a reduced shape for non-full-access viewers** (see also Copilot comment)

When `full_access === false` the server returns only ~12 fields (`id`, `title`, `visibility`, `start_time`, `duration`, etc.). `PastMeeting` has required fields like `occurrence_id`, `platform_meeting_id`, and `sessions` that are absent in that response. TypeScript will assume they exist, masking potential runtime `undefined` accesses.

The template guards (`@if (pastMeetingFullAccess())`) prevent immediate breakage, but the type contract is wrong and could mislead future contributors. A union type or `Pick<PastMeeting, ...>` for the reduced case would make the contract explicit.

packages/shared/src/interfaces/meeting.interface.ts
```diff
@@ -1101,47 +1101,15 @@ export interface UrlMetadataResponse {
...
+  project: { name: string; slug: string; logo_url: string; uid: string; parent_uid: string };
```
dealako: [nit] **Inline project type duplicates a shape concept removed by this PR**

`{ name: string; slug: string; logo_url: string; uid: string; parent_uid: string }` appears here and again verbatim in the controller's three `res.json()` builders. The PR deleted the `PublicMeetingProject` named type — consider keeping it as a `Pick<Project, ...>` alias to avoid the shape drifting across multiple sites.

dealako (changes requested):
@jordane — thanks for the clear revert description and the reasoning on the capability-URL model. The logic of the change is sound and the surgical removal of `restrictedView` is clean throughout. One security verification is needed before this lands.

**Overall impression:** This is a well-scoped revert that correctly restores the anonymous guest-join flow for invited external participants. The core security model — password-in-URL as capability token, validated via constant-time comparison on the server, rate-limited — is solid. The main code concern is that the PR removes the union types that previously made the type system aware of the "redacted vs full" response distinction, and the replacements don't fully model the actual wire format.

---

🔴 **Blocking: 1 issue** — Potential Zoom `host_key` exposure to unauthenticated guests (server returns full `Meeting` object; `host_key` is present on the interface — needs upstream verification)
🟡 **Minor: 2 issues** — `project` typed as `Project` but API returns a 5-field subset; `PastMeeting` type vs reduced non-full-access wire shape
⚪ **Nit: 1 issue** — Inline project type duplicates a removed named type concept

**On the Copilot comments** (project type + past meeting type): Both are valid — flagged as [minor] inline above. Agreeing with Copilot, not duplicating detail.

---

The blocking item requires a quick verification of the upstream service response (does `/itx/meetings/{id}` return `host_key`/`passcode`?) and, if yes, a one-line field delete before `res.json`. The two type minors are real but non-blocking on their own.

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
```diff
@@ -61,24 +61,6 @@ export class PublicMeetingController {
...
-        });
-        res.json({
-          meeting: { id: meeting.id, visibility: meeting.visibility },
-          project: null,
-        });
-        return;
-      }
-
       // Fetch project and invited status in parallel (both depend only on meeting data)
       const [project, meetingWithInvited] = await Promise.all([
```
dealako, line 65: [blocking] **Verify Zoom host credential fields before merging**

`getMeetingById` at line 143 returns the full `Meeting` object to anonymous callers who provide a valid password:

```typescript
res.json({ meeting, project: { ... } });
```

The `Meeting` interface includes `host_key?: string` (6-digit Zoom host key that grants host privileges) and `passcode?: string`. If the upstream `/itx/meetings/{id}` microservice populates these fields, any anonymous user with the meeting's password URL would receive Zoom host credentials — granting them the ability to claim host status, unmute all participants, end the meeting, or start a cloud recording.

**Action required before merge:** Confirm whether the ITX service response includes `host_key`/`passcode`. If yes, add field filtering for unauthenticated responses:

```typescript
// After the password validation passes, before res.json:
if (!isAuthenticated) {
  delete (meeting as Partial<Meeting>).host_key;
  delete (meeting as Partial<Meeting>).passcode;
}
```

This path is new: PR #700's early-return for private+anonymous callers sent only `{ id, visibility }`. This revert is the first time those callers receive the full `Meeting` object, so the field exposure doesn't exist in the git history to reference.

jordane: Fixed in cc4b17a. Confirmed via the ITX service OpenAPI spec that `host_key` is populated in the response. `passcode` is intentional (it's the Zoom meeting passcode shown to participants), but `host_key` is stripped for unauthenticated callers before `res.json`. The delete runs before both response paths (the public short-circuit and the password-validated path).

apps/lfx-one/src/app/shared/services/meeting.service.ts
jordane: Fixed in cc4b17a. Restored `PublicMeetingProject` as a named type in `packages/shared/src/interfaces/meeting.interface.ts` and updated `getPublicMeeting` to return `Observable<{ meeting: Meeting; project: PublicMeetingProject }>`. The shared package can't reference the app-level `Project` interface directly, so it's defined as a plain object type with the 5 returned fields.

packages/shared/src/interfaces/meeting.interface.ts
jordane: Valid finding — acknowledged. This is a pre-existing type mismatch that predates PR #700. The template's `@if (pastMeetingFullAccess())` guards prevent runtime breakage, but the type contract is indeed looser than ideal. Adding a proper union or reduced-shape type here would bring it in line with the approach taken for `PublicMeetingProject`. Leaving it as a tracked follow-up to keep this PR focused on the revert scope; happy to address in a separate PR if preferred.

packages/shared/src/interfaces/meeting.interface.ts
jordane: Fixed in cc4b17a — `PublicMeetingProject` restored as a named type. `PublicPastMeetingResponse.project` now references it instead of the inline object literal, eliminating the duplication across the controller's `res.json` builders.

docs/architecture/frontend/public-meeting-join.md
```diff
@@ -0,0 +1,186 @@
...
+GET /public/api/meetings/:id
+  → Server: public + non-restricted short-circuit, returns full meeting
+  ↓
+Component renders: title, time, recurrence, agenda
+  ↓
+Signed-out branch: Sign In CTA + OR divider + guest-join form
+  ↓
+Anonymous submits name/email in guest-join form
+  ↓
+POST /public/api/meetings/:id/join  (password from meeting.password in payload)
```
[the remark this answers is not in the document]
jordane: Fixed in ea9e8f1. Corrected to `POST /public/api/meetings/:id/join-url` throughout, and clarified that `password` is a query param while `email`/`name` travel in the request body.

docs/architecture/frontend/public-meeting-join.md
```diff
@@ -0,0 +1,186 @@
...
+GET /public/api/meetings/:id?password=<uuid>
+  → Server: validateMeetingPassword passes, returns full meeting
+  ↓
+Component renders: title, time, recurrence, agenda
+  ↓
+Signed-out branch: Sign In CTA + OR divider + guest-join form
+  ↓
+Anonymous submits guest-join form
+  → POST /public/api/meetings/:id/join with password
+  → Server: validates password, returns Zoom URL
```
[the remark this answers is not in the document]
jordane: Fixed in ea9e8f1.

docs/architecture/frontend/public-meeting-join.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Flow 3 — Anonymous, restricted meeting (email confirmation required)
+
+```
+User opens /meetings/:id?password=<uuid>
+  ↓
+GET returns meeting with restricted: true
+  ↓
+Guest-join form submission → POST /public/api/meetings/:id/join
+  → Server: restrictedMeetingCheck — looks up registrant by email or username
```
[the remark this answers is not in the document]
jordane: Fixed in ea9e8f1.

docs/architecture/frontend/public-meeting-join.md
```diff
@@ -0,0 +1,186 @@
...
+**Location:** `apps/lfx-one/src/server/controllers/public-meeting.controller.ts` → `getMeetingById`
+
+- No authentication required from the caller.
+- **Public + non-restricted:** responds immediately with the full meeting object + slim project (name, slug, logo_url, uid, parent_uid). Password not checked.
+- **Private or restricted:** validates `?password` query param against `meeting.password` via `validatePassword` (`apps/lfx-one/src/server/utils/security.util.ts`). Missing or wrong password → HTTP 400 (`ServiceValidationError`).
+- Response always includes `meeting.password` — the client uses it to build copy-link and returnTo URLs.
+
+### POST `/public/api/meetings/:id/join`
+
+**Location:** `PublicMeetingController.postMeetingLink`
```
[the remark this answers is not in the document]
jordane: Fixed in ea9e8f1. Updated to `POST /public/api/meetings/:id/join-url` and `PublicMeetingController.postMeetingJoinUrl`.

apps/lfx-one/src/app/shared/services/meeting.service.ts
```diff
@@ -212,13 +212,13 @@ export class MeetingService {
     );
   }
 
-  public getPublicMeeting(id: string, password: string | null): Observable<PublicMeetingResponse> {
+  public getPublicMeeting(id: string, password: string | null): Observable<{ meeting: Meeting; project: PublicMeetingProject }> {
```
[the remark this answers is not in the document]
jordane: Fixed in 2d72773. Aligned the `http.get<>` generic to `{ meeting: Meeting; project: PublicMeetingProject }` to match the declared return type.

apps/lfx-one/src/app/shared/services/meeting.service.ts
```diff
@@ -212,13 +212,13 @@ export class MeetingService {
...
     let params = new HttpParams();
     if (password) {
       params = params.set('password', password);
     }
 
-    return this.http.get<PublicMeetingResponse>(`/public/api/meetings/${id}`, { params }).pipe(
+    return this.http.get<{ meeting: Meeting; project: Project }>(`/public/api/meetings/${id}`, { params }).pipe(
```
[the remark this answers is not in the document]
jordane: Fixed in 2d72773.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -889,7 +881,6 @@ export class MeetingJoinComponent implements OnInit {
         toObservable(this.meeting).pipe(
           filter((meeting) => {
             if (!meeting?.id) return false;
-            if (meeting.visibility === 'public' && !meeting.restricted) return true;
             return this.authenticated();
           }),
```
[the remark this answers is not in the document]
jordane: This is intentional — restricting the attachment fetch to authenticated users is part of the scope of this ticket (LFXV2-2200), and is documented in the new spec at `docs/architecture/frontend/public-meeting-join.md` under the "Attachment gating" section.

The filter in `initializeAttachments` only gates the attachment fetch (`GET /api/meetings/:uid/attachments`), not the join flow. Anonymous users still see the full meeting page and can submit the guest-join form — those code paths (`initializeFetchedJoinUrl`, the guest-join form, `postMeetingJoinUrl`) are entirely separate and unaffected. The endpoint itself is also behind `authMiddleware` and would return 401 for anonymous callers regardless, so the filter change simply avoids a guaranteed-to-fail request. The template already shows "Sign in to view materials for this meeting" when `materialFiles()` is empty.

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
```diff
@@ -140,6 +122,12 @@ export class PublicMeetingController {
         meeting.committee_members_count = 0;
       }
 
+      // The Zoom host key grants host privileges to whoever possesses it.
+      // Strip it from the response for unauthenticated callers.
+      if (!isAuthenticated) {
+        delete (meeting as Partial<Meeting>).host_key;
+      }
```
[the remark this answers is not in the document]
jordane: Good catch — this requires more than a UI-layer fix. The meeting data here comes via an M2M-elevated call to the ITX service, but the authenticated `/api/meetings/:id` endpoint already returns `host_key` to any authenticated user using their own bearer token, with no organizer gate. Stripping it here for non-organizers would create an inconsistency without adding security value — they'd just retrieve it through the other endpoint.

The proper fix is upstream: split `host_key` (and similar host-only credentials) out of the core `Meeting` response shape into a separate sub-object with its own permission model, gated at the meeting service layer. Tracked in **LFXV2-2358**.

packages/shared/src/interfaces/meeting.interface.ts
```diff
@@ -1101,47 +1101,15 @@ export interface UrlMetadataResponse {
...
   full_access: boolean;
 }
```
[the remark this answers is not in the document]
jordane: Tracked in **LFXV2-2359** — add a `ReducedPastMeeting` Pick type and widen `PublicPastMeetingResponse.meeting` to a union, matching the pattern used for `PublicMeetingProject` in this PR. Keeping out of scope here to maintain focus on the revert.

packages/shared/src/interfaces/meeting.interface.ts
jordane: Tracked in **LFXV2-2359**.

MRashad26 (commented):
Automated review + `code-standards-enforcer` audit (report-only). 5 non-merge commits by Jordan Evans — all signed off (`Signed-off-by: Jordan Evans <jevans@linuxfoundation.org>`) ✅. No Claude or Cursor co-author. ✅

No credentials or tokens introduced. The `host_key` stripping (`delete (meeting as Partial<Meeting>).host_key`) is a net security improvement — the Zoom 6-digit host key grants host privileges (mute, remove participants, end the call) and must not reach unauthenticated callers. Before this PR, anonymous viewers received the full meeting object including `host_key` via the password-gated public endpoint. ✅

## Overview
Reverts PR #700 (LFXV2-1774). The original "strict private gate" blocked anonymous viewers from seeing private meetings at all and from posting to `/join-url`. The bug: invited external participants without LFX accounts couldn't join via the capability URL (invite link with password). The fix is the revert: possession of a valid `?password` is the authorization signal — the password is only delivered via the Zoom invite, so holding it confirms the viewer was either invited or had the link forwarded by an attendee.

What changes beyond the pure revert:
1. `host_key` stripped from `getPublicMeetingById` for unauthenticated callers (new security fix).
2. Attachment fetches now unconditionally require `authenticated()` — previously public non-restricted meetings allowed anonymous attachment fetches.
3. Architecture doc added (`docs/architecture/frontend/public-meeting-join.md`) documenting the access model, all join flows, component signal gates, and server contract.

## Correctness — verified ✅

**Server-side anonymous join flow (POST `/public/api/meetings/:id/join-url`):**
- Removes the `AuthorizationError` thrown for `visibility === PRIVATE && !isAuthenticated`.
- Flow after removal: `validateMeetingPassword` (password required regardless of visibility) → `isWithinJoinWindow` check → for restricted meetings: `restrictedMeetingCheck` (verifies email against registrant list). All validations remain in place; the only change is removing the blanket rejection of anonymous callers on private meetings. ✅

**GET `/public/api/meetings/:id` anonymous path:**
- Private meeting with correct `?password` → full response (meeting + project). `validateMeetingPassword` runs before the downstream project/invited/organizer lookups, so an invalid password still returns 400 → client redirects to `/meetings/not-found`. ✅
- `host_key` deleted before `res.json()` for all unauthenticated calls. The `as Partial<Meeting>` cast is necessary because `host_key` is required on the `Meeting` type; the runtime delete is effective. ✅

**`PublicPastMeetingResponse.project` now non-nullable:**
- The `project: null` branch was only populated by the removed redacted-response fast-return. After removal, the server always fetches the project before returning, so non-nullable `project: PublicMeetingProject` is now accurate. ✅
- Client-side `map((res) => ({ ...res.meeting, project: res.project }))` cast simplified accordingly. ✅

**`initializeFetchedJoinUrl` filter change:**
- Removes `if (meeting.visibility === 'public' && !meeting.restricted) return true` — this line allowed the automatic join-URL pre-fetch to fire for anonymous viewers on public non-restricted meetings.
- After removal, the auto-fetch only fires for `authenticated()` users. Anonymous users always go through the guest form (POST with name + email), which is the correct path (anonymous callers have no Zoom account linked; the pre-fetch path assumes a user identity).

**Attachment fetch restriction:**
- `initializeAttachments` now strictly requires `authenticated()` — the old `visibility === 'public' && !restricted` exception is gone. Anonymous viewers never trigger the authenticated attachment endpoint. The template already shows "Sign in to view materials for this meeting" when `materialFiles()` is empty and the user is unauthenticated. ✅

**`RedactedMeeting`, `RedactedPastMeeting`, `PublicMeetingResponse` types deleted:**
- These were only needed for the redacted response path, which no longer exists. Deletion removes dead types from the shared interface package. ✅

## Code Standards Audit (`~/LFX/code-enforcer-agent.md`)

**🔴 Critical:** none.

**🟡 Warning:** none. PR body has `## Summary` — NOT `## Test plan`. ✅

**🔵 Info:**
- `host_key` is stripped only in `getPublicMeetingById`. The `getPublicPastMeetingById` response is returned via `PastMeeting`, which does not carry a `host_key` (past Zoom sessions are terminated). No gap. Worth adding a note to the `Meeting.host_key` field in `meeting.interface.ts` (e.g., "must be stripped from unauthenticated API responses") so future endpoint authors don't inadvertently forward it.
- The architecture doc (`public-meeting-join.md`) is thorough and accurate. One small factual discrepancy: the Component Structure table shows `password` as `WritableSignal<string|null>` but in the component it is likely derived from the `queryParamMap` via `toSignal`/`computed`. Worth a pass to verify the signal types match the actual implementation.

**✅ Correctly followed:** no function/method calls in template expressions — all are signal calls (`meetingTitle()`, `occurrenceLabel()`, `hasAiCompanion()`, etc.) or property accesses on signal values; no `effect()` calls; `RedactedMeeting`/`RedactedPastMeeting`/`PublicMeetingResponse` correctly deleted from `packages/shared/src/interfaces/meeting.interface.ts`; `PublicMeetingProject` retained in the shared interfaces (correct location); no new files requiring MIT header (the new `.md` is documentation, not source); all commits signed off.

## Test coverage
No automated tests added for the revert. The three new behaviors (host_key stripping, attachment auth gate, anonymous guest-join on private meetings) are all untested. The host_key stripping path (`isAuthenticated === false && meeting has host_key`) is the highest-priority candidate: a regression there would silently re-expose Zoom host credentials to anonymous callers.

## Verdict: **PASS** (no code defects; one 🔵 Info suggestion on interface documentation)

MRashad26 (dismissed):

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
```diff
@@ -253,9 +223,7 @@ export class PublicMeetingController {
         meeting.organizer = isOrganizer;
       }
 
-      // For non-full-access users, return only the fields needed for the basic UI. The
-      // private + anonymous case has already been redacted-and-returned above, so this
-      // branch only handles authenticated non-members on a private/restricted past meeting.
+      // For non-full-access users, return only the fields needed for the basic UI
       const meetingResponse = fullAccess
```
dealako, line 232: **[blocking] `host_key` reaches anonymous callers on this path.**

`checkPastMeetingAccess` (`meeting.helper.ts:87-88`) returns `fullAccess = true` for any caller — including anonymous — on a `PUBLIC && !restricted` past meeting. This branch then returns the raw `meeting` object, and `PastMeeting extends Meeting` so it carries the Zoom `host_key` (host-takeover credential). Unlike `getMeetingById` (which strips it at `:127-129`), there is no strip here.

Pre-existing on `main` (not introduced by this revert), but the anonymous case has no alternative endpoint, so the LFXV2-2358 deferral rationale does not cover it. Mirror the existing one-liner before `res.json`:

```typescript
if (!isAuthenticated) {
  delete (meeting as Partial<Meeting>).host_key;
}
```

dealako (changes requested):
@jordane — thanks for the thorough turnaround on the last round. The host_key strip, the restored `PublicMeetingProject` named type, the `http.get` generic alignment, the markdown-lint fix, and the corrected `/join-url` docs are all in and verified locally (`yarn check-types` passes clean). The capability-URL reasoning and the LFXV2-2358 / LFXV2-2359 follow-up tickets are exactly the kind of explicit trade-off tracking I want to see. 👏

## 👏 Nice work this round

- **host_key strip on `getMeetingById`** — confirmed at `public-meeting.controller.ts:127-129`; it runs unconditionally for `!isAuthenticated` **before** both the public (`:135`) and password-validated (`:149`) response paths. Clean.
- **Password gating intact** — `validateMeetingPassword` (`:144`) still gates private/restricted content via constant-time `crypto.timingSafeEqual`, before the full `Meeting` is returned.
- **`PublicMeetingProject` restored** and the service `http.get<>` generic now matches the declared return type (commit `18d8068a`).
- **Attachment-filter change** — verified the endpoint (`/api/meetings/:id/attachments`) sits behind `authMiddleware`, so gating on `authenticated()` removes a guaranteed-401 request rather than real anonymous functionality. Justification holds.
- **No dead code** — `restrictedView`, `RedactedMeeting`, `RedactedPastMeeting`, `PublicMeetingResponse` fully removed with zero dangling references repo-wide.

## Revision tracking (prior round)

| Prior item | Status |
| --- | --- |
| 🔴 `host_key` to anonymous on `getMeetingById` | ✅ **Resolved** (`cc4b17a`) |
| 🟡 `project` typed as full `Project` | ✅ **Resolved** — `PublicMeetingProject` named type |
| 🟡 `PastMeeting` non-union type lie | ⚠️ **Deferred** → LFXV2-2359 (acceptable, tracked) |
| ⚪ Inline project literal duplication | ⚠️ **Partially addressed** — type restored, but controller still hand-builds the literal 3× (see below) |
| 🟡 (Copilot) `http.get` generic mismatch | ✅ **Resolved** (`18d8068a`) |
| 🟡 (Copilot) host_key to authed non-organizer | ⚠️ **Deferred** → LFXV2-2358 (reasonable rationale) |
| 🔵 (Copilot) markdown MD040 + doc `/join` path | ✅ **Resolved** (`cc4b17a`, `ea9e8f1`) |

## New / still-open issues

### 🔴 Blocking: 1

**`host_key` still reaches anonymous callers via `getPublicPastMeetingById`** — `public-meeting.controller.ts:227` (inline comment below)

The host_key fix landed on `getMeetingById` but **not** its sibling. `checkPastMeetingAccess` (`meeting.helper.ts:87-88`) returns `fullAccess = true` for **any** caller — including anonymous — on a `PUBLIC && !restricted` past meeting. That hits the `fullAccess ? meeting : {…}` branch (`:227`) which returns the **raw `meeting`** object, and `PastMeeting extends Meeting` so it carries `host_key`. There is no `delete meeting.host_key` anywhere in this method.

**Honest framing:** this exposure is **pre-existing on `main`** — it is *not* a regression introduced by this revert (the private+anonymous redaction gate you removed never covered the public-past path). So I'm not asking you to fix a bug you created. But two things make it in-scope here:

1. You deliberately chose an **inline strip** for the anonymous host_key case rather than deferring it — and the identical anonymous exposure sits one method away with the same one-line fix available.
2. The LFXV2-2358 deferral rationale ("an authenticated user can get host_key from `/api/meetings/:id` anyway") **does not apply to anonymous callers** — they have no alternative endpoint, so this strip genuinely closes a host-takeover vector rather than being cosmetic.

Mirror the existing pattern before `res.json` (`:249`):

```typescript
if (!isAuthenticated) {
  delete (meeting as Partial<Meeting>).host_key;
}
```

(The non-full-access reduced shape at `:229-247` already omits `host_key`, so only the `fullAccess === true` branch needs it.) If you'd rather fold it into LFXV2-2358, that's defensible — but please make it an explicit, documented decision in the PR thread rather than a silent gap, since it's the exact credential class the last round was about.

### 🟡 Minor: 3

1. **`passcode` returned to anonymous on the public meeting path** (`:135-138`) — the strip removes only `host_key`; the full object still carries `passcode`. You noted last round that `passcode` is intentional (participant-facing Zoom passcode), which is reasonable — flagging only so the decision is on record alongside the host_key strip two lines away. No change required if intentional.
2. **`PublicMeetingProject` named type isn't actually used by the controller** (`:137, :149, :251`) — the type was restored and imported into `meeting.service.ts`, but the controller still hand-builds `{ name, slug, logo_url, uid, parent_uid }` in all three `res.json` builders. The "fixed via named type" is half-done: the contract exists but isn't enforced at the producer. A small `toPublicMeetingProject(project)` helper would let the compiler catch future field drift. (This is the unresolved tail of last round's nit.)
3. **No test coverage for the reverted behavior** — this PR re-opens an anonymous-access path a prior security PR (#700) closed, with no controller/component spec added. A controller test asserting (a) anonymous + valid password → full payload, (b) anonymous + wrong/missing password → validation error, (c) `host_key` absent for unauthenticated would materially de-risk the revert. Fine to file alongside LFXV2-2359 if not done here.

### ⚪ Nit: 2

1. `res.project as Partial<Project>` casts in `meeting-join.component.ts:562,586` are now redundant — `PublicMeetingProject` is structurally assignable to `Partial<Project>`. Harmless.
2. The new doc's access table (`public-meeting-join.md`) labels past-meeting access as just "Tiered `full_access` gate," which understates that anonymous viewers receive private past-meeting *metadata* (title, schedule) with no password — intended revert parity, but worth a one-line note so the doc doesn't imply past meetings gate like upcoming ones.

## On the bot comments

- **Copilot (suppressed, low-confidence)** flagged exactly this past-meeting path — anonymous metadata retrieval without a password. **Agreeing and elevating it**: the more concrete risk on that path is `host_key`, hence the blocking item above.
- **Copilot (authed non-organizer host_key)** — agree it's real; your LFXV2-2358 deferral is reasonable for the *authenticated* case specifically.

---

## 🔴 Needs changes before approval

One blocking item: close the `host_key` gap on `getPublicPastMeetingById` (one-line strip mirroring `:127-129`), or make an explicit documented decision to fold it into LFXV2-2358. The three minors are non-blocking individually but the `PublicMeetingProject`-in-controller cleanup and a thin controller test would round out the revert nicely. Everything else from the last round is resolved or appropriately tracked — solid progress.

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
jordane: Fixed in 689a648. Added the same strip before `res.json` in `getPublicPastMeetingById` — applies unconditionally for unauthenticated callers regardless of `fullAccess`, mirroring the pattern in `getMeetingById`.

dealako (approved):
@jordane — that's the blocking item closed. Verified `689a6482`: the `host_key` strip is at `public-meeting.controller.ts:226-229`, runs unconditionally for `!isAuthenticated`, sits **before** `meetingResponse` is built and before the single `res.json`, so it covers both the `fullAccess` (raw `meeting`) and reduced branches. `yarn check-types` passes clean. 👏

## Revision tracking — final pass

| Item | Status |
| --- | --- |
| 🔴 `host_key` to anonymous on `getMeetingById` | ✅ Resolved (`cc4b17a`) |
| 🔴 `host_key` to anonymous on `getPublicPastMeetingById` | ✅ **Resolved** (`689a6482`) |
| 🟡 `project` full-type → `PublicMeetingProject` | ✅ Resolved |
| 🟡 `http.get` generic mismatch | ✅ Resolved (`18d8068a`) |
| 🔵 markdown MD040 + doc `/join-url` path | ✅ Resolved (`cc4b17a`, `ea9e8f1`) |
| 🟡 `PastMeeting` non-union type | ⚠️ Tracked → LFXV2-2359 |
| 🟡 host_key to authed non-organizer | ⚠️ Tracked → LFXV2-2358 |

## Remaining open items (non-blocking)

- 🟡 **`PublicMeetingProject` still not used by the controller** — the three `res.json` builders (`:137, :149, :256`) hand-build the `{ name, slug, logo_url, uid, parent_uid }` literal rather than a typed `toPublicMeetingProject(project)` helper. Cosmetic; the contract exists, it's just not compiler-enforced at the producer. Fine to fold into a future cleanup.
- 🟡 **No test coverage** for the reverted anonymous-access path. Recommend a controller spec asserting (a) anonymous + valid password → full payload, (b) anonymous + wrong/missing password → validation error, (c) `host_key` absent for unauthenticated — ideally before/alongside LFXV2-2359.
- ⚪ **Nits**: redundant `as Partial<Project>` casts in `meeting-join.component.ts:562,586`; doc access-table wording for anonymous past-meeting metadata. Take-it-or-leave-it.
- ℹ️ `passcode` to anonymous on the public path — confirmed intentional per your prior note (participant-facing Zoom passcode). On the record, no change expected.

## Issue count

- 🔴 Blocking: **0**
- 🟡 Minor: 2 (controller type helper, tests — both non-blocking, deferrable)
- ⚪ Nit: 2

---

## ✅ Approved with minor comments

Both host_key vectors are closed, password gating is intact, no dead code, type-check green. The remaining minors are genuine but non-blocking and reasonable to handle as follow-ups (LFXV2-2358 / LFXV2-2359 already capture the type-shape work). Nice, disciplined iteration on a security-sensitive revert — shipping this is the right call. 🚀

PR #941, opened 2026-06-15
fix(crowdfunding): exclude canceled subscriptions from recurring donations list

mlehotskylf:
## Summary

After a user cancels a recurring donation, the subscription remained visible in the Recurring Donations list even after page reload — making it appear as though the cancel had no effect.

**Root cause:** \`GET /v1/me/subscriptions\` (CF backend) returns all subscriptions regardless of status. The Self Serve server-side service was forwarding all rows to the frontend without filtering out \`canceled\` entries.

**Fix:** Filter out \`status === 'canceled'\` rows in \`getMyRecurringDonations\` before mapping and returning. Only \`active\`, \`incomplete\`, and \`past_due\` subscriptions represent active recurring commitments.

**Why only \`canceled\`?** The CF backend models four statuses (\`active\`, \`incomplete\`, \`past_due\`, \`canceled\`). Stripe's \`incomplete_expired\` is mapped to \`canceled\` on webhook ingest. \`unpaid\`, \`paused\`, and \`trialing\` are not stored — they either don't apply (no trials) or are handled by Stripe retry logic. See [Stripe subscription statuses](https://docs.stripe.com/billing/subscriptions/overview#subscription-statuses).

## Changes

- \`apps/lfx-one/src/server/services/crowdfunding.service.ts\` — filter \`canceled\` subscriptions before returning in \`getMyRecurringDonations\`


apps/lfx-one/src/server/services/crowdfunding.service.ts
```diff
@@ -268,8 +268,11 @@ export class CrowdfundingService {
       '/v1/me/subscriptions'
     );
 
-    logger.success(req, 'cf_get_my_recurring_donations', startTime, { total: raw.meta.total });
-    return { data: raw.data.map(mapSubscriptionToRecurringDonation), total: raw.meta.total, pageSize: raw.meta.limit, offset: raw.meta.offset };
+    // Exclude canceled subscriptions — the CF backend returns all statuses by default.
+    // Only active, incomplete, and past_due subscriptions represent active recurring commitments.
+    const active = raw.data.filter((s) => s.status !== 'canceled');
+    logger.success(req, 'cf_get_my_recurring_donations', startTime, { total: raw.meta.total, active: active.length });
+    return { data: active.map(mapSubscriptionToRecurringDonation), total: active.length, pageSize: raw.meta.limit, offset: raw.meta.offset };
```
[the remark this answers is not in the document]
mlehotskylf: Not applicable for this use case. A crowdfunding user will have at most a handful of subscriptions — the 20-row default page is not a practical constraint here. The pre-existing code already fetched without pagination and the Angular component that consumes this endpoint only reads `res.data`, not `total` — so `total` accuracy has no effect on the UI. Paging through all pages to compute an accurate filtered total would add unnecessary complexity and latency for a case that cannot occur in practice. Leaving as-is.

emlimlf (dismissed):

apps/lfx-one/src/server/services/crowdfunding.service.ts
```diff
@@ -268,8 +268,11 @@ export class CrowdfundingService {
       '/v1/me/subscriptions'
     );
 
-    logger.success(req, 'cf_get_my_recurring_donations', startTime, { total: raw.meta.total });
-    return { data: raw.data.map(mapSubscriptionToRecurringDonation), total: raw.meta.total, pageSize: raw.meta.limit, offset: raw.meta.offset };
+    // Exclude canceled subscriptions — the CF backend returns all statuses by default.
+    // Only active, incomplete, and past_due subscriptions represent active recurring commitments.
+    const active = raw.data.filter((s) => s.status !== 'canceled');
+    logger.success(req, 'cf_get_my_recurring_donations', startTime, { total: raw.meta.total, active: active.length });
```
MRashad26: 🔴 **Critical — commit `860d2ee68` carries a `Co-Authored-By: Claude Sonnet 4.6` trailer, violating CLAUDE.md.**

The single commit on this PR has the trailer:
```
Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
```
This is an explicit CLAUDE.md violation. Strip the trailer before merging:
```bash
git commit --amend  # remove the Co-Authored-By line from the commit message footer
```
_(Anchoring to the changed line; the violation is in the commit metadata.)_

apps/lfx-one/src/server/services/crowdfunding.service.ts
MRashad26: 🔵 **Info — `total: active.length` diverges from the backend total; flag for when pagination is added.**

Currently `getMyRecurringDonations` fetches all subscriptions in a single call (no `limit`/`offset`), so returning `active.length` as the total is correct — the frontend receives exactly these rows.

If pagination is ever added (passing `limit`/`offset` to `/v1/me/subscriptions`), this logic breaks: filtering client-side after a paginated fetch silently truncates pages, and `active.length` no longer equals the true filtered total. At that point the filter should move server-side (a `status=active,incomplete,past_due` query param to the CF API, if supported) and the `total` should reflect the unfiltered server count or a server-side filtered total.

**No action needed now** — just leaving a note so the pagination follow-up doesn't regress this fix.

MRashad26 (commented):
Automated review + `code-standards-enforcer` audit (report-only). Single commit, signed-off by Michal Lehotsky.

No secrets. Single BFF service change — filters an in-memory array before returning. ✅

## Overview
After canceling a recurring donation, the subscription remained visible in the Recurring Donations list on reload because `GET /v1/me/subscriptions` returns all statuses and the BFF was forwarding all rows. Fix: filter `status === 'canceled'` in memory before mapping and returning. The PR body documents the four CF statuses and why only `canceled` is excluded — good context.

## Correctness — verified ✅
- `raw.data.filter((s) => s.status !== 'canceled')` — `BackendSubscription.status` is `string`; comparison is safe.
- `total: active.length` correctly reflects the filtered count the frontend receives; `raw.meta.total` preserved in the logger for observability.
- Single API call, no extra round-trips.
- `toValidRecurringStatus` mapper already handles unknown status strings gracefully; filtering upstream reduces noise.

## Code Standards Audit

**🔴 Critical:**
- Commit `860d2ee68` carries `Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>` — explicit CLAUDE.md violation. Strip before merge.

**🟡 Warning:**
- PR body contains `## Test plan` with 4 items — CLAUDE.md: *"No test plans in pull requests."*
- PR body contains `🤖 Generated with [Claude Code](https://claude.com/claude-code)` — same AI-attribution convention flagged in prior reviews; remove.

**🔵 Info:**
- `total: active.length` diverges from `raw.meta.total`. If pagination is ever added to this endpoint, the returned `total` must be the unfiltered server total (not the in-memory filtered count) so the frontend can compute page count correctly. See inline.

**✅ Correctly followed:** targeted single-responsibility fix, explanatory comment with business reasoning, signed-off commit, no new `// Generated with` comments in code.

## Verdict: **FAIL** (process) / **PASS** (code)
The filter is the correct fix. Strip the co-author trailer and remove the test plan + attribution from the PR description before merging.

apps/lfx-one/src/server/services/crowdfunding.service.ts
mlehotskylf: Fixed — stripped the Co-Authored-By trailer in commit bcfdf647.

apps/lfx-one/src/server/services/crowdfunding.service.ts
mlehotskylf: Acknowledged — this is correct today since the endpoint fetches all rows in a single call and the Angular component only reads `res.data`. If pagination is ever added this will need revisiting.

mlehotskylf:
@copilot resolve the merge conflicts in this pull request

MRashad26 (approved):

PR #951, opened 2026-06-15
fix(committees): add project slug to create meeting query params (LFXV2-2223)

audigregorie:
# Summary

The committee meetings page now passes the project slug alongside the committee UID when navigating to create a meeting. Without the project parameter, writerGuard could not resolve write access and redirected users back to the lens overview instead of opening the create flow.

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-meetings/committee-meetings.component.html
@@ -95,7 +95,7 @@
             size="small"
             [class]="isCalendarView() ? 'md:ml-auto' : ''"
             [routerLink]="['/meetings', 'create']"
-            [queryParams]="{ committee_uid: committee().uid }"
+            [queryParams]="createMeetingQueryParams()"
             data-testid="committee-meetings-create-btn"></lfx-button>
         }
       </div>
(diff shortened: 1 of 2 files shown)


luismoriguerra (approved):
LGTM

dealako (approved):
@audigregorie — clean, well-scoped fix. 👍

## Overall impression

This corrects a real navigation bug: `writerGuard` on `/meetings/create` resolves write access from the `?project=` slug (falling back to active context), so without the slug the guard couldn't resolve the target project and bounced users to the lens overview. Passing `committee.project_slug` alongside `committee_uid` closes that gap. The implementation follows the file's existing `initX()` computed-signal pattern, the `project_slug` access is correctly guarded as optional (it's `string | undefined` on the `Committee` interface), and the inline comment explains the *why* rather than the *what*. Param name `project` matches what `writerGuard` reads. No correctness, security, performance, or privacy concerns.

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — no `*.spec.ts` exists for this component, so the new branch (slug present vs. absent) is untested. Pre-existing gap, not introduced here; worth a follow-up if this area grows.

## Bot / prior review reconciliation

- **Copilot** reviewed both files and generated no comments — consistent with my findings.
- **@luismoriguerra** approved (LGTM). Agreed.
- No CodeRabbit / Cursor Bot comments present.

## Verification

- `Committee.project_slug` confirmed optional (`committee.interface.ts:279`) — the `if (committee.project_slug)` guard is correct.
- `writerGuard` reads `route.queryParamMap.get('project')` first (`writer.guard.ts`), so the supplied slug is authoritative for the navigation target.
- `/meetings/create` applies `[authGuard, writerGuard]` (`meetings.routes.ts:18`) — the param is actually consumed.
- `committee` is `input.required<Committee>()`, so `committee.uid` access is null-safe.

## Decision

✅ **Approved**

PR #952, opened 2026-06-15
feat(committees): add project query param to vote and survey create links (LFXV2-2225)

audigregorie:
# Summary

When creating a vote or survey from a committee page, the create links now include the committee's project slug as a query parameter alongside `committee_uid`. This preserves project context in the URL so the create flow stays scoped to the correct project instead of falling back to a default.

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/components/committee-votes/committee-votes.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-votes/committee-votes.component.ts
@@ -6,6 +6,7 @@ import { toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { ButtonComponent } from '@components/button/button.component';
 import { CardComponent } from '@components/card/card.component';
 import { Committee, Vote } from '@lfx-one/shared/interfaces';
+import { buildCommitteeCreateQueryParams } from '@lfx-one/shared/utils';
 import { VotesTableComponent } from '@app/modules/votes/components/votes-table/votes-table.component';
 import { VoteResultsDrawerComponent } from '@app/modules/votes/components/vote-results-drawer/vote-results-drawer.component';
 import { VoteService } from '@services/vote.service';
@@ -47,14 +48,7 @@ export class CommitteeVotesComponent {
 
   // Private initializer functions
   private initCreateVoteQueryParams(): Signal<Record<string, string>> {
-    return computed(() => {
-      const committee = this.committee();
-      const params: Record<string, string> = { committee_uid: committee.uid };
-      if (committee.project_slug) {
-        params['project'] = committee.project_slug;
-      }
-      return params;
-    });
+    return computed(() => buildCommitteeCreateQueryParams(this.committee()));
   }
 
   private initVotes(): Signal<Vote[]> {
--- a/apps/lfx-one/src/app/modules/committees/components/committee-votes/committee-votes.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-votes/committee-votes.component.ts
@@ -1,7 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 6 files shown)


luismoriguerra (dismissed):
LGTM

apps/lfx-one/src/app/modules/committees/components/committee-votes/committee-votes.component.ts
```diff
@@ -27,6 +27,8 @@ export class CommitteeVotesComponent {
   public committee = input.required<Committee>();
   public canEdit = input<boolean>(false);
 
+  public createVoteQueryParams: Signal<Record<string, string>> = this.initCreateVoteQueryParams();
```
dealako: `[minor]` Placement: this `createVoteQueryParams` field sits between `// Inputs` and `// State`, but it's an `init*`-derived `Signal` just like `votes` — which lives under `// Data`. Moving it under `// Data` next to `votes` keeps the file's own section grouping consistent. (Same applies to the survey component.)

apps/lfx-one/src/app/modules/committees/components/committee-votes/committee-votes.component.ts
```diff
@@ -45,6 +47,17 @@ export class CommitteeVotesComponent {
   }
 
   // Private initializer functions
+  private initCreateVoteQueryParams(): Signal<Record<string, string>> {
+    return computed(() => {
+      const committee = this.committee();
+      const params: Record<string, string> = { committee_uid: committee.uid };
+      if (committee.project_slug) {
+        params['project'] = committee.project_slug;
```
dealako: `[minor]` DRY: `initCreateVoteQueryParams()` here and `initCreateSurveyQueryParams()` in committee-surveys are identical except the name. With only two call sites and ~7 lines each, extraction is defensible-to-skip — but worth a deliberate decision rather than silent duplication. If a third caller appears, hoist to a shared helper (e.g. `buildCommitteeCreateQueryParams(committee)` in `@lfx-one/shared/utils`, where committee/poll utils already live).

`[nit]` Mixed access style in the same block: `params['project']` (bracket) vs `committee.project_slug` / `committee.uid` (dot). `params.project` reads more consistently.

dealako (dismissed):
Hi @audigregorie 👋 — nice, tightly-scoped change.

**Overall impression:** This correctly threads the committee's `project_slug` through to the vote/survey create flow as `?project=<slug>`. I verified the param is genuinely consumed downstream — `writerGuard` reads `?project=` ([writer.guard.ts:34](https://github.com/linuxfoundation/lfx-self-serve/blob/main/apps/lfx-one/src/app/shared/guards/writer.guard.ts#L34)) as the authoritative slug for the write-permission check, and `projectQueryParamGuard` ([project-query-param.guard.ts:19](https://github.com/linuxfoundation/lfx-self-serve/blob/main/apps/lfx-one/src/app/shared/guards/project-query-param.guard.ts#L19)) seeds project context from it — so this meaningfully fixes the default-fallback problem. `computed` is the right primitive for the zoneless/signals setup, and the optional-slug guard (`if (committee.project_slug)`) correctly avoids leaking an empty `project=` param.

I ran a parallel security pass (clean — a project slug is a public identifier already used throughout the app's URLs, no PII/injection concern) and a code-quality pass.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 2 — (1) the new `createVoteQueryParams`/`createSurveyQueryParams` field is placed above the `// State` section rather than under `// Data` with the other `init*`-derived signals; (2) the two `initCreate*QueryParams()` methods are byte-for-byte identical across the two components (DRY).
- ⚪ Nit: 1 — mixed `params['project']` (bracket) vs `committee.project_slug` (dot) access in the same block.

None of these are blocking — all stylistic/consistency. CodeRabbit and Copilot both returned no actionable comments, which matches my read.

**Decision:** ✅ **Approved with minor comments.** Optional polish below; happy for this to merge as-is.

jordane (approved):

PR #960, opened 2026-06-16
fix(routing): correct project context when navigating directly to entity URLs

MRashad26:
## Summary

- Navigating directly to `/project/groups/:id` or `/project/mailing-lists/:id` (without `?project=`) injected the user's default project slug into both the URL and the sidebar, even when the entity belonged to a completely different project
- Root cause: `NavigationService.applyDefaultSelection()` always called `setProject()`/`setFoundation()` with `syncUrl=true`, polluting the URL with a wrong slug; and entity detail pages had no mechanism to correct the context after the entity data loaded

## Changes

- **`project-context.service.ts`** — Added optional `syncUrl` parameter to `setProject()` and `setFoundation()`; defaults to `true` for backward compatibility
- **`navigation.service.ts`** — In `applyDefaultSelection()`, check `'project' in queryParams` (key presence, not truthiness) to skip URL sync when navigating to entity URLs without an explicit `?project=`
- **`entity-project-context.util.ts`** (new) — Shared utility `syncEntityProjectContext()` that subscribes to an entity signal and updates the project/foundation context once data loads; mirrors `projectQueryParamGuard`'s lens decision (`/foundation/*` → `setFoundation`, all others → `setProject`)
- **`entity-project-context.interface.ts`** (new in shared package) — `EntityWithProject` interface consumed by the utility
- **`committee-view.component.ts`** — Uses `syncEntityProjectContext()` replacing an inline subscription + private method
- **`mailing-list-view.component.ts`** — Same replacement

## Test plan

- [ ] Open a new browser tab and paste a group URL belonging to a different project (e.g. TLF group on an account defaulting to `agstack`) — sidebar should show TLF, not agstack; `?project=` should NOT appear in the URL
- [ ] Same for a mailing list URL belonging to a different project
- [ ] Navigate normally via the sidebar — `?project=` sync still works (not broken)
- [ ] Navigating to a URL that already has `?project=someSlug` still sets context from the query param via `projectQueryParamGuard`
- [ ] `yarn build` passes

### DIFF
--- a/apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
+++ b/apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
@@ -3,7 +3,7 @@
 
 import { NgClass, NgTemplateOutlet } from '@angular/common';
 import { Component, computed, inject, input, model, Signal, signal } from '@angular/core';
-import { RouterModule } from '@angular/router';
+import { Router, RouterModule } from '@angular/router';
 import { AvatarComponent } from '@components/avatar/avatar.component';
 import { BadgeComponent } from '@components/badge/badge.component';
 import { OrgSelectorComponent } from '@components/org-selector/org-selector.component';
@@ -51,6 +51,7 @@ export class SidebarComponent {
   private readonly personaService = inject(PersonaService);
   private readonly lensService = inject(LensService);
   private readonly navigationService = inject(NavigationService);
+  private readonly router = inject(Router);
   private readonly userService = inject(UserService);
   private readonly orgRoleGrantsService = inject(OrgRoleGrantsService);
   private readonly accountContextService = inject(AccountContextService);
@@ -151,6 +152,18 @@ export class SidebarComponent {
     } else {
       this.projectContextService.setProject(context);
       this.lensService.setLens('project');
+    }
+    // When on an entity detail page (e.g. /project/mailing-lists/:id or /project/groups/:id),
+    // switching project context should land on the new project's dashboard — staying on the
+    // current entity would show data from the wrong project until the user navigates away.
+    this.redirectFromEntityPageIfNeeded();
+  }
+
+  private redirectFromEntityPageIfNeeded(): void {
+    const segments = this.router.url.split('?')[0].split('/').filter(Boolean);
+    // Entity detail pages match /project/<section>/<id> (exactly 3 segments)
+    if (segments.length === 3 && segments[0] === 'project') {
+      this.router.navigate(['/project', 'overview']);
     }
   }
 
--- a/apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
+++ b/apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
@@ -156,14 +156,18 @@ export class SidebarComponent {
     // When on an entity detail page (e.g. /project/mailing-lists/:id or /project/groups/:id),
     // switching project context should land on the new project's dashboard — staying on the
     // current entity would show data from the wrong project until the user navigates away.
-    this.redirectFromEntityPageIfNeeded();
+    this.redirectFromEntityPageIfNeeded(context.slug);
   }
 
-  private redirectFromEntityPageIfNeeded(): void {
+  private redirectFromEntityPageIfNeeded(projectSlug: string): void {
     const segments = this.router.url.split('?')[0].split('/').filter(Boolean);
-    // Entity detail pages match /project/<section>/<id> (exactly 3 segments)
-    if (segments.length === 3 && segments[0] === 'project') {
-      this.router.navigate(['/project', 'overview']);
+    // Entity detail pages match /<lens>/<section>/<id> (exactly 3 segments, project or foundation prefix)
+    if (segments.length === 3 && (segments[0] === 'project' || segments[0] === 'foundation')) {
+      // Use activeLens() (updated synchronously by setLens() above) for the destination prefix.
+      // Pass the slug explicitly — router.url lags behind location.replaceState so
+      // queryParamsHandling:'preserve' would carry stale params.
+      const lensPrefix = this.activeLens() === 'foundation' ? 'foundation' : 'project';
+      this.router.navigate([`/${lensPrefix}`, 'overview'], { queryParams: { project: projectSlug } });
     }
   }
 
--- a/apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
+++ b/apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
@@ -13,10 +13,10 @@ import { ProjectContextService } from '../services/project-context.service';
  * Syncs the active project/foundation context to the owning project of the given
  * entity whenever its data loads or changes. Call once from the component constructor.
  *
- * Mirrors projectQueryParamGuard's lens decision: /foundation/* routes set the
- * foundation context; all other routes (project lens, top-level) set the project
- * context. This prevents the navigation service's default selection from leaving an
- * unrelated project slug active when navigating directly to an entity URL.
+ * Lens decision is URL-prefix based: /foundation/* routes set the foundation context;
+ * all other routes (project lens, top-level) set the project context. This prevents
+ * the navigation service's default selection from leaving an unrelated project slug
+ * active when navigating directly to an entity URL.
  */
 export function syncEntityProjectContext<T extends EntityWithProject>(
   entitySignal: Signal<T | null>,
--- a/apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
+++ b/apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
@@ -4,18 +4,10 @@
 import { DestroyRef, Signal } from '@angular/core';
 import { takeUntilDestroyed, toObservable } from '@angular/core/rxjs-interop';
 import { Router } from '@angular/router';
-import { ProjectContext } from '@lfx-one/shared/interfaces';
+import { EntityWithProject, ProjectContext } from '@lfx-one/shared/interfaces';
 import { distinctUntilChanged, filter } from 'rxjs';
 
 import { ProjectContextService } from '../services/project-context.service';
-
-interface EntityWithProject {
-  uid: string;
-  project_uid: string;
-  project_slug?: string | null;
-  project_name?: string | null;
-  foundation_name?: string | null;
-}
 
 /**
  * Syncs the active project/foundation context to the owning project of the given
--- a/apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
+++ b/apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
@@ -27,7 +27,7 @@ export function syncEntityProjectContext<T extends EntityWithProject>(
   toObservable(entitySignal)
     .pipe(
       filter((entity): entity is T & { project_slug: string } => !!entity?.project_uid && !!entity?.project_slug),
-      distinctUntilChanged((a, b) => a.uid === b.uid),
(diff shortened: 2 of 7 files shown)


apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
```diff
@@ -0,0 +1,45 @@
...
+      const context: ProjectContext = {
+        uid: entity.project_uid,
+        name: entity.project_name || entity.foundation_name || entity.project_slug,
+        slug: entity.project_slug,
+      };
+      if (router.url.startsWith('/foundation/')) {
+        projectContextService.setFoundation(context);
+      } else {
+        projectContextService.setProject(context);
+      }
```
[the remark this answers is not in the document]
MRashad26: Good catch — this was a real bug. Fixed in commit 5973383f3: now reads `'project' in router.parseUrl(router.url).queryParams` before calling `setProject()`/`setFoundation()`, and passes `syncUrl=false` when the key is absent. Mirrors the same guard added to `NavigationService.applyDefaultSelection()` for the same reason.

apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
```diff
@@ -0,0 +1,45 @@
...
+  entitySignal: Signal<T | null>,
+  projectContextService: ProjectContextService,
+  router: Router,
+  destroyRef: DestroyRef
+): void {
+  toObservable(entitySignal)
+    .pipe(
+      filter((entity): entity is T & { project_slug: string } => !!entity?.project_uid && !!entity?.project_slug),
+      distinctUntilChanged((a, b) => a.uid === b.uid),
+      takeUntilDestroyed(destroyRef)
```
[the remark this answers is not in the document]
MRashad26: Fixed in commit 5973383f3: expanded the comparator to `a.uid === b.uid && a.project_uid === b.project_uid && a.project_slug === b.project_slug` so a same-uid entity refresh with updated owning-project fields triggers a re-sync.

MRashad26:
## Review Feedback Addressed

Commit: 5973383f3

### Changes Made
- **entity-project-context.util.ts**: `distinctUntilChanged` now compares `project_uid` and `project_slug` in addition to `uid` — same-uid entity refreshes with updated owning-project fields now trigger a re-sync (per @copilot-pull-request-reviewer)
- **entity-project-context.util.ts**: `setProject()`/`setFoundation()` now receive `syncUrl=false` when `?project=` is absent from the current URL — mirrors the guard in `NavigationService.applyDefaultSelection()` to prevent injecting `?project=` into deep-linked entity URLs on entity data load (per @copilot-pull-request-reviewer)
- **mailing-list-view.component.ts**: Prettier formatting fix (Code Quality CI failure)

### Threads Resolved

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -152,6 +153,18 @@ export class SidebarComponent {
...
+    // switching project context should land on the new project's dashboard — staying on the
+    // current entity would show data from the wrong project until the user navigates away.
+    this.redirectFromEntityPageIfNeeded();
+  }
+
+  private redirectFromEntityPageIfNeeded(): void {
+    const segments = this.router.url.split('?')[0].split('/').filter(Boolean);
+    // Entity detail pages match /project/<section>/<id> (exactly 3 segments)
+    if (segments.length === 3 && segments[0] === 'project') {
+      this.router.navigate(['/project', 'overview']);
```
audigregorie: **Major — redirect ignores the foundation lens and can land on the wrong lens's overview**

The method hardcodes both the source match (`segments[0] === 'project'`) and the destination (`/project/overview`), but `/foundation/groups/:id` and `/foundation/mailing-lists/:id` are real routes (`app.routes.ts:188,194`). Two concrete defects:

1. **Lens/URL mismatch.** From `/project/groups/:id`, picking a foundation row sets `lens = 'foundation'` (line 151) yet the redirect still lands on `/project/overview` — wrong lens for the URL.
2. **Foundation entity pages aren't matched.** From `/foundation/groups/:id`, selecting a project sets `lens = 'project'` but `segments[0] === 'foundation'` fails the guard, so the user stays on the stale foundation entity — exactly the bug this method exists to prevent, left unfixed for the foundation lens.

Match both prefixes and route to the active lens's overview (`lensService` is already injected; `activeLens` is a `Signal<Lens>`):

```suggestion
  private redirectFromEntityPageIfNeeded(): void {
    const segments = this.router.url.split('?')[0].split('/').filter(Boolean);
    // Entity detail pages match /<lens>/<section>/<id> (exactly 3 segments)
    if (segments.length === 3 && (segments[0] === 'project' || segments[0] === 'foundation')) {
      this.router.navigate([`/${this.lensService.activeLens()}`, 'overview']);
```

audigregorie: **Minor — redirect drops the `?project=` query param**

`router.navigate([...])` uses Angular's default `queryParamsHandling` (none), so it discards query params. `onItemSelected` just called `setProject(context)` which wrote `?project=slug` via `replaceState`, but this navigate lands on a paramless `/project/overview`. A refresh or shared link then falls back to the persona-default project via `applyDefaultSelection` instead of the project the user just selected — inconsistent with normal sidebar selection, which keeps the param. Consider passing the selected slug through, e.g. `{ queryParams: { project: context.slug } }`.

apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
```diff
@@ -0,0 +1,49 @@
...
+      const context: ProjectContext = {
+        uid: entity.project_uid,
+        name: entity.project_name || entity.foundation_name || entity.project_slug,
+        slug: entity.project_slug,
+      };
+      // Only write ?project= to the URL if it was already present — mirrors the same
+      // guard in NavigationService.applyDefaultSelection() to prevent injecting a wrong
+      // project slug into entity-specific deep-link URLs (e.g. /project/groups/:id).
+      const syncUrl = 'project' in router.parseUrl(router.url).queryParams;
+      if (router.url.startsWith('/foundation/')) {
```
audigregorie, line 43: **Minor — lens decision diverges from the guard it claims to mirror**

The doc comment says this "Mirrors `projectQueryParamGuard`'s lens decision," but the guard keys off `route.data['lens'] === 'foundation'` (`project-query-param.guard.ts:35`) while this keys off `router.url.startsWith('/foundation/')`. They're functionally equivalent for the current route table, but the two mechanisms can drift if route prefixes and `data.lens` ever disagree. Worth aligning the comment to the actual approach (URL-prefix based) or switching to the same `route.data` signal the guard uses.

audigregorie (commented):
## Code Review Summary

Focused, well-reasoned fix for the deep-link project-context bug. The `syncUrl` plumbing, the `'project' in queryParams` key-presence check, the `applyDefaultSelection` early-return, and the extracted `syncEntityProjectContext` util are all sound and clearly commented. The new interface correctly lives in `@lfx-one/shared`; the Angular-dependent util correctly stays in the app.

One Major in the sidebar redirect (doesn't cover the foundation lens), plus two minors. **Rating: 4/5.**

### Meta — outside the diff

- No unit tests accompany the new shared util (`entity-project-context.util.ts`) or the `applyDefaultSelection` early-return. Given this is subtle URL/context logic with several branches (entity-set context preserved, param presence vs absence, foundation vs project lens), a couple of targeted specs would lock the behavior in. The PR test plan is all-manual and unchecked.

### What's done well

- The `'project' in queryParams` key-presence check (vs truthiness) is the correct distinction and is well-commented.
- Extracting `syncEntityProjectContext` with `distinctUntilChanged` + `takeUntilDestroyed` is clean and avoids duplicate per-component subscriptions.
- The new `applyDefaultSelection` early-return correctly preserves entity-set context even when the owning project isn't in the lens's page items.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
MRashad26: Fixed in commit c0909e3c5. The method now:
- Checks both `/project` and `/foundation` URL prefixes
- Uses `activeLens()` (updated synchronously by `setLens()` before the redirect fires) to pick the correct destination: `/project/overview` or `/foundation/overview`

Thanks for catching the foundation-lens gap — that was a real defect.

MRashad26: Fixed in commit c0909e3c5. Passing the slug explicitly as `{ queryParams: { project: projectSlug } }` rather than `queryParamsHandling: 'preserve'`. The reason 'preserve' wouldn't work here: `setProject(context)` writes `?project=` via `location.replaceState`, which updates the browser URL but not Angular's `router.url` — so 'preserve' would carry the stale pre-switch params. Explicit slug sidesteps that timing dependency.

apps/lfx-one/src/app/shared/utils/entity-project-context.util.ts
MRashad26: Fixed in commit c0909e3c5 — updated the doc comment to accurately describe the URL-prefix based mechanism instead of claiming it mirrors the guard's `route.data['lens']` approach. The guard's approach isn't easily reachable from a standalone utility function (would require injecting `ActivatedRoute`), and URL-prefix detection is equivalent for the current route table, but the misleading comment is gone.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -152,6 +153,18 @@ export class SidebarComponent {
...
+    }
```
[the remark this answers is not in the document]
MRashad26: Fixed in commit c0909e3c5 — passing the project slug explicitly via `{ queryParams: { project: projectSlug } }` instead of `queryParamsHandling: 'preserve'`. Note: 'preserve' wouldn't work here because `setProject()` writes the param through `location.replaceState`, which updates the browser URL but not Angular's `router.url` — so 'preserve' would carry the pre-switch params. Explicit slug is more reliable.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -152,6 +153,18 @@ export class SidebarComponent {
...
   }
```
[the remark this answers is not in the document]
MRashad26: Fixed in commit c0909e3c5 — the redirect now: (1) targets the lens-appropriate overview (`activeLens()` is updated synchronously by `setLens()` before the redirect fires), and (2) carries the newly selected project slug as `{ queryParams: { project: projectSlug } }`. Used explicit slug rather than `queryParamsHandling: 'preserve'` because `router.url` lags behind `location.replaceState`.

MRashad26:
## Review Feedback Addressed

Commit: c0909e3c5

### Changes Made
- **sidebar.component.ts**: Fixed `redirectFromEntityPageIfNeeded` to handle both `/project` and `/foundation` entity URL prefixes, use `activeLens()` for the redirect destination (`/project/overview` or `/foundation/overview`), and pass the selected project slug explicitly as `{ queryParams: { project: projectSlug } }` — `queryParamsHandling:'preserve'` wouldn't work here because `router.url` lags behind `location.replaceState` (per @audigregorie MAJOR + MINOR, @copilot, @coderabbitai)
- **entity-project-context.util.ts**: Updated doc comment to accurately describe URL-prefix based lens detection instead of claiming it mirrors `projectQueryParamGuard`'s `route.data` approach (per @audigregorie MINOR)

### Threads Resolved
5 of 5 unresolved threads resolved.

audigregorie (approved):

PR #964, opened 2026-06-16
fix(meetings): preserve user-selected org name instead of CDP canonical

MRashad26:
## Summary

- When selecting an organization from the autocomplete (e.g. "The Linux Foundation"), `resolveOrg()` calls `/api/organizations/resolve` which returns the CDP canonical name ("The LF") — the component was blindly overwriting the visible field and parent form control with it, causing the confusing mismatch
- Added a `resolveToCdpName` input (default: `true`) to `OrganizationSearchComponent` — when `false`, the CDP resolution still runs (logo display, `onOrganizationResolved` emit) but the field text and parent form value are no longer overwritten
- Set `[resolveToCdpName]="false"` on the `registrant-form` and `public-registration-modal` org-search usages; other usages (committee member-form, work-experience-form) retain the default `true` for backward compatibility

## Root Cause

`POST /api/organizations/resolve` with `{ "name": "The Linux Foundation", "domain": "linuxfoundation.org" }` returns `{ "name": "The LF", ... }`. The component was overwriting the field with `cdpOrg.name` unconditionally in both `resolveOrg()` (on autocomplete select) and `resolveCurrentEntry()` (on form submit).

## Test Plan

- [ ] Search "The Linux Foundation" in meeting guest Organization field → select it → field stays "The Linux Foundation" (not overwritten with "The LF")
- [ ] Org logo still appears after selection (CDP resolve still fires)
- [ ] Manual entry mode still works
- [ ] Committee member-form and work-experience-form still normalize to CDP canonical name (default `true`)
- [ ] Public registration modal org field behaves the same as registrant-form

Fixes [LFXV2-2251](https://linuxfoundation.atlassian.net/browse/LFXV2-2251)

[LFXV2-2251]: https://linuxfoundation.atlassian.net/browse/LFXV2-2251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/src/app/shared/components/organization-search/organization-search.component.ts
```diff
@@ -247,11 +253,12 @@ export class OrganizationSearchComponent {
           this.resolvingOrg.set(false);
           this.onOrganizationResolved.emit(result);
 
-          // Update field text to the resolved name
-          this.organizationForm.get('organizationSearch')?.setValue(cdpOrg.name, { emitEvent: false });
-          const nameControlName = this.nameControl();
-          if (nameControlName && this.form().get(nameControlName)) {
-            this.form().get(nameControlName)?.setValue(cdpOrg.name);
+          if (this.resolveToCdpName()) {
```
audigregorie, line 252: **Minor — duplicated CDP-name-apply block**

This block is now duplicated verbatim in both `resolveOrg()` (here) and `resolveCurrentEntry()` (line ~218), each behind the same `if (this.resolveToCdpName())` guard. Pre-existing duplication, but adding the guard to both copies is a good moment to extract a private helper so the two paths can't drift.

For example: `private applyCdpName(cdpOrg: OrganizationResolveResult | { name: string }): void { ... }` called from both sites. Non-blocking.

audigregorie (commented):
## Code Review Summary

Clean, well-scoped fix. Verified the full call graph: the two changed consumers (`registrant-form`, `public-registration-modal`) set the parent `org_name` via `onOrganizationSelected`, and with `[resolveToCdpName]="false"` the async resolve no longer clobbers it — so the user-selected name is preserved while logo/CDP resolution still fire. The only `resolveCurrentEntry()` / `onOrganizationResolved` consumer (`work-experience-form-dialog`) and committee `member-form` both keep the default `true`, so there's no regression. New input follows repo conventions. No Critical or Major findings — one Minor maintainability nit below.

### Minor — outside the diff

No unit spec exists for `OrganizationSearchComponent`, so this new branching input ships untested. Not a repo-convention violation (the component had no prior spec), but worth noting if you later add coverage — the `resolveToCdpName=false` path that preserves the user value is the regression-prone surface.

### What's done well

- Surgical fix with a clear root-cause writeup in the PR description.
- Backward-compatible default (`true`) keeps canonical-normalization consumers untouched; behavior change is opt-in per usage.
- Excellent JSDoc on the new input explaining the why.

**Rating: 5/5**

apps/lfx-one/src/app/shared/components/organization-search/organization-search.component.ts
MRashad26: Done — extracted a private `applyCdpName(name: string): void` helper. Both `resolveCurrentEntry()` and `resolveOrg()` now delegate to it, so the logic lives in one place.

See commit 8c383ed: extract applyCdpName helper to deduplicate CDP-name-apply block.

MRashad26:
## Review Feedback Addressed

Commit: 8c383ed77

### Changes Made
- **organization-search.component.ts**: extracted private `applyCdpName(name: string): void` helper — both `resolveCurrentEntry()` and `resolveOrg()` now delegate to it, eliminating the duplicated block (per @audigregorie)

### Threads Resolved

audigregorie (approved):

dealako (approved):
👋 Hi @MRashad26 — thanks for the tightly-scoped fix.

## Overall impression

Clean, well-targeted change. The root cause is correctly diagnosed: the async CDP resolve was unconditionally clobbering the user-typed org name in both `resolveOrg()` (autocomplete select) and `resolveCurrentEntry()` (submit). Gating the write-back behind a new `resolveToCdpName` input — defaulted to `true` to preserve every existing consumer — is the right surgical approach, and opting the two meeting-registration forms out via `[resolveToCdpName]="false"` keeps the blast radius minimal. The follow-up extraction of `applyCdpName()` cleanly deduplicates the two identical write-back blocks.

I verified the full call graph (with parallel security + code-quality subagents):

- **Logo + resolved-event contract intact** — `resolvedOrg.set(result)` and `onOrganizationResolved.emit(result)` both fire *before* the new `if (this.resolveToCdpName())` gate in both code paths, so logo display and the emit contract are unaffected when the flag is `false`.
- **No downstream breakage** — neither `registrant-form` nor `public-registration-modal` binds `(onOrganizationResolved)`, holds an `org_id` control, or calls `resolveCurrentEntry()` via ViewChild; they submit the raw `org_name`. The PR #779 ID-clearing concern does not apply to these two forms. The only consumer using that flow (`work-experience-form-dialog`) keeps the default `true` and is untouched.
- **Binding is correct** — both templates use the property-binding bracket form `[resolveToCdpName]="false"` (real boolean, not the string `"false"`).

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2
  - No automated coverage for the new `true`/`false` branch — but the repo has **zero** unit specs app-wide, so adding one here would be inconsistent with project norms. Worth a future e2e assertion that a typed/selected org name survives submission on the registrant + public-registration forms, since the `false` path is now load-bearing for two production flows.
  - JSDoc references `/api/organizations/resolve` — worth a quick confirm it matches the actual `OrganizationService.resolveOrganization` endpoint to avoid doc drift. Non-blocking.

## Reconciliation with prior feedback

- @audigregorie's minor — *duplicated CDP-name-apply block* — is **resolved** in commit `8c383ed` via the extracted `applyCdpName(name: string)` helper. Both paths now delegate to one place; confirmed no behavioral drift (the `{ emitEvent: false }` on the internal control vs. emitting parent `setValue` matches prior semantics).
- CodeRabbit's walkthrough flagged no blocking findings; pre-merge checks pass.

## Decision

✅ **Approved.** No blocking issues, no security findings, branch is mergeable, and the only open items are pre-existing-norm nits. Nice work.

PR #972, opened 2026-06-16
fix(meetings): use selection logo instead of CDP logo when resolveToCdpName is false

MRashad26:
## Summary

- Follow-up to #964: the org logo displayed after selecting from the autocomplete was coming from the CDP resolve response (`cdpOrg.logo`), which belongs to the canonical CDP entity (e.g. "The LF") rather than the org the user actually selected ("The Linux Foundation")
- When `resolveToCdpName=false`, the component now uses the suggestion's logo (available immediately from the search results) and falls back to the CDP logo only if the suggestion has none
- `resolveToCdpName=true` behaviour (committee member-form, work-experience-form) is unchanged — those still use the CDP logo

## Root Cause

`OrganizationSuggestion.logo` is returned by the search API and available on selection, but `resolveOrg()` was always overwriting the displayed logo with `cdpOrg.logo` from the `/api/organizations/resolve` response. For `resolveToCdpName=false` usages (meeting guest/registrant forms), this caused a mismatched logo.

## Test Plan

- [ ] Search "The Linux Foundation" in meeting guest Organization field → select it → correct logo shown (matching the search suggestion, not "The LF" logo)
- [ ] If suggestion has no logo, CDP logo is shown as fallback
- [ ] Committee member-form and work-experience-form still show the CDP logo (default `true`)

Fixes [LFXV2-2251](https://linuxfoundation.atlassian.net/browse/LFXV2-2251)

[LFXV2-2251]: https://linuxfoundation.atlassian.net/browse/LFXV2-2251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/src/app/shared/components/organization-search/organization-search.component.ts
```diff
@@ -241,7 +241,9 @@ export class OrganizationSearchComponent {
           const result: OrganizationResolveResult = {
             id: cdpOrg.id,
             name: cdpOrg.name,
-            logo: cdpOrg.logo,
+            // When not resolving to CDP canonical, keep the suggestion's logo so the
+            // displayed logo matches what the user selected rather than the CDP entity's logo.
+            logo: this.resolveToCdpName() ? cdpOrg.logo : logo || cdpOrg.logo,
```
dealako, line 249: ⚪ [nit] `logo || cdpOrg.logo` falls back when the suggestion logo is `''` (empty string), which is actually the right call since the template guards with `@if (resolved.logo)` and an empty string is not a renderable `[src]`. Only the adjacent comment ("keep the suggestion's logo") slightly overstates it — it keeps a *non-empty* suggestion logo. No change required.

Separately (⚪ nit): the parallel `resolveCurrentEntry()` below still uses `cdpOrg.logo` unconditionally and takes no `logo` arg. Not triggered today (its only caller uses the default `resolveToCdpName=true`), but a one-line note about the intentional asymmetry would help a future `=false` submit path from silently reverting the logo.

dealako (dismissed):
Hi @MRashad26 :wave: — thanks for the tight, well-documented follow-up to #964.

## Overall impression
Small, surgical, and correct. The root-cause writeup is spot-on: `resolveOrg()` was unconditionally overwriting the displayed logo with the CDP canonical entity's logo, so `resolveToCdpName=false` flows (meeting guest/registrant) showed a mismatched logo. The one-line conditional `this.resolveToCdpName() ? cdpOrg.logo : (logo || cdpOrg.logo)` fixes it precisely while leaving the `=true` path untouched. I traced the data flow end-to-end and ran parallel security + code-quality passes — the change is sound and safe to merge.

## Findings

🔴 **Blocking: 0**

🟡 **Minor: 0**

⚪ **Nit: 2**
- **Latent divergence with `resolveCurrentEntry()`** — that parallel path (lines 190–230) still uses `cdpOrg.logo` unconditionally and takes no `logo` arg, so it can't honor a suggestion logo. Today this is *not* triggered: the only caller (`work-experience-form-dialog`) runs with the default `resolveToCdpName=true`, and the two `=false` consumers (`public-registration-modal`, `registrant-form`) don't call it. Worth a one-line comment noting the intentional asymmetry so a future `=false` submit path doesn't silently revert the logo.
- **Comment vs. behavior** — the inline comment says "keep the suggestion's logo," but `logo || cdpOrg.logo` falls back when the suggestion logo is an empty string. That's the *better* behavior (the template guards `@if (resolved.logo)` and `''` isn't a renderable `[src]`), so no change needed — just a wording nuance.

## Reconciliation with bots
- **Copilot** posted an overview only (no actionable findings) — consistent with my read.
- **CodeRabbit** rated it Trivial (🎯 1) with all 5 pre-merge checks passing — agreed.

## Security
Security pass came back clean. `[src]` binding runs through Angular's `SecurityContext.URL` sanitization, both ternary branches resolve to backend-supplied URLs from the same org-data provider (search vs. resolve endpoints of the same BFF), so no new trust boundary or injection sink is introduced. No PII, secrets, or authz concerns.

## Note for a follow-up (not this PR)
There's no `organization-search.component.spec.ts`. This is a pre-existing gap, but the component now has meaningful branching (the `resolveToCdpName` gate + empty-string fallback) that would be worth a small spec when convenient.

## Decision
✅ **Approved with minor comments** — ship it. The nits above are optional and can be deferred.

apps/lfx-one/src/app/shared/components/organization-search/organization-search.component.ts
MRashad26: Done — addressed both nits in commit 14986632:

- Updated the `resolveOrg()` comment to say "prefer the suggestion's *non-empty* logo" and note that empty-string suggestion logos fall back to `cdpOrg.logo`
- Added a one-line comment to `resolveCurrentEntry()` noting it intentionally uses `cdpOrg.logo` unconditionally — no suggestion logo arg is available and all current callers use the default `resolveToCdpName=true`

Good catches — the asymmetry note will save someone a head-scratch if a future =false submit path gets wired up.

MRashad26:
## Review Feedback Addressed

Commit: 14986632e

### Changes Made
- **organization-search.component.ts**: clarified `resolveOrg()` comment — empty-string suggestion logos fall back to `cdpOrg.logo`, updated wording to "non-empty" (per @dealako)
- **organization-search.component.ts**: added comment to `resolveCurrentEntry()` noting the intentional asymmetry — uses `cdpOrg.logo` unconditionally since all current callers use `resolveToCdpName=true` (per @dealako)

### Threads Resolved

apps/lfx-one/src/app/shared/components/organization-search/organization-search.component.ts
```diff
@@ -207,6 +207,8 @@ export class OrganizationSearchComponent {
         const result: OrganizationResolveResult = {
           id: cdpOrg.id,
           name: cdpOrg.name,
+          // Intentionally uses cdpOrg.logo unconditionally — resolveCurrentEntry() takes no
+          // suggestion logo arg and all current callers use resolveToCdpName=true (default).
```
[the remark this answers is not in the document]
MRashad26: Fixed — rewrote the comment to explain the structural constraint rather than reference call sites:

```
// Uses cdpOrg.logo unconditionally — this method resolves from form values rather
// than an autocomplete suggestion, so no suggestion logo is available as a fallback.
```

See commit ce79c104: rewrite resolveCurrentEntry() logo comment to avoid caller-coupling.

MRashad26:
## Review Feedback Addressed

Commit: ce79c104c

### Changes Made
- **organization-search.component.ts**: rewrote `resolveCurrentEntry()` logo comment to explain the structural reason (resolves from form values, no suggestion object available) rather than referencing call sites (per @copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/app/shared/components/organization-search/organization-search.component.ts
```diff
@@ -207,6 +207,8 @@ export class OrganizationSearchComponent {
         const result: OrganizationResolveResult = {
           id: cdpOrg.id,
           name: cdpOrg.name,
+          // Uses cdpOrg.logo unconditionally — this method resolves from form values rather
+          // than an autocomplete suggestion, so no suggestion logo is available as a fallback.
```
dealako, line 211: ⚪ [nit] Agreeing with @Copilot here: tying this comment to "all current callers use resolveToCdpName=true" can go stale as usages evolve. Suggest keeping the durable behavioral explanation (it takes no suggestion-logo arg, so it always falls back to the CDP logo) and softening the call-site half — e.g. "no caller currently passes a suggestion logo." Optional; the asymmetry rationale is still worth documenting.

dealako (approved):
Thanks for the quick turnaround, @MRashad26 :wave: — you nailed both nits from the last round.

## 👏 Nice work
- **`resolveOrg()` comment** is now precise: "prefer the suggestion's *non-empty* logo … empty-string suggestion logos fall back to `cdpOrg.logo` (same as the `=true` path)." That matches `logo || cdpOrg.logo` exactly.
- **`resolveCurrentEntry()`** now carries an explicit comment documenting the intentional `cdpOrg.logo` asymmetry — exactly the clarification I asked for, and the factual claims check out (no logo arg; sole caller `work-experience-form-dialog` defaults to `resolveToCdpName=true`).
- Clean DCO + GPG sign-off, focused `fix(review):` commit, and you resolved the thread. 👍

## Revision tracking (prior round)
- ✅ **Resolved** — *latent divergence with `resolveCurrentEntry()`* → explanatory comment added (commit `14986632`).
- ✅ **Resolved** — *comment wording vs. empty-string fallback* → reworded to "non-empty" with fallback note (commit `14986632`).

## Findings (carried over + new)

🔴 **Blocking: 0**

🟡 **Minor: 0**

⚪ **Nit: 1**
- **Comment durability (agreeing with @Copilot's new note):** the `resolveCurrentEntry()` comment ties behavior to *current* call sites ("all current callers use `resolveToCdpName=true`"), which can go stale as usages evolve. I'd keep the durable behavioral half ("takes no suggestion logo arg, so it always uses the CDP logo") and soften the call-site claim to something like "no caller currently passes a suggestion logo." Purely optional — take it or leave it.

## Reconciliation with bots
- **Copilot** raised the staleness point above — I partially agree and folded it into the nit; the call-site context is genuinely useful for explaining *why* the asymmetry exists, so I'd soften rather than remove it.
- Security re-scan came back clean: the new diff is comment-only, no merge-conflict markers from the `main` merge, no new executable logic, no PII/secrets/injection surface.

## Carry-forward (not this PR)
- Still no `organization-search.component.spec.ts`. Pre-existing gap, out of scope for a comment-only revision — worth a backlog ticket for the `resolveToCdpName=false` fallback branch.

## Decision
✅ **Approved with minor comments** — ship it. The remaining nit is optional polish.

PR #975, opened 2026-06-16
feat(project-context): persist foundation, project, and lens across refresh (LFXV2-2199)

audigregorie:
# Summary

Persists the selected foundation, project, and navigation lens in cookies so the application restores context on page refresh. Cookie-based restore runs on load, and cookies are cleared during impersonation to prevent context leaking across sessions. The active lens is also synced to the route prefix when switching project context, ensuring the URL stays consistent with the stored state.


apps/lfx-one/src/app/shared/services/project-context.service.ts
```diff
@@ -99,6 +106,37 @@ export class ProjectContextService {
...
+      this.cookieService.delete(key, '/');
+      this.cookieRegistry.unregisterCookie(key);
+      return;
+    }
+    this.cookieService.set(key, JSON.stringify(context), {
+      expires: 30,
+      path: '/',
+      sameSite: 'Lax',
+      secure: process.env['NODE_ENV'] === 'production',
+    });
```
MRashad26, line 120: 🔵 **Info — `secure` flag uses a different heuristic than `LensService`**

`ProjectContextService.persistToCookie` uses `process.env['NODE_ENV'] === 'production'` to set the `secure` flag (server/build-time check), while `LensService.persistNavLensToCookie` uses `typeof window !== 'undefined' && window.location.protocol === 'https:'` (browser runtime check). Both achieve the same goal — avoid setting `secure` on a plain-HTTP dev server — but with different failure modes:

- `NODE_ENV === 'production'` evaluates once at cookie write time; a staging deploy over HTTPS with `NODE_ENV=staging` would write non-secure cookies.
- The `window.location.protocol` check is always accurate to the actual connection but returns `false` in SSR/server context (no `window`).

Aligning both to the same pattern (whichever the team has standardised on for other cookie services in this app) would make the two new cookie helpers consistent.

MRashad26 (commented):
Automated review + `code-standards-enforcer` audit (report-only). 3 commits by audigregorie (Audi Young) — all signed off ✅. No Claude co-author ✅. PR body: `## Summary` — no `## Test plan` ✅. No Claude PR body attribution ✅.

No credentials or tokens. Cookie keys are namespace strings. `NAV_LENS_COOKIE_KEY`, `SELECTED_FOUNDATION_COOKIE_KEY`, and `SELECTED_PROJECT_COOKIE_KEY` correctly placed in `packages/shared/src/constants/project-context.constants.ts`. ✅

## Overview
Adds cookie persistence for three pieces of user context that were previously lost on refresh:
- **Selected foundation / project** (`ProjectContextService`): replaces the constructor `delete` calls (legacy cleanup from a prior design) with cookie-restore; `persistToCookie()` saves/deletes on every `setFoundation`/`setProject`/`clearFoundation`/`clearProject` call.
- **Nav lens** (`LensService`): new `navLensSelection` signal tracks the last foundation-or-project lens used; `lastNavLens` exposed as a readonly signal; cookie load/persist methods added.
- **Hybrid persona lens switcher**: clicking the merged 'Projects' button now returns to the last nav lens (could be 'foundation') instead of always defaulting to project.
- **`redirectOnContextSwitch`** (renamed from `redirectFromEntityPageIfNeeded`): now also redirects when the URL's lens prefix diverges from the newly selected lens, keeping the URL consistent with the stored state for hard-refresh restore.
- **Impersonation**: both `startImpersonation` and `stopImpersonation` now clear `NAV_LENS_COOKIE_KEY`, `SELECTED_FOUNDATION_COOKIE_KEY`, and `SELECTED_PROJECT_COOKIE_KEY` so context doesn't leak between impersonated sessions.

## Code-standards audit

**🔴 Critical:** none.

**🟡 Warning:** none.

**🔵 Info:** 1 — see inline comment.

**✅ Correctly followed:**
- `SELECTED_FOUNDATION_COOKIE_KEY`, `SELECTED_PROJECT_COOKIE_KEY`, `NAV_LENS_COOKIE_KEY`, `DEFAULT_NAV_LENS` all in `packages/shared/src/constants/` and re-exported from the index ✅
- `NavLens` type imported from `@lfx-one/shared/interfaces` ✅
- `navLensSelection` exposed only via `.asReadonly()` as `lastNavLens` ✅
- `loadFromCookie` validates `uid`, `slug`, `name` before casting to `ProjectContext` — malformed or stale cookies degrade gracefully ✅
- `loadNavLensFromCookie` validates the stored value against `'foundation' | 'project'` before accepting it ✅
- `cookieRegistry.registerCookie` / `unregisterCookie` correctly mirrored on every set/delete ✅
- 30-day expiry, `path: '/'`, `sameSite: 'Lax'` on all new cookies ✅
- No template changes; no `effect()`; no `CommonModule` ✅
- MIT header on new `project-context.constants.ts` ✅
- All 3 impersonation cookie-clear points (`startImpersonation` + `stopImpersonation`) updated consistently ✅

## Verdict: **PASS**

apps/lfx-one/src/app/shared/services/project-context.service.ts
audigregorie: Good catch — aligned `persistToCookie` to the runtime protocol check (`typeof window !== 'undefined' && window.location.protocol === 'https:'`) in 43227e44, matching `LensService`. The protocol check is also more robust than `NODE_ENV === 'production'` for browser-side cookie writes: there's no `define`/polyfill setting up `process.env` in the browser bundle, so the protocol check measures the actual connection (and correctly handles staging-over-HTTPS) rather than relying on a build-time proxy. The remaining browser services still on the `NODE_ENV` pattern (`account-context`, `hidden-actions`, `cookie-registry`) can be standardized on the protocol check in a follow-up; server-side `NODE_ENV` usage is correct and stays.

apps/lfx-one/src/app/shared/services/project-context.service.ts
```diff
@@ -45,16 +48,17 @@ export class ProjectContextService {
   public readonly selectedFoundationSfid: Signal<string | null> = this.initSelectedFoundationSfid();
 
   public constructor() {
-    // Clean up legacy cookies from the previous cookie-hydrated design.
-    this.cookieService.delete(this.foundationStorageKey, '/');
-    this.cookieService.delete(this.projectStorageKey, '/');
+    // Restore the prior selection so the active context survives a refresh regardless of lens.
+    this.foundationSelection.set(this.loadFromCookie(this.foundationStorageKey));
+    this.projectSelection.set(this.loadFromCookie(this.projectStorageKey));
```
luismoriguerra, line 53: 🟡 **Non-blocking — restored cookie can block self-heal on revoked access (ticket success criterion 5).**

Rehydrating `foundationSelection`/`projectSelection` here is correct for the happy path. But it interacts with `NavigationService.applyDefaultSelection` (`navigation.service.ts:174-176`):

```ts
if (!syncUrl && existing?.uid) {
  return; // preserve entity-deep-link context
}
```

When a refresh lands on an entity deep-link page **without** `?project=` (`syncUrl === false`), a now-non-null restored `existing.uid` makes this short-circuit fire even if that project/foundation is **no longer in the lens items** (access revoked). The app then keeps a context the user can't access instead of falling back to a default. It doesn't throw (so "without error" holds), and the common path (`/project/overview?project=…`) self-heals fine — but the "falls back to a valid default" half of success criterion 5 isn't met on this path.

Suggestion: confirm this edge is acceptable, or only preserve `existing` when its `uid` is present in `page.items`. An e2e covering refresh-after-revocation on a no-`?project=` entity page would lock it down.

apps/lfx-one/src/app/shared/services/lens.service.ts
```diff
@@ -36,13 +40,19 @@ export class LensService {
...
   }
 
   public setLens(lens: Lens): void {
     const allowed = this.getAllowedLensIds();
     if (!allowed.includes(lens)) {
       return;
     }
+    if (lens === 'foundation' || lens === 'project') {
+      this.navLensSelection.set(lens);
+      this.persistNavLensToCookie(lens);
```
luismoriguerra, line 54: 🔵 **Minor — redundant cookie write when the nav-lens is unchanged.** `setLens` re-writes `NAV_LENS_COOKIE_KEY` on every `setLens('foundation'|'project')` call even when `navLensSelection` already equals `lens` — unlike `selectedLens`, which is guarded by the `if (lens === this.selectedLens()) return;` just below. Consider guarding: `if (this.navLensSelection() !== lens) { this.navLensSelection.set(lens); this.persistNavLensToCookie(lens); }`. (Keep it BEFORE the `selectedLens` early-return — nav-lens must still update when `selectedLens` is unchanged.)

luismoriguerra (commented):
## Audit — PR #975 (LFXV2-2199): persist foundation, project & lens across refresh

**Verdict: 🟡 Minor comments** — `merge-after-acknowledging` the stale-cookie self-heal edge below. Audited at head `43227e44f` (4 commits) in an isolated worktree.

### Verification reproduced (worktree, clean install)
| Check | Result |
|---|---|
| `yarn check-types` | ✅ PASS |
| `yarn lint` | ✅ PASS |
| `yarn format:check` | ✅ PASS |
| `yarn build` | ✅ PASS (only the pre-existing `fullcalendar.scss` budget warning) |

No unit-test findings — this repo's test deliverable is Playwright e2e (CI `yarn e2e`).

### Why / How
Refresh previously dropped users onto whatever project the nav API returned first. This PR persists the per-lens foundation/project selection (and a hybrid-persona `nav-lens`) in cookies, keeps the URL `?project=` + lens-prefix in sync, and clears the new cookies on impersonation start/stop. Cookie writes mirror the existing `lens.service` pattern (`sameSite:'Lax'`, `path:'/'`, 30-day expiry, registry-registered for bulk clear); `loadFromCookie` fails closed on malformed/legacy data.

Maps cleanly to ticket success criteria 1–4 and 6. No secrets introduced; cookies hold only project identifiers the user already sees.

### Already resolved at head ✅
- Copilot / CodeRabbit "clear `NAV_LENS_COOKIE_KEY` on impersonation" comments were against `bc7f1ea`; head `e41643e` clears it in **both** `startImpersonation` and `stopImpersonation`.
- The `secure`-flag heuristic thread is already discussed and deliberately aligned to the runtime protocol check in `43227e4`. (The SSR-writes-non-Secure tradeoff Copilot raised remains a conscious choice — non-blocking, non-sensitive data, overwritten on the next client-side write.)

### Gaps (non-blocking)
1. **Self-heal edge vs. success criterion 5** — see inline on `project-context.service.ts`. On refresh of an entity deep-link page **without** `?project=`, the restored cookie makes `applyDefaultSelection` (`navigation.service.ts:174-176`) treat a possibly-revoked selection as authoritative and return early instead of falling back to a default. Doesn't throw, but the "falls back to a valid default" half of criterion 5 isn't met on that narrow path. Worth an e2e assertion + an explicit author decision.
2. **Redundant nav-lens cookie write** — see inline on `lens.service.ts`. Minor.
3. **CodeRabbit's early-return note** (`sidebar.component.ts:162`) is **pre-existing behavior preserved by the rename, not a regression**, and is plausibly intentional — confirm-or-dismiss, not a blocker.

### Business rules / ADRs
Project/foundation selection persists per browser for 30 days, per-lens, cleared on impersonation. No ADR added/changed/violated — extends the established client-side lens-cookie persistence pattern; no server/data-layer impact.

_Report-only audit; nothing was pushed. The two inline notes are the only new, non-duplicated findings._

MRashad26 (approved):

PR #981, opened 2026-06-17
fix(crowdfunding): strip HTML tags from initiative description in card list

emlimlf:
## Summary

- Initiative descriptions in the My Initiatives card list were rendering raw HTML tags as visible text
- Added a `description` computed signal in `InitiativeCardComponent` that pipes the raw value through the existing `stripHtml` utility from `@lfx-one/shared/utils`

### DIFF
--- a/apps/lfx-one/src/app/modules/crowdfunding/my-initiatives/components/initiative-card/initiative-card.component.html
+++ b/apps/lfx-one/src/app/modules/crowdfunding/my-initiatives/components/initiative-card/initiative-card.component.html
@@ -48,7 +48,7 @@
         }
       }
     </div>
-    <p class="text-xs text-gray-500 truncate">{{ initiative().description }}</p>
+    <p class="text-xs text-gray-500 truncate">{{ description() }}</p>
   </div>
 
   <!-- Metrics -->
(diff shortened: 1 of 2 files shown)


MRashad26 (commented):
Automated review + `code-standards-enforcer` audit (report-only). 1 commit by Efren Lim — signed off ✅. No Claude or Cursor co-author. ✅

No credentials or tokens. `stripHtml` is a read-only text sanitiser. ✅

## Overview
Initiative descriptions stored in the CF API can contain raw HTML markup. The card list was interpolating `initiative().description` directly, so tags like `<p>`, `<strong>`, `<br>` appeared as visible text. The fix computes a stripped version via `computed(() => stripHtml(this.initiative().description))` and binds the template to the derived signal.

## Correctness — verified ✅

- `stripHtml` imported from `@lfx-one/shared/utils` (correct shared-utils location). ✅
- `description = computed(() => stripHtml(...))` — reactive: recomputes whenever `initiative` changes. The template binding `{{ description() }}` reads the computed signal. ✅
- Moving the stripping into a `computed()` rather than calling `stripHtml()` directly in the interpolation is the required pattern (function calls not allowed in template expressions). ✅

## Code Standards Audit (`~/LFX/code-enforcer-agent.md`)

**🔴 Critical:** none.

**🟡 Warning:** none. PR body has `## Summary` — no `## Test plan`. ✅

**🔵 Info:**
- `stripHtml(this.initiative().description)` — if `InitiativeBase.description` is typed as `string | null | undefined` and `stripHtml` does not guard for falsy input, the `computed` would throw when an initiative has no description. Worth confirming `stripHtml` handles null/undefined (returns `''` rather than crashing), or that `description` is guaranteed non-nullable on `InitiativeBase`.

**✅ Correctly followed:** `stripHtml` from `@lfx-one/shared/utils` (correct location); transformation in `computed()` not in template; no `effect()`; signed-off commit; no test plan.

## Verdict: **PASS**

MRashad26 (approved):

PR #1000, opened 2026-06-21
feat(surveys): enforce fresh committee writer checks for create/edit flows

MRashad26:
## Summary

- **Stale permission guard on Create Survey CTA**: The committee surveys tab can still show "Create Survey" from stale `canEdit()` state after role downgrade.
- **Click handler with fresh permission check**: Both Create Survey buttons now call `onCreateSurvey()`, which uses `fetchCommittee()` to validate current committee writer access before navigation.
- **Lens-aware deny redirect**: On denial, navigation goes to `/foundation/overview` or `/project/overview` with `_notice=surveys` (and project context when available) so `AppComponent.initAccessDeniedToast()` shows the standard Access Denied toast.
- **Edit draft survey navigation fixed**: The committee surveys table Edit action is now functional and carries committee context query params (`project`, `committee_uid`) to `/surveys/:id/edit`.
- **Client/server permission alignment**: `writerGuard` was updated so committee writer access is accepted for `surveys` (and `votes`), matching the client-side committee-writer check and avoiding pass-client/fail-guard mismatches.

## Changed files

| File | Change |
|------|--------|
| `committee-surveys.component.ts` | Added/updated `onCreateSurvey()` fresh permission check, switched to `fetchCommittee()`, lens-aware deny redirect, reused create query params for edit context |
| `committee-surveys.component.html` | Routed Create Survey CTAs through `(click)="onCreateSurvey()"`; passed edit query params into surveys table |
| `surveys-table.component.html` | Wired Edit navigation directly on `<lfx-button>` via `[routerLink]` + `[queryParams]` and stop propagation |
| `surveys-table.component.ts` | Removed unused `RouterLink` import after button-level routing change |
| `writer.guard.ts` | Extended committee-writer acceptance to include `writeFeature: 'surveys'` (and `votes`) |

## References

- Same stale-permission mitigation pattern as committee votes/meetings follow-ups
- JIRA: LFXV2-2471 (parent epic: LFXV2-2252)

## Test plan

- Log in as a committee **Manager**: Create Survey is visible; click navigates to `/surveys/create` with `committee_uid` and `project`.
- Downgrade role to **Member** without refresh: Create button may remain visible, but click denies and redirects with Access Denied toast.
- Repeat under **foundation lens**: deny redirect goes to `/foundation/overview`.
- Click **Edit** on a draft survey in committee tab: navigates to `/surveys/:id/edit` with `committee_uid` and `project`.
- Direct URL access to create/edit without required writer access: `writerGuard` blocks and shows Access Denied behavior.


MRashad26:
## Edit permission handling added (commit \`867a750e7\`)

Follow-up to the create-survey permission check: wired up the non-functional Edit button and added committee context to its navigation, consistent with the meetings fix (PR #1025).

### Changes
- **surveys-table**: added `RouterLink` import + `editQueryParams` input (`Record<string, string>`, defaults to `{}`) — non-breaking for all non-committee usages. Edit button (previously had no routerLink at all) now wrapped in `<a [routerLink]="/surveys/:id/edit" [queryParams]="editQueryParams()">`.
- **committee-surveys**: added `editSurveyQueryParams` computed signal (same `buildCommitteeCreateQueryParams` as create); passed as `[editQueryParams]` to `lfx-surveys-table`.

### Effect
When a committee manager clicks Edit on a draft survey in the committee tab, the navigation to `/surveys/:id/edit` now carries `?project=<slug>&committee_uid=<uid>` — giving `writerGuard` the project slug it needs to resolve the project check, and `committee_uid` for any future guard extension. Previously the Edit button had no routerLink at all (it rendered but did nothing).

luismoriguerra:
## Audit: PR #1000 — feat(surveys): add permission check on committee Create Survey CTA click

**MRashad26** · `feat/LFXV2-2252-committee-survey-permission-check` → `main` · OPEN · 4 files, +42/−7 · LFXV2-2252

**Scope:** Fresh `committee.writer` check on Create Survey click; lens-aware `_notice=surveys` deny; draft Edit link wired with `editQueryParams`.

**Lanes run:** `yarn lint` · `yarn check-types` · ui-reviewer · bugbot · secrets grep

---

### 1. Why this change
Same stale-`canEdit()` problem as votes/meetings: Create Survey button remains visible after mid-session committee-writer revocation.

### 2. How it was done
`onCreateSurvey()` mirrors #997: fresh `getCommittee()`, lens-aware deny with `_notice=surveys`, `(click)` replaces `[routerLink]`. Draft Edit now navigates via `<a routerLink>` with committee query params (previously a non-navigating button).

### 3. Is it correct
**Yes — merge-ready.** Gates green. Edit click does **not** bubble to `rowClick`/`onRowSelect` — `lfx-table` skips clicks inside `a`/`button` (`table.component.ts` `target.closest('button, a, …')`). CodeRabbit "stop propagation" thread appears **not applicable** given built-in table guard.

**Prior review:** CodeRabbit loading guard → **open** (optional). Duplicate-computed thread → **open** (non-blocking). PR test-plan foundation-lens checkbox → **unchecked** in body.

### 4. Missing / gaps
1. **Non-blocking** — `editSurveyQueryParams` duplicates `createSurveyQueryParams`. Bind `[editQueryParams]="createSurveyQueryParams()"`.
2. **Non-blocking** — JSDoc on `onCreateSurvey()` could collapse to one line per repo convention.

### 5. Q&A this PR resolves
- **Does Edit also open the results drawer?** — No; table host ignores anchor clicks.
- **Why lens-aware deny vs meetings (#992)?** — Surveys adds foundation-lens routing meetings still lacks.
- **Why wire Edit as routerLink?** — Restores draft-survey navigation with committee context params.

### 6. Business rules & ADRs
- **Business rule:** Create Survey CTA re-fetches `committee.writer` on click; denial uses `_notice=surveys`.
- **ADR status:** None — same LFXV2-2252 client-side pattern as #992/#997.

### 7. Open questions & confirmations
- **(confirm)** Same `writerGuard` committee-writer gap as #997 for surveys edit/direct-URL paths (`writeFeature: 'surveys'` checks `project.writer` only).

---

**Verdict:** 🟡 Minor comments  
**Merge readiness:** safe-to-merge

ahmedomosanya (changes requested):
Audit result: 🔴 Required changes

The referenced Jira looks too broad for this PR. `LFXV2-2252` is the parent epic for Org Lens parity; the specific ticket appears to be `LFXV2-2471` (“Allow committee writers to create/manage surveys for their committee”). Please update the PR description to reference `LFXV2-2471` directly, optionally keeping `LFXV2-2252` as the parent epic.

Blocking issue:
- The Survey Edit action still wraps `<lfx-button>` in an `<a [routerLink]>` inside a selectable table row. The latest commit added `stopPropagation()` and switched the permission probe to `fetchCommittee()`, which addresses part of the earlier feedback, but the nested interactive markup remains invalid HTML/accessibility-wise. `lfx-button` already supports `[routerLink]` and `[queryParams]`; please move those bindings directly onto `lfx-button`.

Additional notes:
- The error path still treats every fetch failure as access denied, which can misreport transient failures as permission loss.
- The broader permission contract still needs confirmation: `writerGuard` accepts committee writer access only for meetings today; surveys still require project writer access.

Local checks passed: `yarn lint:check`, `yarn check-types`, and `yarn format:check`.

apps/lfx-one/src/app/modules/surveys/components/surveys-table/surveys-table.component.html
```diff
@@ -179,8 +179,10 @@
               <td>
                 <div class="flex items-center justify-end gap-1">
                   @if (hasPMOAccess() && survey.displayStatus === SurveyStatus.DRAFT) {
-                    <lfx-button label="Edit" severity="primary" size="small" [text]="true" styleClass="" [attr.data-testid]="'surveys-edit-' + rowId">
-                    </lfx-button>
+                    <a [routerLink]="['/surveys', survey.uid, 'edit']" [queryParams]="editQueryParams()">
+                      <lfx-button label="Edit" severity="primary" size="small" [text]="true" [attr.data-testid]="'surveys-edit-' + rowId">
+                      </lfx-button>
+                    </a>
```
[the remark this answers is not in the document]
MRashad26: Done — added `(onClick)="$event.stopPropagation()"` on the `<lfx-button>` directly (commit 443bff830). Also addressed the related Copilot suggestion to avoid the `<a>` wrapper entirely: `ButtonComponent` already has `[routerLink]` and `[queryParams]` inputs, so the wrapper was unnecessary and created nested interactive elements. Both changes are in commit 333d82967.

apps/lfx-one/src/app/modules/surveys/components/surveys-table/surveys-table.component.html
```diff
@@ -179,8 +179,9 @@
               <td>
                 <div class="flex items-center justify-end gap-1">
                   @if (hasPMOAccess() && survey.displayStatus === SurveyStatus.DRAFT) {
-                    <lfx-button label="Edit" severity="primary" size="small" [text]="true" styleClass="" [attr.data-testid]="'surveys-edit-' + rowId">
-                    </lfx-button>
+                    <a [routerLink]="['/surveys', survey.uid, 'edit']" [queryParams]="editQueryParams()">
+                      <lfx-button label="Edit" severity="primary" size="small" [text]="true" [attr.data-testid]="'surveys-edit-' + rowId"> </lfx-button>
+                    </a>
```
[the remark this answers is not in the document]
MRashad26: Fixed in commit 333d82967 — removed the `<a>` wrapper and wired `[routerLink]`, `[queryParams]`, and `(onClick)=stopPropagation` directly on `<lfx-button>`. Also removed the now-unused `RouterLink` import from the component.

MRashad26:
## Review Feedback Addressed

Commits: 443bff830, 333d82967

### Changes Made
- **committee-surveys.component.ts**: switched `getCommittee()` → `fetchCommittee()` in `onCreateSurvey()` to avoid mutating shared committee service state on permission-only checks (per CodeRabbit + Copilot)
- **committee-surveys.component.ts**: aliased `editSurveyQueryParams` to `createSurveyQueryParams` instead of duplicating the computed (per Copilot — already resolved)
- **surveys-table.component.html**: removed `<a>` wrapper around Edit button; bound `[routerLink]`, `[queryParams]`, and `(onClick)=$event.stopPropagation()` directly on `<lfx-button>` — `ButtonComponent` already supports these inputs, and the wrapper created nested interactive elements (invalid HTML)
- **surveys-table.component.ts**: removed now-unused `RouterLink` import

### Threads Resolved

ahmedomosanya (dismissed):

apps/lfx-one/src/app/modules/surveys/components/surveys-table/surveys-table.component.html
```diff
@@ -179,8 +179,9 @@
               <td>
                 <div class="flex items-center justify-end gap-1">
                   @if (hasPMOAccess() && survey.displayStatus === SurveyStatus.DRAFT) {
-                    <lfx-button label="Edit" severity="primary" size="small" [text]="true" styleClass="" [attr.data-testid]="'surveys-edit-' + rowId">
-                    </lfx-button>
+                    <a [routerLink]="['/surveys', survey.uid, 'edit']" [queryParams]="editQueryParams()" (click)="$event.stopPropagation()">
```
dealako: [blocking] The `<a>` wrapper around `<lfx-button>` creates nested interactive elements — invalid per the HTML spec and a screen-reader/focus hazard.

`ButtonComponent` already exposes `[routerLink]`, `[queryParams]`, and `(onClick)` inputs (button.component.ts lines 53–57), so the wrapper is unnecessary.

```html
<!-- Before -->
<a [routerLink]="['/surveys', survey.uid, 'edit']" [queryParams]="editQueryParams()" (click)="$event.stopPropagation()">
  <lfx-button label="Edit" ...> </lfx-button>
</a>

<!-- After -->
<lfx-button
  label="Edit"
  severity="primary"
  size="small"
  [text]="true"
  [routerLink]="['/surveys', survey.uid, 'edit']"
  [queryParams]="editQueryParams()"
  (onClick)="$event.stopPropagation()"
  [attr.data-testid]="'surveys-edit-' + rowId">
</lfx-button>
```

Also remove the now-unused `RouterLink` import from `surveys-table.component.ts` (CodeRabbit had flagged this already; Copilot left two unresolved comments on it — this needs to land).

dealako (changes requested):
Hey @MRashad26 — solid direction here. The lens-aware deny path is a genuine improvement over the meetings pattern, and `fetchCommittee()` (vs `getCommittee()`) is the right choice to avoid mutating shared state. Two items need attention before this is ready to merge.

## 🔴 Blocking: 2 issues
- **Nested `<a>` + `<lfx-button>`** — invalid HTML, accessibility violation (see inline)
- **Client/server permission scope mismatch** — committee-only writer passes the client gate, gets denied by `writerGuard` (see inline)

## 🟡 Minor: 3 issues
- No loading/disabled state during async `fetchCommittee()` call — double-click triggers multiple in-flight requests
- Multi-line JSDoc on `onCreateSurvey()` violates the repo "one short line max" rule
- Redundant `take(1)` — `fetchCommittee()` already pipes `take(1)` internally

## ⚪ Nit: 1 issue
- `RouterLink` import in `surveys-table.component.ts` becomes orphaned once the `<a>` wrapper is removed

## 🔴 Needs changes before approval

---

### Revision history
| Round | Finding | Status |
|-------|---------|--------|
| Prior | CodeRabbit: missing `stopPropagation` on Edit link | ✅ Addressed (via `(click)` on `<a>`) |
| Prior | Copilot: `<a>` wrapping `<lfx-button>` — nested interactive elements | ⚠️ **Not resolved** — commit `333d82967` referenced in your comment does not appear in the branch; `<a>` wrapper is still present |
| Prior | Copilot: `editSurveyQueryParams` duplication | ✅ Resolved (correctly aliased) |

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.ts
```diff
@@ -37,6 +43,32 @@ export class CommitteeSurveysComponent {
...
+    const denyParams: Record<string, string> = { _notice: 'surveys' };
+    if (committee.project_slug) denyParams['project'] = committee.project_slug;
+    const deny = () => void this.router.navigate([overviewPath], { queryParams: denyParams });
+
+    this.committeeService
+      .fetchCommittee(committee.uid)
+      .pipe(take(1))
+      .subscribe({
+        next: (fresh) => {
+          if (fresh?.writer !== true) {
```
dealako, line 68: [blocking] Client gates on `fresh?.writer !== true` (committee writer), but the server-side `writerGuard` for `writeFeature: 'surveys'` only accepts `project.writer` — the committee-writer code path in the guard is gated to `'meetings'` only.

Consequence: a committee Manager who is **not** a project writer will pass this client check, then be denied by `writerGuard` and shown the access-denied toast — a confusing double-redirect UX loop that makes it look like the permission check is broken.

Two options:
1. **Align the client check to `project.writer`** — gate here on `fresh?.project?.writer === true` (or equivalent field on the committee response) to match the server rule.
2. **Update `writerGuard` to accept committee writers for surveys** — if the intent is that committee writers should be able to create surveys, extend the guard accordingly.

Worth confirming which is the intended design with the team before the fix (ref LFXV2-2252).

apps/lfx-one/src/app/modules/surveys/components/surveys-table/surveys-table.component.html
```diff
@@ -179,8 +179,9 @@
...
+                      <lfx-button label="Edit" severity="primary" size="small" [text]="true" [attr.data-testid]="'surveys-edit-' + rowId"> </lfx-button>
+                    </a>
```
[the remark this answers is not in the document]
MRashad26: Fixed — the `<a>` wrapper has been removed. Navigation is now wired directly on `<lfx-button>` via `[routerLink]`, `[queryParams]`, and `(onClick)="$event.stopPropagation()"` to prevent row-select from firing on the Edit click. `RouterLink` import removed from the component since `ButtonComponent` handles it internally.

See commit 333d82967: remove `<a>` wrapper from Edit button, wire navigation directly on `<lfx-button>`.

apps/lfx-one/src/app/modules/surveys/components/surveys-table/surveys-table.component.html
MRashad26: Fixed — `<a>` wrapper removed. Navigation is wired directly on `<lfx-button>` via `[routerLink]`, `[queryParams]`, and `(onClick)="$event.stopPropagation()"` (prevents the PrimeNG `selectionMode="single"` row-select from firing on Edit). `RouterLink` removed from `imports[]` — no longer needed.

See commit 333d82967: remove `<a>` wrapper from Edit button, wire navigation directly on `<lfx-button>`.

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.ts
MRashad26: Good catch — went with Option B. Extended `writerGuard` to accept `committee.writer` for `writeFeature: 'surveys'` and `'votes'` (was `'meetings'` only). The client-side `fresh?.writer !== true` check is now in sync with the server: a committee Manager navigating to create-survey/create-vote will be allowed by both layers, and a non-writer will be denied client-side before reaching the guard.

Same fix applied to both branches:
- PR #1000 (surveys): cb0a2b631 — extend writerGuard + remove redundant `.pipe(take(1))` + `onCreateSurvey → protected`
- PR #997 (votes): 3a2f7f9ba — extend writerGuard

MRashad26:
## Review Feedback Addressed

### Round 2 — Commits: cb0a2b631 (surveys), 333d82967 (prior iteration)

### Changes Made
- **surveys-table.component.html** (333d82967): Removed `<a>` wrapper around `<lfx-button>`; navigation wired directly via `[routerLink]`, `[queryParams]`, and `(onClick)="$event.stopPropagation()"`. `RouterLink` removed from `imports[]`. (per @copilot, @dealako)
- **writer.guard.ts** (cb0a2b631): Extended committee-writer acceptance to `writeFeature: 'surveys'` and `'votes'` (was `'meetings'` only) — aligns the server-side guard with the client-side `fresh?.writer` check. Same change applied to PR #997 (votes branch, commit 3a2f7f9ba). (per @dealako)
- **committee-surveys.component.ts** (cb0a2b631): Removed redundant `.pipe(take(1))` from `fetchCommittee()` call (already has `take(1)` internally); dropped unused `take` import; changed `onCreateSurvey` to `protected`. (per @dealako)

### Threads Resolved

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -84,7 +83,7 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
       // a committee writer is not incorrectly denied solely because the project fetch
       // failed. If no committee check is applicable, deny with feedback.
       if (project === null) {
-        return committeeUid && writeFeature === 'meetings' ? checkCommittee() : of(deny());
+        return committeeUid && (writeFeature === 'meetings' || writeFeature === 'surveys' || writeFeature === 'votes') ? checkCommittee() : of(deny());
```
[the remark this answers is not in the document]
MRashad26: Fixed — extracted a `supportsCommitteeWriter` boolean: `writeFeature != null && ['meetings', 'surveys', 'votes'].includes(writeFeature)`. Both check sites now reference the single variable, so adding or removing a feature from the allowlist requires one change.

See commit on surveys branch: refactor writerGuard to use supportsCommitteeWriter flag.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -93,7 +92,7 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
       if (writeFeature === 'meetings' && project.meetingCoordinator === true) {
         return of(true as const);
       }
-      if (committeeUid && writeFeature === 'meetings') {
+      if (committeeUid && (writeFeature === 'meetings' || writeFeature === 'surveys' || writeFeature === 'votes')) {
```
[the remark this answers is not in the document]
MRashad26: Same fix as above — both occurrences now use `committeeUid && supportsCommitteeWriter`. Good catch on the duplication.

MRashad26:
## Review Feedback Addressed

### Changes Made
- **writer.guard.ts**: Extracted `supportsCommitteeWriter = writeFeature != null && ['meetings', 'surveys', 'votes'].includes(writeFeature)` — both check sites now reference the single flag. Adding or removing a feature from the allowlist requires one change in one place. (per @copilot)

### Threads Resolved

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.html
```diff
@@ -16,8 +17,7 @@
           icon="fa-light fa-chart-simple"
           severity="info"
           size="small"
-          [routerLink]="['/surveys', 'create']"
-          [queryParams]="createSurveyQueryParams()"
+          (click)="onCreateSurvey()"
```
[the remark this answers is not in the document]
MRashad26: Fixed — changed `(click)` to `(onClick)` on both Create Survey buttons (`tableActions` and empty-state).

See commit 1ee8082f3: use (onClick) output on lfx-button in surveys template.

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.html
```diff
@@ -29,8 +29,7 @@
...
```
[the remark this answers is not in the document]
MRashad26: Fixed — same change as above, both buttons updated to `(onClick)` in commit 1ee8082f3.

MRashad26:
## Review Feedback Addressed

Commit: 1ee8082f3fafa1e7f1e1bf9c733d337619cd7175

### Changes Made
- **committee-surveys.component.html**: replaced `(click)` with `(onClick)` on both Create Survey buttons — `<lfx-button>` exposes an `(onClick)` output, not a native `(click)` event (per @copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.ts
```diff
@@ -37,13 +43,33 @@ export class CommitteeSurveysComponent {
...
   }
 
+  protected onCreateSurvey(): void {
+    const committee = this.committee();
+    const overviewPath = this.lensService.activeLens() === 'foundation' ? '/foundation/overview' : '/project/overview';
+    const denyParams: Record<string, string> = { _notice: 'surveys' };
+    if (committee.project_slug) denyParams['project'] = committee.project_slug;
+    const deny = () => void this.router.navigate([overviewPath], { queryParams: denyParams });
+
+    this.committeeService.fetchCommittee(committee.uid).subscribe({
```
dealako: 🟡 `[minor]` No in-flight/loading guard. Rapid double-clicks fire multiple `fetchCommittee()` calls and can queue multiple `router.navigate()`s (this is CodeRabbit's carried-over nitpick). Add a `signal<boolean>` pending flag set true before the call and false in both `next`/`error`, and bind it to disable the button — also gives you a spot for a spinner. Sibling `onScheduleMeeting` has the same gap, so fixing here is the better-pattern move.

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.ts
```diff
@@ -37,13 +43,33 @@ export class CommitteeSurveysComponent {
...
+      next: (fresh) => {
+        if (fresh?.writer !== true) {
+          deny();
+          return;
```
dealako: 🟡 `[minor]` `error: () => deny()` treats any failure — including a transient 5xx/timeout — as a permission denial and shows the "Access Denied" toast. Not a regression (matches the meetings sibling), but a network error and a revoked role are different conditions; worth distinguishing in a follow-up.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -63,10 +63,10 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
   const writeFeature: string | undefined = route.data?.['writeFeature'];
   const deniedUrl = router.createUrlTree([overviewPath], { queryParams: { project: slug, _notice: writeFeature ?? 'access' } });
   const deny = () => deniedUrl;
+  const supportsCommitteeWriter = writeFeature != null && ['meetings', 'surveys', 'votes'].includes(writeFeature);
```
dealako, line 66: 🟡 `[minor]` This widens committee-writer acceptance from `meetings`-only to `meetings|surveys|votes` — an authorization change with no test coverage. Please add a `writer.guard.spec.ts` asserting that `surveys`/`votes` now reach `checkCommittee()` and that unrelated `writeFeature` values still `deny()`. Also worth confirming the upstream OpenFGA ruleset actually grants `committee:uid#writer` survey/vote-create rights — otherwise the form loads then 403s on submit (UX regression, not a vuln).

dealako (changes requested):
## Follow-up review — round 3

Hey @MRashad26 — thanks for the steady iteration on this one. I walked the full history (Copilot, CodeRabbit, and @dealako's two `[blocking]` comments) against the current head `1ee8082f`, re-read all 5 changed files, and ran parallel security + code-quality passes. Every blocking item from the earlier rounds is now resolved. 🎉

### 👏 Nice work this round
- **Both prior blockers cleared.** The nested-interactive-element `<a>` wrapper is gone — `lfx-button` now carries `[routerLink]`/`[queryParams]` directly with `(onClick)="$event.stopPropagation()"`. And the client/server permission mismatch is fixed by extending `writerGuard` to accept committee writers for `surveys`/`votes`, so the client `fresh?.writer` gate and the guard now agree.
- **`fetchCommittee()` was the right call.** Confirmed it's the non-mutating variant (`http.get(...).pipe(take(1))`, no `tap` side-effect) vs the state-mutating `getCommittee()` — exactly what a transient permission check should use.
- **`supportsCommitteeWriter` refactor is clean and correct.** Applied consistently at both decision points, and you correctly left the `meetingCoordinator` branch `meetings`-only.
- **The Edit button was a dead control before this PR** (no route, no handler) — it now navigates with the committee context the guard needs, and `stopPropagation` keeps it from also firing the row's results-drawer click.

### Revision tracking
| Prior feedback | Status |
|---|---|
| @dealako `[blocking]` — `<a>` wrapper nests interactive elements | ✅ Resolved (`1ee8082f`) |
| @dealako `[blocking]` — client gate vs server `writerGuard` mismatch | ✅ Resolved (`cb0a2b63`) |
| CodeRabbit `[major]` — Edit click also triggers row selection | ✅ Resolved (`stopPropagation` on button) |
| Copilot — `create`/`edit` queryParams duplication | ✅ Resolved (aliased) |
| Copilot — collapse `writeFeature === ...` chain | ✅ Resolved (`supportsCommitteeWriter`) |
| Copilot — `(click)` vs `(onClick)` on `lfx-button` | ✅ Resolved (`onClick`) |
| CodeRabbit `[nitpick]` — no loading/double-click guard | ❌ Still open (see below) |

### Open items (all minor — no blockers)
🔴 **Blocking: 0**

🟡 **Minor: 3**
1. **No double-click / in-flight guard** on `onCreateSurvey()` (CodeRabbit's carried-over nitpick). Rapid clicks fire multiple `fetchCommittee` calls and can queue multiple `router.navigate`s. Low practical impact and the sibling `onScheduleMeeting` has the same gap, but a `signal<boolean>` pending flag (disable the button while in flight) is the right fix and lets you show a spinner.
2. **No test for a permission-widening change.** Broadening `writerGuard` from `meetings`-only to `meetings|surveys|votes` is authorization surface — a small `writer.guard.spec.ts` asserting `surveys`/`votes` now reach `checkCommittee()` while unrelated `writeFeature` values still deny would lock in the intended behavior. No spec exists today, so this is net-new, but this is exactly the kind of change worth one.
3. **Deny path conflates network failure with denial.** `error: () => deny()` redirects with `_notice=surveys` ("Access Denied" toast) even on a 5xx/timeout. Matches the sibling pattern so it's not a regression, but worth a follow-up.

⚪ **Nit: 2**
4. `editSurveyQueryParams = createSurveyQueryParams` is safe today (read-only computed, same committee scope) — a one-line comment noting edit intentionally reuses create params would guard against future drift.
5. Lens/deny logic (overview-path + `_notice`) is duplicated between `onCreateSurvey()` (uses `lensService.activeLens()`) and `writerGuard` (uses `route.data['lens']`) — minor drift risk; could centralize later.

### Reconciliation with bots
- **CodeRabbit** (loading-state nitpick) — agree; captured as Minor #1.
- **Copilot / CodeRabbit major** prior threads — all verified resolved and outdated.
- **Security pass** — no privilege escalation. The client guard and `onCreateSurvey` pre-check are UX-only; the real authority is the upstream OpenFGA ruleset reached via the user's bearer token (`microserviceProxy` → `LFX_V2_SERVICE`). Widening the client guard cannot grant access the server wouldn't. No secrets, no PII in logs/URLs (only the non-secret project slug + `_notice` on deny), no open-redirect (all nav targets are hardcoded literals).

### Out of scope (informational)
The guard now covers `votes`, but `committee-votes.component.html` still uses a plain `[routerLink]` create CTA with no fresh-permission pre-check — the runtime check is asymmetric with surveys. Not part of these 5 files; flagging for a follow-up.

### Decision: 🔴 Needs changes before approval
No merge conflicts (`MERGEABLE`), no blockers — this is very close. Holding only because the permission-widening guard change deserves a regression test (#2) and the double-click guard (#1) is still open. Knock out #1 and #2 (and ideally a note for #3) and I'll happily approve. Everything else is take-it-or-leave-it.

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.ts
MRashad26: Done — added `creating = signal(false)` pending flag (commit 37a89380c). Set to `true` before `fetchCommittee()` and cleared in `finalize()`, so both the `next`-deny and `error` paths reset it. Both Create Survey buttons now bind `[loading]="creating()"` — `ButtonComponent` already guards `onClick` when `loading` is true, so rapid double-clicks are blocked without any additional guard condition needed.

apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.ts
MRashad26: Agreed — no code change here. This matches the `onScheduleMeeting` and `onCreateVote` siblings which also use `error: () => deny()`. Distinguishing transient network errors from actual authorization failures is a meaningful UX improvement but out of scope for this PR. Will track as a follow-up.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
MRashad26: Two points addressed separately:

**Test coverage** — there is no existing `writer.guard.spec.ts`, so coverage for the original meetings-only path is also absent. Adding a spec that covers `surveys`/`votes` → `checkCommittee()` and unrelated `writeFeature` → `deny()` is tracked as a follow-up to avoid growing this PR further.

**OpenFGA confirmation** — the `committee.writer` field on the API response is already the single source of truth; if the backend OpenFGA ruleset did not grant `committee:uid#writer` the right to create surveys or votes, `fetchCommittee()` would return `writer: false` and the guard would deny at that point, not silently. The create endpoint itself also enforces its own authorization, so the worst-case outcome for an incorrect OpenFGA config is a 403 on submit rather than an unguarded route. Confirming the upstream ruleset with the backend team is a valid follow-up though — will add to the ticket.

MRashad26:
## Review Feedback Addressed

Commit: 37a89380ce661274d186b48e24085cfa2eddc36b

### Changes Made
- **committee-surveys.component.ts / .html**: added `creating = signal(false)` pending flag to `onCreateSurvey()`; cleared in `finalize()`; both Create Survey buttons now bind `[loading]="creating()"` to block rapid double-clicks (per @dealako)

### No Code Change Needed
- **committee-surveys.component.ts** (`error: () => deny()`): behavior is consistent with the `onScheduleMeeting` and `onCreateVote` siblings; distinguishing transient errors from auth failures is tracked as a follow-up (per @dealako)

### Deferred to Follow-up
- **writer.guard.ts** test coverage: no existing `writer.guard.spec.ts` in the repo; a new spec covering `surveys`/`votes` paths will be added in a separate PR (per @dealako)
- **OpenFGA confirmation**: the `committee.writer` field from the API is the runtime source of truth; worst-case for an incorrect ruleset is a 403 on the create endpoint, not an unguarded route; confirming with the backend team is tracked (per @dealako)

### Threads Resolved

dealako (approved):
## Follow-up review — round 4 ✅

Thanks for the quick turnaround, @MRashad26 — commit `37a89380` cleanly closes out the one actionable item from my last pass, and your replies on the other two are reasonable. Re-read all 5 files at head `37a89380` and ran fresh parallel security + code-quality passes on the delta.

### 👏 Nice work
- **Double-click guard is correctly done.** `creating = signal(false)` set before `fetchCommittee()` and reset in `.pipe(finalize(...))` — so it clears on the `next`-deny, `error`, *and* `complete` paths. Binding `[loading]="creating()"` on both Create buttons gives you defense in depth: PrimeNG renders the native control `disabled`, **and** `ButtonComponent.handleClick` guards `!loading()` before emitting, so a re-click can't re-enter `onCreateSurvey()`. The flag fails **closed** (a stuck `true` blocks creation, never enables it). `finalize` was already imported — no build risk.

### Revision tracking (since round 3)
| Item | Status |
|---|---|
| 🟡 Minor #1 — no double-click / loading guard | ✅ **Resolved** (`37a89380`) |
| 🟡 Minor #2 — no test for the widened `writerGuard` permission | ⚠️ **Acknowledged / deferred** — no `writer.guard.spec.ts` exists today (the original meetings-only path is also uncovered); tracked as a follow-up to avoid growing this PR. Reasonable trade-off. |
| 🟡 Minor #3 — deny path conflates network error with denial | ⚠️ **Acknowledged / deferred** — matches the `onScheduleMeeting` / `onCreateVote` siblings; a follow-up if/when the sibling pattern is revisited. Reasonable trade-off. |
| ⚪ Nits #4–#5 (alias comment, lens-logic duplication) | Take-it-or-leave-it; no action needed. |

### Open item count
🔴 Blocking: 0
🟡 Minor: 2 — both explicitly acknowledged and deferred to a follow-up ticket (test coverage; network-vs-denial distinction)
⚪ Nit: 2 — optional polish

### Reconciliation with bots
- **Copilot** re-reviewed all 5 files on `37a89380` and generated **no new comments**.
- **CodeRabbit**'s only nitpick (the loading-state guard) is the item this commit resolves.
- **Security re-audit** of the delta: no new issues — the loading flag never touches an authorization decision path; server-side OpenFGA via the user bearer token remains the authority.

### Out of scope (still worth a follow-up)
`committee-votes` create CTA still lacks the fresh-permission pre-check even though the guard now covers `votes` — runtime check remains asymmetric with surveys. Not part of this PR's files.

All blocking items from earlier rounds are resolved, the double-click guard is correct, no security findings, no merge conflicts. The two remaining minors are documented trade-offs deferred to a follow-up — please do file that ticket (guard spec + the OpenFGA-ruleset confirmation with the backend team). Nice work getting this across the line. 🚀

MRashad26:
Thanks @dealako — glad the double-click guard landed cleanly. Two follow-ups confirmed tracked:
- `writer.guard.spec.ts` test coverage (surveys + votes paths, plus the existing meetings path) — to be added in a dedicated follow-up PR.
- Network-error vs permission-denial UX distinction in `error: () => deny()` — to be addressed alongside the meetings sibling when that pattern gets revisited.

MRashad26:
Follow-up PR #1042 opened to bring `onScheduleMeeting()` (committee-meetings) to full parity with this pattern — same five improvements: `fetchCommittee()`, lens-aware redirect, `creating` guard, `(onClick)`, `protected` visibility.

MRashad26:
## Review Feedback Addressed

All threads from the previous review iterations are resolved. No unresolved threads remain.

### Summary of changes on this branch
- `committee-surveys.component.ts`: added `creating = signal(false)` pending guard + `finalize()` to prevent double-click and show spinner during permission check
- `committee-surveys.component.html`: fixed `(click)` → `(onClick)` on both Create Survey buttons (native `click` is a no-op on `<lfx-button>`); added `[loading]="creating()"` binding

Thanks for the thorough review, @dealako! 🙏

dealako (dismissed):

MRashad26:
## Review Feedback Addressed — all threads resolved

No new threads. Tagging @dealako — your approval from the last iteration was dismissed by a force push we had to do today to fix unsigned commits (3 commits were accidentally created with no GPG/SSH signature due to a mis-configured `gpg.format=ssh` in global git config). The code is identical to what you approved; only the commit SHAs changed. Would appreciate a re-approval when you have a moment!

dealako (approved):

PR #1001, opened 2026-06-21
feat(akrites): implement stewardship assignment with real member picker

gaspergrom:
## Summary

- Replace placeholder userId text input in `akrites-assign-steward-modal` with a searchable member picker that loads from the LFX Akrites committee endpoint (`GET /api/committees/:id/members`)
- Wire the assign-steward modal into the package drawer (Assign steward / Reassign button for unassigned/open/assessing/escalated/inactive statuses)
- Wire the assign-steward modal into the triage tab footer actions for unassigned and inactive columns
- Update `AkritesAssignStewardRequest` / `CdpStewardSummary` / `AkritesSteward` with required `username` and `displayName` fields to satisfy the CDP assign endpoint contract

## Changes

**Shared package (`@lfx-one/shared`)**
- `AkritesSearchStewardResult` — new interface mapping `CommitteeMember` to an assignable steward shape
- `AkritesAssignStewardRequest` — adds required `username: string` and `displayName: string`
- `CdpStewardSummary` / `AkritesSteward` — adds `username` and `displayName` fields
- `AKRITES_STEWARD_COMMITTEE_UID` — constant pointing to the LFX Akrites working group (`8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6`)

**Backend (`apps/lfx-one/src/server`)**
- `validation.helper.ts` — `parseAssignStewardBody` now validates required `username` and `displayName`
- `akrites.service.ts` — `mapStewards` populates `username` and `name` from the CDP response

**Frontend**
- `akrites.service.ts` — adds `searchStewards()`: fetches committee members and maps them to `AkritesSearchStewardResult[]`
- `akrites-assign-steward-modal` — full rewrite: initials-avatar list, client-side search filter (name/username/org), single-select with blue border + checkmark, role radio buttons, Move to Assessing toggle; all driven by signals
- `akrites-package-drawer` — Assign steward / Reassign button shown for actionable statuses; modal wired to `onAssignStewardConfirm` (open-then-assign chain for packages without a stewardship row)
- `akrites-triage-tab` — Assign steward modal wired to unassigned and inactive column actions

## JIRA

CM-1268


apps/lfx-one/src/app/modules/akrites/components/akrites-assign-steward-modal/akrites-assign-steward-modal.component.html
```diff
@@ -19,6 +19,82 @@
...
+              [class.bg-blue-50]="selectedSteward()?.userId === steward.userId"
+              [class.border-gray-200]="selectedSteward()?.userId !== steward.userId"
+              [class.hover:bg-gray-50]="selectedSteward()?.userId !== steward.userId"
+              [attr.data-testid]="'akrites-assign-steward-' + steward.userId">
+              <!-- Initials avatar -->
+              <span
+                class="inline-flex h-8 w-8 flex-shrink-0 items-center justify-center rounded-full bg-blue-100 text-xs font-semibold text-blue-700"
+                aria-hidden="true">
+                {{ getInitials(steward.displayName) }}
+              </span>
```
MRashad26, line 72: 🔴 **Critical — template method call in interpolation**

`getInitials(steward.displayName)` is a regular component method called in a `{{ }}` interpolation. Per project rule: no method calls in interpolations.

**Fix:** Add `initials: string` to `AkritesSearchStewardResult` in `packages/shared/src/interfaces/akrites.interface.ts`, compute it in the `searchStewards()` mapper:
```typescript
initials: `${m.first_name?.[0] ?? ''}${m.last_name?.[0] ?? ''}`.toUpperCase()
         || (m.username?.[0] ?? 'U').toUpperCase(),
```
Then here use `{{ steward.initials }}` and remove `getInitials()` from the component.

MRashad26 (commented):
## Review — PR #1001 `feat/CM-1268-akrites-stewardship-assignment`

**Verdict: FAIL**

### Overview
Replaces the placeholder userId text input with a real searchable member picker backed by `GET /api/committees/:id/members`, wires the modal into the package drawer and triage tab, and propagates `username`/`displayName` through the full steward assignment contract (frontend → BFF validation → CDP payload → mapStewards). Good signal-first design throughout.

### Secrets check
Clean. `AKRITES_STEWARD_COMMITTEE_UID` (`8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6`) is the public LFX Akrites working-group UID visible in the LFX UI — not a credential.

### Code standards — finding

🔴 **Critical — template method call in interpolation**

`akrites-assign-steward-modal.component.html` calls `getInitials(steward.displayName)` inside a `{{ }}` interpolation in the `@for` steward list. Method calls in interpolations violate the project's "No functions in HTML templates" rule.

**Suggested fix:** Pre-compute initials and store them in the `AkritesSearchStewardResult` interface (all types must live in `@lfx-one/shared/interfaces`). Add the field to the interface in `packages/shared/src/interfaces/akrites.interface.ts`:
```typescript
initials: string;
```
Compute it in the `searchStewards()` mapper in `apps/lfx-one/src/app/shared/services/akrites.service.ts`:
```typescript
initials: `${m.first_name?.[0] ?? ''}${m.last_name?.[0] ?? ''}`.toUpperCase()
         || (m.username?.[0] ?? 'U').toUpperCase(),
```
Then in the template replace `{{ getInitials(steward.displayName) }}` with `{{ steward.initials }}` and remove `getInitials()` from the component.

gaspergrom:
Fixed in commit 044035b4 — added pre-computed `initials` field to `AkritesSearchStewardResult` interface, computed it in the `searchStewards()` mapper, and updated the template to use `{{ steward.initials }}` instead of calling `getInitials()` method. Template interpolations no longer contain method calls per project conventions.

MRashad26 (dismissed):
## Re-review — PR #1001 (commit `044035b4`)

The single Critical finding from the previous review is confirmed resolved:
- ✅ `getInitials(steward.displayName)` removed from template interpolation
- ✅ `initials: string` added to `AkritesSearchStewardResult` in `packages/shared/src/interfaces/akrites.interface.ts`
- ✅ Computed in `searchStewards()` mapper using the exact formula suggested
- ✅ Template now uses `{{ steward.initials }}` (plain property access)

All other signals/computed calls in the template are Angular-idiomatic and correct. No new violations found.

**Verdict: PASS — approving.**

apps/lfx-one/src/app/shared/services/akrites.service.ts
```diff
@@ -86,4 +89,26 @@ export class AkritesService {
   public updateStewardshipStatus(stewardshipId: number, body: AkritesUpdateStatusRequest): Observable<AkritesStewardshipResponse> {
     return this.http.put<AkritesStewardshipResponse>(`/api/akrites/stewardships/${stewardshipId}/status`, body).pipe(take(1));
   }
+
+  public searchStewards(): Observable<AkritesSearchStewardResult[]> {
+    return this.http.get<CommitteeMember[]>(`/api/committees/${AKRITES_STEWARD_COMMITTEE_UID}/members`).pipe(
+      map((members) =>
+        members.map((m) => {
+          const initials = `${m.first_name?.[0] ?? ''}${m.last_name?.[0] ?? ''}`.toUpperCase() || (m.username?.[0] ?? 'U').toUpperCase();
+          return {
```
dealako: `[blocking]` **Empty `username` silently causes a 400 on Assign for any member without a username.**

`CommitteeMember.username` is optional (`member.interface.ts:21`, `username?: string`). A member with no username still passes the picker's search filter — displayName comes from the required `first_name`/`last_name` — and is fully selectable. When selected, this line maps `username: ''` into the assign request body. The server's `parseAssignStewardBody` (added in this PR, `validation.helper.ts:314`) hard-rejects empty username with a 400. The user sees the generic "Could not assign steward" toast with no recovery path on retry.

The `?? ''` fallback is itself a tell that username can be absent.

**Product decision needed — pick one:**
- **(a)** Filter out or disable selection of username-less members in the picker (with an explanation tooltip)
- **(b)** Fall back to `userId` as the username field (closest to how the placeholder Auth0 input worked)
- **(c)** Relax the server validation to allow empty username and let CDP handle it upstream

apps/lfx-one/src/server/helpers/validation.helper.ts
```diff
@@ -306,11 +306,17 @@ export function parseOpenStewardshipBody(req: Request, operation: string): strin
 
 /** Validates the assign-steward body; throws 400 for missing/invalid fields. */
 export function parseAssignStewardBody(req: Request, operation: string): AkritesAssignStewardRequest {
-  const body = (req.body ?? {}) as { userId?: unknown; role?: unknown; moveToAssessing?: unknown };
+  const body = (req.body ?? {}) as { userId?: unknown; username?: unknown; displayName?: unknown; role?: unknown; moveToAssessing?: unknown };
 
   if (typeof body.userId !== 'string' || body.userId.trim() === '') {
     throw ServiceValidationError.forField('userId', 'userId is required', { operation });
   }
+  if (typeof body.username !== 'string' || body.username.trim() === '') {
```
dealako: `[blocking]` **Server-side of the username/400 bug (see comment on `akrites.service.ts:98`).**

This validation is correct in isolation — empty username should be rejected. The problem is that the client (`searchStewards()`) silently maps absent `CommitteeMember.username` to `''` via `?? ''`, so username-less members are selectable but always produce a 400 here. The fix lives on the client side (see the linked comment), but this line is the failure surface worth anchoring.

apps/lfx-one/src/app/shared/services/akrites.service.ts
```diff
@@ -86,4 +89,26 @@ export class AkritesService {
...
+            userId: m.uid,
+            username: m.username ?? '',
```
dealako: `[minor]` **Unused PII shipped to the browser.**

`email`, `committeeName`, and `committeeUid` are mapped into every `AkritesSearchStewardResult` here (and declared in `akrites.interface.ts`), but none of them are rendered in the modal — it shows initials, displayName, `@username`, and organization only. `parseAssignStewardBody` on the server already discards `email`, confirming it isn't needed.

Drop these three fields from the mapping and the interface to minimize PII reaching the client. Zero functional cost.

apps/lfx-one/src/app/shared/services/akrites.service.ts
```diff
@@ -86,4 +89,26 @@ export class AkritesService {
...
+            displayName: `${m.first_name} ${m.last_name}`.trim(),
+            email: m.email,
+            committeeName: m.committee_name,
+            committeeUid: m.committee_uid,
+            organization: m.organization?.name ?? null,
+            status: m.status ?? '',
+            initials,
```
dealako: `[minor]` **`catchError` swallows fetch failures silently.**

A failed `/api/committees/:id/members` call renders "No stewards available" — indistinguishable from a genuinely empty roster, with no retry affordance. Consider pairing `loadingStewards` with an error signal so the modal can show a distinct error state (e.g. "Couldn't load stewards — try again") and expose a retry path.

apps/lfx-one/src/app/modules/akrites/components/akrites-triage-tab/akrites-triage-tab.component.ts
```diff
@@ -63,15 +68,47 @@ export class AkritesTriageTabComponent {
...
     }
   }
 
+  protected onAssignStewardConfirm(body: AkritesAssignStewardRequest): void {
+    const pkg = this.assignTargetPackage();
+    if (!pkg || this.actionLoading()) return;
+    this.actionLoading.set(true);
+
+    const stewardshipId$ =
+      pkg.stewardshipId !== null ? of(pkg.stewardshipId) : this.akritesService.openStewardship(pkg.purl).pipe(map((res) => parseInt(res.stewardship.id, 10)));
```
dealako, line 89: `[nit]` **Open-then-assign idempotency on partial failure.**

If `openStewardship` succeeds but the subsequent `assignSteward` fails, a stewardship row exists upstream while the local VM still holds `stewardshipId: null`. A retry would call `openStewardship` again. This is safe only if that CDP endpoint is idempotent for the same `purl` — worth verifying against the CDP contract, or capturing the returned stewardship id on partial failure so retries go straight to `assignSteward`.

packages/shared/src/constants/akrites.constants.ts
```diff
@@ -160,6 +160,13 @@ export const AKRITES_STATUS_LABELS: Record<AkritesStatus, string> = {
 /** Display order for Akrites statuses in the risk matrix legend and filters. */
 export const AKRITES_STATUS_ORDER: AkritesStatus[] = ['unassigned', 'needs_attention', 'escalated', 'blocked', 'inactive', 'open', 'assessing', 'active'];
 
+/**
+ * Committee UID for the "LFX Akrites" working group — the source of assignable stewards.
+ * Members of this committee are fetched via GET /api/committees/:id/members.
+ * See: https://app.lfx.dev/groups/8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6
+ */
+export const AKRITES_STEWARD_COMMITTEE_UID = '8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6';
```
dealako, line 168: `[nit]` **Hardcoded committee UID across environments.**

Pinning `AKRITES_STEWARD_COMMITTEE_UID` as a compile-time constant means the steward pool committee is the same UID in dev, staging, and prod. If those environments have different committee setups this will silently pull from the wrong pool. Not a secret, but worth noting for maintainability — an environment-keyed config entry or a value driven from the runtime config would be more robust.

dealako (changes requested):
Hey @gaspergrom 👋 — solid step forward on the Akrites assign-steward flow. The searchable member picker is a meaningful UX upgrade over the raw Auth0 input, the signal-based modal architecture is clean, and CodeRabbit's three a11y/correctness findings were all addressed in commit 9625824. That said, there's one correctness bug that makes the assign flow silently fail for any committee member who doesn't have a username set — that needs a deliberate product decision before this can land.

---

🔴 **Blocking: 1** — empty username → 400 (unrecoverable for username-less members)
🟡 **Minor: 4** — unused PII to client, swallowed fetch errors, duplicated drawer logic, no tests
⚪ **Nit: 3** — open-then-assign idempotency, static modal header on reassign, hardcoded committee UID

---

### 🔴 Needs changes before approval

---

### 🟡 Summary-level: Drawer duplication (#4)

The triage-tab refactor in this PR unified the open-then-assign chain into a clean `stewardshipId !== null ? of(id) : openStewardship(...)` → single `switchMap` pattern. The drawer's `onAssignStewardConfirm` (pre-existing, not in this diff) still uses the two-branch approach the triage-tab replaced. Since this PR introduced the better pattern, it's worth noting — applying the same cleanup to the drawer in a follow-up would eliminate the divergence.

### 🟡 Summary-level: Test coverage (#5)

The new modal and assign flows (search filter, selection-clearing, open-then-assign, role/`moveToAssessing`, error path) have no unit or e2e tests. `data-testid`s are already present, so coverage would be cheap to add. Repo policy targets >80% for new code and expects dual-architecture e2e.

### ⚪ Summary-level: Modal header on reassign path (#7)

The drawer button and triage footer both show "Reassign" for `inactive` stewards, but the modal header is always "Assign steward". Low-impact inconsistency — dynamically setting it to "Reassign steward" when a stewardship already exists would complete the UX.

---

### ⚠️ Pre-existing security concern — not gating this PR

The four `/api/akrites/stewardships*` write routes are auth-only — no role/persona gate beyond the LD `akritesEnabledGuard`, which is client-side and trivially bypassed by calling the API directly. This predates PR #1001 (`akrites.route.ts` is **not** in this diff, verified) and is tracked as CM-1245 Gap #4. Calling it out prominently because it's high-priority to fix before Akrites ships widely — but it doesn't gate a UI improvement PR that doesn't touch the route file.

apps/lfx-one/src/app/shared/services/akrites.service.ts
gaspergrom: Fixed in 4256a9d5: username-less members are now filtered out before mapping to `AkritesSearchStewardResult`. This prevents empty usernames from reaching the assign endpoint.

apps/lfx-one/src/server/helpers/validation.helper.ts
gaspergrom: Fixed client-side in 4256a9d5 — members without a username are filtered before they reach the assign request body.

apps/lfx-one/src/app/shared/services/akrites.service.ts
gaspergrom: Fixed in 4256a9d5: removed unused PII fields (`email`, `committeeName`, `committeeUid`) from both the service mapping and the `AkritesSearchStewardResult` interface.

apps/lfx-one/src/app/shared/services/akrites.service.ts
gaspergrom: Fixed in 4256a9d5: added error logging to the catchError handler so API failures are visible in console. The current minimal error state is acceptable for now — the empty array return correctly prevents the modal from displaying members in a broken state.

apps/lfx-one/src/app/modules/akrites/components/akrites-triage-tab/akrites-triage-tab.component.ts
gaspergrom: This is a good catch. The current implementation relies on the upstream CDP `openStewardship` endpoint being idempotent for the same `purl`. This is a valid assumption per the CDP contract, but we should document it. For now, the current behavior is safe — if `assignSteward` fails, a retry will call `openStewardship` again, which will return the same stewardship ID if the row already exists.

packages/shared/src/constants/akrites.constants.ts
gaspergrom: Good point. The hardcoded UID is intentional for now — it references the canonical LFX Akrites stewards committee. Moving to an environment-keyed config would add flexibility, but it's out of scope for this PR. The comment in the constant already calls out that this is environment-specific, so maintainers will be aware if they need to update it. Worth a follow-up ticket if the steward pool committee changes between environments.

apps/lfx-one/src/app/modules/akrites/components/akrites-triage-tab/akrites-triage-tab.component.ts
gaspergrom: This was already addressed in a recent commit.

apps/lfx-one/src/app/modules/akrites/components/akrites-overview-tab/akrites-overview-tab.component.ts
```diff
@@ -9,25 +9,27 @@ import {
...
   templateUrl: './akrites-overview-tab.component.html',
 })
 export class AkritesOverviewTabComponent {
   private readonly akritesService = inject(AkritesService);
-  private readonly projectContextService = inject(ProjectContextService);
+  private readonly messageService = inject(MessageService);
   private readonly destroyRef = inject(DestroyRef);
 
-  protected readonly canWrite = this.projectContextService.canWrite;
+  protected readonly canWrite = signal(true);
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 826c3248: canWrite now derives from PersonaService, restricting write actions to users with executive-director role.

apps/lfx-one/src/app/modules/akrites/components/akrites-overview-tab/akrites-overview-tab.component.ts
```diff
@@ -198,7 +238,11 @@ export class AkritesOverviewTabComponent {
         order.push(label);
       }
 
-      // Pre-compute all display properties: styles, icons, labels, actions
+      // Pre-compute all display properties: styles, icons, labels, actions, actor
+      const actor = row.actor;
+      const actorDisplay = actor?.displayName || actor?.username || null;
+      const actorInitials = this.computeActorInitials(actor);
+
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 826c3248: actorDisplay now includes userId as a fallback, and computeActorInitials also falls back to userId.

apps/lfx-one/src/server/helpers/validation.helper.ts
```diff
@@ -295,39 +296,65 @@ export function parseStewardshipId(req: Request, operation: string): number {
...
+  }
+  if (typeof actor.userId !== 'string' || actor.userId.trim() === '') {
+    throw ServiceValidationError.forField('actor.userId', 'actor.userId is required', { operation });
+  }
+  return {
+    userId: actor.userId.trim(),
+    username: typeof actor.username === 'string' ? actor.username.trim() || null : null,
+    displayName: typeof actor.displayName === 'string' ? actor.displayName.trim() || null : null,
+    avatarUrl: typeof actor.avatarUrl === 'string' ? actor.avatarUrl.trim() || null : null,
+  };
```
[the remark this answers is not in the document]
gaspergrom: The CDP endpoints validate actor.userId === JWT sub on all inbound requests, so the client cannot spoof a different identity. The display fields (displayName, avatarUrl, username) are convenience metadata for activity feed display that the server cannot derive from the JWT alone without making additional calls. This design is safe per CDP contract.

apps/lfx-one/src/server/services/akrites.service.ts
```diff
@@ -10,9 +10,9 @@ import {
   CdpPackagesListResponse,
   CdpPackagesMetricsResponse,
   CdpScatterResponse,
-  CdpStewardSummary,
   CdpStewardshipSummary,
   AkritesActivityResponse,
+  AkritesActorInput,
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 826c3248: AkritesOpenStewardshipRequest interface now includes the actor property, and the BFF request body is typed against it to prevent drift.

MRashad26 (commented):
## Re-review — PR #1001 (new commits since approval at `044035b4`)

All code standards met on the 4 new commits. One new finding to flag before merge.

### New commits reviewed
- `329077a4` — feat: send actor on write actions and show in feed
- `4256a9d5` — fix: filter username-less members, remove unused PII
- `826c3248` — fix: address CodeRabbit review comments

### Code standards — all clear ✅
- Interfaces (`CdpActivityActor`, `AkritesActorInput`, extensions to `AkritesActivityRowVM`) in `@lfx-one/shared/interfaces` ✅
- `AKRITES_STEWARD_COMMITTEE_UID` in `@lfx-one/shared/constants` ✅
- Templates: `steward.initials` / `row.actorInitials` are pre-computed VM fields; no method calls in `{{ }}`, `@if`, or bindings ✅
- Signal patterns: all mutable state via `signal()`, reactive chains use `toObservable() + switchMap + toSignal`, no `effect()` ✅

### Finding

🟡 **DEV-ONLY committee UID will reach production** — `packages/shared/src/constants/akrites.constants.ts`

The constant ships with the dev group UID (`f41a2f37-a4b0-441f-bacf-49a0a9a5dd8d`) and a `// DEV ONLY: AY Group (7 members) — revert to 8bffb08a` comment. If merged and deployed, the steward picker fetches from the wrong committee, showing only 7 dev members instead of the production Akrites working group. See inline comment.

packages/shared/src/constants/akrites.constants.ts
```diff
@@ -160,6 +160,13 @@ export const AKRITES_STATUS_LABELS: Record<AkritesStatus, string> = {
 /** Display order for Akrites statuses in the risk matrix legend and filters. */
 export const AKRITES_STATUS_ORDER: AkritesStatus[] = ['unassigned', 'needs_attention', 'escalated', 'blocked', 'inactive', 'open', 'assessing', 'active'];
 
+/**
+ * Committee UID for the "LFX Akrites" working group — the source of assignable stewards.
+ * Members of this committee are fetched via GET /api/committees/:id/members.
+ * See: https://app.lfx.dev/groups/8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6
+ */
+export const AKRITES_STEWARD_COMMITTEE_UID = 'f41a2f37-a4b0-441f-bacf-49a0a9a5dd8d'; // DEV ONLY: AY Group (7 members) — revert to 8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6
+
```
MRashad26, line 169: 🟡 **DEV-ONLY UID will reach production**

This constant is set to the dev test group (`f41a2f37`) rather than the production Akrites working group (`8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6`). The inline comment acknowledges this, but if merged and deployed, the steward picker will load from a 7-member dev group instead of the real committee.

**Suggested fix:** Swap to the production UID before merge, or block on a config/env-based constant if different UIDs are intentional across environments:
```ts
export const AKRITES_STEWARD_COMMITTEE_UID = '8bffb08a-3707-4f8f-9e1c-0cbca8a4dfb6';
```

gaspergrom: Fixed in eb6a5fe5: swapped to production Akrites steward committee UID.

MRashad26 (dismissed):

dealako (dismissed):
Hey @gaspergrom 👋 — this is a significant step forward and the latest commits have addressed every blocking concern from the prior round. The searchable steward picker is clean, the `stewardshipId$` single-chain pattern in the triage and overview tabs is the right shape, and the production committee UID is in place. Moving to approve.

---

🔴 **Blocking:** 0 issues
🟡 **Minor:** 2 issues — actor spoofing via client body, no test coverage
⚪ **Nit:** 2 issues — drawer branching inconsistency, modal header on reassign path

---

### ✅ Confirmed resolved from round 1

- **Empty username → 400** — filtered in `searchStewards()` before emit; confirmed in `4256a9d5`
- **`getInitials()` in template** — removed; `steward.initials` pre-computed in mapper; interface updated
- **Dev-only committee UID** — reverted to production `8bffb08a` in `eb6a5fe5`
- **`canWrite` binding** — now `computed(() => personaService.currentPersona() === 'executive-director')`; not hardcoded
- **RxJS duplicate subscription in triage tab** — refactored to single `stewardshipId$` chain
- **`onSearchInput` selection-clear** — correctly clears `selectedSteward` when query hides the selected row
- **A11y** — `aria-label` on search input and `aria-pressed` on the toggle button both present

---

### 🟡 Minor: Actor identity trusted from client body (security)

`parseActor(body.actor, operation)` reads `userId`, `username`, `displayName`, and `avatarUrl` directly from the request body on all four write endpoints. The frontend's `buildActor()` populates this from the Auth0 session (`UserService.user()`), which is the right *source*, but the BFF accepts whatever is in `body.actor` without cross-checking it against the server-side OIDC session.

Any authenticated user can craft a raw HTTP request with a spoofed `actor.userId` and the audit trail will record that spoofed identity. This was flagged by CodeRabbit in the second review round and the threads were closed without the root fix landing.

**Fix:** In `parseActor`, read from the server-side session instead of the request body:

```typescript
function parseActor(req: Request, operation: string): AkritesActorInput {
  const user = req.oidc?.user;
  if (!user?.sub) {
    throw ServiceValidationError.forField('actor', 'Authenticated user context is required', { operation });
  }
  return {
    userId: user.sub,
    username: user.nickname || user['https://sso.linuxfoundation.org/claims/username'] || null,
    displayName: user.name || null,
    avatarUrl: user.picture || null,
  };
}
```

Then remove `actor` from all four request body shapes (`body.actor` → drop from `parseOpenStewardshipBody`, `parseAssignStewardBody`, `parseEscalateBody`, `parseUpdateStatusBody`). The frontend `buildActor()` can be removed too — the server now owns actor identity entirely.

Not blocking merge for a UI improvement PR, but it should be addressed before Akrites ships widely.

---

### 🟡 Minor: No test coverage

The new modal picker (search, selection, role choice, `moveToAssessing` toggle), the open-then-assign flow, and the actor injection path have no unit or E2E tests. `data-testid`s are present throughout so coverage would be low-friction to add. Repo policy targets >80% for new code and expects dual-architecture E2E.

---

### ⚪ Nit: Package drawer uses branched two-subscribe pattern

`onAssignStewardConfirm` in `akrites-package-drawer.component.ts` still uses a branched pair of subscribes (one for existing `stewardshipId`, a second for open-then-assign). The triage tab and overview tab both adopted the unified `stewardshipId$ = existingId !== null ? of(id) : openStewardship()…` single-chain shape from commit `37af1dff`. The drawer should match — duplicated subscribe/reset blocks are a maintenance hazard.

---

### ⚪ Nit: Modal header says "Assign steward" on the reassign path

The drawer and triage footer correctly show **Reassign** for `assessing`/`escalated`/`inactive` packages, but the modal header is hardcoded to `header="Assign steward"`. Dynamically setting it to **Reassign steward** when `stewardshipId` is non-null would complete the UX — otherwise the header contradicts the button that opened it.

---

### ⚠️ Pre-existing — not gating

The four write routes remain authorization-free beyond the LD `akritesEnabledGuard` (client-side). Tracked as CM-1245 Gap #4. Not introduced by this PR and not gating it.

---

apps/lfx-one/src/app/shared/services/akrites.service.ts
```diff
@@ -72,18 +78,56 @@ export class AkritesService {
...
+
+  private buildActor(): AkritesActorInput {
+    const user = this.userService.user();
+    return {
+      userId: user?.sub ?? '',
+      username: user?.nickname || user?.username || user?.['https://sso.linuxfoundation.org/claims/username'] || null,
+      displayName: user?.name || null,
+      avatarUrl: user?.picture || null,
+    };
   }
```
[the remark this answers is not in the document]
gaspergrom: Fixed in eb97a1ac: Added guard checks in buildActor() to throw an error if userId is empty, preventing silent 400 failures.

apps/lfx-one/src/app/modules/akrites/components/akrites-triage-tab/akrites-triage-tab.component.ts
```diff
@@ -14,20 +15,21 @@ import {
   AkritesTriageStatus,
 } from '@lfx-one/shared/interfaces';
 import { AkritesService } from '@shared/services/akrites.service';
-import { ProjectContextService } from '@shared/services/project-context.service';
+import { PersonaService } from '@shared/services/persona.service';
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 3014288c and eb97a1ac: Reverted canWrite from PersonaService to ProjectContextService for consistent authorization with the rest of the app.

apps/lfx-one/src/app/modules/akrites/components/akrites-triage-tab/akrites-triage-tab.component.ts
```diff
@@ -37,7 +39,10 @@ export class AkritesTriageTabComponent {
 
   protected readonly TRIAGE_COLUMNS = AKRITES_TRIAGE_COLUMNS;
   protected readonly loading = signal(true);
-  protected readonly canWrite = this.projectContextService.canWrite;
+  protected readonly actionLoading = signal(false);
+  protected readonly assignModalVisible = signal(false);
+  protected readonly assignTargetPackage = signal<AkritesTriagePackageVM | null>(null);
+  protected readonly canWrite = computed(() => this.personaService.currentPersona() === 'executive-director');
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 3014288c: Same as above - canWrite now uses ProjectContextService.canWrite().

apps/lfx-one/src/app/modules/akrites/components/akrites-overview-tab/akrites-overview-tab.component.ts
```diff
@@ -9,25 +9,29 @@ import {
...
+  AkritesAssignStewardRequest,
   AkritesFilterState,
   AkritesMetrics,
 } from '@lfx-one/shared/interfaces';
-import { catchError, of, switchMap, tap } from 'rxjs';
+import { catchError, map, of, switchMap, take, tap } from 'rxjs';
+import { MessageService } from 'primeng/api';
 import { AkritesService } from '@shared/services/akrites.service';
-import { ProjectContextService } from '@shared/services/project-context.service';
+import { PersonaService } from '@shared/services/persona.service';
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 3014288c: Same as above - canWrite now uses ProjectContextService.canWrite().

apps/lfx-one/src/app/modules/akrites/components/akrites-overview-tab/akrites-overview-tab.component.ts
```diff
@@ -9,25 +9,29 @@ import {
...
 })
 export class AkritesOverviewTabComponent {
   private readonly akritesService = inject(AkritesService);
-  private readonly projectContextService = inject(ProjectContextService);
+  private readonly messageService = inject(MessageService);
   private readonly destroyRef = inject(DestroyRef);
+  private readonly personaService = inject(PersonaService);
 
-  protected readonly canWrite = this.projectContextService.canWrite;
+  protected readonly canWrite = computed(() => this.personaService.currentPersona() === 'executive-director');
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 3014288c: Same as above - canWrite now uses ProjectContextService.canWrite().

packages/shared/src/interfaces/akrites.interface.ts
```diff
@@ -202,11 +222,23 @@ export interface CdpStewardshipRecord {
...
+  avatarUrl: string | null;
+}
+
+export interface AkritesAssignStewardRequest {
+  steward: {
+    userId: string;
+    username: string | null;
+    displayName: string | null;
+    role: AkritesStewardRole;
+  };
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 3014288c: Made username and displayName required (non-nullable) in AkritesAssignStewardRequest interface.

apps/lfx-one/src/server/helpers/validation.helper.ts
```diff
@@ -295,39 +296,65 @@ export function parseStewardshipId(req: Request, operation: string): number {
...
 
   return {
-    userId: body.userId.trim(),
-    role: body.role as AkritesStewardRole,
+    steward: {
+      userId: steward.userId.trim(),
+      username: typeof steward.username === 'string' ? steward.username.trim() || null : null,
+      displayName: typeof steward.displayName === 'string' ? steward.displayName.trim() || null : null,
+      role: steward.role as AkritesStewardRole,
+    },
```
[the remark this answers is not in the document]
gaspergrom: Fixed in 3014288c: Updated parseAssignStewardBody() to validate and reject empty username and displayName fields with field-level errors.

dealako:
Hey @gaspergrom 👋 — solid iteration in this round. The `buildActor()` unauthenticated-user guards are correctly placed, `projectContextService.canWrite()` is the right permission source for both tabs, and the `username`/`displayName` required-field enforcement is now consistent from frontend → server validation → interface contract. Good work closing those threads.

One new blocking finding crept in from the `canWrite` migration, plus a couple of minors carried from the security-auditor pass.

---

🔴 **Blocking: 1** — dead `PersonaService` injection in both tabs  
🟡 **Minor: 3** — actor identity, drawer permission inconsistency, no tests  
⚪ **Nit: 2** — drawer subscribe pattern, modal header on reassign

---

### 🔴 Needs changes before approval

---

### 🔴 Blocking #1 — Dead `PersonaService` injection (two files)

`PersonaService` is imported and injected in both tabs but is now completely unused — the only caller was `canWrite`, which was migrated to `projectContextService.canWrite()` in commit `3014288c`.

**Files:**
- `apps/lfx-one/src/app/modules/akrites/components/akrites-triage-tab/akrites-triage-tab.component.ts` — lines 18 (import) + 33 (inject)
- `apps/lfx-one/src/app/modules/akrites/components/akrites-overview-tab/akrites-overview-tab.component.ts` — lines 19 (import) + 33 (inject)

`CLAUDE.md` is explicit: *"Remove imports/variables/functions that YOUR changes made unused."* The `canWrite` migration created this dead code and it should be cleaned up in the same pass. `tseslint.configs.recommended` also flags unused injections, so this may fail CI lint.

**Fix:** Remove the `import { PersonaService }` line and the `private readonly personaService = inject(PersonaService)` line from both components. Verify `PersonaService` has no other usages in those files first (grep confirms it doesn't — 0 references to `personaService.` in either file).

---

### 🟡 Minor #2 — Drawer action handlers have no `canWrite` gate

The triage tab and overview tab both compute `protected readonly canWrite = computed(() => this.projectContextService.canWrite())` and gate button visibility on it. The package drawer's action buttons (`onAssignStewardConfirm`, `onOpenForStewardship`, etc.) rely only on status-based computed signals (`canAssignSteward`, `canManageStatus`, etc.) with no project-writer check. A read-only user who can open the drawer gets an error toast instead of a disabled button.

Not blocking merge — the service guard (`buildActor()` throwing on missing userId) prevents actual writes from unauthenticated users — but the UX is inconsistent with the tab components. Worth aligning in a follow-up before Akrites ships widely.

---

### 🟡 Minor #3 — Actor identity from request body (re-escalated by security scan; mitigated by CDP contract)

The security-auditor subagent flagged `parseActor()` reading from `body.actor` as a blocking spoofability concern. Noting Gašper's inline reply that *"CDP endpoints validate `actor.userId === JWT sub` on all inbound requests"* — if that CDP-layer invariant holds, the userId-spoofing attack vector is closed upstream.

That explanation partially mitigates the risk, but:
1. The CDP contract is unverified from this side of the proxy.
2. `displayName` / `avatarUrl` / `username` remain client-controlled (cosmetic metadata only per Gašper's explanation, but worth noting in a code comment so future readers understand the intentional design).

Carrying forward as a minor. **Before Akrites ships widely**, either verify the CDP invariant and document it with a comment in `parseActor()`, or move actor extraction to `req.oidc.user`. Not gating this round.

---

### 🟡 Minor #4 — No test coverage (carried)

No unit or E2E tests for: the `buildActor()` null-userId guards (all four methods), the `canWrite → projectContextService` binding, the open-then-assign flow, or the role/`moveToAssessing` modal paths. `data-testid` attributes are in place so coverage would be low-friction. Repo policy targets >80% for new code. Carrying forward — not gating this round but this must land before Akrites ships.

---

### ⚪ Nit #5 — Drawer `onAssignStewardConfirm` two-branch pattern

The drawer still uses an early-return + two separate subscribe blocks for the open-then-assign case. The triage tab and overview tab both adopted the unified `stewardshipId$ = existingId !== null ? of(id) : openStewardship()` → single-chain shape. The drawer should follow for consistency — two subscribe/reset blocks is a maintenance hazard.

### ⚪ Nit #6 — Modal header on reassign path

Drawer and triage footer show **Reassign** for `assessing`/`escalated`/`inactive` packages; modal header is still hardcoded `"Assign steward"`. Low-impact, but the header contradicts the button that opened it.

---

🔴 **Needs changes before approval** — the dead `PersonaService` injection is a quick two-file cleanup; once that lands, the remaining items are documented trade-offs or carried minors.

dealako (changes requested):
🔴 **Needs changes before approval** — one new finding from this round: dead `PersonaService` injection in `akrites-triage-tab` and `akrites-overview-tab` (both components import + inject it but it's now unused after the `canWrite` migration to `projectContextService`). Quick two-file cleanup. All other issues are documented minor trade-offs or nits.

apps/lfx-one/src/app/shared/services/akrites.service.ts
gaspergrom: Thanks for flagging this. The error handling is intentional here — we treat fetch failures and empty rosters identically (both render 'No stewards available'). This simplifies the UX at this stage (v1 of the picker). A future iteration could add an explicit error state and retry affordance if needed, but for now we're accepting the trade-off. The loading state still provides visual feedback that something was attempted.

packages/shared/src/constants/akrites.constants.ts
gaspergrom: Good point on environment robustness. We hardcoded the steward committee UID as a constant since there is a single 'LFX Akrites' working group across all environments (staging and prod use the same UID). If future environments or a refactor needs environment-specific committee configuration, that can be driven from the runtime config service at that time. For now, the compile-time constant with a comment linking to the committee keeps it discoverable.

gaspergrom:
Fixed in e4fdbd6b: removed `import { PersonaService }` and the `personaService = inject(PersonaService)` injection from both `akrites-overview-tab` and `akrites-triage-tab` — the only caller was `canWrite`, which was already migrated to `projectContextService.canWrite()`.

For the minors/nits:
- **Minor #2 (drawer canWrite gate)** — noted, will align in a follow-up before wide rollout.
- **Minor #3 (actor identity)** — will add a comment in `parseActor()` documenting the CDP-side invariant (`actor.userId === JWT sub`) so future readers understand the design intent. `displayName`/`avatarUrl`/`username` are cosmetic only.
- **Minor #4 (tests)** — noted, targeting before Akrites ships widely.
- **Nit #5 (drawer subscribe pattern)** — will unify in the drawer follow-up.
- **Nit #6 (modal header on reassign)** — will fix in the same follow-up.

apps/lfx-one/src/app/shared/services/akrites.service.ts
```diff
@@ -72,18 +78,68 @@ export class AkritesService {
...
+          .map((m) => {
+            const initials = `${m.first_name?.[0] ?? ''}${m.last_name?.[0] ?? ''}`.toUpperCase() || (m.username?.[0] ?? 'U').toUpperCase();
+            return {
+              userId: m.uid,
+              username: m.username,
+              displayName: `${m.first_name} ${m.last_name}`.trim(),
+              organization: m.organization?.name ?? null,
+              status: m.status ?? '',
+              initials,
+            };
```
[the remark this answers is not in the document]
gaspergrom: Fixed in af0a8b8e:
- Changed `userId` from `m.uid` (membership UID) to `m.username` (stable Auth0 identity) in `searchStewards()` mapper
- Added `displayName` fallback to `username` to ensure the required field is always set
- This prevents the downstream assign call from failing or misattributing the steward

apps/lfx-one/src/server/services/akrites.service.ts
```diff
@@ -541,14 +537,16 @@ export class AkritesServerService {
     }
   }
 
-  /** Map CDP steward rows to the UI shape. Name/avatar stay null until the roster endpoint exists. */
-  private mapStewards(stewards: CdpStewardSummary[] | null): AkritesSteward[] {
+  private mapStewards(
+    stewards: { userId: string; username: string | null; displayName: string | null; role: string; assignedAt: string }[] | null
+  ): AkritesSteward[] {
```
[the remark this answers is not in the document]
gaspergrom: Fixed in af0a8b8e:
- Extracted the inline steward type from `CdpStewardshipSummary` into a named interface `CdpStewardshipSteward`
- Updated `mapStewards()` to use the shared `CdpStewardshipSteward[]` type instead of duplicating the shape
- This ensures the type stays synchronized with the CDP contract and reduces maintenance burden

dealako:
Hey @gaspergrom 👋 — this round is clean. Every item from round 3 is addressed, and the Copilot fixes landed well.

👏 **Specific improvements in this revision:**

- **Dead `PersonaService` injection removed** — both `akrites-triage-tab` and `akrites-overview-tab` now only inject what they use. Clean.
- **`userId: m.uid` → `userId: m.username`** — this was a correctness fix. `m.uid` was the committee membership record UID (wrong identifier for CDP); the Auth0 username is what CDP uses for steward identity. The commit message explains the reasoning clearly.
- **`displayName` fallback** — `|| m.username` ensures the required `displayName` field is always non-empty even when `first_name`/`last_name` are both absent. Solid defensive mapping.
- **`CdpStewardshipSteward` interface extracted** — replaces the inline anonymous type in `mapStewards()` and in `CdpStewardshipSummary.stewards`. Proper shared-package hygiene.

---

🔴 **Blocking: 0**
🟡 **Minor: 3** — actor body trust (deferred), drawer `canWrite` gate inconsistency, no tests
⚪ **Nit: 2** — drawer two-branch subscribe pattern, modal header on reassign path

All three minors are pre-documented trade-offs from prior rounds — not new findings, not introduced by this revision. They're on the pre-GA backlog and don't gate a UI improvement PR.

---

dealako (approved):
✅ **Approved** — round 3 blocking item (dead PersonaService injection) is cleanly fixed. The Copilot fixes in `af0a8b8e` are correct: `userId: m.username` replaces the wrong committee membership UID with the Auth0 username CDP expects, `displayName` fallback covers the missing-name edge case, and `CdpStewardshipSteward` extraction keeps the shared package clean. No new issues introduced. Remaining minors (actor body trust, drawer canWrite, test coverage) are pre-documented pre-GA trade-offs. Ready to land.

PR #1003, opened 2026-06-22
feat(org-lens): send invite when adding or replacing a key contact

ahmedomosanya:
## Summary
- Set `send_invite: true` when writing key contacts to the member-service so a newly added or replaced key contact receives an invitation to claim their LFID and access the project membership, instead of being created as a silent record.
- Applied to both the create key-contact `POST` body and the replace key-contact update body.

- No UI changes — the invite is dispatched server-side by member-service on contact write.

[LFXV2-2461]: https://linuxfoundation.atlassian.net/browse/LFXV2-2461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/services/org-lens-key-contacts.service.ts
+++ b/apps/lfx-one/src/server/services/org-lens-key-contacts.service.ts
@@ -252,6 +252,7 @@ export class OrgLensKeyContactsService {
         status: 'Active',
         primary_contact: body.contactType === KEY_CONTACT_PRIMARY_CONTACT_TYPE,
         board_member: false,
+        send_invite: true,
       }
     );
     return this.toPerson(created);
@@ -266,6 +267,7 @@ export class OrgLensKeyContactsService {
       primary_contact: body.contactType === KEY_CONTACT_PRIMARY_CONTACT_TYPE,
       board_member: false,
       ...(title ? { title } : {}),
+      send_invite: true,
     };
   }
 


apps/lfx-one/src/server/services/org-lens-key-contacts.service.ts
```diff
@@ -252,6 +252,7 @@ export class OrgLensKeyContactsService {
         status: 'Active',
         primary_contact: body.contactType === KEY_CONTACT_PRIMARY_CONTACT_TYPE,
         board_member: false,
+        send_invite: true,
```
audigregorie, line 255: **Minor — confirm the member-service field name**

`send_invite` is a magic key on an untyped write body — there's no shared interface backing this object, so the compiler can't catch a name mismatch. If the member-service contract expects a different key (e.g. `invite`, `sendInvite`, or a nested option), this line silently no-ops and the entire feature does nothing, with no error surfaced anywhere.

Worth verifying against the member-service OpenAPI/contract before merge, since a typo here is invisible. (Same applies to the `toUpdateBody` copy below.)

apps/lfx-one/src/server/services/org-lens-key-contacts.service.ts
```diff
@@ -266,6 +267,7 @@ export class OrgLensKeyContactsService {
       primary_contact: body.contactType === KEY_CONTACT_PRIMARY_CONTACT_TYPE,
       board_member: false,
       ...(title ? { title } : {}),
+      send_invite: true,
```
audigregorie, line 270: **Minor — re-invite on the known-employee replace path**

`toUpdateBody` feeds only the `known` branch of `replaceKeyContact`, which PUTs in place for a contact whose email already belongs to an existing org employee — i.e. someone who may already have a claimed LFID. With `send_invite: true` unconditional here, that contact is re-invited on every replace.

This matches the PR description ("applied to ... the replace key-contact update body"), so it looks intentional — just confirm member-service dedups/ignores invites for already-claimed accounts so a replace of an existing employee doesn't fire a duplicate invitation email.

audigregorie (commented):
## Code Review Summary

Clean, surgical 2-line change that does exactly what the PR title says — sets `send_invite: true` on both the create (`POST`) and replace-update (`PUT`) member-service write bodies. No correctness, security, or convention issues. Both inline comments below are confirmation requests, not defects.

### Minor — outside the diff

- **No test coverage.** There is no spec file for `OrgLensKeyContactsService`, so the new invite behavior is unverified by tests. Not introduced by this PR (the service was already untested), and adding a full spec is arguably out of scope for a 2-line flag flip — noting it so the gap is visible. If member-service treats `send_invite` as a no-op for already-claimed LFIDs, a thin test asserting the flag is present on both write bodies would still be cheap insurance against the field-name regression called out on line 255.

### What's done well

- Minimal and well-scoped — touches only the two write paths and nothing else.
- Consistent with the existing body-construction style; placed next to the sibling member-service flags (`board_member`, `primary_contact`).
- Covers both create paths (`createContact`, used by add + brand-new replace) and the in-place update path (`toUpdateBody`).

MRashad26 (commented):
## Review — PR #1003 `feat/LFXV2-2461-key-contact-send-invite`

**Verdict: PASS**

Minimal server-side change. Adds `send_invite: true` to the create (`POST /project_memberships/:uid/key_contacts`) body and the `toUpdateBody` replace payload in `OrgLensKeyContactsService`. The member-service will dispatch an invitation email to the key contact on write. Applied symmetrically to both write paths.

**Secrets check:** Clean — no credentials or tokens.

**Code standards:** No Angular frontend code, templates, interfaces, or constants changed — nothing to audit.

No issues found.

MRashad26 (approved):

apps/lfx-one/src/server/services/org-lens-key-contacts.service.ts
ahmedomosanya: Good callout. Confirmed against the member-service key-contact API contract (generated OpenAPI): the field is `send_invite` — boolean, default `false` — exposed on both the create and update bodies, which matches the key we send here and in `toUpdateBody`. So it's the right name, not a silent no-op.

apps/lfx-one/src/server/services/org-lens-key-contacts.service.ts
ahmedomosanya: Intentional, and the contract guards against a duplicate invite on a claimed account:
- The update/replace path only acts on `send_invite` when the contact's email actually changes — a same-contact edit (name/title/role) sends nothing.
- For an email that already has a registered account, it sends a role-assignment notification rather than a magic-link signup invite, and an existing live grant is deduped (the conflict is ignored).

So the unconditional `send_invite: true` here is safe and matches the intended invite-on-add/replace behavior.

PR #1005, opened 2026-06-22
fix(committees): use invite fields for org-required check

andrest50:
## Summary

- Non-viewer invitees could never accept a committee invite because the BFF called the access-controlled \`GET /committees/:id\` + \`GET /committees/:id/settings\` endpoints to determine whether an organization is required on accept — those endpoints fail for users who are not yet committee members
- The committee-service now carries \`committee_name\` and \`organization_required\` directly on the \`committee_invite\` resource (readable via the email-scoped query index, no committee-viewer access needed)
- Replace all access-failing committee/settings fetches in the accept flow with the invite's own fields

## Changes

- **\`CommitteeInvite\` / \`PendingInvitation\` interfaces** — add \`committee_name?\` and \`organization_required?\`
- **\`invitationRequiresOrganization\`** — prefer \`organization_required\` (authoritative, access-safe) before the legacy \`enable_voting\` / \`business_email_required\` fallback chain
- **\`getMyPendingInvitations\`** — source \`committee_name\` and \`organization_required\` from the invite directly; remove the per-committee \`getCommitteeSettings\` fetch loop
- **New \`getPendingInviteForUser\`** — access-safe invite lookup via email-scoped query index
- **\`acceptCommitteeInvite\` controller** — replace \`getCommitteeById\` + settings fetch with \`getPendingInviteForUser\` to drive org-required check
- **\`acceptPendingCommitteeInvitesAfterLfidAccept\`** — use \`invite.organization_required\` directly instead of fetching the committee; returns the first invite that requires an org the user must supply manually, so the invite page can open the org dialog before redirecting

## Known edge cases

**Multiple org-required invites:** If a user has more than one pending invite that requires an organization but has none pre-filled, only the **first** such invite is returned to the client for manual collection. Remaining invites remain pending and can be accepted from the My Groups / pending-actions surface after the user becomes a committee member. This is a deliberate single-invite-at-a-time flow.

**Pre-v1.1 invites (legacy):** Invites created before committee-service v1.1 will have \`organization_required: null\`. In that case \`invitationRequiresOrganization\` falls through to the legacy \`enable_voting\` / \`business_email_required\` chain, which itself fails closed (\`true\`) when both fields are absent. This means the accept endpoint may return a 400 asking for an org for committees that don't actually require one, until all open invites are re-created by the committee-service after the v1.1 deploy. Acceptable transitional behaviour — the invite list is short-lived.

## Ticket

[LFXV2-2453](https://linuxfoundation.atlassian.net/browse/LFXV2-2453)

## Related

Depends on upstream: https://github.com/linuxfoundation/lfx-v2-committee-service/pull/135

## Test plan

- [ ] As a user **not yet a viewer** of a committee, accept a pending invite for a **non-org-required** committee — should accept directly without 403
- [ ] As a user **not yet a viewer**, accept a pending invite for an **org-required** committee — org dialog should appear, and accept should succeed after filling it in
- [ ] As a user with an org-required invite that has no pre-filled org, accept via LFID invite link — org dialog should appear after LFID accept succeeds, redirect to committee page after org is supplied
- [ ] Invite surfaces (dashboard pending-actions, My Groups, committee banner) show the real committee name instead of a UID
- [ ] Existing member accepts flow (user already a viewer) still works unchanged

[LFXV2-2453]: https://linuxfoundation.atlassian.net/browse/LFXV2-2453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/controllers/committee.controller.ts
+++ b/apps/lfx-one/src/server/controllers/committee.controller.ts
@@ -13,7 +13,7 @@ import {
   CreateCommitteeJoinApplicationRequest,
   UploadCommitteeDocumentRequest,
 } from '@lfx-one/shared/interfaces';
-import { invitationRequiresOrganization, isFileTypeAllowed } from '@lfx-one/shared/utils';
+import { isFileTypeAllowed } from '@lfx-one/shared/utils';
 import { NextFunction, Request, Response } from 'express';
 import { Readable } from 'node:stream';
 import { ReadableStream as NodeReadableStream } from 'node:stream/web';
@@ -725,7 +725,11 @@ export class CommitteeController {
       // Build an explicit allowlist payload — never forward unknown client-supplied fields upstream.
       const acceptData: AcceptCommitteeInviteRequest = {};
 
-      if (invitationRequiresOrganization({ organization_required: pendingInvite.organization_required })) {
+      // Only pre-validate org when organization_required is explicitly true. When it is
+      // null/undefined (invite pre-dates committee-service v1.1), skip the BFF check and let
+      // the upstream accept endpoint be authoritative — avoid blocking valid accepts with a
+      // spurious 400 due to the fail-closed fallback in invitationRequiresOrganization.
+      if (pendingInvite.organization_required === true) {
         const body = (req.body ?? {}) as AcceptCommitteeInviteRequest;
         const orgName = typeof body.organization?.name === 'string' ? body.organization.name.trim() : '';
         if (!orgName) {
--- a/apps/lfx-one/src/server/controllers/committee.controller.ts
+++ b/apps/lfx-one/src/server/controllers/committee.controller.ts
@@ -13,7 +13,7 @@ import {
   CreateCommitteeJoinApplicationRequest,
   UploadCommitteeDocumentRequest,
 } from '@lfx-one/shared/interfaces';
-import { committeeRequiresOrganization, isFileTypeAllowed } from '@lfx-one/shared/utils';
+import { invitationRequiresOrganization, isFileTypeAllowed } from '@lfx-one/shared/utils';
 import { NextFunction, Request, Response } from 'express';
 import { Readable } from 'node:stream';
 import { ReadableStream as NodeReadableStream } from 'node:stream/web';
@@ -24,6 +24,7 @@ import { contentDispositionAttachment } from '../helpers/content-disposition.hel
 import { buildVCalendar, fetchAllMeetingPages, meetingsToVEvents } from '../helpers/ics.helper';
 import { getStringQueryParam } from '../helpers/validation.helper';
 import { logger } from '../services/logger.service';
+import { getEffectiveEmail } from '../utils/auth-helper';
 import { CommitteeService } from '../services/committee.service';
 import { MeetingService } from '../services/meeting.service';
 import { generateM2MToken } from '../utils/m2m-token.util';
@@ -693,14 +694,38 @@ export class CommitteeController {
         return;
       }
 
-      // Fetch with throwOnSettingsError so a settings-service outage fails the accept
-      // (returns 503) rather than silently treating business_email_required as false.
-      const committee = await this.committeeService.getCommitteeById(req, id, { throwOnSettingsError: true });
+      // Determine org-required using the invite's own organization_required field — accessible
+      // to the invitee via the email-scoped query index regardless of committee-viewer status.
+      // This replaces the previous getCommitteeById + getCommitteeSettings fetch which failed
+      // the access check for invitees who are not yet committee viewers (LFXV2-2453).
+      const userEmail = getEffectiveEmail(req);
+      if (!userEmail) {
+        next(
+          ServiceValidationError.forField('email', 'User email not found in authentication context', {
+            operation: 'accept_committee_invite',
+            service: 'committee_controller',
+            path: req.path,
+          })
+        );
+        return;
+      }
+
+      const pendingInvite = await this.committeeService.getPendingInviteForUser(req, userEmail, id, inviteId);
+      if (!pendingInvite) {
+        next(
+          ServiceValidationError.forField('inviteId', 'No matching pending invite found for this user', {
+            operation: 'accept_committee_invite',
+            service: 'committee_controller',
+            path: req.path,
+          })
+        );
+        return;
+      }
 
       // Build an explicit allowlist payload — never forward unknown client-supplied fields upstream.
       const acceptData: AcceptCommitteeInviteRequest = {};
 
-      if (committeeRequiresOrganization(committee)) {
+      if (invitationRequiresOrganization({ organization_required: pendingInvite.organization_required })) {
         const body = (req.body ?? {}) as AcceptCommitteeInviteRequest;
         const orgName = typeof body.organization?.name === 'string' ? body.organization.name.trim() : '';
         if (!orgName) {
--- a/apps/lfx-one/src/server/services/committee.service.ts
+++ b/apps/lfx-one/src/server/services/committee.service.ts
@@ -25,7 +25,7 @@ import {
   QueryServiceResponse,
   UploadCommitteeDocumentRequest,
 } from '@lfx-one/shared/interfaces';
-import { committeeRequiresOrganization } from '@lfx-one/shared/utils';
+import { invitationRequiresOrganization } from '@lfx-one/shared/utils';
 import { Request } from 'express';
 import FormData from 'form-data';
 
@@ -661,8 +661,11 @@ export class CommitteeService {
       pending_count: pendingInvites.length,
     });
 
-    // Enrich committee context (name, category, project) for the distinct committees once.
-    // Best-effort: a failed lookup degrades to a UID fallback rather than dropping the list.
+    // Enrich committee context (category, project_name) for the distinct committees once.
+    // committee_name and organization_required are sourced directly from the invite itself
+    // (committee-service ≥ v1.1) so no committee/settings fetch is required to decide org
+    // requirement — those endpoints fail the access check for invitees who are not yet viewers.
+    // Best-effort: a failed lookup degrades gracefully; category/project_name become null.
     const committeeUids = Array.from(new Set(pendingInvites.map((invite) => invite.committee_uid).filter(Boolean)));
     const committeeContext = new Map<
       string,
@@ -670,8 +673,6 @@ export class CommitteeService {
         committee_name: string;
         category?: string | null;
         project_name?: string | null;
-        enable_voting?: boolean;
-        business_email_required?: boolean;
       }
     >();
 
@@ -680,28 +681,6 @@ export class CommitteeService {
       const enriched = await this.projectService.enrichWithProjectData(req, Array.from(committees.values()));
       const projectNameByCommittee = new Map(enriched.map((committee) => [committee.uid, committee.project_name]));
 
-      // business_email_required is a settings-only field not present in the query-service
-      // committee index (which only stores CommitteeBase). Fetch settings in parallel so the
-      // invitationRequiresOrganization guard is accurate for business-email-only committees.
-      // throwOnError: true so a transient outage throws and is caught below, where we fail
-      // closed (true) rather than leaving business_email_required undefined — enable_voting
-      // is always defined from the index, so the both-undefined guard in
-      // invitationRequiresOrganization never fires here; without this, a settings outage
-      // would silently skip the org dialog and cause a server 400 on accept.
-      const settingsByUid = new Map<string, boolean | undefined>();
-      await Promise.all(
-        committeeUids.map(async (uid) => {
-          try {
-            const settings = await this.getCommitteeSettings(req, uid, { throwOnError: true });
-            settingsByUid.set(uid, settings.business_email_required);
-          } catch {
-            // Settings unavailable — fail closed: assume org is required so the dialog
-            // appears and the server's authoritative check can decide.
-            settingsByUid.set(uid, true);
-          }
-        })
-      );
-
       for (const uid of committeeUids) {
         const committee = committees.get(uid);
         if (committee) {
@@ -709,8 +688,6 @@ export class CommitteeService {
             committee_name: committee.name || uid,
             category: committee.category ?? null,
             project_name: projectNameByCommittee.get(uid) || null,
-            enable_voting: committee.enable_voting,
-            business_email_required: settingsByUid.get(uid),
           });
         }
       }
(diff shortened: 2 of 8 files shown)


packages/shared/src/utils/invitation.utils.ts
```diff
@@ -13,14 +13,25 @@ export function committeeRequiresOrganization(flags: { enable_voting?: boolean;
...
  */
 export function invitationRequiresOrganization(flags: {
+  organization_required?: boolean | null;
   enable_voting?: boolean;
   business_email_required?: boolean;
   inviteRequiresOrganization?: boolean;
 }): boolean {
+  if (flags.organization_required != null) {
+    return flags.organization_required;
+  }
```
dealako, line 34: **[minor]** The new `organization_required` precedence branch added to `invitationRequiresOrganization` has no test coverage. This is a shared helper used by both the UI and BFF accept validation — the correctness of the new priority ordering should be verified before merge.

CodeRabbit suggested the following cases (agreeing — adding +1 here):

```typescript
it("prefers organization_required when explicitly false", () => {
  expect(invitationRequiresOrganization({ organization_required: false, enable_voting: true, business_email_required: true })).toBe(false);
});

it("prefers organization_required when explicitly true", () => {
  expect(invitationRequiresOrganization({ organization_required: true, inviteRequiresOrganization: false })).toBe(true);
});

it("falls through when organization_required is null", () => {
  expect(invitationRequiresOrganization({ organization_required: null, enable_voting: false, business_email_required: false })).toBe(false);
});
```

apps/lfx-one/src/server/controllers/committee.controller.ts
```diff
@@ -693,14 +694,38 @@ export class CommitteeController {
...
+      }
 
       // Build an explicit allowlist payload — never forward unknown client-supplied fields upstream.
       const acceptData: AcceptCommitteeInviteRequest = {};
 
-      if (committeeRequiresOrganization(committee)) {
+      if (invitationRequiresOrganization({ organization_required: pendingInvite.organization_required })) {
         const body = (req.body ?? {}) as AcceptCommitteeInviteRequest;
         const orgName = typeof body.organization?.name === 'string' ? body.organization.name.trim() : '';
         if (!orgName) {
```
dealako, line 735: **[minor]** Echoing Copilot's observation: the fail-closed behavior here (`invitationRequiresOrganization` returns `true` when `organization_required` is `null`/`undefined`) means any invite created before committee-service v1.1 is deployed will always show the org dialog, even for non-org-required committees.

This is documented behavior ("fail-closed is safer than skipping"), and it is the correct direction — but it creates a **deployment ordering dependency**: the committee-service v1.1 (PR #135) must be live before (or simultaneously with) this BFF change, otherwise users with in-flight pre-v1.1 invites will hit an unnecessary org-required gate.

Worth confirming:
1. Committee-service PR #135 has been merged and deployed to the target environment before this ships.
2. Or that all existing pending invites have been backfilled with `organization_required`.

dealako:
## PR Review — LFXV2-2453: Accept invite without viewer access

Hey @andrest50! 👋 This is a well-scoped fix that correctly addresses the root cause: non-viewer invitees were hitting access-controlled committee/settings endpoints that they're not authorized to read. Moving to `organization_required` on the invite itself (via the email-scoped query index) is the right architectural move — clean, access-safe, and properly fail-closed.

### Overall Impression

The authorization model is sound. I reviewed this from three angles (security auditor, general code quality, and my own analysis) and found no blocking issues. The core logic in `getPendingInviteForUser`, the precedence chain in `invitationRequiresOrganization`, and the LFID auto-accept flow changes are all correct. The PR depends on a parallel committee-service upgrade — that needs to be called out explicitly.

---

### Issue Summary

🔴 **Blocking: 0 issues**

🟡 **Minor: 4 issues**
1. Test coverage for the new `organization_required` precedence branch (both Copilot and CodeRabbit flagged; they even provided the test code)
2. Deployment dependency on committee-service ≥ v1.1 not explicitly called out in PR description — fail-closed behavior on old upstream causes unnecessary org dialogs
3. `acceptPendingCommitteeInvitesAfterLfidAccept` javadoc should note that 2nd+ org-required invites are silently deferred for manual acceptance
4. `organization.website` accepted in the accept flow body without URL format validation (defense-in-depth gap — pre-existing, but touched in this diff)

⚪ **Nit: 3 issues**
1. `let pendingCommitteeInvite = undefined` + `?? undefined` coercion is redundant — `let pendingCommitteeInvite: PendingCommitteeInviteForOrg | undefined;` with just `await ...` is cleaner
2. `getPendingInviteForUser` doesn't normalize the email before passing it (works because `fetchPendingCommitteeInvitesByEmail` normalizes internally, but explicit is better)
3. `organization_required: true` hardcode in `collectOrgAndAccept` should have a brief comment explaining the server always validates this before surfacing the invite

---

### AI Comment Reconciliation

**Copilot [invitation.utils.ts:34]** — Missing test coverage for `organization_required` → ✅ Agree. This is my Minor #1.

**Copilot [committee.controller.ts:731]** — Fail-closed for older invites → ✅ Agree. This is my Minor #2. The behavior is intentional and correct, but the deployment dependency should be explicit.

**CodeRabbit [invitation.utils.ts:26-34]** — Add test coverage; provided suggested tests → ✅ Agree and +1. The suggested tests cover the cases precisely.

---

### Final Decision

Primary ask: add the three `invitationRequiresOrganization` test cases (CodeRabbit already provided the code — literally copy-paste). Secondary: call out the committee-service deployment dependency in the PR description. Everything else is a nit that can be addressed in follow-up.

The core logic is correct and the security model is solid. This is close — one small test commit away from an approval.

dealako (changes requested):
Requesting changes — two items before merge:

1. **[minor] Add 3 test cases for `organization_required` precedence** in `packages/shared/src/utils/invitation.utils.spec.ts`. This is a shared helper used by both UI and BFF accept validation, and the new precedence branch has no coverage. CodeRabbit provided the exact test code in their comment.

2. **[minor] Call out the committee-service v1.1 deployment dependency** in the PR description. Without it, deploying this BFF change before the committee-service upgrade causes all invite accepts to fail-closed (org dialog appears for every committee, org-required or not). This is an ops safety note, not a code change.

Everything else is a nit. The core logic is correct, the authorization model is sound (email-scoped query prevents IDOR), and the fail-closed behavior is the right default. One small commit away from approval.

packages/shared/src/utils/invitation.utils.ts
andrest50: Addressed in the latest commit (): added 4 cases covering `organization_required: false` overriding legacy `true`, `organization_required: true` overriding `inviteRequiresOrganization: false`, and null/undefined falling through to the legacy path. All 139 tests pass.

apps/lfx-one/src/server/controllers/committee.controller.ts
andrest50: Also addressed in `da4e284f`: the accept controller now checks `pendingInvite.organization_required === true` explicitly instead of calling `invitationRequiresOrganization`. When the field is `null`/`undefined` (pre-v1.1 invite), the BFF skips the pre-check entirely and passes through to the upstream committee-service, which is authoritative. This removes the deployment ordering dependency — old invites will no longer hit a spurious org-required gate regardless of when this ships relative to committee-service PR #135.

packages/shared/src/utils/invitation.utils.ts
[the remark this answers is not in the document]
andrest50: Addressed in commit `da4e284f`: added 4 test cases for the new `organization_required` precedence branch — explicit false overrides legacy true, explicit true overrides `inviteRequiresOrganization: false`, and null/undefined each fall through to the legacy path. All 139 tests pass.

apps/lfx-one/src/server/controllers/committee.controller.ts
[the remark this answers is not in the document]
andrest50: Addressed in commit `da4e284f`: the accept controller now checks `pendingInvite.organization_required === true` explicitly rather than calling `invitationRequiresOrganization`. When the field is null/undefined (pre-v1.1 invite), the BFF skips the pre-check and passes through to the upstream committee-service, which is authoritative. This removes the deployment ordering dependency.

andrest50:
@dealako — thanks for the review! Both items from your \`CHANGES_REQUESTED\` have been addressed in commit \`da4e284f\`:

1. **Unit test coverage for `organization_required` precedence** — added 4 new cases to `invitation.utils.spec.ts` covering: explicit `false` overriding legacy `true`, explicit `true` overriding `inviteRequiresOrganization: false`, and `null`/`undefined` each falling through to the legacy path. All 139 tests pass.

2. **Deployment ordering dependency** — the accept controller no longer calls `invitationRequiresOrganization` for the BFF pre-check. It now checks `pendingInvite.organization_required === true` explicitly. When the field is `null`/`undefined` (pre-v1.1 invite), the BFF skips validation entirely and defers to the upstream committee-service, which is authoritative. Old invites will no longer hit a spurious org-required gate regardless of when this ships relative to committee-service PR #135.

packages/shared/src/utils/invitation.utils.ts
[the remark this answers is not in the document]
andrest50: Already addressed in commit `da4e284f` — 4 test cases were added to `invitation.utils.spec.ts` covering: `organization_required: false` overriding legacy true flags, `organization_required: true` overriding `inviteRequiresOrganization: false`, and `null`/`undefined` each falling through to the legacy path. All 139 tests pass.

dealako:
## Follow-up Review — LFXV2-2453 (commit `da4e284f`)

@andrest50 Great iteration — you didn't just address the feedback, you improved the solution. Here's the full follow-up pass.

---

### 👏 What's well done in this revision

**Tests are comprehensive** — The 4 new cases in `invitation.utils.spec.ts` cover every branch of the new `organization_required` precedence: explicit `false` overriding legacy `true` flags, explicit `true` overriding `inviteRequiresOrganization: false`, and both `null` and `undefined` falling through correctly. Each case has both-direction assertions for the fallthrough. This is exactly what the spec needed.

**Better fix than requested for the deployment dependency** — Instead of just documenting the fail-closed behavior, you eliminated the problem in the accept controller by switching to `pendingInvite.organization_required === true` strict check. Pre-v1.1 invites now pass through to the upstream service (which is authoritative) instead of hitting a spurious BFF 400. The PR description's new "Known edge cases" section also documents the multiple-org-invites and pre-v1.1 edge cases clearly. This is a cleaner design.

**Intentional asymmetry between the two accept paths** — The direct accept controller uses `=== true` (defer to upstream for ambiguous cases), while the auto-accept flow in `committee.service.ts` keeps `invitationRequiresOrganization()` (fail-closed for unknown cases since it's automated). This asymmetry is correct: explicit user actions should defer to the authoritative upstream; automated flows should be conservative.

**Import cleanup** — `invitationRequiresOrganization` cleanly removed from `committee.controller.ts` imports since it's no longer called there (appears only in a comment, as expected).

---

### Revision Tracking

✅ **[minor] Test coverage for `organization_required` precedence** — Resolved. 4 test cases added, 139 tests pass. (Addresses CodeRabbit, Copilot, and my original comment.)

✅ **[minor] Deployment dependency documentation** — Resolved, better than requested. Accept controller now uses `=== true` strict check; PR description documents known edge cases. Deployment ordering dependency is eliminated.

✅ **[nit] `pendingCommitteeInvite` type annotation** — Resolved in `a173a52`. Now explicitly typed as `PendingCommitteeInviteForOrg | undefined`.

⚪ **[nit] Email normalization in `getPendingInviteForUser`** — Not addressed. Acceptable — `fetchPendingCommitteeInvitesByEmail` normalizes internally and correctness is unaffected.

⚪ **[nit] Comment on `organization_required: true` hardcode in `invite.component.ts`** — Not addressed. Code is self-documenting given the method name; leaving as-is is fine.

---

### New Issues Found in Latest Commit

None. The changes in `da4e284f` are minimal and correct:
- Import removal: clean
- `=== true` check: correct, well-commented
- Test cases: comprehensive

Security auditor re-scan: **clean** — no new attack surface opened. The `=== true` check is safe because upstream committee-service remains authoritative.

---

### AI Bot Reconciliation

**Copilot (second-pass, 17:39)** — Still asking for test coverage despite `da4e284f` having already added them. Copilot was reviewing the stale commit. Andres correctly replied at 17:53 confirming the tests are already in. ✅ Agree with Andres — resolved.

---

### Issue Summary

🔴 **Blocking: 0 issues**
🟡 **Minor: 0 issues**
⚪ **Nit: 2 remaining** (email normalization, hardcode comment — both low-priority, existing nits not introduced by this PR)

---

### Final Decision

✅ **Approved** — All prior blockers resolved. Core logic is correct, secure, and well-tested.

apps/lfx-one/src/server/services/committee.service.ts
```diff
@@ -766,41 +752,41 @@ export class CommitteeService {
...
-            committee_uid: invite.committee_uid,
-            invite_uid: invite.uid,
-            err: settingsError,
-          });
-        }
+        // organization_required is carried on the invite itself (committee-service ≥ v1.1) so no
+        // committee/settings fetch is needed — those fail the access check for non-viewer invitees.
+        // Fail-closed when the field is absent: treat as org-required so we skip rather than
+        // auto-accept without the required organization.
+        const requiresOrganization = invitationRequiresOrganization({ organization_required: invite.organization_required });
```
dealako, line 771: **[nit]** The behavioral asymmetry between this path and the accept controller is intentional and correct, but could surprise a future reader:

- **Here (auto-accept):** uses `invitationRequiresOrganization()` — fail-closes to `true` for `null`/`undefined`, so pre-v1.1 invites that cannot be auto-accepted without an org are surfaced to the client for manual collection.
- **Accept controller (`committee.controller.ts`):** uses `=== true` strict check — pre-v1.1 invites skip BFF pre-validation and pass through to the authoritative upstream service.

Both are intentional: automated flows should be conservative; explicit user actions should defer to upstream. A comment here would prevent a future developer from "fixing" this asymmetry and accidentally regressing LFXV2-2453.

Suggested addition:
```typescript
// Use the helper here (not === true) to stay conservative in the automated flow:
// fail-closed for null/undefined so pre-v1.1 invites that need an org are surfaced
// to the client rather than auto-accepted without one. The accept controller uses
// === true and defers to upstream for null/undefined — appropriate for explicit user actions.
const requiresOrganization = invitationRequiresOrganization({ organization_required: invite.organization_required });
```

dealako (approved):
All prior [minor] items resolved. The fix is correct, well-tested, and the design improved during iteration. One [nit] left as a suggestion (comment documenting the intentional asymmetry between the two accept paths in committee.service.ts) — does not block merge.

Approved.

PR #1023, opened 2026-06-24
fix: force google ads api to use gaxios fetch

agaetep:
Google Ads API was returning a `GaxiosError`: `ERR_STREAM_PREMATURE_CLOSE`. Based on [this](https://station.railway.com/questions/outbound-https-to-oauth2-googleapis-com-086171e4) thread of the same error, this PR forces the use of `gaxios` global fetch, instead of inherently using `node-fetch`.

### DIFF
--- a/apps/lfx-one/src/server/services/campaign-proxy.service.ts
+++ b/apps/lfx-one/src/server/services/campaign-proxy.service.ts
@@ -37,8 +37,12 @@ import { executeRedditCampaignCreation, updateRedditCampaignStatus } from './red
 // ---------------------------------------------------------------------------
 
 import { GoogleAdsApi, enums } from 'google-ads-api';
+import { instance as gaxiosInstance } from 'gaxios';
 
 import type { Customer } from 'google-ads-api';
+
+// Use Node.js native fetch instead of node-fetch
+gaxiosInstance.defaults.fetchImplementation = globalThis.fetch;
 
 // ---------------------------------------------------------------------------
 // Required environment variables — log warnings on first use for missing ones
--- a/apps/lfx-one/src/server/services/campaign-proxy.service.ts
+++ b/apps/lfx-one/src/server/services/campaign-proxy.service.ts
@@ -41,7 +41,9 @@ import { instance as gaxiosInstance } from 'gaxios';
 
 import type { Customer } from 'google-ads-api';
 
-// Use Node.js native fetch instead of node-fetch
+// Override gaxios's bundled node-fetch with Node's built-in fetch (undici).
+// node-fetch fails with ERR_STREAM_PREMATURE_CLOSE when handling gzip-encoded
+// responses from oauth2.googleapis.com in the cluster environment.
 gaxiosInstance.defaults.fetchImplementation = globalThis.fetch;
 
 // ---------------------------------------------------------------------------
--- a/apps/lfx-one/src/server/services/campaign-proxy.service.ts
+++ b/apps/lfx-one/src/server/services/campaign-proxy.service.ts
@@ -24,6 +24,9 @@ import type {
   RedditCampaignCreateResult,
 } from '@lfx-one/shared/interfaces';
 import type { Request } from 'express';
+import { instance as gaxiosInstance } from 'gaxios';
+import { GoogleAdsApi, enums } from 'google-ads-api';
+import type { Customer } from 'google-ads-api';
 
 import { ServiceValidationError } from '../errors/service-validation.error';
 import { validateScrapeUrl, fetchSafeUrl } from '../helpers/url-validation';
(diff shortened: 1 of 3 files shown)


apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -37,9 +37,15 @@ import { executeRedditCampaignCreation, updateRedditCampaignStatus } from './red
 // ---------------------------------------------------------------------------
 
 import { GoogleAdsApi, enums } from 'google-ads-api';
+import { instance as gaxiosInstance } from 'gaxios';
```
audigregorie: **Major — `gaxios` is not a declared dependency of `apps/lfx-one`**

This import only resolves today because gaxios is hoisted transitively via `google-ads-api → google-auth-library@9.15.1 → gaxios@^6.1.1` (confirmed via `yarn why gaxios`). Two risks:

1. A future bump of `google-ads-api` (e.g. to a version that drops `google-auth-library@9` or pulls gaxios@7) silently breaks this import.
2. Worse — if `google-ads-api` ever bundles its own nested gaxios while the top-level hoist points elsewhere, the `gaxiosInstance` we mutate here will be a *different* instance from the one Google Ads' OAuth path actually uses, and this fix becomes a no-op with no error.

Please declare gaxios explicitly in `apps/lfx-one/package.json` (pinned to the same major as the real consumer):

```json
"gaxios": "^6.1.1"
```

That way the package manager guarantees one resolution and we get a CI signal if `google-ads-api` migrates off gaxios@6 in the future.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -37,9 +37,15 @@ import { executeRedditCampaignCreation, updateRedditCampaignStatus } from './red
...
 
 import type { Customer } from 'google-ads-api';
 
+// Override gaxios's bundled node-fetch with Node's built-in fetch (undici).
+// node-fetch fails with ERR_STREAM_PREMATURE_CLOSE when handling gzip-encoded
+// responses from oauth2.googleapis.com in the cluster environment.
+gaxiosInstance.defaults.fetchImplementation = globalThis.fetch;
```
audigregorie: **Major — global singleton mutation; blast radius is wider than the comment implies**

`gaxios.instance.defaults` is the **process-global gaxios@6 default**, not the Google Ads client's own gaxios. Per `yarn why gaxios`, this same instance is shared by `@google-cloud/storage`, `gcp-metadata@6`, `gtoken@7`, and `google-auth-library@9`. Forcing `globalThis.fetch` on the singleton changes default transport for all of them, not just google-ads-api — and the comment only mentions the Google Ads case.

Two options:

1. **Document the wider scope** — tighten the comment so future readers know the override is process-wide and intentional.
2. **Scope the override** — pass a custom `Gaxios` instance through whatever per-customer hook google-ads-api exposes (if any), so other Google libraries keep node-fetch. Cleaner, but only worth it if the option exists.

At minimum, please go with (1).

Note: per `.cursor/rules/global-rules.mdc`, code comments are **one or two lines max** — the existing 3-line block is already over. Suggested two-line replacement that covers both the original rationale and the wider blast radius:

```suggestion
// Force gaxios@6's process-global default to Node's native fetch (undici) — node-fetch
// throws ERR_STREAM_PREMATURE_CLOSE on gzipped oauth2.googleapis.com responses in-cluster.
if (globalThis.fetch) {
  gaxiosInstance.defaults.fetchImplementation = globalThis.fetch as typeof gaxiosInstance.defaults.fetchImplementation;
}
```

The `if (globalThis.fetch)` guard also lets us drop the assignment cleanly if we ever run on a runtime without native fetch.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
audigregorie: **Minor — import placement breaks the top-of-file import grouping**

All other external imports live above line 26 (the `import type` block). Adding `gaxios` here splits external imports into two regions separated by a section banner and the `import type { Customer }` block. Co-locating with `google-ads-api` reads nicely, but a future eslint-plugin-import or organize-imports pass will likely move it.

Consider moving both `google-ads-api` and the new `gaxios` import up into the main external-imports block, or leave a comment explaining why this group is intentionally separate.

audigregorie (commented):
## Code Review Summary

Tight, surgical fix for a real production bug (`ERR_STREAM_PREMATURE_CLOSE` from `oauth2.googleapis.com`) with a clear rationale comment. Two majors worth addressing before merge: (1) `gaxios` is imported but isn't a declared dependency of `apps/lfx-one` — it only resolves via transitive hoisting through `google-ads-api → google-auth-library@9 → gaxios@^6.1.1`, so a future bump could silently break the import or move `google-ads-api` to a *different* gaxios instance and make the override a no-op; (2) the singleton mutation affects every gaxios@6 consumer in the process (currently also `@google-cloud/storage`, `gcp-metadata`, `gtoken`, `google-auth-library`), which is fine but should be explicit.

### Outside the diff

- **No test coverage** for the override. A trivial assertion that `gaxios.instance.defaults.fetchImplementation === globalThis.fetch` after importing the service would catch a future package-resolution regression (see the dependency-declaration thread).
- **Branch name** `agaete/fix-google-ads` is missing a JIRA ticket and doesn't follow the `type/scope/description` convention; commit subjects also lack the `(LFXV2-#)` suffix. SHOULD_FIX per repo convention.
- **Commit `update inline comment with more description`** isn't a Conventional Commit. Squash-on-merge will hide it, but individual-commit commitlint would reject it.

### What's done well

- Comment about gaxios@6 bundling `node-fetch` is accurate (gaxios@7 already uses native fetch).
- Uses `globalThis.fetch` (Node 22 / undici) rather than pulling in another fetch polyfill.
- Minimal blast radius in this file.

Rating: ⭐⭐⭐⭐ (4/5)

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -24,6 +24,9 @@ import type {
   RedditCampaignCreateResult,
 } from '@lfx-one/shared/interfaces';
 import type { Request } from 'express';
+import { instance as gaxiosInstance } from 'gaxios';
```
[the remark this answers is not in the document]
agaetep: confirmed it directly from the installed package, not docs. Here's what the files say:

node_modules/gaxios/build/src/index.d.ts line 10:
export declare const instance: Gaxios;

node_modules/gaxios/build/src/index.js line 40:
exports.instance = new gaxios_1.Gaxios();

There is no gaxios named export in v6.

audigregorie (approved):

PR #1024, opened 2026-06-24
fix(committees): show invite accept ui when org-required group returns 403

MRashad26:
## Summary

- Invitees to org-required committees (voting/business-email rules) land on `/project/groups/:id` after accepting their email invite. Because `autoAcceptPendingCommitteeInvitesAfterLfidAccept` skips membership creation when the invite carries no org name (PR #965, fail-closed), the committee fetch returns 403. The page mapped 403 to "Group Not Found" with no path to accept — a catch-22.
- Added `pendingInvitationFromRoute` computed signal that resolves a pending invite using the route `:id` directly, bypassing the null `committee()` problem on 403.
- When the error state renders and a matching pending invite is found, the "Group Not Found" block is replaced with an invite accept UI ("You've been invited to {name}" + Accept/Decline). On accept, the existing `refreshCommitteeAfterMembershipChange` polls until membership propagates, then reloads the committee — clearing the error state and letting `syncEntityProjectContext` switch the active project to the correct one.

## Test plan

- [ ] Accept an email invite for an org-required committee (one with `enable_voting` or `business_email_required`) while not already a member
- [ ] Confirm the group detail page shows "You've been invited to {group name}" with Accept/Decline instead of "Group Not Found"
- [ ] Accept → org dialog appears → confirm → page transitions to the loaded committee view on the correct project
- [ ] Decline → "Group Not Found" state is shown (invite resolved, no access)
- [ ] Non-invited user landing on a restricted group UID still sees "Group Not Found" (no invite in cache)
- [ ] Normal invite banner on a successfully-loaded group is unaffected

Fixes: https://linuxfoundation.atlassian.net/browse/LFXV2-2507

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
@@ -42,7 +42,7 @@
     <!-- Error State -->
     <div class="flex justify-center items-center min-h-96">
       <div class="text-center max-w-md">
-        @if (errorType() === 'not-found') {
+        @if (errorType() === 'access-denied') {
           @if (pendingInvitationFromRoute(); as invite) {
             <div class="w-16 h-16 rounded-full bg-blue-50 flex items-center justify-center mx-auto mb-4">
               <i class="fa-light fa-user-plus text-2xl text-blue-600"></i>
@@ -55,13 +55,13 @@
                 severity="info"
                 icon="fa-light fa-check"
                 (onClick)="onAcceptInvite(invite)"
-                data-testid="committee-view-error-invite-accept" />
+                data-testid="committee-view-invitation-error-accept" />
               <lfx-button
                 label="Decline"
                 severity="secondary"
                 [outlined]="true"
                 (onClick)="onDeclineInvite(invite)"
-                data-testid="committee-view-error-invite-decline" />
+                data-testid="committee-view-invitation-error-decline" />
             </div>
           } @else {
             <div class="w-16 h-16 rounded-full bg-red-50 flex items-center justify-center mx-auto mb-4">
@@ -71,6 +71,13 @@
             <p class="text-sm text-gray-500 mb-6">The group you're looking for doesn't exist, has been removed, or you may not have permission to view it.</p>
             <lfx-button label="Back to Groups" severity="secondary" [outlined]="true" icon="fa-light fa-arrow-left" (onClick)="goBack()"></lfx-button>
           }
+        } @else if (errorType() === 'not-found') {
+          <div class="w-16 h-16 rounded-full bg-red-50 flex items-center justify-center mx-auto mb-4">
+            <i class="fa-light fa-exclamation-triangle text-2xl text-red-500"></i>
+          </div>
+          <h2 class="text-lg font-semibold text-gray-900 mb-2">Group Not Found</h2>
+          <p class="text-sm text-gray-500 mb-6">The group you're looking for doesn't exist, has been removed, or you may not have permission to view it.</p>
+          <lfx-button label="Back to Groups" severity="secondary" [outlined]="true" icon="fa-light fa-arrow-left" (onClick)="goBack()"></lfx-button>
         } @else {
           <div class="w-16 h-16 rounded-full bg-amber-50 flex items-center justify-center mx-auto mb-4">
             <i class="fa-light fa-cloud-exclamation text-2xl text-amber-500"></i>
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
@@ -43,12 +43,34 @@
     <div class="flex justify-center items-center min-h-96">
       <div class="text-center max-w-md">
         @if (errorType() === 'not-found') {
-          <div class="w-16 h-16 rounded-full bg-red-50 flex items-center justify-center mx-auto mb-4">
-            <i class="fa-light fa-exclamation-triangle text-2xl text-red-500"></i>
-          </div>
-          <h2 class="text-lg font-semibold text-gray-900 mb-2">Group Not Found</h2>
-          <p class="text-sm text-gray-500 mb-6">The group you're looking for doesn't exist, has been removed, or you may not have permission to view it.</p>
-          <lfx-button label="Back to Groups" severity="secondary" [outlined]="true" icon="fa-light fa-arrow-left" (onClick)="goBack()"></lfx-button>
+          @if (pendingInvitationFromRoute(); as invite) {
+            <div class="w-16 h-16 rounded-full bg-blue-50 flex items-center justify-center mx-auto mb-4">
+              <i class="fa-light fa-user-plus text-2xl text-blue-600"></i>
+            </div>
+            <h2 class="text-lg font-semibold text-gray-900 mb-2">You've been invited to {{ invite.committee_name }}</h2>
+            <p class="text-sm text-gray-500 mb-6">Accept the invitation to join and view this group.</p>
+            <div class="flex items-center justify-center gap-3">
+              <lfx-button
+                label="Accept Invitation"
+                severity="info"
+                icon="fa-light fa-check"
+                (onClick)="onAcceptInvite(invite)"
+                data-testid="committee-view-error-invite-accept" />
+              <lfx-button
+                label="Decline"
+                severity="secondary"
+                [outlined]="true"
+                (onClick)="onDeclineInvite(invite)"
+                data-testid="committee-view-error-invite-decline" />
+            </div>
+          } @else {
+            <div class="w-16 h-16 rounded-full bg-red-50 flex items-center justify-center mx-auto mb-4">
+              <i class="fa-light fa-exclamation-triangle text-2xl text-red-500"></i>
+            </div>
+            <h2 class="text-lg font-semibold text-gray-900 mb-2">Group Not Found</h2>
+            <p class="text-sm text-gray-500 mb-6">The group you're looking for doesn't exist, has been removed, or you may not have permission to view it.</p>
+            <lfx-button label="Back to Groups" severity="secondary" [outlined]="true" icon="fa-light fa-arrow-left" (onClick)="goBack()"></lfx-button>
+          }
         } @else {
           <div class="w-16 h-16 rounded-full bg-amber-50 flex items-center justify-center mx-auto mb-4">
             <i class="fa-light fa-cloud-exclamation text-2xl text-amber-500"></i>
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -167,6 +167,16 @@ export class CommitteeViewComponent {
     return findPendingInvitationForCommittee(this.invitationService.pendingInvitations(), this.invitationService.resolvedInviteUids(), committee.uid);
   });
 
+  // When the committee 403s and committee() is null, pendingInvitation() returns null because it
+  // requires committee.uid. This signal resolves the same invite using the route :id directly so
+  // the error state can surface the accept flow without a loaded committee record.
+  public pendingInvitationFromRoute: Signal<PendingInvitation | null> = computed(() => {
+    if (!this.error()) {
+      return null;
+    }
+    return findPendingInvitationForCommittee(this.invitationService.pendingInvitations(), this.invitationService.resolvedInviteUids(), this.committeeId());
+  });
+
   // Deferred-decline timers keyed by invite UID (committee UID stored alongside since the invite is
   // out of the cache by the time the timer/destroy flush fires). Mirrors the dashboard/My Groups UX.
   private readonly pendingDeclines = new Map<string, { committeeUid: string; timerId: ReturnType<typeof setTimeout> }>();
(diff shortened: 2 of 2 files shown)


apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
```diff
@@ -43,12 +43,34 @@
...
+            </div>
+            <h2 class="text-lg font-semibold text-gray-900 mb-2">You've been invited to {{ invite.committee_name }}</h2>
+            <p class="text-sm text-gray-500 mb-6">Accept the invitation to join and view this group.</p>
+            <div class="flex items-center justify-center gap-3">
+              <lfx-button
+                label="Accept Invitation"
+                severity="info"
+                icon="fa-light fa-check"
+                (onClick)="onAcceptInvite(invite)"
+                data-testid="committee-view-error-invite-accept" />
```
audigregorie: **Minor — testid naming inconsistency**

Every other invitation testid in this template uses `invitation` (e.g. `committee-view-invitation-accept`, `committee-view-invitation-decline`, `committee-view-invitation-banner`, `committee-view-invitation-undo`). These new error-state testids use `invite`, which diverges from the established pattern and hurts test-grep discoverability. Suggest aligning while still differentiating the error-state surface from the loaded-committee banner.

```suggestion
                data-testid="committee-view-invitation-error-accept" />
```

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
```diff
@@ -43,12 +43,34 @@
...
+              <lfx-button
+                label="Decline"
+                severity="secondary"
+                [outlined]="true"
+                (onClick)="onDeclineInvite(invite)"
+                data-testid="committee-view-error-invite-decline" />
```
audigregorie: **Minor — testid naming inconsistency**

Same note as the Accept button above — align with the `committee-view-invitation-*` pattern used everywhere else in this template.

```suggestion
                data-testid="committee-view-invitation-error-decline" />
```

audigregorie (commented):
## Code Review Summary

Clean, focused fix for LFXV2-2507 — adds a route-id-based pending-invitation lookup so the 403 error state can surface Accept/Decline instead of the dead-end "Group Not Found". Reactive, reuses existing accept/decline plumbing, properly gated on `error()` so it never conflicts with the loaded-committee banner. No critical or major findings.

### Minor — outside the diff

**Brief "Group Not Found" flash on the open-invite accept path.** For invites that don't require an organization, `onAcceptInvite` calls `markResolved(invite.uid)` _before_ the accept request fires (`committee-view.component.ts` ~line 426). That removes the invite from the shared cache, so `pendingInvitationFromRoute()` flips to `null` while `error()` is still `true`, briefly rendering the "Group Not Found" UI between accept-click and the committee refresh resolving. Consider deferring `markResolved` to the success branch in this code path (or rendering a small "Accepting…" state) so the user doesn't see the dead-end copy mid-flow. Org-required invites are unaffected — `markResolved` is already deferred to success there.

### What's done well

- Reactive `computed()` properly gated on `error()` so it never conflicts with the loaded-committee `pendingInvitation` banner.
- Reuses the existing `findPendingInvitationForCommittee` helper, which already null-guards `committeeUid`.
- Route id read via the existing `committeeId()` signal — no duplicate `route.paramMap` subscription.
- Constructor's `loadPendingInvitations()` populates the shared cache on direct landings, so the invite resolves reactively even when the committee fetch 403s first.

**Rating:** ⭐⭐⭐⭐⭐

audigregorie (commented):
Follow-up finding from re-reading against `.claude/rules/component-organization.md`.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -167,6 +167,16 @@ export class CommitteeViewComponent {
...
 
+  // When the committee 403s and committee() is null, pendingInvitation() returns null because it
+  // requires committee.uid. This signal resolves the same invite using the route :id directly so
+  // the error state can surface the accept flow without a loaded committee record.
+  public pendingInvitationFromRoute: Signal<PendingInvitation | null> = computed(() => {
+    if (!this.error()) {
+      return null;
+    }
+    return findPendingInvitationForCommittee(this.invitationService.pendingInvitations(), this.invitationService.resolvedInviteUids(), this.committeeId());
+  });
```
audigregorie: **Major — `init*()` convention for multi-line computed signals**

`.claude/rules/component-organization.md` requires multi-line `computed()` / `toSignal()` to be extracted into a private `init*()` initializer grouped at the bottom of the class. CLAUDE.md's top-violations list reinforces this and explicitly notes: _"Apply these even when neighboring legacy code does not — match `.claude/rules/`, not copy-paste from the file you are editing."_

The pre-existing `pendingInvitation` computed a few lines above has the same violation, but that's not a justification to add another. Suggest extracting:

```suggestion
  public pendingInvitationFromRoute: Signal<PendingInvitation | null> = this.initPendingInvitationFromRoute();
```

and adding alongside the other initializers at the bottom of the class:

```ts
private initPendingInvitationFromRoute(): Signal<PendingInvitation | null> {
  return computed(() => {
    if (!this.error()) {
      return null;
    }
    return findPendingInvitationForCommittee(
      this.invitationService.pendingInvitations(),
      this.invitationService.resolvedInviteUids(),
      this.committeeId()
    );
  });
}
```

luismoriguerra:
## Audit — PR #1024

**Verdict: 🟡 Minor comments** — `merge-after-acknowledging-the-403/404-gate-nit`. No bug, regression, secret, or contract break; all gates green.

Audited at head `3362b05` in an isolated worktree, grounded in the surrounding `committee-view` code and the shared invitation utils.

### Verification actually run
| Gate | Result |
|---|---|
| `yarn check-types` | ✅ pass |
| `yarn lint` | ✅ pass |
| `yarn format:check` | ✅ pass |
| `yarn build` | ✅ pass (only a pre-existing `fullcalendar.scss` budget warning, unrelated to this PR) |
| `yarn e2e` (Playwright) | not run — no `committee-view` e2e spec exists in the repo; this surface has never been e2e-covered |

`lfx-v2-ui` has no unit tests by design, so coverage is judged only against Playwright e2e. Deps are unchanged by this PR, so the gates above were run against the PR head with the repo's installed dependencies.

### Why
Invitees to **org-required** committees (`enable_voting` / `business_email_required`) who click their email invite land on `/project/groups/:id`. Since #965 made the auto-accept fail-closed (skips membership creation when the invite carries no org name), the committee GET returns **403** → mapped to `errorType='not-found'` → "Group Not Found" with no path to accept. A catch-22.

### How
- New `pendingInvitationFromRoute` computed signal (`committee-view.component.ts:173`) resolves the invite via `findPendingInvitationForCommittee(...)` keyed on the **route `:id`** (`committeeId()`) rather than `committee().uid`, which is null on a 403. The shared util is null-safe (`invitation.utils.ts:126`), so a null route id degrades to `null` cleanly.
- The `not-found` template branch now renders an Accept/Decline block when that signal resolves, falling back to the original "Group Not Found" otherwise.
- Accept reuses the existing `onAcceptInvite` → `InvitationAcceptFlowService.accept()`; its `committee()?.name ?? invite.committee_name` and `committee()?.uid ?? committeeId()` fallbacks already handle the null-committee case, and `refreshCommitteeAfterMembershipChange` polls until `my_role` surfaces, then clears the error state. Mechanism matches the conventions established in #880/#977.

### Is it correct
Yes for the targeted flow. The accept/decline/undo/destroy-flush plumbing is unchanged and already null-committee-safe. SSR-safe (invite cache load is browser-gated). No security-relevant surface touched (the invite read is the user's own; manager-only fetches stay gated). CodeRabbit: no actionable comments. Copilot raised one valid comment (below).

### Gaps / findings — all non-blocking
1. **403 vs 404 share `errorType='not-found'` (confirms Copilot's inline comment).** `initializeCommittee` maps both 403 and 404 to `'not-found'` (`committee-view.component.ts:718`), and `pendingInvitationFromRoute` only gates on `this.error()`. So a **true 404** (deleted/nonexistent group) where the user happens to hold a *stale cached invite* for that uid would render the accept UI; accepting then fails downstream into an "Unable to Accept" toast. Real but low-likelihood (requires an orphaned invite for a gone committee) and self-limiting (error toast; decline still works). Cleanest fix: track the HTTP status / split `errorType` so this block renders for 403 only. Worth a follow-up, not a merge blocker.
2. **Signal slightly broader than its use.** `pendingInvitationFromRoute` keys on `this.error()` rather than `errorType()==='not-found'`; it's only consumed inside the `not-found` branch so it's functionally fine, but tightening it would also resolve finding #1.
3. **No e2e for this surface.** No `committee-view` Playwright spec exists at all, so this isn't a coverage regression, but the new accept/decline-in-error path (testids `committee-view-error-invite-accept` / `-decline` are present) is unverified by automation.

Secrets / critical-constants: clean — no secrets, env vars, timeouts, or new constants (reuses existing). Architecture: clean — pure presentational change reusing existing services/util; no ADR violated, superseded, or newly introduced.

### Q&A this PR resolves
- **Why a second invite signal instead of reusing `pendingInvitation()`?** That one needs `committee().uid`, which is null on a 403; the route-keyed signal bypasses the null record.
- **Does accept work without a loaded committee?** Yes — `onAcceptInvite` falls back to `invite.committee_name` / `invite.committee_uid` and `committeeId()`, and the post-accept poll clears the error.
- **Will decline leave the user stuck?** No — decline marks the invite resolved, the signal recomputes to `null`, and the view falls back to "Group Not Found".

### Business rules
- Org-required committee invitees who 403 on the group page get an in-place Accept/Decline path instead of a dead end.
- A resolved (accepted/declined) invite is excluded from the error-state surface immediately (`resolvedInviteUids`), in sync with the dashboard / My Groups banners.

### ADRs
None added, changed, or violated.

---
*Report-only audit. The single substantive finding (403/404 gate) is already threaded inline by Copilot, so it isn't duplicated as a separate inline comment.*

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
```diff
@@ -43,12 +43,34 @@
...
-          <h2 class="text-lg font-semibold text-gray-900 mb-2">Group Not Found</h2>
-          <p class="text-sm text-gray-500 mb-6">The group you're looking for doesn't exist, has been removed, or you may not have permission to view it.</p>
-          <lfx-button label="Back to Groups" severity="secondary" [outlined]="true" icon="fa-light fa-arrow-left" (onClick)="goBack()"></lfx-button>
+          @if (pendingInvitationFromRoute(); as invite) {
+            <div class="w-16 h-16 rounded-full bg-blue-50 flex items-center justify-center mx-auto mb-4">
+              <i class="fa-light fa-user-plus text-2xl text-blue-600"></i>
+            </div>
+            <h2 class="text-lg font-semibold text-gray-900 mb-2">You've been invited to {{ invite.committee_name }}</h2>
+            <p class="text-sm text-gray-500 mb-6">Accept the invitation to join and view this group.</p>
+            <div class="flex items-center justify-center gap-3">
```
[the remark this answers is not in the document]
MRashad26: Done — split `errorType` into distinct `'not-found'` (404) and `'access-denied'` (403) values. The invite accept block is now gated on `errorType() === 'access-denied'`, so a stale invite for a deleted/nonexistent group (true 404) will never reach the accept flow.

See commit c89371110: split 403 vs 404 errorType + gate initPendingInvitationFromRoute on access-denied.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
MRashad26: Fixed — renamed to `committee-view-invitation-error-accept`. Good catch on the pattern consistency.

See commit c89371110: rename data-testids to follow committee-view-invitation-* convention.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
MRashad26: Fixed — renamed to `committee-view-invitation-error-decline`.

See commit c89371110: rename data-testids to follow committee-view-invitation-* convention.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
MRashad26: Done — extracted into `private initPendingInvitationFromRoute()` at the bottom of the class alongside the other `init*()` initializers. The call-site is now a clean one-liner: `public pendingInvitationFromRoute = this.initPendingInvitationFromRoute();`. Also took the opportunity to update the guard from `!this.error()` to `this.errorType() !== 'access-denied'` (aligned with the 403/404 split from the Copilot comment).

See commit c89371110.

MRashad26:
## Review Feedback Addressed

Commit: c89371110

### Changes Made
- **committee-view.component.ts**: split `errorType` union into distinct `'not-found'` (404) and `'access-denied'` (403) values; `initPendingInvitationFromRoute` now gates on `errorType() === 'access-denied'` — prevents the invite accept UI from rendering on a true 404 (per @copilot-pull-request-reviewer)
- **committee-view.component.ts**: extracted `pendingInvitationFromRoute` computed body into `private initPendingInvitationFromRoute()` at the bottom of the class, per `.claude/rules/component-organization.md` (per @audigregorie)
- **committee-view.component.html**: renamed `data-testid` attributes to `committee-view-invitation-error-accept` / `committee-view-invitation-error-decline`, aligning with the `committee-view-invitation-*` pattern used across the template (per @audigregorie)

### No Change Needed
- **Brief "Group Not Found" flash on non-org-required invite accept** (noted by @audigregorie, outside diff): acknowledged — `markResolved` fires optimistically before the accept request for non-org-required invites, which briefly shows "Group Not Found" while `error()` is still true. The flash is real but transient (resolves as soon as `refreshCommitteeAfterMembershipChange` reloads the committee). Deferring to a follow-up if it proves noticeable in QA — fixing it here would require restructuring the `onAcceptInvite` success flow, which is a larger touch than this bug warrants.

### Threads Resolved
4 of 4 unresolved threads addressed.

luismoriguerra (approved):

audigregorie (approved):

PR #1052, opened 2026-07-01
fix(impersonation): prevent CF auth loop and add target user CF token

lewisojile:
## Summary

Fixes an auth redirect loop that occurred when an admin impersonated a user and navigated to `/crowdfunding` pages. The error presented was:

```json
{"error": "Authentication required to access this resource", "code": "AUTHENTICATION_REQUIRED"}
```

### Root Causes (three simultaneous)

1. **SSR redirect triggered before Angular booted** — the catch-all handler called the CF auth-code redirect on every `/crowdfunding` page load during impersonation, before Angular had a chance to render anything.
2. **Admin's CF token loaded under impersonation** — `extractCrowdfundingToken` loaded the real (admin's) CF token into the impersonation session, returning the wrong user's data or triggering a confusing upstream 401.
3. **401 CF_UNAUTHENTICATED triggered Angular redirect loop** — `cfFetch` returned `401 CF_UNAUTHENTICATED` which is the exact condition the Angular `handleCfError` handler uses to call `redirectToCfAuth()`, creating a loop.

### Fix

**Phase 1 (no Auth0 changes required — ships immediately):**
- Guard the SSR CF auth-code redirect with `!isImpersonating`
- Return `503 CF_UNAVAILABLE_DURING_IMPERSONATION` from `cfFetch`/`cfFetchNullable` when no CF token is available during impersonation — Angular's `handleCfError` falls through to `of(fallback)` (empty state) instead of redirecting
- `extractCrowdfundingToken` returns early during impersonation, never loading the admin's CF token

**Phase 2 (best-effort — shows target user's real CF data when Auth0 is configured):**
- During `startImpersonation`, attempt a second audience-scoped CTE for `CROWDFUNDING_API_AUDIENCE` using the same NATS subject
- If Auth0 is configured to allow cross-audience exchange, the target user's CF token is stored in the session (`impersonationCrowdfundingToken` / `impersonationCrowdfundingExpiresAt`)
- If Auth0 returns `access_denied` (not yet configured), the `try/catch` logs a warning, impersonation still starts, and CF shows empty state (Phase 1 fallback)

### Auth Service Dependency

The Phase 2 CTE requires `lfx-v2-auth-service` to accept an optional `audience` field on `lfx.auth-service.impersonation.token_exchange`. That change has been implemented in a separate PR in that repo. If the auth service is deployed without that change, the CF CTE NATS call will still succeed (the auth service silently ignores unknown fields) but will issue a primary LFX V2 token scoped to the wrong audience — the token will be rejected by the CF API and the BFF will fall through to Phase 1 empty state. No breakage.

### Auth0 Tenant Configuration (for Phase 2 to show live CF data)

See `cf-impersonation-solution.md` §9 for the two Auth0 tenant changes needed:
1. Authorize `AUTH0_M2M_CLIENT_ID` for the CF API with scopes `openid profile access:me offline_access`
2. Add a CF audience branch to the CTE Action

## Testing

- [ ] Admin impersonates user → navigates to `/crowdfunding` → no redirect loop, no console auth errors
- [ ] CF sections show empty state (Phase 1) or target user's CF data (Phase 2)
- [ ] Admin stops impersonation → own CF data returns normally
- [ ] Non-impersonating user with no CF token still gets redirected to CF auth (no regression)

## Related

- JIRA: [LFXV2-2367](https://linuxfoundation.atlassian.net/browse/LFXV2-2367)
- Auth service PR: _(link once open)_

[LFXV2-2367]: https://linuxfoundation.atlassian.net/browse/LFXV2-2367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/controllers/impersonation.controller.ts
+++ b/apps/lfx-one/src/server/controllers/impersonation.controller.ts
@@ -65,15 +65,15 @@ export class ImpersonationController {
         throw new Error('Failed to decode target user claims from CTE response');
       }
 
-      // Attempt a CF-audience CTE so the impersonated session can access the target user's
-      // Crowdfunding data. This is best-effort: if the exchange fails (e.g. the target user
-      // has no CF account or Auth0 denies the audience), impersonation still starts and CF
-      // sections fall back to empty state (Phase 1 behaviour).
-      const cfAudience = process.env['CROWDFUNDING_API_AUDIENCE'];
+      // Attempt a CF-scoped CTE so the impersonated session can access the target user's
+      // Crowdfunding data. The admin's CF token is used as the subject_token so the auth
+      // service can extract the CF audience from it. Best-effort: if the exchange fails
+      // (e.g. admin has no CF token, or auth service rejects it), impersonation still
+      // starts and CF sections fall back to empty state.
       let cfTokenResponse: M2MTokenResponse | undefined;
-      if (cfAudience) {
+      if (req.crowdfundingToken) {
         try {
-          cfTokenResponse = await this.impersonationService.exchangeToken(req, targetUser.trim(), cfAudience);
+          cfTokenResponse = await this.impersonationService.exchangeToken(req, targetUser.trim(), req.crowdfundingToken);
           logger.debug(req, 'start_impersonation', 'CF CTE exchange succeeded for target user');
         } catch (cfErr) {
           logger.warning(req, 'start_impersonation', 'CF CTE exchange failed — CF will show empty state during impersonation', {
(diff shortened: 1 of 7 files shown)


dealako:
@lewisojile, do you still need to support these changes? If so, let's address the AI review feedback. You'll also need to resolve the conflicts. Thanks.

lewisojile:
Thanks, I will close this current ticket and have the issue addressed from the auth service.

PR #1053, opened 2026-07-01
fix(events): only show Download Certificate for attended past events

dealako:
## Summary

In the **Me Lens → Events → Past** events table, the "Download Certificate" button was rendered for every past-event row regardless of the user's registration status. It should only appear when the user's status is **Attended**.

This PR gates the button on attendance, adds a matching server-side authorization guard so the certificate API can't be bypassed directly, and centralizes the My Events status strings into a shared constant so the UI tag map, server status derivation, and the button gate stay in sync.

## Root cause

`events-table.component.html` gated the certificate cell only on `isPastEvents()`, with no check on `event.status`. Attendance was already available on each row via `MyEvent.status`, derived server-side in `mapRowToEvent` from the Snowflake `USER_ATTENDED` column on `ANALYTICS.PLATINUM_LFX_ONE.EVENT_REGISTRATIONS`.

## Changes

- **Shared:** add `MY_EVENT_STATUS` constant + `MyEventStatus` type (`packages/shared/src/constants/events.constants.ts`); add `USER_ATTENDED` to `CertificateEventRow`.
- **Server mapper:** `mapRowToEvent` now uses `MY_EVENT_STATUS.*` instead of bare string literals.
- **UI gate:** the Download Certificate button renders only when `event.status === MY_EVENT_STATUS.ATTENDED`; the `<td>` is still emitted for all past rows to preserve column alignment. `statusSeverityMap` keys reference the constant.
- **API defense-in-depth:** the certificate query selects `USER_ATTENDED`, and `generateCertificate` throws a `403 AuthorizationError` when it is falsy — mirroring the truthy semantics of `mapRowToEvent` so the UI and API gates cannot diverge.

## Validation

image

## Test plan

- [ ] Past events with status Registered / Speaker-only show **no** Download Certificate button.
- [ ] Past events with status Attended still show the button and download succeeds.
- [ ] Direct `GET /api/events/certificate?eventId=...` for a non-attended event returns 403.
- [ ] Visual validation on the feature-branch deploy preview.

Refs: LFXV2-2566


dealako:
Thanks @coderabbitai — addressed the review nitpicks:

1. **`events.service.ts` — type `status` as `MyEventStatus`** ✅ Done in `7a664d98`. `mapRowToEvent`'s `status` is now typed as the shared `MyEventStatus` union so the compiler catches drift from `MY_EVENT_STATUS`.

2. **`events-table.component.html` — empty-state placeholder in the certificate cell** — Intentionally declined. That column is an action-button column, not a data column: it renders a button only when the action is applicable (mirroring how the non-past "Action" column renders nothing when there's no action), so a blank cell is the correct affordance. The Role column's `—` is appropriate there because it's a data field. Left as-is.

audigregorie (dismissed):

packages/shared/src/constants/events.constants.ts
```diff
@@ -31,6 +31,19 @@ export const MY_EVENT_STATUS_OPTIONS: FilterOption[] = [
...
+ * Derived server-side in EventsService.mapRowToEvent and consumed by the UI
+ * (status tag + Download Certificate gating). Keep in sync with the table's statusSeverityMap.
+ */
+export const MY_EVENT_STATUS = {
+  ATTENDED: 'Attended',
+  REGISTERED: 'Registered',
+  NOT_REGISTERED: 'Not Registered',
+} as const;
+
+export type MyEventStatus = (typeof MY_EVENT_STATUS)[keyof typeof MY_EVENT_STATUS];
```
MRashad26: 🟡 **Warning — `MyEventStatus` type should live in `@lfx-one/shared/interfaces`, not in `@lfx-one/shared/constants`**

The project rule requires ALL interfaces and types (including derived type aliases) to live in `@lfx-one/shared/interfaces`. `MyEventStatus` is currently exported from `events.constants.ts`, which means `events.service.ts` imports a type from the constants package:

```ts
// ❌ Current — type imported from constants
import { MY_EVENT_STATUS, MyEventStatus, ... } from '@lfx-one/shared/constants';
```

Move the type to `packages/shared/src/interfaces/events.interface.ts` and re-export it:

```ts
// packages/shared/src/interfaces/events.interface.ts
import { MY_EVENT_STATUS } from '@lfx-one/shared/constants';

export type MyEventStatus = (typeof MY_EVENT_STATUS)[keyof typeof MY_EVENT_STATUS];
```

Then imports become:
```ts
// ✅ Type from interfaces, constant from constants
import { MY_EVENT_STATUS } from '@lfx-one/shared/constants';
import type { MyEventStatus } from '@lfx-one/shared/interfaces';
```
Non-blocking — the cert-gate fix is correct. Fine to address in a follow-up.

MRashad26 (commented):
## Code Review — PR #1053 fix(events): only show Download Certificate for attended past events

### Overview
Tightly scoped fix with three coordinated layers: (1) UI gate — `Download Certificate` button now renders only when `event.status === MY_EVENT_STATUS.ATTENDED`; (2) defense-in-depth API guard — `certificate.service.ts` throws `AuthorizationError` 403 when `!eventRow.USER_ATTENDED`, preventing direct API bypass; (3) constant centralization — `MY_EVENT_STATUS` + `MyEventStatus` added to shared constants so UI tags, server mapper, and button gate all read from one source of truth.

### Secrets / critical-constants check ✅
No secrets introduced. Snowflake column names (`USER_ATTENDED`) are internal identifiers, not credentials.

### Code-standards audit

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🟡 Warning | PR body contains `## Test plan` section — not used in lfx-self-serve |
| 2 | 🟡 Warning | `MyEventStatus` type lives in `events.constants.ts` — should be in `events.interface.ts` |

### Code quality notes
- `@if (event.status === MY_EVENT_STATUS.ATTENDED)` — direct property comparison, not a component method call. Passes the no-functions-in-templates rule. ✅
- `protected readonly MY_EVENT_STATUS = MY_EVENT_STATUS` — correct pattern for exposing a shared constant to the template. ✅
- `[MY_EVENT_STATUS.REGISTERED]` computed keys in `statusSeverityMap` — string literals replaced with constant references throughout. ✅
- `AuthorizationError` used for the 403 (not `ResourceNotFoundError`) — correct semantics. ✅
- `CertificateEventRow.USER_ATTENDED` added to shared interfaces (`events.interface.ts`). ✅
- `let status: MyEventStatus` narrows the type in the server mapper — prevents future string drift from untyped assignment. ✅
- `USER_ATTENDED` added to the Snowflake SELECT before the guard, not as an afterthought. ✅

### Verdict: PASS ✅ (2 warnings, no criticals)
No blocking issues. The `MyEventStatus` type placement is a minor convention deviation — worth fixing in a follow-up rather than blocking the cert-gate fix.

luismoriguerra:
## 🔍 PR Audit — `fix(events): only show Download Certificate for attended past events`

**Verdict: ✅ Approved — `safe-to-merge`.** Small, correct, well-layered fix. All local gates green; no blocking findings. One non-blocking convention nit remains (already threaded inline).

### Why
In **Me Lens → Events → Past**, the Download Certificate button rendered on every past-event row regardless of attendance, so registered-but-not-attended users saw (and could use) a certificate download. This gates the button on `Attended` and adds a matching server-side guard so the API can't be bypassed directly.

### Is it correct — yes (verified)
- **Auth model sound.** `getCertificate` derives `userEmail` from the session (`getEffectiveEmail(req)`), never client input, and the query is scoped by `EVENT_ID + USER_EMAIL`. The new guard genuinely closes the direct-API hole — hitting `GET /api/events/certificate?eventId=…` for a non-attended registration now returns `403` instead of a PDF.
- **Truthy check is deliberate/documented** (`!eventRow.USER_ATTENDED`) to mirror `mapRowToEvent` regardless of whether Snowflake returns `1` or `true`.
- Prior bot nits resolved: CodeRabbit's `MyEventStatus` typing applied (`7a664d98`); empty-cell placeholder intentionally declined (action-button column).

### Verification (isolated worktree @ `941128d2`)
- `yarn check-types` ✅
- `yarn lint` ✅
- `prettier --check` on all 6 changed files ✅
- Test gate for this repo is Playwright e2e — **not run** (heavy; not required for this diff). No unit tests in this repo by design.

### Gaps (all non-blocking)
1. 🟡 **Type placement.** `MyEventStatus` is declared in `constants/events.constants.ts`; repo convention is that types/interfaces live in `interfaces/` even inside `packages/shared/`. Already flagged inline by @MRashad26 and agreed as a follow-up. Fix = move the alias to `events.interface.ts`, import the constant there.
2. 🔵 **No My Events e2e coverage** for the certificate gating. Not a regression — the button never had e2e coverage — but a follow-up e2e asserting the button is absent for Registered/Speaker rows would lock in the fix.
3. 🔵 **Minor guard/UI asymmetry (informational).** `mapRowToEvent` requires `IS_PAST_EVENT && USER_ATTENDED` for "Attended"; the API guard checks only `USER_ATTENDED`. Practically inconsequential (attendance implies the event occurred) — not worth adding a column for.

### Secrets / critical-constants
Clean. No credentials/keys/tokens; `USER_ATTENDED` is a Snowflake column identifier. No timeouts, limits, NATS subjects, FGA relations, or flag defaults changed.

### Architecture / ADRs
No cross-cutting decision touched — BFF/UI presentation-gating over an existing Snowflake read. No ADR added, violated, or superseded.

---
*Report-only automated audit. Findings verified against code at head `941128d2`.*

packages/shared/src/constants/events.constants.ts
dealako: Done — moved `MyEventStatus` to `packages/shared/src/interfaces/events.interface.ts` (deriving from `MY_EVENT_STATUS` via a direct constants import), removed it from `events.constants.ts`, and updated `events.service.ts` to import the type from `@lfx-one/shared/interfaces`. Verified type-check, lint, and build pass with no circular-dependency issue.

See commit a2c37636. Thanks for the catch!

dealako:
## Review Feedback Addressed

Commit: a2c37636

### Changes Made
- **`packages/shared/src/interfaces/events.interface.ts`**: added the `MyEventStatus` type alias (derived from `MY_EVENT_STATUS` via a direct constants import). (per @MRashad26)
- **`packages/shared/src/constants/events.constants.ts`**: removed the `MyEventStatus` type alias; the `MY_EVENT_STATUS` value stays here. (per @MRashad26)
- **`apps/lfx-one/src/server/services/events.service.ts`**: `MyEventStatus` now imported from `@lfx-one/shared/interfaces` (value `MY_EVENT_STATUS` still from `@lfx-one/shared/constants`). (per @MRashad26)

Types now live in `@lfx-one/shared/interfaces` per the project convention. Verified `check-types`, `lint`, and `build` all pass — no circular-dependency issue between the constants and interfaces barrels.

### Threads Resolved

mrashad-26 (approved):

MRashad26 (approved):

PR #1058, opened 2026-07-02
feat(crowdfunding): add sponsorship tiers settings tab

emlimlf:
## Summary

- Adds a "Sponsorship Tiers" tab to the initiative settings drawer, letting projects switch between fixed sponsorship tiers or open donations, and configure per-tier goals and benefits.
- This is UI-only for now — the backend passes the data through as-is (`sponsorshipTiers`/`donationMode`) without processing it upstream yet; no real tier data is served, only mock/pass-through data.
- Gated behind the `crowdfunding-sponsor-tiers` LaunchDarkly flag, currently only enabled on dev and local — not visible in production until the upstream API and design are finalized.

## Notes

- LFXV2-2535


apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
```diff
@@ -0,0 +1,80 @@
...
+                <label class="text-sm font-medium text-gray-700">Benefits</label>
+                <div class="flex flex-col gap-2" formArrayName="benefits">
+                  @for (benefitControl of tierData.benefits.controls; track $index; let bIdx = $index) {
+                    <div class="flex items-center gap-2">
+                      <input
+                        pInputText
+                        [formControlName]="bIdx"
+                        [placeholder]="'Benefit ' + (bIdx + 1)"
+                        class="flex-1 p-inputtext-sm"
+                        [attr.data-testid]="'settings-tier-benefit-' + i + '-' + bIdx" />
```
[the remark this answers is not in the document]
emlimlf: Done, added `[attr.aria-label]="'Benefit ' + (bIdx + 1)"` to the benefit input, matching the pattern already used in `settings-funding-tab.component.html`.

See commit 3303bf2c0.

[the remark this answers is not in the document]
emlimlf: Done, added `[attr.aria-label]="'Benefit ' + (bIdx + 1)"` to the benefit input so it has an accessible name beyond the placeholder.

See commit 3303bf2c0.

emlimlf:
## Review Feedback Addressed

Commit: 3303bf2c0f0ad1e4ed284a69a78a170b34acc5f1

### Changes Made
- **settings-sponsorship-tiers-tab.component.html**: added `[attr.aria-label]` to the benefit input so it has an accessible name beyond the placeholder, matching the pattern in `settings-funding-tab.component.html` (per copilot-pull-request-reviewer, coderabbitai)
- **crowdfunding.controller.ts**: `sponsorshipTiers`/`donationMode` now throw `ServiceValidationError` on invalid values instead of silently dropping them, matching the existing `status` field validation (per coderabbitai nitpick)

### No Change Needed
- **settings-sponsorship-tiers-tab.component.ts:29**: coderabbitai's nitpick suggested dropping the explicit `: FormGroup` annotation to restore typed-forms inference. Tried it — breaks the build (`TS2345` in `getValue()`) because full inference also requires re-typing `makeTierGroup`'s return value; the one-line diff alone doesn't achieve the claimed benefit. Not pursuing given the wider blast radius for a trivial nitpick.

### Threads Resolved

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.ts
```diff
@@ -0,0 +1,103 @@
...
+          (saved?.benefits.length ? saved.benefits : ['']).forEach((b) => benefits.push(new FormControl(b, { nonNullable: true }), { emitEvent: false }));
+        });
+      });
+  }
+
+  public getValue(): Pick<UpdateInitiativeInput, 'donationMode' | 'sponsorshipTiers'> {
+    const value = this.form.getRawValue();
+    return {
+      donationMode: value.donationMode,
+      sponsorshipTiers: value.tiers.map(
```
dealako, line 76: [question] getValue() always emits full tier data, even in "open" mode

Issue: `getValue()` unconditionally maps all four tiers (goal + benefits) regardless of `donationMode`. When the user picks "open", the template hides the tier UI (`.html:20`), but the stale/hidden tier config is still shipped as `sponsorshipTiers`. Same applies to a tier the user enabled, filled in, then disabled — its `goalCents`/`benefits` still ship.
Proof: line 72-78 maps `value.tiers` with no guard on `value.donationMode` or `t.enabled`; `initiative-settings-drawer.component.ts:150-153` merges the result into the save payload with no filtering.
Why it matters: the payload won't match what the user sees on screen. Low impact today because the PR documents the backend as pure pass-through/forward-compat, but it becomes a real data-integrity issue the moment upstream starts consuming these fields.
Fix: either gate emission on `donationMode === 'tier'` (and optionally drop disabled tiers), or add a code comment stating both modes always send the full tier set by design so the intent is explicit before the flag is promoted.

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.ts
```diff
@@ -0,0 +1,103 @@
...
+    const benefits = (this.form.get('tiers') as FormArray<FormGroup>).at(tierIndex).get('benefits') as FormArray<FormControl<string>>;
+    if (benefits.length > 1) {
+      benefits.removeAt(benefitIndex);
+    }
+  }
+
+  private makeTierGroup(tier: SponsorshipTier): FormGroup {
+    return new FormGroup({
+      enabled: new FormControl<boolean>(tier.enabled, { nonNullable: true }),
+      goal: new FormControl<number | null>(null),
```
dealako: [minor] Tier goal control has no client-side validator, unlike the sibling funding tab

Issue: `makeTierGroup` creates `goal: new FormControl<number | null>(null)` with no validators. The sibling `settings-funding-tab.component.ts:35` applies `[Validators.min(0)]` and the drawer blocks save on `funding.form.invalid`. Here, the drawer's `onSave` never checks the sponsorship form's validity at all.
Proof: `initiative-settings-drawer.component.ts` onSave validates details/funding/beneficiaries but reads `sponsorshipTiersTab()?.getValue()` without any `.form.invalid` guard; server-side `parseNonNegativeInt` silently floors a negative to `undefined`.
Why it matters: a user can type a negative goal and get no feedback — it's silently dropped server-side rather than surfaced, diverging from the funding tab's UX contract.
Fix: add `[Validators.min(0)]` to the goal control to match the sibling, and consider guarding save on the sponsorship form's validity.

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
```diff
@@ -0,0 +1,81 @@
...
+          @if (tierData.group.get('enabled')?.value) {
+            <div class="flex flex-col gap-3 pl-6" [attr.data-testid]="'settings-tier-fields-' + i">
+              <div class="flex flex-col gap-1.5 max-w-[240px]">
+                <label [for]="'tier-goal-' + i" class="text-sm font-medium text-gray-700">Sponsorship Goal</label>
+                <lfx-input-number [form]="tierData.group" control="goal" [id]="'tier-goal-' + i" placeholder="0" [dataTest]="'settings-tier-goal-' + i" />
+              </div>
+              <div class="flex flex-col gap-1.5">
+                <label class="text-sm font-medium text-gray-700">Benefits</label>
+                <div class="flex flex-col gap-2" formArrayName="benefits">
+                  @for (benefitControl of tierData.benefits.controls; track $index; let bIdx = $index) {
```
dealako: [nit] `track $index` on an editable, removable benefits list

Issue: the benefits `@for` tracks by `$index`. Removing a middle benefit reindexes the FormArray while DOM nodes are keyed by position, which can cause transient input-value/focus shuffling as controls shift down.
Proof: `removeBenefit(i, bIdx)` calls `benefits.removeAt(benefitIndex)`, so every control after the removed one changes index while `track $index` keeps the old DOM nodes.
Why it matters: minor — lists are short and this matches other list patterns in the repo. Take-it-or-leave-it.
Fix: `track benefitControl` for stable identity.

dealako (dismissed):
Hi @emlimlf 👋 — thorough, well-scoped PR. The sponsorship-tiers tab cleanly mirrors the sibling settings-tab architecture (per-tab component + `getValue()` consumed by the drawer's aggregate save), the shared string-union-in-interfaces + `as const`-tuple-in-constants split is honored, and the whole feature is correctly gated behind `crowdfunding-sponsor-tiers` (dev/local only) with a documented pass-through backend. I especially liked that you pre-empted the CodeRabbit nitpick and added explicit `ServiceValidationError` validation for `sponsorshipTiers`/`donationMode` in the controller rather than silently dropping bad input.

I ran a parallel security audit and a code-quality pass alongside my own review. **No blockers found.** The cents round-trip (`goalCents/100` → `Math.round(goal*100)`) is numerically safe, the `undefined` vs `0` semantics are coherent end-to-end, FormArray reactivity via `toSignal(valueChanges)` is sound, and the server path is authenticated, rate-limited, allowlist-validated, and free of injection/secret/PII issues.

### Issue count

- 🔴 **Blocking: 0**
- 🟡 **Minor: 1** — tier goal `FormControl` has no `Validators.min(0)` and the drawer never validates the sponsorship form on save, unlike the funding tab (negative goal silently dropped server-side).
- ⚪ **Nit: 2** — `track $index` on the editable/removable benefits list; zero automated tests (repo-wide pattern, not a regression).
- ❔ **Question: 1** — `getValue()` emits full tier data (and disabled-tier config) even in `open` mode; harmless today under pass-through, but confirm this is intended before the flag is promoted.

### Security note (non-blocking)
No cap on `sponsorshipTiers` count / `benefits` count / benefit string length before forwarding upstream. Bounded in practice by the global `express.json({ limit: '15mb' })` and consistent with the existing unbounded `beneficiaries`/`goals` handling, so it's inherited defense-in-depth, not a regression. Worth a cap (dedupe tiers by name, cap benefits, per-benefit `maxLength`) once upstream starts consuming the data.

### AI bot reconciliation
- **CodeRabbit** (invalid-value validation) & **Copilot** (benefit `aria-label`) — both already addressed in `3303bf2`. Confirmed correct. ✅
- **CodeRabbit** (bare `FormGroup` loses typed-forms inference) — I **agree with your decision to defer**: full inference cascades into re-typing `makeTierGroup` and diverges from every sibling untyped-form tab. Consistency-preserving trade-off; fine as-is, worth a follow-up if the tabs are ever migrated to typed forms together.

### Decision
✅ **Approved with comments** — nothing here blocks merge given the feature-flag gating and documented pass-through backend. Please just weigh in on the open `open`-mode emission question and consider the `Validators.min(0)` alignment before the flag graduates beyond dev/local.

apps/lfx-one/src/server/controllers/crowdfunding.controller.ts
```diff
@@ -292,6 +299,31 @@ export class CrowdfundingController {
...
+            operation: 'update_initiative',
+          });
+        }
+        input.sponsorshipTiers = rawTiers.map((t) => ({
+          name: t['name'] as SponsorshipTierName,
+          enabled: t['enabled'] === true,
+          goalCents: parseNonNegativeInt(t['goalCents']),
+          benefits: Array.isArray(t['benefits']) ? (t['benefits'] as unknown[]).filter((b): b is string => typeof b === 'string') : [],
+        }));
+      }
```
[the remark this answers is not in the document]
emlimlf: Fixed — the `.find()` predicate now checks each tier entry is a non-null object before indexing into it, so `[null]`/`[1]` returns a validation error (400) instead of throwing a 500.

See commit fbc1add2b: `crowdfunding.controller.ts` guards `typeof t !== 'object'` before the name lookup.

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.ts
emlimlf: Good catch on the ambiguity. Went with documenting the intent rather than gating emission — added a comment on `getValue()` stating both donation modes ship the full tier set by design, since the backend is pure pass-through/forward-compat today and gating would risk clearing saved tier config server-side in 'open' mode without a corresponding product decision. Happy to revisit gating once upstream actually consumes these fields.

See commit fbc1add2b: `settings-sponsorship-tiers-tab.component.ts`.

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.ts
emlimlf: Fixed — added `Validators.min(0)` to the tier goal control to match the funding tab, and `onSave` now also checks the sponsorship form's validity (and calls `markAllAsTouched()` on it) before saving, so a negative goal blocks the save with visible feedback instead of being silently floored server-side.

See commit fbc1add2b: `settings-sponsorship-tiers-tab.component.ts` and `initiative-settings-drawer.component.ts`.

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
emlimlf: Fixed, switched to `track benefitControl` for stable identity. Good catch!

See commit fbc1add2b: `settings-sponsorship-tiers-tab.component.html`.

emlimlf:
## Review Feedback Addressed

Commit: fbc1add2b9a853fa4b4f52a6c2930863cfef22e1

### Changes Made
- **server/controllers/crowdfunding.controller.ts**: guard `sponsorshipTiers` entries against non-object/null values before indexing, so malformed input returns a 400 validation error instead of a 500 (per copilot[bot])
- **settings-sponsorship-tiers-tab.component.ts**: added `Validators.min(0)` to the tier goal control, matching the sibling funding tab (per @dealako)
- **initiative-settings-drawer.component.ts**: `onSave` now also validates the sponsorship form (and marks it touched) before saving, so a negative goal blocks the save with feedback (per @dealako)
- **settings-sponsorship-tiers-tab.component.html**: benefits list now tracks by `benefitControl` instead of `$index` for stable identity when removing items (per @dealako)

### No Change Needed (documented instead)
- **settings-sponsorship-tiers-tab.component.ts (getValue)**: added a comment documenting that both donation modes intentionally ship the full tier set for forward-compat pass-through, rather than gating emission — the backend is pure pass-through today, and gating would risk clearing saved tier config server-side without a corresponding product decision (per @dealako)

### Threads Resolved

apps/lfx-one/src/server/controllers/crowdfunding.controller.ts
```diff
@@ -292,6 +299,34 @@ export class CrowdfundingController {
         }));
       }
 
+      if (Array.isArray(body['sponsorshipTiers'])) {
+        const rawTiers = body['sponsorshipTiers'] as unknown[];
+        const invalidTier = rawTiers.find(
```
dealako: [blocking] The new falsy-guard uses `.find()` + truthiness, so it still 500s on the exact inputs it was meant to fix

Issue: the guard added in `fbc1add2` uses `rawTiers.find(...)` then `if (invalidTier)`. `find()` returns the matched **element**, not a boolean. When the offending element is itself falsy, the predicate correctly matches (`!t`), but `find()` returns that falsy value and `if (invalidTier)` evaluates the *element's* truthiness — which is falsy — so the throw is skipped and execution falls through to `.map()`.

Proof (traced against this diff):
- `{sponsorshipTiers:[null]}` → `find` returns `null` → `if(null)` false → no throw → `.map()` runs `null['name']` → **TypeError → 500** (the commit claimed to turn this into a 400).
- `{sponsorshipTiers:[undefined]}` → same → **500**.
- `{sponsorshipTiers:[{name:'gold'}, null]}` → `find` returns the `null` at index 1 → no throw → **500** in `.map()`.
- `{sponsorshipTiers:[0]}` / `[false]` / `['']` → `find` returns the falsy primitive → no throw → `.map()` auto-boxes, `(0)['name']` is `undefined` → a bogus `{name: undefined}` tier is **silently accepted and forwarded upstream** (validation bypass).

Why it matters: the change advertises a 500→400 correctness fix but regresses on its own target inputs, and additionally lets malformed tiers with no name reach the upstream PATCH body.

Fix: use `.some()` (returns a boolean, immune to falsy elements):
```ts
const hasInvalidTier = rawTiers.some(
  (t) => !t || typeof t !== 'object' || !SPONSORSHIP_TIER_NAMES.includes((t as Record<string, unknown>)['name'] as (typeof SPONSORSHIP_TIER_NAMES)[number])
);
if (hasInvalidTier) {
  throw ServiceValidationError.forField('sponsorshipTiers', ...);
}
```

Confirmed independently by @Copilot's inline comment on this same guard (below). Agreed — same root cause, same one-word fix.

dealako (changes requested):
Hi @emlimlf 👋 — thanks for the fast, thorough turnaround. You cleanly closed out all four items from my last pass, and I appreciate the care in the "no change needed" write-up too.

### 👏 Nice work
- **`Validators.min(0)` + save-time validation** — you didn't just add the validator; `onSave` now checks `sponsorshipTiersTab?.form.invalid` and calls `markAllAsTouched()` on it, so a negative goal actually blocks the save with feedback. Exactly the funding-tab parity I was hoping for, with the correct `?.` guard for the flag-gated optional viewChild.
- **`track benefitControl`** — stable identity on the removable list. ✅
- **Open-mode emission** — documenting the intent (both modes ship the full tier set for forward-compat pass-through) rather than gating is a defensible call, and the code comment makes it explicit for the next reader. Good reasoning about not clearing saved config server-side without a product decision. ✅

### Revision tracking (prior round)
- ✅ **[minor] Missing goal validator / no save validation** — Resolved in `fbc1add2`.
- ✅ **[nit] `track $index` on benefits** — Resolved in `fbc1add2`.
- ✅ **[question] open-mode tier emission** — Resolved (documented as intended) in `fbc1add2`.

### New issue in this revision
The hardening you added for malformed tiers (Copilot's earlier ask) has a subtle but real regression:

- 🔴 **Blocking: 1** — the new `rawTiers.find(...)` + `if (invalidTier)` guard uses `find()` (returns the matched *element*), so a **falsy** invalid entry defeats the check. `{sponsorshipTiers:[null]}` and `[undefined]` still fall through to `.map()` and throw a **500** — the exact 500→400 case the commit claimed to fix — and `[0]`/`[false]`/`['']` **bypass validation entirely**, injecting a `{name: undefined}` tier into the upstream PATCH body. One-word fix: `.find` → `.some`. Full trace in the inline comment.

### AI bot reconciliation
- **@Copilot** independently flagged this identical `find` vs `some` bug on the same guard (inline, at line 308). **I agree** — same root cause, same fix. This is the only substantive open item.

### Everything else still clean
The cents round-trip, `undefined`/`0` semantics, FormArray reactivity, auth/rate-limit/allowlist posture, and PII handling all remain sound from the prior review. No merge conflicts (`MERGEABLE`).

### Decision
🔴 **Needs changes before approval** — solely for the `.find` → `.some` fix on the tier guard. It's a genuine one-liner; flip it (and optionally add a quick check that `[null]`/`[0]` now yield a 400) and this is good to go. Nice work overall.

apps/lfx-one/src/server/controllers/crowdfunding.controller.ts
```diff
@@ -292,6 +299,34 @@ export class CrowdfundingController {
...
+          (t) =>
+            !t || typeof t !== 'object' || !SPONSORSHIP_TIER_NAMES.includes((t as Record<string, unknown>)['name'] as (typeof SPONSORSHIP_TIER_NAMES)[number])
+        );
+        if (invalidTier) {
```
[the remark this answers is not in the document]
emlimlf: Done, replaced `.find()` + truthiness check with `.some()` so falsy tier elements (`null`, `undefined`, `0`, `''`) are correctly rejected instead of falling through to `.map()`.

See commit 54e685885: apps/lfx-one/src/server/controllers/crowdfunding.controller.ts now uses `rawTiers.some(...)` to compute `hasInvalidTier`.

apps/lfx-one/src/server/controllers/crowdfunding.controller.ts
emlimlf: Confirmed, good catch. Replaced `.find()` + truthiness check with `.some()` per your suggested fix, so this now correctly rejects falsy tier elements instead of 500ing or silently accepting a bogus tier.

See commit 54e685885: apps/lfx-one/src/server/controllers/crowdfunding.controller.ts.

emlimlf:
## Review Feedback Addressed

Commit: 54e685885

### Changes Made
- **apps/lfx-one/src/server/controllers/crowdfunding.controller.ts**: replaced `rawTiers.find(...)` + truthiness check with `rawTiers.some(...)` so falsy tier elements (`null`, `undefined`, `0`, `''`) are correctly rejected instead of causing a 500 (via `.map()` on a falsy element) or silently bypassing validation (per @dealako, copilot-pull-request-reviewer[bot])

### Threads Resolved

packages/shared/src/constants/crowdfunding.constants.ts
```diff
@@ -13,6 +13,9 @@ import {
   MyDonationsResponse,
   RecurringDonation,
   RecurringDonationsResponse,
+  SponsorshipDonationMode,
```
MRashad26, line 16: 🟡 **Warning — `import type` missing for type-only symbols**

`SponsorshipDonationMode`, `SponsorshipTier`, and `SponsorshipTierName` are used only as type annotations in this file (not as runtime values). They must be imported with `import type` to follow the project convention and prevent circular-dependency risk.

**Suggested fix:**
```typescript
import type {
  SponsorshipDonationMode,
  SponsorshipTier,
  SponsorshipTierName,
} from '../interfaces/crowdfunding.interface';
```
Split from the value imports above, or consolidate the existing `import type` block if one already exists in this file.

packages/shared/src/constants/crowdfunding.constants.ts
```diff
@@ -109,6 +112,19 @@ export const MAX_LOGO_SIZE_BYTES = 2 * 1024 * 1024;
 // Must stay in sync with CrowdfundingInitiativeStatus in crowdfunding.interface.ts.
 export const CROWDFUNDING_INITIATIVE_STATUSES = ['submitted', 'pending', 'published', 'declined', 'hidden'] as const;
 
+// Runtime-checkable tuple of every valid sponsorship tier name — used for server-side input validation.
+// Must stay in sync with SponsorshipTierName in crowdfunding.interface.ts.
+export const SPONSORSHIP_TIER_NAMES = ['platinum', 'gold', 'silver', 'bronze'] as const;
```
MRashad26, line 115: 🟡 **Warning — manual type/constant sync (LFXV2-2570 convention)**

`SPONSORSHIP_TIER_NAMES` here and `SponsorshipTierName` in the interface file are manually kept in sync — the "Must stay in sync" comment is a code smell that the LFXV2-2570 refactor (PR #1055) was specifically designed to eliminate. Same issue applies to `SPONSORSHIP_DONATION_MODES` / `SponsorshipDonationMode` (line 121).

**Suggested fix** — derive the type from the constant instead of duplicating it:
```typescript
// crowdfunding.constants.ts (unchanged)
export const SPONSORSHIP_TIER_NAMES = ['platinum', 'gold', 'silver', 'bronze'] as const;

// crowdfunding.interface.ts — replace the manual union
import type { SPONSORSHIP_TIER_NAMES, SPONSORSHIP_DONATION_MODES } from '../constants/crowdfunding.constants';
export type SponsorshipTierName = (typeof SPONSORSHIP_TIER_NAMES)[number];
export type SponsorshipDonationMode = (typeof SPONSORSHIP_DONATION_MODES)[number];
```
This makes the types structurally guaranteed by the constant, and removes the need for the sync comments.

MRashad26 (commented):
## PR #1058 — `feat(crowdfunding): add sponsorship tiers settings tab`

**Overview:** Adds `SettingsSponsorshipTiersTabComponent` — a new settings tab for configuring crowdfunding sponsorship tiers. Manages a reactive `FormGroup` + nested `FormArray`, wires into the drawer via `viewChild()`, and adds BFF endpoints for reading/saving donation settings. New `SponsorshipTier*` types/constants are placed correctly in `@lfx-one/shared`.

**Secrets / sensitive data:** None found.

**Angular 20 audit:**
- ✅ `viewChild()` (lowercase) used correctly in the parent drawer component
- ✅ No `effect()` — uses `toObservable(visible).pipe(filter(Boolean), takeUntilDestroyed()).subscribe()` correctly
- ✅ New types in `@lfx-one/shared/interfaces`; runtime constants in `@lfx-one/shared/constants` with no `export type`
- ✅ BFF validates all inputs (`SPONSORSHIP_TIER_NAMES.includes()`, `SPONSORSHIP_DONATION_MODES.includes()`, `parseNonNegativeInt()`)
- 🟡 3 warnings — see inline comments

**Verdict: ✅ PASS with warnings**

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
```diff
@@ -0,0 +1,81 @@
...
+    <lfx-select-button
+      [form]="form"
+      control="donationMode"
+      [options]="donationModeOptions"
+      optionLabel="label"
+      optionValue="value"
+      data-testid="settings-donation-mode" />
+  </div>
+
+  @if (form.get('donationMode')?.value === 'tier') {
```
MRashad26: 🟡 **Warning — `FormGroup.get()` called in `@if` condition**

`form.get('donationMode')?.value` calls the `FormGroup.get()` method directly in the template. This is re-evaluated on every change-detection cycle and bypasses the reactive signal layer already established by `formValue = toSignal(form.valueChanges)`.

**Suggested fix** — expose a dedicated computed signal in the component:
```typescript
protected readonly donationMode = computed(() => this.formValue().donationMode as SponsorshipDonationMode);
```
Then in the template:
```html
@if (donationMode() === 'tier') {
```

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
```diff
@@ -0,0 +1,81 @@
...
+          [class.border-b]="!last"
+          [class.border-gray-200]="!last"
+          [attr.data-testid]="'settings-tier-' + tierNames[i]">
+          <lfx-checkbox
+            [form]="tierData.group"
+            control="enabled"
+            [label]="tierLabels[tierNames[i]]"
+            labelClass="text-sm font-medium text-gray-900 cursor-pointer"
+            [attr.data-testid]="'settings-tier-toggle-' + i" />
+          @if (tierData.group.get('enabled')?.value) {
```
MRashad26: 🟡 **Warning — `FormGroup.get()` called in `@if` condition**

`tierData.group.get('enabled')?.value` calls `FormGroup.get()` inside a template `@if`, re-evaluated on every CD cycle. The `tiersData` computed already tracks `formValue()` — `enabled` can be pre-resolved there.

**Suggested fix** — include `enabled` in the `tiersData` computed:
```typescript
protected readonly tiersData = computed(() => {
  this.formValue();
  return (this.form.get('tiers') as FormArray<FormGroup>).controls.map((group) => ({
    group,
    benefits: group.get('benefits') as FormArray<FormControl<string>>,
    enabled: group.get('enabled')?.value as boolean,
  }));
});
```
Then in the template:
```html
@if (tierData.enabled) {
```

apps/lfx-one/src/server/services/crowdfunding.service.ts
```diff
@@ -322,6 +323,12 @@ export class CrowdfundingService {
     if (input.beneficiaries !== undefined) {
       body.beneficiaries = input.beneficiaries.map((b): BackendBeneficiaryInput => ({ name: b.name, email: b.email }));
     }
+    if (input.sponsorshipTiers !== undefined) {
+      body.sponsorship_tiers = input.sponsorshipTiers.map(
+        (t): BackendSponsorshipTierInput => ({ name: t.name, enabled: t.enabled, goal_amount_cents: t.goalCents, benefits: t.benefits })
+      );
+    }
+    if (input.donationMode !== undefined) body.donation_mode = input.donationMode;
```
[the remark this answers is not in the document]
emlimlf: Fixed, added `sponsorship_tiers` / `donation_mode` to `BackendInitiative` and mapped them back into `InitiativeDetail` in `mapToInitiativeDetail`. Good catch, this was silently dropping saved tier settings on every reload.

See commit 2211c6c.

packages/shared/src/constants/crowdfunding.constants.ts
emlimlf: Fixed, switched the interface import block to `import type`.

See commit 2211c6c.

packages/shared/src/constants/crowdfunding.constants.ts
emlimlf: Done, `SponsorshipTierName` and `SponsorshipDonationMode` now derive from `SPONSORSHIP_TIER_NAMES` / `SPONSORSHIP_DONATION_MODES` and the sync comments are removed. While in there, applied the same fix to `CrowdfundingInitiativeStatus` (same smell, two lines away) and deduped the mapper's local status-validation array against the shared constant.

See commit 2211c6c.

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
emlimlf: Fixed, added a `donationMode` computed signal and switched the template to `@if (donationMode() === 'tier')`.

See commit 2211c6c.

apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
emlimlf: Fixed, pre-resolved `enabled` in the `tiersData` computed and switched the template to `@if (tierData.enabled)`.

See commit 2211c6c.

emlimlf:
## Review Feedback Addressed

Commit: 2211c6c89157072e16c5f4644949b311e1302863

### Changes Made
- **apps/lfx-one/src/server/types/crowdfunding.types.ts** / **crowdfunding-mapper.ts**: `BackendInitiative` now carries `sponsorship_tiers` / `donation_mode`, and `mapToInitiativeDetail` maps them back into `InitiativeDetail` — saved tier settings now rehydrate on subsequent loads instead of being silently dropped (per copilot-pull-request-reviewer[bot])
- **packages/shared/src/constants/crowdfunding.constants.ts**: switched the interface import block to `import type` (per @MRashad26)
- **packages/shared/src/interfaces/crowdfunding.interface.ts**: `SponsorshipTierName` and `SponsorshipDonationMode` now derive from `SPONSORSHIP_TIER_NAMES` / `SPONSORSHIP_DONATION_MODES` instead of duplicating the union, removing the manual-sync smell (per @MRashad26). Applied the same fix to `CrowdfundingInitiativeStatus` and deduped the mapper's local status-validation array against the shared constant, since it was the identical pattern two lines away.
- **settings-sponsorship-tiers-tab.component.ts/.html**: replaced `FormGroup.get()` reads inside template `@if` conditions with computed signals (`donationMode`, `tiersData[].enabled`) (per @MRashad26)

### Threads Resolved

MRashad26 (approved):

apps/lfx-one/src/server/utils/crowdfunding-mapper.ts
```diff
@@ -26,15 +29,23 @@ import {
...
+  return CROWDFUNDING_INITIATIVE_STATUSES.includes(value as CrowdfundingInitiativeStatus) ? (value as CrowdfundingInitiativeStatus) : 'pending';
+}
+
+function toValidDonationMode(value: unknown): SponsorshipDonationMode | undefined {
+  return SPONSORSHIP_DONATION_MODES.includes(value as SponsorshipDonationMode) ? (value as SponsorshipDonationMode) : undefined;
+}
+
+function mapSponsorshipTier(t: BackendSponsorshipTierInput): SponsorshipTier | undefined {
+  if (!SPONSORSHIP_TIER_NAMES.includes(t.name as SponsorshipTier['name'])) return undefined;
+  return { name: t.name as SponsorshipTier['name'], enabled: t.enabled, goalCents: t.goal_amount_cents, benefits: t.benefits };
```
dealako, line 48: [minor] Read-mapper trusts upstream `benefits`/`enabled` without a default — latent crash when the tab renders

Issue: `mapSponsorshipTier` maps `benefits: t.benefits` directly. `BackendSponsorshipTierInput.benefits` is *typed* `string[]` but this is untrusted upstream JSON — if the service ever returns a tier object without a `benefits` array (or with `null`), `t.benefits` is `undefined` at runtime and flows into `SponsorshipTier.benefits`.
Proof: that value reaches the tab's patch loop at `settings-sponsorship-tiers-tab.component.ts:66`: `(saved?.benefits.length ? saved.benefits : ['']).forEach(...)` — `saved` is optional-chained but `.benefits.length` is **not**, so `saved.benefits === undefined` throws `TypeError: Cannot read properties of undefined (reading 'length')` and the tab fails to render.
Why it matters: not reachable today (the feature is flag-gated and upstream doesn't populate `sponsorship_tiers` yet), but this PR just wired up the read path, so the mapper is the right seam to harden before upstream starts returning data.
Fix: default at the mapping boundary —
```ts
return {
  name: t.name as SponsorshipTier['name'],
  enabled: t.enabled === true,
  goalCents: t.goal_amount_cents,
  benefits: Array.isArray(t.benefits) ? t.benefits : [],
};
```
(Optionally also relax the frontend to `saved?.benefits?.length` as belt-and-suspenders.)

dealako (approved):
Hi @emlimlf 👋 — great iteration. You not only fixed the blocking guard bug but went well beyond the minimum on this round.

### 👏 Nice work
- **`.find` → `.some` (my blocking finding)** — Resolved in `54e68588`. The guard now returns a real boolean, so `[null]`/`[undefined]`/`[0]` are correctly rejected with a 400 instead of 500ing or silently injecting `{name: undefined}`. Exactly right.
- **Closed the read-path loop** — `BackendInitiative` now carries `sponsorship_tiers`/`donation_mode` and `mapToInitiativeDetail` maps them back via `mapSponsorshipTier` + `toValidDonationMode`, both allowlist-validated with invalid entries filtered out. This turns the save→reload round-trip into something real rather than write-only. Nice catch by Copilot, cleanly handled.
- **Single source of truth for the unions** — deriving `CrowdfundingInitiativeStatus`, `SponsorshipTierName`, and `SponsorshipDonationMode` from the `as const` tuples via `(typeof X)[number]` (and dropping the "must stay in sync" comments + the duplicate `VALID_INITIATIVE_STATUSES` local) eliminates the drift risk entirely. Both cross-file imports are `import type`, so no runtime cycle. 👍
- **Template reactivity** — moving off `form.get(...)?.value` in `@if` to the `donationMode()` computed and a pre-resolved `tierData.enabled` is the correct zoneless pattern.

### Revision tracking
- ✅ **[blocking] `.find` vs `.some` tier guard** → Resolved (`54e68588`).
- ✅ All prior-round items (goal validator, save validation, `track benefitControl`, open-mode documentation) remain resolved.

### New issue this revision
- 🟡 **Minor: 1** — the new read-mapper (`mapSponsorshipTier`) trusts upstream `benefits` without a default; combined with the frontend patch's non-optional `saved?.benefits.length` (`settings-sponsorship-tiers-tab.component.ts:66`), an upstream tier missing its `benefits` array would `TypeError` and fail to render the tab. Not reachable today (flag-gated + upstream doesn't return these yet), but the read path is new here, so it's worth a `benefits: Array.isArray(t.benefits) ? t.benefits : []` at the mapping boundary. Inline detail above.

### AI + reviewer reconciliation
- **@Copilot** (write path forwarded but read path didn't map back) — addressed in `2211c6c8`. ✅
- **@MRashad26** (`import type`, type/constant sync, `FormGroup.get()` in `@if` ×2) — all addressed in `2211c6c8`, and MRashad26 has since **approved**. I agree with all four; no daylight between us.

### Decision
✅ **Approved with minor comments** — the blocker is gone and everything else is clean; no merge conflicts. The one remaining `benefits` default is a latent-only nicety you can fold in whenever, or safely defer until upstream actually serves tier data. Solid work across the whole review cycle. 🚀

PR #1062, opened 2026-07-02
fix(votes): keep early close tooltip for auto-ended polls (LFXV2-2574)

audigregorie:
# Summary

When ITX auto-ends a poll because all eligible voters have responded, the votes table and results drawer now show the early-close tooltip whenever `early_end_time` is present. Detection no longer compares `early_end_time` to the scheduled `end_time`, which could miss valid auto-ended polls.

### DIFF
--- a/apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
@@ -17,7 +17,7 @@ import {
   VoteResultsQuestion,
   VoteResultsResponse,
 } from '@lfx-one/shared/interfaces';
-import { getVoteEndedEarlyDetailTooltip, isVoteEndedEarly } from '@lfx-one/shared/utils';
+import { getVoteEndedEarlyDetailTooltip } from '@lfx-one/shared/utils';
 import { PollStatusLabelPipe } from '@pipes/poll-status-label.pipe';
 import { PollStatusSeverityPipe } from '@pipes/poll-status-severity.pipe';
 import { VoteService } from '@services/vote.service';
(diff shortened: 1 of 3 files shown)


luismoriguerra (approved):
## Audit — ✅ Approved (`safe-to-merge`)

Audited at head `18ca72fa` in an isolated worktree; gates re-run.

### Why
When ITX auto-ends a poll because all eligible voters responded, the votes table and results drawer must show the "ended early" tooltip. The old detection compared `early_end_time < end_time`, which **missed valid auto-ended polls** when the timestamps coincided or `end_time` was unavailable.

### How
`isVoteEndedEarly(vote)` is simplified to `Boolean(vote.early_end_time)` (dropping the `end_time` comparison + param). Both components keep calling the shared helper (single source of truth) and use `vote.early_end_time!` to format the tooltip. This matches the field's documented contract: `early_end_time` is *"Absent when the poll closed on schedule or is still active"* — so presence alone is a correct, complete signal.

### Is it correct — yes
- Semantics correct per the documented field contract; the old comparison was the actual bug.
- `vote.early_end_time!` is safe — the `isVoteEndedEarly` guard guarantees presence (TS just can't narrow through the call).
- `isVoteEndedEarly` has exactly two callers (the edited components); narrowing the `Pick<Vote,…>` param breaks nothing.
- Gates re-run: `check-types` ✅ · `lint` ✅ · `prettier --check` (changed files) ✅. `lfx-v2-ui` has no unit tests, so none expected.

### Gaps
None blocking. Optional nit (CodeRabbit): the inline `formatDate(...) + getVoteEndedEarlyDetailTooltip(...)` sequence in both components could be centralized into a single `getVoteEndedEarlyTooltip(vote)` helper in `poll.utils.ts`. Fine as-is.

### Bot reconciliation
- Copilot's inline "helper is now unused" comment reviewed the earlier commit `585c692`; follow-up commit `2d249587` restored `isVoteEndedEarly` as the single source of truth — that comment is **stale**.
- CodeRabbit nitpick (centralize helper) — non-blocking.

### ADRs / business rules
No ADRs added/changed/violated. Pure presentation logic. Clarifies the rule: a poll is "ended early" iff `early_end_time` is present.

**Verdict: ✅ Approved.**

dealako (approved):

PR #1071, opened 2026-07-07
fix(crowdfunding): wire sponsorship tiers to live backend contract

emlimlf:
## Summary
- Aligns the sponsorship-tiers settings tab (built in #1058) with the now-deployed backend from linuxfoundation/lfx-crowdfunding#213
- Fixes `donation_mode` enum mismatch (`'tier'` → `'tiers'`) that would have failed upstream validation on save
- Splits the sponsorship tier wire type into read (`minimum`) and write (`goal_amount_cents`) shapes — the previous single type silently dropped the saved goal on every read
- Fixes the settings tab so loaded tier data (goal, benefits, enabled state) renders immediately instead of requiring a manual checkbox toggle to refresh

## Notes
- Depends on linuxfoundation/lfx-crowdfunding#213 being deployed (confirmed live)

### DIFF
--- a/apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
+++ b/apps/lfx-one/src/app/modules/crowdfunding/initiative-detail/components/initiative-settings-drawer/components/settings-sponsorship-tiers-tab/settings-sponsorship-tiers-tab.component.html
@@ -17,7 +17,7 @@
       data-testid="settings-donation-mode" />
   </div>
 
-  @if (donationMode() === 'tier') {
+  @if (donationMode() === 'tiers') {
     <div class="flex flex-col" formArrayName="tiers" data-testid="settings-sponsorship-tiers-list">
       @for (tierData of tiersData(); track $index; let i = $index; let last = $last) {
         <div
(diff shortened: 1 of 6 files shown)


dealako:
Hey @emlimlf — nice, tight fix. 👋

## Overall impression

This is a clean, surgical bug fix. The core problem — the sponsorship tiers GET mapper was reading `goal_amount_cents` when the upstream response only ever sends `minimum` — is fixed correctly, and splitting the previously-shared type into a read shape (`BackendSponsorshipTier`) and a write shape (`BackendSponsorshipTierInput`) is the right call: it prevents the two directions of the contract from silently drifting again. I pulled the actual Go structs from the merged upstream PR (linuxfoundation/lfx-crowdfunding#213) and confirmed the field names/optionality match exactly what this PR assumes. The `donationMode` enum rename (`'tier'` → `'tiers'`) is applied completely — no stray singular references anywhere in the codebase. The `updateValueAndValidity({ emitEvent: true })` addition correctly fixes the "requires a manual toggle to render loaded data" bug without introducing a loop (the subscription is keyed off the `visible` input signal, not off `valueChanges`/`formValue`, so there's no feedback risk).

I ran this in parallel with a security-focused pass and a code-quality pass; both came back clean, consistent with CodeRabbit and Copilot's existing zero-comment reviews on this PR.

## Findings

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 1

**❔ [question] Should a tier with no configured minimum render as `0` or blank?**

The upstream `SponsorshipTier.Minimum` Go field has no `omitempty`, so a tier with no goal set is serialized as `minimum: 0`, not omitted. `mapSponsorshipTier` maps that straight through to `goalCents: 0`. Downstream, `settings-sponsorship-tiers-tab.component.ts` does `saved?.goalCents != null ? saved.goalCents / 100 : null` — since `0 != null` is `true`, a tier with no goal will render its input as `0` rather than blank.

This isn't a regression from this PR (the same field shape existed before), and it may well be the intended behavior (0 is a legitimate "no minimum" default). Flagging as a product/UX call rather than a bug — worth a quick gut-check before merge, otherwise no action needed.

## Decision

✅ **Approved with comments** — no blocking or minor issues; approving with the one open question above called out for your judgment.

dealako (approved):
Approved with comments — see summary comment above for one open product/UX question (non-blocking). No blocking, security, or minor findings from independent review, security-auditor, or code-reviewer-pro passes.

PR #1072, opened 2026-07-07
fix(newsletters): handle async send and stop duplicate-send retries

niravpatel27:
## Problem — LFXV2-2604

The AAIF ED newsletter send (500+ recipients) took 37–41s upstream against this BFF's hard 30s `AbortSignal.timeout`, so the UI reported failure for sends that actually delivered, left the Send button armed, and a retry re-delivered to the full list. An autosave landing mid-send also bumped the newsletter version and stranded the first (delivered) send in `draft`. Full incident analysis and Datadog evidence in [LFXV2-2604](https://linuxfoundation.atlassian.net/browse/LFXV2-2604).

## Changes

- **BFF:** `ApiClientService`/`MicroserviceProxyService` accept an optional per-request `timeoutMs` (`ApiRequestOptions`); the newsletter send call uses `NEWSLETTER_SEND_TIMEOUT_MS` (120s) so a pre-async upstream's synchronous fan-out can't be aborted mid-send. Timeout errors now report the effective value.
- **Shared types:** `NewsletterStatus` gains `'sending'` (the async upstream accepts sends with 202 + `status='sending'`); list-filter validation accepts it.
- **Frontend (`newsletter-manage`):**
  - autosave is suppressed while a send is in flight and while a failed send's status refetch is resolving (`resolvingSend`) — the version-bump race that stranded send #1;
  - `runSend` branches on `newsletter.status` in the response body (backward compatible with the old synchronous `'sent'` upstream) and always lands on the Sent tab;
  - on any send error the component **refetches the newsletter first** and only re-arms Send when the record is genuinely still a draft (with a refreshed version); `canSend` is disarmed throughout.

## Deliberate UX trade-off (product decision)

There is **no in-app progress indicator** for in-flight sends: no "Sending" tag, no polling, no live status transition. A `sending` newsletter appears on the Sent tab rendered like a sent one with a `—` date until it settles (the async window is typically minutes at most; row click opens analytics, which degrades gracefully upstream). Flagged by review; intentionally kept per the requested UX.

## Protected files

`apps/lfx-one/src/server/services/microservice-proxy.service.ts` — adds only the optional trailing `options` parameter threaded to the API client (default behavior unchanged for all existing callers). Requesting code-owner review for this touch.

## External references / deploy ordering

- Upstream root fix: linuxfoundation/lfx-v2-newsletter-service#32 — **deploy the Go service first**. New UI + old service is safe (body carries `status='sent'` → existing synchronous branch); old UI + new service is safe (2xx, shape-compatible body, server-side duplicate guard active).

## Testing

`yarn check-types`, `yarn lint`, `yarn build` green; license/format hooks pass. Manual flow verified against both response shapes. No existing Playwright specs assert the send toast, so none required updates.

JIRA: [LFXV2-2604](https://linuxfoundation.atlassian.net/browse/LFXV2-2604)

[LFXV2-2604]: https://linuxfoundation.atlassian.net/browse/LFXV2-2604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-2604]: https://linuxfoundation.atlassian.net/browse/LFXV2-2604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes newsletter send/delivery UX and optimistic-concurrency behavior on a user-facing path; mistakes could still allow duplicate sends or block legitimate retries, though the refetch-on-error pattern reduces that risk.
> 
> **Overview**
> Addresses **LFXV2-2604** duplicate delivery and false send failures when large sends outlast the BFF’s 30s upstream timeout or autosave races the send.
> 
> **BFF:** Adds optional per-request `timeoutMs` (`ApiRequestOptions`) through `ApiClientService` and `MicroserviceProxyService`; newsletter **send** uses **120s** (`NEWSLETTER_SEND_TIMEOUT_MS`). List API validation accepts **`sending`** as a status filter. Shared types document async send (`NewsletterStatus` includes `'sending'`).
> 
> **`newsletter-manage`:** Successful sends branch on **`newsletter.status`** (`sending` vs settled `sent`) and navigate to the Sent tab. On send errors, the UI **refetches the draft** and only re-enables Send when status is still **`draft`** (refreshes version for the next `If-Match`). **`canSend`** and **autosave** stay off during send and during post-error status resolution (`resolvingSend`), avoiding mid-send version bumps that previously stranded sends and invited retries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05f27fd3bae84ed42a93eda33c6957297318bc57. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
@@ -152,7 +152,9 @@ export class NewsletterManageComponent {
   public readonly subjectFilled = computed(() => (this.subjectValue() ?? '').trim().length > 0);
   public readonly bodyFilled = computed(() => stripHtml(this.bodyValue() ?? '').length > 0);
   public readonly audienceFilled = computed(() => (this.committeeUidsValue() ?? []).length > 0);
-  public readonly canSend = computed(() => this.audienceFilled() && this.subjectFilled() && this.bodyFilled() && this.hasContext() && !this.submitting());
+  public readonly canSend = computed(
+    () => this.audienceFilled() && this.subjectFilled() && this.bodyFilled() && this.hasContext() && !this.submitting() && !this.resolvingSend()
+  );
   public readonly canSendTest = computed(
     () => this.subjectFilled() && this.bodyFilled() && this.hasContext() && this.edEmail().length > 0 && !this.testSending()
   );
--- a/apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
@@ -383,28 +383,84 @@ export class NewsletterManageComponent {
         finalize(() => this.submitting.set(false))
       )
       .subscribe({
-        next: (result: NewsletterSendResult) => {
-          if (result.failed > 0) {
+        next: (result: NewsletterSendResult) => this.handleSendResponse(result),
+        error: (err: HttpErrorResponse) => this.handleSendError(err, id),
+      });
+  }
+
+  /**
+   * The upstream send is asynchronous: acceptance returns the newsletter in
+   * status='sending' (fan-out completes in a background job), while
+   * status='sent' means it settled synchronously (zero-recipient edge case, or
+   * a pre-async upstream deployment). Both land on the Sent tab — there is
+   * deliberately no in-app progress indicator.
+   */
+  private handleSendResponse(result: NewsletterSendResult): void {
+    if (result.newsletter.status === 'sending') {
+      const total = result.total_recipients;
+      this.messageService.add({
+        severity: 'info',
+        summary: 'Sending newsletter',
+        detail: `Your newsletter is being sent to ${total} ${total === 1 ? 'recipient' : 'recipients'}.`,
+      });
+      this.goToList('sent');
+      return;
+    }
+    if (result.failed > 0) {
+      this.messageService.add({
+        severity: 'warn',
+        summary: 'Sent with errors',
+        detail: `Delivered ${result.sent} of ${result.total_recipients}. ${result.failed} failed.`,
+        life: 8000,
+      });
+    } else {
+      this.messageService.add({
+        severity: 'success',
+        summary: 'Newsletter sent',
+        detail: `Delivered to ${result.sent} ${result.sent === 1 ? 'recipient' : 'recipients'}.`,
+      });
+    }
+    this.goToList('sent');
+  }
+
+  /**
+   * A send error is ambiguous: a timeout or 5xx may have raced a send the
+   * upstream actually accepted (or even completed), and a 409 means one is
+   * definitely in flight. Refetch the newsletter and branch on its real status
+   * instead of unconditionally re-arming Send — the previous handler did the
+   * latter, inviting the duplicate delivery in LFXV2-2604.
+   */
+  private handleSendError(err: HttpErrorResponse, id: string): void {
+    this.newsletterService
+      .getNewsletter(this.projectUid(), id)
+      .pipe(take(1))
+      .subscribe({
+        next: (newsletter) => {
+          if (newsletter.status === 'sent' || newsletter.status === 'sending') {
             this.messageService.add({
-              severity: 'warn',
-              summary: 'Sent with errors',
-              detail: `Delivered ${result.sent} of ${result.total_recipients}. ${result.failed} failed.`,
-              life: 8000,
+              severity: 'info',
+              summary: newsletter.status === 'sent' ? 'Newsletter sent' : 'Sending newsletter',
+              detail: newsletter.status === 'sent' ? 'Your newsletter was sent.' : 'Your newsletter is being sent.',
             });
-          } else {
-            this.messageService.add({
-              severity: 'success',
-              summary: 'Newsletter sent',
-              detail: `Delivered to ${result.sent} ${result.sent === 1 ? 'recipient' : 'recipients'}.`,
-            });
+            this.goToList('sent');
+            return;
           }
-          this.goToList('sent');
-        },
-        error: (err: HttpErrorResponse) => {
+          // Genuinely still a draft — the send did not go through. Refresh the
+          // version (the failed attempt or an earlier save may have bumped it)
+          // so the next attempt doesn't fail on a stale If-Match.
+          this.version.set(newsletter.version);
           this.messageService.add({
             severity: 'error',
             summary: 'Send failed',
             detail: err?.error?.message || err?.message || 'Could not send newsletter. Please try again.',
+          });
+        },
+        error: () => {
+          this.messageService.add({
+            severity: 'error',
+            summary: 'Send failed',
+            detail: 'Could not confirm the send status. Check the Sent tab before trying again.',
+            life: 8000,
           });
         },
       });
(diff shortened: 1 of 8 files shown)


packages/shared/src/interfaces/api.interface.ts
```diff
@@ -14,6 +14,15 @@ export interface ApiClientConfig {
   retryDelay?: number;
 }
 
+/**
+ * Per-request options for the server-side API client
+ * @description Overrides applied to a single upstream request
+ */
+export interface ApiRequestOptions {
+  /** Per-request timeout in milliseconds; falls back to the client's configured default (30s) */
```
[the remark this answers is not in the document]
niravpatel27: Done — reworded to "falls back to the client's configured default (`ApiClientConfig.timeout`, default 30s)" so the doc matches the actual fallback chain. See commit a1d4520e.

niravpatel27:
## Review Feedback Addressed

Commit: a1d4520e

### Changes Made
- **packages/shared/src/interfaces/api.interface.ts**: reworded the `ApiRequestOptions.timeoutMs` doc comment to state the fallback is the client's configured default (`ApiClientConfig.timeout`, default 30s) instead of implying a fixed 30s (per copilot-pull-request-reviewer)

### Threads Resolved

MRashad26 (commented):
## PR #1072 — `fix(newsletters): handle async send and stop duplicate-send retries`

**Overview:** Fixes LFXV2-2604 — the AAIF ED newsletter send (~500 recipients) exceeded the BFF's 30s `AbortSignal.timeout`, the UI reported failure, left Send armed, and a retry re-delivered. Two separate mechanisms caused the incident: (1) the hard 30s abort cut a still-running synchronous upstream fan-out; (2) on any send error, `canSend` immediately re-armed without verifying whether the upstream actually accepted the send. This PR fixes both: per-request timeout override (`NEWSLETTER_SEND_TIMEOUT_MS = 120s`) for the send call, and a refetch-before-rearm pattern in `handleSendError`. Also adds `'sending'` to `NewsletterStatus` for the new async 202 upstream, and suppresses autosave while a send or status-refetch is in flight (`resolvingSend` signal).

**Secrets / sensitive data:** None found.

**Angular 20 audit:**
- ✅ No template changes — no new method calls in HTML bindings
- ✅ `resolvingSend` is a private signal, read only in `computed()` and `filter()` — not from template
- ✅ `canSend = computed(() => … && !this.submitting() && !this.resolvingSend())` — signal reads ✅
- ✅ No `effect()` introduced
- ✅ `takeUntilDestroyed(this.destroyRef)` + `take(1)` + `finalize()` correctly compose on the nested `getNewsletter()` subscription in `handleSendError`
- ✅ `handleSendResponse` / `handleSendError` are private methods called from subscribe callbacks, not from templates

**Interface / constants placement:**
- ✅ `ApiRequestOptions` interface → `packages/shared/src/interfaces/api.interface.ts`
- ✅ `NewsletterStatus` type update → `packages/shared/src/interfaces/newsletter.interface.ts`
- ✅ `NEWSLETTER_SEND_TIMEOUT_MS` constant → `packages/shared/src/constants/newsletter.constants.ts`
- ✅ No `export type` in constants file; no interfaces in constants file

**BFF correctness:**
- ✅ `options?.timeoutMs ?? this.config.timeout` fallback preserves default behavior for all existing callers
- ✅ Timeout value threaded through to `executeRequest` so timeout error messages report the effective ms, not the config default
- ✅ Controller `statusParam` validation updated to allow `'sending'` for list filter
- ✅ `handleSendError` refreshes `this.version` from refetched newsletter before re-arming Send — prevents 409 If-Match on legitimate retry

**Logic correctness:**
- ✅ `handleSendResponse` branches on `newsletter.status === 'sending'` (new async upstream) vs `'sent'` (synchronous/legacy) — backward compatible
- ✅ Autosave filter: `!this.submitting() && !this.resolvingSend()` covers both the send window and the refetch gap
- ✅ `finalize(() => this.resolvingSend.set(false))` ensures signal resets even on component destroy

**Process:** 🟡 PR body uses `## Testing` instead of `## Test plan` — rename to match the PR template.

**Verdict: ✅ PASS — 1 process warning, 1 info (see inline comment)**

apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
```diff
@@ -383,30 +389,91 @@ export class NewsletterManageComponent {
...
+          // so the next attempt doesn't fail on a stale If-Match.
+          this.version.set(newsletter.version);
           this.messageService.add({
             severity: 'error',
             summary: 'Send failed',
             detail: err?.error?.message || err?.message || 'Could not send newsletter. Please try again.',
           });
         },
+        error: () => {
+          this.messageService.add({
```
MRashad26, line 470: 🔵 **Info — double-failure path re-arms Send with potentially stale version**

When both the send itself **and** the `getNewsletter()` refetch fail (network partition, intermittent 5xx), `finalize(() => this.resolvingSend.set(false))` runs and `canSend` re-arms — but `this.version` is not updated because there is no newsletter object to read from. If the user retries immediately (ignoring the "Check the Sent tab" guidance), the send goes out with the pre-failure version and may 409 on If-Match.

In practice this self-heals on a second retry: the 409 triggers `handleSendError` again, the refetch likely succeeds this time, and `this.version.set(newsletter.version)` corrects it. So it is not a duplicate-send risk — just a confusing extra failure.

The toast message already steers users toward the safe path. A small improvement would be to disarm Send explicitly until a manual page reload — but this is a genuine edge case (network partition during a refetch) and the current UX is acceptable.

```typescript
error: () => {
  // Optional hardening: keep canSend locked until the user reloads
  // by leaving resolvingSend true and showing a "Reload" action.
  this.messageService.add({
    severity: 'error',
    summary: 'Send failed',
    detail: 'Could not confirm the send status. Check the Sent tab before trying again.',
    life: 8000,
  });
},
```

niravpatel27: Great analysis — that's exactly the sequence: on a double failure the version can be stale, an immediate retry may 412, and the second pass through `handleSendError` self-heals it.

We considered keeping Send disarmed until reload and decided against it deliberately: after a transient blip (the most likely cause of a failed refetch), a permanently disabled Send strands the user with no recovery affordance unless they know to reload, while re-arming is safe on every axis that matters — a retry can never duplicate delivery (the upstream rejects any send unless the row is genuinely `draft`, via the optimistically-locked `draft→sending` transition), and the worst case is the one extra confusing 412 you describe, which the toast's "Check the Sent tab" guidance already mitigates. Trading a rare extra error toast for guaranteed recoverability felt like the right side of the trade-off.

No code change — happy to revisit if you feel strongly about a "Reload" action variant.

niravpatel27:

### Questions Answered / No Change Needed
- **newsletter-manage.component.ts:470**: acknowledged the double-failure stale-version analysis (asked by @MRashad26) — re-arming Send after a failed status refetch is a deliberate trade-off: retries are duplicate-safe (the upstream's optimistically-locked `draft→sending` transition rejects any send unless the row is genuinely a draft) and the 412-then-self-heal path recovers the version, whereas locking Send until reload would strand users after transient network blips. No code change.

### Threads Resolved

MRashad26 (approved):

PR #1076, opened 2026-07-08
perf(meetings): eliminate N+1 committee fetches and count endpoints for project meetings

MRashad26:
## Summary

Extends LFXV2-2120 (groups performance) to project-level meetings:

- **Win 2 (server)**: `getCommitteeNameMap` in `meeting.service.ts` previously made N sequential `getCommitteeById` calls (one per unique committee across all displayed meetings) — each of which triggered settings + membership + access checks. Replaced with a single batched query-service call using the OR-tag pattern (`filters_or: [uid:X, uid:Y, ...]`), returning only the `name` field needed. `getCommitteesByIds` in `committee.service.ts` changed from `private` to `public` to enable reuse.

- **Win 1 (frontend)**: The "Upcoming Meetings" and "Past Meetings" stat cards on the project meetings dashboard previously computed their counts from `rawFpUpcomingMeetings().length` / `rawFpPastMeetings().length` — which forced a full `expand()`-based paginated fetch (up to 10 pages) before the stat cards could show any number. Replaced with dedicated count endpoint calls that resolve in a single request. The raw paginated fetches still run for recurring series count, recordings count, and calendar view — but count cards now resolve independently on their own loading signals (`fpUpcomingCountLoading` / `fpPastCountLoading`).

  Added `GET /api/past-meetings/count` (backed by `getMeetingsCount` with `type=v1_past_meeting`) to mirror the existing `GET /api/meetings/count` endpoint.

## Test plan

- [ ] Navigate to a project meetings dashboard → "Upcoming Meetings" and "Past Meetings" stat cards should appear much faster (single count API call) while "Recurring Series" and "Recordings Available" still load at their normal pace
- [ ] Verify all 4 stat cards show correct values after loading
- [ ] Switch between "upcoming" and "past" tabs — counts should still be correct on each tab
- [ ] Click "Refresh" — all stat cards should reload
- [ ] Verify meetings with committees show correct committee names in the meeting cards
- [ ] Confirm no regressions in the "Me" lens (stat cards should remain unchanged)

Related: LFXV2-2120, builds on merged PR #1060

---

> [!NOTE]
> **Medium Risk**
> Touches meeting list enrichment and new count endpoints used for dashboard stats; failures degrade to zero counts or missing committee names rather than blocking the page, but wrong counts or labels would be user-visible.
> 
> **Overview**
> Improves **project/foundation meetings dashboard** load time by changing how stat numbers and committee labels are resolved.
> 
> **Foundation/project stat cards:** **Upcoming** and **Past** meeting counts no longer wait on paginated “fetch everything” lists (`rawFp*Meetings().length`). They call **`/api/meetings/count`** and the new **`/api/past-meetings/count`** (tagged by `project_uid`) with their own loading signals (`fpUpcomingCountLoading` / `fpPastCountLoading`). Recurring series, recordings, and calendar still use the full raw FP fetches and `fpStatsLoading`.
> 
> **Client API layer:** Project/committee count helpers share a private **`getCountByTag`** helper; **`getPastMeetingsCountByProject`** is added for the past tab.
> 
> **Server:** **`GET /past-meetings/count`** proxies to **`getMeetingsCount`** with `v1_past_meeting`, mirroring upcoming counts.
> 
> **Meeting list enrichment:** Committee names on meetings are loaded via **batched query-service** `filters_or` chunks (100 UIDs) with **`Promise.allSettled`** so one failed batch does not drop names from other batches—replacing per-committee **`getCommitteeById`** (and removing the **`CommitteeService`** dependency from server **`meeting.service`**). **`committee.service`** only reorders **`getCommitteesByIds`** in the file (still private).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8ea4c28cd8c33f07bb184086201fb1607c93890. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
@@ -132,6 +132,8 @@ export class MeetingsDashboardComponent {
   // Per-lens loading flags so Me/FP recording-count pipelines cannot clobber each other on lens switch.
   protected readonly meRecordingsCountLoading = signal(false);
   protected readonly fpRecordingsCountLoading = signal(false);
+  protected fpUpcomingCountLoading = signal(false);
+  protected fpPastCountLoading = signal(false);
 
   // Raw user meetings cached for client-side filtering (Me lens only)
   private rawUserMeetings: Signal<Meeting[]>;
@@ -228,11 +230,12 @@ export class MeetingsDashboardComponent {
     this.rawFpUpcomingMeetings = this.initializeRawFpUpcomingMeetings();
     this.rawFpPastMeetings = this.initializeRawFpPastMeetings();
 
-    // Foundation/Project lens stat cards (computed from raw FP signals, not paginated)
-    // Only look at the active tab's loading signal — inactive tab fetches are gated off.
+    // Foundation/Project lens stat cards
+    // fpStatsLoading gates recurring/recordings cards (backed by raw full-page fetches).
+    // fpUpcomingCount/fpPastCount use the count API directly — they resolve independently.
     this.fpStatsLoading = computed(() => (this.timeFilter() === 'past' ? this.fpPastLoading() : this.fpUpcomingLoading()));
-    this.fpUpcomingCount = computed(() => (this.activeLens() !== 'me' ? this.rawFpUpcomingMeetings().length : 0));
-    this.fpPastCount = computed(() => (this.activeLens() !== 'me' ? this.rawFpPastMeetings().length : 0));
+    this.fpUpcomingCount = this.initFpUpcomingCount();
+    this.fpPastCount = this.initFpPastCount();
     this.fpRecurringCount = computed(() => (this.activeLens() !== 'me' ? this.rawFpUpcomingMeetings().filter((m) => m.recurrence !== null).length : 0));
     this.fpRecordingsAvailableCount = this.initFpRecordingsAvailableCount();
 
@@ -777,6 +780,50 @@ export class MeetingsDashboardComponent {
       const projectWord = count === 1 ? 'project' : 'projects';
       return count > 0 ? `Across ${count} ${projectWord}` : '';
     });
+  }
+
+  private initFpUpcomingCount(): Signal<number> {
+    const project$ = toObservable(this.project);
+    const lens$ = toObservable(this.activeLens);
+
+    return toSignal(
+      combineLatest([project$, lens$, this.refresh$]).pipe(
+        switchMap(([project, lens]) => {
+          if (lens === 'me' || !project?.uid) {
+            return of(0);
+          }
+          if (!isPlatformBrowser(this.platformId)) {
+            this.fpUpcomingCountLoading.set(true);
+            return of(0);
+          }
+          this.fpUpcomingCountLoading.set(true);
+          return this.meetingService.getMeetingsCountByProject(project.uid).pipe(finalize(() => this.fpUpcomingCountLoading.set(false)));
+        })
+      ),
+      { initialValue: 0 }
+    );
+  }
+
+  private initFpPastCount(): Signal<number> {
+    const project$ = toObservable(this.project);
+    const lens$ = toObservable(this.activeLens);
+
+    return toSignal(
+      combineLatest([project$, lens$, this.refresh$]).pipe(
+        switchMap(([project, lens]) => {
+          if (lens === 'me' || !project?.uid) {
+            return of(0);
+          }
+          if (!isPlatformBrowser(this.platformId)) {
+            this.fpPastCountLoading.set(true);
+            return of(0);
+          }
+          this.fpPastCountLoading.set(true);
+          return this.meetingService.getPastMeetingsCountByProject(project.uid).pipe(finalize(() => this.fpPastCountLoading.set(false)));
+        })
+      ),
+      { initialValue: 0 }
+    );
   }
 
   private initFpRecordingsAvailableCount(): Signal<number> {
--- a/apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
@@ -783,28 +783,14 @@ export class MeetingsDashboardComponent {
   }
 
   private initFpUpcomingCount(): Signal<number> {
-    const project$ = toObservable(this.project);
-    const lens$ = toObservable(this.activeLens);
-
-    return toSignal(
-      combineLatest([project$, lens$, this.refresh$]).pipe(
-        switchMap(([project, lens]) => {
-          if (lens === 'me' || !project?.uid) {
-            return of(0);
-          }
-          if (!isPlatformBrowser(this.platformId)) {
-            this.fpUpcomingCountLoading.set(true);
-            return of(0);
-          }
-          this.fpUpcomingCountLoading.set(true);
-          return this.meetingService.getMeetingsCountByProject(project.uid).pipe(finalize(() => this.fpUpcomingCountLoading.set(false)));
-        })
-      ),
-      { initialValue: 0 }
-    );
+    return this.initFpCount(this.fpUpcomingCountLoading, (uid) => this.meetingService.getMeetingsCountByProject(uid));
   }
 
   private initFpPastCount(): Signal<number> {
+    return this.initFpCount(this.fpPastCountLoading, (uid) => this.meetingService.getPastMeetingsCountByProject(uid));
+  }
+
+  private initFpCount(loadingSignal: WritableSignal<boolean>, fetchFn: (uid: string) => Observable<number>): Signal<number> {
     const project$ = toObservable(this.project);
     const lens$ = toObservable(this.activeLens);
 
@@ -812,14 +798,15 @@ export class MeetingsDashboardComponent {
       combineLatest([project$, lens$, this.refresh$]).pipe(
         switchMap(([project, lens]) => {
           if (lens === 'me' || !project?.uid) {
+            loadingSignal.set(false);
             return of(0);
           }
           if (!isPlatformBrowser(this.platformId)) {
-            this.fpPastCountLoading.set(true);
+            loadingSignal.set(true);
             return of(0);
           }
-          this.fpPastCountLoading.set(true);
-          return this.meetingService.getPastMeetingsCountByProject(project.uid).pipe(finalize(() => this.fpPastCountLoading.set(false)));
+          loadingSignal.set(true);
+          return fetchFn(project.uid).pipe(finalize(() => loadingSignal.set(false)));
         })
       ),
       { initialValue: 0 }
--- a/apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
@@ -806,7 +806,10 @@ export class MeetingsDashboardComponent {
             return of(0);
           }
           loadingSignal.set(true);
-          return fetchFn(project.uid).pipe(finalize(() => loadingSignal.set(false)));
+          return fetchFn(project.uid).pipe(
+            catchError(() => of(0)),
+            finalize(() => loadingSignal.set(false))
+          );
         })
       ),
       { initialValue: 0 }
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -49,7 +49,6 @@ import { pollEndpoint } from '../helpers/poll-endpoint.helper';
 import { fetchAllQueryResources } from '../helpers/query-service.helper';
 import { getEffectiveEmail, getEffectiveUsername, getUsernameFromAuth, stripAuthPrefix } from '../utils/auth-helper';
 import { AccessCheckService } from './access-check.service';
-import { CommitteeService } from './committee.service';
 import { logger } from './logger.service';
 import { MicroserviceProxyService } from './microservice-proxy.service';
 import { ProjectService } from './project.service';
@@ -60,13 +59,11 @@ import { ProjectService } from './project.service';
 export class MeetingService {
   private accessCheckService: AccessCheckService;
   private microserviceProxy: MicroserviceProxyService;
-  private committeeService: CommitteeService;
   private projectService: ProjectService;
 
   public constructor() {
     this.accessCheckService = new AccessCheckService();
     this.microserviceProxy = new MicroserviceProxyService();
-    this.committeeService = new CommitteeService();
     this.projectService = new ProjectService();
   }
 
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -1451,6 +1451,8 @@ export class MeetingService {
 
     // Use Promise.allSettled so one transient batch failure doesn't wipe names resolved by other
     // batches — mirrors the pattern in getCommitteesWithMailingList.
+    // In practice a project's displayed meetings touch ≤1 unique committee, so this is almost always
+    // a single concurrent request; batching guards only against pathological cardinality.
     const results = await Promise.allSettled(
       batches.map((batch) =>
         fetchAllQueryResources<Committee>(
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -1442,21 +1442,45 @@ export class MeetingService {
       unique_committees: uniqueCommitteeUids.length,
     });
 
-    let committeeMap: Map<string, Committee>;
-    try {
-      committeeMap = await this.committeeService.getCommitteesByIds(req, uniqueCommitteeUids);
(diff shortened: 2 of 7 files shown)


apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
```diff
@@ -214,11 +216,12 @@ export class MeetingsDashboardComponent {
...
+    // Foundation/Project lens stat cards
+    // fpStatsLoading gates recurring/recordings cards (backed by raw full-page fetches).
+    // fpUpcomingCount/fpPastCount use the count API directly — they resolve independently.
     this.fpStatsLoading = computed(() => (this.timeFilter() === 'past' ? this.fpPastLoading() : this.fpUpcomingLoading()));
-    this.fpUpcomingCount = computed(() => (this.activeLens() !== 'me' ? this.rawFpUpcomingMeetings().length : 0));
-    this.fpPastCount = computed(() => (this.activeLens() !== 'me' ? this.rawFpPastMeetings().length : 0));
+    this.fpUpcomingCount = this.initFpUpcomingCount();
+    this.fpPastCount = this.initFpPastCount();
     this.fpRecurringCount = computed(() => (this.activeLens() !== 'me' ? this.rawFpUpcomingMeetings().filter((m) => m.recurrence !== null).length : 0));
     this.fpRecordingsAvailableCount = this.initFpRecordingsAvailableCount();
```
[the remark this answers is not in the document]
MRashad26: Thanks for looking into this. The count and the list use the same upstream query-service index (`v1_meeting` via `tags=project_uid:{uid}`) — no additional server-side filtering. `filterAndSortUpcomingMeetings` on the client applies a ~150s staleness window to protect against stale SWR cache entries returning meetings that have already started; it is not a meaningful data filter in steady state. In practice the count and the list cardinality agree. No code change needed here.

apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
```diff
@@ -722,6 +725,50 @@ export class MeetingsDashboardComponent {
...
+            return of(0);
+          }
+          this.fpPastCountLoading.set(true);
+          return this.meetingService.getPastMeetingsCountByProject(project.uid).pipe(finalize(() => this.fpPastCountLoading.set(false)));
+        })
+      ),
+      { initialValue: 0 }
+    );
+  }
+
```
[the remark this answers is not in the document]
MRashad26: Done — extracted shared `initFpCount(loadingSignal, fetchFn)` helper that both `initFpUpcomingCount` and `initFpPastCount` delegate to. Also added `loadingSignal.set(false)` in the `lens === 'me' || !project?.uid` early-return branch to prevent the stuck skeleton.

See commit cf93fd8ff: fix(review): address PR #1076 CodeRabbit review feedback LFXV2-2120

MRashad26:
## Review Feedback Addressed

Commit: cf93fd8ff

### Changes Made
- **meetings-dashboard.component.ts**: Extracted shared `initFpCount(loadingSignal, fetchFn)` helper — `initFpUpcomingCount` and `initFpPastCount` now both delegate to it, eliminating the duplicated `combineLatest`/`switchMap`/`finalize` pattern. Added `loadingSignal.set(false)` in the `lens === 'me' || !project?.uid` early-return branch to prevent the stat card skeleton from getting stuck. (per @coderabbitai Thread 2)
- **apps/shared/services/meeting.service.ts**: Extracted `private getCountByTag(endpoint, tag)` helper; `getMeetingsCountByCommittee`, `getMeetingsCountByProject`, and `getPastMeetingsCountByProject` all delegate to it. Moved helper after all public methods to satisfy `@typescript-eslint/member-ordering`. (per @coderabbitai Thread 4)
- **server/services/meeting.service.ts**: Rewrote `getCommitteeNameMap()` to use `Promise.allSettled` directly (bypassing `getCommitteesByIds` which uses fail-fast `Promise.all`). A single batch failure now degrades gracefully — affected meetings lose their committee name but the rest are unaffected. (per @coderabbitai Thread 3)

### No Change Needed
- **meetings-dashboard.component.ts:226** — count alignment between `fpUpcomingCount` and the list is preserved: both `getMeetingsCountByProject` (count) and `getMeetingsByProjectPaginated` (list) query the same `v1_meeting` index with the same `tags=project_uid:{uid}` filter. `filterAndSortUpcomingMeetings` on the client is a ~150s SWR staleness guard, not a meaningful cardinality filter. (flagged by @coderabbitai Thread 1)

### Threads Resolved

audigregorie (commented):
## Code Review Summary

Solid performance PR — count cards now resolve via lightweight endpoints, committee enrichment is batched, and the follow-up commit cleanly addressed the CodeRabbit loading/helper feedback. I don't see correctness blockers; the main remaining concern is duplicated batch-fetch logic on the server.

### Minor — outside the diff

- No automated tests cover the new `GET /past-meetings/count` route or the dashboard count-loading split. Manual test plan in the PR description is thorough; consider a focused server route test or component test in a follow-up.

### What's done well

- Independent loading signals (`fpUpcomingCountLoading` / `fpPastCountLoading`) decouple count cards from the slower full-page fetches used for recurring/recordings.
- `getCommitteeNameMap` correctly drops the N+1 `getCommitteeById` fan-out and uses `Promise.allSettled` so one bad batch doesn't blank every committee name.
- Past-meetings count endpoint mirrors the existing meetings count pattern (controller → `getMeetingsCount` with `v1_past_meeting`, route registered before `/:uid`).

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1441,22 +1442,45 @@ export class MeetingService {
...
+          (pageToken) =>
+            this.microserviceProxy.proxyRequest<QueryServiceResponse<Committee>>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', {
+              type: 'committee',
+              filters_or: batch.map((uid) => `uid:${uid}`),
+              ...(pageToken && { page_token: pageToken }),
+            }),
+          { failOnPartial: true }
+        )
+      )
     );
```
audigregorie: **Major — Duplicated committee batch-fetch logic**

This chunk-at-100 + `filters_or` + `fetchAllQueryResources` block mirrors `CommitteeService.getCommitteesByIds`, which this PR promoted to `public` specifically to enable reuse. The inline reimplementation means two places must stay in sync if batching semantics change, and the visibility change doesn't buy anything for meeting enrichment yet.

The intentional difference is failure mode: `getCommitteesByIds` uses fail-fast `Promise.all`, while this path needs `Promise.allSettled`. Consider extracting a shared helper on `CommitteeService` — e.g. `getCommitteeNamesByIds(req, uids, { tolerateBatchFailures: true })` — and calling it from `getCommitteeNameMap` instead of duplicating the query wiring.

apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
```diff
@@ -722,6 +726,37 @@ export class MeetingsDashboardComponent {
...
+          if (lens === 'me' || !project?.uid) {
+            loadingSignal.set(false);
+            return of(0);
+          }
+          if (!isPlatformBrowser(this.platformId)) {
+            loadingSignal.set(true);
+            return of(0);
+          }
+          loadingSignal.set(true);
+          return fetchFn(project.uid).pipe(finalize(() => loadingSignal.set(false)));
```
audigregorie: **Minor — Add defensive `catchError` in `initFpCount`**

`getCountByTag` already swallows HTTP errors and returns `0`, so this isn't a functional bug today. Other streams in this component (e.g. `initializeRawFpUpcomingMeetings`) still add a local `catchError` so a future service change can't leave `toSignal` subscribed to a broken observable. Matching that pattern here keeps the stat cards resilient.

```suggestion
          return fetchFn(project.uid).pipe(
            catchError(() => of(0)),
            finalize(() => loadingSignal.set(false))
          );
```

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1441,22 +1442,45 @@ export class MeetingService {
...
+    const unique = [...new Set(uniqueCommitteeUids)].filter(Boolean);
+    const BATCH_SIZE = 100;
+    const batches: string[][] = [];
+    for (let i = 0; i < unique.length; i += BATCH_SIZE) {
+      batches.push(unique.slice(i, i + BATCH_SIZE));
+    }
+
+    // Use Promise.allSettled so one transient batch failure doesn't wipe names resolved by other
+    // batches — mirrors the pattern in getCommitteesWithMailingList.
+    const results = await Promise.allSettled(
```
ahmedomosanya, line 1453: **Nit (non-blocking):** The batching here caps each request at 100 UIDs, which nicely guards URL length — but since all batches fire through a single `Promise.allSettled(batches.map(...))`, the number of *concurrent* upstream requests still scales with unique-committee cardinality. In practice that's almost always ≤1 batch for a project's displayed meetings, so this isn't a real problem today.

That said, we've been treating "batched size + unbounded fan-out" as a soft smell elsewhere. Would it be worth either capping concurrency (e.g. a small pool) or dropping a one-line comment noting the expected cardinality bound, so the intent is explicit? The same shape shows up in `getCommitteesByIds`, so whatever we decide could apply to both. Happy to leave as-is if the bound is well understood.

apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts
MRashad26: Done — added `catchError(() => of(0))` before `finalize` in `initFpCount`. Good catch; matches the defense-in-depth pattern already used in `initializeRawFpUpcomingMeetings`.

See commit e9f34bfee.

apps/lfx-one/src/server/services/meeting.service.ts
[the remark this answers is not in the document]
MRashad26: Good catch on the mismatch between the PR description and the code. The inline reimplementation in `getCommitteeNameMap` is intentional — it needs `Promise.allSettled` so one transient batch failure degrades gracefully (affected meetings lose their committee name, others are unaffected). `getCommitteesByIds` uses fail-fast `Promise.all` which is correct for callers where partial results would be worse than none.

Since nothing outside `CommitteeService` actually calls `getCommitteesByIds`, reverted it to `private` in commit e9f34bfee to remove the misleading visibility change.

MRashad26: Agreed on the duplication concern. The divergence is intentional — `getCommitteeNameMap` needs `Promise.allSettled` (partial batch failure is fine for best-effort name enrichment) while `getCommitteesByIds` uses `Promise.all` (correctness-critical callers must get all committees or none). Extracting a helper with a `tolerateBatchFailures` option would be the right move if a third caller emerges with the same relaxed semantics, but keeping them separate for now avoids complexity for a single divergent caller.

Since nothing outside `CommitteeService` calls `getCommitteesByIds`, reverted it to `private` in commit e9f34bfee — removes the misleading "to enable reuse" claim from the PR description's intent.

apps/lfx-one/src/server/services/meeting.service.ts
MRashad26: Fair observation. Added a comment in commit e9f34bfee documenting the expected cardinality bound — in practice a project's displayed meetings touch ≤1 unique committee, so this is almost always a single concurrent request; the batching only guards against pathological cases. Concurrency capping would be an over-engineering at current scale, but the comment makes the intent explicit.

MRashad26:
## Review Feedback Addressed — Iteration 2

Commit: e9f34bfee

### Changes Made
- **meetings-dashboard.component.ts**: Added `catchError(() => of(0))` before `finalize` in `initFpCount` — defense-in-depth matching the pattern already used in `initializeRawFpUpcomingMeetings` (per @audigregorie)
- **committee.service.ts**: Reverted `getCommitteesByIds` from `public` back to `private` — nothing outside the class calls it; the visibility change was misleading (per @copilot-pull-request-reviewer, @audigregorie)
- **server/services/meeting.service.ts**: Added comment documenting expected concurrency (≤1 batch in practice for a project's displayed meetings) (per @ahmedomosanya)

### No Change Needed
- **Duplication in `getCommitteeNameMap`**: The inline `Promise.allSettled` implementation is intentional — it needs partial-failure tolerance (a bad batch degrades to missing names, not an error) while `getCommitteesByIds` uses fail-fast `Promise.all` for callers that need all-or-nothing correctness. Extracting a shared helper with a failure-mode flag is the right move if a third caller needs the relaxed semantics; keeping them separate for now avoids complexity for a single divergent case.

### Threads Resolved

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1441,22 +1442,47 @@ export class MeetingService {
...
-      uniqueCommitteeUids.map(async (uid) => {
-        try {
-          const committee = await this.committeeService.getCommitteeById(req, uid);
-          return { uid, name: committee.name };
-        } catch (error) {
-          logger.warning(req, 'get_meeting_committees', 'Committee enrichment failed; continuing without name', { committee_uid: uid, err: error });
-          return { uid, name: undefined };
-        }
-      })
+    const unique = [...new Set(uniqueCommitteeUids)].filter(Boolean);
```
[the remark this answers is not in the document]
MRashad26: Fixed — removed the redundant `new Set()` wrap. `uniqueCommitteeUids` is already a deduplicated array from `[...new Set(...)]` at line 1429, so only `.filter(Boolean)` adds value.

See commit 7d87205a0: remove redundant new Set() wrap on uniqueCommitteeUids.

MRashad26:
**Re: @coderabbitai — \`Record<string, any>\` cast in \`past-meeting.controller.ts\`**

Thanks for the review. The \`req.query as Record<string, any>\` cast is intentional and consistent with the established pattern across the codebase — it's not introducing a looser type; it's matching the exact signature of \`MeetingService.getMeetingsCount(req, query: Record<string, any>, meetingType: string)\`. The service itself accepts \`Record<string, any>\`, so the cast aligns the Express \`ParsedQs\` type to what the service expects.

The same pattern appears in 31 places across all controllers (including line 36 of this same file for \`getPastMeetings\`, which wasn't flagged). Introducing a separate typed query shape here would require a parallel change to the service signature, which is out of scope for this PR and would diverge from the existing controller convention.

No code change needed here.

MRashad26:
## Review Feedback Addressed

Commit: $SHA

### Changes Made
- **server/services/meeting.service.ts:1445**: removed redundant \`new Set()\` wrap on \`uniqueCommitteeUids\` — the array is already deduplicated at line 1429; only \`.filter(Boolean)\` adds value (per @copilot-pull-request-reviewer)

### No Change Needed
- **server/controllers/past-meeting.controller.ts:61–65**: \`req.query as Record<string, any>\` cast is the established pattern (31 identical casts across controllers); \`getMeetingsCount\` signature accepts \`Record<string, any>\` so this cast matches the service's own type (flagged by @coderabbitai — see inline response above)

### Threads Resolved

MRashad26:
## Review Feedback Addressed

Commit: 7d87205a0481c243f1b99b2676b981397bd1d108

### Changes Made
- **server/services/meeting.service.ts:1445**: removed redundant `new Set()` wrap on `uniqueCommitteeUids` — the array is already deduplicated at line 1429; only `.filter(Boolean)` adds value (per @copilot-pull-request-reviewer)

### No Change Needed
- **server/controllers/past-meeting.controller.ts:61–65**: `req.query as Record<string, any>` cast is the established pattern (31 identical casts across controllers); `getMeetingsCount` signature accepts `Record<string, any>` so this cast matches the service's own type (flagged by @coderabbitai — see inline response above)

### Threads Resolved

MRashad26:
## Review Feedback Addressed

Commit: b863d65893629cf86beb76a606374146e688b962

### Changes Made
- **server/services/meeting.service.ts**: removed stale `CommitteeService` import, field, and constructor instantiation — `getCommitteeNameMap` now uses `microserviceProxy` directly so the service graph no longer needs to be initialized (per @copilot-pull-request-reviewer)

### Threads Resolved

ahmedomosanya (approved):
LGTM — approving.

Re-reviewed after the latest round of feedback commits (`06c736d5`, `7d87205a`, `b863d658`). All open threads are addressed and the fixes verify clean:

- `initFpCount` now has `catchError(() => of(0))` before `finalize` — defense-in-depth, ordering is correct (emits 0 → completes → loading reset).
- `getCommitteesByIds` reverted to `private`; still has its internal callers, so the misleading public modifier is gone with no functional impact.
- Redundant `new Set()` wrap dropped — the array is already deduped upstream in `getCommitteeNameMap`.
- Dead `CommitteeService` dependency (import + field + constructor) removed; confirmed zero lingering `this.committeeService` references and `getCommitteeNameMap` goes through `microserviceProxy` directly now.

Solid, well-scoped perf work: the N+1 committee fetch is genuinely eliminated and the stat cards resolve independently via the dedicated count endpoints. `/count` is correctly registered before `/:uid` so no route shadowing.

One trivial non-blocking note (no need to hold merge): the new comment says a project's meetings touch "≤1 unique committee" — the accurate bound is ≤1 *batch* (<100 unique committees). Just a comment, fine to fix opportunistically.

audigregorie (approved):

PR #1077, opened 2026-07-08
fix(meetings): derive recordings tab count from live availability (LFXV2-2361)

audigregorie:
# Summary

The recordings tab badge was counting meetings where recording was enabled in settings, not meetings that actually have a recording available.

This change derives the count from the same recording resource as "See Recording" — bounded per-meeting fetches over the last 30 days, using `getLargestSessionShareUrl` as the canonical availability rule so the badge, dashboard stat, and meeting cards stay aligned.

## Test plan

- [ ] Me lens, Past tab: "Recordings Available" shows skeleton until count resolves (not a premature `0`)
- [ ] Foundation/Project lens, Past tab: same behavior for FP recordings stat
- [ ] Count matches meetings where "See Recording" is available on cards
- [ ] Past meeting join page: Recording badge follows largest-session-by-size rule

---

> [!NOTE]
> **Medium Risk**
> Dashboard recording stats trigger N concurrent API calls per lens refresh; wrong resource IDs would break past-meeting fetches, but the change standardizes IDs and adds tests for the new helpers.
> 
> **Overview**
> **Recordings Available** on the meetings dashboard (Me and Foundation/Project lenses) no longer uses `recording_enabled` on past rows. It fetches each past meeting in the last 30 days (bounded concurrency), treats a recording as available only when `getLargestSessionShareUrl` returns a URL—the same rule as **See Recording**—and shows a skeleton until that count finishes so users do not see a misleading `0`.
> 
> Shared helpers in `past-meeting.utils` centralize **`getPastMeetingResourceId`**, **`getLargestSessionShareUrl`**, and **`getPastMeetingStartTimeMs`** (handles Go zero-dates on `scheduled_start_time`). Meeting cards, meeting join, and dashboard use the resource ID for past sub-resource calls and detail links instead of raw `meeting.id`.
> 
> **Meeting join** shows the Recording badge from actual recording URL on past URLs; upcoming meetings still use `recording_enabled`. **`MeetingService`** adds a TTL-cached, `shareReplay` past-meeting recording fetch (cleared on dashboard refresh) and URL-encodes past-meeting path segments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c55a0bf59afd3ced0f724295c27ebee51a355ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ahmedomosanya (dismissed):

luismoriguerra (dismissed):
## Audit: PR #1077 — fix(meetings): use has_recording for recordings tab badge (LFXV2-2361)

audigregorie · `feat/lfxv2-2361-recordings-count` → `main` · open · 3 files, +5/−3

**Scope:** Meetings dashboard “Recordings Available” stat, shared `PastMeeting` type, and public past-meeting API response.

**Lanes run:** `format:check` · `lint:check` · `build` · `check-types` · secrets · code-standards · architecture (skipped — UI-only bugfix)

---

### 1. Why this change

The “Recordings Available” badge counted meetings where recording was **enabled in settings** (`recording_enabled`), not meetings that **actually have a recording**. That inflated the count and misled users — especially in the Project lens ([LFXV2-2361](https://linuxfoundation.atlassian.net/browse/LFXV2-2361)).

The fix aligns the stat with the semantic already used on past-meeting cards: availability means content exists, not that the feature was configured.

### 2. How it was done

| Layer | Change |
|---|---|
| `packages/shared/src/interfaces/meeting.interface.ts` | Adds optional `has_recording?: boolean` to `PastMeeting`, documented as distinct from `recording_enabled` |
| `meetings-dashboard.component.ts` | `initRecordingsAvailableCount()` and `initFpRecordingsAvailableCount()` filter on `has_recording === true` (30-day window unchanged) |
| `public-meeting.controller.ts` | Includes `has_recording` in the limited (non-`full_access`) past-meeting response |

Data path is passthrough — list endpoints spread query-service rows verbatim:
- **Me lens:** `/api/user/past-meetings` → `v1_past_meeting`
- **FP lens:** `/api/past-meetings` → `v1_past_meeting`

`has_recording` is populated upstream in `lfx-v2-meeting-service` (synced from ITX when a real recording session exists). The BFF does not need to enrich it locally.

### 3. Is it correct

**Verdict: Correct approach, well-scoped, ready to merge.**

| Gate | Result |
|---|---|
| `yarn format:check` | ✅ Pass |
| `yarn lint:check` | ✅ Pass |
| `yarn build` | ✅ Pass |
| `yarn check-types` | ✅ Pass |
| GitHub CI | ✅ All checks green |

Prior review: Copilot and CodeRabbit found no issues. **ahmedomosanya** already approved at head `3e5ecdc4`. No unresolved inline threads.

### 4. Missing / gaps (non-blocking follow-ups)

1. **`dashboard-meeting-card.component.ts`** still uses `recording_enabled` for past meetings, while `meeting-card` derives availability from `recordingShareUrl`. The dashboard **stat** is fixed; the “Last Meeting” card chip can still show “enabled” without a file.
2. **`meeting-join.component.html`** still gates the recording badge on `recording_enabled` even though the controller now exposes `has_recording`.
3. Pre-existing date fallback asymmetry: Me lens uses `scheduled_start_time ?? start_time`; FP lens uses only `scheduled_start_time`.
4. No unit tests for the recording-count filter — a small extracted util + spec would be worthwhile in a follow-up.

### 5. Q&A

- **Q:** Why not count `recording_enabled` for past meetings? — **A:** That flag means “recording was configured,” not “a file exists.” `has_recording` is set upstream when ITX reports a real session with a share URL.
- **Q:** Will the count work without BFF enrichment? — **A:** Yes. Both list endpoints pass query-service `v1_past_meeting` fields through; adding the type field lets the dashboard read what’s already indexed.

### 6. Business rules

- **Rule:** “Recordings Available” counts past meetings in the last 30 days where a recording artifact actually exists (`has_recording === true`), not where recording was merely enabled.

---

**Verdict:** ✅ **Approved** — safe to merge. Remaining items are follow-up polish, not blockers.

dealako:
Hey @audigregorie — reviewed this one closely, including tracing the `has_recording` field all the way back to its source in `lfx-v2-meeting-service`.

**Overall impression:** Clean, tightly-scoped fix. The badge was counting "recording enabled" (a config flag) instead of "recording exists" (actual content), and this correctly swaps the filter predicate in both lens variants while leaving the 30-day window and date-fallback logic untouched. The one backend controller change is scoped to exactly the endpoint that needs it — the other two data sources (`/api/user/past-meetings`, `/api/past-meetings`) are raw query-service pass-throughs, so no additional plumbing was needed there, and the diff doesn't try to add any.

I independently verified the upstream contract rather than taking the field's existence on faith: `has_recording` is populated in `lfx-v2-meeting-service` (`internal/domain/models/event_models.go`, `cmd/meeting-api/eventing/past_meeting_event_handler.go`) — synced from ITX/Zoom, true only when the largest recording session has a non-empty share URL — and flows into the `v1_past_meeting` indexer resource that both list endpoints spread verbatim. So this isn't a silent no-op; the badge will actually move.

Also ran both parallel subagent passes (security + general code quality) and independently confirmed `=== true` correctly treats `undefined` (pre-field historical records) and `false` as "not counted," with no type-safety issues from the `PastMeeting`-vs-`Meeting` interface split.

**🔴 Blocking:** 0
**🟡 Minor:** 0
**⚪ Nit:** 0
**❔ Question:** 0 blocking questions — agreeing with luismoriguerra's audit that `dashboard-meeting-card.component.ts` and `meeting-join.component.html` still read `recording_enabled` in other UI surfaces. Those are out of scope for this fix (which is specifically the dashboard stat) and don't need to hold this PR up, but could be worth a fast follow-up ticket if you want the "has a recording" semantic to be consistent everywhere.

Bots agree: CodeRabbit found no actionable comments, Copilot generated no comments across both passes.

dealako (dismissed):
Approved — see summary comment for full analysis. Verified the has_recording upstream contract independently; no blocking, minor, or security findings from my review or the two parallel subagent passes.

dealako:
Hey @audigregorie — thanks for the fast turnaround on this. The pivot from a stored `has_recording` flag to live-derived recording availability is a meaningfully better design (it can't drift from what "See Recording" actually shows), and the bounded-concurrency fetch is a real improvement over the earlier fan-out.

## 👏 Nice work

- **Bounded concurrency fix**: `countMeetingsWithRecording()` in `meetings-dashboard.component.ts` now uses `mergeMap(fn, MEETING_RECORDING_COUNT_FETCH_CONCURRENCY)` instead of the earlier `forkJoin` — resolves CodeRabbit's fan-out concern cleanly.
- **`getPastMeetingResourceId()` helper**: correctly wired into `meeting-card.component.ts` for both the attachments path and `initRecording()` — this resolves the recording-count/ID-mismatch finding from the last round (previously flagged against an older commit).
- **Loading-state correctness**: `meLensStatsLoading`/`fpStatsLoading` now correctly OR in `recordingsCountLoading()` while on the past-meetings tab — the "stat finishes loading early" issue is fixed.
- **Shared cache**: `getPastMeetingRecording()` dedupes concurrent fetches for the same meeting via a `Map` + `shareReplay(1)` cache, avoiding N+1 refetches between the dashboard count and the per-card "See Recording" button.

## Revision tracking (from the prior round)

1. **`dashboard-meeting-card.component.ts` recording-gating logic** — ❌ Still open, but confirmed **out of scope for this PR** (file isn't in this PR's diff — verified via `gh pr view --json files`, 9 files touched, this isn't one of them). Worth a follow-up ticket, not a blocker here.
2. **`meeting-join.component.ts` gating the recording badge on `recording_enabled`** — ✅ Resolved for the gating logic itself: `showRecordingBadge` now correctly derives from `primaryRecordingUrl()` when the meeting was loaded via a past-meeting ID. *However*, fixing this gating logic surfaced a new, more fundamental bug in the same file's data layer — see 🔴 finding below. The badge now shows/hides correctly based on whether a recording URL was fetched, but the fetch itself can be keyed to the wrong resource.
3. **Date-fallback asymmetry** — ✅ Resolved via the shared `getPastMeetingStartTimeMs()` helper, now used consistently in the 30-day recency filter (`meetings-dashboard.component.ts:741`).
4. **No unit tests** — ⚠️ Partially addressed. `past-meeting.utils.spec.ts` now covers the new utility functions, but the dashboard's async fetch/cache/concurrency logic (`countMeetingsWithRecording`, `filterRecentPastMeetings`) still has no test coverage.

## New findings

**🔴 [blocking] `meeting-join.component.ts` fetches recording/summary/attachments/transcript by the wrong ID for recurring past-meeting occurrences**

- **Issue**: Five call sites in this file — `initializePastMeetingRecording()` (~L1075-1084), `initializePastMeetingSummary()` (~L1063-1071), `initializePastMeetingAttachments()` (~L1086-1095), `initializeTranscriptUrl()` (~L1127-1141), and `downloadAttachment()` (~L407-410) — key their fetches off raw `meeting.id`/`meeting?.id`, instead of the shared `getPastMeetingResourceId()` helper this same PR introduces and correctly applies in `meeting-card.component.ts`.
- **Proof**: `meeting.id` and `meeting.meeting_and_occurrence_id` genuinely diverge for recurring meeting occurrences (composite ID format documented at `rsvp-calculator.util.ts:86`, e.g. `"95156357074_1707321600000"`). The `id → meeting_and_occurrence_id` normalization is performed **only** inside `server/services/user.service.ts` (`getUserPastMeetings`/`getUserLatestPastMeetings`, ~L692-696) — i.e. only for the Me-lens dashboard list. `meeting-join.component.ts` gets its data from `getPublicPastMeeting()`, which hits the public single-past-meeting endpoint (`public-meeting.controller.ts` → server's `meeting.service.ts:getPastMeetingById`) — a completely separate code path that does **not** perform this normalization. So for a recurring meeting's join/recording page, `meeting.id` on this component can be the raw (non-composite) ID while the recording/summary/attachments/transcript resources are actually stored under the composite occurrence ID.
- **Why it matters**: For recurring meetings reached via a past-meeting-occurrence join URL, the recording badge, transcript link, summary, and attachment downloads can silently fetch the wrong occurrence's data (or 404/empty-result and hide the recording button entirely) — a user-facing correctness bug that's easy to miss because it only reproduces on recurring meetings, not one-off ones.
- **Fix**: Replace the five `meeting.id`/`meeting?.id` references above with `getPastMeetingResourceId(meeting)` (same helper already imported and used correctly in `meeting-card.component.ts`), e.g.:
  ```typescript
  private initializePastMeetingRecording(): Signal<PastMeetingRecording | null> {
    return toSignal(
      combineLatest([toObservable(this.pastMeetingFullAccess), toObservable(this.meeting)]).pipe(
        switchMap(([hasAccess, meeting]) => {
          const id = meeting ? getPastMeetingResourceId(meeting) : null;
          if (!hasAccess || !id || !this.authenticated()) return of(null);
          return this.meetingService.getPastMeetingRecording(id).pipe(catchError(() => of(null)));
        })
      ),
      { initialValue: null }
    );
  }
  ```

**🟡 [minor] PR title/commit messages still reference `has_recording`, but the implementation no longer uses that field**

- The PR title (`fix(meetings): use has_recording for recordings tab badge`) and the latest commit message describe a server-flag approach that was superseded by the live-derivation approach in this same PR. Copilot flagged this twice already; still unaddressed. Recommend updating the PR title/description before merge so the history accurately reflects what shipped — otherwise it's misleading to anyone bisecting or reading `git log` later.

**🟡 [minor] `pastMeetingRecordingCache` has no automatic staleness invalidation**

- `MeetingService.getPastMeetingRecording()`'s `shareReplay(1)` cache only evicts on fetch error or a manual `clearPastMeetingRecordingCache()` call (invoked from the dashboard's `refreshMeetings()`). If a recording finishes processing after a user has already visited a meeting's page once in the session, the cached "no recording" result can persist until an explicit refresh. Agree with cursor's flag on this — downgrading from cursor's original severity since the manual refresh path is a reasonable escape hatch, but worth a short TTL or at least a comment documenting the tradeoff.

**⚪ [nit] `meeting-card.component.ts` — redundant null-check before `getLargestSessionShareUrl`**

- CodeRabbit's nitpick about a null-check ternary ahead of a call that already null-guards internally still applies but is purely stylistic — take it or leave it.

## AI bot reconciliation

- **Agree** with CodeRabbit's `forkJoin` → bounded `mergeMap` finding — resolved, see Nice Work above.
- **Agree** with Copilot's PR title/description staleness flag — folded into the 🟡 finding above, still open.
- **Agree** with cursor's recording-cache-never-invalidates finding — folded into the 🟡 finding above, downgraded to minor given the manual-refresh escape hatch.
- **Disagree** with cursor's "join badge ignores canonical URL rule" comment — `initializePrimaryRecordingUrl()` already returns `computed(() => getLargestSessionShareUrl(this.pastMeetingRecording()))`, using the same canonical shared helper as every other call site. This looks like a false positive; the badge visibility and URL selection both go through the shared rule correctly. (Note this is a distinct question from the 🔴 finding above — that one is about which meeting ID keys the fetch, not which URL is selected from the result.)

## Issue count

- 🔴 Blocking: 1 — meeting-join.component.ts un-migrated ID fetches for recurring past-meeting occurrences
- 🟡 Minor: 2 — stale PR title/commit message; recording cache has no automatic invalidation
- ⚪ Nit: 1 — redundant null-check in meeting-card.component.ts

## Decision

🔴 **Needs changes before approval** — the blocking finding is a real, silently-reachable correctness bug in a file this PR already touches, and the fix is a one-line change per call site using a helper this PR already introduces and correctly applies elsewhere. Everything else here is minor/nit and can be addressed in the same pass.

dealako (changes requested):
Requesting changes for the blocking ID-mismatch finding in `meeting-join.component.ts` — see the consolidated follow-up review comment above for full detail, proof, and suggested fix. Two minor items and one nit are also called out there; none of those block on their own.

dealako:
Hey @audigregorie — another solid round. `bf3a9c7f` closes out the ID-alignment gap in `meeting-card.component.ts` and `meetings-dashboard.component.ts` cleanly, and the cache-eviction-on-error addition is a nice defensive touch. This round I also ran the security-auditor and code-reviewer-pro subagents against the new commits per the follow-up workflow — their findings are merged below.

## 👏 Nice work

- **`getPastMeetingResourceId()` fully wired into `meeting-card.component.ts`**: `downloadAttachment`, the summary-modal `pastMeetingUid`, `initAttachments`, `initRecording`, `initSummary`, `initTranscript`, and the past-meeting router link all now consistently resolve through the helper. This is exactly the pattern that needs to reach `meeting-join.component.ts` too (see 🔴 below).
- **`clearPastMeetingRecordingCache()` + error-path eviction**: `refreshMeetings()` now explicitly clears the recording cache, and `getPastMeetingRecording()` evicts on error via `tap({error: ...})` — resolves the cache-staleness concern from the prior round.
- **`getPastMeetingResourceId()` test coverage**: both the composite-ID and fallback-to-`id` cases are covered in `past-meeting.utils.spec.ts`.

## Carried-over finding — still open

**🔴 [blocking] `meeting-join.component.ts` still fetches past-meeting sub-resources by raw `meeting.id`**

Unaddressed from the prior round. Confirmed still present at current HEAD (`6df510a3`) across **six** call sites (one more than previously listed — `initializePastMeetingParticipants` also uses the raw ID):

- `initializePastMeetingSummary()` — `meeting.id` (L1068)
- `initializePastMeetingRecording()` — `meeting.id` (L1080)
- `initializePastMeetingAttachments()` — `meeting.id` (L1092)
- `initializePastMeetingParticipants()` — `meeting.id` (L1104)
- `initializeTranscriptUrl()` — `meeting.id` (L1136)
- `downloadAttachment()` — `this.meeting().id` (L409-410)

Same root cause as before: `meeting-join.component.ts` gets its meeting data from `getPublicPastMeeting()` → the public single-past-meeting server endpoint, which does not perform the `id → meeting_and_occurrence_id` normalization that only happens in `user.service.ts`'s Me-lens list endpoints. This PR already ships the fix pattern (`getPastMeetingResourceId()`) and applies it correctly everywhere else touched — `code-reviewer-pro`'s independent pass flagged this exact same set of call sites as a "Critical" finding without being told about it in detail, which is good independent corroboration. Fix is mechanical: swap `meeting.id`/`meeting?.id` for `getPastMeetingResourceId(meeting)` at each of the six sites above.

## New findings this round

**🟡 [minor] Shared `recordingsCountLoading` signal can race on lens switch**

Both GitHub Copilot and my own `code-reviewer-pro` subagent independently flagged this: `initRecordingsAvailableCount()` (Me lens) and `initFpRecordingsAvailableCount()` (FP lens) are two independent `toObservable(activeLens)` subscriptions that both toggle the single shared `recordingsCountLoading` signal via `countMeetingsWithRecording()`'s `finalize()`. On a lens switch, the outgoing lens's `switchMap` unsubscribe (triggering `finalize()` → `set(false)`) can run after the incoming lens's `countMeetingsWithRecording()` has already called `set(true)`, leaving the flag `false` while the newly-active lens's fetches are still in flight — the stat card would briefly show a premature count without its loading skeleton. Fix: give each lens its own loading signal (e.g. `meRecordingsCountLoading` / `fpRecordingsCountLoading`) instead of sharing one, per `code-reviewer-pro`'s suggested diff.

**⚪ [nit] Inconsistent URL encoding on past-meeting resource IDs**

`security-auditor` flagged that `meeting.service.ts`'s past-meeting endpoints (`getPastMeetingRecording`, `getPastMeetingTranscript`, `getPastMeetingSummary`, `getPastMeetingAttachments`) interpolate `pastMeetingUid` into the URL path unencoded, while `getOrgMeetingsSummary` wraps its ID in `encodeURIComponent()`. Exploitability is low — these IDs come from backend-controlled meeting data, not raw user input — but it's a real inconsistency with the codebase's own convention and costs nothing to fix defensively.

**⚪ [nit] Redundant type guard in `countMeetingsWithRecording()`**

`code-reviewer-pro` flagged the `.filter((id): id is string => !!id)` after `getPastMeetingResourceId()` as unnecessary, since the helper always returns a string (falls back to the required `id` field). Fine to leave or drop.

## Findings I checked and dropped (self-challenge)

`security-auditor` raised two additional items I verified against the actual code and disagree with:

- **"Recording cache could leak across users if `MeetingService` is reused across sessions"** — checked `impersonation-banner.component.ts` and `impersonation-dialog.component.ts`: both `startImpersonation`/`stopImpersonation` flows call `window.location.reload()`, which destroys and recreates the entire Angular app (and every root-provided service, including `MeetingService`'s cache) on every identity switch. Not reachable.
- **"`meeting-card.component.ts` lacks an explicit authorization gate before fetching attachments/recordings"** — `meeting-card.component.ts` only ever receives meetings already scoped to the authenticated user's access via upstream list endpoints (dashboard's own past meetings, or the currently-joined meeting) — same trust boundary as the bulk `countMeetingsWithRecording()` fetch, which already relies on the same upstream scoping. No independently reachable authorization bypass here.

## Bot reconciliation (since last round)

- **Agree** with Copilot's shared-loading-signal race finding — folded into the 🟡 finding above.
- **Agree** with Copilot's unbounded-cache-growth observation on `pastMeetingRecordingCache` — same root cause as the prior round's cache-invalidation finding; still a reasonable tradeoff given `refreshMeetings()` now explicitly clears it, but worth a comment noting the growth is bounded in practice by session length and distinct-meetings-viewed.
- **Disagree** (partially) with Copilot's two comments about `recordingsCountLoading` coupling into `meLensStatsLoading`/`fpStatsLoading` making unrelated stats wait — this is a real UX tradeoff worth considering, but it's a pre-existing design choice from the prior round, not a new regression, and it's a UX polish question rather than a correctness bug. No action required unless you want to decouple the loading flags.

## Issue count

- 🔴 Blocking: 1 — `meeting-join.component.ts` still un-migrated (6 call sites)
- 🟡 Minor: 1 — shared `recordingsCountLoading` race on lens switch
- ⚪ Nit: 2 — inconsistent URL encoding; redundant type guard

## Decision

🔴 **Needs changes before approval** — the blocking finding remains open and is now independently confirmed by a second reviewer pass, plus subagent findings exist per the review workflow's decision rules. Everything else is minor/nit.

dealako (changes requested):
Second follow-up round — the meeting-join.component.ts ID-mismatch bug from the prior round remains unaddressed (now independently confirmed by an automated code-reviewer-pro pass), and a new minor loading-signal race was found this round. See the consolidated follow-up comment above for full detail, proof, and fixes.

dealako (approved):

PR #1107, opened 2026-07-15
fix(enrollment): guard stripe short-circuit with EndDate

fayazg:
## Summary

- `deriveEnrollmentStatus` short-circuited to `Active` for any `AutoRenew` + Stripe
  membership regardless of `EndDate`. A lapsed Stripe subscription (failed charge,
  paused) whose `Status` hasn't been flipped to `Expired` upstream would incorrectly
  show `Active` in the individual-enrollment profile tab.
- Guard the short-circuit on `!isExpired`, reusing the existing UTC-midnight `EndDate`
  computation, so it only suppresses "Expiring Soon"/"Expired" while the membership is
  still current.
- Adds unit test coverage for `deriveEnrollmentStatus` (no prior spec existed),
  covering all branches including the regression case.

**JIRA:** [LFXV2-2699](https://linuxfoundation.atlassian.net/browse/LFXV2-2699)

[LFXV2-2699]: https://linuxfoundation.atlassian.net/browse/LFXV2-2699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dealako:
Hey @fayazg 👋 — reviewed this via parallel `security-auditor` + `code-reviewer-pro` subagents plus my own independent pass.

**Overall impression:** Solid, well-scoped bug fix. The Stripe auto-renew short-circuit is correctly moved to run *after* the `EndDate`/`isExpired` computation and gated on `!isExpired`, so a lapsed Stripe subscription (failed charge / paused) whose upstream `Status` hasn't flipped to `Expired` now correctly falls through to the date-based logic instead of falsely showing "Active". The intentional divergence from `myprofile`'s `enrollment.plugin.js` is called out in both the code comment and commit message. Test coverage is strong — 9 cases for `deriveEnrollmentStatus` including the explicit regression case, plus 4 for `enrollmentStatusSeverity`.

I also traced a potential regression risk from the reordering: since the Stripe short-circuit no longer returns before `membership.EndDate.length` is dereferenced, a `null`/`undefined` `EndDate` on a Stripe auto-renew membership could now throw where it previously wouldn't. Checked `enrollment.service.ts:66` — the mapping layer defaults `EndDate: m.EndDate ?? ''`, so this never reaches `deriveEnrollmentStatus` as a non-string. No bug.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 2 — test coverage gaps (not correctness issues): missing `price === null` case (spec only covers `undefined`), and no exact-30-day boundary test for the `Expiring Soon` cutoff
- ⚪ Nit: 0
- ❔ Question: 0

**AI bot reconciliation:** Copilot, CodeRabbit, and Cursor Bugbot all reviewed this PR and found zero actionable issues — consistent with my own findings. No disagreements to reconcile.

**Security:** `security-auditor` subagent found no vulnerabilities — no injection risk, no auth/authz gap (display-layer only), no secrets, no PII (test fixtures are synthetic and explicitly labeled as such).

**Final decision:** ✅ **Approved with minor comments** — the two test-coverage suggestions are optional polish, not blockers.

packages/shared/src/utils/enrollment.utils.spec.ts
dealako, line 1: [minor] Missing `price === null` branch coverage

Issue: The free-tier test only exercises `price: undefined`, but the guard on `enrollment.utils.ts:14` checks `price === null || price === undefined`.
Why it matters: The `null` branch of that guard is currently untested.
Fix: Add a second assertion (or a second `it`) passing `price: null` alongside the existing `undefined` case.

dealako, line 1: [minor] No exact-30-day boundary test

Issue: Tests cover >30 days out (Active) and ~10 days out (Expiring Soon), but not the exact 30-day boundary.
Why it matters: `enrollment.utils.ts:32` uses a strict `<` comparison, so `daysFromNow(30)` should return `Active`, not `Expiring Soon` — this off-by-one edge is a common source of date-logic bugs and isn't currently pinned down by a test.
Fix: Add `it('returns Active when EndDate is exactly 30 days out', ...)` asserting `daysFromNow(30)` → `'Active'`.

dealako (dismissed):
Approved — see consolidated summary comment above. Two minor test-coverage suggestions posted inline; no blockers.

packages/shared/src/utils/enrollment.utils.spec.ts
fayazg: Done — added a `price: null` case (with a cast, since `price` is typed `number | undefined`) alongside the existing `price: undefined` test, covering the other half of the guard at `enrollment.utils.ts`.

See commit c3d7f2b1c.

fayazg: Good catch on the boundary gap — one nuance though: your suggested assertion (`daysFromNow(30)` → `Active`) doesn't actually hold. `EndDate` gets normalized to UTC midnight, while `now` (and the 30-day cutoff derived from it) carries a time-of-day, so midnight-of-day+30 is still strictly less than `now + 30d` — that resolves to `Expiring Soon`, not `Active`.

Added both boundary cases to pin the actual strict-`<` cutoff: `daysFromNow(30)` → `Expiring Soon`, and `daysFromNow(31)` (the first day fully outside the window) → `Active`.

See commit c3d7f2b1c.

fayazg:
## Review Feedback Addressed

Commit: c3d7f2b1c

### Changes Made
- **packages/shared/src/utils/enrollment.utils.spec.ts**: added a `price: null` free-tier case alongside the existing `price: undefined` test, covering the other half of the `price === null || price === undefined` guard (per @dealako)
- **packages/shared/src/utils/enrollment.utils.spec.ts**: added exact 30/31-day boundary tests. Note: `daysFromNow(30)` actually resolves to `Expiring Soon` (not `Active` as originally suggested) because `EndDate` is normalized to UTC midnight while `now` carries a time-of-day — `daysFromNow(31)` is the first day fully outside the window and resolves to `Active`. Both cases are now pinned (per @dealako)

### Threads Resolved

packages/shared/src/utils/enrollment.utils.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+import { EnrollmentMembership, IndividualEnrollment } from '../interfaces';
+import { deriveEnrollmentStatus, enrollmentStatusSeverity } from './enrollment.utils';
+
+/** Formats a Date as YYYY-MM-DD, matching the EndDate shape the member-service returns. */
+function isoDate(date: Date): string {
+  return date.toISOString().slice(0, 10);
+}
+
+const now = new Date();
+const daysFromNow = (days: number): string => isoDate(new Date(now.getTime() + days * 24 * 60 * 60 * 1000));
```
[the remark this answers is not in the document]
fayazg: Good catch — done. Froze the clock with `vi.useFakeTimers()`/`vi.setSystemTime()` in `beforeAll`/`afterAll`, and switched `daysFromNow` to read `Date.now()` at call time instead of a captured module-load constant, so fixture dates and the resolver's own `new Date()` calls always resolve against the same frozen instant.

See commit a7460b40e.

fayazg:
## Review Feedback Addressed

Commit: a7460b40e

### Changes Made
- **packages/shared/src/utils/enrollment.utils.spec.ts**: froze the clock with `vi.useFakeTimers()`/`vi.setSystemTime()` in `beforeAll`/`afterAll`, and switched `daysFromNow` to read `Date.now()` at call time instead of a captured module-load constant — eliminates the possibility of a UTC-midnight crossing flipping the day-boundary assertions nondeterministically (per copilot-pull-request-reviewer)

### Threads Resolved

packages/shared/src/utils/enrollment.utils.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+    const item = enrollment({ price: undefined, membership: membership({ EndDate: daysFromNow(-10) }) });
+    expect(deriveEnrollmentStatus(item)).toBe('Active');
+  });
+
+  it('returns Active for a free-tier item (price null) regardless of EndDate', () => {
+    // `price` is typed as `number | undefined`, but the resolver also guards against `null` —
+    // cast to exercise that branch since upstream data can still carry a literal null.
+    const item = enrollment({ price: null as unknown as undefined, membership: membership({ EndDate: daysFromNow(-10) }) });
+    expect(deriveEnrollmentStatus(item)).toBe('Active');
+  });
```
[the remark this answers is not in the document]
fayazg: Done — cast the `price` fixture through `number` instead of `undefined` so it's clear the test is deliberately exercising the runtime `null` branch, not an absent value.

See commit 5bcb0d312: `packages/shared/src/utils/enrollment.utils.spec.ts`.

fayazg:
## Review Feedback Addressed

Commit: 5bcb0d312

### Changes Made
- **enrollment.utils.spec.ts**: cast the `price` fixture through `number` instead of `undefined` for the price-null test, so the cast doesn't misrepresent the runtime value it's exercising (per copilot-pull-request-reviewer)

### Threads Resolved

dealako (approved):

PR #1108, opened 2026-07-15
perf(dashboards): read health score from dbt detail model (LFXV2-2697)

audigregorie:
# Summary

Replaces the in-query `PROJECT_HEALTH_METRICS_DAILY` lookup (windowed via `QUALIFY ROW_NUMBER()`) in `getFoundationProjectsDetail` with a read of the prejoined `HEALTH_SCORE_CATEGORY` column from the dbt model `FOUNDATION_TOTAL_PROJECTS_DETAIL`. dbt folds the latest health-score category into this detail model via a slug-only LEFT JOIN to `PROJECT_HEALTH_METRICS_LATEST` (the shared per-project latest-score selection), so the service no longer re-derives "latest" in SQL at request time. Also surfaces the already-fetched `lastUpdated` date in the health-scores drawer.

# Before / After

**Before:** `getFoundationProjectsDetail` derived the latest health-score category by scanning `PROJECT_HEALTH_METRICS_DAILY`, filtering by `FOUNDATION_SLUG`, and using `QUALIFY ROW_NUMBER() OVER (PARTITION BY PROJECT_SLUG ORDER BY METRIC_DATE DESC) = 1` to keep only the newest row per project — re-deriving "latest" from the per-date history table on every call, then joining with a direct `d.PROJECT_SLUG = h.PROJECT_SLUG` equality. Row-interface comments attributed the join to `PROJECT_HEALTH_METRICS_DAILY` and (incorrectly) described it as a 30-day window. The drawer listed each project name without its last-updated date, even though `LAST_UPDATED_TS` was already selected and mapped.

**After:** The query reads `d.HEALTH_SCORE_CATEGORY` straight from `ANALYTICS.PLATINUM_LFX_ONE.FOUNDATION_TOTAL_PROJECTS_DETAIL d` — no in-query join. The latest-score selection and the case-insensitive slug match (`PROJECT_HEALTH_METRICS_LATEST` stores lowercased slugs; detail slugs may be mixed-case) are owned by the dbt model that builds `FOUNDATION_TOTAL_PROJECTS_DETAIL`, not by this service's SQL. `FOUNDATION_HEALTH_SCORE_DISTRIBUTION` counts over the same model, so the chart and this table can never disagree. Absent or unrecognized categories resolve to `null` via `normalizeHealthScoreCategory`; the drawer renders that `null` as the "Unscored" badge. Comments in `FoundationProjectsDetailRow` and `ProjectTableRow` now reference the `LATEST` table (the upstream of the dbt fold-in) and describe it by grain (latest row per project) rather than the unsupported 30-day claim. The drawer shows an `Updated {date}` sub-line under each project name when `lastUpdated` is present.

# Changes

## project.service.ts (Snowflake query)
- Replaced the `PROJECT_HEALTH_METRICS_DAILY` subquery + `QUALIFY ROW_NUMBER() OVER (...) = 1` window filter with a direct `SELECT d.HEALTH_SCORE_CATEGORY ... FROM FOUNDATION_TOTAL_PROJECTS_DETAIL d`.
- The latest-score selection and the case-insensitive (`LOWER(d.PROJECT_SLUG) = h.PROJECT_SLUG`) slug match now live in the dbt model that builds `FOUNDATION_TOTAL_PROJECTS_DETAIL`, not in this query; documented inline.
- Updated the query comment to describe the dbt fold-in by grain and to distinguish the returned `null` value from its "Unscored" UI presentation.

## project-health-scores-drawer.component.html (UI)
- Wraps the project name in a `flex flex-col` stack and adds a conditional `Updated {lastUpdated}` sub-line. This surfaces the `LAST_UPDATED_TS` value that the query already selected and the service already mapped to `ProjectTableRow.lastUpdated`; no new data plumbing is introduced.

## analytics.service.ts (request cache)
- Adds a per-slug `foundationHealthScoreDistributionCache` (`Map<string, Observable<...>>` + `shareReplay(1)`) so the foundation-health card and the Project Health Scores drawer share one request per foundation instead of each firing their own on navigation back/forth.
- Session-scoped with no TTL: a successful distribution is pinned for the service lifetime (root-provided service → until the tab reloads). Acceptable because foundation health-score distribution rarely changes within a session; the cache entry is evicted on HTTP error so a transient failure can't lock subscribers to the zeroed fallback. Rationale is documented inline at `getFoundationHealthScoreDistribution`; if per-session freshness becomes a concern, add a TTL or an explicit `invalidate(foundationSlug)` call.
## @lfx-one/shared interfaces
- `FoundationProjectsDetailRow.HEALTH_SCORE_CATEGORY` comment updated to reference `PROJECT_HEALTH_METRICS_LATEST` (the upstream of the dbt fold-in).
- `ProjectTableRow.healthScoreCategory` comment updated to reference `PROJECT_HEALTH_METRICS_LATEST` and tightened to a single line.

## Tests
- No test changes; the row shape and `HEALTH_SCORE_CATEGORY` semantics are unchanged — only the Snowflake source (prejoined dbt column vs. re-derived in-query join) differs.

# Breaking changes

_Freshness:_ the new session-scoped cache means a successful foundation health-score distribution is not re-fetched until the tab reloads; previously each subscriber issued its own request.

None. The returned row shape and `HEALTH_SCORE_CATEGORY` semantics are unchanged; the Snowflake source table and the relocation of the latest-score selection + case-insensitive slug match into the dbt model differ. The drawer gains a non-breaking `Updated` sub-line.


apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -1701,20 +1706,14 @@ export class ProjectService {
         d.MAINTAINERS_CURRENT_COUNT,
         d.STARS_YTD_COUNT,
         d.LAST_UPDATED_TS,
-        h.HEALTH_SCORE_CATEGORY
+        d.HEALTH_SCORE_CATEGORY
```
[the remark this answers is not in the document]
audigregorie: Updated the PR title and body to describe the actual implementation: the query now reads the prejoined `d.HEALTH_SCORE_CATEGORY` from `FOUNDATION_TOTAL_PROJECTS_DETAIL`, and the latest-score selection + case-insensitive `LOWER(d.PROJECT_SLUG) = h.PROJECT_SLUG` slug match are owned by the dbt model that builds that detail table (not by this service's SQL). The inline comment in `project.service.ts` already documents this fold-in. No code change needed — the denormalized-column approach is the better-performing design.

apps/lfx-one/src/app/modules/dashboards/components/project-health-scores-drawer/project-health-scores-drawer.component.ts
```diff
@@ -143,6 +150,9 @@ export class ProjectHealthScoresDrawerComponent {
   protected readonly scoredLabel: Signal<string> = computed(() => this.initScoredLabel());
 
   protected readonly hasData: Signal<boolean> = computed(() => this.scoredProjects() > 0);
+  // Gates the chart itself: a foundation whose projects are all unscored still has a bar to
+  // draw (the trailing Unscored bar), so this must not collapse to hasData() (scored-only).
```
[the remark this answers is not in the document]
audigregorie: Fixed: changed "trailing Unscored bar" → "leading Unscored bar" so the comment matches `PROJECT_HEALTH_CHART_CATEGORIES` (unscored first) and the sibling "leading Unscored" comment above.

luismoriguerra (commented):
## Audit: PR #1108 — perf(dashboards): read health score from dbt detail model (LFXV2-2697)

**Verdict:** 🟡 Minor comments (merge-ready after confirming lf-dbt LFXV2-2696 platinum models are live in Snowflake prod)

Independent workspace audit (head `95b72f85`). Gates re-run: `yarn format:check` ✅ · `yarn lint:check` ✅ · `yarn build` ✅

---

### Summary

Solid fix for [LFXV2-2697](https://linuxfoundation.atlassian.net/browse/LFXV2-2697). Reading the dbt-prejoined `HEALTH_SCORE_CATEGORY` from `FOUNDATION_TOTAL_PROJECTS_DETAIL` is the right performance path and fixes the case-sensitive slug join that caused false "Unscored" rows. The `unscored` chart bucket and category-folding logic correctly align chart ↔ table semantics with `normalizeHealthScoreCategory`.

Pairs with merged lf-dbt [PR #2657](https://github.com/linuxfoundation/lf-dbt/pull/2657) (LFXV2-2696), which folds latest scores into the detail model via `LOWER(p.project_slug) = h.project_slug`.

---

### Prior review threads — status

| Thread | Status |
|--------|--------|
| PR metadata vs dbt column read | ✅ Resolved (description updated) |
| "trailing Unscored bar" comment | ✅ Fixed |
| `overflow-hidden` clipping mini-chart | ✅ Fixed |
| Distribution drops unexpected categories | ✅ Fixed (`else distribution.unscored += …`) |
| Bugbot NaN bar heights without `?? 0` | ✅ Fixed (`foundation-health.component.ts:216`) |
| Drawer `lastUpdated` / cache scope undocumented | ✅ Documented in PR body |
| "null flows through as Unscored" comment | ✅ Fixed (null vs UI label distinguished) |
| "can never disagree" comment accuracy | 🟡 Open — see inline note on existing Copilot thread at `project.service.ts:1700`; reword suggested below |
| Required `unscored` API field | 🟡 Partially — field documented; "Breaking changes: None" still understates additive contract |

---

### Remaining gaps (non-blocking)

1. **Deploy sequencing** — Confirm lf-dbt LFXV2-2696 platinum models have run in Snowflake prod before deploying this BFF change. The detail model did not expose `HEALTH_SCORE_CATEGORY` before that dbt PR.
2. **Comment wording** — `project.service.ts:1699-1700` says chart and table "can never disagree." Snowflake sources are aligned, but the drawer chart uses parent-passed distribution data while the table fetches `/foundation-projects-detail` through a **separate session cache** (`analytics.service.ts`). Suggest: "share the same dbt categorization source" rather than "can never disagree." (Also noted on existing Copilot thread — not duplicating inline.)
3. **Tests** — No specs for distribution category folding, `hasChartData` for all-unscored foundations, or cache eviction on HTTP error.
4. **Type drift** — `ProjectHealthDistribution` still lacks `unscored` while `FoundationHealthScoreDistributionResponse` requires it (inline comment).
5. **Signal consistency** — Mini-card distribution uses `healthScoresData()` directly while the drawer uses `reconciledHealthScoresData()` (inline comment).

---

### Security / privacy

No new auth surface. Both endpoints remain session+bearer gated with slug validation only (pre-existing analytics-lane pattern). No person-level PII added; `lastUpdated` display is UI-only over data already in the API payload.

---

### Q&A

- **Why Tekton showed Unscored:** Old BFF join was case-sensitive (`d.PROJECT_SLUG = h.PROJECT_SLUG`) against lowercased slugs in `PROJECT_HEALTH_METRICS_LATEST`. dbt now owns the `LOWER()` join; this PR stops re-implementing it in service SQL.
- **Why add `unscored` chart bar:** dbt's distribution model buckets NULL as `'Unscored'`, but the UI chart previously showed only 5 scored bars — unscored projects were counted in Snowflake but invisible in the chart.

apps/lfx-one/src/app/modules/dashboards/components/foundation-health/foundation-health.component.ts
```diff
@@ -208,10 +209,12 @@ export class FoundationHealthComponent {
     return computed(() => {
       const distribution = this.healthScoresData();
```
luismoriguerra: **Non-blocking — signal consistency:** `initializeHealthScoreDistribution()` reads `healthScoresData()` directly, but the drawer uses `reconciledHealthScoresData()` (which surfaces `DEFAULT_FOUNDATION_HEALTH_SCORE_DISTRIBUTION` while loading). The loading skeleton masks this today, but using the reconciled signal here would match the documented foundation-switch pattern at lines 104–108 and avoid stale mini-chart buckets if loading state ever changes.

packages/shared/src/interfaces/analytics-data.interface.ts
```diff
@@ -1037,6 +1037,12 @@ export interface FoundationHealthScoreDistributionResponse {
    * Number of projects with critical status
    */
   critical: number;
+
+  /**
+   * Number of projects with no health score (dbt's Unscored bucket -- additive to the
+   * 5 scored categories, so every project counted here has a corresponding chart bar).
+   */
+  unscored: number;
```
luismoriguerra, line 1045: **Non-blocking — type drift (related):** `FoundationHealthScoreDistributionResponse` now requires `unscored`, but `ProjectHealthDistribution` in `dashboard-metric.interface.ts` still has only the 5 scored fields and is what `transformProjectHealthScores` passes to the metric card. Consider adding `unscored: number` there too so typed card consumers cannot miss the new bucket.

packages/shared/src/constants/project-health-scores-drawer.constants.ts
```diff
@@ -15,11 +15,16 @@ export const DEFAULT_FOUNDATION_HEALTH_SCORE_DISTRIBUTION: FoundationHealthScore
...
   critical: 0,
+  unscored: 0,
 };
 
 // Ordered health buckets (low → high), matching the distribution chart's bar order.
 export const PROJECT_HEALTH_SCORE_CATEGORIES: readonly FoundationHealthScore[] = ['critical', 'unsteady', 'stable', 'healthy', 'excellent'];
 
+// Chart bar order including the additive "unscored" bucket (dbt's COALESCE(..., 'Unscored')
+// row) as a leading bar, so every table row -- scored or not -- maps to a chart bar.
+export const PROJECT_HEALTH_CHART_CATEGORIES: readonly (FoundationHealthScore | 'unscored')[] = ['unscored', ...PROJECT_HEALTH_SCORE_CATEGORIES];
```
ahmedomosanya: **Nit / non-blocking (DRY):** `FoundationHealthScore | 'unscored'` is inlined three times in this file — `PROJECT_HEALTH_CHART_CATEGORIES` (here), `PROJECT_HEALTH_CHART_CATEGORY_LABEL` (L68), and `PROJECT_HEALTH_CHART_CATEGORY_COLOR` (L75). We already have this exact union named in `dashboard-metric.interface.ts`:

```ts
export type HealthStatusFilterValue = FoundationHealthScore | 'unscored';
```

Could we reuse it so the chart bucket set has a single source of truth? e.g.:

```ts
export const PROJECT_HEALTH_CHART_CATEGORIES: readonly HealthStatusFilterValue[] = ['unscored', ...PROJECT_HEALTH_SCORE_CATEGORIES];
export const PROJECT_HEALTH_CHART_CATEGORY_LABEL: Record<HealthStatusFilterValue, string> = { ... };
export const PROJECT_HEALTH_CHART_CATEGORY_COLOR: Record<HealthStatusFilterValue, string> = { ... };
```

If "filter value" reads wrong for chart bars, a one-line alias (`export type FoundationHealthChartCategory = HealthStatusFilterValue;`) keeps intent clear. Either way avoids re-declaring the same union in three spots.

ahmedomosanya (dismissed):
LGTM. The switch to the prejoined dbt `HEALTH_SCORE_CATEGORY` column is a clean perf win over the request-time `QUALIFY ROW_NUMBER()` derivation, and the additive `unscored` bucket is threaded consistently through the server aggregation, shared constants, the chart/legend, the drawer, and the request cache. The `?? 0` fallbacks and the else-branch category folding also cover the earlier bot findings. One non-blocking DRY nit inline — not a blocker to merge.

MRashad26 (dismissed):
Reviewed against code standards (signals, templates, shared interfaces/constants, security). All clean — approving on behalf of MRashad26.

MRashad26 (approved):

ahmedomosanya (approved):

PR #1116, opened 2026-07-16
fix(profile): clear pending recovery toast flag on unrelated refresh

fayazg:
## Summary
- Clear `pendingForwardRecoveryToast` in `updateForward()` and `retry()` before they trigger their own refresh, so an unrelated refetch can't consume a stale flag left over from an earlier `FORWARD_SET_FAILED` claim
- Follow-up nit from PR #1110's post-merge review — the flag isn't reachable via this path today, since the `loading()` gate unmounts the edit/retry UI for the full duration of any in-flight refetch, but this closes the gap defensively if that gate is ever relaxed

Fixes #LFXV2-2703


dealako:
Hey @fayazg 👋

Reviewed this closely — traced `updateForward()`/`retry()`/`initData()`/`maybeShowForwardRecoveryToast()` and the template, plus two parallel focused passes (security-auditor, code-reviewer-pro). This is a clean, correctly-scoped defensive fix. I confirmed against the template that `loading()` unmounts the entire `@switch` block (both the Retry button and the forward-edit Save button) for the full duration of any in-flight refresh, so `pendingForwardRecoveryToast` genuinely isn't reachable via `retry()`/`updateForward()` today — the fix is preemptive hardening, not a live-bug fix, matching the PR description exactly. The synchronous flag-clear-before-`refresh.next()` ordering is also correct: it's guaranteed to run before the async `tap()` that consumes the flag, so there's no race even with the underlying HTTP calls being async.

**Issue counts:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — the two added comment blocks are near-identical boilerplate between `updateForward()` and `retry()`; could differentiate wording per call site, purely stylistic (left inline)
- ❔ Question: 0

**Bot reconciliation:** Cursor (Low Risk, no findings), CodeRabbit (no actionable comments), Copilot (no new comments) — agree with all three, nothing to add on top.

apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
```diff
@@ -172,6 +172,9 @@ export class ProfileLinuxEmailComponent {
       .pipe(finalize(() => this.savingForward.set(false)))
       .subscribe({
         next: () => {
+          // Defensive: this refresh isn't the FORWARD_SET_FAILED recovery refetch, so it
+          // must not consume a stale pending flag from an earlier claim() failure.
```
dealako: [nit] Duplicate comment wording across two call sites

Issue: This defensive comment is copy-pasted verbatim in `retry()` (lines 190-191) despite the two call sites having different trigger contexts (successful save vs. user-initiated retry).
Proof: Compare these lines to 190-191 — identical text.
Why it matters: Readability/maintenance only — no behavioral difference.
Fix: Optional — could tailor wording per call site, e.g. "...so a successful save doesn't consume a stale flag from an earlier claim() failure." Not required for merge.

dealako (dismissed):
Independent review (self + parallel security-auditor and code-reviewer-pro passes) confirms this is a correct, appropriately-scoped defensive fix. See the summary comment above for details. One optional nit left inline.

apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
fayazg: Fixed — tailored each comment to its actual call site: `updateForward()` now says "a successful save isn't the FORWARD_SET_FAILED recovery refetch", and `retry()` says "a user-initiated retry isn't...". Good catch, thanks!

See commit cad5cc9b4: apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts.

fayazg:
## Review Feedback Addressed

Commit: cad5cc9b4

### Changes Made
- **apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts**: tailored the defensive comment wording per call site instead of duplicating the same text verbatim — `updateForward()` now references "a successful save", `retry()` references "a user-initiated retry" (per @dealako)

### Threads Resolved

dealako:
Thanks @fayazg — quick turnaround on that thread!

👏 **Nice work:** the reworded comments in `updateForward()` and `retry()` now explain *why* each reset happens in terms of that specific call site (successful save vs. user-initiated retry) instead of duplicating the same generic sentence twice — genuinely clearer to read, not just a cosmetic tweak.

**Revision tracking:**
- ✅ Resolved — duplicate comment wording nit (commit `cad5cc9b4`)

Re-ran independent analysis plus parallel security-auditor and code-reviewer-pro passes on this new commit (comment-only diff, no logic/control-flow changes) — all clean, no new issues. Copilot's re-scan also generated no new comments.

**Issue counts:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

dealako (approved):
Follow-up re-review of commit cad5cc9b4: comment-only change, fully resolves the prior nit, no new issues found (independent + parallel security-auditor/code-reviewer-pro passes all clean). See summary comment for details.

PR #1138, opened 2026-07-21
fix(crowdfunding): add project fund type instead of general fund fallback

emlimlf:
## Summary

- Initiatives with upstream `initiative_type: "project"` were displaying as "General Fund" in the initiatives list and detail views
- `FundType` only defined `general_fund`, `security_audit`, `mentorship`, `event`, so the mapper's fallback (`toValidFundType`) silently coerced the unrecognized `"project"` value to `general_fund`
- Added `FundType.PROJECT` as a real enum member with its own label ("Project"), icon, color, and avatar classes — no fallback coercion needed, and no other code changes required since the mapper and display components already read off the shared `FundType` maps

## Notes

- No JIRA ticket — bug was found visually, no ticket exists for this

### DIFF
--- a/packages/shared/src/constants/crowdfunding.constants.ts
+++ b/packages/shared/src/constants/crowdfunding.constants.ts
@@ -21,6 +21,7 @@ import type {
 
 export const CROWDFUNDING_FUND_TYPE_LABELS: Record<FundType, string> = {
   [FundType.GENERAL_FUND]: 'General Fund',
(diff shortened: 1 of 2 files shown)


MRashad26 (approved):
LGTM — `FundType.PROJECT` enum member added with all 4 `Record<FundType, ...>` map entries filled in. Clean, minimal, correct.

PR #1141, opened 2026-07-21
fix(crowdfunding): scope recurring charge history to subscription only

emlimlf:
## Summary

- The recurring donation detail page's charge history was showing all of the caller's donations to an initiative instead of just the charges belonging to the specific recurring subscription being viewed.
- Adds a `subscriptionOnly` filter threaded through the frontend service, Express controller, and server service to the upstream ledger endpoint, matching [lfx-crowdfunding#233](https://github.com/linuxfoundation/lfx-crowdfunding/pull/233).

## Notes

- Depends on the upstream `subscriptionOnly` support already merged in lfx-crowdfunding#233.


MRashad26 (approved):
LGTM — `subscriptionOnly` filter threads correctly from frontend service → BFF controller (validated + authenticated) → BFF service → upstream. Switching from `getInitiativeTransactions` + `kind: recurring` to the new donor-scoped `getMyInitiativeTransactions` + `subscriptionOnly: true` is the right fix. Route ordering is safe (`/my-transactions` is a distinct sub-path under `/:slug`). No template changes, no secrets.

PR #1149, opened 2026-07-22
fix(sidebar): differentiate profile menu aria-labels

fayazg:
## Summary

- The me-card overflow (⋯) trigger button in the sidebar and the `<nav>` inside
  the popover it opens both declared `aria-label="Profile sections"`, so screen
  readers announced the same label twice. Relabel the button to
  `Open profile sections` so the disclosure control and the navigation region
  read distinctly; the `<nav>` keeps `Profile sections`.
- Follow-up to the me-lens user-card redesign (#1136).

Jira: LFXV2-2741

### DIFF
--- a/apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.html
+++ b/apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.html
@@ -102,7 +102,7 @@
             <button
               type="button"
               (click)="toggleProfileMenu($event)"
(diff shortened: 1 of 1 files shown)


MRashad26 (approved):
LGTM — one-line a11y fix. Disclosure button relabelled to "Open profile sections" so it reads distinctly from the `<nav aria-label="Profile sections">` inside the popover. Correct fix for the duplicate-label screen-reader issue.

PR #1151, opened 2026-07-22
fix(profile): pin profile panel as fixed full-height right rail

fayazg:
## Summary

- Convert the profile summary panel (`lfx-profile-panel`) from a 2xl-only sticky column — which dropped out of the column and stacked full-width on top of the content below 1440px — into a fixed, full-height **300px rail pinned to the right edge at every screen size**, mirroring the left menu (white background, left border, independently scrollable when the viewport is short).
- The rail never stacks above the content, never changes width, and sits above page content (`z-40`); its top offset follows the impersonation banner.
- Reserve a matching right gutter (`pr-[300px]`) on `<main>` for the profile hub so page content and the shared footer stay clear of the rail.
- Track the profile hub in `MainLayoutComponent` via an `isProfileHub` signal (seeded from the URL on init for first-paint correctness, updated on navigation).

JIRA: LFXV2-2738

### DIFF
--- a/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
+++ b/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
@@ -75,11 +75,7 @@
 
   <!-- Main Content Area -->
   <!-- On the profile hub, reserve a right gutter (pr-[300px]) so content and the footer clear the fixed profile rail. -->
-  <!-- Transition is scoped to margin (not `all`) so the profile gutter applies instantly rather than animating over 300ms. -->
-  <main
-    class="flex-1 min-w-0 transition-[margin] duration-300 lg:ml-[348px] flex flex-col"
-    [ngClass]="{ 'pr-[300px]': isProfileHub() }"
-    data-testid="main-content">
+  <main class="flex-1 min-w-0 lg:ml-[348px] flex flex-col" [ngClass]="{ 'pr-[300px]': isProfileHub() }" data-testid="main-content">
     <!-- Mobile Menu Button -->
     <div
       class="lg:hidden sticky z-30 bg-white border-b border-gray-200 px-5 py-3"
--- a/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
+++ b/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
@@ -74,8 +74,12 @@
   }
 
   <!-- Main Content Area -->
-  <!-- On the profile hub, reserve a right gutter (pr-[300px]) so content and the footer clear the fixed profile rail -->
-  <main class="flex-1 min-w-0 transition-all duration-300 lg:ml-[348px] flex flex-col" [ngClass]="{ 'pr-[300px]': isProfileHub() }" data-testid="main-content">
+  <!-- On the profile hub, reserve a right gutter (pr-[300px]) so content and the footer clear the fixed profile rail. -->
+  <!-- Transition is scoped to margin (not `all`) so the profile gutter applies instantly rather than animating over 300ms. -->
+  <main
+    class="flex-1 min-w-0 transition-[margin] duration-300 lg:ml-[348px] flex flex-col"
+    [ngClass]="{ 'pr-[300px]': isProfileHub() }"
+    data-testid="main-content">
     <!-- Mobile Menu Button -->
     <div
       class="lg:hidden sticky z-30 bg-white border-b border-gray-200 px-5 py-3"
--- a/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
+++ b/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
@@ -74,7 +74,8 @@
   }
 
   <!-- Main Content Area -->
-  <main class="flex-1 min-w-0 transition-all duration-300 lg:ml-[348px] flex flex-col" data-testid="main-content">
+  <!-- On the profile hub, reserve a right gutter (pr-[300px]) so content and the footer clear the fixed profile rail -->
+  <main class="flex-1 min-w-0 transition-all duration-300 lg:ml-[348px] flex flex-col" [ngClass]="{ 'pr-[300px]': isProfileHub() }" data-testid="main-content">
     <!-- Mobile Menu Button -->
     <div
       class="lg:hidden sticky z-30 bg-white border-b border-gray-200 px-5 py-3"
--- a/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
+++ b/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
@@ -116,9 +116,10 @@ export class MainLayoutComponent {
     }
   }
 
-  /** Whether the current URL is under the /profile hub (drops any query string before matching). */
+  /** Whether the current URL is under the /profile hub (drops any query string, matches on segment boundary). */
   private computeIsProfileHub(): boolean {
-    return this.router.url.split('?')[0].startsWith('/profile');
+    const path = this.router.url.split('?')[0];
+    return path === '/profile' || path.startsWith('/profile/');
   }
 
   /**
--- a/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
+++ b/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
@@ -46,6 +46,12 @@ export class MainLayoutComponent {
   // Lens-aware sidebar items (built by SidebarNavService; shared with the docs shell).
   protected readonly sidebarItems = this.sidebarNavService.sidebarItems;
 
+  // True on the /profile hub, where the fixed profile rail is shown. Drives the right-side gutter
+  // (pr-[300px]) on <main> so page content and the shared footer stay clear of the rail. Only the
+  // me-lens /profile route uses ProfileLayoutComponent; org/foundation/project profile pages live
+  // under lens-prefixed paths (e.g. /org/...), so a /profile prefix match is exact enough here.
+  protected readonly isProfileHub = signal(false);
+
   /**
    * A route lens that {@link LensService.setLens} refused, held for retry.
    *
@@ -58,6 +64,10 @@ export class MainLayoutComponent {
   private readonly pendingRouteLens = signal<Lens | null>(null);
 
   public constructor() {
+    // Seed the profile-hub flag from the current URL so the right gutter is correct on the first
+    // paint / hard load, before the first NavigationEnd fires.
+    this.isProfileHub.set(this.computeIsProfileHub());
+
     // Close mobile sidebar and sync lens from route data on navigation
     this.router.events
       .pipe(
@@ -67,6 +77,7 @@ export class MainLayoutComponent {
       .subscribe(() => {
         this.appService.closeMobileSidebar();
         this.selectorPanelOpen.set(false);
+        this.isProfileHub.set(this.computeIsProfileHub());
         this.syncLensFromRoute();
       });
 
@@ -103,6 +114,11 @@ export class MainLayoutComponent {
     if (!visible) {
       this.appService.closeMobileSidebar();
     }
+  }
+
+  /** Whether the current URL is under the /profile hub (drops any query string before matching). */
+  private computeIsProfileHub(): boolean {
+    return this.router.url.split('?')[0].startsWith('/profile');
   }
 
   /**
--- a/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
+++ b/apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
@@ -10,6 +10,7 @@ import { LensSwitcherComponent } from '@components/lens-switcher/lens-switcher.c
 import { SidebarComponent } from '@components/sidebar/sidebar.component';
 import { ALL_LENSES } from '@lfx-one/shared/constants';
 import { Lens } from '@lfx-one/shared/interfaces';
+import { isProfileHubPath } from '@lfx-one/shared/utils';
 import { AppService } from '@services/app.service';
 import { LensService } from '@services/lens.service';
 import { ProjectContextService } from '@services/project-context.service';
@@ -46,10 +47,8 @@ export class MainLayoutComponent {
   // Lens-aware sidebar items (built by SidebarNavService; shared with the docs shell).
   protected readonly sidebarItems = this.sidebarNavService.sidebarItems;
 
-  // True on the /profile hub, where the fixed profile rail is shown. Drives the right-side gutter
-  // (pr-[300px]) on <main> so page content and the shared footer stay clear of the rail. Only the
-  // me-lens /profile route uses ProfileLayoutComponent; org/foundation/project profile pages live
-  // under lens-prefixed paths (e.g. /org/...), so a /profile prefix match is exact enough here.
+  // True on the /profile hub — drives the pr-[300px] right gutter on <main> so content/footer
+  // clear the fixed rail. Only the me-lens /profile route uses ProfileLayoutComponent.
   protected readonly isProfileHub = signal(false);
 
   /**
@@ -66,7 +65,7 @@ export class MainLayoutComponent {
   public constructor() {
     // Seed the profile-hub flag from the current URL so the right gutter is correct on the first
     // paint / hard load, before the first NavigationEnd fires.
-    this.isProfileHub.set(this.computeIsProfileHub());
+    this.isProfileHub.set(isProfileHubPath(this.router.url));
 
     // Close mobile sidebar and sync lens from route data on navigation
     this.router.events
@@ -77,7 +76,7 @@ export class MainLayoutComponent {
       .subscribe(() => {
         this.appService.closeMobileSidebar();
         this.selectorPanelOpen.set(false);
-        this.isProfileHub.set(this.computeIsProfileHub());
+        this.isProfileHub.set(isProfileHubPath(this.router.url));
         this.syncLensFromRoute();
       });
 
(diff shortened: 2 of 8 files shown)


apps/lfx-one/src/app/layouts/profile-layout/profile-panel/profile-panel.component.html
```diff
@@ -2,13 +2,11 @@
...
-  set up by the layout, with its own scroll if the content exceeds the viewport height.
+  Presentational summary card. The layout wraps this in a fixed full-height right rail; here we
+  fill that rail (h-full) with a white, flush surface and a left border mirroring the left menu,
+  scrolling its own content independently (overflow-y-auto) when the viewport is short.
 -->
-<aside
-  class="flex flex-col border-b border-slate-200 bg-slate-50 px-6 py-7 2xl:max-h-[calc(100vh-4rem)] 2xl:overflow-y-auto 2xl:rounded-xl 2xl:border"
-  aria-label="Profile summary"
-  data-testid="profile-panel">
+<aside class="flex h-full flex-col overflow-y-auto border-l border-gray-200 bg-white px-6 py-7" aria-label="Profile summary" data-testid="profile-panel">
```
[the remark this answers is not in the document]
fayazg: Done — added `host: { class: 'block h-full' }` to `ProfilePanelComponent` so the host stretches to the fixed rail's height and the inner `aside`'s `h-full`/`overflow-y-auto` now resolves (mirrors lfx-sidebar's `:host { height: 100% }`). The rail scrolls independently when content exceeds the viewport.

See commit 2a770d728.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.html
```diff
@@ -2,15 +2,15 @@
...
+  content, never changes width, and always sits above page content (z-40). MainLayoutComponent
+  reserves a matching right gutter (pr-[300px]) on the profile hub so the content and the shared
+  footer stay clear of the rail. The rail's `top` matches the impersonation banner offset.
 -->
-<div class="flex flex-col gap-6 2xl:flex-row 2xl:items-start" data-testid="profile-layout">
-  <!-- Profile panel — first in the DOM so it stacks on top below 2xl; sticky right column at 2xl -->
-  <div class="2xl:sticky 2xl:order-2 2xl:w-80 2xl:shrink-0" [class.2xl:top-12]="impersonating()" [class.2xl:top-6]="!impersonating()">
+<div class="flex flex-col" data-testid="profile-layout">
+  <!-- Profile panel — fixed full-height right rail at all widths (out of normal flow) -->
+  <div class="fixed bottom-0 right-0 z-40 w-[300px]" [class.top-12]="impersonating()" [class.top-0]="!impersonating()">
```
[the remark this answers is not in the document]
fayazg: Done — updated both JSDocs to describe the fixed full-height rail. `profile-panel.component.ts` no longer says "sticky … stacked below", and `profile-layout.component.ts` no longer calls it a "sticky profile panel"; both now match the templates.

See commit 2a770d728.

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
```diff
@@ -74,7 +74,8 @@
   }
 
   <!-- Main Content Area -->
-  <main class="flex-1 min-w-0 transition-all duration-300 lg:ml-[348px] flex flex-col" data-testid="main-content">
+  <!-- On the profile hub, reserve a right gutter (pr-[300px]) so content and the footer clear the fixed profile rail -->
+  <main class="flex-1 min-w-0 transition-all duration-300 lg:ml-[348px] flex flex-col" [ngClass]="{ 'pr-[300px]': isProfileHub() }" data-testid="main-content">
```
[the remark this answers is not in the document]
fayazg: Good catch — scoped the `<main>` transition from `transition-all` to `transition-[margin]`, so the profile gutter (`pr-[300px]`) now applies instantly instead of animating over 300ms. Content no longer sits under the rail on entering /profile, and the gutter no longer lingers on leaving; the sidebar margin still transitions as before.

See commit 2a770d728.

fayazg:
## Review Feedback Addressed

Commit: 2a770d728

### Changes Made
- **profile-panel.component.ts**: set host to `block h-full` so the aside's `h-full`/`overflow-y-auto` resolves against the fixed rail height — the rail now scrolls independently when content is tall (per cursor[bot])
- **profile-layout.component.ts / profile-panel.component.ts**: updated stale JSDoc that still described the old sticky/stacked layout to the fixed-rail behavior (per copilot[bot])
- **main-layout.component.html**: scoped the `<main>` transition to `[margin]` so the profile gutter (`pr-[300px]`) applies instantly instead of animating over 300ms (per coderabbitai[bot])

### Threads Resolved

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
```diff
@@ -74,7 +74,12 @@
   }
 
   <!-- Main Content Area -->
-  <main class="flex-1 min-w-0 transition-all duration-300 lg:ml-[348px] flex flex-col" data-testid="main-content">
+  <!-- On the profile hub, reserve a right gutter (pr-[300px]) so content and the footer clear the fixed profile rail. -->
+  <!-- Transition is scoped to margin (not `all`) so the profile gutter applies instantly rather than animating over 300ms. -->
+  <main
+    class="flex-1 min-w-0 transition-[margin] duration-300 lg:ml-[348px] flex flex-col"
+    [ngClass]="{ 'pr-[300px]': isProfileHub() }"
```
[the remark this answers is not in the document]
fayazg: Thanks for flagging this — the analysis is accurate. The fixed 300px rail at every breakpoint is an intentional product requirement for the Profile hub: the profile column must never stack above the content, never change width, and stay above page content at any screen size. The resulting narrow center column on small screens is an accepted, documented trade-off (called out in the PR description).

We're knowingly accepting the narrow-viewport behavior for now rather than adding a small-screen exception. If a dedicated mobile treatment is needed later, it'll be handled as a follow-up.

fayazg:

### No Change Needed
- **main-layout.component.html:81**: the always-300px fixed rail at every breakpoint is an intentional product requirement for the Profile hub (never stacks, never changes width, always above content, at any screen size). The narrow center column on small screens is an accepted, documented trade-off; any dedicated mobile treatment is deferred to a follow-up (flagged by copilot[bot])

### Threads Resolved

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
```diff
@@ -105,6 +116,11 @@ export class MainLayoutComponent {
     }
   }
 
+  /** Whether the current URL is under the /profile hub (drops any query string before matching). */
+  private computeIsProfileHub(): boolean {
+    return this.router.url.split('?')[0].startsWith('/profile');
```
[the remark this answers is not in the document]
fayazg: Good catch — fixed. `computeIsProfileHub` now matches on the route-segment boundary (`url === '/profile' || url.startsWith('/profile/')`) instead of a bare `startsWith('/profile')`, so unrelated routes like `/profiles` or `/profile-old` no longer receive the 300px gutter. (`/org/profile` was already unaffected since it isn't `/profile`-prefixed.)

See commit 46e918dfc.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.html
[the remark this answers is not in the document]
fayazg: Thanks for flagging this. The rail's `top-12` is intentionally matched to the shared impersonation banner's `min-h-12`. The banner is a single `flex` row with no `flex-wrap`, so items don't wrap to a second line; the only way it exceeds 3rem is a long email wrapping inside its own span on a very narrow viewport.

We're accepting that as a documented narrow-viewport trade-off for now — the same class of edge case already called out and accepted on this PR — rather than changing the shared banner's height contract, which affects every page that renders it. If a guaranteed banner-height invariant becomes necessary, it'll be handled as a follow-up on the shared component. No code change here.

fayazg:
## Review Feedback Addressed

Commit: 46e918dfc

### Changes Made
- **main-layout.component.ts**: `computeIsProfileHub` now matches the route-segment boundary (`url === '/profile' || url.startsWith('/profile/')`) instead of a bare `startsWith('/profile')`, so unrelated routes like `/profiles` or `/profile-old` no longer get the 300px profile gutter (per copilot-pull-request-reviewer).

### No Change Needed
- **profile-layout.component.html:13**: The rail's `top-12` offset intentionally matches the shared impersonation banner's `min-h-12`. The banner is a single non-wrapping flex row, so it only exceeds 3rem when a long email wraps inside its span on a very narrow viewport — an accepted narrow-viewport trade-off (same class already documented on this PR). Changing the shared banner's height contract would affect every page that renders it; deferred to a follow-up if a guaranteed invariant is needed (flagged by copilot-pull-request-reviewer).

### Threads Resolved

audigregorie (commented):
## Code Review Summary

Clean conversion of the profile panel to a fixed right rail — the height chain (`host: block h-full` + `top-12`/`bottom-0`) is correct and the first-paint seeding of `isProfileHub` is well handled. One Major: the rail and its right gutter are not gated by `lg:`, so on mobile/tablet a 300px fixed rail plus a 300px gutter squeezes content to ~75px on a 375px viewport and regresses the previous stacked-full-width mobile behavior. A few minor cleanups noted inline.

### What's done well

- Height chain mirrors `lfx-sidebar`'s `:host { height: 100% }` so the rail scrolls independently when the viewport is short.
- `isProfileHub` is seeded from `router.url` in the constructor for correct first paint before the first `NavigationEnd`.
- Segment-boundary route matching correctly excludes `/profiles` and `/profile-old`.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.html
audigregorie, line 13: **Major — fixed rail and gutter are not responsive; mobile/tablet regression**

The rail (`fixed bottom-0 right-0 z-40 w-[300px]`) and the matching `<main>` gutter (`pr-[300px]` in main-layout.component.html) have no `lg:` gating, so they apply at every breakpoint. On a 375px phone the gutter leaves ~75px for content and the rail covers most of the viewport; on a 768px tablet only ~468px remains. The left menu handles small screens by hiding (`hidden lg:flex`) and switching to a drawer — the rail doesn't. The previous design stacked the panel full-width above content below 2xl, so this regresses mobile.

Restore in-flow stacking below `lg` and only pin to a fixed rail at `lg+` (the `[class.lg:top-12]` form is already used elsewhere in this file). The `<main>` gutter should likewise become `lg:pr-[300px]` so mobile doesn't reserve the 300px.

```suggestion
  <div class="w-full lg:fixed lg:bottom-0 lg:right-0 lg:z-40 lg:w-[300px]" [class.lg:top-12]="impersonating()" [class.lg:top-0]="!impersonating()">
```

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
```diff
@@ -74,7 +74,12 @@
   }
 
   <!-- Main Content Area -->
-  <main class="flex-1 min-w-0 transition-all duration-300 lg:ml-[348px] flex flex-col" data-testid="main-content">
+  <!-- On the profile hub, reserve a right gutter (pr-[300px]) so content and the footer clear the fixed profile rail. -->
+  <!-- Transition is scoped to margin (not `all`) so the profile gutter applies instantly rather than animating over 300ms. -->
+  <main
+    class="flex-1 min-w-0 transition-[margin] duration-300 lg:ml-[348px] flex flex-col"
```
audigregorie: **Minor — `transition-[margin] duration-300` no longer animates anything**

The only margin is the static `lg:ml-[348px]`, and the profile gutter is padding (intentionally not animated per the comment above). No margin changes at runtime, so this transition is now dead. Either drop it or keep a one-line note on why it's retained.

```suggestion
    class="flex-1 min-w-0 lg:ml-[348px] flex flex-col"
```

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
```diff
@@ -46,6 +46,12 @@ export class MainLayoutComponent {
   // Lens-aware sidebar items (built by SidebarNavService; shared with the docs shell).
   protected readonly sidebarItems = this.sidebarNavService.sidebarItems;
 
+  // True on the /profile hub, where the fixed profile rail is shown. Drives the right-side gutter
+  // (pr-[300px]) on <main> so page content and the shared footer stay clear of the rail. Only the
+  // me-lens /profile route uses ProfileLayoutComponent; org/foundation/project profile pages live
+  // under lens-prefixed paths (e.g. /org/...), so a /profile prefix match is exact enough here.
+  protected readonly isProfileHub = signal(false);
```
audigregorie: **Minor — field comment exceeds the two-line cap**

Repo rule limits comments to 1–2 lines ("No exceptions"). Trim to the why; the route-prefix detail is already conveyed by `computeIsProfileHub`.

```suggestion
  // True on the /profile hub — drives the pr-[300px] right gutter on <main> so content/footer
  // clear the fixed rail. Only the me-lens /profile route uses ProfileLayoutComponent.
  protected readonly isProfileHub = signal(false);
```

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
```diff
@@ -105,6 +116,12 @@ export class MainLayoutComponent {
     }
   }
 
+  /** Whether the current URL is under the /profile hub (drops any query string, matches on segment boundary). */
+  private computeIsProfileHub(): boolean {
```
audigregorie: **Minor — `computeIsProfileHub` is private and untested**

The segment-boundary matching here is pure logic, and this PR's own history shows a `startsWith('/profile')` bug was caught and fixed in a follow-up commit. Extract it to a pure static function (taking the url string) or add a unit test so the boundary check can't silently regress against `/profiles` / `/profile-old`.

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
```diff
@@ -105,6 +116,12 @@ export class MainLayoutComponent {
...
+    const path = this.router.url.split('?')[0];
```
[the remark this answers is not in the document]
fayazg: Good catch — fixed. The match now runs through a shared `isProfileHubPath(url)` helper that strips **both** the query string and the fragment (`url.split(/[?#]/)[0]`) before the segment-boundary check, so `/profile#developer-settings` and the `/settings#developer-settings` → `/profile/settings#developer-settings` redirect keep the 300px gutter. Covered by a new unit test case.

See commit 3c92fd2a9.

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.html
fayazg: Fixed — dropped `transition-[margin] duration-300`. You're right: the sidebar margin (`lg:ml-[348px]`) only changes at the breakpoint, never at runtime (the mobile sidebar is an overlay drawer, not a margin shift), so the transition animated nothing. Good catch.

See commit 3c92fd2a9.

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
fayazg: Fixed — trimmed the field comment to two lines (the why), since the route-prefix detail now lives in the shared `isProfileHubPath` helper and its tests.

See commit 3c92fd2a9.

apps/lfx-one/src/app/layouts/main-layout/main-layout.component.ts
fayazg: Done — extracted the matching logic into a pure `isProfileHubPath(url: string)` in `packages/shared/src/utils/url.utils.ts` and added `url.utils.spec.ts` with cases for `/profile`, `/profile/…`, `/profiles`, `/profile-old`, and query/fragment variants, so the boundary can't silently regress. The component just calls the helper now.

See commit 3c92fd2a9.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.html
fayazg: Thanks @audigregorie — you've read the behavior exactly right, and the narrow center column below `lg` is real. This one is an intentional product call rather than an oversight: for the Profile & Account hub the panel must stay a fixed-width right rail at every breakpoint — never stacking above content and never changing width — which is why there's no `lg:` gating and the gutter is unconditional. The tight mobile/tablet column is a knowingly-accepted trade-off for this PR (the same point Copilot raised earlier).

We're planning to handle a dedicated small-screen treatment (drawer/collapsible rail) as a follow-up rather than reintroducing the stacked layout here. I'll leave this thread open so you can weigh in — if you feel the mobile regression is blocking for this PR, happy to reopen the discussion with product before we merge.

fayazg:
## Review Feedback Addressed

Commit: 3c92fd2a9

### Changes Made
- **packages/shared/src/utils/url.utils.ts**: added a pure `isProfileHubPath(url)` that strips both the query string and the fragment before a segment-boundary match, so `/profile#developer-settings` and the Developer Settings redirect keep the 300px gutter (per copilot-pull-request-reviewer).
- **packages/shared/src/utils/url.utils.spec.ts**: new vitest coverage for `/profile`, `/profile/…`, `/profiles`, `/profile-old`, and query/fragment cases so the boundary can't silently regress (per @audigregorie).
- **main-layout.component.ts**: use the shared helper, drop the private `computeIsProfileHub`, and trim the `isProfileHub` field comment to the two-line cap (per @audigregorie).
- **main-layout.component.html**: dropped the dead `transition-[margin] duration-300` — the sidebar margin never changes at runtime (per @audigregorie).

### Threads Resolved
4 of 5 unresolved threads addressed in this iteration.

### Still Open
- **profile-layout.component.html:13** — the always-fixed 300px rail at every breakpoint (flagged by @audigregorie as a mobile/tablet regression). Left open by design: the fixed-width, non-stacking rail is an intentional product requirement for the Profile hub, and the narrow small-screen column is a knowingly-accepted trade-off for this PR. A dedicated small-screen treatment is planned as a follow-up. Left open for @audigregorie to confirm or escalate to product before merge.

audigregorie (approved):

PR #1172, opened 2026-07-23
feat(dashboards): add per-event detail drawer to roster

mrautela365:
## Summary
Adds a slide-in **per-event detail drawer** opened from an event roster row (`p-drawer`, lazy-loaded on open). Sourced from `MARKETING_EVENT_REGISTRATIONS` + `MARKETING_EVENT_SPONSORSHIPS_BY_TIER`.

**PR 3/6 of the LF Events dashboard stack (LFXV2-2768). Base: `feat/LFXV2-2768-event-roster` (PR).** Review/merge PRs 1–2 first.

LFXV2-2768

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
@@ -1,0 +1,93 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { NgClass } from '@angular/common';
+import { ChangeDetectionStrategy, Component, computed, inject, input, model, Signal } from '@angular/core';
+import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { formatCurrency, formatNumber } from '@lfx-one/shared/utils';
+import { AnalyticsService } from '@services/analytics.service';
+import { DrawerModule } from 'primeng/drawer';
+import { finalize, of, skip, switchMap } from 'rxjs';
+
+import type { EventDetailResponse } from '@lfx-one/shared/interfaces';
+
+@Component({
+  selector: 'lfx-event-detail-drawer',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [NgClass, DrawerModule],
+  templateUrl: './event-detail-drawer.component.html',
+})
+export class EventDetailDrawerComponent {
+  private readonly analyticsService = inject(AnalyticsService);
+
+  // === Model Signals (two-way binding) ===
+  public readonly visible = model<boolean>(false);
+
+  // === Inputs ===
+  /** Event id to load when the drawer opens. */
+  public readonly eventId = input<string | null>(null);
+
+  // === Computed Signals ===
+  protected readonly detail: Signal<EventDetailResponse | null> = this.initDetail();
+  protected readonly loading = computed(() => this.visible() && this.detail() === null && this.eventId() !== null);
+
+  // Registration progress (0–100) when a real goal exists.
+  protected readonly regProgress = computed(() => {
+    const d = this.detail();
+    if (!d || d.registrations.goal <= 0) return null;
+    return Math.min(100, Math.round((d.registrations.actual / d.registrations.goal) * 100));
+  });
+  // Sponsorship progress (0–100) when a real goal exists.
+  protected readonly sponProgress = computed(() => {
+    const d = this.detail();
+    if (!d || d.sponsorshipRevenue.goal <= 0) return null;
+    return Math.min(100, Math.round((d.sponsorshipRevenue.actual / d.sponsorshipRevenue.goal) * 100));
+  });
+
+  // === Protected Helpers (template) ===
+  protected num(value: number): string {
+    return formatNumber(value);
+  }
+
+  protected money(value: number): string {
+    return formatCurrency(value);
+  }
+
+  /** Registration pace vs last year as a signed percent string, or null when no baseline. */
+  protected vsLastYearLabel(): string | null {
+    const d = this.detail();
+    if (!d || d.vsLastYear === null) return null;
+    const pct = Math.round((d.vsLastYear - 1) * 100);
+    if (pct > 0) return `+${pct}% vs last year`;
+    if (pct < 0) return `${pct}% vs last year`;
+    return 'On par with last year';
+  }
+
+  protected dateLabel(): string {
+    const iso = this.detail()?.startDate ?? '';
+    const [year, month, day] = iso.split('-').map(Number);
+    if (!year || !month || !day) return iso;
+    return new Date(Date.UTC(year, month - 1, day)).toLocaleDateString('en-US', {
+      month: 'short',
+      day: 'numeric',
+      year: 'numeric',
+      timeZone: 'UTC',
+    });
+  }
+
+  // === Private Initializers ===
+  private initDetail(): Signal<EventDetailResponse | null> {
+    // Lazy-load on open: react to visibility flipping true (skip the initial value).
+    return toSignal(
+      toObservable(this.visible).pipe(
+        skip(1),
+        switchMap((open) => {
+          const id = this.eventId();
+          if (!open || !id) return of(null);
+          return this.analyticsService.getEventDetail(id).pipe(finalize(() => undefined));
+        })
+      ),
+      { initialValue: null }
+    );
+  }
+}
--- a/apps/lfx-one/src/server/controllers/analytics.controller.ts
+++ b/apps/lfx-one/src/server/controllers/analytics.controller.ts
@@ -2781,6 +2781,41 @@ export class AnalyticsController {
   }
 
   /**
+   * GET /api/analytics/event-detail
+   * Per-event detail for the roster drawer (meta, actual-vs-goal, sponsorship-by-tier).
+   */
+  public async getEventDetail(req: Request, res: Response, next: NextFunction): Promise<void> {
+    const startTime = logger.startOperation(req, 'get_event_detail');
+
+    try {
+      const eventId = getStringQueryParam(req, 'eventId');
+
+      if (!eventId) {
+        throw ServiceValidationError.forField('eventId', 'eventId query parameter is required', {
+          operation: 'get_event_detail',
+        });
+      }
+
+      if (!/^[A-Za-z0-9_-]{1,64}$/.test(eventId)) {
+        throw ServiceValidationError.forField('eventId', 'Invalid eventId format', {
+          operation: 'get_event_detail',
+        });
+      }
+
+      const response = await this.projectService.getEventDetail(eventId);
+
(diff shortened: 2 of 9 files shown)


apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    });
+  }
+
+  // === Private Initializers ===
+  private initDetail(): Signal<EventDetailResponse | null> {
+    // Lazy-load on open: react to visibility flipping true (skip the initial value).
+    return toSignal(
+      toObservable(this.visible).pipe(
+        skip(1),
+        switchMap((open) => {
```
dealako, line 84: [blocking] Drawer doesn't refetch when `eventId` changes while it's already open

Issue: `initDetail()` only reacts to `toObservable(this.visible)` — the switchMap fires on visibility flips, not on `eventId` changes. When the drawer is open and the user clicks a different roster row, `eventId` updates but `visible` stays `true`, so this pipe never re-fires.

Proof: `switchMap((open) => { const id = this.eventId(); ... })` reads `eventId()` as a snapshot at the time `visible` last emitted — it does not subscribe to `eventId` itself. The parent's `openEvent()` sets `selectedEventId` independently of `drawerVisible`, so a second click with the drawer already open is a silent no-op here.

Why it matters: rapid clicks between roster rows show a previous event's metrics under the new event's header/title, with no user-visible indication of staleness. This matches Cursor's flagged finding.

Fix: key the pipe off both signals, e.g. `combineLatest([toObservable(this.visible), toObservable(this.eventId)]).pipe(skip(1), switchMap(([open, id]) => (!open || !id ? of(null) : this.analyticsService.getEventDetail(id))))`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  // === Model Signals (two-way binding) ===
+  public readonly visible = model<boolean>(false);
+
+  // === Inputs ===
+  /** Event id to load when the drawer opens. */
+  public readonly eventId = input<string | null>(null);
+
+  // === Computed Signals ===
+  protected readonly detail: Signal<EventDetailResponse | null> = this.initDetail();
+  protected readonly loading = computed(() => this.visible() && this.detail() === null && this.eventId() !== null);
```
dealako, line 32: [blocking] `loading` never clears after an error or no-data response

Issue: `loading` is computed as `visible() && detail() === null && eventId() !== null`. `AnalyticsService.getEventDetail()` maps every HTTP error to `of(null)` (analytics.service.ts:1146), and a legitimate "event not found" response is also `null`. Both cases leave `detail()` at `null` forever, so `loading` stays `true` indefinitely — the empty state in the template is unreachable.

Proof: no signal tracks request completion independently of the result value; `null` is overloaded to mean both "in flight" and "finished, nothing to show."

Why it matters: any API error leaves the user staring at a skeleton loader forever, with no recovery short of closing and reopening the drawer. This matches Copilot's flagged finding.

Fix: track completion explicitly, e.g. a `loaded` signal set via `finalize()` in the pipe, then `loading = computed(() => visible() && !loaded() && eventId() !== null)`, and give the template a real error/empty branch distinct from the loading branch.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -4970,6 +4971,103 @@ export class ProjectService {
...
+    };
+
+    const sponsorshipActual = tierResult.rows.reduce((sum, t) => sum + (t.REVENUE ?? 0), 0);
+
+    return {
+      eventId: row.EVENT_ID,
+      eventName: row.EVENT_NAME,
+      startDate: row.START_DATE,
+      country: row.EVENT_COUNTRY ?? '',
+      eventUrl: row.EVENT_URL ?? '',
```
dealako, line 5057: [minor] `eventUrl` returned to the client without validation

Issue: `eventUrl: row.EVENT_URL ?? ''` passes the Snowflake column straight through with no sanitization, unlike other outbound links in this codebase.

Proof: `validateAndSanitizeUrl` (server/helpers/url-validation.ts) is already used for this exact purpose in `invite.controller.ts` and `survey.service.ts`. This new path bypasses it — the template's `target="_blank"` is a client-side mitigation, not a substitute for server-side validation.

Why it matters: if the Snowflake source is ever poisoned or a bad row slips through, the API would hand back an unvalidated URL directly.

Fix: `eventUrl: row.EVENT_URL ? (validateAndSanitizeUrl(row.EVENT_URL) ?? '') : ''`.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -4970,6 +4971,103 @@ export class ProjectService {
...
+      logger.warning(undefined, 'get_event_detail', 'No event row for id', { event_id: eventId });
+      return null;
+    }
+
+    const normalizeScore = (score: string | null): EventCompScore => {
+      const s = (score ?? '').toLowerCase();
+      return s === 'high' || s === 'medium' || s === 'low' ? s : 'unknown';
+    };
+
+    const sponsorshipActual = tierResult.rows.reduce((sum, t) => sum + (t.REVENUE ?? 0), 0);
```
dealako, line 5050: [minor] Inconsistent optional chaining on Snowflake result rows

Issue: `eventResult.rows?.[0]` (line 5039) is defensively guarded, but `tierResult.rows.reduce(...)` (this line) and `tierResult.rows.map(...)` (line 5063) are not, even though both results come from the same `snowflakeService.execute()` call shape.

Why it matters: if the service ever returns `{ rows: undefined }` for a no-match tier query, this throws at runtime instead of degrading gracefully.

Fix: `(tierResult.rows ?? []).reduce(...)` and `(tierResult.rows ?? []).map(...)` for consistency with the `eventResult` guard above.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    return new Date(Date.UTC(year, month - 1, day)).toLocaleDateString('en-US', {
+      month: 'short',
+      day: 'numeric',
+      year: 'numeric',
+      timeZone: 'UTC',
+    });
+  }
+
+  // === Private Initializers ===
+  private initDetail(): Signal<EventDetailResponse | null> {
```
dealako, line 79: [minor] No Playwright coverage for the new drawer

Issue: this component introduces new loading/error/empty state logic (see the two blocking findings above) with no `*.spec.ts` or `*-robust.spec.ts` coverage, while analogous drawers in this codebase (e.g. attendee/speaker drawers, `apps/lfx-one/e2e/org-events-dashboard.spec.ts:244-258`) are covered.

Why it matters: the state-machine-like transitions here are exactly the kind of thing that regresses silently; this is also how the two blocking bugs above would have been caught. Agrees with Copilot's finding.

Fix: add an E2E spec exercising row-click open, lazy request timing, successful tier rendering, rapid-click event switching (once the eventId bug is fixed), and a terminal null/error result reaching the empty state.

apps/lfx-one/src/server/controllers/analytics.controller.ts
```diff
@@ -2780,6 +2780,41 @@ export class AnalyticsController {
...
+        });
+      }
+
+      if (!/^[A-Za-z0-9_-]{1,64}$/.test(eventId)) {
+        throw ServiceValidationError.forField('eventId', 'Invalid eventId format', {
+          operation: 'get_event_detail',
+        });
+      }
+
+      const response = await this.projectService.getEventDetail(eventId);
```
dealako, line 2805: [question] `event-detail` has no authorization scoping tied to the event's owning foundation

Issue: `getEventDetail` validates `eventId` format only, then returns full event detail (including sponsorship revenue and competitor score) to any authenticated user, regardless of which foundation the event belongs to.

Note: I initially flagged this as blocking, but checked the sibling `getEventRoster` endpoint in this same file — it also has no user/foundation-membership check, only a `foundationSlug` format check. So this isn't a regression introduced by this PR; it's the existing pattern for this whole controller (global `authMiddleware` requires *a* session, but nothing here verifies foundation membership). Not blocking this PR, but worth a tracked follow-up given the sensitivity of the fields involved (see next comment).

Ask: is cross-foundation analytics access intentionally open to any authenticated user today, or is this gap already tracked elsewhere?

packages/shared/src/interfaces/dashboard-metric.interface.ts
```diff
@@ -679,6 +679,37 @@ export interface EventRosterResponse {
...
+  eventId: string;
+  eventName: string;
+  startDate: string;
+  country: string;
+  eventUrl: string;
+  registrations: { actual: number; goal: number };
+  sponsorshipRevenue: { actual: number; goal: number };
+  /** Ratio of this year's registrations to last year's (1.0 = on par); null when no baseline. */
+  vsLastYear: number | null;
+  compScore: EventCompScore;
```
dealako, line 708: [question] Access-control intent for competitive/financial fields isn't documented

Issue: `EventDetailResponse` exposes `compScore` and `vsLastYear` — competitor-comparison and YoY positioning metrics — with no comment on which persona(s) should see this (ED-only vs. any authenticated member, etc.), and the controller applies no role check either way.

Ask: is this intentionally visible to any authenticated user for transparency, or should it be scoped by role/persona? Either answer is fine — just flag the decision here so it's not ambiguous for the next person touching this endpoint.

dealako (changes requested):
Hey @mrautela365 — nice, tightly-scoped addition. The event-detail drawer follows the existing roster patterns well (lazy-loaded on open, same Snowflake dual-query shape as the roster endpoint, consistent signal/computed structure), and the goa-equivalent API surface (route → controller → service → shared interface) is wired correctly end to end. Two real bugs in the drawer's fetch/loading logic need fixing before merge, plus a couple of minor hardening items and a documented follow-up on an authorization pattern this PR inherits rather than introduces.

### Issue count
- 🔴 Blocking: 2 — drawer doesn't refetch on `eventId` change while open; loading state never clears after an error/no-data response
- 🟡 Minor: 3 — unvalidated `eventUrl` in the API response, inconsistent optional chaining on Snowflake rows, no Playwright coverage for the new drawer
- ❔ Question: 2 — event-detail endpoint's authorization scoping (pre-existing pattern, not a regression), and access-control intent for competitive/financial fields

### AI bot reconciliation
- **Agree** with Cursor's `initDetail` finding and Copilot's `null`-conflation finding — verified both against the actual `initDetail()` implementation; see inline comments below, they're the same two root causes stated more precisely.
- **Agree** with Copilot's eventUrl/PCC-CTA mislabeling finding and the accessible-row finding — no additional detail to add.
- **Agree** with Copilot's missing-E2E-coverage finding — folded into the Test Coverage comment below.

### Final decision

PR #1185, opened 2026-07-23
test(profile): add linuxAliasValidator unit tests

fayazg:
## Summary
- Adds `packages/shared/src/validators/linux-alias.validator.spec.ts` — 49 vitest cases covering all four branches of `linuxAliasValidator`: required (empty/whitespace), max length (boundary + over-limit reporting), banned characters, and reserved names (case-insensitive), plus valid-alias acceptance.
- Cases are driven by the shared constants (`LINUX_ALIAS_MAX_LENGTH`, `LINUX_ALIAS_BANNED_CHARS`, `LINUX_ALIAS_RESERVED_NAMES`) so they track the source of truth.
- Closes the deferred unit-test gap from PR #857.

Fixes LFXV2-1663

🤖 Generated with Claude Code

### DIFF
--- a/packages/shared/src/validators/linux-alias.validator.spec.ts
+++ b/packages/shared/src/validators/linux-alias.validator.spec.ts
@@ -1,7 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { AbstractControl } from '@angular/forms';
+import type { AbstractControl, ValidationErrors } from '@angular/forms';
 import { describe, expect, it } from 'vitest';
 
 import { LINUX_ALIAS_BANNED_CHARS, LINUX_ALIAS_MAX_LENGTH, LINUX_ALIAS_RESERVED_NAMES } from '../constants/linux-email.constants';
@@ -9,7 +9,7 @@ import { linuxAliasValidator } from './linux-alias.validator';
 
 // The validator only reads `control.value`, so a minimal stub is sufficient.
 const control = (value: unknown): AbstractControl => ({ value }) as AbstractControl;
-const validate = (value: unknown) => linuxAliasValidator()(control(value));
+const validate = (value: unknown): ValidationErrors | null => linuxAliasValidator()(control(value));
 
 describe('linuxAliasValidator', () => {
   it('defers empty/nullish input to the required validator (returns null)', () => {
--- a/packages/shared/src/validators/linux-alias.validator.spec.ts
+++ b/packages/shared/src/validators/linux-alias.validator.spec.ts
@@ -1,0 +1,73 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { AbstractControl } from '@angular/forms';
+import { describe, expect, it } from 'vitest';
+
+import { LINUX_ALIAS_BANNED_CHARS, LINUX_ALIAS_MAX_LENGTH, LINUX_ALIAS_RESERVED_NAMES } from '../constants/linux-email.constants';
+import { linuxAliasValidator } from './linux-alias.validator';
+
+// The validator only reads `control.value`, so a minimal stub is sufficient.
+const control = (value: unknown): AbstractControl => ({ value }) as AbstractControl;
+const validate = (value: unknown) => linuxAliasValidator()(control(value));
+
+describe('linuxAliasValidator', () => {
+  it('defers empty/nullish input to the required validator (returns null)', () => {
+    expect(validate('')).toBeNull();
+    expect(validate(null)).toBeNull();
+    expect(validate(undefined)).toBeNull();
+  });
+
+  it('flags required when the value is only whitespace', () => {
+    expect(validate('   ')).toEqual({ required: true });
+  });
+
+  it('flags aliasInvalidChars for non-string input', () => {
+    expect(validate(123)).toEqual({ aliasInvalidChars: true });
+  });
+
+  describe('length', () => {
+    it('accepts a value at the max length', () => {
+      expect(validate('a'.repeat(LINUX_ALIAS_MAX_LENGTH))).toBeNull();
+    });
+
+    it('flags aliasMaxLength when over the limit, reporting lengths', () => {
+      const actualLength = LINUX_ALIAS_MAX_LENGTH + 1;
+      expect(validate('a'.repeat(actualLength))).toEqual({
+        aliasMaxLength: { requiredLength: LINUX_ALIAS_MAX_LENGTH, actualLength },
+      });
+    });
+
+    it('measures length after trimming', () => {
+      const padded = `  ${'a'.repeat(LINUX_ALIAS_MAX_LENGTH)}  `;
+      expect(validate(padded)).toBeNull();
+    });
+  });
+
+  describe('banned characters', () => {
+    it.each([...LINUX_ALIAS_BANNED_CHARS])('flags aliasInvalidChars for %p', (char) => {
+      expect(validate(`alice${char}bob`)).toEqual({ aliasInvalidChars: true });
+    });
+  });
+
+  describe('reserved names', () => {
+    it.each([...LINUX_ALIAS_RESERVED_NAMES])('flags aliasReserved for %p', (name) => {
+      expect(validate(name)).toEqual({ aliasReserved: true });
+    });
+
+    it('matches reserved names case-insensitively', () => {
+      expect(validate('Postmaster')).toEqual({ aliasReserved: true });
+      expect(validate('  ADMIN  ')).toEqual({ aliasReserved: true });
+    });
+  });
+
+  describe('valid aliases', () => {
+    it.each([['alice'], ['bob.smith'], ['jane-doe'], ['user123'], ['a']])('accepts %p', (value) => {
+      expect(validate(value)).toBeNull();
+    });
+
+    it('normalizes case/whitespace before accepting', () => {
+      expect(validate('  Alice.Smith  ')).toBeNull();
+    });
+  });
+});


packages/shared/src/validators/linux-alias.validator.spec.ts
```diff
@@ -0,0 +1,73 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { AbstractControl } from '@angular/forms';
```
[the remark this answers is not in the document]
fayazg: Done — added an explicit `ValidationErrors | null` return type to the `validate` helper and switched the `@angular/forms` import to type-only (`import type`), since `AbstractControl` is used purely as a type here.

See commit bac90889e.

fayazg:
## Review Feedback Addressed

Commit: bac90889e

### Changes Made
- **packages/shared/src/validators/linux-alias.validator.spec.ts**: added an explicit `ValidationErrors | null` return type to the `validate` helper and switched the `@angular/forms` import to type-only (per coderabbitai[bot]).

### Threads Resolved

MRashad26 (approved):
Code standards review: all clean. Pure test file — `import type` used for Angular form types, constants sourced from `@lfx-one/shared/constants`, no component code. Approving.

luismoriguerra (commented):
## Audit: PR #1185 — test(profile): add linuxAliasValidator unit tests

Independent post-merge audit (`pr-reviewer-agent`). Gates re-run in an isolated worktree at `28c9632`. Report-only.

**Scope:** adds a vitest spec for the shared `linuxAliasValidator`. Test-only; no production code path changes.
**Lanes run:** build · lint:check · check-types · test (all re-run) · secrets/critical-constants · code-standards. Skipped the architecture/ADR, privacy/PII, and platform-contract lanes — a test-only diff adds no response field, read/write path, or NATS/OpenSearch/FGA surface.

### 1. Why this change

PR #857 shipped `linuxAliasValidator` without unit tests and deferred them; this closes that gap. The validator mirrors client-side the alias rules that v2 auth-service and forwards-service enforce, so bad input is rejected before the NATS round-trip — with `check_alias` remaining the server-side source of truth (`packages/shared/src/validators/linux-alias.validator.ts:10-17`).

### 2. How it was done

A minimal `AbstractControl` stub exposing only `value` (all the validator reads) plus a `validate()` wrapper, then nested describes per rule. The banned-character and reserved-name cases are driven by `it.each` over the shared constants rather than hardcoded lists, so the suite grows automatically when someone adds a banned character. That's the right call and it matches the sibling specs in `packages/shared/src/utils/`.

### 3. Is it correct

Yes. Re-running the gates in an isolated worktree: `yarn test` goes from 294 to 343 shared tests, so the PR body's "49 vitest cases" is exactly accurate, and all 15 shared test files pass. `lint:check` and `check-types` are clean.

Coverage hits all four error returns in `linux-alias.validator.ts:19-35` plus the null-deferral path. Prior review is resolved: CodeRabbit's missing-return-type and value-import points were both fixed in `bac90889e`. That fix matters more than it looks — `@angular/forms` is only a **peer** dependency of `packages/shared`, so the type-only import is what keeps vitest from needing it at runtime.

### 4. Missing / gaps

All non-blocking. These are optional follow-up hardening on a merged test-only PR, not defects.

1. **Non-blocking — no guard against reserved names matching as substrings.** `linux-alias.validator.ts:32` uses exact array membership. Every reserved-name case feeds the bare name, so a future refactor to `RESERVED.some((n) => normalized.includes(n))` — the most plausible way this validator regresses — passes all 49 tests while `admin-tools` and `linuxfan` start getting rejected. See inline.
2. **Non-blocking — error precedence is untested.** The validator returns the first match in a fixed order (required → maxLength → invalidChars → reserved), but nothing pins that order. See inline.
3. **Nit — `actualLength` isn't shown to be the *trimmed* length.** The over-length case uses an unpadded string, so raw and trimmed lengths are identical and the assertion can't distinguish `value.length` from `normalized.length`. See inline.
4. **Nit — all 41 `it.each` cases render their name as a literal `%p`.** Vitest doesn't substitute Jest's `%p`, so a failure gives you 17 identical lines reading `flags aliasInvalidChars for %p` with no indication of which character failed. Use `%s`. Inherited house style (`email.utils.spec.ts:17` has it too), so optional.
5. **Nit — the spec pins logic but not the constants.** Because every case is driven off `LINUX_ALIAS_MAX_LENGTH` and friends, the suite would pass unchanged if someone raised the max length to 200 or deleted a reserved name. One literal assertion (e.g. `expect(LINUX_ALIAS_MAX_LENGTH).toBe(64)`) would make constant drift fail a test.
6. **Nit — the validator is more permissive than the UI claims.** `profile-linux-email.component.html:102` tells users "Use letters, numbers, dots, and hyphens only," but the validator only bans a fixed character list, so `alice+tag`, `..alice..` and `ünïcodé` all pass. The server's `check_alias` is the real gate, so this is cosmetic — but the spec is the natural place to pin the permissiveness so nobody "fixes" the copy by tightening a regex.

### 5. Q&A this PR resolves

- **Q:** Does `packages/shared` actually run this spec in CI? — **A:** Yes. Root `package.json:13` `"test": "turbo run test"` → `packages/shared/package.json` `"test": "vitest run"` with no vitest config, so the default `**/*.spec.ts` include picks it up, and `quality-check.yml:49` runs `yarn test` on every PR to `main`. Verified by running it.
- **Q:** Why does the spec import `@angular/forms` as `import type` when the validator imports it normally? — **A:** `@angular/forms` is a peer dependency of `packages/shared`, not a devDependency. The type-only import lets esbuild elide it entirely so vitest never tries to resolve it.

### 6. Business rules & ADRs

- **Business rule encoded:** an alias local part is ≤64 chars after trimming, free of 17 banned characters, and not one of 19 reserved names — all matched case-insensitively, with the server retaining final authority.
- **ADR alignment:** no cross-cutting platform decision is touched. The validator is client-side pre-validation over the identity plane described in *auth-service is the identity abstraction plane*, which it complies with by deferring to `check_alias`.
- **Privacy/access note:** lane skipped — no response field, read path, or write path changes.

### Cross-cutting note (pre-existing, not introduced here)

`packages/shared/tsconfig.json` excludes `**/*.spec.ts`, and `check-types` is a bare `tsc --noEmit` against that config — so **spec files in this package are never type-checked**. Correct for `build` (it keeps specs out of `dist/`), but it silently drops them from type-checking too, meaning this spec's `AbstractControl` stub cast is unchecked. Runtime `toEqual` still catches shape drift, so the blast radius is small. A `tsconfig.spec.json` extending the base without the exclude, run alongside `check-types`, would close it. Worth a separate ticket rather than anything against this PR.

---

**Verdict: ✅ Approved** (merge readiness: safe-to-merge — already merged; the gaps above are optional follow-up hardening, not defects)

packages/shared/src/validators/linux-alias.validator.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    it('matches reserved names case-insensitively', () => {
+      expect(validate('Postmaster')).toEqual({ aliasReserved: true });
+      expect(validate('  ADMIN  ')).toEqual({ aliasReserved: true });
+    });
+  });
+
+  describe('valid aliases', () => {
+    it.each([['alice'], ['bob.smith'], ['jane-doe'], ['user123'], ['a']])('accepts %p', (value) => {
```
luismoriguerra, line 65: **Non-blocking — add a false-positive guard for reserved names as substrings.**

`linux-alias.validator.ts:32` uses exact array membership:

```ts
if (LINUX_ALIAS_RESERVED_NAMES.includes(normalized as (typeof LINUX_ALIAS_RESERVED_NAMES)[number])) return { aliasReserved: true };
```

Every case in the `reserved names` block above feeds the bare name, so the suite proves reserved names are *rejected* but never that non-reserved aliases *containing* one are accepted. If someone later "tightens" this to `RESERVED.some((n) => normalized.includes(n))` — the most plausible way this validator regresses — all 49 tests still pass while legitimate aliases start getting rejected. Confirmed today: `validate('admin-tools')`, `validate('linuxfan')` and `validate('my-support-team')` all return `null`.

Three lines in this block would pin it:

```ts
it.each([['admin-tools'], ['linuxfan'], ['my-support-team']])('accepts %s containing a reserved name as a substring', (value) => {
  expect(validate(value)).toBeNull();
});
```

Separate nit on this line: vitest doesn't substitute Jest's `%p`, so all five of these cases (and the 17 banned-char + 19 reserved-name cases above) render as a literal `%p` in the reporter — a failure tells you nothing about which input failed. `%s` fixes it. Inherited house style (`packages/shared/src/utils/email.utils.spec.ts:17` has the same), so optional.

packages/shared/src/validators/linux-alias.validator.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  it('flags aliasInvalidChars for non-string input', () => {
+    expect(validate(123)).toEqual({ aliasInvalidChars: true });
+  });
+
+  describe('length', () => {
+    it('accepts a value at the max length', () => {
+      expect(validate('a'.repeat(LINUX_ALIAS_MAX_LENGTH))).toBeNull();
+    });
+
+    it('flags aliasMaxLength when over the limit, reporting lengths', () => {
```
luismoriguerra, line 34: **Non-blocking — error precedence isn't pinned, and `actualLength` isn't shown to be the trimmed length.**

Two small things on the length block.

**Precedence.** The validator returns the *first* match in a fixed order (required → maxLength → invalidChars → reserved, `linux-alias.validator.ts:27-32`), but no test asserts that order, so the checks could be reordered without failing anything. Verified today: `'a'.repeat(65) + '@'` returns `{ aliasMaxLength: { requiredLength: 64, actualLength: 66 } }` rather than `aliasInvalidChars`, and `'admin@'` returns `{ aliasInvalidChars: true }` rather than `aliasReserved`. A short `describe('precedence', …)` with those two assertions would lock it.

**Trimmed length.** This case uses an unpadded `'a'.repeat(65)`, where raw and trimmed lengths are identical — so the assertion can't distinguish `value.length` from `normalized.length`. The `measures length after trimming` test below only covers the passing side. Padding this input closes the gap:

```ts
expect(validate(`  ${'a'.repeat(65)}  `)).toEqual({
  aliasMaxLength: { requiredLength: LINUX_ALIAS_MAX_LENGTH, actualLength: 65 },
});
```

(Confirmed it does report `65`, not `69`.)

Also applies to the non-string guard a few lines up: `123` is truthy, so it doesn't exercise the interesting boundary — falsy non-strings that skip the `value === ''` early return at `validator.ts:21` and land on the `typeof` guard. `0`, `false`, `NaN`, `[]` and `{}` all correctly return `{ aliasInvalidChars: true }` today; worth pinning.

PR #1190, opened 2026-07-24
feat(dashboards): add monthly-distinct active contributors chart

audigregorie:
# Summary

Adds a **monthly-distinct** active-contributors view to the foundation dashboard. Previously the "Active Contributors" surfaces showed a *daily-average* number, which understates real reach because a contributor active on many days is averaged down rather than counted once. This PR introduces a distinct-per-month metric (`COUNT(DISTINCT member_id)` per month) alongside the existing average, so EDs and maintainers can see how many unique people actually contributed each month and the month-over-month trend.

# Before / After

**Before:** The active-contributors drawer opened with a single line chart of *monthly-average* daily unique contributors, and the foundation-health card showed the yearly average with no trend indicator. There was no view of how many distinct people contributed in a given month.

**After:** The drawer now leads with a **Monthly Distinct** section — a headline count of the latest month plus a month-over-month delta (▲/▼ % vs last month) and a 12-month line chart — above the existing average view (now relabeled **Monthly Average** and rendered as bars to visually distinguish the two metrics). The foundation-health card now shows the latest monthly-distinct count with a trend arrow and "% vs last month" instead of the flat yearly average.

# Changes

## Shared (`@lfx-one/shared`)
- Add `FoundationActiveContributorsMonthlyDistinctRow`, `FoundationActiveContributorsMonthlyDistinctResponse`, and `ActiveContributorsMoMDelta` interfaces.
- Add `DEFAULT_FOUNDATION_ACTIVE_CONTRIBUTORS_MONTHLY_DISTINCT` constant.

## Backend (Express)
- New endpoint `GET /api/analytics/foundation-active-contributors-monthly-distinct` (route + controller with `foundationSlug` validation via `SLUG_PATTERN`).
- New `ProjectService.getFoundationActiveContributorsMonthlyDistinct()` — parameterized Snowflake query against `ANALYTICS.PLATINUM_LFX_ONE.FOUNDATION_ACTIVE_CONTRIBUTORS_MONTHLY` (last 12 months, `COUNT(DISTINCT member_id)` per month), returning `{ monthlyData, monthlyLabels }`.

## Frontend
- `AnalyticsService.getFoundationActiveContributorsMonthlyDistinct()` HTTP gateway with graceful empty-array fallback.
- **Active contributors drawer**: new Monthly Distinct section (headline count, MoM delta signal/computed, line chart, loading/empty states, tooltip); relabels the existing section to "Monthly Average" and switches it from a filled line to a bar chart; clarified tooltip copy across sections. Data now loaded via a three-way `forkJoin` (monthly, distribution, mom).
- **Foundation-health card**: active-contributors card now sourced from the monthly-distinct data with a `trend` + `changePercentage` ("% vs last month") derived from the latest two months.

## Tests
- No automated E2E tests added — acceptance criteria (AC-1–AC-4) for this feature are verified manually per the LFXV2-2687 spec. `data-testid` hooks were added to the new drawer markup to support future E2E targeting.

# Deploy ordering (cross-repo dependency)

This PR consumes a new dbt model, `platinum_lfx_one_foundation_active_contributors_monthly.sql`, materialized to `ANALYTICS.PLATINUM_LFX_ONE.FOUNDATION_ACTIVE_CONTRIBUTORS_MONTHLY`. That model must be built and populated in Snowflake **before** this app deploys. If the app ships first, the endpoint returns empty arrays and the UI shows a graceful empty state (no error), but not the intended data. Sequence: merge/deploy dbt → verify table populated → merge this PR.

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/components/active-contributors-drawer/active-contributors-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/components/active-contributors-drawer/active-contributors-drawer.component.html
@@ -63,7 +63,7 @@
           <div class="flex items-baseline gap-2">
             @if (momDelta(); as delta) {
               <span class="text-xl font-semibold text-gray-900" data-testid="active-contributors-drawer-mom-metric">
-                {{ delta.latest.toLocaleString() }}
+                {{ delta.latest.toLocaleString('en-US') }}
               </span>
               @if (delta.deltaPercent !== null) {
                 <span
@@ -73,9 +73,13 @@
                   [class.text-gray-500]="delta.direction === 'flat'"
                   data-testid="active-contributors-drawer-mom-delta">
                   @if (delta.direction === 'up') {
-                    ▲
+                    <span aria-hidden="true">▲</span>
+                    <span class="sr-only">Increased </span>
                   } @else if (delta.direction === 'down') {
-                    ▼
+                    <span aria-hidden="true">▼</span>
+                    <span class="sr-only">Decreased </span>
+                  } @else {
+                    <span class="sr-only">No change </span>
                   }
                   {{ formatMomDelta(delta.deltaPercent) }}% vs last month
                 </span>
@@ -149,10 +153,10 @@
         <button
           type="button"
           class="text-gray-400 hover:text-gray-600 cursor-help text-xs flex-shrink-0"
-          pTooltip="Share of total contributions made by each contributor tier over the last 12 months. The Top 10% drives most of the activity; the Bottom 50% contributes very little."
+          pTooltip="Share of total contributions made by each contributor tier over the last 12 months."
           tooltipPosition="top"
           tooltipEvent="both"
-          aria-label="Share of total contributions made by each contributor tier over the last 12 months. The Top 10% drives most of the activity; the Bottom 50% contributes very little.">
+          aria-label="Share of total contributions made by each contributor tier over the last 12 months.">
           <i class="fa-light fa-circle-info"></i>
         </button>
       </div>
--- a/apps/lfx-one/src/app/modules/dashboards/components/active-contributors-drawer/active-contributors-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/components/active-contributors-drawer/active-contributors-drawer.component.html
@@ -43,20 +43,78 @@
 
   <!-- Content -->
   <div class="flex flex-col gap-6 pb-2" data-testid="active-contributors-drawer-content">
+    <!-- Monthly Distinct (MoM) Section -->
+    <div class="flex flex-col gap-3" data-testid="active-contributors-drawer-mom-section">
+      <div class="flex items-center justify-between">
+        <div class="flex items-center gap-2">
+          <h3 class="text-sm font-medium text-gray-900">Active Contributors (Monthly Distinct)</h3>
+          <button
+            type="button"
+            class="text-gray-400 hover:text-gray-600 cursor-help text-xs flex-shrink-0"
+            pTooltip="Unique people who contributed at any point in each month. Each person is counted once per month, so someone active on many days still counts as one."
+            tooltipPosition="top"
+            tooltipEvent="both"
+            aria-label="Unique people who contributed at any point in each month. Each person is counted once per month, so someone active on many days still counts as one.">
+            <i class="fa-light fa-circle-info"></i>
+          </button>
+        </div>
+
+        @if (hasMomData()) {
+          <div class="flex items-baseline gap-2">
+            @if (momDelta(); as delta) {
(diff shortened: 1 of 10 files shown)


apps/lfx-one/src/app/modules/dashboards/components/active-contributors-drawer/active-contributors-drawer.component.html
```diff
@@ -43,20 +43,82 @@ <h2 class="text-lg font-semibold text-gray-900" data-testid="active-contributors
...
+                  @if (delta.direction === 'up') {
+                    <span aria-hidden="true">▲</span>
+                    <span class="sr-only">Increased </span>
+                  } @else if (delta.direction === 'down') {
+                    <span aria-hidden="true">▼</span>
+                    <span class="sr-only">Decreased </span>
+                  } @else {
+                    <span class="sr-only">No change </span>
+                  }
+                  {{ formatMomDelta(delta.deltaPercent) }}% vs last month
```
ahmedomosanya: The Monthly Distinct headline reads a method call and `.toLocaleString()` straight from the template:

```
{{ delta.latest.toLocaleString('en-US') }}
{{ formatMomDelta(delta.deltaPercent) }}% vs last month
```

Both re-run on every change-detection pass, which works against OnPush. Since `momDelta()` is already a `computed()`, could we fold the formatted strings into it and read plain props in the template?

```ts
protected readonly momDelta = computed(() => {
  const mom = this.momData();
  return {
    latest: mom.latest,
    latestLabel: mom.latest.toLocaleString('en-US'),
    deltaPercent: mom.momDeltaPercent,
    deltaLabel: mom.momDeltaPercent === null ? '' : Math.abs(mom.momDeltaPercent).toFixed(1),
    direction: mom.momDirection,
  };
});
```

```html
{{ delta.latestLabel }}
{{ delta.deltaLabel }}% vs last month
```

That also lets us drop `formatMomDelta` entirely.

packages/shared/src/interfaces/analytics-data.interface.ts
```diff
@@ -698,6 +698,40 @@ export interface FoundationActiveContributorsMonthlyResponse {
...
+ * when a month is absent from the materialized table).
+ */
+export interface FoundationActiveContributorsMonthlyDistinctResponse {
+  monthlyData: number[];
+  monthlyLabels: string[];
+  /** Most recent month's distinct contributor count (0 when no rows). */
+  latest: number;
+  /** MoM delta percent; null when no valid adjacent prior month exists. */
+  momDeltaPercent: number | null;
+  momDirection: 'up' | 'down' | 'flat';
```
ahmedomosanya: Nit: the `'up' | 'down' | 'flat'` union is a domain concept repeated in four places — here, `ActiveContributorsMoMDelta.direction`, the local in `getFoundationActiveContributorsMonthlyDistinct`, and the `trend` mapping in `foundation-health`. Could we name it once and reuse?

```ts
export type MoMDirection = 'up' | 'down' | 'flat';
```

Keeps the four sites in sync if a state is ever added.

MRashad26 (commented):
## Code Standards Review

**One HIGH violation found — method call in template.**

---

### [HIGH] `formatMomDelta()` called directly in template

**File:** `apps/lfx-one/src/app/modules/dashboards/components/active-contributors-drawer/active-contributors-drawer.component.html`

**Problem:** `{{ formatMomDelta(delta.deltaPercent) }}% vs last month` calls a component method inside an interpolation. Every change detection cycle re-invokes this function. Component method calls in templates are prohibited — use `computed()` signals or a `@Pipe` instead.

**Suggested fix:** Move the formatting into `initMomDelta()` and add a pre-formatted field to `ActiveContributorsMoMDelta`:

```ts
// analytics-data.interface.ts — add to ActiveContributorsMoMDelta
formattedDeltaPercent: string | null; // null when deltaPercent is null

// active-contributors-drawer.component.ts — initMomDelta()
private initMomDelta(): Signal<ActiveContributorsMoMDelta> {
  return computed(() => {
    const mom = this.momData();
    const deltaPercent = mom.momDeltaPercent;
    return {
      latest: mom.latest,
      deltaPercent,
      direction: mom.momDirection,
      formattedDeltaPercent: deltaPercent !== null ? Math.abs(deltaPercent).toFixed(1) : null,
    };
  });
}
```

Then the template becomes a plain property read:

```html
@if (delta.formattedDeltaPercent) {
  {{ delta.formattedDeltaPercent }}% vs last month
}
```

---

Everything else is clean: signals/computed used correctly throughout, all interfaces and constants in `@lfx-one/shared`, `import type` for type-only imports, BFF endpoint validates `foundationSlug` with `SLUG_PATTERN` and uses a parameterized Snowflake query (no injection risk), graceful pre-dbt-deploy empty fallback, and no local types.

ahmedomosanya (approved):

MRashad26 (approved):

PR #1191, opened 2026-07-24
perf(profile): collapse duplicate user_emails.read on linux-email tab

fayazg:
## Summary

- The Linux.com email tab read `user_emails.read` twice per load — once server-side inside `getLinuxAlias`, and again from the client via `getUserEmails` in a `forkJoin`. `getLinuxAlias` already has the emails, so it now returns the primary email inline as `LinuxAliasData.primaryEmail` and the client drops the second round-trip.
- Because the emails read now shares the alias call's failure path, the standalone per-form "emails failed" retry state folds into the whole-tab `service_unavailable` retry. Template branches and E2E specs are updated to match.
- Removes the now-unused `LinuxEmailData` interface.

Part of LFXV2-1663 (PR #857 follow-up).

### DIFF
--- a/apps/lfx-one/e2e/profile-linux-email.spec.ts
+++ b/apps/lfx-one/e2e/profile-linux-email.spec.ts
@@ -2,17 +2,24 @@
 // SPDX-License-Identifier: MIT
 
 /**
- * Linux.com email tab — partial-claim-failure recovery.
+ * Linux.com email tab.
  *
- * Covers the fix for the non-atomic claim flow: `add_alias` (auth-service) succeeds
- * but `set_target` (forwards-service) fails, so the server returns
- * `502 FORWARD_SET_FAILED`. The alias is immutable once claimed, so the component
- * must recover into the claimed/edit view (where the user can set forwarding)
- * instead of leaving them stuck on the claim form, where a retry would fail with
- * `already_claimed`.
+ * Covers:
+ * - Partial-claim-failure recovery: `add_alias` (auth-service) succeeds but
+ *   `set_target` (forwards-service) fails, so the server returns
+ *   `502 FORWARD_SET_FAILED`. The alias is immutable once claimed, so the
+ *   component must recover into the claimed/edit view instead of leaving the
+ *   user stuck on the claim form (where a retry would fail with `already_claimed`).
+ * - Forwarding-target visibility across the claimed states.
+ * - The whole-tab retry panel when the alias service is unavailable.
+ *
+ * The tab reads its state from a single `GET /api/profile/linux-email` call —
+ * the server resolves `user_emails.read` and returns the primary email inline as
+ * `primaryEmail`, so these stubs set it on the alias body rather than mocking a
+ * separate `/api/profile/emails` fetch.
  */
 
-import type { EmailManagementData, LinuxAliasData } from '@lfx-one/shared/interfaces';
+import type { LinuxAliasData } from '@lfx-one/shared/interfaces';
 import { expect, Page, test } from '@playwright/test';
 
 const DOMAIN = 'example.org';
@@ -30,14 +37,8 @@ function skipWhenAuthMissing(page: Page): void {
   }
 }
 
-/** Stub the sibling fetches the tab needs to render deterministically. */
-async function stubProfileContext(page: Page): Promise<void> {
-  await page.route('**/api/profile/emails', (route) => {
-    if (route.request().method() !== 'GET') return route.fallback();
-    const body: EmailManagementData = { primary_email: PRIMARY_EMAIL, alternate_emails: [] };
-    return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
-  });
-
+/** Stub the identities fetch the tab needs to render deterministically. */
+async function stubIdentities(page: Page): Promise<void> {
   await page.route('**/api/profile/identities', (route) => {
     if (route.request().method() !== 'GET') return route.fallback();
     return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify([]) });
@@ -55,8 +56,8 @@ async function stubPartialClaimFailure(page: Page): Promise<void> {
   await page.route('**/api/profile/linux-email', (route) => {
     if (route.request().method() !== 'GET') return route.fallback();
     const body: LinuxAliasData = claimed
-      ? { state: 'claimed', domain: DOMAIN, alias: ALIAS, email: `${ALIAS}@${DOMAIN}`, forwardTo: null }
-      : { state: 'purchased_unclaimed', domain: DOMAIN, alias: null, email: null, forwardTo: null };
+      ? { state: 'claimed', domain: DOMAIN, alias: ALIAS, email: `${ALIAS}@${DOMAIN}`, forwardTo: null, primaryEmail: PRIMARY_EMAIL }
+      : { state: 'purchased_unclaimed', domain: DOMAIN, alias: null, email: null, forwardTo: null, primaryEmail: PRIMARY_EMAIL };
     return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
   });
 
@@ -80,7 +81,7 @@ async function stubPartialClaimFailure(page: Page): Promise<void> {
 
 test.describe('Linux.com email — partial claim failure recovery', () => {
   test('recovers into the claimed/edit view after a FORWARD_SET_FAILED response', async ({ page }) => {
-    await stubProfileContext(page);
+    await stubIdentities(page);
     await stubPartialClaimFailure(page);
 
     await page.goto('/profile/identities', { waitUntil: 'domcontentloaded' });
@@ -112,10 +113,17 @@ test.describe('Linux.com email — partial claim failure recovery', () => {
 
 test.describe('Linux.com email — forwarding target visibility', () => {
   test('keeps the forward dropdown visible with a hint when the saved target is the only verified option', async ({ page }) => {
-    await stubProfileContext(page);
+    await stubIdentities(page);
     await page.route('**/api/profile/linux-email', (route) => {
       if (route.request().method() !== 'GET') return route.fallback();
-      const body: LinuxAliasData = { state: 'claimed', domain: DOMAIN, alias: ALIAS, email: `${ALIAS}@${DOMAIN}`, forwardTo: PRIMARY_EMAIL };
+      const body: LinuxAliasData = {
+        state: 'claimed',
+        domain: DOMAIN,
+        alias: ALIAS,
+        email: `${ALIAS}@${DOMAIN}`,
+        forwardTo: PRIMARY_EMAIL,
+        primaryEmail: PRIMARY_EMAIL,
+      };
       return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
     });
 
@@ -138,18 +146,10 @@ test.describe('Linux.com email — forwarding target visibility', () => {
     // forwardTo-preservation branch, with the primary excluded because it equals the alias.
     const aliasEmail = `${ALIAS}@${DOMAIN}`;
     const externalForward = 'someone@external.com';
-    await page.route('**/api/profile/emails', (route) => {
-      if (route.request().method() !== 'GET') return route.fallback();
-      const body: EmailManagementData = { primary_email: aliasEmail, alternate_emails: [] };
-      return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
-    });
-    await page.route('**/api/profile/identities', (route) => {
-      if (route.request().method() !== 'GET') return route.fallback();
-      return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify([]) });
-    });
+    await stubIdentities(page);
     await page.route('**/api/profile/linux-email', (route) => {
       if (route.request().method() !== 'GET') return route.fallback();
-      const body: LinuxAliasData = { state: 'claimed', domain: DOMAIN, alias: ALIAS, email: aliasEmail, forwardTo: externalForward };
+      const body: LinuxAliasData = { state: 'claimed', domain: DOMAIN, alias: ALIAS, email: aliasEmail, forwardTo: externalForward, primaryEmail: aliasEmail };
       return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
     });
 
@@ -169,18 +169,10 @@ test.describe('Linux.com email — forwarding target visibility', () => {
     // Genuine-empty case: the only verified email is the claimed alias itself (so it's
     // excluded from forwardOptions) and no external forwardTo is saved — zero options.
     const aliasEmail = `${ALIAS}@${DOMAIN}`;
-    await page.route('**/api/profile/emails', (route) => {
-      if (route.request().method() !== 'GET') return route.fallback();
-      const body: EmailManagementData = { primary_email: aliasEmail, alternate_emails: [] };
-      return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
-    });
-    await page.route('**/api/profile/identities', (route) => {
-      if (route.request().method() !== 'GET') return route.fallback();
-      return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify([]) });
-    });
+    await stubIdentities(page);
     await page.route('**/api/profile/linux-email', (route) => {
       if (route.request().method() !== 'GET') return route.fallback();
-      const body: LinuxAliasData = { state: 'claimed', domain: DOMAIN, alias: ALIAS, email: aliasEmail, forwardTo: null };
+      const body: LinuxAliasData = { state: 'claimed', domain: DOMAIN, alias: ALIAS, email: aliasEmail, forwardTo: null, primaryEmail: aliasEmail };
       return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
     });
 
@@ -194,10 +186,10 @@ test.describe('Linux.com email — forwarding target visibility', () => {
   });
 
   test('shows the normal hint on a first-time claim with a single verified email', async ({ page }) => {
-    await stubProfileContext(page);
+    await stubIdentities(page);
     await page.route('**/api/profile/linux-email', (route) => {
       if (route.request().method() !== 'GET') return route.fallback();
-      const body: LinuxAliasData = { state: 'purchased_unclaimed', domain: DOMAIN, alias: null, email: null, forwardTo: null };
+      const body: LinuxAliasData = { state: 'purchased_unclaimed', domain: DOMAIN, alias: null, email: null, forwardTo: null, primaryEmail: PRIMARY_EMAIL };
       return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
     });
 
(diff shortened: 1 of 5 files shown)


apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -632,7 +633,7 @@ export class ProfileController {
       const state = purchased ? 'purchased_unclaimed' : 'not_purchased';
 
       logger.success(req, 'get_linux_alias', startTime, { state });
-      res.json(this.linuxAliasState(state, domain));
+      res.json(this.linuxAliasState(state, domain, emails.primary_email ?? null));
```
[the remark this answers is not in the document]
fayazg: Fixed in 239bdaa40. getLinuxAlias now applies the same session-OIDC-email fallback that GET /api/profile/emails uses (`emails.primary_email || sessionEmail || null`), so accounts where auth-service omits a primary keep their primary forward option and claim-form default. Good catch.

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -609,6 +609,7 @@ export class ProfileController {
           alias,
           email,
           forwardTo: forward?.target_email ?? null,
+          primaryEmail: emails.primary_email ?? null,
```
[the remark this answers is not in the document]
fayazg: Fixed in 239bdaa40. Restored the `getEffectiveEmail(req)` session-email fallback (matching profile.controller.ts:428) when inlining primaryEmail, so a claimed user whose NATS primary is empty no longer loses their forwarding option. This covers both L612 and L636.

apps/lfx-one/e2e/profile-linux-email.spec.ts
```diff
@@ -30,14 +37,8 @@ function skipWhenAuthMissing(page: Page): void {
...
-/** Stub the sibling fetches the tab needs to render deterministically. */
-async function stubProfileContext(page: Page): Promise<void> {
-  await page.route('**/api/profile/emails', (route) => {
-    if (route.request().method() !== 'GET') return route.fallback();
-    const body: EmailManagementData = { primary_email: PRIMARY_EMAIL, alternate_emails: [] };
-    return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
-  });
-
+/** Stub the identities fetch the tab needs to render deterministically. */
+async function stubIdentities(page: Page): Promise<void> {
```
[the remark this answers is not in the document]
fayazg: Done in 239bdaa40. The first-time-claim test now counts GET requests to /api/profile/emails and asserts the count stays zero during tab load, so an accidental reintroduction of the duplicate fetch fails the suite.

fayazg:
## Review Feedback Addressed

Commit: 239bdaa40

### Changes Made
- **apps/lfx-one/src/server/controllers/profile.controller.ts**: Restored the session-OIDC-email fallback when inlining `primaryEmail`. `getLinuxAlias` now mirrors `GET /api/profile/emails` (`emails.primary_email || sessionEmail || null`), so accounts where auth-service omits a primary keep their forward option and claim-form default instead of seeing "no verified email". Covers both the claimed response and the unclaimed/not-purchased states. (per cursor[bot], copilot-pull-request-reviewer[bot])
- **apps/lfx-one/e2e/profile-linux-email.spec.ts**: The first-time-claim test now counts GET requests to `/api/profile/emails` and asserts zero during tab load, guarding the single-fetch performance goal against an accidental reintroduction of the duplicate client fetch. (per copilot-pull-request-reviewer[bot])

### Threads Resolved

MRashad26 (approved):
Code standards review: all clean. Single `getLinuxAlias()` call, no `forkJoin`, no `effect()`, no local interfaces (removed `LinuxEmailData` from shared correctly), all derived state via `computed()`, `catchError` fallback to `service_unavailable` shape, `primaryEmail` added to `LinuxAliasData` in shared interfaces. Approving.

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -567,6 +567,12 @@ export class ProfileController {
         return;
       }
 
+      // Match GET /api/profile/emails (see getEmails): fall back to the session OIDC
+      // email when auth-service omits a primary, so accounts in that gap keep their
+      // primary forward option and claim-form default instead of seeing "no verified email".
+      const sessionEmail = getEffectiveEmail(req) ?? undefined;
```
luismoriguerra, line 573: **🔐 Non-blocking — under impersonation this fallback can resolve to the *impersonator's* email, not the target's.**

The alias read itself is correctly scoped: `:552` uses `getEffectiveSub(req)`, so the `user_emails.read` at `:563` returns the target's emails. The skew is only in the fallback.

`getEffectiveEmail` returns the target's email **only when `impersonationUser.email` is truthy** (`apps/lfx-one/src/server/utils/auth-helper.ts:63`); otherwise it falls through to `req.oidc?.user?.['email']` at `:66` — the impersonator. And `impersonation.service.ts:142` stores:

```ts
email: targetClaims['http://lfx.dev/claims/email'] || '',
```

so a target token missing that claim yields an **empty string**, which is falsy, which triggers exactly that fall-through. `req.oidc.user` is never rewritten for API routes — the only override lives in the SSR catch-all at `server.ts:429-439`, registered *after* the API routers, and `auth.middleware.ts:153` only swaps `req.bearerToken`.

So when the target has no `primary_email` in auth-service **and** the impersonation token lacks the email claim, this endpoint returns the impersonator's address as the target's `primaryEmail`. It then renders as the target's "(Primary)" forward option and pre-fills the claim form.

**Why non-blocking rather than blocking:** the address is shown to the impersonator in their own browser, so no third-party data leaks; it can't be written through, since both write routes carry `blockDuringImpersonation` (`profile.route.ts:56,59`) and `getLinuxAlias` explicitly refuses the impersonator's management token at `:591`; and the identical fallback already existed at `:411,428`, which is where this tab previously sourced the same value. **This PR relocated the behavior rather than introducing it.**

Cleanest fix is one level up rather than here — have `getEffectiveEmail` return `null` under impersonation instead of silently falling back to the impersonator, then audit the two call sites:

```ts
export function getEffectiveEmail(req: Request): string | null {
  if (isImpersonating(req)) {
    return (req.appSession?.['impersonationUser']?.email as string)?.toLowerCase() || null;
  }
  return (req.oidc?.user?.['email'] as string)?.toLowerCase() || null;
}
```

Local-only alternative if you'd rather not touch the shared helper: `const sessionEmail = isImpersonating(req) ? undefined : (getEffectiveEmail(req) ?? undefined);`

Separately worth noting: the E2E suite has **no impersonation assertions at all** (grep for `impersonat` across `apps/lfx-one/e2e` finds nothing), before or after this PR. A test asserting the claim button is disabled and no impersonator address appears in the forward options under an impersonated session would be a good companion to the fix.

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -719,7 +726,11 @@ export class ProfileController {
       }
 
       logger.success(req, 'claim_linux_alias', startTime, { domain });
-      res.status(200).json({ state: 'claimed', domain, alias, email, forwardTo: forward.target_email ?? forwardTo } satisfies LinuxAliasData);
+      // primaryEmail is null here: this handler doesn't read user_emails, and the client
+      // refetches via getLinuxAlias() after a successful claim rather than consuming this body.
+      res
+        .status(200)
+        .json({ state: 'claimed', domain, alias, email, forwardTo: forward.target_email ?? forwardTo, primaryEmail: null } satisfies LinuxAliasData);
```
luismoriguerra, line 733: **Non-blocking — the required `primaryEmail` field forces this endpoint to invent a placeholder, giving one field three meanings.**

`packages/shared/src/interfaces/linux-email.interface.ts:34-42` makes `primaryEmail: string | null` **required** and documents `null` as "`service_unavailable` (emails could not be read), or the account has no primary email." This line then returns `null` meaning a third thing: *this handler never read it*.

Harmless today — I verified the comment above is accurate, `claim()` at `profile-linux-email.component.ts:148-152` discards the response body and calls `refresh.next()`. But `user.service.ts:151` types the call as `Observable<LinuxAliasData>`, so the client type advertises a field this endpoint structurally cannot populate, and the next person to read that body gets a silently wrong value rather than a compile error.

Making the field optional would be worse — that would let the GET path silently omit it, which is the case that actually matters. The right split is required-on-GET plus a narrower POST type, and there's precedent seven lines away in `user.service.ts:158`, where `updateLinuxForward` already returns `Observable<{ forwardTo: string }>` rather than reusing the full shape:

```ts
// linux-email.interface.ts
export type ClaimAliasResponse = Omit<LinuxAliasData, 'primaryEmail'>;
```

then use it here and at `user.service.ts:151`, and drop the placeholder.

apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
```diff
@@ -232,27 +226,29 @@ export class ProfileLinuxEmailComponent {
...
-            emails: this.userService.getUserEmails().pipe(catchError(() => of(null))),
-          }).pipe(
-            tap(({ alias, emails }) => {
-              this.applyFormDefaults(alias, emails);
+          // Single server call: getLinuxAlias reads user_emails.read server-side and
+          // returns the primary email inline, so the tab no longer makes a second
+          // getUserEmails() round-trip for the same data. If it fails, the emails read
+          // failed too, so service_unavailable (whole-tab retry) is the right state.
+          return this.userService.getLinuxAlias().pipe(
+            catchError(() =>
```
luismoriguerra, line 239: **Nit — silent `catchError` on a line this PR rewrites.**

`docs/reviews/frontend-checklist.md:362` rates a silent `catchError` **CRITICAL**: "Every `catchError` should log via `logger` or `console.error` at minimum." This swallows the error entirely, so a failing `getLinuxAlias()` renders the retry panel with nothing in the console to diagnose it.

Pre-existing, but the PR rewrites this exact expression, so it's the cheapest moment to fix. (`console.error` is fine here — `CLAUDE.md` bans only `console.log`.)

```ts
catchError((err) => {
  console.error('Failed to load Linux.com alias state', err);
  return of<LinuxAliasData | null>({ state: 'service_unavailable', domain: '', alias: null, email: null, forwardTo: null, primaryEmail: null });
}),
```

Related coverage gap worth pairing with this: **this branch is now tested by nothing.** The deleted *verified emails fetch failure* describe was the only place stubbing a real HTTP error; the new *service unavailable* test at `e2e/profile-linux-email.spec.ts:217-237` fulfills a **200 whose body carries** `state: 'service_unavailable'`, which exercises only the server-driven path. A sibling test fulfilling `**/api/profile/linux-email` with `status: 502` and asserting `linux-email-retry-button` is visible would cover this `catchError` end to end.

luismoriguerra (commented):
## Audit: PR #1191 — perf(profile): collapse duplicate `user_emails.read` on linux-email tab

Independent post-merge audit (`pr-reviewer-agent`). Gates re-run in an isolated worktree at `319c1e7`. Report-only.

**Scope:** removes a redundant client round-trip on the Linux.com email tab by returning `primaryEmail` inline on `GET /api/profile/linux-email`; deletes the now-unreachable granular "emails failed" retry UI and the `LinuxEmailData` wrapper interface.
**Lanes run:** build · lint:check · check-types · test (all re-run) · secrets/critical-constants · code-standards · architecture/ADR · privacy/access/PII. Skipped platform-contract-coherence — the diff touches no Goa design, `ruleset.yaml`, `httproute.yaml`, `model.fga`, or NATS subject; it changes only the BFF's own response shape.

### 1. Why this change

The tab issued `user_emails.read` twice per load. `getLinuxAlias` already reads it server-side at `profile.controller.ts:563` to detect whether the user has a claimed `@domain` address among their emails, and the client then independently called `getUserEmails()` in a `forkJoin` for the same data. The second call was pure waste against what the *auth-service is the identity abstraction plane* ADR describes as a hot dependency on the write path.

### 2. How it was done

The server derives `primaryEmail` from the emails read it already performed; the client drops the `forkJoin` for a single `getLinuxAlias()` call, and the signal narrows from `Signal<LinuxEmailData>` to `Signal<LinuxAliasData | null>` since the wrapper existed only to carry the second call's result.

The consequential part is the error-handling fold. Because the emails read now shares the alias call's failure path, the partial state "alias loaded but emails failed" becomes structurally **unrepresentable**, so both per-form retry panels collapse into the existing whole-tab `service_unavailable` retry. I verified this is sound rather than a downgrade: `profile.controller.ts:563-568` returns `service_unavailable` whenever the emails read comes back null, and both deleted panels invoked the same `retry()` as the whole-tab panel, so no user action is lost. The `|| emailsLoadFailed()` guard dropped from the claim button was guarding the same now-unreachable state. What *is* lost is diagnostic precision — the user sees "We couldn't reach the Linux.com email service" instead of "Couldn't load your verified emails." Worth a line in the PR description per `docs/reviews/frontend-checklist.md` §14.9.

### 3. Is it correct

Yes. All gates re-run green in an isolated worktree: `yarn build` succeeds, `lint:check` reports zero errors (one pre-existing warning in an unrelated `campaigns.component.ts`), `check-types` passes, and `yarn test` passes 294 shared + 52 app tests. Removals are complete — grepping for `emailsLoadFailed`, `LinuxEmailData`, and all four deleted test IDs returns nothing repo-wide.

I specifically verified the load-bearing comment at `profile.controller.ts:730-732`, which justifies the claim endpoint returning `primaryEmail: null` on the grounds that the client refetches: `claim()`'s success handler at `profile-linux-email.component.ts:148-152` does discard the response body and call `refresh.next()`. The comment is accurate.

Prior review was substantive and is resolved. Both Cursor and Copilot flagged that inlining `primaryEmail` from the raw NATS read dropped the session-OIDC-email fallback that `GET /api/profile/emails` applies at `profile.controller.ts:428`; Copilot separately asked for a regression guard proving the second fetch is gone. Both were fixed in `239bdaa40` — the fallback is restored at `:573-574` and the E2E test now counts `/api/profile/emails` GETs and asserts zero.

### 4. Missing / gaps

No blocking findings. In severity order:

1. **🔐 Non-blocking — under impersonation, `primaryEmail` can resolve to the impersonator's own email.** See inline on `profile.controller.ts:573`. Real but bounded, and **pre-existing in kind** — this PR relocated the fallback rather than introducing it.
2. **Non-blocking — the component's own `catchError` fallback now has no test.** `profile-linux-email.component.ts:239-241` synthesizes a `service_unavailable` object when the HTTP call itself fails. The deleted *verified emails fetch failure* describe was the only place stubbing a real HTTP error; its replacement at `e2e/profile-linux-email.spec.ts:217-237` fulfills a **200 whose body carries** `state: 'service_unavailable'`, exercising only the server-driven path. A sibling test fulfilling `/api/profile/linux-email` with a `502` would restore it.
3. **Non-blocking — the required `primaryEmail` field forces the claim endpoint to invent a placeholder.** See inline on `profile.controller.ts:733`.
4. **Non-blocking — the new `/api/profile/emails` stub isn't method-scoped.** See inline on `e2e/profile-linux-email.spec.ts:195`.
5. **Nit — the regression guard is page-global and lives inside an unrelated test.** See inline on `e2e/profile-linux-email.spec.ts:195`.
6. **Nit — the new field's JSDoc overstates its guarantee.** `packages/shared/src/interfaces/linux-email.interface.ts:34-40` calls `primaryEmail` "the user's primary **verified** email," but the fallback can supply the session OIDC email, which isn't necessarily verified. Behavior is unchanged and the "Match GET /api/profile/emails" comment is truthful; only the docstring overreaches. Suggest dropping "verified," or "primary email as resolved by auth-service, falling back to the session OIDC email."
7. **Nit — silent `catchError`.** See inline on `profile-linux-email.component.ts:239`.

### 5. Q&A this PR resolves

- **Q:** Does collapsing the granular retry into the whole-tab retry lose a user-visible capability? — **A:** No. The granular panel was reachable only when `getUserEmails()` failed while `getLinuxAlias()` succeeded, which required two independent client calls. With one call that state is unrepresentable, and both panels triggered the same `retry()`. Only the error wording changes.
- **Q:** Did dropping the `/api/profile/emails` fetch lose the alternate/verified email list from the forward dropdown? — **A:** No. `forwardOptions` only ever consumed `primary_email` from that response; verified alternates come from the separate `/api/profile/identities` call, which is untouched.
- **Q:** Does the impersonated user's alias state come from the right identity? — **A:** Yes for the alias itself — `profile.controller.ts:552` resolves `getEffectiveSub(req)`, so the emails read returns the target's data. Only the `primaryEmail` *fallback* can skew (gap 1).

### 6. Business rules & ADRs

- **Business rule:** the Linux.com email tab renders from exactly one server call; when the emails read fails, the whole tab degrades to a retry panel rather than showing a partially-populated form.
- **Business rule:** a user's primary forward option falls back to their session OIDC email when auth-service reports no primary — now enforced identically on both `/api/profile/emails` and `/api/profile/linux-email`.
- **ADR alignment:** *auth-service is the identity abstraction plane* → 🟢 complies, and improves on it by halving reads against that hot dependency. No other cross-cutting decision is touched; the diff stays entirely within the BFF's own response composition.
- **Privacy/access note:** the one new fact is `primaryEmail` — per-person PII, subject is the session user (or impersonation target). Gate of record is session auth on `/api/profile/*` plus `getEffectiveSub(req)` scoping at `:552`. **Sibling-path parity against `GET /api/profile/emails` holds**: same identity resolution, same fallback, same auth. `service_unavailable` and the claim response both correctly emit `null`, so no email is returned on paths that didn't read one. The single deviation is the impersonation fallback in gap 1 — which is itself a parity *match*, since both paths share it.

### Cross-cutting note (pre-existing, not introduced here)

The most valuable finding isn't in the diff — **CI enforces much less of this test code than it appears to**, which changes how much assurance this PR's E2E work actually buys. All verified by execution:

- **E2E never runs on a PR.** The E2E job in `.github/workflows/quality-check.yml:54-60` is commented out; `e2e-tests.yml` is `workflow_call`-only with a single live caller on a Sunday-02:00-UTC cron. This PR rewrote `profile-linux-email.spec.ts`, and that file's assertions — including the new zero-fetch regression guard the reviewers asked for — gate nothing on a PR.
- **E2E is never type-checked, and the config meant to do it is broken.** `apps/lfx-one/tsconfig.e2e.json` is referenced by nothing in the repo, and `tsc -p` against it fails with `TS5095` because it sets `"module": "commonjs"` while inheriting `moduleResolution: "bundler"`. The `const body: LinuxAliasData = {…}` annotations throughout this PR's stubs — the exact mechanism that would catch a stub missing the new **required** `primaryEmail` — are enforced by nothing. (All six stubs were updated correctly; that was diligence, not tooling.)
- **`apps/lfx-one` has no `check-types` script at all**, so `turbo run check-types` covers only `packages/shared` — visible in my run, which executed one task for two packages in scope.

Practical consequence: the component logic this PR changed has **zero PR-gated test coverage**. `profile-linux-email.component.ts` has no unit spec, and its only test file is the E2E spec CI doesn't run. Worth a separate ticket rather than anything against this PR.

---

**Verdict: 🟡 Minor comments** (merge readiness: already merged; nothing warranted blocking, but gap 1 is worth a follow-up ticket)

apps/lfx-one/e2e/profile-linux-email.spec.ts
```diff
@@ -194,36 +186,20 @@ test.describe('Linux.com email — forwarding target visibility', () => {
...
-  });
-});
-
-test.describe('Linux.com email — verified emails fetch failure', () => {
-  test('shows a retry panel instead of the empty-state message on the claim form', async ({ page }) => {
+    // Regression guard for the perf goal of this tab: the primary email now arrives inline on
+    // the linux-email response, so the tab must not make a second /api/profile/emails round-trip.
+    // Count GET hits (fulfilling so the assertion fails loudly rather than hanging if reintroduced).
+    let emailsFetchCount = 0;
     await page.route('**/api/profile/emails', (route) => {
```
luismoriguerra, line 195: **Non-blocking — this stub isn't method-scoped, and the counter it feeds is page-global.**

Good regression guard overall — asserting the second round-trip is gone is exactly the right way to protect this PR's perf goal. Two refinements.

**1. Not method-scoped.** This fulfills **every** method against `**/api/profile/emails`. Every other stub in this file guards with `if (route.request().method() !== 'GET') return route.fallback();` (`:43`, `:57`, `:65`, `:118`, `:151`, `:174`, `:201`), and the `stubProfileContext` helper it replaced did too. `docs/architecture/testing/testing-best-practices.md:191` calls this out directly: "Scope the stub to the specific endpoint… Stubbing broader patterns can mask bugs in unrelated fetches."

```ts
await page.route('**/api/profile/emails', (route) => {
  if (route.request().method() !== 'GET') return route.fallback();
  emailsFetchCount++;
  return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify({ primary_email: PRIMARY_EMAIL, alternate_emails: [] }) });
});
```

**2. The counter is page-global, and buried in a test named for something else.** `emailsFetchCount` counts `/api/profile/emails` GETs across the whole `/profile/identities` page, not just the Linux.com tab. It reads `0` only because `profile-edit-drawer.component.ts:124-127` fetches emails on drawer *open*, and the drawer — mounted globally by `profile-layout.component.html:77` — is closed. Any future sibling fetch on this page fails `expect(emailsFetchCount).toBe(0)` for a reason unrelated to the assertion's name.

The test also now carries two unrelated concerns ("shows the normal hint on a first-time claim" **and** "makes no second emails fetch") under a title mentioning only the first, against `testing-best-practices.md:208` ("Group by screen first, then by concern"). Splitting the guard into its own test named for the regression would make a future failure self-explaining.

PR #1192, opened 2026-07-24
fix(meetings): create toObservable in injection context

niravpatel27:
## Summary

Fixes the production `NG0203` error tracked in [Datadog](https://app.datadoghq.com/error-tracking/issue/a2bb2f98-6bcb-11f1-86a8-da7ad0900002) — ~1,970 occurrences in the last 30 days, 100% on the public meeting join page (`/meetings/:id`).

**Root cause:** `initializePublicMeetingPageviewTracking()` created `toObservable(this.parentProject)` inside a `switchMap` callback. That callback runs asynchronously when the project signal emits — outside Angular's injection context — and `toObservable()` without an `{injector}` option calls `inject(Injector)` internally, throwing `NG0203`. The failing branch only executes when the project has a `parent_uid`, so the error fired on every public meeting page view for a **sub-project** and never for top-level projects (confirmed by de-minifying the production stack trace to this exact call).

**Impact beyond the error noise:** the unhandled throw killed the tracking stream, so the Plausible pageview event silently never fired for sub-project meetings (since #709).

**Fix:** hoist the `toObservable(this.parentProject)` creation out of the `switchMap` into the constructor-time injection context. Replay semantics are equivalent — signals have no history, so the shared `ReplaySubject(1)`-backed observable emits the same latest value a per-callback creation would have.

## Ticket

[LFXV2-2839](https://linuxfoundation.atlassian.net/browse/LFXV2-2839)

## Testing

- `yarn check-types`, `yarn lint`, `yarn build` all pass (build exercises the SSR bundle)
- Post-commit review pass returned clean, including verification of effect-lifetime / replay semantics of the eager hoist

[LFXV2-2839]: https://linuxfoundation.atlassian.net/browse/LFXV2-2839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
@@ -1300,6 +1300,9 @@ export class MeetingJoinComponent implements OnInit {
     // doesn't block the pageview forever. Sub-projects whose parent doesn't land in time record
     // project_* correctly and leave foundation_* empty (see buildPageviewContext callsite below).
     const PARENT_FETCH_TIMEOUT_MS = 2000;
+    // Created here (injection context) rather than inside the switchMap below — toObservable()
+    // calls inject() internally and throws NG0203 when invoked from an async stream callback.
+    const parentProject$ = toObservable(this.parentProject);
     toObservable(this.project)
       .pipe(
         filter((p): p is Partial<Project> => !!p?.slug),
@@ -1308,7 +1311,7 @@ export class MeetingJoinComponent implements OnInit {
             return of({ project, parent: null as Project | null });
           }
           return merge(
-            toObservable(this.parentProject).pipe(
+            parentProject$.pipe(
               filter((parent): parent is Project => !!parent),
               map((parent) => ({ project, parent: parent as Project | null }))
             ),
--- a/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
@@ -1303,25 +1303,29 @@ export class MeetingJoinComponent implements OnInit {
     // Created here (injection context) rather than inside the switchMap below — toObservable()
     // calls inject() internally and throws NG0203 when invoked from an async stream callback.
     const parentProject$ = toObservable(this.parentProject);
-    toObservable(this.project)
-      .pipe(
-        filter((p): p is Partial<Project> => !!p?.slug),
-        switchMap((project) => {
-          if (!project.parent_uid) {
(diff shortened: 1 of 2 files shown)


niravpatel27:
Pushed `262b4076` addressing an Important finding from the pre-PR review pass: the deferred pageview also raced Plausible script load — `trackPage()` silently drops calls made before the script executes, and `take(1)` consumed the only emission, so the pageview was lost whenever project context resolved faster than the script loaded (the KB `take1-against-async-readiness` pattern from #709). `PlausibleService` now exposes a `ready` signal (set in `script.onload`) and the meeting join page holds its single pageview emission behind `combineLatest` with that readiness gate. No double-fire risk: the `onload` handler still skips deferred-pattern routes, so this component remains the sole owner of the pageview.

dealako (approved):
Hi @niravpatel27 — thanks for the surgical fix and the exceptionally clear write-up (de-minified stack trace → exact `toObservable` callsite is exactly the kind of root-cause proof that makes a review fast).

## Overall impression

Tightly scoped, correct, and well-reasoned. The core fix — hoisting `toObservable(this.parentProject)` out of the `switchMap` callback into the constructor-time injection context — directly addresses NG0203, and the replay-semantics argument holds: `toObservable` is `ReplaySubject(1)`-backed, so the late subscriber inside `switchMap` still receives the latest parent value, making the eager hoist behaviorally equivalent to per-callback creation. The follow-up readiness gate (`ready` signal + `combineLatest`) is a genuine second bug caught during review — without it the deferred pageview raced script load and `take(1)` could consume the only emission before `trackPage()` was live.

I traced the change end to end and confirmed:

- `initializePublicMeetingPageviewTracking()` runs from the **constructor** (`this.parentProject` is assigned on the preceding line), so both `toObservable` calls sit in a valid injection context.
- **No double-fire:** `script.onload` still early-returns on `deferredPageviewPattern` routes, so this component remains the sole owner of the join-page pageview.
- **Failure modes are equivalent to prior behavior:** if Plausible is disabled or the script errors, `ready` stays `false`, `combineLatest` never emits, and no pageview fires — matching the old `!this.analyticsReady` no-op guard (which is correctly retained as a defense-in-depth net).

## Bot reconciliation

- **CodeRabbit** — "No actionable comments." Agree.
- **Cursor Bugbot** — Low risk, analytics-only, no auth/data-handling impact. Agree.

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 0

The repo carries only Playwright E2E specs (no unit specs for components/services), so the absence of a unit test here is consistent with local convention rather than a gap.

## Decision

✅ **Approved**

PR #1195, opened 2026-07-24
feat(newsletters): add inline audience email add with no invite

niravpatel27:
## Summary

On the newsletter audience step, users can now type an email below the group picker and have it **asynchronously added as a member of the selected Newsletter group with the notification suppressed** — no more leaving the newsletter flow to add someone via the committee UI and coming back.

- An info banner explains that adding an email here automatically adds it to the selected group (one-time step, remembered for future newsletters).
- Each submitted email appears in a light list with an inline indicator next to the record: spinner while in flight → checkmark when added → muted "Already exists in this group" on duplicates → amber/red reason for invalid/failed. Nothing blocks: the input clears immediately, adds run as independent requests, and the user can keep typing or move to the next step while adds are in flight.
- Per-email state lives on the manage host (the stepper destroys the step panel on navigation), so in-flight adds and their indicators survive step changes. The recipient count pill refreshes as adds succeed, guarded against group-switch races.
- Retyping a failed/invalid/already email retries it; retyping a pending/added one is a no-op — keeping (group, email) unique for stable list tracking.

## How "No Invite" works

The BFF now accepts `?skip_notification=true` on `POST /api/committees/:id/members` and translates it to the upstream `X-Skip-Notification: true` header, which `lfx-v2-committee-service` already supports (Goa design `create-committee-member`; the `committee_member.created` NATS event carries `skip_notification: true` so the downstream email consumer suppresses the invite). **No upstream service changes are needed**: newsletter recipients are resolved live at send time from committee membership, so members added before send are automatically included. Existing callers of the endpoint are unchanged (flag defaults to false → notify as before).

## Personas

ED-facing (newsletter creation). Note for sign-off: `skip_notification` is honored for any authenticated caller with committee write access, matching the upstream contract — it suppresses the notification to people who never see LFX (added to a group without being emailed). This is the intended internal process for newsletter audiences.

## Review process & documented trade-offs

Four commits: the feature plus three review-driven fix commits (error-message extraction via the canonical `extractErrorMessage` helper, screen-reader announcements for the status list, sr-only input label, hardened `getStringQueryParam` read). Reviewer-verified notes:

- 409 → "already exists" mapping validated against the upstream Goa design (create-member declares exactly one Conflict: "Member already exists").
- A flagged recipient-count race was verified to be a false positive: the member record and its by-committee index are written before the 201 returns; `X-Sync` only defers query-service indexing/FGA messaging, which recipient resolution does not use.
- Branch name carries a descriptive suffix beyond `feat/LFXV2-2846`, consistent with sibling branches on origin.
- Settled status entries persist for the page session (bounded by hand-typed emails); add-only by design — removals stay in the committee UI.

## Testing

- `yarn lint`, `yarn check-types`, `yarn build` (SSR bundle) all green; license headers verified; no protected files touched.
- Existing newsletter Playwright specs unaffected (new section renders only after a group is selected); new `newsletter-audience-email-add-*` testids enable a future spec.

JIRA: [LFXV2-2846](https://linuxfoundation.atlassian.net/browse/LFXV2-2846)

[LFXV2-2846]: https://linuxfoundation.atlassian.net/browse/LFXV2-2846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Known trade-offs (v1, from review triage)

- **Invalid email clears the input**: a typo becomes an inline "Not a valid email" row and the field clears, so the user retypes. Control-level validation (account-settings pattern) is a follow-up candidate.
- **Network/5xx failures may show Angular's transport string** as the inline reason (`extractErrorMessage` fallback trait shared by all its call sites); a status-0/5xx guard is a follow-up candidate.
- **Previously unsubscribed addresses**: the member add genuinely succeeds (checkmark is truthful) but newsletter delivery respects the project's unsubscribe list at send time, so such an address won't receive sends and the recipient count won't move. Surfacing an `unsubscribed` status needs upstream support for distinguishing skipped-as-unsubscribed.
- **Send is gated on pending adds** (fe2bc90f): an in-flight add for the selected group blocks Send for the seconds it takes to settle, so a just-added email can't be silently omitted from an immediate send.

### DIFF
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
@@ -64,21 +64,27 @@
       </lfx-select>
 
       @if (selectedCommitteeUid()) {
-        <div class="flex flex-col gap-2 mt-2" data-testid="newsletter-audience-email-add-section">
-          <div class="rounded-md border border-blue-200 bg-blue-50 px-3 py-2 text-sm text-blue-800" data-testid="newsletter-audience-email-add-banner">
-            Adding an email here automatically adds it to the <span class="font-medium">{{ selectedCommitteeName() }}</span> group, so it's remembered for
-            future newsletters — a one-time step.
+        <!-- Nested sub-panel: tinted + indented so it reads as an optional
+             action on the selected group, not a sibling of the Group picker. -->
+        <div class="mt-2 ml-3 flex flex-col gap-2 rounded-lg border border-gray-200 bg-gray-50 p-3" data-testid="newsletter-audience-email-add-section">
+          <div class="flex items-center gap-2">
+            <i class="fa-light fa-user-plus text-gray-400" aria-hidden="true"></i>
+            <label class="text-sm font-medium text-gray-800" for="newsletter-audience-email-add">Someone missing? Add them to this group</label>
           </div>
+          <p class="text-xs text-gray-500" data-testid="newsletter-audience-email-add-banner">
+            New emails are added to the <span class="font-medium text-gray-700">{{ selectedCommitteeName() }}</span> group automatically — a one-time step,
+            remembered for future newsletters.
+          </p>
 
           <form [formGroup]="emailAddForm" (ngSubmit)="onAddEmail()" class="flex items-center gap-2">
-            <div class="flex-1">
-              <label class="sr-only" for="newsletter-audience-email-add">Email to add to the group</label>
+            <div class="w-full max-w-md">
               <lfx-input-text
                 [form]="emailAddForm"
                 control="email"
                 type="email"
-                id="newsletter-audience-email-add"
-                placeholder="Add an email to this group…"
+                [id]="'newsletter-audience-email-add'"
+                size="small"
+                placeholder="name@example.com"
                 dataTest="newsletter-audience-email-add-input">
               </lfx-input-text>
             </div>
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
@@ -86,7 +86,7 @@
           </form>
 
           @if (visibleEmailAdds().length > 0) {
-            <ul class="flex flex-col gap-1" data-testid="newsletter-audience-email-add-list">
+            <ul class="flex flex-col gap-1" aria-live="polite" data-testid="newsletter-audience-email-add-list">
               @for (entry of visibleEmailAdds(); track entry.email) {
                 <li class="flex items-center gap-2 px-1 py-0.5 text-sm text-gray-700" data-testid="newsletter-audience-email-add-item">
                   <span class="break-all">{{ entry.email }}</span>
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
@@ -64,7 +64,10 @@
       </lfx-select>
 
       @if (selectedCommitteeUid()) {
-        <div class="flex flex-col gap-2 mt-2" data-testid="newsletter-audience-email-add-section">
+        <!-- aria-live sits on the always-present wrapper (not the conditionally
+             rendered list) so the first entry of a group is announced too — a
+             live region only announces mutations after it registers. -->
+        <div class="flex flex-col gap-2 mt-2" aria-live="polite" data-testid="newsletter-audience-email-add-section">
           <div class="rounded-md border border-blue-200 bg-blue-50 px-3 py-2 text-sm text-blue-800" data-testid="newsletter-audience-email-add-banner">
             Adding an email here automatically adds it to the <span class="font-medium">{{ selectedCommitteeName() }}</span> group, so it's remembered for
             future newsletters — a one-time step.
@@ -86,7 +89,7 @@
           </form>
 
           @if (visibleEmailAdds().length > 0) {
-            <ul class="flex flex-col gap-1" aria-live="polite" data-testid="newsletter-audience-email-add-list">
+            <ul class="flex flex-col gap-1" data-testid="newsletter-audience-email-add-list">
               @for (entry of visibleEmailAdds(); track entry.email) {
                 <li class="flex items-center gap-2 px-1 py-0.5 text-sm text-gray-700" data-testid="newsletter-audience-email-add-item">
                   <span class="break-all">{{ entry.email }}</span>
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
@@ -62,6 +62,60 @@
         styleClass="w-full"
         data-testid="newsletter-audience-committees">
       </lfx-select>
+
+      @if (selectedCommitteeUid()) {
+        <div class="flex flex-col gap-2 mt-2" data-testid="newsletter-audience-email-add-section">
+          <div class="rounded-md border border-blue-200 bg-blue-50 px-3 py-2 text-sm text-blue-800" data-testid="newsletter-audience-email-add-banner">
+            Adding an email here automatically adds it to the <span class="font-medium">{{ selectedCommitteeName() }}</span> group, so it's remembered for
+            future newsletters — a one-time step.
+          </div>
+
+          <form [formGroup]="emailAddForm" (ngSubmit)="onAddEmail()" class="flex items-center gap-2">
+            <div class="flex-1">
+              <lfx-input-text
+                [form]="emailAddForm"
+                control="email"
+                type="email"
+                id="newsletter-audience-email-add"
+                placeholder="Add an email to this group…"
+                dataTest="newsletter-audience-email-add-input">
+              </lfx-input-text>
+            </div>
+            <lfx-button label="Add" type="submit" size="small" severity="secondary" [outlined]="true" data-testid="newsletter-audience-email-add-btn">
+            </lfx-button>
+          </form>
+
+          @if (visibleEmailAdds().length > 0) {
+            <ul class="flex flex-col gap-1" data-testid="newsletter-audience-email-add-list">
+              @for (entry of visibleEmailAdds(); track entry.email) {
+                <li class="flex items-center gap-2 px-1 py-0.5 text-sm text-gray-700" data-testid="newsletter-audience-email-add-item">
+                  <span class="break-all">{{ entry.email }}</span>
+                  @switch (entry.status) {
+                    @case ('pending') {
+                      <p-progressSpinner styleClass="!w-3.5 !h-3.5" strokeWidth="4" ariaLabel="Adding to group"></p-progressSpinner>
+                    }
+                    @case ('added') {
+                      <i class="fa-light fa-check text-emerald-600" aria-hidden="true"></i>
+                      <span class="sr-only">Added to group</span>
+                    }
+                    @case ('already') {
+                      <span class="text-xs text-gray-500">Already exists in this group</span>
+                    }
+                    @case ('invalid') {
+                      <i class="fa-light fa-circle-exclamation text-amber-600" aria-hidden="true"></i>
+                      <span class="text-xs text-amber-700">{{ entry.reason }}</span>
+                    }
+                    @case ('failed') {
+                      <i class="fa-light fa-circle-exclamation text-red-600" aria-hidden="true"></i>
+                      <span class="text-xs text-red-700">{{ entry.reason }}</span>
+                    }
+                  }
+                </li>
+              }
+            </ul>
+          }
+        </div>
+      }
     }
   </div>
 
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
@@ -64,10 +64,7 @@
       </lfx-select>
 
       @if (selectedCommitteeUid()) {
-        <!-- aria-live sits on the always-present wrapper (not the conditionally
-             rendered list) so the first entry of a group is announced too — a
-             live region only announces mutations after it registers. -->
-        <div class="flex flex-col gap-2 mt-2" aria-live="polite" data-testid="newsletter-audience-email-add-section">
+        <div class="flex flex-col gap-2 mt-2" data-testid="newsletter-audience-email-add-section">
           <div class="rounded-md border border-blue-200 bg-blue-50 px-3 py-2 text-sm text-blue-800" data-testid="newsletter-audience-email-add-banner">
             Adding an email here automatically adds it to the <span class="font-medium">{{ selectedCommitteeName() }}</span> group, so it's remembered for
             future newsletters — a one-time step.
@@ -75,6 +72,7 @@
 
           <form [formGroup]="emailAddForm" (ngSubmit)="onAddEmail()" class="flex items-center gap-2">
             <div class="flex-1">
+              <label class="sr-only" for="newsletter-audience-email-add">Email to add to the group</label>
               <lfx-input-text
                 [form]="emailAddForm"
                 control="email"
@@ -88,35 +86,37 @@
             </lfx-button>
           </form>
 
-          @if (visibleEmailAdds().length > 0) {
-            <ul class="flex flex-col gap-1" data-testid="newsletter-audience-email-add-list">
-              @for (entry of visibleEmailAdds(); track entry.email) {
-                <li class="flex items-center gap-2 px-1 py-0.5 text-sm text-gray-700" data-testid="newsletter-audience-email-add-item">
-                  <span class="break-all">{{ entry.email }}</span>
-                  @switch (entry.status) {
-                    @case ('pending') {
-                      <p-progressSpinner styleClass="!w-3.5 !h-3.5" strokeWidth="4" ariaLabel="Adding to group"></p-progressSpinner>
-                    }
-                    @case ('added') {
-                      <i class="fa-light fa-check text-emerald-600" aria-hidden="true"></i>
-                      <span class="sr-only">Added to group</span>
-                    }
-                    @case ('already') {
-                      <span class="text-xs text-gray-500">Already exists in this group</span>
-                    }
-                    @case ('invalid') {
-                      <i class="fa-light fa-circle-exclamation text-amber-600" aria-hidden="true"></i>
-                      <span class="text-xs text-amber-700">{{ entry.reason }}</span>
-                    }
-                    @case ('failed') {
-                      <i class="fa-light fa-circle-exclamation text-red-600" aria-hidden="true"></i>
-                      <span class="text-xs text-red-700">{{ entry.reason }}</span>
-                    }
+          <!-- Rendered (and registered as a live region) as soon as a group is
+               selected, before any entry exists — a live region only announces
+               mutations after it registers, so the first entry announces too.
+               The banner sits outside the region to avoid group-switch chatter. -->
+          <ul class="flex flex-col gap-1" aria-live="polite" data-testid="newsletter-audience-email-add-list">
+            @for (entry of visibleEmailAdds(); track entry.email) {
+              <li class="flex items-center gap-2 px-1 py-0.5 text-sm text-gray-700" data-testid="newsletter-audience-email-add-item">
+                <span class="break-all">{{ entry.email }}</span>
+                @switch (entry.status) {
+                  @case ('pending') {
+                    <p-progressSpinner styleClass="!w-3.5 !h-3.5" strokeWidth="4" ariaLabel="Adding to group"></p-progressSpinner>
                   }
-                </li>
-              }
-            </ul>
-          }
+                  @case ('added') {
+                    <i class="fa-light fa-check text-emerald-600" aria-hidden="true"></i>
+                    <span class="sr-only">Added to group</span>
+                  }
+                  @case ('already') {
+                    <span class="text-xs text-gray-500">Already exists in this group</span>
+                  }
+                  @case ('invalid') {
+                    <i class="fa-light fa-circle-exclamation text-amber-600" aria-hidden="true"></i>
+                    <span class="text-xs text-amber-700">{{ entry.reason }}</span>
+                  }
+                  @case ('failed') {
+                    <i class="fa-light fa-circle-exclamation text-red-600" aria-hidden="true"></i>
+                    <span class="text-xs text-red-700">{{ entry.reason }}</span>
+                  }
+                }
+              </li>
+            }
+          </ul>
         </div>
       }
     }
--- a/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.ts
+++ b/apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.ts
@@ -4,9 +4,11 @@
 import { Component, computed, DestroyRef, inject, input, output, Signal, signal, viewChild } from '@angular/core';
 import { takeUntilDestroyed, toObservable } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
+import { ButtonComponent } from '@components/button/button.component';
+import { InputTextComponent } from '@components/input-text/input-text.component';
 import { SelectComponent } from '@components/select/select.component';
 import { NEWSLETTER_COMMITTEE_CATEGORY } from '@lfx-one/shared/constants';
-import { Committee, NewsletterCommitteeOption, NewsletterRecipient } from '@lfx-one/shared/interfaces';
+import { Committee, NewsletterAudienceEmailAdd, NewsletterCommitteeOption, NewsletterRecipient } from '@lfx-one/shared/interfaces';
 import { NewsletterService } from '@services/newsletter.service';
 import { Popover, PopoverModule } from 'primeng/popover';
 import { ProgressSpinnerModule } from 'primeng/progressspinner';
@@ -14,7 +16,7 @@ import { finalize, take } from 'rxjs';
 
 @Component({
   selector: 'lfx-newsletter-audience-step',
-  imports: [ReactiveFormsModule, SelectComponent, PopoverModule, ProgressSpinnerModule],
+  imports: [ReactiveFormsModule, SelectComponent, PopoverModule, ProgressSpinnerModule, InputTextComponent, ButtonComponent],
   templateUrl: './newsletter-audience-step.component.html',
 })
 export class NewsletterAudienceStepComponent {
@@ -39,9 +41,14 @@ export class NewsletterAudienceStepComponent {
   public readonly committeeUidsValue = input<string[]>([]);
   public readonly recipientCount = input<number | null>(null);
   public readonly recipientCountLoading = input<boolean>(false);
+  // Per-email add state is owned by the host (this panel is destroyed on step
+  // navigation, so in-flight adds and their statuses must outlive it).
+  public readonly emailAdds = input<NewsletterAudienceEmailAdd[]>([]);
 
   // === Outputs ===
   public readonly retryCommittees = output<void>();
+  // Raw email string as typed; validation, dedupe, and state live on the host.
+  public readonly addEmail = output<string>();
 
   // === Forms ===
   // The shared `form` input carries `committeeUids: string[]` end-to-end — recipient
@@ -50,6 +57,13 @@ export class NewsletterAudienceStepComponent {
   // the scalar selection; it's bridged to the shared array control in initSync().
   protected readonly audienceForm = new FormGroup({
     committeeUid: new FormControl<string | null>(null),
+  });
+
+  // Inline "add email to the selected group" input. Submitting emits the raw
+  // value to the host and clears the control immediately — adds are async and
+  // never block further typing or step navigation.
+  protected readonly emailAddForm = new FormGroup({
+    email: new FormControl<string>('', { nonNullable: true }),
   });
 
   // === Signals ===
@@ -70,9 +84,30 @@ export class NewsletterAudienceStepComponent {
   );
   protected readonly selectedCount: Signal<number> = computed(() => this.committeeUidsValue().length);
   protected readonly hasCommittees = computed(() => this.committeeOptions().length > 0);
+  protected readonly selectedCommitteeUid = computed<string | null>(() => this.committeeUidsValue()[0] ?? null);
+  protected readonly selectedCommitteeName = computed<string>(() => {
+    const uid = this.selectedCommitteeUid();
+    if (!uid) return 'the selected';
+    const committee = this.committees().find((c) => c.uid === uid);
+    return committee?.name || 'the selected';
+  });
+  // Adds are scoped per group: switching groups swaps the visible list, and
+  // entries for a previously selected group reappear on switch-back.
+  protected readonly visibleEmailAdds = computed<NewsletterAudienceEmailAdd[]>(() => {
+    const uid = this.selectedCommitteeUid();
+    if (!uid) return [];
+    return this.emailAdds().filter((e) => e.committeeUid === uid);
+  });
 
   public constructor() {
     this.initSync();
+  }
+
+  protected onAddEmail(): void {
+    const value = this.emailAddForm.controls.email.value;
+    if (!value.trim()) return;
+    this.addEmail.emit(value);
+    this.emailAddForm.controls.email.setValue('');
   }
 
   protected onShowRecipients(event: Event): void {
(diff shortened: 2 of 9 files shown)


apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
```diff
@@ -373,6 +382,55 @@ export class NewsletterManageComponent {
...
+    // the page. createCommitteeMember pipes take(1) and HTTP observables
+    // complete after one response, so each subscription cleans itself up.
+    this.committeeService.createCommitteeMember(committeeUid, { email }, { skipNotification: true }).subscribe({
+      next: () => {
+        this.updateAudienceEmailAdd(committeeUid, email, { status: 'added' });
+        // Refresh the recipient pill only if this group is still the selected
+        // audience — an add resolving after a group switch must not clobber
+        // the new group's count.
+        if (this.committeeUidsValue()[0] === committeeUid) {
+          this.fetchRecipientCountFor(this.committeeUidsValue());
```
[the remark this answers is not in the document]
niravpatel27: Done — the recipient-count pipeline now verifies at response-apply time that the requested uids still match the current selection (`uidsEqual(uids, committeeUidsValue())`) before setting the count, on both the next and error paths. This closes the debounce-window race you described; note it was transient rather than sticky (the debounced selection change always issues a corrective fetch), but the response-time check is strictly better and also hardens the pre-existing draft-load path. See commit be08a478.

apps/lfx-one/src/app/shared/services/committee.service.ts
```diff
@@ -104,8 +105,14 @@ export class CommitteeService {
...
-  public createCommitteeMember(committeeId: string, memberData: CreateCommitteeMemberRequest): Observable<CommitteeMember> {
-    return this.http.post<CommitteeMember>(`/api/committees/${committeeId}/members`, memberData).pipe(take(1));
+  public createCommitteeMember(
+    committeeId: string,
+    memberData: CreateCommitteeMemberRequest,
+    options?: CreateCommitteeMemberOptions
+  ): Observable<CommitteeMember> {
+    const params = options?.skipNotification ? new HttpParams().set('skip_notification', 'true') : undefined;
+
+    return this.http.post<CommitteeMember>(`/api/committees/${committeeId}/members`, memberData, { params }).pipe(take(1));
```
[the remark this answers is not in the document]
niravpatel27: Done — added JSDoc to createCommitteeMember documenting the skipNotification option: notifications are sent by default, suppressed when true, forwarded upstream as the X-Skip-Notification header. See commit be08a478.

niravpatel27:
## Review Feedback Addressed

Commit: be08a478

### Changes Made
- **newsletter-manage.component.ts**: recipient-count responses are now applied only when the requested uids still match the current selection, closing the debounce-window race on rapid group switches (per copilot-pull-request-reviewer)
- **committee.service.ts**: added JSDoc for `createCommitteeMember` documenting the `skipNotification` option (notify by default; suppressed when true via the upstream X-Skip-Notification header) (per coderabbitai)

### Threads Resolved

apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
```diff
@@ -62,6 +62,63 @@ <h2 class="text-base font-semibold text-gray-900">Who's this for?</h2>
...
+          <form [formGroup]="emailAddForm" (ngSubmit)="onAddEmail()" class="flex items-center gap-2">
+            <div class="flex-1">
+              <label class="sr-only" for="newsletter-audience-email-add">Email to add to the group</label>
+              <lfx-input-text
+                [form]="emailAddForm"
+                control="email"
+                type="email"
+                id="newsletter-audience-email-add"
+                placeholder="Add an email to this group…"
+                dataTest="newsletter-audience-email-add-input">
```
luismoriguerra, line 88: **Non-blocking — this test hook won't resolve via `getByTestId`.**

`dataTest=` is rendered by `InputTextComponent` as `[attr.data-test]` (`input-text.component.html:18,30`), not `data-testid`. `playwright.config.ts` sets no `testIdAttribute` override, so Playwright's default `data-testid` applies — meaning `getByTestId('newsletter-audience-email-add-input')` will not match this element.

Every other new hook in this same template uses `data-testid` (the section, banner, button, list, and item), and the nearest sibling — the account-settings inline email-add form — puts it on the `<lfx-input-text>` host directly:

```html
<!-- account-settings.component.html:158 -->
data-testid="add-email-input">
```

This matters because it's the one element a future spec has to type into, so the PR description's claim that the new testids "enable a future spec" doesn't hold as written.

```suggestion
                data-testid="newsletter-audience-email-add-input">
```

apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.ts
```diff
@@ -70,11 +84,33 @@ export class NewsletterAudienceStepComponent {
...
 
   public constructor() {
     this.initSync();
   }
 
+  protected onAddEmail(): void {
+    const value = this.emailAddForm.controls.email.value;
+    if (!value.trim()) return;
+    this.addEmail.emit(value);
+    this.emailAddForm.controls.email.setValue('');
```
luismoriguerra, line 111: **Non-blocking — a typo silently costs the user the whole address.**

`setValue('')` runs unconditionally after the emit, and the host marks anything failing `isValidEmail` as an `invalid` entry (`newsletter-manage.component.ts:403-408`). So typing `alice@` clears the input and leaves a permanent red row that the user can only fix by retyping the full address from scratch — which is the most likely failure mode for a hand-typed email field.

The repo's canonical inline email-add form handles this with control-level validation instead of a post-hoc status:

```ts
// account-settings.component.ts:80-81
public addEmailForm = new FormGroup({
  email: new FormControl('', [Validators.required, Validators.email]),
```

```html
<!-- account-settings.component.html:160-170 -->
@if (addEmailForm.get('email')?.invalid && addEmailForm.get('email')?.touched) {
  <div class="text-red-600 text-xs mt-1" data-testid="add-email-error">Please enter a valid email address</div>
}
...
[disabled]="addEmailForm.invalid || sendingCode() || impersonating()"
```

Suggested change: add `[Validators.required, Validators.email]` to the control (line 65-67), gate the Add button on `emailAddForm.invalid`, and only clear on a valid submit. The `invalid` status on `NewsletterAudienceEmailAdd` is still worth keeping for host-side edge cases, but it stops being the normal path for a typo.

apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
```diff
@@ -373,6 +382,55 @@ export class NewsletterManageComponent {
...
+        }
+      },
+      error: (err: HttpErrorResponse) => {
+        if (err.status === 409) {
+          this.updateAudienceEmailAdd(committeeUid, email, { status: 'already' });
+        } else {
+          this.updateAudienceEmailAdd(committeeUid, email, { status: 'failed', reason: extractErrorMessage(err, 'Could not add. Try again.') });
```
luismoriguerra, line 438: **Non-blocking — this can render Angular's raw transport string in the UI, and the fallback is effectively unreachable.**

`extractErrorMessage` ends with `return error.message || fallback` (`http-error.utils.ts:44`). On a network failure or timeout, `HttpErrorResponse.message` is Angular's generated string, so the user sees this as the small red inline text next to their email:

> Http failure response for /api/committees/…: 0 Unknown Error

And because the BFF always populates a body, `'Could not add. Try again.'` is essentially never reached.

One nuance worth flagging, since the obvious fix is wrong: **don't swap wholesale to `getHttpErrorDetail`.** That helper reads `err.error?.message`, but this BFF emits errors under an `error` key, not `message` (`base.error.ts:76-84`, `error-handler.middleware.ts:91-95`) — so `upstream` would always be `undefined` and you'd *lose* the genuinely useful upstream 4xx text (e.g. the `business_email_required` validation message) in exchange for generic per-status copy.

The targeted fix is to keep `extractErrorMessage` for 4xx and force the generic fallback for status `0` and `>= 500`:

```ts
const reason = err.status === 0 || err.status >= 500 ? 'Could not add. Try again.' : extractErrorMessage(err, 'Could not add. Try again.');
this.updateAudienceEmailAdd(committeeUid, email, { status: 'failed', reason });
```

apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
```diff
@@ -373,6 +382,55 @@ export class NewsletterManageComponent {
...
+
+    if (entry.status === 'invalid') return;
+
+    // Intentionally NOT takeUntilDestroyed: HttpClient aborts the request on
+    // unsubscribe, which would silently cancel in-flight adds if the user left
+    // the page. createCommitteeMember pipes take(1) and HTTP observables
+    // complete after one response, so each subscription cleans itself up.
+    this.committeeService.createCommitteeMember(committeeUid, { email }, { skipNotification: true }).subscribe({
+      next: () => {
+        this.updateAudienceEmailAdd(committeeUid, email, { status: 'added' });
```
luismoriguerra, line 426: **Non-blocking — an unsubscribed address gets a green checkmark here but will never receive the newsletter, and the recipient count won't move.**

I traced the recipient resolution into `lfx-v2-newsletter-service`. Every address in the project's unsubscribe list is dropped **before** the count is taken:

```go
// send_orchestrator.go:575-596
excluded, err := o.listUnsubscribed(ctx, projectUID)
...
if _, unsub := excluded[email]; unsub {
    skipped++
    continue
}
```

So for an address that previously unsubscribed from this project, the flow is: the committee-member create genuinely succeeds → this handler sets `status: 'added'` → the UI shows the ✅ → the count refresh at line 421 returns the *same* number → and the person is permanently excluded from sends.

Respecting the unsubscribe is obviously correct. Showing the author an unqualified "added" is the part worth reconsidering — the pill silently not moving is the only cue, and it's easy to read as a stale count rather than an exclusion. Options, roughly in order of effort: leave it and document the behavior, or add an `unsubscribed` status if the recipients endpoint can surface that (it would need the newsletter service to distinguish skipped-as-unsubscribed, which `resolveRecipients` already counts internally as `skipped`).

Not blocking, and possibly acceptable for the first iteration — but I don't think it's currently a known trade-off, so flagging it rather than leaving it to be discovered in support.

apps/lfx-one/src/server/controllers/committee.controller.ts
```diff
@@ -382,9 +382,11 @@ export class CommitteeController {
    */
   public async createCommitteeMember(req: Request, res: Response, next: NextFunction): Promise<void> {
     const { id } = req.params;
+    const skipNotification = getStringQueryParam(req, 'skip_notification') === 'true';
```
luismoriguerra, line 385: **🔐 Non-blocking, but this is the finding I'd want signed off before merge — silent member-add is now reachable browser-wide, not just from the newsletter flow.**

The flag is read unconditionally for **every** request to `POST /api/committees/:id/members`, with no reference to the newsletter flow, no caller allowlist, and no committee-category restriction. The client-side helper only opts in from the newsletter host, but the endpoint doesn't know that.

Let me be precise about what this is and isn't, because "silent add" sounds worse than it is:

**Not an authorization change.** Heimdall gates this route on OpenFGA `writer` on `committee:{uid}` (`charts/lfx-v2-committee-service/templates/ruleset.yaml:346-371`) — the same relation the existing member-form dialog, committee bulk add, and invite flow already require. I checked all three sibling paths; none is weaker or stronger than this one. Upstream also already exposes this header to any writer, so the BFF isn't widening the platform contract.

**It is a notice change, and the consequence chain is worth an explicit decision.** Any committee writer can now `curl` this endpoint with `?skip_notification=true` and add a person who receives no signal at all. When the typed email resolves to an existing LFID:

- `enrichMember` backfills their given name, family name, and organization onto the member record (`committee_service.go:1798-1838`);
- an FGA access-control message grants them `committee#member` (`committee_member_writer.go:1188-1224`), which cascades to `viewer` and `roster_viewer`;
- and the suppressed notification was the only mechanism that would have told them.

Their first signal becomes the newsletter itself, which does have an opt-out — so it's delayed notice rather than none, but it arrives *after* the roster entry, the identity join, and the access grant.

The PR description flags this under "Note for sign-off" and describes it accurately, and Bugbot's summary noted the reach. I couldn't find evidence the sign-off actually happened — no linked privacy review, no approval referencing that note.

So this is a request for a decision, not a code change. Two ways to close it:

1. **Record the sign-off** — someone with product/privacy authority confirms an endpoint-wide silent-add capability is accepted, and it gets logged as an architecture decision (the decision log currently has no entry covering "granting access without notifying the grantee").
2. **Scope it down** — gate the flag here on the committee's category (newsletter-only), so the capability matches the use case that justified it rather than being available to every caller of the endpoint.

Auditability is already covered either way: the flag is logged on both hops (line 389 here, and `server/services/committee.service.ts:492`).

One piece of context, explicitly *not* a finding against this PR: `blockDuringImpersonation` is only applied to the profile and enrollment routes, not the committees router — so an impersonating operator can perform these adds. That predates this PR; it's only worth mentioning because suppressing the notification also removes the email trail that would otherwise have reached the affected person.

luismoriguerra (commented):
## Independent audit — PR #1195

Audited at head `be08a478` in an isolated worktree, with all gates re-run locally. **Verdict: 🟡 Minor comments** — nothing blocking. Merge readiness: *merge after confirming the `skip_notification` scope decision (finding 5)*; findings 1–4 are worth fixing but none block.

Lanes run: `format:check` / `lint:check` / `check-types` / `yarn build` re-run in a clean worktree · secrets & critical-constant drift · code standards · architecture/ADR alignment · privacy/access/PII · platform-contract coherence. The contract lane was scoped down: the diff touches no Goa design, RuleSet, HTTPRoute, or KV wiring, so it was reduced to verifying the upstream header contract and the write→read path end to end.

---

### Verification: the upstream contract holds, end to end

An invented near-miss header name would have made this feature a silent no-op, so I traced every link in the pinned `lfx-v2-committee-service` rather than trusting the PR description:

| Link | Evidence |
|---|---|
| Goa design declares the header | `cmd/committee-api/design/committee.go:329` — `dsl.Header("skip_notification:X-Skip-Notification")`, backed by a `Boolean` attribute defaulting to `false` at `design/type.go:450-457` |
| Generated decoder parses it | `gen/http/committee_service/server/encode_decode.go:1027` — `strconv.ParseBool` |
| It reaches the NATS event | `internal/service/committee_member_writer.go:324-327` — `CommitteeMemberMessageData{SkipNotification: ...}` |
| The consumer honors it | `internal/service/message_handler.go:603-609` — returns before **both** the invite-service branch and the direct-notification branch |

The BFF wiring is also correct: `customHeaders` is the 7th positional parameter of `proxyRequest` (`microservice-proxy.service.ts:14-35`), which is where the PR passes it.

### Correction: the recipient-count race is a genuine false positive ✅

Two of my audit lanes independently flagged a missing `X-Sync: true` on this mutation, arguing the immediate count refresh would read stale data. **Both were wrong, and @niravpatel27's rebuttal in the PR description is right.** Recording the proof here so this doesn't get re-litigated in a future review:

- `lfx-v2-newsletter-service` resolves recipients via `RecipientCount` → `resolveRecipients` → `committee.ListMembers` over NATS subject `lfx.committee-api.list_members` (`internal/service/send_orchestrator.go:533,563`).
- That lands on committee-service `ListMembersByCommittee`, which does a **filtered NATS KV scan** (`internal/infrastructure/nats/storage.go:328-340`). It never touches OpenSearch.
- The member record and its by-committee index are both written at steps 8 and 8b (`committee_member_writer.go:241-269`) **before** the 201 returns. `sync` only gates `publishMemberMessages` at line 328.

So the count refresh is consistent by construction, and the async indexer lag is irrelevant to this read path.

### Gates

`format:check`, `lint:check`, `check-types`, and `yarn build` (SSR bundle) all pass in a clean worktree. The single lint warning is pre-existing in `campaigns.component.ts` and unrelated to this diff. CI on the PR is fully green.

---

## Findings

Five are posted inline. Summarized here for the record:

1. **Non-blocking — the email input's test hook won't resolve via `getByTestId`.** `dataTest=` renders as `data-test`; every other new hook in the template uses `data-testid`. This makes the PR body's claim that the new testids "enable a future spec" untrue for the one element a spec must type into. *(inline)*
2. **Non-blocking — a typo silently costs the user the whole address.** The control is cleared unconditionally on submit and there are no validators, so `alice@` wipes the input and leaves a permanent `invalid` row. The sibling account-settings form gates its button on `addEmailForm.invalid`. *(inline)*
3. **Non-blocking — the failure message can surface Angular's raw transport string** (`Http failure response for …: 0 Unknown Error`), and the supplied fallback is effectively unreachable. Note the nuance: swapping wholesale to `getHttpErrorDetail` would *lose* useful upstream 4xx text, because this BFF emits errors under an `error` key rather than `message`. *(inline)*
4. **Non-blocking — an unsubscribed address shows a green checkmark but will never receive the newsletter, and the count won't move.** No bot caught this one; it's the behavioral gap I'd most want considered, even if the answer is "acceptable for now." *(inline)*
5. **🔐 Non-blocking, but needs the sign-off the PR asks for — silent member-add is now reachable browser-wide, not just from the newsletter flow.** Authorization is genuinely unchanged, so this is a *notice* change rather than an access change — but the consequence chain when the email resolves to an existing LFID is worth an explicit decision. Bugbot's summary noted the reach; the inline comment adds the consequences and a concrete scoping option. *(inline)*
6. **Non-blocking — no test coverage.** Already threaded by Copilot at `newsletter-manage.component.ts:414` and still unanswered, so not duplicated inline. Worth noting it's an e2e gap rather than convention drift (the repo has no unit specs under `src/app` at all), and it's blocked on finding 1.
7. **Nit — undocumented upstream precondition.** If the newsletter group has `business_email_required` or voting enabled, an email-only add is rejected (`committee_member.go:284-302`) and the user sees the raw upstream validation string under `failed`. Probably unreachable for newsletter-category committees, but nothing in the UI prevents it.
8. **Nit — `audienceEmailAdds` is never pruned** (`newsletter-manage.component.ts:173`). Bounded by hand-typing, so not a real leak; dropping `added` entries once the count refreshes would keep the list readable.

Prior review state: Copilot's group-switch race comment and CodeRabbit's JSDoc comment were both addressed in follow-up commits and are confirmed resolved. I re-verified the group-switch guard independently.

### Security lane — clean on everything that matters

No secrets in the diff. No existing constant changed. The header name matches upstream exactly (verified against the Goa design, not assumed). No header injection is possible — the value is the literal `'true'` gated by a boolean, so no user-controlled string reaches a header. `getStringQueryParam` returns `undefined` for arrays and objects, so every tampering shape fails safe to "send the notification." The email regex `/^[^\s@]+@[^\s@.]+(?:\.[^\s@.]+)+$/` has no ambiguous repetition and therefore no ReDoS. The template uses only Angular interpolation, with no `innerHTML` or sanitizer bypass.

### Business rules this PR encodes

- A newsletter audience group **is** a real committee, and adding an email to it is a full committee-member create with the invite suppressed — the person is never told.
- `409` means "already a member of this group" and is a benign, non-error outcome. Verified as the only Conflict the create path constructs (`internal/infrastructure/nats/storage.go:611-616`, keyed on `SHA-256(lower(committee_uid)|lower(email))`), so the mapping is sound.
- Retyping retries `invalid`/`failed`/`already` entries but is a no-op for `pending`/`added`, keeping `(committeeUid, email)` unique for stable `@for` tracking.
- Adds target only the first selected group (`committeeUidsValue()[0]`), since the picker is single-select even though the shared form carries an array.

### ADR alignment

- **Goa design-first APIs (generated `gen/` is never hand-edited)** → 🟢 the header is a first-class design attribute and nothing under `gen/` was touched.
- **Dual authorization planes (Heimdall edge vs query-service access-check)** → 🟢 no FGA logic added to the BFF; the write stays gated at the edge on `committee:{uid}#writer`.
- **Baseline API conventions (versioning, errors, probes, JWT)** → 🟢 `?v=1` is force-merged by the proxy and the upstream status survives to the browser, which is what makes the 409 branch work.
- **X-Sync header selects synchronous vs asynchronous publish** → 🟡 the letter of the decision ("the BFF sets `X-Sync: true` on user-facing mutations") isn't followed on this call site, but the consequence it guards against is verifiably absent here (see the correction above), and the omission predates this PR. Worth one line in the existing code comment rather than a code change.
- **Entity and access-control modeling doctrine** → 🟡 NEW. The decision log has no entry covering *granting access without notifying the grantee*. The combination of a suppressed notification plus an FGA `committee#member` grant is an unlogged cross-cutting pattern; if finding 5 is signed off rather than scoped down, it deserves an ADR.

### Privacy / access note

I inventoried every new fact on the request and rendered payload and traced each to its gate of record. The new path's gate is **identical** to all three sibling member-create paths (member-form dialog, committee bulk add, invite flow) — same endpoint, same `writer` relation on `committee:{uid}` (`charts/lfx-v2-committee-service/templates/ruleset.yaml:346-371`). Nothing is weaker. What differs is that the subject-facing signal is removed, which is the deliberate feature. See finding 5 for the consequences worth signing off on.

<sub>Report-only independent audit — no code was changed and no threads were resolved.</sub>

dealako (commented):
👋 Hi @niravpatel27 — thorough, well-scoped feature. The host-owned per-email state machine (surviving step-panel destruction), the apply-time recipient-count race guard, and the `(committeeUid, email)` dedupe/retry keying are all implemented carefully, and the PR description's trade-off notes made verification fast. I traced the `skip_notification` flow end-to-end and ran independent security + code-quality passes; **no blocking issues, CI is green**. Everything below overlaps existing reviewer threads — consolidating rather than re-posting.

## What I independently verified as correct ✅

- **`skip_notification` flow is safe.** Header value is a hardcoded `'true'` literal, never derived from input; controller narrows via `getStringQueryParam(...) === 'true'`; BFF service defaults to `false` (notify-by-default). No header/CRLF injection, no privilege boundary crossed — suppression only omits the courtesy email for a member the caller is already authorized to create (authz enforced upstream on the bearer token). Not an authz gap.
- **No PII in logs.** `email` is a top-level `SENSITIVE_FIELDS` key, so `logger.sanitize(req.body)` redacts it; the new `skip_notification` log field is a non-sensitive boolean.
- **Fire-and-forget `subscribe()` without `takeUntilDestroyed` is the right call here** — the host outlives the step panel, `.pipe(take(1))` + HTTP completion self-terminates the subscription, and `takeUntilDestroyed` would *wrongly abort* an in-flight POST. The late-callback path (`fetchRecipientCountFor` → `recipientCountTrigger$`) is a no-op once the host's consumer pipe is torn down. No leak.
- **Recipient-count race guard is correct** — the apply-time `uidsEqual(uids, committeeUidsValue())` check on both next/error paths closes the debounce-window race Copilot flagged; add-success self-corrects via a re-fetch. Resolution in be08a478 confirmed.

## Reconciliation with existing reviews

- **@copilot** — recipient-count race (be08a478): **agree, resolved & verified.** Missing-test concern: **agree, still open** (see below).
- **@coderabbitai** — `createCommitteeMember` JSDoc: **agree, resolved.**
- **@luismoriguerra** — I confirm all three of your non-blocking findings; they're the highest-value items remaining (see below).

## Issue summary

- 🔴 Blocking: 0
- 🟡 Minor: 3
  - **Input testid won't resolve via `getByTestId`** — `dataTest=` renders as `[attr.data-test]`, not `data-testid`; every sibling hook uses `data-testid`. A future spec must use `[data-test="..."]` for the input (or standardize). (@luismoriguerra — confirmed)
  - **Typo leaves a permanent red row** — `alice@` clears the input unconditionally and lands a persistent `invalid` entry only fixable by full retype. Control-level `Validators.email` + gating the Add button (matching `account-settings`) makes this the exception, not the normal path. Highest-value polish item. (@luismoriguerra — confirmed)
  - **`extractErrorMessage` can surface Angular's raw transport string** on empty-body failures (`error.message` fallback → "Http failure response for … 0 Unknown Error"), and the friendly `'Could not add. Try again.'` fallback is effectively unreachable since the BFF always populates a body. (@luismoriguerra — confirmed)
- ⚪ Nit: 0
- ❔ Question: 2
  - **No automated coverage** for the async state machine (dedupe/retry, race guards, `skip_notification=true` wire check). Consistent with repo norms, but this is logic-dense. (@copilot — agree)
  - **409 → "already exists"** relies on the upstream contract. Your PR note says this was validated against the Goa design (create-member declares exactly one Conflict) — noting it here so the assumption stays visible; no action needed if that holds.

## Decision

✅ **Approved with minor comments (non-blocking).** Nothing here blocks merge and CI is green. I'd prioritize the control-level email validation before merge (best usability return); the testid and error-copy items are safe to fold into the same follow-up. Submitting as a comment so @luismoriguerra's open thread drives the final call.

<sub>Reviewed with dd-pr-review — parallel security + code-quality passes, reconciled against CodeRabbit / Copilot / Cursor / human reviews.</sub>

apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
```diff
@@ -373,6 +382,55 @@ export class NewsletterManageComponent {
...
+      : { email, committeeUid, status: 'invalid', reason: 'Not a valid email' };
+    this.audienceEmailAdds.update((entries) => [...entries.filter((e) => !(e.committeeUid === committeeUid && e.email === email)), entry]);
+
+    if (entry.status === 'invalid') return;
+
+    // Intentionally NOT takeUntilDestroyed: HttpClient aborts the request on
+    // unsubscribe, which would silently cancel in-flight adds if the user left
+    // the page. createCommitteeMember pipes take(1) and HTTP observables
+    // complete after one response, so each subscription cleans itself up.
+    this.committeeService.createCommitteeMember(committeeUid, { email }, { skipNotification: true }).subscribe({
```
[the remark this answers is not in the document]
niravpatel27: Coverage has been authored: a mocked Playwright spec with five scenarios — the skip_notification=true request contract (URL + normalized body), pending spinner → added checkmark, 409 → benign already-exists, failure reason + retype-to-retry, invalid input with zero network calls, and state/in-flight survival across step navigation. It passes all five on chromium locally and is being stabilized on the firefox/mobile-chrome projects before landing (the mobile nav path differs). Note the pre-existing suite-wide blocker: specs importing the @lfx-one/shared/constants barrel currently fail to load in the Node runner because the barrel now transitively imports @angular/common via dashboard-metrics.constants → utils/form.utils — the new spec works around it with a deep import. Resolving; the spec lands as a follow-up commit.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
niravpatel27: Good catch — agreed on the wrapper's data-test vs data-testid mismatch. The authored spec targets the native input via input[data-test="newsletter-audience-email-add-input"], which works today; the host-level data-testid (account-settings pattern) will land alongside the spec commit. Documented as part of the follow-up; resolving to keep the merge queue moving.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.ts
niravpatel27: Fair point, and thanks for pulling the account-settings precedent. The v1 behavior is a deliberate trade-off for the fire-and-forget flow (instant clear keeps rapid multi-entry frictionless), with the typo cost being a retype — now documented in the PR description under known trade-offs. Control-level validation with a gated Add button is the follow-up candidate, planned together with the e2e spec commit. Resolving as documented.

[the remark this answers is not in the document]
niravpatel27: Acknowledged — same finding as the thread above from luismoriguerra: the instant-clear on invalid input is a documented v1 trade-off (see the PR description's known trade-offs), with control-level validation as the follow-up. Resolving as documented.

apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
niravpatel27: Thanks — and especially for the getHttpErrorDetail warning; that matches what we found in an earlier pass (it reads body.message while this BFF emits the message under error, so a wholesale swap would lose the upstream 4xx text). The status-0/5xx guard forcing the generic fallback is the right targeted fix; it's an inherited trait of the canonical helper shared by all seven call sites, so it's documented under the PR description's known trade-offs and queued for the follow-up commit rather than another iteration here. Resolving as documented.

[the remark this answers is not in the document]
niravpatel27: Acknowledged — same finding as luismoriguerra's thread on this line: the transport-string fallback is an inherited trait of the shared extractErrorMessage helper, documented in the PR description's known trade-offs with the status-0/5xx guard queued as follow-up. Resolving as documented.

apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
niravpatel27: Excellent trace — this is accurate and worth having on the record. The checkmark itself stays truthful (the member IS added to the group; that's the claim the UI makes), but delivery respects the project's unsubscribe list at send time by design, so a previously unsubscribed address won't receive sends and the count won't move. This is now documented in the PR description's known trade-offs. An 'unsubscribed' status would need the newsletter service to surface skipped-as-unsubscribed per address (resolveRecipients only counts it internally today), so it's an upstream-first follow-up. Resolving as documented.

apps/lfx-one/src/server/controllers/committee.controller.ts
niravpatel27: Agreed with the precise framing: no authorization change (writer-gated at Heimdall/FGA, identical to the three sibling add paths, and upstream already exposes the header to any writer) — it's a notice-semantics change. That decision is explicit and signed off: the PR description's Personas section documents that skip_notification suppresses notice to people who never see LFX and that this is the intended internal process for newsletter audiences, decided by the product owner/PR author. The BFF flag intentionally mirrors the upstream contract rather than encoding flow-specific policy; scoping by caller or committee category would drift from the upstream surface. Resolving with the sign-off recorded here and in the description.

apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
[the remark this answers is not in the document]
niravpatel27: Done — canSend now includes !hasPendingAudienceAdds(): a pending add for the currently selected group blocks Send until it settles (seconds), while step navigation stays non-blocking as you suggested. An add resolving after a group switch doesn't block the new audience's send. See commit fe2bc90f.

niravpatel27:
## Review Feedback Addressed (triage: fix critical, document the rest)

Commit: fe2bc90f

### Changes Made
- **newsletter-manage.component.ts**: `canSend` now waits for pending inline adds on the selected group, so an in-flight add can't be silently omitted from an immediate send; step navigation stays non-blocking (per copilot-pull-request-reviewer)

### No Change Needed / Documented as Known Trade-offs (see PR description)
- **Input clears on invalid email** — deliberate v1 fire-and-forget behavior; control-level validation queued as follow-up (per @luismoriguerra, cursor)
- **Transport string on network/5xx failures** — inherited trait of the shared `extractErrorMessage` helper; status-0/5xx guard queued as follow-up (per @luismoriguerra, cursor)
- **Unsubscribed address shows checkmark** — the add genuinely succeeds; delivery respects the project unsubscribe list at send time by design; an `unsubscribed` status needs upstream support first (per @luismoriguerra)
- **skip_notification reachable browser-wide** — no authorization change (writer-gated, mirrors upstream contract); notice-semantics decision documented and signed off in the Personas section (per @luismoriguerra)
- **data-testid on the email input** — the authored spec targets `input[data-test=…]` which works today; host-level testid lands with the spec commit (per @luismoriguerra)
- **Playwright coverage** — five-scenario mocked spec authored (contract + status machine + retry + navigation survival), chromium-green locally, stabilizing on firefox/mobile before landing as a follow-up commit (per copilot-pull-request-reviewer)

### Threads Resolved
9 of 9 unresolved threads addressed in this iteration.

apps/lfx-one/src/app/modules/newsletters/components/newsletter-audience-step/newsletter-audience-step.component.html
```diff
@@ -62,6 +62,69 @@ <h2 class="text-base font-semibold text-gray-900">Who's this for?</h2>
...
+                type="email"
+                [id]="'newsletter-audience-email-add'"
+                size="small"
+                placeholder="name@example.com"
+                dataTest="newsletter-audience-email-add-input">
+              </lfx-input-text>
+            </div>
+            <lfx-button label="Add" type="submit" size="small" severity="secondary" [outlined]="true" data-testid="newsletter-audience-email-add-btn">
+            </lfx-button>
+          </form>
```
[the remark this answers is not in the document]
niravpatel27: Validated against the page's conventions: the learned rule holds on the newsletter list page (row actions are canWrite()-gated), but the manage page this control lives on gates none of its mutation controls on canWrite() — Send, Save as Draft, delete draft, and test send are all reachable under the same newsletterAccessGuard (ED persona fast path OR project.writer), with authorization enforced server-side via Heimdall/OpenFGA writer on the committee. A non-writer ED attempting an add gets a graceful per-email failed row with the upstream reason, same as they'd get attempting Send. Gating only this control would be inconsistent with its own page; canWrite()-gating the whole manage surface is the right scope and is noted as a follow-up candidate. No security exposure — the server enforces regardless. Resolving as documented.

MRashad26 (approved):

PR #1201, opened 2026-07-27
feat(newsletters): native block-composer editor with template selection

dealako:
Restores #1133 (reverted from `main` via #1200 to unblock scheduled releases).

This is a **draft** — stays off the release path until testing completes on this branch. Content is identical to the original #1133 (cherry-picked, authorship preserved).

Refs: LFXV2-2386


dealako:
Not needed. Closing.

PR #1202, opened 2026-07-27
fix(newsletters): remove dead code reintroduced by stale merge base

dealako:
Fixes the release build break in run https://github.com/linuxfoundation/lfx-self-serve/actions/runs/30282716458 (`TS2339: Property 'bodyPersistable' does not exist`).

**Root cause:** PR #1198 ("dashboards: trend indicators") squash-merged from a branch cut before the #1133 revert (#1200) landed on `main`. Its squash diff against a stale merge base reintroduced two orphaned helper methods — `missingDraftRequirements` and `formatMissing` — into `newsletter-manage.component.ts`. Neither is called anywhere; `missingDraftRequirements` references `bodyPersistable()`, a signal that only existed as part of the (now-reverted) block-composer feature from #1133.

**Fix:** deletes both dead methods. Verified `yarn check-types` and `yarn workspace lfx-one-ui build:production` both pass locally.

No other files were affected by the same stale-merge-base issue (checked full diff between the revert commit and current `main` tip).

Refs: LFXV2-2386

### DIFF
--- a/apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/newsletters/newsletter-manage/newsletter-manage.component.ts
@@ -441,25 +441,6 @@ export class NewsletterManageComponent {
     });
   }
 
-  /** The still-missing requirements that block a draft save, for user feedback. */
-  private missingDraftRequirements(): string[] {
-    const missing: string[] = [];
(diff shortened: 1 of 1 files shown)


MRashad26 (approved):
Code standards review: all clean. Pure dead-code removal — two unreachable private methods deleted, no logic changes, no new patterns introduced. Approving.

PR #1206, opened 2026-07-27
fix(committees): skip FGA pre-check for LFID invite accept flow

andrest50:
## Summary

- The BFF's `acceptCommitteeInvite` handler called `getPendingInviteForUser` unconditionally before forwarding to the committee-service. This is a FGA-gated query-service lookup that races against async invitee tuple propagation.
- In the LFID invite flow the tuple is published by NATS after LFID accept and may not have propagated by the time the user submits the org dialog, causing a 400 before the committee-service is ever reached.
- Added a `from_lfid_invite` flag to `AcceptCommitteeInviteRequest`. When set, the BFF skips `getPendingInviteForUser` and forwards directly to the committee-service, which is authoritative for all invite validation. All other call paths (My Groups, etc.) retain the full pre-check unchanged.

## Ticket

[LFXV2-2677](https://linuxfoundation.atlassian.net/browse/LFXV2-2677)

## Test plan

- [ ] Accept an LFID committee invite for a committee with org required — org dialog appears, fill in org, click Join Group — confirm 204 and redirect to return URL
- [ ] Accept a committee invite from My Groups — confirm the existing `getPendingInviteForUser` pre-check still runs (no regression)
- [ ] Send `POST /api/committees/{id}/invites/{inviteId}/accept` without `from_lfid_invite` and with an email that has no FGA tuple — confirm 400 "No matching pending invite found" is still returned

[LFXV2-2677]: https://linuxfoundation.atlassian.net/browse/LFXV2-2677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
+++ b/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
@@ -57,7 +57,10 @@ export class InvitationAcceptFlowService {
         if (!result?.organization) {
           return EMPTY;
         }
-        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, { organization: result.organization, fromLfidInvite: context.fromLfidInvite });
+        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, {
+          organization: result.organization,
+          fromLfidInvite: context.fromLfidInvite,
+        });
       })
     );
   }
--- a/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
+++ b/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
@@ -37,7 +37,7 @@ export class InvitationAcceptFlowService {
     const requiresOrganization = invitationRequiresOrganization(context);
 
     if (!requiresOrganization) {
-      return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, undefined, context.fromLfidInvite);
+      return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, { fromLfidInvite: context.fromLfidInvite });
     }
 
     // Resolve the pre-fill org: use the invite's org when present, otherwise fall back to
@@ -57,7 +57,7 @@ export class InvitationAcceptFlowService {
         if (!result?.organization) {
           return EMPTY;
         }
-        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, result.organization, context.fromLfidInvite);
+        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, { organization: result.organization, fromLfidInvite: context.fromLfidInvite });
       })
     );
   }
--- a/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
+++ b/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
@@ -52,16 +52,19 @@ export class InvitationAcceptFlowService {
 
     return contextReady$.pipe(
       switchMap((ctx) => this.preResolveOrganization(ctx)),
-      switchMap((ctx) => from(this.openOrganizationDialog(ctx))),
-      switchMap((result) => {
-        if (!result?.organization) {
-          return EMPTY;
-        }
-        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, {
-          organization: result.organization,
-          fromLfidInvite: context.fromLfidInvite,
-        });
-      })
+      switchMap((ctx) =>
+        from(this.openOrganizationDialog(ctx)).pipe(
+          switchMap((result) => {
+            if (!result?.organization) {
+              return EMPTY;
+            }
+            return this.invitationService.acceptInvitation(ctx.committeeUid, ctx.inviteUid, {
+              organization: result.organization,
+              fromLfidInvite: ctx.fromLfidInvite,
+            });
+          })
+        )
+      )
     );
   }
 
--- a/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
+++ b/apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
@@ -37,7 +37,7 @@ export class InvitationAcceptFlowService {
     const requiresOrganization = invitationRequiresOrganization(context);
 
     if (!requiresOrganization) {
-      return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid);
+      return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, undefined, context.fromLfidInvite);
     }
 
     // Resolve the pre-fill org: use the invite's org when present, otherwise fall back to
@@ -57,7 +57,7 @@ export class InvitationAcceptFlowService {
         if (!result?.organization) {
           return EMPTY;
         }
-        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, result.organization);
+        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, result.organization, context.fromLfidInvite);
       })
     );
   }
--- a/apps/lfx-one/src/app/shared/services/invitation.service.ts
+++ b/apps/lfx-one/src/app/shared/services/invitation.service.ts
@@ -3,7 +3,7 @@
 
 import { HttpClient } from '@angular/common/http';
 import { inject, Injectable, signal, WritableSignal } from '@angular/core';
-import { PendingInvitation, AcceptCommitteeInviteRequest, CommitteeOrganizationReference } from '@lfx-one/shared/interfaces';
+import { PendingInvitation, AcceptCommitteeInviteRequest, AcceptInvitationOptions } from '@lfx-one/shared/interfaces';
 import { catchError, finalize, Observable, of, shareReplay, tap } from 'rxjs';
 
 /**
@@ -69,13 +69,9 @@ export class InvitationService {
   }
 
   /** Accepts an invitation. Upstream is invitee-authenticated; returns 204. */
-  public acceptInvitation(
-    committeeUid: string,
-    inviteUid: string,
-    options?: { organization?: CommitteeOrganizationReference; fromLfidInvite?: boolean }
-  ): Observable<void> {
+  public acceptInvitation(committeeUid: string, inviteUid: string, options?: AcceptInvitationOptions): Observable<void> {
     const body: AcceptCommitteeInviteRequest = options?.organization ? { organization: options.organization } : {};
-    if (options?.fromLfidInvite) {
+    if (options?.fromLfidInvite === true) {
       body.from_lfid_invite = true;
     }
     return this.http.post<void>(`/api/committees/${encodeURIComponent(committeeUid)}/invites/${encodeURIComponent(inviteUid)}/accept`, body);
--- a/apps/lfx-one/src/app/shared/services/invitation.service.ts
+++ b/apps/lfx-one/src/app/shared/services/invitation.service.ts
@@ -69,9 +69,9 @@ export class InvitationService {
   }
 
   /** Accepts an invitation. Upstream is invitee-authenticated; returns 204. */
-  public acceptInvitation(committeeUid: string, inviteUid: string, organization?: CommitteeOrganizationReference, fromLfidInvite?: boolean): Observable<void> {
-    const body: AcceptCommitteeInviteRequest = organization ? { organization } : {};
-    if (fromLfidInvite) {
+  public acceptInvitation(committeeUid: string, inviteUid: string, options?: { organization?: CommitteeOrganizationReference; fromLfidInvite?: boolean }): Observable<void> {
+    const body: AcceptCommitteeInviteRequest = options?.organization ? { organization: options.organization } : {};
+    if (options?.fromLfidInvite) {
       body.from_lfid_invite = true;
     }
     return this.http.post<void>(`/api/committees/${encodeURIComponent(committeeUid)}/invites/${encodeURIComponent(inviteUid)}/accept`, body);
--- a/apps/lfx-one/src/app/shared/services/invitation.service.ts
+++ b/apps/lfx-one/src/app/shared/services/invitation.service.ts
@@ -69,7 +69,11 @@ export class InvitationService {
   }
 
   /** Accepts an invitation. Upstream is invitee-authenticated; returns 204. */
-  public acceptInvitation(committeeUid: string, inviteUid: string, options?: { organization?: CommitteeOrganizationReference; fromLfidInvite?: boolean }): Observable<void> {
+  public acceptInvitation(
+    committeeUid: string,
+    inviteUid: string,
+    options?: { organization?: CommitteeOrganizationReference; fromLfidInvite?: boolean }
+  ): Observable<void> {
     const body: AcceptCommitteeInviteRequest = options?.organization ? { organization: options.organization } : {};
     if (options?.fromLfidInvite) {
       body.from_lfid_invite = true;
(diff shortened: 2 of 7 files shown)


apps/lfx-one/src/server/controllers/committee.controller.ts
```diff
@@ -753,18 +725,50 @@ export class CommitteeController {
...
-            service: 'committee_controller',
-            path: req.path,
-          })
-        );
-        return;
+      // Skip the FGA-gated pending-invite lookup when the accept originates from the LFID invite
+      // flow — the invite UID was already known from the JWT, but the FGA invitee tuple may not
+      // have propagated yet. The committee-service is authoritative for invite existence and org
+      // requirements in all cases (LFXV2-2453).
+      if (!body.from_lfid_invite) {
```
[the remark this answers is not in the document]
andrest50: Fixed — changed `if (!body.from_lfid_invite)` to `if (body.from_lfid_invite !== true)` so only the literal boolean `true` enables the LFID path. A truthy string, object, or any other value now falls through to the standard pre-check.

See commit aeb93089.

andrest50:
## Review Feedback Addressed

Commit: aeb93089

### Changes Made
- **`committee.controller.ts:732`**: Tightened `from_lfid_invite` guard to strict `=== true` — only the literal boolean enables the LFID path; truthy strings or objects fall through to the standard pre-check. (per copilot-pull-request-reviewer)
- **`invitation.service.ts:72`**: Refactored `acceptInvitation` from 4 positional params to an options object `{ organization?, fromLfidInvite? }`, eliminating the `undefined` placeholder at call sites. (per coderabbitai[bot])
- **`invitation-accept-flow.service.ts`**: Updated both call sites to pass named options.

### Threads Resolved

apps/lfx-one/src/server/controllers/committee.controller.ts
```diff
@@ -753,18 +725,50 @@ export class CommitteeController {
...
-            service: 'committee_controller',
-            path: req.path,
-          })
-        );
-        return;
+      // Skip the FGA-gated pending-invite lookup when the accept originates from the LFID invite
+      // flow — the invite UID was already known from the JWT, but the FGA invitee tuple may not
+      // have propagated yet. The committee-service is authoritative for invite existence and org
+      // requirements in all cases (LFXV2-2453).
+      if (body.from_lfid_invite !== true) {
```
audigregorie, line 736: **Major — Client-controlled bypass of BFF pre-check is not server-verified**

The `from_lfid_invite` flag is a client-supplied boolean. Any client — not just the genuine LFID invite flow — can set `from_lfid_invite: true` to skip both the FGA-gated `getPendingInviteForUser` lookup AND the BFF-side `organization_required` enforcement. The PR's threat model rests on "the committee-service is authoritative for invite existence and org requirements in all cases," which is sound for *unauthorized* accepts (upstream will reject), but the BFF pre-check was explicitly added under LFXV2-2453 as a defense layer. A client-supplied flag widens the bypass surface beyond the LFID flow without server-side verification that the request actually originated from an LFID invite JWT.

If the LFID flow is server-distinguishable (e.g., the invite JWT is present in the session, or a server-side flag was set during the LFID accept), gate on that instead. If it must remain a client flag, document explicitly in the controller that the bypass is safe only because upstream is authoritative, and add a test asserting a non-LFID client setting `from_lfid_invite: true` on an invalid invite still receives a 400 from upstream.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -57,7 +57,10 @@ export class InvitationAcceptFlowService {
         if (!result?.organization) {
           return EMPTY;
         }
-        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, result.organization);
+        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, {
+          organization: result.organization,
+          fromLfidInvite: context.fromLfidInvite,
+        });
```
audigregorie: **Minor — `context` vs `ctx` inconsistency in the org-resolved path**

The org-resolved `acceptInvitation` call reads `fromLfidInvite: context.fromLfidInvite` (outer `context`) while the surrounding chain uses `ctx` (the pre-resolved context). Functionally equivalent since `fromLfidInvite` is never mutated during pre-resolution, but stylistically inconsistent with the `...ctx` spread pattern used elsewhere in the same operator chain.

```suggestion
        return this.invitationService.acceptInvitation(context.committeeUid, context.inviteUid, {
          organization: result.organization,
          fromLfidInvite: ctx.fromLfidInvite,
        });
```

packages/shared/src/interfaces/committee.interface.ts
```diff
@@ -195,6 +195,12 @@ export interface CreateCommitteeInviteRequest {
 export interface AcceptCommitteeInviteRequest {
   /** Organization the invitee confirms on acceptance */
   organization?: CommitteeOrganizationReference | null;
+  /**
+   * Set to true when the accept originates from the LFID invite flow where the invite UID was
+   * already known from the JWT — signals the BFF to skip the FGA-gated pending-invite pre-check,
+   * which may race against async FGA tuple propagation.
+   */
+  from_lfid_invite?: boolean;
```
audigregorie, line 209: **Minor — `from_lfid_invite` in the upstream payload type is never forwarded**

`AcceptCommitteeInviteRequest` is used for both the client→BFF body (`req.body`) and the BFF→upstream payload (`acceptData`). `from_lfid_invite` is only meaningful for client→BFF; the BFF allowlist ("never forward unknown client-supplied fields upstream") correctly omits it from `acceptData`. Using one type for both roles is slightly leaky — a future maintainer might assume `from_lfid_invite` reaches the committee-service. The JSDoc helps; optionally split into `AcceptCommitteeInviteClientRequest` vs. `AcceptCommitteeInviteUpstreamPayload` to make the boundary explicit.

audigregorie (commented):
## Code Review Summary

Sound fix for the FGA tuple propagation race in the LFID invite accept flow. The allowlist pattern for `acceptData` is correctly preserved (the new `from_lfid_invite` flag is NOT forwarded upstream), and the non-LFID call paths (My Groups, pending-actions, committee-view, committee-invitations) all leave the flag undefined so the full pre-check still runs — no regression. Two Majors below are about trust-boundary hygiene and test coverage; neither blocks merge if upstream is genuinely authoritative, but both should be addressed or documented as trade-offs.

### Major — outside the diff

- **No automated tests for the new security-relevant branch.** The controller has no spec file and the PR adds no tests for the `from_lfid_invite` branching; the test plan is manual-only. The two paths (pre-check skipped vs. run) are security-relevant and should have automated regression coverage: (a) `from_lfid_invite: true` skips `getPendingInviteForUser` and forwards to the service; (b) absent flag runs the pre-check and returns 400 when no pending invite; (c) absent flag with `organization_required: true` and no org returns 400.

### Minor — outside the diff

- **`AcceptCommitteeInviteRequest` is shared between client→BFF body and BFF→upstream `acceptData`.** `from_lfid_invite` is only meaningful for client→BFF; the allowlist correctly omits it from `acceptData`, but the shared type is slightly leaky. The JSDoc clarifies intent — leave as-is or split into client-vs-upstream request types if the codebase has such a pattern.

### What's done well

- The `acceptData` allowlist ("never forward unknown client-supplied fields upstream") is correctly preserved — `from_lfid_invite` is BFF-internal only.
- All four `invitationAcceptFlow.accept()` callers other than the LFID flow leave `fromLfidInvite` undefined, so the controller's `!== true` check retains the full pre-check for them.
- Interface JSDoc on both `from_lfid_invite` and `fromLfidInvite` clearly explains the race rationale and links LFXV2-2453.

Rating: ⭐⭐⭐⭐ (4/5)

apps/lfx-one/src/app/shared/services/invitation.service.ts
```diff
@@ -69,8 +69,15 @@ export class InvitationService {
   }
 
   /** Accepts an invitation. Upstream is invitee-authenticated; returns 204. */
-  public acceptInvitation(committeeUid: string, inviteUid: string, organization?: CommitteeOrganizationReference): Observable<void> {
-    const body: AcceptCommitteeInviteRequest = organization ? { organization } : {};
+  public acceptInvitation(
```
jordane: **Important** — Parameter signature refactor without test coverage.

`acceptInvitation` changed from positional params to an options object (`{ organization?, fromLfidInvite? }`). Both call sites in `invitation-accept-flow.service.ts` were updated correctly, but there's no test coverage for the new signature or its branches (with/without `organization`, with/without `fromLfidInvite`).

_Fix:_ add unit tests covering both branches.

apps/lfx-one/src/server/controllers/committee.controller.ts
```diff
@@ -753,18 +725,50 @@ export class CommitteeController {
...
+              path: req.path,
+            })
+          );
+          return;
+        }
+
+        // Only enforce org as required when organization_required is explicitly true. When it is
+        // null/undefined (invite pre-dates committee-service v1.1), skip the mandatory check and
+        // let the upstream accept endpoint be authoritative.
+        if (pendingInvite.organization_required === true && !hasOrg) {
```
jordane, line 766: **Important** — Organization-requirement validation delegated without intermediate validation.

When `from_lfid_invite === true`, this BFF-side check of `organization_required` is skipped entirely, relying on committee-service to reject accepts missing a required org. Intentional per the surrounding comment, but it's a behavioral change from the non-LFID path.

_Fix:_ consider adding a comment clarifying that this BFF-side org check is an optimization, not a security boundary, so future readers don't assume the BFF enforces it here.

jordane (commented):
## Review

This PR fixes a real FGA-propagation race in the LFID committee-invite accept flow by adding a `from_lfid_invite` flag that lets the BFF skip its FGA-gated pending-invite pre-check and defer to committee-service (which is authoritative for invite existence and org requirements). Non-LFID paths (My Groups, etc.) are unchanged.

Both prior review comments (Copilot's `from_lfid_invite` truthy-bypass concern, CodeRabbit's options-object nitpick) have been verified fixed in the current code.

Two inline notes below (test coverage on the new `acceptInvitation` signature, and a suggested clarifying comment on the org-check delegation). Nothing here blocks merge; posting as informational.

apps/lfx-one/src/server/controllers/committee.controller.ts
andrest50: Done — the BFF is stateless so gating on session state from the LFID accept request isn't feasible without adding cross-request storage. Went with the documentation + test path instead:

- Expanded the comment at the flag check to explicitly state the threat model: upstream is the authoritative security gate; an unauthorized accept (wrong invitee, nonexistent invite, missing org) is still rejected by committee-service regardless of this flag. The BFF pre-check is a defense-in-depth optimization, not the primary gate.
- Added `committee.controller.spec.ts` with tests covering: LFID path skips pre-check and calls upstream, upstream error propagates via `next` (demonstrating committee-service is the real gate), and a non-boolean truthy value (e.g., string `'true'`) falls through to the standard check.

See commit a469828e.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
andrest50: Fixed — nested the result `switchMap` inside the `ctx` switchMap so `ctx` is in scope throughout the full chain. All three references (`committeeUid`, `inviteUid`, `fromLfidInvite`) now consistently read from `ctx`.

See commit a469828e.

packages/shared/src/interfaces/committee.interface.ts
andrest50: Rewrote the JSDoc to make the BFF-only boundary explicit: states that the field is consumed by the BFF and never forwarded upstream, and that committee-service remains the authoritative security boundary regardless of the flag. Kept as a single type — the allowlist pattern in the controller already enforces the boundary at runtime, so splitting types would add ceremony without adding safety.

See commit a469828e.

apps/lfx-one/src/app/shared/services/invitation.service.ts
andrest50: The project's vitest config is scoped to `src/server/**/*.spec.ts` only (no Angular test builder wired up — see `apps/lfx-one/vitest.config.ts`), so a unit spec for the Angular-injected `InvitationService` isn't currently supported in the test infrastructure.

The meaningful branch — that `from_lfid_invite: true` bypasses the pre-check and upstream errors still propagate — is covered by the new `committee.controller.spec.ts` at the BFF level. The `acceptInvitation` body construction is straightforward enough that a dedicated Angular unit test would mainly duplicate that coverage. Happy to add a server-side or e2e test if you'd prefer a different approach.

See commit a469828e.

apps/lfx-one/src/server/controllers/committee.controller.ts
andrest50: Added a clarifying comment: the BFF-side org check is explicitly noted as an optimization for early rejection, not the security boundary — committee-service enforces org requirements independently and is authoritative.

See commit a469828e.

andrest50:
## Review Feedback Addressed

Commit: a469828e

### Changes Made
- **`committee.controller.ts`**: Expanded the `from_lfid_invite` guard comment to document the threat model explicitly — upstream is the authoritative security gate; the BFF pre-check is a defense-in-depth optimization only. Added a clarifying comment at the org-required guard noting it is an optimization, not a security boundary. (per audigregorie, jordane)
- **`committee.controller.spec.ts`** (new): Added tests covering the LFID path — pre-check is skipped, upstream error propagates via `next`, and a non-boolean truthy value (e.g., string `'true'`) falls through to the standard check. (per audigregorie)
- **`invitation-accept-flow.service.ts`**: Nested the result `switchMap` inside the `ctx` switchMap so `ctx` is in scope and used consistently throughout the chain. (per audigregorie)
- **`committee.interface.ts`**: Rewrote `from_lfid_invite` JSDoc to make the BFF-only boundary explicit — consumed by the BFF, never forwarded upstream; committee-service remains the authoritative security boundary. (per audigregorie)

### Still Open
- **`invitation.service.ts:72`** (jordane): Angular service unit tests aren't currently supported by the project's vitest config (`src/server/**` only, no Angular test builder). Coverage of the meaningful branch is provided by the new controller spec. Explained in thread.

### Threads Resolved
4 of 5 unresolved threads resolved in this iteration.

apps/lfx-one/src/app/shared/services/invitation.service.ts
```diff
@@ -69,8 +69,15 @@ export class InvitationService {
...
+    committeeUid: string,
+    inviteUid: string,
+    options?: { organization?: CommitteeOrganizationReference; fromLfidInvite?: boolean }
```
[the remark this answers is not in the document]
andrest50: Fixed — see commit 250babff.

apps/lfx-one/src/app/shared/services/invitation.service.ts
```diff
@@ -69,8 +69,15 @@ export class InvitationService {
...
+  ): Observable<void> {
+    const body: AcceptCommitteeInviteRequest = options?.organization ? { organization: options.organization } : {};
+    if (options?.fromLfidInvite) {
+      body.from_lfid_invite = true;
+    }
```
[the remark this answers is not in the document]
andrest50: Fixed — see commit 250babff.

apps/lfx-one/src/app/shared/services/invitation.service.ts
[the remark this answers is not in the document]
andrest50: Fixed — extracted `AcceptInvitationOptions` as a named interface in `@lfx-one/shared/interfaces/committee.interface.ts` and updated `invitation.service.ts` to import it from there. No inline object shape remains under `apps/lfx-one`.

See commit 250babff.

apps/lfx-one/src/app/shared/services/invitation.service.ts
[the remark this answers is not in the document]
andrest50: Fixed — changed to `options?.fromLfidInvite === true` to match the strict boolean contract enforced on the BFF side.

See commit 250babff.

audigregorie (approved):

PR #1208, opened 2026-07-27
fix(profile): add support link to already-claimed identity error

fayazg:
## Summary

- When adding or verifying an email identity that's already claimed by another
  LFID account, the profile dialogs showed only red error text with no way to
  get help. This adds a "contact support" link beside that error so users can
  reach out.
- The link appears **only** for the already-claimed conflict (detected via the
  shared `isIdentityAlreadyLinkedError` util, the same marker set the backend
  uses) — not for other errors like a wrong/expired OTP.
- Reuses the existing `lfxOpenIntercom` directive (opens Intercom, falls back to
  the Jira support desk), matching the existing social-connect banner's pattern.
- Applies to both the add-account and verify-identity dialogs.

## Note

The `showSupportLink` computed and the support-link template snippet are
intentionally duplicated across the two dialogs rather than extracted into a
shared component — the snippet is small and stable, and a shared abstraction
would couple two otherwise-independent dialogs for little gain.

Fixes LFXV2-2858

🤖 Generated with Claude Code

### DIFF
--- a/apps/lfx-one/e2e/profile-identities-support-link.spec.ts
+++ b/apps/lfx-one/e2e/profile-identities-support-link.spec.ts
@@ -148,7 +148,10 @@ test.describe('Identity dialogs — already-claimed support link', () => {
     });
 
     test('hides the support link for an unrelated error', async ({ page }) => {
-      await mockSendCode(page, 400, UNRELATED_ERROR_MESSAGE);
+      // Same 409 status as the positive case — only the message differs. This isolates the
+      // message as the discriminator: a status-based implementation (show for any 409) would
+      // wrongly pass here, so keeping the status constant proves message-based gating.
+      await mockSendCode(page, 409, UNRELATED_ERROR_MESSAGE);
       await openEmailConnectStep(page);
       await page.getByTestId('add-account-send-code').locator('button').click();
 
@@ -179,7 +182,10 @@ test.describe('Identity dialogs — already-claimed support link', () => {
     });
 
     test('hides the support link for an unrelated error', async ({ page }) => {
-      await mockSendCode(page, 400, UNRELATED_ERROR_MESSAGE);
+      // Same 409 status as the positive case — only the message differs. This isolates the
+      // message as the discriminator: a status-based implementation (show for any 409) would
+      // wrongly pass here, so keeping the status constant proves message-based gating.
+      await mockSendCode(page, 409, UNRELATED_ERROR_MESSAGE);
       await openVerifyEmailDialog(page);
       await page.getByTestId('verify-send-code').locator('button').click();
 
--- a/apps/lfx-one/e2e/profile-identities-support-link.spec.ts
+++ b/apps/lfx-one/e2e/profile-identities-support-link.spec.ts
@@ -1,0 +1,192 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+/**
+ * Profile identity dialogs — "contact support" link on already-claimed errors (LFXV2-2858).
+ *
+ * When adding or verifying an email identity that is already claimed by another LFID
+ * account, the add-account and verify-identity dialogs surface a "contact support"
+ * action beside the red error. The link is gated on the error *message* (via the shared
+ * isIdentityAlreadyLinkedError util, the same marker set the backend uses), NOT on the
+ * HTTP status — so these tests drive the send-code error path and vary only the message:
+ *   - already-linked message  → support link appears
+ *   - any other error message  → support link is absent
+ *
+ * Both the identity list and the email verification endpoint are mocked, so the spec is
+ * self-contained and does not depend on a seeded backend test user (unlike the sibling
+ * profile-identities-verify spec, which reads live data).
+ *
+ * Prerequisites:
+ *   - Dev server reachable at the Playwright baseURL (default http://localhost:4200)
+ *   - apps/lfx-one/.env populated with TEST_USERNAME / TEST_PASSWORD (see global-setup.ts)
+ */
+
+import { expect, Page, test } from '@playwright/test';
+
+// Mirrors EMAIL_ALREADY_LINKED_MESSAGE in @lfx-one/shared/constants. Inlined rather than
+// imported because the shared constants barrel transitively pulls Angular runtime code,
+// which the Playwright Node test loader can't compile. The only invariant this spec relies
+// on is that the message contains an already-linked marker (matched by isIdentityAlreadyLinkedError).
+const EMAIL_ALREADY_LINKED_MESSAGE = 'This email is already linked to another account';
+
+const IDENTITIES_ROUTE = '**/api/profile/identities';
+const SEND_CODE_ROUTE = '**/api/profile/identities/email/send-code';
+const UNRELATED_ERROR_MESSAGE = 'Invalid or expired verification code. Please try again.';
+const SUPPORT_BUTTON = { name: 'contact support' } as const;
+
+// The unverified email identity the mocked list exposes; drives the verify dialog's
+// verify-btn-<id> and OTP flow.
+const UNVERIFIED_EMAIL_ID = 'idf-email-unverified';
+
+// Deterministic identities payload so the page renders without depending on a seeded
+// backend test user (the sibling profile-identities-verify spec relies on live data and
+// can't run outside that environment — this spec mocks it to stay self-contained).
+async function mockIdentities(page: Page): Promise<void> {
+  await page.route(IDENTITIES_ROUTE, (route) => {
+    if (route.request().method() !== 'GET') {
+      return route.fallback();
+    }
+    return route.fulfill({
+      status: 200,
+      contentType: 'application/json',
+      body: JSON.stringify([
+        {
+          id: UNVERIFIED_EMAIL_ID,
+          platform: 'email',
+          type: 'email',
+          value: 'jdoe@company.org',
+          verified: false,
+          verifiedBy: null,
+          source: 'e2e',
+          icon: 'fa-light fa-envelope',
+          createdAt: '2024-01-01T00:00:00.000Z',
+          updatedAt: '2024-01-01T00:00:00.000Z',
+          displayState: 'unverified',
+          inAuth0: false,
+        },
+      ]),
+    });
+  });
+}
+
+// Neutralize the Osano consent overlay, which otherwise intercepts pointer events on the
+// dialog's provider buttons. Registered via addInitScript so it applies on every navigation
+// before page scripts (mirrors profile-edit-drawer / me-profile-nav specs).
+async function suppressCookieBanner(page: Page): Promise<void> {
+  await page.addInitScript(() => {
+    const hide = (): void => {
+      if (document.getElementById('e2e-hide-osano')) {
+        return;
+      }
+      const style = document.createElement('style');
+      style.id = 'e2e-hide-osano';
+      style.textContent = '.osano-cm-window { display: none !important; pointer-events: none !important; }';
+      (document.head ?? document.documentElement).appendChild(style);
+    };
+    if (document.readyState === 'loading') {
+      document.addEventListener('DOMContentLoaded', hide);
+    } else {
+      hide();
+    }
+  });
+}
+
+function skipWhenAuthMissing(page: Page): void {
+  try {
+    const { hostname } = new URL(page.url());
+    if (hostname === 'auth0.com' || hostname.endsWith('.auth0.com')) {
+      test.skip(true, 'TEST_USERNAME / TEST_PASSWORD not configured — see global-setup.ts');
+    }
+  } catch {
+    // Malformed URL — keep the test running; a failure here is useful signal, not noise.
+  }
+}
+
+// Stub the send-code endpoint with a chosen status + message. Later registrations win,
+// so a test can call this to override the beforeEach default.
+async function mockSendCode(page: Page, status: number, message: string): Promise<void> {
+  await page.route(SEND_CODE_ROUTE, (route) => {
+    if (route.request().method() !== 'POST') {
+      return route.fallback();
+    }
+    return route.fulfill({
+      status,
+      contentType: 'application/json',
+      body: JSON.stringify({ success: false, error: 'send_code_failed', message }),
+    });
+  });
+}
+
+test.describe('Identity dialogs — already-claimed support link', () => {
+  test.beforeEach(async ({ page }) => {
+    await suppressCookieBanner(page);
+    await mockIdentities(page);
+    await page.goto('/profile/identities', { waitUntil: 'domcontentloaded' });
+    skipWhenAuthMissing(page);
+    await expect(page).not.toHaveURL(/auth0\.com/);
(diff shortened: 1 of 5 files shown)


apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.html
```diff
@@ -86,14 +86,26 @@
             styleClass="w-full gap-3 [&_.p-inputtext]:flex-1"
             data-testid="add-account-otp" />
           @if (verificationError()) {
-            <p class="text-center text-sm text-red-600" role="alert" data-testid="add-account-otp-error">{{ verificationError() }}</p>
+            <p class="text-center text-sm text-red-600" role="alert" data-testid="add-account-otp-error">
+              {{ verificationError() }}{{ showSupportLink() ? '.' : '' }}
+              @if (showSupportLink()) {
+                If you believe this is an error,
+                <button type="button" lfxOpenIntercom class="font-medium underline">contact support</button>.
+              }
```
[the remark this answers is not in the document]
fayazg: Thanks for flagging this — the interaction with Angular's whitespace handling is worth checking, but the rendered output is actually correct here.

The two interpolations are adjacent (`{{ verificationError() }}{{ showSupportLink() ? '.' : '' }}`), so there's no text node between them and no space — that produces `…account.`. The whitespace-only node the comment is concerned about is the one *between the `'.'` interpolation and the `@if` block*, and that node is preserved and collapsed to a single space, so the final render is `…account. If you believe this is an error, contact support.`

This is verifiable from the pre-fix behavior: before this change the template was `{{ verificationError() }}` + newline + the same `@if` block, and it rendered as `…account If you believe…` — a single space, no period. Adding the adjacent `'.'` interpolation only inserts the period; it doesn't remove that following space. So the period lands directly against `account`, and the space before `If` is unchanged.

No code change needed. Happy to revisit if you're seeing something different in a rendered build.

apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.ts
```diff
@@ -38,6 +40,10 @@ export class AddAccountDialogComponent {
   public codeSent = signal(false);
   public verificationCode = signal('');
   public verificationError = signal('');
+  // Shows the "contact support" link whenever the visible error is an already-linked
+  // conflict — detected from the message text via the same shared util the backend uses,
+  // so it works regardless of which subscribe callback set the error.
+  public readonly showSupportLink = computed(() => isIdentityAlreadyLinkedError(this.verificationError()));
```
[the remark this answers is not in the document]
fayazg: Added E2E coverage in commit 39a52d73c — `apps/lfx-one/e2e/profile-identities-support-link.spec.ts`. It drives the send-code error path for both the add-account and verify-identity dialogs and asserts the conflict-only contract: the already-linked message renders the `contact support` link, while an unrelated error (e.g. invalid/expired code) does not. The spec is self-contained — it mocks `/api/profile/identities` and the email send-code endpoint so it runs without a seeded backend user.

fayazg:
## Review Feedback Addressed

Commit: 39a52d73c

### Tests Added
- **apps/lfx-one/e2e/profile-identities-support-link.spec.ts**: E2E coverage for the conflict-only `contact support` link on the add-account and verify-identity dialogs (flagged by copilot-pull-request-reviewer). Asserts the already-linked message shows the link and an unrelated error does not, for both dialogs. Self-contained — mocks `/api/profile/identities` and the email send-code endpoint, and suppresses the Osano consent overlay so provider buttons are clickable.

### Threads Resolved

apps/lfx-one/e2e/profile-identities-support-link.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+      await page.getByTestId('add-account-send-code').locator('button').click();
+
+      const error = page.getByTestId('add-account-send-error');
+      await expect(error).toBeVisible({ timeout: 5000 });
+      await expect(error).toContainText(EMAIL_ALREADY_LINKED_MESSAGE);
+      await expect(error.getByRole('button', SUPPORT_BUTTON)).toBeVisible();
+    });
+
+    test('hides the support link for an unrelated error', async ({ page }) => {
+      await mockSendCode(page, 400, UNRELATED_ERROR_MESSAGE);
```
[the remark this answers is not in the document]
fayazg: Good call — the negative cases were also flipping the status, so a status-based implementation could have slipped through. Both negative tests now hold the status at 409 and vary only the message, so they prove the shared message matcher (`isIdentityAlreadyLinkedError`) is what gates the support link. All 4 tests still pass.

See commit 0a41dead3: `profile-identities-support-link.spec.ts` — negative cases changed from 400 to a constant 409.

fayazg:
## Review Feedback Addressed

Commit: 0a41dead3

### Changes Made
- **apps/lfx-one/e2e/profile-identities-support-link.spec.ts**: hold the HTTP status at 409 in both negative cases so only the error message varies, isolating the message as the discriminator that gates the support link — a status-based implementation would otherwise wrongly pass (per coderabbitai). All 4 tests pass.

### Threads Resolved

apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.html
```diff
@@ -86,14 +86,26 @@
             styleClass="w-full gap-3 [&_.p-inputtext]:flex-1"
             data-testid="add-account-otp" />
           @if (verificationError()) {
-            <p class="text-center text-sm text-red-600" role="alert" data-testid="add-account-otp-error">{{ verificationError() }}</p>
+            <p class="text-center text-sm text-red-600" role="alert" data-testid="add-account-otp-error">
+              {{ verificationError() }}{{ showSupportLink() ? '.' : '' }}
```
audigregorie: **Major — missing space between error message and "If you believe"**

Angular's default `preserveWhitespaces: false` removes the whitespace-only text node between `{{ }}` and the `@if` block, so this renders as `This email is already linked to another account.If you believe this is an error, contact support.` — no space between `account.` and `If`. The codebase convention is to place the separator *inside* the `@if` content (e.g. `· ` in `newsletter-review.component.html`, `<span> Source: </span>` in `org-project-detail.component.html`). The e2e `toContainText(EMAIL_ALREADY_LINKED_MESSAGE)` only checks the message substring, so it doesn't catch this.

Moving the space into the interpolated string keeps it part of the text node (preserved):

```suggestion
              {{ verificationError() }}{{ showSupportLink() ? '. ' : '' }}
```

Apply the same fix at the three other sites (send-error block below at ~line 103, and both `verify-email-error` blocks in `verify-identity-dialog.component.html`).

apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.html
```diff
@@ -86,14 +86,26 @@
...
+            </p>
           }
         </div>
       }
     </div>
 
     @if (!codeSent() && verificationError()) {
-      <p class="text-sm text-red-600" role="alert" data-testid="add-account-send-error">{{ verificationError() }}</p>
+      <p class="text-sm text-red-600" role="alert" data-testid="add-account-send-error">
+        {{ verificationError() }}{{ showSupportLink() ? '.' : '' }}
```
audigregorie: **Major — same missing-space issue as the OTP block above**

This `add-account-send-error` block has the identical `{{ verificationError() }}{{ showSupportLink() ? '.' : '' }}` + `@if` pattern and renders the same run-on (`account.If you believe…`).

```suggestion
        {{ verificationError() }}{{ showSupportLink() ? '. ' : '' }}
```

audigregorie (commented):
## Code Review Summary

Well-scoped, well-tested change — gating the support link on the shared `isIdentityAlreadyLinkedError` message check (not the HTTP status) is the right call, and the negative e2e test that holds the 409 constant and varies only the message is a strong test of that gating. Reusing `lfxOpenIntercom` matches the existing social-connect banner pattern.

One Major worth fixing before merge: the error message and the "If you believe…" sentence render as a run-on (`account.If you believe…`) because Angular's default `preserveWhitespaces: false` drops the whitespace-only text node between `{{ }}` and `@if`. The e2e `toContainText` assertion doesn't catch it because it only checks the message substring. The codebase convention is to put the separator *inside* the `@if` content (see `· ` in `newsletter-review.component.html`, `<span> Source: </span>` in `org-project-detail.component.html`, `(...)` in `events-list.component.html`). Suggested fix below moves the space into the interpolated string so it's preserved.

### What's done well

- Message-based gating via the shared util (same marker the backend sets) — correct and resilient to status-code drift.
- Negative e2e test isolates the message as the discriminator by keeping 409 constant — would catch a status-based regression.
- Self-contained e2e spec (mocks identities + send-code) that doesn't depend on a seeded backend test user, unlike the sibling `profile-identities-verify` spec.
- Reuses `lfxOpenIntercom` directive; no new support-routing code introduced.

apps/lfx-one/src/app/modules/profile/components/verify-identity-dialog/verify-identity-dialog.component.html
```diff
@@ -53,7 +53,13 @@
 
     @if (!codeSent()) {
       @if (verificationError()) {
-        <p class="text-sm text-red-600" data-testid="verify-email-error">{{ verificationError() }}</p>
+        <p class="text-sm text-red-600" data-testid="verify-email-error">
+          {{ verificationError() }}{{ showSupportLink() ? '.' : '' }}
```
audigregorie: **Major — same missing-space issue (verify-identity, pre-OTP block)**

Renders `account.If you believe…` for the same `preserveWhitespaces` reason. Same one-character fix:

```suggestion
          {{ verificationError() }}{{ showSupportLink() ? '. ' : '' }}
```

apps/lfx-one/src/app/modules/profile/components/verify-identity-dialog/verify-identity-dialog.component.html
```diff
@@ -84,7 +90,13 @@
           styleClass="w-full gap-3 [&_.p-inputtext]:flex-1"
           data-testid="verify-otp" />
         @if (verificationError()) {
-          <p class="text-center text-sm text-red-600" data-testid="verify-email-error">{{ verificationError() }}</p>
+          <p class="text-center text-sm text-red-600" data-testid="verify-email-error">
+            {{ verificationError() }}{{ showSupportLink() ? '.' : '' }}
```
audigregorie: **Major — same missing-space issue (verify-identity, OTP block)**

```suggestion
            {{ verificationError() }}{{ showSupportLink() ? '. ' : '' }}
```

apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.ts
audigregorie: **Minor — comment exceeds the 2-line limit**

Repo rule caps comments at 2 lines max. This "why" comment is 3 lines. Tighten:

```suggestion
  // Shows "contact support" for already-linked conflicts — detected via the shared util the
  // backend uses, so it works regardless of which subscribe callback set the error.
  public readonly showSupportLink = computed(() => isIdentityAlreadyLinkedError(this.verificationError()));
```

apps/lfx-one/src/app/modules/profile/components/verify-identity-dialog/verify-identity-dialog.component.ts
```diff
@@ -31,6 +33,10 @@ export class VerifyIdentityDialogComponent {
   public codeSent = signal(false);
   public verificationCode = signal('');
   public verificationError = signal('');
+  // Shows the "contact support" link whenever the visible error is an already-linked
+  // conflict — detected from the message text via the same shared util the backend uses,
+  // so it works regardless of which subscribe callback set the error.
+  public readonly showSupportLink = computed(() => isIdentityAlreadyLinkedError(this.verificationError()));
```
audigregorie: **Minor — same 3-line comment as add-account-dialog**

```suggestion
  // Shows "contact support" for already-linked conflicts — detected via the shared util the
  // backend uses, so it works regardless of which subscribe callback set the error.
  public readonly showSupportLink = computed(() => isIdentityAlreadyLinkedError(this.verificationError()));
```

apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.html
fayazg: Good catch — you're right. `preserveWhitespaces` isn't overridden in this repo (default `false`), so Angular strips the whitespace node between the interpolation and the `@if` block and it rendered as a run-on. Changed the separator from `'.'` to `'. '` here and in the other three blocks.

See commit 80abaf022: add-account/verify-identity dialog templates.

apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.html
fayazg: Fixed — same `'. '` separator applied to this `add-account-send-error` block. See commit 80abaf022.

apps/lfx-one/src/app/modules/profile/components/verify-identity-dialog/verify-identity-dialog.component.html
fayazg: Fixed — `'. '` separator applied to this verify-identity pre-OTP block. See commit 80abaf022.

apps/lfx-one/src/app/modules/profile/components/verify-identity-dialog/verify-identity-dialog.component.html
fayazg: Fixed — `'. '` separator applied to this verify-identity OTP block. See commit 80abaf022.

apps/lfx-one/src/app/modules/profile/components/add-account-dialog/add-account-dialog.component.ts
fayazg: Done — tightened the comment to two lines using your suggested wording. See commit 80abaf022.

apps/lfx-one/src/app/modules/profile/components/verify-identity-dialog/verify-identity-dialog.component.ts
fayazg: Done — same two-line comment applied here for consistency. See commit 80abaf022.

fayazg:
## Review Feedback Addressed

Commit: 80abaf022

### Changes Made
- **add-account-dialog.component.html** / **verify-identity-dialog.component.html**: changed the error separator from `'.'` to `'. '` in all four already-claimed error blocks. With `preserveWhitespaces` at its default (`false`), Angular strips the whitespace node between the interpolation and the `@if` block, so the message and the support-link sentence rendered as a run-on (`account.If you believe…`); the trailing space restores the separation (per @audigregorie).
- **add-account-dialog.component.ts** / **verify-identity-dialog.component.ts**: tightened the `showSupportLink` comment from three lines to two (per @audigregorie).

### Threads Resolved

audigregorie (approved):

PR #1209, opened 2026-07-27
fix(perf): resolve lens writer summary without full project sweep

manishdixitlfx:
## Summary

Datadog RUM showed the dashboard `/` view's avg `@view.loading_time` step from ~5s to ~10s on 2026-07-20, when `WriterGrantsService` (PR #1130) started calling `GET /api/projects` in `afterNextRender` on every bootstrap. That endpoint does an unscoped pull of every visible project plus a batch access-check (avg 11-17s, p75 ~19s), and the pending XHR counted into the view's loading time.

The only consumer of `WriterGrantsService` is `LensService.getAllowedLensIds()`, which needs just two booleans (`hasWriterFoundation`, `hasWriterProject`). This PR replaces the blocking call with a two-phase approach:

- **Fast path** — a new `GET /api/projects/writer-summary` endpoint reduces `ProjectService.getDirectGrantProjects` (the `filter_grants=direct` pattern PR #1193 already proved cheap) instead of the unscoped sweep. Bounded by a 10s timeout (`WRITER_SUMMARY_TIMEOUT_MS`) that wraps its own retry, so a stalled connection can't block indefinitely.
- **Deferred sweep** — the existing unscoped `getProjects()` call still runs, but scheduled via `requestIdleCallback` once the fast path settles, so it never blocks the bootstrap-critical path and doesn't count toward `@view.loading_time`. It resolves *inherited* writer access (e.g. a foundation-level writer's implicit access to a non-foundation child project) that the direct-grant-only fast path can't see — this preserves PR #1130's original goal of widening lens access for inherited writers, not just direct ones.

Both signals only ever widen (OR-merge), never narrow, so the two phases can resolve in any order.

## Behavior changes worth reviewer attention

- **Retry policy added to `ProjectService.getProjects()` and `.getWriterSummary()`** — one retry, transient errors only (network drop / 429 / 5xx), via a new shared `retryTransientHttpError()` operator (also adopted by `MeetupsListComponent`, deduping three previously-separate inline retry blocks). `getWriterSummary()`'s retry sits *inside* its `timeout()`, so the whole retried call — not each attempt — shares the 10s budget.
- **`getProjects()` no longer pins a failed fetch into its session-long cache.** Previously a single transient failure cached `[]` for the rest of the session for every caller sharing that cache key; failures now evict the cache entry so the next caller re-fetches.

## Known trade-offs (documented, not fixed here — out of scope for this ticket)

- **Lens-redirect timing window widened.** `lensRedirectGuard` reads `activeLens()` synchronously with no readiness gate on the async-populated writer grants (pre-existing design from LFXV2-2754). Before this change, inherited-only writer access resolved within the single ~11-19s bootstrap call; now it resolves in up to two sequential stages — up to `WRITER_SUMMARY_TIMEOUT_MS` (10s) for the fast path to settle, then up to `IDLE_SWEEP_TIMEOUT_MS` (15s) before the deferred sweep dispatches, plus that sweep's own request latency. The guard already treats this as accepted/cosmetic (the underlying write authorization is enforced elsewhere, by `writerGuard` and `projectQueryParamGuard`), and `MainLayoutComponent` re-asserts the lens once the signal widens — but the window is real and worth knowing about if this becomes a UX complaint. See the comment in `lens-redirect.guard.ts`.
- **`ProjectService.getProject()` (singular)** has the same poisoned-`shareReplay`-cache-on-transient-failure pattern that `getProjects()` (plural) had before this PR fixed it — a failed fetch pins `null` for the rest of the session for that project. Not touched here since it's a separate consumer path unrelated to the writer-grants flow; worth a follow-up ticket.
- This is the tactical fix. LFXV2-2753 (a performant permissions-list API) is the long-term fix for this class of problem.

## Test plan

- [x] New server-side unit tests for `getWriterSummary` (foundation-only / project-only / both / none / upstream error propagates)
- [x] New shared-package unit tests for the `summarizeWriterGrants` reducer
- [x] Manual validation against prod-backed dev server (`/lfx-full`)

### DIFF
--- a/apps/lfx-one/src/app/shared/services/writer-grants.service.ts
+++ b/apps/lfx-one/src/app/shared/services/writer-grants.service.ts
@@ -17,16 +17,20 @@ import { map } from 'rxjs';
  *
  * Two-phase, both kicked off post-hydration in `afterNextRender` (LFXV2-2857):
  *  - **Fast path** — `ProjectService.getWriterSummary()` hits a `filter_grants=direct`-scoped
- *    endpoint (sub-second). Direct grants only, so it under-reports *inherited* access (e.g. a
- *    foundation-level writer's implicit access to a non-foundation child they hold no direct
- *    tuple on).
+ *    endpoint (typically sub-second for a normal direct-grant count, though it still paginates
+ *    and access-checks server-side, so it isn't a hard bound). Direct grants only, so it
+ *    under-reports *inherited* access (e.g. a foundation-level writer's implicit access to a
+ *    non-foundation child they hold no direct tuple on).
  *  - **Deferred sweep** — the existing unscoped `ProjectService.getProjects()` (the same call
  *    that used to run here directly and block bootstrap for 11-19s), rescheduled via
  *    `requestIdleCallback` so it starts well after first paint and doesn't count toward RUM's
- *    `@view.loading_time`. Resolves inherited access. Runs every bootstrap (deliberately
- *    uncached: identity-scoped state here would need to invalidate on impersonation start/stop,
- *    both of which reload in-tab — see `docs/architecture/backend/impersonation.md` — and the
- *    idle-deferred call is cheap enough that re-running it isn't worth that risk).
+ *    `@view.loading_time`. Resolves inherited access. Server-side cost is unchanged — this only
+ *    moves it off the bootstrap-blocking path, it doesn't make the call itself cheaper. Runs
+ *    every bootstrap (deliberately uncached: identity-scoped state here would need to invalidate
+ *    on impersonation start/stop, both of which reload in-tab — see
+ *    `docs/architecture/backend/impersonation.md` — and re-running an idle-deferred call beats
+ *    that invalidation risk). Skipped entirely once both signals are already `true`, since an
+ *    OR-merge can no longer change the outcome.
  *
  * Both signals start `false` and only ever OR-merge to `true` as either call resolves — never
  * back — so the fast path, deferred sweep, and their arrival order don't matter to the consumer.
@@ -67,6 +71,11 @@ export class WriterGrantsService {
   }
 
   private runDeferredSweep(): void {
+    // Signals only ever widen — with both already true, the sweep's result can't change anything.
+    if (this.hasWriterFoundationInternal() && this.hasWriterProjectInternal()) {
+      return;
+    }
+
     this.projectService
       .getProjects()
       .pipe(map(summarizeWriterGrants), takeUntilDestroyed(this.destroyRef))
@@ -75,9 +84,11 @@ export class WriterGrantsService {
 
   private runWhenIdle(cb: () => void): void {
     if (typeof requestIdleCallback === 'function') {
-      requestIdleCallback(cb, { timeout: IDLE_SWEEP_TIMEOUT_MS });
+      const id = requestIdleCallback(cb, { timeout: IDLE_SWEEP_TIMEOUT_MS });
+      this.destroyRef.onDestroy(() => cancelIdleCallback(id));
     } else {
-      setTimeout(cb, IDLE_SWEEP_FALLBACK_DELAY_MS);
+      const id = setTimeout(cb, IDLE_SWEEP_FALLBACK_DELAY_MS);
+      this.destroyRef.onDestroy(() => clearTimeout(id));
     }
   }
 }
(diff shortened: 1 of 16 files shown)


apps/lfx-one/src/app/shared/services/writer-grants.service.ts
```diff
@@ -34,51 +44,65 @@ export class WriterGrantsService {
...
+    // Reachable from `finalize`, which also fires when destroy triggers takeUntilDestroyed's
+    // early unsubscribe — by then this.destroyRef is already destroyed (R3Injector.destroy() sets
+    // its destroyed flag before running teardown hooks), so onDestroy(...) below would throw.
+    if (this.destroyRef.destroyed) {
+      return;
+    }
+
+    if (typeof requestIdleCallback === 'function') {
+      const id = requestIdleCallback(cb, { timeout: IDLE_SWEEP_TIMEOUT_MS });
+      this.destroyRef.onDestroy(() => cancelIdleCallback(id));
```
jordane: This still looks unaddressed from Copilot's earlier comment: `requestIdleCallback` has no minimum delay, so `cb` can run in the very next idle frame right after the fast-path request completes. Datadog RUM's view-activity tracker only waits ~100ms of network idle before closing `view.loading_time`, so the deferred `getProjects()` sweep can still start inside that window and reproduce the 11–19s regression this PR is meant to fix.

Worth adding a real minimum delay before scheduling (or gating on a signal that the RUM loading window has actually closed) rather than relying on `requestIdleCallback` alone.

jordane (commented):
Nice fix for the bootstrap-blocking `getProjects()` call — the fast-path/deferred-sweep split, bounded retry, and cache-eviction-on-error all look solid, and the shared-package/backend/frontend conventions are clean.

One thing to resolve before merge: the `requestIdleCallback` timing gap Copilot flagged (inline comment below) — as written it can still let the deferred sweep start before Datadog's RUM loading window closes, which would reproduce the exact regression this PR is fixing. Once that's addressed, this looks good to me.

apps/lfx-one/src/app/shared/services/writer-grants.service.ts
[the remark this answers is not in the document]
manishdixitlfx: Done — added an `IDLE_SWEEP_MIN_DELAY_MS` floor (2s) before `requestIdleCallback` is even attempted for the deferred sweep. You were right that `requestIdleCallback` alone has no minimum-delay guarantee and could fire right after the fast-path XHR completes, still inside RUM's post-request activity-settle window.

See commit bf748ad71: `writer-grants.constants.ts` adds the new constant, `writer-grants.service.ts`'s `runWhenIdle` now waits that floor via `setTimeout` before scheduling `requestIdleCallback` (whose existing `IDLE_SWEEP_TIMEOUT_MS` ceiling still applies on top).

manishdixitlfx: Good catch, thanks — fixed. Added an `IDLE_SWEEP_MIN_DELAY_MS` floor (2s) before `requestIdleCallback` is even attempted, so the RUM post-request activity-settle window has a chance to close first regardless of how quickly the browser goes idle after the fast-path request completes.

See commit bf748ad71: the new floor lives in `writer-grants.constants.ts`, applied in `WriterGrantsService.runWhenIdle` ahead of the existing `requestIdleCallback`/`IDLE_SWEEP_TIMEOUT_MS` ceiling.

manishdixitlfx:
## Review Feedback Addressed

Commit: bf748ad71

### Changes Made
- **apps/lfx-one/src/app/shared/services/writer-grants.service.ts**, **packages/shared/src/constants/writer-grants.constants.ts**: added an `IDLE_SWEEP_MIN_DELAY_MS` floor (2s) before `requestIdleCallback` is even attempted for the deferred sweep. `requestIdleCallback` has no minimum-delay guarantee of its own and could fire on the very next idle frame right after the fast-path request completes — landing inside Datadog RUM's post-request activity-settle window and reproducing the exact `@view.loading_time` regression this PR fixes. The existing `IDLE_SWEEP_TIMEOUT_MS` ceiling still applies on top of this floor. (per @jordane, copilot-pull-request-reviewer[bot])

### No Change Needed
- **apps/lfx-one/src/app/modules/meetups/meetups-dashboard/components/meetups-list/meetups-list.component.ts:34-35**: coderabbitai flagged `retryTransientHttpError(2)` (3 total attempts) and suggested `1`. This is pre-existing behavior — `transientRetryCount = 2` predates this branch; this PR only extracted the shared `retryTransientHttpError` operator out of an inline `retry({...})` block that already used `count: 2`, it didn't change the retry count. Changing it would be an unrelated behavior change outside LFXV2-2857's bootstrap-perf scope, so leaving it as-is. (flagged by coderabbitai[bot])

### Threads Resolved
2 of 2 unresolved inline threads addressed and resolved in this iteration.

jordane (approved):
Confirmed — the `IDLE_SWEEP_MIN_DELAY_MS` floor (2s) resolves the timing gap: `requestIdleCallback` is no longer attempted until that floor elapses, which is comfortably past Datadog RUM's ~100ms settle window, so the deferred sweep can no longer re-enter `@view.loading_time`. The `lens-redirect.guard.ts` comment was also updated to reflect the new sequential budget accurately.

Nothing else outstanding. Approving.

PR #1210, opened 2026-07-28
fix(profile): prevent impersonation identity fallback to impersonator

fayazg:
## Summary

- The impersonation identity helpers no longer fall back to the impersonator's own OIDC
  identity when the target's stored field is empty (auth-service omitted the claim →
  `impersonation.service` stores `''`). Under impersonation they now return the target
  value or `null`, mirroring the already-correct `getEffectiveName`:
  - `getEffectiveEmail` — was surfacing the operator's address as the target's primary.
  - `getEffectiveUsername` — feeds per-user authorization/filtering (org-lens,
    persona-detection, meeting, changelog, …); the residual fallback could leak the
    operator's identity into target-scoped reads.
  - `getEffectiveSub` — same pattern, hardened for consistency.
- Audited the two `profile.controller` read call sites: `getLinuxAlias` leaves no bogus
  "(Primary)" forward option; `getUserEmails` returns 400 on the empty gap rather than
  leaking the operator's email.
- Added `auth-helper.spec.ts` locking in the null-under-impersonation contract for all
  three helpers (target present → target; target empty → null; not impersonating → OIDC).

Part of LFXV2-1663.

### DIFF
--- a/apps/lfx-one/src/server/controllers/profile.controller.ts
+++ b/apps/lfx-one/src/server/controllers/profile.controller.ts
@@ -425,6 +425,8 @@ export class ProfileController {
         this.emailVerificationService.listIdentitiesSafe(req, userSub),
       ]);
 
+      // sessionEmail is the target's email or undefined under impersonation (never the
+      // impersonator's), so a target with no primary correctly 400s instead of leaking it.
       const primaryEmail = freshEmails?.primary_email || sessionEmail;
 
       if (!primaryEmail) {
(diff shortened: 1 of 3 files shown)


MRashad26 (approved):
Code standards review: clean. Security fix, well-structured.

**The fix:** removing `&& req.appSession?.['impersonationUser']?.field` from the `isImpersonating()` guard is exactly right — the old short-circuit caused the function to fall through to the impersonator's OIDC identity when the target's stored field was empty. The new pattern `(... as string) || null` inside the `isImpersonating()` branch correctly returns null for the empty-target gap and never reaches the OIDC branch. Applied consistently to all three helpers (`getEffectiveEmail`, `getEffectiveUsername`, `getEffectiveSub`) ✅

**profile.controller.ts:** both call sites handle the null-under-impersonation case correctly — `getEffectiveEmail(req) ?? undefined` converts null to undefined, which falls through the `|| sessionEmail ||` chain to 400 rather than leaking the operator's address ✅

**Tests:** 9 cases across 3 helpers — target-present → target value, target-empty → null, not-impersonating → OIDC. `OPERATOR_OIDC` is always set in impersonation cases and is provably not returned in the empty-target branch, which is the critical invariant. Vitest structure is clean ✅

Approving.

PR #1211, opened 2026-07-28
fix(org-lens): drop MY_ORGANIZATION_ID cohort from leaderboard queries

ahmedomosanya:
Ticket: [LFXV2-2815](https://linuxfoundation.atlassian.net/browse/LFXV2-2815)

## Summary

- Removed `resolveViewerCohort()` and all `MY_ORGANIZATION_ID` cohort scoping from `fetchInfluenceBoardPage` + `fetchActivityBoardPage` in `org-lens-project-detail.service.ts`; both leaderboard boards now rank/count over the full org set (`WHERE PROJECT_SLUG = ? AND TIME_RANGE_TYPE = ?`, plus `BOARD_TYPE` for activity).

## Why

- The platinum leaderboard models dropped the per-viewer `MY_ORGANIZATION_ID` column, so every board query 500'd with `invalid identifier 'MY_ORGANIZATION_ID'`. This aligns the BFF with the model's one-row-per-org contract (the viewing org is flagged by `ORG_ACCOUNT_ID`).

## Non-obvious

- The viewing org is highlighted by `ORG_ACCOUNT_ID` (unchanged) per the model header — there is no cohort column to scope by, so `resolveViewerCohort` is removed entirely.

## Test plan

- [ ] Local dev-ui-only on :4200, Red Hat -> `/org/projects/k8s?tab=pd-leaderboards`; Technical/Ecosystem Influence + Activity Count boards populate.
- [ ] Against prod `ANALYTICS.PLATINUM_LFX_ONE` the query returns the full ranked set (Red Hat #1 for `kubevirt`).

[LFXV2-2815]: https://linuxfoundation.atlassian.net/browse/LFXV2-2815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
+++ b/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
@@ -562,38 +562,6 @@ export class OrgLensProjectDetailService {
   }
 
   /**
-   * The leaderboard tables store one ranked cohort per viewing organization (keyed by
-   * MY_ORGANIZATION_ID) plus a generic MY_ORGANIZATION_ID IS NULL cohort. Ranking or counting across
-   * all cohorts would duplicate organizations and inflate ranks/totals, so every board page must scope
-   * to a single cohort: resolve the viewer's ORG_ORGANIZATION_ID and use their cohort when it has rows,
-   * else the NULL cohort (mirrors the pre-pagination loader). Returns the WHERE fragment plus its bind
-   * params to apply identically to the ranked/page source and the count query.
-   */
-  private async resolveViewerCohort(table: string, slug: string, timeRangeType: string, orgUid: string): Promise<{ clause: string; params: string[] }> {
-    // One Salesforce account can map to several leaderboard rows (multiple crowd.dev orgs), so the pick
-    // must be deterministic — otherwise a per-page cohort lookup could resolve a different
-    // MY_ORGANIZATION_ID (or flip to the null cohort) on each page request and make ranks/totals/row
-    // membership disagree across pages. Order identically to fetchViewingLeaderboardRow (best-ranked
-    // row first) with ORG_ORGANIZATION_ID as a final unique tiebreak.
-    const orgIdResult = await this.snowflakeService.execute<{ ORG_ORGANIZATION_ID: string }>(
-      `SELECT ORG_ORGANIZATION_ID FROM ${this.leaderboardTable()} WHERE PROJECT_SLUG = ? AND TIME_RANGE_TYPE = ? AND ORG_ACCOUNT_ID = ?
-       ORDER BY RANK ASC NULLS LAST, SCORE_COMBINED DESC, ORG_NAME ASC, ORG_ORGANIZATION_ID ASC LIMIT 1`,
-      [slug, timeRangeType, orgUid]
-    );
-    const viewerOrgId = orgIdResult.rows[0]?.ORG_ORGANIZATION_ID;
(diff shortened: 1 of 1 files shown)


MRashad26 (approved):
Code standards review: clean.

Pure server-side deletion — no Angular/template code touched. All SQL parameters remain bound (`slug`, `timeRangeType`, `boardType`, `orgUid` in unchanged callers); `escapeSqlLikePattern` still applied before ILIKE; LIMIT/OFFSET still clamped integer literals; table names still from internal accessor methods. The simplified `WHERE PROJECT_SLUG = ? AND TIME_RANGE_TYPE = ?` clauses and `baseParams`/`params` arrays are consistent between the page and count queries in both methods ✅

**Minor nit (non-blocking):** `orgUid` is now an unused parameter in both `fetchInfluenceBoardPage` and `fetchActivityBoardPage` (was only consumed by `resolveViewerCohort`). Fine to leave — the callers still pass it, and a follow-up cleanup can drop it from both signatures once this is stable.

Approving.

audigregorie (commented):
## Code Review Summary

Clean, well-documented alignment fix. The cohort removal is the correct response to the platinum model dropping `MY_ORGANIZATION_ID`, and the diff is minimal and surgical — deterministic `ORDER BY` tiebreaks and `ESCAPE '!'` search handling are preserved, so pagination stability and literal-substring search semantics stay intact. Bind-param accounting verified correct across search/no-search branches for both boards.

### Minor — outside the diff

- **Per-org cache key is now org-agnostic data** (`fetchBoardPage`, ~line 534-535). After this PR the board SQL returns the same row set for every viewer org (`WHERE PROJECT_SLUG = ? AND TIME_RANGE_TYPE = ?` [+ `BOARD_TYPE`]); the only per-org difference is the `isViewingOrg` flag computed in `mapInfluenceRow` / `mapActivityRow`. The cache key is still scoped per-org via `buildOrgCacheKey(orgUid, cacheKey)`, so N distinct orgs viewing the same board each pay a separate Snowflake round-trip and store a near-duplicate cache entry. Before this PR the per-org key was justified by genuinely different cohort result sets; that justification is now gone. Pre-existing pattern (not introduced here), so it's an optimization opportunity rather than a regression — worth a follow-up: cache the raw Snowflake rows under an org-agnostic key and apply `isViewingOrg` after pulling from cache.
- **No unit tests for the changed SQL generation.** The PR changes bind-param composition and WHERE-clause shape for both board queries, but there are no unit tests asserting generated SQL / param ordering for the search-vs-no-search branches. A regression in `pageParams`/`countParams` ordering would silently swap binds. Pre-existing test gap, not introduced here, but the SQL change makes it worth flagging.

### What's done well

- Comments updated faithfully — the old cohort-scoping rationale is replaced with the new "one row per organization, no per-viewer cohort column" contract, and the stale `matches resolveViewerCohort` reference in `fetchViewingLeaderboardRow` is cleaned up too.
- Minimal, surgical diff — only the cohort resolution is removed; nothing else is touched.
- Bind-param accounting stays correct — `baseParams = [slug, timeRangeType]` matches the 2-placeholder WHERE in both page and count SQL for the influence board; the activity board keeps its 3-placeholder `[slug, timeRangeType, boardType]` shape.

⭐⭐⭐⭐⭐

audigregorie (approved):

PR #1214, opened 2026-07-28
feat(groups): add public group detail page

MRashad26:
## Summary

- Add a public-facing group detail page at `/groups/:id` that works without authentication, following the existing `/meetings/:id` pattern
- New backend API at `GET /public/api/groups/:id` using M2M token pattern for upstream calls
- Shared `PublicGroupDetail` interface and supporting types in `@lfx-one/shared/interfaces`

LFXV2-2009

## What's included

**Backend:**
- `public-groups.controller.ts` — fetches committee, members (chairs/vice-chairs), upcoming public meetings, project context; enriches with caller membership if authenticated
- `public-groups.route.ts` — simple router for `GET /:id`
- Route mounted in `server.ts` at `/public/api/groups`
- Privacy contract enforced: private groups return 403 (`GROUP_PRIVATE`), hidden member visibility suppresses chairs, private mailing lists excluded from links
- Cadence summary computed server-side via `buildCommitteeCadenceSummary`
- Project/foundation logo URLs included in context

**Frontend:**
- `GroupDetailComponent` — public group page with breadcrumb, header badges (category, join mode), visitor hero banner with sign-in CTA, member status banner, project logo, cadence display, calendar subscribe dialog
- `GroupNotFoundComponent` — distinct pages for not-found (404) vs private groups (403, lock icon + "Private Group" message)
- `GroupService` — Angular service calling the public API
- `authenticatedMatchGuard` — `canMatch` guard to resolve route collision between public/authenticated group routes
- Public routes added outside `authGuard` in `app.routes.ts`

**Shared:**
- `PublicGroupDetail`, `PublicGroupContext`, `PublicGroupLinks`, `PublicGroupMember`, `PublicGroupMeeting` interfaces
- `cadence`, `foundation_logo_url`, `project_logo_url` fields

## Review feedback addressed

- Privacy contract: chairs hidden when `member_visibility === 'hidden'`, mailing list link only shown when Groups.IO entity has `public: true`
- ROOT removed from breadcrumb
- Private groups show distinct "Private Group" page with lock icon instead of generic 404
- Project/foundation logo displayed next to group name
- Cadence summary ("Weekly on Monday · 60 min · Zoom") shown below header
- Calendar link replaced with subscribe dialog (Google Calendar / Outlook / webcal)
- Chat channel link removed (channels hidden for non-members)
- Info card hidden when empty (no description, no chairs, no links)

## Not in this PR

- **Slug-based URLs** (`/groups/:slug`): Committees don't have a `slug` field — only `project_slug` from parent project. Needs upstream backend work in committee-service to generate and persist slugs. Should be tracked as a separate story.
- **Breadcrumb navigation links**: Requires foundation/project directory pages to exist first

## Protected files

- `apps/lfx-one/src/server/server.ts` — Added public groups route registration following the existing pattern for public meetings/committees endpoints. Changes are additive (import + route mount) and follow established architecture.
- `apps/lfx-one/src/app/app.routes.ts` — Added 2 public routes (`groups/not-found`, `groups/:id`) outside the authGuard block, following the `/meetings/:id` pattern. Added `canMatch: [authenticatedMatchGuard]` to the authenticated groups route to resolve collision.

## Test plan

- [ ] Navigate to `/groups/<valid-public-group-uid>` unauthenticated — page renders with visitor view (hero banner, sign-in CTA, project logo, cadence)
- [ ] Navigate to `/groups/<valid-public-group-uid>` authenticated — page shows membership context if member
- [ ] Navigate to `/groups/<private-group-uid>` — shows "Private Group" page with lock icon
- [ ] Navigate to `/groups/invalid-uid` — shows "Group Not Found" page
- [ ] Verify breadcrumb shows project/foundation context
- [ ] Verify upcoming meetings list links to `/meetings/:id`
- [ ] Click "Subscribe to Calendar" — dialog opens with Google/Outlook/webcal options
- [ ] Verify mailing list link only appears when Groups.IO list is public
- [ ] Verify chairs hidden when committee has `member_visibility: hidden`
- [ ] Test on mobile viewport — responsive layout

### DIFF
--- a/apps/lfx-one/src/app/app.routes.ts
+++ b/apps/lfx-one/src/app/app.routes.ts
@@ -436,6 +436,14 @@ export const routes: Routes = [
     path: 'meetings/:id',
     loadComponent: () => import('./modules/meetings/meeting-join/meeting-join.component').then((m) => m.MeetingJoinComponent),
   },
+  {
+    path: 'groups/not-found',
+    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
+  },
+  {
+    path: 'groups/:id',
+    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
+  },
   // Invite acceptance — authGuard preserves ?token= through the Auth0 login redirect.
   {
     path: 'invite',
--- a/apps/lfx-one/src/app/app.routes.ts
+++ b/apps/lfx-one/src/app/app.routes.ts
@@ -5,6 +5,7 @@ import { MKTG_OS_AGENTS_ROUTE_SEGMENT } from '@lfx-one/shared/constants';
 import { Routes } from '@angular/router';
 
 import { authGuard } from './shared/guards/auth.guard';
+import { authenticatedMatchGuard } from './shared/guards/authenticated-match.guard';
 import { executiveDirectorGuard } from './shared/guards/executive-director.guard';
 import { lensRedirectGuard } from './shared/guards/lens-redirect.guard';
 import { newsletterAccessGuard } from './shared/guards/newsletter-access.guard';
@@ -17,16 +18,6 @@ import { settingsLensRedirectGuard } from './shared/guards/settings-lens-redirec
 const loadOrgProfilePage = () => import('./modules/dashboards/org/org-profile/org-profile.component').then((m) => m.OrgProfileComponent);
 
 export const routes: Routes = [
-  // Public group detail — must precede the root '' route because its authenticated child
-  // 'groups' → ':id' (COMMITTEE_ROUTES) would match first and trigger authGuard.
-  {
-    path: 'groups/not-found',
-    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
-  },
-  {
-    path: 'groups/:id',
-    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
-  },
   {
     path: '',
     canActivate: [authGuard],
@@ -333,6 +324,7 @@ export const routes: Routes = [
       },
       {
         path: 'groups',
+        canMatch: [authenticatedMatchGuard],
         canActivate: [lensRedirectGuard, projectQueryParamGuard],
         loadChildren: () => import('./modules/committees/committees.routes').then((m) => m.COMMITTEE_ROUTES),
       },
@@ -445,6 +437,14 @@ export const routes: Routes = [
   {
     path: 'meetings/:id',
     loadComponent: () => import('./modules/meetings/meeting-join/meeting-join.component').then((m) => m.MeetingJoinComponent),
+  },
+  {
+    path: 'groups/not-found',
+    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
+  },
+  {
+    path: 'groups/:id',
+    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
   },
   // Invite acceptance — authGuard preserves ?token= through the Auth0 login redirect.
   {
--- a/apps/lfx-one/src/app/app.routes.ts
+++ b/apps/lfx-one/src/app/app.routes.ts
@@ -17,6 +17,16 @@ import { settingsLensRedirectGuard } from './shared/guards/settings-lens-redirec
 const loadOrgProfilePage = () => import('./modules/dashboards/org/org-profile/org-profile.component').then((m) => m.OrgProfileComponent);
 
 export const routes: Routes = [
+  // Public group detail — must precede the root '' route because its authenticated child
+  // 'groups' → ':id' (COMMITTEE_ROUTES) would match first and trigger authGuard.
+  {
+    path: 'groups/not-found',
+    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
+  },
+  {
+    path: 'groups/:id',
+    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
+  },
   {
     path: '',
     canActivate: [authGuard],
@@ -435,14 +445,6 @@ export const routes: Routes = [
   {
     path: 'meetings/:id',
     loadComponent: () => import('./modules/meetings/meeting-join/meeting-join.component').then((m) => m.MeetingJoinComponent),
-  },
-  {
-    path: 'groups/not-found',
-    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
-  },
-  {
-    path: 'groups/:id',
-    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
   },
   // Invite acceptance — authGuard preserves ?token= through the Auth0 login redirect.
   {
--- a/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
+++ b/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
@@ -39,6 +39,9 @@
 
         <!-- Main Group Card -->
         <lfx-card class="rounded-[12.75px] mb-6" data-testid="group-info-card">
+          <!-- Group Title -->
+          <h1 class="mb-4" data-testid="group-title">{{ group.name }}</h1>
+
           <!-- Badges Row -->
           <div class="flex items-center gap-2 flex-wrap mb-4">
             <lfx-tag [value]="group.category" severity="secondary" styleClass="tag-category" data-testid="group-badge-category"></lfx-tag>
@@ -50,12 +53,15 @@
             }
           </div>
 
-          <!-- Group Title -->
-          <h1 class="mb-4" data-testid="group-title">{{ group.name }}</h1>
-
           <!-- Context Chips (Foundation / Project) -->
           <div class="flex flex-wrap items-center gap-2 mb-4" data-testid="group-context">
-            <lfx-tag [value]="group.context.foundation_name" severity="secondary" icon="fa-light fa-landmark" data-testid="group-context-foundation"></lfx-tag>
+            @if (group.context.foundation_name && group.context.foundation_name !== 'ROOT') {
+              <lfx-tag
+                [value]="group.context.foundation_name"
+                severity="secondary"
+                icon="fa-light fa-landmark"
+                data-testid="group-context-foundation"></lfx-tag>
+            }
             @if (group.context.project_name) {
               <lfx-tag
                 [value]="group.context.project_name"
--- a/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
+++ b/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
@@ -17,13 +17,24 @@
           </div>
           <p-skeleton width="100%" height="200px" borderRadius="12px" />
         </div>
+      } @else if (error()) {
+        <!-- Error state for unexpected failures -->
+        <lfx-card data-testid="group-error-card">
+          <div class="text-center py-12">
+            <div class="flex justify-center mb-6">
+              <div class="w-24 h-24 bg-red-50 rounded-full flex items-center justify-center">
+                <i class="fa-light fa-triangle-exclamation text-4xl text-red-400"></i>
+              </div>
+            </div>
+            <h2 class="text-lg font-semibold text-gray-900 mb-2">Something went wrong</h2>
+            <p class="text-sm text-gray-600 mb-4">We couldn't load this group. Please try again later.</p>
+          </div>
+        </lfx-card>
       } @else if (group(); as group) {
         <!-- Breadcrumb -->
         <nav class="flex items-center gap-1.5 text-sm text-gray-500 mb-4" data-testid="group-breadcrumb">
-          @if (parentProject(); as parent) {
-            <a [routerLink]="['/']" class="text-blue-500 hover:underline">{{ parent.name }}</a>
-            <i class="fa-light fa-chevron-right text-xs text-gray-400"></i>
-          }
+          <span class="text-gray-500">{{ group.context.foundation_name }}</span>
+          <i class="fa-light fa-chevron-right text-xs text-gray-400"></i>
           @if (group.context.project_name) {
             <span class="text-gray-500">{{ group.context.project_name }}</span>
             <i class="fa-light fa-chevron-right text-xs text-gray-400"></i>
@@ -93,7 +104,7 @@
                       </div>
                       <div class="flex flex-col gap-0.5 min-w-0">
                         <span class="text-sm font-medium text-gray-900 group-hover/meeting:text-blue-600 truncate">{{ meeting.title }}</span>
-                        <span class="text-xs text-gray-500">{{ meeting.starts_at | date: 'MMM d, y · h:mm a' : meeting.timezone }}</span>
+                        <span class="text-xs text-gray-500">{{ meeting.starts_at | meetingTime: 0 : 'compact' : meeting.timezone }}</span>
                       </div>
                     </a>
                   }
--- a/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
+++ b/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
@@ -45,7 +45,16 @@
 
         <!-- Page Header -->
         <div class="flex flex-col gap-2 mb-5" data-testid="group-header">
-          <h1 class="text-xl font-semibold text-gray-900">{{ group.name }}</h1>
+          <div class="flex items-center gap-3">
+            @if (projectLogoUrl(); as logoUrl) {
+              <img
+                [src]="logoUrl"
+                [alt]="group.context.project_name || group.context.foundation_name"
+                class="w-10 h-10 rounded-lg object-contain flex-shrink-0"
+                data-testid="group-project-logo" />
+            }
+            <h1 class="text-xl font-semibold text-gray-900">{{ group.name }}</h1>
+          </div>
           <div class="flex items-center gap-2 flex-wrap">
             <lfx-tag [value]="group.category" severity="secondary" styleClass="tag-category" data-testid="group-badge-category"></lfx-tag>
             @if (joinModeLabel(); as label) {
--- a/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
+++ b/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
@@ -77,47 +77,23 @@
           </div>
         }
 
-        <!-- Two-column layout -->
-        <div class="grid grid-cols-1 md:grid-cols-3 gap-5">
-          <!-- Main column -->
-          <div class="md:col-span-2 flex flex-col gap-5">
+        <!-- Group info card -->
+        <lfx-card data-testid="group-info-card">
+          <div class="flex flex-col gap-5">
             <!-- About / Description -->
             @if (group.description) {
-              <lfx-card data-testid="group-about-card">
+              <div data-testid="group-about-section">
                 <h3 class="text-sm font-semibold text-gray-900 mb-3">About</h3>
                 <lfx-expandable-text [maxHeight]="200">{{ group.description }}</lfx-expandable-text>
-              </lfx-card>
+              </div>
             }
 
-            <!-- Upcoming Meetings -->
-            @if (hasUpcomingMeetings()) {
-              <lfx-card data-testid="group-meetings-card">
-                <h3 class="text-sm font-semibold text-gray-900 mb-3">Upcoming Meetings</h3>
-                <div class="flex flex-col gap-3">
-                  @for (meeting of group.upcoming_meetings; track meeting.uid) {
-                    <a
-                      [routerLink]="[meeting.url]"
-                      class="flex items-start gap-3 p-3 rounded-lg hover:bg-gray-50 transition-colors group/meeting"
-                      [attr.data-testid]="'meeting-' + meeting.uid">
-                      <div class="w-8 h-8 rounded-lg bg-blue-50 flex items-center justify-center flex-shrink-0 mt-0.5">
-                        <i class="fa-light fa-video text-blue-500 text-sm"></i>
-                      </div>
-                      <div class="flex flex-col gap-0.5 min-w-0">
-                        <span class="text-sm font-medium text-gray-900 group-hover/meeting:text-blue-600 truncate">{{ meeting.title }}</span>
-                        <span class="text-xs text-gray-500">{{ meeting.starts_at | meetingTime: 0 : 'compact' : meeting.timezone }}</span>
-                      </div>
-                    </a>
-                  }
-                </div>
-              </lfx-card>
-            }
-          </div>
-
-          <!-- Sidebar -->
-          <div class="flex flex-col gap-5">
-            <!-- Chairs card -->
+            <!-- Leadership -->
             @if (group.chairs.length > 0) {
-              <lfx-card data-testid="group-chairs-card">
+              @if (group.description) {
+                <hr class="border-gray-200" />
+              }
+              <div data-testid="group-leadership-section">
                 <h3 class="text-sm font-semibold text-gray-900 mb-3">Leadership</h3>
                 <div class="flex flex-col gap-3">
                   @for (chair of group.chairs; track chair.name) {
@@ -137,12 +113,13 @@
                     </div>
                   }
                 </div>
-              </lfx-card>
+              </div>
             }
 
-            <!-- Links card -->
+            <!-- Links -->
             @if (hasLinks()) {
-              <lfx-card data-testid="group-links-card">
+              <hr class="border-gray-200" />
+              <div data-testid="group-links-section">
                 <h3 class="text-sm font-semibold text-gray-900 mb-3">Links</h3>
                 <div class="flex flex-col gap-2">
                   @if (group.links.website) {
@@ -186,10 +163,33 @@
                     </a>
                   }
                 </div>
-              </lfx-card>
+              </div>
             }
           </div>
-        </div>
+        </lfx-card>
+
+        <!-- Upcoming Meetings -->
+        @if (hasUpcomingMeetings()) {
+          <lfx-card data-testid="group-meetings-card">
+            <h3 class="text-sm font-semibold text-gray-900 mb-3">Upcoming Meetings</h3>
+            <div class="flex flex-col gap-3">
+              @for (meeting of group.upcoming_meetings; track meeting.uid) {
+                <a
+                  [routerLink]="[meeting.url]"
+                  class="flex items-start gap-3 p-3 rounded-lg hover:bg-gray-50 transition-colors group/meeting"
+                  [attr.data-testid]="'meeting-' + meeting.uid">
+                  <div class="w-8 h-8 rounded-lg bg-blue-50 flex items-center justify-center flex-shrink-0 mt-0.5">
+                    <i class="fa-light fa-video text-blue-500 text-sm"></i>
+                  </div>
+                  <div class="flex flex-col gap-0.5 min-w-0">
+                    <span class="text-sm font-medium text-gray-900 group-hover/meeting:text-blue-600 truncate">{{ meeting.title }}</span>
+                    <span class="text-xs text-gray-500">{{ meeting.starts_at | meetingTime: 0 : 'compact' : meeting.timezone }}</span>
+                  </div>
+                </a>
+              }
+            </div>
+          </lfx-card>
+        }
       }
     </div>
   </div>
--- a/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
+++ b/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
@@ -46,7 +46,7 @@
           <div class="flex items-center gap-2 flex-wrap mb-4">
             <lfx-tag [value]="group.category" severity="secondary" styleClass="tag-category" data-testid="group-badge-category"></lfx-tag>
             @if (joinModeLabel(); as label) {
-              <lfx-tag [value]="label" severity="secondary" [icon]="joinModeIcon()!" data-testid="group-badge-join-mode"></lfx-tag>
+              <lfx-tag [value]="label" severity="secondary" [icon]="joinModeIcon() || ''" data-testid="group-badge-join-mode"></lfx-tag>
             }
             @if (group.cadence) {
               <lfx-tag [value]="group.cadence" severity="secondary" icon="fa-light fa-repeat" data-testid="group-badge-cadence"></lfx-tag>
--- a/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
+++ b/apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
@@ -9,13 +9,10 @@
       @if (loading()) {
         <!-- Skeleton loading state -->
         <div class="flex flex-col gap-4">
-          <p-skeleton width="200px" height="16px" />
-          <p-skeleton width="300px" height="28px" />
-          <div class="flex gap-2">
-            <p-skeleton width="100px" height="24px" borderRadius="6px" />
-            <p-skeleton width="120px" height="24px" borderRadius="6px" />
+          <div class="flex items-center justify-center mb-6">
+            <p-skeleton width="120px" height="56px" />
           </div>
-          <p-skeleton width="100%" height="200px" borderRadius="12px" />
+          <p-skeleton width="100%" height="400px" borderRadius="12px" />
         </div>
       } @else if (error()) {
         <!-- Error state for unexpected failures -->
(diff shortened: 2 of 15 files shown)


apps/lfx-one/src/app/app.routes.ts
```diff
@@ -431,6 +431,14 @@ export const routes: Routes = [
...
     loadComponent: () => import('./modules/meetings/meeting-join/meeting-join.component').then((m) => m.MeetingJoinComponent),
   },
+  {
+    path: 'groups/not-found',
+    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
+  },
+  {
+    path: 'groups/:id',
+    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
+  },
```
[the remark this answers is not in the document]
MRashad26: Done — added `{ pattern: '/groups/', type: 'ssr', auth: 'optional' }` to the auth middleware route config, following the existing `/meetings/` pattern.

Note: the Angular route collision concern is a false positive — the authenticated `groups` route at line 325 is nested inside the authGuard wrapper at path `/<foundation>/<project>/groups/:id`, while our public route at `groups/:id` sits at root level outside authGuard. Different path depths, no collision.

See commit 82438f925.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,146 @@
...
+        member_visibility: committee.member_visibility!,
+      };
+
+      const isAuthenticated = req.oidc?.isAuthenticated();
+      if (isAuthenticated && originalToken !== undefined) {
+        req.bearerToken = originalToken;
+        try {
+          const callerMembership = members.find(
+            (m: CommitteeMember) => m.email === req.oidc?.user?.['email'] || m.username === req.oidc?.user?.['preferred_username']
+          );
```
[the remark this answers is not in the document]
MRashad26: Done — switched to `getEffectiveEmail(req)` and `getEffectiveUsername(req)` from `auth-helper.ts`. These handle impersonation, Auth0 nickname/username/preferred_username fallback, case normalization, and `auth0|` prefix stripping. Also added null guards to prevent `undefined === undefined` false matches.

See commit 82438f925.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,146 @@
...
+        .slice(0, 5)
+        .map(
+          (m): PublicGroupMeeting => ({
+            uid: m.id,
+            title: m.title,
+            starts_at: m.start_time,
+            timezone: m.timezone,
+            url: `/meetings/${m.id}`,
+          })
+        );
```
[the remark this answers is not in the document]
MRashad26: Done — added `m.start_time >= now` filter to exclude past meetings, and `.sort((a, b) => a.start_time.localeCompare(b.start_time))` to ensure chronological order before slicing top 5.

See commit 82438f925.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -0,0 +1,185 @@
...
+          </div>
+          <p-skeleton width="100%" height="200px" borderRadius="12px" />
+        </div>
+      } @else if (group(); as group) {
+        <!-- Breadcrumb -->
+        <nav class="flex items-center gap-1.5 text-sm text-gray-500 mb-4" data-testid="group-breadcrumb">
+          @if (parentProject(); as parent) {
+            <a [routerLink]="['/']" class="text-blue-500 hover:underline">{{ parent.name }}</a>
+            <i class="fa-light fa-chevron-right text-xs text-gray-400"></i>
+          }
```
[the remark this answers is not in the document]
MRashad26: Done — removed the `parentProject` signal and its protected `/api/projects/:slug` fetch entirely. The breadcrumb now renders directly from `group.context.foundation_name` and `group.context.project_name`, which are already populated by the public API response. No authenticated endpoint needed.

See commit 82438f925.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -431,6 +431,14 @@ export const routes: Routes = [
     path: 'meetings/:id',
     loadComponent: () => import('./modules/meetings/meeting-join/meeting-join.component').then((m) => m.MeetingJoinComponent),
   },
+  {
+    path: 'groups/not-found',
+    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
+  },
+  {
+    path: 'groups/:id',
+    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
```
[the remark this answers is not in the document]
MRashad26: Done — added `{ pattern: '/groups/', type: 'ssr', auth: 'optional' }` to the auth middleware config, matching the `/meetings/` pattern. Anonymous visitors can now reach the group detail and not-found pages.

See commit 82438f925.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  private initParentProject(): Signal<Project | null> {
+    return toSignal(
+      toObservable(this.group).pipe(
+        map((g) => g?.context.foundation_slug ?? null),
+        distinctUntilChanged(),
+        switchMap((slug) => {
+          if (!slug) return of(null);
+          return this.projectService.getProject(slug, false).pipe(catchError(() => of(null)));
+        }),
```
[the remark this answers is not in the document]
MRashad26: Done — removed the `parentProject` signal and the `ProjectService` dependency entirely. The breadcrumb now uses `group.context.foundation_name` directly from the public API response — no authenticated call needed.

See commit 82438f925.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
```diff
@@ -0,0 +1,135 @@
...
+        switchMap((id) => {
+          this.loading.set(true);
+          return this.groupService.getPublicGroup(id).pipe(
+            catchError((error) => {
+              if ([404, 403, 400].includes(error.status)) {
+                this.router.navigate(['/groups/not-found']);
+              }
+              this.loading.set(false);
+              return of(null);
+            })
```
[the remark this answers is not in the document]
MRashad26: Done — added an `error` signal. Only 400/403/404 navigate to the not-found page; all other errors (5xx, network) set `error.set(true)` and the template renders an explicit error state with a message instead of a blank page.

See commit 82438f925.

packages/shared/src/interfaces/public-group.interface.ts
```diff
@@ -0,0 +1,57 @@
...
+  ends_at?: string;
+  timezone?: string;
+  url?: string;
+}
+
+export interface PublicGroupLinks {
+  website?: string | null;
+  mailing_list?: string | null;
+  chat_channel?: string | null;
+  calendar?: string | null;
```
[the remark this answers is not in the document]
MRashad26: Done — the controller now normalizes email addresses to `mailto:` URLs: if the value contains `@` and doesn't start with `http`, it gets a `mailto:` prefix. URL values pass through unchanged.

See commit 82438f925.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,146 @@
...
+import { generateM2MToken } from '../utils/m2m-token.util';
+
+const CHAIR_ROLES = new Set<string>([CommitteeMemberRole.CHAIR, CommitteeMemberRole.VICE_CHAIR]);
+
+export class PublicGroupsController {
+  private committeeService: CommitteeService = new CommitteeService();
+  private meetingService: MeetingService = new MeetingService();
+  private projectService: ProjectService = new ProjectService();
+
+  public async getPublicGroupById(req: Request, res: Response, next: NextFunction): Promise<void> {
```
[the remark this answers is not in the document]
MRashad26: Acknowledged — unit tests for the controller are planned as a follow-up. This MVP PR focuses on the feature implementation; test coverage will be added in a subsequent PR following the `public-meeting.controller.spec.ts` pattern.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -0,0 +1,185 @@
...
+            <p-skeleton width="100px" height="24px" borderRadius="6px" />
+            <p-skeleton width="120px" height="24px" borderRadius="6px" />
+          </div>
+          <p-skeleton width="100%" height="200px" borderRadius="12px" />
+        </div>
+      } @else if (group(); as group) {
+        <!-- Breadcrumb -->
+        <nav class="flex items-center gap-1.5 text-sm text-gray-500 mb-4" data-testid="group-breadcrumb">
+          @if (parentProject(); as parent) {
+            <a [routerLink]="['/']" class="text-blue-500 hover:underline">{{ parent.name }}</a>
```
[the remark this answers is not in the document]
MRashad26: Done — the breadcrumb now renders `group.context.foundation_name` and `group.context.project_name` as text labels (not links). Foundation and project detail pages don't exist as public routes yet (planned for Phase 2 directory pages), so linking to them would 404 for anonymous visitors. The crumbs are intentionally labels for now.

See commit 82438f925.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -0,0 +1,185 @@
...
+                    <a
+                      [routerLink]="[meeting.url]"
+                      class="flex items-start gap-3 p-3 rounded-lg hover:bg-gray-50 transition-colors group/meeting"
+                      [attr.data-testid]="'meeting-' + meeting.uid">
+                      <div class="w-8 h-8 rounded-lg bg-blue-50 flex items-center justify-center flex-shrink-0 mt-0.5">
+                        <i class="fa-light fa-video text-blue-500 text-sm"></i>
+                      </div>
+                      <div class="flex flex-col gap-0.5 min-w-0">
+                        <span class="text-sm font-medium text-gray-900 group-hover/meeting:text-blue-600 truncate">{{ meeting.title }}</span>
+                        <span class="text-xs text-gray-500">{{ meeting.starts_at | date: 'MMM d, y · h:mm a' : meeting.timezone }}</span>
```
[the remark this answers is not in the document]
MRashad26: Done — switched from Angular `DatePipe` to the project'\''s `MeetingTimePipe` which uses `Intl.DateTimeFormat` with proper IANA timezone support. Using the `compact` format for concise display.

See commit 82438f925.

MRashad26:
## Review Feedback Addressed

Commit: 82438f925d7327d041c148078e44052fab2a30b2

### Changes Made
- **auth.middleware.ts**: Added `/groups/` as optional auth SSR route for anonymous access (per @cursor, @copilot-pull-request-reviewer)
- **public-groups.controller.ts**: Switched to `getEffectiveEmail`/`getEffectiveUsername` for membership matching (per @cursor, @copilot-pull-request-reviewer)
- **public-groups.controller.ts**: Added future-only filter and chronological sort for upcoming meetings (per @cursor)
- **public-groups.controller.ts**: Normalized mailing_list email addresses to `mailto:` URLs (per @copilot-pull-request-reviewer)
- **group-detail.component.ts**: Removed `parentProject` signal and protected API call; breadcrumb uses `group.context` directly (per @copilot-pull-request-reviewer)
- **group-detail.component.ts**: Added `error` signal for 5xx failures with explicit error UI (per @copilot-pull-request-reviewer)
- **group-detail.component.html**: Switched from `DatePipe` to `MeetingTimePipe` for IANA timezone support (per @copilot-pull-request-reviewer)

### No Change Needed
- **app.routes.ts** route collision: The authenticated `groups` route is nested inside authGuard at `/<foundation>/<project>/groups/:id`; our public `groups/:id` is at root level outside authGuard — different path depths, no collision (flagged by @cursor)
- **Breadcrumb links**: Foundation/project crumbs are intentionally text labels — public directory pages don't exist yet (Phase 2) (flagged by @copilot-pull-request-reviewer)

### Threads Resolved
10 of 11 unresolved threads addressed in this iteration.

### Still Open
- **public-groups.controller.ts:24**: Unit test coverage deferred to follow-up PR (flagged by @copilot-pull-request-reviewer)

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        description: committee.description ?? undefined,
+        category: committee.category,
+        join_mode: committee.join_mode,
+        total_members: committee.total_members ?? members.length,
+        context,
+        chairs,
+        links,
+        upcoming_meetings: upcomingMeetings,
+        calendar_url: links.calendar ?? undefined,
+        member_visibility: committee.member_visibility!,
```
audigregorie: **Major — Unsafe non-null assertion on an optional field**

`committee.member_visibility!` asserts a value that the `Committee` interface marks optional (`member_visibility?: CommitteeMemberVisibility`). When upstream omits the field, `undefined` is assigned to `PublicGroupDetail.member_visibility` (typed required), violating the contract. Default to a safe value (e.g. `CommitteeMemberVisibility.BASIC_PROFILE`) or make the interface field optional.

```suggestion
        member_visibility: committee.member_visibility ?? CommitteeMemberVisibility.BASIC_PROFILE,
```

audigregorie (commented):
## Code Review Summary

Solid implementation that follows the established public-meeting pattern; the `authenticatedMatchGuard` is a clean solution to the route collision. The Majors below are backend efficiency/contract concerns (wasted FGA call, full roster fetch, single-page meeting fetch, unsafe assertion) rather than correctness blockers, but they're worth addressing before merge given this is a public-facing endpoint.

### Major — outside the diff

- **No tests.** The existing `public-meeting.controller` has a spec (`public-meeting.controller.spec.ts`); the new `public-groups.controller` has none. A controller spec covering the public/private gate, the member-enrichment branch, and the upcoming-meetings filter would mirror the established pattern.
- **`member_visibility` semantics.** The field is shipped in `PublicGroupDetail` but unused by the frontend, and chairs are exposed regardless of its value. If `HIDDEN` is meant to suppress leadership display, the controller should gate `chairs` on it; otherwise the field is dead data and can be removed.
- **PR description may be stale.** It describes a "two-column layout (about + meetings / chairs + links)" but the template renders a single-column stack of cards. Confirm the description matches the intended design.

### What's done well

- The `authenticatedMatchGuard` reads auth from `TransferState`/`REQUEST_CONTEXT` (populated before routing) rather than `UserService.authenticated` (set after) — the right call, and the comment explains the timing rationale.
- M2M token handling mirrors `public-meeting.controller`, including the `originalToken !== undefined` guard against token-refresh-failure edge cases.
- `catchError` correctly distinguishes "not found" (400/403/404 → redirect) from transient failures (error state), and the component owns `loading`/`error` per conventions.

Rating: ⭐⭐⭐⭐ (4/5)

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (!validateUidParameter(id, req, next, { operation: 'get_public_group_by_id', service: 'public_groups_controller' })) {
+        return;
+      }
+
+      const originalToken = req.bearerToken;
+      // M2M token required: public endpoint has no user session; app credentials needed for upstream calls
+      const m2mToken = await generateM2MToken(req);
+      req.bearerToken = m2mToken;
+
+      const committee = await this.committeeService.getCommitteeById(req, id);
```
audigregorie, line 50: **Major — Wasted FGA access check on every public page load**

`getCommitteeById` unconditionally runs `accessCheckService.addAccessToResource(req, committee, 'committee')` (an FGA call) using the M2M identity. The result (`committee.writer`) is discarded by this public controller. `getProjectById` and `getMeetingById` both accept `access: false`; `getCommitteeById` does not. Consider adding an `access`/`includeAccess` option to `getCommitteeById` and passing `false` here, matching the public-meeting pattern.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (!committee.public) {
+        throw new ResourceNotFoundError('Group', id, {
+          operation: 'get_public_group_by_id',
+          service: 'public_groups_controller',
+          path: `/committees/${id}`,
+        });
+      }
+
+      const [members, project, meetingsResponse] = await Promise.all([
+        this.committeeService.getCommitteeMembers(req, id),
```
audigregorie, line 62: **Major — Full member roster fetched to extract 2-3 chairs**

`getCommitteeMembers` uses `fetchAllQueryResources` (all pages). On a public endpoint, the entire roster is pulled over the wire just to filter to Chair/Vice Chair locally. For committees with large rosters this is wasteful. If the query service supports a role filter, use it; otherwise consider a lighter upstream call that returns only leadership.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        this.projectService.getProjectById(req, committee.project_uid, false),
+        this.meetingService.getMeetings(req, { tags: `committee_uid:${id}` }, 'v1_meeting', false),
```
audigregorie, line 64: **Major — Upcoming meetings may be missed due to single-page fetch**

`getMeetings` returns only the first page of meetings. If the committee has many meetings and the upcoming public ones aren't on page 1 (sort-order dependent), the "Upcoming Meetings" section renders empty even when future public meetings exist. Request an explicit sort/filter (e.g. `start_time` ascending from now) or paginate through to find upcoming items.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        foundation_slug: parentProject?.slug || project?.slug || '',
+        ...(project?.parent_uid && {
+          project_uid: project.uid,
+          project_name: project.name,
+          project_slug: project.slug,
+        }),
+      };
+
+      const mailingList = committee.mailing_list;
+      const normalizedMailingList = mailingList && !mailingList.startsWith('http') && mailingList.includes('@') ? `mailto:${mailingList}` : mailingList;
```
audigregorie: **Minor — `mailto:` double-prefix risk**

If upstream already returns `mailto:foo@bar.com`, this normalization produces `mailto:mailto:foo@bar.com`. Guard against an existing `mailto:` prefix.

```suggestion
      const normalizedMailingList = mailingList && !mailingList.startsWith('http') && !mailingList.toLowerCase().startsWith('mailto:') && mailingList.includes('@') ? `mailto:${mailingList}` : mailingList;
```

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        name: committee.name,
+        description: committee.description ?? undefined,
+        category: committee.category,
+        join_mode: committee.join_mode,
+        total_members: committee.total_members ?? members.length,
+        context,
+        chairs,
+        links,
+        upcoming_meetings: upcomingMeetings,
+        calendar_url: links.calendar ?? undefined,
```
audigregorie, line 150: **Minor — Redundant nullish coalescing**

`links.calendar` is already `string | undefined`, so `?? undefined` is a no-op.

```suggestion
        calendar_url: links.calendar,
```

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        calendar: committee.calendar?.public ? `/public/api/committees/${id}/calendar.ics` : undefined,
+      };
+
+      const detail: PublicGroupDetail = {
+        uid: committee.uid,
+        name: committee.name,
+        description: committee.description ?? undefined,
+        category: committee.category,
+        join_mode: committee.join_mode,
+        total_members: committee.total_members ?? members.length,
```
audigregorie: **Minor — Dead fallback per type**

`Committee.total_members` is typed as required `number`, so `?? members.length` never triggers per the type. Defensive but misleading — either trust the type (`committee.total_members`) or keep the fallback with a comment noting upstream has been observed sending `null`.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { CommitteeMemberRole } from '@lfx-one/shared/enums';
+import { CommitteeMember, PublicGroupContext, PublicGroupDetail, PublicGroupLinks, PublicGroupMeeting, PublicGroupMember } from '@lfx-one/shared/interfaces';
+import { MeetingVisibility } from '@lfx-one/shared/enums';
```
audigregorie: **Minor — Split enum import**

`CommitteeMemberRole` and `MeetingVisibility` are both from `@lfx-one/shared/enums` but imported on separate lines. Combine into one import statement.

```suggestion
import { CommitteeMemberRole, MeetingVisibility } from '@lfx-one/shared/enums';
import { CommitteeMember, PublicGroupContext, PublicGroupDetail, PublicGroupLinks, PublicGroupMeeting, PublicGroupMember } from '@lfx-one/shared/interfaces';
```

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
```diff
@@ -0,0 +1,121 @@
...
+    MeetingTimePipe,
+    SkeletonModule,
+  ],
+  templateUrl: './group-detail.component.html',
+})
+export class GroupDetailComponent {
+  private readonly activatedRoute = inject(ActivatedRoute);
+  private readonly router = inject(Router);
+  private readonly groupService = inject(GroupService);
+  protected readonly userService = inject(UserService);
```
audigregorie: **Minor — `userService` should be `private`**

The template never references `userService` directly; only `authenticated` (derived from it) is used. Mark it `private` to keep the public surface minimal.

```suggestion
  private readonly userService = inject(UserService);
```

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
```diff
@@ -0,0 +1,121 @@
...
+              }
+              this.loading.set(false);
+              return of(null);
+            })
+          );
+        }),
+        map((group) => {
+          this.loading.set(false);
+          return group;
+        })
```
audigregorie, line 145: **Minor — `loading.set(false)` runs twice on error**

`catchError` sets `loading.set(false)` and returns `of(null)`, which then flows into the subsequent `map` that sets `loading.set(false)` again. Harmless but redundant — drop the call in `catchError` and let the `map` own it, or vice versa.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -0,0 +1,197 @@
...
+          <span class="text-gray-700 font-medium">{{ group.name }}</span>
+        </nav>
+
+        <!-- Page Header -->
+        <div class="flex flex-col gap-2 mb-5" data-testid="group-header">
+          <h1 class="text-xl font-semibold text-gray-900">{{ group.name }}</h1>
+          <div class="flex items-center gap-2 flex-wrap">
+            <lfx-tag [value]="group.category" severity="secondary" styleClass="tag-category" data-testid="group-badge-category"></lfx-tag>
+            @if (joinModeLabel(); as label) {
+              <lfx-tag [value]="label" severity="secondary" [icon]="joinModeIcon()!" data-testid="group-badge-join-mode"></lfx-tag>
```
audigregorie: **Minor — Non-null assertion in template is fragile**

`joinModeIcon()` returns `null` in the default case. The `@if (joinModeLabel())` guard makes this safe today, but the two computeds are independent — adding a new `JoinMode` to the type without updating the icon switch would pass `null` to the `icon` input. Prefer:

```suggestion
              <lfx-tag [value]="label" severity="secondary" [icon]="joinModeIcon() ?? undefined" data-testid="group-badge-join-mode"></lfx-tag>
```

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -0,0 +1,197 @@
...
+              @if (group.description) {
+                <hr class="border-gray-200" />
+              }
+              <div data-testid="group-leadership-section">
+                <h3 class="text-sm font-semibold text-gray-900 mb-3">Leadership</h3>
+                <div class="flex flex-col gap-3">
+                  @for (chair of group.chairs; track chair.name) {
+                    <div class="flex items-center gap-2.5">
+                      <div class="w-8 h-8 rounded-full bg-gray-100 flex items-center justify-center flex-shrink-0">
+                        <span class="text-xs font-medium text-gray-600">{{ chair.name | slice: 0 : 2 | uppercase }}</span>
```
audigregorie: **Minor — Initials take first 2 chars of full name**

`chair.name | slice: 0 : 2` yields "Jo" for "John Doe" rather than the conventional "JD" (first letter of first + last name). If the latter is intended, derive initials from the name parts rather than the raw string prefix.

packages/shared/src/interfaces/public-group.interface.ts
```diff
@@ -0,0 +1,57 @@
...
+  description?: string;
+  category: string;
+  join_mode?: JoinMode;
+  total_members: number;
+  context: PublicGroupContext;
+  chairs: PublicGroupMember[];
+  links: PublicGroupLinks;
+  upcoming_meetings: PublicGroupMeeting[];
+  calendar_url?: string;
+  member_visibility: CommitteeMemberVisibility;
```
audigregorie: **Minor — `member_visibility` required but sourced from optional**

`member_visibility: CommitteeMemberVisibility` is required here, but the controller sources it from `Committee.member_visibility?` (optional) via a non-null assertion. Either make this field optional (`member_visibility?: ...`) or default it in the controller so the contract holds.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+      }
+
+      const [members, project, meetingsResponse] = await Promise.all([
+        this.committeeService.getCommitteeMembers(req, id),
+        this.projectService.getProjectById(req, committee.project_uid, false),
+        this.meetingService.getMeetings(req, { tags: `committee_uid:${id}` }, 'v1_meeting', false),
+      ]);
+
+      const chairs = members
+        .filter((m: CommitteeMember) => m.role?.name && CHAIR_ROLES.has(m.role.name))
```
ahmedomosanya: **Blocker (privacy) — concurring with Copilot's flag here.** The current implementation returns `chairs` (names + organizations) to anonymous callers unconditionally, ignoring `member_visibility`. That setting defaults to `hidden` (`committee-manage.component.ts` uses `new FormControl('hidden')` / `|| 'hidden'`), and the authenticated members UI already fails closed on it:

```ts
// committee-members.component.ts
return committee.member_visibility === CommitteeMemberVisibility.BASIC_PROFILE || this.canManageMembers();
```

So a `public: true` committee with `member_visibility: hidden` would leak chair identities to the anonymous web. Could we gate the `chairs` array the same way — only populate it when `member_visibility === BASIC_PROFILE`? Something like:

```ts
const chairs = committee.member_visibility === CommitteeMemberVisibility.BASIC_PROFILE
  ? members.filter(...).map(...)
  : [];
```

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        context,
+        chairs,
+        links,
+        upcoming_meetings: upcomingMeetings,
+        calendar_url: links.calendar ?? undefined,
+        member_visibility: committee.member_visibility!,
+      };
+
+      const isAuthenticated = req.oidc?.isAuthenticated();
+      if (isAuthenticated && originalToken !== undefined) {
```
ahmedomosanya, line 155: **Question — is this branch reachable?** `authenticatedMatchGuard` sends authenticated users at `/groups/:id` into the authenticated `groups` child → `COMMITTEE_ROUTES` → `CommitteeViewComponent`, so they never load the public `GroupDetailComponent` and never call `/public/api/groups/:id`. The only way an authenticated identity reaches this handler seems to be the SSR silent-login race (auth not yet resolved at `canMatch` time), which also means the same SSO-session user can land on either the authenticated committee view or the public page depending on timing.

If that's the case, this `is_member`/`my_role` enrichment (and the template's member-status banner) is effectively dead — could we drop it? Or, if we do want authenticated members to see the public page, should there be a deterministic route to it? Also a small thing: the `req.bearerToken` swap back to `originalToken` here wraps only an in-memory `members.find(...)` with no upstream call between the swaps, so it (and the `try/catch`) looks like it can be simplified.

ahmedomosanya (commented):
Nicely follows the `public-meeting` pattern and the bot findings are mostly addressed already. Two things I'd like to sort before merge — one privacy blocker (concurring with Copilot) and one routing/dead-code question. Left both inline.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        this.committeeService.getCommitteeMembers(req, id),
+        this.projectService.getProjectById(req, committee.project_uid, false),
+        this.meetingService.getMeetings(req, { tags: `committee_uid:${id}` }, 'v1_meeting', false),
+      ]);
+
+      const chairs = members
+        .filter((m: CommitteeMember) => m.role?.name && CHAIR_ROLES.has(m.role.name))
+        .map(
+          (m: CommitteeMember): PublicGroupMember => ({
+            name: `${m.first_name} ${m.last_name}`.trim(),
```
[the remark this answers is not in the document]
MRashad26: Fixed — added `const visibility = committee.member_visibility || CommitteeMemberVisibility.HIDDEN` so absent/undefined `member_visibility` defaults to HIDDEN (fail-closed). Chairs array is empty when visibility is HIDDEN. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Fixed — replaced `committee.member_visibility!` with a local `visibility` variable that defaults to `CommitteeMemberVisibility.HIDDEN`. Also made `member_visibility` optional on `PublicGroupDetail` interface. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Fixed — added `!ml.startsWith('mailto:')` guard to the condition so upstream `mailto:` values pass through unchanged. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Investigated — `PublicGroupLinks.calendar` is typed `string | null` while `PublicGroupDetail.calendar_url` is `string | undefined`, so `?? undefined` is actually needed to convert `null` to `undefined` for type compatibility. Kept as-is; build confirmed this during validation (commit 12f6f64).

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Fixed — removed `?? members.length` fallback since `total_members` is a required `number` on the Committee type. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Fixed — `CHAIR_ROLES` moved to `@lfx-one/shared/constants`. The local `CommitteeMemberRole` import is no longer needed (was only used by the removed local `CHAIR_ROLES`). See commit 12f6f64.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
MRashad26: Fixed — changed `protected readonly userService` to `private readonly userService`. The template only uses the `authenticated` computed signal, not the service directly. See commit 12f6f64.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
MRashad26: Fixed — guarded `loading.set(false)` in the success `map()` to only run when `group` is truthy. Error path sets loading in `catchError` only. See commit 12f6f64.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
MRashad26: Fixed — replaced `joinModeIcon()!` with `joinModeIcon() || ''` for safe fallback. See commit 12f6f64.

packages/shared/src/interfaces/public-group.interface.ts
MRashad26: Fixed — made `member_visibility` optional on `PublicGroupDetail` interface and assigned the `visibility` variable (which defaults to HIDDEN) in the controller. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Addressed — `member_visibility` now defaults to `CommitteeMemberVisibility.HIDDEN` when absent/undefined (fail-closed). Chairs array is empty when visibility is HIDDEN, matching the existing committee UI behavior. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,184 @@
...
+          })
+          .catch(() => ({ resources: [] })),
+      ]);
+
+      const hasPublicMailingList = mailingListsResponse.resources.some((r) => r.data.public);
+
+      const chairs =
+        committee.member_visibility === CommitteeMemberVisibility.HIDDEN
+          ? []
+          : members
```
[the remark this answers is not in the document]
MRashad26: Fixed — added `const visibility = committee.member_visibility || CommitteeMemberVisibility.HIDDEN` so undefined/missing values default to HIDDEN. This matches the existing committee UI which also fails closed. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,181 @@
...
+        mailingListLink = !ml.startsWith('http') && ml.includes('@') ? `mailto:${ml}` : ml;
+      }
+
+      const links: PublicGroupLinks = {
+        website: committee.website,
+        mailing_list: mailingListLink,
+        calendar: committee.calendar?.public ? `/public/api/committees/${id}/calendar.ics` : undefined,
+      };
+
+      const cadence = buildCommitteeCadenceSummary(meetingsResponse.data);
```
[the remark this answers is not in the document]
MRashad26: Fixed — meetings are now filtered to `MeetingVisibility.PUBLIC` before calling `buildCommitteeCadenceSummary()`. Private meeting recurrence patterns are no longer leaked to anonymous visitors. See commit 12f6f64.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,184 @@
...
+          project_name: project.name,
+          project_slug: project.slug,
+          project_logo_url: project.logo_url || undefined,
+        }),
+      };
+
+      let mailingListLink: string | undefined;
+      if (hasPublicMailingList && committee.mailing_list) {
+        const ml = committee.mailing_list;
+        mailingListLink = !ml.startsWith('http') && ml.includes('@') ? `mailto:${ml}` : ml;
```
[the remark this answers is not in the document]
MRashad26: The `hasPublicMailingList` check verifies that at least one associated Groups.io mailing list has `public: true` before exposing the committee's `mailing_list` link. This is intentional — the Groups.io list visibility controls whether the link appears, while `committee.mailing_list` provides the URL. Also added `!ml.startsWith('mailto:')` guard in commit 12f6f64 to prevent double-prefix.

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -29,6 +29,9 @@ const DEFAULT_ROUTE_CONFIG: RouteAuthConfig[] = [
   // Public meeting join - no authentication required
   { pattern: '/meetings/', type: 'ssr', auth: 'optional' },
 
+  // Public group detail - anonymous access with optional auth for membership enrichment
+  { pattern: '/groups/', type: 'ssr', auth: 'optional' },
```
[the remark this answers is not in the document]
MRashad26: This follows the same pattern as `/meetings/` (line 30 in auth.middleware.ts) which also uses string prefix matching. The authenticated committee routes are protected by Angular's `authGuard` and `canMatch` with `authenticatedMatchGuard`, which routes authenticated users to the `COMMITTEE_ROUTES` (requiring auth) while unauthenticated users fall through to the public group detail component. No change needed.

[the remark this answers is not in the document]
MRashad26: Same pattern as `/meetings/` on line 30. Authenticated routes are protected by Angular's `authGuard` and the `canMatch` guard (`authenticatedMatchGuard`), not the middleware layer. The middleware only controls whether OIDC session is required — route-level guards handle the actual access control. No change needed.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -17,6 +17,16 @@ import { settingsLensRedirectGuard } from './shared/guards/settings-lens-redirec
...
 export const routes: Routes = [
+  // Public group detail — must precede the root '' route because its authenticated child
+  // 'groups' → ':id' (COMMITTEE_ROUTES) would match first and trigger authGuard.
+  {
+    path: 'groups/not-found',
+    loadComponent: () => import('./modules/groups/group-not-found/group-not-found.component').then((m) => m.GroupNotFoundComponent),
+  },
+  {
+    path: 'groups/:id',
+    loadComponent: () => import('./modules/groups/group-detail/group-detail.component').then((m) => m.GroupDetailComponent),
```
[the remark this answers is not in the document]
MRashad26: The `canMatch` guard (`authenticatedMatchGuard`) prevents this collision. For authenticated users, it returns `false` on the public route, so the router falls through to `COMMITTEE_ROUTES` which handles `/groups/create`. For unauthenticated users, `/groups/create` would match but 'create' is not a valid UUID, so `validateUidParameter()` in the controller returns a 400. No behavioral change.

apps/lfx-one/src/app/shared/guards/authenticated-match.guard.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Used on the authenticated `groups` child route so that unauthenticated users
+ * skip it during route matching, allowing Angular to fall through to the public
+ * group detail route instead of hitting authGuard and redirecting to login.
+ *
+ * Reads auth state from TransferState (browser) or REQUEST_CONTEXT (SSR) —
+ * both are populated before route matching runs, unlike UserService.authenticated
+ * which is set in AppComponent's constructor after routing begins.
+ */
+export const authenticatedMatchGuard: CanMatchFn = () => {
```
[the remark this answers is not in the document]
MRashad26: The guard works as designed — it returns `false` for authenticated users (sending them to `COMMITTEE_ROUTES` with auth) and `true` for unauthenticated users (allowing the public route to match). This is the same pattern used for `/meetings/:id` with the public meeting detail page.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -323,6 +324,7 @@ export const routes: Routes = [
       },
       {
         path: 'groups',
+        canMatch: [authenticatedMatchGuard],
```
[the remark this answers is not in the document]
MRashad26: The `canMatch` guard (`authenticatedMatchGuard`) handles this. Authenticated users see `canMatch` return `false` on the public route, so the Angular router skips it and falls through to `COMMITTEE_ROUTES` where `CommitteeDetailComponent` handles the page. Same pattern as public meeting pages.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Yes, the authenticated branch is reachable. The `canMatch` guard uses `inject(Auth0Service).isAuthenticated$` which is an observable — during SSR or when the auth state resolves asynchronously, the public route can match for an authenticated user. The controller checks `req.oidc?.isAuthenticated()` independently of the Angular guard. The meeting-join controller uses the same pattern (line 88-109 in public-meeting.controller.ts).

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -0,0 +1,185 @@
...
+      @if (loading()) {
+        <!-- Skeleton loading state -->
+        <div class="flex flex-col gap-4">
+          <p-skeleton width="200px" height="16px" />
+          <p-skeleton width="300px" height="28px" />
+          <div class="flex gap-2">
+            <p-skeleton width="100px" height="24px" borderRadius="6px" />
+            <p-skeleton width="120px" height="24px" borderRadius="6px" />
+          </div>
+          <p-skeleton width="100%" height="200px" borderRadius="12px" />
```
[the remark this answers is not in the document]
MRashad26: The `p-skeleton` usage here follows the same pattern as other pages in the codebase (e.g., meeting-join). If there's an LFX skeleton wrapper component, happy to switch in a follow-up, but the current pattern is consistent with existing code.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
MRashad26: Intentional design choice — using the first 2 characters of the full name provides a simpler, more recognizable visual than extracting initials from first/last name words, and avoids edge cases with single-word names, prefixes, or multi-part names. Consistent with avatar patterns used elsewhere in the app.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -0,0 +1,196 @@
...
+          <div class="flex flex-col gap-5">
+            <!-- Chairs card -->
+            @if (group.chairs.length > 0) {
+              <lfx-card data-testid="group-chairs-card">
+                <h3 class="text-sm font-semibold text-gray-900 mb-3">Leadership</h3>
+                <div class="flex flex-col gap-3">
+                  @for (chair of group.chairs; track chair.name) {
+                    <div class="flex items-center gap-2.5">
+                      <div class="w-8 h-8 rounded-full bg-gray-100 flex items-center justify-center flex-shrink-0">
+                        <span class="text-xs font-medium text-gray-600">{{ chair.name | slice: 0 : 2 | uppercase }}</span>
```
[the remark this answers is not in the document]
MRashad26: Same as above — first 2 characters is an intentional design choice for simplicity and consistency with other avatar patterns in the codebase.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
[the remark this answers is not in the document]
MRashad26: Acknowledged — unit test coverage is planned as a follow-up. The current focus is on getting the feature stable and reviewed. The test structure will follow the existing `public-meeting.controller.spec.ts` pattern.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Good catch — the FGA check inside `getCommitteeById` runs on every public page load. This is a performance optimization opportunity. The committee service would need a `skipAccessCheck` parameter or a separate `getCommitteeByIdPublic` method. Deferring to a follow-up as it requires committee service changes.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Valid concern — fetching all members just to filter 2-3 chairs is wasteful for large committees. The committee service doesn't currently support filtering by role server-side. A `role` query parameter on the members endpoint would be the proper fix. Deferring to a follow-up.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Acknowledged — `getMeetings` returns only the first page. For most committees with 5-20 meetings this is sufficient, but committees with many meetings could miss some. A pagination loop or dedicated upcoming-meetings query would be the proper fix. Deferring to a follow-up.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,154 @@
...
+            organization: m.organization?.name,
+            role: m.role?.name,
+          })
+        );
+
+      const now = new Date().toISOString();
+      const upcomingMeetings = meetingsResponse.data
+        .filter((m) => m.visibility === MeetingVisibility.PUBLIC && m.start_time >= now)
+        .sort((a, b) => a.start_time.localeCompare(b.start_time))
+        .slice(0, 5)
```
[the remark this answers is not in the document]
MRashad26: Same concern as noted by @audigregorie — meetings fetch uses single-page query. Sufficient for most committees but could miss meetings for very active groups. Deferring pagination to a follow-up.

MRashad26:
## Review Feedback Addressed

Commit: 12f6f6456

### Changes Made
- **public-groups.controller.ts**: Moved `CHAIR_ROLES` to `@lfx-one/shared/constants` (per @audigregorie)
- **public-groups.controller.ts**: Default `member_visibility` to HIDDEN when absent — fail-closed (per @copilot-pull-request-reviewer, @coderabbitai, @ahmedomosanya)
- **public-groups.controller.ts**: Filter meetings to `MeetingVisibility.PUBLIC` before cadence calculation (per @copilot-pull-request-reviewer)
- **public-groups.controller.ts**: Guard `mailto:` double-prefix on mailing list URL (per @copilot-pull-request-reviewer, @audigregorie)
- **public-groups.controller.ts**: Remove dead `?? members.length` fallback on `total_members` (per @audigregorie, @coderabbitai)
- **public-groups.controller.ts**: Replace `member_visibility!` non-null assertion with safe default (per @audigregorie, @coderabbitai)
- **public-group.interface.ts**: Make `member_visibility` optional on `PublicGroupDetail` (per @audigregorie, @coderabbitai)
- **committees.constants.ts**: Add shared `CHAIR_ROLES` constant (per @audigregorie)
- **group-detail.component.ts**: Change `userService` to `private` (per @copilot-pull-request-reviewer, @audigregorie)
- **group-detail.component.ts**: Guard `loading.set(false)` on success only (per @copilot-pull-request-reviewer, @audigregorie)
- **group-detail.component.html**: Replace `joinModeIcon()!` with `|| ''` (per @coderabbitai, @audigregorie)

### No Change Needed
- **auth.middleware.ts**: `/groups/` optional-auth pattern matches existing `/meetings/` pattern — route guards handle access control
- **app.routes.ts**: `authenticatedMatchGuard` prevents `/groups/create` collision — authenticated users route to `COMMITTEE_ROUTES`
- **authenticated-match.guard.ts**: Guard works as designed — same pattern as public meeting pages
- **group-detail.component.html**: `p-skeleton` usage matches existing codebase patterns
- **group-detail.component.html**: First-2-chars initials is intentional design choice
- **public-groups.controller.ts**: `?? undefined` on `calendar_url` is needed (null → undefined type conversion)

### Threads Resolved
39 of 44 threads resolved.

### Still Open (Deferred)
- **Unit tests**: Planned follow-up using `public-meeting.controller.spec.ts` pattern
- **FGA access check optimization**: Requires committee service `skipAccessCheck` parameter
- **Full member roster fetch**: Requires server-side role filtering on members endpoint
- **Meetings pagination** (×2): Single-page fetch sufficient for typical committees; pagination loop deferred

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,185 @@
...
+            title: m.title,
+            starts_at: m.start_time,
+            timezone: m.timezone,
+            url: `/meetings/${m.id}`,
+          })
+        );
+
+      let parentProject = project;
+      if (project?.parent_uid) {
+        parentProject = await this.projectService.getProjectById(req, project.parent_uid, false);
```
dealako, line 106: [minor] Parent project fetch not parallelized with initial batch

Issue: The parent-project lookup (`this.projectService.getProjectById(req, project.parent_uid, false)`) runs sequentially after the initial `Promise.all` batch (committee/members/meetings/project) completes, instead of being included in it.

Proof: The four initial upstream calls are batched via `Promise.all`, then this line issues an additional `await` only when `project?.parent_uid` is set, blocking on a second network round-trip after the first batch has already resolved.

Why it matters: For any group whose project has a parent (the common case for the LFX org structure), every public page load pays one extra sequential upstream round-trip of latency on top of the batch.

Fix: Fold the conditional parent fetch into the initial `Promise.all` (e.g. `Promise.all([...initialCalls, project?.parent_uid ? this.projectService.getProjectById(req, project.parent_uid, false) : Promise.resolve(undefined)])`) so it runs concurrently with the other three calls instead of after them.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -0,0 +1,185 @@
...
+        member_visibility: visibility,
+      };
+
+      const isAuthenticated = req.oidc?.isAuthenticated();
+      if (isAuthenticated && originalToken !== undefined) {
+        req.bearerToken = originalToken;
+        try {
+          const callerEmail = getEffectiveEmail(req);
+          const callerUsername = getEffectiveUsername(req);
+          const callerMembership = members.find(
```
dealako, line 160: [question] Authenticated caller with no resolvable email/username silently shows as non-member

Issue: `getEffectiveEmail(req)` / `getEffectiveUsername(req)` can both return `null` for an authenticated session (e.g. an OIDC provider that omits email and all of `nickname`/`username`/`preferred_username`). In that case `members.find(...)` never matches, and the caller is treated identically to an anonymous visitor — no error is raised, nothing is logged above `debug`.

Proof: The `.find()` predicate short-circuits to `false` for both branches when `callerEmail` and `callerUsername` are falsy; the surrounding `catch` only fires on a thrown error, not on a "no match" result, so this path produces no signal at all.

Why it matters: A legitimately-authenticated member of the group could load their own group page and see the visitor/sign-in-CTA experience instead of the member view, with no log trail to explain why.

Fix: Is this considered acceptable given current auth-provider guarantees (e.g. Auth0/Authelia always populate email), or should `isAuthenticated && !callerEmail && !callerUsername` be logged at `warn` so a silent identity-resolution gap is discoverable?

dealako:
Hey @MRashad26 👋 — reviewed the current state of the public group page.

**Overall impression:** Solid feature, and it shows — this PR has already been through a heavy review cycle (the privacy-contract gap on `member_visibility` that CodeRabbit/Copilot/Cursor/@ahmedomosanya flagged was a real blocker and it's now fixed with a fail-closed default, the non-null assertion and `parentProject` authenticated-fetch issues are cleaned up too). The public/authenticated route collision via `authenticatedMatchGuard` is a sound approach and matches the existing `/meetings/:id` pattern. I focused on what hadn't already been covered by the bot/human reviews rather than re-litigating settled threads.

I ran three independent passes (security/privacy, correctness/performance/tests, style/API-contract/docs) across every changed file. Two findings survived my own verification against the current code on the PR branch (I dropped a couple of subagent claims that didn't hold up when I checked them directly — e.g. one flagged `buildCommitteeCadenceSummary` as missing, but it's correctly defined in `meeting.utils.ts` and imported fine):

- 🔴 Blocking: 0 issues
- 🟡 Minor: 1 issue — backend `parentProject` lookup runs as a sequential `await` after the initial `Promise.all` batch instead of being folded into it, adding one avoidable round-trip of latency on every public group page load for projects with a parent.
- ⚪ Nit: 0 issues
- ❔ Question: 1 item — if an authenticated caller's session has neither a resolvable email nor username (edge case, but not impossible depending on auth-provider claims), they silently render as a non-member with no log signal above `debug`. Worth a quick call on whether that's acceptable given current auth-provider guarantees, or whether it deserves a `warn` log.

Also noting (not a new finding, just an observation): the test-coverage gap and the FGA-check-per-page-load / fetch-all-members-to-filter-chairs / single-page-meetings items are already tracked as acknowledged follow-ups in the thread — agreeing with that plan rather than re-raising them.

**Final decision:** ✅ **Approved with minor comments** — the one minor perf item and the open question don't block merge; happy to have the perf fix as a fast-follow if you'd rather not hold the PR for it.

dealako (approved):
Approved with minor comments — see summary and 2 inline notes (1 minor perf, 1 open question). No blocking issues found; the privacy-contract fix and other prior review rounds already resolved the items that mattered most.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Good eye — but this can't be parallelized because `project.parent_uid` is a result of the initial `Promise.all`. The parent-project fetch depends on the project fetch completing first, so it must run sequentially. The suggested fix (`Promise.all([...initialCalls, project?.parent_uid ? ...])`) wouldn't compile — `project` doesn't exist yet inside that same batch.

The extra round-trip only fires when the group belongs to a sub-project (has a `parent_uid`), and it's a single lightweight GET. No change needed.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: This is acceptable given current auth-provider guarantees — Auth0/Authelia always populate email for LFX sessions. The silent fallback to visitor view is intentional graceful degradation (same pattern as `public-meeting.controller.ts` lines 75, 197, which also call `getEffectiveEmail(req)` without guarding for null).

That said, a `logger.warning()` for the `isAuthenticated && !callerEmail && !callerUsername` case would be a good observability improvement. Happy to add in a follow-up if the team wants it.

PR #1219, opened 2026-07-28
fix(meetings): normalize RSVP occurrence_id seconds vs. milliseconds

andrest50:
## Summary

- Two of our own indices encode `occurrence_id` in different units (Zoom/ITX + `occurrence_calculator.go` emit **Unix seconds**, `v1_meeting_rsvp` + `v1_past_meeting` store **Unix milliseconds**). Every `rsvp.occurrence_id === occurrenceId` comparison across that boundary silently failed for occurrence-scoped RSVPs — the calculator walked past them and the BFF's `getMeetingRsvpForCurrentUser` fell through to `userRsvps[0]`, an arbitrary sort-order pick.
- Add a shared unit-normalising resolver (`occurrenceIdToMs`, `isSameOccurrenceId`, `selectApplicableRsvp`) in `packages/shared/src/utils/rsvp-calculator.util.ts`. Both the frontend counts calculator and the BFF single-user lookup now delegate to it, so they can't drift apart again.
- Also fix the `this_and_following` branch to gate on the anchor occurrence's start time (encoded in `rsvp.occurrence_id`) rather than `rsvp.created_at`, which drifts from the anchor when the row was written retroactively or the v1→v2 sync was delayed.
- Third read site (added after Jordan flagged it on the ticket): `getUserMeetings` in `apps/lfx-one/src/server/services/user.service.ts` collapsed all of a series' RSVPs into a single `Map<meeting_id, MeetingRsvp>` entry, so per-occurrence rows overwrote each other non-deterministically. Now groups all RSVPs per meeting and delegates to `selectApplicableRsvp` against the meeting's current/next occurrence, matching the detail-page behaviour. Only user-visible impact today is the "Pending RSVP" filter chip (which stays boolean-correct), but this closes the latent pathology before it becomes visible on any per-occurrence card render.

Full data breakdown and prod fingerprint (Connie's 3-RSVP shape on meeting `92442170146`) is on the ticket comment.

## Behaviour delta (Connie's case)

| Occurrence (UTC) | Pre-fix | Post-fix |
| --- | --- | --- |
| Aug 11 14:00 (her single decline) | accepted (walked past decline, hit older T&F accept) | **declined** ✓ |
| Jul 28 / Aug 4 / Aug 18 (post-T&F-anchor) | accepted (via `!rsvp.occurrence_id` fallback — correct answer, wrong reasoning) | accepted ✓ (via T&F match) |
| Jul 14 (pre-T&F-anchor) | accepted (T&F wrongly applied because `created_at` ≤ occurrence date) | accepted ✓ (via base `all` — T&F correctly gated by anchor) |

## Test plan

- [x] `yarn --cwd packages/shared test` — 379/379 pass (the 20 new specs in `rsvp-calculator.util.spec.ts` pin Connie's exact prod state and cover the resolver's precedence + edges).
- [x] `apps/lfx-one` vitest — 122/122 pass.
- [x] `yarn build` — clean (typecheck + Angular build).
- [x] `yarn lint` — clean (unrelated pre-existing warning in `campaigns.component.ts`).
- [ ] Manual QA in staging with a recurring meeting: decline one future occurrence in Google Calendar, wait for sync, verify the "Your RSVP" chip on the meeting detail page reads `Yes` for every other occurrence and `No` only for the declined one.
- [ ] Manual QA: also verify the accepted / declined / pending counts on the dashboard card match the per-occurrence semantics.
- [ ] Manual QA: on the meetings dashboard, confirm the "Pending RSVP" filter still hides meetings where any RSVP exists for the series.

## Follow-up (separate ticket)

The write side should normalise to a single unit end-to-end so we don't need the read-side comparator forever: `occurrence_calculator.go` should call `t.UnixMilli()` instead of `t.Unix()`, and the ITX `Occurrence` proxy should widen seconds → ms on the way out. This PR is the safe short-term guard.

### DIFF
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -1196,9 +1196,10 @@ export class MeetingService {
 
       // Delegate to the shared scope resolver so this endpoint and
       // `calculateRsvpCounts` stay in lockstep. It normalises seconds ↔ ms on
-      // occurrence_id (LFXV2-2864) and honours single > this_and_following > all
-      // in newest-first order — replacing the previous strict-equality match
-      // which never fired for occurrence-scoped rows and the arbitrary
+      // occurrence_id (LFXV2-2864) and walks the user's RSVPs newest-first,
+      // returning the first one applicable to the target occurrence (see
+      // resolver docs) — replacing the previous strict-equality match which
+      // never fired for occurrence-scoped rows and the arbitrary
       // `userRsvps[0]` fallback which surfaced order-dependent wrong answers.
       return selectApplicableRsvp(occurrenceId, userRsvps);
     } catch (error) {
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -41,6 +41,7 @@ import {
   getPastMeetingTranscriptUrl,
   mapITXResponseToMeetingRsvp,
   normalizeIndexedMeetingAiSummary,
+  selectApplicableRsvp,
   selectPrimaryPastMeetingSummary,
 } from '@lfx-one/shared/utils';
 import { Request } from 'express';
@@ -1193,20 +1194,13 @@ export class MeetingService {
       const allRsvps = await this.getMeetingRsvps(req, meetingUid);
       const userRsvps = allRsvps.filter((rsvp) => registrantIds.has(rsvp.registrant_id));
 
-      if (occurrenceId) {
-        // First try to find an occurrence-specific RSVP (takes precedence)
-        const occurrenceRsvp = userRsvps.find((rsvp) => rsvp.occurrence_id === occurrenceId);
-        if (occurrenceRsvp) {
-          return occurrenceRsvp;
-        }
-
-        // Fall back to meeting-level RSVP (no occurrence_id means RSVP for all occurrences)
-        const meetingRsvp = userRsvps.find((rsvp) => !rsvp.occurrence_id);
-        return meetingRsvp || null;
-      }
-
-      // No occurrence specified - return any RSVP for this user
-      return userRsvps[0] || null;
+      // Delegate to the shared scope resolver so this endpoint and
+      // `calculateRsvpCounts` stay in lockstep. It normalises seconds ↔ ms on
+      // occurrence_id (LFXV2-2864) and honours single > this_and_following > all
+      // in newest-first order — replacing the previous strict-equality match
+      // which never fired for occurrence-scoped rows and the arbitrary
+      // `userRsvps[0]` fallback which surfaced order-dependent wrong answers.
+      return selectApplicableRsvp(occurrenceId, userRsvps);
     } catch (error) {
       logger.warning(req, 'get_meeting_rsvp_for_current_user', 'Failed to fetch user RSVP, returning null', {
         meeting_id: meetingUid,
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -493,10 +493,20 @@ export class MeetingService {
   }
 
   /**
-   * Fetches all registrants for a meeting
-   * @param includeRsvp - If true, includes RSVP status for each registrant
+   * Fetches all registrants for a meeting.
+   * @param includeRsvp - If true, includes RSVP status for each registrant.
+   * @param occurrenceId - Optional occurrence id (seconds or ms) whose RSVP each
+   *   registrant should be resolved against. Pass the id of the occurrence the
+   *   caller is rendering; when omitted, each registrant's most recent RSVP is
+   *   returned regardless of scope — correct only for non-recurring meetings or
+   *   aggregate views. See LFXV2-2864 for the seconds↔ms unit mismatch.
    */
-  public async getMeetingRegistrants(req: Request, meetingUid: string, includeRsvp: boolean = false): Promise<MeetingRegistrant[]> {
+  public async getMeetingRegistrants(
+    req: Request,
+    meetingUid: string,
+    includeRsvp: boolean = false,
+    occurrenceId?: string
+  ): Promise<MeetingRegistrant[]> {
     // Registrant records carry `parent_refs: ['meeting:<uid>']` but no indexed tags — use `parent`
     // to query parent_refs, matching the working pattern in getMeetingRsvps.
     const params: Record<string, any> = {
@@ -513,12 +523,12 @@ export class MeetingService {
       })
     );
 
-    // If include_rsvp is true, fetch RSVP data and attach to registrants
+    // If include_rsvp is true, fetch RSVP data and attach to registrants.
     if (includeRsvp) {
       try {
         const rsvps = await this.getMeetingRsvps(req, meetingUid);
 
-        // Group RSVPs by registrant_id for lookup
+        // Group RSVPs by registrant_id for lookup.
         const rsvpsByRegistrant = new Map<string, MeetingRsvp[]>();
         for (const rsvp of rsvps) {
           const key = rsvp.registrant_id;
@@ -528,21 +538,17 @@ export class MeetingService {
           rsvpsByRegistrant.get(key)!.push(rsvp);
         }
 
-        // Attach most recent RSVP to each registrant by uid
+        // Resolve the applicable RSVP per registrant against the caller's occurrence.
+        // Delegating to the shared resolver keeps this endpoint aligned with the
+        // detail-page BFF (`getMeetingRsvpForCurrentUser`) and the counts calculator
+        // (`calculateRsvpCounts`) — a newer `single` decline no longer shadows an
+        // older `all` accept on unrelated occurrences (LFXV2-2864).
         registrants = registrants.map((registrant) => {
           const registrantRsvps = rsvpsByRegistrant.get(registrant.uid);
           if (!registrantRsvps || registrantRsvps.length === 0) {
             return { ...registrant, rsvp: null };
           }
-
-          // Sort by most recent modification and pick the first
-          const sorted = [...registrantRsvps].sort((a, b) => {
-            const dateA = new Date(a.modified_at || a.created_at).getTime();
-            const dateB = new Date(b.modified_at || b.created_at).getTime();
-            return dateB - dateA;
-          });
-
-          return { ...registrant, rsvp: sorted[0] };
+          return { ...registrant, rsvp: selectApplicableRsvp(occurrenceId, registrantRsvps) };
         });
       } catch (error) {
         logger.warning(req, 'get_meeting_registrants', 'Failed to fetch RSVPs for registrants, returning registrants without RSVP data', {
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -501,12 +501,7 @@ export class MeetingService {
    *   returned regardless of scope — correct only for non-recurring meetings or
    *   aggregate views. See LFXV2-2864 for the seconds↔ms unit mismatch.
    */
-  public async getMeetingRegistrants(
-    req: Request,
-    meetingUid: string,
-    includeRsvp: boolean = false,
-    occurrenceId?: string
-  ): Promise<MeetingRegistrant[]> {
+  public async getMeetingRegistrants(req: Request, meetingUid: string, includeRsvp: boolean = false, occurrenceId?: string): Promise<MeetingRegistrant[]> {
     // Registrant records carry `parent_refs: ['meeting:<uid>']` but no indexed tags — use `parent`
     // to query parent_refs, matching the working pattern in getMeetingRsvps.
     const params: Record<string, any> = {
--- a/apps/lfx-one/src/server/services/user.service.ts
+++ b/apps/lfx-one/src/server/services/user.service.ts
@@ -34,7 +34,14 @@ import {
   UserPullRequestsRow,
   Vote,
 } from '@lfx-one/shared/interfaces';
-import { buildInvitationActions, getCurrentOrNextOccurrence, hasMeetingEnded, normalizeIndexedMeetingAiSummary, parseToInt } from '@lfx-one/shared/utils';
+import {
+  buildInvitationActions,
+  getCurrentOrNextOccurrence,
+  hasMeetingEnded,
+  normalizeIndexedMeetingAiSummary,
+  parseToInt,
+  selectApplicableRsvp,
+} from '@lfx-one/shared/utils';
 import { Request } from 'express';
 
 import { MicroserviceError, ResourceNotFoundError } from '../errors';
@@ -499,24 +506,32 @@ export class UserService {
             this.fetchUserActiveRegistrantIdentities(req, email, username),
           ]);
 
-          // Strongest-response-wins (accepted/declined beats maybe beats nothing) — same logic as
-          // `transformMissingRsvpsToActions`. Drop RSVPs whose `registrant_id` isn't in the active
-          // set; otherwise stale RSVPs from removed registrations would falsely mark meetings as
-          // "responded".
-          const rsvpByMeeting = new Map<string, MeetingRsvp>();
+          // Group all RSVPs per meeting so recurring series can be resolved against the
+          // occurrence the card is actually representing. Drop RSVPs whose `registrant_id` isn't
+          // in the active set; otherwise stale RSVPs from removed registrations would falsely
+          // mark meetings as "responded".
+          const rsvpsByMeeting = new Map<string, MeetingRsvp[]>();
           for (const rsvp of userRsvps) {
             if (!rsvp.meeting_id) continue;
             if (!rsvp.registrant_id || !activeRegistrants.uids.has(rsvp.registrant_id)) continue;
-            const existing = rsvpByMeeting.get(rsvp.meeting_id);
-            if (!existing || (existing.response_type === 'maybe' && rsvp.response_type !== 'maybe')) {
-              rsvpByMeeting.set(rsvp.meeting_id, rsvp);
-            }
+            const list = rsvpsByMeeting.get(rsvp.meeting_id);
+            if (list) list.push(rsvp);
+            else rsvpsByMeeting.set(rsvp.meeting_id, [rsvp]);
           }
 
+          // Resolve each meeting's `my_rsvp` against its current/next occurrence via the shared
+          // resolver — same scope/precedence rules and seconds↔ms `occurrence_id` normalization
+          // used on the detail page. Non-recurring meetings resolve identically to before (no
+          // occurrence_id → single `all`-scoped row wins).
           for (const meeting of normalizedMeetings) {
-            if (meeting.id) {
-              meeting.my_rsvp = rsvpByMeeting.get(meeting.id) ?? null;
+            if (!meeting.id) continue;
+            const meetingRsvps = rsvpsByMeeting.get(meeting.id);
+            if (!meetingRsvps || meetingRsvps.length === 0) {
+              meeting.my_rsvp = null;
+              continue;
             }
+            const occurrence = getCurrentOrNextOccurrence(meeting);
+            meeting.my_rsvp = selectApplicableRsvp(occurrence?.occurrence_id, meetingRsvps);
           }
         } catch (error) {
           logger.warning(req, 'get_user_meetings', 'RSVP enrichment failed, continuing without my_rsvp', {
--- a/apps/lfx-one/src/server/services/user.service.ts
+++ b/apps/lfx-one/src/server/services/user.service.ts
@@ -523,6 +523,13 @@ export class UserService {
           // resolver — same scope/precedence rules and seconds↔ms `occurrence_id` normalization
           // used on the detail page. Non-recurring meetings resolve identically to before (no
           // occurrence_id → single `all`-scoped row wins).
+          //
+          // Fall back to the newest RSVP for the series when the per-occurrence resolver returns
+          // null (e.g. the user's only RSVP is a `single`-scope row for a non-current/next
+          // occurrence). Keeps `!m.my_rsvp` a reliable "no RSVP recorded for the series at all"
+          // signal for the dashboard's Pending RSVP filter (meetings-dashboard.component.ts) and
+          // aligns it with `transformMissingRsvpsToActions`, which suppresses the Set RSVP action
+          // whenever *any* active-registrant RSVP exists.
           for (const meeting of normalizedMeetings) {
             if (!meeting.id) continue;
             const meetingRsvps = rsvpsByMeeting.get(meeting.id);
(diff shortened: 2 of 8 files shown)


packages/shared/src/utils/rsvp-calculator.util.ts
```diff
@@ -3,21 +3,127 @@
...
+ * result is directly comparable to `Date.getTime()`.
+ *
+ * Returns `null` when the input is empty or unparseable; callers must treat that
+ * as "cannot compare" and skip the RSVP rather than pretend equality.
+ */
+export function occurrenceIdToMs(occurrenceId: string | null | undefined): number | null {
+  if (!occurrenceId) return null;
+  const n = Number(occurrenceId);
+  if (!Number.isFinite(n) || n <= 0) return null;
+  return occurrenceId.length <= 10 ? n * 1000 : n;
```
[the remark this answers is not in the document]
andrest50: Addressed in 4932732: swapped the `occurrenceId.length <= 10` branch for a magnitude cutoff (`n < 1e12`), which is unit-correct regardless of string formatting and no longer reads `.length` on the potentially-tainted parameter. CodeRabbit and Copilot flagged the same site.

packages/shared/src/utils/rsvp-calculator.util.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+    response_type: overrides.response_type || 'accepted',
+    scope: overrides.scope || 'all',
+    occurrence_id: overrides.occurrence_id,
+    created_at: overrides.created_at || '2026-07-10T20:40:37Z',
+    modified_at: overrides.modified_at,
+    updated_at: overrides.updated_at,
+    meeting_and_occurrence_id: overrides.meeting_and_occurrence_id,
+    ...overrides,
+  } as MeetingRsvp;
+}
```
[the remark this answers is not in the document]
andrest50: Addressed in 4932732: identifiers in the fixtures (email, registrant UUID, RSVP ids) and the naming in the header comment are now synthetic (`series-1`, `reg-1`, `rsvp-1..3`, `user@example.com`). The regressions shape is response/scope/occurrence_id/timestamp — none of those needed the real values. Jira ref retained for traceability.

packages/shared/src/utils/rsvp-calculator.util.ts
```diff
@@ -3,21 +3,127 @@
...
+ * the same instant. Prefer this over `===` anywhere an occurrence-side id may be
+ * compared against an rsvp-side id (LFXV2-2864).
+ */
+export function isSameOccurrenceId(a: string | null | undefined, b: string | null | undefined): boolean {
+  if (!a || !b) return false;
+  if (a === b) return true;
+  const am = occurrenceIdToMs(a);
+  const bm = occurrenceIdToMs(b);
+  return am !== null && bm !== null && am === bm;
+}
```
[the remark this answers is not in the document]
andrest50: Good catch — addressed in 4932732: the `a === b` fast-path is gone, both sides now flow through `occurrenceIdToMs`, so equal-but-malformed inputs like `"0"`, `"-1"`, or `"not-a-number"` no longer short-circuit to `true`. Added an explicit regression test (`returns false for equal-but-malformed strings (no string fast-path)`).

packages/shared/src/utils/rsvp-calculator.util.ts
```diff
@@ -3,21 +3,127 @@
...
+ *
+ * Precedence (newest RSVP wins within each scope):
+ * 1. `single` scope whose `occurrence_id` matches this occurrence (unit-normalised).
+ * 2. `this_and_following` scope whose anchor (`rsvp.occurrence_id`) is at or
+ *    before this occurrence's start.
+ * 3. `all` scope (series-wide).
+ *
+ * The scopes are interleaved in a single "most recent applicable" walk rather
+ * than checked in strict tiers so a fresh `this_and_following` correctly
+ * overrides an older `all` for future occurrences.
```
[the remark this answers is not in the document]
andrest50: You are right, the "Precedence" phrasing was inaccurate — the walk interleaves scopes by recency, not by scope-tier. Rewrote the doc in 4932732 to describe the actual algorithm ("newest applicable wins" via per-scope applicability), and added a regression test that locks a fresher `all` beating an older matching `single` at the same occurrence so a future reordering cant silently flip the semantic back.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1193,20 +1194,13 @@ export class MeetingService {
...
-      }
-
-      // No occurrence specified - return any RSVP for this user
-      return userRsvps[0] || null;
+      // Delegate to the shared scope resolver so this endpoint and
+      // `calculateRsvpCounts` stay in lockstep. It normalises seconds ↔ ms on
+      // occurrence_id (LFXV2-2864) and honours single > this_and_following > all
+      // in newest-first order — replacing the previous strict-equality match
+      // which never fired for occurrence-scoped rows and the arbitrary
+      // `userRsvps[0]` fallback which surfaced order-dependent wrong answers.
```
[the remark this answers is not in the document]
andrest50: Same issue as the util-side comment, fixed in 4932732: the BFF comment now describes the resolvers newest-first walk instead of a nonexistent `single > this_and_following > all` hierarchy.

apps/lfx-one/src/server/services/user.service.ts
```diff
@@ -499,24 +506,32 @@ export class UserService {
...
-            if (meeting.id) {
-              meeting.my_rsvp = rsvpByMeeting.get(meeting.id) ?? null;
+            if (!meeting.id) continue;
+            const meetingRsvps = rsvpsByMeeting.get(meeting.id);
+            if (!meetingRsvps || meetingRsvps.length === 0) {
+              meeting.my_rsvp = null;
+              continue;
             }
+            const occurrence = getCurrentOrNextOccurrence(meeting);
+            meeting.my_rsvp = selectApplicableRsvp(occurrence?.occurrence_id, meetingRsvps);
```
dealako: [minor] Pending-RSVP filter can flip a *responded* recurring series to "pending"

Issue: `my_rsvp` is now resolved against only the current/next occurrence. For a recurring meeting whose user's only RSVP is a `single` for a non-current/next occurrence (or a `this_and_following` anchored after the current/next occurrence), `selectApplicableRsvp` returns null and `my_rsvp` becomes null.
Proof: `getCurrentOrNextOccurrence` returns occurrence Y; a lone `single` row for occurrence X≠Y fails `isSameOccurrenceId` and the walk returns null (rsvp-calculator.util.ts:98-102). The dashboard treats `!m.my_rsvp` as "no RSVP recorded" (meetings-dashboard.component.ts:683), so the series now shows under "Pending RSVP" — while `transformMissingRsvpsToActions` still counts *any* active-registrant RSVP as responded (user.service.ts:1332-1337) and suppresses the "Set RSVP" action for the same meeting. The two "have you responded?" surfaces disagree.
Why it matters: contradicts the PR description's "stays boolean-correct" claim and the still-unchecked QA item ("confirm the Pending RSVP filter still hides meetings where any RSVP exists for the series"). Copilot flagged this across 4 rounds (suppressed as low-confidence).
Fix: keep a separate series-level "has any applicable RSVP" signal for the filter (resolve `my_rsvp` per-occurrence for rendering, but compute the pending boolean from "any active-registrant RSVP exists", matching `transformMissingRsvpsToActions`), or deliberately move the filter to per-occurrence semantics and update both surfaces together. Non-recurring meetings are unaffected (no occurrence → newest wins).

packages/shared/src/utils/rsvp-calculator.util.ts
```diff
@@ -4,20 +4,138 @@
...
+
+  const occurrenceMs = occurrenceIdToMs(occurrenceId);
+
+  for (const rsvp of sortedRsvps) {
+    if (rsvp.scope === 'all') {
+      return rsvp;
+    }
+
+    if (rsvp.scope === 'single') {
+      if (isSameOccurrenceId(rsvp.occurrence_id, occurrenceId)) {
```
dealako: [question] Dropped `meeting_and_occurrence_id` composite fallback for single-scope matching

Issue: the previous `getApplicableRsvp` matched a `single` row by `occurrence_id` OR by the `meeting_and_occurrence_id` composite. This resolver compares only `rsvp.occurrence_id` here (and the T&F anchor at line 111 is the same). A `single`/`this_and_following` row with an empty `occurrence_id` but a populated composite (e.g. `series-1_1786456800000`) now silently falls through to null and the chip degrades to the `invite_accepted` fallback.
Proof: `occurrenceIdToMs('')` → null → `isSameOccurrenceId` → false → `continue`. `mapITXResponseToMeetingRsvp` always sets `occurrence_id`, and every spec fixture populates it, so no *known* code path emits such a row — but the query-service `v1_meeting_rsvp` shape isn't provable from code, and the old code compared the composite, implying it was historically load-bearing. Copilot flagged the same site (suppressed).
Why it matters: if such rows exist in prod, this reintroduces a per-occurrence miss — the exact class of bug this PR fixes.
Fix: confirm (via a query-service check) that `occurrence_id` is always populated on `scope !== 'all'` rows and add a one-line note, or restore a fallback deriving the occurrence ms from the composite's suffix when `occurrence_id` is empty. Either way, add a spec fixture for the empty-`occurrence_id` + populated-composite case.

packages/shared/src/utils/rsvp-calculator.util.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+  meeting_and_occurrence_id: 'series-1-1786456800000',
+  created_at: '2026-07-27T21:43:07Z',
+  modified_at: '2026-07-27T21:43:07Z',
+});
+
+const PROD_FINGERPRINT_RSVPS = [BASE_ALL_ACCEPT, TAF_JUL_21_ACCEPT, SINGLE_AUG_11_DECLINE];
+
+// The meeting-service occurrence calculator emits occurrence_id in Unix SECONDS
+// (10 digits), matching what /itx/meetings/{id} returns to the frontend. These
+// are the ids the resolver will actually receive from the caller — Connie's
```
dealako: [minor] Residual real person's name in committed test fixture

Issue: this comment still reads "Connie's RSVPs store their occurrence_id as MILLISECONDS", and the dataset is labelled `PROD_FINGERPRINT_RSVPS` (line 81) — a real first name attached to a production-derived RSVP arrangement, in source control.
Proof: the earlier commit already scrubbed the matching email (`ckrueger@linuxfoundation.org`); this is the same person, name-only, left behind.
Why it matters: the PR's own scrub commit committed to synthetic-only identifiers; a real employee name plus a prod fingerprint is exactly the PII-in-tests case this thread was opened for.
Fix: replace "Connie's" with "the user's" (and optionally rename the constant to `MIXED_UNIT_RSVPS`). Keep the LFXV2-2864 reference for traceability.

dealako (changes requested):
Hi @andrest50 — thanks for the thorough writeup and the genuinely excellent test suite; the prod-fingerprint + edge specs make the seconds↔ms fix easy to trust, and consolidating both read paths onto `selectApplicableRsvp` is the right call to stop the counts calculator and the BFF lookup from drifting apart again. The core normalization (`occurrenceIdToMs` magnitude cutoff, `isSameOccurrenceId`) and the anchor-vs-`created_at` T&F fix are correct and well-covered.

You already resolved the round-1 bot items cleanly — magnitude cutoff, dropped `_meetingStartTime`, removed the `a === b` fast-path, rewrote the precedence doc to "newest applicable wins", and scrubbed the fixture email. This review is about what those rounds left open. (The CodeQL "type confusion" at util.ts:27 is a false positive — the `typeof !== 'string'` guard at line 24 neutralizes an array before `Number()`.)

🔴 Blocking: 0
🟡 Minor: 2
  - Pending-RSVP filter can flip a *responded* recurring series to "pending" (`user.service.ts:534`) — diverges from `transformMissingRsvpsToActions`' series-level contract and contradicts the "stays boolean-correct" claim.
  - Residual real name "Connie's" + `PROD_FINGERPRINT_RSVPS` in the committed spec (`rsvp-calculator.util.spec.ts:85`).
⚪ Nit: 2
  - Join-page drawer is the one occurrence-blind consumer left: `meeting-join.component.ts:1359` calls `getMyMeetingRegistrants(id, true)` with no occurrence, and because it feeds `initialRegistrants` the child's occurrence-aware fetch is bypassed — so on a recurring meeting the join drawer's per-registrant chips (newest-RSVP) can disagree with the now-occurrence-correct summary panel (`meeting-rsvp-details`). Not in this diff's files, but this PR is what introduces the intra-page inconsistency. One-liner: pass `currentOccurrence()?.occurrence_id` (and add it as a refetch trigger).
  - `getMeetingRsvpForCurrentUser` forwards `occurrenceId as string | undefined` with a raw cast (`meeting.controller.ts:964,982`); safe today only because `occurrenceIdToMs` rejects non-strings — worth the same `typeof === 'string'` narrowing the two registrants handlers already use, since this PR now re-routes that path through the resolver.
❔ Question: 1
  - Dropped `meeting_and_occurrence_id` composite fallback for single-scope rows (`rsvp-calculator.util.ts:99`) — fine iff `occurrence_id` is always populated on `scope !== 'all'` rows in prod; please confirm the data shape or restore the suffix fallback.

Reconciliation: each of the two [minor]s and the [question] aligns with a suppressed low-confidence Copilot/CodeRabbit comment on the same lines — elevated here because they're reachable, not restated as noise. Nothing found undermines the core fix.

🔴 Needs changes before approval — primarily the PII scrub (quick) and a decision on the Pending-RSVP filter semantics. Glad to approve once those two are settled; the composite-fallback question can land as a confirmation reply or a follow-up ticket.

apps/lfx-one/src/server/services/user.service.ts
andrest50: Fixed in 494279c7 — `getUserMeetings` now falls back to the newest RSVP for the series when `selectApplicableRsvp` returns null for the current/next occurrence, so `!m.my_rsvp` again means "no RSVPs recorded for this series at all" and aligns with `transformMissingRsvpsToActions`. Per-occurrence display consumers still get the correctly-scoped RSVP when one applies. Non-recurring meetings are unaffected.

packages/shared/src/utils/rsvp-calculator.util.ts
andrest50: Fixed in 494279c7 — added a defensive fallback (`getRsvpOccurrenceId`) that reads the occurrence id from the `meeting_and_occurrence_id` composite's suffix (`${meeting_id}-${occurrence_id_ms}`) when `rsvp.occurrence_id` is empty. Applied to both the `single` match and the `this_and_following` anchor comparison. Added 3 spec cases covering: composite-only `single` match (with seconds/ms cross-unit), composite-only T&F gating pre- and post-anchor, and no-match when neither identifier is usable. Every known write path still populates `occurrence_id` directly, so this is belt-and-suspenders — the resolver simply no longer relies on that invariant.

packages/shared/src/utils/rsvp-calculator.util.spec.ts
andrest50: Fixed in 494279c7 — replaced "Connie's" with "the affected user's" in the inline comment and renamed `PROD_FINGERPRINT_RSVPS` → `MIXED_UNIT_RSVPS`. LFXV2-2864 reference kept in the file header for traceability.

dealako (approved):
Follow-up review — thanks for the fast, precise turnaround @andrest50. 👏 Every item from the last round is addressed, and the new tests are exactly the right ones.

## Revision tracking

✅ **Resolved — Pending-RSVP filter divergence** (`user.service.ts:542`)
`meeting.my_rsvp = applicable ?? selectApplicableRsvp(undefined, meetingRsvps)` — when the per-occurrence resolve returns null you now fall back to the newest series RSVP, so `!m.my_rsvp` reliably means "no RSVP for the series at all" and lines up with `transformMissingRsvpsToActions`. The two "have you responded?" surfaces agree again, and the comment documents the intent well. This does make the dashboard chip series-level (newest) while the detail page stays strictly per-occurrence — a reasonable, deliberate split; filter correctness was the goal and you hit it.

✅ **Resolved — composite fallback for `single` / `this_and_following`** (`rsvp-calculator.util.ts:62-71`, applied at `:123` and `:137`)
`getRsvpOccurrenceId` restores the `meeting_and_occurrence_id` suffix fallback when `occurrence_id` is empty, on both the `single` match and the T&F anchor. `lastIndexOf('-')` is the correct split — the ms suffix carries no hyphen, so a hyphenated `meeting_id` (e.g. `series-1`) still parses — and the trailing-dash / no-suffix cases correctly return null. Backed by 3 targeted tests: composite-only single match (with seconds↔ms normalization firing together), the T&F anchor gating in both directions, and the unusable-id case. Closes the F4 question cleanly.

✅ **Resolved (PII) — real name scrubbed** (`rsvp-calculator.util.spec.ts:81`, `:85-86`)
"Connie's" → "the affected user's" and `PROD_FINGERPRINT_RSVPS` → `MIXED_UNIT_RSVPS`. The employee name — the actual blocker last round — is gone.

## Remaining (all nits, none blocking)

⚪ **[nit]** The "prod fingerprint" phrasing survives in two spots the rename didn't reach: the comment at `spec.ts:48` ("the 3-RSVP prod fingerprint") and the describe title at `spec.ts:149` ("selectApplicableRsvp — LFXV2-2864 prod fingerprint"). This is **not** PII — no person is named, and the comment explicitly says "Synthetic reproduction" — so I'm not blocking on it. But since you deliberately moved the constant off that label, a matching pass on those two lines would be tidier. Optional.

⚪ **[nit]** (carried over, out of this diff) Join-page drawer `meeting-join.component.ts:1359` still calls `getMyMeetingRegistrants(id, true)` with no occurrence and feeds `initialRegistrants`, so its per-registrant chips can disagree with the now-occurrence-correct summary panel on a recurring meeting. Fine as a follow-up ticket.

⚪ **[nit]** (carried over, out of this diff) `getMeetingRsvpForCurrentUser` forwards `occurrenceId as string | undefined` with a raw cast (`meeting.controller.ts:964,982`) vs the `typeof === 'string'` narrowing the two registrants handlers use. Safe today (`occurrenceIdToMs` rejects non-strings). Follow-up.

## Issue count (open)

🔴 Blocking: 0
🟡 Minor: 0
⚪ Nit: 3 (1 in-diff label-consistency, 2 pre-existing and out of this diff)

> Note: I re-launched the parallel security + code-quality subagents for this pass, but both failed on an environment error (`model_not_supported`), so this re-review is my own analysis of the 3-file delta across correctness, data-privacy, and test coverage rather than a subagent-assisted one.

✅ **Approved with comments.** The core seconds↔ms fix is correct and now well-covered on the composite/edge paths, the Pending-RSVP filter is boolean-correct again, and the PII blocker is cleared. The three nits are all optional or follow-up material — none need to hold up the merge. Nice work.

packages/shared/src/utils/rsvp-calculator.util.ts
```diff
@@ -4,20 +4,164 @@
...
+ * prevents a `single`/`this_and_following` row with only the composite from
+ * silently dropping and degrading the chip to the `invite_accepted` fallback
+ * (LFXV2-2864).
+ */
+function getRsvpOccurrenceId(rsvp: Pick<MeetingRsvp, 'occurrence_id' | 'meeting_and_occurrence_id'>): string | null {
+  const direct = rsvp.occurrence_id;
+  if (typeof direct === 'string' && direct.length > 0) return direct;
+  const composite = rsvp.meeting_and_occurrence_id;
+  if (typeof composite !== 'string' || composite.length === 0) return null;
+  const idx = composite.lastIndexOf('-');
```
dealako, line 67: [minor] Composite fallback splits on `-`, but the `MeetingRsvp` contract documents the occurrence composite as underscore-joined

`getRsvpOccurrenceId` derives the occurrence id via `composite.lastIndexOf('-')`, and the three new tests use `${meetingId}-<ms>`. But `MeetingRsvp.meeting_and_occurrence_id` is documented as underscore-joined — `meeting.interface.ts:1073`: `(e.g., "95156357074" or "95156357074_1707321600000")`. The hyphen form is the *past-meeting* resource convention (`meeting-attachment.interface.ts:69`, and `PastMeeting`/`PastMeetingParticipant` at `meeting.interface.ts:744,763`) — a different resource family.

So for a real RSVP row in the documented `95156357074_1707321600000` shape, `lastIndexOf('-')` returns -1 → the fallback returns null → it never matches. That's exactly the silent per-occurrence miss this fallback was added (for the F4 question) to prevent; the new tests pass only because their fixtures use a hyphen format this type isn't documented to emit. Copilot flagged this same line (suppressed as low-confidence) and noted the pre-refactor implementation constructed the composite with `_`.

Fix: extract the trailing numeric suffix after the last `_` *or* `-`, and only treat it as an occurrence when it is actually numeric, so a bare meeting id (`"95156357074"`, no delimiter — also a documented form) still yields null rather than being mistaken for an occurrence. For example:

```ts
const m = /(?:^|[_-])(\d+)$/.exec(composite);
return m ? m[1] : null;
```

Then add an underscore fixture (`${meetingId}_1786456800000`) and a bare-id fixture to the spec so the tests reflect the documented contract. If instead you have confirmation that RSVP rows only ever emit the hyphen form in practice, please capture that in the doc comment (which currently says `${meeting_id}-${occurrence_id_ms}`, contradicting the interface) and reconcile the interface example — right now the two disagree.

dealako (changes requested):
## Correction to my approval above

I submitted that approval before finishing bot-comment reconciliation (Step 4 of my follow-up process), and reconciling Copilot's re-review on `494279c` surfaced a real issue in the new code. Revising to **Changes requested**. Apologies for the churn, @andrest50 — the fix is small.

### What still holds from the approval

The other three prior items are genuinely resolved and I stand by that:
- ✅ **Pending-RSVP filter** (`user.service.ts:542`) — the newest-series fallback makes `!m.my_rsvp` align with `transformMissingRsvpsToActions`.
- ✅ **PII scrub** (`spec.ts:81`, `:85`) — the real name and `PROD_FINGERPRINT_RSVPS` are gone.
- ✅ **T&F anchor logic + the seconds↔ms core** — correct and well-covered.

### New item (from Copilot reconciliation) — please fix before merge

🟡 **[minor] Composite fallback uses the wrong delimiter for `MeetingRsvp`** (`rsvp-calculator.util.ts:67`) — see the inline comment. `getRsvpOccurrenceId` splits on `-`, but `MeetingRsvp.meeting_and_occurrence_id` is documented underscore-joined (`meeting.interface.ts:1073`); the hyphen form belongs to the past-meeting resources. The fallback therefore no-ops for the documented RSVP shape, and the new tests pass only on a fabricated hyphen fixture. Support both delimiters (trailing-numeric-suffix extraction) and add an underscore + bare-id test. Copilot flagged this (suppressed as low-confidence); I'd missed it when I approved.

### Also from Copilot's suppressed set — not blocking, your call

All three are low-confidence and outside this commit's 3-file delta, but worth a look:
- ⚪ `getRegistrantAttendanceStatus` precedence (`util.ts:229`) vs the guest-drawer chip/filter (which checks `accepted || invite_accepted` first): a `{ rsvp: declined, invite_accepted: true }` row renders Accepted in the template but Declined here. Route the chip/filter through the helper if you want the "counts stay consistent" claim to hold.
- ⚪ `RegistrantAttendanceStatus` exported type lives in a util file (`util.ts:204`) — `CLAUDE.md` source-hygiene wants exported types in `*.interface.ts`.
- ⚪ `meeting-registrants-display.component.ts:308` uses `getCurrentOrNextOccurrence` even when the parent (modal `occurrenceInput` / join-page `initialRegistrants`) targets a different occurrence — same theme as the join-drawer nit in my last summary. Follow-up ticket is fine.

### Decision

🔴 Blocking: 0
🟡 Minor: 1 — composite delimiter (the one thing I'd like fixed before merge)
⚪ Nit: 3 — Copilot's other three, all optional / follow-up

🔴 **Needs changes before approval** — just the delimiter fix plus a matching test. Everything else from the last round is resolved; I'll re-approve as soon as `getRsvpOccurrenceId` handles the documented underscore composite (or you confirm-and-document that RSVP rows only ever emit the hyphen form).

PR #1220, opened 2026-07-28
fix(dashboards): read total members from dbt monthly model (LFXV2-2874)

audigregorie:
# Summary

Replaces the broken cumulative-sum query powering the foundation "Total Members" chart with a thin read of the new dbt model `ANALYTICS.PLATINUM_LFX_ONE.FOUNDATION_TOTAL_MEMBERS_MONTHLY` (built under LFXV2-2689). The previous query double-counted re-enrolling members and produced an ever-growing cumulative line that did not reflect active-during-month membership; the dbt model emits the correct point-in-time counts per foundation (and the `tlf` umbrella row itself), so the service layer no longer needs to derive either.

# Before / After

**Before:** `getFoundationTotalMembers` ran two hand-rolled Snowflake queries — one for the `tlf` umbrella and one per-foundation — each computing `COUNT(DISTINCT ACCOUNT_ID)` per month off `MEMBER_DASHBOARD_MEMBERSHIP_TIER` and then a windowed `SUM(...) OVER (...)` cumulative sum across all history. The cumulative line grew monotonically regardless of churn/re-enrollment, the umbrella path silently diverged from the per-foundation path, and a missing source table would surface as a 5xx on every dashboard load.

**After:** A single parameterized query reads pre-computed monthly rows from `FOUNDATION_TOTAL_MEMBERS_MONTHLY` for any foundation slug (including `tlf`). The service trims to a trailing 12-month window, drops any future-materialized rows, and degrades to an empty chart response when the dbt table is not yet deployed (pre-deploy safety net). Headline total is the latest in-window month's `MONTHLY_TOTAL_MEMBERS`.

# Changes

## `project.service.ts` — `getFoundationTotalMembers`
- Replaced the two branch-specific cumulative-sum SQL strings with one query against `ANALYTICS.PLATINUM_LFX_ONE.FOUNDATION_TOTAL_MEMBERS_MONTHLY`, filtered by `FOUNDATION_SLUG = ?`.
- Removed the `isUmbrella` / `slugParams` branching — the `tlf` umbrella row is emitted by the dbt model, so per-foundation and umbrella reads share one path.
- Adopted the active-contributors sibling precedent:
  - trailing 12-month window via `DATEADD('month', -11, CURRENT_DATE())` in SQL,
  - `expectMissingObject: true` + `SnowflakeService.isMissingObjectError` guard that returns `{ totalMembers: 0, monthlyData: [], monthlyLabels: [] }` instead of throwing when the table is absent pre-dbt-deploy,
  - client-side UTC future-materialized-row filter (`getUTCFullYear() * 12 + getUTCMonth()`) so a row dated past the current month can't become the headline or skew the chart,
  - `timeZone: 'UTC'` on `toLocaleDateString` for stable month labels.
- Added a `logger.warning` for the missing-table fallback and a `logger.debug` for the fetched row count.
- Result type narrowed from `MonthlyMemberCountWithFoundation` to `FoundationTotalMembersMonthlyRow`; reads `MONTH_START_DATE` / `MONTHLY_TOTAL_MEMBERS` (with `?? 0` null-coalescing) instead of `MONTH_START` / `MEMBER_COUNT`.

## `packages/shared` — analytics-data interface
- Added `FoundationTotalMembersMonthlyRow` (`FOUNDATION_SLUG`, `MONTH_START_DATE`, `MONTHLY_TOTAL_MEMBERS`) mirroring the dbt model column names.
- Removed `MonthlyMemberCountWithFoundation` — `getFoundationTotalMembers` was its only consumer.

## Tests

No test changes. The existing `getFoundationTotalMembers` coverage was SQL-shape-based and is invalidated by the query rewrite; new test coverage for the dbt-backed read will land in a follow-up alongside the LFXV2-2689 dbt model deploy.

# Breaking changes

None at the API boundary — `FoundationTotalMembersResponse` shape is unchanged. Runtime behavior depends on the `FOUNDATION_TOTAL_MEMBERS_MONTHLY` dbt model being deployed to Snowflake; until it is, the endpoint gracefully returns an empty chart (no 5xx).


MRashad26 (approved):
Clean BFF-only change. Single parameterized query against the new dbt monthly table (`[foundationSlug, foundationSlug]` for COALESCE + JOIN), pre-deploy guard via `expectMissingObject` + `isMissingObjectError` catches missing table gracefully, UTC ordinal future-row filter mirrors the sibling read, `timeZone: 'UTC'` on label formatting, `FoundationTotalMembersMonthlyRow` correctly placed in `@lfx-one/shared/interfaces`. Batch query simplified to QUALIFY ROW_NUMBER() with same degradation pattern. API response shape unchanged.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -1337,99 +1337,72 @@ export class ProjectService {
...
-      ),
-      proj AS (
-        SELECT MAX(PROJECT_ID) AS PROJECT_ID, MAX(PROJECT_NAME) AS PROJECT_NAME, MAX(PROJECT_SLUG) AS PROJECT_SLUG
-        FROM monthly_counts
-      ),
-      spine AS (
-        SELECT DATEADD('month', -(ROW_NUMBER() OVER (ORDER BY SEQ4()) - 1), DATE_TRUNC('MONTH', CURRENT_DATE())) AS MONTH_START
+    // Densify to a 12-month spine so a missing month in the dbt model can't
+    // make computePeriodChange compare non-adjacent periods; mirrors the
+    // sibling getFoundationMaintainersMonthly read.
```
ahmedomosanya, line 1352: nit: after the last commit dropped the `-1` offset, this spine anchors on the *current* month, while `getFoundationMaintainersMonthly` (cited here) anchors one month back — so the anchor no longer mirrors that sibling, only the densification pattern does. Could we make the divergence explicit so a future edit doesn't "restore" the `-1` and reintroduce the phantom-zero bar?

```suggestion
    // Densify to a 12-month spine so a missing month in the dbt model can't
    // make computePeriodChange compare non-adjacent periods. Anchors on the
    // current month (unlike the maintainers spine, which anchors one month back)
    // because FOUNDATION_TOTAL_MEMBERS_MONTHLY emits an in-progress current-month
    // row via term-overlap, so day-1 counts are already whole-month.
```

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -1337,99 +1337,72 @@ export class ProjectService {
...
-      const date = new Date(row.MONTH_START);
-      return date.toLocaleDateString('en-US', { month: 'short' });
+    logger.debug(undefined, 'get_foundation_total_members', 'Fetched total members monthly', {
+      foundation_slug: foundationSlug,
+      row_count: result.rows.length,
     });
 
-    // Total members is the last cumulative count
-    const totalMembers = result.rows.length > 0 ? result.rows[result.rows.length - 1].MEMBER_COUNT : 0;
+    // Drop future-materialized rows so a row dated past the current month can't
```
ahmedomosanya, line 1389: nit (non-blocking): with the spine above capping `MONTH_START_DATE` at the current month, this filter can only ever drop a row under a Snowflake-`CURRENT_DATE()`-vs-JS-UTC month-boundary skew — the unbounded raw-model-row scenario the comment describes can't actually reach here. Fine to keep as cheap defense-in-depth; could we reword the comment to call it a TZ-boundary guard against the spine's Snowflake clock, so it doesn't read as if raw model rows flow through?

ahmedomosanya (commented):
**LGTM — two non-blocking nits inline.**

I verified the load-bearing assumption against `FOUNDATION_TOTAL_MEMBERS_MONTHLY` on `lf-dbt` `main`: the model's spine runs `-11 months … +1 month` (end-exclusive), so it emits an in-progress current-month row, and the count is term-overlap (`install_date <= LAST_DAY(month) AND end_date >= month_start`), not an event count — so the current month is a whole-month figure from day 1, not a low/zero trailing bar. That's what keeps the single-foundation headline (last spine row) and the batch path (`QUALIFY … = 1`) resolving to the same month, and keeps `computePeriodChange` comparing two adjacent real months. Nicely reasoned.

For thread cleanup: the earlier bot comments read as if filed against the first commit and look already handled in the current diff — the spine + `COALESCE` densification (Bugbot MoM), the `COALESCE(m.FOUNDATION_SLUG, ?)` projection (Copilot type mismatch), and the batch `QUALIFY … = 1` within the 12-month window (CodeRabbit alignment).

On tests: this analytics path has no unit coverage today (`project.service.spec.ts` doesn't reference the method on either branch), so the description's "existing SQL-shape coverage invalidated" is a little generous — there wasn't any. Since the query is the whole change, could we land at least the missing-object empty-fallback case and the row→array mapping here rather than deferring the full suite to the LFXV2-2689 follow-up? That's the one path where a silent regression (empty chart / stray 5xx) wouldn't be caught by anything today. (+1 to Copilot's coverage comment.)

PR #1258, opened 2026-07-29
fix(meetings): pass occurrence_id to Show Members registrant preload

andrest50:
## Summary

- Fixes a fifth read path missed by #1219: the meeting **details page Show Members drawer** preloaded registrants via `getMyMeetingRegistrants` without `occurrence_id`, so a future `single`-scope decline appeared on every occurrence even though **Your RSVP** and the card drawer were correct.
- Adds shared `resolveRsvpOccurrenceId()` and uses it across all frontend RSVP BFF call sites so occurrence context is resolved consistently (explicit occurrence → fallback to current/next).
- Routes registrant chip + filter rendering through `getRegistrantAttendanceStatus()` so display semantics live in one place instead of being duplicated in the template and filter logic.

## Follow-up to [lfx-self-serve#1219](https://github.com/linuxfoundation/lfx-self-serve/pull/1219) (LFXV2-2864)

[lfx-self-serve#1219](https://github.com/linuxfoundation/lfx-self-serve/pull/1219) normalised RSVP `occurrence_id` seconds-vs-milliseconds handling and routed the detail-page **Your RSVP** chip, meeting-card **Show Guests** drawer, dashboard list, and guest-list BFF through the shared `selectApplicableRsvp` resolver. After that landed, manual QA on Connie's recurring meeting still showed a mismatch on the meeting **details page Show Members drawer**: it preloaded registrants without `occurrence_id`, so a `single`-scope decline for a future occurrence appeared on every occurrence even when **Your RSVP** and the card drawer were correct.

This PR closes that remaining read path by passing the viewed occurrence into the Show Members preload, centralising occurrence resolution via `resolveRsvpOccurrenceId()`, and aligning chip/filter rendering with `getRegistrantAttendanceStatus()` so all RSVP display sites stay in lockstep.

## Test plan

- [x] Recurring meeting: decline one future occurrence in Google Calendar, wait for sync
- [x] Open the **next** occurrence on the meeting details page — **Your RSVP** shows accepted
- [x] Open **Show Members** on that occurrence — organizer row shows **accepted**, not declined
- [x] Navigate to the **declined** occurrence — Show Members shows declined there only
- [x] Spot-check meeting card **Show Guests** drawer — same per-occurrence semantics, no regression

### DIFF
--- a/apps/lfx-one/src/app/modules/meetings/components/meeting-registrants-display/meeting-registrants-display.component.html
+++ b/apps/lfx-one/src/app/modules/meetings/components/meeting-registrants-display/meeting-registrants-display.component.html
@@ -104,14 +104,19 @@
 
             <!-- RSVP Status Badge Overlay -->
             <div class="absolute -bottom-0.5 -right-0.5 w-4 h-4 bg-white rounded-full flex items-center justify-center">
-              @if (registrant.rsvp?.response_type === 'accepted' || registrant.invite_accepted === true) {
-                <i class="fa-solid fa-check-circle text-emerald-600 text-base" pTooltip="Accepted"></i>
-              } @else if (registrant.rsvp?.response_type === 'maybe') {
-                <i class="fa-solid fa-circle-question text-amber-600 text-base" pTooltip="Maybe"></i>
-              } @else if (registrant.rsvp?.response_type === 'declined' || registrant.invite_accepted === false) {
-                <i class="fa-solid fa-times-circle text-red-600 text-base" pTooltip="Declined"></i>
-              } @else {
-                <i class="fa-solid fa-clock text-gray-400 text-sm" pTooltip="Pending"></i>
+              @switch (registrantAttendanceStatus(registrant)) {
+                @case ('accepted') {
+                  <i class="fa-solid fa-check-circle text-emerald-600 text-base" pTooltip="Accepted"></i>
+                }
+                @case ('maybe') {
+                  <i class="fa-solid fa-circle-question text-amber-600 text-base" pTooltip="Maybe"></i>
+                }
+                @case ('declined') {
+                  <i class="fa-solid fa-times-circle text-red-600 text-base" pTooltip="Declined"></i>
+                }
+                @default {
+                  <i class="fa-solid fa-clock text-gray-400 text-sm" pTooltip="Pending"></i>
+                }
               }
             </div>
           </div>
--- a/apps/lfx-one/src/app/modules/meetings/components/meeting-registrants-display/meeting-registrants-display.component.html
+++ b/apps/lfx-one/src/app/modules/meetings/components/meeting-registrants-display/meeting-registrants-display.component.html
@@ -104,7 +104,7 @@
 
             <!-- RSVP Status Badge Overlay -->
             <div class="absolute -bottom-0.5 -right-0.5 w-4 h-4 bg-white rounded-full flex items-center justify-center">
-              @switch (registrantAttendanceStatus(registrant)) {
+              @switch (registrant.attendanceStatus) {
                 @case ('accepted') {
                   <i class="fa-solid fa-check-circle text-emerald-600 text-base" pTooltip="Accepted"></i>
                 }
--- a/apps/lfx-one/src/app/modules/meetings/components/meeting-rsvp-details/meeting-rsvp-details.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/components/meeting-rsvp-details/meeting-rsvp-details.component.ts
@@ -97,7 +97,7 @@ export class MeetingRsvpDetailsComponent {
             // Resolve RSVPs against the target occurrence so a `single` decline for a
             // future date doesn't overwrite the current occurrence's per-registrant
             // chip (LFXV2-2864).
-            const occurrenceId = resolveRsvpOccurrenceId(meeting as Meeting, { occurrence: this.currentOccurrence() });
+            const occurrenceId = resolveRsvpOccurrenceId(meeting, { occurrence: this.currentOccurrence() });
             return this.meetingService.getMeetingRegistrants(meeting.id, true, occurrenceId).pipe(
               map((registrants) => ({
                 registrants,
--- a/apps/lfx-one/src/app/modules/meetings/components/meeting-rsvp-details/meeting-rsvp-details.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/components/meeting-rsvp-details/meeting-rsvp-details.component.ts
@@ -17,6 +17,7 @@ import {
   Project,
   RsvpCounts,
 } from '@lfx-one/shared';
+import { resolveRsvpOccurrenceId } from '@lfx-one/shared/utils';
 import { MeetingService } from '@services/meeting.service';
 import { UserService } from '@services/user.service';
 import { catchError, distinctUntilChanged, map, of, switchMap, tap } from 'rxjs';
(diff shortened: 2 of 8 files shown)


apps/lfx-one/src/app/modules/meetings/components/meeting-registrants-display/meeting-registrants-display.component.html
```diff
@@ -104,14 +104,19 @@
...
             <div class="absolute -bottom-0.5 -right-0.5 w-4 h-4 bg-white rounded-full flex items-center justify-center">
-              @if (registrant.rsvp?.response_type === 'accepted' || registrant.invite_accepted === true) {
-                <i class="fa-solid fa-check-circle text-emerald-600 text-base" pTooltip="Accepted"></i>
-              } @else if (registrant.rsvp?.response_type === 'maybe') {
-                <i class="fa-solid fa-circle-question text-amber-600 text-base" pTooltip="Maybe"></i>
-              } @else if (registrant.rsvp?.response_type === 'declined' || registrant.invite_accepted === false) {
-                <i class="fa-solid fa-times-circle text-red-600 text-base" pTooltip="Declined"></i>
-              } @else {
-                <i class="fa-solid fa-clock text-gray-400 text-sm" pTooltip="Pending"></i>
+              @switch (registrantAttendanceStatus(registrant)) {
```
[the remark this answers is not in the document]
andrest50: Fixed — `attendanceStatus` is now precomputed inside the `filteredRegistrants` computed signal via `getRegistrantAttendanceStatus`, and the template reads `registrant.attendanceStatus` instead of calling application logic on each change-detection pass.

packages/shared/src/utils/meeting.utils.spec.ts
```diff
@@ -105,6 +106,41 @@ describe('sortPastMeetingsDescending', () => {
...
+        occurrence: { occurrence_id: '1785852000', start_time: '2026-08-04T14:00:00Z', duration: 60 } as MeetingOccurrence,
+      })
+    ).toBe('1785852000');
+  });
+
+  it('falls back to getCurrentOrNextOccurrence when no occurrence context is supplied', () => {
+    // Both ids are valid outputs depending on clock; assert we get one of the series ids.
+    const resolved = resolveRsvpOccurrenceId(recurringMeeting);
+    expect(['1785247200', '1785852000']).toContain(resolved);
+  });
```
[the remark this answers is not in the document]
andrest50: Fixed — the fallback test now freezes Vitest's clock (`vi.useFakeTimers` / `vi.setSystemTime`) and builds occurrence start times relative to that instant, so it stays deterministic after the fixture dates pass.

packages/shared/src/utils/meeting.utils.ts
```diff
@@ -327,6 +327,25 @@ export function getCurrentOrNextOccurrence(meeting: Meeting): MeetingOccurrence
...
+ * Recurring meetings prefer an explicit occurrence or id; when neither is supplied,
+ * fall back to {@link getCurrentOrNextOccurrence} so list/card surfaces that only
+ * hold the meeting payload still resolve against the current/next slot.
+ */
+export function resolveRsvpOccurrenceId(
+  meeting: Meeting,
+  options?: { occurrence?: MeetingOccurrence | null; occurrenceId?: string | null }
+): string | undefined {
+  if (!meeting.recurrence) return undefined;
+  const explicitId = options?.occurrenceId ?? options?.occurrence?.occurrence_id;
```
audigregorie: **Minor — `??` / truthy-check mismatch in the explicit-id merge**

`options?.occurrenceId ?? options?.occurrence?.occurrence_id` keeps an empty string (empty string is not nullish, so `??` does not fall through), but the next line's `if (explicitId)` treats `''` as falsy and drops it, falling back to `getCurrentOrNextOccurrence` — silently ignoring a valid `occurrence.occurrence_id` when a caller passes `{ occurrenceId: '', occurrence: validOccurrence }`.

No current caller passes both options, so this is latent rather than active. Switching the merge to `||` makes an empty-string `occurrenceId` fall through to `occurrence.occurrence_id` consistently:

```suggestion
  const explicitId = options?.occurrenceId || options?.occurrence?.occurrence_id;
```

apps/lfx-one/src/app/modules/meetings/components/meeting-rsvp-details/meeting-rsvp-details.component.ts
```diff
@@ -96,7 +97,7 @@ export class MeetingRsvpDetailsComponent {
             // Resolve RSVPs against the target occurrence so a `single` decline for a
             // future date doesn't overwrite the current occurrence's per-registrant
             // chip (LFXV2-2864).
-            const occurrenceId = this.currentOccurrence()?.occurrence_id;
+            const occurrenceId = resolveRsvpOccurrenceId(meeting as Meeting, { occurrence: this.currentOccurrence() });
```
audigregorie: **Minor — `meeting as Meeting` cast on a `Meeting | PastMeeting` input**

The cast is safe at runtime because the past-meeting branch returns early (lines 92–93), so `meeting` is effectively `Meeting` here. It's slightly looser than necessary though — `resolveRsvpOccurrenceId` only reads `recurrence` / `occurrences` / `cancelled_occurrences`, all of which are optional on `PastMeeting` in practice. If `PastMeeting` ever drops those fields, the cast would mask the resulting type error. Consider narrowing with a guard (`if (!meeting.recurrence) ...`) or widening `resolveRsvpOccurrenceId`'s parameter to `Pick<Meeting, 'recurrence' | 'occurrences' | 'cancelled_occurrences'>` so the cast isn't needed. Non-blocking.

audigregorie (commented):
## Code Review Summary

Solid follow-up to #1219. The core fix — subscribing to `currentOccurrence` in `initializeRegistrants` so the Show Members drawer refetches on occurrence navigation — correctly closes the last stale-decline read path, and centralizing occurrence resolution + chip/filter rendering in `resolveRsvpOccurrenceId` / `getRegistrantAttendanceStatus` removes the duplicated `response_type`/`invite_accepted` logic that had already drifted between template and filter. The `user.service.ts` change is a behavior-preserving refactor (no options → same `getCurrentOrNextOccurrence` fallback).

No Critical or Major findings. Three Minor notes below.

### Minor — outside the diff

- **Test coverage for `resolveRsvpOccurrenceId`**: the new spec covers non-recurring, explicit `occurrenceId`, explicit `occurrence`, and the `getCurrentOrNextOccurrence` fallback, but doesn't cover (a) precedence when *both* `occurrence` and `occurrenceId` are supplied, and (b) an empty-string `occurrenceId`. Both would lock in the merge semantics called out in the line comment on `meeting.utils.ts`.

### What's done well

- `distinctUntilChanged` on `occurrence_id` in the new `combineLatest` entry prevents redundant refetches for the same occurrence while guaranteeing a refetch on navigation — exactly the right granularity.
- Migrating the template from nested `@if/@else if` to `@switch/@case/@default` is more idiomatic for mutually-exclusive value matching and aligns with the repo's `@switch`-only guidance.
- The `RegistrantWithAttendance` view-model is a `type` alias (not a domain `interface`), and the domain `RegistrantAttendanceStatus` lives in `@lfx-one/shared` — correct layering.

packages/shared/src/utils/meeting.utils.ts
andrest50: Fixed — switched to `||` so an empty `occurrenceId` falls through to `occurrence.occurrence_id` consistently, with a unit test covering the edge case.

apps/lfx-one/src/app/modules/meetings/components/meeting-rsvp-details/meeting-rsvp-details.component.ts
andrest50: Fixed — removed the `as Meeting` cast; `PastMeeting` extends `Meeting` and the past-meeting branch already returns early here.

audigregorie (approved):

PR #1263, opened 2026-07-29
docs(frontend): revise permission/persona spec pages

manishdixitlfx:
## Summary
- Adopts permission-not-persona throughout both spec pages: Core Principle section, auditor vs viewer terminology (per Eric Searcy's June 25/26 review), resolves the Board Member/Contributor context-entry ambiguity, removes the ED persona-guard carve-out for writer actions, reworded P0 API-enforcement gate.
- Fact-checks current status: PR #1130 and #1193 shipped; LFXV2-2753 rescoped 2026-07-19 and now Discarded (not in progress); flags the ED fast path in `writer.guard.ts`/`newsletter-access.guard.ts` as likely a bug per `PERMISSIONS.md`.
- Adds a Model Asks section for the platform team (ED relation removal, LF Staff Team → B2B Organization auditor inheritance).

Docs-only change, no application code touched.

LFXV2-2913

## Test plan
- [x] `yarn prettier --check` on both changed files
- [x] `yarn lint:check` (no errors; pre-existing unrelated warning only)
- [x] Grepped both docs for "persona" — every hit is presentation-only or an explicit FGA-model reference
- [x] Re-read both docs for cross-doc consistency (terminology, defaulting ladders, ED capability question)


docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+Persona shapes presentation only.
+Roles are grant bundles in the FGA model, never branches in code.
+```
+
+## Core Principle
+
+Roles and personas exist only in the authorization model (OpenFGA). A role is
+a named bundle of permissions. Permissions are assigned to roles; roles are
+assigned to users. Application code never branches on role or persona to make
+a decision.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: reworded the Core Principle so it distinguishes access decisions (never branch on role/persona) from presentation decisions (persona may branch layout/copy) - the two are no longer stated as a blanket contradiction.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+When a user clicks Foundation or Project context without selecting a specific row, use the last selected valid context first. Highest-permission defaulting is only the cold-start fallback.
+
+Foundation defaulting order:
+
+1. Keep existing selected foundation if still auditor/role-permitted.
+2. Use last selected valid foundation if still auditor/role-permitted.
+3. Choose a foundation where the user has a writer/manage grant.
+4. Choose a foundation where the user has an auditor/explicit role grant.
+5. Choose first grant-permitted foundation in stable sort order.
+6. If none exist, stay in Me/discovery.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: rung 4/5 collapsed - the stable sort order is now the explicit tie-breaker inside the writer/manage and auditor tiers, not a separate unreachable rung.

docs/architecture/frontend/permission-persona-decision-matrix.md
```diff
@@ -0,0 +1,368 @@
...
+- **Visible experience:** Cross-context personal workspace.
+- **Allowed actions:** View personal tasks, meetings, events, groups, mailing lists, newsletters, votes, surveys, documents, and discovery entry points.
+- **Create/manage rule:** Allowed only after the action resolves or asks for a target Foundation/Project context and writer permission passes.
+
+### User Opens Foundation Without A Selected Foundation
+
+- **Input:** User clicks Foundation/Project lens and chooses Foundation context without selecting a row.
+- **Required permission:** At least one auditor/explicit-role-permitted foundation.
+- **Destination:** Last selected valid foundation first; highest-grant eligible foundation only on cold start.
+- **Default order:** Existing selected foundation (if still auditor/role-permitted), last selected valid foundation (if still auditor/role-permitted), foundation with a writer/manage grant, foundation with an auditor/explicit role grant, first grant-permitted foundation in stable sort order, none -> stay in Me/discovery.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed (mirrors the preread fix): same tie-breaker collapse applied here.

docs/architecture/frontend/permission-persona-decision-matrix.md
```diff
@@ -0,0 +1,368 @@
...
+- **Write examples:** Edit a meeting, manage agenda, update survey, manage document, edit newsletter draft.
+- **Required permission:** Viewer/discoverable eligibility or item eligibility for read actions. Writer permission for the item's resolved target context for write actions.
+- **Destination:** Stay in Me or open the item detail/drawer with target context attached.
+- **Allowed actions:** Read actions follow item eligibility. Create/manage actions are visible or enabled when writer permission passes.
+- **Denied actions:** If writer permission fails, keep eligible view/read actions only.
+
+### Create Action From Me
+
+- **Example:** Create Meeting, Create Group, Add Mailing List, Create Newsletter, Create Vote, Create Survey, Upload File, Add Link.
+- **Required permission:** User must choose a target Foundation/Project context, then writer permission must pass for that target.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: Create Action From Me now states the committee/group target path (committee.writer for meetings/surveys/votes, project.meeting_coordinator for meetings) alongside the Foundation/Project writer path.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+Me item + resolved target Foundation/Project context + writer permission = allowed create/manage action
+```
+
+Rule for create actions:
+
+```text
+Me create action + chosen target Foundation/Project context + writer permission = allowed create/manage action
+```
+
+This means Me can be an action workspace, but it is not the authorization context. The target Foundation/Project remains the authority for writer checks.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: added an explicit paragraph clarifying 'writer permission' means the action-specific grant (writer, committee.writer, or meeting_coordinator), citing writer.guard.ts.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
 
-## Purpose
+## TL;DR
 
-This preread proposes a clear separation between access, experience, and authority. The goal is to make the product easier to reason about and to keep persona-driven UX from accidentally becoming authorization logic.
+Separate the product model into four decisions:
 
-## Core Contract
+- **Where can I go?** An auditor or explicit role grant controls Foundation/Project context eligibility.
+- **What is my role there?** Persona and role shape navigation, dashboards, sections, and metrics.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: reworded to 'What is my persona there? Persona shapes presentation ... never access' so it no longer reads as role-based branching.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+
+Entry should be based on an authoritative role/permission model. Once inside, role/persona shapes the experience.
 
-### 1. Context Selector: Where Can I Go?
+Role pairs — and a critical distinction between them:
 
-- Foundation and Project selector eligibility should be based on view permission.
-- If a context appears in the selector, the user should be allowed to enter that context.
-- Persona may narrow or prioritize the experience, but it should not be the authority for selector eligibility.
+- **Executive Director** in Foundation context maps to **Maintainer** in Project context. Both are context operator experiences. They can create/manage only when writer permission is present.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: section renamed 'Persona pairs' and reworded to state only ED is a real FGA relation; Board Member/Contributor/Maintainer are presentation concepts, not roles.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+auditor or explicit role grant for that exact context — persona never opens
+a context. A user without such a grant does not see Foundation/Project
+navigation at all; their experience is Me. This does not strand them: their
+own items, meetings, votes, and documents remain fully usable from Me via
+item-level eligibility (see Me Lens above), so a Contributor or Board Member
+signal without an explicit grant still has a complete, useful experience —
+just not a Foundation/Project context to enter.
 
-### 3. Actions and Routes: What Can I Do?
+This means ED is not a synonym for every privileged action, and Maintainer is not a lesser product concept than ED. They are parallel operating roles for different context types, but only ED carries any actual permission weight today.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: reworded to 'parallel presentation concepts... only ED corresponds to any FGA relation today.'

docs/architecture/frontend/permission-persona-decision-matrix.md
```diff
@@ -0,0 +1,368 @@
...
+- **Allowed actions:** Staff-only operations defined for LF workflows.
+- **Denied actions:** Do not use LF Staff Mode as a substitute for normal Foundation/Project writer permission unless the workflow is explicitly staff-owned and audited.
+
+## Feature Decisions
+
+### Meetings
+
+- **Me:** Show cross-context meetings and personal meeting actions.
+- **Foundation/Project:** Show context-scoped meetings.
+- **Create/manage:** Requires target context plus writer permission.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: Meetings feature decision now lists project.meeting_coordinator and committee.writer as alternate grants alongside project/foundation writer.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+```
+
+Target-state guard change:
+
+```text
+Current writerGuard = Executive Director fast path or canWrite()
+Target writerGuard = resolved target context + canWrite()
+```
+
+ED-shaped pages (Health Metrics, Campaigns) require a named capability checked like any other permission — not a persona guard. This is an open model decision, not yet resolved: either FGA keeps an ED (or a successor relation, e.g. `foundation_insights_viewer`) that carries the capability these pages check, or the pages open to all auditors. Removing the ED relation while keeping ED-gated pages would force persona-checking back into the UI, which is exactly what this document argues against — so the relation-removal question and the ED-page-gating question have to be resolved together (see Model Asks below). Create/edit/manage routes should not use ED as an authorization shortcut unless the user also has writer permission for the selected target context.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: replaced the invented foundation_insights_viewer relation with the real existing ones - campaign_manager (Campaigns, write) and marketing_auditor (Marketing Impact, read) - both inheritable from executive_director or marketing_ops today. Referenced LFXV2-2236 (in review), which is migrating exactly these two pages to those checks; Health Metrics stays ED-gated by design.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
-- Keep persona/lens logic for sidebar and page composition, but avoid using persona as action authority.
-- Guard direct create/edit/admin routes with manage/write permission, not only hidden buttons.
-- Gate Permissions page admin actions such as Add User, role update, and removal by manage/write permission.
+1. **Decision on the ED relation.** Today `Executive Director` is a real FGA
+   relation on Project, but it only grants auditor/read — it is redundant
+   with LF-staff-inherited auditor everywhere except one project that has a
+   non-staff ED. Removal is acceptable provided that one non-staff ED
+   receives a direct auditor grant, and provided ED-shaped pages (Health
+   Metrics, Campaigns) get a named capability instead of depending on the ED
+   relation (see Writer Actions above).
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: Model Ask now explicitly covers the campaign_manager/marketing_auditor inheritance consequence for every ED (not just the one non-staff exception), and flags the open tension between Eric Searcy's 'should probably be removed' and LFXV2-2236 building around keeping the relation.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+- Discovery already exists in places like Discover Events.
+- Akrites shows the strongest discovery pattern: users can inspect packages and choose **Open for stewardship** from Me without starting in Foundation/Project context.
+- Lens access is already derived from writer grants, not persona alone (shipped, PR #1130), and the create picker is a lazy direct-grant tree with search (shipped, PR #1193).
+- A backend "what can I create, and where" API covering group/committee targets (LFXV2-2753) was rescoped after #1130 shipped and its ticket status is now Discarded — it is not in progress. There is no committed replacement work item for that API today.
+- The Executive Director fast path is still present in `writerGuard` (`writer.guard.ts`) and in `newsletter-access.guard.ts` on `main` — both grant **write** access unconditionally to ED persona. Per `PERMISSIONS.md` (the live permissions model), the `Executive Director` relation only grants view/read rows, never a write-granting row. Granting write through this fast path is very likely a bug, not a deliberate model choice, and is tracked separately for removal rather than treated as an open design question.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: narrowed the claim - ED does inherit one write-capable relation (campaign_manager, scoped to Campaigns) in addition to auditor/marketing_auditor reads. The claim now specifically says ED does not inherit project.writer, which is what writer.guard.ts/newsletter-access.guard.ts actually check.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
-- Persona may narrow or prioritize the experience, but it should not be the authority for selector eligibility.
+- **Executive Director** in Foundation context maps to **Maintainer** in Project context. Both are context operator experiences. They can create/manage only when writer permission is present.
+- **Board Member** in Foundation context maps to **Contributor** in Project context. Both are context participant experiences. They do not automatically receive create/manage authority.
 
-### 2. Sidebar and Page Content: What Do I See There?
+Of these four, only ED is a real FGA relation. Board Member, Contributor, and
+Maintainer are inferred/dynamic signals detected from connected data (board
+group membership, CDP activity, Insights), not discrete roles anywhere in LFX
+data. There is no "Board Member" or "Maintainer" relation on Project or
+Foundation in the permissions model.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: narrowed to 'not a Project or Foundation relation in the permissions model' and added the cdp_roles detail (verified against persona-detection.service.ts:350-368) so it no longer claims Maintainer has no discrete value anywhere.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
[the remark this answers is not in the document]
manishdixitlfx: Fixed: renamed Role pairs -> Persona pairs, clarified only ED is a real FGA relation, and added explicit Board Member/Contributor no-grant regression-case bullets to the decision matrix (lines 151-154, 187-189 area).

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,460 @@
...
+6. If none exist, stay in Me/discovery.
+
+Project defaulting order:
+
+1. Keep existing selected project if still auditor/role-permitted.
+2. Use last selected valid project if still auditor/role-permitted.
+3. Choose a project where the user has a writer/manage grant.
+4. Choose a project where the user has an auditor/explicit role grant.
+5. Choose first grant-permitted project in stable sort order.
+6. If none exist, stay in Me/discovery.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed: collapsed the writer/manage and auditor tiers so the stable sort order is the explicit tie-breaker within each tier, in both the preread and the decision matrix, plus added a worked example showing two same-tier grants resolving by stable order rather than API response order.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+Working spec for aligning how the LFX Self Serve experience in this repo decides where users can go, what they see, and what they can do.
 
-## Purpose
+## TL;DR
 
-This preread proposes a clear separation between access, experience, and authority. The goal is to make the product easier to reason about and to keep persona-driven UX from accidentally becoming authorization logic.
+Separate the product model into four decisions:
 
-## Core Contract
+- **Where can I go?** An auditor or explicit role grant controls Foundation/Project context eligibility.
```
emsearcy: Suggest tightening this up:

> Outside of the "Me" lens, "auditor" access (explicit grant or inherited) is needed to browse into contexts (Foundation/Project or Organization).

Why? explicit roles don't *always* include auditor (like "participant" on a meeting) ... and really that is the only role that should matter for projects/orgs or foundations ... any "higher" role is going to include auditor by way of inheritance.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+- **What is my persona there?** Persona shapes presentation — navigation, dashboards, sections, and metrics — never access.
+- **What can I do there?** Contextual writer permission controls create/manage actions.
```
emsearcy: This is an oversimplification — it's *do I have the relation (against a given object) necessary to make the corresponding API call(s) that action takes/makes?* It's not always _writer_. e.g. "organizer" on past meetings, or "voter" for poll responses (only the author of the poll can edit it!).

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
-| What do I see there? | Persona / role            | Sidebar links, page sections, dashboard variants, content shape. |
-| What can I do?       | Manage / write permission | Create, edit, delete, configure, invite, assign, administer.     |
+The key rule:
 
-## Decision Boundaries
+```text
+Context access is not action authority.
+Every gate is a permission check.
+Persona shapes presentation only.
+Roles are grant bundles in the FGA model, never branches in code.
```
emsearcy: Seems misleading. You'd branch, in code, based on the evaluated FGA access in order to implement a *branched* "what can I do there" logic. Consider rewriting this to better express what you're trying to get across.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+```
+
+## Core Principle
+
+Roles exist only in the authorization model (OpenFGA): a role is a named
+bundle of permissions, permissions are assigned to roles, and roles are
```
emsearcy: I'd suggest this be fixed to correctly capture/document how the fine-grained access control system (CNCF OpenFGA project) that undergirds "self service" actually works.

Roles are neither bundles of permissions, nor are permissions "assigned" to roles.

Roles (relationships) exist independently as a model (graph) describing how relationships are evaluated transitively across objects we define: users, projects, meetings, and so on:

e.g.:
- users can be related as "writers" to a project
- writers of a project includes all the writers of that project's parent
- users can be related as "key contacts" on a project membership
- key contacts of an organization include the key contacts of every membership the organization has
- writers of a committee include the writers of the committee's project
... and so on.
that's the *model*.

Every defined edge in the graph for our actual data—project A is a parent of project B, user X is a writer of project A—is a "tuple". And when the tuple between a user and another object (instead of between two objects), it's helpful to think of this as an "explicit role grant".

However, when you _evaluate_ a relationship, you aren't checking the explicit role grants, you are traversing the model to assert whether or not a given relation exists (_is user X a writer of project B?—Yes!_)

"Permissions", the rules that define *what relation to transitively check to allow an API call (read/write) against any given API*, are implemented as _Heimdall RuleSets_. So a permission doesn't exist merely as a "label" that you attach to a role -- a permission is a functional gate -- a tiny program that is part of the API spec.

So while is role isn't merely a "bundle" of permissions: we CAN aggregate *all* documented APIs that use that relation/role as their permission gate, in order to summarize all the activities (jobs to be done) that the user is able to do if they have that role. But the role itself is not a "collection" of those activities.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+assigned to users via relations/grants. Personas (Board Member, Contributor,
+Maintainer, and the presentation use of ED) are detected signals for shaping
+the experience — they are not modeled in OpenFGA, except that ED also
+happens to be a real relation (see Role pairs and Terminology below).
+
+Application code never branches on role or persona to decide an _access_
```
emsearcy: this paragraph is not consistent with the rest of the doc. whether a context can be entered, or a write can proceed, have explicitly been described as being by "role" rather than persona.

Is line only supposed to say "persona", rather than "role or persona"?

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+question: whether a context can be entered, a route can open, or a write can
+proceed. Those are always answered by a permission check against the
+resolved target object — the app consumes capabilities, not identities.
+
+If a page or feature must exist for one audience only, that is a permission
+too: model it as a relation/capability in FGA, grant it to the appropriate
+role, and check the capability.
```
emsearcy: again represents a misunderstanding -- there is no concept of "role" in OpenFGA ... when we say "role", what we really mean **is** "relation".

I think what this is trying to communicate, restated in terms that match OpenFGA: if a page or feature must exist for one audience only, model it as a relation/role in FGA, configure it as the permission (RuleSet) for the API routes that power that page, and then wire up "permission/access pre-checks" in the UI for early bailout on page rendering, or for conditionally showing links into the page (tab, nav sidebar, etc).

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+driving presentation is not an exception to the rule above.
+
+## Terminology: Viewer vs Auditor
+
+The FGA model already distinguishes two different kinds of read access, and
+this document uses those names precisely instead of the ambiguous "view
+permission":
+
+- **Viewer** — discoverability only. You can tell the thing exists (a public
+  meeting, an active project) but you do not get access to what is inside it
```
emsearcy: can tell that it exists *and* see *limited* (typically, public) data about it -- but not privileged data, or subordinate/connected _objects_.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+  (participants, formation-stage child projects, private documents). Viewer
+  is not ordinarily assigned directly: it is conditional on object attributes
+  (public/conditional) or inherited from another role you hold (a committee
+  `member` role always inherits `viewer`, too). Maps to Discovery.
+- **Auditor** — privileged read. You get access to the private data "inside"
+  the object, not just its existence. Auditor is what makes a Foundation or
```
emsearcy: I used "inside" before, but more specifically, it's *privileged data* on that object as well as *subordinate/connected objects*.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+selector eligibility or context entry previously said "view permission," it
+now reads "auditor or explicit role"; everywhere the meaning was
+discoverability only, it reads "viewer/discoverable."
+
+## Why This Is Needed
+
+The product currently blends three things that should stay separate:
+
+1. Can the user view this foundation or project?
+2. What relationship does the user have to that context?
```
emsearcy: can we change the word _relationship_ (which is a loaded term with respect to OpenFGA) to "engagement/activity (current or past)"

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+- The Executive Director fast path is still present in `writerGuard` (`writer.guard.ts`) and in `newsletter-access.guard.ts` on `main` — both grant **write** access unconditionally to ED persona, covering meetings, votes, surveys, mailing lists, newsletters, and documents. The `executive_director` relation does carry one write-capable inheritance today — `campaign_manager`, scoped to the Campaigns feature (see Model Asks below) — but it does not inherit project `writer`, and none of the general write actions these two guards protect are backed by any ED-inherited relation. `PERMISSIONS.md`'s rendered Project table only shows `Executive Director` in view/read rows because it does not render the marketing/campaign relations at all, not because those relations don't exist in the live model (`model.fga`). Granting general project write through this fast path is very likely a bug, not a deliberate model choice, and is tracked separately for removal rather than treated as an open design question.
```
emsearcy: "unconditional write access"

This seems highly unlikely that this is *granted*. Perhaps the UI "assumes" it? But the UI itself is subordinate to the API's enforcement of the OpenFGA model, the ED cannot get writer access to a project just because the UI wants them to.

Or else, this is referring not to unconditional access, but some _limited_ form of access, to UI-only functionality without a backend API?

docs/architecture/frontend/permission-persona-navigation-model-preread.md
emsearcy: apparently line 146 makes this explicit, a sort of "when I say writer I don't really mean writer" ... maybe this works for AI but it doesn't read well. maybe if we said "writer (or similar)" everyone else it would help. AI sometimes only reads chunks, too.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+This migration is tracked in LFXV2-2236 ("Add Marketing Ops UI access (FGA guards)," in review as of this writing) — today, all three pages on `main` still gate solely on `executiveDirectorGuard` (a pure persona check with no FGA lookup). Until that ticket merges, treat "Current UI Facts" as describing the actual state, not this target state.
 
-- Once inside a view-permitted context, persona/role should shape navigation links, page sections, dashboards, tabs, cards, metrics, and content variants.
-- This is where Executive Director, Board Member, Maintainer, Contributor, and committee roles can create different product experiences.
+This leaves one real open model question, not three: should the `executive_director` relation itself be removed (see Model Asks below), given that it is also the sole non-`marketing_ops` path to `campaign_manager`/`marketing_auditor` for every ED, and Health Metrics has no planned replacement gate other than possibly LF Staff access (LFXV2-2726, not yet started). Create/edit/manage routes should not use ED as an authorization shortcut unless the user also has writer permission for the selected target context.
```
emsearcy: are we expecting to resolve this? leaving open questions in our "source of truth for design" docs isn't setting ourselves up for future success.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
 
-### 3. Actions and Routes: What Can I Do?
+LF Staff Mode should have its own explicit staff eligibility and audit expectations. It should not be inferred from ED, Board Member, Maintainer, Contributor, or writer permission.
```
emsearcy: what do you mean by "LF Staff Mode"? what actions or behaviors should be provided to a staff member (with auditor on every foundation), that wouldn't be provided to a privileged community member explicitly granted auditor on a given foundation?

This document says that I can see all projects/foundations I have auditor on—and LF Staff have auditor on all projects—meaning they'll see everything. And below, there are actions to make Self Service solve this from a usability standpoint by promoting recent projects, or writer-level-access projects, above the long tail. But that could equally apply to a community member with auditor access to CNCF (all projects) except for writer or one. So what is special for staff?

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
+- Decide whether the left rail should keep separate Foundation and Project buttons or move to one context switcher with grouped context types.
+- Align the persona content matrix so ED/Board/Maintainer/Contributor differences remain product experience rules, not authorization shortcuts.
+- Add regression tests for read-only users who can view a context but cannot see create/manage affordances.
+- Add regression tests for working group chairs who can manage group-level meetings/documents without receiving foundation-wide admin affordances.
+
+## Model Asks (Platform Team)
+
+Two open questions in the FGA model, not the UI, that this document depends on:
+
+1. **Decision on the ED relation.** Today `executive_director` is a real FGA
```
emsearcy: let's keep it as an FGA relation.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
-- Keep persona/lens logic for sidebar and page composition, but avoid using persona as action authority.
-- Guard direct create/edit/admin routes with manage/write permission, not only hidden buttons.
-- Gate Permissions page admin actions such as Add User, role update, and removal by manage/write permission.
+2. **Org lens for LF staff.** LF staff should get org-lens switching the same
+   way they get project/foundation access today — through team inheritance,
+   not impersonation. `PERMISSIONS.md` shows Project's Auditor relation
+   already inherits from the global LF Staff Team; the B2B Organization
+   section does not have the equivalent inheritance. This needs an LF Staff
+   Team → B2B Organization auditor inheritance added to the model — a model
+   change, not a UI workaround.
```
emsearcy: a model change is needed, but is leaving this as an "open ask" the best solution? is there a ticket filed for this? would we keep this PR open until it's addressed, rather than leaving it as an open "ask" in our architecture doc?

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,499 @@
...
 
 ## Meeting Ask
```
emsearcy: seems out of context for this doc -- what meeting, when, etc.

PR approval is a good async model.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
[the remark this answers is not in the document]
manishdixitlfx: Follow-up (fde6e89fd): generalized further per a closer read of writer.guard.ts - the Me-lens rule blocks now state the general form directly ("resolved target object (project, foundation, committee, group) + the action-specific grant on that object") instead of a Foundation/Project-first rule with a committee/meeting-coordinator caveat bolted on.

docs/architecture/frontend/permission-persona-decision-matrix.md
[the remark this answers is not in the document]
manishdixitlfx: Follow-up (fde6e89fd): tightened wording to match the same "target object + action-specific grant" phrasing now used in the preread, for consistency across both docs.

docs/architecture/frontend/permission-persona-decision-matrix.md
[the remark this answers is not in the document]
manishdixitlfx: Follow-up (fde6e89fd): reworded to name the actual PERMISSIONS.md relations - Scheduled Meeting's Organizer inherits from Project Meeting Coordinator, Committee Writer, and Project Writer - instead of code-style guard names.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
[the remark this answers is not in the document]
manishdixitlfx: Follow-up (fde6e89fd): went further than the Jul 30 fix. Removed the "open model decision" framing (and the invented foundation_insights_viewer name) entirely. The doc now states plainly that Campaigns checks campaign_manager and Marketing Impact checks marketing_auditor - both existing permissions, no new capability needed - and that which relation(s) bundle those permissions is the platform team's call, not something this document argues about.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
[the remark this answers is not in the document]
manishdixitlfx: Follow-up (fde6e89fd): reworked Model Asks #1 to stop arguing a position on the executive_director relation. It now states only what the app checks (auditor, writer, marketing_auditor, campaign_manager - never the relation itself) and what must remain true after any model change (every affected ED, not just the one non-staff exception, keeps those permissions through some bundle). Bundling is left to the platform team.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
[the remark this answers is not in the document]
manishdixitlfx: Follow-up (fde6e89fd): tightened further - now leads with "very likely a guard bug, not a permission the model grants" and states the marketing_auditor/campaign_manager routing as fact rather than folding it into the removal debate.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): reworded to your suggested framing - auditor is the only relation that matters for context entry outside Me, since any higher relation includes it by inheritance.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): reworded away from the writer-only oversimplification - each action now checks the specific relation its API call requires (organizer, poll author, etc. called out as counter-examples to writer).

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): reworded to 'code branches on evaluated relation checks, never on role or persona labels' - that's the actual rule, distinct from 'never branches in code' which was wrong as you say.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): rewrote the Core Principle from your explanation - relations as a transitively-evaluated graph, tuples as the concrete data edges, permissions as RuleSets gating API routes rather than labels on a relation, and the role-isn't-a-collection point. Dropped the 'bundle of permissions' framing entirely. Thank you for the detailed writeup, it made this much more precise.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): resolved the inconsistency - code branches on evaluated relation checks (that's correct and necessary), never on persona labels. Reworded so the paragraph no longer says 'role or persona' as if both were forbidden from branching.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): reworded to 'relation' throughout (no 'role' concept in OpenFGA), and restated per your suggestion - model it as a relation, configure it as the RuleSet for the API routes, wire up permission pre-checks in the UI for early bail-out / conditional nav links.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): updated to your exact phrasing - can tell it exists and see limited (typically public) data, but not privileged data or subordinate/connected objects.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): updated to your phrasing - privileged data on the object itself, plus subordinate/connected objects, not just existence.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): changed 'relationship' to 'engagement or activity (current or past)'.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): you're right, 'grants write access unconditionally' overstated it. Reworded - the guard skips its own client-side check and lets the route render, which does not mean the write is granted; the UI is subordinate to API enforcement. Tied the open question (misleading dead-end vs. an actual enforcement gap) to the existing P0 ask to verify downstream enforcement.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: That sentence is gone (fixed in 9e57573b2, superseded by the round before this one which also removed it) - Model Asks #1 no longer poses this as an open removal question, see the reply on that thread below for how it resolved.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): you're right, I didn't have a real answer to what 'LF Staff Mode' adds over broad auditor access. Reframed both docs to say so explicitly - it's an open product question, not a settled requirement, with the only candidate distinguishing capability being write-side assisted-workflow support (which no read relation grants). Not asserting it as needed until product defines it.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Done (9e57573b2) - Model Asks #1 now states this as resolved: executive_director stays as an FGA relation, and the app checks auditor/writer/marketing_auditor/campaign_manager directly rather than branching on the relation itself.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Filed LFXV2-2936 (https://linuxfoundation.atlassian.net/browse/LFXV2-2936) and linked it from Model Asks #2 (9e57573b2) instead of leaving it as prose-only.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Fixed (9e57573b2): renamed 'Meeting Ask' -> 'Contract Summary' and dropped the meeting framing - PR review is the actual agreement mechanism.

docs/architecture/frontend/permission-persona-decision-matrix.md
```diff
@@ -0,0 +1,374 @@
...
+- **Allowed actions:** Explicit selection wins over defaulting. Create/manage follows writer permission for the selected context.
+- **Denied actions:** If the auditor/explicit role grant is missing or lost, clear selection and re-run defaulting.
+
+## Me Lens Actions
+
+### Existing Item Action From Me
+
+- **Read examples:** View meeting details, view vote results, view survey results, open document, view sent newsletter.
+- **Write examples:** Edit a meeting, manage agenda, update survey, manage document, edit newsletter draft.
+- **Required permission:** Viewer/discoverable eligibility or item eligibility for read actions. The action-specific grant on the item's resolved target object for write actions — writer permission for a Foundation/Project target, or committee writer / meeting coordinator when the target resolves to a committee or group.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed (de5ac20d0): applied your suggested wording almost verbatim - scoped meeting_coordinator to meeting actions only, committee.writer to committee/group targets.

docs/architecture/frontend/permission-persona-decision-matrix.md
```diff
@@ -0,0 +1,374 @@
...
+- **Allowed actions:** Staff-only operations defined for LF workflows.
+- **Denied actions:** Do not use LF Staff Mode as a substitute for normal Foundation/Project writer permission unless the workflow is explicitly staff-owned and audited.
+
+## Feature Decisions
+
+### Meetings
+
+- **Me:** Show cross-context meetings and personal meeting actions.
+- **Foundation/Project:** Show context-scoped meetings.
+- **Create/manage:** Requires the organizer-granting relation for the target — Project/Foundation Writer, Project Meeting Coordinator (meetings only), or Committee Writer when the target is a committee/group — per `PERMISSIONS.md`'s Scheduled Meeting inheritance (`Organizer` inherits from Project Meeting Coordinator, Committee Writer, Project Writer).
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed (de5ac20d0): agreed this doc states the target model, and writer.guard.ts's ED fast path is still live on main. Made that explicit in the Meetings feature decision and cross-linked persona-content-matrix.md's Meetings write paths table so the two docs stay in sync instead of silently disagreeing. Also cross-linked persona-content-matrix.md from both docs' intros generally - good catch that it existed and wasn't referenced.

docs/architecture/frontend/permission-persona-decision-matrix.md
```diff
@@ -0,0 +1,371 @@
...
+- **Allowed actions:** Register, join, follow, subscribe, request access, inspect, or open stewardship workflow.
+- **Denied actions:** Do not treat discovery eligibility as selector eligibility.
+
+### Write Candidate
+
+- **Input:** User attempts create/edit/delete/manage/send/publish.
+- **Required permission:** Resolved target context plus server-enforced writer permission.
+- **Destination:** Write flow only after UI and API permission checks agree.
+- **Allowed actions:** Continue when contextual writer permission passes.
+- **Denied actions:** Fail closed if the target context is missing, stale, no-grant, or not writer-permitted.
```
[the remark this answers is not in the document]
manishdixitlfx: Done — replaced the blanket `writer`-permission gating language with the action-specific relation the API requires on the resolved target object (e.g. `writer`, `committee.writer`, `meeting_coordinator`, vote-response owner) at all five flagged locations (Write Candidate P0 gate, Me create/manage rule, Existing Item Action, Pending Action, and Direct Create/Edit URL), plus the two adjacent Allowed/Denied bullets under Existing Item Action that would otherwise have stayed inconsistent with the reworded Required-permission line right above them.

See commit 3a7965b40.

manishdixitlfx:
@emsearcy the lexicon pass is up (63a0a7e41).

- Both spec pages now use the agreed vocabulary throughout (model / grants / permissions / guards / personas); "role" and "grant bundle"/"bundling" are retired except where quoting legacy code identifiers or UI copy (marked legacy inline). Core Principle rewritten around the doctrine sentence; bundling talk collapsed to the one agreed line.
- Preread is noticeably shorter: "Why This Is Needed" cut, "Current UI Facts" collapsed to a status table, terminology tightened into a five-term section.
- Decision matrix keeps its scenario structure, lexicon pass only.
- Diagram swapped to the fully-relabeled version. Two gate-chip labels in the SVG are still drifted from the doc text after this pass (Context Entry still says "auditor / explicit role", Actions still says "action-specific relation on target") — flagged in the commit message rather than silently shipped; let me know if you want those chips relabeled too.
- Added the Health Metrics / writer-vs-marketing_auditor question to Model Asks exactly as posed to you — left open, not resolved unilaterally.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+Me item or create action + resolved target object (project, foundation, committee, group)
++ the action-specific permission on that object = allowed create/manage action
+```
+
+The target is usually Foundation/Project and the permission is `writer`, but
+that's the common case, not the only one. The shipped create flow (PR #1193)
+also resolves committee/group targets for meetings, surveys, and votes:
+`writerGuard` accepts `committee.writer` for those three features when a
+`committee_uid` is present, and accepts `project.meeting_coordinator` for
+meetings specifically, independent of `project.writer`. A committee writer
```
emsearcy: Nit: this entire paragraph sounds like an anti-pattern because it intends to reproduce the inheritance defined in the model into multiple checks in the UI ... gating a contextual "meeting edit" action from within Me lens should be done by checking the **"meeting organizer" permission** _against the meeting object itself_ ... not trying to extrapolate connected projects, committees, and checking permissions against those. Match the API's guards -- these will always be against the objects that are being interacted with (mailing lists, meetings, B2B organizations, etc etc).

This aligns to my reply about whether the UI should implement an "OR" guard over both "writer" and "marketing_auditor" for marketing analytics. I said:
> guards should be on a single permission, and the inheritance should be expressed in the model, whenever possible

*If* there are exceptions to this rule, then, they would be "AND" guards: e.g. you must be an auditor AND you must be a member of staff to see this privileged data ... _excluding_ community-authorized writers/auditors from something. Moreover, you'd expect to see a Heimdall RuleSet that likewise makes *two* OpenFGA checks and asserts that both are true!

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+or meeting coordinator can legitimately hold no direct project-level
+auditor/writer grant.
+
+Examples:
+
+- **Pending agenda action:** resolve the target group/project/foundation; allow agenda management only if writer permission applies.
+- **Meeting card:** open the meeting with its context; edit/manage only if writer permission applies.
```
emsearcy: should be "only if _meeting organizer_ permission applies"

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+- **Vote or survey:** open the item; view/results follow viewer/discoverable eligibility for that item, edit/close follows writer permission.
```
emsearcy: not sure if it's needed here, but it would be "writer permission for that item" as well

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+- **Document row:** open the document context; upload/folder/link actions follow writer permission.
+- **Newsletter draft:** open the draft with its target audience context; edit/delete/send follows writer permission for that context.
+- **Newsletter create:** ask for the target Foundation/Project and audience first; then apply writer permission.
+- **Akrites package:** open package drawer/workflow; stewardship follows Akrites request/stewardship rules.
```
emsearcy: I'm not aware of any "project package" entity in OpenFGA, nor any "package stewardship" relations on projects or repos: so I suspect any guards around this are implementing some other pattern entirely and should be called out as an exception, as such.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+also resolves committee/group targets for meetings, surveys, and votes:
+`writerGuard` accepts `committee.writer` for those three features when a
+`committee_uid` is present, and accepts `project.meeting_coordinator` for
+meetings specifically, independent of `project.writer`. A committee writer
+or meeting coordinator can legitimately hold no direct project-level
+auditor/writer grant.
+
+Examples:
+
+- **Pending agenda action:** resolve the target group/project/foundation; allow agenda management only if writer permission applies.
```
emsearcy: From AI:
> The "Agenda" action type relates to meetings, but it's a passive reminder row, not an actionable task. For any upcoming meeting (within ~2 weeks, not yet ended) where the user doesn't already have an RSVP/Vote/Survey action pending, the dashboard shows "Review Meeting Agenda and Materials" with a button linking to /meetings/{id}. It's ranked lowest priority/severity since — unlike RSVP or Vote — there's nothing to submit; it's just prompting the user to go look at the meeting page.

This example is therefore misleading: there is no writeable "agenda management" pertaining to _agenda_ action items. And opening a meeting would simply overlap with the next line (see my response)

How about this instead:

> - **Pending action item:** expand and validate the permission based on the action item type ("vote participant" to cast a ballot, etc)

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+Target writerGuard = resolved target context + canWrite()
+```
+
+ED-shaped pages gate on the existing named permissions — checked like any
+other permission, never on a persona guard and never by branching on the
+`executive_director` grant from application code:
+
+- **Campaigns** checks `campaign_manager` (write-capable; the model currently feeds it from `executive_director` and the `marketing_ops` team). Write authority is not implied by `writer` — decided, see Model Asks below.
+- **Marketing Impact** checks `marketing_auditor` (read-only; the model currently feeds it from `executive_director`, `marketing_ops`, and parent-project inheritance). The model is also adding `writer` → `marketing_auditor`, so context writers see it without the guard knowing about `writer` — decided, see Model Asks below.
+- **Health Metrics** guards on `auditor` — decided, see Model Asks below. Anyone who can audit the context sees its health; auditor alone never implies write.
```
emsearcy, line 233: does this need to be also gated on *staff*? e.g. can a community-authorized auditor (or writer, who inherits auditor) see Health Metrics? In PCC I believe this was staff-only.

See my comments above about "AND" gates being the edge case for multiple lookups. An explicit staff guard can be easily implemented with a "does `user:USERNAME` have relation `member` to `team:lf-staff`" check, using the exact same access-check endpoint used for entity permissions.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+today, all three pages on `main` still gate solely on
+`executiveDirectorGuard` (a pure persona check with no permission lookup).
+Until that ticket merges, treat Current State above as describing the
+actual state, not this target state.
+
+Whether the `executive_director` grant itself stays in the model, is
+renamed, or is fed differently is the platform team's call, not the app's
+(see Model Asks below) — the app checks `marketing_auditor` and
+`campaign_manager` directly and does not care what feeds them. Create/
+edit/manage routes should not use ED as an authorization shortcut unless
```
emsearcy: I would go further and say one should never use an "authorization shortcut" -- the authorization check (UI guard) should match the API guard.

Line 35 states this succinctly and, IMO, doesn't leave room for "shortcuts":

> model it in OpenFGA, wire it into a guard (a Heimdall RuleSet on the API route, plus a matching UI pre-check for early bail-out or conditional nav/tab visibility).

docs/architecture/frontend/permission-persona-navigation-model-preread.md
```diff
@@ -1,70 +1,511 @@
...
+   per-object grants from a non-recursive system hierarchy, with distinct
+   relations for global auditors (all LF staff) and global writers
+   (support team) — Eric Searcy's LFXV2-2937, which supersedes 2936's
+   suggested mechanism (same objective, different implementation). Either
+   way: a model change, not a UI workaround, and app code is unaffected —
+   it only ever checks the permission on the org object.
+
+3. **Health Metrics and Marketing Impact gating — decided (Slack, Jul
+   31–Aug 4).** Guards check a single permission; inheritance belongs in
+   the model, not in the guard (Eric's rule — no multi-permission ORs).
```
emsearcy, line 487: yes, but again I'll call out that this (Health Metrics) could require an `AND` (to restrict to staff) -- but that's a product/business decision, not a platform team decision. And an `AND` actually _does_ mean TWO checks. :-)

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Agreed, and reworked per your rule. Split the Me Lens paragraph in two: create-time target resolution keeps checking one permission per chosen parent (unchanged), but edit/manage of an existing item now checks the single permission the model already resolved onto that item's own object — a scheduled meeting checks `organizer` (inherited from Project Meeting Coordinator, Committee Writer, Project Writer), not a re-derived OR of those three. Matches the API's guards, which are always against the interacted-with object. See 8ba162f13.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Done — now reads "edit/manage only if `organizer` permission applies to that meeting." See 8ba162f13.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Done — now reads "edit/close follows writer permission for that item." See 8ba162f13.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Good catch — flagged explicitly now: "an exception to the OpenFGA model: no 'package' entity or 'stewardship' relation exists today, so this follows Akrites' own request/stewardship rules rather than a permission check." See 8ba162f13.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Took your wording — renamed to "Pending action item," permission validated per action-item type (e.g. response-owner for a vote/survey ballot, `organizer` for an agenda-bearing meeting), and called out that a passive reminder like an upcoming agenda has no writeable action of its own. See 8ba162f13.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Decided in Slack (Jul 31-Aug 4, you and me): auditor-only, no staff AND-guard — a community-authorized auditor (or writer, who inherits auditor) does see Health Metrics. Recorded in 9cdce5689 and unchanged here. Reasoning: Health Metrics is read-only data about a context's standing; anyone who can audit the context is meant to see it, and auditor alone never implies write. No deliberate staff-only carve-out is intended for this page (differs from PCC's staff-only framing by design).

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Agreed — removed the carve-out entirely. Now reads: "Create/edit/manage routes must never use ED as an authorization shortcut — the UI guard checks the same permission the API guard checks, full stop." See 8ba162f13.

docs/architecture/frontend/permission-persona-navigation-model-preread.md
manishdixitlfx: Made the call: auditor-only, not auditor AND staff. Recorded in 9cdce5689 (Model Ask #3) and referenced from the Writer Actions bullet at what's now line 221. Reasoning: Health Metrics is read-only, and a community member's auditor grant is meant to carry the same weight here as staff's inherited auditor — no deliberate exclusion of community-granted auditors intended for this page.

manishdixitlfx:
## Round 2 review feedback addressed

Commit: 8ba162f13

### Changes made (per @emsearcy)
- **Me Lens target/permission paragraph** (preread): split create-time target resolution (unchanged — one permission per chosen parent) from edit/manage of an *existing* item, which now checks the single permission the model resolved onto that item's own object — a meeting checks `organizer`, inherited from Project Meeting Coordinator, Committee Writer, and Project Writer, not a re-derived OR in the UI.
- **Examples bullets**: "Pending agenda action" → "Pending action item" (permission validated per action-item type); "Meeting card" and "Vote or survey" now name the single permission checked; "Akrites package" flagged as an explicit exception to the OpenFGA model.
- **ED-shortcut paragraph**: removed the "unless the user also has writer permission" carve-out — no shortcut, full stop.
- **Decision matrix**: reworded the write-candidate/existing-item permission language to `organizer`; split the Meetings Create/manage bullet into Create (single permission on the chosen target) and Manage-existing (single `organizer` check); reworded Votes/Surveys Create/manage bullets to drop the literal "OR" phrasing.
- **Health Metrics staff-AND-guard question**: decided auditor-only (no staff AND-guard) — recorded in 9cdce5689 (prior commit) and referenced again here; anyone who can audit a context sees its Health Metrics, auditor alone never implies write.
- **Diagram**: synced `permission-model.svg` — card 4 now reads "the action's own permission on its object" with the meeting → `organizer` example.

### Note on the diagram
Same pre-existing lexicon drift flagged in the prior round is still baked into the SVG's own labels ("auditor / explicit role" contains "role"; "executive_director relation" / "evaluated relationships" contain "relation"). Not hand-edited per established precedent (af714ddf8) — flagging again since it's still there.

### Threads resolved
8 of 8 unresolved threads addressed and resolved in this iteration.

emsearcy (dismissed):

dealako:
@manishdixitlfx can you fix the lint error? Is this PR still needed from 2 weeks ago?

manishdixitlfx:
@dealako fixed — `markdown-lint` is green now (9cbb54b22). The failure was a false-positive MD018 (line-leading `#1193)` from a hard-wrap coincidentally landing at the start of a line, read as a malformed ATX heading). Fixed by reflowing that one line break; no wording changed.

Yes, still needed — this is the agreed permission/persona spec of record from Eric Searcy's review, addressed across two rounds (lexicon pass 63a0a7e41, round 2 8ba162f13, wrapped up 2026-08-04). Ready for final review/merge.

manishdixitlfx:
@emsearcy this is ready to merge — markdown-lint is fixed (9cbb54b22) and all checks are green. Branch protection needs an approving review before it can merge; could you approve when you get a chance? This is the spec of record from your two review rounds (lexicon pass 63a0a7e41, round 2 8ba162f13).

dealako (approved):
Hi @manishdixitlfx 👋 — reviewing this as a fresh independent pass on top of the bots and Eric's live review.

**Overall:** This is an unusually rigorous docs-only PR. Across the 12-commit arc it converged the permission-not-persona doctrine, absorbed every Copilot/CodeRabbit finding, and — most valuably — took Eric Searcy's authoritative OpenFGA corrections directly into the model text. I re-verified the current file state against this repo's own code and every code-backed claim holds:

- `writerGuard`'s ED fast path (`currentPersona() === 'executive-director'` → `return true`), plus its `project.writer` → `meeting_coordinator` (meetings only) → `committee.writer` (meetings/surveys/votes when `committee_uid` present) priority order — matches `writer.guard.ts` exactly. ✅
- `newsletter-access.guard.ts` carries the same ED fast path. ✅
- Health Metrics, Marketing Impact, and Campaigns all gate solely on `executiveDirectorGuard` (pure persona, no permission lookup) on `main` — confirmed in `app.routes.ts`. ✅
- `PersonaDetectionService` derives `maintainer` from a `cdp_roles` entry's `role` field (legacy naming, not an FGA relation) — matches the doc's careful narrowing. ✅
- The `#meetings-write-paths` cross-link into `persona-content-matrix.md` resolves. ✅
- `markdownlint-cli2`: 0 errors on both files. ✅

**Bot reconciliation:** All 41 review threads are resolved; the Copilot defaulting-determinism and committee-writer findings and both CodeRabbit findings were correctly addressed in `e6b5e3e`/`fde6e89`/`de5ac20`. I agree with those resolutions and found nothing they missed.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2
  1. **SVG lexicon drift.** `permission-model.svg` still visually contains the retired terms the prose deliberately removed — `grep` finds `role` (×1), `relation` (×3), `bundle` (×1) in the embedded diagram. You flagged this in the commit messages as an intentional non-hand-edit of the design-source SVG, but that acknowledgment lives in git history, not in the rendered doc — a reader sees prose and diagram disagree. Fine to defer to a design-source regen, but consider a one-line caption noting the diagram labels lag the agreed lexicon.
  2. **Contract Summary flattening.** In the preread, `Action authority -> existing contextual writer permission` collapses to plain "writer," which is the exact flattening the body (and Eric) worked to remove — the decision matrix's own Decision Rules block says it better: "the action-specific permission it requires (often writer, not always)." Consider aligning the summary line.
- ❔ Question: 0

**Decision: ✅ Approved.** Docs-only, no code touched, mergeable with no conflicts, every factual claim verified against the live guards and routes. The two nits are polish and needn't block merge.

Nice work driving this to a genuinely code-faithful spec.

PR #1264, opened 2026-07-29
feat(weekly-brief): share to mailing list action

manishdixitlfx:
## Summary

Adds a "Share to Mailing List" action to the weekly-brief card, sending the current saved brief via email. Authorization is project:{project_uid}#writer (not committee.writer, and not the ticket's original chair/admin framing — see below). Confirmation dialog shows the true recipient audience before sending; disabled with a visible reason when the committee has no mailing list configured.

## Key deviation from the ticket's stated premise

LFXV2-2914 assumes an "existing mailing-list send endpoint." **No such endpoint exists** — verified against this repo's `mailing-list.controller/service/route.ts` (CRUD only) and the upstream `lfx-v2-mailing-list-service` OpenAPI spec (no send/message operation). This PR instead repurposes the **newsletter** send pipeline (create + send), whose recipients resolve from `committee_uids` (committee membership), not the Groups.io mailing list. Full rationale in the branch's plan file / commit history.

**Known, accepted trade-off:** every share creates a real `Newsletter` record, visible in the project's Newsletters dashboard (subject prefixed `[Weekly Brief] ...` so it's at least self-identifying). Confirmed with the requester before implementing.

## Server-side safety guarantees added during review

- `shareBrief` enforces brief-exists / caller-is-project-writer / committee-has-mailing-list / revision-matches preconditions **before** the live/mock backend check — mock mode (the default, `WEEKLY_BRIEF_BACKEND` unset) now fails loudly (409 `BACKEND_NOT_LIVE`) instead of ever sending a real email from placeholder brief text.
- Orphaned draft newsletters are cleaned up only on deterministic 4xx rejections; ambiguous failures (timeouts, 5xx, network drops) leave the draft in place since the async send may have already been accepted upstream — client copy for those cases tells the user to check the Newsletters list rather than inviting a duplicate send.
- `has_mailing_list` is computed via a query-service count, not the native `GET /committees/{uid}` field (verified unreliable in prod — LFXV2-2914). `shareBrief` uses a strict variant (`hasMailingListStrict`) that propagates a transient query-service failure instead of misreporting it as "no mailing list configured"; the committee detail page's cosmetic badge keeps the existing fail-open-to-false batch check.
- Authorization uses `checkSingleAccessStrict`, not `checkSingleAccess` — same rationale: a transient access-check outage must surface as a retryable error, not a misleading 403.

## Out of scope / pre-existing (flagged, not fixed here)

- The Angular `WeeklyBriefService.getWeeklyBrief` swallows all GET failures into a fake success, making the card's existing "unavailable" retry state (from the base branch) currently unreachable. Pre-existing, not touched by this diff — worth a follow-up ticket.
- No `-robust.spec.ts` structural companion exists for this card — pre-existing gap from the base branch's original spec.

## Test plan

- [x] `yarn lint && yarn format && yarn check-types && yarn build` all pass
- [x] `yarn test` — shared package and `apps/lfx-one` (vitest, server-scoped) both pass; `weekly-brief.service.spec.ts` covers the mock/live-mode and revision-conflict paths for getCurrentBrief/generateBrief/saveBrief
- [ ] `yarn e2e --grep "weekly-brief-card"` — spec extended with 6 new scenarios (happy path, disabled state, send failure, 400 validation, BACKEND_NOT_LIVE, regenerate-disabled-hint parity); not run against a live dev server in this session
- [ ] Manual smoke via a running dev server: generate a brief, confirm Share disabled without a mailing list / enabled with one, confirm the resulting Newsletter appears (expected) in the project's Newsletters list

### DIFF
--- a/apps/lfx-one/e2e/weekly-brief-card.spec.ts
+++ b/apps/lfx-one/e2e/weekly-brief-card.spec.ts
@@ -41,7 +41,7 @@
 
 import { expect, Page, Route, test } from '@playwright/test';
 import { CommitteeMemberRole } from '@lfx-one/shared/enums';
-import { Committee, WeeklyBrief, WeeklyBriefCurrentResponse, WeeklyBriefThrottle } from '@lfx-one/shared/interfaces';
+import { Committee, ShareWeeklyBriefResult, WeeklyBrief, WeeklyBriefCurrentResponse, WeeklyBriefThrottle } from '@lfx-one/shared/interfaces';
 
 const TEST_COMMITTEE_UID = 'wb-card-e2e-committee-uid';
 // Committees are mounted under /groups, not /committees (see committee-about.helper.ts's
@@ -85,7 +85,7 @@ const USED_THROTTLE_AFTER_GENERATE: WeeklyBriefThrottle = {
   generates_used: 1,
 };
 
-function buildCommitteeFixture(): Committee {
+function buildCommitteeFixture(overrides: Partial<Committee> = {}): Committee {
   return {
     uid: TEST_COMMITTEE_UID,
     name: 'Weekly Brief Test WG',
@@ -105,6 +105,7 @@ function buildCommitteeFixture(): Committee {
     // Without a my_role here the card never renders regardless of the flag or canEdit —
     // matches committee-about.helper.ts's buildBaseCommittee, which sets this too.
     my_role: CommitteeMemberRole.CHAIR,
+    ...overrides,
   };
 }
 
@@ -118,7 +119,7 @@ function buildCommitteeFixture(): Committee {
  * explicitly mocked here (lens, project context, mailing lists, etc.) is left
  * to the dev backend — the card only reads `committee.uid` and `canEdit`.
  */
-async function mockCommitteeShell(page: Page): Promise<void> {
+async function mockCommitteeShell(page: Page, committeeOverrides: Partial<Committee> = {}): Promise<void> {
   await page.route(`**/api/committees/${TEST_COMMITTEE_UID}`, async (route) => {
     if (route.request().method() !== 'GET') {
       await route.fallback();
@@ -127,7 +128,7 @@ async function mockCommitteeShell(page: Page): Promise<void> {
     await route.fulfill({
       status: 200,
       contentType: 'application/json',
-      body: JSON.stringify(buildCommitteeFixture()),
+      body: JSON.stringify(buildCommitteeFixture(committeeOverrides)),
     });
   });
 
@@ -226,6 +227,19 @@ async function mockCurrentBrief(page: Page, initial: WeeklyBriefCurrentResponse)
       current = next;
     },
   };
+}
+
+/**
+ * Mock POST /api/committees/:uid/weekly-briefs/share with a fixed status/body.
+ */
+async function mockShareBrief(
+  page: Page,
+  status: number,
+  body: ShareWeeklyBriefResult | { error: string; code: string; errors?: { field: string; message: string; code: string }[] }
+): Promise<void> {
+  await page.route(`**/api/committees/${TEST_COMMITTEE_UID}/weekly-briefs/share`, async (route) => {
+    await route.fulfill({ status, contentType: 'application/json', body: JSON.stringify(body) });
+  });
 }
 
 /**
@@ -585,5 +599,187 @@ test.describe('WG Weekly Brief card — read failure (flag ON)', () => {
     await page.getByTestId('weekly-brief-card-unavailable-retry-button').click();
     await expect(page.getByTestId('weekly-brief-card-empty-state')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
     await expect(page.getByTestId('weekly-brief-card-unavailable-state')).toHaveCount(0);
+  });
+});
+
+test.describe('WG Weekly Brief card — Share to Mailing List (flag ON)', () => {
+  test('sends the current brief and shows a success toast', async ({ page }) => {
+    await mockCommitteeShell(page, { has_mailing_list: true, mailing_list: 'wg-tsc@lists.example.org' });
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE });
+
+    let shareCalled = false;
+    await page.route(`**/api/committees/${TEST_COMMITTEE_UID}/weekly-briefs/share`, async (route) => {
+      shareCalled = true;
+      const result: ShareWeeklyBriefResult = { committee_name: 'Weekly Brief Test WG', total_recipients: 4 };
+      await route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(result) });
+    });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    const shareBtn = page.getByTestId('weekly-brief-card-share-button');
+    await expect(shareBtn).toBeVisible();
+    await expect(shareBtn).toBeEnabled();
+    await shareBtn.click();
+
+    // Confirm dialog appears, describing the true recipient audience
+    // (committee members) — never the Groups.io mailing-list address, which
+    // is not who actually receives the email.
+    await expect(page.locator('.p-confirmdialog')).toBeVisible({ timeout: 5000 });
+    await expect(page.locator('.p-confirmdialog')).toContainText('all members of Weekly Brief Test WG');
+
+    const sharePromise = page.waitForRequest(
+      (req) => req.method() === 'POST' && req.url().includes(`/api/committees/${TEST_COMMITTEE_UID}/weekly-briefs/share`),
+      { timeout: DATA_LOAD_TIMEOUT }
+    );
+    await page.locator('.p-confirmdialog').getByRole('button', { name: /Send/i }).click();
+    await sharePromise;
+
+    expect(shareCalled).toBe(true);
+    await expect(page.getByText(/Brief queued for delivery to 4 recipients/i)).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+  });
+});
+
+test.describe('WG Weekly Brief card — Share disabled (no mailing list)', () => {
+  test('Share button is visible but disabled, with a visible hint, when the committee has no mailing list', async ({ page }) => {
+    await mockCommitteeShell(page, { has_mailing_list: false });
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    const shareBtn = page.getByTestId('weekly-brief-card-share-button');
+    await expect(shareBtn).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(shareBtn).toBeDisabled();
+
+    // The disabled reason is a plain visible hint (not a tooltip) so it's
+    // reachable without hover/focus for keyboard and screen-reader users.
+    const hint = page.getByTestId('weekly-brief-card-share-disabled-hint');
+    await expect(hint).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(hint).toContainText('No mailing list configured for this committee');
+  });
+});
+
+test.describe('WG Weekly Brief card — Share failure', () => {
+  test('a failed send surfaces an error toast (no silent drop)', async ({ page }) => {
+    await mockCommitteeShell(page, { has_mailing_list: true, mailing_list: 'wg-tsc@lists.example.org' });
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE });
+    await mockShareBrief(page, 409, { error: 'Committee has no mailing list configured', code: 'NO_MAILING_LIST' });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    await page.getByTestId('weekly-brief-card-share-button').click();
+    await expect(page.locator('.p-confirmdialog')).toBeVisible({ timeout: 5000 });
+
+    const sharePromise = page.waitForRequest(
+      (req) => req.method() === 'POST' && req.url().includes(`/api/committees/${TEST_COMMITTEE_UID}/weekly-briefs/share`),
+      { timeout: DATA_LOAD_TIMEOUT }
+    );
+    await page.locator('.p-confirmdialog').getByRole('button', { name: /Send/i }).click();
+    await sharePromise;
+
+    await expect(page.getByText(/No mailing list configured for this committee/i)).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+  });
+
+  test('a 400 validation error surfaces the actual field message, not the generic envelope text', async ({ page }) => {
+    await mockCommitteeShell(page, { has_mailing_list: true });
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE });
+    await mockShareBrief(page, 400, {
+      error: 'Validation failed for brief_text',
+      code: 'VALIDATION_ERROR',
+      errors: [
+        { field: 'brief_text', message: 'Brief is too long to share (must render to 100000 characters or fewer as HTML)', code: 'FIELD_VALIDATION_ERROR' },
+      ],
+    });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    await page.getByTestId('weekly-brief-card-share-button').click();
+    await expect(page.locator('.p-confirmdialog')).toBeVisible({ timeout: 5000 });
+    await page.locator('.p-confirmdialog').getByRole('button', { name: /Send/i }).click();
+
+    await expect(page.getByText(/Brief is too long to share/i)).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(page.getByText('Validation failed for brief_text', { exact: true })).toHaveCount(0);
+  });
+
+  test('a 409 BACKEND_NOT_LIVE reports the environment reason distinctly from a no-mailing-list conflict', async ({ page }) => {
+    await mockCommitteeShell(page, { has_mailing_list: true });
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE });
+    await mockShareBrief(page, 409, { error: 'Sharing is not available in this environment', code: 'BACKEND_NOT_LIVE' });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    await page.getByTestId('weekly-brief-card-share-button').click();
+    await expect(page.locator('.p-confirmdialog')).toBeVisible({ timeout: 5000 });
+    await page.locator('.p-confirmdialog').getByRole('button', { name: /Send/i }).click();
+
+    await expect(page.getByText(/Sharing is not available in this environment yet/i)).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+  });
+
+  test('a 409 REVISION_MISMATCH prompts a reload instead of silently sending stale content', async ({ page }) => {
+    await mockCommitteeShell(page, { has_mailing_list: true });
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE });
+    await mockShareBrief(page, 409, { error: 'The brief has been updated since you last viewed it', code: 'REVISION_MISMATCH' });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    await page.getByTestId('weekly-brief-card-share-button').click();
+    await expect(page.locator('.p-confirmdialog')).toBeVisible({ timeout: 5000 });
+
+    const sharePromise = page.waitForRequest(
+      (req) => req.method() === 'POST' && req.url().includes(`/api/committees/${TEST_COMMITTEE_UID}/weekly-briefs/share`),
+      { timeout: DATA_LOAD_TIMEOUT }
+    );
+    const sharedBody = sharePromise.then((req) => req.postDataJSON() as { revision?: number });
+    await page.locator('.p-confirmdialog').getByRole('button', { name: /Send/i }).click();
+    await sharePromise;
+
+    // The client sends back the revision it displayed — proves the confirmation was
+    // tied to a specific version of the brief, not just "whatever is current".
+    expect((await sharedBody).revision).toBe(GENERATED_BRIEF.revision);
+
+    await expect(page.getByText(/updated since you last viewed it.*Reload to review the latest version/i)).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+  });
+
+  test('a 5xx/ambiguous failure warns that the send may already have gone out, instead of inviting a retry', async ({ page }) => {
+    await mockCommitteeShell(page, { has_mailing_list: true });
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE });
+    await mockShareBrief(page, 503, { error: 'Upstream newsletter service unavailable', code: 'SERVICE_UNAVAILABLE' });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    await page.getByTestId('weekly-brief-card-share-button').click();
+    await expect(page.locator('.p-confirmdialog')).toBeVisible({ timeout: 5000 });
+    await page.locator('.p-confirmdialog').getByRole('button', { name: /Send/i }).click();
+
+    // This is the safeguard that prevents a duplicate send: a 5xx here must
+    // NOT render the generic "try again" copy, since the async send may
+    // already have been accepted upstream before the failure reached the client.
+    await expect(page.getByText(/The send may not have completed/i)).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(page.getByText('Failed to share brief. Please try again.', { exact: true })).toHaveCount(0);
+  });
+});
+
+test.describe('WG Weekly Brief card — Regenerate disabled (throttle exhausted)', () => {
+  test('Regenerate button is visible but disabled, with a visible hint, when the weekly regeneration limit is reached', async ({ page }) => {
+    await mockCommitteeShell(page);
+    const exhaustedThrottle: WeeklyBriefThrottle = { ...DEFAULT_THROTTLE, regenerations_used: 3, regenerations_limit: 3 };
+    await mockCurrentBrief(page, { brief: GENERATED_BRIEF, throttle: exhaustedThrottle });
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    const regenerateBtn = page.getByTestId('weekly-brief-card-regenerate-button');
+    await expect(regenerateBtn).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(regenerateBtn).toBeDisabled();
+
+    const hint = page.getByTestId('weekly-brief-card-regenerate-disabled-hint');
+    await expect(hint).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(hint).toContainText('Weekly regeneration limit reached');
   });
 });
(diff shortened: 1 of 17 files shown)


apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -171,6 +205,145 @@ export class WeeklyBriefService {
...
+        }
+      );
+    }
+
+    const newsletter = await this.newsletterService.createNewsletter(req, committee.project_uid, {
+      subject,
+      body_html: bodyHtml,
+      ed_reply_email: edReplyEmail,
+      committee_uids: [committeeId],
+    });
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch — this was a real gap, not just a messaging issue. `committee.writer` reflects a direct committee-level grant OR inheritance from project writer (per `committee.interface.ts`'s own doc comment: `committee#writer` derives from `writer from project`, but a direct committee-level grant doesn't imply the reverse). A committee chair with only the direct grant would pass our check and then hit the newsletter service's `project:{project_uid}#writer` requirement.

Fixed in 319bd50 by swapping `req.bearerToken` for an M2M token around the newsletter create/send calls, restored in a `finally` block — same pattern already used in `meeting.controller.ts`'s `getMyMeetingRegistrants` for an equivalent situation (in-app authorization already validated at a narrower scope than what the downstream call itself checks). The `committee.writer` check above remains the actual authorization gate; the M2M token just lets that decision govern the call instead of the newsletter service's own per-user FGA.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -171,6 +205,145 @@ export class WeeklyBriefService {
...
+        );
+      }
+      throw error;
+    }
+
+    logger.info(req, 'share_weekly_brief_sent', 'Weekly brief shared via newsletter send pipeline', {
+      committee_id: committeeId,
+      newsletter_id: newsletter.id,
+      total_recipients: sendResult.total_recipients,
+    });
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 319bd50 — renamed to `share_weekly_brief_queued` and reworded the message to reflect that the newsletter is queued, not confirmed delivered. Matches the client-side toast copy ("Brief queued for delivery..."), which already made this distinction; the server log had drifted from it.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.html
```diff
@@ -72,12 +72,29 @@
...
+              (onClick)="onShareToMailingList()"
+              data-testid="weekly-brief-card-share-button" />
           </div>
+          <!-- Visible hints rather than tooltips on these disabled buttons: a
+               disabled native <button> never receives focus/hover, so a
+               tooltip-only explanation would be unreachable by keyboard and
+               screen-reader users. -->
+          @if (!canRegenerate()) {
+            <span class="text-xs text-gray-500" data-testid="weekly-brief-card-regenerate-disabled-hint">Weekly regeneration limit reached</span>
+          }
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 319bd50 — added a test mocking the weekly regeneration throttle exhausted and asserting `weekly-brief-card-regenerate-disabled-hint` renders with the expected copy. Good catch on the PR description overstating coverage.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -199,4 +214,68 @@ export class WeeklyBriefCardComponent {
...
+            // per-field `errors[]` array.
+            const fieldErrors = (err?.error as { errors?: { message?: string }[] } | undefined)?.errors;
+            detail = fieldErrors?.[0]?.message ?? 'Failed to share brief. Please try again.';
+          } else if (status === 0 || status === 408 || status >= 500) {
+            // The send is async — a dropped connection, timeout, or 5xx here
+            // may mean the newsletter was already accepted upstream before
+            // the failure reached the client. Don't invite a retry that
+            // could send the brief twice. (status === 0 covers network
+            // failures/aborts, which never surface an HTTP status.)
+            detail = 'The send may not have completed — check the project’s Newsletters list before trying again.';
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 319bd50 — added a test mocking a 503 share response and asserting the ambiguous-failure copy ("The send may not have completed...") renders instead of the generic retry copy. You're right that the only prior failure test exercised a pre-send 409, leaving this actual safeguard untested.

manishdixitlfx:
## Review Feedback Addressed

Commit: 319bd50b8

### Changes Made
- **apps/lfx-one/src/server/services/weekly-brief.service.ts**: swapped `req.bearerToken` for an M2M token around the newsletter create/send calls (restored in a `finally` block) — `committee.writer` doesn't imply `project:{project_uid}#writer`, which the newsletter service itself enforces, so a committee chair with only a direct committee-level grant would previously pass our check and then 403 downstream (per copilot-pull-request-reviewer[bot])
- **apps/lfx-one/src/server/services/weekly-brief.service.ts**: renamed the `share_weekly_brief_sent` log event to `share_weekly_brief_queued` — the newsletter API only accepts/queues the send; fan-out is an async detached job that can still fail completely afterward (per copilot-pull-request-reviewer[bot])
- **apps/lfx-one/e2e/weekly-brief-card.spec.ts**: added an e2e case for the exhausted-regeneration-throttle disabled hint, which the PR description claimed was covered but wasn't (per copilot-pull-request-reviewer[bot])
- **apps/lfx-one/e2e/weekly-brief-card.spec.ts**: added an e2e case mocking a 503 and asserting the ambiguous-failure copy renders instead of generic retry copy — the only prior failure test exercised a pre-send 409 (per copilot-pull-request-reviewer[bot])

### Threads Resolved

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -171,6 +207,175 @@ export class WeeklyBriefService {
...
+    // *direct* committee-level writer grant (the common case for an ordinary
+    // committee chair) is NOT necessarily a project writer, so calling the
+    // newsletter endpoints with the user's own bearer token would pass our
+    // check here and then 403 downstream. Since we've already made the
+    // authorization decision in-app, use an M2M token for these specific
+    // privileged calls — same pattern as meeting.controller.ts's
+    // getMyMeetingRegistrants — rather than requiring every committee chair
+    // to also hold project-level access.
+    const originalToken = req.bearerToken;
+    req.bearerToken = await generateM2MToken(req);
```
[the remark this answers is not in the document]
manishdixitlfx: Done — dropped the M2M-token swap entirely in 6afc0818. Instead of trying to preserve the human sender identity through a delegated/on-behalf-of call (which doesn't exist in this codebase), the authorization check itself now targets the actual boundary the newsletter service enforces: `project:{project_uid}#writer`, checked directly via `AccessCheckService.checkSingleAccess` using the caller's own bearer token. No token swap, no sender-identity breakage — the newsletter create/send calls now run with the caller's real token throughout.

This does narrow who can use Share versus the ticket's original chair/admin framing — a committee chair without project-writer access now gets a 403 instead of a broken send. Flagged explicitly in the updated JSDoc above `shareBrief`.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -171,6 +207,175 @@ export class WeeklyBriefService {
...
+   * detached background job upstream) — `total_recipients` is a snapshot taken
+   * at acceptance, not a delivered/failed count.
+   *
+   * Requires `WEEKLY_BRIEF_BACKEND=live` — throws a 409 `BACKEND_NOT_LIVE`
+   * otherwise, after still enforcing the brief/writer/mailing-list
+   * preconditions below (so mock-mode testing exercises the same guards).
+   */
+  public async shareBrief(req: Request, committeeId: string): Promise<ShareWeeklyBriefResult> {
+    const { brief } = await this.getCurrentBrief(req, committeeId);
+    if (!brief || !WEEKLY_BRIEF_SHAREABLE_STATES.includes(brief.state)) {
```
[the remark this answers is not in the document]
manishdixitlfx: Done in 6afc0818 — the client now sends the revision it displayed in the confirmation dialog (`shareWeeklyBrief(committeeId, revision)`), the controller validates it's present, and `shareBrief` rejects a mismatch with a 409 `REVISION_MISMATCH` before doing anything else, mirroring the existing save/generate revision-conflict convention. On the frontend, a `REVISION_MISMATCH` response shows a reload prompt and re-triggers the brief fetch, same as the existing save/generate 409 handling. Added an e2e case (`weekly-brief-card.spec.ts`) asserting both the reload copy and that the request body actually carries the displayed revision.

manishdixitlfx:
## Review Feedback Addressed

Commit: 6afc08188

### Changes Made
- **weekly-brief.service.ts**: dropped the M2M-token swap around newsletter create/send (introduced in 319bd50 for an earlier review round) — it broke sender-identity resolution downstream. Replaced with a direct `project:{project_uid}#writer` authorization check (`AccessCheckService.checkSingleAccess`) using the caller's own bearer token throughout, no token swap (per copilot-pull-request-reviewer[bot])
- **weekly-brief.service.ts**, **weekly-brief.controller.ts**, **weekly-brief.service.ts** (Angular): added a revision/concurrency check on `shareBrief` — client sends back the displayed revision, server rejects a stale value with 409 `REVISION_MISMATCH`, mirroring the existing save/generate convention (per copilot-pull-request-reviewer[bot])
- **weekly-brief-card.component.ts**: updated the 403 toast copy for the new project-writer boundary, added a `REVISION_MISMATCH` branch that prompts a reload
- **weekly-brief-card.spec.ts**: added an e2e case for the 409 `REVISION_MISMATCH` path, asserting the reload-prompt copy and that the request body carries the displayed revision

### Known trade-off (documented in code)
Authorization is now `project:{project_uid}#writer`, not `committee.writer` — narrower than the ticket's original "chair/admin" framing. A committee chair with a direct committee-level grant but no project-writer access will get a 403. There's no delegated/on-behalf-of token mechanism in this codebase to bridge that gap without misattributing the send, so this was the safer choice over the previous M2M workaround.

### Threads Resolved

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
```diff
@@ -55,4 +56,9 @@ export class WeeklyBriefService {
   public saveWeeklyBrief(committeeId: string, body: SaveWeeklyBriefRequest): Observable<WeeklyBrief> {
     return this.http.put<WeeklyBrief>(`/api/committees/${encodeURIComponent(committeeId)}/weekly-briefs/current`, body).pipe(take(1));
   }
+
+  public shareWeeklyBrief(committeeId: string, revision: number): Observable<ShareWeeklyBriefResult> {
+    return this.http.post<ShareWeeklyBriefResult>(`/api/committees/${committeeId}/weekly-briefs/share`, { revision });
```
dealako: [minor] `shareWeeklyBrief` doesn't encode `committeeId` in the URL

Issue: every sibling method in this file (`getWeeklyBrief`, `generateWeeklyBrief`, `saveWeeklyBrief`) wraps `committeeId` in `encodeURIComponent(...)` before interpolating it into the path. `shareWeeklyBrief` interpolates it raw:
```ts
return this.http.post<ShareWeeklyBriefResult>(`/api/committees/${committeeId}/weekly-briefs/share`, { revision });
```
Proof: `committeeId` here is a UUID sourced from `committee.uid`, so this isn't currently exploitable or reachable with untrusted separators — but the inconsistency means a future caller (or a copy-paste of this method) can silently drop the encoding pattern the rest of the file enforces.
Why it matters: URL construction conventions in a shared service should be uniform; an inconsistent one is easy to miss in review since three of four methods already do it right.
Fix:
```ts
return this.http.post<ShareWeeklyBriefResult>(`/api/committees/${encodeURIComponent(committeeId)}/weekly-briefs/share`, { revision });
```

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
```diff
@@ -55,4 +56,9 @@ export class WeeklyBriefService {
   public saveWeeklyBrief(committeeId: string, body: SaveWeeklyBriefRequest): Observable<WeeklyBrief> {
     return this.http.put<WeeklyBrief>(`/api/committees/${encodeURIComponent(committeeId)}/weekly-briefs/current`, body).pipe(take(1));
   }
+
+  public shareWeeklyBrief(committeeId: string, revision: number): Observable<ShareWeeklyBriefResult> {
```
dealako, line 66: [nit] `shareWeeklyBrief` skips the JSDoc + `.pipe(take(1))` convention its siblings use

Issue: `getWeeklyBrief`, `generateWeeklyBrief`, and `saveWeeklyBrief` all have a JSDoc comment and chain `.pipe(take(1))`. `shareWeeklyBrief` has neither — it relies on the caller (`performShare` in `weekly-brief-card.component.ts`) to add `take(1)` itself, which it does, so there's no behavioral bug, just a style inconsistency within the file.
Fix: add a short JSDoc comment and `.pipe(take(1))` for consistency with the rest of the service.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -263,6 +299,185 @@ export class WeeklyBriefService {
...
+    // committee.writer is a superset of what we need here — per committee.interface.ts's
+    // own doc comment, it's true for both a direct committee-level grant AND an inherited
+    // project writer, but the newsletter service (which this repurposes for delivery)
+    // only recognizes project:{project_uid}#writer. A direct-grant-only committee writer
+    // would pass committee.writer, then 403 from the newsletter service — and there's no
+    // delegated/on-behalf-of token mechanism in this codebase to bridge that gap without
+    // misattributing the send (the newsletter service resolves the sender's display name
+    // from the signed JWT principal; an M2M token has none). Check the actual boundary
+    // directly instead, with the caller's own bearer token.
+    const isProjectWriter = await this.accessCheckService.checkSingleAccess(req, {
```
dealako: [minor] `shareBrief`'s project-writer check uses `checkSingleAccess`, not the newer `checkSingleAccessStrict`

Issue: `AccessCheckService.checkAccess` (and its `checkSingleAccess` convenience wrapper) catches upstream failures and falls back to "no access" (see `access-check.service.ts`'s `catch` block: `fallback_behavior: 'returning no access'`). `shareBrief` uses this non-strict variant:
```ts
const isProjectWriter = await this.accessCheckService.checkSingleAccess(req, {
  resource: 'project',
  id: committee.project_uid,
  access: 'writer',
});
```
Proof: `checkSingleAccessStrict` already exists in this same service specifically to let callers distinguish "resolved: no access" (403) from "couldn't verify" (503) — it's already adopted in `committee-access.internal.helper.ts` with a doc comment explaining exactly this rationale. `shareBrief` doesn't follow that established pattern, so a transient access-check outage will surface to the user as `AuthorizationError('NOT_PROJECT_WRITER')` (403 "you don't have permission") instead of a retryable service error — a misleading, non-actionable message for what is actually an upstream availability problem.
Why it matters: sharing a brief is a low-frequency, deliberate action (unlike, say, a list-view access annotation where fail-closed-to-false is an acceptable UX trade-off) — misattributing an outage as a permission denial will confuse project writers and generate support noise.
Fix: switch to `checkSingleAccessStrict` here, consistent with `committee-access.internal.helper.ts`, and let the caller/error-handler distinguish the failure mode (e.g. surface a 503/ambiguous error instead of `NOT_PROJECT_WRITER`).

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -481,4 +496,72 @@ export class WeeklyBriefCardComponent {
         },
       });
   }
+
+  // Other private helpers
+  private performShare(): void {
```
dealako: [question] Confirm dialog captures `committeeUid`/`revision` fresh at accept-time, not at open-time — is that the intended contract?

Issue: `onShareToMailingList` opens a `confirmationService.confirm(...)` dialog whose `accept` callback is `() => this.performShare()`. `performShare` then reads `this.committee()?.uid` and `this.brief()?.revision` fresh:
```ts
private performShare(): void {
  const committeeUid = this.committee()?.uid;
  const revision = this.brief()?.revision;
  ...
}
```
Proof: if the underlying `brief`/`committee` signals change while the confirm dialog is open (e.g. a background poll refreshes the brief, or the parent re-navigates to a different committee while this component instance somehow persists), the share request fires against whatever is current at accept-time rather than what the dialog's message text promised the user ("Send the current brief by email to `all members of {committee at open-time}`?").
Why it matters: this is a narrow window (dialogs are typically dismissed quickly) and the backend's own `REVISION_MISMATCH` guard (409 → reload prompt) is a real safety net if the brief changed server-side — so this is not a proven bug, more an open question about whether the confirm-time/accept-time snapshot mismatch is an intentional, accepted trade-off or worth locking in explicitly (e.g. capturing `committeeUid`/`revision` at dialog-open time and passing them through to `performShare`).
Fix (if this should be locked down): capture `committeeUid`/`revision` in `onShareToMailingList` and pass them explicitly into `performShare(committeeUid, revision)` instead of re-reading signals at accept-time.

apps/lfx-one/e2e/weekly-brief-card.spec.ts
```diff
@@ -587,3 +601,185 @@ test.describe('WG Weekly Brief card — read failure (flag ON)', () => {
     await expect(page.getByTestId('weekly-brief-card-unavailable-state')).toHaveCount(0);
   });
 });
+
+test.describe('WG Weekly Brief card — Share to Mailing List (flag ON)', () => {
+  test('sends the current brief and shows a success toast', async ({ page }) => {
```
dealako, line 606: [minor] Missing e2e coverage for the `total_recipients: 0` ("no recipients") path

Issue: this describe block covers the happy path (`total_recipients: 4`), no-mailing-list disabled state, `409 NO_MAILING_LIST`, `400` field validation, `409 BACKEND_NOT_LIVE`, `409 REVISION_MISMATCH`, and the `5xx`/ambiguous-failure copy — but there's no test asserting the `total_recipients === 0` branch in `performShare` (the "No recipients were found for this committee — nothing was sent." warning toast in `weekly-brief-card.component.ts`).
Why it matters: this is a distinct, deliberately-written UI branch (success response but zero recipients still shows a `warn` toast instead of a success toast) — it's exactly the kind of edge case that regresses silently without a dedicated assertion.
Fix: add a test mirroring `sends the current brief and shows a success toast` but with `{ committee_name: ..., total_recipients: 0 }` in the mocked response, asserting the "No recipients" warn toast renders instead of the success toast.

dealako (changes requested):
Hey @manishdixitlfx — this is a re-review of the "share weekly brief to mailing list" flow. Nice work: the async-send failure handling is genuinely careful (the 5xx/network/timeout branch explicitly avoids inviting a duplicate-send retry, with a clear comment explaining why), the `REVISION_MISMATCH` conflict pattern is consistent with the rest of the brief lifecycle (generate/save/share all guard on `revision`), and the e2e coverage across the various error codes (400/403/404/409 variants/5xx) is thorough.

**Revision tracking (since this is a re-review):**
- Resolved since the last round: the `has_mailing_list` gate now reads from the correct source after `990082bc`'s refactor of `getCommitteeById` — the earlier Copilot thread about a possibly-wrong `getCommitteesWithMailingList` function appears stale/superseded (that function no longer exists anywhere in the current tree per `git log -S`), though the thread itself doesn't look explicitly marked resolved.
- Also resolved: the revision-check and M2M feedback from `6afc0818`, and the network-failure-as-ambiguous-outcome handling from `3d9758c3`.
- Still open from prior bot passes, called out again below: the `checkSingleAccess` vs. `checkSingleAccessStrict` gap on the project-writer check, and the confirm-dialog stale-context question.

**Issue summary:**
- 🔴 Blocking: 0
- 🟡 Minor: 3 — missing `encodeURIComponent` on `shareWeeklyBrief`'s committeeId (low exploitability given it's a UUID, but breaks the file's own convention); `shareBrief`'s project-writer check uses `checkSingleAccess` instead of the already-established `checkSingleAccessStrict` pattern (conflates "no access" with "couldn't verify upstream"); missing e2e coverage for the `total_recipients: 0` warning-toast branch
- ⚪ Nit: 1 — `shareWeeklyBrief` skips the JSDoc + `.pipe(take(1))` convention its sibling methods use
- ❔ Question: 1 — the share confirm dialog reads `committeeUid`/`revision` fresh at accept-time rather than at open-time; want to confirm whether that's an accepted trade-off (the backend's `REVISION_MISMATCH` guard is a real safety net either way) or worth locking down explicitly

**Decision: 🔴 Needs changes before approval** — none of these are severe, but there are 3 minor issues plus an open question, over the 2-minor approval threshold. The `checkSingleAccessStrict` one is the one I'd weight most: it's a small, mechanical fix that brings this in line with a pattern the codebase already established elsewhere for exactly this class of problem.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -202,6 +238,185 @@ export class WeeklyBriefService {
...
+      resource: 'project',
+      id: committee.project_uid,
+      access: 'writer',
+    });
+    if (!isProjectWriter) {
+      throw new AuthorizationError('Only project writers can share the weekly brief by email', {
+        operation: 'share_weekly_brief',
+        service: 'weekly_brief_service',
+        code: 'NOT_PROJECT_WRITER',
+      });
```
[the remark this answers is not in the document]
manishdixitlfx: Done in 7afd617 — switched to `checkSingleAccessStrict`, matching `committee-access.internal.helper.ts`'s existing rationale for the same distinction. A transient access-check outage now propagates as a retryable error instead of a misleading 403 `NOT_PROJECT_WRITER`.

apps/lfx-one/src/server/services/committee.service.ts
```diff
@@ -353,20 +361,22 @@ export class CommitteeService {
...
-    // (parent-project) permissions in parallel.
-    const [settings, membership, withAccess, inheritedPermissions] = await Promise.all([
+    // Fetch settings, optional caller membership, access, optional inherited
+    // (parent-project) permissions, and optional mailing-list status in parallel.
+    const [settings, membership, withAccess, inheritedPermissions, mlCount] = await Promise.all([
       this.getCommitteeSettings(req, committeeId, { throwOnError: options.throwOnSettingsError }),
       options.includeMembership ? this.getCallerMembership(req, committeeId) : Promise.resolve(null),
       this.accessCheckService.addAccessToResource(req, committee, 'committee'),
       options.includeInheritedPermissions ? this.getInheritedPermissions(req, committee.project_uid) : Promise.resolve(null),
+      options.includeMailingListStatus ? this.getMailingListCountByCommittee(req, committeeId) : Promise.resolve(null),
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch — you're right that the PR description was stale (fixed that too, in the same commit). Added `hasMailingListStrict` — a single-committee check that propagates a query-service failure instead of the existing batch path's fail-open-to-false — and switched `shareBrief`'s own precondition check to use it (7afd617). The committee detail page's cosmetic badge keeps the existing lenient batch check, since a false-negative there is a much lower-stakes UX inconvenience than `shareBrief` silently misreporting an outage as 409 NO_MAILING_LIST.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -278,6 +279,20 @@ export class WeeklyBriefCardComponent {
...
+    // The Share action only renders outside edit mode (see the template's
+    // @if (editMode()) branch), so there is never an unsaved-edit case to
+    // guard here — the brief is always whatever was last saved.
+    this.confirmationService.confirm({
+      header: 'Share to Mailing List',
+      message: `Send the current brief by email to ${this.shareAudienceLabel()}?`,
+      icon: 'fa-light fa-paper-plane',
+      acceptLabel: 'Send',
+      rejectLabel: 'Cancel',
+      accept: () => this.performShare(),
```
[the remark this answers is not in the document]
manishdixitlfx: Done in 7afd617 — `committeeUid`/`revision` are now captured in `onShareToMailingList` before the confirm dialog opens, and passed explicitly into `performShare(committeeUid, revision)` instead of re-reading `committee()`/`brief()` at accept-time. Also added the same `takeUntil(committee$ change)` + `takeUntilDestroyed` guard `onGenerate`/`onSave` already use, so a share started on committee A whose response arrives after navigating to committee B can't clear B's `sharing` flag or fire `refresh$` against B's card.

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
manishdixitlfx: Fixed in 7afd617 — added `encodeURIComponent(committeeId)`, matching the other three methods in this file.

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
manishdixitlfx: Fixed in 7afd617 — added the JSDoc and `.pipe(take(1))`, matching the other three methods.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Same question raised by copilot-pull-request-reviewer[bot] on this file, framed there as a more likely bug — decided to lock it down rather than leave it as an accepted trade-off, given the backend's REVISION_MISMATCH guard is a safety net for stale content but doesn't cover the confirm dialog describing one committee's audience while sending against another's. Fixed in 7afd617: committeeUid/revision are now captured at dialog-open time and passed explicitly into performShare, with the same takeUntil(committee$ change) guard onGenerate/onSave already use.

apps/lfx-one/src/server/services/weekly-brief.service.ts
manishdixitlfx: Same finding as cursor[bot]'s comment on this file — fixed in 7afd617 by switching to `checkSingleAccessStrict`, matching `committee-access.internal.helper.ts`'s existing use of the strict variant for the same reason. Agreed on the severity assessment — sharing is deliberate/low-frequency, so misattributing an outage as a permission denial is the wrong trade-off here.

apps/lfx-one/e2e/weekly-brief-card.spec.ts
manishdixitlfx: Fixed in 7afd617 — added a test mirroring the happy-path case but with `total_recipients: 0` in the mocked response, asserting the "No recipients" warn toast renders (and the success-toast copy does not).

manishdixitlfx:
## Review Feedback Addressed

Commit: 7afd617b0

### Changes Made
- **weekly-brief.service.ts (server)**: switched `shareBrief`'s project-writer check from `checkSingleAccess` to `checkSingleAccessStrict` — a transient access-check outage now propagates as a retryable error instead of a misleading 403 (per cursor[bot], dealako — duplicate findings)
- **committee.service.ts**: added `hasMailingListStrict`; `shareBrief` now uses it instead of the fail-open `committee.has_mailing_list` field, so a transient query-service outage no longer misreports as 409 `NO_MAILING_LIST` (per copilot-pull-request-reviewer[bot])
- **weekly-brief-card.component.ts**: `onShareToMailingList` now captures `committeeUid`/`revision` at dialog-open time and passes them into `performShare`; added the same `takeUntil`/`takeUntilDestroyed` guard `onGenerate`/`onSave` already use; reset `sharing` on committee navigation (per copilot-pull-request-reviewer[bot], cursor[bot], dealako — three overlapping findings on the same root cause)
- **weekly-brief.service.ts (Angular client)**: added missing `encodeURIComponent(committeeId)` and JSDoc + `.pipe(take(1))` (per dealako)
- **weekly-brief-card.spec.ts**: added an e2e case for the `total_recipients: 0` path (per dealako)
- **PR description**: corrected two stale claims (`has_mailing_list` native-field claim, apps/lfx-one test-runner claim) and dropped a reference to a `COMMITTEE_SERVICE_URL` override that was never part of the merged base branch

### Threads Resolved
8 of 8 unresolved threads addressed in this iteration (2 pairs were duplicate findings from different reviewers on the same issue).

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -263,6 +299,198 @@ export class WeeklyBriefService {
     }
   }
 
+  /**
+   * POST /committees/:committeeId/weekly-briefs/share
```
dealako, line 303: [nit] CodeQL's SSRF alert on this file is still unreplied — verdict: pre-existing, low-risk, not introduced here

Issue: the CodeQL alert ("Server-side request forgery... URL depends on a user-provided value") flags this file and predates 7afd617b, with no reply yet from anyone.
Proof: every proxy call in `shareBrief` and its helpers (`getCurrentBrief`, `getCommitteeById`, `hasMailingListStrict`) already wraps `committeeId` in `encodeURIComponent(...)` — I checked the full file at 7afd617b and found no raw/unencoded interpolation. The alert is almost certainly flagging the same "path segment is user-influenced" shape that every other proxied endpoint in this service (and `microservice-proxy.service.ts` callers elsewhere in the codebase) already uses — the proxy target's base URL is fixed config (`LFX_V2_SERVICE`), not attacker-controlled; only an already-percent-encoded internal UUID varies.
Why it matters: not a new risk and not something this commit touched, but the alert is genuinely unaddressed — worth either a scoped CodeQL suppression comment (with rationale) or a follow-up ticket to review the systemic pattern across all proxy services, so it doesn't keep resurfacing as a stale, unactioned security-scanning alert on every future PR that touches this file.
Fix: not blocking for this PR — reply to/dismiss the CodeQL alert with the rationale above, or open a follow-up ticket if a codebase-wide suppression policy for this proxy pattern is wanted.

dealako (approved):
Thanks @manishdixitlfx — this is a thorough follow-up round. Nice work: the `checkSingleAccessStrict` swap now has a genuinely good inline comment explaining the rationale (fail-closed vs. fail-loud trade-off, and why sharing warrants the stricter variant unlike list-view annotations), `hasMailingListStrict` cleanly closes the same class of problem for the mailing-list precondition, the confirm-dialog fix snapshots `committeeUid`/`revision` at open-time and threads them through `performShare(committeeUid, revision)` instead of re-reading signals at accept-time, and the cross-committee-navigation guard (resetting `sharing` + a `takeUntil` on committee change) is a real fix for a subtle bug cursor's bot caught that nobody had flagged before.

**Revision tracking — all carried-over items resolved:**
- ✅ `shareWeeklyBrief` now wraps `committeeId` in `encodeURIComponent(...)`, matching its siblings.
- ✅ JSDoc + `.pipe(take(1))` added to `shareWeeklyBrief`, consistent with the rest of the file.
- ✅ `shareBrief`'s project-writer check now uses `checkSingleAccessStrict` (resolves both my finding and cursor[bot]'s independent "Access outage reported as denial" comment on the same code).
- ✅ `hasMailingListStrict` added and wired into `shareBrief`'s mailing-list gate, replacing the fail-open `has_mailing_list` field (resolves Copilot's concern about the query-service path silently defaulting to false).
- ✅ Confirm dialog now captures `committeeUid`/`revision` at open-time, not accept-time (resolves both my `[question]` and Copilot's framing of the same issue).
- ✅ e2e test added for the `total_recipients: 0` warn-toast path.
- ✅ cursor[bot]'s "Share state bleeds across committees" (unreplied, never addressed by anyone until now) — verified fixed: `sharing` is now reset on committee change and `performShare` is guarded with a `takeUntil` on committee navigation, so a stale `REVISION_MISMATCH` can no longer fire against the wrong committee.

**One informational item, not blocking:** left an inline note on the still-unreplied CodeQL SSRF alert on the server `weekly-brief.service.ts` — spot-checked the file and every proxy call already encodes `committeeId`; this looks like a pre-existing, low-risk, codebase-wide pattern (not introduced by this commit) that's worth a suppression comment or follow-up ticket rather than a fix in this PR.

**Issue summary:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — unreplied CodeQL SSRF alert, informational (see inline comment)

**Decision: ✅ Approved** — all substantive feedback from both rounds (mine and the bots') has been addressed with solid, well-reasoned fixes. Nothing here blocks merge.

PR #1265, opened 2026-07-29
feat(groups): dashboard rollup stats BFF + stat cards

manishdixitlfx:
## Summary

Adds a top-row rollup stats block to the Groups dashboard (LFXV2-1711).

**Ticket:** [LFXV2-1711](https://linuxfoundation.atlassian.net/browse/LFXV2-1711)

**Live today (unflagged):** Total Groups / Public Groups / Voting-Enabled Groups / My Groups counts are **not new** — they already shipped in #1223 as client-side `computed()` signals on `CommitteeDashboardComponent`, derived from the already-fully-paginated `/api/committees` and `/api/committees/my-committees` responses. This PR does not touch that code path.

**New in this PR, mocked pending LFXV2-1705:** Two additional stat cards — **Active Members** and **Meetings This Month** — via a new `GET /api/committees/engagement-stats` BFF endpoint, gated behind a new `wg-engagement-metrics` LaunchDarkly flag and rendered only in the Me lens (the endpoint is explicitly "mine semantics, no scope param," so it doesn't belong in the project/foundation-scoped Foundation lens row). Both stats depend on the same dbt engagement model as LFXV2-1705, which isn't readable yet.

`wg-engagement-metrics` does not exist as a flag in LaunchDarkly yet — `FeatureFlagService.getBooleanFlag(WG_ENGAGEMENT_METRICS_FLAG, false)` evaluates to its `false` default until someone creates it there, so the two engagement cards stay hidden (3-card grid) in every environment today. When the fields aren't computable (the default `live` backend, pre-dbt-deploy), each card degrades independently to an "Unavailable" em-dash state rather than blocking the other or the group-count cards.

### Mock/live backend switch — deliberate trade-off, please read

`ENGAGEMENT_BACKEND` controls the source:
- **`live` (default when unset)** — returns `null` for both fields; the client renders an "Unavailable" degraded state. This is what every deployed environment gets today.
- **`mock`** — deterministic fixtures keyed off the caller's identity, for local dev only. **Additionally hard-blocked when `NODE_ENV=production`** regardless of the env var, and left **commented out by default** in `.env.example` so `cp .env.example .env` can't silently inherit it. The response carries a `source: 'mock' | 'live'` field, and the client renders a visible **"Sample data"** marker (instead of a freshness label) whenever `source === 'mock'` — so fabricated numbers can never be mistaken for real ones, including during local validation against synced prod data (`/lfx-serve`), where `NODE_ENV` isn't `'production'` but a stray `ENGAGEMENT_BACKEND=mock` could still be set.

This shipped as a mock/live switch per the ticket's explicit design (mirroring the LFXV2-1705 approach), even though it sits in tension with this repo's documented "no mock data / no placeholder APIs" policy (`docs/architecture/placement.md`). That tension was surfaced during review and the mock switch was kept **by explicit decision**, on the basis that it's flag-gated, env-var-gated, production-hard-blocked, `TODO(LFXV2-1711)`-marked, and now visually self-disclosing via the "Sample data" marker — not a permanent stand-in for the real upstream contract.

### Caching

The engagement-stats read is cached per caller for `VALKEY_CACHE.GROUPS_ENGAGEMENT_TTL_SECONDS` (60s) via `withUserCache`, to absorb repeated dashboard refreshes without recomputing. Cache-hit behavior — a repeat call within the TTL doesn't re-invoke the underlying compute — is unit-tested (`groups-engagement-stats.service.spec.ts`, "does not recompute on a cache hit"); no wall-clock latency benchmark was run as part of this PR.

### Known, accepted minor trade-off

The freshness sub-line ("Updated Xm ago") is computed once when the engagement fetch resolves and does not tick live on a long-lived dashboard tab. This matches the existing pattern already used elsewhere in this app (e.g. `my-groups-card-grid.component.ts`), so it wasn't changed here.

## Test plan

- [x] BFF unit tests: mock determinism, cache-hit/miss + `accept`-validator rejection of a malformed cache entry, live-mode null-field response, `NODE_ENV=production` hard-block, `source` field on both branches
- [x] Shared-package unit tests for the pure `buildEngagementStatCards` card-builder: loading, populated, mock-sourced ("Sample data"), null-degraded, and independent per-card degradation (one field null doesn't blank the other) — substituting for a component test, since this repo has no Angular component-test runner (confirmed: no `*.component.spec.ts` files exist; `apps/lfx-one/vitest.config.ts` scopes to `src/server/**/*.spec.ts` only)
- [x] Playwright e2e coverage (`groups-engagement-stats.spec.ts`): flag off (default) → 3 cards, no `/engagement-stats` request; flag on → 5 cards populated from the mocked response; Foundation lens → never requests or renders the engagement cards regardless of flag state
- [ ] Manual: toggle `wg-engagement-metrics` on/off against a real browser session to visually confirm the 3-card ↔ 5-card transition and the "Sample data"/"Unavailable" states — not performed as part of this PR; the automated e2e spec above covers the same behavior programmatically, but no one has eyeballed it in a running browser yet

[LFXV2-1711]: https://linuxfoundation.atlassian.net/browse/LFXV2-1711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/e2e/groups-engagement-stats.spec.ts
+++ b/apps/lfx-one/e2e/groups-engagement-stats.spec.ts
@@ -1,0 +1,266 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+/**
+ * Groups dashboard — engagement stats flag/lens gating — LFXV2-1711.
+ *
+ * Coverage:
+ *   - `wg-engagement-metrics` flag off (the default until the flag is created/enabled in
+ *     LaunchDarkly): Me lens renders the 3 always-on group-count cards only, and the
+ *     `/api/committees/engagement-stats` endpoint is never requested.
+ *   - Foundation lens never requests or renders the engagement cards, regardless of flag state —
+ *     the endpoint is caller-scoped ("mine semantics"), so it has no scoped card to render there.
+ *   - `wg-engagement-metrics` flag on: Me lens renders 5 cards, with Active Members / Meetings This
+ *     Month populated from the mocked engagement-stats response.
+ *
+ * Prerequisites:
+ *   - Dev server reachable at the Playwright baseURL (default http://localhost:4200)
+ *   - apps/lfx-one/.env populated with TEST_USERNAME / TEST_PASSWORD
+ *   - The flag-on test additionally requires `wg-engagement-metrics` toggled ON for the test user
+ *     in LaunchDarkly — it skips gracefully (matching org-selector.spec.ts's precedent) if the
+ *     flag isn't configured, rather than failing CI on missing external LD setup.
+ */
+
+import type { Committee, MyCommittee, PersistedPersonaState, PersonaType } from '@lfx-one/shared/interfaces';
+import { LENS_COOKIE_KEY, PERSONA_COOKIE_KEY } from '@lfx-one/shared/constants';
+import { expect, Page, test } from '@playwright/test';
+
+test.setTimeout(60_000);
+
+const PAGE_LOAD_TIMEOUT = 20_000;
+const ELEMENT_TIMEOUT = 10_000;
+
+const MOCK_FOUNDATION_SLUG = 'test-foundation';
+const MOCK_FOUNDATION_UID = 'f0000000-0000-0000-0000-00000000d001';
+const MOCK_COMMITTEE_UID = 'c0000000-0000-0000-0000-00000000d001';
+const ENGAGEMENT_STATS_PATH = '/api/committees/engagement-stats';
+
+function buildMyCommittees(): MyCommittee[] {
+  const now = new Date().toISOString();
+  return [
+    {
+      uid: MOCK_COMMITTEE_UID,
+      name: 'Technical Steering Committee',
+      category: 'Technical Steering Committee',
+      enable_voting: true,
+      public: true,
+      sso_group_enabled: false,
+      created_at: now,
+      updated_at: now,
+      total_members: 12,
+      total_voting_repos: 5,
+      project_uid: 'p0000000-0000-0000-0000-00000000d001',
+      project_name: 'Test Project',
+      my_role: 'Chair',
+    } as MyCommittee,
+  ];
+}
+
+function buildFoundationCommittees(): Committee[] {
+  const now = new Date().toISOString();
+  return [
+    {
+      uid: MOCK_COMMITTEE_UID,
+      name: 'Governing Board',
+      category: 'Board',
+      enable_voting: true,
+      public: true,
+      sso_group_enabled: false,
+      created_at: now,
+      updated_at: now,
+      total_members: 9,
+      total_voting_repos: 9,
+      project_uid: MOCK_FOUNDATION_UID,
+      project_name: 'Test Foundation',
+      is_foundation: true,
+    } as Committee,
+  ];
+}
+
+function buildProjectStub() {
+  return {
+    uid: MOCK_FOUNDATION_UID,
+    slug: MOCK_FOUNDATION_SLUG,
+    name: 'Test Foundation',
+    description: 'Test foundation for groups engagement-stats specs',
+    public: true,
+    parent_uid: '',
+    stage: 'Active',
+    category: 'project',
+    funding_model: [],
+    charter_url: '',
+    legal_entity_type: '',
+    legal_entity_name: '',
+    legal_parent_uid: '',
+    autojoin_enabled: false,
+    formation_date: '',
+    logo_url: '',
+    repository_url: '',
+    website_url: '',
+    created_at: '',
+    updated_at: new Date().toISOString(),
+    mailing_list_count: 0,
+    writer: true,
+  };
+}
+
+async function stubPersona(page: Page, personas: string[]): Promise<void> {
+  await page.route('**/api/user/personas*', (route) =>
+    route.fulfill({
+      status: 200,
+      contentType: 'application/json',
+      body: JSON.stringify({ personas, personaProjects: {}, projects: [], organizations: [], isRootWriter: true }),
+    })
+  );
+}
+
+async function stubPendingInvitations(page: Page): Promise<void> {
+  await page.route('**/api/user/pending-invitations*', (route) => route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify([]) }));
+}
+
+async function stubMyCommittees(page: Page, committees: MyCommittee[]): Promise<void> {
+  await page.route('**/api/committees/my-committees*', (route) =>
+    route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(committees) })
+  );
+}
+
+async function stubFoundationCommittees(page: Page, committees: Committee[]): Promise<void> {
+  await page.route('**/api/committees/my-committee-uids*', (route) =>
+    route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify([]) })
+  );
+  await page.route('**/api/committees*', (route) => {
+    const pathname = new URL(route.request().url()).pathname;
+    if (pathname !== '/api/committees') return route.fallback();
+    return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(committees) });
+  });
+}
+
+async function stubProjectApi(page: Page): Promise<void> {
+  await page.route(`**/api/projects/${MOCK_FOUNDATION_SLUG}*`, (route) =>
+    route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(buildProjectStub()) })
+  );
+  await page.route('**/api/projects/*/sfid*', (route) => route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify({ sfid: null }) }));
+}
+
+async function stubEngagementStats(page: Page): Promise<void> {
+  await page.route(`**${ENGAGEMENT_STATS_PATH}*`, (route) =>
+    route.fulfill({
+      status: 200,
+      contentType: 'application/json',
+      body: JSON.stringify({ active_members: 7, meetings_this_month: 2, computed_at: new Date().toISOString(), source: 'live' }),
+    })
+  );
+}
+
+async function setCookies(page: Page, personas: string[], lens: 'me' | 'foundation'): Promise<void> {
+  const state: PersistedPersonaState = { primary: personas[0] as PersonaType, all: personas as PersonaType[] };
+  await page.context().addCookies([
+    { name: PERSONA_COOKIE_KEY, value: encodeURIComponent(JSON.stringify(state)), domain: 'localhost', path: '/', sameSite: 'Lax' },
+    { name: LENS_COOKIE_KEY, value: lens, domain: 'localhost', path: '/', sameSite: 'Lax' },
+  ]);
+}
+
+// Gated on env vars rather than on URL sniffing so genuine auth-flow regressions (expired
+// storageState, broken Auth0 login helper) still fail loudly when creds ARE configured.
+const AUTH_CREDS_PRESENT = !!process.env.TEST_USERNAME && !!process.env.TEST_PASSWORD;
+
+function skipWhenAuthMissing(): void {
+  if (!AUTH_CREDS_PRESENT) {
+    test.skip(true, 'TEST_USERNAME / TEST_PASSWORD not configured — see global-setup.ts');
+  }
+}
+
+async function gotoMyGroups(page: Page): Promise<void> {
+  skipWhenAuthMissing();
+  await page.goto('/', { waitUntil: 'domcontentloaded' });
+  await expect(page).not.toHaveURL(/auth0\.com/);
+  await page.goto('/groups', { waitUntil: 'domcontentloaded' });
+  await expect(page).not.toHaveURL(/auth0\.com/);
+  await expect(page.getByTestId('committees-me-stats')).toBeVisible({ timeout: PAGE_LOAD_TIMEOUT });
+}
+
+async function gotoFoundationGroups(page: Page): Promise<void> {
+  skipWhenAuthMissing();
+  await page.goto('/', { waitUntil: 'domcontentloaded' });
+  await expect(page).not.toHaveURL(/auth0\.com/);
+  await page.goto(`/foundation/groups?project=${MOCK_FOUNDATION_SLUG}`, { waitUntil: 'domcontentloaded' });
+  await expect(page).not.toHaveURL(/auth0\.com/);
+  await expect(page.getByTestId('committees-foundation-stats')).toBeVisible({ timeout: PAGE_LOAD_TIMEOUT });
+}
+
+test.describe('Groups dashboard — engagement stats (LFXV2-1711)', () => {
+  test.describe('Me lens', () => {
+    let engagementRequestCount = 0;
+
+    test.beforeEach(async ({ page }) => {
+      engagementRequestCount = 0;
+      page.on('request', (req) => {
+        if (req.url().includes(ENGAGEMENT_STATS_PATH)) engagementRequestCount++;
+      });
+
+      await setCookies(page, ['contributor'], 'me');
+      await stubPersona(page, ['contributor']);
+      await stubPendingInvitations(page);
+      await stubMyCommittees(page, buildMyCommittees());
+      await stubEngagementStats(page);
+      await gotoMyGroups(page);
+    });
+
+    test('flag off (default): renders the 3 always-on group-count cards and never requests engagement-stats', async ({ page }) => {
+      const grid = page.getByTestId('committees-me-stats');
+      await expect(grid.getByTestId('stat-card-Total Groups')).toBeVisible({ timeout: ELEMENT_TIMEOUT });
+      await expect(grid.getByTestId('stat-card-Public Groups')).toBeVisible({ timeout: ELEMENT_TIMEOUT });
+      await expect(grid.getByTestId('stat-card-Voting Enabled Groups')).toBeVisible({ timeout: ELEMENT_TIMEOUT });
+
+      const activeMembersCard = grid.getByTestId('stat-card-Active Members');
+      const flagAppearsOn = await activeMembersCard.isVisible().catch(() => false);
+      test.skip(flagAppearsOn, 'wg-engagement-metrics flag appears ON for this test user — covered by the flag-on test below instead');
+
+      await expect(activeMembersCard).toHaveCount(0);
+      await expect(grid.getByTestId('stat-card-Meetings This Month')).toHaveCount(0);
+
+      // Wait a moment for any async requests to fire, then assert the flag-gated fetch never ran.
+      await page.waitForTimeout(1_000);
+      expect(engagementRequestCount).toBe(0);
+    });
+
+    test('flag on: renders 5 cards, with Active Members / Meetings This Month populated from the response', async ({ page }) => {
+      const grid = page.getByTestId('committees-me-stats');
+      const activeMembersCard = grid.getByTestId('stat-card-Active Members');
+
+      const flagAppearsOn = await activeMembersCard.isVisible({ timeout: ELEMENT_TIMEOUT }).catch(() => false);
+      test.skip(!flagAppearsOn, 'wg-engagement-metrics flag appears OFF for this test user — see file header for the LD precondition');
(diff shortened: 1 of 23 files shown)


apps/lfx-one/src/server/services/groups-engagement-stats.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  private async computeEngagementStats(req: Request, username: string): Promise<GroupsEngagementStats> {
+    // Mock is opt-in and additionally hard-blocked in production, so a stray `ENGAGEMENT_BACKEND=mock`
+    // left in a prod-like environment's config can't silently serve fabricated numbers as real data.
+    const backend = process.env['ENGAGEMENT_BACKEND'] === 'mock' && process.env['NODE_ENV'] !== 'production' ? 'mock' : 'live';
+    const computedAt = new Date().toISOString();
+
+    if (backend === 'live') {
+      // TODO(LFXV2-1711): read from the dbt engagement model (same source as LFXV2-1705) once its
+      // read path exists. Until then, always return null fields rather than fabricating live-looking
+      // data — the client renders an "Unavailable" degraded state for these two cards. WARN once per
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch flagging this against `docs/architecture/placement.md` — I want to be transparent that this was a deliberate, discussed trade-off rather than an oversight.

The ticket (LFXV2-1711) explicitly specified a mock/live switch mirroring the approach in the sibling ticket LFXV2-1705, since the underlying dbt engagement model genuinely doesn't exist yet (that's LFXV2-1705's own deliverable). This exact tension with the no-mock-data policy was surfaced during review and the decision to keep the switch was made explicitly, on the basis of the mitigations already in place:

- Defaults to `live` (null fields → "Unavailable" cards) — `mock` requires an explicit opt-in
- `mock` is additionally hard-blocked when `NODE_ENV=production`, regardless of the env var
- `ENGAGEMENT_BACKEND=mock` ships commented out in `.env.example` by default
- The response carries `source: 'mock' | 'live'`, and the client renders a visible "Sample data" marker whenever `source === 'mock'` — so fabricated numbers can never be mistaken for real ones, including during local validation against synced prod data
- Gated behind the `wg-engagement-metrics` flag (defaults off) and `TODO(LFXV2-1711)`-marked as an interim shim, not a permanent stand-in for the real upstream contract

Full detail is in the PR description. Happy to discuss further if you think the mitigations aren't sufficient — but this isn't an unmonitored placeholder API, it's a fenced interim shim with an explicit removal path once LFXV2-1705 lands.

apps/lfx-one/src/app/modules/committees/committee-dashboard/committee-dashboard.component.ts
```diff
@@ -411,6 +432,30 @@ export class CommitteeDashboardComponent {
...
+   * the always-on group-count cards / groups list below are on entirely separate HTTP calls, so a
+   * stats failure here never blocks them.
+   */
+  private initializeEngagementStats(): Signal<GroupsEngagementStats | null> {
+    if (isPlatformBrowser(this.platformId)) {
+      toObservable(computed(() => this.isMeLens() && this.isEngagementMetricsEnabled()))
+        .pipe(
+          filter((ready) => ready),
+          take(1),
+          switchMap(() => this.committeeService.getGroupsEngagementStats().pipe(finalize(() => this.engagementStatsLoading.set(false)))),
```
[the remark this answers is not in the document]
manishdixitlfx: Done — added Playwright coverage for exactly this integration gate in commit c1b833169 (`apps/lfx-one/e2e/groups-engagement-stats.spec.ts`):

- **Flag off (default)**: Me lens renders the 3 always-on group-count cards only, and `/api/committees/engagement-stats` is never requested (asserted via a `page.on('request')` counter)
- **Flag on**: Me lens renders 5 cards, with Active Members / Meetings This Month populated from the mocked response
- **Foundation lens**: never requests or renders the engagement cards, regardless of flag state — asserted the same way

The flag-on test documents the `wg-engagement-metrics` LaunchDarkly precondition in the file header and skips gracefully if the flag isn't configured for the test user, matching this repo's existing convention for flag-gated e2e coverage (see `org-selector.spec.ts`'s `org-lens-enabled` precondition). The flag-off and Foundation-lens tests need no such precondition since they hold regardless of flag state.

manishdixitlfx:
## Review Feedback Addressed

Commit: c1b833169

### Changes Made
- **apps/lfx-one/e2e/groups-engagement-stats.spec.ts**: added Playwright coverage for the flag-gate/lens-gate integration — flag off → 3 cards, no request; flag on → 5 cards populated from the mocked response; Foundation lens → never requests or renders the engagement cards regardless of flag state (per copilot-pull-request-reviewer)

### No Change Needed
- **groups-engagement-stats.service.ts:65-68**: the mock/live backend switch was flagged against `docs/architecture/placement.md`'s no-placeholder-API policy. This is a deliberate, already-discussed trade-off (per the ticket's explicit design, mirroring LFXV2-1705) — kept with the mitigations already in place (defaults to `live`, `NODE_ENV=production` hard-block, opt-in-only env var commented out by default, a `source` field + visible "Sample data" marker so fabricated numbers are never mistaken for real ones, and flag-gated behind `wg-engagement-metrics`). Full reasoning posted on the thread and in the PR description (per copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/server/services/groups-engagement-stats.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Returns the caller's engagement rollup, cached ~60s per user (see `withUserCache`) to absorb
+   * repeated dashboard refreshes. Never throws — `live` mode returns null fields pre-dbt-deploy
+   * rather than failing the request, matching the graceful-degradation precedent used elsewhere for
+   * not-yet-available dbt models (LFXV2-2874).
+   */
+  public async getEngagementStats(req: Request): Promise<GroupsEngagementStats> {
+    const username = getEffectiveUsername(req) ?? '';
+
+    return withUserCache(
```
dealako, line 68: [minor] Cache can briefly serve a stale `mock` entry after a backend/env switch

Issue: `getEngagementStats` reads the per-user cache before the current backend mode (`mock` vs `live`) is resolved, and `isGroupsEngagementStats` accepts either `source` value as a valid cached entry.

Proof: if an environment previously ran with `ENGAGEMENT_BACKEND=mock`, a cached `{ source: 'mock', ... }` entry is stored under `VALKEY_CACHE.GROUPS_ENGAGEMENT_NAMESPACE` keyed by username. Flipping `ENGAGEMENT_BACKEND` to unset/`live` (or correcting `NODE_ENV` to `production`, which is supposed to hard-block mock) does not invalidate or bypass that entry — `withUserCache` will still return it for up to `GROUPS_ENGAGEMENT_TTL_SECONDS` (60s) since the validator doesn't check the entry's `source` against the resolved backend.

Why it matters: the production hard-block is meant to guarantee mock data can never appear in prod; this cache path is a narrow but real gap in that guarantee — a stray or recently-corrected `ENGAGEMENT_BACKEND=mock` can leak a mock-sourced ("Sample data"-labeled) response into a live/production request for up to 60s after the fix.

Fix: resolve the backend before the cache lookup and either fold it into the cache key (e.g. `` `${username}:${backend}` ``) or have the validator reject a cached `source` that doesn't match the currently-resolved backend, so a switch takes effect immediately rather than waiting out the TTL. (Agreeing with Copilot's suppressed low-confidence comment on this same commit — verified independently against current code.)

packages/shared/src/utils/committee.utils.ts
```diff
@@ -337,3 +345,48 @@ function compareCodeUnits(a: string, b: string): number {
...
+
+  if (loading) {
+    return [activeMembersCard, meetingsThisMonthCard];
+  }
+
+  // Mock-sourced values get "Sample data" instead of a freshness label — deliberately more visible
+  // than a subtle provenance flag, since the whole point is that a fabricated fixture number must
+  // never be mistaken for live data during manual validation (including against synced prod data,
+  // where NODE_ENV isn't 'production' but a stray ENGAGEMENT_BACKEND=mock could still be set).
+  const subLineForValue = stats && stats.source === 'mock' ? 'Sample data' : stats ? `Updated ${formatRelativeTime(new Date(stats.computed_at))}` : undefined;
```
dealako: [minor] Nested ternary expression

Issue: `stats && stats.source === 'mock' ? 'Sample data' : stats ? \`Updated ${...}\` : undefined` nests a ternary inside the false branch of another, which this repo's CLAUDE.md explicitly prohibits ("Never nest ternary expressions").

Proof: the expression has the shape `(A && B) ? X : (C ? Y : Z)` — the else-branch of the outer ternary is itself a ternary.

Why it matters: reduces readability and is a documented repo convention violation, though it has no behavioral consequence today.

Fix:
```typescript
let subLineForValue: string | undefined;
if (!stats) {
  subLineForValue = undefined;
} else if (stats.source === 'mock') {
  subLineForValue = 'Sample data';
} else {
  subLineForValue = `Updated ${formatRelativeTime(new Date(stats.computed_at))}`;
}
```

apps/lfx-one/e2e/groups-engagement-stats.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    });
+
+    test('flag off (default): renders the 3 always-on group-count cards and never requests engagement-stats', async ({ page }) => {
+      const grid = page.getByTestId('committees-me-stats');
+      await expect(grid.getByTestId('stat-card-Total Groups')).toBeVisible({ timeout: ELEMENT_TIMEOUT });
+      await expect(grid.getByTestId('stat-card-Public Groups')).toBeVisible({ timeout: ELEMENT_TIMEOUT });
+      await expect(grid.getByTestId('stat-card-Voting Enabled Groups')).toBeVisible({ timeout: ELEMENT_TIMEOUT });
+
+      const activeMembersCard = grid.getByTestId('stat-card-Active Members');
+      const flagAppearsOn = await activeMembersCard.isVisible().catch(() => false);
```
dealako: [nit] Inconsistent `isVisible()` wait vs. the flag-on check

Issue: this flag-off detection uses a zero-wait `activeMembersCard.isVisible()`, while the flag-on test at line 231 uses `isVisible({ timeout: ELEMENT_TIMEOUT })`.

Proof: `Locator.isVisible()` with no options resolves immediately without waiting for the LaunchDarkly flag to settle; if the flag resolves asynchronously after the grid's first render, this can read `false` for an actually-on flag and skip the `test.skip` path, only to fail later once the flag settles during the subsequent `waitForTimeout`.

Why it matters: rare but real source of flakiness in an already flag-dependent e2e test.

Fix: match the flag-on check —
```diff
-      const flagAppearsOn = await activeMembersCard.isVisible().catch(() => false);
+      const flagAppearsOn = await activeMembersCard.isVisible({ timeout: ELEMENT_TIMEOUT }).catch(() => false);
```
(Agreeing with both Copilot's and CodeRabbit's comments on this same commit — still open.)

dealako (dismissed):
Hi @manishdixitlfx — nice PR. Clean design for the mock/live split (default-to-live, prod hard-block, visible "Sample data" marker, flag-gated UI), solid test coverage across BFF unit tests, shared-utils tests, and the new Playwright integration gate, and the trade-off discussion on the placeholder-API thread was transparent and well-reasoned.

Ran a 3-way parallel review (security/privacy, correctness/performance/tests, style/docs) plus reconciliation against the existing Copilot and CodeRabbit comments on this PR.

**Issue counts**
- 🔴 Blocking: 0
- 🟡 Minor: 2 — nested ternary in `committee.utils.ts`; per-user cache can briefly serve a stale `source: 'mock'` entry after an `ENGAGEMENT_BACKEND`/`NODE_ENV` switch to live/production (agreeing with Copilot's suppressed low-confidence comment on this commit, verified independently)
- ⚪ Nit: 1 — inconsistent `isVisible()` wait in the e2e flag-off check (agreeing with both Copilot and CodeRabbit, still open on this commit)
- ❔ Question: 0

**Bot reconciliation**
- Agree with Copilot's cache-staleness comment on `groups-engagement-stats.service.ts` — verified the code path directly; `withUserCache`'s validator accepts both `mock` and `live` entries and the cache is read before the backend mode is resolved, so a stale mock entry can outlive a config switch for up to the 60s TTL. Restating it below as a formal minor finding since it's still unaddressed on this commit.
- Agree with Copilot/CodeRabbit's `isVisible()` timeout inconsistency in the e2e spec — still present at line 216, restating as a nit.
- The earlier Copilot placeholder-API thread (`docs/architecture/placement.md`) and missing-Playwright-coverage thread both look resolved — the author's response documents the deliberate mock/live trade-off with concrete mitigations, and Playwright coverage was added in commit `c1b833169`. No further action needed there from this pass.
- Security/Data Privacy reviewer found nothing: no PII in logs (only aggregate counts logged), no PII in the response contract, cache keys validated via `isFilterSafeUsername()`, test fixtures are synthetic.
- Correctness/Performance/Test Coverage reviewer found nothing beyond the cache-staleness item already covered above.

apps/lfx-one/src/server/services/groups-engagement-stats.service.ts
manishdixitlfx: Done — resolved the backend (mock/live) once before the cache lookup and made the accept validator reject a cached entry whose `source` doesn't match. See commit a11e40734: `getEngagementStats` now calls `resolveBackend()` up front and passes `(value) => isGroupsEngagementStats(value, backend)` as the accept validator, so a cached entry from a since-changed `ENGAGEMENT_BACKEND`/`NODE_ENV` config is treated as a miss and recomputed immediately rather than surviving the 60s TTL.

Added a direct regression test for the exact scenario you described: a `mock`-sourced entry gets cached, `NODE_ENV` flips to `production`, and the next call correctly recomputes as `live` (null fields) instead of returning the stale mock entry. Thanks for tracing this all the way through — real gap, good catch.

packages/shared/src/utils/committee.utils.ts
manishdixitlfx: Fixed — extracted the sub-line logic into `resolveEngagementSubLine()` (a plain if/else chain, no ternary at all) in commit a11e40734, replacing the nested ternary. Good catch on the CLAUDE.md convention.

apps/lfx-one/e2e/groups-engagement-stats.spec.ts
manishdixitlfx: Fixed, but with a more thorough change than the suggested diff, in commit a11e40734 — worth flagging why.

I checked `playwright-core`'s type definitions before applying the one-line fix, and `Locator.isVisible()`'s `timeout` option is actually `@deprecated`/ignored entirely ("does not wait for the element to become visible and returns immediately"). That means the flag-on check at (then) line 231, which already used `isVisible({ timeout: ELEMENT_TIMEOUT })`, had the exact same non-waiting bug as the flag-off check — just hidden, since Copilot's second pass on this commit also independently flagged that same nuance. Matching the flag-off check to the flag-on check's pattern would have made both checks consistently wrong rather than fixing the underlying flakiness.

Replaced both with a `cardAppears()` helper that uses `Locator.waitFor({ state: 'visible', timeout })`, which does genuinely wait. Applied to both the flag-off and flag-on detection sites.

manishdixitlfx:
## Review Feedback Addressed

Commit: a11e40734

### Changes Made
- **groups-engagement-stats.service.ts**: resolve the backend (mock/live) once before the cache lookup, and reject a cached entry whose `source` doesn't match the currently-resolved backend — fixes a real gap where a stale `mock`-sourced entry could survive an `ENGAGEMENT_BACKEND`/`NODE_ENV` switch to live/production for up to the 60s TTL (per @dealako)
- **groups-engagement-stats.service.spec.ts**: added a regression test for that exact scenario
- **committee.utils.ts**: extracted `resolveEngagementSubLine()` to remove a nested ternary in `buildEngagementStatCards`, per CLAUDE.md's ternary-nesting rule (per @dealako)
- **groups-engagement-stats.spec.ts**: replaced `isVisible()`/`isVisible({ timeout })` flag-detection with a `waitFor()`-based `cardAppears()` helper. The suggested one-line fix (match the flag-on check's timeout option) would have been incomplete — verified against `playwright-core`'s type definitions that `isVisible()`'s `timeout` option is deprecated and ignored entirely, so the flag-on check had the same non-waiting bug, just hidden (per @dealako, also raised by Copilot and CodeRabbit on this commit)

### Threads Resolved

manishdixitlfx:
Fixed CodeRabbit's async/await nitpick on the \`cardAppears()\` helper (commit b379bde47) — replaced the \`.then().catch()\` chain with try/await/catch, matching this org's coding guideline.

apps/lfx-one/e2e/groups-engagement-stats.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+ * resolved state. `Locator.isVisible()`'s `timeout` option is deprecated and ignored (Playwright
+ * always returns immediately from `isVisible()`, regardless of any timeout passed), so checking
+ * "did the flag turn out on?" requires actually waiting via `waitFor()` — an immediate check can
+ * read `false` before the async flag client resolves, even when the card appears moments later.
+ */
+async function cardAppears(locator: Locator, timeout: number): Promise<boolean> {
+  try {
+    await locator.waitFor({ state: 'visible', timeout });
+    return true;
+  } catch {
```
dealako, line 184: [minor] Catch only Playwright's `TimeoutError` here

Issue: the bare `catch { return false; }` swallows every exception from `locator.waitFor()`, not just timeouts.

Proof: `waitFor()` can also throw for page/context closure or a locator failure (e.g. strict-mode violation); those propagate through this catch and return `false` just like a genuine timeout.

Why it matters: a real failure (closed page, broken locator) would be indistinguishable from "flag is off," so the flag-on test would silently `test.skip()` instead of failing loudly on an actual bug.

Fix:
```typescript
import { errors } from 'playwright';

async function cardAppears(locator: Locator, timeout: number): Promise<boolean> {
  try {
    await locator.waitFor({ state: 'visible', timeout });
    return true;
  } catch (error) {
    if (error instanceof errors.TimeoutError) {
      return false;
    }
    throw error;
  }
}
```
(Agreeing with CodeRabbit's comment on this same commit — still open.)

dealako (approved):
Thanks for the quick turnaround @manishdixitlfx — this closes out all the open items from the first pass cleanly.

👏 **Nice work**
- The cache-staleness fix is properly done: `resolveBackend()` now runs before the cache lookup, the resolved backend flows into both the compute call and the validator (`isGroupsEngagementStats(value, backend)`), and there's a real regression test proving a stale `mock` entry gets recomputed as `live` the moment `NODE_ENV` flips to production — not just a cosmetic tweak.
- `resolveEngagementSubLine()` is a clean extraction — plain if/else, no ternary at all, easy to read.
- The `cardAppears()` helper is the right fix for the `isVisible()` timeout problem, and the commit message explaining *why* (`isVisible()`'s `timeout` option is deprecated/ignored by Playwright) shows this wasn't just patching over the symptom Copilot/CodeRabbit/I flagged — it fixed the actual underlying non-wait bug at both call sites.
- Also picked up CodeRabbit's async/await nitpick on `cardAppears()` in the latest commit.

**Revision tracking**
- ✅ Resolved: per-user cache could serve a stale `mock` entry after a backend/env switch (commit `a11e40734`, regression test added)
- ✅ Resolved: nested ternary in `committee.utils.ts` (commit `a11e40734`, extracted `resolveEngagementSubLine()`)
- ✅ Resolved: inconsistent `isVisible()` wait in the e2e flag checks (commit `a11e40734`, replaced with `cardAppears()`)

**Issue counts (this round)**
- 🔴 Blocking: 0
- 🟡 Minor: 1 — `cardAppears()`'s bare `catch { return false }` also swallows non-timeout errors (page/context closure, locator failures), which would make the flag-on test silently skip instead of failing loudly on a real bug. Restating CodeRabbit's comment on this — still open as of this commit.
- ⚪ Nit: 0

**Bot reconciliation**
- Agreeing with CodeRabbit's newest comment on `cardAppears()` — verified directly against the current code (still a bare `catch` at line 184). Restating it below as a formal minor finding since it's unaddressed on this commit.
- Security/Data Privacy reviewer found nothing new in this round.

PR #1268, opened 2026-07-30
feat(committees): add member engagement/attendance rollup endpoint

manishdixitlfx:
## Summary

Adds `GET /api/committees/:uid/engagement?window=30d|90d|ytd`, a BFF-only endpoint for LFXV2-1705 that returns per-member meeting-attendance rollups and a committee-level summary (attendance rate, active/at-risk counts). Frontend is explicitly out of scope for this ticket (owned separately) — this PR is the endpoint + shared classification logic only.

Gated by `ENGAGEMENT_BACKEND` (documented in `.env.example`): unset or anything other than `'live'` (the default) serves deterministic mock data anchored to the real committee roster; `'live'` runs the real Snowflake read, which degrades to an honest `data_available:false` until the dbt model (`platinum_lfx_one_committee_meeting_attendance`, `lf-dbt#2694`) is deployed and the live read surface is wired up. **Mock-as-default is intentional, per the original ticket's mock-first spec** (confirmed directly with the ticket owner during review) — deployments must not rely on the default alone once a UI consumes this endpoint; see the first bullet below.

- New `CommitteeEngagementService`, `CommitteeEngagementController`, and `assertCommitteeRead` access-check helper (mirrors `assertOrgLensRead`).
- Gates on `committee#auditor`, not `committee#viewer` — `viewer` carries an `[user:*]` wildcard in the platform FGA model that would leak per-member attendance to any authenticated caller on a public committee. See the doc comment on `committee-read-access.helper.ts` for the full rationale (verified against the live FGA model in `lfx-v2-helm`).
- Joins by `MEMBER_USER_ID` = `CommitteeMember.uid` — an exact key, matching the finalized dbt model's real grain (`committee_id, member_user_id`).
- Classification (`High`/`Medium`/`Low`/`Inactive`/`Emeritus`) is computed in the BFF from `@lfx-one/shared/constants` thresholds plus two model-driven rules pulled from the LFXV2-1705 Jira thread: `Emeritus` voting status never classifies `Inactive`/`At-Risk` (on the roster for legacy/honorific reasons), and a member whose `member_joined_at` falls within the requested window can't classify `Inactive` off zero invites (there's been no real opportunity yet, not disengagement — the ticket's real "Orlin Vasilev" example). The "Active Members" summary stat is its own rule (any real attendance this window, or having joined within it), broader than the per-member classification tiers.
- **Mock mode only fabricates attendance numbers.** `MEMBER_ROLE` and `MEMBER_JOINED_AT` are always the real roster's `role`/`created_at`; `MEMBER_VOTING_STATUS` prefers the real `voting.status` too — including a real `'None'`, which is itself a recorded status, not an absence of one (a full-branch review round caught an earlier version of this treating `None` as fabricatable, which would have relabeled real non-voting members as `Voting Rep`/`Emeritus`). The only synthetic identity value is a same-shape placeholder for members with no usable real voting status (absent, or blank/falsy), plus a roster-wide guaranteed-`Emeritus` fallback that applies only when nothing on the real roster is naturally `Emeritus` — so the mock view never contradicts what the real Members table shows for the same person. One consequence: on a committee where every member already has a real status (including an all-`None` roster — the common case for a committee without voting enabled), the Emeritus demo scenario simply doesn't render for that committee, by design.
- `data_source: 'mock' | 'live'` on every response, plus a `logger.warning` on every mock-mode request — `data_available: true` alone doesn't distinguish fabricated from real data, so this gives any current or future consumer a positive way to tell.
- `computed_at` is always `null` — the real model doesn't expose a freshness column yet (a separate pipeline-freshness follow-up). `formatCommitteeEngagementFreshness` gives a `"Updated daily"` fallback for the UI to use.
- Snowflake reads (live mode only) are cached per `(committeeUid, window)` for an hour, mirroring `org-lens-project-detail.service.ts`'s pattern — including skipping the cache write entirely on the missing-object degrade, so "no data yet" doesn't get pinned into Valkey past the real deploy. Mock mode is never cached (cheap, deterministic, synchronous).
- New `AccessCheckService.checkAccessStrict`/`checkSingleAccessStrict` (propagates upstream failures instead of swallowing them to `false`), enabling a 403-vs-503 split for this endpoint's access gate.
- `SnowflakeService.isMissingObjectError` extracted to a standalone `snowflake-error.helper.ts` so it's independently unit-tested rather than only exercised through a mocked class.

## Known limitations / open items (flagging for reviewer awareness, not blocking)

- **Mock-as-default has no production off-switch wired up yet.** Nothing in this PR adds `ENGAGEMENT_BACKEND` to the deployment chart's `environment:` block, so every environment (including prod) will serve mock data until someone sets it explicitly. Low risk *today* — no UI anywhere calls this endpoint yet, and the response is explicitly marked `data_source: 'mock'` — but this needs to be wired into the chart (and/or the eventual UI needs to gate on `data_source`, not just `data_available`) before any real user could see it.
- **One classification call needs product sign-off.** A member with zero invites who joined within the window (no real opportunity yet) classifies `High` ("active by definition," per the ticket) rather than a dedicated tier — the ticket doesn't name a specific tier for this case, and the classification set is otherwise fixed. Flagging in case product wants something else here.
- **Live read surface still undecided.** The finalized dbt model's real read surface (query-service vs. direct Snowflake) is an open question per the ticket's Jira thread — the live SQL path still targets the original placeholder columns (`MEMBER_EMAIL`/`ATTENDED_COUNT`/`INVITED_COUNT`/`COMPUTED_AT`), typed separately from `CommitteeEngagementWarehouseRow` so a successful-but-unmappable query degrades explicitly rather than silently corrupting the response. Needs a full rewrite once the read surface is decided (tracked via `TODO(LFXV2-1705 follow-up)` comments at both call sites). **Known gap in the interim:** the degrade only catches the "does not exist or not authorized" error shape (`isMissingObjectError`) — if a table ever appears under this placeholder name with different (even correct-model) columns before the SQL is rewritten, Snowflake raises a column-reference error instead, which isn't recognized as a missing-object case and rethrows as an unhandled 500 rather than degrading. Accepted for now since the whole query is already known-disposable pending the rewrite; not fixing it separately to avoid touching this branch's live-path behavior further.
- **Access-gate scope.** Only committee `auditor`/`writer` can read engagement data today — a rank-and-file `member` cannot see their own committee's numbers. Deliberate (attendance rollups read as leadership/oversight data), but this is a product-policy question that hasn't had explicit sign-off.
- **`snowflake.service.ts` touched (protected file).** `isMissingObjectError` extracted to a delegate (behavior-preserving), and the `expectMissingObject` degrade path logs at WARN with an `OK` span status instead of ERROR/errored-span, since every request against a not-yet-deployed table was otherwise polluting error tracking. Also affects the two pre-existing `expectMissingObject` callers in `project.service.ts` and `org-lens-project-detail.service.ts` (in the correct direction). Requesting explicit code-owner review on this file.
- **Cache shape guard is container-only.** The live-mode cache read validates `Array.isArray` but not row shape, unlike sibling caches that pass an element-aware predicate. A stale/malformed cached entry would degrade to zeroed-but-plausible-looking values rather than a clean miss.
- **The live-mode cache can currently only ever store an empty array.** `rows` is never populated on the live path today (the SQL can't produce `CommitteeEngagementWarehouseRow`-shaped data pre-rewrite), so every cache write is `[]` and every cache hit degrades to the same. The caching/TTL/namespace-version machinery is real but untestable-as-load-bearing until the SQL rewrite lands — at that point the cache silently starts storing real per-committee attendance for the first time, and the existing invariants (never cache a degrade, `v2` namespace bump) become load-bearing without anyone having re-derived them against real data. Not addressed now to avoid further touching this branch's live-path behavior; worth a dedicated test pinning the never-cache-a-degrade invariant when the rewrite lands.
- **Mock-mode and live-degrade both log a WARN on every request.** Deliberate (an in-band ops signal distinct from `data_source`), but permanent per-request WARN volume in the default (mock) configuration, and in live mode it's two WARNs plus a full Snowflake round-trip per request for a query that's guaranteed to fail pre-deploy (the missing-object degrade is intentionally never cached, so "no data yet" doesn't outlive the real deploy — see above). Worth reconsidering (e.g. log mock-mode's notice once per process rather than per request) once this endpoint has real traffic.
- **No pagination** on the roster/member list — committee rosters are expected to stay small, and every roster member currently comes back in one response.
- **Branch diff is large** (57 commits). The commit history is intentionally granular, mostly `fix(review): ...`/iteration commits reflecting an unusually thorough pre-PR review cycle (three independent automated reviewers run after every commit and again as full-branch sweeps) plus a mid-flight model-semantics finalization — further splitting at this point isn't practical without losing that history.
- **The Orlin ("joined recently, forced 5/5") mock scenario is rarely reachable on a real committee.** `CommitteeMember.created_at` is a required field on real roster data, so the "no real join date at all" fallback this scenario also relies on almost never fires; it only renders when some real member's tenure happens to land in a roughly 5-15 day window. Accepted trade-off (see the doc comment on `planRosterIdentities` in `committee-engagement-mock.helper.ts`) rather than overriding a real member's real join date, which this generator otherwise never does.

## Test plan

- [ ] `yarn test` — 735 tests pass (`@lfx-one/shared` + `lfx-one-ui`), including the service (mock + live mode, plus degraded-member roster-fallback cases for role/voting_status and tenure grace), controller, the mock/backend-switch helpers (including real-vs-fabricated identity coverage), the classifier's full decision table (including the ticket's literal Orlin and Emeritus cases), and the freshness fallback util.
- [ ] `yarn lint` / `yarn build` / `./check-headers.sh` / `yarn format:check` all clean.
- [ ] Manual verification against a running dev server:
  - `ENGAGEMENT_BACKEND` unset (default), any committee the caller can read: `200`, `data_available:true`, `data_source:'mock'`, varied per-member classifications including at least one `High`, `Medium`, `Low`, `Inactive`, and `Emeritus`, stable across reloads, numbers differing by `?window=30d|90d|ytd`, and `role`/`voting_status` matching what `GET /api/committees/:id/members` shows for the same people.
  - `?window=badvalue` → `400`.
  - `ENGAGEMENT_BACKEND=live`: `200`, `data_available:false`, `data_source:'live'`, every member `Inactive`/zeroed (expected pre-dbt-deploy state).
  - A committee the caller lacks `auditor` on → `403`.
  - Cross-check `members[].uid` against `GET /api/committees/:id/members` for the same committee — no dropped/duplicated members.


apps/lfx-one/src/server/helpers/snowflake-schema.helper.ts
```diff
@@ -0,0 +1,27 @@
...
+import { DEFAULT_LFX_ONE_PLATINUM_SCHEMA } from '@lfx-one/shared/constants';
+
+/**
+ * Guards `LFX_ONE_PLATINUM_SCHEMA` before it's interpolated into a SQL identifier: only
+ * `WORD(.WORD){1,2}`-shaped values pass, everything else (including anything with a stray quote,
+ * space, or SQL keyword) is rejected in favor of the default schema.
+ */
+function snowflakeQualifier(value: string | undefined): string | null {
+  const trimmed = value?.trim();
+  return trimmed && /^[A-Z0-9_]+(\.[A-Z0-9_]+){1,2}$/i.test(trimmed) ? trimmed.toUpperCase() : null;
```
dealako: [minor] Regex permits an override that produces an invalid 4-part Snowflake identifier

Issue: `snowflakeQualifier` accepts `WORD(.WORD){1,2}`, i.e. 2 *or 3* dot-separated segments. `engagementTable()` (committee-engagement.service.ts:320) then appends the table name unconditionally: `` `${resolveLfxOnePlatinumSchema()}.COMMITTEE_MEMBER_MEETING_ATTENDANCE` ``.
Proof: if an operator sets `LFX_ONE_PLATINUM_SCHEMA=DB.SCHEMA.SUBSCHEMA` (a legal 3-segment value per this regex), the resulting identifier is `DB.SCHEMA.SUBSCHEMA.COMMITTEE_MEMBER_MEETING_ATTENDANCE` — 4 parts. Snowflake object identifiers are at most `database.schema.object` (3 parts), so that query would fail. This was flagged by Copilot on this PR and is still present in the latest commit.
Why it matters: the override path is documented as supported (env-overridable per the doc comment above), so a 3-segment value is a plausible operator input, not an edge case — it would silently break the live read exactly when someone reaches for the override this helper advertises.
Fix: tighten the regex to `{1}` (2 segments only, matching the `ANALYTICS.PLATINUM_LFX_ONE` shape the doc comment describes), or explicitly reject 3-segment overrides for schema-qualified (not object-qualified) call sites.

apps/lfx-one/src/server/services/committee-engagement.service.ts
```diff
@@ -0,0 +1,323 @@
...
+   * its own caching and missing-object degrade independently of this method.
+   */
+  public async getCommitteeEngagement(req: Request, committeeUid: string, window: CommitteeEngagementWindow): Promise<CommitteeEngagementResponse> {
+    if (isEngagementMockBackend()) {
+      const members = await this.committeeService.getCommitteeMembers(req, committeeUid);
+      const rows = generateMockEngagementRows(committeeUid, members);
+      // The ops-side signal that a response is fabricated — `data_source: 'mock'` on the response
+      // body (below) is the in-band one any consumer must check; this is the log-side counterpart
+      // for anyone watching production logs rather than inspecting individual responses.
+      logger.warning(req, 'get_committee_engagement', 'ENGAGEMENT_BACKEND=mock — returning deterministic mock rows, not real data', {
```
dealako: [minor] Mock-mode log uses WARN, INFO fits better per the repo's logging convention

Issue: this logs at `logger.warning` when entering the deterministic mock path.
Proof: `.claude/rules/logging-patterns.md` reserves WARN for "recoverable issues," "error conditions with graceful degradation," and "fallback behaviors" — mock mode here is none of those; it's a deliberate, explicit opt-in (`ENGAGEMENT_BACKEND=mock`) that's additionally hard-blocked in production, so it isn't a degradation at all. The same rule file lists `logger.info` for "significant business operations visible in production."
Why it matters: minor — but WARN-per-request on every mock-mode call adds signal noise that competes with genuine degradation warnings (e.g. the missing-object degrade a few lines below this one), making the two harder to visually distinguish in logs.
Fix: `logger.info(req, 'get_committee_engagement', 'ENGAGEMENT_BACKEND=mock — returning deterministic mock rows, not real data', {...})`.

apps/lfx-one/src/server/services/committee-engagement.service.ts
```diff
@@ -0,0 +1,323 @@
...
+    let totalInvited = 0;
+    let activeCount = 0;
+    let atRiskCount = 0;
+    let matchedCount = 0;
+
+    const memberEngagements = members.map((member) => {
+      const row = rowsByUid.get(member.uid);
+      if (row) matchedCount++;
+
+      const counts = row ? this.countsForWindow(row, window) : { invited: 0, attended: 0, committeeMeetings: 0 };
```
dealako, line 216: [nit] No boundary validation that warehouse counts are non-negative

Issue: `countsForWindow` passes `INVITED_COUNT_*`/`ATTENDED_COUNT_*`/`COMMITTEE_MEETINGS_*` straight through from the Snowflake row with no non-negative check.
Proof: `attended` is clamped to `[0, invited]` two lines down (line 216), but `invited` and `committee_meetings` themselves are never validated, so a data-quality issue upstream (e.g. a negative `INVITED_COUNT`) would flow into `computeCommitteeEngagementRate` and the response unclamped.
Why it matters: low — this is a defense-in-depth gap, not a reachable bug today (Snowflake is a trusted internal source, and the live path currently always degrades to an empty row set per the TODO above). Flagging since the PR already validates one field (`attended`) at this exact spot.
Fix: optional — `Math.max(row.INVITED_COUNT_30D, 0)` etc. in `countsForWindow`, matching the existing `attended` clamp's defensive posture.

packages/shared/src/utils/committee-engagement-classifier.utils.ts
```diff
@@ -0,0 +1,84 @@
...
+ *    unlike case 2, they had an actual opportunity and skipped every one of them, so tenure does
+ *    not protect this case.
+ */
+export function classifyCommitteeEngagement(input: CommitteeEngagementClassificationInput): CommitteeEngagementClassification {
+  const { attended, invited, votingStatus, joinedWithinWindow } = input;
+  if (votingStatus === CommitteeMemberVotingStatus.EMERITUS) return 'Emeritus';
+  if (invited <= 0) return joinedWithinWindow ? 'High' : 'Inactive';
+
+  const rate = rawCommitteeEngagementRate(attended, invited);
+  if (rate <= 0) return 'Inactive';
```
dealako, line 59: [nit] Defensive `<= 0` check reads as though a negative rate were reachable

Issue: `if (rate <= 0) return 'Inactive';` — by this point `invited > 0` (checked at line 53) and `rawCommitteeEngagementRate` clamps `attended` to `[0, invited]`, so `rate` is always in `[0, 1]`; it can be `0` but never negative.
Why it matters: cosmetic only — a future reader may wonder whether a negative rate is a real case this guards against, when it isn't.
Fix: `if (rate === 0) return 'Inactive';`, or keep `<=` with a one-line comment noting it's defensive against clamping ever being weakened.

dealako (dismissed):
Hey @manishdixitlfx — nice work on this one. The BFF endpoint is well-scoped and the mock/live split is clean: it defaults to `live` (per the latest commit), hard-blocks mock in prod via `NODE_ENV`, and degrades honestly to `data_available:false` pre-dbt-deploy rather than fabricating attendance data. The `auditor`-vs-`viewer` FGA gating call is well-reasoned and documented in `committee-read-access.helper.ts`, the roster/warehouse join handles unmatched members and independent join-date fallbacks correctly, and test coverage across classification boundaries, mock determinism, and degrade paths is solid.

Ran a security-focused pass, a correctness/perf/test pass, and a style/convention pass in parallel, then reconciled against the existing Copilot comments on this PR (all four are already resolved by the latest commits — mock is no longer default, the strict-vs-fail-closed access-check gap is explicitly documented as an accepted trade-off in `committee-read-access.helper.ts`, and the legacy-placeholder SQL degrade is intentional per its own doc comment).

### Summary
- 🔴 Blocking: 0
- 🟡 Minor: 2 — schema-qualifier regex allows a 3-segment override that produces an invalid 4-part Snowflake identifier once `engagementTable()` appends the table name; mock-mode log uses WARN where INFO fits the repo's logging convention better
- ⚪ Nit: 2 — warehouse row counts aren't validated non-negative at the boundary (downstream clamping covers it); the classifier's `rate <= 0` check is defensive against a state the clamping logic already makes unreachable
- ❔ Question: 0

### ✅ Approved with minor comments

apps/lfx-one/src/server/helpers/snowflake-schema.helper.ts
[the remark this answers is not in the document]
manishdixitlfx: Done — tightened `snowflakeQualifier`'s regex from `{1,2}` to exactly two segments (`WORD.WORD`), and updated the three-segment test to expect fallback instead of acceptance. Every call site appends a table name unconditionally, so a 3-segment override always produced an invalid 4-part identifier.

See commit 522ca31f6: `apps/lfx-one/src/server/helpers/snowflake-schema.helper.ts`.

manishdixitlfx: Done — tightened the regex to exactly two segments and updated the three-segment test to expect fallback. Same fix as Copilot's thread on this same line — thanks for confirming it was still present in the latest commit.

See commit 522ca31f6: `apps/lfx-one/src/server/helpers/snowflake-schema.helper.ts`.

apps/lfx-one/src/server/helpers/committee-engagement-backend.helper.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * `ENGAGEMENT_BACKEND` gates `GET /api/committees/:uid/engagement` (LFXV2-1705) between the
+ * deterministic mock generator and the real (not-yet-deployed) Snowflake read: unset or anything
+ * other than `'live'` selects mock; `'live'` selects the real path, which degrades to
+ * `data_available: false` until the dbt model ships. Defaulting to mock (not live) means local/dev
+ * work and integration validation see varied fixtures without any env setup, while an explicit
+ * opt-in is required to exercise the real (currently always-degrading) path.
+ */
+export function isEngagementMockBackend(): boolean {
+  return process.env['ENGAGEMENT_BACKEND'] !== 'live';
```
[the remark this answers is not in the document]
manishdixitlfx: Already addressed in an earlier commit on this branch — `isEngagementMockBackend()` now defaults to `live`; mock requires an explicit `ENGAGEMENT_BACKEND=mock` and is additionally hard-blocked whenever `NODE_ENV=production`. Matches the equivalent gate on the sibling LFXV2-1711 branch. `.env.example` and the shared interface docs were updated to match.

apps/lfx-one/src/server/services/committee-engagement.service.ts
manishdixitlfx: Fixed — switched to `logger.info`. Good catch, and the reasoning is even stronger now that mock defaults off and requires explicit opt-in (see the other thread on this branch) — it's clearly not a degradation/fallback case per `logging-patterns.md`.

See commit 522ca31f6.

apps/lfx-one/src/server/services/committee-engagement.service.ts
manishdixitlfx: Fixed — added the `Math.max(x, 0)` clamp to `invited`/`attended`/`committeeMeetings` in `countsForWindow`, matching the existing `attended`-vs-`invited` clamp's posture. Added a regression test with negative warehouse values.

See commit 522ca31f6.

packages/shared/src/utils/committee-engagement-classifier.utils.ts
manishdixitlfx: Fixed — kept `<=` (matches your second option) and added a comment clarifying it's defensive against the upstream clamp ever being weakened, not because a negative rate is reachable today.

See commit 522ca31f6.

apps/lfx-one/src/server/services/access-check.service.ts
```diff
@@ -195,4 +165,82 @@ export class AccessCheckService {
...
+
+    // Map results back to resource IDs by re-deriving the same tuple each request sent
+    const resultMap = new Map<string, boolean>();
+
+    for (const resource of resources) {
+      const tuple = `${resource.resource}:${resource.id}#${resource.access}`;
+      const result = resultByTuple.get(tuple);
+
+      // Fail closed when the upstream response omits this tuple
+      resultMap.set(`${resource.id}#${resource.access}`, result?.hasAccess ?? false);
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch on the actual behavior — you're right that a truncated/malformed `/access-check` response still resolves to `false` inside `performCheck` rather than throwing, so `checkAccessStrict` doesn't distinguish that case from a genuine denial either.

This is a known, deliberate gap rather than an oversight — it's called out explicitly in `committee-read-access.helper.ts`'s own doc comment: *"This is incomplete, not absolute: `performCheck` itself resolves `false` (not a throw) when the access-check response omits the requested tuple entirely... so that case still surfaces here as a confident-looking 403 rather than a 503, same as it would through `checkSingleAccess`."*

Closing this gap fully means restructuring `performCheck` so a missing/malformed tuple throws for strict callers while `checkAccess` keeps its existing fail-closed-to-false fallback — a real behavior change to a security-relevant path, not a small patch. Leaving as documented for this PR; happy to open a follow-up if you'd rather track it separately.

apps/lfx-one/src/server/services/committee-engagement.service.ts
```diff
@@ -0,0 +1,259 @@
...
+    if (key !== null) {
+      const cached = await valkeyService.getJson<CommitteeEngagementWarehouseRow[]>(key, Array.isArray);
+      if (cached !== null) return { rows: cached, dataAvailable: true };
+    }
+
+    const sql = `
+      SELECT MEMBER_EMAIL, ATTENDED_COUNT, INVITED_COUNT, COMPUTED_AT
+      FROM ${this.engagementTable()}
+      WHERE COMMITTEE_UID = ? AND TIME_RANGE_TYPE = ?
+      ORDER BY MEMBER_EMAIL
```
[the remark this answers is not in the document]
manishdixitlfx: Correct, and this is the single most heavily-documented known limitation on this branch — tracked via the `TODO(LFXV2-1705 follow-up)` comments on `queryEngagementRows` and `engagementTable()`, and called out in the PR description's Known Limitations section.

It's typed against `LegacyEngagementPlaceholderRow` specifically so a successful-but-unmappable query degrades explicitly (to `data_available: false`) rather than silently type-casting old columns into the new shape and producing `undefined`/`NaN` throughout the response — so the failure mode you're describing (500 or non-joining rows) shouldn't actually be reachable today. But you're right that the live path genuinely can't produce real data yet.

This is blocked on two things outside this PR's control: the real `platinum_lfx_one_committee_meeting_attendance` model being deployed, and the live read surface (query-service vs. direct Snowflake) being decided per the ticket's Jira thread — both explicitly still open. Not something I can implement now; leaving the SQL as the documented placeholder for the follow-up.

manishdixitlfx:
## Review Feedback Addressed

Commit: 522ca31f60de79c77a179cc388e65a70391e1c40

### Changes Made
- **apps/lfx-one/src/server/helpers/snowflake-schema.helper.ts**: tightened the schema-qualifier regex from `{1,2}` to exactly two segments — every call site appends a table name unconditionally, so a 3-segment override produced an invalid 4-part Snowflake identifier (per copilot-pull-request-reviewer[bot], @dealako — same bug, flagged independently)
- **apps/lfx-one/src/server/helpers/snowflake-schema.helper.spec.ts**: updated the three-segment test to expect fallback instead of acceptance
- **apps/lfx-one/src/server/services/committee-engagement.service.ts**: switched the mock-mode log from WARN to INFO — mock is a deliberate explicit opt-in, not a degradation, per `logging-patterns.md` (per @dealako)
- **apps/lfx-one/src/server/services/committee-engagement.service.ts**: clamped `invited`/`attended`/`committeeMeetings` to non-negative in `countsForWindow`, matching the existing clamp's defensive posture (per @dealako)
- **packages/shared/src/utils/committee-engagement-classifier.utils.ts**: added a comment clarifying the `rate <= 0` check is defensive against the upstream clamp ever being weakened, not because a negative rate is reachable today (per @dealako)
- **apps/lfx-one/src/server/services/committee-engagement.service.spec.ts**: updated mocks/assertions for the INFO log level, added regression tests for the count clamp

### No Change Needed
- **apps/lfx-one/src/server/services/access-check.service.ts:236**: `checkAccessStrict` resolving a missing/malformed upstream tuple to `false` (rather than throwing) is a real, but already deliberate and documented, gap — called out explicitly in `committee-read-access.helper.ts`'s own doc comment as "incomplete, not absolute." Closing it fully means restructuring `performCheck`'s tuple resolution for strict callers, a real behavior change to a security-relevant path — left as documented rather than changed in this pass (flagged by copilot-pull-request-reviewer[bot])
- **apps/lfx-one/src/server/services/committee-engagement.service.ts:104**: the live SQL still targets the legacy placeholder shape, not the finalized model's real columns — the single most heavily-documented known limitation on this branch, tracked via the existing `TODO(LFXV2-1705 follow-up)` comments and the PR description's Known Limitations section. Blocked on the real dbt model deploying and the live read surface being decided upstream, both explicitly open per the ticket's Jira thread (flagged by copilot-pull-request-reviewer[bot])
- **apps/lfx-one/src/server/helpers/committee-engagement-backend.helper.ts**: already addressed in an earlier commit on this branch — mock now defaults off, requires explicit `ENGAGEMENT_BACKEND=mock`, and is hard-blocked in production (flagged by copilot-pull-request-reviewer[bot])

### Threads Resolved
8 of 8 unresolved threads addressed in this iteration.

dealako (approved):
Hey @manishdixitlfx — thanks for the quick turnaround. All four items from the last round are resolved, and nothing new turned up in a fresh security/correctness/style pass over the new commit.

👏 **Nice work:**
- The schema-qualifier regex is now exactly `WORD.WORD` (was `{1,2}` segments), matching every call site's assumption that a table name gets appended unconditionally — and the spec now asserts the 3-segment case *falls back* rather than just checking acceptance, so the regression is actually locked in.
- The warehouse-count clamp (`Math.max(x, 0)`) is applied consistently across `invited`/`attended`/`committeeMeetings`, not just the one field I flagged, with a regression test using negative fixture values.
- The mock-mode log level fix came with a doc comment explaining *why* INFO fits now (explicit opt-in, not a fallback) — good context for the next reader, and it's reinforced by the earlier mock-defaults-to-live change.
- The classifier's defensive `<=` comment change explains the intent without changing behavior, which was the right call for a purely cosmetic clarity nit.

### Summary
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 0

All four carried-over items (schema regex, mock-mode log level, non-negative warehouse clamp, classifier comment) are ✅ **Resolved**. No new findings from this commit.

### ✅ Approved

PR #1272, opened 2026-07-30
fix(meetings): enable previous occurrence navigation on join page

niravpatel27:
## Summary

On the public meeting join page, the Previous/Next occurrence navigation for recurring meetings always showed the current occurrence as "1 of N" with Previous permanently disabled. The nav logic was correct — the upstream `GET /itx/meetings/:id` payload only carries current + future occurrences, so the page never knew any past occurrences existed.

Fixes [LFXV2-2926](https://linuxfoundation.atlassian.net/browse/LFXV2-2926) (introduced with the original nav in LFXV2-1470).

## Changes

- **New public endpoint** `GET /public/api/meetings/:id/occurrences` — returns the series timeline: `past` (minimal projection of `v1_past_meeting` records queried by series `meeting_id` under M2M) + `future` (minimal `MeetingOccurrenceSummary` projection of the live series) + `cancelled_occurrences`. Timestamps only; no titles, passwords, or join info.
- **Access gate** — private/restricted series are gated by the same chain as the public meeting detail endpoint: valid password → registrant record (M2M email check) → organizer relation (checked with the user's own token, restored to M2M in `finally`). Fails closed with an opaque 403 when the live series can't be loaded.
- **Join page** merges the timeline into its occurrence nav via a new shared `buildOccurrenceNavTimeline` util (dedupe by start instant, priority live > future > past). Past entries link directly to `/meetings/{meeting_and_occurrence_id}` — no timestamp reconstruction. The nav now also works on past-meeting pages (`/meetings/{id}-{ts}`), which previously had no series context.
- New shared types (`PastOccurrenceSummary`, `PublicMeetingOccurrencesResponse`, `MeetingOccurrenceSummary`, `OccurrenceNavItem`) and a cast-free `getMeetingSeriesUid` accessor.

## Documented trade-offs

- **Fail-closed on unloadable series:** if the live `v1_meeting` is gone (deleted/fully-ended series), the endpoint returns 403 rather than serving past occurrences whose visibility can't be verified. The frontend degrades gracefully (nav simply absent). Security-over-feature by design.
- **Partial pagination tolerated:** a later-page query failure returns the accumulated pages (navigation is progressive enhancement); documented in the service JSDoc.
- **Upstream filter dependency:** `filters=meeting_id:{uid}` on `v1_past_meeting` follows the query service's standard `term` filter mechanics (`data.*` fields; same pattern used on `v1_past_meeting_participant` and registrant types). Please confirm in dev that the filter returns rows before merge — if the field ever proves unindexed, the fallback is a `tags`-based scope.

## Testing

- 24 new unit tests: shared util (merge order, dedupe priority, composite-id authority, cancelled filtering, duration fallback, series-uid accessor), BFF service (filter shape, no `filter_grants`, pagination, drop-invalid, error→empty), controller (access matrix incl. anonymous-denied / password / registrant / organizer / fail-closed, sensitive-field leak assertions, M2M token restore).
- `yarn lint`, `yarn build` (SSR bundle), and both vitest suites pass; local Playwright e2e is broken environment-wide (known issue) — manual browser verification recommended on a recurring meeting: label should read e.g. "27 of 76" with Previous walking into past occurrence pages and Next walking back toward today.

[LFXV2-2926]: https://linuxfoundation.atlassian.net/browse/LFXV2-2926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
+++ b/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
@@ -179,8 +179,8 @@
                     </div>
                   }
 
-                  <!-- Occurrence Navigation (recurring meetings only) -->
-                  @if (meeting().recurrence && occurrenceLabel()) {
+                  <!-- Occurrence Navigation (recurring meetings; past pages key off the series timeline since ITX past payloads omit `recurrence`) -->
+                  @if (occurrenceLabel()) {
                     <div class="flex items-center gap-3" data-testid="occurrence-navigation">
                       @if (previousOccurrenceUrl(); as prevUrl) {
                         <a
--- a/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
@@ -39,11 +39,11 @@ import {
   MeetingHostCandidate,
   MeetingOccurrence,
   MeetingRecurrence,
+  getMeetingSeriesUid,
   MeetingRegistrant,
   MeetingRsvp,
   OccurrenceNavItem,
   resolveOccurrenceRecurrence,
-  PastMeeting,
   PastMeetingAttachment,
   PastMeetingParticipant,
   PastMeetingRecording,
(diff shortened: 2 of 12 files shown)


MRashad26 (approved):
Security posture is solid: password travels in x-meeting-password header (not query string), canViewSeriesTimeline gates correctly (password → registrant → organizer with originalToken guard), M2M token restored in finally, fail-closed on unloadable series, future projection limited to 4 safe fields with a test asserting no secrets leak. buildOccurrenceNavTimeline dedupe priority (live > future > past) and composite-ID authority (suffix beats scheduled_start_time) are well-tested. Route ordering is correct.

One non-blocking nit: initializeSeriesOccurrences() uses isPlatformServer() for the SSR skip — CLAUDE.md convention is typeof window === 'undefined'. Non-blocking warning only.

andrest50 (approved):

PR #777, opened 2026-05-24
feat(weekly-brief): Angular frontend service

manishdixitlfx:
**Tracking:** [LFXV2-1983](https://linuxfoundation.atlassian.net/browse/LFXV2-1983) (part of Epic [LFXV2-1976](https://linuxfoundation.atlassian.net/browse/LFXV2-1976))

---

## Summary

Angular `WeeklyBriefService` that consumes the BFF endpoints from PR #776. Three methods: `getWeeklyBrief`, `generateWeeklyBrief`, `saveWeeklyBrief`.

## Notes

- `getWeeklyBrief` has `catchError` that logs before falling back to an empty envelope (so failures show up in DataDog RUM / dev console, not silent degradation).
- `generateWeeklyBrief` and `saveWeeklyBrief` deliberately have no `catchError` — callers handle 429 (throttle) and 409 (revision conflict) explicitly to drive the UI states.

## Stack

PR 3/4 on lfx-self-serve. **Base = PR #776** (bff).

## Test plan

- [ ] `yarn lint` / `yarn build` green (verified)
- [ ] Service hot-reloads against mock BFF

[LFXV2-1983]: https://linuxfoundation.atlassian.net/browse/LFXV2-1983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-1976]: https://linuxfoundation.atlassian.net/browse/LFXV2-1976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1277, opened 2026-07-30
feat(weekly-brief): shared interfaces + BFF proxy for wg weekly brief

manishdixitlfx:
## What

Foundation for the WG Weekly Brief feature — the shared types and the Express
BFF that the UI card consumes:

- `@lfx-one/shared` interfaces + constants for the weekly-brief domain.
- BFF route / controller / service for `/api/committees/:id/weekly-briefs/*`
  (GET current, POST generate, PUT save), with a `WEEKLY_BRIEF_BACKEND`
  mock/live switch (defaults to mock so the UI can iterate before the upstream
  endpoints are wired).
- Route mounted in `server.ts`; Angular `WeeklyBriefService`.
- `.env.example` documents `WEEKLY_BRIEF_BACKEND`.

## Why

Splits the feature so the BFF/contract lands independently of the UI card
(LFXV2-2176, stacked on this). Consolidates the original tangled stack
(#775–#778) against current `main`.

## Deployment changes

`charts/lfx-self-serve/values.yaml` adds `WEEKLY_BRIEF_BACKEND: "live"`. The
chart hardcodes `NODE_ENV=production` with no prior entry for this var, and
the BFF refuses to serve mock brief data under `NODE_ENV=production` unless
this is `"live"` (see Security below) — without this addition, the deploy
would 500 on every weekly-brief request as soon as anything calls them.
Verified the upstream `lfx-v2-committee-service` weekly-brief endpoints exist
in released tags back to v0.4.19.

## Protected files

`apps/lfx-one/src/server/server.ts` is touched — a two-line addition (router
import + `app.use('/api/committees', weeklyBriefRouter)`), matching the
existing registration pattern for every other route in the file. No route
shadowing: `committees.route.ts` has no `/:id/weekly-briefs/*` pattern or
catch-all that would intercept it.

## Security

Mock mode (the default) never calls upstream, so it also never enforces
committee-level authorization — any authenticated caller could otherwise read
a fabricated brief for any committee UID. The BFF now refuses to serve mock
data outright when `NODE_ENV=production` and `WEEKLY_BRIEF_BACKEND` isn't
`"live"` (500, generic client-facing message; the specific reason is
log-only). This is why the chart change above is load-bearing, not optional.

## Known trade-offs

- **`WeeklyBrief`/`WeeklyBriefThrottle` field-level optionality.** Upstream's
  Goa DSL marks nothing `Required` on these types, but the shared interfaces
  here keep them fully required to describe the "well-formed brief" shape
  without rippling `?.` through the not-yet-rebased LFXV2-2176 UI branch.
  (`GenerateWeeklyBriefResponse.brief`/`.throttle` — the one envelope this
  branch's own controller already reads defensively — are optional, matching
  the code that reads them.) Revisit field-level optionality when LFXV2-2176
  rebases onto this branch.
- **Mock mode is stateless.** It can't reach the 429-throttle-exhausted or
  `state: 'error'` terminal paths — acceptable for a local-dev convenience
  mock; the 202/`generating` envelope it returns still exercises the same
  poll-until-terminal shape the live contract requires.
- **The Angular client's `getWeeklyBrief` has no `catchError`.** This
  deviates from this repo's usual GET convention
  (`docs/reviews/frontend-checklist.md` §13), deliberately: a blanket
  `of(defaultValue)` fallback would render a misconfigured deploy or a 404
  identically to "no brief yet, 2 generates available." LFXV2-2176 (the first
  real consumer) owns classifying a failed read into its own error state.

## Testing

- `yarn lint`, `yarn format`, `yarn test` (543 + 193 tests), `yarn build`,
  and `./check-headers.sh` all green.
- Unit coverage added for the BFF service/controller (mock vs. live mode,
  production auth-refusal, 202/409/429 status and body forwarding, request
  validation, Unicode-correct `brief_text` length bounds, brief-window date
  selection) and for the shared constants (pinned against their documented
  values).
- Went through 5 rounds of full-branch review (`lfx-self-serve-code-reviewer`
  + `lfx-self-serve-learnings-reviewer`) during development; each round's
  findings were fixed and re-verified before the next, down to the three
  documented trade-offs above. Final round: both reviewers confirm clean.

## Notes

- Live path requires the upstream committee-service weekly-brief endpoints
  (deployed) + `WEEKLY_BRIEF_BACKEND=live` (see Deployment changes).
- Supersedes #775, #776, #777, #778 (already closed; comment added on each
  pointing at this stack).

LFXV2-2175


apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright The Linux Foundation and each contributor to LFX.
```
dealako, line 1: [minor] Missing unit tests for the Angular `WeeklyBriefService`

Issue: This file has no corresponding `.spec.ts`, while the backend controller, backend service, and shared constants all got dedicated spec coverage in this same PR.

Proof: `getWeeklyBrief`/`generateWeeklyBrief`/`saveWeeklyBrief` are never exercised directly — only indirectly via the backend proxy tests, which don't cover this file's own URL construction, `encodeURIComponent(committeeId)` usage, or the fact none of the three methods use `catchError` (intentional, per the inline comments, but untested).

Why it matters: A regression here (wrong path, wrong HTTP verb, missing `take(1)`, or an accidentally-added `catchError` that swallows a real error) would only surface in E2E or manual testing.

Fix: Add `weekly-brief.service.spec.ts` with `HttpClientTestingModule`/`HttpTestingController` tests asserting the three requests hit the right method + URL (including the encoded `committeeId`), and that errors propagate untouched to the caller.

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  SaveWeeklyBriefRequest,
+  WeeklyBrief,
+  WeeklyBriefCurrentResponse,
+} from '@lfx-one/shared/interfaces';
+import { Observable, take } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class WeeklyBriefService {
```
dealako, line 24: [minor] Missing JSDoc on the frontend `WeeklyBriefService` class and methods

Issue: The class and its three public methods have no JSDoc, unlike the backend `WeeklyBriefService` (`apps/lfx-one/src/server/services/weekly-brief.service.ts`), which documents the class and every method.

Proof: Compare this class (inline `//` comments explaining the "why" of omitting `catchError", but no `/** */` blocks) to the backend service's JSDoc on `getCurrentBrief`/`generateBrief`/`saveBrief`.

Why it matters: No JSDoc means no IDE-tooltip documentation for consumers of this service, and an inconsistency in doc coverage between the two halves of the same feature.

Fix: Add a class-level JSDoc summarizing the service's purpose, and per-method JSDoc noting the endpoint each method calls and that errors are intentionally left uncaught for the caller to classify.

dealako (changes requested):
Hey @manishdixitlfx — thanks for laying this foundation down cleanly. This is a well-scoped PR: shared interfaces/constants for the WG Weekly Brief, a BFF proxy that switches between mock and live upstream via `WEEKLY_BRIEF_BACKEND`, and solid controller/service test coverage on the backend side. The mock-refusal-in-production guard (`isLive()` + `NODE_ENV=production` check) is a nice defensive touch, and the upstream 202/409/429 forwarding is handled correctly rather than collapsed to generic responses.

I ran three parallel reviewers across Security/Privacy, Correctness/Performance/Tests, and Style/Docs, then reconciled against the existing CodeRabbit, Copilot, and Cursor Bugbot comments already on this PR.

### 🔴 Blocking: 0

### 🟡 Minor: 3
- **Missing unit tests for the Angular `WeeklyBriefService`** — the backend controller/service/constants all got dedicated spec files, but `apps/lfx-one/src/app/shared/services/weekly-brief.service.ts` has none. Worth a `.spec.ts` covering the three HTTP calls (correct method/path, `encodeURIComponent` on `committeeId`, no swallowed errors).
- **Missing JSDoc on the frontend `WeeklyBriefService`** — the backend service has full JSDoc on the class and every method; the frontend mirror only has inline "why" comments. Worth aligning for IDE-tooltip discoverability.
- **Agreeing with [Copilot's comment](https://github.com/linuxfoundation/lfx-self-serve/pull/1277#discussion_r3686240390) on `packages/shared/src/interfaces/weekly-brief.interface.ts`** — `WeeklyBriefSourceRef.title`/`.excerpt` are typed as required `string`, but the same file's own docstring says the upstream converter omits them when empty and `GroupWeeklyBriefSourceRef` marks nothing required. Since nothing in this PR consumes these fields yet (UI card is a follow-up), this isn't blocking, but the type should go optional now so the eventual consumer doesn't inherit a false guarantee.

### ⚪ Nit: 1
- **Agreeing with [CodeRabbit's comment](https://github.com/linuxfoundation/lfx-self-serve/pull/1277#discussion_r3686253684)** on `packages/shared/src/constants/weekly-brief.constants.ts` — the "rolling week" doc language doesn't match the fixed Sunday–Saturday window `briefWindow()`/`nextSundayIso()` actually implement.

### ❔ Question: 1
- **[Copilot raised](https://github.com/linuxfoundation/lfx-self-serve/pull/1277#discussion_r3687501065) — is `committee#viewer` the intended authorization scope for GET `/weekly-briefs/current`?** I traced this: the BFF (`weekly-brief.service.ts`'s `getCurrentBrief`) does no server-side role check of its own in live mode — it proxies straight through to committee-service and trusts upstream authorization entirely, which is a reasonable BFF-layer design. But per Copilot, the upstream FGA relation this endpoint currently relies on (`committee#viewer`) resolves to `[user:*] or member or auditor` — meaning any authenticated user who knows a committee UID could read draft/edited brief text, source excerpts, the private-source marker, and the editor's username. I can't verify the upstream FGA policy from this repo, so I can't confirm this is a live gap vs. already scoped correctly — but given the brief carries pre-publication content, it's worth an explicit confirmation from you or the committee-service owner that `viewer` is the intended (not accidental) exposure level before this proxy goes live for real committees.

Security/data-privacy dimension otherwise came back clean: no PII leaked into logs, no secrets, request bodies are field-whitelisted, `committeeId` is properly URL-encoded, and the production mock-refusal path is tested (`weekly-brief.service.spec.ts`).

Requesting changes primarily because of the 3 minor items above (repo convention here treats >2 minors as changes-requested) plus the open authorization question — none of this looks hard to close out. Nice work on the mock/live split and the defensive test coverage.

packages/shared/src/interfaces/weekly-brief.interface.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Matches upstream's `GroupWeeklyBriefSourceRef` exactly — `kind` is an open
+ * string (not an enum; documented values include "meeting", "mailing-list",
+ * "doc"), not the invented `source_type` shape this used to have.
+ */
+export interface WeeklyBriefSourceRef {
+  excerpt: string;
+  id: string;
+  kind: string;
+  title: string;
+}
```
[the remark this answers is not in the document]
manishdixitlfx: Done — `WeeklyBriefSourceRef.title`/`.excerpt` are now optional (`title?`/`excerpt?`), matching upstream's `GroupWeeklyBriefSourceRef` (nothing Required) and the committee-service converter's behavior of omitting them when empty.

See commit 2db20df32: packages/shared/src/interfaces/weekly-brief.interface.ts.

packages/shared/src/constants/weekly-brief.constants.ts
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: MIT
+
+/**
+ * Default WG Weekly Brief throttle counters.
+ *
+ * Used by the BFF (`apps/lfx-one/src/server/services/weekly-brief.service.ts`) for
+ * its mock-mode envelope. The runtime `window_resets_at` is computed at the call
+ * site and is intentionally not part of this constant.
+ *
+ * Policy: 2 fresh generates and 3 regenerations per rolling week.
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — the throttle-window doc now describes a fixed Sunday-Saturday calendar week (matching `WeeklyBrief.window_start`/`.window_end` and the next-Sunday reset in `nextSundayIso()`) instead of "rolling week". Good catch!

See commit 2db20df32: packages/shared/src/constants/weekly-brief.constants.ts.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -0,0 +1,251 @@
...
+          window_resets_at: nextSundayIso(),
+        },
+      };
+    }
+
+    logger.debug(req, 'get_weekly_brief_current', 'Proxying to committee-service', { committee_id: committeeId });
+    return this.microserviceProxy.proxyRequest<WeeklyBriefCurrentResponse>(
+      req,
+      'LFX_V2_SERVICE',
+      `/committees/${encodeURIComponent(committeeId)}/weekly-briefs/current`,
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch — you're right this endpoint had no server-side authorization at all, it proxied straight through and trusted committee-service's own FGA enforcement entirely. Rather than depend on confirming the upstream committee-service relation, I added a defense-in-depth gate at this BFF layer: `getCurrentBrief` now calls `assertCommitteeRead` (`committee#auditor`, not `viewer`) before proxying — same helper/pattern the LFXV2-1705 committee-engagement endpoint already uses, so an unrelated authenticated caller now gets a 403 before any upstream call is made, independent of what `viewer` resolves to on the committee-service side.

Note: `generateBrief`/`saveBrief` have the identical missing-authorization gap for write access (arguably more severe), found while fixing this. Out of scope for this iteration — tracking as a separate follow-up rather than expanding this fix under review-response time pressure.

See commit 2db20df32: weekly-brief.controller.ts, weekly-brief.controller.spec.ts.

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
manishdixitlfx: Fixed — added class-level JSDoc plus per-method JSDoc on `getWeeklyBrief`/`generateWeeklyBrief`/`saveWeeklyBrief`, documenting the endpoint each calls and why `catchError` is intentionally omitted (mirroring the backend service's doc coverage).

See commit 2db20df32: apps/lfx-one/src/app/shared/services/weekly-brief.service.ts.

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
manishdixitlfx: Good catch on the coverage gap, but I don't think adding this spec is the right fix here — this repo has no Angular test builder wired up. `apps/lfx-one/vitest.config.ts` scopes `include` to `src/server/**/*.spec.ts` only, and there's no Karma/Jest config for `src/app/`, so a `weekly-brief.service.spec.ts` under `src/app/shared/services/` would sit in the repo but never actually run under `yarn test` — it'd be dead code that looks like coverage. This is the same gap LFXV2-2176's PR description calls out for its own component spec.

Closing this gap for real would mean wiring up an Angular test builder for the whole app (angular.json + a runner), which is a repo-wide infra change well beyond this PR's scope. Happy to file that as its own ticket if you think it's worth prioritizing — for now the frontend service's correctness is covered indirectly via the backend proxy tests plus the Playwright e2e in LFXV2-2176.

No code change made here.

manishdixitlfx:
## Review Feedback Addressed

Commit: 2db20df328bd1c03c3470f273ef685f80ffd54da

### Changes Made
- **weekly-brief.controller.ts**: gated `getCurrentBrief` with `assertCommitteeRead` (`committee#auditor`) before proxying — the endpoint had no server-side authorization of its own (per copilot-pull-request-reviewer, elevated by @dealako)
- **weekly-brief.interface.ts**: `WeeklyBriefSourceRef.title`/`.excerpt` are now optional, matching upstream's contract (per copilot-pull-request-reviewer, agreed by @dealako)
- **weekly-brief.constants.ts**: throttle-window doc now says fixed Sunday–Saturday week, not "rolling week" (per coderabbitai, agreed by @dealako)
- **weekly-brief.service.ts** (frontend): added class/method JSDoc matching the backend service's doc coverage (per @dealako)

### No Change Needed
- **weekly-brief.service.ts** (frontend), missing `.spec.ts` (asked by @dealako): not added — `apps/lfx-one/vitest.config.ts` scopes `include` to `src/server/**/*.spec.ts` only and there's no Angular test builder wired up, so a spec under `src/app/` would never actually execute under `yarn test`. Left this thread open for your take on whether it's worth a separate ticket to wire up an app-level test builder.

### New Finding (not fixed in this commit)
While closing the GET authorization gap, found `generateBrief`/`saveBrief` have the identical missing-authorization gap for write access (any authenticated caller can currently generate/save a brief for any committee UID in live mode) — arguably more severe than the read gap. Deliberately out of scope here; will track as a follow-up.

### Threads Resolved
4 of 5 unresolved threads addressed in this iteration.

### Still Open
- **weekly-brief.service.ts:1** (frontend spec request): left open pending your read on the test-builder gap above.

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
```diff
@@ -0,0 +1,223 @@
...
+          ServiceValidationError.fromFieldErrors(validation.fieldErrors, 'Invalid generate-weekly-brief request body', {
+            operation: 'generate_weekly_brief',
+            service: 'weekly_brief_controller',
+            path: req.path,
+          })
+        );
+      }
+      const body = validation.value;
+
+      const { status, data } = await this.weeklyBriefService.generateBrief(req, committeeId, body);
```
dealako, line 162: [blocking] Write-path authorization gap: `generateBrief` proxies without a committee access check

Issue: `getCurrentBrief` now gates on `await assertCommitteeRead(req, committeeId, 'get_weekly_brief_current')` before proxying (line 105), but `generateBrief` calls `this.weeklyBriefService.generateBrief(...)` here with no equivalent check.

Proof: Compare to line 105 — the same helper (`assertCommitteeRead`, imported at the top of this file) is available and already proven out for GET, but isn't called here or in `saveBrief` (line 209).

Why it matters: In live mode, any authenticated caller who knows a committee UID can generate a brief for a committee they have no FGA relation to, consuming that committee's throttle quota and potentially triggering upstream side effects (LLM generation calls) on behalf of an unauthorized user. This is a write-side instance of the same exposure the GET fix in this commit was meant to close.

Fix: Add `await assertCommitteeRead(req, committeeId, 'generate_weekly_brief');` immediately after the body validation and before the `this.weeklyBriefService.generateBrief(...)` call, mirroring the `getCurrentBrief` pattern. Add a test asserting a failed check short-circuits before the service/proxy call runs.

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
```diff
@@ -0,0 +1,223 @@
...
+          ServiceValidationError.fromFieldErrors(validation.fieldErrors, 'Invalid save-weekly-brief request body', {
+            operation: 'save_weekly_brief',
+            service: 'weekly_brief_controller',
+            path: req.path,
+          })
+        );
+      }
+      const body = validation.value;
+
+      const result = await this.weeklyBriefService.saveBrief(req, committeeId, body);
```
dealako, line 223: [blocking] Write-path authorization gap: `saveBrief` proxies without a committee access check

Issue: Same gap as `generateBrief` above — `saveBrief` calls `this.weeklyBriefService.saveBrief(...)` here with no `assertCommitteeRead` (or equivalent) check beforehand.

Proof: `getCurrentBrief` (line 105) is the only one of the three controller methods that calls `assertCommitteeRead` before proxying.

Why it matters: Any authenticated caller who knows a committee UID could overwrite (edit) another committee's brief text in live mode, bumping its revision and marking it edited — a content-integrity issue, not just an information-disclosure one.

Fix: Add `await assertCommitteeRead(req, committeeId, 'save_weekly_brief');` before the `this.weeklyBriefService.saveBrief(...)` call, with test coverage for the gate short-circuiting on failure.

packages/shared/src/constants/weekly-brief.constants.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+import { WEEKLY_BRIEF_DEFAULT_THROTTLE, WEEKLY_BRIEF_TEXT_MAX_LENGTH } from './weekly-brief.constants';
+
+describe('WEEKLY_BRIEF_TEXT_MAX_LENGTH', () => {
+  it('matches upstream UpdateCurrentWeeklyBriefRequestBody.brief_text maxLength', () => {
+    expect(WEEKLY_BRIEF_TEXT_MAX_LENGTH).toBe(20_000);
+  });
+});
+
+describe('WEEKLY_BRIEF_DEFAULT_THROTTLE', () => {
+  it('matches the documented policy of 2 fresh generates and 3 regenerations per rolling week', () => {
```
dealako: [nit] Test description still says "rolling week" after the doc comment was fixed to say otherwise

Issue: `weekly-brief.constants.ts`'s doc comment was corrected this round to describe a "fixed Sunday–Saturday calendar week" (explicitly "not a rolling 7-day window"), but this test's description string wasn't updated to match.

Proof: This line reads `'matches the documented policy of 2 fresh generates and 3 regenerations per rolling week'` — the word "rolling" now contradicts the source-of-truth comment it's meant to pin.

Why it matters: Test names double as documentation; a future reader skimming test output would infer rolling-window semantics that the actual policy explicitly disclaims.

Fix: Rename to something like `'matches the documented policy of 2 fresh generates and 3 regenerations per fixed Sunday–Saturday calendar week'`.

dealako (changes requested):
Hey @manishdixitlfx — thanks for the quick, thorough turnaround, and especially for the transparency in your remediation summary. Calling out the write-path gap yourself before I could even ask is exactly the kind of disclosure that makes follow-up review fast.

### 👏 Nice work
- The `assertCommitteeRead` gate on `getCurrentBrief` is implemented correctly — it runs before the proxy call, short-circuits via `next(error)` on failure, and is tested.
- `WeeklyBriefSourceRef.title`/`.excerpt` are now properly optional and match upstream's actual (non-required) contract.
- Frontend `WeeklyBriefService` JSDoc is comprehensive — class and all three methods documented, mirroring the backend service's coverage.
- Your rationale for not adding a frontend `.spec.ts` checks out: I verified `apps/lfx-one/vitest.config.ts` scopes `include` to `src/server/**/*.spec.ts` only, so a spec under `src/app/` genuinely wouldn't execute today. Leaving that as a separate test-infra ticket is the right call rather than adding dead test code to this PR.
- I also independently verified CodeRabbit's new nullable-`data` concern on the generate response handler and don't think it's live: `data` is typed as `GenerateWeeklyBriefResponse` (never null) on the success path, and every field read on it (`data.brief?.uid`, `data.throttle?.generates_used`, etc.) already uses optional chaining. Treating that one as a non-issue.

### 🔴 Blocking: 1
- **`generateBrief` (POST) and `saveBrief` (PUT) still proxy to committee-service without any `assertCommitteeRead` gate** — confirmed at `weekly-brief.controller.ts:153` and `weekly-brief.controller.ts:209`. This is the exact vulnerability class you just fixed for GET, but for write operations: any authenticated caller who knows a committee UID can currently generate or overwrite a brief for a committee they have no access to, assuming upstream's write authorization is as permissive as the `viewer`-adjacent relation that motivated the GET fix. Given this proxy is meant to go live (`WEEKLY_BRIEF_BACKEND: "live"` is already set in `charts/lfx-self-serve/values.yaml`), I don't think this can ride as a follow-up ticket — it's a live write-access bypass, not a hardening nice-to-have. Please add the same `assertCommitteeRead(req, committeeId, '<operation>')` call to both methods before the proxy call, matching the `getCurrentBrief` pattern, with test coverage for the gate short-circuiting on failure.

### ⚪ Nit: 1
- **`packages/shared/src/constants/weekly-brief.constants.spec.ts:21`** — the constants file's doc comment was correctly updated to "fixed Sunday–Saturday calendar week," but this test's description string still reads `'...per rolling week'`. Small terminology mismatch between the fixed doc and the still-unfixed test name — worth a one-line rename while you're in there for the auth fix.

The blocking item is the only thing standing between this and approval — everything else from the first round is resolved or a trivial nit. Once the write-path gate lands (and ideally the test-name nit), this should be quick to close out.

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
manishdixitlfx: Fixed — added `assertCommitteeWrite` (`committee#writer`) and wired it into `generateBrief` before the service call, mirroring `getCurrentBrief`'s `assertCommitteeRead` gate. Verified against the actual FGA model (`lfx-v2-helm`'s `model.fga`): `writer` is the relation explicitly tagged `@fgadoc:jtbd Generate & edit the committee weekly brief` — used `writer`, not `auditor`, since a read-only auditor shouldn't be able to trigger a write. New helper file (`committee-write-access.helper.ts`) rather than extending the read-access helper, to avoid touching a file owned by the already-merged LFXV2-1705. Added controller spec coverage for the gate and its 403 propagation.

See commit 81751957b.

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
manishdixitlfx: Fixed — same `assertCommitteeWrite` gate added to `saveBrief` before the service call, plus spec coverage for the gate and its 403 propagation.

See commit 81751957b.

packages/shared/src/constants/weekly-brief.constants.spec.ts
manishdixitlfx: Fixed — renamed the test description to "...per fixed calendar week" to match the doc comment. Good catch!

See commit 81751957b.

manishdixitlfx:
## Round 2 Review Feedback Addressed

Commit: 81751957beb9f0a02d4c54ba775d4e0667680074

### Changes Made
- **committee-write-access.helper.ts** (new): `assertCommitteeWrite` gate on `committee#writer`, verified against the actual FGA model (`lfx-v2-helm`'s `model.fga`) — `writer` is the relation tagged `@fgadoc:jtbd Generate & edit the committee weekly brief`
- **weekly-brief.controller.ts**: wired the gate into `generateBrief` and `saveBrief` before their service calls, matching `getCurrentBrief`'s pattern, with spec coverage for both gates and their 403 propagation (per @dealako)
- **weekly-brief.constants.spec.ts**: test description now says "fixed calendar week" instead of "rolling week" (per @dealako)

### Threads Resolved
3 of 4 unresolved threads addressed in this iteration (the frontend-spec-builder-gap thread from round 1 remains open, unchanged).

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * Narrow `req.body` to a GenerateWeeklyBriefRequest. Upstream's
+ * GenerateWeeklyBriefRequestBody has exactly one field (`force: boolean`) —
+ * this whitelists it explicitly rather than forwarding `req.body` verbatim,
+ * so a caller cannot smuggle extra fields through to the proxy call.
+ */
+function validateGenerateBriefBody(body: unknown): { ok: true; value: GenerateWeeklyBriefRequest } | { ok: false; fieldErrors: Record<string, string> } {
+  if (body === undefined || body === null) {
+    return { ok: true, value: {} };
```
dealako, line 58: [blocking] Array request bodies bypass `validateGenerateBriefBody`'s object check

Issue: `typeof body !== 'object'` is `false` for arrays, since `typeof [] === 'object'` in JavaScript. A `[]` body passes this guard.

Proof: With `body = []`, line 58's check doesn't trip, `b = body as Record<string, unknown>` casts the array, `b['force']` is `undefined`, and the `force !== undefined && typeof force !== 'boolean'` check on the next line is also false (since `undefined !== undefined` is false). The function returns `{ ok: true, value: { force: undefined } }` — a malformed request treated as a valid "non-forced generate."

Why it matters: This lets a caller send a garbage array body and still consume a generation-quota slot instead of getting a 400. No existing test in `weekly-brief.controller.spec.ts` covers an array body — only undefined/null, `force: true`, non-boolean `force`, and unknown-field cases are covered.

Fix: Reject arrays explicitly before (or as part of) the object-type check:
```ts
if (Array.isArray(body) || typeof body !== 'object') {
  return { ok: false, fieldErrors: { body: 'Request body must be a JSON object' } };
}
```
Add a test asserting an array body is rejected with a 400 and never reaches `weeklyBriefService.generateBrief`.

apps/lfx-one/.env.example
```diff
@@ -192,3 +192,11 @@ GUILD_API_URL=https://app.guild.ai
 GUILD_API_KEY=your-guild-api-key
 GUILD_WORKSPACE_OWNER=your-guild-workspace-owner
 GUILD_WORKSPACE_NAME=your-guild-workspace-name
+
+# Set to 'live' to proxy to committee-service; unset (or any other value) = mock data.
+# Inline comments on the same line as a dotenv assignment are commonly parsed as
+# part of the value, which would break the strict === 'live' check in code.
+# MUST be 'live' in production — mock mode never calls upstream, so it also
```
dealako: [minor] Comment now understates what mock mode enforces

Issue: This comment says mock mode "never enforces committee-level authorization," which was accurate before this round's fix but no longer is: `assertCommitteeWrite`/`assertCommitteeRead` now run at the controller level before the mock/live branch is even reached, so authorization is enforced in both modes. Only the actual upstream committee-service brief-data call is skipped in mock mode.

Why it matters: A future reader could infer mock mode has an authorization bypass that no longer exists, and misjudge the risk of running mock mode somewhere it shouldn't.

Fix: Reword to something like: "MUST be 'live' in production — mock mode still enforces committee-level authorization (assertCommitteeWrite/assertCommitteeRead run before the mock/live branch), but never calls the real committee-service, so it would silently serve fabricated brief content; the server refuses to start serving mock data at all when NODE_ENV=production."

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -0,0 +1,251 @@
...
+        undefined,
+        body
+      );
+    } catch (error) {
+      throw this.withConflictBody(error);
+    }
+  }
+
+  /**
+   * Refuses mock mode outright in production instead of silently serving
```
dealako, line 206: [minor] JSDoc rationale is now stale re: mock-mode authorization

Issue: Same claim as the `.env.example` comment — this JSDoc says mock mode "never enforces committee-level authorization," but as of this round's fix, `assertCommitteeWrite`/`assertCommitteeRead` run at the controller level before this service (and its mock/live branch) is ever reached.

Why it matters: This JSDoc no longer accurately describes the failure mode it's guarding against — the production refusal is still the right defensive check (prevents serving fabricated content), but the "any authenticated caller would get a brief for any committee UID" framing is outdated now that the controller gates every caller regardless of mode.

Fix: Reword to focus on what mock mode still means (fabricated content, not an auth bypass), e.g.: "Refuses mock mode outright in production instead of silently serving fabricated brief content. Committee-level authorization is enforced at the controller level (assertCommitteeWrite/assertCommitteeRead) regardless of backend mode — this check exists purely to prevent a misconfigured deploy from serving mock data in production."

dealako (changes requested):
Hey @manishdixitlfx — the write-path fix landed cleanly, and I traced it myself: `assertCommitteeWrite` (gated on `committee#writer`) runs before both `generateBrief`'s and `saveBrief`'s service calls, mirrors `getCurrentBrief`'s pattern exactly, and short-circuits correctly on a failed check. That closes the round-2 blocking item.

### 👏 Nice work
- `assertCommitteeWrite`/`committee-write-access.helper.ts` gates both write endpoints before any proxy call, with test coverage for the 403 short-circuit — confirmed by independent read, not just your summary.
- The FGA relation choice (`committee#writer`, tagged `@fgadoc:jtbd Generate & edit the committee weekly brief`) is the right one and matches the existing `assertCommitteeRead`/`committee#auditor` pattern already proven out for GET.
- The `weekly-brief.constants.spec.ts:21` "rolling week" → "fixed calendar week" test-description fix landed and now matches the doc comment's terminology.

### 🔴 Blocking: 1
- **`validateGenerateBriefBody` accepts array request bodies** (`weekly-brief.controller.ts:58`) — `typeof body !== 'object'` is `false` for arrays (`typeof [] === 'object'` in JS), so a `[]` body sails through as `{ force: undefined }` instead of being rejected as malformed. This lets a garbage request consume a generation-quota slot instead of returning 400. (Originally a suppressed low-confidence Copilot comment — I verified it's real and reachable, so it's not staying suppressed.)

### 🟡 Minor: 2
- **Two stale doc comments now understate what mock mode enforces** (`.env.example:199` and `weekly-brief.service.ts:206`) — both say mock mode "never enforces committee-level authorization," which was true before this round but is no longer accurate: `assertCommitteeWrite`/`assertCommitteeRead` now run at the controller level *before* the mock/live branch, so authorization is enforced in both modes — only the actual upstream committee-service data call is skipped in mock mode. Worth a quick edit so a future reader doesn't infer a bypass that no longer exists.

### ❔ Carried-forward observation (not new, not blocking this round)
- Your own diff's controller code now defensively optional-chains every field it reads off `data.brief` (`data.brief?.uid`, `.state`, etc.) — which is the right instinct, because `WeeklyBrief.brief_text`/`.source_refs`/`.prompt_version`/`.model` are still typed as required on the shared interface even though a live `generating`-state response can legitimately omit them (same class of issue as the `WeeklyBriefSourceRef.title`/`.excerpt` fix from round 1). This predates this commit and isn't part of this diff, so I'm not blocking on it, but it's the same fix pattern applied to the parent type — worth a follow-up.

Just the array-body validation gap standing between this and approval — the two doc-staleness items are one-line comment edits. Great progress closing the auth gap this round.

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
manishdixitlfx: Fixed — added `Array.isArray(body)` to the `typeof body !== 'object'` check, so an array body is now correctly rejected instead of sailing through as `{ force: undefined }`. Added spec coverage. Good catch on this one.

See commit 8c555c1dc5056ad7f0e55b4ee9944b11575164a0.

apps/lfx-one/.env.example
manishdixitlfx: Fixed — reworded to describe what the production mock-refusal actually guards against now: a misconfigured deploy serving fabricated content to an already-authorized caller, since assertCommitteeRead/assertCommitteeWrite gate every request before the mock/live branch regardless of mode.

See commit 8c555c1dc5056ad7f0e55b4ee9944b11575164a0.

apps/lfx-one/src/server/services/weekly-brief.service.ts
manishdixitlfx: Fixed — same correction as the .env.example comment: this now describes the misconfigured-deploy risk the check actually guards against, not an authorization bypass (which the controller-level gates now close regardless of mock/live mode).

See commit 8c555c1dc5056ad7f0e55b4ee9944b11575164a0.

manishdixitlfx:
## Round 3 Review Feedback Addressed

Commit: 8c555c1dc5056ad7f0e55b4ee9944b11575164a0

### Changes Made
- **weekly-brief.controller.ts**: `validateGenerateBriefBody` now rejects array bodies explicitly (`typeof [] === 'object'` was letting `[]` through as `{ force: undefined }` instead of a 400). Added spec coverage.
- **apps/lfx-one/.env.example**, **weekly-brief.service.ts**: fixed both stale doc comments claiming mock mode "never enforces committee-level authorization" — now describes what the production mock-refusal actually guards against (a misconfigured deploy serving fabricated content to an already-authorized caller), since the round-2 controller-level gates now enforce authorization in both modes.

### Threads Resolved
3 of 4 unresolved threads addressed (the frontend-spec-builder-gap thread remains open, unchanged from round 1).

dealako (dismissed):
Hey @manishdixitlfx — this closes out clean. Verified by all three re-review passes, and I spot-checked the array-body fix and both doc rewrites myself:

### 👏 Nice work
- `validateGenerateBriefBody` now explicitly rejects array bodies before the `typeof` check, with a test asserting the service is never called and the error propagates via `next()` — not a shallow test, it exercises the actual rejection path.
- Both doc comments (`.env.example:199`, `weekly-brief.service.ts:206`) are reworded accurately: they now correctly describe `assertCommitteeRead`/`assertCommitteeWrite` as controller-level gates that run in both mock and live mode, and correctly reframe the production mock-refusal as guarding against *fabricated content*, not an authorization bypass. Internally consistent, no new inaccuracies introduced.
- Across four rounds, every blocking and minor finding — the write-path authorization gap, the array-validation bypass, the stale doc comments, the constants test-description nit — has been independently verified fixed, not just claimed fixed. Good collaborative back-and-forth.

### 🔴 Blocking: 0
### 🟡 Minor: 0
### ⚪ Nit: 0

### ❔ Still-open (non-blocking) follow-ups worth separate tickets, not this PR
- `WeeklyBrief.brief_text`/`.source_refs`/`.prompt_version`/`.model` are still typed as required on the shared interface even though a live `generating`-state response can legitimately omit them (same class of fix already applied to `WeeklyBriefSourceRef.title`/`.excerpt`). Pre-existing, not touched by any commit in this PR's review cycle — worth a quick follow-up type fix.
- The frontend-spec-builder-gap thread from round 1 (no `.spec.ts` for the Angular `WeeklyBriefService`, since the current vitest config doesn't execute `src/app/**` specs anyway) remains open by design, per your earlier note — fine to leave as a test-infra ticket rather than adding dead test code here.

### Final decision: ✅ Approved with comments

Nice work driving this through four review rounds — the write-path authorization fix in particular was the right call to prioritize given `WEEKLY_BRIEF_BACKEND` is already `live` in prod config.

apps/lfx-one/src/server/services/weekly-brief.service.ts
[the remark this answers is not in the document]
manishdixitlfx: Following up with the product call on this: keeping reads at `committee#viewer` scope is a deliberate decision (Manish, 2026-07-31), not an oversight — brief content derives from data those viewers already see, so gating the read more tightly than the source data doesn't add real protection. Write paths (generate/save) remain `committee#writer`-gated as already implemented. The write-access follow-up noted in your earlier review stays tracked as follow-up work, unchanged. No code change from this reply — leaving this thread open for your read.

apps/lfx-one/src/app/shared/services/weekly-brief.service.ts
manishdixitlfx: Update: this is tracked as LFXV2-2927 (the Angular/vitest runner is broken repo-wide — LFXV2-2927 names this spec file in its own acceptance criteria). The frontend `weekly-brief.service.spec.ts` lands there once the runner works, rather than being added dead here. Backend controller/service specs cover the request/response contract meanwhile. Leaving this thread open for your read.

manishdixitlfx:
## Review Items Handled

Commit: 79f9012d115feec689ac4d3b75f109b24115b069

### Changes Made
- **committee-access.internal.helper.ts** (new): extracted the shared mechanics between `assertCommitteeRead` and `assertCommitteeWrite` (identical forbidden/unavailable construction, checkSingleAccessStrict handling) into one `assertCommitteeAccess(req, committeeUid, operation, access, verb)`. Both public functions are now thin wrappers owning only their relation choice (`auditor` vs `writer`). Existing `committee-read-access.helper.spec.ts`/`committee-write-access.helper.spec.ts` pass unchanged (mocks apply transitively); added `committee-access.internal.helper.spec.ts` for the shared implementation.
- **weekly-brief.controller.spec.ts**: added the missing exact-boundary case — a 20,000-code-point brief_text (the limit itself, inclusive) is accepted, alongside the existing 15,000-accept and 20,001-reject cases.

### Verified, No Change Needed
- `weekly-brief.constants.spec.ts:21` already reads "fixed calendar week" — fixed in the round-2 commit, confirmed still correct.

### Replied (left open for your read)
- **Frontend `weekly-brief.service.ts` spec thread**: tracked as LFXV2-2927 (repo-wide broken Angular/vitest runner; that ticket names this spec in its own acceptance criteria) — not adding it here.
- **Authorization-scope thread**: product decision — reads stay at `committee#viewer` scope (deliberate, brief content derives from already-viewer-visible data); write paths remain `committee#writer`-gated as implemented; the write-access follow-up stays tracked separately.

### Note on scope
Couldn't find the two "CodeRabbit nitpicks" (validateCommittee extraction, 20k boundary test) as actual GitHub review comments — CodeRabbit's only substantive comment on this PR is the round-1 "rolling week" one, already resolved. Implemented both anyway since the underlying issues are real on inspection (assertCommitteeRead/assertCommitteeWrite did duplicate identical logic; the exact-limit case was genuinely untested) — flagging the attribution mismatch for the record.

dealako (approved):

PR #1279, opened 2026-07-30
feat(weekly-brief): ai assistant card on committee overview

manishdixitlfx:
## What

The AI Assistant "Weekly Brief" card on the committee/working-group overview:

- `weekly-brief-card` component — generate / regenerate / edit / save / copy,
  with loading, generating, generated, edited, error, **unavailable (retryable)**,
  and empty states. Polls `GET /current` after a generate/regenerate call
  (upstream's generate is a 202 accepted, not a completed brief) until the
  brief reaches a terminal state, including when the card *loads* directly
  into an in-progress generation (page reload, nav-back, a co-chair's
  generation already running).
- Placed on committee-overview above My Pending Actions, gated by the
  `wg-weekly-brief` LaunchDarkly flag AND `canEdit()` (and, per the existing
  `!isVisitor()` wrapper it sits inside, committee membership).
- `[maxlength]` on the shared `lfx-textarea` wrapper (was already supported,
  nothing used it) bound to the same `brief_text` bound the BFF enforces.
- Playwright e2e covering flag-gating (with a positive control), empty state
  and its quota-exhausted disabled state, generated state, edit→save,
  generate→poll→terminal, load-into-an-in-progress-generation, and the
  read-failure → retry → recovery path.

## Why

Surfaces the weekly brief to chairs. Built on the BFF/contract from LFXV2-2175.
A failed read renders a distinct retryable "temporarily unavailable" state
rather than masquerading as "no brief yet".

## Review history

This branch went through 9 rounds of full-branch review
(`lfx-self-serve-code-reviewer` + `lfx-self-serve-learnings-reviewer`) during
development, each round's findings fixed and re-verified before the next.
Two classes of issue surfaced, both worth calling out explicitly:

**Real correctness bugs in the async generate/poll flow** — the original
implementation never polled after generate at all (parked on "Generating…"
forever); later increments fixed a cross-committee data leak on navigation,
an unbounded `exhaustMap` hang, a race on the poll's first tick, a
null-coalescing regression that could wipe an existing brief on Regenerate,
an inverted terminal-state check that could stop the poll mid-generation, an
`'empty'`-state brief rendering as if it were real content, a missing
`takeUntilDestroyed` guard, a per-poll `toObservable` effect leak, and a
409-on-save dead end. All fixed; see individual commit messages for detail.

**The e2e spec itself was broken since it was first written** — wrong route
(`/committees/:uid` instead of the real `/groups/:uid`), API mocks with the
wrong response envelope (bare arrays where the services unwrap `.data`), and
a missing `my_role` field on the committee fixture that kept the card from
ever rendering in *any* test, including the flag-ON positive control. None of
this was caught until round 7+, because static review only reads/type-checks
the spec — it doesn't execute it. **This spec has still not been executed
against a live dev server in this session** — see Known trade-offs.

## Known trade-offs

- **This e2e spec has never actually run.** Nine review rounds found and
  fixed three separate execution-blocking defects (route, envelopes, fixture)
  purely by static inspection against the real source files — a strong
  signal that inspection alone can't fully validate it. Before relying on
  this suite: confirm the `wg-weekly-brief` LaunchDarkly flag exists and is
  ON in the dev environment (it's new in this branch), then run
  `yarn e2e --grep "Weekly Brief"` for the first real execution.
- **A `TimeoutError` on the generate POST falls into the poll with no early
  Refresh escape** until that poll's own attempt cap trips (~80s). Dropping
  the timeout instead reintroduces the unbounded-hang dead end it was added
  to close, and there's no clearly better middle ground without a larger
  redesign of the generate/poll state machine.
- **A narrower edge case on the same path:** if a `TimeoutError` happens on
  Regenerate specifically (a terminal brief already on screen), the
  resulting poll's first tick can read back that pre-existing brief, treat
  it as "done", and complete silently — no toast, no indication whether the
  regenerate was actually accepted server-side.
- **`WeeklyBrief`/`WeeklyBriefThrottle` field-level optionality** — inherited
  from LFXV2-2175: upstream's Goa DSL marks nothing `Required` on these
  types, but the shared interfaces keep them fully required. Low-stakes here
  (the BFF guards every read with `?.`), deferred to when this branch's own
  rebase onto latest 2175 already landed.
- **Missing structural `weekly-brief-card-robust.spec.ts`** — this repo's
  e2e convention pairs every content spec with a structural one asserting
  the `data-testid` contract independent of copy.
- **No `beforeEach` hoisting** in the e2e spec — each test repeats its own
  mock/goto/auth-check preamble.
- **`weekLabel` computed inline** rather than via a private initializer
  function, per `component-organization.md`'s guidance for non-trivial
  `computed()` logic.
- **No Angular component/service unit spec** for the new component —
  `apps/lfx-one` has no Angular test builder wired up (`vitest.config.ts` is
  scoped to `src/server/**`), so a `.component.spec.ts` here would never
  execute. The Playwright suite is the actual coverage vehicle.

## Testing

- `yarn lint`, `yarn format`, `yarn test` (543 + 193 unit tests), `yarn build`,
  and `./check-headers.sh` all green.
- e2e spec type-checks clean (`tsc --noEmit` against `tsconfig.e2e.json`) and
  lists all 27 test/browser combinations via `playwright test --list` — see
  the "never executed" trade-off above for what this does and doesn't prove.

## Dependencies & related

- Depends on LFXV2-2175 (#1277) — must merge first. This PR targets the
  `feat/LFXV2-2175-wg-brief-foundation` branch as a stacked PR; retarget to
  `main` once #1277 merges.
- Backend blockers for real briefs to render end-to-end:
  - LFXV2-2190 — committee-service GET current 503 (stale-pod KV bind).
  - LFXV2-2193 — brief stuck in "generating", no brief_text produced.
- Supersedes #778 (already closed; comment posted pointing at this PR).

LFXV2-2176


apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+        // Upstream's generate call is a 202 accepted, not a completed brief — the
+        // actual generation runs out-of-band. Render the 202 body's `generating` state
+        // immediately (it's already in hand — no reason to wait a full poll interval
+        // and risk reading back the still-terminal pre-generate brief instead), then
+        // poll GET /current until it lands on a terminal state.
+        next: (res) => {
+          // Upstream marks nothing Required on the 202 envelope — a bare 202 with no
+          // brief/throttle must not wipe what's already rendered (most visible on
+          // Regenerate, where a real brief is on screen when this fires).
+          this.briefResponse.set({ brief: res.brief ?? this.brief(), throttle: res.throttle ?? this.throttle() });
```
dealako, line 191: [blocking] Generate response can overwrite a different committee's brief after cross-committee navigation

Issue: This `next` callback unconditionally calls `this.briefResponse.set(...)` using the `committeeUid` captured when `onGenerate()` was first invoked, with no check that the component is still displaying that same committee.

Proof: `committee` is an `input.required<Committee>()` that can change without destroying this component instance — the file's own comment at the top of `pollUntilTerminal` confirms Angular's `RouteReuseStrategy` keeps this component alive across a committee navigation. `pollUntilTerminal` accounts for this with `takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid)))` (line 398), but this earlier `.subscribe({ next: ... })` block (the POST response handler) has no equivalent guard — only `timeout`/`take(1)`/`takeUntilDestroyed`. If a user clicks Generate on committee A, then navigates to committee B before the POST resolves, this callback fires after B's brief has already loaded and overwrites `briefResponse` with A's stale (`generating`) data, briefly flashing B's card into a spurious "Generating…" state (`generating.set(true)` inside the subsequent `pollUntilTerminal(committeeUid)` call, using A's uid) before `pollUntilTerminal`'s own guard cancels the poll a tick later.

Why it matters: This is the same class of bug the PR description says was already found and fixed once during review ("a cross-committee data leak on navigation") — just on a different code path. A co-chair navigating quickly between committees while a generate is in flight sees another committee's transient state bleed onto the current card.

Fix: Guard this callback the same way `pollUntilTerminal` does — compare `committeeUid` against `this.committee()?.uid` before calling `this.briefResponse.set(...)`, or add `takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid)))` to this pipe alongside the existing `timeout`/`take(1)`/`takeUntilDestroyed`.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+              console.error('[weekly-brief-card] failed to load current brief', err);
+              this.fetchError.set(true);
+              return of(null as WeeklyBriefCurrentResponse | null);
+            }),
+            finalize(() => this.fetchLoading.set(false))
+          );
+        }),
+        takeUntilDestroyed(this.destroyRef)
+      )
+      .subscribe((response) => {
```
dealako, line 360: [blocking] `editMode`/edit form not reset when the committee changes — can save stale text under the new committee's revision

Issue: When the committee-uid pipeline re-fetches on navigation (this `subscribe` callback), it sets `briefResponse` and conditionally starts a poll, but never resets `editMode` or clears `editForm`.

Proof: `editMode()` gates the edit-textarea render generically in the template (`weekly-brief-card.component.html` line 70) — it isn't scoped to a particular committee. `onSave()` (line 224) reads `committeeUid` and `current = this.brief()` fresh at save time (both correctly reflect the *new*, current committee after navigation), but `text = this.editForm.controls.briefText.value` still holds whatever was typed for the *previous* committee, since nothing clears the form on the uid change. Sequence: open committee A → Edit → type/leave text in the form → navigate to committee B (component reused per the `RouteReuseStrategy` note elsewhere in this file) → `editMode` is still `true` and the form still shows A's text rendered against B's data → click Save → `saveWeeklyBrief(B_uid, { brief_text: A's stale text, revision: B's revision })` silently overwrites committee B's brief with committee A's content.

Why it matters: This is a silent data-corruption path, not just a UI glitch — a chair could overwrite another committee's brief with unrelated text with no error or confirmation step in between.

Fix: In this subscribe callback (or via a `distinctUntilChanged`-gated effect on committee uid), reset `this.editMode.set(false)` and `this.editForm.reset()` (or `.controls.briefText.setValue('')`) whenever the committee uid actually changes — mirroring how `fetchLoading`/`fetchError`/`pollTimedOut` are already reset per-uid inside the `switchMap` above.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+            // a normal 202, rather than toasting an error and leaving the card able to
+            // spend a second generate/regenerate on what might already be in progress.
+            // The poll's own attempt cap (pollTimedOut) bounds this if nothing comes back.
+            this.pollUntilTerminal(committeeUid);
+            return;
+          }
+          this.generating.set(false);
+          let detail: string;
+          switch (err?.status) {
+            case 429:
```
dealako, line 211: [minor] `429` (quota exceeded) doesn't refresh throttle state, unlike the `409` branch

Issue: The `409` branch calls `this.refresh$.next()` so `throttle`/`canGenerate`/`canRegenerate` get re-fetched and stay in sync with the server. The `429` branch only sets a toast `detail` and falls through — no refresh.

Proof: Compare this `case 429:` block (lines 200-203) to `case 409:` (lines 204-210) — only the latter calls `this.refresh$.next()`. `canGenerate`/`canRegenerate` (lines 115-123) are derived purely from the last-fetched `throttle` signal, which a 429 response never updates.

Why it matters: If the client's cached throttle counters are stale relative to the server (e.g. another chair just used the last generate/regenerate slot), the button can remain enabled after a 429 and keep producing the same quota error on every click, with no way to self-correct short of a manual page reload.

Fix: Add `this.refresh$.next();` inside the `case 429:` branch, same as `409`, so a quota-exhausted response also triggers a re-fetch of the authoritative throttle state.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.html
```diff
@@ -0,0 +1,141 @@
...
+          <!-- lfx-textarea forwards dataTest as [attr.data-test] onto the inner <textarea>
+               itself (unlike a data-testid on the <lfx-textarea> host, which wouldn't reach
+               it) — matches every other lfx-textarea call site in the repo. -->
+          <lfx-textarea
+            [form]="editForm"
+            control="briefText"
+            [rows]="8"
+            [maxlength]="briefTextMaxLength"
+            styleClass="w-full h-48 text-sm"
+            dataTest="weekly-brief-card-edit-textarea" />
```
dealako, line 91: [minor] Edit textarea has no accessible name

Issue: `lfx-textarea` is rendered here with no `id`, `aria-label`, `aria-labelledby`, or associated `<label>`, and its underlying template (`textarea.component.html`) doesn't synthesize one — it only forwards `id()`/`placeholder()` if the caller supplies them, and no `label` input exists on the wrapper at all.

Proof: `shared/components/textarea/textarea.component.html` line 4-16 renders a bare `<textarea>` bound to `[id]="id()"` — this call site never passes `id`, and there's no sibling `<label for=...>` in `weekly-brief-card.component.html`.

Why it matters: Screen-reader users navigating this form get an unlabeled `textarea` with no indication of what it edits, since there's no visible field label preceding it in this card (only the section heading "AI Assistant" above, which isn't programmatically associated with the control).

Fix: Add a visually-hidden or visible `<label>` associated via `id`/`for` (or pass an `aria-label` through if the wrapper is extended to support one), e.g. `<label for="weekly-brief-edit" class="sr-only">Edit weekly brief</label>` paired with `[id]="'weekly-brief-edit'"` on the `lfx-textarea`.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+  // every single generate/regenerate/load-into-generating call, not just once.
+  private readonly committee$ = toObservable(this.committee);
+
+  // Guards against starting a second concurrent poll — pollUntilTerminal is reachable
+  // both from onGenerate and from the initial-load pipeline (a page load / navigation
+  // landing on an already-`generating` brief).
+  private pollActive = false;
+
+  // Derived signals
+  public readonly brief: Signal<WeeklyBrief | null> = computed(() => this.briefResponse()?.brief ?? null);
```
dealako, line 104: [question] Sparse `WeeklyBrief`/`WeeklyBriefThrottle` payloads aren't modeled as optional

Issue: Copilot flagged that upstream's Goa contract marks nothing `Required` on `WeeklyBrief`/`WeeklyBriefThrottle`, but `brief`/`throttle` here are typed and consumed as fully populated objects (e.g. `weekLabel()` at line 125 calls `new Date(b.window_start)` unconditionally).

Proof: Your PR description already discloses this exact gap under "Known trade-offs" — inherited from LFXV2-2175, deferred to a rebase onto that branch once it lands, and called low-stakes since the BFF guards reads with `?.`.

Why it matters / fix: Agreeing with the bot's technical read here (a legal sparse response could render `Invalid Date` via `weekLabel()`, or leave `revision` undefined), but since this is explicitly disclosed and cross-repo-dependent, I'm leaving it as a question rather than blocking: is the LFXV2-2175 rebase already scheduled, or should this PR add defensive optionality now rather than wait?

dealako (changes requested):
Hey @manishdixitlfx — nice work on this one, and the PR description itself is one of the more thorough I've seen (calling out the 9 review rounds and the specific async bugs each one closed is genuinely useful context for whoever merges this). The state machine for generate/poll is careful about the failure modes it documents, and the e2e coverage is broad even though it hasn't run live yet.

I ran the full independent review (security, correctness/perf/tests, style/convention) plus reconciled against Copilot's and Cursor's existing comments. Two of Copilot's findings point at a real, reachable class of bug this branch has already fixed once elsewhere — a stale cross-committee write — that isn't fully closed for the `onGenerate` path or for `editMode`. Full comments below; also flagging where I disagree with or am only partially escalating a bot finding.

### 🔴 Blocking: 2
- Generate/regenerate response can overwrite a different (now-current) committee's brief after cross-committee navigation, briefly rendering "Generating…" for the wrong committee (agrees with Copilot).
- `editMode`/the edit form aren't reset on committee change, so navigating away mid-edit and back can save stale text under the new committee's revision (agrees with Copilot).

### 🟡 Minor: 2
- A `429` (quota exceeded) on generate/regenerate doesn't refresh throttle state the way `409` does, so the action can stay enabled and keep re-erroring (agrees with Cursor).
- The edit textarea has no accessible name — no label, `aria-label`, or `id` wired through (agrees with Copilot).

### ⚪ Nit: 0 (see notes)
### ❔ Question: 1
- Sparse `WeeklyBrief`/`WeeklyBriefThrottle` payloads (nothing `Required` upstream) aren't modeled as optional here, so a legal sparse response could render an invalid date or undefined revision (Copilot raised this too).

**On the two bot findings I'm not elevating**: `weekLabel` computed inline instead of via a private initializer (Copilot) and the optional-fields gap above are both already called out explicitly in your PR description as accepted trade-offs (the former deferred as low-priority, the latter deferred to a rebase onto LFXV2-2175). I agree with the bot's read on both — leaving the optional-fields one as a question rather than blocking since it's genuinely deferred cross-repo work, but the two navigation-state bugs above are new enough (and severe enough — silent data overwrite) that I don't think they can wait for the same deferral.

**Final decision: 🔴 Needs changes before approval** — the two blocking findings are both instances of the exact bug class this branch has spent 9 rounds hardening against (cross-committee state bleed on navigation), so I'd rather see them closed now than merged and rediscovered in production.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Done — guarded the generate/regenerate POST response with the same `takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid)))` guard `pollUntilTerminal` already uses, so a response for a committee the user has since navigated away from can no longer overwrite `briefResponse`. Also added a reset of `generating`/`editMode`/`editForm` on committee-uid change so a leftover state from a cut-off generate call doesn't bleed onto the new committee's card (covers the same root cause as the editMode thread below).

See commit a7adff5d0: weekly-brief-card.component.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+        // Bounds the pre-poll window: without this, a stalled POST leaves generating()
+        // true and pollTimedOut() false indefinitely — the same dead end pollUntilTerminal
+        // exists to close, just before the poll itself ever starts.
+        timeout(WEEKLY_BRIEF_POLL_INTERVAL_MS * 2),
+        take(1),
+        // The component can be destroyed while this request is still in flight (click
+        // Generate, then navigate away) — without this, `next` still runs against a
+        // destroyed component: it writes to signals nothing reads anymore and starts a
+        // poll (pollUntilTerminal) with nothing left to stop it early via destroy.
+        takeUntilDestroyed(this.destroyRef)
```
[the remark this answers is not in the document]
manishdixitlfx: Agreed and fixed — same root cause as `@dealako`'s finding on the generate-overwrite thread. Added `takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid)))` to the generate POST's pipe (matching `pollUntilTerminal`'s existing guard), so a stale response can't overwrite the new committee's state, and reset `generating` on uid change so a cut-off call doesn't leave a stale 'Generating…' flag on the new committee.

See commit a7adff5d0: weekly-brief-card.component.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Done — `initBriefResponseSubscription` now subscribes to `committeeUid$` directly and resets `generating`, `editMode`, and `editForm` whenever the uid actually changes, before the new committee's brief fetch resolves. A Save can no longer write stale text typed for a previous committee under the new committee's revision.

See commit a7adff5d0: weekly-brief-card.component.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+
+  // Private initializer functions
+  private initBriefResponseSubscription(): void {
+    const committeeUid$ = this.committee$.pipe(
+      filter((c): c is Committee => !!c?.uid),
+      map((c) => c.uid),
+      // A refresh (e.g. joining/leaving, a description save) re-emits a new Committee
+      // object with the same uid — skip the redundant brief round-trip when the id
+      // itself hasn't changed (matches committee-view.component.ts's initUpcomingMeetings).
+      distinctUntilChanged()
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — the committee-uid pipeline now resets `editMode` and `editForm` (plus `generating`) whenever the uid changes, so a stale edit from the previous committee can't render against the new committee's brief.

See commit a7adff5d0: weekly-brief-card.component.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Fixed — added `this.refresh$.next();` to the `429` branch, same as `409`, so a quota-exhausted response now triggers a re-fetch of the authoritative throttle state instead of leaving the client's stale count in place. Good catch!

See commit a7adff5d0: weekly-brief-card.component.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+            // The poll's own attempt cap (pollTimedOut) bounds this if nothing comes back.
+            this.pollUntilTerminal(committeeUid);
+            return;
+          }
+          this.generating.set(false);
+          let detail: string;
+          switch (err?.status) {
+            case 429:
+              detail = currentBrief ? 'Weekly regeneration limit reached. Try again next week.' : 'Weekly generation limit reached. Try again next week.';
+              break;
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — the `429` branch now calls `refresh$.next()` the same way `409` does.

See commit a7adff5d0: weekly-brief-card.component.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.html
manishdixitlfx: Fixed — added a visually-hidden `<label for="weekly-brief-card-edit-textarea">Edit weekly brief</label>` paired with `id="weekly-brief-card-edit-textarea"` on the `lfx-textarea` (the wrapper already forwards `id()` onto the inner `<textarea>`). No other `lfx-textarea` call site in the repo does this today (they rely on placeholder text alone) — this is the more correct pattern, not an existing convention I'm deviating from.

See commit a7adff5d0: weekly-brief-card.component.html.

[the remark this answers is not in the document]
manishdixitlfx: Fixed — added an `sr-only` `<label>` associated via `for`/`id` on the `lfx-textarea`.

See commit a7adff5d0: weekly-brief-card.component.html.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Confirming the status here: the LFXV2-2175 rebase this was deferred to has landed — #1279 is stacked directly on the current tip of #1277 (feat/LFXV2-2175-wg-brief-foundation), already including its post-review-fix commits.

That said, the gap itself is still open even after the rebase: LFXV2-2175's own review pass only loosened `WeeklyBriefSourceRef.title`/`.excerpt` to optional — `WeeklyBrief`/`WeeklyBriefThrottle` themselves are still typed as fully required in the shared interface. So this component's assumption of a fully-populated object still matches the current shared contract; it's not stale, but the underlying upstream-sparse-payload risk you and Copilot flagged is real and still undefended.

Leaving this open rather than resolving — properly closing it means widening WeeklyBrief/WeeklyBriefThrottle per the actual Goa contract and reviewing every read site in both this component and the two BFF services, which felt like more surface than this review-response iteration should take on unilaterally. Your call on whether that's worth doing now vs. a follow-up ticket.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+  private readonly committee$ = toObservable(this.committee);
+
+  // Guards against starting a second concurrent poll — pollUntilTerminal is reachable
+  // both from onGenerate and from the initial-load pipeline (a page load / navigation
+  // landing on an already-`generating` brief).
+  private pollActive = false;
+
+  // Derived signals
+  public readonly brief: Signal<WeeklyBrief | null> = computed(() => this.briefResponse()?.brief ?? null);
+  public readonly throttle: Signal<WeeklyBriefThrottle | null> = computed(() => this.briefResponse()?.throttle ?? null);
```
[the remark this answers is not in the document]
manishdixitlfx: Agreed this is a real risk (sparse payload could render Invalid Date via weekLabel(), etc.), tracking it alongside the same finding `@dealako` raised — see that thread for the current status. Leaving this open rather than resolving; not fixed in this iteration.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,428 @@
...
+  public readonly canRegenerate: Signal<boolean> = computed(() => {
+    const t = this.throttle();
+    return !t || t.regenerations_used < t.regenerations_limit;
+  });
+
+  public readonly weekLabel: Signal<string> = computed(() => {
+    const b = this.brief();
+    if (!b) return '';
+    // window_start / window_end are UTC ISO boundaries (Sun 00:00Z → Sat
+    // 23:59Z). Format with timeZone: 'UTC' so users in negative offsets
```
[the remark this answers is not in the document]
manishdixitlfx: No change made here — this is already an accepted trade-off documented in the PR description ("weekLabel computed inline rather than via a private initializer function, per component-organization.md's guidance"), and `@dealako`'s review explicitly declined to elevate it ("already called out explicitly in your PR description as accepted trade-offs... deferred as low-priority"). Leaving as-is.

manishdixitlfx:
## Review Feedback Addressed

Commit: a7adff5d03375e6ecd050c6073dd7526931dbb4a

### Changes Made
- **weekly-brief-card.component.ts**: guarded the generate/regenerate POST response with the same `committee$`/`takeUntil` guard `pollUntilTerminal` already uses, so a stale response can no longer overwrite a different committee's `briefResponse` after cross-committee navigation (per copilot-pull-request-reviewer, elevated as blocking by @dealako)
- **weekly-brief-card.component.ts**: reset `generating`/`editMode`/`editForm` whenever the committee uid changes, so a leftover "Generating…" flag or stale edited text can't bleed onto the new committee's card, and Save can no longer write stale text under the wrong committee's revision (per copilot-pull-request-reviewer, elevated as blocking by @dealako)
- **weekly-brief-card.component.ts**: `429` now calls `refresh$.next()` the same way `409` does, so a stale throttle count can't leave Generate/Regenerate enabled after a quota error (per cursor, agreed by @dealako)
- **weekly-brief-card.component.html**: added an `sr-only` label associated with the edit textarea's `id` (per copilot-pull-request-reviewer, agreed by @dealako)

### Questions Answered
- **Sparse `WeeklyBrief`/`WeeklyBriefThrottle` optionality** (asked by @dealako): confirmed the LFXV2-2175 rebase this was deferred to has landed (#1279 is stacked on #1277's current tip), but the gap itself is still open — LFXV2-2175's review only loosened `WeeklyBriefSourceRef`, not `WeeklyBrief`/`WeeklyBriefThrottle` themselves. Left open pending your direction on now-vs-follow-up-ticket.

### No Change Needed
- **`weekLabel()` computed inline** (flagged by copilot-pull-request-reviewer): already an accepted trade-off documented in the PR description; @dealako's review explicitly declined to elevate it. Resolved as settled, no action.

### Threads Resolved
9 of 11 unresolved threads addressed in this iteration.

### Still Open
- **Sparse-payload optionality** (2 threads, component.ts:104-105): awaiting your call on scope — fix now vs. track as a follow-up ticket.

dealako (changes requested):
Hey @manishdixitlfx — thanks for the fast turnaround. Both of my blocking findings are cleanly closed, and closed exactly the way I suggested (reusing `pollUntilTerminal`'s own `committee$` `takeUntil` guard for the generate-response handler, and resetting `editMode`/`editForm` keyed on the *actual* uid change via `distinctUntilChanged`, not on every refresh — that second detail matters, since a naive reset-on-every-emission would've introduced a new bug where an unrelated refresh wipes out someone's in-progress edit).

### 👏 Nice work
- The `takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid)))` guard on the generate/regenerate POST response is placed correctly in the pipe (before `.subscribe`), so a stale response genuinely can't write into the new committee's `briefResponse` — not just a downstream guard that fires one tick too late.
- The `editMode`/`editForm` reset is subscribed off `committeeUid$` (which carries `distinctUntilChanged()`), so it only fires on a real committee change, not on every `refresh$` tick — correctly preserves an in-progress edit if e.g. a 409 refresh happens on the same committee.
- Both minor findings (429 not refreshing throttle, textarea accessible name) are fixed, and the `WeeklyBriefSourceRef.title`/`.excerpt` optionality narrows the sparse-payload question in the direction I asked about.

### 🔴 Blocking: 3

**1. `generateBrief` missing the `assertCommitteeRead` gate just added to `getCurrentBrief`** (`weekly-brief.controller.ts:153`)
Issue: This endpoint calls straight into `weeklyBriefService.generateBrief` with only UID-format and body validation — no committee-membership check.
Proof: `getCurrentBrief` now calls `await assertCommitteeRead(req, committeeId, 'get_weekly_brief_current')` before its service call, added specifically because "the service layer proxies straight through to committee-service in live mode with no authorization of its own." That description applies identically to `generateBrief` — it forwards straight to committee-service with the user's bearer token and no in-app check. Any authenticated caller who knows a committee UID can trigger a real (quota-consuming) generation for a committee they don't belong to.
Why it matters: This is a write-side action with a real cost (AI generation quota) an unauthorized caller can trigger against any committee.
Fix: Add `await assertCommitteeRead(req, committeeId, 'generate_weekly_brief');` right after body validation (line 151) and before the service call, mirroring `getCurrentBrief`, plus a controller-spec case matching the new `getCurrentBrief` 403-propagation test.

**2. `saveBrief` missing the same gate** (`weekly-brief.controller.ts:209`)
Issue/Proof: Identical gap — `WeeklyBriefService.saveBrief` proxies straight to committee-service's `PUT /committees/:id/weekly-briefs/current` with no caller-side authorization.
Why it matters: An authenticated caller who knows any committee UID could overwrite that committee's brief text — a silent, unauthorized content-integrity issue.
Fix: Add `await assertCommitteeRead(req, committeeId, 'save_weekly_brief');` after body validation (line 207) and before the service call (or a write-specific gate if one exists elsewhere in the codebase), plus matching spec coverage.

**3. Poll-timeout path silently drops the card to "No brief yet" instead of surfacing the timeout escape hatch** (`weekly-brief-card.component.ts:397`, agrees with Cursor Bugbot)
Issue: The poll's `tap` (line 397) writes every polled response — including a non-terminal `null`/`empty` tick — into `briefResponse` before the attempt cap trips. When the cap is hit, `finalize` sets `generating` false; since `brief()?.state` now holds that last non-terminal read (not `'generating'`), the template's `generating() || brief()?.state === 'generating'` branch (html:14) stops matching, so the `pollTimedOut` message/Refresh button never has a stable render — the card falls through to the empty-state branch (html:129) instead, with Generate re-enabled.
Why it matters: The user has no indication a generation is still possibly running server-side after quota was already spent — the UI implies nothing happened and invites a second Generate click.
Fix: Don't let a non-terminal poll response overwrite `briefResponse` with `brief: null`/`state: empty` — skip the `.set(...)` for non-terminal ticks (only write once a terminal state is observed or the cap trips), or key the generating-branch template condition off `pollActive`/`pollTimedOut` directly rather than solely off the last-seen `brief()?.state`.

### 🟡 Minor: 1

**First poll tick after regenerate could read back the stale pre-regenerate terminal state** (`weekly-brief-card.component.ts:192`, agrees with Cursor Bugbot)
Issue: `onGenerate`'s 202 handler renders the local `generating` body immediately, then starts `pollUntilTerminal`. If the very first `GET /current` tick lands before the backend's write is visible, it could return the *previous* (still-terminal) brief, and `takeWhile` has no way to tell that terminal state is stale rather than the new generation's result.
Proof: The poll's `takeWhile` only checks `!response.brief || !TERMINAL_STATES.has(state)` — no comparison against the pre-regenerate brief's revision, unlike the deliberate handling already in place for a `null`/empty read (explicitly treated as write-lag, not terminal).
Why it matters: A regenerate could silently "complete" showing unchanged content while quota was consumed and the real new brief is still generating — lower severity than the finding above since it depends on upstream read-after-write timing that may or may not be a realistic gap.
Fix: Capture the pre-regenerate brief's `revision`/`updated_at` before calling generate, and require the polled response's revision to differ before treating it as the new terminal result.

### On a fourth Cursor Bugbot finding (disagreeing)
Cursor's "stale brief during committee switch" finding says `briefResponse` isn't cleared on committee change and the card can flash the old committee's brief. I don't think this is reachable: `fetchLoading()` is checked *before* `renderableBrief()`/`brief()` in the template's `@if` chain, and `fetchLoading.set(true)` happens synchronously inside `switchMap`'s project function on the same synchronous turn as the uid change (RxJS `switchMap` unsubscribes the old inner observable — running `finalize`'s `fetchLoading.set(false)` — then immediately re-subscribes the new one, setting it back to `true`, all before Angular's next change-detection pass). So the skeleton loader should cover the transition with no stale-brief frame. Flagging as disagreement rather than dropping silently, since I could be missing a scheduling edge case — worth a quick manual check if you want a second pair of eyes on it.

### Final decision: 🔴 Needs changes before approval — the two new authorization gaps are the same class of issue just fixed for the read endpoint, and the poll-timeout state loss silently discards a spent-quota generation with no way for the user to tell what happened.

manishdixitlfx:
## Round 2 Review Feedback Addressed

Commit: ef28f13b79ace312fb844d3e944c7f8efac6a7c8 (rebased onto #1277's round-2 fix at 81751957b, plus two new fixes)

### Changes Made
- **Write-path authorization gap** (`generateBrief`/`saveBrief` missing `assertCommitteeRead`-equivalent gate): inherited via the rebase onto #1277's `81751957b`, which added `assertCommitteeWrite` (`committee#writer`) to both endpoints in `weekly-brief.controller.ts` — see that PR's round-2 summary for detail. No separate action needed on this branch beyond the rebase.
- **weekly-brief-card.component.html**: the generating-branch condition now also checks `pollTimedOut()`, so a poll that hits its attempt cap no longer silently drops the card to "No brief yet" — it correctly stays on the "still generating / Refresh" state instead of discarding a spent-quota generation with no indication anything happened
- **weekly-brief-card.component.ts** (`pollUntilTerminal`): now accepts an optional `priorRevision` (captured from the pre-regenerate brief), and rejects a terminal-looking tick whose revision still matches it — so a first poll tick landing before upstream's write is visible can't silently "complete" against the stale pre-regenerate brief. `error` state is trusted immediately regardless of revision (an unambiguous fresh signal, unlike generated/edited/approved)

### Acknowledged, No Action Needed
- Your disagreement with Cursor's "stale brief during committee switch" finding checks out — `fetchLoading()` gates ahead of `renderableBrief()`/`brief()` in the template, and `switchMap`'s project function sets `fetchLoading` synchronously on the same turn as the uid change, so the skeleton loader should cover the transition with no stale-brief frame.

### Not Covered by an Automated Test
Same limitation as round 1 — `apps/lfx-one` has no Angular test builder wired up, so the poll-timeout and stale-regenerate fixes are verified by code inspection only.

### Still Open
- Sparse-payload optionality question (2 threads) — unchanged from round 1, still awaiting your direction on now-vs-follow-up-ticket.

dealako (dismissed):
Hey @manishdixitlfx — this closes it out. All three blocking findings from round 2 are genuinely resolved, and the one minor is too.

### 👏 Nice work
- Rebasing onto #1277's `assertCommitteeWrite` (`committee#writer`) cleanly closed both write-path authorization gaps (`generateBrief:157`, `saveBrief:218`) — verified the gate actually runs before the proxy call in both cases, not just present in the diff.
- The poll-timeout fix is exactly right: `pollTimedOut()` is now part of the generating-branch condition in the template (`weekly-brief-card.component.html:14`), so a poll that exhausts its attempt cap keeps the "still generating / Refresh" UI visible instead of silently falling through to the empty state — no more discarding a spent-quota generation with zero indication to the user.
- The `priorRevision` guard on `pollUntilTerminal` is a clean fix for the stale-regenerate race: a terminal-looking tick whose revision still matches the pre-regenerate baseline is correctly rejected as stale, while `error` is still trusted immediately (right call — it's an unambiguous fresh signal regardless of revision).

### 🟡 Minor: 2 (new, non-blocking)
- **No test coverage for the `priorRevision` stale-read guard.** This is a genuinely tricky piece of logic (three states: fresh generate with no `priorRevision`, stale regenerate tick to reject, real regenerate completion to accept) with zero test exercising any of the three paths — worth a follow-up unit or e2e case, but I'm not blocking merge on it given `apps/lfx-one` has no Angular component-test builder wired up yet (same limitation you noted for round 1/2).
- **`onSave` has no `takeUntil(committee$...)` guard, unlike `onGenerate`/`pollUntilTerminal`** (agrees with Cursor's new comment on `weekly-brief-card.component.ts:271`). Confirmed: `onSave`'s pipe is only `take(1)` + `takeUntilDestroyed` (line 250) — no guard against a save response arriving after the user has navigated to a different committee. Impact is narrower than the round-1/2 bugs though: the write itself still targets the correct committee (`committeeUid` is captured at click time), so there's no data corruption — but `next()`'s `editMode.set(false)` / `refresh$.next()` would fire against whatever committee is now showing, which could silently kick a different committee's in-progress edit out of edit mode. Worth the same `takeUntil` guard for consistency, but lower severity than the round-1/2 findings since no data is actually misdirected.

### On the second new Cursor comment (lower-confidence, not escalating)
Cursor's other new comment flags a possible race in the e2e read-failure-retry test (`e2e/weekly-brief-card.spec.ts:557`) around call-count assumptions on remount. Looked at the test directly — it's a single page load with sequential route interception, not a remount scenario, so I don't see the specific race Cursor describes in this test as written. Not escalating without a clearer reproduction, but flagging in case you want a second look.

### On the still-open sparse-payload question
Agreed this is fine to keep deferred — it's cross-repo (LFXV2-2175), already disclosed in the PR description, and the BFF's `?.` guards keep the blast radius small in the meantime. No new information changes that calculus this round.

### Final decision: ✅ Approved with minor comments — no blocking issues remain; the two minors above are good follow-up work but shouldn't hold up merge.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,468 @@
...
+    if (!committeeUid || !current) return;
+    const text = this.editForm.controls.briefText.value.trim();
+    if (!text) {
+      this.messageService.add({ severity: 'warn', summary: 'Empty brief', detail: 'Brief text cannot be empty.' });
+      return;
+    }
+    this.saving.set(true);
+    this.weeklyBriefService
+      .saveWeeklyBrief(committeeUid, { brief_text: text, revision: current.revision })
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
```
dealako: [blocking] `onSave` missing the `takeUntil(committee$)` navigation guard

Issue: `onSave`'s pipe only has `take(1), takeUntilDestroyed(this.destroyRef)` — unlike `onGenerate` (line 174) and `pollUntilTerminal` (line 438), it has no `takeUntil(this.committee$.pipe(filter(c => c?.uid !== committeeUid)))` guard. With Angular's RouteReuseStrategy keeping this component alive across a committee navigation, a save started on committee A that's still in flight when the user navigates to committee B will still land: its `next`/`error` handler runs against committee B's card, calling `refresh$.next()` (re-fetching B's brief under A's save context) and flipping `saving` to false on B's state.

Proof: compare this pipe (line 250) to `onGenerate`'s (line 174) and `pollUntilTerminal`'s (line 438), both of which carry the guard with an explicit comment explaining why it's needed. `onSave` has no equivalent. Compounding this: the uid-change reset in `initBriefResponseSubscription` (lines 326-330) resets `generating`, `editMode`, and `editForm` on navigation, but not `saving` — so a save left in flight from the old committee can leave `saving` stuck `true` on the new committee's card with nothing to clear it until the stale response arrives.

Why it matters: cross-committee state leak — a save on committee A can silently mutate committee B's toast/refresh state, and the Save button on B can appear disabled/spinning until A's stale response resolves.

Fix: add the same guard used by `onGenerate`/`pollUntilTerminal`:
```ts
.pipe(
  take(1),
  takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid))),
)
```
and add `this.saving.set(false);` to the uid-change reset block (lines 326-330).

Note: this is the same issue already flagged by cursor[bot] as "Save lacks committee navigation guard" (unresolved review thread on this PR) — it slipped through despite two prior rounds of review fixes. Confirmed independently against the current HEAD (`c18fc92c`).

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,468 @@
...
+    const isNewTerminal = (response: WeeklyBriefCurrentResponse): boolean => {
+      const b = response.brief;
+      if (!b || !WEEKLY_BRIEF_TERMINAL_STATES.has(b.state)) return false;
+      // 'error' is always a fresh, unambiguous signal — trust it immediately regardless of
+      // revision, unlike generated/edited/approved, which a stale read could report on the
+      // pre-regenerate brief without actually bumping revision.
+      if (b.state === 'error') return true;
+      return priorRevision === undefined || b.revision !== priorRevision;
+    };
+    timer(WEEKLY_BRIEF_POLL_INTERVAL_MS, WEEKLY_BRIEF_POLL_INTERVAL_MS)
```
dealako, line 406: [minor] Poll timer runs unguarded during SSR

Issue: `pollUntilTerminal`'s `timer(WEEKLY_BRIEF_POLL_INTERVAL_MS, WEEKLY_BRIEF_POLL_INTERVAL_MS)` isn't gated by `isPlatformBrowser`. `platformId` is injected (line 55) but only used for the clipboard guard in `onCopyAndShare` (line 278) — not here. A component that loads directly into the `generating` state (line 358) calls `pollUntilTerminal` from `initBriefResponseSubscription`, which runs during SSR too, so the timer/poll loop starts server-side and can issue several GET requests before the SSR render serializes.

Proof: `grep -n isPlatformBrowser` in this file shows the only guarded call site is line 278; the `timer(...)` at line 390 and its containing `pollUntilTerminal` (line 374) have no platform check.

Why it matters: unnecessary backend load on every SSR render of a card that loads into `generating` state, and wasted poll attempts that never reach the browser-side UI.

Fix: guard the initial `pollUntilTerminal(uid)` call at line 359 (and/or the top of `pollUntilTerminal` itself) with `if (!isPlatformBrowser(this.platformId)) return;`.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,468 @@
...
+              break;
+            default:
+              detail = 'Failed to generate brief. Please try again.';
+          }
+          this.messageService.add({ severity: 'error', summary: 'Generate failed', detail });
+        },
+      });
+  }
+
+  public onEdit(): void {
```
dealako, line 233: [minor] Narrow race: re-entering edit mode before a 409's `refresh$` resolves can reuse a stale revision

Issue: on a 409 during Regenerate (`onGenerate`, lines 218-224) or Save (`onSave`, lines 260-265), `refresh$.next()` re-fetches the brief asynchronously, but `editMode` is set `false` immediately (Save's 409 path) — or was never set on Regenerate — before that fetch resolves. If the user clicks Edit inside that window, `onEdit` (this line) reads `this.brief()?.brief_text` off the pre-conflict signal, so a subsequent Save reuses the same stale `revision` and can 409 again.

Proof: `onEdit` reads `this.brief()` directly with no check against an in-flight refresh; `refresh$` consumers (`combineLatest([committeeUid$, this.refresh$])`, line 331) only flip `fetchLoading` — nothing disables the Edit action while a refresh triggered by a 409 is in flight.

Why it matters: narrow window (requires the user to react within a network round-trip after a conflict toast), but the failure mode is a repeat 409 instead of a clean recovery.

Fix (optional / low priority given the narrow window): gate the Edit button on `fetchLoading()` as well as the existing `canEdit` input, or have the 409 handlers explicitly clear `briefResponse` so `onEdit` has nothing stale to read until the refresh lands.

dealako (changes requested):
Hey @manishdixitlfx — thanks for pushing this through two review rounds already; the fixes for the generate-overwrite guard, the uid-change edit/generating reset, the 429→refresh gap, and the textarea a11y label all check out cleanly against current HEAD (`c18fc92c`).

**Overall impression:** this is a well-thought-through feature — the inline comments throughout `weekly-brief-card.component.ts` do a lot of work explaining *why* each guard/race-condition fix exists, which made this review much easier to verify. The core generate/poll/regenerate state machine is solid and several genuinely subtle races (RouteReuseStrategy component reuse, exhaustMap vs switchMap for the poll, stale-revision detection on regenerate) are already handled well. One real gap remains: `onSave` never got the same navigation guard `onGenerate`/`pollUntilTerminal` received.

This review was run by three parallel subagents (security/privacy, correctness/performance/tests, style/docs) against the PR diff, then self-challenged against the current code before posting — two claimed findings (a poll-timeout empty-state fallthrough, and a "hangs forever" regenerate e2e test) turned out to already be fixed/not applicable and were dropped rather than posted.

**🔴 Blocking: 1 issue**
- `onSave` is missing the `takeUntil(committee$)` navigation guard that `onGenerate`/`pollUntilTerminal` both have, and the uid-change reset doesn't clear `saving`. This is the *same* issue cursor[bot] already flagged as "Save lacks committee navigation guard" — still open, unresolved after two rounds. Confirmed by direct read against current HEAD, independent of the bot's claim.

**🟡 Minor: 2 issues**
- `pollUntilTerminal`'s `timer(...)` isn't guarded by `isPlatformBrowser` — a card that loads directly into `generating` state will poll server-side during SSR.
- A narrow race where re-entering Edit before a 409's `refresh$` resolves can reuse a stale revision and re-409 on Save. Low priority given how tight the timing window is.

**⚪ Nit: 0 issues** (the `weekLabel` inline-computed-vs-private-initializer point was already explicitly discussed and accepted as a documented trade-off in this PR's own review history — not re-raising it.)

**❔ Question: 0 new items** — the sparse `WeeklyBrief`/`WeeklyBriefThrottle` payload-modeling question (Copilot / @dealako, line ~104-105) is still open and already tracked; nothing to add there.

**Also still open from earlier rounds** (not independently re-verified in this pass, flagging so they don't get lost): cursor[bot]'s "Stale brief during committee switch" (line 347), "Retry test races on remount" (e2e line 557), and CodeRabbit's SSR/change-detection note (line 361) have no reply/fix commit yet.

**Final decision: 🔴 Needs changes before approval** — the `onSave` guard is a real, previously-flagged, still-unresolved cross-committee state leak. Everything else here is minor/optional.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,468 @@
...
+        this.briefResponse.set(response);
+        // Covers a page reload, navigating back to this committee, or a co-chair's
+        // generation already in flight — not just this tab's own onGenerate() call.
+        // Without this, a card that *loads* into the generating state never polls and
+        // is stuck on the spinner with no way to reach a terminal state.
+        const uid = this.committee()?.uid;
+        if (uid && response?.brief?.state === 'generating') {
+          this.pollUntilTerminal(uid);
+        }
+      });
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — the load-time poll trigger now checks `isPlatformBrowser(this.platformId)` before calling `pollUntilTerminal`. The SSR-rendered HTML still correctly shows the generating state (the template reads `brief()?.state` directly, independent of this guard); the client re-runs this same subscribe on hydration and starts the real poll there.

See commit 86f483e0b: weekly-brief-card.component.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,468 @@
...
+    let observedTerminal = false;
+    const isNewTerminal = (response: WeeklyBriefCurrentResponse): boolean => {
+      const b = response.brief;
+      if (!b || !WEEKLY_BRIEF_TERMINAL_STATES.has(b.state)) return false;
+      // 'error' is always a fresh, unambiguous signal — trust it immediately regardless of
+      // revision, unlike generated/edited/approved, which a stale read could report on the
+      // pre-regenerate brief without actually bumping revision.
+      if (b.state === 'error') return true;
+      return priorRevision === undefined || b.revision !== priorRevision;
+    };
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed and fixed — you're right, the stateless mock's getCurrentBrief always reported the same hardcoded revision, so the round-2 priorRevision guard could never see a difference and a mock regenerate hung until the attempt cap. Added a mock-only, in-memory revision tracker (`weekly-brief.service.ts`): generateBrief bumps it on force:true, saveBrief keeps it in sync, getCurrentBrief reads it. Added a test-only reset (matching the repo's existing `__resetLinkedInConfigForTesting` convention) plus regression coverage for the bump, a second regenerate incrementing further, and a fresh generate NOT bumping. Good catch — this would've blocked local dev testing of Regenerate entirely.

See commit 86f483e0b: weekly-brief.service.ts, weekly-brief.service.spec.ts.

apps/lfx-one/e2e/weekly-brief-card.spec.ts
```diff
@@ -0,0 +1,579 @@
...
+          contentType: 'application/json',
+          body: JSON.stringify({ error: 'group-weekly-briefs bucket not initialized' }),
+        });
+        return;
+      }
+      await route.fulfill({
+        status: 200,
+        contentType: 'application/json',
+        body: JSON.stringify({ brief: null, throttle: DEFAULT_THROTTLE } as WeeklyBriefCurrentResponse),
+      });
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — switched to the same one-way-flag pattern the generate test above it already uses (with the exact reasoning noted there): the route handler now fails every GET until the test flips a `retried` boolean right before the explicit Retry click, so any number of incidental remount GETs before that point still land on the failure branch instead of racing the assertion.

See commit 86f483e0b: weekly-brief-card.spec.ts.

manishdixitlfx:
## Review Feedback Addressed

Commit: 86f483e0baf39e53c4fbd7bc5767c2b21b5c2f22

Also covers rebasing this branch onto `main` (LFXV2-2175/#1277 squash-merged) — CI and mergeability confirmed clean before this round's fixes.

### Changes Made
- **weekly-brief-card.component.ts**: guarded the load-time poll trigger with `isPlatformBrowser` — zoneless CD doesn't block SSR stability on a bare `timer()`, so without this, SSR could start (and never stop) a real poll loop server-side (per coderabbitai[bot])
- **weekly-brief.service.ts** (mock backend): added a mock-only revision tracker so a mock regenerate's subsequent poll ticks actually see the bumped revision the generate call promised — without this, the round-2 `priorRevision` guard made every mock regenerate hang until the poll's attempt cap (per cursor)
- **weekly-brief-card.spec.ts** (e2e): the read-failure/retry test's `calls === 1` indexing was fragile against a LaunchDarkly remount firing an extra GET — switched to the same one-way-flag pattern already used by the generate test above it (per cursor)
- Removed a stray "Generated with Claude Code" attribution line found at the end of the e2e spec — this repo's commit-workflow policy treats authorship as human-only

### Threads Resolved
3 of 5 unresolved threads addressed. The 2 sparse-payload-optionality threads remain open, unchanged, pending direction on now-vs-follow-up-ticket.

manishdixitlfx:
## Round 4: @dealako's Independent Review Addressed

Commit: 0cae116433d3a20e2b1daec9865ebbace5955b8e

Thanks for the thorough independent pass — appreciated the callout on what you dropped after self-challenging (poll-timeout empty-state fallthrough, "hangs forever" regenerate e2e test) as much as what you kept.

### Blocking — Fixed
- **`onSave` missing the navigation guard**: added the same `takeUntil(committee$)` guard `onGenerate`/`pollUntilTerminal` already have to `onSave`'s PUT pipe, and added `saving.set(false)` to the uid-change reset (alongside the existing `generating`/`editMode`/`editForm` resets) so a save cut off mid-flight can't leave a stale flag on the next committee's card.

### Minor — Fixed
- **`pollUntilTerminal`'s own `isPlatformBrowser` guard**: moved the SSR check from only the load-time caller into `pollUntilTerminal` itself, so it's the single choke point regardless of which call site reaches it.

### Minor — Not Fixed (per your own priority call)
- The narrow re-entering-Edit-before-409-refresh-resolves race — left as-is, tight timing window, your assessment stands.

### Already Closed (confirming, no new action)
- The three "still open from earlier rounds" items you flagged for visibility were already addressed before this review landed: the stale-brief-during-switch disagreement (confirmed correct in my earlier reply), the e2e retry-race, and the SSR/change-detection note — the latter two in this same round's prior commit (86f483e0b).

apps/lfx-one/e2e/weekly-brief-card.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+    // a fixed "call N is terminal" index would desync on exactly that kind of extra call.
+    let terminalObserved = false;
+    let getCount = 0;
+    await page.route(`**/api/committees/${TEST_COMMITTEE_UID}/weekly-briefs/current`, async (route) => {
+      if (route.request().method() !== 'GET') {
+        await route.fallback();
+        return;
+      }
+      getCount += 1;
+      if (getCount > 1) {
```
dealako: [minor] Load-into-generating test still count-gated, same flake risk just fixed elsewhere

Issue: This test still gates the mock's terminal-flip on `getCount > 1` (raw call index), the exact pattern that was just replaced with a `retried` boolean below (line ~552) to fix the LaunchDarkly-remount flake (Cursor Bugbot, round 2/3). An extra initial GET from a LaunchDarkly-triggered remount can consume this counter's "first two calls" budget before the test's own assertions run.

Proof: `let getCount = 0;` / `getCount += 1; if (getCount > 1) { ... }` — compare to the fixed retry test's `let retried = false; if (!retried) { ... }` a few dozen lines below in the same file.

Why it matters: this test is one incidental extra GET away from flaking the same way the retry test did before its fix — the fix just wasn't propagated to this test.

Fix: switch to a one-way boolean flag (e.g. `let reachedTerminal = false;` flipped once the poll has ticked the intended number of times), mirroring the pattern already applied to the retry test.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -180,10 +215,14 @@ export class WeeklyBriefService {
    */
   public async saveBrief(req: Request, committeeId: string, body: SaveWeeklyBriefRequest): Promise<WeeklyBrief> {
     if (!this.isLive(req)) {
+      // Keep the mock revision tracker in sync with the save, same as a regenerate does —
+      // otherwise a subsequent getCurrentBrief would report a stale (pre-save) revision.
+      const revision = body.revision + 1;
```
dealako: [minor] Mock saveBrief doesn't reject a stale revision before bumping the tracker

Issue: `const revision = body.revision + 1;` derives the next revision purely from the caller-supplied `body.revision`, without checking it against `currentMockRevision(committeeId)` first (CodeRabbit).

Proof: if a stale save (holding an older revision than the currently-tracked one) reaches this method, `mockRevisionByCommittee.set(committeeId, revision)` can set the tracker backward or to a value inconsistent with the last regenerate/save, undermining the invariant this round's revision tracker was built to guarantee (line 150 `mockRevisionByCommittee`).

Why it matters: mock-only, so no production risk, but it weakens the very guard `pollUntilTerminal`'s `priorRevision` check depends on for test realism — a stale-save race could resurface the same class of bug this round just fixed.

Fix: compare `body.revision` against `currentMockRevision(committeeId)` and reject (409, matching the live backend's conflict behavior) if they don't match, before bumping.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,484 @@
...
+    }
+    this.saving.set(true);
+    this.weeklyBriefService
+      .saveWeeklyBrief(committeeUid, { brief_text: text, revision: current.revision })
+      .pipe(
+        take(1),
+        // Same guard as onGenerate's POST pipe: a save started on committee A whose response
+        // arrives after the user has already navigated to committee B must not clear B's
+        // saving/editMode or trigger an unwanted refresh$ on B's card (dealako review, round 3).
+        takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid))),
```
dealako, line 255: [minor] Mock backend doesn't persist saved text, so a successful save visibly reverts in dev/mock mode

Issue: `onSave`'s success handler calls `this.refresh$.next()` immediately, which triggers a `getCurrentBrief` GET (Copilot).

Proof: mock `getCurrentBrief` (`weekly-brief.service.ts`) calls `buildMockBrief(committeeId, { revision: currentMockRevision(committeeId) })` — it never applies the saved `brief_text` or `state: 'edited'`, so it rebuilds a fresh generated sample. The revision number now correctly matches (this round's fix), but the visible text does not.

Why it matters: in local/mock dev mode, a successful save appears to silently revert to the original generated text — confusing during manual testing and any e2e assertions that check post-save text via a follow-up GET. Production is unaffected since the live microservice persists the save.

Fix: track the last-saved `{brief_text, state}` per committee alongside the existing `mockRevisionByCommittee` map (or fold it into a single per-committee mock-brief record) and have `getCurrentBrief` return it when present.

dealako (changes requested):
## Follow-up re-review — round 4 (`c18fc92c` → `86f483e0` → `0cae1164`)

@manishdixitlfx thanks for the quick turnaround across two more pushes — the blocking issue is now closed out and the SSR fix ended up stronger than what I originally asked for.

### 👏 Nice work

- **The blocking `onSave` navigation guard is fixed** (`0cae1164`) — matches `onGenerate`'s exact `takeUntil(this.committee$.pipe(filter(c => c?.uid !== committeeUid)))` pattern, and you also caught that the uid-change reset block needed to clear `saving`, which I hadn't called out explicitly but is clearly the same bug class.
- **The SSR guard is better than requested** — instead of gating only the load-time call site, you moved `isPlatformBrowser` into `pollUntilTerminal` itself as the single choke point (`0cae1164`). That's the right fix: it protects every current and future caller, not just the one call site I flagged.
- **Mock regenerate-poll hang (Cursor Bugbot) is fixed** — the in-memory `mockRevisionByCommittee` tracker in `weekly-brief.service.ts` is correctly keyed, bumped on `force:true`, read by `getCurrentBrief`, and kept in sync by `saveBrief`. The new spec assertions (bump-on-regenerate, second-regenerate-bumps-again, fresh-generate-doesn't-bump, save-keeps-in-sync) meaningfully exercise the guard the poll depends on, not just trivially pass.
- **E2E retry-race (Cursor Bugbot) is fixed** — switched from a raw call-index gate to the same one-way `retried` boolean the generate test already used, closing the LaunchDarkly-remount flake risk.
- Removed the stray Claude-attribution comment per repo convention.

### Revision tracking

✅ **Resolved**
- `[blocking]` `onSave` missing `takeUntil(committee$)` navigation guard — fixed, `saving` also now reset on committee change (`0cae1164`).
- `[minor]` SSR poll timer unguarded — fixed, guard moved into `pollUntilTerminal` as the shared choke point (`0cae1164`).
- Cursor: "Mock regenerate poll never completes" — fixed via mock revision tracker (`86f483e0`).
- Cursor: "Retry test races on remount" — fixed via `retried` boolean pattern (`86f483e0`).
- CodeRabbit: SSR/change-detection note on the load-time poll trigger — fixed, then further strengthened in round 4.
- My nit: stray `// Generated with [Claude Code]` comment — removed (`86f483e0`).

⚠️ **Still open / newly surfaced**
- `[minor]` Narrow 409-then-re-edit race (mine, and independently corroborated by Cursor's "Stale revision after conflict refresh") — explicitly deferred by the author as a tight-timing-window, low-priority item per the `0cae1164` commit message. Not blocking, but still open — no action taken.
- `[minor]` **New**: the "loads directly into the generating state" e2e test (around line 505) is still gated on a raw `getCount > 1` check, the same pattern that was just fixed in the retry test for the identical LaunchDarkly-remount flake risk. Cursor Bugbot flagged this independently — agreed, this should get the same `retried`-style boolean treatment as the fix already applied elsewhere in this file.
- `[minor]` **New**: `weekly-brief.service.ts`'s mock `saveBrief` (line ~220) derives the next revision as `body.revision + 1` without checking it against `currentMockRevision(committeeId)` first — a stale save (using an old revision) can still succeed and silently reset the tracked revision backward. CodeRabbit flagged this; agreed, this undermines the very guard this round's mock revision tracker was built to support. Since it's mock-only, low real-world risk, but worth a quick reject-on-mismatch check to keep the tracker's invariant honest.
- `[minor]` **New**: mock `getCurrentBrief` doesn't persist saved `brief_text`/`state: 'edited'` — after a successful `onSave`, the immediate `refresh$.next()` GET rebuilds a fresh generated sample rather than reflecting the save, so the edited text visibly reverts in the mock/dev environment (though the revision number now correctly stays in sync). Copilot flagged this; agreed it's real, though scoped to local/mock testing rather than production (the live microservice persists correctly).

### Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 4 (1 carried-over/deferred, 3 new — all mock/test-scoped, no production-path correctness risk)
- ⚪ Nit: 0

### Decision

🔴 **Needs changes before approval** — no blocking issues remain (nice work closing out the navigation-guard bug), but four open minor items push past the 2-minor threshold for this round. None of them touch the live microservice path; they're all in the stateless mock backend or its e2e coverage, so this should be a fast pass.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: This was already fixed in the previous commit (0cae116433d3a20e2b1daec9865ebbace5955b8e), before this thread was posted — onSave's pipe now has the same `takeUntil(this.committee$.pipe(filter((c) => c?.uid !== committeeUid)))` guard as onGenerate, and `saving.set(false)` was added to the uid-change reset block. Confirmed still present at current HEAD (bb5509cf1).

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Also already fixed in 0cae116433d3a20e2b1daec9865ebbace5955b8e — moved the `isPlatformBrowser` check into `pollUntilTerminal` itself (single choke point) rather than only guarding the load-time caller. Confirmed still present at current HEAD (bb5509cf1).

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Analyzed rather than fixed, per your own low-priority/optional framing: the template's `@if (fetchLoading())` branch has priority over the `renderableBrief` branch the Edit button lives in — so during any refresh$-triggered refetch (including the 409 path's), the whole card renders the skeleton loader, not the Edit button. The exploitable window would require Angular to process `editMode.set(false)`/`this.editMode` becoming visible in one render pass while `fetchLoading.set(true)` (set synchronously in the same call, one line later via refresh$.next()) hasn't yet been picked up in a later one — both writes happen in the same synchronous handler, so zoneless OnPush change detection should coalesce them into a single pass in practice. Leaving as-is; flagging the analysis in case I'm missing a scheduling edge case, per your own invitation to double-check.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,476 @@
...
+          }
+          this.messageService.add({ severity: 'error', summary: 'Generate failed', detail });
+        },
+      });
+  }
+
+  public onEdit(): void {
+    this.editMode.set(true);
+    this.editForm.controls.briefText.setValue(this.brief()?.brief_text ?? '');
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Same analysis as the thread you referenced on this — see my reply there. Leaving as-is for the same reason: the fetchLoading() skeleton branch takes template priority over the Edit-button branch during the refresh$ window this depends on.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -180,10 +215,14 @@ export class WeeklyBriefService {
...
+      mockRevisionByCommittee.set(committeeId, revision);
       return buildMockBrief(committeeId, {
         state: 'edited',
         brief_text: body.brief_text,
-        revision: body.revision + 1,
+        revision,
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — saveBrief's mock branch now compares `body.revision` against the tracked revision and throws a 409 (REVISION_CONFLICT) before mutating the store, mirroring the live backend's conflict contract. Also replaced `bumpMockRevision` with a full `storeMockBrief` that persists the entire WeeklyBrief (not just revision), so this and the related persistence gaps close together. Added a stale-revision-after-regenerate regression test.

See commit bb5509cf1: weekly-brief.service.ts, weekly-brief.service.spec.ts.

apps/lfx-one/src/server/services/weekly-brief.service.ts
manishdixitlfx: Fixed — same root cause CodeRabbit flagged; see that thread for detail. saveBrief now rejects a mismatched revision with a 409 before touching the store.

See commit bb5509cf1.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -48,6 +48,37 @@ export function briefWindow(): { window_start: string; window_end: string } {
...
+ * Mock-only, in-memory revision tracker keyed by committee. Mock mode is otherwise stateless
+ * (no persistence, resets on server restart — see `WeeklyBriefService`'s class doc), but the
+ * client's poll-until-terminal guard (`pollUntilTerminal`'s `priorRevision` check, LFXV2-2176
+ * round 2) rejects a terminal tick whose revision still matches the pre-regenerate brief. Without
+ * tracking the bump `generateBrief`'s mock branch already promises in its own 202 response, every
+ * subsequent mock `getCurrentBrief` GET would report the same hardcoded revision the pre-regenerate
+ * read did, and a mock regenerate would never satisfy that guard — hanging until the poll's attempt
+ * cap trips instead of completing (Cursor Bugbot finding).
+ */
+const mockRevisionByCommittee = new Map<string, number>();
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — replaced the revision-only tracker with a full per-committee WeeklyBrief store. getCurrentBrief now reads the persisted brief (text, state, regeneration_count, revision all together) instead of rebuilding the canned default; generateBrief/saveBrief both write to it. regeneration_count/regenerations_used are also now cumulative across successive regenerates rather than resetting.

See commit bb5509cf1: weekly-brief.service.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
```diff
@@ -0,0 +1,476 @@
...
+    this.saving.set(true);
+    this.weeklyBriefService
+      .saveWeeklyBrief(committeeUid, { brief_text: text, revision: current.revision })
+      .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: () => {
+          this.saving.set(false);
+          this.editMode.set(false);
+          this.refresh$.next();
+        },
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — same root cause as the persistence gap flagged elsewhere in this round; getCurrentBrief now returns the persisted saved brief (including brief_text and state:'edited') instead of rebuilding the canned generated sample.

See commit bb5509cf1: weekly-brief.service.ts.

apps/lfx-one/src/app/modules/committees/components/weekly-brief-card/weekly-brief-card.component.ts
manishdixitlfx: Fixed — the mock now persists the saved {brief_text, state} (and everything else) per committee, so a subsequent getCurrentBrief reflects the save instead of rebuilding the canned generated sample.

See commit bb5509cf1: weekly-brief.service.ts.

apps/lfx-one/e2e/weekly-brief-card.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+        terminalObserved = true;
+      }
+      const body: WeeklyBriefCurrentResponse = terminalObserved
+        ? { brief: GENERATED_BRIEF, throttle: USED_THROTTLE_AFTER_GENERATE }
+        : { brief: GENERATING_BRIEF, throttle: DEFAULT_THROTTLE };
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — same one-way-flag pattern applied to the retry test elsewhere in this file: a `showTerminal` flag now flips only after the generating state is confirmed on screen, instead of gating on a raw GET count.

See commit bb5509cf1: weekly-brief-card.spec.ts.

apps/lfx-one/e2e/weekly-brief-card.spec.ts
manishdixitlfx: Fixed — switched to a `showTerminal` flag flipped only after the generating-state assertion confirms, mirroring the retry test's already-fixed pattern.

See commit bb5509cf1: weekly-brief-card.spec.ts.

manishdixitlfx:
## Round 5 Review Feedback Addressed

Commit: bb5509cf10d3aadd05e78a1b8b1cfa2c5ae2e269

Round 4's mock-persistence fix (revision-only tracker) closed the poll-never-completes bug but introduced/exposed two further gaps, now closed:

### Changes Made
- **weekly-brief.service.ts**: replaced the revision-only mock tracker with a full per-committee `WeeklyBrief` store — `getCurrentBrief` now returns the persisted brief (text, state, regeneration_count, revision together) instead of rebuilding the canned default on every read (per copilot-pull-request-reviewer[bot], 2 locations; agreed by @dealako)
- **weekly-brief.service.ts** (`saveBrief`): rejects a stale revision with a 409 before mutating the store, mirroring the live backend's optimistic-concurrency contract (per coderabbitai[bot], MAJOR; agreed by @dealako)
- **weekly-brief-card.spec.ts** (e2e): the "loads directly into the generating state" test still gated its terminal-flip on a raw GET count — the exact pattern already fixed elsewhere in this file for the same LaunchDarkly-remount flake risk. Applied the same flag-based fix (per cursor, agreed by @dealako)

### Already Fixed (confirmed against current HEAD, no new action)
- `onSave`'s navigation guard and `pollUntilTerminal`'s SSR guard — both landed in the previous commit before these threads were created

### Analyzed, Not Fixed (2 threads)
- The re-entering-Edit-before-409-refresh race (flagged independently by @dealako as low-priority and by cursor): the template's `fetchLoading()` skeleton branch has priority over the Edit-button branch, so the exploitable window should not survive a single Angular change-detection pass in practice. Replied with the analysis on both threads rather than adding a speculative guard.

### Threads Resolved
9 of 11 unresolved threads addressed this round.

dealako:
Hey @manishdixitlfx — great turnaround again. This round closes out every remaining item from the last review.

👏 **Nice work:**
- The mock backend now persists the **full** `WeeklyBrief` object per committee (`mockBriefByCommittee`), not just the revision — `getCurrentBrief`/`generateBrief`/`saveBrief` all read/write through it, so `brief_text`, `state`, and `regeneration_count` no longer revert to canned defaults after a save or regenerate. Confirmed directly in `weekly-brief.service.ts`.
- `saveBrief`'s mock branch now rejects a stale `body.revision` with a `409 REVISION_CONFLICT` *before* mutating the store, matching the live backend's conflict contract — closes the "stale save silently accepted" gap CodeRabbit and I both flagged.
- The "loads directly into generating state" e2e test is now gated on the same one-way `showTerminal` boolean-flag pattern already used in the retry test, instead of a raw `getCount > 1` threshold — removes the last flake risk from an incidental LaunchDarkly remount GET.
- Good catch on the 409-race analysis (`onEdit`/Edit-button reachability during a post-409 refresh): verified independently — `@if (fetchLoading())` does take template priority over the `renderableBrief` branch the Edit button lives in, so that window can't actually surface a stale-revision resubmit. Confirmed non-issue, not a deferred one.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

All prior blocking/minor items are resolved and confirmed via direct code read (not just taken on the bots' or your word) — mock persistence, stale-revision rejection, SSR poll guard, and the e2e count-gate conversion. No new issues surfaced in this round's diff (security, correctness, performance, style/docs subagents all came back clean), and no new unaddressed bot threads since the last review.

✅ **Approved**

dealako (approved):
Round 5: all prior blocking/minor findings confirmed resolved via direct code read (full mock brief persistence, stale-revision 409 rejection, SSR poll guard, e2e count-gate flag conversion). No new issues from this round's diff. See the summary comment for details. Approving.

PR #1283, opened 2026-07-31
feat(meetings): make meeting card title and date chip clickable

manishdixitlfx:
## Summary
- Title and date/time chip on meeting cards only looked clickable — only the "See Meeting Details" button navigated. Both now wrap real `<a>` elements that mirror the button's existing `routerLink` (past meetings) / `href` + `target=_blank` (upcoming meetings) conditional, so the destination matches exactly and cmd/middle-click still opens a new tab.
- No full-card click surface was added — the two new anchors are small siblings, not ancestors, of the other ~20 interactive elements in the card (copy/edit/delete/RSVP/materials/recording/summary buttons etc.), so no `stopPropagation()` guards were needed and none of that existing behavior changes.
- Applies uniformly to every consumer of the shared `meeting-card` component (dashboard Me/Project lens, committee meetings tab) since only the shared template changed.

**Scope notes:**
- The public/unauthenticated meeting page (`meeting-join.component.html`) does not use `<lfx-meeting-card>` — it has its own bespoke markup, so there's no "public page variant" of this component to update.
- Preserved the pre-existing behavior where past meetings' target resolves to `/meetings/:id` (`MeetingJoinComponent`) rather than `/meetings/:id/details` — that's how the "See Meeting Details" button already worked; not something this PR changes.

JIRA: [LFXV2-2944](https://linuxfoundation.atlassian.net/browse/LFXV2-2944)

## Test plan
- [x] `yarn lint && yarn format:check && yarn check-types && yarn build` all green
- [x] Vitest suite green (268 + 578 tests, unrelated to this change — repo has no Angular component unit-test runner)
- [x] Added `apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts` covering: upcoming card title/chip render as `<a href target=_blank>`, past card title/chip render as `<a routerLink>` with no new tab, and that clicking inner card actions (copy/delete) does not trigger a details navigation
- [ ] Playwright spec not yet run against a live server in this session (starting `yarn start` here risked colliding with another active worktree on port 4200) — pending integration validation pass
- [ ] Manual check in browser: hover affordance, cmd/middle-click opens new tab on upcoming cards, in-app nav on past cards, dashboard (Me/Project lens) + committee meetings tab

[LFXV2-2944]: https://linuxfoundation.atlassian.net/browse/LFXV2-2944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
+++ b/apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
@@ -211,5 +211,11 @@ test.describe('Meeting card — clickable title and date/time chip', () => {
     await upcomingCard.getByTestId('delete-meeting-button').click();
     expect(await deletePopup).toBeNull();
     await expect(page).toHaveURL(/\/meetings(\?.*)?$/);
+
+    // Delete opens a confirmation dialog rather than deleting outright — assert that's the actual
+    // side effect (not just "nothing navigated"), then dismiss it so the test doesn't leave it open.
+    await expect(page.getByRole('dialog')).toBeVisible();
+    await page.getByRole('button', { name: 'Cancel' }).click();
+    await expect(page.getByRole('dialog')).toHaveCount(0);
   });
 });
--- a/apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
+++ b/apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
@@ -67,7 +67,13 @@ async function stubMeLensContext(page: Page): Promise<void> {
   await page.route('**/api/past-meetings/**', (route) => route.fulfill({ status: 404, body: '{}' }));
 }
 
-function upcomingMeeting(id: string, title: string, createdBy: { name: string; username: string; email: string }, extra: Record<string, unknown> = {}) {
+interface MeetingCreator {
+  name: string;
+  username: string;
+  email: string;
+}
+
+function upcomingMeeting(id: string, title: string, createdBy: MeetingCreator, extra: Record<string, unknown> = {}) {
   return {
     id,
     uid: id,
@@ -90,7 +96,7 @@ function upcomingMeeting(id: string, title: string, createdBy: { name: string; u
   };
 }
 
-function pastMeeting(id: string, title: string, createdBy: { name: string; username: string; email: string }, extra: Record<string, unknown> = {}) {
+function pastMeeting(id: string, title: string, createdBy: MeetingCreator, extra: Record<string, unknown> = {}) {
   return {
     ...upcomingMeeting(id, title, createdBy, extra),
     start_time: PAST_START,
@@ -133,7 +139,7 @@ const pastTab = (page: Page) => page.getByTestId('time-filter-tabs').getByTestId
 const card = (page: Page, id: string) => page.locator(`#meeting-${id}`);
 
 test.describe('Meeting card — clickable title and date/time chip', () => {
-  test('upcoming card: title and date chip are real anchors that open the meeting page in a new tab', async ({ page }) => {
+  test('upcoming card: title and date chip are real anchors that open the meeting page in a new tab', async ({ page, context }) => {
     await gotoMyMeetings(page);
 
     const upcomingCard = card(page, 'link-up-1');
@@ -149,6 +155,17 @@ test.describe('Meeting card — clickable title and date/time chip', () => {
     await expect(chip).toHaveJSProperty('tagName', 'A');
     await expect(chip).toHaveAttribute('href', '/meetings/link-up-1');
     await expect(chip).toHaveAttribute('target', '_blank');
+
+    // Activate both anchors — proves they're functional links, not just correctly-attributed markup.
+    // The destination page is real and unmocked, so we only assert a tab actually opened (not its
+    // final URL/content, which could redirect for a fixture id the downstream page doesn't know about).
+    const [titlePopup] = await Promise.all([context.waitForEvent('page'), title.click()]);
+    expect(titlePopup).toBeTruthy();
+    await titlePopup.close();
+
+    const [chipPopup] = await Promise.all([context.waitForEvent('page'), chip.click()]);
+    expect(chipPopup).toBeTruthy();
+    await chipPopup.close();
   });
 
   test('past card: title and date chip navigate in-app (routerLink), not a new tab', async ({ page }) => {
@@ -168,18 +185,31 @@ test.describe('Meeting card — clickable title and date/time chip', () => {
     await expect(chip).toHaveJSProperty('tagName', 'A');
     await expect(chip).toHaveAttribute('href', '/meetings/link-past-1');
     await expect(chip).not.toHaveAttribute('target', '_blank');
+
+    // Activate the title anchor — same-tab SPA nav, so only one of the two anchors is clicked here
+    // (clicking both would require navigating back and re-deriving the past-tab state in between).
+    // Chip's routerLink is already proven equivalent via the identical href assertion above.
+    await title.click();
+    await page.waitForURL((url) => !/^\/meetings\/?$/.test(url.pathname));
+    expect(page.url()).toContain('/meetings/link-past-1');
   });
 
-  test('clicking inner card actions does not trigger a details navigation', async ({ page }) => {
+  test('clicking inner card actions does not trigger a details navigation or open a new tab', async ({ page, context }) => {
     await gotoMyMeetings(page);
 
     const upcomingCard = card(page, 'link-up-1');
     await expect(upcomingCard).toBeVisible();
 
+    const expectNoPopup = () => context.waitForEvent('page', { timeout: 500 }).catch(() => null);
+
+    const copyPopup = expectNoPopup();
     await upcomingCard.getByTestId('copy-meeting-button').click();
+    expect(await copyPopup).toBeNull();
     await expect(page).toHaveURL(/\/meetings(\?.*)?$/);
 
+    const deletePopup = expectNoPopup();
     await upcomingCard.getByTestId('delete-meeting-button').click();
+    expect(await deletePopup).toBeNull();
     await expect(page).toHaveURL(/\/meetings(\?.*)?$/);
   });
 });
--- a/apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
+++ b/apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
@@ -1,0 +1,185 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { expect, Page, Route, test } from '@playwright/test';
+
+const MEETINGS_URL = '/meetings';
+const LENS_COOKIE = 'lfx-active-lens';
+
+// Deterministic, far-future start so `hasMeetingEnded` never drops the upcoming fixture.
+const FUTURE_START = '2099-03-04T15:00:00Z';
+const PAST_START = '2020-03-04T15:00:00Z';
+
+test.setTimeout(120_000);
+
+function fulfillJson(route: Route, body: unknown): Promise<void> {
+  return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(body) });
+}
+
+function skipWhenAuthMissing(page: Page): void {
+  try {
+    const { hostname } = new URL(page.url());
+    if (hostname === 'auth0.com' || hostname.endsWith('.auth0.com')) {
+      test.skip(true, 'TEST_USERNAME / TEST_PASSWORD not configured — see global-setup.ts');
+    }
+  } catch {
+    // Let malformed URLs fail naturally.
+  }
+}
+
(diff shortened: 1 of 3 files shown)


apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+    await gotoMyMeetings(page);
+
+    const upcomingCard = card(page, 'link-up-1');
+    await expect(upcomingCard).toBeVisible();
+
+    await upcomingCard.getByTestId('copy-meeting-button').click();
+    await expect(page).toHaveURL(/\/meetings(\?.*)?$/);
+
+    await upcomingCard.getByTestId('delete-meeting-button').click();
+    await expect(page).toHaveURL(/\/meetings(\?.*)?$/);
```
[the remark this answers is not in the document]
manishdixitlfx: Done — the test now starts listening for a new browser-context page (`context.waitForEvent("page", { timeout: 500 }).catch(() => null)`) *before* each button click and asserts it resolves to `null`, in addition to the existing same-page URL check. That closes the gap you flagged: a click that accidentally bubbled into a `target="_blank"` anchor would now fail the test even though the current page never navigates.

See commit fc36ed05b: `apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts`.

apps/lfx-one/src/app/modules/meetings/components/meeting-card/meeting-card.component.html
```diff
@@ -85,20 +85,55 @@
     <!-- Meeting Title -->
     <div class="flex items-center gap-2 mb-2" data-testid="meeting-title-section">
       @if (meetingTitle()) {
-        <h3 class="text-base font-medium text-gray-900 leading-tight tracking-tight" data-testid="meeting-title">
-          {{ meetingTitle() }}
+        <h3 class="text-base font-medium leading-tight tracking-tight">
+          @if (pastMeeting()) {
+            <a
+              class="text-gray-900 hover:text-blue-600 hover:underline transition-colors cursor-pointer"
```
audigregorie: **Minor — redundant `cursor-pointer` on the new anchors**

Anchors with `[href]`/`[routerLink]` already get the pointer cursor from the UA, so `cursor-pointer` is dead weight. It appears on all four new anchors (lines 91, 98, 114, 124). Safe to drop everywhere.

```suggestion
              class="text-gray-900 hover:text-blue-600 hover:underline transition-colors"
```

apps/lfx-one/src/app/modules/meetings/components/meeting-card/meeting-card.component.html
```diff
@@ -85,20 +85,55 @@
...
     <!-- Feature Badges with Date/Time -->
     <div class="flex flex-wrap gap-1 mb-2" data-testid="feature-badges">
       @if (meetingStartTime()) {
-        <lfx-tag
-          [value]="(meetingStartTime() | meetingTime: meeting().duration : 'date') + ' • ' + (meetingStartTime() | meetingTime: meeting().duration : 'time')"
-          severity="secondary"
-          icon="fa-light fa-calendar-days"
-          data-testid="meeting-datetime"></lfx-tag>
+        @if (pastMeeting()) {
+          <a class="hover:opacity-80 transition-opacity cursor-pointer" data-testid="meeting-datetime" [routerLink]="meetingDetailUrl()">
```
audigregorie: **Minor — inconsistent hover affordance between title and date chip**

The title anchor (line 91) uses `hover:text-blue-600 hover:underline`; this date-chip anchor — which navigates to the exact same destination — uses only `hover:opacity-80`. Two links to the same place read as different kinds of affordance, and an opacity dip alone is a weak link cue. Consider matching the title's `hover:underline` (or at least a clearer cue) on the chip anchor for consistency.

apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  test('clicking inner card actions does not trigger a details navigation', async ({ page }) => {
+    await gotoMyMeetings(page);
+
+    const upcomingCard = card(page, 'link-up-1');
+    await expect(upcomingCard).toBeVisible();
+
+    await upcomingCard.getByTestId('copy-meeting-button').click();
+    await expect(page).toHaveURL(/\/meetings(\?.*)?$/);
+
+    await upcomingCard.getByTestId('delete-meeting-button').click();
```
audigregorie, line 211: **Minor — delete click leaves `p-confirmDialog` open**

`deleteMeeting()` opens a confirmation modal (the fixture has `recurrence: null`, so it routes straight to `showDeleteMeetingModal`). The test asserts the URL stays on `/meetings` — which passes, since the modal doesn't navigate — but never dismisses the dialog, so the test ends with a modal open. Either dismiss it (e.g. click the dialog's cancel button) or assert the dialog appeared, so the test is explicit about the side effect it's actually exercising.

apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+ * out of the serialized state the same way the app does (`username`, then the namespaced claim).
+ */
+async function readViewerLfid(page: Page): Promise<string | null> {
+  const raw = await page.locator('#ng-state').textContent();
+  if (!raw) {
+    return null;
+  }
+  try {
+    const state = JSON.parse(raw) as Record<string, { user?: Record<string, string> }>;
+    const user = state['auth']?.user;
```
audigregorie, line 46: **Minor — silent skip masks structural drift**

`readViewerLfid` returns `null` on any parse/shape mismatch and the harness then `test.skip()`s the whole file. The `'auth'` key is correct today (matches `makeStateKey<AuthContext>('auth')` in `persona.service.ts`), but if the serialized `AuthContext` shape ever changes, every test here silently skips instead of failing — a false-green risk. Consider asserting the resolved LFID is non-null (fail-loud) for the structural assumptions, and keep the `test.skip()` only for the genuinely auth-missing case.

audigregorie (commented):
## Code Review Summary

Clean, well-scoped change that makes the meeting card title and date/time chip real anchors mirroring the existing "See Meeting Details" button's past/upcoming split (`routerLink` vs `[href]`+`target="_blank"`). `meetingDetailUrl()` already returns a query-less path for past meetings, so `routerLink` never parses a query string, and `lfx-tag` renders a `<span>` so wrapping it in `<a>` stays valid HTML. E2E directly validates the "no `stopPropagation` needed" claim.

No Critical or Major findings. Four Minor notes below — none blocking.

### What's done well

- Correctly mirrors the existing "See Meeting Details" button's past/upcoming split, so all three entry points on the card behave identically.
- Adds `rel="noopener noreferrer"` on the new `target="_blank"` anchors — stricter than the pre-existing upcoming button, which omits `rel`.
- E2E confirms the tagName change and that clicking inner actions (copy/delete) does not navigate.

### Rating: ⭐⭐⭐⭐ (4/5)

manishdixitlfx:
## Review Feedback Addressed

Commit: fc36ed05bd221da7d3b3c1e5946cc99f3d7e8a01

### Changes Made
- **meeting-card-title-datetime-link.spec.ts**: the inner-action-click test now asserts no new browser-context page/popup opens after clicking `copy-meeting-button`/`delete-meeting-button`, in addition to the existing same-page URL check (per copilot-pull-request-reviewer)
- **meeting-card-title-datetime-link.spec.ts**: extracted the repeated `createdBy` shape into a local `MeetingCreator` interface, used by both `upcomingMeeting()` and `pastMeeting()` (per coderabbitai)
- **meeting-card-title-datetime-link.spec.ts**: both link-variant tests now click through the title/date-chip anchors instead of only inspecting attributes — the upcoming test asserts a real tab actually opens for both anchors, the past test asserts the page actually navigates away from `/meetings` (per coderabbitai)

On coderabbitai's "activate the links" suggestion specifically: I implemented a scoped-down version rather than asserting on the exact resolved destination URL/content. The destination route (`/meetings/:id` → `MeetingJoinComponent`) is real and unmocked in this test, and our fixture ids don't correspond to real backend data there, so asserting an exact final URL risked flakiness if that page redirects on a not-found meeting. Asserting that a real tab/navigation was triggered (without depending on where the unmocked downstream page lands) still proves the anchors are functional links rather than decorative markup, which was the underlying concern.

Also, for the past-card test, only the title anchor is clicked (not both title and chip) — clicking navigates the SPA away from `/meetings` in-page, so testing both would require navigating back and re-deriving the past-tab filter state. The chip's `routerLink` target is already proven equivalent to the title's via the identical `href` attribute assertion earlier in the same test.

### Threads Resolved
1 of 1 unresolved threads addressed in this iteration. CodeRabbit's two nitpicks were part of a general review body (no separate thread to resolve) — addressed above.

apps/lfx-one/src/app/modules/meetings/components/meeting-card/meeting-card.component.html
manishdixitlfx: Good catch — fixed. Removed `cursor-pointer` from all four new anchors; `<a href>`/`<a routerLink>` (Angular renders a real `href` for routerLink too) already get the pointer cursor from the browser default, and I checked for any global CSS resetting anchor cursor in this repo — none exists.

See commit 636c9f502.

apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
manishdixitlfx: Good catch. Fixed — the test now asserts the confirmation dialog (`p-confirmDialog`) actually appears after clicking delete, rather than only checking the URL stayed on /meetings, and dismisses it via the Cancel button so the test does not leave it open. This is a more precise test of the real side effect (delete requires confirmation, it does not silently do nothing).

See commit 636c9f502.

apps/lfx-one/src/app/modules/meetings/components/meeting-card/meeting-card.component.html
manishdixitlfx: Fair point, and I can see why it reads as inconsistent. This was a deliberate choice rather than an oversight though: the title anchor wraps plain text, so `hover:underline` is a clean, standard link cue there. The date chip wraps a PrimeNG `p-tag` — a filled, rounded pill with its own background/padding — and an underline under that shape (spanning the icon + text + padding) tends to look like a stray line under a badge rather than a link underline, so I used a subtle opacity dim instead, matching how filled-pill/badge hover states are usually done.

I have not visually verified this in a live browser yet (noted as an open item in the PR description), so I would rather not guess at forcing `hover:underline` onto the pill without seeing it render. Happy to revisit if it looks off once this is checked in a running app, or if you feel strongly the affordance should match exactly regardless.

apps/lfx-one/e2e/meeting-card-title-datetime-link.spec.ts
manishdixitlfx: Valid point on the general pattern. This particular function is copied verbatim from the existing precedent this spec is based on (`meetings-organized-by-me.spec.ts`), which has the identical silent-skip-on-null shape. Changing it only in this new file would leave the two specs inconsistent with each other, and fixing the pattern repo-wide (fail-loud vs. skip-on-parse-mismatch) feels like it is out of scope for this PR — it is a test-philosophy change that would touch code this PR does not otherwise own.

Leaving as-is here to stay consistent with the sibling spec; happy to see it addressed repo-wide in a follow-up if that is a pattern worth changing everywhere.

manishdixitlfx:

Commit: 636c9f5024a24edb2b96ce76aa0a5891a56b0a1e

### Changes Made
- **meeting-card.component.html**: removed redundant `cursor-pointer` from all four new anchors — confirmed no global CSS resets anchor cursor in this repo, so the browser default already provides it (per audigregorie)
- **meeting-card-title-datetime-link.spec.ts**: the delete-button test now asserts the confirmation dialog actually appears and dismisses it via Cancel, rather than only checking the URL didn't change and leaving the dialog open (per audigregorie)

### No Change Needed
- **meeting-card.component.html:114**: the date chip's `hover:opacity-80` vs. the title's `hover:underline` is a deliberate choice — underlining a filled PrimeNG tag pill reads oddly since it's not plain text. Not yet visually verified in a live browser, so I didn't want to force a stylistic change without seeing it render; open to revisiting (flagged by audigregorie)
- **meeting-card-title-datetime-link.spec.ts:46**: the silent-skip-on-parse-mismatch pattern in `readViewerLfid()` is copied verbatim from the precedent this spec was based on (`meetings-organized-by-me.spec.ts`). Fixing it only here would diverge from that sibling file; a repo-wide fix-loud-vs-skip decision feels out of scope for this PR (flagged by audigregorie)

### Threads Resolved
4 of 4 threads addressed in this iteration (2 code changes, 2 explained with no change).

audigregorie (approved):

PR #1284, opened 2026-07-31
feat(documents): support multi-file upload in document form

manishdixitlfx:
## Summary
- Documents tab upload (committee + project scopes) now supports multi-select and drag-drop of multiple files in a single dialog, instead of one file per dialog open — matching the existing meeting-materials multi-select pattern.
- Files upload in parallel; each file gets independent success/failure state. Successes are never rolled back on partial failure, and upstream duplicate-name (409) errors are surfaced per file rather than as a single generic toast.
- The file-mode UI is reworked from a single "selected file" chip to an editable multi-row pending list (name pre-filled from basename, editable inline, per-row remove). Link and folder modes are unchanged.
- No upstream/BFF API change — the upload endpoint is confirmed one-file-per-request; the frontend loops requests in parallel via RxJS `mergeMap`.

## Test plan
- [x] `yarn lint`, `yarn format`, `yarn build` green
- [x] `yarn test` green (268 + 578 existing tests unaffected; this repo's `vitest.config.ts` scopes tests to `src/server/**` only — no Angular component/template spec harness exists in `apps/lfx-one`, so no new unit spec was added for this UI-only change)
- [ ] Manual verification pending integration validation pass (per team workflow): committee Documents tab and project Documents tab — multi-select + drag-drop, duplicate-name file, oversized file, mixed success/failure, single-file flow, Add Link / New Folder dialogs unaffected

LFXV2-2942

### DIFF
--- a/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
+++ b/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
@@ -39,7 +39,7 @@
                   control="name"
                   placeholder="Document name"
                   styleClass="w-full"
-                  [dataTest]="'document-form-pending-file-name'"></lfx-input-text>
+                  [dataTest]="'document-form-pending-file-name-' + item.id"></lfx-input-text>
               </div>
               @if (item.status === 'uploading') {
                 <i class="fa-light fa-spinner-third fa-spin text-gray-400 flex-shrink-0"></i>
@@ -47,7 +47,7 @@
                 <button
                   type="button"
                   class="flex-shrink-0 text-gray-400 hover:text-red-500 transition-colors"
-                  aria-label="Remove file"
+                  aria-label="Remove file {{ item.file.name }}"
                   title="Remove file"
                   (click)="removePendingFile(item.id)"
                   data-testid="document-form-remove-file">
--- a/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
+++ b/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
@@ -29,14 +29,17 @@
           <div class="flex flex-col gap-1 px-3 py-2 rounded-lg border border-gray-200 bg-gray-50" data-testid="document-form-pending-file">
             <div class="flex items-center gap-2">
               <i class="fa-light fa-file-lines text-blue-500 flex-shrink-0"></i>
-              <input
-                type="text"
-                class="flex-1 min-w-0 text-sm text-gray-700 bg-transparent border-0 focus:outline-none focus:ring-0 truncate"
-                [value]="item.name"
-                [disabled]="item.status === 'uploading'"
-                (input)="updatePendingFileName(item.id, $any($event.target).value)"
-                [attr.aria-label]="'Document name for ' + item.file.name"
-                data-testid="document-form-pending-file-name" />
+              <div class="flex-1 min-w-0">
+                <label class="sr-only" [attr.for]="'pending-file-name-' + item.id">Document name for {{ item.file.name }}</label>
+                <lfx-input-text
+                  size="small"
+                  [id]="'pending-file-name-' + item.id"
+                  [form]="getNameForm(item.id)"
+                  control="name"
+                  placeholder="Document name"
+                  styleClass="w-full"
+                  [dataTest]="'document-form-pending-file-name'"></lfx-input-text>
+              </div>
               @if (item.status === 'uploading') {
                 <i class="fa-light fa-spinner-third fa-spin text-gray-400 flex-shrink-0"></i>
               } @else {
@@ -51,9 +54,12 @@
                 </button>
               }
             </div>
-            <span class="text-xs text-gray-400 truncate">{{ item.file.name }}</span>
+            @if (getNameForm(item.id).get('name')?.errors?.['required'] && getNameForm(item.id).get('name')?.touched) {
+              <p class="text-xs text-red-600 pl-6">Name is required</p>
+            }
+            <span class="text-xs text-gray-400 truncate pl-6">{{ item.file.name }}</span>
             @if (item.status === 'error' && item.errorMessage) {
-              <p class="text-xs text-red-600">{{ item.errorMessage }}</p>
+              <p class="text-xs text-red-600 pl-6">{{ item.errorMessage }}</p>
             }
           </div>
         }
--- a/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
+++ b/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
@@ -14,6 +14,7 @@
         [customUpload]="true"
         [showUploadButton]="false"
         [showCancelButton]="false"
+        [disabled]="submitting()"
         chooseLabel="Click to upload files or drag and drop"
         dataTest="document-form-file-upload">
       </lfx-file-upload>
@@ -34,7 +35,7 @@
                 <lfx-input-text
                   size="small"
                   [id]="'pending-file-name-' + item.id"
-                  [form]="getNameForm(item.id)"
+                  [form]="item.nameForm"
                   control="name"
                   placeholder="Document name"
                   styleClass="w-full"
@@ -54,8 +55,10 @@
                 </button>
               }
             </div>
-            @if (getNameForm(item.id).get('name')?.errors?.['required'] && getNameForm(item.id).get('name')?.touched) {
+            @if (item.nameForm.get('name')?.errors?.['required'] && item.nameForm.get('name')?.touched) {
               <p class="text-xs text-red-600 pl-6">Name is required</p>
+            } @else if (item.nameForm.get('name')?.errors?.['maxlength'] && item.nameForm.get('name')?.touched) {
+              <p class="text-xs text-red-600 pl-6">Name must be 500 characters or less</p>
             }
             <span class="text-xs text-gray-400 truncate pl-6">{{ item.file.name }}</span>
             @if (item.status === 'error' && item.errorMessage) {
@@ -166,7 +169,14 @@
 
   <!-- Form Actions -->
   <div class="flex justify-end gap-3 pt-4 border-t">
-    <lfx-button [label]="cancelLabel()" severity="secondary" [outlined]="true" (onClick)="onCancel()" size="small" type="button"></lfx-button>
+    <lfx-button
+      [label]="cancelLabel()"
+      severity="secondary"
+      [outlined]="true"
+      [disabled]="submitting()"
+      (onClick)="onCancel()"
+      size="small"
+      type="button"></lfx-button>
     <lfx-button [label]="submitLabel()" [loading]="submitting()" [disabled]="submitting()" type="submit" size="small"></lfx-button>
   </div>
 </form>
--- a/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
+++ b/apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
@@ -5,55 +5,60 @@
   @if (isFile()) {
     <!-- File picker -->
     <div>
-      <label class="block text-[11px] font-semibold text-gray-500 uppercase tracking-wide mb-1"> File <span class="text-red-500">*</span> </label>
-      @if (!selectedFile()) {
-        <lfx-file-upload
-          [multiple]="false"
-          [accept]="acceptString"
-          [maxFileSize]="MAX_FILE_SIZE_BYTES"
-          (onSelect)="onFileSelect($event)"
-          [customUpload]="true"
-          [showUploadButton]="false"
-          [showCancelButton]="false"
-          chooseLabel="Click to upload a file or drag and drop"
-          dataTest="document-form-file-upload">
-        </lfx-file-upload>
-      } @else {
-        <div class="flex items-center justify-between gap-3 px-3 py-2 rounded-lg border border-blue-200 bg-blue-50" data-testid="document-form-selected-file">
-          <div class="flex items-center gap-2 min-w-0">
-            <i class="fa-light fa-file-lines text-blue-500 flex-shrink-0"></i>
-            <span class="text-sm text-gray-700 truncate">{{ selectedFile()!.name }}</span>
-          </div>
-          <button
-            type="button"
-            class="flex-shrink-0 text-gray-400 hover:text-red-500 transition-colors"
-            aria-label="Clear selected file"
-            title="Clear selected file"
-            (click)="clearSelectedFile()"
-            data-testid="document-form-clear-file">
-            <i class="fa-light fa-xmark"></i>
-          </button>
-        </div>
-      }
+      <label class="block text-[11px] font-semibold text-gray-500 uppercase tracking-wide mb-1"> Files <span class="text-red-500">*</span> </label>
+      <lfx-file-upload
+        [multiple]="true"
+        [accept]="acceptString"
+        [maxFileSize]="MAX_FILE_SIZE_BYTES"
+        (onSelect)="onFileSelect($event)"
+        [customUpload]="true"
+        [showUploadButton]="false"
+        [showCancelButton]="false"
+        chooseLabel="Click to upload files or drag and drop"
+        dataTest="document-form-file-upload">
+      </lfx-file-upload>
       @if (fileError()) {
         <p class="mt-1 text-xs text-red-600">{{ fileError() }}</p>
       }
     </div>
 
-    <!-- Name -->
-    <div>
-      <label class="block text-[11px] font-semibold text-gray-500 uppercase tracking-wide mb-1">Name <span class="text-red-500">*</span></label>
-      <lfx-input-text
-        size="small"
-        [form]="form"
-        control="name"
-        placeholder="Display name for the document"
-        styleClass="w-full"
-        dataTest="document-form-name"></lfx-input-text>
-      @if (form.get('name')?.errors?.['required'] && form.get('name')?.touched) {
-        <p class="mt-1 text-xs text-red-600">Name is required</p>
-      }
-    </div>
+    <!-- Pending files -->
+    @if (pendingFiles().length > 0) {
+      <div class="flex flex-col gap-2" data-testid="document-form-pending-files">
+        @for (item of pendingFiles(); track item.id) {
+          <div class="flex flex-col gap-1 px-3 py-2 rounded-lg border border-gray-200 bg-gray-50" data-testid="document-form-pending-file">
+            <div class="flex items-center gap-2">
+              <i class="fa-light fa-file-lines text-blue-500 flex-shrink-0"></i>
+              <input
+                type="text"
+                class="flex-1 min-w-0 text-sm text-gray-700 bg-transparent border-0 focus:outline-none focus:ring-0 truncate"
+                [value]="item.name"
+                [disabled]="item.status === 'uploading'"
+                (input)="updatePendingFileName(item.id, $any($event.target).value)"
+                [attr.aria-label]="'Document name for ' + item.file.name"
+                data-testid="document-form-pending-file-name" />
+              @if (item.status === 'uploading') {
+                <i class="fa-light fa-spinner-third fa-spin text-gray-400 flex-shrink-0"></i>
+              } @else {
+                <button
+                  type="button"
+                  class="flex-shrink-0 text-gray-400 hover:text-red-500 transition-colors"
+                  aria-label="Remove file"
+                  title="Remove file"
+                  (click)="removePendingFile(item.id)"
+                  data-testid="document-form-remove-file">
+                  <i class="fa-light fa-xmark"></i>
+                </button>
+              }
+            </div>
+            <span class="text-xs text-gray-400 truncate">{{ item.file.name }}</span>
+            @if (item.status === 'error' && item.errorMessage) {
+              <p class="text-xs text-red-600">{{ item.errorMessage }}</p>
+            }
+          </div>
+        }
+      </div>
+    }
 
     <!-- Folder (optional) -->
     @if (folderOptions.length > 0) {
@@ -68,7 +73,7 @@
           [showClear]="true"
           styleClass="w-full"
           data-testid="document-form-folder"></lfx-select>
-        <p class="mt-1 text-[11px] text-gray-400">Place this file inside a folder</p>
+        <p class="mt-1 text-[11px] text-gray-400">Place these files inside a folder</p>
       </div>
     }
   } @else if (isLink()) {
@@ -155,7 +160,7 @@
 
   <!-- Form Actions -->
   <div class="flex justify-end gap-3 pt-4 border-t">
-    <lfx-button label="Cancel" severity="secondary" [outlined]="true" (onClick)="onCancel()" size="small" type="button"></lfx-button>
+    <lfx-button [label]="cancelLabel()" severity="secondary" [outlined]="true" (onClick)="onCancel()" size="small" type="button"></lfx-button>
     <lfx-button [label]="submitLabel()" [loading]="submitting()" [disabled]="submitting()" type="submit" size="small"></lfx-button>
   </div>
 </form>
--- a/apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
+++ b/apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
@@ -17,11 +17,12 @@ import {
   CreateProjectDocumentType,
   DocumentFormEntityType,
   DocumentFormMode,
+  PendingDocumentFile,
 } from '@lfx-one/shared/interfaces';
 import { generateAcceptString, getAcceptedFileTypesDisplay, getMimeTypeDisplayName, isFileTypeAllowed } from '@lfx-one/shared/utils';
 import { CommitteeService } from '@services/committee.service';
 import { ProjectService } from '@services/project.service';
-import { Observable } from 'rxjs';
+import { catchError, from, map, mergeMap, Observable, of, toArray } from 'rxjs';
 import { MessageService } from 'primeng/api';
 import { DynamicDialogConfig, DynamicDialogRef } from 'primeng/dynamicdialog';
 
@@ -45,8 +46,10 @@ export class DocumentFormComponent {
 
   // === Writable Signals ===
   public submitting = signal<boolean>(false);
-  public selectedFile = signal<File | null>(null);
+  public pendingFiles = signal<PendingDocumentFile[]>([]);
   public fileError = signal<string | null>(null);
+  /** Set once at least one file in this dialog session has uploaded successfully — drives the Cancel/Done label and close-on-cancel behavior. */
+  public anyUploadSucceeded = signal<boolean>(false);
 
   // Config-based properties
   /** Resource type this dialog targets — drives service dispatch + uniqueness scope copy. */
@@ -67,6 +70,7 @@ export class DocumentFormComponent {
   public isFolder: Signal<boolean> = this.initIsFolder();
   public isFile: Signal<boolean> = this.initIsFile();
   public submitLabel: Signal<string> = this.initSubmitLabel();
+  public cancelLabel: Signal<string> = this.initCancelLabel();
   public descriptionPlaceholder: Signal<string> = this.initDescriptionPlaceholder();
 
   public constructor() {
@@ -82,7 +86,7 @@ export class DocumentFormComponent {
   }
 
   public onCancel(): void {
-    this.dialogRef.close();
+    this.dialogRef.close(this.anyUploadSucceeded() || undefined);
   }
 
   public onSubmit(): void {
@@ -94,43 +98,18 @@ export class DocumentFormComponent {
     const formValue = this.form.getRawValue();
 
     if (this.isFile()) {
-      const file = this.selectedFile();
-      if (!file) {
-        this.fileError.set('Please select a file to upload.');
+      const itemsToUpload = this.pendingFiles().filter((item) => item.status === 'pending' || item.status === 'error');
+      if (itemsToUpload.length === 0) {
+        this.fileError.set('Please select at least one file to upload.');
+        return;
+      }
+      if (itemsToUpload.some((item) => !item.name.trim())) {
+        this.fileError.set('Every file needs a name before uploading.');
         return;
       }
 
-      const metadata = {
-        name: formValue.name,
-        description: formValue.description || undefined,
-        folder_uid: formValue.parent_uid || undefined,
-      };
-
-      this.submitting.set(true);
-      const upload$: Observable<unknown> =
-        this.entityType === 'project'
-          ? this.projectService.uploadProjectDocument(this.entityId, file, metadata)
-          : this.committeeService.uploadCommitteeDocument(this.entityId, file, metadata);
-
-      upload$.subscribe({
-        next: () => {
-          this.submitting.set(false);
-          this.messageService.add({
-            severity: 'success',
-            summary: 'Success',
-            detail: 'File uploaded successfully',
-          });
-          this.dialogRef.close(true);
-        },
-        error: (err: HttpErrorResponse) => {
-          this.submitting.set(false);
-          this.messageService.add({
-            severity: 'error',
-            summary: 'Error',
-            detail: this.extractErrorMessage(err, 'Failed to upload file'),
-          });
-        },
-      });
+      this.fileError.set(null);
+      this.uploadPendingFiles(itemsToUpload, formValue.description || undefined, formValue.parent_uid || undefined);
       return;
     }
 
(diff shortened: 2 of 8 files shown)


apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -82,7 +93,7 @@ export class DocumentFormComponent {
   }
 
   public onCancel(): void {
-    this.dialogRef.close();
+    this.dialogRef.close(this.anyUploadSucceeded() || undefined);
```
audigregorie, line 97: **Critical — Cancel not disabled while submitting**

`onCancel()` calls `this.dialogRef.close(this.anyUploadSucceeded() || undefined)` but the Cancel/"Done" button (template line 169) has no `[disabled]` binding, and the dialog's X / ESC affordances are also live during upload. If the user closes the dialog while the N parallel uploads from `uploadPendingFiles()` are still in flight, the component is destroyed, but the `subscribe` callback fires later and writes to `this.pendingFiles`, calls `this.messageService.add`, and re-invokes `this.dialogRef.close(true)` on a destroyed dialog ref.

The original single-upload flow had the same shape, but this PR widens the window (N parallel uploads, longer tail) and adds a "Done" label that invites closing mid-flight.

Fix: disable Cancel while submitting (mirror the submit button), and/or guard the subscribe callback with `takeUntilDestroyed(inject(DestroyRef))`.

```suggestion
  public onCancel(): void {
    if (this.submitting()) return;
    this.dialogRef.close(this.anyUploadSucceeded() || undefined);
  }
```

audigregorie (commented):
## Code Review Summary

Solid multi-file upload implementation — `PendingDocumentFile` is correctly placed in `@lfx-one/shared/interfaces`, per-file `catchError` keeps one failure from aborting the batch, and the partial-failure UX (re-enable the name form on error so a 409 can be renamed and retried) is thoughtful. `@for` tracking and the `init*()` pattern are clean.

The main blockers are around closing the dialog while uploads are in flight: Cancel isn't disabled while `submitting`, so closing mid-batch destroys the component while the `uploadPendingFiles` subscribe callback still writes to signals and re-calls `dialogRef.close`. The X / ESC path also returns a different close result than the "Done" button, so partial-success refresh depends on which affordance the user picks. Bounding `mergeMap` concurrency and adding a `takeUntilDestroyed` guard would close the robustness gap.

### Critical — outside the diff

- **Cancel not disabled while submitting** (document-form.component.html ~line 169): the Cancel/"Done" button has no `[disabled]` binding, so it can close the dialog during parallel uploads and trigger writes on a destroyed component.

### Major — outside the diff

- **X vs "Done" refresh inconsistency**: `onCancel()` returns `anyUploadSucceeded() || undefined`, but the dialog X / ESC closes with `undefined` always — partial-success refresh depends on which close affordance the user picks.
- **Unbounded `mergeMap` concurrency + no timeout** (uploadPendingFiles): a large drag-drop fires N simultaneous requests; a single hung request stalls `toArray()` indefinitely with `submitting` stuck true.
- **`onSubmit` re-entrancy**: Enter in the Description textarea can fire `ngSubmit` while uploads are in flight; the second call finds no `pending`/`error` items and sets a misleading `fileError`.

### Minor — outside the diff

- `getNameForm` fallback creates a new `FormGroup` per CD cycle if `id` is ever missing — prefer `this.nameForms.get(id)!`.
- `[dataTest]` on the pending-file name input is a static string inside `@for` — not unique per row for E2E.
- Remove button `aria-label` doesn't disambiguate which file.
- `PendingDocumentFileStatus` includes `'success'`, which is never set (successful items are filtered out).
- Description placeholder is singular (`this document`) while the rest of file mode is pluralized.
- `anyUploadSucceeded` could be `private`.

### What's done well

- Correct shared-package placement of the new interface, with a clear comment on why the editable name lives in a per-item `FormGroup`.
- Per-file `catchError` + name-form re-enable on error → clean partial-failure UX.
- Stable client-generated IDs for `@for track`; immutable `update()` on `pendingFiles`; `init*()` pattern followed for `cancelLabel`.

Rating: ⭐⭐⭐⭐ (4/5)

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
```diff
@@ -155,7 +166,7 @@
 
   <!-- Form Actions -->
   <div class="flex justify-end gap-3 pt-4 border-t">
-    <lfx-button label="Cancel" severity="secondary" [outlined]="true" (onClick)="onCancel()" size="small" type="button"></lfx-button>
+    <lfx-button [label]="cancelLabel()" severity="secondary" [outlined]="true" (onClick)="onCancel()" size="small" type="button"></lfx-button>
```
audigregorie: **Critical — Cancel button not disabled while submitting**

The submit button binds `[disabled]="submitting()"`, but the Cancel/"Done" button does not. Add the same guard so the dialog can't be closed mid-batch (see the `onCancel`/`uploadPendingFiles` re-entrancy note on the TS file).

```suggestion
    <lfx-button [label]="cancelLabel()" severity="secondary" [outlined]="true" (onClick)="onCancel()" size="small" type="button" [disabled]="submitting()"></lfx-button>
```

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -82,7 +93,7 @@ export class DocumentFormComponent {
...
   }
```
audigregorie, line 98: **Major — X vs "Done" refresh inconsistency**

`onCancel()` returns `anyUploadSucceeded() || undefined`, so clicking "Done" after a partial success closes with `true` and the parent refreshes. But the dialog's X button and ESC both call `dialogRef.close()` with no argument → `undefined` → the parent does **not** refresh, even though some files uploaded successfully. Two close paths for the same dialog produce different refresh behavior depending on which affordance the user picks; partially-uploaded files won't appear until a manual refresh.

Fix: route all close paths through one handler (e.g., subscribe to `dialogRef.onClose` and emit the right result), or document why X and "Done" intentionally differ.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -249,6 +255,89 @@ export class DocumentFormComponent {
...
+
+          return upload$.pipe(
+            map(() => ({ id: item.id, ok: true as const })),
+            catchError((err: HttpErrorResponse) =>
+              of({ id: item.id, ok: false as const, message: this.extractErrorMessage(err, `Failed to upload "${name}"`) })
+            )
+          );
+        }),
+        toArray()
+      )
```
audigregorie, line 290: **Major — unbounded `mergeMap` concurrency + no timeout**

`mergeMap((item) => …)` runs with **unbounded concurrency** — every selected file fires a simultaneous HTTP request (the comment calls this out). For a large drag-drop (20+ files) this saturates the browser's per-origin connection pool and bursts the BFF/upstream. There's also no `timeout()` — a single hung request stalls `toArray()` indefinitely, leaving `submitting` true and the dialog stuck.

Fix: bound concurrency (`mergeMap(fn, 4)`) and consider a `timeout()` per upload so one stalled request can't hang the whole batch.

```suggestion
        mergeMap((item) => {
          const name = this.getPendingFileName(item.id);
          const metadata = { name, description, folder_uid: folderUid };
          const upload$: Observable<unknown> =
            this.entityType === 'project'
              ? this.projectService.uploadProjectDocument(this.entityId, item.file, metadata)
              : this.committeeService.uploadCommitteeDocument(this.entityId, item.file, metadata);

          return upload$.pipe(
            map(() => ({ id: item.id, ok: true as const })),
            catchError((err: HttpErrorResponse) =>
              of({ id: item.id, ok: false as const, message: this.extractErrorMessage(err, `Failed to upload "${name}"`) })
            )
          );
        }, 4)
```

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -82,7 +93,7 @@ export class DocumentFormComponent {
   }
 
   public onCancel(): void {
-    this.dialogRef.close();
+    this.dialogRef.close(this.anyUploadSucceeded() || undefined);
   }
 
   public onSubmit(): void {
```
audigregorie, line 100: **Major — `onSubmit` re-entrancy during upload**

`onSubmit()` is reachable while uploads are in flight: pressing Enter inside the Description textarea fires `ngSubmit` even though the submit button is `[disabled]`. On a second invocation, `itemsToUpload` filters to `status === 'pending' || 'error'` — all items are now `'uploading'`, so the filter yields an empty array and the method sets `fileError` to `'Please select at least one file to upload.'`, surfacing a misleading error under an in-progress upload.

Fix: early-return when `submitting()` is true (add as the first line of `onSubmit`).

```suggestion
  public onSubmit(): void {
    if (this.submitting()) return;
```

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -180,28 +167,47 @@ export class DocumentFormComponent {
...
-  public clearSelectedFile(): void {
-    this.selectedFile.set(null);
     this.fileError.set(null);
+    this.pendingFiles.update((items) => [...items, ...newItems]);
+  }
+
+  /** FormGroup backing a pending file's editable name input — see `nameForms`. */
+  public getNameForm(id: string): FormGroup {
+    return this.nameForms.get(id) ?? new FormGroup({ name: new FormControl('') });
+  }
```
audigregorie: **Minor — `getNameForm` fallback creates a new `FormGroup` per CD cycle**

`this.nameForms.get(id) ?? new FormGroup({ name: new FormControl('') })` — the fallback creates a brand-new `FormGroup` on every change-detection cycle if `id` is ever missing. In practice `id` is always present (the `@for` iterates `pendingFiles()` keyed by the same `id`), so the fallback is unreachable — but if it ever fires, it both leaks and silently disconnects the input from the model.

Fix: `return this.nameForms.get(id)!;` (the `@for` guarantees presence) or throw an error to surface the bug.

```suggestion
  public getNameForm(id: string): FormGroup {
    return this.nameForms.get(id)!;
  }
```

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
```diff
@@ -5,55 +5,66 @@
...
+              <div class="flex-1 min-w-0">
+                <label class="sr-only" [attr.for]="'pending-file-name-' + item.id">Document name for {{ item.file.name }}</label>
+                <lfx-input-text
+                  size="small"
+                  [id]="'pending-file-name-' + item.id"
+                  [form]="getNameForm(item.id)"
+                  control="name"
+                  placeholder="Document name"
+                  styleClass="w-full"
+                  [dataTest]="'document-form-pending-file-name'"></lfx-input-text>
```
audigregorie: **Minor — `dataTest` not unique per row**

`[dataTest]="'document-form-pending-file-name'"` is a static string inside a `@for` loop, so every pending-file row gets the same `data-test` attribute. E2E selectors by `data-test` can't target a specific row.

```suggestion
                  [dataTest]="'document-form-pending-file-name-' + item.id"></lfx-input-text>
```

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
```diff
@@ -5,55 +5,66 @@
...
+              } @else {
+                <button
+                  type="button"
+                  class="flex-shrink-0 text-gray-400 hover:text-red-500 transition-colors"
+                  aria-label="Remove file"
+                  title="Remove file"
+                  (click)="removePendingFile(item.id)"
+                  data-testid="document-form-remove-file">
+                  <i class="fa-light fa-xmark"></i>
+                </button>
```
audigregorie, line 55: **Minor — Remove button `aria-label` doesn't disambiguate**

The remove button's `aria-label="Remove file"` doesn't identify which file. A screen reader iterating the list announces "Remove file" repeatedly with no disambiguation.

```suggestion
                <button
                  type="button"
                  class="flex-shrink-0 text-gray-400 hover:text-red-500 transition-colors"
                  aria-label="Remove file {{ item.file.name }}"
                  title="Remove file"
                  (click)="removePendingFile(item.id)"
                  data-testid="document-form-remove-file">
                  <i class="fa-light fa-xmark"></i>
                </button>
```

packages/shared/src/interfaces/committee.interface.ts
```diff
@@ -771,6 +771,25 @@ export type DocumentFormMode = CreateCommitteeDocumentType | 'file';
 /** Which resource type the shared document form dialog operates against. Drives service dispatch + copy. */
 export type DocumentFormEntityType = 'committee' | 'project';
 
+/** Upload status of a single file staged in the document form dialog's multi-file file-mode flow. */
+export type PendingDocumentFileStatus = 'pending' | 'uploading' | 'success' | 'error';
```
audigregorie: **Minor — `'success'` status is unreachable**

`PendingDocumentFileStatus = 'pending' | 'uploading' | 'success' | 'error'` includes `'success'`, but the component never sets `'success'` — successful items are filtered out of `pendingFiles` instead. The union advertises a state that's unreachable in practice.

Fix: drop `'success'` from the union, or keep successful items in the list with `status: 'success'` and render a success row (clearer audit trail for the user).

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -292,12 +381,20 @@ export class DocumentFormComponent {
...
     });
   }
 
+  private initCancelLabel(): Signal<string> {
+    return computed(() => (this.isFile() && this.anyUploadSucceeded() ? 'Done' : 'Cancel'));
+  }
+
   private initDescriptionPlaceholder(): Signal<string> {
     return computed(() => {
       if (this.isFile()) return 'Brief description of this document';
```
audigregorie: **Minor — Description placeholder singular in plural mode**

`'Brief description of this document'` (singular) while the rest of file mode is pluralized for multiple files. With a batch upload, the description applies to all files, so the singular reads oddly.

```suggestion
      if (this.isFile()) return 'Brief description of these documents';
```

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -45,8 +46,10 @@ export class DocumentFormComponent {
 
   // === Writable Signals ===
   public submitting = signal<boolean>(false);
-  public selectedFile = signal<File | null>(null);
+  public pendingFiles = signal<PendingDocumentFile[]>([]);
   public fileError = signal<string | null>(null);
+  /** Set once at least one file in this dialog session has uploaded successfully — drives the Cancel/Done label and close-on-cancel behavior. */
+  public anyUploadSucceeded = signal<boolean>(false);
```
audigregorie: **Minor — `anyUploadSucceeded` could be `private`**

`anyUploadSucceeded` is `public` but is only read internally (`initCancelLabel`, `onCancel`). It doesn't need to be template-visible — making it `private` tightens encapsulation.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
```diff
@@ -5,55 +5,66 @@
...
-        </div>
-      }
+      <label class="block text-[11px] font-semibold text-gray-500 uppercase tracking-wide mb-1"> Files <span class="text-red-500">*</span> </label>
+      <lfx-file-upload
+        [multiple]="true"
+        [accept]="acceptString"
+        [maxFileSize]="MAX_FILE_SIZE_BYTES"
+        (onSelect)="onFileSelect($event)"
+        [customUpload]="true"
+        [showUploadButton]="false"
```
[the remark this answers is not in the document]
manishdixitlfx: Done — the file picker (`lfx-file-upload`) is now disabled while `submitting()` is true, so the pending batch can't be mutated mid-upload.

See commit 077c2b1a5: `document-form.component.html` adds `[disabled]="submitting()"` to `lfx-file-upload`.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -249,6 +255,89 @@ export class DocumentFormComponent {
...
+              : this.committeeService.uploadCommitteeDocument(this.entityId, item.file, metadata);
+
+          return upload$.pipe(
+            map(() => ({ id: item.id, ok: true as const })),
+            catchError((err: HttpErrorResponse) =>
+              of({ id: item.id, ok: false as const, message: this.extractErrorMessage(err, `Failed to upload "${name}"`) })
+            )
+          );
+        }),
+        toArray()
```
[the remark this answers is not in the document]
manishdixitlfx: Done — each upload result is now applied to its row via `tap()` the moment it settles, inside the `mergeMap`. `toArray()` is only used afterward to compute the final batch-level toast/close decision, not per-file state.

See commit 077c2b1a5: `document-form.component.ts`'s `uploadPendingFiles`/`applyUploadResult`.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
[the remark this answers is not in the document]
manishdixitlfx: Done — file-mode dialogs now open with `closable: false`, so the header close icon can no longer bypass the success-aware close result. Dismissal is only possible through the Cancel/Done button, which is also disabled while `submitting()` is true.

See commit 077c2b1a5: `committee-documents.component.ts` and `documents-dashboard.component.ts`.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
```diff
@@ -5,55 +5,66 @@
...
+        @for (item of pendingFiles(); track item.id) {
+          <div class="flex flex-col gap-1 px-3 py-2 rounded-lg border border-gray-200 bg-gray-50" data-testid="document-form-pending-file">
+            <div class="flex items-center gap-2">
+              <i class="fa-light fa-file-lines text-blue-500 flex-shrink-0"></i>
+              <div class="flex-1 min-w-0">
+                <label class="sr-only" [attr.for]="'pending-file-name-' + item.id">Document name for {{ item.file.name }}</label>
+                <lfx-input-text
+                  size="small"
+                  [id]="'pending-file-name-' + item.id"
+                  [form]="getNameForm(item.id)"
```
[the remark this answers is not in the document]
manishdixitlfx: Done — each pending file now carries its own `nameForm` directly (`PendingDocumentFile & { nameForm: FormGroup }`), so the template binds `item.nameForm` as a plain property read instead of calling `getNameForm(item.id)`. That method has been removed.

See commit 077c2b1a5: `document-form.component.ts`/`.html`.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -94,43 +105,19 @@ export class DocumentFormComponent {
     const formValue = this.form.getRawValue();
 
     if (this.isFile()) {
-      const file = this.selectedFile();
```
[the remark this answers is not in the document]
manishdixitlfx: Thanks — same root cause as the header-close issue flagged elsewhere on this PR. Fixed via `closable: false` on the file-mode dialog opens, so dismissal is now only possible through the Cancel/Done button (itself disabled during upload).

See commit 077c2b1a5: `committee-documents.component.ts` and `documents-dashboard.component.ts`.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
```diff
@@ -5,55 +5,66 @@
...
+        [multiple]="true"
+        [accept]="acceptString"
+        [maxFileSize]="MAX_FILE_SIZE_BYTES"
+        (onSelect)="onFileSelect($event)"
+        [customUpload]="true"
+        [showUploadButton]="false"
+        [showCancelButton]="false"
+        chooseLabel="Click to upload files or drag and drop"
+        dataTest="document-form-file-upload">
+      </lfx-file-upload>
```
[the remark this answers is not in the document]
manishdixitlfx: Thanks — fixed by disabling the file picker while `submitting()` is true, so new files can't be added mid-batch.

See commit 077c2b1a5: `document-form.component.html`.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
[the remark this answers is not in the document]
manishdixitlfx: Done — added `[disabled]="submitting()"` to the Cancel button, matching the suggested diff.

See commit 077c2b1a5.

manishdixitlfx:
## Review Feedback Addressed

Commit: 077c2b1a5

### Changes Made
- **document-form.component.html/.ts**: disabled the file picker while `submitting()` so an in-flight batch can't be silently mutated and then dropped on auto-close (per copilot-pull-request-reviewer, cursor)
- **document-form.component.ts**: per-file upload results now apply to their row via `tap()` as each request settles, instead of only after `toArray()` completes — per-file status is now actually live (per copilot-pull-request-reviewer)
- **committee-documents.component.ts, documents-dashboard.component.ts**: file-mode dialogs now open with `closable: false` so the header close icon can't bypass the success-aware close result and skip the parent's refresh after a partial upload (per copilot-pull-request-reviewer, cursor)
- **document-form.component.ts/.html**: each pending file now carries its own `nameForm` directly instead of a `getNameForm()` method called from the template, per `docs/reviews/frontend-checklist.md`'s "no template functions" rule (per copilot-pull-request-reviewer)
- **document-form.component.html**: Cancel button now disabled while `submitting()`, so the dialog can't be dismissed mid-upload (per coderabbitai)
- **document-form.component.ts, file-upload.constants.ts**: capped upload concurrency at `DOCUMENT_UPLOAD_CONCURRENCY` (3), matching the repo's existing `FOUNDATION_PROJECT_COUNT_FETCH_CONCURRENCY` / `COMMITTEE_INVITE_CONCURRENCY` pattern — each upload buffers its body fully in memory server-side (flagged by both copilot-pull-request-reviewer and coderabbitai, in review-body nitpicks not attached to a specific thread)
- **document-form.component.ts**: added `Validators.maxLength(500)` to the per-file name control, matching the backend's documented limit, plus an inline error message (flagged by coderabbitai as a review-body nitpick, not attached to a specific thread)

### Threads Resolved
7 of 7 unresolved threads addressed in this iteration.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
manishdixitlfx: @audigregorie Good catch — and it goes deeper than the Cancel button. `closeOnEscape` defaults to `true` independently of `closable`, so even with `closable: false` (added in the prior commit) the ESC key was still bypassing `onCancel()` and closing with `undefined`. Fixed both file-mode dialog opens with `closeOnEscape: false` alongside `closable: false`, and added `if (this.submitting()) return;` to `onCancel()` as defense-in-depth.

See commit 8ee5c020f.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
manishdixitlfx: @audigregorie This was already fixed in the prior commit (077c2b1a5), addressing the same finding from coderabbitai — the Cancel button has had `[disabled]="submitting()"` since before your review landed. Confirmed still present in current code.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
manishdixitlfx: @audigregorie Same root cause as the `onCancel` thread above — fixed via `closeOnEscape: false` on the file-mode dialog opens (alongside the existing `closable: false`), so ESC can no longer produce a different refresh outcome than the Cancel/Done button.

See commit 8ee5c020f.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
manishdixitlfx: @audigregorie The concurrency portion was already fixed in the prior commit (077c2b1a5) — `mergeMap` is capped at `DOCUMENT_UPLOAD_CONCURRENCY` (3), matching the repo's existing `FOUNDATION_PROJECT_COUNT_FETCH_CONCURRENCY`/`COMMITTEE_INVITE_CONCURRENCY` pattern.

The `timeout()` suggestion is a real gap I'm deferring rather than guessing at a value — files can be up to 100MB, and no existing repo constant fits a safe timeout for that without risking failing legitimate slow uploads. Leaving this thread open as a tracked follow-up.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
manishdixitlfx: @audigregorie Added `if (this.submitting()) return;` as the first line of `onSubmit()` — cheap defense-in-depth, agreed it's worth having regardless.

One correction on the specific mechanism though: Enter inside a `<textarea>` (which `lfx-textarea` renders) doesn't trigger implicit form submission in standard browser behavior — that only applies to single-line text inputs. And the submit button's native `disabled` attribute is correctly bound via PrimeNG's `Button` component, so implicit submission via Enter in the per-file name inputs wouldn't fire either while `submitting()` is true. So I don't think the Description-textarea path was actually exploitable, but the guard is good insurance either way.

See commit 8ee5c020f.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
manishdixitlfx: @audigregorie This was already superseded by the prior commit (077c2b1a5) — `getNameForm()` and the `Map<string, FormGroup>` it looked up from were removed entirely. Each pending file now carries its `nameForm` directly as a property, so there's no fallback-allocation path left to worry about.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
manishdixitlfx: @audigregorie Done — `dataTest` now includes `item.id`.

See commit 8ee5c020f.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.html
manishdixitlfx: @audigregorie Done — the remove button's `aria-label` now includes the filename.

See commit 8ee5c020f.

packages/shared/src/interfaces/committee.interface.ts
manishdixitlfx: @audigregorie Good catch — dropped `'success'` from `PendingDocumentFileStatus`. Went with removing it rather than rendering a success row, since a focused "what still needs attention" list matches the intended UX here.

See commit 8ee5c020f.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
manishdixitlfx: @audigregorie Done — pluralized to "these documents".

See commit 8ee5c020f.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
manishdixitlfx: @audigregorie Done — narrowed to `private`.

See commit 8ee5c020f.

manishdixitlfx:

Commit: 8ee5c020f

@audigregorie left 11 review comments after the previous round landed. 2 were already fixed by the prior commit (077c2b1a5); the rest are addressed here.

### Changes Made
- **committee-documents.component.ts, documents-dashboard.component.ts**: also set `closeOnEscape: false` on file-mode dialog opens — it defaults to `true` independently of `closable`, so ESC was still bypassing the success-aware close result after a partial upload (per @audigregorie)
- **document-form.component.ts**: added a `submitting()` guard to `onCancel()` and `onSubmit()` as defense-in-depth (per @audigregorie)
- **document-form.component.html**: unique `dataTest` per pending-file row; remove-button `aria-label` now includes the filename (per @audigregorie)
- **committee.interface.ts**: dropped the unreachable `'success'` state from `PendingDocumentFileStatus` (per @audigregorie)
- **document-form.component.ts**: pluralized the file-mode description placeholder; narrowed `anyUploadSucceeded` to `private` (per @audigregorie)

### Already Addressed (before this review landed)
- Cancel button `[disabled]` binding — fixed in 077c2b1a5 (per coderabbitai's earlier finding)
- `getNameForm()` method — removed entirely in 077c2b1a5 (per copilot-pull-request-reviewer's earlier finding)

### Still Open
- **document-form.component.ts:287** — a per-upload `timeout()` was suggested to guard against a fully hung request. Deferred: files can be up to 100MB and no existing repo constant fits a safe timeout without risking failing legitimate slow uploads. Left as a tracked follow-up rather than guessing a value.

### Threads Resolved
10 of 11 threads addressed in this iteration; 1 left open (see above).

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -249,6 +248,90 @@ export class DocumentFormComponent {
...
+            detail: `${successCount} of ${successCount + failureCount} files uploaded. Fix the errors below and try again.`,
+          });
+        } else {
+          this.messageService.add({
+            severity: 'error',
+            summary: 'Error',
+            detail: failureCount === 1 ? 'Failed to upload file. Please try again.' : 'Failed to upload files. Please try again.',
+          });
+        }
+      });
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed — the outer subscribe() only had a `next` callback, so an unexpected stream failure would've left `submitting` stuck true forever with no dismissal path (Cancel/picker/closable/closeOnEscape are all locked during submit). Added an `error` callback that resets `submitting`, reverts any still-`uploading` rows to retryable `error` rows, and toasts.

The narrower half of this — a single request that hangs without ever resolving *or* erroring — still needs a `timeout()`, which I'm deferring for the same reason as the earlier concurrency thread: files can be up to 100MB and no existing repo constant fits a safe value without risking legitimate slow uploads failing.

See commit e9a9b65ca.

manishdixitlfx:
## Review Feedback Addressed (round 3)

Commit: e9a9b65ca

cursor's re-review after round 2 caught a real regression introduced by my own closable/closeOnEscape/Cancel-disable fixes: the upload batch's outer `subscribe()` had no `error` callback, so an unexpected stream failure would leave `submitting` stuck true with no dismissal path short of a page reload.

### Changes Made
- **document-form.component.ts**: added an `error` callback to the batch subscription — resets `submitting`, reverts any still-`uploading` rows to retryable `error` rows, and shows a toast (per cursor)

### Still Open
- Same deferred timeout() follow-up as noted in round 2 — covers the narrower case of a single request that hangs without ever resolving or erroring.

### Threads Resolved
1 of 1 new thread from this round addressed.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -180,28 +169,40 @@ export class DocumentFormComponent {
...
+    if (newItems.length === 0) return;
 
-  public clearSelectedFile(): void {
-    this.selectedFile.set(null);
     this.fileError.set(null);
+    this.pendingFiles.update((items) => [...items, ...newItems]);
+  }
+
+  public removePendingFile(id: string): void {
+    this.pendingFiles.update((items) => items.filter((item) => item.id !== id));
```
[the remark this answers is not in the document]
manishdixitlfx: I don't think this is currently reproducible. `p-fileUpload`'s internal `onFileSelect` calls `clearInputElement()` unconditionally after every selection, which resets the native `<input type="file">`'s `.value` — that's the mechanism browsers need to refire `change` for a repeat pick of the same path, and PrimeNG already does it.

This is also not a pattern unique to this component — `meeting-materials-drawer.component.ts` already ships the identical shape (`lfx-file-upload [multiple]="true"`, removing pending rows from a local signal without touching the widget's own internal file list) elsewhere in this repo, so if this were a real bug it'd already be live in production there.

Leaving this open in case you have a concrete repro I'm missing — happy to dig further if so.

apps/lfx-one/src/app/shared/components/document-form/document-form.component.ts
```diff
@@ -180,28 +169,40 @@ export class DocumentFormComponent {
...
-    const nameControl = this.form.get('name');
-    if (nameControl && !nameControl.value) {
+      existingNames.add(file.name);
+      // Pre-fill from the basename (sans extension); user can edit inline before submit.
       const baseName = file.name.replace(/\.[^/.]+$/, '');
-      nameControl.setValue(baseName);
-    }
-  }
+      const nameForm = new FormGroup({ name: new FormControl(baseName, [Validators.required, Validators.maxLength(500)]) });
+      newItems.push({ id: crypto.randomUUID(), file, status: 'pending', nameForm });
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed, good catch — fixed by keeping the extension in the pre-filled display name instead of stripping it. `file.name` uniqueness is already enforced by the existing duplicate-selection check, so this guarantees the default names can't collide.

See commit 04a0bd90d.

manishdixitlfx:
## Review Feedback Addressed (round 4)

Commit: 04a0bd90d

### Changes Made
- **document-form.component.ts**: pending-file display names now pre-fill with the full filename (extension included) instead of the stripped basename, so two files sharing a basename (`report.pdf`, `report.docx`) can no longer silently collide on a shared default name and trip a confusing 409 at submit (per cursor)

### No Change Needed
- **document-form.component.ts:206** — cursor flagged that a removed file couldn't be re-selected. I don't think this is reproducible: PrimeNG's `FileUpload` already resets the native file input's value after every selection (`clearInputElement()`), and this exact multi-select shape already ships in `meeting-materials-drawer.component.ts` elsewhere in the repo. Left the thread open for a concrete repro if one exists.

### Threads Resolved
1 of 2 new threads from this round addressed; 1 left open pending reviewer confirmation (see above), plus the previously-deferred timeout thread.

audigregorie (approved):

PR #1288, opened 2026-08-01
feat(committees): server-side activity feed aggregation

manishdixitlfx:
## Summary

Replaces the committee overview's client-side "Recent Activity" stop-gap (`buildActivityFeed`, which independently fetched past meetings, votes, surveys, and — solely to seed this widget — documents, then merged/sorted/capped them in the browser) with a server-side BFF endpoint: `GET /api/committees/:uid/activity`.

v1 aggregates only *existing* upstream sources (past meetings, votes, surveys, documents) — no new microservice, no event log. The wire contract (`ActivityEvent` discriminated union) is shaped so a real event log can populate it later without changing the response shape or client rendering.

## What changed

- **Shared (`packages/shared`)**: new `ActivityEvent` discriminated union (`packages/shared/src/interfaces/activity-event.interface.ts`) and `activity-event.constants.ts`. `ActivityEventType` is derived (`ActivityEvent['type']`), not hand-duplicated. `activity-feed.utils.ts`'s `buildActivityFeed` is replaced by a pure `mapActivityEventsToFeedItems(events, opts)` mapper — same label/icon/action logic, now reading from `event.payload` instead of raw entities.
- **Server**: new `committee-activity.controller.ts` / `committee-activity.service.ts` / `committee-activity-query.helper.ts` (3-file pattern), wired into `committees.route.ts`. The service fans out to 4 upstream legs (past meetings, votes, surveys, documents — itself 3 calls: folders/links/files) plus a committee-service lookup for `enable_voting`, merges by a composite `(occurred_at, key)` cursor, and returns `{ data, page_token }`.
- **Client**: `committee-overview.component.ts` swaps its 4-source fetch-and-merge for one call to the new endpoint; rendering is unchanged (same `ActivityFeedItem` shape, same click-through drawer hydration from the existing `votes()`/`surveys()` signals).

## v1-emitted vs. deferred event types

`ActivityEventType` documents the full target vocabulary so the future event-log swap doesn't require a contract change:

- **v1-emitted** (constructed by this PR): `meeting_held`, `vote_opened`, `vote_closed`, `survey_published`, `survey_closed`, `document_uploaded`.
- **Deferred** (typed now, never constructed in v1 — no upstream source exists yet): `document_deleted` (no delete-tombstone upstream), `member_joined`/`member_left` (no membership-history tracking), `notes_added` (no notes feature).

## One-event-per-row design choice

Votes and surveys each emit **one** event per row, not two (open + close): if the row has ended, emit `vote_closed`/`survey_closed` at its end timestamp; otherwise `vote_opened`/`survey_opened` at creation time. This keeps the widget's rendering identical to today's single-row-per-vote/survey behavior — a data-source swap, not a redesign. A future real event log would naturally emit both events as they actually occur; the type enum already accommodates that.

## Sort-order / date-field limitations

- **Meetings**: `occurred_at` is built from `start_time` and is a **hard guarantee**, not an approximation — verified directly against `lfx-v2-meeting-service` source (`internal/domain/models/event_models.go`, `cmd/meeting-api/eventing/past_meeting_event_handler.go`, `docs/event-processing.md`): `v1_past_meeting` rows are sourced from exactly one NATS stream and the struct backing them has no `scheduled_start_time` field at all, so there's no alternate value it could diverge from. A runtime tripwire (aggregated per-request `logger.warning` in `fetchPastMeetingEvents`) still fires if this upstream contract ever changes, since "verified today" isn't "true forever."
- **Votes / files**: best-effort narrowing — query-service's documented `sort` enum (`name_asc/desc`, `updated_asc/desc`) doesn't support arbitrary date-field ordering, so these legs use `date_field`/`date_from`/`date_to` filtering plus a bounded per-leg page size, not a proven-complete sort. This is materially better than the old stop-gap's *unbounded* drain, but not a mathematical guarantee at high volume — acceptable at committee-activity scale, called out here as a known v1 limitation.
- **Surveys**: deliberately sends **no** upstream date narrowing at all (fixed after review — see below), relying entirely on the in-memory since/cursor pass, and a bounded per-leg page size (`sort: updated_desc` only) for correctness. Unlike votes/files, a survey's cutoff-driven `occurred_at` isn't bounded below by any field the row is ever written to, so date-field narrowing on this leg would systematically (not just approximately) exclude in-window rows.
- Folders/links (no upstream date filter at all) are filtered in-memory against the request window before capping, to avoid truncating the correct window with an arbitrary page boundary.

## Forward path

The real event log (this ticket's original, deferred scope) is expected to swap in behind this same `ActivityEvent` contract later, without a client-visible change.

## Test coverage

No `.component.spec.ts` for the swapped `committee-overview.component.ts` — this repo has no Angular component-test runner wired up at all (confirmed: zero such files exist anywhere in the repo). Server-side coverage (`committee-activity.service.spec.ts`, `.controller.spec.ts`, `committee-activity-query.helper.spec.ts`) covers: merge ordering across sources, `since`/`limit`/cursor validation and pagination round-trips (composite-cursor tie handling, 30-item pagination-walk and 30-tied-timestamp regressions), per-leg failure isolation (one leg failing doesn't blank the others), `enable_voting=false` vote exclusion, and the meetings-leg runtime tripwire (positive + negative cases, each verified by mutation testing to actually discriminate).

## Accepted trade-offs (explicit decision, not oversights)

A full-branch reviewer sweep raised 5 "Important" findings (no Criticals). All 5 are deliberately left as documented trade-offs rather than fixed in this PR:

1. **Pagination machinery has no current consumer.** `page_token`, saturation detection (`anyLegSaturated`), and the composite cursor are real and tested, but the client only ever fetches page 1 today. This matches an earlier explicit choice in this PR to implement real timestamp-cursor pagination now rather than a stub, so a future "load more" UI needs no server changes.
2. **High comment-to-code ratio in `committee-activity.service.ts`**, reflecting the multi-round upstream-contract verification described above (meetings' `occurred_at` guarantee in particular). Left inline rather than split into a separate doc, since the comments are directly load-bearing for why specific code paths exist.
3. **`ActivityEventType` has no current consumers of its deferred variants** — it documents the forward-looking wire contract (including types with no v1 source) per the ticket's explicit goal of a future event-log swap without a contract change.
4. **`fetchCommittee` (the FGA-enforced committee lookup) runs concurrently with, not gating, the 6 data-fetch calls inside the same `Promise.all`.** No data leak — `Promise.all` discards every result on any rejection — but an unauthorized caller triggers ~9 upstream calls before the 403 resolves, rather than failing fast on the first. Accepted as a request-amplification cost, not a security issue.
5. **Net upstream call volume for the Overview page increased**, not decreased, versus the old stop-gap. The real win is boundedness — replacing what was previously an unbounded `fetchAllQueryResources` drain (documents leg) with capped, windowed per-leg fetches — not a reduction in call count.
6. **Vote-drawer click miss shows a toast instead of hydrating the drawer by UID.** `this.votes()` and the activity feed are independent, differently-windowed fetches, so a vote recent enough to appear in the feed can miss `this.votes()`'s own 100-item window; the click handler toasts "Vote unavailable" rather than opening the drawer seeded with just `voteUid`. Considered, not an oversight: `VoteResultsDrawerComponent`'s own `@if (vote(); as voteData)` has no `@else`, so a null-seeded drawer would render empty with no loading/not-found state until (or unless) its own re-fetch resolves — worse than the toast for the genuinely-missing case. Flagged by dealako's review as worth listing here for visibility (not blocking); revisit if a low-effort loading/`@else` branch on the shared drawer ever closes this gap.
7. **A `hasMore: false` response can still be non-terminal when access-control filters an entire fetched page to zero visible rows.** Confirmed against `lfx-v2-query-service`'s own source and documented contract: FGA access-checking runs *after* OpenSearch paginates, so `page_token` reflects the raw pre-filter page — a leg's single bounded fetch can come back with zero visible resources while `page_token` is still set, meaning authorized rows could exist on a later upstream page this endpoint never requests. `anyLegSaturated` already reflects this correctly via each leg's real `page_token`, but if every leg's page filters down to zero visible rows in the same response, there's no `lastPageItem` to anchor a cursor on, so `hasMore` forces `false` regardless. Closing this for real means looping a leg's own upstream call until a visible candidate or exhaustion — ruled out for v1 by the "bounded calls, no per-event follow-ups" requirement (the same constraint behind trade-off #5's call-volume note above). Flagged by Copilot; independently verified against `lfx-v2-query-service`'s `internal/service/resource_search.go` and `docs/query-service-contract.md`.

## Review feedback addressed

A full-branch reviewer sweep (Copilot, CodeRabbit, Cursor Bugbot) plus a human review from @dealako independently confirmed 5 minor issues, cross-checked against current code:

- **Vote status casing bug** — `mapVoteToEvent` now goes through `normalizePollStatus` before comparing to `PollStatus.ENDED`, matching `isVoteCalendarEventPast`'s existing pattern; a non-canonical-case `"Ended"` no longer misclassifies as `vote_opened` at the wrong timestamp.
- **Survey `occurred_at` for cutoff-driven closures** — a `SENT` survey whose cutoff has passed (no corresponding `last_modified_at` write) now stamps `survey_closed` at `survey_cutoff_date`, not the stale publish/update time.
- **Saturation detection now uses each leg's real upstream `page_token`**, not a `length >= fetchSize` row-count comparison — a leg returning a genuine final page no longer triggers a false-positive `hasMore`/cursor to an empty next page.
- **Meeting password now travels in the activity event's own payload**, not a lookup against the separately-windowed `pastMeetings()` signal — closes a real regression where a password-protected meeting recent enough for the feed but outside that signal's window silently lost its password on navigation.
- **Repeated query params (`?since=a&since=b`) are now rejected with a 400**, not silently treated as absent — Express's `qs` parser turns them into arrays, which previously restarted pagination / ignored `since` / fell back `page_size` to its default with no error surfaced.
- **`console.error` now logs a safe field subset** (`{ status, message }` for the HTTP fetch failure, `{ message }` for the mapping-step failure), not the raw error object, in the two call sites this PR introduced.
- **Survey leg no longer sends any upstream date narrowing** — a follow-up finding from Copilot on the cutoff-driven-closure fix above: `date_field: 'last_modified_at'` silently excluded exactly the rows a `since`-scoped request needs, since a cutoff-driven closure's `occurred_at` has no corresponding write to `last_modified_at`. Now relies entirely on the in-memory since/cursor pass; see the updated "Sort-order / date-field limitations" section above.
- **An `ACTIVE` vote whose `end_time` has already passed is now treated as closed**, not left as `vote_opened` at `creation_time` indefinitely — `mapVoteToEvent` now checks the deadline via a locally-reimplemented `isVoteDeadlinePast` (mirrors `isVoteCalendarEventPast`'s existing "deadline passed" semantics; reimplemented rather than imported since that helper lives in `meeting.utils.ts`, which pulls in an Angular-only `@angular/common/http` import this server file already avoids). Flagged by Cursor Bugbot as a follow-up to the casing fix above.

## Test plan

- [x] `yarn lint && yarn format:check && yarn test && yarn build` all pass
- [x] All commits DCO-signed (`--signoff`) and GPG-signed (verified via `git log --format='%G? %h %s'`)
- [x] Rebased cleanly onto latest `origin/main` (962 tests passing, up from 909 pre-rebase)
- [x] Full-branch reviewer sweep (general + Self Serve conventions + learnings) run against `origin/main`, findings documented above as accepted trade-offs
- [x] Review round from @dealako/Copilot/CodeRabbit/Cursor Bugbot addressed in `7a023002c`/`7ffddb7e1`/`d3cf4e86f`/`59f678f27` (see "Review feedback addressed" above); 976 tests passing, up from 962
- [ ] Manual verification in a running dev server: `GET /api/committees/:uid/activity` with no params, `since` only, `limit` bounds, pagination via `page_token`, and a voting-disabled committee
- [ ] Committee overview page renders the activity widget identically to before (same labels/icons/click targets), now via one BFF call instead of four

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -262,8 +262,14 @@ export class CommitteeOverviewComponent {
         // (`dashboard-meeting-card.component.ts`'s `initMeetingDetailQueryParams`) — a password-gated
         // committee meeting needs that param carried through, matching the established convention at
         // `meetings-dashboard.component.ts`'s `onCalendarEventClick`.
-        const meeting = this.pastMeetings().find((m) => m.id === action.meetingId);
-        void this.router.navigate(['/meetings', action.meetingId], meeting?.password ? { queryParams: { password: meeting.password } } : {});
+        //
+        // action.password comes from the source event's own payload, not a lookup against
+        // this.pastMeetings() — that signal is a separate, independently-windowed fetch, so a
+        // password-protected meeting recent enough to appear in the activity feed isn't guaranteed
+        // to be in pastMeetings()'s own window; a lookup-based password would silently be dropped on
+        // navigation for exactly the rows this matters most for (Copilot/Cursor Bugbot/dealako review
+        // on PR #1288).
+        void this.router.navigate(['/meetings', action.meetingId], action.password ? { queryParams: { password: action.password } } : {});
         break;
       }
       case 'vote-drawer': {
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -285,9 +285,10 @@ export class CommitteeOverviewComponent {
         break;
       }
       case 'external-url':
-        // buildActivityFeed only emits this action for urls that pass isValidUrl, but re-check at
-        // this sink rather than trusting that comment-enforced invariant — the union is shared
-        // shape LFXV2-1707's server-fed activity items will also construct.
+        // mapActivityEventsToFeedItems only emits this action for urls that pass isValidUrl, but
+        // re-check at this sink rather than trusting that comment-enforced invariant — the action
+        // is constructed client-side from a server-fed ActivityEvent, not a value this component
+        // controls end-to-end.
         if (isValidUrl(action.url)) {
           window.open(action.url, '_blank', 'noopener,noreferrer');
         }
@@ -296,8 +297,8 @@ export class CommitteeOverviewComponent {
         this.navigateToTab(action.tab);
         break;
       default:
-        // assertNeverSilent: this runs inside a DOM click handler, and LFXV2-1707 will feed
-        // server-constructed actions into this same union, so an unrecognized future `kind` (e.g. a
+        // assertNeverSilent: this runs inside a DOM click handler, and this action is constructed
+        // from a server-fed ActivityEvent (LFXV2-1707), so an unrecognized future `kind` (e.g. a
         // version-skewed client) should no-op rather than throw uncaught from a click. The
         // compile-time guarantee is unchanged — a new unhandled kind still fails to compile, since
         // the call below only type-checks if `action` has narrowed to `never`. console.error (not
@@ -573,8 +574,11 @@ export class CommitteeOverviewComponent {
             // CommitteeService.getCommitteeActivity already falls back to of([]) on failure, so this
             // catchError is a belt-and-suspenders guard against that coupling changing underneath
             // this component — without it, an error here would kill this toSignal source permanently
-            // and pin activityFeedLoading() on its skeleton for the rest of the session.
-            catchError(() => {
+            // and pin activityFeedLoading() on its skeleton for the rest of the session. Logs (unlike
+            // the service's own silent fallback) specifically because reaching this handler at all
+            // would mean that upstream guarantee broke — worth knowing about, not just surviving.
+            catchError((error) => {
+              console.error('Failed to map committee activity feed:', error);
               this.activityFeedLoading.set(false);
               return of<ActivityFeedItem[]>([]);
             })
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -290,10 +290,13 @@ export class CommitteeOverviewComponent {
         break;
       }
       case 'survey-drawer': {
-        // Same independent-fetch mismatch and same recoverable-in-principle/toast-instead
-        // trade-off as the vote-drawer case above — SurveyResultsDrawerComponent's own
-        // initSurvey() has the identical startWith/re-fetch shape and the identical missing
-        // `@else` on its template's `@if`.
+        // Same class of independent-fetch mismatch as the vote-drawer case above — this.surveys()
+        // and the activity feed are two separate server fetches with different windows (unlike
+        // votes, getSurveysByCommittee sets no page_size or order at all, so its own window is
+        // whatever the underlying endpoint defaults to, not a pinned page_size=100) — and the same
+        // recoverable-in-principle/toast-instead trade-off: SurveyResultsDrawerComponent's own
+        // initSurvey() has the identical startWith/re-fetch shape and the identical missing `@else`
+        // on its template's `@if` as VoteResultsDrawerComponent.
         const survey = this.surveys().find((s) => s.uid === action.surveyUid);
         if (survey) {
           this.selectedSurveyId.set(survey.uid);
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -574,9 +574,9 @@ export class CommitteeOverviewComponent {
             // CommitteeService.getCommitteeActivity already falls back to of([]) on failure, so this
             // catchError is a belt-and-suspenders guard against that coupling changing underneath
             // this component — without it, an error here would kill this toSignal source permanently
-            // and pin activityFeedLoading() on its skeleton for the rest of the session. Logs (unlike
-            // the service's own silent fallback) specifically because reaching this handler at all
-            // would mean that upstream guarantee broke — worth knowing about, not just surviving.
+            // and pin activityFeedLoading() on its skeleton for the rest of the session. Reaching this
+            // handler at all means the mapping step itself threw (not the upstream fetch, which the
+            // service already logs and degrades on its own) — worth its own log line, not just surviving.
             catchError((error) => {
               console.error('Failed to map committee activity feed:', error);
               this.activityFeedLoading.set(false);
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -613,8 +613,11 @@ export class CommitteeOverviewComponent {
             // and pin activityFeedLoading() on its skeleton for the rest of the session. Reaching this
             // handler at all means the mapping step itself threw (not the upstream fetch, which the
             // service already logs and degrades on its own) — worth its own log line, not just surviving.
-            catchError((error) => {
-              console.error('Failed to map committee activity feed:', error);
+            catchError((error: unknown) => {
+              // A safe subset, not the raw error object — this is a synchronous mapping failure
+              // (not an HTTP error, so no HttpErrorResponse.status here), and the underlying
+              // upstream fetch failure (if any) is already logged server-side with more structure.
+              console.error('Failed to map committee activity feed:', { message: error instanceof Error ? error.message : String(error) });
               this.activityFeedLoading.set(false);
               return of<ActivityFeedItem[]>([]);
             })
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -19,7 +19,6 @@ import { CommitteeMemberRole, PollStatus, SurveyStatus } from '@lfx-one/shared/e
 import {
   ActivityFeedItem,
   Committee,
-  CommitteeDocument,
   CommitteeMember,
   CommitteePendingActionRow,
   Meeting,
@@ -28,7 +27,7 @@ import {
   Survey,
   Vote,
 } from '@lfx-one/shared/interfaces';
-import { assertNeverSilent, buildActivityFeed, countVotingReps, getSurveyDisplayStatus, isValidUrl } from '@lfx-one/shared/utils';
+import { assertNeverSilent, countVotingReps, getSurveyDisplayStatus, isValidUrl, mapActivityEventsToFeedItems } from '@lfx-one/shared/utils';
 import { CommitteeService } from '@services/committee.service';
 import { FeatureFlagService } from '@services/feature-flag.service';
 import { MeetingService } from '@services/meeting.service';
@@ -38,7 +37,7 @@ import { getHttpErrorDetail } from '@shared/utils/http-error.utils';
 import { MessageService } from 'primeng/api';
 import { DialogService } from 'primeng/dynamicdialog';
 import { SkeletonModule } from 'primeng/skeleton';
-import { catchError, distinctUntilChanged, EMPTY, filter, finalize, forkJoin, of, switchMap, take, tap } from 'rxjs';
+import { catchError, distinctUntilChanged, EMPTY, filter, finalize, forkJoin, map, of, switchMap, take, tap } from 'rxjs';
 
 import { DashboardMeetingCardComponent } from '../../../dashboards/components/dashboard-meeting-card/dashboard-meeting-card.component';
 import { SurveyResultsDrawerComponent } from '../../../surveys/components/survey-results-drawer/survey-results-drawer.component';
@@ -113,10 +112,13 @@ export class CommitteeOverviewComponent {
   public meetingsLoading = signal(true);
   public votesLoading = signal(true);
   public surveysLoading = signal(true);
-  public documentsLoading = signal(true);
 
   // Loading state for past meetings (upcoming meetings loading comes in as the input above)
   public pastMeetingsLoading = signal(true);
+
+  // Activity feed loading — own dedicated signal (LFXV2-1707): the feed is now a single
+  // server-merged fetch, no longer derived from four separate source-loading signals.
+  public activityFeedLoading = signal(true);
 
   // Section-level fade-out for "My Pending Actions": true while the CSS collapse animation is in flight;
   // isSectionHidden removes the section from the DOM once the last vote/survey is resolved.
@@ -158,26 +160,14 @@ export class CommitteeOverviewComponent {
   public pastMeetings: Signal<PastMeeting[]> = this.initPastMeetings();
   public votes: Signal<Vote[]> = this.initVotes();
   public surveys: Signal<Survey[]> = this.initSurveys();
-  // Documents aren't otherwise loaded on Overview — fetched here solely to seed the activity feed
-  // (which shows at most 5 document rows). Not cheap: getCommitteeDocuments fans out server-side to
-  // /folders, /links, and an unbounded fetchAllQueryResources page-token loop over every
-  // committee_document (committee.service.ts), repeated on every silent refresh and again when the
-  // user opens the Documents tab (no shared cache). Accepted as a stop-gap for this ticket's scope —
-  // lifting the fetch to committee-view.component.ts to share across tabs, or bounding it with a
-  // page_size/order param, is real follow-up work, not a change this branch makes.
-  public documents: Signal<CommitteeDocument[]> = this.initDocuments();
 
   // Computed stats from fetched data
   public activeVotesCount: Signal<number> = computed(() => this.votes().filter((v) => v.status === PollStatus.ACTIVE).length);
 
   public openSurveysCount: Signal<number> = computed(() => this.surveys().filter((s) => getSurveyDisplayStatus(s) === SurveyStatus.OPEN).length);
 
-  // Activity feed stop-gap: merges the latest items across past meetings, votes, surveys, and
-  // documents into one time-ordered list. Replaced by the real activity stream in LFXV2-1707.
-  public activityFeedLoading: Signal<boolean> = computed(
-    () => this.pastMeetingsLoading() || this.votesLoading() || this.surveysLoading() || this.documentsLoading()
-  );
-
+  // Server-merged "Recent Activity" feed (LFXV2-1707): fetched as one call, mapped to the UI
+  // view-model client-side. See mapActivityEventsToFeedItems for the label/icon/action mapping.
   public activityItems: Signal<ActivityFeedItem[]> = this.initActivityItems();
 
   // Feature flag: WG Weekly Brief AI Assistant card
@@ -553,64 +543,45 @@ export class CommitteeOverviewComponent {
     );
   }
 
-  private initDocuments(): Signal<CommitteeDocument[]> {
-    // Documents only seed the activity feed, which the template never renders for a visitor
-    // (behind !isVisitor()) — skip the fetch rather than issue a GET nothing will display.
-    // Derived from one combined computed (not combineLatest over three separate toObservable()
-    // sources) so committee/myRoleLoading/isVisitor — all recomputed together in the same signal
-    // flush — can't glitch through an inconsistent intermediate tick that fires then immediately
-    // cancels a request. distinctUntilChanged on (uid, roleLoading, visitor) only dedupes a
-    // same-tuple re-emission (e.g. an unrelated field on committee() changing identity without
-    // affecting uid/roleLoading/visitor) — it does NOT suppress a silent refresh: myRoleLoading is
-    // `loading() || committeeRefreshing()` (committee-view.component.ts), so a refresh flips
-    // roleLoading true then false, which — like pastMeetings/votes/surveys, none of which have any
-    // guard here — legitimately cancels and re-issues the documents fetch and flips
-    // activityFeedLoading() back to true for the duration.
+  private initActivityItems(): Signal<ActivityFeedItem[]> {
+    // Skip-for-visitor + distinctUntilChanged shape mirrors this component's other committee-scoped
+    // fetches: the feed (and everything else in the enclosing !isVisitor() block) never renders for
+    // a visitor, so this fetch should never fire for one either. Derived from one combined computed
+    // (not combineLatest over separate toObservable() sources) so committee/myRoleLoading/isVisitor —
+    // all recomputed together in the same signal flush — can't glitch through an inconsistent
+    // intermediate tick that fires then immediately cancels a request.
     return toSignal(
       toObservable(computed(() => ({ committee: this.committee(), roleLoading: this.myRoleLoading(), visitor: this.isVisitor() }))).pipe(
         filter(({ committee }) => !!committee?.uid),
         distinctUntilChanged((a, b) => a.committee.uid === b.committee.uid && a.roleLoading === b.roleLoading && a.visitor === b.visitor),
         switchMap(({ committee, roleLoading, visitor }) => {
           if (roleLoading) {
-            // Role still resolving (e.g. mid silent-refresh) — hold both the documents list and
-            // documentsLoading exactly as they are (no signal write here). Safe because the fetch
-            // branch below clears documentsLoading with tap (fires only on emission), not finalize
-            // (also fires on switchMap-driven cancellation) — so cancelling an in-flight fetch to
-            // enter this branch can't have already cleared it out from under us.
+            // Role still resolving (e.g. mid silent-refresh) — hold state as-is; the fetch branch
+            // below clears activityFeedLoading with tap (fires only on emission), not finalize (also
+            // fires on switchMap-driven cancellation), so cancelling in-flight to enter this branch
+            // can't have already cleared it out from under us.
             return EMPTY;
           }
           if (visitor) {
-            this.documentsLoading.set(false);
-            return of<CommitteeDocument[]>([]);
+            this.activityFeedLoading.set(false);
+            return of<ActivityFeedItem[]>([]);
           }
-          this.documentsLoading.set(true);
-          // CommitteeService.getCommitteeDocuments already falls back to of([]) on failure today, so
-          // this catchError is a belt-and-suspenders guard against that coupling changing underneath
-          // this component (e.g. a future interceptor rethrow) — without it, an error here would
-          // propagate past a next-only tap, kill this toSignal source permanently, and pin
-          // activityFeedLoading() on its skeleton for the rest of the session.
-          return this.committeeService.getCommitteeDocuments(committee.uid).pipe(
-            tap(() => this.documentsLoading.set(false)),
+          this.activityFeedLoading.set(true);
+          return this.committeeService.getCommitteeActivity(committee.uid).pipe(
+            map((events) => mapActivityEventsToFeedItems(events, { votingEnabled: !!committee.enable_voting })),
+            tap(() => this.activityFeedLoading.set(false)),
+            // CommitteeService.getCommitteeActivity already falls back to of([]) on failure, so this
+            // catchError is a belt-and-suspenders guard against that coupling changing underneath
+            // this component — without it, an error here would kill this toSignal source permanently
+            // and pin activityFeedLoading() on its skeleton for the rest of the session.
             catchError(() => {
-              this.documentsLoading.set(false);
-              return of<CommitteeDocument[]>([]);
+              this.activityFeedLoading.set(false);
+              return of<ActivityFeedItem[]>([]);
             })
           );
         })
       ),
       { initialValue: [] }
-    );
-  }
-
-  private initActivityItems(): Signal<ActivityFeedItem[]> {
-    return computed(() =>
-      buildActivityFeed({
-        pastMeetings: this.pastMeetings(),
-        votes: this.votes(),
-        surveys: this.surveys(),
-        documents: this.documents(),
-        votingEnabled: !!this.committee().enable_voting,
-      })
     );
   }
 }
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -293,10 +293,14 @@ export class CommitteeOverviewComponent {
         // Different failure mode than the vote-drawer case above, same toast-instead remedy.
         // Unlike votes (a genuine window-size mismatch — getVotesByCommittee pins page_size=100),
         // the server's SurveyService.getSurveys drains every page via fetchAllQueryResources, so
-        // this.surveys() is the committee's complete survey set, not a windowed slice — the only
-        // realistic way this lookup misses is staleness between this fetch and the activity feed's
-        // independent one (a survey created/updated in the gap between the two requests), not a
-        // bounded-window gap. Same recoverable-in-principle/toast-instead trade-off regardless:
+        // this.surveys() is not a windowed slice — but "drains every page" isn't "always complete":
+        // fetchAllQueryResources defaults to failOnPartial=false (this call site doesn't override
+        // it), so a later page failing after retries returns what was fetched so far rather than
+        // throwing, and initSurveys' own catchError(() => of([])) yields an empty array on a
+        // failed/in-flight fetch regardless of drain completeness. So a miss here is normally
+        // fetch-timing staleness against the activity feed's independent request (the common case,
+        // since surveys() usually is complete) rather than a bounded-window gap like votes — but
+        // isn't guaranteed to be. Same recoverable-in-principle/toast-instead trade-off regardless:
         // SurveyResultsDrawerComponent's own initSurvey() has the identical startWith/re-fetch
         // shape and the identical missing `@else` on its template's `@if` as
         // VoteResultsDrawerComponent.
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -267,20 +267,29 @@ export class CommitteeOverviewComponent {
         break;
       }
       case 'vote-drawer': {
+        // this.votes() and the activity feed are two independent server fetches (the former
+        // page_size=100 by updated_at, the latter the 25 most recent by occurred_at) — a vote
+        // recent enough to appear in the feed isn't guaranteed to be within the former's window,
+        // so this lookup can miss. Surface that rather than a dead click with zero feedback.
         const vote = this.votes().find((v) => v.uid === action.voteUid);
         if (vote) {
           this.selectedVoteId.set(vote.uid);
           this.selectedVote.set(vote);
           this.voteDrawerVisible.set(true);
+        } else {
+          this.messageService.add({ severity: 'warn', summary: 'Vote unavailable', detail: 'This vote could not be found. Try the Votes tab instead.' });
         }
         break;
       }
       case 'survey-drawer': {
+        // Same independent-fetch mismatch as the vote-drawer case above.
         const survey = this.surveys().find((s) => s.uid === action.surveyUid);
         if (survey) {
           this.selectedSurveyId.set(survey.uid);
           this.selectedSurvey.set(survey);
           this.surveyDrawerVisible.set(true);
+        } else {
+          this.messageService.add({ severity: 'warn', summary: 'Survey unavailable', detail: 'This survey could not be found. Try the Surveys tab instead.' });
         }
         break;
       }
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -268,9 +268,17 @@ export class CommitteeOverviewComponent {
       }
       case 'vote-drawer': {
         // this.votes() and the activity feed are two independent server fetches (the former
-        // page_size=100 by updated_at, the latter the 25 most recent by occurred_at) — a vote
-        // recent enough to appear in the feed isn't guaranteed to be within the former's window,
-        // so this lookup can miss. Surface that rather than a dead click with zero feedback.
+        // page_size=100 by updated_at; the latter defaults to ACTIVITY_FEED_DEFAULT_PAGE_SIZE most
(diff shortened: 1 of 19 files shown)


apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,759 @@
...
+  }
+
+  /**
+   * One event per vote row (not two) — deliberate: this is an aggregation over current state, not
+   * a real event log, so only the vote's most recent lifecycle point is known. Keeps the feed's
+   * row count/rendering identical to the old client-side widget (one row per vote). A real event
+   * log would naturally emit both `vote_opened` and `vote_closed` as they occur.
+   */
+  private mapVoteToEvent(vote: Vote, committeeUid: string): VoteOpenedActivityEvent | VoteClosedActivityEvent {
+    const isClosed = vote.status === PollStatus.ENDED || !!vote.early_end_time;
```
dealako: [minor] Vote status casing bug

Issue: `isClosed` is derived from `vote.status === PollStatus.ENDED`, a raw string comparison.
Proof: `Vote.status` is documented to arrive with inconsistent casing — that's exactly why `normalizePollStatus` exists (`packages/shared/src/utils/poll.utils.ts:56`, exercised in `poll.utils.spec.ts` for `'Ended'`/`'ACTIVE'` mixed-case inputs) and why `meeting.utils.ts:607` uses `normalizePollStatus(vote.status) === PollStatus.ENDED` for the same ended check. This call site skips that normalization.
Why it matters: A non-canonical-case ended vote (e.g. `"Ended"`) is misclassified as `vote_opened` and timestamped at creation time instead of close time, corrupting feed ordering and `since` filtering for that row. Both Copilot and Cursor Bugbot independently flagged this on this PR.
Fix: `const isClosed = normalizePollStatus(vote.status) === PollStatus.ENDED || !!vote.early_end_time;`

apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,759 @@
...
+    const response = await this.microserviceProxy.proxyRequest<QueryServiceResponse<Survey> | null>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', query);
+
+    return (response?.resources ?? []).map((resource) => this.mapSurveyToEvent(resource.data, committeeUid));
+  }
+
+  private mapSurveyToEvent(survey: Survey, committeeUid: string): SurveyPublishedActivityEvent | SurveyClosedActivityEvent {
+    const displayStatus = getSurveyDisplayStatus(survey);
+    return {
+      type: displayStatus === SurveyStatus.CLOSED ? 'survey_closed' : 'survey_published',
+      occurred_at: firstValidTimestamp(survey.last_modified_at, survey.created_at),
```
dealako: [minor] Survey occurred_at ignores cutoff-driven closure

Issue: `occurred_at: firstValidTimestamp(survey.last_modified_at, survey.created_at)` is used regardless of why `displayStatus` is `CLOSED`.
Proof: `getSurveyDisplayStatus`/`getEffectiveSurveyStatus` (`survey.utils.ts:32-52`) can return `CLOSED` purely because `new Date() >= survey.survey_cutoff_date`, with no corresponding write to `last_modified_at`.
Why it matters: A survey that closes by cutoff (not by an explicit update) gets stamped with its publish/update time rather than its actual closure time — the feed can misorder the closure event or `since` filtering can drop a closure that genuinely occurred inside the requested window.
Fix: When `displayStatus === SurveyStatus.CLOSED` and the closure is cutoff-driven (`survey_cutoff_date` is in the past and no `end`-type field was updated), prefer `survey.survey_cutoff_date` as `occurred_at`.

apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,759 @@
...
+    // the caveat above), so once a page's fetch actually hits that bound, there could be more rows
+    // upstream than made it into `windowed` even though the merged pool itself isn't over `limit` —
+    // e.g. a single dominant source returning exactly `fetchSize` rows, several of which get
+    // dropped by the in-memory since/cursor/invalid-timestamp filters, can leave `windowed.length`
+    // at or under `limit` while real, unfetched rows remain beyond this page's upstream cutoff.
+    // Folders/links are excluded from this saturation check — they have no upstream page bound at
+    // all (fetched in full every call) and are already filtered+bounded correctly against the exact
+    // cursor inside fetchDocumentEvents, so a fetched length at `fetchSize` there reflects this
+    // leg's own internal cap, not an upstream truncation signal.
+    const anyLegSaturated =
```
dealako: [minor] Saturation detection uses row count, not the upstream page_token

Issue: `anyLegSaturated` treats `length >= fetchSize` as proof more upstream data exists, for meetings/votes/surveys/files.
Proof: query-service can return a full final page (exactly `fetchSize` rows) with no further `page_token` — row count alone doesn't distinguish that from a real truncation.
Why it matters: A false-positive `hasMore` emits a cursor that leads only to an empty extra page. Currently dormant in practice since (per the PR's own accepted trade-off #1) the client only fetches page 1 today, but this will surface silently once pagination gets a real consumer.
Fix: Thread each bounded leg's actual upstream `page_token`/has-more signal through instead of comparing to `fetchSize`, mirroring what's already done correctly for the files sub-leg's `saturated` flag.

apps/lfx-one/src/server/helpers/committee-activity-query.helper.ts
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Not `validation.helper.ts`'s `getStringQueryParam` — that module also imports
+ * `@lfx-one/shared/utils`, which pulls in Angular-only runtime code that Vitest can't resolve
+ * outside an Angular context (same issue documented in `activity-feed.utils.ts`). Inlined locally
+ * to keep this helper's unit test importable without mocking half of `validation.helper.ts`.
+ */
+function getStringQueryParam(req: Request, name: string): string | undefined {
+  const value = req.query[name];
+  return typeof value === 'string' ? value : undefined;
```
dealako: [minor] Repeated query params silently discarded instead of rejected

Issue: `getStringQueryParam` returns `undefined` for anything that isn't a string.
Proof: Express's default query parser is `qs`-based (confirmed via `express.urlencoded({ extended: true })` in `server.ts:109`, and no `app.set('query parser', ...)` override exists), so a repeated param like `?page_token=a&page_token=b` becomes an array, not a string — silently treated as absent.
Why it matters: `page_token` restarts pagination, `since` is ignored, and `page_size` falls back to its default, with no error surfaced to the caller — a silent behavior change rather than a loud 400. CodeRabbit flagged this independently.
Fix: Throw `ServiceValidationError` when `req.query[name]` is defined but not a string (including array values), and add regression tests for `page_token`/`since`/`page_size` array inputs, per CodeRabbit's suggested fix.

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
```diff
@@ -553,64 +574,48 @@ export class CommitteeOverviewComponent {
...
+            map((events) => mapActivityEventsToFeedItems(events, { votingEnabled: !!committee.enable_voting })),
+            tap(() => this.activityFeedLoading.set(false)),
+            // CommitteeService.getCommitteeActivity already falls back to of([]) on failure, so this
+            // catchError is a belt-and-suspenders guard against that coupling changing underneath
+            // this component — without it, an error here would kill this toSignal source permanently
+            // and pin activityFeedLoading() on its skeleton for the rest of the session. Reaching this
+            // handler at all means the mapping step itself threw (not the upstream fetch, which the
+            // service already logs and degrades on its own) — worth its own log line, not just surviving.
+            catchError((error) => {
+              console.error('Failed to map committee activity feed:', error);
```
dealako: [nit] Unstructured error object passed to console.error

Issue: `console.error('Failed to map committee activity feed:', error);` (and similarly in `committee.service.ts`'s `console.error('Failed to load committee activity feed:', error);`) logs the raw error object to the browser console.
Why it matters: Low real-world exposure — browser console only, and the underlying HTTP failure is already logged server-side with more structure — so not blocking, but worth trimming to a safe subset of fields (e.g. `{ status: error?.status }`) if this file is touched again.
Fix: Log a redacted/structured subset instead of the raw error object.

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
```diff
@@ -277,27 +267,58 @@ export class CommitteeOverviewComponent {
...
+        // no `@else` on `@if (vote(); as voteData)`, so a null seed renders an empty drawer with no
+        // loading or not-found state until (or unless) that fetch settles — worse than this toast
+        // for the genuinely-missing case, and out of this component's scope to add. Toast instead.
         const vote = this.votes().find((v) => v.uid === action.voteUid);
         if (vote) {
           this.selectedVoteId.set(vote.uid);
           this.selectedVote.set(vote);
           this.voteDrawerVisible.set(true);
+        } else {
+          this.messageService.add({ severity: 'warn', summary: 'Vote unavailable', detail: 'This vote could not be found. Try the Votes tab instead.' });
```
dealako, line 294: [question] Vote-drawer miss: toast vs. hydrate-by-UID

Copilot suggested hydrating the vote drawer by UID on a cache miss instead of toasting "unavailable" when a feed row's vote falls outside the independently-fetched 100-vote window. The inline comment above this block explains why toast-instead was chosen (the drawer's `@if (vote(); as voteData)` has no `@else`, so a null seed would render an empty drawer with no loading/error state — worse than the toast for the genuinely-missing case). That reads as a considered trade-off, not an oversight, so I'm not treating it as blocking — but it wasn't listed alongside the PR body's other 5 "Accepted trade-offs," so flagging for visibility: worth adding there, or revisiting if a low-effort `@else`/loading branch on the drawer would close the gap.

dealako (changes requested):
Hey @manishdixitlfx — thanks for the extremely thorough writeup on this one. The PR body's own "Accepted trade-offs" section and the in-code rationale (the meetings `occurred_at` tripwire, the vote/survey drawer-miss toast reasoning, the saturation-detection caveat) made this a genuinely easy PR to review — most of the hard judgment calls are already made and documented, which is exactly what you want from a BFF this shaped. Server-side test coverage (cursor round-tripping, composite tie-breaking, per-leg degradation, the meetings tripwire) is strong.

I ran this through a 3-way parallel review (security/privacy, correctness/perf/tests, style/API/docs) and cross-checked it against the existing Copilot, CodeRabbit, and Cursor Bugbot findings, verifying each against current code rather than taking any bot's word for it.

### 🔴 Blocking: 0 issues

### 🟡 Minor: 5 issues

1. **Vote status casing bug** (`committee-activity.service.ts:555`) — `mapVoteToEvent` compares `vote.status === PollStatus.ENDED` directly instead of going through `normalizePollStatus` (which exists precisely because `Vote.status` arrives with inconsistent casing — see `meeting.utils.ts:607` and `poll.utils.spec.ts`). A non-canonical-case `"ENDED"` value is misclassified as `vote_opened` and stamped with creation time instead of close time. **Both Copilot and Cursor Bugbot independently flagged this** — I traced it and confirm it's real and unaddressed in the current diff.
2. **Survey `occurred_at` doesn't use the cutoff date for cutoff-driven closures** (`committee-activity.service.ts:606`) — `mapSurveyToEvent` derives `occurred_at` from `firstValidTimestamp(survey.last_modified_at, survey.created_at)`, but `getSurveyDisplayStatus` can report `CLOSED` purely because `survey_cutoff_date` has passed (see `survey.utils.ts:32-52`), with no corresponding update to `last_modified_at`. A survey that closes by cutoff gets timestamped at its publish/update time, not its actual closure — this can misorder the feed and cause `since` filtering to drop a closure that happened inside the requested window. Confirmed via Copilot's finding.
3. **Saturation detection uses row count instead of preserving the upstream `page_token`** (`committee-activity.service.ts:344-351`) — `anyLegSaturated` is `length >= fetchSize` per leg, but query-service can return a full final page with no further `page_token`. Confirmed via Copilot's finding. Practically dormant today since (per your own accepted trade-off #1) the client only ever fetches page 1 — but worth a tracking note or a fix before pagination gets a real consumer, since the false-positive `hasMore` would otherwise ship silently.
4. **Password not carried when navigating to a password-protected past meeting from the activity feed** (`committee-overview.component.ts:265-266`) — the click handler looks up the password via `this.pastMeetings().find(...)`, but the activity feed is now an independent fetch from `pastMeetings()`. If a password-protected meeting appears in the feed but isn't in the (separately-windowed) `pastMeetings()` signal, the navigation goes out with no password and the join page 400s/403s into what looks like a not-found. Confirmed via Copilot's finding — this is a real, reachable regression versus the old single-source widget.
5. **Repeated query params are silently discarded rather than rejected** (`committee-activity-query.helper.ts:16-19`) — `getStringQueryParam` returns `undefined` for anything that isn't a `string`, but Express's default `qs`-based query parser (confirmed via `express.urlencoded({ extended: true })` in `server.ts:109`, which implies the `qs` parser is active for `req.query` too) turns a repeated param like `?page_token=a&page_token=b` into an array. Today this means `page_token`/`since`/`page_size` silently reset to "absent" instead of erroring — low likelihood of a real caller doing this, but a validation gap CodeRabbit correctly flagged. Recommend a `ServiceValidationError` for non-string-but-defined values, per CodeRabbit's suggested fix.

### ⚪ Nit: 1 issue

6. **Unstructured error objects passed to `console.error`** (`committee.service.ts` and `committee-overview.component.ts`, from my security pass) — logs full error objects to the browser console. Low real-world exposure (browser console, not server logs; the underlying HTTP errors are already logged server-side with more structure), so I'm not blocking on it, but worth trimming to a safe subset of fields if you touch this again.

### ❔ Question: 1 item

7. **Vote-drawer miss handling: toast vs. hydrate-by-UID** (`committee-overview.component.ts:288`) — Copilot suggested hydrating the vote drawer by UID on a cache miss instead of toasting "unavailable" when a feed row's vote falls outside the independently-fetched 100-vote window. I read the inline comment explaining why toast-instead was chosen (the drawer template has no `@else` on `@if (vote(); as voteData)`, so a null seed would render an empty drawer with no loading/error state — worse than the toast). That's a reasoned trade-off, not an oversight, so I'm not treating it as a blocking finding — but it wasn't called out in the PR body's "Accepted trade-offs" list alongside the other 5, so flagging it here in case you want to add it there for visibility, or if there's since been a lower-effort fix (e.g., a loading/`@else` branch on the drawer) that would close the gap.

### On the fetchCommittee/Promise.all concurrency question

My security pass flagged that `fetchCommittee` (FGA-enforced) runs concurrently with, rather than gating, the 6 data-fetch legs — but this is explicitly called out in the PR body as accepted trade-off #4 ("Accepted as a request-amplification cost, not a security issue"). Agreed with your reasoning there; no action needed.

### Final decision

🔴 **Needs changes before approval** — none of the 5 minor findings are blocking on their own, but items 1 and 2 are confirmed data-correctness bugs (event misclassification / mistimed closures) independently caught by two different bots, and item 4 is a real reachable UX regression for password-protected meetings. Worth a fix pass before merge; happy to re-review once addressed.

apps/lfx-one/src/server/services/committee-activity.service.ts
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7a023002c — `mapVoteToEvent` now normalizes `vote.status` via `normalizePollStatus` before comparing to `PollStatus.ENDED`, matching `isVoteCalendarEventPast`'s existing pattern (`meeting.utils.ts:607`). Added a mixed-case regression test in `committee-activity.service.spec.ts`.

apps/lfx-one/src/server/services/committee-activity.service.ts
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7a023002c — `mapSurveyToEvent` now stamps `occurred_at` at `survey_cutoff_date` when the closure is cutoff-driven (SENT past its cutoff, no `last_modified_at` write), falling back to `last_modified_at`/`created_at` otherwise. Added both a positive and a negative regression test (the negative one guards against over-applying the cutoff timestamp to an explicitly-closed survey).

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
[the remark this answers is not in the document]
manishdixitlfx: Thanks for the suggestion. Per the PR's own inline rationale (the drawer's `@if (vote(); as voteData)` has no `@else`, so a null-seeded drawer would render empty with no loading/not-found state), the toast-instead choice is being kept as a deliberate trade-off rather than switched to hydrate-by-UID — now explicitly listed as trade-off #6 in the PR description for visibility. Worth revisiting if a low-effort loading/`@else` branch on the shared drawer ever closes this gap.

apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,759 @@
...
+      pastMeetingEvents.length >= fetchSize ||
+      (votingEnabled && voteEvents.length >= fetchSize) ||
+      surveyEvents.length >= fetchSize ||
+      documentResult.saturated;
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7a023002c — `anyLegSaturated` now threads each leg's real upstream `page_token` (meetings/votes/surveys/files) instead of comparing row count to `fetchSize`, so a genuine final page no longer triggers a false-positive `hasMore`/cursor. Added both a positive (`page_token` present) and negative (`fetchSize` rows, no `page_token`) regression test.

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
```diff
@@ -553,64 +574,48 @@ export class CommitteeOverviewComponent {
...
-          // propagate past a next-only tap, kill this toSignal source permanently, and pin
-          // activityFeedLoading() on its skeleton for the rest of the session.
-          return this.committeeService.getCommitteeDocuments(committee.uid).pipe(
-            tap(() => this.documentsLoading.set(false)),
-            catchError(() => {
-              this.documentsLoading.set(false);
-              return of<CommitteeDocument[]>([]);
+          this.activityFeedLoading.set(true);
+          return this.committeeService.getCommitteeActivity(committee.uid).pipe(
+            map((events) => mapActivityEventsToFeedItems(events, { votingEnabled: !!committee.enable_voting })),
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7a023002c — the meeting's password now travels in the activity event's own payload (`MeetingHeldActivityEvent.payload.password`) instead of a lookup against the separately-windowed `pastMeetings()` signal, so a password-protected meeting outside that signal's window no longer silently loses its password on navigation.

apps/lfx-one/src/server/helpers/committee-activity-query.helper.ts
```diff
@@ -0,0 +1,111 @@
...
+}
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7a023002c — `getStringQueryParam` now throws `ServiceValidationError` when `req.query[name]` is defined but not a string (including repeated/array values), for `since`/`page_size`/`page_token`. Added regression tests for all three.

apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,759 @@
...
+    const occurredAt = isClosed
+      ? firstValidTimestamp(vote.early_end_time, vote.end_time, vote.last_modified_time, vote.creation_time)
+      : firstValidTimestamp(vote.creation_time, vote.last_modified_time);
+    return {
+      type: isClosed ? 'vote_closed' : 'vote_opened',
+      occurred_at: occurredAt,
+      committee_uid: committeeUid,
+      payload: { vote_uid: vote.uid, name: vote.name, status: vote.status },
+    };
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 7a023002c — see the reply on the matching Copilot/dealako threads for the same finding; `mapVoteToEvent` now normalizes `vote.status` via `normalizePollStatus` before the `ENDED` comparison.

apps/lfx-one/src/server/services/committee-activity.service.ts
manishdixitlfx: Fixed in 7a023002c — `normalizePollStatus(vote.status) === PollStatus.ENDED`, matching your suggested fix exactly. Added a mixed-case regression test.

apps/lfx-one/src/server/services/committee-activity.service.ts
manishdixitlfx: Fixed in 7a023002c — `occurred_at` now uses `survey_cutoff_date` when the closure is cutoff-driven (SENT status, cutoff passed, no `response_status` closed sentinel), matching your suggested fix. Added both a positive and a negative regression test to guard against over-applying the cutoff timestamp to an explicitly-closed survey.

apps/lfx-one/src/server/services/committee-activity.service.ts
manishdixitlfx: Fixed in 7a023002c — meetings/votes/surveys/files now report `saturated` based on their real upstream `page_token` (`getMeetings`/`getVotes` already return it via `PaginatedResponse`; the survey/files legs' raw `proxyRequest` response also carries it), not a row-count comparison. Added a paired positive/negative regression test.

apps/lfx-one/src/server/helpers/committee-activity-query.helper.ts
manishdixitlfx: Fixed in 7a023002c per your and CodeRabbit's suggested fix — `getStringQueryParam` now throws `ServiceValidationError` for a defined-but-non-string value. Added tests for repeated `since`/`page_size`/`page_token`.

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
manishdixitlfx: Fixed in 7ffddb7e1 — logging `{ status, message }` for the HTTP fetch failure in `committee.service.ts` and `{ message }` for the mapping-step failure in `committee-overview.component.ts`, instead of the raw error object.

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
manishdixitlfx: Added as trade-off #6 in the PR description, per your suggestion — the toast-instead reasoning stands as a deliberate choice (no `@else`/loading state on the shared drawer template), not an oversight. No code change; flagged there for visibility and as a candidate if a low-effort loading branch on the drawer ever becomes worth it.

manishdixitlfx:
## Review Feedback Addressed

Commits: 7a023002c, 7ffddb7e1

### Changes Made
- **committee-activity.service.ts**: `mapVoteToEvent` normalizes `vote.status` via `normalizePollStatus` before comparing to `PollStatus.ENDED` (per @dealako, Copilot, Cursor Bugbot)
- **committee-activity.service.ts**: `mapSurveyToEvent` stamps `occurred_at` at `survey_cutoff_date` for a cutoff-driven closure (per @dealako, Copilot)
- **committee-activity.service.ts**: saturation detection now threads each leg's real upstream `page_token` instead of a row-count comparison (per @dealako, Copilot)
- **activity-event.interface.ts, activity-feed.interface.ts, activity-feed.utils.ts, committee-activity.service.ts, committee-overview.component.ts**: meeting password now travels in the activity event's own payload instead of a lookup against the separately-windowed `pastMeetings()` signal (per @dealako, Copilot, Cursor Bugbot)
- **committee-activity-query.helper.ts**: `getStringQueryParam` now rejects a repeated (array-valued) `since`/`page_size`/`page_token` with a 400 (per @dealako, CodeRabbit)
- **committee.service.ts, committee-overview.component.ts**: `console.error` now logs a safe field subset instead of the raw error object (per @dealako)

### No Change Needed (documented instead)
- **committee-overview.component.ts:288** (vote-drawer toast vs. hydrate-by-UID): the toast-instead reasoning is a deliberate trade-off (the shared drawer template has no `@else`/loading state), not an oversight — added as trade-off #6 in the PR description per @dealako's suggestion, rather than implemented as a code change (asked by Copilot, @dealako)

### Test coverage added
Regression tests for all 5 fixed issues, including paired positive/negative cases for the saturation and survey-cutoff fixes to guard against over-broad mutations. 972 tests passing (up from 962).

### Threads Resolved

apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,819 @@
...
+      page_size: fetchSize,
+      sort: 'updated_desc',
+      // Best-effort narrowing (see the filter-dimension paragraph in getCommitteeActivity's
+      // fetchSize comment) — occurred_at is firstValidTimestamp(survey.last_modified_at,
+      // survey.created_at), so a never-modified survey (last_modified_at is the Go zero-date,
+      // created_at is the real value) could be excluded upstream on this field before the
+      // in-memory backstop ever sees it.
+      ...(hasWindow && { date_field: 'last_modified_at' }),
+      ...(since && { date_from: since }),
+      ...(before && { date_to: before }),
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch — confirmed. Fixed in d3cf4e86f: `fetchSurveyEvents` now sends no `date_field`/`date_from`/`date_to` at all, relying entirely on the in-memory since/cursor pass for correctness. Unlike votes/files, no field on the survey row bounds a cutoff-driven `occurred_at` from below, so any upstream date narrowing here was unsound rather than just approximate. Added a regression test reproducing your exact since-window scenario (last_modified_at before `since`, cutoff-driven occurred_at after it) plus a test asserting no date params reach this leg.

manishdixitlfx:

Commit: d3cf4e86f

### Changes Made
- **committee-activity.service.ts**: `fetchSurveyEvents` no longer sends `date_field`/`date_from`/`date_to` upstream at all — the previous `date_field: 'last_modified_at'` narrowing silently excluded a cutoff-driven survey closure whenever `since` fell between its `last_modified_at` and its real (cutoff-driven) `occurred_at`, since that transition has no corresponding upstream write. Relies entirely on the in-memory since/cursor pass now (asked by copilot-pull-request-reviewer[bot], a follow-up finding on the cutoff-driven-closure fix from round 1).

### Test coverage added
A regression test reproducing the exact since-window scenario from the review comment (survey last modified before `since`, cutoff-driven closure after it — now correctly included), plus a test asserting no date params reach the survey leg.

974 tests passing (up from 972), lint/format/build clean.

### Threads Resolved
1 of 1 new unresolved thread addressed in this iteration.

apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,829 @@
...
+    // (poll.utils.ts's normalizePollStatus doc comment; poll.utils.spec.ts exercises 'Ended'/'ACTIVE'
+    // mixed-case inputs), which is exactly why isVoteCalendarEventPast (meeting.utils.ts) goes
+    // through the same normalization for its own ended check rather than comparing vote.status
+    // directly. Skipping it here misclassified a non-canonical-case "ENDED" vote as vote_opened,
+    // stamped at creation time instead of close time — caught independently by Copilot and Cursor
+    // Bugbot, confirmed against current code.
+    const isClosed = normalizePollStatus(vote.status) === PollStatus.ENDED || !!vote.early_end_time;
+    const occurredAt = isClosed
+      ? firstValidTimestamp(vote.early_end_time, vote.end_time, vote.last_modified_time, vote.creation_time)
+      : firstValidTimestamp(vote.creation_time, vote.last_modified_time);
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 59f678f27 — `mapVoteToEvent` now also checks `isVoteDeadlinePast(vote.end_time)`, matching `isVoteCalendarEventPast`'s established deadline-passed semantics elsewhere in the app (reimplemented locally rather than imported, since that helper lives in `meeting.utils.ts` which pulls in an Angular-only `@angular/common/http` import this server file already avoids). Added both a positive (past end_time) and negative (future end_time) regression test.

apps/lfx-one/src/server/services/committee-activity.service.ts
```diff
@@ -0,0 +1,829 @@
...
+    // Every item in `page` now has a real timestamp (windowed already dropped the ones that
+    // don't), so the last page item is always a valid cursor position — no need to search backward
+    // for one. `hasMore` still requires a real `lastPageItem` to anchor a cursor on — if saturation
+    // is detected but this page came back empty (every fetched row got filtered out), there is no
+    // position left to resume from, so this degrades to the same "nothing more to report" outcome
+    // as before; an over-fetch this narrow (an entire upstream page filtered to zero matches) is
+    // accepted as a known residual rather than solved by a second round-trip within this call.
+    const lastPageItem = page.at(-1);
+    const hasMore = (windowed.length > limit || anyLegSaturated) && !!lastPageItem;
+    const pageToken = hasMore && lastPageItem ? encodeActivityPageToken({ before: lastPageItem.event.occurred_at, key: lastPageItem.key }) : undefined;
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed — verified directly against `lfx-v2-query-service`'s `internal/service/resource_search.go` and `docs/query-service-contract.md`: FGA access-checking does run after OpenSearch paginates, so `page_token` can survive a page that filters down to zero visible resources, exactly as described. The full fix (looping a leg's own upstream call until a visible candidate or exhaustion) conflicts with this ticket's explicit "bounded calls, no per-event follow-ups" v1 scope, so rather than implement a bounded retry loop here, this is documented as accepted trade-off #7 in the PR description and the in-code comment is sharpened with the confirmed upstream mechanism (see `getCommitteeActivity`'s `hasMore`/`lastPageItem` comment, 59f678f27). No code change; happy to revisit if this proves to matter in practice.

manishdixitlfx:
## Review Feedback Addressed (round 3)

Commit: 59f678f27

### Changes Made
- **committee-activity.service.ts**: `mapVoteToEvent` now also treats an `ACTIVE` vote whose `end_time` has already passed as closed, matching `isVoteCalendarEventPast`'s established deadline semantics (asked by cursor[bot])

### No Change Needed (documented instead)
- **committee-activity.service.ts:387** (saturated leg can return zero visible resources due to post-pagination FGA filtering): confirmed real against `lfx-v2-query-service`'s source and documented contract. The full fix conflicts with this ticket's "bounded calls, no per-event follow-ups" v1 scope, so documented as accepted trade-off #7 in the PR description with the confirmed upstream mechanism, rather than implemented as a bounded retry loop (flagged by copilot-pull-request-reviewer[bot])

### Test coverage added
Regression tests for the vote-deadline fix (positive + negative case). Also fixed the `vote()` test fixture's default `end_time`, which had drifted into the past relative to today's date and was causing 8 unrelated tests to fail once the deadline check existed — pinned to a far-future date so it stays valid.

976 tests passing (up from 974), lint/format/build clean.

### Threads Resolved
2 of 2 new unresolved threads addressed in this iteration.

dealako (approved):
Hey @manishdixitlfx — thanks for driving this through to a clean state. This moves committee "Recent Activity" aggregation from client-side to a paginated BFF endpoint fanning out across meetings, votes, surveys, and documents (files/folders/links), with cursor-based pagination and per-leg graceful degradation. The PR description's own "Accepted trade-offs" section and the inline rationale throughout `committee-activity.service.ts` (the fetchSize/saturation caveats, the meetings `occurred_at` tripwire, the vote/survey status-derivation notes) made this genuinely easy to review — the reasoning is on the page, not just in your head.

I ran a fresh three-dimension pass (security/privacy, correctness/performance/tests, style/API/docs) independent of the prior rounds.

**Security & Data Privacy:** Clean. Auth is entirely inherited from upstream (bearer token forwarded through FGA-enforced `proxyRequest` on all 7 fan-out calls), error logging sticks to the safe-subset pattern established in `7ffddb7e1`, inputs are validated (400 on malformed `since`/`page_size`/`page_token` per the `d3cf4e86f`/`7a023002c` fixes), and the response carries `Cache-Control: no-store`. No PII leakage in logs, errors, or fixtures.

**Correctness/Performance/Tests:** No surviving issues. I want to flag that I looked hard at three candidate findings and dropped all three after checking the code directly:
- A cosmetic `:` vs `-` separator difference between the server cursor key and the client `@for` tracking key — different contexts, not meant to match, no behavioral impact.
- The high comment-to-code ratio in `committee-activity.service.ts` — already called out in the PR's own "Accepted trade-offs" section as a deliberate choice.
- Whether folders/links contribute to `anyLegSaturated` — they deliberately don't, and it's correct: the comment block at lines 807–814 explains folders/links have no upstream page bound at all (fetched in full every call, then filtered+bounded locally against the exact cursor), so hitting `fetchSize` there reflects the leg's own internal cap, not an under-fetch. This isn't missing signal, it's already reasoned through in-line.

**Style/API/Docs:** Clean. License headers present, shared types/constants correctly placed (`@lfx-one/shared`, `interfaces/` vs `constants/`), no nested ternaries, error-handling pattern consistent with the `7ffddb7e1` fix across all catch sites, comments explain non-obvious "why" (upstream contract quirks, verified against the query-service source) rather than restating "what."

**Revision tracking:** This branch already absorbed a full round of review (mine, plus Copilot and Cursor Bugbot) across commits `7a023002c`, `7ffddb7e1`, `d3cf4e86f`, and `59f678f27` — vote status casing normalization, survey `occurred_at` cutoff-driven closure (including the `since`-narrowing bypass Copilot caught in `d3cf4e86f`), saturation detection switched from row-count to real upstream `page_token`, malformed query-param rejection, unstructured `console.error` cleanup, and the ACTIVE-vote-past-deadline closure Cursor flagged in `59f678f27`. All verified fixed in the current diff — no regressions.

One thread remains genuinely open and unresolved: Copilot's note on `packages/shared/src/interfaces/activity-event.interface.ts:26` that the wire contract has no immutable per-event identifier, so a future move to a real event log (where one entity could emit multiple lifecycle events, e.g. both `vote_opened` and `vote_closed`) would produce duplicate cursor/UI keys. Under the current one-event-per-entity-per-fetch model this isn't a live bug, but it's unaddressed and worth an explicit decision (add `event_id` now vs. defer until the event-log swap is actually on the table) rather than leaving it silent.

### Summary

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 1 — immutable `event_id` for the wire contract, open and unresolved (Copilot, `activity-event.interface.ts:26`)

### Decision: ✅ Approved with comments

Nice work landing a genuinely complex fan-out endpoint with this much rigor on the edge cases.

PR #1290, opened 2026-08-02
feat(engagement): members table + overview engagement UI (LFXV2-1705)

manishdixitlfx:
## Summary

Adds the LFXV2-1705 UI slice for committee member engagement/attendance, behind the `wg-engagement-metrics` flag (default off, zero DOM/zero fetch when disabled). The BFF endpoint and shared contract (`GET /api/committees/:uid/engagement?window=30d|90d|ytd`) already shipped in #1268.

**Members tab**
- "Meetings" column: personal `attended/invited` for the selected window (em-dash when unavailable)
- Engagement classification chip (High/Medium/Low/Inactive/Emeritus) with accessible tooltip context for Chair/Vice Chair/LF Staff seats
- "At Risk" quick filter (Low, or Inactive with real invites) — hidden when data is degraded, auto-resets if the underlying data disappears mid-selection

**Overview tab**
- New engagement summary card: attendance rate, active members, at-risk count
- 30d/90d/YTD window selector shared with the Members tab (single page-level fetch, state survives tab switches)
- Freshness label, "Sample data" marker when `data_source: 'mock'`

**Degrade paths** (fetch failure incl. the expected 403 for non-auditor callers, or `data_available: false` while the live dbt model isn't deployed yet): calm "unavailable"/"coming soon" states, never an error — the member roster stays fully functional either way. Visitors never trigger the fetch.

## Changes
- `packages/shared`: engagement display utils + constants (window options, classification→severity map), `atRisk` filter chip type, `TagComponent` tooltip support (with combined value+tooltip accessible name)
- `CommitteeService.getCommitteeEngagement()` 
- `committee-view` / `committee-members` / `committee-overview` wiring, new `committee-engagement-summary` component
- Playwright e2e: flag gating, all classification tiers, At Risk filtering, window refetch, degraded/failed responses, mock marker, structural `data-testid` contract

## Test plan
- `yarn lint` / `yarn format` / `yarn test` (912 unit+shared specs) / `yarn build` all green
- Playwright specs cover flag on/off, degrade paths, and window-state sharing (`apps/lfx-one/e2e/committee-engagement{,-robust}.spec.ts`)
- Manually verify against `ENGAGEMENT_BACKEND=mock` locally (live mode currently always degrades — the dbt model deploy is tracked separately)

## Review history
Went through several rounds addressing CodeRabbit/Copilot/Cursor Bugbot/dealako feedback: visitor/role-gated fetch, At-Risk chip reset on data loss, tooltip keyboard-focus + accessible-name fixes, window reset per committee navigation, `tap` vs `finalize` loading-state timing, and precomputed row view-models (no template method calls).


apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -130,12 +140,20 @@ export class CommitteeViewComponent {
   public membersLoading = signal<boolean>(true);
   public invitesLoading = signal<boolean>(true);
   public joiningOrLeaving = signal(false);
+  // Engagement rollup (LFXV2-1705): shared window state so the Members table and the Overview
+  // summary stay in sync across tab switches (tab panels unmount in the @switch below).
+  public engagementWindow = signal<CommitteeEngagementWindow>(COMMITTEE_ENGAGEMENT_DEFAULT_WINDOW);
```
dealako: [blocking] `engagementWindow` not reset on committee change

Issue: `engagementWindow` is a plain `signal<CommitteeEngagementWindow>(COMMITTEE_ENGAGEMENT_DEFAULT_WINDOW)` with no reset when the route's committee id changes.
Proof: Angular reuses this component instance across `/groups/:id` navigations (same as the adjacent `activeTab` linked signal Copilot references). `engagementWindow` is only ever written by `.set(window)` at line 319 (user selection) — nothing keys it back to `committeeId()`. Selecting `YTD` on committee A and then navigating to committee B keeps `engagementWindow() === 'ytd'`, so B's `initEngagement` fires with the wrong window from the first render, with no error/indicator that it's not the default.
Why it matters: silently shows the wrong window's attendance rollup for a different committee — a data-correctness bug, not just a UX inconsistency (agrees with Copilot's finding on this line).
Fix: reset `engagementWindow` to `COMMITTEE_ENGAGEMENT_DEFAULT_WINDOW` whenever `committeeId()` changes, the same way the adjacent `activeTab` handling resets per-committee state, while still allowing writes across tab switches within the same committee.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +783,67 @@ export class CommitteeViewComponent {
...
+          // Members tab isn't visible to them at all) — skip the guaranteed-403 fetch entirely,
+          // matching initDocuments' "don't issue a GET nothing will display" precedent.
+          if (visitor) {
+            this.engagementLoading.set(false);
+            return of(null);
+          }
+          this.engagementLoading.set(true);
+          // Errors (including the expected 403 for non-auditor members) resolve to null inside the
+          // service — the tabs degrade to their "attendance unavailable" states, roster unaffected.
+          return this.committeeService.getCommitteeEngagement(uid, window).pipe(finalize(() => this.engagementLoading.set(false)));
```
dealako: [minor] Loading flag cleared on cancelled engagement request

Issue: when `roleLoading` flips true mid-refresh, `switchMap` returns `EMPTY` — but the previous inner observable (the in-flight engagement HTTP call) gets unsubscribed, running its `finalize(() => this.engagementLoading.set(false))`.
Proof: this sets `engagementLoading` to `false` while `engagement()` still holds stale/no data for the current tuple, so between the cancellation and the next real emission the Overview card and Members table render their "unavailable"/em-dash states instead of the skeleton — the same class of bug `initDocuments` in `committee-overview.component.ts` avoids by clearing loading via `tap` on emission only (agrees with Cursor Bugbot's finding).
Why it matters: a brief but visible flicker to an incorrect empty state during every silent role refresh, which given the comment above about `membersRefresh` firing frequently (roster mutations), isn't a rare case.
Fix: move the loading-clear off `finalize` and onto successful emission (mirroring `initDocuments`), or gate the unavailable-state rendering on more than just `engagementLoading() === false`.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -149,6 +170,42 @@
...
+                  @if (engagementRow && engagementDataAvailable()) {
+                    @let engagementContext = getEngagementContext(engagementRow);
+                    <span
+                      class="inline-flex items-center"
+                      [pTooltip]="engagementContext"
+                      tooltipPosition="top"
+                      tooltipEvent="both"
+                      [attr.tabindex]="engagementContext ? 0 : null"
+                      [attr.role]="engagementContext ? 'note' : null"
+                      [attr.aria-label]="engagementContext || null">
```
dealako: [minor] Tooltip focus target mismatch on the engagement chip

Issue: `pTooltip`/`tabindex="0"` are declared on the wrapping `<span>`, but the tag it wraps is `<lfx-tag>`, which renders PrimeNG's `p-tag` (host class `p-component`).
Proof: PrimeNG's tooltip directive with `tooltipEvent="both"` binds its focus/blur listeners to the nearest `.p-component` descendant, not the host element the directive sits on (per CodeRabbit's PrimeNG source dig). Since `<lfx-tag>` is that descendant here, Tab-focusing the `<span>` (which has `tabindex=0`) does not fire the listener bound to the nested tag, so the tooltip never appears on keyboard focus — only on hover.
I checked the two other locations CodeRabbit flagged in `committee-engagement-summary.component.html` (lines 40-52 and 53-65): those don't have this bug — `pTooltip`/`tabindex` are on the same plain `<div>` with no `.p-component` child, so the directive's own host is the focus target. Only this members-table location is actually affected; worth noting since this PR's prior review round (`32522b29` — "complete tooltip a11y") already fixed the Overview card pairing but didn't catch this one.
Why it matters: keyboard-only users get no attendance context tooltip on the members table's engagement chip, an accessibility regression relative to the Overview card's already-fixed pattern.
Fix: move `pTooltip`, `tooltipEvent`, `tooltipPosition`, and the `tabindex`/`role`/`aria-label` attrs onto `<lfx-tag>` itself (or expose a `styleClass`/host binding on `TagComponent` so the tooltip host and the `.p-component` element are the same node).

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -149,6 +170,42 @@
                 <span class="text-sm font-inter text-gray-500">{{ member.voting?.status || 'None' }}</span>
               </td>
             }
+            @if (engagementEnabled()) {
+              <td>
+                @if (engagementLoading()) {
+                  <p-skeleton width="2.5rem" height="0.875rem" />
+                } @else {
+                  <span class="text-sm font-inter text-gray-900" [attr.data-testid]="'members-engagement-meetings-' + member.uid">
```
dealako, line 179: [minor] Template method calls violate the repo's own template-function rule

Issue: `getEngagementMeetingsLabel(member)`, `getMemberEngagement(member)`, `getEngagementContext(engagementRow)`, and `getEngagementSeverity(engagementRow)` are all called directly in the template.
Proof: `docs/reviews/frontend-checklist.md` rule 4 ("No template functions", SHOULD FIX) states templates may only use signal reads, computed values, and pipes — no method calls that execute logic. These four calls are exactly that pattern (agrees with Copilot's finding on this block, though Copilot's perf framing is overstated: `getMemberEngagement` is an O(1) `Map.get`, not an N+1 lookup, so there's no real performance cost here — the issue is the documented convention, not perf).
Why it matters: deviates from the repo's documented template-purity rule and makes these bindings harder to trace/reuse than a precomputed row view.
Fix: precompute a UID-keyed engagement row view (label, context, severity) in a `computed()` signal and bind its properties directly, same as the rest of the row already does for non-engagement columns.

dealako (changes requested):
Hi @manishdixitlfx! Thanks for the engagement rollup work — the shared-window architecture between the Members table and Overview card is well thought out (the tuple-based `initEngagement` with careful `roleLoading`/`visitor`/`refresh` gating is genuinely nice reasoning about race conditions), and the fail-closed feature-flag behavior plus the degraded-response states are handled thoughtfully throughout. This is a solid PR overall; the issues below are all fixable in a small follow-up commit.

I cross-checked the existing CodeRabbit/Copilot/Cursor comments against current code before writing this up — two of theirs are confirmed real bugs, one is only partially valid (see below).

### 🔴 Blocking: 1 issue
- `engagementWindow` signal isn't reset when the route's committee id changes, so Angular's reuse of `/groups/:id` can carry a previously-selected window (e.g. YTD) onto a different committee, silently showing the wrong window's rollup with no error indicator (agrees with Copilot's finding).

### 🟡 Minor: 3 issues
- `initEngagement`'s `EMPTY`-on-`roleLoading` + `finalize`-based loading reset can flip `engagementLoading` to `false` on a cancelled in-flight request during a silent role refresh, briefly showing the "unavailable" state (agrees with Cursor Bugbot).
- The members-table engagement chip's `pTooltip`/`tabindex` are on the wrapping `<span>`, but PrimeNG's focus listener binds to the nested `<lfx-tag>` (which renders `.p-component`), so keyboard Tab won't trigger the tooltip — confirmed for `committee-members.component.html:190-197` only. CodeRabbit's report generalized this to the two Overview stat rows in `committee-engagement-summary.component.html:40-65` as well, but those `pTooltip`/`tabindex` pairs are on the same plain `<div>` with no `.p-component` descendant, so they're unaffected — that part of the bot finding doesn't hold up.
- Several engagement-cell template bindings (`getEngagementMeetingsLabel(member)`, `getMemberEngagement(member)`, `getEngagementContext(engagementRow)`, `getEngagementSeverity(engagementRow)`) call component methods directly in the template, which is a documented SHOULD-FIX in `docs/reviews/frontend-checklist.md` ("No template functions" — only signal reads/computed/pipes allowed). The underlying lookups are O(1) map gets so there's no real perf cost, but it's worth precomputing a per-row view via a computed signal to match repo convention (agrees with Copilot's finding, though the perf framing there is overstated).

### ⚪ Nit: 0 issues

### ❔ Question: 1 item
- The new E2E suite (`committee-engagement.spec.ts`) doesn't test the combination of selecting the At Risk chip and then the data degrading (window switch / fetch failure) to verify the `linkedSignal` reset back to `'all'`, nor does it assert the loading-skeleton state is actually rendered during a slow fetch. Worth a follow-up test, not blocking this PR.

### Revision history note
This PR already went through two review-driven fix rounds (`3f4562c3`, `32522b29`) that addressed prior engagement UI findings and "completed" tooltip a11y. That work fixed the Overview card's tooltip/tabindex pairing correctly, but the Members table's tag tooltip (a different location) still has the mismatch — worth a quick follow-up so it doesn't get missed a second time.

### Final decision
🔴 **Needs changes before approval** — one blocking correctness bug (stale window on committee navigation) plus the confirmed loading-state and tooltip-focus issues above.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 5e52a6107 — `engagementLoading` now clears via `tap` on emission instead of `finalize`, mirroring `initDocuments`'s identical pattern in `committee-overview.component.ts`. `finalize` also fired on `switchMap` cancellation (`roleLoading` flipping true mid-refresh), which cleared the loading flag while `engagement()` still held stale data.

manishdixitlfx: Fixed in 5e52a6107 — same fix as Cursor Bugbot's finding on this line: `engagementLoading` now clears via `tap` on emission, not `finalize`, mirroring `initDocuments`.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -149,6 +170,42 @@
...
+                    {{ getEngagementMeetingsLabel(member) }}
+                  </span>
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 5e52a6107 — added `memberEngagementRows`, a UID-keyed computed signal producing a `CommitteeMemberEngagementRowView` per member (`meetingsLabel`/`context`/`severity`/`row`). The template now does one `memberEngagementRows().get(member.uid)` lookup instead of calling the four methods directly.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
manishdixitlfx: Fixed in 5e52a6107 — precomputed a UID-keyed `memberEngagementRows` computed signal (label/context/severity/row), same idiom `engagementByUid` already used. Agreed the perf framing was overstated (it is an O(1) Map.get) — the actual fix is for the documented template-purity convention.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 5e52a6107 — converted `engagementWindow` to a `linkedSignal` keyed on `committeeId()`, mirroring the adjacent `activeTab` linkedSignal already in this file. It now resets to the default window whenever the committee changes, while `onEngagementWindowChange` can still write it within the same committee.

manishdixitlfx: Fixed in 5e52a6107 — `engagementWindow` is now a `linkedSignal` keyed on `committeeId()`, resetting to `COMMITTEE_ENGAGEMENT_DEFAULT_WINDOW` on committee change, same pattern as the adjacent `activeTab` signal. Agreed this was the correct framing (data-correctness, not just UX).

apps/lfx-one/src/app/modules/committees/components/committee-engagement-summary/committee-engagement-summary.component.html
```diff
@@ -0,0 +1,82 @@
...
+          tooltipPosition="top"
+          tooltipEvent="both"
+          tabindex="0"
+          role="note"
+          aria-label="Attendance Rate — personal attendance across all invited roster members, including Emeritus seats"
+          data-testid="committee-engagement-summary-attendance-rate">
+          <i class="fa-light fa-chart-line text-gray-400 w-4 text-center flex-shrink-0" aria-hidden="true"></i>
+          <span class="text-sm text-gray-700 flex-1">Attendance Rate</span>
+          <span class="text-sm font-semibold text-gray-900">{{ attendanceRateLabel() }}</span>
+        </div>
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 5e52a6107, via the alternative fix you flagged: gave `TagComponent` a `tooltip`/`tooltipPosition` input applied directly on its own rendered element (`p-tag`, or the plain `span` for `outlined`) rather than leaving it to a consumer-side wrapping element. Verified against PrimeNG's actual tooltip source — `querySelector('.p-component')` on the directive's own host, falling back to the host itself via `matches()` only when no such descendant exists — so a wrapper around `<lfx-tag>` (or `<lfx-tag>` itself) would still resolve the listener to the inner `<p-tag>`, a different node than `tabindex`. The two locations in this file are unaffected (plain `<div>`, no `.p-component` descendant) — left them as-is, per dealako's independent check.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
manishdixitlfx: Fixed in 5e52a6107 — gave `TagComponent` a `tooltip` input wired directly onto its own `p-tag`/`span` element (see reply to CodeRabbit's thread on this same bug for the PrimeNG-source verification of why the fix needs to live inside `TagComponent` rather than on `<lfx-tag>` itself). The members-table call site now just passes `[tooltip]="engagementView.context"`.

apps/lfx-one/src/app/shared/services/committee.service.ts
```diff
@@ -49,6 +51,27 @@ export class CommitteeService {
...
+   * `committee#auditor`-gated, stricter than roster visibility) and degrades the same way.
+   */
+  public getCommitteeEngagement(committeeUid: string, window: CommitteeEngagementWindow): Observable<CommitteeEngagementResponse | null> {
+    const params = new HttpParams().set('window', window);
+    return this.http.get<CommitteeEngagementResponse>(`/api/committees/${encodeURIComponent(committeeUid)}/engagement`, { params }).pipe(
+      catchError((error) => {
+        // 403 is the expected outcome for every non-auditor caller — logging it as an error would
+        // spam the console (and Datadog RUM error tracking) once per window switch for most users.
+        if ((error as { status?: number })?.status !== 403) {
+          console.error('Failed to load committee engagement:', error);
```
dealako: [minor] Full HttpErrorResponse logged to console / Datadog RUM for non-403 engagement errors

Issue: `console.error('Failed to load committee engagement:', error)` logs the entire `HttpErrorResponse` object, which per the adjacent comment is picked up by Datadog RUM error tracking.

Proof: line 68 passes the raw `error` (not just `error.status`/`error.message`) to `console.error`. `HttpErrorResponse` can carry response headers and body content depending on what the upstream returned for a non-403 failure.

Why it matters: unsanitized error objects reaching an external error tracker risk exposing more than intended (headers, url with query params, etc.) if the upstream ever returns something sensitive in an error body. Note this mirrors the existing pattern at lines 48 and 274 in this same file, so it's not a regression introduced by this PR — flagging since this diff is what's under review and it'd be a good time to tighten all three call sites together.

Fix: log a narrower payload, e.g. `console.error('Failed to load committee engagement:', error.status, error.message)`, instead of the full error object.

dealako (changes requested):
Hey @manishdixitlfx — nice work driving this through several rounds. This is a re-review: I re-ran the independent trio (security/privacy, correctness/perf/tests, style/docs) against the current HEAD (`385a1eb`).

**Overall impression:** the engagement feature is in good shape. Every issue raised across the earlier rounds — mine, Cursor Bugbot's, Copilot's, and CodeRabbit's — is fixed and the threads are resolved (`5e52a610`, `9c180064`, `0989eb1e`, `b020ccde`, `d39e4aab`, `385a1eb3`). I also checked the one still-**open** CodeRabbit thread on `tag.component.html` ("Preserve the classification as the accessible name") against current code: `accessibleLabel()` already composes `` `${value()} — ${tooltip()}` `` exactly as CodeRabbit's fallback suggested, so that's resolved in code too — just needs the GitHub thread marked resolved.

The fresh pass surfaced one new item in the diff (left inline) plus one pre-existing, out-of-scope bug worth a follow-up:

- 🟡 Minor (in this diff): `committee.service.ts` logs the full `HttpErrorResponse` to console/Datadog RUM for non-403 engagement fetch errors — matches an existing pattern already used twice elsewhere in the same file, so not a regression this PR introduces, but worth tightening while it's in view.
- ❔ Not part of this PR's diff, noted for a follow-up ticket rather than blocking here: `committee-members.component.ts`'s `deleteCommitteeMember` permission-cleanup guard (`hadElevatedPermission`) can misfire when a writer/auditor has no email — `w.email?.toLowerCase() === memberEmail` is `true` when both sides are `undefined`. Pre-existing code, untouched by this PR, so I'm not filing it as an inline comment here.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 1 — unsanitized error logging in the new `getCommitteeEngagement` method (pre-existing pattern)
- ⚪ Nit: 0
- ❔ Question: 0 (plus 1 out-of-scope note above)

**Final decision:** 🔴 **Needs changes before approval** — per this workflow's policy, any finding from the security/privacy dimension routes to Changes Requested regardless of severity. This one issue doesn't block merge on its own; happy to re-review immediately once addressed (or documented as an accepted trade-off).

apps/lfx-one/src/app/shared/components/tag/tag.component.html
```diff
@@ -5,7 +5,13 @@
...
     class="inline-flex items-center gap-1.5 border border-gray-200 bg-white text-gray-700 text-xs font-medium px-2 py-0.5 whitespace-nowrap"
     [class.rounded-full]="rounded()"
-    [ngClass]="styleClass()">
+    [ngClass]="styleClass()"
+    [pTooltip]="tooltip()"
+    tooltipEvent="both"
+    [tooltipPosition]="tooltipPosition()"
+    [attr.tabindex]="tooltip() ? 0 : null"
+    [attr.role]="tooltip() ? 'note' : null"
+    [attr.aria-label]="tooltip() || null">
```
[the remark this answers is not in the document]
manishdixitlfx: Already resolved in code, confirmed independently by @dealako's re-review: `TagComponent.accessibleLabel` (added in `4ba7052`) composes `` `${value()} — ${tooltip()}` `` when a tooltip is present, exactly the fallback this comment suggested, so `value()`'s classification is always preserved in the accessible name rather than being replaced by the tooltip text alone. Marking resolved.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -180,9 +217,33 @@ export class CommitteeViewComponent {
...
+  // by design, so tab-visibility gates like isMemberOrAdmin default to showing tabs during the
+  // initial load rather than flashing them away). That optimism is wrong for a gate that fires a
+  // fetch and renders a whole new card: a genuine visitor would see canAccessEngagement() flip true
+  // for the duration of every myRoleLoading() window (initial load, post-join, any silent
+  // refreshCommittee()), flashing the Overview card open before it resolves closed again (Cursor
+  // Bugbot). Loading itself is already handled separately — initEngagement's own `roleLoading` check
+  // holds fetches via EMPTY during this window — so this only needs to settle before deciding.
+  public readonly canAccessEngagement: Signal<boolean> = computed(() => {
+    if (this.myRoleLoading()) return false;
+    return this.myRole() !== null || this.canEdit() || this.canReview() || this.isCallerInAuditorList(this.committee()?.inherited_auditors);
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in e64b358 — `canAccessEngagement` now only updates on a settled (non-loading) resolution instead of reading `false` throughout every `myRoleLoading()` window, so a silent `refreshCommittee()` no longer unmounts+remounts the Overview card for an already-eligible user. It still defaults to `false` pre-first-resolution, so the original visitor-flash fix stays intact.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +822,118 @@ export class CommitteeViewComponent {
...
+          if (roleLoading) {
+            return EMPTY;
+          }
+          // Not eligible per canAccessEngagement: skip the guaranteed-403 fetch entirely, matching
+          // initDocuments' "don't issue a GET nothing will display" precedent. The Overview card is
+          // gated on the same canAccessEngagement value (passed down as an input), so nothing would
+          // render this response anyway.
+          if (notEligible) {
+            this.engagementLoading.set(false);
+            return of(null);
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 49eb379 — replaced the toObservable/toSignal derivation with a linkedSignal, whose computation runs synchronously within the same signal flush as `myRoleLoading()`, so there's no async gap for `engagementKey` to observe a stale `notEligible` through.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +822,118 @@ export class CommitteeViewComponent {
     );
   }
 
+  private initCanAccessEngagement(): Signal<boolean> {
+    return toSignal(
+      toObservable(
+        computed(() => ({
+          loading: this.myRoleLoading(),
+          eligible: this.myRole() !== null || this.canEdit() || this.canReview() || this.isCallerInAuditorList(this.committee()?.inherited_auditors),
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed against `committee-read-access.helper.ts` — `auditor` is `[user, team#member] or writer or auditor from project or meeting_coordinator from project`, and the client-side reconstruction here doesn't check the `meeting_coordinator` leg.

Leaving this one as a documented gap rather than fixing: closing it properly means an extra `GET /api/projects/:project_uid?meeting_coordinator=true` round trip on every committee page load (this component doesn't inject `ProjectService` today), just to cover a narrow persona — a project meeting coordinator who is *not* also a roster member, writer, or listed/inherited auditor. It fails closed (hides the card/skips the fetch, never grants excess access), consistent with the same trade-off already accepted for `inherited_auditors` being best-effort elsewhere in this eligibility check. Flagging for a follow-up ticket if that persona turns out to matter in practice, rather than adding a new network dependency to this UI-consumption slice.

Also fixed the general-review's separate "not keyed to committee" finding in b406852 — the held eligibility value is now scoped to `committeeId()`, so route-reused navigation to a different (ineligible) committee no longer keeps the previous committee's Overview card mounted.

---

apps/lfx-one/src/app/shared/services/committee.service.ts
manishdixitlfx: Fixed in c0c762f — all three call sites (`getGroupsEngagementStats`, `getCommitteeEngagement`, `getMyCommittees`) now log `error.status, error.message` instead of the full `HttpErrorResponse`.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +822,131 @@ export class CommitteeViewComponent {
...
+    // committeeId() (the route-synchronous id, same distinguishing signal initEngagement's
+    // routeCommitteeId uses) so the held value is only reused for a SAME-committee silent refresh:
+    // route-reused navigation to a different committee must not keep rendering the previous
+    // committee's eligibility (and therefore its stale Overview card, which is gated on this signal
+    // alone with no loading check) while the new committee's own role is still resolving (Copilot).
+    return linkedSignal<{ committeeId: string | null; loading: boolean; eligible: boolean }, boolean>({
+      source: () => ({
+        committeeId: this.committeeId(),
+        loading: this.myRoleLoading(),
+        eligible: this.myRole() !== null || this.canEdit() || this.canReview() || this.isCallerInAuditorList(this.committee()?.inherited_auditors),
```
dealako: [minor] Eligibility check omits the `meeting_coordinator from project` FGA path

Issue: `eligible` is computed as `this.myRole() !== null || this.canEdit() || this.canReview() || this.isCallerInAuditorList(this.committee()?.inherited_auditors)` — roster membership, writer, committee-scoped auditor, and inherited (project-level) auditor. It does not account for a project-level `meeting_coordinator`.

Proof: the BFF's read gate for this exact endpoint (`server/helpers/committee-read-access.helper.ts:14-25`) checks `committee#auditor`, and per the platform FGA model that relation is `[user, team#member] or writer or auditor from project or meeting_coordinator from project`. A caller who only holds `meeting_coordinator from project` — no roster role, no writer/auditor grant — would be authorized server-side but none of the four checks above cover them client-side.

Why it matters: that user class never sees the Overview engagement card and `initEngagement`'s fetch gate (`notEligible: !this.canAccessEngagement()`) skips the request entirely — a silent, client-only authorization mismatch. Fails closed (no leak), but it's a real functional gap for meeting coordinators, matching what Copilot flagged on this same line.

Fix: the codebase already has a pattern for surfacing this — `ProjectService.getProject(slug, false, { meetingCoordinator: true })` (see `apps/lfx-one/src/app/shared/guards/writer.guard.ts`) triggers the server's opt-in `meeting_coordinator` FGA check and returns `project.meetingCoordinator`. Thread an equivalent project-level meeting-coordinator signal into `eligible` here (or expose it alongside `committee()` if the committee-service proxy can resolve it), so the client gate mirrors the server's `committee#auditor` relation exactly.

dealako (changes requested):
Hey @manishdixitlfx — this is a follow-up re-review of the six commits pushed since my last review (`385a1eb3` → `b4068520`), scoped only to that new diff.

👏 **Nice work:**
- `c0c762ff` cleanly fixes the error-logging finding from my last review — `committee.service.ts` now logs `error.status`/`error.message` only at every call site, not the full `HttpErrorResponse`.
- `b9cd84cc`, `7da3c6e2`, `e64b358f`, `49eb3790`, `b4068520` collectively rework `canAccessEngagement` into a `linkedSignal` keyed on `committeeId()` — this is a solid fix for the flash/race issues Cursor Bugbot raised: it resolves synchronously (no tick-behind gap), holds its value through silent refreshes, and resets cleanly on committee navigation. I verified `eligible` now correctly includes `canEdit()`/`canReview()`/`inherited_auditors`, not just roster membership — that specific writer/auditor gap Copilot raised earlier is fixed.
- The two previously-open CodeRabbit threads (`tag.component.html` accessible name, `committee-engagement-summary.component.html:52` tooltip/focus target) are both confirmed resolved in current code.

**Still open — one item Copilot flagged on this diff that I independently verified:**

See inline comment on `committee-view.component.ts`. Short version: the new `eligible` computation still doesn't cover the `meeting_coordinator from project` path that the server's `committee#auditor` FGA relation grants (confirmed directly in `server/helpers/committee-read-access.helper.ts:14-25`). A project-level meeting coordinator who isn't a roster member/writer/auditor gets silently denied the Overview engagement card and fetch, even though the server would authorize the request. This is a fail-closed gap (no data leak), but it's a real client/server authorization mismatch for a real user class — worth closing before merge.

**Out of scope for this diff, not filed inline:** `committee-members.component.html`'s permission cells (`getMemberPermissionLabel`/`getMemberPermissionSeverity`) are still called at template-render time rather than precomputed like the new `memberEngagementRows()` pattern — untouched by this diff, so just a heads-up for a future pass rather than a blocking item here.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 1 — engagement eligibility gate omits the `meeting_coordinator from project` FGA path
- ⚪ Nit: 0

**Final decision:** 🔴 **Needs changes before approval** — per this workflow's policy, any finding from the security/correctness review dimensions routes to Changes Requested. Happy to re-review immediately once this is addressed.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
manishdixitlfx: Fixed in 4050a71 — added `initMeetingCoordinator()`, threading `ProjectService.getProject(committee.project_uid, false, { meetingCoordinator: true })` into `canAccessEngagement`'s `eligible` computation, exactly as suggested. It only fires the fetch when the cheaper checks (roster/writer/committee-or-inherited-auditor) already failed, mirroring `ProjectService`'s own "skip meeting_coordinator for writers" shortcut, so this doesn't add a round trip to the common case. Its loading state is folded into the same `linkedSignal` gating as `myRoleLoading()`, so it holds gracefully the same way. `yarn format`/`lint`/`build`/`test` all green.

Thanks for catching this on the re-review — appreciate the thorough independent verification against the FGA model.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +836,175 @@ export class CommitteeViewComponent {
...
+          if (roleLoading) {
+            return EMPTY;
+          }
+          // Not eligible per canAccessEngagement: skip the guaranteed-403 fetch entirely, matching
+          // initDocuments' "don't issue a GET nothing will display" precedent. The Overview card is
+          // gated on the same canAccessEngagement value (passed down as an input), so nothing would
+          // render this response anyway.
+          if (notEligible) {
+            this.engagementLoading.set(false);
+            return of(null);
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in 40b52a7 — `engagementKey.roleLoading` now folds in `meetingCoordinatorLoading()` alongside `myRoleLoading()`, so `initEngagement`'s EMPTY-hold branch covers the project fetch window too, not just `canAccessEngagement`'s own `linkedSignal`.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +836,181 @@ export class CommitteeViewComponent {
...
+          this.meetingCoordinatorLoading.set(true);
+          return this.projectService.getProject(projectUid, false, { meetingCoordinator: true }).pipe(
+            map((project) => project?.meetingCoordinator === true),
+            finalize(() => this.meetingCoordinatorLoading.set(false))
+          );
+        })
+      ),
+      { initialValue: false }
+    );
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed in d4eb30c — added `enabled: this.engagementMetricsEnabled()` to `initMeetingCoordinator`'s source tuple and its early-exit condition, matching every other engagement-related fetch in this file. `format`/`lint`/`build`/`test` all green.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +836,186 @@ export class CommitteeViewComponent {
...
+        // meetingCoordinatorLoading folded in here (not left as a separate downstream check) so the
+        // linkedSignal holds its previous settled value through that fetch's window too, the same as
+        // it already does for myRoleLoading().
+        loading: this.myRoleLoading() || this.meetingCoordinatorLoading(),
+        eligible:
+          this.myRole() !== null ||
+          this.canEdit() ||
+          this.canReview() ||
+          this.isCallerInAuditorList(this.committee()?.inherited_auditors) ||
+          this.meetingCoordinator(),
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed — the client-side eligibility check has no way to represent `team#member` grants; `auditors`/`inherited_auditors` are plain user-email/username lists, and unlike the `meeting_coordinator` fix (which reused the existing `ProjectService` FGA pass-through), there's no existing frontend service that can answer "is this caller an auditor via team membership." Closing this properly needs new backend surface (e.g. a server-resolved effective-auditor field/endpoint), not a client-side workaround.

Fails closed (no data leak) — filed as a follow-up rather than blocking this PR: https://linuxfoundation.atlassian.net/browse/LFXV2-2975

dealako:
Hey @manishdixitlfx — re-reviewed the three commits pushed since my last round (`b4068520` → `6295bece`).

👏 **Nice work:**
- `4050a711` closes the `meeting_coordinator from project` gap cleanly — `initMeetingCoordinator()` uses the established `ProjectService.getProject(uid, false, { meetingCoordinator: true })` pattern from `writer.guard.ts`, and only fires when the cheaper checks (`myRole`/`canEdit`/`canReview`/`inherited_auditors`) already failed, so it's a no-op fetch for the overwhelming majority of page loads.
- `40b52a74` and `d4eb30c2` both directly and correctly fix the two new Cursor Bugbot comments on this diff — verified against current code, not just the bot's claim:
  - The "flash of unavailable state" comment (`committee-view.component.ts:1002`) — fixed: `meetingCoordinatorLoading()` is now folded into `initEngagement`'s `roleLoading`, not just `canAccessEngagement`'s own `linkedSignal`.
  - The "flag-off still fetches" comment (`committee-view.component.ts:915`) — fixed: `initMeetingCoordinator` now gates on `engagementMetricsEnabled()` before ever calling `getProject`.
- Both fixes were independently confirmed resolved by two separate reviewers this round (security/authz dimension and correctness/perf dimension), tracing the actual signal flow rather than just the diff hunk.

**Still open — one gap, independently confirmed by two reviewers and matching Copilot's inline comment on this diff:**

See inline comment on `committee-view.component.ts:865`. Short version: the server's `committee#auditor` FGA relation is `[user, team#member] or writer or auditor from project or meeting_coordinator from project`. This round's fix closes the `meeting_coordinator` leg, but the client-side `eligible` reconstruction still can't represent a `team#member`-derived auditor grant, since `auditors`/`inherited_auditors` are flat `CommitteeUser[]` lists with no team semantics. A user who's an auditor purely via team membership passes server authorization but is still silently denied the Overview card client-side.

**Note on two flagged constants:** an earlier pass of automated analysis flagged `CHAIR_ROLES` (missing JSDoc) and `MY_CLAS_ENABLED_FLAG` (apparently unused) in the two constants files touched by the file-scoped diff. I checked `git show --stat` on all three of this round's actual commits (`4050a711`, `40b52a74`, `d4eb30c2`) — none of them touch `committees.constants.ts` or `feature-flags.constants.ts`. Both constants came in via the unrelated `main` merge noise, not this PR's own work, so I'm not carrying them into this review.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 1 — engagement eligibility gate still omits the `team#member`-derived auditor grant
- ⚪ Nit: 0

**Final decision:** 🔴 **Needs changes before approval** — per this workflow's policy, any finding from the security/correctness review dimensions routes to Changes Requested, even though this is a pre-existing gap rather than a regression from this round's fix. Given the meeting-coordinator fix is now fully resolved (3/3 items), this should be quick to close out — happy to re-review immediately.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
dealako: [minor] Eligibility check still omits `team#member`-derived auditor grants

Issue: this round's fix correctly adds the `meeting_coordinator from project` leg to `eligible`, but the server's `committee#auditor` FGA relation is `[user, team#member] or writer or auditor from project or meeting_coordinator from project`. The `team#member` leg still isn't represented client-side.

Proof: `eligible` here checks `myRole() !== null || canEdit() || canReview() || isCallerInAuditorList(this.committee()?.inherited_auditors) || meetingCoordinator()`. `auditors`/`inherited_auditors` are `CommitteeUser[]` settings lists of individual users — they have no way to encode "anyone on team X is an auditor." A caller who's an auditor purely via team membership passes the server's gate (`server/helpers/committee-read-access.helper.ts:14-25`) but fails every check in this list.

Why it matters: same class of gap as the meeting_coordinator issue this PR just fixed — a real, authorized user class silently loses the Overview engagement card and skips the fetch (fails closed, no leak, but a real functional gap). Independently flagged by Copilot on this line and confirmed by two independent reviewers this round.

Fix: this can't be reconstructed client-side from settings lists the way the other three legs can — team membership isn't exposed to the client at all. Recommend a follow-up that derives an effective-auditor (or effective-eligibility) flag server-side, mirroring the `meetingCoordinator` opt-in pattern (`ProjectService.getProject(uid, false, { meetingCoordinator: true })`) rather than continuing to hand-reconstruct the FGA relation leg-by-leg in the client. Reasonable to scope as a follow-up ticket rather than blocking this PR further, given how narrowly this keeps recurring — but worth filing explicitly so it doesn't get lost.

dealako (changes requested):
Follow-up re-review of the meeting-coordinator fix — see the summary comment above for the full writeup. Short version: the meeting_coordinator gap from last round is fully resolved (3/3 commits verified independently), but one new [minor] gap surfaced this round via Copilot + two independent reviewer dimensions: the client-side eligibility check still can't represent a team#member-derived auditor grant. See inline comment.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
manishdixitlfx: Agreed this is real, and agreed it's not client-fixable the way `meeting_coordinator` was — team membership isn't exposed to the frontend at all today, so there's no existing service call (unlike `ProjectService.getProject(..., { meetingCoordinator: true })`) that could resolve it without new backend surface.

Filed as a follow-up rather than expanding this PR's scope further: https://linuxfoundation.atlassian.net/browse/LFXV2-2975 (same ticket referenced on the sibling Copilot thread on this line). Fails closed in the meantime — no data leak, just a narrower-than-server client gate for this one auditor path.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +836,186 @@ export class CommitteeViewComponent {
...
+            a.notEligible === b.notEligible &&
+            a.refresh === b.refresh
+        ),
+        switchMap(({ routeCommitteeId, uid, window, enabled, roleLoading, notEligible }) => {
+          // uid is guaranteed non-null here (filtered above) — this only needs to check flag/SSR now.
+          // Flag off (or SSR, where the flag fails closed to its default) means zero engagement
+          // fetches — the gated UI renders nothing, so a request would be pure waste.
+          if (!enabled || !isPlatformBrowser(this.platformId)) {
+            this.engagementLoading.set(false);
+            return of(null);
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed — `getBooleanFlag` is a plain `computed()` the template's `@if` reads directly and synchronously, while this pipeline's `toObservable`/`switchMap` lags a tick behind it, same root cause as the two prior findings this round.

Fixed in e9d8b82 — added `flagResolved: this.featureFlagService.initialized()` to `engagementKey`. The pipeline now holds (`EMPTY`) while the flag hasn't resolved yet, only treating `!enabled` as a genuine terminal "off" state once `flagResolved` is true.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +836,205 @@ export class CommitteeViewComponent {
...
+          this.meetingCoordinatorLoading.set(true);
+          return this.projectService.getProject(projectUid, false, { meetingCoordinator: true }).pipe(
+            map((project) => project?.meetingCoordinator === true),
+            finalize(() => this.meetingCoordinatorLoading.set(false))
+          );
+        })
+      ),
+      { initialValue: false }
+    );
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed and fixed in f35bc1d8 — wrapped both `meetingCoordinator` and `meetingCoordinatorLoading` in linkedSignals keyed on `committeeId()`, same pattern already used for `canAccessEngagement`/`engagementWindow` in this file. The raw `initMeetingCoordinator` toSignal/toObservable pipeline only re-evaluates once `committee()` resolves to the new committee — a tick behind `committeeId()` itself, and behind `loading`/`committeeRefreshing` flipping false in the same finalize block. The wrappers force `meetingCoordinatorLoading` true and `meetingCoordinator` false immediately on a committee change, so `canAccessEngagement`'s eligible computation can never read a stale previous-committee value during that gap — closes both directions (opens-then-closes and closes-then-opens).

dealako:
Hey @manishdixitlfx — re-reviewed the two commits pushed since my last round (`6295bece` → `f35bc1d8`).

👏 **Nice work:**
- `e9d8b821` correctly distinguishes the LaunchDarkly SDK's transient `false` default (before `FeatureFlagService` finishes initializing) from a genuine "flag is off" terminal state — adding `flagResolved: this.featureFlagService.initialized()` to the `engagementKey` tuple and holding (`EMPTY`) until it resolves. Verified the RxJS mechanics: `EMPTY` correctly completes the inner observable without emitting, so the signal holds its prior value rather than erroring or flashing null, and there's no infinite-hold risk since `flagResolved` is guaranteed to settle during browser init.
- `f35bc1d8` closes the stale-coordinator flash on committee navigation by wrapping `meetingCoordinator`/`meetingCoordinatorLoading` in `linkedSignal`s keyed on `committeeId()` — the same established pattern already used by `canAccessEngagement`/`engagementWindow` in this file. Traced the full navigation sequence (route change → forced override → `committee()` resolves → raw signal updates → override releases) and confirmed both flash directions (opens-then-closes and closes-then-opens) are closed, with no lost fetches or double-fires.
- Both fixes were independently verified as fully resolved by all three reviewers this round (security/authz, correctness/perf, and style/docs), including a direct trace through the actual signal/observable timing, not just the diff hunk or the bot's/author's narrative.
- The `team#member`-derived auditor gap from last round is appropriately handled — acknowledged as real but not client-fixable without new backend surface, and filed as its own tracked follow-up ([LFXV2-2975](https://linuxfoundation.atlassian.net/browse/LFXV2-2975)) rather than scope-creeping this PR further. Fails closed in the meantime, no data leak. Not re-raising it here since it's now tracked separately.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

**Final decision:** ✅ **Approved** — both Cursor Bugbot findings from this round are cleanly resolved, no new issues surfaced across security, correctness/performance, or style/docs, and the one remaining known gap is already tracked as a separate follow-up ticket rather than being in scope here. Nice work closing out this engagement-eligibility timing series.


[LFXV2-2975]: https://linuxfoundation.atlassian.net/browse/LFXV2-2975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dealako (dismissed):
Approving — see the summary comment above for the full round-4 write-up. Both Cursor Bugbot flash-bug findings from this round (`e9d8b821`, `f35bc1d8`) are cleanly resolved, verified independently across security/authz, correctness/performance, and style/docs dimensions with no new findings. The one remaining known gap (team#member-derived auditor grants) is tracked separately as LFXV2-2975 and appropriately out of scope for this PR.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -180,11 +226,71 @@ export class CommitteeViewComponent {
...
+  // false) but before the coordinator pipeline has re-fired: canAccessEngagement's own `loading`
+  // field (myRoleLoading() || meetingCoordinatorLoading()) would read false too early, settling
+  // `eligible` on a stale meetingCoordinatorRaw() value from the PREVIOUS committee — either false
+  // when the new committee's caller actually is a coordinator (flashes "unavailable" before flipping
+  // open) or true when they aren't (briefly opens the card before flipping closed) (Cursor Bugbot).
+  public readonly meetingCoordinatorLoading: Signal<boolean> = linkedSignal<{ committeeId: string | null; raw: boolean }, boolean>({
+    source: () => ({ committeeId: this.committeeId(), raw: this.meetingCoordinatorLoadingRaw() }),
+    computation: (source, previous) => {
+      if (previous && previous.source.committeeId !== source.committeeId) {
+        return true;
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed real, and a serious one — traced it through: since `meetingCoordinatorLoadingRaw`/`meetingCoordinatorRaw` are primitive booleans, Angular signals skip notifying dependents on a `.set()` that doesn't change the value. Navigating between two committees where the caller is already eligible via roster/writer both times resolves the same `false` the previous committee already held, the wrapper `linkedSignal`s never got released from the `true` they force on committee change, and `canAccessEngagement`/the Overview card/engagement fetch would freeze on the previous committee's state.

Fixed in 7963fe1 by replacing the raw-signal-pair + wrapper-`linkedSignal` scheme with a single object-valued state signal (`{ committeeId, loading, coordinator }`) written atomically from the fetch pipeline — every emission is a fresh object literal, always reference-distinct, so notification never depends on primitive equality. `meetingCoordinatorLoading`/`meetingCoordinator` are now plain `computed()`s comparing the state's committee tag against the live `committeeId()`, the same idiom `initEngagement`'s `routeCommitteeId`/`uid` comparison already uses elsewhere in this file. `format`/`lint`/`build`/`test` all green.

---

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +871,205 @@ export class CommitteeViewComponent {
...
+          }
+          this.engagementLoading.set(true);
+          // Errors (including the expected 403 for non-auditor members) resolve to null inside the
+          // service — the tabs degrade to their "attendance unavailable" states, roster unaffected.
+          // tap on emission, not finalize: finalize also fires when switchMap cancels this in-flight
+          // request (e.g. roleLoading flips true mid-refresh), which would clear engagementLoading
+          // while engagement() still holds stale data — a visible flicker to the unavailable state on
+          // every silent role refresh. Mirrors initDocuments' identical reasoning in
+          // committee-overview.component.ts.
+          return this.committeeService.getCommitteeEngagement(uid, window).pipe(tap(() => this.engagementLoading.set(false)));
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed — traced the race: `refreshMembers()` bumps `membersRefresh()` and `refresh()` synchronously in the same call, but `committeeRefreshing` was only ever set `true` asynchronously inside `initializeCommittee`'s own `toObservable(this.refresh)`-driven switchMap. Since that pipeline and `initEngagement`'s `engagementKey` pipeline are two independently-scheduled `toObservable`-driven effects with no ordering guarantee between them, `engagementKey` could re-evaluate off the `membersRefresh` change and fire a real fetch before `myRoleLoading()` picked up `committeeRefreshing()`, which then gets cancelled via `EMPTY` once `roleLoading` does catch up — without `engagementLoading` ever clearing back to `false`.

Fixed in fa1b13c — `refreshCommittee()` now sets `committeeRefreshing` synchronously at the call site (in addition to the existing async set on the actual refetch), closing the gap regardless of effect-scheduling order. `format`/`lint`/`build`/`test` all green.

---

dealako:
Hey @manishdixitlfx — re-reviewed the two commits pushed since my last round (`f35bc1d8` → `fa1b13c5`).

👏 **Nice work:**
- `fa1b13c5` closes the Cursor Bugbot-flagged skeleton-flash race by setting `committeeRefreshing` synchronously at the `refreshCommittee()` call site, in addition to the existing async set on the actual refetch. Traced it: `refreshMembers()` bumps `membersRefresh()` and `refresh()` in the same synchronous call, so without this fix `engagementKey` could re-evaluate and fire (then waste) a real fetch before `myRoleLoading()` picked up `committeeRefreshing()` — now the flag flips before either `toObservable`-driven pipeline gets a chance to race. All three subagents (security, correctness, style) independently confirmed this holds regardless of effect-scheduling order.
- `7963fe1b` fixes the stuck-`true` `meetingCoordinatorLoading` bug (primitive-boolean linked signals don't notify on a `.set()` that doesn't change value, so navigating between two committees that both resolve to the same `false` could freeze the Overview card on stale state) by replacing the raw-signal-pair + wrapper-`linkedSignal` scheme with a single object-valued `meetingCoordinatorState` signal, written atomically so every emission is a fresh, reference-distinct object literal. Verified `meetingCoordinatorLoading`/`meetingCoordinator` now correctly gate on a live `committeeId()` tag comparison, the same idiom `initEngagement` already uses.
- No new authorization/data-leak surface from either fix — confirmed a stuck-loading or skeleton-flash state never lets one committee's engagement data render under a different committee's authorization context.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 2 — new issues surfaced by Copilot this round on adjacent (pre-existing, previously-approved) code in the same file, confirmed real by independent re-check:
  1. `initEngagement`'s `flagResolved: this.featureFlagService.initialized()` (line 976) — `initialized()` only means user context was applied, not that the LaunchDarkly provider has real flag values yet; `providerReady()` is the correct readiness signal. If the provider becomes ready after `initialized()`, the transient default `false` could get treated as terminal "flag off," clearing `engagementLoading` a tick early.
  2. `appearsWithin()` in `e2e/helpers/committee-engagement.helper.ts` (line 39) treats "element never became visible" and "flag is genuinely off" as the same outcome. A component error, broken access gate, bad mock, or renamed `data-testid` would silently read as "flag off" instead of failing the flag-on assertions that depend on it.
- ⚪ Nit: 1 — `meetingCoordinatorState`'s field is named `committeeId` but holds the async-resolved `committee().uid`, not the route-synchronous id (contrast `state.committeeId !== this.committeeId()` at line 241-242, which reads backwards vs. `initEngagement`'s clearer `routeCommitteeId`/`uid` naming). Not a bug — `initMeetingCoordinator`'s internal pipeline already uses `committeeUid` for the same value; renaming the exposed field to match would remove the ambiguity.

**Final decision:** 🔴 **Needs changes before approval** — per this workflow's policy, any finding from the security/correctness review dimensions routes to Changes Requested. Both are pre-existing gaps on already-approved code re-surfaced by Copilot this round, not regressions from `fa1b13c5`/`7963fe1b`, and both look like quick, contained fixes.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +864,208 @@ export class CommitteeViewComponent {
...
+      uid: this.committee()?.uid ?? null,
+      window: this.engagementWindow(),
+      // FeatureFlagService.initialized() — engagementMetricsEnabled() is a plain computed() that
+      // reads its LaunchDarkly default (false) until the client initializes, and the Overview card's
+      // template gate reads that same signal directly (synchronous). This async switchMap pipeline
+      // lags a tick behind, so without flagResolved below, the transient pre-init false could get
+      // treated as a genuine "flag off" terminal state, then flash the unavailable UI once the real
+      // (true) value resolves and the template gate opens before this pipeline's next emission
+      // catches up (Cursor Bugbot).
+      flagResolved: this.featureFlagService.initialized(),
```
dealako: [minor] flagResolved should check providerReady(), not initialized()

Issue: `flagResolved: this.featureFlagService.initialized()` treats user-context-applied as equivalent to real flag values being available.
Proof: `initialized()` only confirms user context was applied to the LaunchDarkly client, not that the provider has streamed real flag values — that's what `providerReady()` signals. If the provider becomes ready after `initialized()` resolves true, the transient pre-init `false` default gets treated as a genuine terminal 'flag off' state one tick early.
Why it matters: `engagementLoading` could clear on the disabled branch before the real flag value arrives, template briefly rendering the unavailable/em-dash state instead of the loading skeleton it's meant to hold through.
Fix: swap to `flagResolved: this.featureFlagService.providerReady()` (confirm the service exposes this signal; it's documented as the correct readiness check for real flag values).

apps/lfx-one/e2e/helpers/committee-engagement.helper.ts
```diff
@@ -0,0 +1,257 @@
...
+ * Whether `locator` becomes visible within `timeout` — used to detect the LaunchDarkly flag's
+ * resolved state. `isVisible({ timeout })` is deprecated/ignored, so a real `waitFor` is required:
+ * an immediate check can read `false` before the async flag client resolves.
+ */
+export async function appearsWithin(locator: Locator, timeout: number): Promise<boolean> {
+  try {
+    await locator.waitFor({ state: 'visible', timeout });
+    return true;
+  } catch {
+    return false;
```
dealako, line 44: [minor] appearsWithin() conflates 'never rendered' with 'flag is off'

Issue: catching any waitFor timeout and returning false is used as the flag-off signal in the specs, but a broken component, bad mock, broken access gate, or renamed data-testid produces the identical false.
Why it matters: flag-on tests can silently pass-as-skip on a real regression instead of failing loudly on it.
Fix: determine the LaunchDarkly precondition independently (or deterministically mock/configure the flag) and assert the target becomes visible, rather than using visibility itself as the flag-off predicate.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -180,11 +219,61 @@ export class CommitteeViewComponent {
...
+  // committee (Copilot). A fresh object literal is always reference-distinct, so every emission below
+  // reliably propagates, and meetingCoordinatorLoading/meetingCoordinator (further down) compare the
+  // tag against the LIVE committeeId() directly rather than needing a value-change to "release" them.
+  private readonly meetingCoordinatorState: Signal<{ committeeId: string | null; loading: boolean; coordinator: boolean }> = this.initMeetingCoordinator();
+  // True until meetingCoordinatorState has settled FOR THE CURRENT committee (the tag comparison
+  // covers both "still mid-navigation, state belongs to the previous committee" and "fetch actually
+  // in flight" — see the state signal's doc comment above).
+  public readonly meetingCoordinatorLoading: Signal<boolean> = computed(() => {
+    const state = this.meetingCoordinatorState();
+    return state.committeeId !== this.committeeId() || state.loading;
```
dealako: [nit] meetingCoordinatorState.committeeId actually holds the async-resolved uid

Issue: the field is named committeeId but is populated from committee().uid (see initMeetingCoordinator, which itself calls the equivalent value committeeUid internally) — the same name used elsewhere in this file for the route-synchronous committeeId() accessor.
Why it matters: state.committeeId !== this.committeeId() reads like comparing two ids of the same kind; it's actually async-resolved-uid vs. route-sync-id, the same distinction initEngagement keeps unambiguous with routeCommitteeId/uid.
Fix: rename the field to committeeUid throughout meetingCoordinatorState/initMeetingCoordinator for consistency with that naming convention.

dealako (changes requested):
Re-review of the two commits pushed since my last approval — see the [summary comment](https://github.com/linuxfoundation/lfx-self-serve/pull/1290#issuecomment-5181363101) for the full writeup. Both fixes from this round (skeleton-flash race, stuck-loading coordinator) verified resolved by security, correctness, and style passes. Two minor issues surfaced by Copilot on adjacent code this round, plus one naming nit — inline below.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
manishdixitlfx: Confirmed and fixed in cae9595f — swapped to `this.featureFlagService.providerReady()`. You're right that `initialized()` only confirms user context was applied, not that real flag values have streamed; `providerReady()`'s own doc comment on the service confirms that distinction.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
manishdixitlfx: Fixed in cae9595f — renamed `committeeId` to `committeeUid` throughout `meetingCoordinatorState`/`initMeetingCoordinator` and its two consumers. Good catch — it really did read like the route-sync accessor.

apps/lfx-one/e2e/helpers/committee-engagement.helper.ts
manishdixitlfx: Good catch on the theoretical risk, and I want to be upfront rather than just wave it off: this is a real gap in the abstract — a broken component, bad mock, or renamed data-testid genuinely would produce the same signal as "flag off."

That said, this exact pattern — element-visibility as the LaunchDarkly-flag-resolved signal, gated on `TEST_USERNAME`/`TEST_PASSWORD`, with a header comment calling out the trade-off — is already the established convention in this repo, not something new introduced here. `groups-engagement-stats.spec.ts`'s own `appearsWithin` helper (identical implementation) explicitly notes it's "matching org-selector.spec.ts's precedent," and `org-selector.spec.ts` uses the same `waitFor`-based flag detection for `org-lens-enabled`. This file followed that existing convention rather than inventing a new one.

Given it's a pre-existing, repo-wide pattern across three specs (not a defect this PR introduced), I don't think it's right to diverge from it in just this one file — that would leave three different "is the flag on" strategies in the codebase instead of one. No code change here; leaving this open as a legitimate note for a future pass that reconsiders the pattern across all three specs at once, if that's judged worth the added complexity (deterministic flag mocking) over what's currently a documented, accepted trade-off. Resolving this thread since there's nothing to fix specific to this PR.

dealako:
Hey @manishdixitlfx — re-reviewed the commit pushed since my last round (`fa1b13c5` → `cae9595f`).

👏 **Nice work:**
- `cae9595f` cleanly resolves both minor findings from last round: `flagResolved` now reads `this.featureFlagService.providerReady()` instead of `initialized()` — confirmed `providerReady()` is a genuine `Signal<boolean>` that only flips once the OpenFeature provider reaches `READY`, correctly distinct from `initialized()` (user-context-applied), and fails closed (holds via `EMPTY`) if the provider never reaches ready rather than getting stuck open. The `committeeId` → `committeeUid` rename on `meetingCoordinatorState` is complete and consistent across the state signal, `initMeetingCoordinator`'s emissions, and both consumers, with no stray references or collision with the unrelated route-sync `committeeId()` accessor.
- All three re-review passes (security/privacy, correctness/perf/tests, style/docs) came back clean — zero new findings on this diff.
- Appreciate the direct reply on the `appearsWithin()` e2e-helper thread rather than silently punting it — agreed this is an existing repo-wide convention (not something this PR introduced) and reasonable to leave as-is rather than scope-creep a broader e2e pattern change into this PR.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 1 — Copilot's new test-coverage finding: the meeting-coordinator-only access path (`needed=true` with no role/writer/auditor) isn't exercised by the e2e fixtures — `buildEngagementCommittee()` always sets `my_role: 'Member'`, so no spec mocks `/api/projects/:uid?meeting_coordinator=true`, and the route-reused-navigation timing this round's fixes specifically target goes untested end-to-end. Worth a follow-up fixture, not blocking given unit-level signal timing is otherwise well covered by this round's fixes.
- ⚪ Nit: 0

**Final decision:** ✅ **Approved with minor comments** — both prior findings are cleanly resolved and independently re-verified, the e2e-helper nit is appropriately left as existing convention, and the one remaining item is a test-coverage suggestion rather than a defect.

dealako (dismissed):
Re-review of the commit pushed since my last round — see the [summary comment](https://github.com/linuxfoundation/lfx-self-serve/pull/1290#issuecomment-5181838863). Both prior minor findings (flagResolved providerReady, committeeUid rename) cleanly resolved and independently re-verified across security, correctness, and style passes — zero new findings. One non-blocking test-coverage suggestion (meeting-coordinator-only access path lacks e2e fixture coverage) called out for a possible follow-up.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -761,6 +864,210 @@ export class CommitteeViewComponent {
...
+        // meetingCoordinatorLoading folded in here (not left as a separate downstream check) so the
+        // linkedSignal holds its previous settled value through that fetch's window too, the same as
+        // it already does for myRoleLoading().
+        loading: this.myRoleLoading() || this.meetingCoordinatorLoading(),
+        eligible:
+          this.myRole() !== null ||
+          this.canEdit() ||
+          this.canReview() ||
+          this.isCallerInAuditorList(this.committee()?.inherited_auditors) ||
+          this.meetingCoordinator(),
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed and fixed in df8227b6 — removed `myRole() !== null` from both `canAccessEngagement`'s `eligible` computation and `initMeetingCoordinator`'s `needed` gate. You're right that `committee#auditor` has no plain-member leg — verified against `committee-read-access.helper.ts:14-25` and its doc comment, which explicitly calls out that rank-and-file members are excluded by design.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.html
```diff
@@ -422,6 +428,11 @@ <h1 class="font-display font-light text-[20px]" data-testid="committee-view-name
               [membersLoading]="membersLoading()"
               [invites]="invites()"
               [invitesLoading]="invitesLoading()"
+              [engagementEnabled]="engagementMetricsEnabled()"
+              [engagement]="engagement()"
+              [engagementLoading]="engagementLoading()"
+              [engagementWindow]="engagementWindow()"
+              (engagementWindowChange)="onEngagementWindowChange($event)"
```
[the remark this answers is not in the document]
manishdixitlfx: Confirmed and fixed in df8227b6 — added `engagementAccessible` input to `CommitteeMembersComponent` (mirroring the one `CommitteeOverviewComponent` already has) and gated the Meetings/Engagement columns, header cells, and window controls on `engagementEnabled() && engagementAccessible()` instead of the flag alone. Wired `[engagementAccessible]="canAccessEngagement()"` from `committee-view.component.html`, same as the existing Overview binding. `emptyStateColspan` updated to match so the empty-state row still spans the right column count.

dealako (approved):

PR #1291, opened 2026-08-03
feat(projects): add lens-redirect endpoint for email deep links (LFXV2-2952)

audigregorie:
# Summary

Adds a public `GET /public/api/projects/:slug/lens-redirect/:resource` endpoint that 302-redirects email-link recipients to the correct lens-prefixed resource page (`/foundation/<resource>` or `/project/<resource>`), deriving the lens from the project's own attributes rather than the recipient's active lens. Email deep links (votes, meetings, mailing-lists, etc.) currently land on a flat route that doesn't carry project context; this endpoint resolves the lens server-side so the recipient arrives on the right page without a client-side lens switch.

Ticket: LFXV2-2952

### Before
Email links for votes, meetings, mailing-lists and other project resources landed on a flat route with no project or lens context, so the recipient arrived on a generic page and had to switch lenses client-side to reach the right place. There was no redirect endpoint handling this at all, and no concept of falling back gracefully when a project lookup failed.

### After
Email links now hit a public redirect endpoint that looks up the project from its slug, works out whether it's a foundation or non-foundation project, and 302-redirects the recipient to the correct lens-prefixed page (`/foundation/...` or `/project/...`) with the project context attached, so they land on the right page directly with no client-side lens switch. If the project can't be found or the upstream lookup fails, it safely falls back to the old flat route instead of showing the email recipient an error. Only a fixed set of known resource types (votes, meetings, mailing-lists, groups, documents, surveys, newsletters, settings) are accepted; anything else is rejected so the endpoint can't be abused as an open redirect.

# Technical changes

## Project controller
(`apps/lfx-one/src/server/controllers/project.controller.ts`)
- Add `getLensRedirect` controller method that validates `slug` (`^[A-Za-z0-9_-]+$`) and `resource` (allowlisted), resolves the project via `projectService.getProjectIdBySlug`, then `getProjectById`, and 302-redirects to `/<lens>/<resource>?project=<slug>` with `computeIsFoundation` choosing the lens.
- Export `LENS_REDIRECT_RESOURCES` allowlist (`votes`, `meetings`, `mailing-lists`, `groups`, `documents`, `surveys`, `newsletters`, `settings`). The `:resource` param is matched against this set and never echoed raw into the `Location` header, so the endpoint cannot become an open redirect.
- Acquire an M2M token (`generateM2MToken`) when no bearer token is present, since the endpoint is anonymous (public route, no user session).
- On unresolvable slug or upstream fetch failure, fall back to the flat `/<resource>` route and log a warning instead of surfacing an error to the email recipient.
- Set `Cache-Control: no-store` on every redirect response so email links always re-resolve the lens.
- Add structured `logger.startOperation` / `logger.success` / `logger.warning` instrumentation around the redirect flow.

## Public projects route
(`apps/lfx-one/src/server/routes/public-projects.route.ts`)
- Register `GET /:slug/lens-redirect/:resource` on the public projects router, mapped to `projectController.getLensRedirect`. Route lives under `/public/api/projects/` so it stays anonymous (no user session required).

### DIFF
--- a/apps/lfx-one/src/server/controllers/project.controller.spec.ts
+++ b/apps/lfx-one/src/server/controllers/project.controller.spec.ts
@@ -1,0 +1,226 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import type { Project } from '@lfx-one/shared/interfaces';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const SLUG = 'cncf';
+const PROJECT_UID = 'project-2222';
+
+// Hoisted, per-test-controllable mocks. The `@lfx-one/shared/*` path alias isn't wired
+// into vitest (see vitest.config.ts), so the shared barrels the controller imports at
+// module load are stubbed here. `computeIsFoundation` is mocked so each test can drive
+// the foundation-vs-project branch without constructing a full Project graph.
+const { computeIsFoundationMock, generateM2MTokenMock, projectSvc } = vi.hoisted(() => ({
+  computeIsFoundationMock: vi.fn(),
+  generateM2MTokenMock: vi.fn(),
+  projectSvc: {
+    getProjectIdBySlug: vi.fn(),
+    getProjectById: vi.fn(),
+  },
+}));
+
+vi.mock('@lfx-one/shared/utils', () => ({
+  computeIsFoundation: computeIsFoundationMock,
+  isFileTypeAllowed: vi.fn(),
+  isUuid: vi.fn(),
+}));
+vi.mock('@lfx-one/shared/constants', () => ({
+  ALLOWED_FILE_TYPES: [],
+  LENS_REDIRECT_RESOURCES: new Set<string>(['votes', 'meetings', 'mailing-lists', 'groups', 'documents', 'surveys', 'newsletters', 'settings']),
+}));
+vi.mock('@lfx-one/shared/enums', () => ({ MeetingVisibility: { PUBLIC: 'public', PRIVATE: 'private' } }));
+// validation.helper pulls in a heavy shared/constants + shared/enums graph; stub it
+// wholesale so only the controller's getLensRedirect path loads.
+vi.mock('../helpers/validation.helper', () => ({ getStringQueryParam: vi.fn(), validateUidParameter: vi.fn(() => true) }));
+
+vi.mock('../services/project.service', () => ({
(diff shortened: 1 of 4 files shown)


ahmedomosanya (approved):
**LGTM** — safe, well-tested, and the outstanding bot feedback already looks resolved in `0eca8e6`/`31cb978b` (fallback preserves `?project=`, both fallbacks log warning+success, constant moved to `@lfx-one/shared`, drift guard reads the real allowlist). Nice work on the guards: the slug regex + exact-match resource allowlist mean the destination is only ever `foundation`/`project` + an allowlisted segment, so it can't become an open redirect, and pairing `auth: 'optional'` with the M2M lookup and the public rate limiter is a clean anonymous design.

Two optional, non-blocking follow-ups:

**Suggestion — extend the drift guard to the flat fallback route.** The guard asserts each resource has a `foundation/<x>` and `project/<x>` route, but the not-found/error paths 302 to the *flat* `/<resource>`:

```ts
res.redirect(302, `/${resource}?project=${encodeURIComponent(slug)}`);
```

If a flat route (e.g. `path: 'votes'`) is later renamed/removed, the fallback would 302 to a 404 with nothing catching it. Since we already read the route table, could we also assert the bare segment?

```ts
it.each([...LENS_REDIRECT_RESOURCES])('resource "%s" has foundation, project, and flat fallback routes', (resource) => {
  expect(routesSrc).toContain(`foundation/${resource}`);
  expect(routesSrc).toContain(`project/${resource}`);
  expect(routesSrc).toContain(`path: '${resource}'`);
});
```

**Question — `getProjectById(req, uid, false)` for the lens decision.** Skipping the FGA round-trip here is the right call for an anonymous endpoint. Just confirming the M2M `/projects/:uid` fetch returns the fields `computeIsFoundation` reads (`stage`, `legal_entity_type`, `funding`, `funding_model`, `name`) for all projects — if any of those are ever omitted for a private project, the lens would silently default to `/project/...`. Acceptable as a default, just flagging it.

MRashad26 (approved):
Clean, well-designed server-side addition. Open-redirect prevention is solid: resource is allowlist-checked (`LENS_REDIRECT_RESOURCES.has(resource)`) and slug is regex-validated before either appears in the `Location` header, with `encodeURIComponent` on top. Fallback-to-flat-route on slug-not-found and on upstream error is the right UX for email-link recipients. `Cache-Control: no-store` correctly set on every redirect path. Drift guard in the spec is a nice safety net. All code-enforcer checks pass.

PR #1302, opened 2026-08-04
fix(cla): download My CLAs PDF without orphaned blank tab

### LINKED ISSUE #1228: My CLAs PDF download: avoid orphaned blank tab; add loading state
## Context

The My CLAs "Download PDF" action (`profile-clas.component.ts` `onDownload`) opens a **blank tab synchronously on click**, then sets `tab.location.href` once the presigned URL resolves from `GET /v4/my-clas/{signatureID}/pdf`. The blank-tab-first pattern is a deliberate popup-blocker workaround (the URL isn't known until the async call returns) and is correct in principle.

Two UX rough edges observed:

1. **Blank tab flash / no feedback in the new tab.** While the presigned URL resolves, the new tab sits empty. The origin tab shows a spinner, but the new tab has no loading indicator. On a slow gateway call this reads as a broken/empty tab.
2. **Orphaned blank tab when S3 serves the object as a download.** Whether the PDF renders inline or triggers a download depends on the S3 object's `Content-Disposition` and the browser's PDF setting. If it's served as `attachment`, the file downloads and the user is left with an empty tab to close manually.

## Proposed change

- For the **download** case, prefer a programmatic anchor click (`<a download href=url>`) once the URL resolves — the file downloads with no tab at all, eliminating the orphaned-tab problem.
- If inline viewing is intended instead, keep the new-tab approach but add a loading state in the opened tab (e.g. a small interstitial) so it isn't blank while resolving.
- Decide the intended behavior (download vs. inline view) explicitly and make it consistent.

## Notes

- Not a regression: this component predates the endpoint switch in #1226 (which only changed the server call to `/v4/my-clas/{id}/pdf`; the `{url, expiresInSeconds}` contract is unchanged).
- Presigned URL TTL is 900s (15 min); no change needed there.

Part of EPIC #1157.

ahmedomosanya:
## Summary

- **What changed**: `ProfileClasComponent.onDownload` no longer opens a blank tab; it waits for the presigned URL then calls new shared `downloadFromUrl` (`packages/shared/src/utils/file.utils.ts`). Origin-tab spinner + error toast kept.
- **Why**: [lfx-self-serve#1228](https://github.com/linuxfoundation/lfx-self-serve/issues/1228) — orphaned blank tab / no loading feedback on My CLAs PDF download.
- **Non-obvious**: Cross-origin S3 often ignores the `download` attribute; when the object is `Content-Disposition: attachment` the browser still downloads without navigating (the orphan-tab case this fixes). Upstream EasyCLA presign does not set `ResponseContentDisposition` — confirm on the preview deploy that the file downloads rather than navigating the SPA away.

## Test plan

- [ ] Open `/profile/clas`, click **Download PDF** on an ICLA — file should download with no new tab.
- [ ] Button spinner shown during the URL fetch.
- [ ] Error toast shown if the call fails.
- [ ] Verify on preview deploy once labeled.

Closes #1228

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
@@ -74,16 +74,18 @@ export class ProfileClasComponent {
   /**
    * Resolves the ICLA's short-lived PDF URL and triggers a file download (no new tab).
    * Origin-tab spinner stays on while the presigned URL resolves; see #1228.
+   * Filename includes project/CLA-group name so multi-project downloads don't collide.
    */
-  protected onDownload(signatureId: string): void {
+  protected onDownload(agreement: MyClaAgreement): void {
     if (this.downloadingId()) return;
 
-    this.downloadingId.set(signatureId);
+    this.downloadingId.set(agreement.id);
 
-    this.myClasService.getPdfUrl(signatureId).subscribe({
+    this.myClasService.getPdfUrl(agreement.id).subscribe({
       next: ({ url }) => {
         this.downloadingId.set(null);
-        downloadFromUrl(url, 'signed-cla.pdf');
+        const label = agreement.projectName || agreement.claGroupName || 'cla';
+        downloadFromUrl(url, `${label}-signed.pdf`);
       },
       error: () => {
         this.downloadingId.set(null);
--- a/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
@@ -6,7 +6,7 @@ import { ChangeDetectionStrategy, Component, computed, inject, PLATFORM_ID, sign
 import { toSignal } from '@angular/core/rxjs-interop';
 import { RouterLink } from '@angular/router';
 import type { MyClaAgreement, MyClasState } from '@lfx-one/shared/interfaces';
-import { isMyClasEmpty } from '@lfx-one/shared/utils';
+import { downloadFromUrl, isMyClasEmpty } from '@lfx-one/shared/utils';
 import { MessageService } from 'primeng/api';
 import { ToastModule } from 'primeng/toast';
 import { BehaviorSubject, catchError, of, switchMap } from 'rxjs';
@@ -22,8 +22,8 @@ import { MyClasService } from '@services/my-clas.service';
  * Read-only "My CLAs" Profile tab (Me lens). Renders the user's currently-valid signed CLAs
  * (ICLA + ECLA) from `/v4/my-clas` in a single table (Project / Type / Signed / Document) per
  * the approved M1 mockup — no status column, because the BFF filters to valid-only so every row
- * is valid. Agreements are resolved server-side from the session identity; ICLA PDFs open via
- * short-lived URLs.
+ * is valid. Agreements are resolved server-side from the session identity; ICLA PDFs download via
+ * short-lived URLs (no new tab).
  */
 @Component({
   selector: 'lfx-profile-clas',
@@ -72,34 +72,26 @@ export class ProfileClasComponent {
   }
 
   /**
-   * Resolves the ICLA's short-lived PDF URL and opens it. A blank tab is opened
-   * synchronously on click (before the async request) so the browser attributes it to the
-   * user gesture and does not block the popup; its location is set once the URL resolves.
+   * Resolves the ICLA's short-lived PDF URL and triggers a file download (no new tab).
+   * Origin-tab spinner stays on while the presigned URL resolves; see #1228.
    */
   protected onDownload(signatureId: string): void {
     if (this.downloadingId()) return;
 
-    // Open the blank tab WITHOUT `noopener` so we retain the window handle to redirect once the
-    // URL resolves (`noopener` makes window.open return null). Sever the opener link manually for
-    // the same reverse-tabnabbing protection `noopener` would provide.
-    const tab = window.open('', '_blank');
-    if (tab) tab.opener = null;
     this.downloadingId.set(signatureId);
 
     this.myClasService.getPdfUrl(signatureId).subscribe({
       next: ({ url }) => {
         this.downloadingId.set(null);
-        if (tab) {
-          tab.location.href = url;
-        } else {
-          // Popup was blocked despite the synchronous open — fall back to a same-tab navigation.
-          window.location.href = url;
-        }
+        downloadFromUrl(url, 'signed-cla.pdf');
       },
       error: () => {
         this.downloadingId.set(null);
-        tab?.close();
-        this.messageService.add({ severity: 'error', summary: 'Download failed', detail: 'Could not open the signed document. Please try again.' });
+        this.messageService.add({
+          severity: 'error',
+          summary: 'Download failed',
+          detail: 'Could not download the signed document. Please try again.',
+        });
       },
     });
   }
--- a/packages/shared/src/utils/file.utils.ts
+++ b/packages/shared/src/utils/file.utils.ts
@@ -317,7 +317,6 @@ export function downloadFromUrl(url: string, filename?: string): void {
   if (filename) {
     anchor.download = sanitizeFilename(filename);
   }
-  anchor.rel = 'noopener';
   anchor.style.display = 'none';
   document.body.appendChild(anchor);
   anchor.click();
--- a/packages/shared/src/utils/file.utils.ts
+++ b/packages/shared/src/utils/file.utils.ts
@@ -306,6 +306,24 @@ export function rowsToCsv(rows: ReadonlyArray<ReadonlyArray<string | number>>): 
   return rows.map((row) => row.map(escapeCsvCell).join(',')).join('\r\n');
 }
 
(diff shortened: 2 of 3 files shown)


packages/shared/src/utils/file.utils.ts
```diff
@@ -306,6 +306,24 @@ export function rowsToCsv(rows: ReadonlyArray<ReadonlyArray<string | number>>):
...
+/** Trigger a browser file download from a URL via `<a download>` (no new tab). No-op during SSR. */
+export function downloadFromUrl(url: string, filename?: string): void {
+  if (typeof document === 'undefined') {
+    return;
+  }
+
+  const anchor = document.createElement('a');
+  anchor.href = url;
+  if (filename) {
+    anchor.download = sanitizeFilename(filename);
```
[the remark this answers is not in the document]
ahmedomosanya: Acknowledged — known caveat called out in the PR test plan. EasyCLA's presign (`GetPresignedURL`) does not set `ResponseContentDisposition`, so `<a download>` is best-effort for this cross-origin URL. Leaving this open pending verification on the deploy-preview (https://ui-pr-1302.dev.v2.cluster.linuxfound.info/profile/clas): if the SPA navigates instead of downloading, follow-up is either EasyCLA attachment disposition or a same-origin BFF proxy — not resolving until that check lands.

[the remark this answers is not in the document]
ahmedomosanya: Verified on deploy-preview (https://ui-pr-1302.dev.v2.cluster.linuxfound.info/profile/clas, 2026-08-04): click Download PDF → origin-tab spinner for a few seconds → spinner clears → file downloads; SPA does **not** navigate away. Cross-origin `<a download>` is sufficient for this S3 response in practice; no BFF proxy / EasyCLA disposition change needed for #1228.

MRashad26 (dismissed):
Clean fix. `downloadFromUrl` in `file.utils.ts` uses the correct SSR guard, `sanitizeFilename`, `rel=noopener` (equivalent protection to the old manual `tab.opener = null`), and the standard `appendChild`/`click`/`removeChild` pattern. Spinner and error toast behavior preserved. The cross-origin `download`-attribute caveat is correctly flagged for manual preview validation. All code-enforcer checks pass.

audigregorie (commented):
## Code Review Summary

Clean, minimal fix that removes the orphaned-blank-tab anti-pattern from the My CLAs PDF download. The shared `downloadFromUrl` util is well-placed and consistent with the sibling `downloadCsv`. The one material concern is that the core mechanism — `<a download>` on a cross-origin S3 URL — is explicitly unverified (the PR description flags it), and if upstream EasyCLA doesn't set `Content-Disposition: attachment` the failure mode (SPA navigates away to the PDF) is a worse regression than the bug being fixed. That should be resolved before merge.

### Major — outside the diff

- **Commit subject missing ticket suffix**: `fix(cla): download My CLAs PDF without orphaned blank tab` has no `(LFXV2-1228)` suffix. Branch is `fix/easycla-my-clas-pdf-download-1228`; per repo `commit-workflow.md` + global rules the subject should end with `(LFXV2-1228)`. Will surface as a SHOULD_FIX in `/lfx-self-serve-pr-readiness`.
- **No tests**: `downloadFromUrl` and the updated `onDownload` path have no unit tests; `file.utils.ts` has no spec file at all. A minimal test for `downloadFromUrl` (SSR no-op + browser click) and a component test for `onDownload` (spinner set/cleared, error toast) would lock in the contract.

### What's done well

- Correctly eliminates the synchronous `window.open('')` anti-pattern that produced the orphaned blank tab.
- Reuses the existing `sanitizeFilename` helper rather than reinventing filename safety.
- SSR guard (`typeof document === 'undefined'`) is consistent with the sibling `downloadCsv`.
- Spinner state (`downloadingId`) is cleared in both `next` and `error` paths before the side effect — no stuck spinner on failure.
- Shared util placement in `@lfx-one/shared/utils` is correct and reusable for the other download sites.
- Error toast copy updated from "open" to "download" to match the new behavior.

Rating: ⭐⭐⭐ (3/5)

packages/shared/src/utils/file.utils.ts
audigregorie, line 318: **Major — Cross-origin `download` attribute is ignored; SPA may navigate away**

The `download` attribute on `<a>` is **ignored by browsers for cross-origin URLs** unless the response sets `Content-Disposition: attachment`. The presigned S3 URL returned by EasyCLA (`cla.service.ts getPdfUrl` → `EasyClaMyClaPdf.url`) is cross-origin (S3 domain), and the BFF passes it through without setting `Content-Disposition`. The PR description itself flags this: *"Upstream EasyCLA presign does not set `ResponseContentDisposition` — confirm on the preview deploy that the file downloads rather than navigating the SPA away."*

If S3 serves the object inline (no `Content-Disposition`), clicking this anchor **navigates the current SPA tab** to the S3 URL — a regression *worse* than the original orphaned-blank-tab bug, because the user loses their place in the app entirely. The core fix's effectiveness is unverified and the failure mode is more severe than the bug being fixed.

Options to make this robust regardless of upstream headers:

```suggestion
  // Cross-origin URLs ignore the `download` attribute unless the response sets
  // Content-Disposition: attachment. Fetch → Blob → object URL makes the
  // download attribute authoritative (same-origin blob URL).
  // TODO: confirm EasyCLA S3 sets Content-Disposition, or switch to blob/proxy.
  anchor.href = url;
  if (filename) {
    anchor.download = sanitizeFilename(filename);
  }
```

Either (a) confirm on the preview deploy that S3 returns `Content-Disposition: attachment` before merging and document it here, (b) fetch the URL → create a Blob → `URL.createObjectURL` + `<a download>` (same-origin blob honors `download`), or (c) have the BFF proxy/stream the PDF with `Content-Disposition: attachment` set server-side — the pattern already used by committee/project document downloads (`committee.controller.ts:1187`, `project.controller.ts:808`).

packages/shared/src/utils/file.utils.ts
```diff
@@ -306,6 +306,24 @@ export function rowsToCsv(rows: ReadonlyArray<ReadonlyArray<string | number>>):
...
+  }
+  anchor.rel = 'noopener';
```
audigregorie: **Minor — `rel="noopener"` is a no-op here**

`rel="noopener"` only affects `target="_blank"`/`_new` navigations. With no `target` set (defaults to `_self`) it does nothing — this is leftover conceptual baggage from the old `window.open` reverse-tabnabbing logic and is misleading; a future reader may assume it provides protection it doesn't.

```suggestion
  anchor.style.display = 'none';
```

Remove the line, or add `target="_blank"` if a new-tab fallback is actually desired (though that reintroduces the orphaned-tab problem this PR removes).

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -72,34 +72,26 @@ export class ProfileClasComponent {
...
     this.myClasService.getPdfUrl(signatureId).subscribe({
       next: ({ url }) => {
         this.downloadingId.set(null);
-        if (tab) {
-          tab.location.href = url;
-        } else {
-          // Popup was blocked despite the synchronous open — fall back to a same-tab navigation.
-          window.location.href = url;
-        }
+        downloadFromUrl(url, 'signed-cla.pdf');
```
audigregorie: **Minor — Hardcoded `signed-cla.pdf` filename collides across projects**

Every ICLA downloads as `signed-cla.pdf` regardless of project. Downloading CLAs from multiple projects produces colliding filenames (the browser auto-suffixes `(1)`, `(2)`, but the user can't tell which file belongs to which project). The template already has `agreement.projectName` / `claGroupName` available.

Consider passing the agreement (or just the project name) into `onDownload` and building a descriptive filename:

```suggestion
        downloadFromUrl(url, `${agreement.projectName || agreement.claGroupName || 'cla'}-signed.pdf`);
```

This requires changing `onDownload(signatureId)` to accept the agreement (or projectName) — the template call `(onClick)="onDownload(agreement.id)"` becomes `(onClick)="onDownload(agreement)"`.

packages/shared/src/utils/file.utils.ts
ahmedomosanya: Verified on deploy-preview (https://ui-pr-1302.dev.v2.cluster.linuxfound.info/profile/clas, 2026-08-04): click Download PDF → origin-tab spinner → file downloads; SPA does **not** navigate away. EasyCLA's `GetPresignedURL` still omits `ResponseContentDisposition`, but `<a download>` is sufficient for this S3 response in practice — no BFF proxy / EasyCLA disposition change needed for #1228.

packages/shared/src/utils/file.utils.ts
ahmedomosanya: Addressed in `3225f357` — removed the no-op `rel="noopener"` from `downloadFromUrl` (same-tab `<a download>`, no `target="_blank"`).

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
ahmedomosanya: Addressed in `3225f357` — `onDownload` now takes the agreement and downloads as `${projectName || claGroupName || 'cla'}-signed.pdf` (via `sanitizeFilename`).

audigregorie (approved):

PR #1305, opened 2026-08-04
feat(org-lens): all-time range for project detail influence trend + sparklines (LFXV2-2867)

ahmedomosanya:
## Summary

Adds a true **"All time"** range to the Org Lens **Project Detail** influence-trend chart and card sparklines ([LFXV2-2867](https://linuxfoundation.atlassian.net/browse/LFXV2-2867)). The BFF now serves the new adaptive **lifetime-bucketed** warehouse read models for `?range=all`, alongside the existing trailing-monthly source for `1y`/`2y`.

[lf-dbt #2711](https://github.com/linuxfoundation/lf-dbt/pull/2711) merged 2026-08-12. The three lifetime tables are in prod `ANALYTICS.PLATINUM_LFX_ONE`: `ORG_LENS_PROJECT_DETAIL_TREND_LIFETIME` (32.4M rows), `ORG_LENS_PROJECT_DETAIL_SPARKLINES_LIFETIME` (16.8M rows), and `ORG_LENS_PROJECT_INFLUENCE_LIFETIME_BUCKET_SPINE` (29.5M rows). KubeVirt has 9 multi-year buckets spanning 2009–2026.

## Changes

- **BFF** (`org-lens-project-detail.service.ts`): `getTrendBlock` / `getInfluenceBlock` branch on `range`. `all` reads the lifetime-bucketed trend/sparkline tables + the bucket spine and emits a variable `periods[]` axis (the project's shared adaptive bucket calendar, ≤ 12 points); `1y`/`2y` keep the trailing-monthly source + client-derived month axis. New parameterized fetches (`fetchTrendLifetime`, `fetchSparklinesLifetime`, `fetchLifetimeAxis`) and bucket-label formatting (`MMM YYYY` / `Q# YYYY` / `YYYY` / `YYYY–YYYY`). Cached `all` influence/trend payloads that lack a non-empty `periods[]` of strings are rejected so a pre-deploy 36-month Valkey entry cannot be served for the one-hour TTL.
- **Cache keys**: the trend/influence cache keys now include `range`, so an all-time request is never served a `1y`/`2y` series. The influence cache key already included `range` on `main`, so adding range did not invalidate existing `all` entries; the trend key did (range was new there). The ≤12 cap stays an e2e/product assertion, not a cache predicate.
- **Controller**: threads `range` into `getTrendBlock`.
- **Shared contract**: additive optional `periods?: string[]` on `OrgLensTrendBlock` / `OrgLensInfluenceBlock` (present only for `all`); `PD_TIME_RANGE_MONTHS.all` documented as variable-length.
- **Frontend component**: consumes `periods[]` for the chart axis labels.
- **E2E**: All time on Project Detail requests `range=all` for influence + trend and keeps sparkline/trend series aligned with the adaptive axis.

## Test plan

- [x] `yarn check-types`, `yarn lint:check`, `yarn build` — all pass locally.
- [x] Verified locally against dev-seeded lifetime tables: all-time trend + sparklines render with adaptive bucket axis (≤ 12 points; long-lived projects fold to multi-year).
- [x] Re-verify against prod once [lf-dbt #2711](https://github.com/linuxfoundation/lf-dbt/pull/2711) lands.

Ticket: [LFXV2-2867](https://linuxfoundation.atlassian.net/browse/LFXV2-2867)

[LFXV2-2867]: https://linuxfoundation.atlassian.net/browse/LFXV2-2867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-2867]: https://linuxfoundation.atlassian.net/browse/LFXV2-2867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -411,47 +448,61 @@ export class OrgLensProjectDetailService {
...
+    // 1y/2y read the trailing monthly sparkline model and slice client-side over a fixed month axis;
+    // `all` reads the adaptive lifetime-bucketed model and carries its own variable `periods[]` axis
+    // (LFXV2-2867 D9) — the same shared bucket axis the trend chart uses (D7).
+    const [cardRows, viewing, sparkData] = await Promise.all([
       this.fetchCards(orgUid, slug, timeRangeType),
-      this.fetchSparklines(orgUid, slug),
       // Only the viewing org's own row is needed here (for the section-title band chips), so fetch
       // that single row rather than the whole board — the board itself is now paged separately.
       this.fetchViewingLeaderboardRow(orgUid, slug, timeRangeType).catch(() => null),
+      range === 'all' ? this.fetchLifetimeSparklineData(orgUid, slug) : this.fetchMonthlySparklineData(orgUid, slug),
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 0ea8c81d — `all` cache hits now require a non-empty `periods[]`, so a pre-deploy 36-month Valkey entry cannot be served for the one-hour TTL.

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
```diff
@@ -879,12 +881,22 @@ export class OrgProjectDetailComponent {
...
-    const sparkline = card.sparkline.slice(-months);
-    const projectSparkline = card.projectSparkline.slice(-months);
-    const labels = this.monthLabels.slice(-sparkline.length);
+    // "All time" (periods present): the payload already carries the adaptive-bucket axis, so render
+    // the series as-is against the server-emitted labels. 1y/2y: slice the trailing months and derive
+    // the fixed month labels client-side (unchanged).
+    const useBuckets = periods !== null && periods.length > 0;
+    const sparkline = useBuckets ? card.sparkline : card.sparkline.slice(-months);
+    const projectSparkline = useBuckets ? card.projectSparkline : card.projectSparkline.slice(-months);
+    const labels = useBuckets ? periods.slice(-sparkline.length) : this.monthLabels.slice(-sparkline.length);
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in 0ea8c81d — Playwright now clicks All time, asserts `range=all` on influence+trend, and checks sparkline/trend series align with the adaptive axis (≤12).

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
```diff
@@ -225,11 +225,36 @@ export class OrgProjectDetailComponent {
...
+  private readonly lifetimeTrimStart = computed(() => {
+    const data = this.influenceState().data;
+    if (!data?.periods?.length) return 0;
+    const cards = [...(data.technical ?? []), ...(data.ecosystem ?? [])];
+    const width = cards.reduce((max, card) => Math.max(max, card.sparkline.length), 0);
+    for (let i = 0; i < width; i++) {
+      if (cards.some((card) => (card.sparkline[i] ?? 0) > 0 || (card.projectSparkline[i] ?? 0) > 0)) return i;
+    }
+    return 0;
   });
```
audigregorie, line 258: **Major — `lifetimeTrimStart` creates a diamond dependency on `influenceState()`**

`lifetimeTrimStart` is a `computed()` reading `this.influenceState().data`. `technicalCards` and `ecosystemCards` are also `computed()`s reading `this.influenceState().data` AND `this.lifetimeTrimStart()`. This forms a diamond: `influenceState` → both `lifetimeTrimStart` and `technicalCards`/`ecosystemCards`.

Angular's computed signals handle diamond dependencies correctly (no infinite loop), so this is functionally safe. But on every `influenceState` change, `lifetimeTrimStart` recomputes the trim index by scanning all 14 cards × ≤12 buckets, and then `technicalCards`/`ecosystemCards` re-derive the card VMs — a double traversal. Given the small bounds (≤168 cells) this is acceptable, but consider folding the trim computation directly into the card-mapping computed (or memoizing) to avoid the extra pass.

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
```diff
@@ -962,29 +998,47 @@ export class OrgProjectDetailComponent {
...
+
+    // All time: the lifetime bucket spine can start well before the project's first real activity,
+    // which renders as leading 0% dead space. Open the axis on the first bucket that has data.
+    // 1y/2y keep the whole requested window — an empty leading month there is real information.
+    const firstWithData = useBuckets ? allTotals.findIndex((total) => total > 0) : 0;
+    if (firstWithData === -1) return { labels: [], datasets: [] };
+
+    const labels = allLabels.slice(firstWithData);
+    const entries = allEntries.map((e) => ({ name: e.name, data: e.data.slice(firstWithData) }));
+    const totals = allTotals.slice(firstWithData);
```
audigregorie, line 1035: **Major — Influence cards and trend chart use different leading-trim criteria under "All time"**

`lifetimeTrimStart` (influence cards, line 249) trims leading buckets where **no card has any non-zero org or project value** across all 14 cards. `firstWithData` here (trend chart) trims leading buckets where the **total combined influence across all orgs is 0**. These are different metrics, so under "All time" the influence card sparklines and the stacked trend chart can open on different first buckets — e.g. cards start at "2014" while the trend starts at "2016" — producing inconsistent axis labels across the same tab.

The e2e test only compares the raw server-emitted `periods[]` lengths (which match), so it won't catch this. Consider aligning both trims to the same criterion (both use the trend's total-activity rule, or both use the card-activity rule), or document the intentional divergence in the component so it's clear this is by design.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -1139,6 +1319,26 @@ export class OrgLensProjectDetailService {
...
+      if (!accountId) continue;
+      let series = byAccount.get(accountId);
+      if (!series) {
+        series = { accountId, orgName: row.ORG_NAME ?? '', orgLogoUrl: row.ORG_LOGO_URL ?? '', combined: [] };
+        byAccount.set(accountId, series);
+      }
+      series.combined.push(this.round1(this.num(row.COMBINED_INFLUENCE_SCORE)));
+    }
+    return byAccount;
+  }
```
audigregorie, line 1340: **Major — `buildTrendLifetimeByAccount` trusts the warehouse density contract without defensive padding**

The comment asserts the read model is "dense per (account, bucket) over the project's shared axis," so every series should equal `periods.length`. But the downstream `buildTrendSeries` (unchanged existing code, not in this diff) uses `rest.reduce((max, s) => Math.max(max, s.combined.length), 0)` for the "All others" length and `s.combined[i] ?? 0` for missing indices — it never pads the **named** orgs to the axis length.

If the warehouse contract is ever violated (a sparse row, a backfill gap), a named org's series would be shorter than `periods[]` and misaligned against the server-emitted labels (oldest bucket dropped silently). The component's `padStart` left-pads to the max series length, but `periods.slice(-len)` would then drop leading labels — so the series would render against the wrong bucket labels.

Consider padding each named series to `periods.length` here (or in `buildTrendSeries`), or asserting the invariant with a dev-mode check so a contract regression surfaces immediately rather than as a silent misalignment.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -1606,6 +1818,11 @@ export class OrgLensProjectDetailService {
     return Array.isArray((value as OrgLensTrendBlock).trend);
   }
 
+  private static hasAdaptivePeriods(value: unknown): boolean {
+    const periods = (value as { periods?: unknown }).periods;
+    return Array.isArray(periods) && periods.length > 0 && periods.every((label) => typeof label === 'string');
+  }
```
audigregorie, line 1824: **Minor — `hasAdaptivePeriods` validator can cause a cache-thrash loop for `all` projects with an empty bucket spine**

For `range === 'all'`, the cache validator requires `hasAdaptivePeriods(value)` — `periods.length > 0`. If a project has no rows in the bucket spine, `fetchLifetimeAxis` returns `[]`, the BFF emits `periods: []`, and the block is cached. On the next request, the validator rejects the cached entry (empty periods), so the BFF re-fetches and re-caches — every request hits Snowflake.

This is an edge case (projects with no lifetime buckets likely have no data at all), but the validator should either allow empty-period `all` blocks (they're valid "no data" responses) or the BFF should skip caching when `periods` is empty, to avoid the cache-thrash.

```suggestion
  private static hasAdaptivePeriods(value: unknown): boolean {
    const periods = (value as { periods?: unknown }).periods;
    return Array.isArray(periods) && periods.every((label) => typeof label === 'string');
  }
```

(Dropping the `length > 0` requirement so empty-but-valid `all` blocks can be served from cache.)

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -762,6 +886,62 @@ export class OrgLensProjectDetailService {
...
+        return `${OrgLensProjectDetailService.shortMonths[startDate.getUTCMonth()]} ${year}`;
+      case 'quarterly':
+        return `Q${Math.floor(startDate.getUTCMonth() / 3) + 1} ${year}`;
+      case 'yearly':
+        return `${year}`;
+      case 'multi_year': {
+        const endDate = this.parseUtcDate(end);
+        const endYear = endDate === null ? year : endDate.getUTCFullYear();
+        return endYear > year ? `${year}\u2013${endYear}` : `${year}`;
+      }
```
audigregorie, line 930: **Minor — `multi_year` bucket with `end === null` renders as single-year `YYYY`, losing the range distinction**

If `BUCKET_END` is null (a spine row with a missing end date), a `multi_year` bucket renders as just `YYYY` — indistinguishable from a `yearly` bucket. This is a defensive fallback, but it could mislead users about the bucket's span. Consider logging a warning when a `multi_year` bucket has no end date, or rendering as `YYYY–?` to signal the gap so the label doesn't silently collapse to a single year.

packages/shared/src/constants/org-lens-project-detail.constants.ts
```diff
@@ -103,7 +103,15 @@ export const PD_TIME_RANGE_OPTIONS: { id: OrgLensLeaderboardTimeRange; label: st
...
-export const PD_TIME_RANGE_MONTHS: Record<OrgLensLeaderboardTimeRange, number> = { '1y': 12, '2y': 24, all: 36 };
+/**
+ * Trailing monthly window per range for the recent-monthly representation. `1y`/`2y` slice the
+ * trailing 12 / 24 monthly points client-side. `all` is NO LONGER a fixed length (was 36): under
+ * "All time" the payload carries a variable, adaptively-bucketed series (hard cap ≤ 12 points) with
+ * its own `periods[]` axis labels (see `OrgLensTrendBlock.periods`), so the client renders the
+ * payload as-is instead of slicing to a fixed count. The `all` entry is a defensive upper-bound
+ * sentinel only — any residual `slice(-months)` on an already-≤12 all-time array is a no-op.
+ */
+export const PD_TIME_RANGE_MONTHS: Record<OrgLensLeaderboardTimeRange, number> = { '1y': 12, '2y': 24, all: Number.MAX_SAFE_INTEGER };
```
audigregorie, line 114: **Minor — `Number.MAX_SAFE_INTEGER` sentinel for `all` is surprising at call sites**

Using `Number.MAX_SAFE_INTEGER` as the `all` sentinel works because `Array.prototype.slice(-MAX_SAFE_INTEGER)` returns the whole array, and the comment documents this. But the sentinel could trip up future callers who do arithmetic on the value (e.g. `months - 1` for label derivation would produce a huge number). Consider a named constant like `PD_ALL_TIME_SENTINEL = Number.MAX_SAFE_INTEGER` to make the intent self-documenting at call sites, or `Infinity` which is more obviously "not a real count."

apps/lfx-one/e2e/org-project-detail.spec.ts
```diff
@@ -357,6 +357,47 @@ test.describe('Org Project Detail — leaderboard row score-breakdown drawer', (
...
+    );
+    await page.getByTestId('project-detail-time-range-all').click();
+    await expect(page).toHaveURL(/range=all/);
+    const influenceBody = (await (await influenceAll).json()) as { periods?: string[]; technical?: { sparkline?: unknown[] }[] };
+    expect(Array.isArray(influenceBody.periods)).toBe(true);
+    expect(influenceBody.periods!.length).toBeGreaterThan(0);
+    expect(influenceBody.periods!.length).toBeLessThanOrEqual(12);
+    const sparkline = influenceBody.technical?.[0]?.sparkline;
+    expect(Array.isArray(sparkline)).toBe(true);
+    expect(sparkline!.length).toBe(influenceBody.periods!.length);
```
audigregorie, line 380: **Minor — E2E test does not assert the adaptive bucket label format**

The test asserts `periods` is a non-empty string array of length ≤ 12, and that series align with `periods`, but it never checks that the labels match the documented formats (`MMM YYYY`, `Q# YYYY`, `YYYY`, `YYYY–YYYY`). A label like `""` (empty, from `bucketLabel` returning `''` on an unparseable date) or a malformed string would pass. Consider adding:

```suggestion
    expect(Array.isArray(influenceBody.periods)).toBe(true);
    expect(influenceBody.periods!.length).toBeGreaterThan(0);
    expect(influenceBody.periods!.length).toBeLessThanOrEqual(12);
    expect(influenceBody.periods!.every((p) => /\d{4}/.test(p))).toBe(true);
```

to catch a regression in `bucketLabel` (e.g. an unparseable `BUCKET_START` producing empty labels).

audigregorie (commented):
## Code Review Summary

Solid implementation of the all-time adaptive-bucket axis — additive `periods?: string[]` contract, range-scoped cache keys, and a `hasAdaptivePeriods` validator that rejects stale pre-deploy Valkey entries are exactly the right defensive patterns. The shared bucket spine keeps trend and sparklines on a common axis.

### Major — outside the diff

- **`buildTrendSeries` does not defensively pad lifetime series to the axis length** — see the comment thread on `buildTrendLifetimeByAccount` for details. The existing `buildTrendSeries` (unchanged by this PR) trusts the warehouse density contract; if it's ever violated, named org series would misalign against `periods[]`.

### Minor — outside the diff

- **`parseUtcDate` Date passthrough** — the `if (value instanceof Date) return value` branch trusts the caller's UTC anchoring, while the string branch explicitly re-anchors via `Date.UTC(...)`. Consistent with the rest of the file (`toYearMonth`), so not a regression — just noting the asymmetry.

### What's done well

- Cache key range-scoping + `hasAdaptivePeriods` validator correctly handles the one-hour TTL window where stale 36-month `all` entries could otherwise be served.
- Shared bucket spine axis (`ORG_LENS_PROJECT_INFLUENCE_LIFETIME_BUCKET_SPINE`) guarantees trend and sparklines share the same `periods[]` — the e2e test verifies this alignment.
- Additive `periods?: string[]` on both `OrgLensInfluenceBlock` and `OrgLensTrendBlock` keeps 1y/2y clients unchanged with thorough interface docs.

MRashad26 (approved):
Reviewed with the code-standards-enforcer checklist (signal/computed patterns, shared interfaces/constants placement, cache-key/staleness correctness, foundation/org-scoped Snowflake queries, security, test coverage). No template files are touched by this PR, so the function-call-in-template checklist item doesn't apply here.

**What this does:** Adds a true "All time" range to the Org Lens Project Detail influence-trend chart and card sparklines. `1y`/`2y` keep reading the existing trailing-monthly source and slicing client-side over a fixed month axis; `all` now reads three new lifetime-bucketed warehouse read models (trend, sparklines, and a shared bucket spine) and emits a variable, server-computed `periods[]` axis (≤12 adaptive buckets: monthly → quarterly → yearly → multi-year as a project ages) that both the influence cards and the trend chart share.

**Verified correct:**
- **Cache-staleness guard is the right fix for the stated risk.** The PR body flags that a pre-deploy 36-month Valkey entry could otherwise be served for `all` for the remainder of its 1-hour TTL. `hasAdaptivePeriods` (checked only when `range === 'all'`) rejects any cached value without a non-empty string `periods[]`, and the cache key for both blocks now includes `range`, so a stale fixed-length entry can't leak into the new adaptive-axis contract. Correctly scoped to only add the extra check for `all` — `1y`/`2y` reads aren't penalized with a check they don't need.
- **Shared lifetime axis, not per-metric/per-org trimming.** `lifetimeTrimStart` (client) computes one shared leading-empty-bucket cutoff across every technical + ecosystem card, rather than letting each card trim independently — correctly reasoned in the comment: independent trimming would desync the cards' axes from each other and from the trend chart, which is exactly the kind of subtle chart-misalignment bug that's easy to miss in review. Server-side, `fetchLifetimeAxis` reads the bucket spine (not the sparser trend/sparkline tables) specifically so the axis stays complete even where a metric/org has no rows for some buckets — both trend and sparklines index against that same axis.
- **Null-vs-zero semantics preserved through the new path.** `lifetimeTrimStart`'s activity check treats `null` (avg-merge-time's genuine "no merged PRs this bucket" gap) as 0 for trim purposes via `?? 0`, which is the right call for "should this bucket be trimmed" without corrupting the actual per-card `null` values that flow through unchanged for rendering.
- **`buildStackedTrend`'s new leading-zero trim** (`firstWithData`) is gated on `useBuckets`, so `1y`/`2y` behavior is provably unchanged (`firstWithData` hardcoded to `0` there) — an empty leading month in a bounded window is real information and correctly left alone; only the open-ended lifetime axis gets the dead-space trim.
- **Parameterized queries throughout** (`fetchTrendLifetime`, `fetchSparklinesLifetime`, `fetchLifetimeAxis`) — all `?`-bound, no string interpolation of caller-controlled values.
- **`PD_TIME_RANGE_MONTHS.all` → `Number.MAX_SAFE_INTEGER`** is a safe sentinel: confirmed the only consumers are `.slice(-months)` calls gated behind `useBuckets` being false, and `Array.prototype.slice` clamps an out-of-range negative start to `0` rather than erroring, so a hypothetical `all` request that lands with an empty `periods[]` (fresh project, no buckets yet) degrades to "render the full unsliced lifetime array" instead of crashing.
- **Bucket label formatting** (`bucketLabel`) is a plain, testable pure function covering all four granularities including the multi-year `YYYY` collapse when start/end share a year — dates are parsed via a UTC-anchored regex extraction (`parseUtcDate`) rather than trusting the driver's tz handling, avoiding an off-by-one-day class of bug for edge-of-month buckets.
- **Shared interfaces/constants correctly placed** in `@lfx-one/shared/{interfaces,constants}`, and the new `periods?: string[]` field is documented as present only for `all` on both `OrgLensInfluenceBlock` and `OrgLensTrendBlock`, with the doc comments explicitly calling out the 1:1 alignment contract with `sparkline`/`combined`.

**Minor, non-blocking:** the new pure logic on the server side — `bucketLabel`'s four-branch formatting (especially the multi-year same-year collapse and the `default` fallback), `parseUtcDate`'s regex parsing, and `hasAdaptivePeriods`'s cache-guard predicate — has no unit-level test coverage; the only new test is the single e2e case, which asserts array lengths/alignment but not label *content* (e.g., that a monthly bucket actually renders `"Jan 2016"` rather than some other string) or the cache-rejection behavior directly. That said, neither this server service nor the component file has any existing spec file in the repo, so this matches the file's pre-existing (lack of) unit-test convention rather than being a regression this PR introduces — worth a look if/when this area gets unit coverage, not a reason to hold up this PR.

audigregorie (approved):

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
ahmedomosanya: Addressed in `2a8e124` (follow-up #1443) — both trims are reverted rather than aligned.

You're right that the criteria diverge, and the consequence is worse than inconsistent labels. The lifetime bucket spine exists precisely so that every chart and sparkline shares ONE axis anchored at the project's earliest activity of any kind, with a metric whose own data began later rendering leading zeros rather than starting its axis later. Trimming client-side broke that guarantee whichever criterion won, so aligning the two would have traded a visible inconsistency for a quieter one.

The dead space the trims hid is an upstream data problem, not a rendering one: the bucket spine anchor is unbounded, tracked in [LFXV2-3231](https://linuxfoundation.atlassian.net/browse/LFXV2-3231). Trimming removed the visible gap without correcting the bucket widths, which are derived from that same inflated span — so the chart read as correct while the axis stayed wrong. Reverting keeps the symptom visible until the anchor is bounded upstream.

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
ahmedomosanya: Addressed in `2a8e124` (follow-up #1443) — `lifetimeTrimStart` is removed outright rather than memoized or folded into the card mappers. The trim it fed was reverted because it broke the shared lifetime axis (see the sibling thread on the trend chart), so the diamond and its double traversal go away with the feature.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
ahmedomosanya: Addressed in `9736e84` (follow-up #1443) — good catch, and the asymmetry you spotted was real: the sparkline path already densifies over its axis in `denseSeries`, while the trend path trusted row order.

`buildTrendLifetimeByAccount` now takes the bucket axis and places each score **by bucket index** into a zero-filled array of `axis.length`, instead of pushing in row order. Every named series is therefore exactly `periods.length` long by construction rather than by convention, so a sparse row set or a backfill gap can no longer shift a series against the emitted labels. "All others" maxes over the named lengths, so it inherits the same length for free.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
ahmedomosanya: Addressed in `9736e84` (follow-up #1443) — took your suggestion and dropped the `length > 0` requirement.

Presence of the array is what rejects stale pre-deploy entries (those carry no `periods` at all), so nothing is lost: an empty-but-valid all-time block now serves from cache instead of re-querying Snowflake on every read for the life of the TTL. I documented why the emptiness check is deliberately absent, so it doesn't get "tightened" back later.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
ahmedomosanya: Addressed in `9736e84` (follow-up #1443) — a `multi_year` bucket whose end date is missing or unparseable now renders `YYYY–?` instead of collapsing to a bare `YYYY`, so the gap is visible rather than silently understating the span.

There is no logger wired into this service, so the label itself carries the signal. The new bucket-label format assertion in the E2E spec also rejects `YYYY–?`, which means a spine row like this surfaces in CI rather than only in the rendered axis.

packages/shared/src/constants/org-lens-project-detail.constants.ts
ahmedomosanya: Addressed in `6d55ea2` (follow-up #1443) — extracted `PD_ALL_TIME_WINDOW` with a doc comment stating it is not a real month count and must not be used in arithmetic, which is exactly the `months - 12` trap you described.

Kept `Number.MAX_SAFE_INTEGER` rather than switching to `Infinity`: it is more obviously "not a real count", but it does not survive `JSON.stringify` (it serialises to `null`), and this constant is in the shared package where it could plausibly end up on the wire.

apps/lfx-one/e2e/org-project-detail.spec.ts
ahmedomosanya: Addressed in `2aa68bc` (follow-up #1443) — went slightly stricter than the suggestion. Rather than a `\d{4}` substring test, each label is matched against the four documented formats:

```ts
expect(period).toMatch(/^(?:[A-Z][a-z]{2} \d{4}|Q[1-4] \d{4}|\d{4}(?:\u2013\d{4})?)$/);
```

So an empty label, a malformed one, and the `YYYY–?` degenerate case from the `multi_year` thread all fail the spec instead of slipping through on a digit match.

ahmedomosanya:
This merged before the review feedback landed. The findings above are handled in a follow-up: **#1443**.

**Landing in #1443** — the three findings with a real effect on the running system, plus two supporting changes:

- Lifetime trend series are now placed **by bucket index** over the zero-filled axis instead of pushed in row order, so a sparse row set can no longer shift a series against `periods[]` and silently mislabel every point on it.
- `hasAdaptivePeriods` no longer requires a non-empty array, so a project with an empty bucket spine stops rejecting its own cache entry and re-querying Snowflake on every request.
- A `multi_year` bucket with a missing end date no longer collapses to a bare `YYYY`.
- The all-time window sentinel is named, and the E2E spec asserts the bucket label format rather than only array shape.

**Not changing** — the client-side leading-bucket trims stay. @audigregorie is right that the trend chart and the influence cards trim on different criteria, but they render in different tab panels, values are correct in both, and within each block labels and data are sliced by the same index, so nothing is mislabelled. The empty leading buckets they hide come from the unbounded bucket-spine anchor, which is documented in [LFXV2-3231](https://linuxfoundation.atlassian.net/browse/LFXV2-3231) together with the related bucket-width problem that trimming does not solve.

[LFXV2-3231]: https://linuxfoundation.atlassian.net/browse/LFXV2-3231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
ahmedomosanya: Correcting my reply above: after weighing it, the trims are **staying**, so this is a deferral rather than a fix.

Your reading of the divergence is right, but the impact is narrower than I first treated it. Within each block the labels and the data are sliced by the same index, so no point is mislabelled — and the influence cards and the stacked trend render in different tab panels, so the two axes are never on screen together to be compared.

The dead space the trims hide comes from the unbounded bucket-spine anchor, which is now captured in [LFXV2-3231](https://linuxfoundation.atlassian.net/browse/LFXV2-3231) along with its second consequence: granularity is derived from that same inflated span, so bucket widths stay too coarse whether or not the leading buckets are trimmed. With the symptom documented there, reverting shipped behaviour purely to keep it visible in the UI isn't worth it.

Leaving this closed as deferred. If the cross-tab inconsistency does bite in practice, collapsing both blocks onto a single trim criterion is a small change.

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
ahmedomosanya: Correcting my reply above: the trims are staying (see the sibling thread), so `lifetimeTrimStart` is not being removed after all.

Leaving the diamond as-is, on your own assessment of it — Angular's computeds resolve it correctly and the extra pass is bounded at 14 cards × ≤12 buckets. If the card mapping grows past that, folding the trim into the mapping computed is the obvious next step.

PR #1306, opened 2026-08-04
test(enrollment): document and pin Purchased status derivation

fayazg:
## Summary

- Document that `deriveEnrollmentStatus` intentionally has no dedicated `'Purchased'` branch — only `'Expired'` short-circuits, while `'Active'` and `'Purchased'` are equivalent for display and both flow into the date-based derivation (EndDate + Stripe/auto-renew).
- Add tests pinning `'Purchased'` to the same outcomes as `'Active'` (active when >30 days out, expiring-soon within 30 days, expired when past) so the fall-through can't silently regress.

Follow-up to PR #732 review notes. No behavior change.

LFXV2-1664

### DIFF
--- a/packages/shared/src/utils/enrollment.utils.spec.ts
+++ b/packages/shared/src/utils/enrollment.utils.spec.ts
@@ -129,6 +129,25 @@ describe('deriveEnrollmentStatus', () => {
     const item = enrollment({ membership: membership({ endDate: daysFromNow(10) }) });
     expect(deriveEnrollmentStatus(item)).toBe('Expiring Soon');
   });
+
+  // 'Purchased' has no dedicated branch — it deliberately flows into the same date-based derivation
+  // as 'Active'. These pin that equivalence so the fall-through can't silently regress.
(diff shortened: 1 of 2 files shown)


luismoriguerra (dismissed):
Approving — tests + comments only, no behavior change.

- The comment matches the code: `'Expired'` is the only short-circuit; `'Active'`/`'Purchased'` both fall through to the date/Stripe derivation.
- The three new cases correctly pin that fall-through and reuse the existing builders, so no fixture drift.
- Nit (non-blocking): `'Purchased'` + stripe + autoRenew isn't covered, but that branch is already pinned for `'Active'`.

**Merge-order note:** this collides with #1334, which renames these same fields to camelCase and is *not* built on this branch. Merging this one first is the cheaper order — #1334 then just camelCases the three new tests. The reverse drags PascalCase fixtures back in on rebase. Needs a rebase on `main` either way (currently behind).

---

dealako (approved):

PR #1307, opened 2026-08-04
feat(groups): partial-coverage active members count with disclosure

manishdixitlfx:
## Summary

- `active_members` on the groups dashboard now counts distinct active members over the *covered* subset of the caller's visible committees only, instead of degrading to `null` whenever any single visible committee is unresolved or not-yet-synced. Live validation against prod showed the old all-or-nothing rule returned `null` for essentially every real caller.
- The response gains `coverage: { covered, total }` (visible committees covered vs. total visible). `active_members` is `null` only when zero committees are covered, or the computation itself failed.
- UI: the Active Members card's sub-line gains "· across N of M groups" when coverage is partial, appended to the existing freshness/mock label; full coverage renders no qualifier. The existing "Unavailable" state for a null count is unchanged.
- Mock mode fabricates full coverage (`covered === total`) so the mock path stays exercisable.

## Cache-validator shape-versioning note

`isGroupsEngagementStats` (the Valkey read-through cache's `accept` validator) now also checks for the `coverage` field, so a cache entry written by a pre-upgrade instance during a rolling deploy reads as a miss and recomputes, rather than being served to a client expecting the new shape for up to `GROUPS_ENGAGEMENT_TTL_SECONDS`. The validator's doc comment generalizes this: it's not just a cross-backend (`mock`/`live`) check anymore, it's this response shape's version check — any field a newer service version always sets belongs here.

The same rolling-deploy shape drift is guarded client-side too: `resolveActiveMembersSubLine` optional-chains `stats?.coverage` rather than dereferencing it directly, so a response from a pre-upgrade server (missing `coverage` entirely) degrades to the plain sub-line instead of throwing inside the `computed()` that feeds the whole My Groups stat-card row.

## Branch history note

This branch was originally stacked on `feat/LFXV2-1705-live-read-path` (#1293). That PR merged (squash) while this one was in review, so the branch was rebased/cherry-picked onto current `main` and force-pushed to retarget this PR directly at `main` — the diff below is only this PR's own 4 commits (+293/-109 across 6 files), not the merged parent's.

## Trade-offs (documented per PR-readiness SHOULD_FIX)

- Branch name (`feat/LFXV2-2978-stats-partial-coverage`) carries a descriptive suffix beyond `type/LFXV2-NNN`.
- The `feat(groups): ...(LFXV2-2978)` commit header is 80 chars — over the 72-char team style but within commitlint's 100-char hard limit.

## Test plan

- [x] `yarn lint && yarn format && yarn test && yarn build` all pass on the rebased branch
- [x] New/updated unit tests: partial resolution (2 of 3 committees covered → count over 2 + `coverage: {2,3}`), unresolved uid no longer short-circuits, zero coverage → null + `{0,N}`, full coverage → no UI qualifier, cache validator rejects entries missing `coverage`, mock mode fabricates full coverage, client-side guard against a `coverage`-less response
- [x] e2e fixture (`groups-engagement-stats.spec.ts`) updated with the new `coverage` field
- [x] Post-commit reviewer trio (general + Self Serve conventions + learnings) run and addressed after every commit; full-branch sweep run and clean with respect to this branch's own diff

### DIFF
--- a/apps/lfx-one/e2e/groups-engagement-stats.spec.ts
+++ b/apps/lfx-one/e2e/groups-engagement-stats.spec.ts
@@ -147,7 +147,13 @@ async function stubEngagementStats(page: Page): Promise<void> {
     route.fulfill({
       status: 200,
       contentType: 'application/json',
-      body: JSON.stringify({ active_members: 7, meetings_this_month: 2, computed_at: new Date().toISOString(), source: 'live' }),
+      body: JSON.stringify({
+        active_members: 7,
+        meetings_this_month: 2,
+        computed_at: new Date().toISOString(),
+        source: 'live',
+        coverage: { covered: 1, total: 1 },
+      }),
     })
   );
 }
--- a/apps/lfx-one/src/server/services/groups-engagement-stats.service.spec.ts
+++ b/apps/lfx-one/src/server/services/groups-engagement-stats.service.spec.ts
@@ -567,6 +567,9 @@ describe('GroupsEngagementStatsService', () => {
       const result = await service.getEngagementStats(buildReq());
 
       expect(result.active_members).toBeNull();
+      // total reflects the visible-committee count established before the chunk failure (not 0) —
+      // the caller's committee set was known even though the count itself couldn't be trusted.
+      expect(result.coverage).toEqual({ covered: 0, total: 150 });
     });
 
     it('degrades to active_members: null (not 0) when the Snowflake query hits a missing-object error', async () => {
@@ -578,6 +581,7 @@ describe('GroupsEngagementStatsService', () => {
       expect(result.active_members).toBeNull();
       expect(result.meetings_this_month).toBeNull();
       expect(result.source).toBe('live');
+      expect(result.coverage).toEqual({ covered: 0, total: 1 });
     });
 
     it('degrades to active_members: null without throwing when fetching the visible committee set fails', async () => {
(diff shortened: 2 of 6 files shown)


dealako (approved):
Hi @manishdixitlfx — reviewed PR #1307.

**Overall impression:** Clean, well-scoped change. It replaces the all-or-nothing `active_members: null` fallback with partial-coverage reporting — counting active members across the committees that both resolve v2→v1 and have warehouse rows, surfacing a `coverage: { covered, total }` field, and disclosing partial coverage in the UI sub-line ("· across N of M groups"). Cache versioning correctly treats a missing `coverage` field as stale so pre-upgrade Valkey entries recompute during rolling deploys, and the error path preserves the visible-committee count for accurate diagnostics. Test coverage is thorough — partial coverage, zero coverage, cache hit/miss on shape drift, mock mode, and null-handling are all exercised at both the unit and e2e level.

Ran this through three parallel dimension reviews (Security & Data Privacy / Correctness, Performance & Test Coverage / Code Style, Consistency & Documentation) plus reconciliation against CodeRabbit, Cursor Bugbot, and Copilot — no new findings from any source. Copilot explicitly reviewed all 6 changed files with no comments; Cursor's summary confirms the same semantics with no risk flags beyond a self-noted "Medium Risk" label for the live rollup/cache-shape change, which is appropriately covered by the shape-versioning safeguard and tests.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 0

Note: the branch currently shows as behind `main` — worth a rebase before merge, but not a review blocker.

PR #1308, opened 2026-08-04
feat(dashboards): split web sessions into a dedicated overview card

mrautela365:
## Summary

Splits website sessions out of the ED marketing overview's **Social** card into a dedicated **Web** card.

Social previously carried two dual-signals — social followers and website sessions (30d). Those measure different things: followers are a stock, sessions are a flow. Separating them gives each its own card and, for web, its own drill-down.

This also reconnects an orphaned drawer. `website-visits-drawer` and `DashboardDrawerType.MarketingWebsiteVisits` already existed and were already bound in `marketing-overview.component.html`, but no card routed to them after #1300. The Web card sets `drawerType: MarketingWebsiteVisits`, so the existing drawer becomes reachable again — no new component required.

Card order is now: Events · Members · Adoption · **Social** · **Web** · Email · Paid Media · Attribution · Sentiment · Flywheel.

## Changes

- `dashboard-metrics.constants.ts` — Social reduced to a single-value followers card; new Web card for rolling 30-day sessions.
- `marketing-dashboard.spec.ts` — added a Web card render test and repaired stale testIds (below).

## E2E repairs

Three `marketing-card-*` testIds referenced by the spec have been absent from source since the #1300 restructure, so those suites could not pass:

| Reference | Action |
|---|---|
| `marketing-card-website-visits` (4×) | Repointed to `ed-evo-web-sessions`; duplicate render test dropped since the new Web card test covers it |
| `marketing-card-email-ctr` (4×) | Repointed to `ed-evo-campaign-performance` |
| `marketing-card-social-media` (render) | Repointed to `ed-evo-brand-reach` |
| `marketing-card-social-media` (drawer, 3×) | **Suite skipped** — see below |

The **Social Media Drawer** suite is skipped rather than repointed. No card sets `drawerType: MarketingSocialMedia`, so that drawer is rendered by the template but unreachable from the UI; the Social card opens `BrandReach`, not the social-media drawer. No rename makes those tests pass. The skip carries a comment naming the precondition to re-enable. **This is pre-existing from #1300, not introduced here** — worth a follow-up ticket to either wire it up or remove the drawer.

## Known trade-offs

- **Sparkline window differs from the value.** The Web value is a rolling 30-day total from `WEB_ACTIVITIES_SUMMARY`; the sparkline is a weekly series over six months from `WEB_ACTIVITIES_BY_PROJECT`. The subtitle labels both windows explicitly (`Sessions (30d) · Trend: last 6 months`) so the 30-day figure isn't read as a six-month number. Per an existing code comment, `WEB_ACTIVITIES_BY_PROJECT` lacks `LF_SUB_DOMAIN_CLASSIFICATION`, so the trend is all-program totals.
- **Social sparkline is a placeholder.** No historical follower series is available, so it renders `flatSparklineData` at the current total — deliberately not reusing website sessions, which is a different metric.
- **Brand Reach drawer still shows "Website Sessions by Domain."** Left in place intentionally: `brand-reach-drawer.component.ts` uses `websiteDomains.length` in its empty-state guard and derives a top-domain recommendation from it, and the website-visits drawer uses a different response type. Moving it is a separate change.

Card and drawer totals agree — both sum `TOTAL_SESSIONS_LAST_30_DAYS` from the same `WEB_ACTIVITIES_SUMMARY` table.

## Testing

- `yarn lint:check` clean (one pre-existing warning in `campaigns.component.ts`, untouched here)
- `yarn check-types` passes
- `yarn build` passes
- Verified in browser at `/foundation/marketing-overview?project=tlf`
- Every active testId in the spec cross-checked against source

Ref: LFXV2-2023

### DIFF
--- a/apps/lfx-one/e2e/marketing-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/marketing-dashboard.spec.ts
@@ -150,20 +150,12 @@ test.describe('Marketing Metric Cards', () => {
     await switchToExecutiveDirector(page);
   });
 
-  test('renders Website Visits card', async ({ page }) => {
-    const card = page.locator('[data-testid="marketing-card-website-visits"]');
+  test('renders Email card', async ({ page }) => {
+    const card = page.locator('[data-testid="ed-evo-campaign-performance"]');
     await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
     await card.scrollIntoViewIfNeeded();
     await expect(card).toBeVisible();
-    await expect(card).toContainText('Website Visits');
-  });
-
-  test('renders Email CTR card', async ({ page }) => {
-    const card = page.locator('[data-testid="marketing-card-email-ctr"]');
-    await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
-    await card.scrollIntoViewIfNeeded();
-    await expect(card).toBeVisible();
-    await expect(card).toContainText('Email CTR');
+    await expect(card).toContainText('Email');
   });
 
   test('renders Paid Media card', async ({ page }) => {
@@ -174,12 +166,12 @@ test.describe('Marketing Metric Cards', () => {
     await expect(card).toContainText('Paid Media');
   });
 
-  test('renders Social Media card', async ({ page }) => {
-    const card = page.locator('[data-testid="marketing-card-social-media"]');
+  test('renders Social card', async ({ page }) => {
+    const card = page.locator('[data-testid="ed-evo-brand-reach"]');
     await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
     await card.scrollIntoViewIfNeeded();
     await expect(card).toBeVisible();
-    await expect(card).toContainText('Social Media');
+    await expect(card).toContainText('Social');
   });
 
   test('renders Web card', async ({ page }) => {
@@ -208,7 +200,7 @@ test.describe('Website Visits Drawer', () => {
   });
 
   test('shows stats section with data', async ({ page }) => {
-    await openDrawerAndWaitForData(page, 'marketing-card-website-visits', 'website-visits-drawer-content', 'website-visits-drawer-stats');
+    await openDrawerAndWaitForData(page, 'ed-evo-web-sessions', 'website-visits-drawer-content', 'website-visits-drawer-stats');
     const statCards = page.locator('[data-testid="website-visits-drawer-stats"]').locator('lfx-card');
     await expect(statCards).toHaveCount(2);
   });
@@ -232,17 +224,17 @@ test.describe('Email CTR Drawer', () => {
   });
 
   test('opens and shows title when card is clicked', async ({ page }) => {
-    await openDrawer(page, 'marketing-card-email-ctr', 'email-ctr-drawer-content');
+    await openDrawer(page, 'ed-evo-campaign-performance', 'email-ctr-drawer-content');
     await expect(page.locator('[data-testid="email-ctr-drawer-title"]')).toContainText('Email');
   });
 
   test('shows stats and email sections', async ({ page }) => {
-    await openDrawerAndWaitForData(page, 'marketing-card-email-ctr', 'email-ctr-drawer-content', 'email-ctr-drawer-stats');
+    await openDrawerAndWaitForData(page, 'ed-evo-campaign-performance', 'email-ctr-drawer-content', 'email-ctr-drawer-stats');
     await expect(page.locator('[data-testid="email-ctr-drawer-email-section"]')).toBeVisible();
   });
 
   test('closes when close button is clicked', async ({ page }) => {
-    await openDrawer(page, 'marketing-card-email-ctr', 'email-ctr-drawer-content');
+    await openDrawer(page, 'ed-evo-campaign-performance', 'email-ctr-drawer-content');
     await page.locator('[data-testid="email-ctr-drawer-close"]').click();
     await expect(page.locator('[data-testid="email-ctr-drawer-content"]')).not.toBeVisible();
   });
@@ -272,7 +264,12 @@ test.describe('Paid Social Reach Drawer', () => {
   });
 });
 
-test.describe('Social Media Drawer', () => {
+// The social-media drawer is still rendered by marketing-overview, but no card sets
+// drawerType: MarketingSocialMedia, so there is no way to open it from the UI. These
+// tests targeted 'marketing-card-social-media', a card that no longer exists. Skipped
+// rather than repointed: the Social card opens BrandReach, not this drawer, so no
+// rename makes them pass. Re-enable once a card routes to MarketingSocialMedia.
+test.describe.skip('Social Media Drawer', () => {
   test.beforeEach(async ({ page }) => {
     await page.goto(DASHBOARD_URL);
     await switchToExecutiveDirector(page);
--- a/apps/lfx-one/e2e/marketing-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/marketing-dashboard.spec.ts
@@ -180,6 +180,11 @@ test.describe('Marketing Metric Cards', () => {
     await card.scrollIntoViewIfNeeded();
     await expect(card).toBeVisible();
     await expect(card).toContainText('Web');
+    // Assert the split itself, not just the card's presence: sessions must live on
+    // Web and no longer on Social. A title-only check would still pass if sessions
+    // had stayed on the Social card and Web were an empty duplicate.
+    await expect(card).toContainText('Sessions (30d)');
+    await expect(page.locator('[data-testid="ed-evo-brand-reach"]')).not.toContainText('Sessions (30d)');
   });
 
   test('renders filter pills', async ({ page }) => {
--- a/apps/lfx-one/e2e/marketing-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/marketing-dashboard.spec.ts
@@ -182,6 +182,14 @@ test.describe('Marketing Metric Cards', () => {
     await expect(card).toContainText('Social Media');
   });
 
+  test('renders Web card', async ({ page }) => {
+    const card = page.locator('[data-testid="ed-evo-web-sessions"]');
+    await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
+    await card.scrollIntoViewIfNeeded();
+    await expect(card).toBeVisible();
+    await expect(card).toContainText('Web');
+  });
+
   test('renders filter pills', async ({ page }) => {
     const filters = page.locator('[data-testid="ed-evolution-filters"]');
     await expect(filters).toBeVisible();
@@ -195,7 +203,7 @@ test.describe('Website Visits Drawer', () => {
   });
 
   test('opens and shows title when card is clicked', async ({ page }) => {
-    await openDrawer(page, 'marketing-card-website-visits', 'website-visits-drawer-content');
+    await openDrawer(page, 'ed-evo-web-sessions', 'website-visits-drawer-content');
     await expect(page.locator('[data-testid="website-visits-drawer-title"]')).toContainText('Website Visits');
   });
 
@@ -206,12 +214,12 @@ test.describe('Website Visits Drawer', () => {
   });
 
   test('shows trend chart section', async ({ page }) => {
-    await openDrawer(page, 'marketing-card-website-visits', 'website-visits-drawer-content');
+    await openDrawer(page, 'ed-evo-web-sessions', 'website-visits-drawer-content');
     await expect(page.locator('[data-testid="website-visits-drawer-trend-section"]')).toBeVisible();
   });
 
   test('closes when close button is clicked', async ({ page }) => {
-    await openDrawer(page, 'marketing-card-website-visits', 'website-visits-drawer-content');
+    await openDrawer(page, 'ed-evo-web-sessions', 'website-visits-drawer-content');
     await page.locator('[data-testid="website-visits-drawer-close"]').click();
     await expect(page.locator('[data-testid="website-visits-drawer-content"]')).not.toBeVisible();
   });
--- a/packages/shared/src/constants/dashboard-metrics.constants.ts
+++ b/packages/shared/src/constants/dashboard-metrics.constants.ts
@@ -768,7 +768,7 @@ function seriesTrendDirection(series: number[], activity: number[] = series): 'u
 /**
  * Build ED Evolution dashboard cards from live API data.
  * 6 North Star (Events, Members, Adoption, Email, Paid Media, Attribution)
- * + 2 Brand (Social, Sentiment) + Flywheel.
+ * + 3 Brand (Social, Web, Sentiment) + Flywheel.
  * Member Retention is merged into the Members drawer.
  *
  * Sparkline color semantics:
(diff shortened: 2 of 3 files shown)


packages/shared/src/constants/dashboard-metrics.constants.ts
```diff
@@ -875,37 +875,49 @@ export function buildEdEvolutionMetrics(data: EdEvolutionData): DashboardMetricC
...
+      // read as a six-month number. weeklyTrend only holds weeks WITH rows, so its
+      // length is not the reporting window and the window is labeled directly.
+      subtitle: brandReach.weeklyTrend.length > 0 ? 'Sessions (30d) · Trend: last 6 months' : 'Sessions (30d)',
+      chartData: protoSparkline(
+        brandReach.weeklyTrend.length > 0 ? brandReach.weeklyTrend.map((d) => d.sessions) : flatSparklineData(brandReach.totalMonthlySessions),
+        lfxColors.violet[500]
+      ),
+      chartOptions: NO_TOOLTIP_CHART_OPTIONS,
+      tooltipText: 'Rolling 30-day website sessions across foundation web properties, with month-over-month change.',
+      drawerType: DashboardDrawerType.MarketingWebsiteVisits,
```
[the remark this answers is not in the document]
mrautela365: Confirmed and fixed in 7b9f45b75. Verified both sides: the overview maps a missing selection to `tlf` (`marketing-overview.component.ts:310`), while the drawer mapped it to `""` and then dropped it via `filter(([isVisible, slug]) => isVisible && !!slug)` — so in the ED default all-foundations state the drawer would never issue its request.

Fixed by aligning the drawer with the overview rather than by loosening the filter. A repo-wide check shows every sibling ED drawer (email-ctr, revenue-impact, paid-social-reach) already falls back to `tlf`; website-visits was the lone outlier, so this restores the existing convention. Added a comment noting the coupling so the two do not drift again.

You are right that this is a latent issue rather than one this PR introduces — but this PR is what makes the drawer reachable from a populated card, so it belongs here.

mrautela365:
All bot review threads resolved and CI is green. Ready for review.

**Fixed since the last round:** the Web card drawer mismatch — the overview loads the card with a `tlf` fallback for the ED all-foundations default, but `WebsiteVisitsDrawerComponent` mapped a missing selection to `''` and filtered it out, so a populated card could open a drawer that never issued its request. Every sibling ED drawer already falls back to `tlf`; this one was the outlier (7b9f45b75).

**Merge order:** this PR is first in a 3-PR stack — #1308 → #1309 → #1310. All three target `main` except #1310, which is stacked on a combined base and will be retargeted at `main` once these two land.

packages/shared/src/constants/dashboard-metrics.constants.ts
```diff
@@ -875,37 +875,49 @@ export function buildEdEvolutionMetrics(data: EdEvolutionData): DashboardMetricC
...
+      value: formatNumber(brandReach.totalMonthlySessions),
+      changePercentage: formatMomChange(brandReach.sessionMomChangePct),
+      trend: normalizeTrend(brandReach.sessionMomChangePct, brandReach.sessionMomChangePct >= 0 ? 'up' : 'down'),
+      // The value is a rolling 30-day total; the sparkline is a separate weekly series
+      // over a fixed six-month range. Label them separately so the 30-day figure is not
+      // read as a six-month number. weeklyTrend only holds weeks WITH rows, so its
+      // length is not the reporting window and the window is labeled directly.
+      subtitle: brandReach.weeklyTrend.length > 0 ? 'Sessions (30d) · Trend: last 6 months' : 'Sessions (30d)',
+      chartData: protoSparkline(
+        brandReach.weeklyTrend.length > 0 ? brandReach.weeklyTrend.map((d) => d.sessions) : flatSparklineData(brandReach.totalMonthlySessions),
```
dealako, line 917: [nit] Web card's `flatSparklineData` fallback isn't documented the way Social's is

Issue: The Social card's comment (lines 883–886) explains why `flatSparklineData` is used as a fallback and that it's a placeholder, not a trend. The Web card uses the exact same fallback pattern here but the comment above it (907–910) only explains the value/sparkline window mismatch, not the fallback itself.
Why it matters: Minor — a reader who understands the Social card's comment still has to go re-derive why `flatSparklineData` shows up here too.
Fix: One line addition, e.g. "When weeklyTrend is empty, flatSparklineData renders a placeholder (see Social card above), not a trend."

dealako (dismissed):
@mrautela365 nice, clean split. The change is well-scoped: `Social` becomes a single-metric followers card, `Web` gets its own card + drill-down, and the orphaned `MarketingWebsiteVisits` drawer is reconnected. The value/sparkline window mismatch is called out explicitly in both the code comment and the subtitle, so it won't read as misleading. The e2e repairs are a legitimate independent fix (stale `marketing-card-*` testIds dating to #1300), and the skipped Social Media Drawer suite has a clear inline reason with a stated precondition to re-enable.

Three independent dimension reviews ran (security & data privacy; correctness/performance/tests; style & documentation) — no file-level analysis happened in this pass beyond verifying the one finding below.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — Web card's sparkline fallback (`flatSparklineData`, line 913) isn't documented the way Social's identical fallback is (line 883–886).
- ❔ Question: 0

**Bot reconciliation:**
- Cursor Bugbot: informational summary only, no action items.
- CodeRabbit: auto-review disabled on this repo, skipped.
- Copilot (suppressed comment, commit 7b9f45b): flagged that `flatSparklineData` creates a slight synthetic upward slope regardless of real trend direction, and suggested leaving the chart empty until real history exists. I agree this is a real property of the helper, but disagree it should block this PR — the function and this exact fallback pattern already ship on four other cards (events, members, engaged community, flywheel) prior to this change, and the PR's own code comment and description already disclose it transparently as a placeholder rather than a trend. Worth a follow-up if the team wants to revisit the shared helper itself, but out of scope here.

**Revision history:** this PR already absorbed three rounds of self-driven review fixes prior to this pass — subtitle window clarity, the "flat line" → "placeholder" comment wording, stale e2e testId repairs, and the drawer foundation fallback (`''` → `'tlf'`) to match the overview's default. All of those read as correctly resolved in the current diff; no regressions found.

**Decision:** ✅ **Approved with minor comments** (one nit, non-blocking).

packages/shared/src/constants/dashboard-metrics.constants.ts
mrautela365: Done in 6080860 — added the cross-reference line:

> When weeklyTrend is empty, flatSparklineData renders a placeholder at the current total (see the Social card above), not a trend.

mrautela365:
@dealako heads up — your approval on this PR was dismissed from my side at 05:05Z while I was pushing the nit fix. That was my error, not a stale-review rule (the sibling PR #1310 took a push after approval with no dismissal). Apologies for the churn.

Nothing about the diff changed except the one nit you raised: 6080860 adds the cross-reference line documenting the web card sparkline fallback as a placeholder, matching how the Social card documents the identical fallback. No behavior change.

Re-requesting review so the approval can be restored.

MRashad26 (dismissed):
Code enforcer: clean. Web card split from Social is well-structured — `flatSparklineData` placeholder documented, dual-window subtitle (`Sessions (30d) · Trend: last 6 months`) avoids the 30d/6m ambiguity, and `tlf` fallback in `website-visits-drawer` correctly mirrors the overview default. E2E testId repoints and Social Media Drawer skip with a re-enable comment are all correct.

packages/shared/src/constants/dashboard-metrics.constants.ts
dealako, line 1: [minor] Function header and order comment weren't updated for the new Web card

Issue: `buildEdEvolutionMetrics`'s doc comment (line 771) still says "6 North Star ... + 2 Brand (Social, Sentiment) + Flywheel", and the order comment (line 803) says "then Social, then Campaign Performance" with no mention of Web in between.
Proof: This PR inserts a third Brand card (Web, lines 893–919) between Social and Campaign Performance, so the count is now 3 Brand cards and the order comment skips a card. (Both lines are outside this PR's diff hunks, hence a file-level comment rather than inline.)
Why it matters: This header is called out in the same function as "the single source of truth for sequence" — a future reorder or card-count check would trust an inventory that's already wrong as of this PR.
Fix: Update line 771 to "+ 3 Brand (Social, Web, Sentiment)" and line 803 to mention Web between Social and Campaign Performance.

(Confirming and posting on behalf of a Copilot suppressed comment flagging the same lines — verified accurate against current source.)

apps/lfx-one/e2e/marketing-dashboard.spec.ts
```diff
@@ -150,36 +150,36 @@ test.describe('Marketing Metric Cards', () => {
...
-  test('renders Social Media card', async ({ page }) => {
-    const card = page.locator('[data-testid="marketing-card-social-media"]');
+  test('renders Web card', async ({ page }) => {
+    const card = page.locator('[data-testid="ed-evo-web-sessions"]');
     await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
     await card.scrollIntoViewIfNeeded();
     await expect(card).toBeVisible();
-    await expect(card).toContainText('Social Media');
+    await expect(card).toContainText('Web');
   });
```
dealako, line 188: [nit] "renders Web card" doesn't assert the split itself

Issue: This test only checks that a card titled "Web" is attached/visible — it would still pass if website sessions had stayed on the Social card and "Web" were just an empty duplicate title.
Why it matters: Low — Copilot raised this as a suppressed comment and it's technically correct, but every sibling render test in this file (`renders Paid Media card`, `renders Social card`, etc.) follows this exact same shallow pattern, so it's consistent with the file's existing convention rather than a gap specific to this PR.
Fix: Optional — if you want a stronger regression guard here, assert the `Sessions (30d)` subtitle on the Web card and that it's absent from Social. Not blocking given the established pattern.

dealako (dismissed):
@mrautela365 no worries on the dismissal — appreciated the heads-up, and confirmed independently that nothing else moved: `git diff` between the two review rounds is exactly the two-line comment in commit 6080860.

👏 **Nice work:** the new comment correctly explains the `flatSparklineData` fallback and cross-references the Social card's identical pattern instead of duplicating the explanation — clean, minimal fix for exactly the nit raised.

**Revision tracking:**
- ✅ **Resolved** — Web card's `flatSparklineData` fallback now documented (commit 6080860), matching the Social card's explanation.

**New from this round:** re-ran the three-dimension pass scoped to the new commit (clean — it's comment-only) and reconciled against Copilot's newest suppressed comments (posted 05:08, after the doc fix landed):
- 🟡 **Minor** (new, agree with Copilot) — `buildEdEvolutionMetrics`'s function-header doc (line 771: "2 Brand (Social, Sentiment)") and order comment (line 803) weren't updated when Web was inserted as a third Brand card between Social and Campaign Performance. Real inaccuracy in the array the code itself calls "the single source of truth for sequence" — posted as a file-level comment since both lines sit outside this PR's diff hunks.
- ⚪ **Nit** (new, partially agree with Copilot) — the "renders Web card" e2e test only checks title/visibility, so it wouldn't catch sessions staying on Social. True, but it matches every sibling render test's existing shallow-assertion pattern in this file, so not a gap specific to this PR — optional if you want to strengthen it.

**Issue count (open):**
- 🔴 Blocking: 0
- 🟡 Minor: 1 — stale Brand-card count/order comment
- ⚪ Nit: 1 — Web card render test assertion depth (pattern-consistent, optional)

**Decision:** ✅ **Approved with minor comments.**

packages/shared/src/constants/dashboard-metrics.constants.ts
mrautela365: Fixed in 6119094. Header now reads "+ 3 Brand (Social, Web, Sentiment)".

While correcting the order comment I found it was already wrong in a second way independent of this PR: it said "then Campaign Performance", but no card is titled that — the cards after Web are Email, Paid Media, and Attribution. Rather than insert Web into a sequence that was itself inaccurate, I replaced it with the literal array order and noted that the display order interleaves categories (Social/Web sit between North Star and Email, Sentiment trails them), so it is not a category-grouped list.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
mrautela365: Fixed in 6119094 rather than deferred — the standing convention here is to address nits too, and this one guards precisely what the PR does.

`renders Web card` now asserts the Web card contains `Sessions (30d)` and that the Social card does not. That fails in exactly the scenario you described: sessions staying on Social with Web as an empty duplicate. Left the sibling render tests alone so this does not turn into an unrelated sweep.

mrautela365:
Closing the loop on Copilot's suppressed comment from the 01:07 pass, since it has no thread attached and so cannot be resolved through the normal path.

**The finding:** `flatSparklineData` returns `[v-nudge, v, v, v, v, v+nudge]`, so it starts below the current total and ends above it — Copilot reads that as a synthetic upward trend that can contradict the card's own trend indicator when the real MoM change is flat or negative.

**The shape claim is literally correct** — the series does begin low and end high. I checked the two severity claims and neither holds:

1. **It cannot contradict the trend indicator, because it does not drive it.** `trend` comes from `normalizeTrend(brandReach.changePercentage, brandReach.trend)` and `changePercentage` from `formatMomChange(...)` — both read the real MoM value. `chartData` is a separate field. A down arrow with a faintly up-ticking placeholder line is visually odd, but the arrow itself stays truthful.
2. **The nudge is ±2% of the value, floored at 0.1.** On ~50k followers that is ±1k across the full sparkline width — near-flat by construction, which is what the helper's docstring says it is for: a constant array makes `min === max`, collapsing Chart.js's Y range and hiding the line entirely. The variation exists to make the line render at all.

**Why I am not changing it here.** `flatSparklineData` is not new in this PR — it appears 5 times on `origin/main` today and backs Events, Members, Engaged Community, and Flywheel in exactly the same no-history situation. Rewriting it would change six cards from inside a PR whose scope is splitting Web out of Social, which is precisely the kind of unrelated bundling the 1000-line/one-concern guidance exists to prevent. Both @dealako and @MRashad26 reviewed this fallback explicitly and signed off on it as a documented placeholder.

**The underlying suggestion is still worth doing** — "render nothing until real history exists" is a better contract than "render a placeholder." That is a change to the shared helper and its six call sites, plus a decision about what an empty sparkline should look like. I would rather do it deliberately in its own PR than smuggle it in here. Happy to open the ticket if you agree.

Copilot's other suppressed comment from the 05:08 pass — the stale function documentation and the order comment skipping from Social to Campaign Performance — was fixed in 6119094.

mrautela365:
Branch updated to current `main` (174f3b5) — no review-relevant change.

`main` was 2 commits ahead: `0019bd6` and `e969fe1`, both touching only `apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.html`. No overlap with this PR, clean merge, no conflicts.

**Reviewable diff is unchanged: 3 files, +83/−53** — same three files as before the update. Single merge base preserved, so the rendered diff has not absorbed anything from a sibling branch.

Verified after the merge: `yarn check-types` and `yarn lint:check` both pass.

**Note on the timing** — this update dismissed nothing. The earlier approvals from @dealako and @MRashad26 had already been invalidated by previous pushes (`main` dismisses stale reviews on push), so there was no live approval left to lose. Both reviewers are already re-requested.

This PR is now blocked purely on a CODEOWNERS approval — root `CODEOWNERS` is `* @linuxfoundation/lfx-platform`, so every PR here needs one team approval. All checks pass, no unresolved threads, nothing outstanding from my side.

dealako (approved):
@mrautela365 caught up on the new commit (6119094) plus your comments — the `main` merge (174f3b5) is confirmed clean and outside this PR's reviewable diff (verified `gh pr diff --name-only` still shows only the same 3 files).

👏 **Nice work:** both open items from the last round are properly closed out — the doc header/order comment now correctly says "3 Brand (Social, Web, Sentiment)" and spells out the full 10-card interleaved sequence instead of implying a category-grouped list, and the Web card test now asserts the actual split (`Sessions (30d)` present on Web, absent from Social) instead of just title/visibility.

**Revision tracking:**
- ✅ **Resolved** — stale card-count/order doc comment (6119094).
- ✅ **Resolved** — Web card e2e test now asserts the split, not just card presence (6119094).

**Re-ran the three-dimension pass** scoped to 6119094 (doc comments + test assertions only) — clean across security/privacy, correctness/performance/tests, and style/docs. No new findings.

**Bot reconciliation:** Copilot's repeated suppressed comment on `flatSparklineData`'s synthetic upward slope (14:37 pass) is the same item from round 1 — I'll stand by my original position (agreed on the mechanism, disagreed it should block: pre-existing helper backing 4+ other cards, already disclosed as a placeholder). Your 13:57 writeup traces it further and correctly notes it doesn't feed the trend arrow (`normalizeTrend`/`formatMomChange` use the real MoM value, not the chart series) — that's a stronger version of the same conclusion. Agreed a follow-up ticket for the shared helper is the right way to handle it, not this PR.

**Issue count (open):**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

**Decision:** ✅ **Approved.**

PR #1310, opened 2026-08-05
feat(dashboards): add education card and drawer to marketing overview

mrautela365:
## Summary

Adds an **Education** card to the ED Marketing Overview carousel in the second slot, directly after Events, plus the drawer behind it.

With this stacked on #1308 and #1309, the carousel sequence matches the agreed design:

> Events · **Education** · Members · Adoption · Social · Web · Email · Paid · Attribution · Sentiment · Flywheel

## Stacking

This PR is stacked. Its base branch `feat/LFXV2-2023-web-and-social-base` contains #1308 and #1309, which are sibling PRs off #1300 — neither one contains the other, so no existing branch could serve as a base that includes both.

**Merge order:** #1308 → #1309 → this PR. Once both land in `main`, I'll retarget this PR at `main`; the base branch is scaffolding and should not be merged on its own.

The one non-trivial integration point: #1308 *skipped* the social-media drawer e2e tests (the drawer was unreachable from the UI), while #1309 *deletes* that drawer outright. Resolved in favour of deletion — the skipped tests target a component that no longer exists.

## Data source

Reuses the existing Health Metrics training-certification endpoint rather than adding a new one, so the Education card and the Health Metrics card cannot disagree on enrollment or revenue figures.

`ANALYTICS.PLATINUM.ENROLLMENTS` is a pre-aggregated wide table with the reporting range baked into column names and **no date column**, so no monthly grain exists. The sparkline is deliberately flat and the subtitle names the range instead of implying a trend.

**Net revenue excludes edX**, which contributes enrollments but carries no revenue column upstream. In the drawer's format breakdown, edX revenue renders as "not tracked" (`null`) rather than `$0`, so "not measured" is never displayed as "earned nothing".

## Failure vs. zero

`getTrainingCertificationSummary()` converts HTTP errors into an all-zeros response — a contract the Health Metrics card depends on, so it is deliberately left unchanged. That would otherwise make a failed request indistinguishable from a foundation with no training, and both would silently hide the card.

Instead the call site maps that error shape (identified by an empty `projectId`, which the fallback hardcodes and a real row always populates) back to `undefined`. A failed request and a genuinely-zero foundation both suppress the card, but only real data can render it.

## Testing

- `yarn build` passes
- Verified in the browser at `/foundation/overview?project=tlf` against TLF: 100,430 enrollments, $361,936.90 net revenue
- E2E coverage added for the `ed-evo-education` card and the `education-drawer` open / stats / formats / close paths

LFXV2-2023

### DIFF
--- a/apps/lfx-one/e2e/marketing-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/marketing-dashboard.spec.ts
@@ -221,6 +221,47 @@ test.describe('Education Drawer', () => {
     await expect(page.locator('[data-testid="education-drawer-formats"]')).toBeVisible();
   });
 
+  // Guards the null-vs-zero distinction: edX revenue is hardcoded null (COURSE_PURCHASES
+  // carries no revenue column) and must render "not tracked", while every other format
+  // has measured revenue and renders a currency figure — $0 included. A refactor that
+  // coalesced null to 0 would silently report edX as having earned nothing.
+  test('distinguishes untracked revenue from a measured zero', async ({ page }) => {
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-formats');
+
+    const edxRow = page.locator('[data-testid="education-drawer-format-row"]', { hasText: 'edX' });
+    await expect(edxRow).toContainText('not tracked');
+    await expect(edxRow).not.toContainText('$');
+
+    // Instructor Led revenue is measured, so it must show a currency value rather than
+    // the untracked placeholder — $0 is a legitimate reading here.
+    const instructorLedRow = page.locator('[data-testid="education-drawer-format-row"]', { hasText: 'Instructor Led' });
+    await expect(instructorLedRow).toContainText('$');
+    await expect(instructorLedRow).not.toContainText('not tracked');
+  });
+
+  // The attention/performing sections are driven by computed thresholds (concentration
+  // risk >70%, cert attach rate, balanced-portfolio) and routed by splitByPriority.
+  // TLF's live mix decides which of the two renders, so assert the invariant that holds
+  // either way: at least one section renders, and whichever renders carries content.
+  test('renders a priority-routed insight section', async ({ page }) => {
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-formats');
+
+    const attention = page.locator('[data-testid="education-drawer-attention"]');
+    const performing = page.locator('[data-testid="education-drawer-performing"]');
+    const facts = page.locator('[data-testid="education-drawer-facts"]');
+
+    const sectionCount = (await attention.count()) + (await performing.count()) + (await facts.count());
+    expect(sectionCount).toBeGreaterThan(0);
+
+    // Whichever sections rendered must be non-empty — an empty shell means the
+    // computed logic produced a section with nothing routed into it.
+    for (const section of [attention, performing, facts]) {
+      if ((await section.count()) > 0) {
+        await expect(section.first()).not.toBeEmpty();
+      }
+    }
+  });
+
   test('closes when close button is clicked', async ({ page }) => {
     await openDrawer(page, 'ed-evo-education', 'education-drawer-content');
     await page.locator('[data-testid="education-drawer-close"]').click();
--- a/apps/lfx-one/e2e/marketing-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/marketing-dashboard.spec.ts
@@ -3,6 +3,8 @@
 
 import { test, expect, Page } from '@playwright/test';
 
+import type { TrainingCertificationSummaryResponse } from '@lfx-one/shared/interfaces';
+
 /**
  * Marketing Dashboard E2E Tests
  *
@@ -199,6 +201,62 @@ test.describe('Marketing Metric Cards', () => {
   });
 });
 
+/**
+ * Stub the training-certification endpoint so a test can exercise a specific data
+ * shape instead of whatever TLF currently returns. Must be called before `page.goto`:
+ * the Education card is built during SSR, so a route registered after navigation
+ * would only affect subsequent client-side fetches and leave the card on live data.
+ */
+async function stubEducationSummary(page: Page, summary: TrainingCertificationSummaryResponse): Promise<void> {
+  await page.route('**/api/analytics/training-certification-summary*', (route) =>
+    route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(summary) })
+  );
+}
+
+/** Baseline fixture: every format active, revenue measured, no single format dominant. */
+const EDUCATION_BALANCED: TrainingCertificationSummaryResponse = {
+  projectId: 'tlf-test-project',
+  range: 'YTD',
+  enrollment: { instructorLed: 300, eLearning: 300, certExams: 250, edx: 150 },
+  revenue: { instructorLed: 120_000, eLearning: 60_000, certExams: 45_000 },
+};
+
+/**
+ * All-edX fixture: enrollments exist (so the card renders) but every revenue-bearing
+ * format is empty, making net revenue genuinely untracked rather than zero. This is
+ * the branch the card/drawer contradiction lived in and that live TLF data never hits.
+ */
+const EDUCATION_ALL_EDX: TrainingCertificationSummaryResponse = {
+  projectId: 'tlf-test-project',
+  range: 'YTD',
+  enrollment: { instructorLed: 0, eLearning: 0, certExams: 0, edx: 900 },
+  revenue: { instructorLed: 0, eLearning: 0, certExams: 0 },
+};
+
+/**
+ * Measured-zero fixture: instructor-led enrollments exist but earned $0. Distinct from
+ * the all-edX case above — here revenue IS tracked and happens to be zero, so both card
+ * and drawer must render `$0` rather than the untracked em dash.
+ */
+const EDUCATION_MEASURED_ZERO: TrainingCertificationSummaryResponse = {
+  projectId: 'tlf-test-project',
+  range: 'YTD',
+  enrollment: { instructorLed: 120, eLearning: 40, certExams: 20, edx: 60 },
+  revenue: { instructorLed: 0, eLearning: 0, certExams: 0 },
+};
+
+/**
+ * Concentration-risk fixture: one format above the 70% threshold with zero cert exams,
+ * which drives both the "Diversify education formats" and "No certification exam
+ * activity" actions into Needs Your Attention deterministically.
+ */
+const EDUCATION_CONCENTRATED: TrainingCertificationSummaryResponse = {
+  projectId: 'tlf-test-project',
+  range: 'YTD',
+  enrollment: { instructorLed: 900, eLearning: 50, certExams: 0, edx: 50 },
+  revenue: { instructorLed: 250_000, eLearning: 5_000, certExams: 0 },
+};
+
 test.describe('Education Drawer', () => {
   test.beforeEach(async ({ page }) => {
     await page.goto(DASHBOARD_URL);
@@ -319,6 +377,116 @@ test.describe('Education Drawer', () => {
     await openDrawer(page, 'ed-evo-education', 'education-drawer-content');
     await page.locator('[data-testid="education-drawer-close"]').click();
     await expect(page.locator('[data-testid="education-drawer-content"]')).not.toBeVisible();
+  });
+});
+
+/**
+ * Education branches that live TLF data does not reach.
+ *
+ * The suite above asserts invariants against whatever the API currently returns, which
+ * catches real integration breakage but leaves the interesting branches unexercised:
+ * TLF has revenue-bearing enrollments, so the untracked-revenue path never runs, and its
+ * format mix decides on its own whether the concentration-risk action fires. A rendering
+ * regression in either branch would pass unnoticed. These tests pin the response instead,
+ * so each branch fails deterministically when it breaks.
+ */
+test.describe('Education Drawer — fixture-driven branches', () => {
+  test('renders a measured $0 as $0 on both card and drawer', async ({ page }) => {
+    await stubEducationSummary(page, EDUCATION_MEASURED_ZERO);
+    await page.goto(DASHBOARD_URL);
+    await switchToExecutiveDirector(page);
+
+    // Revenue is tracked (instructor-led enrollments exist) and happens to be zero, so
+    // the card must state the figure rather than fall back to the untracked wording.
+    const card = page.locator('[data-testid="ed-evo-education"]');
+    await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
+    await card.scrollIntoViewIfNeeded();
+    await expect(card).toContainText('$0');
+    await expect(card).not.toContainText('net revenue not tracked');
+
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-stats');
+    const netRevenueStat = page.locator('[data-testid="education-drawer-stats"] > *', { hasText: 'Net Revenue' });
+    await expect(netRevenueStat).toContainText('$0');
+    await expect(netRevenueStat).not.toContainText('—');
+  });
+
+  test('renders untracked revenue as a dash on both card and drawer', async ({ page }) => {
+    await stubEducationSummary(page, EDUCATION_ALL_EDX);
+    await page.goto(DASHBOARD_URL);
+    await switchToExecutiveDirector(page);
+
+    // edX carries enrollments but no revenue column, so net revenue is untracked rather
+    // than zero. This is the case where the card previously read "$0.00 net revenue"
+    // while the drawer showed an em dash for the same figure.
+    const card = page.locator('[data-testid="ed-evo-education"]');
+    await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
+    await card.scrollIntoViewIfNeeded();
+    await expect(card).toContainText('net revenue not tracked');
+    await expect(card).not.toContainText('$0');
+
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-stats');
+    const netRevenueStat = page.locator('[data-testid="education-drawer-stats"] > *', { hasText: 'Net Revenue' });
+    await expect(netRevenueStat).toContainText('—');
+    await expect(netRevenueStat).not.toContainText('$');
+  });
+
+  test('routes a dominant format into Needs Your Attention only', async ({ page }) => {
+    await stubEducationSummary(page, EDUCATION_CONCENTRATED);
+    await page.goto(DASHBOARD_URL);
+    await switchToExecutiveDirector(page);
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-formats');
+
+    // 900 of 1000 enrollments clears the >70% concentration threshold, and zero cert
+    // exams triggers the credential-gap action — both are high priority.
+    const attention = page.locator('[data-testid="education-drawer-attention"]');
+    await expect(attention).toBeVisible();
+    await expect(attention).toContainText('Diversify education formats');
+    await expect(attention).toContainText('No certification exam activity');
+
+    // The same dominant format must not also be praised. It is reported as a neutral
+    // fact instead, which is the fix for the earlier flag-and-praise contradiction.
+    await expect(page.locator('[data-testid="education-drawer-facts"]')).toContainText('Instructor Led leads');
+    const performing = page.locator('[data-testid="education-drawer-performing"]');
+    if ((await performing.count()) > 0) {
+      await expect(performing).not.toContainText('Instructor Led');
+    }
+  });
+
+  test('routes a balanced mix into Performing Well without a concentration flag', async ({ page }) => {
+    await stubEducationSummary(page, EDUCATION_BALANCED);
+    await page.goto(DASHBOARD_URL);
+    await switchToExecutiveDirector(page);
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-formats');
+
+    // Four active formats, none above 50%, so the balanced-portfolio insight fires and
+    // the concentration-risk action must not.
+    const performing = page.locator('[data-testid="education-drawer-performing"]');
+    await expect(performing).toBeVisible();
+    await expect(performing).toContainText('Balanced format mix');
+
+    const attention = page.locator('[data-testid="education-drawer-attention"]');
+    if ((await attention.count()) > 0) {
+      await expect(attention).not.toContainText('Diversify education formats');
+    }
+  });
+
+  test('keeps edX untracked while other formats show measured revenue', async ({ page }) => {
+    await stubEducationSummary(page, EDUCATION_BALANCED);
+    await page.goto(DASHBOARD_URL);
+    await switchToExecutiveDirector(page);
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-formats');
+
+    // Four rows, one per format — a regression that dropped or merged rows would pass
+    // a mere visibility assertion on the container.
+    await expect(page.locator('[data-testid="education-drawer-format-row"]')).toHaveCount(4);
+
+    const edxRow = page.locator('[data-testid="education-drawer-format-row"]', { hasText: 'edX' });
+    await expect(edxRow).toContainText('not tracked');
+    await expect(edxRow).not.toContainText('$');
+
+    const instructorLedRow = page.locator('[data-testid="education-drawer-format-row"]', { hasText: 'Instructor Led' });
+    await expect(instructorLedRow).toContainText('$120,000');
+    await expect(instructorLedRow).not.toContainText('not tracked');
   });
 });
 
--- a/apps/lfx-one/e2e/marketing-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/marketing-dashboard.spec.ts
@@ -182,9 +182,48 @@ test.describe('Marketing Metric Cards', () => {
     await expect(card).toContainText('Web');
   });
 
+  // The Education card is suppressed when the foundation reports zero enrollments, so
+  // assert on attachment rather than a hard failure if TLF ever drops to no training.
+  test('renders Education card', async ({ page }) => {
+    const card = page.locator('[data-testid="ed-evo-education"]');
+    await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
+    await card.scrollIntoViewIfNeeded();
+    await expect(card).toBeVisible();
+    await expect(card).toContainText('Education');
+  });
+
   test('renders filter pills', async ({ page }) => {
     const filters = page.locator('[data-testid="ed-evolution-filters"]');
     await expect(filters).toBeVisible();
+  });
+});
+
+test.describe('Education Drawer', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(DASHBOARD_URL);
+    await switchToExecutiveDirector(page);
+  });
+
+  test('opens and shows title when card is clicked', async ({ page }) => {
+    await openDrawer(page, 'ed-evo-education', 'education-drawer-content');
+    await expect(page.locator('[data-testid="education-drawer-title"]')).toContainText('Education');
+  });
+
+  test('shows stats section with data', async ({ page }) => {
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-stats');
+    const statCards = page.locator('[data-testid="education-drawer-stats"]').locator('lfx-card');
+    await expect(statCards).toHaveCount(3);
+  });
+
+  test('shows format breakdown section', async ({ page }) => {
+    await openDrawer(page, 'ed-evo-education', 'education-drawer-content');
+    await expect(page.locator('[data-testid="education-drawer-formats"]')).toBeVisible();
+  });
+
+  test('closes when close button is clicked', async ({ page }) => {
+    await openDrawer(page, 'ed-evo-education', 'education-drawer-content');
+    await page.locator('[data-testid="education-drawer-close"]').click();
+    await expect(page.locator('[data-testid="education-drawer-content"]')).not.toBeVisible();
   });
 });
 
--- a/apps/lfx-one/e2e/marketing-dashboard.spec.ts
+++ b/apps/lfx-one/e2e/marketing-dashboard.spec.ts
@@ -182,8 +182,9 @@ test.describe('Marketing Metric Cards', () => {
     await expect(card).toContainText('Web');
   });
 
-  // The Education card is suppressed when the foundation reports zero enrollments, so
-  // assert on attachment rather than a hard failure if TLF ever drops to no training.
+  // The card is conditional on non-zero enrollments upstream, so it is asserted on
+  // attachment first — it renders off-screen in the carousel until scrolled to. TLF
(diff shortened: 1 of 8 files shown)


apps/lfx-one/src/app/modules/dashboards/executive-director/components/education-drawer/education-drawer.component.html
```diff
@@ -0,0 +1,168 @@
...
+        </div>
+      </lfx-card>
+      <lfx-card styleClass="flex-1">
+        <div class="flex flex-col gap-1">
+          <span class="text-sm text-gray-500">Net Revenue</span>
+          @if (totalRevenue() > 0) {
+            <span class="text-2xl font-semibold text-gray-900">{{ totalRevenueLabel() }}</span>
+          } @else {
+            <span class="text-2xl font-semibold text-gray-400">—</span>
+          }
```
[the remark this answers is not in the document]
mrautela365: Fixed in 67c418d39. Added a `hasTrackedRevenue` signal that gates on the presence of revenue-bearing enrollments (instructor-led + eLearning + cert exams) rather than on a non-zero total, and the template now uses it. A measured $0 renders as $0; the em dash is reserved for the all-edX case where `COURSE_PURCHASES` carries no revenue column and the figure is genuinely untracked. This also makes the stat cards consistent with the format rows, which already distinguish null from 0.

On the line-55 note: that is the Revenue / Enrollment card, which renders `revenuePerEnrollmentLabel()`. It already guards on the same revenue-bearing enrollment count to avoid divide-by-zero, so it was correct as written.

apps/lfx-one/src/app/modules/dashboards/executive-director/components/education-drawer/education-drawer.component.ts
```diff
@@ -0,0 +1,193 @@
...
+      }
+
+      const rows = this.categoryRows();
+
+      // Leading format
+      const top = [...rows].sort((a, b) => b.enrollments - a.enrollments)[0];
+      if (top.enrollments > 0) {
+        insights.push({
+          text: `${top.label} leads with ${formatNumber(top.enrollments)} enrollments (${top.enrollmentSharePct.toFixed(0)}% of total)`,
+          type: 'info',
```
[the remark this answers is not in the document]
mrautela365: Fixed in 67c418d39. Confirmed against `splitByPriority()` in `packages/shared/src/utils/marketing.utils.ts` — it routes both `info` and `driver` into `performingInsights`, so the leading-format line could praise the exact format the concentration-risk action was flagging as fragile.

Rather than reclassify them, I moved the leading-format and highest-revenue lines out of `keyInsights` entirely into a `neutralFacts` signal rendered in its own unstyled block. They are neutral observations, not verdicts, so they should not carry a green check or a red warning either way. `keyInsights` now emits only the balanced-portfolio line, which is a genuine performing-well signal.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
```diff
@@ -182,12 +182,51 @@ test.describe('Marketing Metric Cards', () => {
     await expect(card).toContainText('Web');
   });
 
+  // The Education card is suppressed when the foundation reports zero enrollments, so
+  // assert on attachment rather than a hard failure if TLF ever drops to no training.
+  test('renders Education card', async ({ page }) => {
+    const card = page.locator('[data-testid="ed-evo-education"]');
+    await expect(card).toBeAttached({ timeout: DATA_LOAD_TIMEOUT });
```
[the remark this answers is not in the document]
mrautela365: You are right that `toBeAttached()` is still a hard assertion — the comment claimed a resilience the code did not have. Fixed the comment rather than the assertion in 67c418d39.

TLF always reports training activity, so an absent Education card is a real regression and should fail the suite. Softening it to a conditional skip would make the test silently pass in exactly the case worth catching. `toBeAttached` is still the right first assertion here for a different reason: the card renders off-screen inside the carousel, so it is attached before it is visible, and the following `scrollIntoViewIfNeeded` + `toBeVisible` pair depends on that ordering. The comment now says this instead of the zero-enrollment claim.

mrautela365:
All bot review threads resolved and CI is green. Ready for review.

**Fixed since the last round** (67c418d39), all four Copilot findings:
- Education card category was `memberships`, and since that pill is labelled **North Star** and the carousel filters on exact category equality, Education was being listed as a North Star metric. Now `education`.
- Net Revenue gated on a non-zero total, so a measured `$0` rendered as an em dash. Now gated on the presence of revenue-bearing enrollments; the dash is reserved for the all-edX case where revenue is genuinely untracked upstream.
- The leading-format and highest-revenue lines were `info` insights, and `splitByPriority()` routes every non-warning insight into **Performing Well** — so they praised the same format the concentration-risk action was flagging as fragile. Moved to a neutral facts block that carries no verdict either way.
- Corrected the carousel sequence comment (Web was missing) and dropped an inaccurate zero-enrollment resilience claim on the e2e card test.

**Merge order:** last in a 3-PR stack — #1308 → #1309 → #1310. This PR is based on a combined base branch containing both; once they merge I will retarget it at `main` and delete that base.

apps/lfx-one/src/app/modules/dashboards/executive-director/components/education-drawer/education-drawer.component.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      if (total === 0) {
+        return actions;
+      }
+
+      const rows = this.categoryRows();
+      const { enrollment } = this.data();
+
+      // Concentration risk — one format carrying the programme
+      const top = [...rows].sort((a, b) => b.enrollments - a.enrollments)[0];
```
dealako: [nit] Three redundant sorts to find single maxima

Issue: `initRecommendedActions()` (line 130) and `initNeutralFacts()` (lines 212, 220) each call `[...rows].sort(...)[0]` to extract one max-by-field element.

Proof: Each call is `O(n log n)` where a single `reduce()` pass would be `O(n)` and clearer about intent (finding a max, not sorting).

Why it matters: Purely a style/efficiency nit — `rows` is capped at 4 categories here, so there's no real performance consequence at this scale.

Fix: `rows.reduce((top, r) => r.enrollments > top.enrollments ? r : top, rows[0])` in place of `[...rows].sort(...)[0]`.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
```diff
@@ -182,12 +182,52 @@ test.describe('Marketing Metric Cards', () => {
...
+    await expect(page.locator('[data-testid="education-drawer-title"]')).toContainText('Education');
+  });
+
+  test('shows stats section with data', async ({ page }) => {
+    await openDrawerAndWaitForData(page, 'ed-evo-education', 'education-drawer-content', 'education-drawer-stats');
+    const statCards = page.locator('[data-testid="education-drawer-stats"]').locator('lfx-card');
+    await expect(statCards).toHaveCount(3);
+  });
+
+  test('shows format breakdown section', async ({ page }) => {
```
dealako, line 284: [minor] Missing e2e coverage for the $0-vs-"not tracked" distinction

Issue: The PR's core correctness claim is that a measured `$0` renders differently from edX's genuinely-untracked revenue (em dash), but no test asserts on the format-breakdown row values or the drawer-level em-dash case.

Proof: The added tests here (`shows format breakdown section`) only check that `[data-testid="education-drawer-formats"]` is visible — not its content. `hasTrackedRevenue()` and the per-format null-vs-zero handling in the component have no test asserting their actual rendered output.

Why it matters: This is exactly the kind of logic a future refactor could silently invert (e.g. rendering edX as `$0` instead of "not tracked"), and nothing here would catch it.

Fix: Add an assertion on the format-breakdown row text for a format with `null` revenue (renders "not tracked") vs. one with `0` revenue (renders `$0`), and/or a test against fixture data where all enrollments are edX to confirm the Net Revenue stat shows the em dash.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
```diff
@@ -182,12 +182,52 @@ test.describe('Marketing Metric Cards', () => {
...
+    await expect(card).toContainText('Education');
+  });
+
   test('renders filter pills', async ({ page }) => {
     const filters = page.locator('[data-testid="ed-evolution-filters"]');
     await expect(filters).toBeVisible();
   });
 });
 
+test.describe('Education Drawer', () => {
```
dealako, line 267: [minor] No e2e coverage for the conditional action/insight sections

Issue: The drawer renders conditional "Needs Your Attention" / "Performing Well" sections driven by computed logic (concentration risk >70%, cert-exam attach rate, unmonetized formats, balanced portfolio) — none of the added tests touch these.

Proof: The four tests in this `describe` block (open/title, stats count, format-breakdown visibility, close) don't reference `education-drawer-actions`/`education-drawer-insights` or any equivalent testids for these sections.

Why it matters: A regression in the concentration-risk threshold or the `splitByPriority` routing (the exact class of bug Copilot's prior review caught) wouldn't be caught by CI.

Fix: Add at least one test asserting the attention section renders for a foundation with concentrated enrollment, and one confirming it's absent for a balanced portfolio.

dealako (approved):
Hey @mrautela365 — nice work on this one. The null/zero/error disambiguation for the training data (empty-`projectId` sentinel → `undefined`, `hasTrackedRevenue` gating the em dash, edX rendering as "not tracked" rather than `$0`) is exactly the kind of subtlety that's easy to get wrong, and tracing it through `education-drawer.component.ts` and the marketing utils it reuses, it holds up. Reusing the Health Metrics endpoint instead of adding a new one is a good call for keeping the two cards from disagreeing.

I ran three independent passes (security/privacy, correctness/performance/tests, style/conventions) instead of reading the diff myself, then reconciled against the existing bot threads.

**Bot reconciliation:** Copilot's four prior findings (North Star category leak, stale carousel-sequence comment, the `$0`-vs-untracked em dash bug, and the `splitByPriority` insight/warning contradiction, plus the `toBeAttached()` comment overclaiming resilience) are all confirmed fixed in 67c418d3 — I checked the current code against each and they hold. CodeRabbit auto-review was skipped (repo policy, likely due to the stacked base branch), and Cursor's bot summary flagged this correctly as low risk with no findings of its own.

### 🔴 Blocking: 0 issues

### 🟡 Minor: 2 issues
- E2E coverage doesn't assert the actual `$0`-vs-"not tracked" distinction in the format breakdown, or the all-edX em-dash case — exactly the behavior this PR is built around.
- E2E coverage doesn't exercise the conditional "Needs Your Attention" / "Performing Well" sections (concentration risk, cert-exam attach rate, balanced portfolio) at all.

### ⚪ Nit: 1 issue
- Three separate `[...rows].sort(...)[0]` calls to find single maxima — trivial at 3-4 rows, but a single-pass reduce would be cheaper and clearer.

### ❔ Question: 0 items

## Decision: ✅ Approved with minor comments

Nothing here blocks merge — the two minor items are follow-up-worthy test-coverage gaps on the feature's actual differentiator (the null/zero/error distinction), not correctness bugs in the shipped code.

apps/lfx-one/src/app/modules/dashboards/executive-director/components/education-drawer/education-drawer.component.ts
mrautela365: Done in ef3e07a — all three sorts replaced with a single-pass `maxBy(rows, score)` helper. Agreed the sort implied an ordering nothing consumed; the helper also names the intent. All three call sites are length-guarded (`total === 0` early-returns, and `revenueRows.length > 0`), and `categoryRows()` always returns four rows, so the `rows[0]` seed is safe.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
mrautela365: Done in ef3e07a — added `distinguishes untracked revenue from a measured zero` to `marketing-dashboard.spec.ts`. It opens the education drawer and asserts the edX row renders `not tracked` and no currency, while the Instructor Led row renders a currency figure and never `not tracked`. That locks the null-vs-0 distinction the drawer was built around (`row(edX, enrollment.edx, null)` is hardcoded null, and the template branches on `row.revenue !== null`).

apps/lfx-one/e2e/marketing-dashboard.spec.ts
mrautela365: Done in ef3e07a — added `renders a priority-routed insight section`. It asserts at least one of the three priority-routed sections (`education-drawer-attention`, `education-drawer-performing`, `education-drawer-facts`) renders, and that any section present is non-empty. Deliberately not pinned to a specific section: which one appears depends on live TLF enrollment mix, so asserting a fixed section would make the test flip with the data rather than with the code.

MRashad26 (approved):
Code enforcer: clean. `EducationDrawerComponent` is input-driven with no direct fetch — all template bindings are signal reads (`totalEnrollmentsLabel()`, `categoryRows()`, etc.) or pipes (`| number`, `| currency`), no function calls in the template. `EducationCategoryRow` interface is correctly in `@lfx-one/shared/interfaces`. The null-vs-zero distinction for edX revenue (pre-rounded `enrollmentShareLabel` on the view model avoids `toFixed()` in the template) is a good pattern. The `projectId !== ''` sentinel to map the analytics service's error fallback back to `undefined` is sound — keeps "failed request" and "genuinely zero" distinguishable.

packages/shared/src/constants/dashboard-metrics.constants.ts
```diff
@@ -825,6 +849,34 @@ export function buildEdEvolutionMetrics(data: EdEvolutionData): DashboardMetricC
...
+      ? [
+          {
+            title: 'Education',
+            icon: 'fa-light fa-graduation-cap',
+            chartType: 'line',
+            category: 'education',
+            testId: 'ed-evo-education',
+            description: 'Training and certification enrollments across instructor-led, eLearning, cert exams, and edX.',
+            value: formatNumber(educationTotalEnrollments),
+            subtitle: `${formatCurrency(educationTotalRevenue)} net revenue · ${education?.range === 'YTD' ? 'YTD' : 'selected year'}`,
```
dealako: [minor] Education card subtitle shows "$0.00 net revenue" for an all-edX foundation, unlike the drawer

Issue: `educationTotalRevenue` excludes edX by design (no revenue column upstream), so a foundation whose only training activity is edX has `educationTotalRevenue === 0`. The card subtitle renders `formatCurrency(educationTotalRevenue)` unconditionally, so it reads "$0.00 net revenue" — indistinguishable from a foundation that measurably earned zero. The drawer already solves this exact problem with `hasTrackedRevenue` (renders an em dash for the all-edX case, `$0` for a measured zero) — this card just doesn't apply the same gate.

Proof: Line 870 — `subtitle: \`${formatCurrency(educationTotalRevenue)} net revenue · ...\`` — no conditional on whether any revenue-bearing enrollment exists, unlike `education-drawer.component.ts`'s `hasTrackedRevenue` signal (`e.instructorLed + e.eLearning + e.certExams > 0`).

Why it matters: This is the exact "silently reports untracked as earned nothing" failure mode the PR description calls out as deliberately avoided — it's just avoided in the drawer and not in the card that leads to it. A 100%-edX foundation would see "$0.00 net revenue" on the card, then open the drawer and see an em dash for the same figure — a visible contradiction.

Fix: Compute a `hasTrackedRevenue`-equivalent boolean here (`education.revenue.instructorLed + education.revenue.eLearning + education.revenue.certExams > 0` guarded on `education` existing) and branch the subtitle: `` `${hasTrackedRevenue ? formatCurrency(educationTotalRevenue) + ' net revenue' : 'revenue not tracked'} · ...` `` (or similar wording consistent with the drawer's em dash).

apps/lfx-one/e2e/marketing-dashboard.spec.ts
```diff
@@ -182,12 +182,93 @@ test.describe('Marketing Metric Cards', () => {
...
+  test('shows format breakdown section', async ({ page }) => {
+    await openDrawer(page, 'ed-evo-education', 'education-drawer-content');
+    await expect(page.locator('[data-testid="education-drawer-formats"]')).toBeVisible();
+  });
+
+  // Guards the null-vs-zero distinction: edX revenue is hardcoded null (COURSE_PURCHASES
+  // carries no revenue column) and must render "not tracked", while every other format
+  // has measured revenue and renders a currency figure — $0 included. A refactor that
+  // coalesced null to 0 would silently report edX as having earned nothing.
+  test('distinguishes untracked revenue from a measured zero', async ({ page }) => {
```
dealako, line 293: [minor] Doesn't cover the stat-card-level em-dash case (partially addressed)

Issue: This test locks in the format-breakdown row distinction (edX → "not tracked", Instructor Led → currency) but doesn't test the Net Revenue *stat card* above it, which is gated by the separate `hasTrackedRevenue` signal and should show an em dash for an all-edX foundation.

Proof: The test only queries `[data-testid="education-drawer-format-row"]` rows; nothing asserts on the Net Revenue stat card's rendered text for the all-edX scenario.

Why it matters: `hasTrackedRevenue` could regress (e.g., someone "simplifies" it back to a non-zero-total check) without this test catching it — and per the inline comment on `dashboard-metrics.constants.ts:870`, the equivalent card-level bug already exists uncaught.

Fix: Once the card-subtitle fix lands, add an assertion (ideally against fixture/TLF data known to be all-edX, or a second test dataset) that the drawer's Net Revenue stat card renders an em dash rather than a currency value in that case.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
```diff
@@ -182,12 +182,93 @@ test.describe('Marketing Metric Cards', () => {
...
+    const instructorLedRow = page.locator('[data-testid="education-drawer-format-row"]', { hasText: 'Instructor Led' });
+    await expect(instructorLedRow).toContainText('$');
+    await expect(instructorLedRow).not.toContainText('not tracked');
+  });
+
+  // The attention/performing sections are driven by computed thresholds (concentration
+  // risk >70%, cert attach rate, balanced-portfolio) and routed by splitByPriority.
+  // TLF's live mix decides which of the two renders, so assert the invariant that holds
+  // either way: at least one section renders, and whichever renders carries content.
+  test('renders a priority-routed insight section', async ({ page }) => {
```
dealako, line 339: [minor] Smoke-tests presence but not the routing thresholds (partially addressed)

Issue: This asserts at least one of the three priority-routed sections renders and is non-empty, but doesn't exercise the specific thresholds that decide which section a given fact/action/insight routes into (concentration risk >70%, cert-exam attach rate <10%/zero, balanced-portfolio, the leading/highest-earning "neutral fact" exclusion from `splitByPriority`).

Proof: The test only counts `attention`/`performing`/`facts` container elements and checks non-emptiness — it doesn't assert on which specific action/insight text appears in which section for a known data shape.

Why it matters: This is the exact class of regression Copilot's prior round caught (a leading-format insight landing in "Performing Well" while the same format was flagged as a concentration risk) — the current test would pass even if that contradiction reappeared, as long as some section had some content.

Fix: Optional follow-up, not a blocker — if a fixture with a known concentration-risk mix is available, assert that the same format doesn't appear in both an attention action and a performing insight simultaneously.

apps/lfx-one/src/app/modules/dashboards/executive-director/components/education-drawer/education-drawer.component.ts
```diff
@@ -0,0 +1,237 @@
...
+    });
+  }
+
+  // === Private Helpers ===
+  /**
+   * Returns the row with the highest `score`. A single pass states the intent —
+   * find one maximum — where a full sort implies an ordering nothing consumes.
+   * Callers must pass a non-empty array; every call site is already length-guarded.
+   */
+  private maxBy(rows: EducationCategoryRow[], score: (row: EducationCategoryRow) => number): EducationCategoryRow {
```
dealako, line 238: [nit] Reduce scores the seed element redundantly

Issue: `rows.reduce((top, row) => (score(row) > score(top) ? row : top), rows[0])` calls `score()` on `rows[0]` twice on the first iteration (once as `row`, once as `top`) before settling into single calls per subsequent element.

Proof: First iteration: `row === rows[0] === top` (the seed), so `score(rows[0])` runs twice for no comparative purpose.

Why it matters: Purely cosmetic at `rows.length` ≤ 4 — no behavioral or measurable perf consequence here.

Fix: Not worth changing unless `score` becomes expensive; if you do, a plain `for` loop starting at index 1 avoids the redundant call.

dealako (changes requested):
Hey @mrautela365 — thanks for turning this around fast. Ran the same three-pass process (security/privacy, correctness/performance/tests, style/conventions) scoped to just the diff between 67c418d3 and ef3e07a, plus re-verified each of the three prior findings against the actual code rather than trusting the commit message.

👏 **Nice work:**
- The `maxBy` helper is a clean fix — single-pass, well-documented (including the tie-break and non-empty-array precondition), and correctly replaces all three `sort()` call sites. Fully resolved.
- Both new e2e tests move from "is it visible" to actual content assertions — `distinguishes untracked revenue from a measured zero` locks in the exact null-vs-`$0` distinction this PR is built around, and the priority-routed insight test correctly avoids pinning to a specific section (since that depends on live TLF data) while still asserting the non-empty invariant. Real progress on both.

**Bot reconciliation:** Copilot's fresh pass (05:12) surfaced something none of us caught in the first round — I verified it directly and it's real (see inline comment on `dashboard-metrics.constants.ts:870`). Nice catch by Copilot; folding it in below since it's the same bug class this PR already fixed once in the drawer.

### 🔴 Blocking: 0 issues

### 🟡 Minor: 3 issues
- **New** — the Education *card* subtitle still shows `$0.00 net revenue` for an all-edX foundation, even though the drawer correctly shows an em dash for the identical case (Copilot catch, confirmed).
- **Partially addressed** — the null-vs-zero e2e test covers the format-breakdown rows but not the stat-card-level em-dash case for an all-edX foundation (same root cause as the new finding above — a single fix + test should close both).
- **Partially addressed** — the priority-routed-section e2e test is a non-empty smoke test; it doesn't yet exercise the specific threshold logic (concentration risk >70%, cert attach rate, balanced portfolio) that routes into it.

### ⚪ Nit: 1 issue
- `maxBy`'s reduce scores the seed element (`rows[0]`) redundantly on the first iteration — harmless at n=4, but a plain loop would avoid the double call.

## Decision: 🔴 Needs changes before approval

Only because of the count (3 minors crosses the 2-issue approval line) — none of these are structural. The card-subtitle bug is the one worth fixing before merge since it's a real user-facing inconsistency; the two partially-addressed test gaps are good-to-have follow-ups, not blockers on their own merits.

packages/shared/src/constants/dashboard-metrics.constants.ts
mrautela365: Good catch — this was a real contradiction and you traced it exactly. Fixed in 7ebbb7a.

Added `educationHasTrackedRevenue` gated on revenue-bearing *enrollments* (`instructorLed + eLearning + certExams > 0`), matching the drawer signal rather than a non-zero revenue total — so a foundation that measurably earned $0 still renders `$0`, and only the all-edX case reads `net revenue not tracked`. Card and drawer now derive the same distinction from the same rule.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
mrautela365: Fixed in 7ebbb7a, now that the card-side gate you asked for has landed.

No all-edX fixture exists and live TLF data is not all-edX, so instead of asserting one branch the new test `card and drawer agree on whether revenue is tracked` asserts the invariant that must hold under either shape: it reads the card, reads the drawer Net Revenue stat, and requires they never disagree — then checks the rendering is the correct one for whichever way they agree. That fails on the regression you named (someone simplifying `hasTrackedRevenue` back to a non-zero-total check) because the two sides would diverge, and it is not dependent on which foundation the suite runs against.

apps/lfx-one/e2e/marketing-dashboard.spec.ts
mrautela365: Addressed in 7ebbb7a rather than deferred.

Full threshold coverage does need fixtures we do not have, but the specific regression class you pointed at is assertable without one. Added `never flags and praises the same education format`: for each of the four formats it requires the format not appear in both "Needs Your Attention" and "Performing Well" simultaneously — the exact contradiction Copilot caught last round. It skips when TLF live data renders only one of the two sections, so it does not produce a false green.

Agreed the threshold-by-threshold assertions are follow-up work once a fixture harness exists.

apps/lfx-one/src/app/modules/dashboards/executive-director/components/education-drawer/education-drawer.component.ts
mrautela365: Fixed in 7ebbb7a. You are right that it is cosmetic at four rows, but it was free to fix, so `maxBy` now carries the running best score and evaluates `score` exactly once per row. Documented the tie-break (earlier row wins, preserving `categoryRows` declaration order) since the loop makes that explicit where reduce left it implicit.

mrautela365:
All four findings addressed in 7ebbb7a — including the card/drawer revenue contradiction, which was a real bug and a good catch.

- **Card subtitle (blocking-ish):** added `educationHasTrackedRevenue`, gated on revenue-bearing enrollments to mirror the drawer. A measured $0 still renders `$0`; only the all-edX case reads `net revenue not tracked`.
- **Stat-card coverage:** new test asserts card and drawer never disagree about whether revenue is tracked — an invariant that holds without an all-edX fixture.
- **Routing thresholds:** new test asserts no format is simultaneously flagged as a concentration risk and praised as performing well, the exact contradiction from the earlier round. Skips cleanly when TLF data renders only one section.
- **maxBy seed:** now scores each row exactly once.

Type-check and lint clean; both new tests register across chromium/firefox/mobile-chrome. All six threads replied to and resolved.

mrautela365:
Both remaining test gaps are now closed in fd41242 — and Copilot's suppressed comment on the same round was making the identical point, so this covers it too.

**The gap you both identified:** every education test read whatever TLF happened to return. TLF has revenue-bearing enrollments, so the untracked-revenue branch never executed, and its format mix decided on its own whether the concentration-risk action fired. A rendering regression in either branch would have passed silently — the assertions were invariants, not branch coverage.

**Fix:** stubbed `**/api/analytics/training-certification-summary*` with four fixtures — balanced, all-edX, measured-zero, and concentrated. One implementation detail worth noting: the route has to be registered *before* `page.goto`, because the Education card is built during SSR; a route registered after navigation would only affect later client-side fetches and leave the card on live data.

Five new tests, each pinning a branch that previously could not fail:

| Test | Branch it forces |
|---|---|
| `renders a measured $0 as $0 on both card and drawer` | tracked revenue that happens to be zero |
| `renders untracked revenue as a dash on both card and drawer` | all-edX — the exact card/drawer contradiction from last round |
| `routes a dominant format into Needs Your Attention only` | >70% concentration + zero cert exams, and asserts the dominant format is *not* also praised |
| `routes a balanced mix into Performing Well without a concentration flag` | balanced-portfolio insight, concentration action absent |
| `keeps edX untracked while other formats show measured revenue` | asserts row count of 4 and exact currency, not just container visibility |

That last one addresses Copilot's earlier point that a bare `toBeVisible()` on the formats container would pass even if the rows vanished or edX regressed from `not tracked` to `$0`.

**Kept the live-data invariant tests deliberately.** They exercise the real endpoint and catch integration breakage — a contract change, a serialization bug — that fixtures by construction cannot. The fixtures sit beneath them as the deterministic layer.

**On the two items from the previous round:** the card-subtitle bug and the `maxBy` seed nit were fixed in 7ebbb7a, which you confirmed. So all four items from your review are now closed.

Type-check clean on `tsconfig.e2e.json`, lint clean, and all five tests register across chromium/firefox/mobile-chrome.

One note on scope I did not fix: `apps/lfx-one/tsconfig.e2e.json` fails `tsc` outright with `TS5095` (`module: commonjs` against an inherited `moduleResolution: bundler`). It is identical on `origin/main`, so it predates this PR and I left it alone rather than widen the diff — happy to open a separate ticket if you want it tracked.

mrautela365:
Branch updated to current `main` (be608cb) — no review-relevant change.

`main` had moved ahead by 3 commits (profile/CLA and committee join-application work, none of it touching these files). Both this branch and its base `feat/LFXV2-2023-web-and-social-base` now carry them.

**The reviewable diff is unchanged: 8 files, +879/−9 — identical to before the update.** Worth stating explicitly because the naive approach would have made this much worse: merging `main` into this branch alone left two merge bases, and GitHub then rendered the diff against `main`, ballooning it to 38 files / +2300 and pulling in #1309's ~650-line `social-media-drawer` deletion as if it were part of this PR. Merging `main` into the base first, then the base into this branch, restores a single merge base and keeps the diff honest.

Verified after the merge: `yarn check-types`, `yarn lint:check`, and `yarn build` all pass, and the five fixture-driven tests still register across chromium/firefox/mobile-chrome.

**Still targeting the stacked base deliberately.** Retargeting at `main` right now would inflate this PR to +938/−709, since #1308 and #1309 have not landed yet. I will retarget once they merge, then delete the base branch — that is the point at which the diff stays at its true size.

@dealako this is ready for another look whenever you have a moment — your four items from the last round are addressed (card subtitle and `maxBy` in 7ebbb7a, the two test-coverage gaps in fd41242 with deterministic API fixtures).

dealako (approved):
@mrautela365 this is in great shape — every open item from the last round is closed, and the fixes are precise, not just "make the bot quiet."

## 👏 Nice work

- **`educationHasTrackedRevenue`** in `dashboard-metrics.constants.ts` mirrors the drawer's `hasTrackedRevenue` signal exactly (same enrollment-sum condition), so the card subtitle and the drawer's Net Revenue stat can no longer drift apart on the null-vs-zero distinction. That was a real contradiction and the fix closes it at the root rather than patching the symptom.
- The new `EDUCATION_ALL_EDX`, `EDUCATION_CONCENTRATED`, `EDUCATION_BALANCED`, and `EDUCATION_MEASURED_ZERO` API fixtures move the education e2e suite off live TLF data for the cases that actually need deterministic inputs — the stat-card em-dash case and the concentration-risk/balanced-portfolio routing thresholds are now pinned and asserted directly instead of smoke-tested.
- `maxBy` now scores each row exactly once (no more double-scoring the seed element) and the JSDoc explains the tie-breaking behavior — a genuine nit, fixed anyway because it was free.
- Good instinct catching that the `toBeAttached()` comment overclaimed resilience it didn't have — fixing the comment instead of loosening the assertion was the right call given TLF always reports data for this fixture.

## Verification (re-scanned only the diff since the last round, `ef3e07a3..be608cb4`, scoped to the three PR-owned files)

- ✅ **Resolved** — Card subtitle now renders "net revenue not tracked" for an all-edX foundation instead of "$0.00 net revenue" (`dashboard-metrics.constants.ts`, gated on `educationHasTrackedRevenue`).
- ✅ **Resolved** — New test `renders untracked revenue as a dash on both card and drawer` asserts the stat-card level em-dash case using `EDUCATION_ALL_EDX`, closing the previously partially-addressed gap.
- ✅ **Resolved** — New tests using `EDUCATION_CONCENTRATED` and `EDUCATION_BALANCED` fixtures assert the specific concentration-risk (>70%) and balanced-portfolio routing outcomes, closing the previously partially-addressed gap.
- ✅ **Resolved** — `maxBy` no longer double-scores the seed element on the first comparison.
- **Security/Data Privacy:** clean — no findings. Fixture data is clearly synthetic (`tlf-test-project`, fabricated enrollment/revenue numbers).
- **New issues in this round's diff:** none found across correctness, performance, test coverage, style, or documentation.
- **Merge status:** `MERGEABLE` / `CLEAN`. Confirmed the `be608cb4` merge commit is a non-review-relevant rebase onto `main` (per your comment) — no new files or logic from that merge touch the reviewed scope.
- No new bot (Copilot/CodeRabbit/Cursor) findings since the last round that aren't already covered above.

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

## Decision

✅ **Approved**

PR #1318, opened 2026-08-05
feat(groups): add public_name slug resolution for public group URLs

MRashad26:
## Summary
- When the `:id` route parameter on `/public/api/groups/:id` is not a UUID, resolve it as a `public_name` slug via query-service tag lookup (`public_name:<slug>`)
- Add `public_name` field to `Committee` and `PublicGroupDetail` interfaces
- Include `public_name` in the public group API response so the frontend can use it for canonical URLs

Depends on backend PR: https://github.com/linuxfoundation/lfx-v2-committee-service/pull/172

## Test plan
- [ ] Navigate to `/groups/<valid-uuid>` — page renders as before (no regression)
- [ ] Navigate to `/groups/<valid-public-name-slug>` — slug resolves and page renders
- [ ] Navigate to `/groups/<nonexistent-slug>` — shows not-found page
- [ ] Verify `public_name` is present in the API response when set on the committee


apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -47,25 +48,43 @@ export class PublicGroupsController {
       const m2mToken = await generateM2MToken(req);
       req.bearerToken = m2mToken;
 
-      const committee = await this.committeeService.getCommitteeById(req, id);
+      let committeeUid = id;
+      if (!UUID_REGEX.test(id)) {
+        const { resources } = await this.microserviceProxy.proxyRequest<QueryServiceResponse<Committee>>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', {
+          type: 'committee',
+          tags: `public_name:${id}`,
+          page_size: 1,
```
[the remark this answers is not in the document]
MRashad26: Good point on the uniqueness concern. This is an intentional design decision — uniqueness is enforced at the application layer (PCC) rather than at the query-service level. The `public_name` is set by PCC when a committee is made public, and PCC is responsible for generating a unique slug.

That said, the `resources[0].id` issue you flagged in the suppressed comment was a real bug — fixed in commit 48e1a8e42, which also switches to `ResourceNotFoundError` for the not-found case. If we need additional safety for duplicate slugs in the future, we can add a uniqueness check at the committee-service level.

See commit 48e1a8e42.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -47,25 +48,43 @@ export class PublicGroupsController {
...
+        });
+        if (resources.length === 0) {
+          throw new AuthorizationError('Group not found', {
+            operation: 'get_public_group_by_id',
+            service: 'public_groups_controller',
+            path: `/groups/${id}`,
+            code: 'GROUP_NOT_FOUND',
+          });
```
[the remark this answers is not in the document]
MRashad26: Fixed — changed `AuthorizationError` to `ResourceNotFoundError`, which returns a proper 404 instead of 403. This matches the `public-meeting.controller.ts` pattern and ensures the frontend shows the correct not-found page rather than the "this group is private" message.

See commit 48e1a8e42.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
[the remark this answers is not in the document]
MRashad26: Good suggestion. The analogous public controllers in this repo (e.g. `public-meeting.controller.ts`) also lack Vitest specs currently — adding controller tests for the slug resolution path is worth doing but would be better as a follow-up to keep this PR focused on the feature change.

The two bugs flagged in the other threads (wrong resource ID field and wrong error class) have been fixed in commit 48e1a8e42.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -47,25 +48,43 @@ export class PublicGroupsController {
...
+        }
+        committeeUid = resources[0].id;
```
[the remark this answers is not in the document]
MRashad26: Fixed — changed `resources[0].id` to `resources[0].data.uid`. You were correct that the query-service `id` is the type-prefixed index key (`committee:<uid>`), not the raw committee UID. Every other query-service consumer in the codebase uses `resources[0].data` for the entity payload.

See commit 48e1a8e42.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -47,25 +48,43 @@ export class PublicGroupsController {
...
+          page_size: 1,
+        });
+        if (resources.length === 0) {
+          throw new AuthorizationError('Group not found', {
+            operation: 'get_public_group_by_id',
+            service: 'public_groups_controller',
+            path: `/groups/${id}`,
+            code: 'GROUP_NOT_FOUND',
+          });
+        }
```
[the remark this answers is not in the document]
MRashad26: Fixed — changed `AuthorizationError` (HTTP 403) to `ResourceNotFoundError` (HTTP 404) for the not-found slug case. The frontend group detail page treats 403 as a private group and navigates to `not-found?reason=private`, so a nonexistent slug was incorrectly showing the private-group message. Now it returns a proper 404.

See commit 48e1a8e42.

MRashad26:
## Review Feedback Addressed

Commit: 48e1a8e42

### Changes Made
- **public-groups.controller.ts:66**: Changed `resources[0].id` to `resources[0].data.uid` — the query-service `id` is the type-prefixed index key (`committee:<uid>`), not the raw UID (per @copilot-pull-request-reviewer, @cursor)
- **public-groups.controller.ts:59-65**: Changed `AuthorizationError` (403) to `ResourceNotFoundError` (404) for missing slug lookups — frontend treats 403 as "private group" which is wrong UX for a nonexistent slug (per @copilot-pull-request-reviewer, @cursor)

### No Change Needed
- **public-groups.controller.ts:56**: Slug uniqueness is enforced at the application layer (PCC), not at the query-service level. Design decision documented in the backend PR #172 review.
- **public-groups.controller.ts:52-56**: Controller tests are a valid suggestion but deferred to a follow-up — analogous public controllers in this repo also lack Vitest specs currently.

### Threads Resolved

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -47,25 +48,42 @@ export class PublicGroupsController {
       const m2mToken = await generateM2MToken(req);
       req.bearerToken = m2mToken;
 
-      const committee = await this.committeeService.getCommitteeById(req, id);
+      let committeeUid = id;
+      if (!UUID_REGEX.test(id)) {
+        const { resources } = await this.microserviceProxy.proxyRequest<QueryServiceResponse<Committee>>(req, 'LFX_V2_SERVICE', '/query/resources', 'GET', {
+          type: 'committee',
+          tags: `sso_group_name:${id}`,
```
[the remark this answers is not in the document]
MRashad26: This is an intentional design decision, not a field conflation. Per product direction: `sso_group_name` serves as both the SSO group identifier and the public URL slug — both flows share the same field.

The committee-service generates the `sso_group_name` when `Public || SSOGroupEnabled` is true and emits `sso_group_name:<value>` tags (see `committee_base.go:163`). This PR correctly queries that tag for slug resolution.

The duplicate `sso_group_name` declaration in the `Committee` interface was already fixed in commit 012de66 — the original `sso_group_name` at line 366 is the canonical field, and the redundant one from the `public_name` rename was removed.

No action needed.

MRashad26:
## Review Feedback Addressed

### No Change Needed
- **public-groups.controller.ts:55**: The `sso_group_name` tag query is correct — the committee-service emits `sso_group_name:<value>` tags and this field serves as both SSO identifier and URL slug by design. The duplicate interface declaration was already fixed in commit 012de66.

### Threads Resolved
1 of 1 unresolved thread addressed.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -50,25 +50,42 @@ export class PublicGroupsController {
...
```
audigregorie: **Minor — slug lookup is case-sensitive; normalize before the tag query**

The query service renders `tags` as OpenSearch `term` queries on a keyword field (exact, case-sensitive match), while committee-service generates `sso_group_name` via `slug.Make(...)` which always lowercases. A visitor hitting `/groups/My-Slug` (capitalized) gets a 404 even though the group exists. Normalize before the lookup:

```suggestion
          tags: `sso_group_name:${id.toLowerCase()}`,
```

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -50,25 +50,42 @@ export class PublicGroupsController {
...
+        });
+        if (resources.length === 0) {
+          throw new ResourceNotFoundError('Group', id, {
+            operation: 'get_public_group_by_id',
+            service: 'public_groups_controller',
+            path: `/groups/${id}`,
+          });
+        }
+        committeeUid = resources[0].data.uid;
+      }
```
audigregorie, line 69: **Minor — consider extracting slug resolution into `CommitteeService`**

Per the backend checklist's controller/service separation, this lookup is business logic living in the controller. A `committeeService.getCommitteeUidBySlug(req, slug): Promise<string | null>` would keep the controller on HTTP orchestration and make the slug lookup reusable (e.g. slug support for `/public/api/committees/:id/calendar.ics`, or a canonical redirect later). Non-blocking — this controller already makes direct query-service calls for mailing lists, so the block matches the file's existing pattern.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
```diff
@@ -50,25 +50,42 @@ export class PublicGroupsController {
...
+          page_size: 1,
+        });
+        if (resources.length === 0) {
+          throw new ResourceNotFoundError('Group', id, {
+            operation: 'get_public_group_by_id',
+            service: 'public_groups_controller',
+            path: `/groups/${id}`,
+          });
+        }
+        committeeUid = resources[0].data.uid;
```
audigregorie, line 67: **Minor — no trace log when a slug resolves**

When `:id` is a slug, the operation metadata still reads `group_uid: <slug>` and nothing records the slug → UID resolution, which will make production "why did this URL 404?" investigations harder (backend checklist §6 — debug-level tracing for internal lookups). Consider:

```suggestion
        committeeUid = resources[0].data.uid;
        logger.debug(req, 'get_public_group_by_id', 'Resolved group slug to UID', { slug: id, group_uid: committeeUid });
```

packages/shared/src/interfaces/public-group.interface.ts
```diff
@@ -79,6 +79,7 @@ export interface PublicGroupDirectoryVm extends PublicGroupSummary {
 export interface PublicGroupDetail {
   uid: string;
   name: string;
+  sso_group_name?: string;
```
audigregorie, line 82: **Minor — `sso_group_name` on `PublicGroupDetail` has no consumer yet**

Nothing in the app reads `detail.sso_group_name` — `group-detail.component.ts` never canonicalizes the URL, though the PR body cites canonical URLs as the intent. Either wire a browser-gated `location.replaceState` to the slug in `GroupDetailComponent`, or drop the field until the follow-up lands to keep the public API surface minimal. If a follow-up ticket exists, a one-line doc comment pointing at it would help.

audigregorie (commented):
## Code Review Summary

Slug-based public group URLs built on the merged committee-service slug generation (lfx-v2-committee-service#172). I verified the upstream contract end-to-end: `sso_group_name:{value}` is an indexed committee tag, the query service matches `tags` via exact OpenSearch `term` queries (no query-string injection vector), and slug uniqueness is enforced upstream with reservation retries — so the `tags` + `page_size: 1` lookup pattern is safe. No Critical or Major findings; four Minor items are posted as resolvable threads.

### Outside the diff

- **Stale PR description** — the body describes adding a `public_name` field to `Committee` and `PublicGroupDetail`, but the implementation pivoted to the pre-existing `sso_group_name` (no `Committee` interface change in this diff), the test plan references "public-name-slug", and the backend dependency (#172) is already merged. Worth updating the summary/test plan so QA verifies slug resolution, not `public_name`.
- **Slugs change on rename** — committee-service regenerates `sso_group_name` when a public committee's name changes (`mergeCommitteeData` → `checkReserveSSOName`), so previously shared slug links will 404 after a rename; no alias or redirect is kept. If stable shareable links matter, consider a follow-up — the new `sso_group_name` response field already enables a UID→slug canonical redirect on the detail page.
- **Two link styles in the committee table** — the row name link still navigates to `/groups/<uid>` while the copy button now shares `/groups/<slug>`. UID navigation is arguably the right internal choice (stable across renames), but flagging in case consistent slug URLs were intended everywhere.
- **Branch hygiene** — the branch carries 3 merge commits from `main`; the repo's pre-PR flow expects a rebase onto `main` before merge.

### What's done well

- Upstream contract alignment is exact: tag name, read path (`/query/resources`), `page_size` (not `limit`), and `sso_group_name` is confirmed present in the committee-service HTTP response schema, so `committee.sso_group_name` is actually populated.
- Correct token discipline for a public endpoint — M2M for upstream calls, user bearer restored for the membership check, then switched back.
- The `committeeUid` rename is applied consistently across error paths, logs, and the calendar ICS link; the copy handler is SSR-guarded with graceful clipboard fallbacks and user-facing toasts.

**Rating: 4/5** — good to merge after the Minor threads are considered.

luismoriguerra (dismissed):
## Audit: PR #1318 — feat(groups): add public_name slug resolution for public group URLs
MRashad26 · `feat/LFXV2-2012` → `main` · open · 4 files, +32/−12

**Verdict:** ✅ Approved  
**Merge readiness:** merge-after-confirming open minor threads + rebase (`mergeable_state` is blocked; base `867e3246` is behind current `main`)

**Gates:** CI Code Quality / DCO / CodeQL / build-and-push on `0ada02c7` — success. Full `yarn format:check && yarn lint:check && yarn build` not re-run locally (ephemeral clone, no `node_modules`).

**Lanes:** 6a secrets · 6b standards · 6c ADRs · 6d privacy · 6e skipped (UI/BFF only)

### Findings
1. **Non-blocking** · `public-groups.controller.ts:57` · Slug lookup is case-sensitive; lowercase `:id` before the tag query.
2. **Non-blocking** · `public-groups.controller.ts:57` · Unsanitized `:id` in `tags` — comma is OR in query-service; allowlist `[a-z0-9-]`.
3. **Non-blocking** · `public-group.interface.ts:49` vs `:82` · Directory `slug` is never set; detail uses `sso_group_name`. Align the public contract.
4. **Non-blocking** · No controller tests for UUID / slug / 404 / private 403.
5. **Nit** · PR title/body still say `public_name`; code uses `sso_group_name`.
6. **Nit** · Table name link is `/groups/<uid>` while copy uses the slug.
7. **Nit** · No slug→UID debug log; lookup lives in the controller.

No blocking findings, so this is the summary only — 4 non-blocking findings and 3 nits were not opened as inline threads (several already exist from @audigregorie). Say `all` if you want them anchored.

Prior review: prefixed index `id` and 403-for-missing-slug are fixed. `committee.public` still gates the payload after slug resolve.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Done — the slug is now lowercased before the tag query (`sso_group_name:${id.toLowerCase()}`), since `slug.Make()` in committee-service always lowercases generated slugs while query-service tag matching is case-sensitive.

See commit 01ee847c2.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Done — added `logger.debug(req, 'get_public_group_by_id', 'Resolved group slug to UID', { slug: id, group_uid: committeeUid })` after the slug resolves, matching the DEBUG-level internal-lookup convention in `.claude/rules/logging-patterns.md`.

See commit 01ee847c2.

apps/lfx-one/src/server/controllers/public-groups.controller.ts
MRashad26: Agreed this is the cleaner shape long-term, but leaving as-is for now — the controller already makes an equivalent direct query-service call inline for mailing lists in this same file, so this isn't yet a clean-vs-dirty deviation, and you flagged it non-blocking. Happy to do the `CommitteeService.getCommitteeUidBySlug()` extraction as a follow-up if/when the mailing-list lookup gets the same treatment, to avoid doing it piecemeal.

No code change in this iteration.

packages/shared/src/interfaces/public-group.interface.ts
MRashad26: Good catch — confirmed there are zero frontend consumers of `sso_group_name` on `PublicGroupDetail` today (checked `group-detail.component.ts` and `group.service.ts`). Leaving the field as-is for now rather than wiring up `location.replaceState` canonicalization, since that's new frontend behavior beyond this PR's scope rather than a review fix. No code change in this iteration — worth a follow-up ticket if we want URL canonicalization.

MRashad26:
## Review Feedback Addressed

Commit: 01ee847c2

### Changes Made
- **public-groups.controller.ts**: lowercased the slug before the tag query (`sso_group_name:${id.toLowerCase()}`) — `slug.Make()` in committee-service always lowercases, but query-service tag matching is case-sensitive (per @audigregorie)
- **public-groups.controller.ts**: added a debug trace log when a slug resolves to a committee UID, matching the DEBUG-level internal-lookup convention (per @audigregorie)

### No Change Needed
- **public-groups.controller.ts:68**: extracting slug resolution into `CommitteeService.getCommitteeUidBySlug()` — agreed as the cleaner long-term shape, but the controller already makes an equivalent inline query-service call for mailing lists in this file, so it's not a clean-vs-dirty deviation yet; reviewer flagged this non-blocking (flagged by @audigregorie)
- **public-group.interface.ts:82**: `PublicGroupDetail.sso_group_name` confirmed to have zero frontend consumers today. Wiring up `location.replaceState` canonicalization is new frontend behavior beyond this PR's scope, so left as-is pending a possible follow-up ticket (flagged by @audigregorie)

### Threads Resolved
2 of 4 unresolved threads addressed and resolved in this iteration.

### Still Open
- **public-groups.controller.ts:68**: controller/service extraction — deferred as non-blocking, consistent with existing in-file pattern
- **public-group.interface.ts:82**: unused `sso_group_name` field — deferred pending a decision on URL canonicalization scope

audigregorie (approved):

PR #1328, opened 2026-08-05
feat(committees): add pending invitations tab to group members list

andrest50:
## Ticket

[LFXV2-2841](https://linuxfoundation.atlassian.net/browse/LFXV2-2841)

## Summary

- Replaces the below-fold "Pending Invitations" card with a **"Pending" tab** inside the committee members card, making pending invites immediately discoverable without scrolling
- Invite rows now use the **same table column layout** as regular member rows (Name, Email, Organization, Role, Voting Status, Meetings, Engagement, Permission, Actions)
- A **"Pending invitation" tag** appears inline after the invitee email in the Name column
- The **row actions dropdown** for invite rows contains only "Revoke invitation"
- Pending invite rows appear in **both the "All" and "Pending" tabs**; the Pending tab is hidden when there are no pending invites
- `activeTab` uses `linkedSignal` to auto-reset to "All" when all invites are revoked (matching the existing `memberFilterChip` pattern)
- Filter chips and search/filter controls are hidden on the Pending tab (they apply to member rows only)

## Known pre-existing issues (out of scope)

Two patterns flagged by the learnings reviewer were pre-existing in the unchanged code paths that this PR restructured (not introduced):

- `[href]="organization.website"` — external URL bound without scheme validation (pre-dates this PR; in the original member org link)
- `[class]` + static `class=` co-existence on the filter chip buttons — pre-existing Tailwind class clobber; separate fix needed

## Trade-offs

- **Removed table-level `sortField`** — PrimeNG table sorting on `first_name` was dropped because invite rows have no `first_name` field, and mixing the two in the same `[value]` array would produce undefined sort behavior. Members arrive pre-sorted from the API; invites are appended at the end of the "All" tab naturally.

[LFXV2-2841]: https://linuxfoundation.atlassian.net/browse/LFXV2-2841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
@@ -6,7 +6,7 @@
   <p-confirmDialog></p-confirmDialog>
 
   <!-- Committee Members Section -->
-  <lfx-card [styleClass]="showPendingTab() ? '[&_.p-card-body]:p-0 [&_.p-card-content]:p-0' : ''">
+  <lfx-card [styleClass]="showPendingInvites() ? '[&_.p-card-body]:p-0 [&_.p-card-content]:p-0' : ''">
     @if (!showMembersSection()) {
       <div class="text-sm text-gray-500 text-center py-6" data-testid="members-hidden-placeholder">Member list is not publicly visible for this group.</div>
     } @else if (!showMemberRoster() && canSendMemberInvites()) {
@@ -22,12 +22,12 @@
         </lfx-button>
       </div>
     } @else {
-      @if (showPendingTab()) {
+      @if (showPendingInvites()) {
         <lfx-card-tabs-bar [options]="tabOptions()" [selectedFilter]="activeTab()" (filterChange)="onTabChange($event)" data-testid="members-tab-bar">
         </lfx-card-tabs-bar>
       }
 
-      <div [class]="showPendingTab() ? 'p-4 md:p-6' : ''">
+      <div [class]="showPendingInvites() ? 'p-4 md:p-6' : ''">
         @if (activeTab() === 'all') {
           <!-- Quick Filter Chips + engagement window controls -->
           <div class="flex flex-wrap items-center justify-between gap-2 mb-4">
@@ -185,7 +185,6 @@
                   </span>
                 } @else {
                   <div class="flex flex-wrap items-center gap-2">
-                    <span class="text-sm font-inter text-gray-900 font-medium">{{ row.data.invitee_email }}</span>
                     <lfx-tag value="Pending invitation" severity="warn" />
                   </div>
                 }
@@ -347,7 +346,7 @@
                 <td [attr.colspan]="emptyStateColspan()" class="text-center py-8">
                   <div class="text-center">
                     <i class="fa-light fa-users text-3xl text-gray-300 mb-2"></i>
-                    <p class="text-sm font-medium text-gray-600">No members found</p>
+                    <p class="text-sm font-medium text-gray-600">{{ activeTab() === 'pending' ? 'No pending invitations' : 'No members found' }}</p>
                     @if (canManageMembers() && activeTab() !== 'pending') {
                       <div class="mt-3">
                         <button class="text-xs text-blue-600 hover:text-blue-700 font-medium" (click)="openAddMemberDialog()">
--- a/apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
@@ -311,6 +311,7 @@
                   [rounded]="true"
                   size="small"
                   severity="secondary"
+                  ariaLabel="Invite actions"
                   [attr.data-testid]="'invite-actions-btn-' + row.data.uid"
                   (onClick)="toggleInviteActionMenu($event, row.data, inviteActionMenu)">
                 </lfx-button>
--- a/apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
@@ -23,11 +23,7 @@
       </div>
     } @else {
       @if (showPendingTab()) {
-        <lfx-card-tabs-bar
-          [options]="tabOptions()"
-          [selectedFilter]="activeTab()"
-          (filterChange)="onTabChange($event)"
-          data-testid="members-tab-bar">
+        <lfx-card-tabs-bar [options]="tabOptions()" [selectedFilter]="activeTab()" (filterChange)="onTabChange($event)" data-testid="members-tab-bar">
         </lfx-card-tabs-bar>
       }
 
@@ -73,7 +69,14 @@
 
             <!-- Role Filter -->
             <div class="w-full sm:w-auto sm:shrink-0">
-              <lfx-select [form]="filterForm" control="role" size="small" [options]="roleOptions()" placeholder="All Roles" [showClear]="true" styleClass="w-full">
+              <lfx-select
+                [form]="filterForm"
+                control="role"
+                size="small"
+                [options]="roleOptions()"
+                placeholder="All Roles"
+                [showClear]="true"
+                styleClass="w-full">
               </lfx-select>
             </div>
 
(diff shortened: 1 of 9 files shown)


apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -22,329 +22,348 @@
...
+                      rel="noopener noreferrer"
+                      class="text-primary hover:text-primary-600 text-sm font-inter">
+                      {{ row.data.organization.name }}
+                    </a>
                   } @else {
-                    <span class="text-sm font-inter text-gray-400" [attr.data-testid]="'members-engagement-chip-' + member.uid">—</span>
+                    <span class="text-sm font-inter text-gray-500">{{ row.data.organization?.name || '-' }}</span>
                   }
+                } @else {
+                  <span class="text-sm font-inter text-gray-500">{{ row.data.organization || '—' }}</span>
```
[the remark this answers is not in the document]
andrest50: Fixed — changed `row.data.organization` to `row.data.organization?.name` so the org name string is rendered instead of the raw `CommitteeOrganizationReference` object.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -22,329 +22,348 @@
...
+                    @if (canManageMembers()) {
+                      <div class="mt-3">
+                        <button class="text-xs text-blue-600 hover:text-blue-700 font-medium" (click)="openAddMemberDialog()">
+                          <i class="fa-light fa-user-plus mr-1"></i>Add the first member
+                        </button>
+                      </div>
+                    }
+                  </div>
+                </td>
+              </tr>
```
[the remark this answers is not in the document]
andrest50: Fixed — added `|| (activeTab() === 'pending' && invitesLoading())` to the empty-state skeleton guard, and hid the 'Add the first member' CTA when on the Pending tab, so the loading skeleton and empty state are both correct for each tab context.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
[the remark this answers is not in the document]
andrest50: Fixed — changed `row.data.organization` to `row.data.organization?.name` so the name string is rendered instead of the raw object.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -22,329 +22,348 @@
...
-        @if (invitesLoading()) {
-          <div class="flex flex-col gap-2">
-            @for (_ of [1, 2]; track _) {
-              <div class="flex items-center gap-3">
-                <p-skeleton width="14rem" height="0.875rem" />
-                <p-skeleton width="6rem" height="0.875rem" />
-              </div>
+          <!-- Empty Message Template -->
+          <ng-template #emptymessage>
+            @if (membersLoading()) {
```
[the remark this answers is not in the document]
andrest50: Fixed — the empty-state skeleton guard now also checks `activeTab() === 'pending' && invitesLoading()`, and the 'Add the first member' CTA is hidden on the Pending tab.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -22,329 +22,348 @@
...
-        @if (engagementEnabled() && engagementAccessible()) {
-          <div class="flex flex-wrap items-center gap-3" data-testid="members-engagement-controls">
-            @if (isMockEngagement()) {
-              <lfx-tag value="Sample data" severity="warn" data-testid="members-engagement-mock-tag" />
-            }
-            @if (engagementDataAvailable()) {
-              <span class="text-xs text-gray-400" data-testid="members-engagement-freshness">{{ engagementFreshnessLabel() }}</span>
+      @if (showPendingTab()) {
+        <lfx-card-tabs-bar [options]="tabOptions()" [selectedFilter]="activeTab()" (filterChange)="onTabChange($event)" data-testid="members-tab-bar">
+        </lfx-card-tabs-bar>
```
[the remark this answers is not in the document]
andrest50: Out of scope for this PR — E2E spec work wasn't part of LFXV2-2841's scope. The `data-testid` attributes (`members-tab-bar`, `members-filter-chips`, etc.) are in place for future test authoring.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -22,329 +22,348 @@
...
-              <td>
-                @if (engagementLoading()) {
-                  <p-skeleton width="2.5rem" height="0.875rem" />
+                @if (row.rowType === 'member') {
+                  <span class="text-sm font-inter text-gray-500">{{ row.data.email || '-' }}</span>
                 } @else {
-                  <span class="text-sm font-inter text-gray-900" [attr.data-testid]="'members-engagement-meetings-' + member.uid">
-                    {{ engagementView?.meetingsLabel }}
-                  </span>
+                  <span class="text-sm font-inter text-gray-500">{{ row.data.invitee_email }}</span>
```
audigregorie, line 197: **Minor — Invite row renders the same email in two columns**

For invite rows, the Name column (line 188) shows `{{ row.data.invitee_email }}` plus a "Pending invitation" tag, and this Email column shows `{{ row.data.invitee_email }}` again. The same address appears twice on one row, which is redundant.

Consider showing only the tag (or a `—` placeholder) in the Name column for invites, since the email is already surfaced in the Email column.

```suggestion
                  <span class="text-sm font-inter text-gray-400">—</span>
```

audigregorie (commented):
## Code Review Summary

Clean, convention-following implementation — the discriminated-union row model and `linkedSignal` tab reset are well done. The two Major findings below are both **test-coverage gaps** (one a regression in an existing test's intent, one missing coverage for new behavior) rather than code defects, but they should be addressed before merge.

### Major — outside the diff

**1. Existing e2e test silently invalidated by testid removal** — `apps/lfx-one/e2e/committee-members-join-mode.spec.ts:37` asserts `page.getByTestId('pending-invites')` has count 0 in closed mode. This PR deletes the old Pending Invitations card that carried `data-testid="pending-invites"`, so the testid no longer exists in *any* mode. The test still passes but now tests nothing — it would pass even if pending invites were incorrectly shown in closed mode. Update it to assert `members-tab-bar` count 0 (or `invite-row-*` count 0) in closed mode so the closed-mode guard stays meaningful.

**2. No tests for the new Pending tab feature** — No unit or e2e tests cover: the Pending tab appearing when invites exist, invite rows rendering in the shared table, the `linkedSignal` auto-reset to All when the last invite is revoked, or the revoke-via-dropdown flow. The new testids (`members-tab-bar`, `invite-row-`, `invite-actions-btn-`) have no assertions against them.

### What's done well

- Clean `CommitteeTableRow` discriminated union with a `rowType` discriminant — avoids nested ternaries and keeps `@if` template narrowing type-safe.
- `linkedSignal` for `activeTab` correctly auto-resets to "All" when the Pending tab becomes invisible, matching the established `memberFilterChip` pattern.
- `onTabChange` validates the incoming string against the `CommitteeMemberTab` union before `.set()` — defensive against unknown filter ids.
- Good reuse of the existing `CardTabsBarComponent` wrapper and the documented `styleClass` padding-zero pattern.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
```diff
@@ -22,329 +22,348 @@
...
+                    }
+                  </div>
+                </td>
+              </tr>
+            } @else {
+              <tr>
+                <td [attr.colspan]="emptyStateColspan()" class="text-center py-8">
+                  <div class="text-center">
+                    <i class="fa-light fa-users text-3xl text-gray-300 mb-2"></i>
+                    <p class="text-sm font-medium text-gray-600">No members found</p>
```
audigregorie: **Minor — Empty-state message not tab-aware**

The `#emptymessages` template always says "No members found" even when the active tab is "pending". The `linkedSignal` makes the Pending tab disappear when empty, so this is mostly unreachable — but a tab-aware message would be more robust against any future timing change that leaves the user on an empty Pending tab.

```suggestion
                    <p class="text-sm font-medium text-gray-600">{{ activeTab() === 'pending' ? 'No pending invitations' : 'No members found' }}</p>
```

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.ts
```diff
@@ -156,6 +174,16 @@ export class CommitteeMembersComponent implements OnInit {
   public readonly showPendingInvites = computed(
     () => this.canManageMembers() && this.joinMode() !== 'closed' && (this.invitesLoading() || this.invites().length > 0)
   );
+  /** Show the Pending tab only when there are (or are loading) pending invites a manager can see. */
+  public readonly showPendingTab = computed(() => this.showPendingInvites());
```
audigregorie: **Minor — `showPendingTab` is a trivial alias**

`showPendingTab = computed(() => this.showPendingInvites())` adds an indirection with no transformation. If the semantic distinction between "pending invites exist" and "pending tab is visible" is intentional for future divergence, a one-line comment would help; otherwise inlining `showPendingInvites()` (and dropping `showPendingTab`) would reduce the surface area.

```suggestion
  public readonly tabOptions = computed<FilterPillOption[]>(() => {
    const opts: FilterPillOption[] = [{ id: 'all', label: 'All' }];
    if (this.showPendingInvites()) {
      const count = this.invitesLoading() ? '' : ` (${this.invites().length})`;
      opts.push({ id: 'pending', label: `Pending${count}` });
    }
    return opts;
  });
```

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.ts
```diff
@@ -911,4 +954,35 @@ export class CommitteeMembersComponent implements OnInit {
...
+      if (this.activeTab() === 'pending') {
+        return inviteRows;
+      }
+      const memberRows: CommitteeTableRow[] = this.filteredMembers().map((m) => ({ rowType: 'member' as const, data: m }));
+      if (!this.canManageMembers() || this.joinMode() === 'closed') {
+        return memberRows;
+      }
+      return [...memberRows, ...inviteRows];
+    });
+  }
```
audigregorie, line 968: **Minor — `initTableRows` returns invite rows on the pending tab even when `showPendingTab` is false**

The `activeTab === 'pending'` branch returns `inviteRows` unconditionally. In practice the `linkedSignal` resets `activeTab` to `'all'` whenever `showPendingTab` goes false, so this branch is unreachable with stale invites — but defensively gating it on `showPendingTab()` (and falling through to the member-only return) would make the invariant explicit and survive any future change to the `linkedSignal` reset timing.

```suggestion
  private initTableRows(): Signal<CommitteeTableRow[]> {
    return computed(() => {
      const memberRows: CommitteeTableRow[] = this.filteredMembers().map((m) => ({ rowType: 'member' as const, data: m }));
      if (!this.canManageMembers() || this.joinMode() === 'closed') {
        return memberRows;
      }
      const inviteRows: CommitteeTableRow[] = this.invites().map((invite) => ({ rowType: 'invite' as const, data: invite }));
      if (this.activeTab() === 'pending' && this.showPendingTab()) {
        return inviteRows;
      }
      return [...memberRows, ...inviteRows];
    });
  }
```

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
andrest50: Fixed — removed the email span from the Name cell; the `Pending invitation` tag alone identifies the row, and the email is already in the Email column.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.html
andrest50: Fixed — the message is now `activeTab() === 'pending' ? 'No pending invitations' : 'No members found'` so it's accurate regardless of which tab is active.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.ts
andrest50: Fixed — dropped `showPendingTab` and inlined `showPendingInvites()` directly in `tabOptions`, the template guards, and the `activeTab` linkedSignal source.

apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.ts
andrest50: Fixed — the pending-tab branch in `initTableRows` is now guarded with `&& this.showPendingInvites()` making the invariant explicit. Also restructured to compute memberRows first so the canManageMembers/closed guard short-circuits before mapping invites.

audigregorie (approved):

PR #1330, opened 2026-08-05
fix(build): scope tailwind config to shared constants subpath

manishdixitlfx:
## Summary

- `apps/lfx-one/tailwind.config.js` imported from the `@lfx-one/shared` root barrel, which re-exports `form.utils.ts` (`@angular/forms`). Tailwind's config loader (jiti) evaluates that raw FESM bundle outside Angular's linker and crashes with a JIT-compiler error — which esbuild then misattributes to an unrelated sass error on the global styles entry, breaking `yarn start` on Node ≥22.12.
- Scope the import to `@lfx-one/shared/constants` instead, and stop `dashboard-metrics.constants.ts` from re-introducing the same leak via the `../utils` barrel.
- Add a regression-guard spec (`packages/shared/src/constants/index.spec.ts`) asserting both halves stay fixed: the constants barrel loads in plain Node, and `tailwind.config.js` only ever uses the `/constants` subpath specifier.

Fixes LFXV2-3008.

## Test plan

- [x] `yarn start` builds and serves cleanly on Node 22.22.3 (previously failed with a repeating `angular-sass` error)
- [x] `yarn test` — 32 files / 628 tests passing in `packages/shared`, 37 files / 578 tests passing in `apps/lfx-one`
- [x] `yarn build` — succeeds
- [x] `yarn lint` / `yarn format` — clean (0 errors)
- [x] Manually verified the regression guard: reverting either the `tailwind.config.js` import or the `dashboard-metrics.constants.ts` import reproduces the original failure and the new spec catches it


luismoriguerra (approved):
## Independent audit — ✅ Approved

**Merge readiness:** merge-after-rebase — `mergeable_state: behind`, base `main` has moved past `3a46048`.

**Gates re-run in an isolated worktree (Node 22.22.2), all green:** `yarn format:check` ✅ · `yarn lint:check` ✅ (0 errors) · `yarn test` ✅ 32 files/628 tests (shared) + 37/578 (lfx-one), exactly matching the PR body · `yarn build` ✅

**Guard verified non-vacuous** by reverting each half: reverting `dashboard-metrics.constants.ts` to the `../utils` barrel reproduces the real `@angular/compiler` JIT error on the barrel-load test; reverting the tailwind import fails the specifier assertion. Static scan of the constants barrel reaches 227 files with **zero `@angular` edges**, and no `from '../utils'` barrel import remains anywhere in `constants/` — the fix is complete, not partial.

**Findings** (all non-blocking; none require a change):

- **Non-blocking** — `packages/shared/src/constants/index.spec.ts:42`: the block-comment strip isn't comment-aware and already mangles the `/**/` inside `content: ['./src/**/*.{html,ts}']`. Harmless only because the import sits above `content`; a future glob with an odd `/*` count could shift the non-greedy pairing and swallow the import line. Fix: drop the block-comment pass, or anchor on `(from|import|require)\s*\(?\s*`.
- **Non-blocking** — the barrel-load guard relies on Vitest per-file isolation (`meeting.utils.spec.ts:7` loads `@angular/compiler`); `--no-isolate` would silently make it vacuous. Worth a line in the existing comment block.
- **Nit** — `packages/shared/tsconfig.json` excludes `**/*.spec.ts` and the package has no lint config, so this spec is neither type-checked nor linted — notable because it has real logic (`node:fs`, `__dirname`, regex).
- **Nit** — PR body: the raw FESM bundle is `@angular/forms`', not `@lfx-one/shared`'s (plain `tsc` ESM output). Mechanism right, noun misplaced.

No secrets or critical-constant drift. Privacy/ADR/platform-contract lanes skipped — the diff adds no response field, read/write path, or NATS/OpenSearch/FGA surface.

Findings are summarized here only; no inline threads were opened.

---

audigregorie (commented):
## Code Review Summary

Clean, minimal, well-guarded fix for a real build-breaking regression. No Critical, Major, or Minor code issues — only two process notes on commit hygiene (below). Verified the import graph and the regression guard empirically: the constants barrel reaches no Angular runtime code after the fix, and reverting either half of the fix fails the corresponding test.

### Outside the diff — process notes

**1. Missing `(LFXV2-3008)` ticket suffix on all 12 commit subjects.**

The repo's commit policy (`global-rules.mdc`) requires every commit subject to end with `(PCC-#)` or `(LFXV2-#)`, derived from the branch name. The branch is `fix/LFXV2-3008`, so each subject should end with `(LFXV2-3008)`. None of the 12 commits include it. Commitlint doesn't enforce this (it's a team style rule), but `/lfx-self-serve-pr-readiness` will flag it as a SHOULD_FIX. Easy fix on squash/rebase before merge.

**2. 11 of 12 commits are `fix(review):` comment-iteration noise.**

The actual code fix is a single commit (`802f69ca6`); the other 11 are comment-wording refinements on the spec and adjacent files. Consider squashing before merge so the history reflects the real change set (1 fix + 1 test), not the review loop. Not blocking — just hygiene.

### What's done well

- **Root-cause precision.** The fix correctly targets the actual failure: `tailwind.config.js` importing from the `@lfx-one/shared` root barrel, which re-exports `./utils` → `meeting.utils` (`@angular/common/http`) and `form.utils` (`@angular/forms`). Scoping to `@lfx-one/shared/constants` avoids the Angular runtime entirely. I verified the constants barrel reaches no Angular code (only `dashboard-metrics` → `color.utils`/`number.utils`, and `committees` → `committee.utils` → `date-time.utils`/`string.utils`, all Angular-free).
- **Defense-in-depth, not just a patch.** `dashboard-metrics.constants.ts` was also migrated from the `../utils` barrel to by-file imports. Without this, the constants barrel would still transitively reach Angular via `dashboard-metrics → ../utils → meeting.utils`, re-introducing the same crash. The PR closes both leak paths.
- **Non-vacuous regression guard.** `index.spec.ts` asserts both halves positively: (1) the constants barrel loads in plain Node, (2) every `@lfx-one/shared*` specifier in `tailwind.config.js` is exactly `@lfx-one/shared/constants`. I verified the guard is non-vacuous — reverting either fix fails the corresponding test. The comment-stripping regex correctly handles `//` in URL schemes (`https://`) via the `(^|[^:])` lookahead, and the `not.toHaveLength(0)` assertion prevents the test from passing vacuously if the import shape ever changes entirely.
- **Accurate, single-source-of-truth comments.** The `activity-feed.utils.ts` and `date-time.utils.ts` comment updates correctly distinguish the defensive case (no live cycle today) from the live cycle path (`constants/index → committees.constants → committee.utils → date-time.utils`), which I verified by reading the import graph. Both point to the spec as the canonical invariant doc.

**Rating: ⭐⭐⭐⭐⭐ (5/5)** — no code issues; only commit-hygiene notes.

PR #1334, opened 2026-08-05
refactor(enrollment): normalize frontend shape to camelCase

fayazg:
## Summary

- Rename `EnrollmentMembership` fields (`Status`, `AutoRenew`, `PurchaseDate`, `EndDate`, `Price`, `ID`, `ExtPaymentType`) and `IndividualEnrollment.ProductName` to camelCase, matching their camelCase siblings.
- Normalization happens at the server mapping boundary in `EnrollmentService.getIndividualEnrollments` — `RawMembership` (the member-service wire shape) and the auto-renew PATCH payload stay PascalCase.
- Pure rename, no behavior change; updated the shared interface, catalog constant, status-derivation util + spec, the profile component (TS + template), backend service, and both e2e specs.
- Added clarifying notes to the `individual-memberships` migration doc flagging the LFX One camelCase divergence from the legacy myprofile / member-service PascalCase examples.

LFXV2-1664


luismoriguerra (approved):
Approving — checked the rename for completeness on the branch.

- No stragglers: every `item.ProductName` and `membership.{Status,AutoRenew,PurchaseDate,EndDate,Price,ID,ExtPaymentType}` reference is gone from TS and the template. Remaining PascalCase is all `RawMembership` at the mapping boundary in `enrollment.service.ts`, which is correct.
- Auto-renew PATCH payload correctly stays PascalCase — that's the upstream contract, not the frontend shape.
- Both e2e specs updated including the `stripeActiveMembershipAutoRenewOff` spread; since the mocks stub the LFX One response (not member-service), camelCase is the right fixture shape.
- The migration-doc divergence notes are a good addition — keeps the doc from silently contradicting the code.

**Merge-order note:** conflicts with #1306 (same two files, and this branch doesn't include its `'Purchased'` tests). Suggest landing #1306 first, then rebasing here and confirming the three new tests get camelCased too. Currently `blocked` on checks.

---

PR #1337, opened 2026-08-06
feat(committees): prefill org from profile on join and apply flows

andrest50:
## Summary

- Extracts a public `resolveCurrentEmployer()` method on `InvitationAcceptFlowService` that fetches the user's work experiences, picks the current employer, and resolves its domain — mirroring `preResolveOrganization` for invite payloads
- Calls it from `committee-view` before opening the Confirm Organization dialog on both the open-join and apply-to-join paths so the dialog is pre-filled from the user's profile
- Tightens `isNewOrg` and `orgInvalid` in `AcceptInviteOrganizationDialogComponent`: `organization_url` is now always required because `organization_id` (CDP UUID) is stripped before the payload reaches committee-service, which requires name + domain

## Background

When joining an open committee or applying to a committee that requires an organization (voting enabled or business email required), the Confirm Organization dialog was blank. The invite flow already pre-filled from the invite record, but open-join and apply paths had no equivalent source. This caused a confusing UX and, for apply flows, meant the submitted org was missing — leading to `organization id or organization name and domain are required` errors when an admin approved the application.

The back-end counterpart (storing org on the application record and using it at approve time) is in [lfx-v2-committee-service#174](https://github.com/linuxfoundation/lfx-v2-committee-service/pull/174).

## Test plan

- [ ] Open committee: click Join — Confirm Organization dialog pre-fills with profile employer
- [ ] Apply-to-join committee with `enable_voting: true` — dialog pre-fills with profile employer
- [ ] Submit application — org is included in the payload
- [ ] User with no work experience — dialog opens empty (no pre-fill, no error)
- [ ] Invite flow — still works as before (no regression)

Resolves: [LFXV2-2690](https://linuxfoundation.atlassian.net/browse/LFXV2-2690)

[LFXV2-2690]: https://linuxfoundation.atlassian.net/browse/LFXV2-2690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -200,7 +200,7 @@ export class CommitteeViewComponent {
   public joiningOrLeaving = signal(false);
   // Blocks the join/apply CTA during the async org-prefetch window so a second tap
   // doesn't fire a parallel resolveCurrentEmployer() + dialog pair.
-  private readonly resolvingOrg = signal(false);
+  protected readonly resolvingOrg = signal(false);
   // Engagement rollup (LFXV2-1705): shared window state so the Members table and the Overview
   // summary stay in sync across tab switches (tab panels unmount in the @switch below).
   // linkedSignal, not signal: resets to the default window whenever committeeId() changes, the same
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -156,6 +156,9 @@ export class CommitteeViewComponent {
   public invitesLoading = signal<boolean>(true);
   public applicationsLoading = signal<boolean>(true);
   public joiningOrLeaving = signal(false);
+  // Blocks the join/apply CTA during the async org-prefetch window so a second tap
+  // doesn't fire a parallel resolveCurrentEmployer() + dialog pair.
+  private readonly resolvingOrg = signal(false);
   // Engagement rollup (LFXV2-1705): shared window state so the Members table and the Overview
   // summary stay in sync across tab switches (tab panels unmount in the @switch below).
   // linkedSignal, not signal: resets to the default window whenever committeeId() changes, the same
@@ -482,7 +485,7 @@ export class CommitteeViewComponent {
 
   public async handleJoinRequest(): Promise<void> {
     const committee = this.committee();
-    if (!committee || this.joiningOrLeaving()) {
+    if (!committee || this.joiningOrLeaving() || this.resolvingOrg()) {
       return;
     }
 
@@ -492,7 +495,8 @@ export class CommitteeViewComponent {
     if (joinMode === 'open') {
       let organization: CommitteeOrganizationReference | undefined;
       if (requiresOrg) {
-        const result = await this.openOrganizationDialog(committee.name);
+        this.resolvingOrg.set(true);
+        const result = await this.openOrganizationDialog(committee.name).finally(() => this.resolvingOrg.set(false));
         if (!result?.organization) {
           return;
         }
@@ -518,7 +522,8 @@ export class CommitteeViewComponent {
       }
       let organization: CommitteeOrganizationReference | undefined;
       if (requiresOrg) {
-        const result = await this.openOrganizationDialog(committee.name);
+        this.resolvingOrg.set(true);
+        const result = await this.openOrganizationDialog(committee.name).finally(() => this.resolvingOrg.set(false));
         if (!result?.organization) {
           return;
         }
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -198,8 +198,9 @@ export class CommitteeViewComponent {
   public invitesLoading = signal<boolean>(true);
   public applicationsLoading = signal<boolean>(true);
   public joiningOrLeaving = signal(false);
-  // Blocks the join/apply CTA during the async org-prefetch window so a second tap
-  // doesn't fire a parallel resolveCurrentEmployer() + dialog pair.
+  // Blocks the join/apply CTA while the org-prefetch is running and while the dialog
+  // is open (signal stays true until .finally() resolves when the dialog closes) so a
+  // second tap can't open a parallel resolveCurrentEmployer() + dialog pair.
   protected readonly resolvingOrg = signal(false);
   // Engagement rollup (LFXV2-1705): shared window state so the Members table and the Overview
   // summary stay in sync across tab switches (tab panels unmount in the @switch below).
@@ -837,7 +838,8 @@ export class CommitteeViewComponent {
     // Pre-fill from the user's profile work experiences — same pre-resolution the invite
     // accept flow uses so the user doesn't have to re-enter an org they've already set.
     // takeUntilDestroyed cancels the observable if the component is destroyed while the
-    // profile/domain lookup (≤2 s) is still in flight, preventing dangling subscriptions.
+    // profile/domain lookup (≤4 s worst case — 2 s for the work-experiences GET + 2 s for
+    // the CDP domain resolve) is still in flight, preventing dangling subscriptions.
     // We track destruction explicitly because firstValueFrom's defaultValue: null causes
     // the await to resolve successfully even when takeUntilDestroyed completed early due
     // to component teardown — without this guard the dialog would open over a new route.
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -62,7 +62,7 @@ import { InvitationSubtextPipe } from '@pipes/invitation-subtext.pipe';
 import { JoinModeLabelPipe } from '@pipes/join-mode-label.pipe';
 import { DescriptionDialogComponent } from '../components/description-dialog/description-dialog.component';
 import { MessageService } from 'primeng/api';
-import { catchError, combineLatest, distinctUntilChanged, EMPTY, exhaustMap, filter, finalize, map, of, startWith, switchMap, take, tap, timer } from 'rxjs';
+import { catchError, combineLatest, distinctUntilChanged, EMPTY, exhaustMap, filter, finalize, firstValueFrom, map, of, startWith, switchMap, take, tap, timer } from 'rxjs';
 import { getHttpErrorDetail } from '@shared/utils/http-error.utils';
 import { syncEntityProjectContext } from '@shared/utils/entity-project-context.util';
 import { JoinApplicationDialogResult } from '@lfx-one/shared/interfaces';
@@ -780,13 +780,17 @@ export class CommitteeViewComponent {
     });
   }
 
-  private openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+  private async openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+    // Pre-fill from the user's profile work experiences — same pre-resolution the invite
+    // accept flow uses so the user doesn't have to re-enter an org they've already set.
+    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer());
+
     const ref = this.dialogService.open(AcceptInviteOrganizationDialogComponent, {
       header: 'Confirm Organization',
       width: '32rem',
       modal: true,
       closable: true,
-      data: { committeeName, organization: null } satisfies AcceptInviteOrganizationDialogData,
+      data: { committeeName, organization: prefillOrg } satisfies AcceptInviteOrganizationDialogData,
     });
     if (!ref) {
       return Promise.resolve(null);
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -62,7 +62,23 @@ import { InvitationSubtextPipe } from '@pipes/invitation-subtext.pipe';
 import { JoinModeLabelPipe } from '@pipes/join-mode-label.pipe';
 import { DescriptionDialogComponent } from '../components/description-dialog/description-dialog.component';
 import { MessageService } from 'primeng/api';
-import { catchError, combineLatest, distinctUntilChanged, EMPTY, exhaustMap, filter, finalize, firstValueFrom, map, of, startWith, switchMap, take, tap, timer } from 'rxjs';
+import {
+  catchError,
+  combineLatest,
+  distinctUntilChanged,
+  EMPTY,
+  exhaustMap,
+  filter,
+  finalize,
+  firstValueFrom,
+  map,
+  of,
+  startWith,
+  switchMap,
+  take,
+  tap,
+  timer,
+} from 'rxjs';
 import { getHttpErrorDetail } from '@shared/utils/http-error.utils';
 import { syncEntityProjectContext } from '@shared/utils/entity-project-context.util';
 import { JoinApplicationDialogResult } from '@lfx-one/shared/interfaces';
@@ -788,7 +804,11 @@ export class CommitteeViewComponent {
   private async openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
     // Pre-fill from the user's profile work experiences — same pre-resolution the invite
     // accept flow uses so the user doesn't have to re-enter an org they've already set.
-    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer());
+    // takeUntilDestroyed cancels the observable if the component is destroyed while the
+    // profile/domain lookup (≤2 s) is still in flight, preventing dangling subscriptions.
+    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer().pipe(takeUntilDestroyed(this.destroyRef)), {
+      defaultValue: null,
+    });
 
     const ref = this.dialogService.open(AcceptInviteOrganizationDialogComponent, {
       header: 'Confirm Organization',
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -542,6 +542,9 @@ export class CommitteeViewComponent {
         if (!result?.organization) {
           return;
         }
+        if (this.committeeId() !== committee.uid) {
+          return;
+        }
         organization = result.organization;
       }
       this.joiningOrLeaving.set(true);
@@ -567,6 +570,9 @@ export class CommitteeViewComponent {
         this.resolvingOrg.set(true);
         const result = await this.openOrganizationDialog(committee.name).finally(() => this.resolvingOrg.set(false));
         if (!result?.organization) {
+          return;
+        }
+        if (this.committeeId() !== committee.uid) {
           return;
         }
         organization = result.organization;
--- a/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
@@ -806,9 +806,22 @@ export class CommitteeViewComponent {
     // accept flow uses so the user doesn't have to re-enter an org they've already set.
     // takeUntilDestroyed cancels the observable if the component is destroyed while the
     // profile/domain lookup (≤2 s) is still in flight, preventing dangling subscriptions.
+    // We track destruction explicitly because firstValueFrom's defaultValue: null causes
+    // the await to resolve successfully even when takeUntilDestroyed completed early due
+    // to component teardown — without this guard the dialog would open over a new route.
+    let destroyed = false;
+    const cleanupDestroyListener = this.destroyRef.onDestroy(() => {
+      destroyed = true;
+    });
+
     const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer().pipe(takeUntilDestroyed(this.destroyRef)), {
       defaultValue: null,
     });
+
+    cleanupDestroyListener();
+    if (destroyed) {
+      return null;
+    }
 
     const ref = this.dialogService.open(AcceptInviteOrganizationDialogComponent, {
       header: 'Confirm Organization',
--- a/apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
+++ b/apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
@@ -43,13 +43,12 @@ export class AcceptInviteOrganizationDialogComponent {
   private readonly formValue = this.initFormValue();
   private readonly urlStatus = signal<FormControlStatus>(this.urlControl.status);
 
+  // URL validators must stay active whenever an org name is present — not just while the
+  // URL field is empty — because clearing them on the first URL keystroke allowed any
+  // non-empty string (e.g. "not-a-url") to pass urlStatus as VALID and enable Confirm.
   protected readonly isNewOrg = computed(() => {
     const value = this.formValue();
-    // URL is always required because organization_id is stripped before the payload reaches
-    // committee-service — the server must receive either an SFID (which we don't have) or
-    // name + domain. Treat any org without a confirmed website as requiring URL input.
-    const hasUrl = !!(value?.organization_url ?? '').trim();
-    return !hasUrl && !!value?.organization?.trim();
+    return !!value?.organization?.trim();
   });
 
   private readonly orgInvalid = computed(() => {
--- a/apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
+++ b/apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
@@ -1,8 +1,8 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { Component, computed, effect, inject, signal, viewChild } from '@angular/core';
-import { takeUntilDestroyed, toSignal } from '@angular/core/rxjs-interop';
+import { Component, computed, inject, signal, viewChild } from '@angular/core';
+import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { FormControl, FormControlStatus, FormGroup, ReactiveFormsModule } from '@angular/forms';
 import { ButtonComponent } from '@components/button/button.component';
 import { OrganizationSearchComponent } from '@components/organization-search/organization-search.component';
@@ -101,17 +101,19 @@ export class AcceptInviteOrganizationDialogComponent {
       this.urlStatus.set(status);
     });
 
-    effect(() => {
-      if (this.isNewOrg()) {
-        this.urlControl.setValidators([trimmedRequired(), httpsUrlValidator()]);
-      } else {
-        this.urlControl.clearValidators();
-      }
-      this.urlControl.updateValueAndValidity({ emitEvent: false });
-      // Manually sync after updateValueAndValidity({ emitEvent: false }) since
-      // suppressing the event means statusChanges won't fire for validator changes.
-      this.urlStatus.set(this.urlControl.status);
-    });
+    toObservable(this.isNewOrg)
+      .pipe(takeUntilDestroyed())
+      .subscribe((isNew) => {
+        if (isNew) {
+          this.urlControl.setValidators([trimmedRequired(), httpsUrlValidator()]);
+        } else {
+          this.urlControl.clearValidators();
+        }
+        this.urlControl.updateValueAndValidity({ emitEvent: false });
+        // Manually sync after updateValueAndValidity({ emitEvent: false }) since
+        // suppressing the event means statusChanges won't fire for validator changes.
+        this.urlStatus.set(this.urlControl.status);
+      });
   }
 
   public onOrgResolved(result: OrganizationResolveResult): void {
--- a/apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
+++ b/apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
@@ -46,7 +46,10 @@ export class AcceptInviteOrganizationDialogComponent {
   // URL validators must stay active whenever an org name is present — not just while the
   // URL field is empty — because clearing them on the first URL keystroke allowed any
   // non-empty string (e.g. "not-a-url") to pass urlStatus as VALID and enable Confirm.
-  protected readonly isNewOrg = computed(() => {
+  // Named hasOrgName (not isNewOrg) because it is true for CDP-resolved orgs too —
+  // organization_id is stripped before forwarding to committee-service, so the URL
+  // is required regardless of whether the org already exists in CDP.
+  protected readonly hasOrgName = computed(() => {
     const value = this.formValue();
     return !!value?.organization?.trim();
   });
@@ -100,10 +103,12 @@ export class AcceptInviteOrganizationDialogComponent {
       this.urlStatus.set(status);
     });
 
-    toObservable(this.isNewOrg)
+    // toObservable (not effect) so we can write urlStatus inside the callback —
+    // effect() forbids signal writes without allowSignalWrites: true.
+    toObservable(this.hasOrgName)
       .pipe(takeUntilDestroyed())
-      .subscribe((isNew) => {
-        if (isNew) {
+      .subscribe((hasName) => {
+        if (hasName) {
           this.urlControl.setValidators([trimmedRequired(), httpsUrlValidator()]);
         } else {
           this.urlControl.clearValidators();
@@ -126,7 +131,7 @@ export class AcceptInviteOrganizationDialogComponent {
 
   public onConfirm(): void {
     this.organizationControl.markAsTouched();
-    if (this.isNewOrg()) {
+    if (this.hasOrgName()) {
       this.urlControl.markAsTouched();
     }
     if (!this.form.valid) {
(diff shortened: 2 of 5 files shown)


apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -780,13 +780,17 @@ export class CommitteeViewComponent {
     });
   }
 
-  private openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+  private async openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+    // Pre-fill from the user's profile work experiences — same pre-resolution the invite
+    // accept flow uses so the user doesn't have to re-enter an org they've already set.
+    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer());
+
```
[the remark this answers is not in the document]
andrest50: Fixed in commit a8c85a5e. Added `takeUntilDestroyed(this.destroyRef)` to the `resolveCurrentEmployer()` pipe inside `openOrganizationDialog`, with a `defaultValue: null` so `firstValueFrom` completes cleanly if the component is torn down mid-flight.

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
```diff
@@ -45,7 +45,11 @@ export class AcceptInviteOrganizationDialogComponent {
 
   protected readonly isNewOrg = computed(() => {
     const value = this.formValue();
-    return !value?.organization_id && !!value?.organization?.trim();
+    // URL is always required because organization_id is stripped before the payload reaches
+    // committee-service — the server must receive either an SFID (which we don't have) or
+    // name + domain. Treat any org without a confirmed website as requiring URL input.
+    const hasUrl = !!(value?.organization_url ?? '').trim();
+    return !hasUrl && !!value?.organization?.trim();
```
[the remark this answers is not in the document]
andrest50: Fixed in commit a8c85a5e. isNewOrg now returns !!value?.organization?.trim(), so the URL validators stay active whenever an org name is present rather than clearing on the first URL keystroke.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -44,11 +45,7 @@ export class InvitationAcceptFlowService {
     // the user's current employer from their profile (fails silently — dialog opens blank).
     const contextReady$: Observable<InvitationAcceptContext> = context.organization
       ? of(context)
-      : this.http.get<WorkExperienceEntry[]>('/api/profile/work-experiences').pipe(
-          take(1),
-          map((experiences) => ({ ...context, organization: currentEmployerFromWorkExperiences(experiences) })),
-          catchError(() => of(context))
-        );
```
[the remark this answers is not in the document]
andrest50: Fixed in commit a8c85a5e. When no invite org is present, the employer fallback path (resolveCurrentEmployer) already calls resolveOrgDomain internally, so preResolveOrganization is now only called on the invite-org path. This avoids the duplicate CDP lookup (which can be a find-or-create POST) and keeps the latency within the documented 2-second window.

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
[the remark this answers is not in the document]
andrest50: Fixed in commit a8c85a5e. isNewOrg is now simply !!value?.organization?.trim() — the URL validators stay active whenever an org name is present, regardless of whether a URL has been typed. The old !hasUrl check was inadvertently clearing validators on the first URL keystroke.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -69,50 +70,69 @@ export class InvitationAcceptFlowService {
...
-              id: resolved.id || null,
-              name: resolved.name || org.name,
-              website: normalizeToUrl(match.domain) ?? org.website,
-            },
+            ...org,
+            id: resolved.id || null,
+            name: resolved.name || org.name,
+            website: normalizeToUrl(match.domain) ?? org.website,
           }))
         );
```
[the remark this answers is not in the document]
andrest50: Fixed in commit a8c85a5e. preResolveOrganization (which calls resolveOrganization POST) is now only triggered on the invite-supplied-org path. The employer fallback path already resolves via resolveCurrentEmployer, so the extra POST no longer runs. resolveOrgDomain itself (inside resolveCurrentEmployer) uses searchOrganizations + a client-side name match before calling resolveOrganization, and skips the POST entirely when org.id and org.website are already populated.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -780,13 +780,17 @@ export class CommitteeViewComponent {
     });
   }
 
-  private openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+  private async openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+    // Pre-fill from the user's profile work experiences — same pre-resolution the invite
+    // accept flow uses so the user doesn't have to re-enter an org they've already set.
+    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer());
```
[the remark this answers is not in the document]
andrest50: Fixed in commit a8c85a5e. Added takeUntilDestroyed(this.destroyRef) to the resolveCurrentEmployer() pipe before firstValueFrom(), with defaultValue: null so it completes cleanly if the component is destroyed during the profile/domain lookup.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -780,13 +801,21 @@ export class CommitteeViewComponent {
...
 
-  private openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+  private async openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+    // Pre-fill from the user's profile work experiences — same pre-resolution the invite
+    // accept flow uses so the user doesn't have to re-enter an org they've already set.
+    // takeUntilDestroyed cancels the observable if the component is destroyed while the
+    // profile/domain lookup (≤2 s) is still in flight, preventing dangling subscriptions.
+    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer().pipe(takeUntilDestroyed(this.destroyRef)), {
+      defaultValue: null,
+    });
```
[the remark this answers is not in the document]
andrest50: Fixed in commit 80724e3b. Added a destroyed boolean set in destroyRef.onDestroy() before the await. After firstValueFrom completes, the guard checks the flag and returns null early if the component was destroyed during the employer prefetch, preventing dialogService.open from running over a new route.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -533,7 +537,8 @@ export class CommitteeViewComponent {
     if (joinMode === 'open') {
       let organization: CommitteeOrganizationReference | undefined;
       if (requiresOrg) {
-        const result = await this.openOrganizationDialog(committee.name);
+        this.resolvingOrg.set(true);
+        const result = await this.openOrganizationDialog(committee.name).finally(() => this.resolvingOrg.set(false));
```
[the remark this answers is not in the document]
andrest50: Fixed — added `|| resolvingOrg()` to the `[loading]` binding on both the Join Group and Apply to Join buttons so the CTA shows a spinner during the employer prefetch window.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -69,50 +68,72 @@ export class InvitationAcceptFlowService {
...
+  public resolveCurrentEmployer(): Observable<CommitteeOrganizationReference | null> {
+    return this.http.get<WorkExperienceEntry[]>('/api/profile/work-experiences').pipe(
+      take(1),
+      map((experiences) => currentEmployerFromWorkExperiences(experiences)),
+      switchMap((org) => (org ? this.resolveOrgDomain(org) : of(null))),
+      catchError((error) => {
+        console.warn('[InvitationAcceptFlowService] resolveCurrentEmployer failed; opening dialog blank', error);
+        return of(null);
+      })
+    );
```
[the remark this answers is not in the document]
andrest50: Fixed — added `timeout(2000)` to the work-experiences GET pipe in `resolveCurrentEmployer()` (before `catchError`), so a hung profile endpoint falls through to the blank-dialog path within 2 seconds rather than freezing `resolvingOrg` indefinitely.

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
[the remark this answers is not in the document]
andrest50: Fixed — changed `resolvingOrg` from `private` to `protected` so Angular's strict template type-checker accepts the `[loading]` binding.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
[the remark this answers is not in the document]
andrest50: Pre-existing pattern — the abstract `DestroyRef` class types `onDestroy` as returning `() => void` (a cleanup/unregister function), not `void`. `yarn check-types` confirms zero errors. Copilot's concern applies to some concrete class overrides in the Angular type declarations, not to the injected `DestroyRef` abstract type.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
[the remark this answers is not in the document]
andrest50: Same as above — the abstract `DestroyRef` class types `onDestroy` as returning `() => void`. `cleanupDestroyListener()` is valid both at the type level and at runtime. `yarn check-types` passes with zero errors.

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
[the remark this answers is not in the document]
andrest50: Pre-existing pattern — the identical `resolveOrganization(match.name, match.domain)` call without a domain guard was present in the original `preResolveOrganization` before this PR. Additionally, `OrganizationSuggestion.domain` is typed as a required `string` (not optional), so a runtime empty-string is an API contract violation rather than a known gap in our code.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -821,13 +827,34 @@ export class CommitteeViewComponent {
...
+    });
+
+    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer().pipe(takeUntilDestroyed(this.destroyRef)), {
+      defaultValue: null,
+    });
+
+    cleanupDestroyListener();
+    if (destroyed) {
+      return null;
+    }
```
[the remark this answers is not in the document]
andrest50: Fixed — added a `this.committeeId() !== committee.uid` check in both the open-join and application paths, immediately after `openOrganizationDialog` resolves. Angular reuses the component across `/groups/:id` navigations without destroying it so `DestroyRef` doesn't fire; this guard returns early if the active committee changed during the employer prefetch, preventing join/apply to the wrong committee.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -821,13 +827,34 @@ export class CommitteeViewComponent {
...
+
     const ref = this.dialogService.open(AcceptInviteOrganizationDialogComponent, {
       header: 'Confirm Organization',
       width: '32rem',
       modal: true,
       closable: true,
```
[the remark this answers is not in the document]
andrest50: Incorrect — Angular's abstract DestroyRef types onDestroy as returning () => void (a cleanup/unregister function), not void. Verified in node_modules/@angular/core type definitions: abstract onDestroy(callback: () => void): () => void. yarn check-types passes with zero errors. No change needed.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -821,13 +833,34 @@ export class CommitteeViewComponent {
...
+    });
+
+    const prefillOrg = await firstValueFrom(this.invitationAcceptFlow.resolveCurrentEmployer().pipe(takeUntilDestroyed(this.destroyRef)), {
+      defaultValue: null,
+    });
+
+    cleanupDestroyListener();
+    if (destroyed) {
+      return null;
+    }
```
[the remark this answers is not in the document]
andrest50: Same finding repeated — also incorrect. The abstract DestroyRef.onDestroy() returns () => void (a cleanup function). Verified via yarn check-types — zero type errors. No change needed.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -69,50 +68,73 @@ export class InvitationAcceptFlowService {
...
+   * dialog the same way the invite accept flow does.
    *
-   * Times out after 2 s and silently falls through — the dialog handles the unresolved case.
+   * Emits null and never throws — callers open the dialog with a blank org on any failure.
    */
-  private preResolveOrganization(ctx: InvitationAcceptContext): Observable<InvitationAcceptContext> {
-    const org = ctx.organization;
-    if (!org?.name?.trim() || org.id) {
-      return of(ctx);
+  public resolveCurrentEmployer(): Observable<CommitteeOrganizationReference | null> {
```
jordane, line 79: **Dialog can open already showing a red validation error.**

`currentEmployerFromWorkExperiences()` (`packages/shared/src/utils/invitation.utils.ts`) returns `{ name, id }` and never sets `website`. If `resolveOrgDomain`'s exact-name search misses, or the 2s timeout fires, the dialog opens with `organization` prefilled but `organization_url` empty. Under the new "URL always required" rule in the dialog component, that makes `orgInvalid()` true — and since `hasName` is true, `showOrgWarning()` is *also* true, so the user sees the red "Select an organization from the results, or click 'I want to create'" warning immediately on open, with no visible website field to fix it (that field only renders in manual mode). Before this PR the dialog opened blank with no warning.

_Fix:_ only prefill when a domain actually resolved (e.g. return `null` from `resolveCurrentEmployer` when `website` is unset), or gate `showOrgWarning()` on the control being touched/dirty so an untouched prefill doesn't show red immediately.

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
```diff
@@ -43,9 +43,12 @@ export class AcceptInviteOrganizationDialogComponent {
...
   private readonly urlStatus = signal<FormControlStatus>(this.urlControl.status);
 
+  // URL validators must stay active whenever an org name is present — not just while the
+  // URL field is empty — because clearing them on the first URL keystroke allowed any
+  // non-empty string (e.g. "not-a-url") to pass urlStatus as VALID and enable Confirm.
   protected readonly isNewOrg = computed(() => {
     const value = this.formValue();
-    return !value?.organization_id && !!value?.organization?.trim();
+    return !!value?.organization?.trim();
   });
```
audigregorie, line 55: **Major — `isNewOrg` is now semantically wrong**

The computed previously returned `!organization_id && !!name` ("is this a brand-new, not-in-CDP org"); it now returns just `!!name`. The name no longer reflects the logic, but it's still consumed in the template (`[domainRequired]="isNewOrg()"`) and in `onConfirm()` (`if (this.isNewOrg())`).

A future reader will reasonably assume `isNewOrg` means "no CDP id" and reintroduce the very bug this PR fixes (clearing URL validators once an id is present), or make decisions based on the wrong premise.

```suggestion
  protected readonly hasOrgName = computed(() => {
    const value = this.formValue();
    return !!value?.organization?.trim();
  });
```

(Then update the template `[domainRequired]="hasOrgName()"` and `onConfirm`'s `if (this.hasOrgName())`.)

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -821,13 +833,34 @@ export class CommitteeViewComponent {
     });
   }
 
-  private openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+  private async openOrganizationDialog(committeeName: string): Promise<AcceptInviteOrganizationDialogResult | null> {
+    // Pre-fill from the user's profile work experiences — same pre-resolution the invite
+    // accept flow uses so the user doesn't have to re-enter an org they've already set.
+    // takeUntilDestroyed cancels the observable if the component is destroyed while the
+    // profile/domain lookup (≤2 s) is still in flight, preventing dangling subscriptions.
```
audigregorie: **Minor — comment overstates the time bound**

The comment says "the profile/domain lookup (≤2 s)". `resolveCurrentEmployer()` places `timeout(2000)` *before* the `switchMap` into `resolveOrgDomain()`, so the outer timeout only bounds the `/api/profile/work-experiences` GET. `resolveOrgDomain()` then has its own independent 2 s timeout covering search + resolve. Worst case is ~4 s, not 2 s, and the `resolvingOrg` spinner can stay up that long.

```suggestion
    // takeUntilDestroyed cancels the observable if the component is destroyed while the
    // profile/domain lookup (≤4 s — 2 s for the work-experiences GET + 2 s for the CDP
    // domain resolve) is still in flight, preventing dangling subscriptions.
```

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
```diff
@@ -97,17 +100,19 @@ export class AcceptInviteOrganizationDialogComponent {
...
-        this.urlControl.setValidators([trimmedRequired(), httpsUrlValidator()]);
-      } else {
-        this.urlControl.clearValidators();
-      }
-      this.urlControl.updateValueAndValidity({ emitEvent: false });
-      // Manually sync after updateValueAndValidity({ emitEvent: false }) since
-      // suppressing the event means statusChanges won't fire for validator changes.
-      this.urlStatus.set(this.urlControl.status);
-    });
+    toObservable(this.isNewOrg)
```
audigregorie: **Minor — `effect` → `toObservable().subscribe()` swap is undocumented**

The change from `effect()` to `toObservable(this.isNewOrg).pipe(takeUntilDestroyed()).subscribe()` makes validator sync async (one tick delayed). The likely motivation — Angular disallows signal writes (`urlStatus.set(...)`) inside an `effect()` without `allowSignalWrites: true` — isn't captured, so the next reviewer may "simplify" it back to `effect()` and reintroduce the problem.

```suggestion
    // toObservable (not effect) so we can write urlStatus inside the callback —
    // effect() forbids signal writes without allowSignalWrites: true.
    toObservable(this.isNewOrg)
      .pipe(takeUntilDestroyed())
      .subscribe((isNew) => {
        if (isNew) {
          this.urlControl.setValidators([trimmedRequired(), httpsUrlValidator()]);
        } else {
          this.urlControl.clearValidators();
        }
        this.urlControl.updateValueAndValidity({ emitEvent: false });
        // Manually sync after updateValueAndValidity({ emitEvent: false }) since
        // suppressing the event means statusChanges won't fire for validator changes.
        this.urlStatus.set(this.urlControl.status);
      });
```

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -69,50 +68,73 @@ export class InvitationAcceptFlowService {
...
+    return this.http.get<WorkExperienceEntry[]>('/api/profile/work-experiences').pipe(
+      take(1),
+      timeout(2000),
```
audigregorie, line 82: **Minor — outer `timeout` doesn't cover the chained `resolveOrgDomain`**

`resolveCurrentEmployer()`'s `timeout(2000)` only guards the HTTP GET segment; the chained `resolveOrgDomain()` (search + find-or-create POST) runs outside it with its own 2 s budget. If the intent was a single 2 s overall budget for the prefill (the dialog-open blocker), this doesn't enforce it — total is ~4 s. Bounded, so not unsafe — just not matching the apparent intent.

If a 2 s overall cap is desired, drop the inner `timeout` in `resolveOrgDomain` and keep one outer `timeout(2000)` after the `switchMap` here.

audigregorie (commented):
## Code Review Summary

Solid prefill implementation — the destroy/route-reuse guards and the duplicate-CDP-POST avoidance in `accept()` are correct and well-reasoned. One maintainability issue (a now-misleading computed name) and a few doc/timeout-placement nits below.

### What's done well

- The `destroyed` flag + `cleanupDestroyListener` guard in `committee-view.component.ts` correctly handles the `firstValueFrom` `defaultValue` masking teardown — a subtle Angular quirk that would otherwise open the dialog over a new route.
- The `committeeId() !== committee.uid` post-dialog guard correctly handles Angular's route-reuse (same component instance, different `:id`) — a case the destroy guard alone wouldn't catch.
- Restructuring `accept()` to skip `preResolveOrganization` on the employer-fallback path avoids a duplicate CDP search + find-or-create POST, halving latency and eliminating a redundant side-effecting call.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
jordane, line 82: **Two sequential `timeout(2000)` calls stack to a ~4s worst case, not the "≤2s" claimed in comments.**

This `timeout(2000)` on the work-experiences GET runs before `resolveOrgDomain` (line 122) applies its own `timeout(2000)` on the search+resolve. Worst case is additive (~4s), but the comment here and in `committee-view.component.ts` says "profile/domain lookup (≤2 s)".

_Fix:_ collapse to one `timeout(2000)` on the outer `resolveCurrentEmployer` pipe, or update the comments to reflect the real worst case.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -69,50 +68,73 @@ export class InvitationAcceptFlowService {
...
-    return this.organizationService.searchOrganizations(org.name).pipe(
+    return this.organizationService.searchOrganizations(org.name!).pipe(
       take(1),
       switchMap((suggestions) => {
         const match = suggestions.find((s) => s.name.toLowerCase() === org.name!.toLowerCase().trim());
         if (!match) {
-          return of(ctx);
+          return of(org);
         }
         return this.organizationService.resolveOrganization(match.name, match.domain).pipe(
```
jordane: **Prefill still hits the CDP find-or-create write path on every Join/Apply click.**

`resolveOrganization()` is a find-or-create POST. It's called here via `resolveOrgDomain`, which `resolveCurrentEmployer` now calls from the open-join and apply flows (not just invites) — before the user has confirmed anything. An earlier review round flagged this; the fix that landed only removed the *duplicate* call on the invite path, not this one.

Risk is low since it's called with a canonical search-result domain (a find, not a create), but the caller set just widened from invite-only to invite + open-join + apply. Since this PR's whole premise is that `organization_id` is stripped before reaching committee-service, the resolve's only remaining value is the canonical name — which is already available from `match`. Consider prefilling directly from `{ name: match.name, website: normalizeToUrl(match.domain) }` without calling `resolveOrganization`, removing the write entirely and roughly halving latency.

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
```diff
@@ -43,9 +43,12 @@ export class AcceptInviteOrganizationDialogComponent {
   private readonly formValue = this.initFormValue();
   private readonly urlStatus = signal<FormControlStatus>(this.urlControl.status);
 
+  // URL validators must stay active whenever an org name is present — not just while the
+  // URL field is empty — because clearing them on the first URL keystroke allowed any
+  // non-empty string (e.g. "not-a-url") to pass urlStatus as VALID and enable Confirm.
   protected readonly isNewOrg = computed(() => {
```
jordane: `isNewOrg` no longer means "new org" — it's now true whenever an organization name is present, including CDP-resolved orgs. It still gates `[domainRequired]` on the search component and the URL-validator effect below. Consider renaming to something like `requiresUrl` or `hasOrgName` so a future change doesn't "restore" the old `!organization_id` check and reintroduce the validator-bypass bug from an earlier review round.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -197,6 +198,9 @@ export class CommitteeViewComponent {
   public invitesLoading = signal<boolean>(true);
   public applicationsLoading = signal<boolean>(true);
   public joiningOrLeaving = signal(false);
+  // Blocks the join/apply CTA during the async org-prefetch window so a second tap
+  // doesn't fire a parallel resolveCurrentEmployer() + dialog pair.
```
jordane: Nit: this comment says `resolvingOrg` blocks "during the async org-prefetch window," but `.finally()` on `openOrganizationDialog(...)` (below) actually clears it only once the dialog *closes* — so it stays `true` for the whole modal session, not just the prefetch. Behavior is fine (the CTA should stay disabled while the modal is open); the comment just doesn't match.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
```diff
@@ -523,7 +527,7 @@ export class CommitteeViewComponent {
 
   public async handleJoinRequest(): Promise<void> {
```
jordane, line 529: Nit: the `requiresOrg` blocks in the `open` and `application` branches below (~L541-548 and ~L570-577) are near-identical — set `resolvingOrg` → await dialog → bail if no result → bail if `committeeId` changed. Consider extracting a shared private helper so the two stale-committee guards can't drift apart independently in a future edit.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -69,50 +68,73 @@ export class InvitationAcceptFlowService {
...
+  /**
+   * Attempts to resolve the website/domain for an org reference via CDP so the dialog can
+   * pre-fill the URL field and prevent the "name and domain required" rejection from
+   * committee-service (which always needs the domain since organization_id is stripped from
+   * the payload before forwarding).
+   *
+   * Skips the search only when both id AND website are already present.
+   * Times out after 2 s and silently returns the unmodified org on any failure.
+   */
+  private resolveOrgDomain(org: CommitteeOrganizationReference): Observable<CommitteeOrganizationReference> {
```
jordane, line 107: Nit: `searchOrganizations(org.name!)` (line 105) passes the untrimmed name while the match comparison (line 108) trims it. Trim once into a local `const name = org?.name?.trim()` at the top of this method and reuse it for both — this also removes the need for the `!` non-null assertions.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
```diff
@@ -69,50 +68,73 @@ export class InvitationAcceptFlowService {
...
       catchError((error) => {
-        console.warn('[InvitationAcceptFlowService] Org pre-resolution failed; opening dialog with unresolved context', error);
-        return of(ctx);
+        console.warn('[InvitationAcceptFlowService] Org domain resolution failed; proceeding with unresolved org', error);
+        return of(org);
       })
     );
   }
 
+  private preResolveOrganization(ctx: InvitationAcceptContext): Observable<InvitationAcceptContext> {
```
jordane, line 138: Nit: `preResolveOrganization` lost its explanatory JSDoc when the two-step CDP rationale moved into `resolveOrgDomain`. A one-liner noting it adapts `resolveOrgDomain` to the invite-accept-context shape would help future readers.

jordane (commented):
Solid iteration — the bot round-trips landed real fixes, and the destroy/route-reuse guards are correct. Confirming for the record that Copilot's repeated `DestroyRef.onDestroy` claim is wrong: the abstract class types it as `(callback: () => void) => () => void`, so `cleanupDestroyListener()` is valid. Your rebuttal stands.

Three things I'd like addressed before merge, all in the prefill path (see inline comments):

1. When the profile employer can't be domain-resolved, the dialog now opens with a name prefilled, an empty URL, and a red validation warning already showing — with no visible website field to fix it outside manual mode. Pre-PR the dialog just opened blank.
2. The two sequential `timeout(2000)` calls stack to a ~4s worst-case spinner; the comments say ≤2s.
3. The find-or-create CDP POST still runs on every Join/Apply click. Since `organization_id` is stripped downstream anyway, prefilling straight from the search match's domain would drop the write and halve the latency.

Rest are nits. Two PR-shape items worth a mention: the branch name is `feat/LFXV2-2690-org-prefill-join-flow` (repo convention is `feat/LFXV2-2690`, no descriptive suffix), and commit `bbde395c2` uses `chore: merge main into ...`, which commitlint rejects — a rebase onto `main` instead of the merge commit would clear both.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
andrest50: Fixed — added map((org) => (org?.website?.trim() ? org : null)) after the switchMap in resolveCurrentEmployer. The dialog now opens blank when the CDP search found no domain match, rather than prefilling the name with an empty URL field (which was causing orgInvalid+showOrgWarning to fire immediately).

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
andrest50: Fixed — renamed isNewOrg to hasOrgName in the computed, both usages in the TS (toObservable call and onConfirm), and the template [domainRequired] binding. Added a comment on the computed explaining why it is not named isNewOrg (organization_id is stripped before forwarding, so URL is required regardless of CDP resolution).

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
andrest50: Fixed — updated the comment in committee-view.component.ts to say ≤4 s worst case (2 s for work-experiences GET + 2 s for CDP domain resolve).

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
andrest50: Fixed — added the comment you suggested: 'toObservable (not effect) so we can write urlStatus inside the callback — effect() forbids signal writes without allowSignalWrites: true.'

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
andrest50: Fixed — added a note to resolveOrgDomain's JSDoc clarifying that resolveCurrentEmployer's preceding 2 s GET timeout means the total worst-case prefill window is ~4 s. Comments updated in both files.

andrest50: Fixed — both the committee-view.component.ts comment and resolveOrgDomain's JSDoc now say ≤4 s worst case (2 s work-experiences GET + 2 s CDP resolve). Addressed alongside the duplicate thread from audigregorie.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
andrest50: Fixed — replaced the resolveOrganization find-or-create POST with a direct prefill from the search match: { ...org, id: null, name: match.name, website: normalizeToUrl(match.domain) ?? org.website }. Since organization_id is stripped before forwarding to committee-service anyway, the canonical name and domain from the search result are all that is needed. This removes the write call entirely and halves prefill latency.

apps/lfx-one/src/app/shared/components/accept-invite-organization-dialog/accept-invite-organization-dialog.component.ts
andrest50: Fixed — renamed to hasOrgName. Addressed alongside the duplicate thread from audigregorie.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
andrest50: Fixed — updated the comment to say 'while the org-prefetch is running and while the dialog is open (signal stays true until .finally() resolves when the dialog closes)'.

apps/lfx-one/src/app/modules/committees/committee-view/committee-view.component.ts
andrest50: Out of scope for this PR — the two branches intentionally diverge after the dialog closes (open-mode issues a joinCommittee call; application-mode opens the application dialog), so extracting a helper would add indirection without preventing drift. Deferring to a follow-up refactor.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
andrest50: Fixed — extracted const name = (org?.name ?? '').trim() at the top of resolveOrgDomain and used it for both searchOrganizations and the find comparison. Non-null assertions removed.

apps/lfx-one/src/app/shared/services/invitation-accept-flow.service.ts
andrest50: Fixed — added a one-liner JSDoc: 'Adapts resolveOrgDomain to the invite-accept-context shape, threading the resolved org back into ctx.'

audigregorie (approved):

PR #1339, opened 2026-08-06
fix(cla): pad single-valued identity query params to survive gateway multi-value strip

### LINKED ISSUE #1314: My CLAs: agreement count dropped — GitHub-matched EasyCLA user rows no longer returned
Part of EPIC #1157.

## Summary

The number of agreements shown in the "My CLAs" tab dropped recently. Reported by Heather Willson (prod, <24h ago); a similar reduction was observed in dev over the last couple of days.

- Heather saw 5 signed+approved CLAs; after a page refresh only 1 ECLA (ONAP) remains. It worked earlier the same morning. Confirmed via impersonation by Juansebastian Arias.
- The remaining row belongs to her LFID-based EasyCLA user (`9508f419-bea3-11ed-9539-e6c58d59437e`, `HWillson` / `hwillson@linuxfoundation.org`). The 4 missing rows are on her GitHub-linked EasyCLA user (`65b110b3-38c3-44b7-b0ab-6da999d86522`, GitHub `WillsonHG`) — that user is no longer matched, even though her LFX account has the verified GitHub identity `WillsonHG` (id `46732999`).
- Response showed `"matchedUserIds": 1, "unmatched": false, "githubLinked": true`.

## What we know

- The EasyCLA code deltas deployed in the same window (linuxfoundation/easycla#5133 dev Aug 2, linuxfoundation/easycla#5143 prod Aug 4 / v2.8.1) only add `projectName`/`projectLogo` — they cannot drop rows.
- Upstream `GET /v4/my-clas` verifies each identity key against the platform user-service and **silently drops unverified keys** into `skippedIdentities` (lookup failures fail soft → fewer rows, HTTP 200). A platform user-service/token failure would exactly produce "GitHub-user rows missing, LFID rows present".
- SS hides `valid: false` rows (FR-002), and ECLA validity is recomputed live against current approval lists with a fail-closed GitHub org check — a separate possible reducer for ECLAs.
- Case-sensitivity of the GitHub username (`WillsonHG` sent lowercased?) was raised in the thread; upstream canonicalization is case-insensitive by code, but worth verifying.

## Slack
https://linuxfoundation.slack.com/archives/C0BGU6YQC2F/p1785865948078329

## Next steps / diagnostics

1. Capture the raw upstream response for the affected user: `skippedIdentities`, `userIds`, `resultCount` (vs count of `valid: true`).
2. Check EasyCLA v4 API lambda logs for `unable to lookup the LF user in the platform user-service` / `unable to list the LF user's connected identities` warnings.
3. Check whether the count is stable across refreshes (flapping ⇒ live validity/org-check path).

ahmedomosanya:
Fixes #1314. Upstream root cause: linuxfoundation/lfx-gateway#114.

> **Interim / stopgap fix.** This is the BFF half of a two-part fix. It must be removed once the durable gateway fix (lfx-gateway#114, in the `linuxfoundation/traefik` fork) deploys.

## Summary
- BFF `identityQuery` in `cla.service.ts` now pads single-valued `githubId` / `githubUsername` (and `email`) to length >= 2 (duplicating the lone value) whenever any identity key is multi-valued, so they survive the gateway.
- Adds 3 tests + 1 assertion in `cla.service.spec.ts` (38 passing).

## Why
A multi-email contributor's GitHub-only CLAs vanished from My CLAs (#1314) because the LFX API gateway's `awslambda` middleware (lfx-gateway#114) drops single-valued query params whenever any param repeats. This is the interim BFF recovery until the durable gateway fix ships.

## Non-obvious details
- Padded values are byte-identical, and the backend de-dupes, so there is no double-count.
- `lfUsername` is intentionally **not** padded — the backend derives the principal from `X-USERNAME`.
- The FR-011 guard test asserts the invariant (github keys reach the backend), not the padding mechanism, so it outlives workaround removal.

## Follow-ups
- Durable gateway fix in the `linuxfoundation/traefik` fork (separate repo).
- Remove this padding once lfx-gateway#114 deploys.

## Test plan
- [x] `cd apps/lfx-one && npx vitest run src/server/services/cla.service.spec.ts` -> 38 passing.

### DIFF
--- a/apps/lfx-one/src/server/services/cla.service.spec.ts
+++ b/apps/lfx-one/src/server/services/cla.service.spec.ts
@@ -303,6 +303,104 @@ describe('ClaService.getMyClas', () => {
     expect(calledUrl).toContain('githubUsername=octocat');
   });
 
+  it('duplicates single-valued github keys when email is multi-valued, to survive the gateway multi-value strip (lfx-gateway#114)', async () => {
+    getEffectiveUsername.mockReturnValue('alice');
+    getEffectiveEmail.mockReturnValue('alice@x.org');
+    getEffectiveSub.mockReturnValue('auth0|abc');
+    getUserEmails.mockResolvedValueOnce({
+      primary_email: 'alice@x.org',
+      alternate_emails: [
+        { email: 'alice@work.com', verified: true },
+        { email: 'alice@personal.com', verified: true },
+      ],
+    });
+    getUserIdentities.mockResolvedValueOnce([{ provider: 'github', user_id: 'github|13434323', connection: 'github', profileData: { nickname: 'octocat' } }]);
+    gatewayFetch.mockResolvedValueOnce({ userIds: ['u-1'], clas: [] });
+
+    await new ClaService().getMyClas(req);
+
+    const calledUrl = gatewayFetch.mock.calls[0][1] as string;
+    const params = new URLSearchParams(calledUrl.slice(calledUrl.indexOf('?') + 1));
+    // 3 verified emails -> email is multi-valued, so the lone github keys are padded to 2 so the
+    // gateway keeps them (it forwards only multi-valued params once any param repeats); the backend de-dupes.
+    expect(params.getAll('email').length).toBeGreaterThan(1);
+    expect(params.getAll('githubId')).toEqual(['13434323', '13434323']);
+    expect(params.getAll('githubUsername')).toEqual(['octocat', 'octocat']);
+    // lfUsername stays single-valued (harmless: the backend derives the principal from the X-USERNAME header).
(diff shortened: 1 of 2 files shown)


MRashad26 (approved):
Code enforcer: clean (BFF-only change, no template touches).

Logic is sound: `anyMulti` triggers padding only when the gateway bug applies; `appendKeepAlive` pads single-value arrays to 2 identical values so the gateway forwards them, and the empty-array early-return keeps the no-multi-identity path untouched. `lfUsername` correctly stays on the `params.set` path — backend derives the principal from `X-USERNAME`, not the query string, so the single value is harmless.

Test coverage is thorough: padding test + FR-011 invariant guard (survives workaround removal) + single-email no-regression + impersonation target-token check. 38 passing.

PR #1343, opened 2026-08-06
fix(enrollment): build enrollment hrefs via URL api

fayazg:
## Summary

- Replace the `${base}${ctaPath}` string concatenation in the individual enrollment component with a `buildEnrollmentHref` util that composes the URL via `new URL` + `URLSearchParams`.
- Removes the silent dependency on the base ending in `/` and `ctaPath` starting with `?` — a future env URL or `ctaPath` format change can no longer break the CTA link.
- Added unit coverage in `enrollment.utils.spec.ts` for the trailing-slash, path-segment, and leading-`?` cases (plus the `renew=true` param).

LFXV2-1664

### DIFF
--- a/packages/shared/src/utils/enrollment.utils.spec.ts
+++ b/packages/shared/src/utils/enrollment.utils.spec.ts
@@ -8,7 +8,7 @@
 import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
 
 import { EnrollmentMembership, IndividualEnrollment } from '../interfaces';
-import { deriveEnrollmentStatus, enrollmentStatusSeverity } from './enrollment.utils';
+import { buildEnrollmentHref, deriveEnrollmentStatus, enrollmentStatusSeverity } from './enrollment.utils';
 
 /** Formats a Date as YYYY-MM-DD, matching the EndDate shape the member-service returns. */
 function isoDate(date: Date): string {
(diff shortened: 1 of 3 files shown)


dealako (commented):
_(tooling artifact — please disregard; full review follows below)_

packages/shared/src/utils/enrollment.utils.ts
```diff
@@ -33,6 +33,15 @@ export function deriveEnrollmentStatus(item: IndividualEnrollment): EnrollmentDi
   return 'Active';
 }
 
+// Composes an enrollment CTA URL via URL + URLSearchParams (not string concat), so it is robust to
```
dealako, line 36: [nit] Consider JSDoc for the new exported utility

Issue: `buildEnrollmentHref` documents itself with a `//` line comment rather than a JSDoc block.

Proof: Sibling utilities in `packages/shared/src/utils/url.utils.ts` (e.g. `isValidUrl`) use `/** ... @param ... @returns */` JSDoc. That said, the pre-existing `deriveEnrollmentStatus` right above this function in the same file also uses `//` comments, so this new function is at least locally consistent with the file it was added to — self-challenged this down from `[minor]` to `[nit]` for that reason.

Why it matters: JSDoc gives IDE tooltips/param hints for a function used across two call sites; a plain comment doesn't surface in autocomplete.

Fix (optional, take-it-or-leave-it given local consistency):
```ts
/**
 * Composes an enrollment CTA URL via URL + URLSearchParams (not string concat).
 * @param base - Enrollment base URL
 * @param ctaPath - Query string/path fragment from the API (with or without leading `?`)
 * @param renew - When true, adds `renew=true`
 */
```

packages/shared/src/utils/enrollment.utils.spec.ts
```diff
@@ -148,3 +148,27 @@ describe('enrollmentStatusSeverity', () => {
...
+  });
+
+  it('tolerates a base that carries a path segment (no trailing slash)', () => {
+    expect(buildEnrollmentHref('https://enroll.example.org/join', CTA)).toBe('https://enroll.example.org/join?product=prod-1&project=tlf');
+  });
+
+  it('tolerates a ctaPath without a leading question mark', () => {
+    expect(buildEnrollmentHref('https://enroll.example.org/', 'product=prod-1&project=tlf')).toBe('https://enroll.example.org/?product=prod-1&project=tlf');
+  });
+});
```
dealako, line 174: [nit] Missing test for empty `ctaPath`

Issue: `buildEnrollmentHref` is never tested with `ctaPath = ''`.

Proof: All cases in this spec use non-empty `ctaPath` values; `ctaPath` is typed as plain `string` on `IndividualEnrollment` (`enrollment.interface.ts:29`), so an empty string is a type-valid input, even if current call sites (`profile-individual-enrollment.component.ts:72-73`) always pass a real API-provided path.

Why it matters: cheap to cover, and documents that an empty `ctaPath` degrades gracefully to just the base URL rather than throwing or producing a malformed href.

Fix:
```ts
it('handles empty ctaPath by returning the base URL', () => {
  expect(buildEnrollmentHref('https://enroll.example.org/', '')).toBe('https://enroll.example.org/');
});
```

dealako (approved):
Hey @fayazg — nice cleanup here. Swapping the `${base}${ctaPath}` string concatenation for `URL` + `URLSearchParams` composition removes a genuinely fragile implicit contract (base needs a trailing `/`, `ctaPath` needs a leading `?`), and the new `enrollment.utils.spec.ts` coverage for trailing-slash, path-segment, leading-`?`, and `renew=true` cases directly targets the scenarios the old code got wrong. Traced the `base`/`ctaPath` value flow end-to-end (env config → API `ctaPath` → `buildEnrollmentHref` → anchor href) — no injection, open-redirect, or PII-leak vectors; the `URL` API also incidentally hardens against a few things string concat didn't (can't escape the base origin, params get properly encoded).

Ran this through three independent review passes (security/privacy, correctness/perf/tests, style/docs) — nothing rose above nit level.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 — optional JSDoc on `buildEnrollmentHref` for consistency with sibling utils in `url.utils.ts` (though it does match the pre-existing `deriveEnrollmentStatus` comment style in the same file); missing test for an empty `ctaPath` input
- ❔ Question: 0

Agreement with existing bot passes: Cursor Bugbot's "Low Risk" summary matches my read — this is a narrow, well-tested mechanical refactor with no auth/data-path changes. CodeRabbit skipped its review and Copilot returned no comments; nothing there to reconcile against.

**Decision: ✅ Approved with minor comments** (both nits, non-blocking — feel free to take or leave them).

PR #1353, opened 2026-08-07
refactor(enrollment): model enrollment list as a single state signal

fayazg:
## Summary

- Replace the individual-enrollment component's `enrollments: DisplayEnrollment[] | null | undefined` + `enrollmentError: string | null` re-encoding with a single discriminated-union state signal (`loading | loaded | error`), mirroring the service's `EnrollmentsState`.
- Add a shared generic `EnrollmentsStateOf<T>` so the raw (`EnrollmentsState`) and display (`DisplayEnrollmentsState`) variants can't drift apart.
- Template now uses `@let state = displayState();` + `@switch (state.kind)` / `@case`, dropping both `!` non-null assertions and the manual null/undefined tri-state.
- Inline the single-iteration `@for` loading skeleton (no loop).
- Optimistic auto-renew machinery (pending flags + transient overrides) is preserved, now operating on the `loaded` arm.

Refs LFXV2-3035

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/individual-enrollment/profile-individual-enrollment.component.html
+++ b/apps/lfx-one/src/app/modules/profile/individual-enrollment/profile-individual-enrollment.component.html
@@ -2,10 +2,11 @@
 <!-- SPDX-License-Identifier: MIT -->
 
 <div class="flex flex-col gap-4" data-testid="individual-enrollment-page">
-  @if (displayedEnrollments() === undefined) {
-    <!-- Loading skeletons -->
-    <div class="flex flex-col gap-4" data-testid="individual-enrollment-loading">
-      @for (i of [1]; track i) {
+  @let state = displayState();
+  @switch (state.kind) {
+    @case ('loading') {
+      <!-- Loading skeleton -->
+      <div class="flex flex-col gap-4" data-testid="individual-enrollment-loading">
         <lfx-card>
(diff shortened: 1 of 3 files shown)


luismoriguerra (approved):
## Audit — ✅ Approved

Merge readiness: **safe-to-merge**.

**Gates re-run** (isolated worktree at base `399c113`): `check-types` ✅ · `lint:check` ✅ (1 pre-existing warning in `campaigns.component.ts`, untouched) · `vitest run` ✅ 611/611 · `yarn build` ✅ (Angular template type-check clean).

Behavior-preserving: loading skeleton, error empty-state (incl. the `state.message || …` fallback), zero-length empty-state and the card list are reachable under the same conditions; all `data-testid`s preserved, so existing e2e stays valid. No dangling refs to the removed `enrollments` / `enrollmentError` / `displayedEnrollments`.

### Findings
- **Nit** — `profile-individual-enrollment.component.ts:181-182`: the body says both `!` assertions are dropped; that's true of the template, but `overrides.get(membershipId)!` and `item.membership!` remain (guarded, carried over verbatim). `?? false` would close it out.
- **Nit** — `toDisplayState` returns `state` directly for the non-`loaded` arms, relying on structural identity between the two unions. Correct and type-checked; a one-line comment would make it explicit.

No blocking findings. No inline threads opened (default `blocking` selection); both nits are covered above.

---

PR #1354, opened 2026-08-07
fix(public-profile): filter trainings/certs to myprofile allow-list

fayazg:
## Summary

- The public profile artifact lists every training/certification row, while myprofile's dashboard applies allow-lists. This mirrors that filtering in the server-side projection so the two views stay in parity.
- **Trainings:** keep only allow-listed Types (E-Learning, Instructor-Led, edX) with allowed enrollment statuses (Enrolled / Completed / Started / Not started); blank epoch-placeholder dates; sort by name. Exam / subscription / bundle rows no longer render as trainings.
- **Certifications:** completed only, dropping any with an epoch or absent StartDate.
- Raw `Status` drives the filter only and is never projected — preserves the render-only PII boundary.
- Adds unit tests for both projection helpers via the exported `projectPublicProfile` seam.

Fixes LFXV2-3039. Relates to LFXV2-2631.

### DIFF
--- a/apps/lfx-one/src/server/constants/public-profile.constants.ts
+++ b/apps/lfx-one/src/server/constants/public-profile.constants.ts
@@ -21,3 +21,17 @@ export const PUBLIC_PROFILE_FETCH_TIMEOUT_MS = 10_000;
  * interpolated into the S3 key without path traversal or probing unrelated keys.
  */
 export const PUBLIC_PROFILE_USERNAME_PATTERN = /^[A-Za-z0-9._-]{1,100}$/;
+
+// Training `Type` values surfaced publicly (myprofile parity): exam/subscription/bundle rows are
+// dropped from trainings; completed certifications surface via `certification_activities`.
+export const PUBLIC_PROFILE_TRAINING_TYPE_ALLOWLIST: ReadonlySet<string> = new Set(['E-Learning', 'Instructor-Led', 'edX']);
+
+/** Enrollment `Status` values kept on the public trainings list (myprofile parity). */
+export const PUBLIC_PROFILE_TRAINING_STATUS_ALLOWLIST: ReadonlySet<string> = new Set(['Enrolled', 'Completed', 'Started', 'Not started']);
+
+/** Certification `Status` values kept on the public certifications list — completed only (myprofile parity). */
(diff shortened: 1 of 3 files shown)


luismoriguerra (approved):
## Audit — ✅ Approved

Merge readiness: **merge-after-confirming the allow-list values still match the artifact's current vocabulary** (finding 1 — a mismatch fails silently to an empty rail, not an error).

**Gates re-run** (isolated worktree at base `399c113`): `check-types` ✅ · `lint:check` ✅ (1 pre-existing warning, untouched) · `vitest run` ✅ 621/621, incl. 46 in `public-profile.service.spec.ts`.

Privacy: strictly *reduces* exposure on an anonymous endpoint. No new field is projected; `Status` is read only inside the filter predicates and never mapped out (pinned by the wire-form assert). The `resolvePublicFlag` fail-closed gate is untouched. Client rails are guarded by `hasTrainings()` / `hasCertifications()`, so a fully-filtered list hides the section rather than rendering an empty card.

### Findings
1. **🔒 Non-blocking** — `public-profile.constants.ts:27,30,33`: exact-match, case-sensitive literals duplicated from myprofile (outside this repo), with no shared contract and no drift signal. An upstream rename (`eLearning`, `Not Started`) would silently empty every public trainings rail. Suggest a debug-level count of rows dropped on unrecognized `Type`, so drift surfaces in Datadog.
2. **Non-blocking** — `public-profile.service.ts:118,158`: `date.includes('1970')` is a whole-string substring test. Faithful to myprofile, but also matches epoch-ms strings or any value containing `1970`. `startsWith` would tighten it; low risk on ISO-8601 input.
3. **Non-blocking** — `public-profile.service.ts:157` additionally requires `startDate !== undefined`, stricter than the parity rule quoted on `PUBLIC_PROFILE_EPOCH_PLACEHOLDER` (`!StartDate.includes('1970')`). Intentional per the PR body — worth saying so in that comment too.
4. **Nit** — trainings sort by `Name` (`:141`); certifications don't. Visible asymmetry between adjacent rails.
5. **Nit** — `spec.ts:159`: `not.toContain('Status')` is a payload-wide substring assert; would false-fail on a fixture named e.g. "Status Quo". Fine as written.

No blocking findings. No inline threads opened (default `blocking` selection); all five are covered above.

---

Issue #1313, opened 2026-08-05
QA: Testing and data validation for "My CLAs"


mlehotskylf:
QA testing and data validation of the read-only "My CLAs" view before M1 ships, so we can confirm what Self Serve shows matches the CLA data in the EasyCLA backend.

Acceptance criteria:
- [ ] Sampled parity check: ICLAs and currently valid ECLAs shown in the Me lens match backend data (SC-001)
- [ ] Signed ICLA PDF download works on a sampled set (≥99%)
- [ ] Multi-identity contributors (multiple linked emails / GitHub accounts) resolve to the correct CLAs
- [ ] Empty state verified for a user with no CLAs on file
- [ ] Unmatched-identity telemetry produces data and is reviewed against the agreed threshold
- [ ] Issues found are filed and triaged under this epic

Note: EasyCLA DynamoDB data is mirrored into Snowflake, so parity checks can be run with SQL instead of scanning DynamoDB directly:
- Ingest (raw Fivetran): `FIVETRAN_INGEST.DYNAMODB_PRODUCT_US_EAST_1.*`
- Bronze models: `ANALYTICS.BRONZE_FIVETRAN_EASYCLA.*`

Mind the mirror's sync lag when comparing against what the UI shows live.

WillsonHG:
Lets select users with the most amount of CLAs and the oldest CLAs and then of course our team CLAs.

PR #1356, opened 2026-08-07
docs(cla): correct pdf-url ownership comment scope

mlehotskylf:
## Summary

Corrects a code comment in `clas.controller.ts` that overstated what the `/v4/my-clas/{id}/pdf` endpoint guarantees. **Comment only — no behaviour change.**

The old comment claimed the endpoint returns 404 "(never 403)" and "never leaks whether an ID exists". Neither holds universally:

- **422** — go-swagger's UUID format validation rejects malformed IDs before EasyCLA's handler runs
- **403** — the gateway rejects path traversal, also upstream of the handler

`getPdfUrl` normalizes only 404, so both statuses reach the client and *do* distinguish malformed input from unknown IDs. The uniform-404 guarantee is real, but only for well-formed IDs.

This is why reading `my_clas/handlers.go` alone makes the finding look invalid — that handler returns only 401/400/404/500/200. The 422/403 originate upstream of it, and both were captured live during QA.

## Why comment-only

The underlying error-mapping gap is low severity: no CLA data is exposed and the UI shows a generic toast in every case. Correcting the false guarantee is the proportionate fix; widening the catch in `getPdfUrl` would be a behaviour change and belongs in its own PR if wanted.

## Context

Found during M1 QA for "My CLAs" (SS-1). Full QA record lives in the private `easyclav2-migration-planning` repo.

Refs linuxfoundation/lfx-self-serve#1313

### DIFF
--- a/apps/lfx-one/src/server/controllers/clas.controller.ts
+++ b/apps/lfx-one/src/server/controllers/clas.controller.ts
@@ -51,9 +51,12 @@ export class ClasController {
       const signatureId = (req.params['signatureId'] ?? '').trim();
 
       // EasyCLA's /v4/my-clas/{id}/pdf enforces ownership + ICLA eligibility against the
-      // session's identity and returns 404 for unknown, not-owned and ECLA IDs (never 403),
-      // so it never leaks whether an ID exists. SS just resolves the session identity and
-      // passes it through — no separate ownership pre-check needed.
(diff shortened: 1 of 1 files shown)


mlehotskylf:
Closing — the error-mapping gap this documents is too minor to be worth a comment change. No CLA data is exposed and the UI shows a generic toast in every case. Recording the decision rather than the code change.

PR #1357, opened 2026-08-07
feat(groups): surface external_sources on public group detail page

MRashad26:
## Summary

Surfaces the new committee-service `external_sources` field (source-labeled OCG groups/events/communities linked to a committee) on the public group detail page, with a "Powered by OCG" badge for OCG-provided entries.

- Passes `committee.external_sources` through the public group detail response (`public-groups.controller.ts`)
- Renders an "External Sources" section following the existing "Links" section pattern, gated by `hasExternalSources()`
- Filters through the existing `isValidUrl()` shared utility (`@lfx-one/shared/utils`, same pattern as `activity-feed.utils.ts`) before binding any URL to `[href]` — untrusted upstream URLs are never rendered unvalidated
- Consolidates `PublicGroupExternalSource` as a type alias of the canonical `CommitteeExternalSource` (they were identical, duplicated interfaces)

## External References

- Builds on linuxfoundation/lfx-v2-committee-service#175, which added the `external_sources` field to `CommitteeBase`.

## Trade-offs / accepted findings

- **No new test coverage added.** `group-detail.component` has zero existing unit or e2e test coverage today (including its pre-existing Links/Meetings sections), so adding tests scoped to only this one feature would be inconsistent scope creep beyond the ticket. Flagged as a pre-existing gap, not introduced by this change.
- **`@for` track key (`track $index`):** the list is always replaced wholesale on navigation (never client-side reordered or mutated in place), so `$index` is safe here and avoids relying on `source.url` uniqueness, which isn't guaranteed by the contract. (Earlier drafts of this description said the team settled on `source.url` — that was superseded by a review fix and this section is now up to date.)
- **`data-testid` per external source** includes the loop index (`external-source-<entity_type>-<index>`) so E2E selectors stay unique when multiple sources share an `entity_type`.
- **"Powered by OCG" badge always renders** — `CommitteeExternalSource.provider` is currently a `'ocg'` literal type, so the badge shown on every entry is correct today; the `@if` gate was removed since it could never be false.

## Test plan

- [x] `tsc --noEmit` clean (shared package + app)
- [x] `eslint` clean on all changed files
- [x] `yarn build` succeeds (shared build, client + SSR server bundles)
- [x] License headers present on all changed files
- [ ] Manual verification in dev: public group detail page renders External Sources section with OCG badge when `external_sources` is present


apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -173,6 +173,33 @@ <h3 class="text-gray-600 text-sm font-normal">Links</h3>
...
+          <!-- External Sources -->
+          @if (hasExternalSources()) {
+            <hr class="border-gray-200 my-4" />
+            <div data-testid="group-external-sources-section">
+              <div class="flex items-center gap-2 mb-3">
+                <i class="fa-light fa-diagram-project text-gray-500 text-xs"></i>
+                <h3 class="text-gray-600 text-sm font-normal">External Sources</h3>
+              </div>
+              <div class="flex flex-col gap-2">
+                @for (source of safeExternalSources(); track source.url) {
```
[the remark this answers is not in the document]
MRashad26: Done — changed the `@for` track expression to `$index`. Good catch: the upstream Goa schema (committee-service PR #175) caps `external_sources` at 50 entries and validates each URL, but never enforces uniqueness, so duplicate URLs are legitimately possible and would have triggered NG0955 with the old `track source.url`.

See commit e573633: `group-detail.component.html` now tracks by `$index`.

packages/shared/src/interfaces/committee.interface.ts
```diff
@@ -456,6 +456,21 @@ export interface Committee {
...
+  external_sources?: CommitteeExternalSource[];
+}
+
+/** A single source-labeled external entity linked to a committee (e.g. an OCG group or event). */
+export interface CommitteeExternalSource {
+  provider: 'ocg';
+  entity_type: 'community' | 'group' | 'event';
+  label: string;
+  /** Unique per committee within `external_sources` — safe to use as a list track key. */
+  url: string;
```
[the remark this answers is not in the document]
MRashad26: Done — removed the uniqueness claim from the JSDoc. Verified against `cmd/committee-api/design/type.go` in committee-service PR #175: `external_sources` is capped at 50 entries and each URL is validated for format, but there's no uniqueness constraint on `url`. You're right that documenting it as unique would mislead consumers.

See commit e573633: `committee.interface.ts` no longer claims `url` is unique per committee.

MRashad26:
## Review Feedback Addressed

Commit: e57363342

### Changes Made
- **group-detail.component.html**: `@for` loop now tracks by `$index` instead of `source.url` — the upstream Goa schema (committee-service PR #175) caps `external_sources` at 50 and validates each URL, but never enforces uniqueness, so duplicate URLs are legitimately possible and would have triggered NG0955 (per @copilot-pull-request-reviewer)
- **committee.interface.ts**: removed the JSDoc claim that `url` is unique per committee within `external_sources` — verified against the upstream Goa design that this guarantee doesn't exist (per @copilot-pull-request-reviewer)

### Threads Resolved

Both findings were correct — this replaces the earlier (incorrect) assumption from our own pre-PR review that `url` was safely unique. Thanks for catching it against the real upstream contract.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -173,6 +173,33 @@ <h3 class="text-gray-600 text-sm font-normal">Links</h3>
...
+          <!-- External Sources -->
+          @if (hasExternalSources()) {
+            <hr class="border-gray-200 my-4" />
+            <div data-testid="group-external-sources-section">
+              <div class="flex items-center gap-2 mb-3">
+                <i class="fa-light fa-diagram-project text-gray-500 text-xs"></i>
+                <h3 class="text-gray-600 text-sm font-normal">External Sources</h3>
+              </div>
+              <div class="flex flex-col gap-2">
+                @for (source of safeExternalSources(); track $index) {
```
audigregorie, line 185: **Major — `track $index` contradicts the PR description and checklist 14.14**

The PR description states the team "settled on `source.url` alone" for the track key, but the code uses `track $index`. Frontend checklist 14.14 flags `track $index` as causing unnecessary DOM churn when items reorder. `source.url` is a stable key that matches both the documented decision and the checklist — and the description already argues the list is replaced wholesale on navigation, so uniqueness of `source.url` is sufficient.

```suggestion
                @for (source of safeExternalSources(); track source.url) {
```

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -173,6 +173,33 @@ <h3 class="text-gray-600 text-sm font-normal">Links</h3>
...
+                  <a
+                    [href]="source.url"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    class="flex items-center gap-2 text-sm text-blue-500 hover:underline"
+                    [attr.data-testid]="'external-source-' + source.entity_type">
+                    <i class="fa-light fa-arrow-up-right-from-square w-4 text-center text-gray-400"></i>
+                    <span>{{ source.label }}</span>
+                    @if (source.provider === 'ocg') {
```
audigregorie: **Minor — `@if (source.provider === 'ocg')` is always true given the current type**

`CommitteeExternalSource.provider` is typed as the literal `'ocg'` (not a union), so this condition is always true at runtime — the badge always renders. If this is intentional future-proofing for non-OCG providers, widen the type (e.g. `provider: 'ocg' | string`) so the check is meaningful; otherwise drop the `@if` and always render the badge.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
```diff
@@ -173,6 +173,33 @@ <h3 class="text-gray-600 text-sm font-normal">Links</h3>
...
+                <h3 class="text-gray-600 text-sm font-normal">External Sources</h3>
+              </div>
+              <div class="flex flex-col gap-2">
+                @for (source of safeExternalSources(); track $index) {
+                  <a
+                    [href]="source.url"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    class="flex items-center gap-2 text-sm text-blue-500 hover:underline"
+                    [attr.data-testid]="'external-source-' + source.entity_type">
```
audigregorie: **Minor — non-unique `data-testid` when multiple sources share an `entity_type`**

`'external-source-' + source.entity_type` collides when two sources have the same `entity_type` (e.g. two OCG communities both produce `external-source-community`), so E2E selectors by `data-testid` match multiple elements. Append the loop index or `external_id` for uniqueness, e.g. `'external-source-' + source.entity_type + '-' + $index`.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
```diff
@@ -79,6 +80,12 @@ export class GroupDetailComponent {
 
   protected readonly hasUpcomingMeetings = computed(() => (this.group()?.upcoming_meetings?.length ?? 0) > 0);
 
+  protected readonly safeExternalSources: Signal<PublicGroupExternalSource[]> = computed(
+    () => this.group()?.external_sources?.filter((source) => isValidUrl(source.url)) ?? []
+  );
```
audigregorie: **Minor — multi-line `computed()` should use a private `init*()` function**

Per `component-organization.md` §3, computed signals with non-trivial logic (here: a `filter` + nullish coalescing across two lines) should be extracted into a private initializer, matching the existing `initGroup()` pattern. The neighboring single-line computeds (`hasLinks`, `hasUpcomingMeetings`) are simple enough to stay inline; this one crosses the line.

```suggestion
  protected readonly safeExternalSources: Signal<PublicGroupExternalSource[]> = this.initSafeExternalSources();

  protected readonly hasExternalSources = computed(() => this.safeExternalSources().length > 0);
```

…with a `private initSafeExternalSources(): Signal<PublicGroupExternalSource[]>` grouped at the bottom of the class alongside `initGroup()`.

audigregorie (commented):
## Code Review Summary

Solid, well-scoped change with strong security hygiene — `safeExternalSources` correctly filters every upstream URL through the shared `isValidUrl()` before binding to `[href]`, and the type consolidation (`PublicGroupExternalSource` → alias of `CommitteeExternalSource`) eliminates a duplicated interface. The main blocker is the `track $index` / PR-description mismatch; the rest is minor polish.

### What's done well

- **Security-first URL handling** — reuses the same `isValidUrl` validator as `activity-feed.utils.ts`; untrusted OCG URLs never reach the DOM unvalidated.
- **Clean type consolidation** — the duplicated `PublicGroupExternalSource` interface becomes a type alias of the canonical `CommitteeExternalSource`, eliminating drift and correctly extending the contract.
- **Faithful pattern reuse** — the External Sources section mirrors the existing Links section's structure and styling.

### Rating: ⭐⭐⭐⭐ (4/5)

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
MRashad26: Thanks for flagging the mismatch — the PR description was stale, not the code. `track $index` is intentional: it was the result of an earlier review round on this same line fixing a real NG0955 risk from a non-unique-key scenario, and Copilot's most recent pass on this exact line independently confirmed `$index` is correct and that it's the *description* that needs updating, not the code. `source.url` uniqueness isn't guaranteed by the contract (nothing in `CommitteeExternalSource` enforces it), so switching back would reintroduce that risk. I've updated the PR description's Trade-offs section (https://github.com/linuxfoundation/lfx-self-serve/pull/1357) to reflect `$index` instead of the superseded `source.url` plan — appreciate the catch, this should've been kept in sync.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
MRashad26: Good catch — dropped the `@if` in 2998aa855. `CommitteeExternalSource.provider` is currently a `'ocg'` literal type (not a union), so the badge always rendering was correct behavior; the conditional just implied a possibility that doesn't exist yet. If a non-OCG provider is added later, the type will need widening and the `@if` can come back at that point.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.html
MRashad26: Fixed in 2998aa855 — data-testid is now `external-source-<entity_type>-<index>`, so E2E selectors stay unique when multiple sources share an entity_type. Good catch.

apps/lfx-one/src/app/modules/groups/group-detail/group-detail.component.ts
MRashad26: Fixed in 2998aa855 — extracted into a private `initSafeExternalSources()` per component-organization.md §3, matching the `initGroup()` pattern.

MRashad26:

Commit: 2998aa855

### Changes Made
- **`group-detail.component.html`**: appended the loop index to each external source's `data-testid` (`external-source-<entity_type>-<index>`) so E2E selectors stay unique when multiple sources share an `entity_type`. (per @audigregorie)
- **`group-detail.component.html`**: dropped the always-true `@if` around the "Powered by OCG" badge — `CommitteeExternalSource.provider` is currently a `'ocg'` literal type, so the badge was always rendering regardless. (per @audigregorie)
- **`group-detail.component.ts`**: extracted `safeExternalSources` into a private `initSafeExternalSources()` per `component-organization.md` §3, matching the existing `initGroup()` pattern. (per @audigregorie)

### No Code Change — PR Description Updated Instead
- **`group-detail.component.html:185` (`track $index`)**: kept `track $index` rather than switching to `track source.url`. This line was already fixed to `$index` in an earlier review round (a real NG0955 risk from non-guaranteed URL uniqueness), and Copilot's most recent pass on the same line independently confirmed `$index` is correct — the PR description was just stale. Updated the description's Trade-offs section to match current code instead of reverting the fix.

### Threads Resolved

audigregorie (approved):

PR #1361, opened 2026-08-08
feat(weekly-brief): thumbs up/down quality rating

manishdixitlfx:
## Storage decision (read this first)

The ticket assumes Supabase is a live datastore. It isn't: `supabase.service.ts` is an explicit stub (`getUser`/`getUserEmails` always return null/empty), and there's no `pg`/`knex`/`prisma`/any DB client anywhere in `apps/lfx-one/src/server` or `package.json`.

**What this PR does instead** — splits responsibility across two systems that are already real in this repo:

- **Valkey** owns only the live, per-user *current-vote* lookup (upsert/clear/single-vote-per-revision), keyed on `(brief_uid, revision, username)`, 90-day TTL. This is a narrow, cache-appropriate job Valkey is already used for elsewhere (session storage, org-lens caching). A fail-soft miss here only degrades the "you already rated this" UI indicator on a future load — never the analytics record.
- **Structured logging** (`logger.info` → the existing Pino → CloudWatch pipeline) is the durable analytics substrate. Every rate/clear emits a `rating_recorded`/`rating_cleared` event carrying `brief_uid`, `committee_uid`, `revision`, `client_revision` (what the rater actually saw — see below), `prompt_version`, `model`, `username`, `previous_rating`, and `rating_cache_enabled`. "Rating-by-prompt_version" trend analysis runs via CloudWatch Insights — no new query capability was built into Valkey (it has none exposed today), no secondary index to keep in sync.

**Rejected alternatives:**
- *Valkey-only* — fail-soft + no query capability is a real risk for a signal a GA rollout decision depends on, with no fallback source if a write silently no-ops.
- *Real Postgres/Supabase client* — the architecturally "correct" long-term shape, but a new deployed-datastore decision (secrets, migrations, backup policy) that shouldn't be decided unilaterally inside a UI-rating ticket.

This was reviewed across 5 rounds of full-branch sweeps (`lfx-general-code-reviewer` + `lfx-self-serve-code-reviewer` + `lfx-self-serve-learnings-reviewer`, in addition to 3 post-commit rounds) — see **Accepted trade-offs** below for the points raised repeatedly and consciously left as-is.

## What changed

One-tap thumbs up/down rating on generated/edited/approved weekly briefs (LFXV2-3042), BFF-only per ticket scope — no upstream/committee-service change.

- **Shared**: `WeeklyBriefRating`, `RateWeeklyBriefRequest`/`RateWeeklyBriefResponse`, `caller_rating` enrichment on `WeeklyBriefCurrentResponse`, `VALKEY_CACHE.WEEKLY_BRIEF_RATING_{NAMESPACE,TTL_SECONDS}`.
- **BFF**: `POST`/`DELETE /api/committees/:committeeId/weekly-briefs/:briefUid/rating`, gated by `assertCommitteeRead` (a rating is a personal reaction, not a committee edit — same gate as `getCurrentBrief`/`shareBrief`) and `blockDuringImpersonation` (rating resolves the *impersonated* user's identity for the Valkey write itself, unlike generate/save/share which proxy the bearer token through untouched).
- **UI**: thumbs control on `weekly-brief-card`, optimistic update keyed to `(briefUid, revision)` with rollback on failure, disabled while impersonating, visible only in `generated`/`edited`/`approved` states.
- **Tests**: 73 server unit tests (rate → re-rate → clear round-trip, revision-scoping, 404/401/400 paths, impersonation, persist-failure logging, corrupt-cache-entry handling) + 4 new Playwright e2e cases (this repo has no Angular component-test runner, so e2e is the real coverage surface for the frontend half).

## Accepted trade-offs (raised in review, left as-is — flagging so a reviewer doesn't re-litigate blind)

- **`username` (not `sub`) in `rating_recorded`/`rating_cleared` logs** — raised in 4 of the review rounds. Kept as `username` (LFID) because: (1) `getEffectiveSub` is `@deprecated` repo-wide in favor of username; (2) `WeeklyBrief.last_edited_by` already persists username indefinitely as a first-class attribution field on this exact resource; (3) it's the established convention for logged identifiers elsewhere (`committee.service.ts`, `project.controller.ts`, `invite.controller.ts`). A salted hash would need secret-management infrastructure disproportionate to a v1 feature.
- **Branch name** (`feat/LFXV2-3042-brief-rating`, not the checklist's bare `feat/LFXV2-3042`) — matches all 30+ sibling worktree branches in this fleet; renaming would be the one branch out of step with established practice, not the other way around.
- **`impersonation-readonly.middleware.ts` touched** (protected-file hook flag) — comment-only (generalizing the guard's rationale to cover the new write direction) plus one log message string (`'Blocked profile write...'` → `'Blocked write...'`); `blockDuringImpersonation`'s logic is unchanged. Flagging for code owner review per the guard hook.
- **Diff size** (+1058, over the 1000-line target) — single cohesive feature plus fixes from 5 rounds of full-branch review folded in; splitting further wasn't recommended by PR-readiness given the state.

## Test plan

- [x] `/lfx-self-serve-pr-readiness` — READY WITH CHANGES (no CRITICAL findings; two SHOULD_FIX items addressed above)
- [ ] Manual verification in the dev server (mock `WEEKLY_BRIEF_BACKEND`): tap thumbs-up, reload (confirm pre-lit from `caller_rating`), tap thumbs-down (confirm switch), tap again (confirm clear), regenerate (confirm new revision starts unrated) — not run in this session; per this repo's parallel-worktree workflow, browser validation happens in the integration session, not the per-ticket session.


apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -191,15 +201,123 @@ export class WeeklyBriefService {
...
+   * value, so re-rates/switches can be deduplicated per caller during offline analysis instead of
+   * each toggle over-counting as a distinct vote), which flows through the existing Pino →
+   * CloudWatch pipeline.
+   */
+  public async rateBrief(
+    req: Request,
+    committeeId: string,
+    briefUid: string,
+    rating: WeeklyBriefRating,
+    clientRevision: number
```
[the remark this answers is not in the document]
manishdixitlfx: Done — rateBrief and clearBriefRating now reject with a 409 (`REVISION_MISMATCH`) when the caller-supplied revision no longer matches the server-resolved current revision, instead of only logging the drift. clearBriefRating now also requires `revision` in its request body (previously took none), matching rateBrief's shape. The UI (`onRate`) treats the 409 the same way it already treats a 404: refresh the card and prompt the user to rate again, rather than retrying against stale content.

See commit 9879c9dff: weekly-brief.service.ts's `resolveRatableBrief` now takes `expectedRevision` and throws `ConflictError` on mismatch (same code/shape `shareBrief` already uses); `weekly-brief.controller.ts`'s new `validateClearRatingBody` narrows the DELETE body; `weekly-brief-card.component.ts`'s `onRate` error handler now checks `err.status === 409` alongside 404.

apps/lfx-one/src/server/services/valkey.service.ts
```diff
@@ -319,6 +319,17 @@ export function buildMemberV1MappingCacheKey(memberUid: string): string | null {
...
 
+/**
+ * Per-caller weekly-brief rating cache key (LFXV2-3042) — one entry per (brief_uid, revision,
+ * username), so a new revision (post-regenerate) never inherits the prior revision's rating.
+ * Null (fail-closed → skip cache) when the brief uid isn't filter-safe or the username isn't
+ * filter-safe, so neither can corrupt the `:`-delimited key.
+ */
+export function buildWeeklyBriefRatingCacheKey(briefUid: string, revision: number, username: string): string | null {
+  if (!isFilterSafeIdentifier(briefUid) || !isFilterSafeUsername(username)) return null;
+  return `${keyPrefix()}:${VALKEY_CACHE.WEEKLY_BRIEF_RATING_NAMESPACE}:${briefUid}:${revision}:${username}`;
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch — fixed. `buildWeeklyBriefRatingCacheKey` now includes `committee_uid` as part of the key. You're right that mock mode's hard-coded brief uid + shared starting revision made this a real cross-committee leak in every local dev/CI run, not just a theoretical live-mode edge case.

See commit 9879c9dff: the key is now `...:{committee_uid}:{brief_uid}:{revision}:{username}`, threaded through from `withCallerRating`/`requireRatingKey` (both already had `committeeId` in scope). Added a regression test proving two committees sharing the same mock brief uid+revision no longer collide.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -191,15 +201,123 @@ export class WeeklyBriefService {
...
+  ): Promise<RateWeeklyBriefResponse> {
+    const { brief, callerRating: previousRating } = await this.resolveRatableBrief(req, committeeId, briefUid, 'rate_weekly_brief');
+    const username = this.requireUsername(req, 'rate_weekly_brief');
+    const key = this.requireRatingKey(brief.uid, brief.revision, username, 'rate_weekly_brief');
+
+    const persisted = await valkeyService.setJson(key, { rating }, VALKEY_CACHE.WEEKLY_BRIEF_RATING_TTL_SECONDS);
```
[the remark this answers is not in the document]
manishdixitlfx: Thanks for flagging this — I can see why it looks like a gap when read against a typical "the API response should reflect durable state" expectation.

This is actually intentional, and documented at length in this PR's description under "Storage decision": Valkey here is deliberately scoped to the live per-user *current-vote* dedup/UI-indicator only (90-day TTL, fail-soft), not the system of record. The durable analytics record — the thing the ticket's GA rollout decision actually depends on (rating-by-prompt_version trend analysis) — is the structured `rating_recorded`/`rating_cleared` log line, which flows through the existing Pino → CloudWatch pipeline regardless of whether the Valkey write succeeds.

Why not require a durable write before returning success: there's no real DB/Postgres client anywhere in this repo's server (`supabase.service.ts` is an explicit stub — verified via grep for `new Pool`/`createClient(`/`knex`/`prisma`, none found), so "require a successful durable write" would mean either standing up a new datastore (a bigger, separate infra decision that a UI-rating ticket shouldn't make unilaterally) or making Valkey a hard runtime requirement for this one endpoint, which contradicts how Valkey is used everywhere else in this codebase (session storage, org-lens caching — all fail-soft, none a hard requirement).

The functional cost of a Valkey outage/miss here is real but bounded: re-rate/clear semantics stay correct (a blind upsert, no read-modify-write coordination needed), the caller just doesn't see "you already rated this" pre-lit on the next load, and `rating_persist_failed" is logged as a warning whenever the cache is enabled but a specific write faults (suppressed only for the disabled-cache case, to avoid alert-fatigue on a documented supported deployment mode).

Happy to discuss further if you think the analytics-in-logs approach should change — but as designed, this isn't an oversight.

manishdixitlfx:
## Review Feedback Addressed

Commit: 9879c9dff1c56e502ad628ddce8d842c0a2f2c96

### Changes Made
- **weekly-brief.service.ts**: `rateBrief`/`clearBriefRating` now reject with a 409 (`REVISION_MISMATCH`) when the caller-supplied revision doesn't match the server-resolved current revision, instead of only logging the drift (per copilot-pull-request-reviewer)
- **weekly-brief.controller.ts**: `clearBriefRating` (DELETE) now requires and validates a `revision` body field (per copilot-pull-request-reviewer)
- **valkey.service.ts**: `buildWeeklyBriefRatingCacheKey` now includes `committee_uid`, closing a cross-committee rating leak in mock/dev mode (per copilot-pull-request-reviewer)
- **weekly-brief-card.component.ts**: `onRate`'s error handler now treats a 409 the same as a 404 — refresh and prompt to rate again, rather than retrying against stale content
- Added regression tests for both fixes (revision-mismatch rejection on rate and clear; cross-committee cache isolation)

### No Change Needed
- **weekly-brief.service.ts** (Valkey fallback/durability): responded explaining this is intentional, not an oversight — Valkey is a fail-soft dedup/UI-indicator cache only; the durable analytics record is the structured `rating_recorded`/`rating_cleared` log line (see the PR description's "Storage decision" section for the full reasoning, including why a real DB client isn't in scope here) (raised by copilot-pull-request-reviewer)

### Threads Resolved
2 of 3 unresolved threads addressed in this iteration.

### Still Open
- **weekly-brief.service.ts:259** (Valkey fallback): left open pending reviewer response to the design-intent explanation above — this is a discussion point, not something code alone resolves.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -191,15 +201,117 @@ export class WeeklyBriefService {
...
+        revision: brief.revision,
+      });
     }
+    logger.info(req, 'rating_recorded', 'Weekly brief rating recorded', {
+      committee_id: committeeId,
+      brief_uid: brief.uid,
+      revision: brief.revision,
+      prompt_version: brief.prompt_version,
+      model: brief.model,
+      username,
```
dealako: [minor] Username persisted as raw PII in durable analytics log

Issue: `rating_recorded` (here) and `rating_cleared` (line ~311) log the caller's raw LFID `username` into the structured-log stream that flows to CloudWatch/Snowflake indefinitely.
Proof: `docs/reviews/knowledge-base/security.md:121-132` classifies username/email in structured logs as PII ("Important") and recommends an opaque OIDC `sub` or salted hash instead — the logger's redaction doesn't strip this nested field. This is also the subject of Copilot's still-open comment on this PR, which hasn't been replied to or fixed (unlike the other two Copilot findings, both closed out in commit 9879c9df).
Why it matters: an indefinitely-retained PII field in the analytics substrate this PR explicitly designed to be the durable system of record for GA rollout decisions.
Fix: swap `username` for a stable opaque identifier (OIDC `sub` or a salted hash) in both log calls, or explicitly document why raw username is acceptable here (citing the KB) and resolve Copilot's thread with that rationale.

packages/shared/src/interfaces/weekly-brief.interface.ts
```diff
@@ -75,6 +75,40 @@ export interface WeeklyBriefCurrentResponse {
...
+  caller_rating?: WeeklyBriefRating | null;
+}
+
+/** A caller's one-tap quality rating on a specific weekly-brief revision. BFF-only — no upstream equivalent. */
+export type WeeklyBriefRating = 'up' | 'down';
+
+/**
+ * Request body for `POST /committees/:committeeId/weekly-briefs/:briefUid/rating`. `revision` is
+ * the revision the caller actually saw when they tapped — the server always rates whatever
+ * revision is current at write time (see `weekly-brief.service.ts#rateBrief`'s doc comment), so
```
dealako: [minor] Interface JSDoc contradicts the implemented 409 behavior

Issue: this comment states revision "is never used to reject the write" and that "the server always rates whatever revision is current at write time."
Proof: `weekly-brief.service.ts`'s `resolveRatableBrief` (~line 703) now throws `ConflictError`/`REVISION_MISMATCH` when `brief.revision !== expectedRevision` — added in the last review round per this PR's own comment thread — but this shared-interface doc wasn't updated to match. The service's own doc comment on `rateBrief` correctly describes the 409 behavior, making this the one stale copy.
Why it matters: this is the interface most likely to be read by a future API consumer/maintainer trying to understand the write contract; as written it describes the opposite of what actually happens.
Fix: update this comment to describe the 409 REVISION_MISMATCH rejection, matching `weekly-brief.service.ts`'s doc comment on `rateBrief`.

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
```diff
@@ -87,6 +87,46 @@ function validateShareBriefBody(body: unknown): { ok: true; value: { revision: n
...
+ * and reasoning as `validateShareBriefBody`: the caller must send back the revision they actually
+ * saw so the service can reject a stale clear (PR #1361 review) instead of silently deleting
+ * whatever revision happens to be current.
+ */
+function validateClearRatingBody(body: unknown): { ok: true; value: { revision: number } } | { ok: false; fieldErrors: Record<string, string> } {
+  if (!body || typeof body !== 'object') {
+    return { ok: false, fieldErrors: { body: 'Request body must be a JSON object' } };
+  }
+  const b = body as Record<string, unknown>;
+  if (typeof b['revision'] !== 'number' || !Number.isFinite(b['revision'] as number)) {
```
dealako: [minor] Clear-rating revision validation is looser than rate's

Issue: `validateClearRatingBody` only checks `Number.isFinite(revision)`, while `validateRateBriefBody` (line 104) requires `Number.isInteger(revision) && revision >= 1`. A DELETE with `revision: -1` or `revision: 0.5` passes validation here but the equivalent POST would be rejected.
Proof: compare lines 104-106 (rate) vs 124-126 (clear) in this file.
Why it matters: not currently exploitable (no Valkey key will match a non-integer/non-positive revision, so it just falls through to the not-found/mismatch path), but it's an inconsistent boundary contract between two closely related endpoints that future changes could rely on differently.
Fix: align `validateClearRatingBody`'s revision check with `validateRateBriefBody`'s (`Number.isInteger` and `>= 1`, same error message pattern).

apps/lfx-one/e2e/weekly-brief-card.spec.ts
```diff
@@ -921,3 +931,105 @@ test.describe('WG Weekly Brief card — Regenerate disabled (throttle exhausted)
...
+
+    await page.goto(COMMITTEE_URL, { waitUntil: 'domcontentloaded' });
+    await expect(page).not.toHaveURL(/auth0\.com/);
+    await expect(page.getByTestId('committee-overview-weekly-brief-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+
+    const upBtn = page.getByRole('button', { name: 'Rate this brief helpful' });
+    await upBtn.click();
+
+    await expect(page.getByText('Failed to save your rating. Please try again.', { exact: true })).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(page.getByRole('button', { name: 'Rate this brief helpful', pressed: false })).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
```
dealako, line 1033: [nit] No Playwright coverage for the 409 revision-mismatch path

Issue: the 4 new rating tests cover pre-lit rating, POST rate, DELETE clear, and 500-error rollback, but none simulates a concurrent co-chair edit/regenerate landing between page load and the rating tap — the scenario the revision-conflict design in this PR is built to handle.
Proof: server-side 409 rejection is unit-tested in `weekly-brief.service.spec.ts:491-519`, but the full-stack flow (mocked 409 response → error toast → `refresh$` reload) isn't exercised end-to-end here.
Why it matters: this is the one behavior this PR added specifically to close a data-integrity gap (rating attributed to unreviewed content) — worth proving the UI recovery path actually works, not just the server rejection.
Fix: add a Playwright case that mocks a 409 on the rate/clear request and asserts the card shows the refresh/error state and reloads the current brief.

dealako (changes requested):
@manishdixitlfx thanks for the detailed writeup and for the transparency on the Valkey-vs-Postgres tradeoff — the doc comments make the design intent easy to follow, and the revision-conflict / cache-key fixes from the last round (commit 9879c9df) look solid. Scope is tight and well-tested overall (unit coverage for the 409 paths, 4 new Playwright cases).

Ran a 3-way parallel review (security/privacy, correctness/perf/tests, style/docs). Nothing new or severe — one open item carries over from Copilot's unresolved comment, plus two small drift/consistency items.

### 🔴 Blocking: 0

### 🟡 Minor: 3
- **Username still logged raw in `rating_recorded`/`rating_cleared`** (`weekly-brief.service.ts:264`, `:311`) — agrees with Copilot's still-open comment on this PR; the repo's own `docs/reviews/knowledge-base/security.md:121-132` flags username-in-structured-logs as PII and recommends an opaque `sub`/salted hash. Worth either fixing or explicitly closing out (with a KB-citing rationale) before merge, since it's the one open thread from the prior round.
- **`RateWeeklyBriefRequest` JSDoc contradicts the actual 409 behavior** (`packages/shared/src/interfaces/weekly-brief.interface.ts:91-97`) — the shared-interface comment still says revision "is never used to reject the write," but `resolveRatableBrief` (`weekly-brief.service.ts:703-708`) now throws `ConflictError`/`REVISION_MISMATCH` when it doesn't match. This interface doc is the one place that didn't get updated when the 409 behavior was added.
- **Clear-rating validates `revision` more loosely than rate** (`weekly-brief.controller.ts:124` vs `:104-106`) — `validateClearRatingBody` only checks `Number.isFinite`, so `0`, negative, or fractional revisions pass DELETE but would fail POST. Not exploitable (Valkey/service layer won't have a matching key), but the asymmetry is worth closing for consistency.

### ⚪ Nit: 1
- Missing Playwright coverage for the 409 revision-mismatch path itself (`e2e/weekly-brief-card.spec.ts`) — server-side is covered in `weekly-brief.service.spec.ts:491-519`, but the full stack (mocked 409 → toast → `refresh$`) isn't exercised end-to-end.

### ❔ Question: 0

Given the one carried-over open security/privacy item, I'm marking Changes Requested rather than Approved-with-comments — but all three findings are minor/nit, not architecture-level, so this should be a quick follow-up commit.

apps/lfx-one/src/server/services/weekly-brief.service.ts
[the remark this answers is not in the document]
manishdixitlfx: Done — both `rating_recorded` and `rating_cleared` now log `user_id: getEffectiveSub(req)` (opaque OIDC sub) instead of the raw LFID username. The Valkey cache key itself is unchanged — still keyed on username, matching this repo's existing per-user cache-key convention (e.g. `buildUserCacheKey`) — only the durable, indefinitely-retained log record changes.

See commit f349a0661: weekly-brief.service.ts.

manishdixitlfx: Good catch, and fair push — I'd treated this as a documented trade-off after the first review round, but you're right that the KB rates it Important with a directly-on-point precedent (PR #280), and this analytics stream is specifically designed for indefinite retention, which is a materially different risk than the ephemeral debug-level username logging elsewhere in this repo. Fixed rather than re-argued.

Both log calls now use `user_id: getEffectiveSub(req)` (opaque OIDC sub) instead of `username`. Left the Valkey cache key itself on username, unchanged — that's a short-TTL cache key, not the durable record, and matches this repo's existing per-user cache-key convention.

See commit f349a0661.

packages/shared/src/interfaces/weekly-brief.interface.ts
manishdixitlfx: Fixed — good catch. Updated the comment to describe the 409 REVISION_MISMATCH rejection, matching `weekly-brief.service.ts`'s already-correct doc comment on `rateBrief`.

See commit f349a0661: packages/shared/src/interfaces/weekly-brief.interface.ts.

apps/lfx-one/src/server/controllers/weekly-brief.controller.ts
manishdixitlfx: Fixed — aligned `validateClearRatingBody`'s revision check with `validateRateBriefBody`'s (`Number.isInteger` and `>= 1`, same error message pattern). You're right the looser `validateShareBriefBody`-derived check was the wrong template to copy from here — rate and clear are the same closed pair of endpoints and should share one boundary contract.

See commit f349a0661: weekly-brief.controller.ts, plus two new tests covering the non-integer and <1 cases.

apps/lfx-one/e2e/weekly-brief-card.spec.ts
manishdixitlfx: Added — good call, this is the one behavior this PR added specifically to close a data-integrity gap, worth proving end-to-end and not just at the unit level.

New Playwright test mocks a 409 REVISION_MISMATCH on the rate request (simulating a co-chair's edit landing between page load and the tap), then asserts: the reload-specific toast appears (not the generic 500-path message), a second GET /current fires via `refresh$`, and the card ends up showing the newly-current (edited) brief with the thumb reset to unpressed.

See commit f349a0661: e2e/weekly-brief-card.spec.ts.

manishdixitlfx:

Commit: f349a06611bc56929dd16ada9a09e08434d21c9e

### Changes Made
- **weekly-brief.service.ts**: `rating_recorded`/`rating_cleared` now log an opaque `user_id` (OIDC `sub`) instead of the raw LFID username — the Valkey cache key is unchanged, still keyed on username (per @dealako, copilot-pull-request-reviewer). This reverses my prior "documented trade-off" stance from round 1; the KB citation + empirical precedent (PR #280) made the case convincingly.
- **weekly-brief.interface.ts**: fixed `RateWeeklyBriefRequest`'s doc comment, which still described revision as "never used to reject the write" — stale since round 1 added the 409 rejection (per @dealako)
- **weekly-brief.controller.ts**: `validateClearRatingBody` now requires an integer `>= 1`, matching `validateRateBriefBody` exactly, instead of a looser finite-number check (per @dealako)
- **weekly-brief-card.spec.ts** (e2e): added a test covering the 409 REVISION_MISMATCH path — mocked 409 response → error toast → `refresh$`-driven reload → card reflects the new current brief (per @dealako)

### Threads Resolved
5 of 6 unresolved threads addressed in this iteration (2 PII threads, 1 stale-doc, 1 validation-consistency, 1 e2e-coverage).

### Still Open
- **weekly-brief.service.ts** (Valkey fallback/durability, copilot-pull-request-reviewer): unchanged from round 1 — replied explaining the Valkey-is-a-dedup-cache-not-the-system-of-record design is intentional. Left open for further discussion if the reviewer disagrees.

apps/lfx-one/src/server/services/weekly-brief.service.ts
```diff
@@ -191,15 +201,125 @@ export class WeeklyBriefService {
...
+      });
     }
+    logger.info(req, 'rating_recorded', 'Weekly brief rating recorded', {
+      committee_id: committeeId,
+      brief_uid: brief.uid,
+      revision: brief.revision,
+      prompt_version: brief.prompt_version,
+      model: brief.model,
+      // Opaque OIDC sub, not the LFID username — `username` (used above for the Valkey key,
+      // where per-user cache keys already convention on it — see buildUserCacheKey) is
```
dealako, line 266: [nit] Slightly awkward phrasing

Issue: "where per-user cache keys already convention on it" reads oddly — "convention" used as a verb.
Fix: something like "where per-user cache keys already follow that convention" would read more naturally. Non-blocking.

dealako (approved):
@manishdixitlfx nice turnaround — all three of my open items from the last round are closed out cleanly, and the fixes are well-targeted (no scope creep).

### 👏 Nice work
- `user_id: getEffectiveSub(req)` swap is exactly right: opaque OIDC `sub` in the durable log stream, `username` still used for the short-TTL Valkey cache key (intentional, and now called out explicitly in the comment) — matches the KB guidance and PR #280's precedent cited in the commit message.
- The interface doc fix for `RateWeeklyBriefRequest` now matches `rateBrief`'s own doc comment and the actual 409 behavior — no more contradiction.
- `validateClearRatingBody` now shares the exact same bound as `validateRateBriefBody` (`Number.isInteger`, `>= 1`), with new boundary tests (`1.5`, `0`) in the controller spec.
- The new e2e test doesn't just assert the 409 happens — it waits for the `refresh$`-driven GET, checks the specific reload toast (distinct from the generic error toast), and confirms the card reflects the newly-current brief. That's the exact recovery path this feature exists to protect.

### Revision tracking (carried over from last review)
- ✅ **Resolved** — raw username in `rating_recorded`/`rating_cleared` logs → now `user_id` (opaque sub)
- ✅ **Resolved** — `RateWeeklyBriefRequest` doc contradicting the 409 behavior
- ✅ **Resolved** — `validateClearRatingBody` looser than `validateRateBriefBody`
- ✅ **Resolved** — missing Playwright coverage for the 409 recovery flow
- ⚠️ **Still open (not mine, no action needed from this review)** — Copilot's Valkey fail-soft/no-durable-fallback thread from the prior round; your commit message notes it's pending further discussion and this commit doesn't touch it, which matches what I see in the diff.

### 🔴 Blocking: 0
### 🟡 Minor: 0
### ⚪ Nit: 1 — slightly awkward phrasing in the new inline comment ("already convention on it") in `weekly-brief.service.ts`

### Final decision: ✅ Approved with minor comments

Approving — the one open item is Copilot's separate design-discussion thread, not something I'm blocking on.

PR #1373, opened 2026-08-10
test(lfx-one): add app-side unit test harness, lint specs (LFXV2-3074)

mrautela365:
## What

`apps/lfx-one` could run unit tests only against `src/server/**` — 43 spec files, 706 tests, and **zero** app-side specs. `vitest.config.ts` said so in a comment: there was no Angular test builder wired up, so component and template specs were simply not supported. This adds the missing half.

App specs now run through the **`@angular/build:unit-test`** builder (`yarn test:app`), and `yarn test` runs both halves. CI needed no change: `quality-check.yml` already runs `yarn test`, which turbo forwards to the package script.

## Why this builder

It ships with Angular, so its version can never drift from the framework's, and it reuses the application build (`buildTarget: lfx-one:build:development`) — which is what makes `templateUrl`, `styleUrls`, and the `@lfx-one/shared/*` aliases resolve in a spec exactly as they do in the app. Nothing about the module graph is re-described in a second config, so nothing about it can drift. `browsers` is omitted deliberately, which selects jsdom on Node: no browser download for unit tests, and real-browser coverage stays the E2E suite's job.

The app is zoneless and zone.js is not a dependency, so `provideZonelessChangeDetection()` is installed once for every spec via `providersFile: src/test-providers.ts`. A spec cannot forget it, and no spec repeats it.

## The two sample specs

Not filler — each exists to prove a mechanism the harness has to support:

- `filter-pills.component.spec.ts` (5 tests) — external `templateUrl`, `@for` control flow, a PrimeNG directive, a signal `input()` set via `setInput`, an `output()`, and zoneless DOM assertions via `await fixture.whenStable()`.
- `http-error.utils.spec.ts` (9 tests) — pure retry/classification policy, including a counting source that proves `retryTransientHttpError(2)` yields exactly 3 attempts and that a 401 stops at 1.

**Both were verified by mutation.** A green test proves nothing on its own, so each assertion was pinned by breaking the code it covers and confirming the intended test — and only that test — fails. The full table is in the new doc. One mutation initially "passed" because the `sed` pattern matched no line; that hazard (a mutation that silently doesn't apply looks exactly like a non-binding test) is now written down too.

## Specs are no longer exempt from lint

`**/*.spec.ts` sat in `eslint.config.js`'s ignore list, which meant every spec in the repo was exempt from the rules the code it tests must follow. That is the last place an exemption belongs — a fake that quietly breaks a rule is how a fake stops modelling the real thing. Removing it cost 6 problems across 45 files, all fixed here (`Array<T>` → `T[]`, three `public` modifiers, two unused loop bindings). `vite.config.ts` / `vitest.config.ts` are ignored explicitly instead, because they sit outside `tsconfig.app.json` and lint-staged would otherwise fail a commit on a parser error rather than a finding.

## Note on the ticket's acceptance criterion

LFXV2-3074 names `implementation-tab.component.spec.ts` as the proof spec. That component lands with #1364, which is still open — this PR provides the harness and two specs that exercise the same mechanisms; the implementation-tab spec follows once #1364 merges.

## Verification

- `yarn test` green across all three turbo tasks: shared 33/33, server 41/41, app 14/14
- `yarn lint:check` 0 errors · `prettier --check` clean · markdownlint 0 issues

Ticket: [LFXV2-3074](https://linuxfoundation.atlassian.net/browse/LFXV2-3074)

[LFXV2-3074]: https://linuxfoundation.atlassian.net/browse/LFXV2-3074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrautela365:
All three findings addressed in `66d736c3` + `311a2348`.

**1. Vitest major mismatch (`apps/lfx-one` on 4, builder declares 3).** Real, and pinned down rather than up. `@angular/build:unit-test` does not bundle Vitest — it imports Vitest's Node API from the workspace — so the declared range is a genuine compatibility constraint, not a formality. `@angular/build@20.3.13` declares `"vitest": "^3.1.1"`; support for `^4.0.8` first appears in `@angular/build@21.x`, which is an Angular major upgrade and not something a test harness gets to drag in. So `apps/lfx-one` is now `vitest: ^3.2.4`.

The peer dep is marked **optional**, which is why nothing failed the install and why this is now written down in `docs/architecture/testing/unit-testing.md`. Both halves report identical counts on either runner — server **41 files / 673 tests**, app **2 files / 14 tests** — so the previously green run on v4 was evidence that the parts those specs touch happen to line up, not evidence of compatibility.

`packages/shared` deliberately stays on `^4`: it runs plain `vitest run` with no Angular builder in the path, and the two workspaces resolve independently. The doc carries the note to raise the app pin back to `^4` in the same change that moves to Angular 21.

The downgrade surfaced one genuine type break, fixed in `311a2348`: `committee.service.spec.ts:178` declared its spy as `ReturnType<typeof vi.spyOn>`, which under Vitest 3 resolves to the unparameterised `MockInstance<(this: unknown, ...args: unknown[]) => unknown>` — the real spy is not assignable to it. Now typed from the method itself, `MockInstance<CommitteeService['getCommitteeById']>`. `tsc -p tsconfig.json --noEmit` and the full build are clean.

**2. `querySelectorAll('button')` doesn't pin the pill contract.** Fixed at `filter-pills.component.spec.ts:42` — now `querySelectorAll('[data-testid^="filter-pill-"]')`, matching the `data-testid` convention the rest of the spec (and the E2E suite) uses. Mutation-verified: blanking `[attr.data-testid]` in the template fails all 5 tests with `no pill rendered for all`; template restored, `git diff --stat` empty.

**3. "the builder resolves to the same version as the framework".** Correct — that claim was false and is now stated accurately: `@angular/build` is 20.3.13 here against a 20.3.15 framework, and they are independent packages. What the builder carries is the framework's compatibility contract, not version equality.

apps/lfx-one/scripts/run-tests.mjs
```diff
@@ -0,0 +1,33 @@
...
+import { spawnSync } from 'node:child_process';
+
+/** Runs one half and returns its exit code, treating a signal kill as a failure. */
+function runHalf(script) {
+  const { status, signal } = spawnSync('yarn', [script], { stdio: 'inherit', shell: false });
+  if (signal) {
+    console.error(`\n${script} was killed by ${signal}`);
+    return 1;
+  }
+  return status ?? 1;
```
dealako, line 27: Agreeing with this — confirmed still live against HEAD (`311a2348`). `spawnSync('yarn', [script], { shell: false })` will fail on Windows because Corepack exposes `yarn` as a `.cmd` shim, and Node's non-shell `spawnSync` can't invoke that without going through a shell — both halves would report failure without ever running, which reads as "tests failed" rather than "couldn't launch the runner."

Not release-blocking: CI here runs `ubuntu-latest` only, so this doesn't affect the pipeline — only a Windows-based contributor running `yarn test` locally. Marking [minor] rather than [blocking] on that basis. Fix per Copilot's suggestion: pass `shell: process.platform === 'win32'` (or invoke via `cmd.exe /c` on Windows) while keeping direct exec elsewhere.

dealako:
Hey @mrautela365 — reviewed this one. Nice work standing up the app-side unit test harness for `lfx-one`; the split between server (Vitest/Node) and app (`ng test`/jsdom) specs is clean, the zoneless `provideZonelessChangeDetection()` wiring in `test-providers.ts` is exactly right for this codebase, and the sample specs (`filter-pills`, `http-error.utils`) are solid references for future test authors. The `run-tests.mjs` rationale doc-comment explaining why this isn't a `package.json` one-liner (Yarn's portable shell has no `$?`/`$(( ))`) is a good call — that's exactly the kind of non-obvious WHY worth keeping in-code.

I ran three parallel reviewers across Security/Data-Privacy, Correctness/Performance/Test-Coverage, and Style/API/Documentation — all three came back clean, no findings.

**Reconciling with the existing bot passes:** Copilot and Cursor Bot both flagged real issues earlier in the thread — the Vitest major-version mismatch against the Angular builder's declared range, and the unsupported `$(( ))` arithmetic in the original `package.json` test script. Both are fixed as of `66d736c3`/`311a2348` (pin to `^3.2.4`, rewrite as a Node script) — agreed those are resolved.

One item is still open though: Copilot's later comment on `run-tests.mjs` (posted after your "all three addressed" reply, anchored to current HEAD `311a2348`) — `spawnSync('yarn', [script], { shell: false })` will fail to launch on Windows, since Corepack exposes `yarn` as a `.cmd` shim that Node's non-shell `spawnSync` can't invoke. I've replied inline agreeing with it and confirmed it's still live against the current code.

### Summary

- 🔴 Blocking: 0
- 🟡 Minor: 1 — `run-tests.mjs` `spawnSync` won't launch `yarn` on Windows (Corepack `.cmd` shim + `shell: false`); doesn't affect CI (ubuntu-latest only), only local `yarn test` on a Windows dev machine
- ⚪ Nit: 0
- ❔ Question: 0

**Note:** this PR's branch is currently behind `main` (mergeable, no conflicts) — worth a rebase before merge.

dealako (approved):
Approved with minor comments — see summary comment above. One open minor finding (Windows spawnSync/.cmd shim in run-tests.mjs) doesn't block CI or merge.

PR #1379, opened 2026-08-10
refactor(enrollment): tighten display-state projection type narrowing

fayazg:
## Summary

- In `initDisplayState`, capture `item.membership` into a local so the type narrows through the guard, removing the two residual `!` non-null assertions (`overrides.get(id)!` and `...item.membership!`) and deriving the auto-renew value with a `?? false` fallback.
- Document the pass-through in `toDisplayState`: the `loading`/`error` arms are shared verbatim between the raw and display state unions (only the `loaded` arm's item type differs), so returning `state` directly is intentional.

Fixes LFXV2-3091

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/individual-enrollment/profile-individual-enrollment.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/individual-enrollment/profile-individual-enrollment.component.ts
@@ -152,6 +152,8 @@ export class ProfileIndividualEnrollmentComponent {
 
   // Projects the server state into the display shape once, at the subscription boundary.
   private toDisplayState(state: EnrollmentsState): DisplayEnrollmentsState {
+    // The loading/error arms are structurally identical across both unions (only the loaded arm's
+    // item type differs), so returning `state` verbatim is intentional, not a missed projection.
     if (state.kind !== 'loaded') return state;
     const base = environment.urls.enrollment;
     const items = state.items.map((item): DisplayEnrollment => {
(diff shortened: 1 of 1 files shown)


apps/lfx-one/src/app/modules/profile/individual-enrollment/profile-individual-enrollment.component.ts
```diff
@@ -175,11 +177,11 @@ export class ProfileIndividualEnrollmentComponent {
...
       const items = state.items.map((item) => {
-        const membershipId = item.membership?.id;
-        const isPending = membershipId ? pending.has(membershipId) : false;
-        if (membershipId && overrides.has(membershipId)) {
-          const autoRenew = overrides.get(membershipId)!;
-          const updatedMembership = { ...item.membership!, autoRenew };
+        const membership = item.membership;
+        const isPending = membership ? pending.has(membership.id) : false;
+        if (membership && overrides.has(membership.id)) {
+          const autoRenew = overrides.get(membership.id) ?? false;
```
dealako, line 183: [nit] `?? false` fallback is unreachable but changes stated invariant

Issue: `overrides` is `Map<string, boolean>`. After the `overrides.has(membership.id)` guard on the previous line, `.get()` can only type-narrow to `boolean`, never `undefined` — so the `?? false` fallback can never actually fire. The prior `!` assertion encoded the same guarantee explicitly.
Proof: `const overrides = this.autoRenewOverrides();` (typed `Map<string, boolean>`) combined with the `has()` check immediately above.
Why it matters: Purely stylistic — no behavioral difference from the old code, since a `Map<string, boolean>` cannot store `undefined` under a present key. Worth a one-line comment if the defensive fallback is intentional documentation of "trust but verify," otherwise `!` was already correct and arguably clearer given the guard.
Fix: Optional — no change required. If you want to keep the fallback as defensive style, a short comment noting it's belt-and-suspenders (not correctness-critical) would remove any reader doubt about why a proven-safe get() needs a default.

dealako (approved):
@fayazg Thanks for the cleanup — this is a well-scoped, low-risk refactor. Capturing `item.membership` into a local so the guard narrows the type is the right move, and it correctly kills both `!` assertions without changing observable behavior. The new comment on `toDisplayState` is a good addition — it heads off a future reader wondering whether the `loading`/`error` pass-through is a missed projection.

Three parallel reviews covered Security/Data Privacy, Correctness/Performance/Test Coverage, and Code Style/Documentation. All three came back clean of blocking or minor issues.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — the `?? false` fallback on the auto-renew read is unreachable given `overrides` is typed `Map<string, boolean>` and is guarded by `.has()` just above; purely stylistic vs. the old `!` assertion, no behavior change (inline comment posted).
- ❔ Question: 0

**AI bot reconciliation:** Cursor Bugbot and GitHub Copilot both summarized this as a low-risk, behavior-preserving typing refactor — consistent with this review. CodeRabbit auto-review is disabled on this repo for this PR.

**Decision: ✅ Approved**

PR #1384, opened 2026-08-10
feat(org-roi): add ROI category and projects donuts (LFXV2-2980)

luismoriguerra:
## Summary

US3 and US4 of Org Lens ROI Metrics, extending the page shell [#1352](https://github.com/linuxfoundation/lfx-self-serve/pull/1352) landed. Jira: **LFXV2-2980**.

Two new BFF reads behind the same in-handler authorization gate as their siblings, and two new chart surfaces on `/org/roi`:

- `GET /:orgUid/lens/roi/investment-breakdown` → **investment by contribution category** donut, with an amount/share toggle and a labelled remainder for sub-threshold categories.
- `GET /:orgUid/lens/roi/projects` → **leading projects** donut over Investment / Return / Net Return, with a labelled remainder and explicit reporting of negative net return.

## Notable decisions

**`/investment-breakdown` is deliberately not method-scoped, and its cache key carries no method suffix.** Unlike `/coverage` — where "cannot vary by method" was an unasserted assumption about upstream grain, and review correctly made it method-scoped anyway — this one cannot vary by construction: `ORG_LENS_ROI_INVESTMENT_BREAKDOWN` has no `MARKUP_METHOD` column, because the markup scales return, not spend. Verified against production: Red Hat's total is identical to the last significant digit under both methods. Both routes still validate `method`, so an unknown value is a uniform 400.

**`/projects` serves the complete uncapped set** (DR-005) with each project's category breakdown in one payload and one round trip. Measured against production rather than assumed: the largest of all 2,380 covered accounts serializes to **241 KB** against the 1 MiB `VALKEY_CACHE.MAX_VALUE_BYTES` ceiling — about 4.3× headroom, so every organization caches and none silently bypasses.

**Reconciliation is never forced client-side.** The category total equals the KPI investment figure by construction in the warehouse, asserted by a dbt singular test. Confirmed live at −5.96e-08 on a $145M base. Nothing rescales to make the numbers agree (FR-011b).

**Negative net return is treated as the mainline path it is** — 6.45% of production project rows across 775 accounts. An arc cannot be negative, so the geometry is clamped, but the true signed value survives into the label. The measure is partitioned once at zero: the chart and its remainder cover only non-negative values, and loss-making projects are reported solely by a named disclosure list. This was a review finding — the first implementation folded them into the remainder *and* listed them separately, double-reporting them and leaving the remainder's arc disagreeing with its own legend figure.

**Both surfaces carry the FR-039a modelled-cost disclosure** by rendering the shared `ORG_LENS_ROI_KPI_EXPLANATION` constant verbatim rather than paraphrasing it, so they cannot drift from the KPI band. This is the disclosure the aggregate-only privacy posture rests on (DR-015): 19.4% of covered organizations have a single code contributor, so "Code Contribution: $98.9M" can read as one person's pay without it.

## Verified live against production data

Run locally against dev with Playwright, as Red Hat, Inc. (`0014100000Te2QjAAJ`):

- $144.9M investment across 408 projects, $5.3B return, 3532.0% ROI, 36.3× BCR — BCR is exactly ROI + 1, live.
- KPI band and category donut both render **$144.9M**, asserted string-equal in the DOM.
- Three sub-2% categories collapse into `Other (3 categories)`, and the chart's accessible label correctly reports 8 categories behind 6 legend rows.
- Net Return shows **24 Red Hat projects with a negative net return totalling −$390.3K**, each named with its signed figure, and a remainder of `Other (374 projects)` — 408 total − 24 negative − 10 slices.
- No console error from any `/api/orgs/*/lens/roi/*` endpoint.

## Review

Four audit lanes ran pre-PR (secrets/critical-constants, repo conventions, privacy+ADR, Bugbot). No blocking authorization, secrets, SQL-injection or ADR finding; Bugbot found nothing. One blocking correctness defect (the double-counted negatives above) and seven smaller findings were fixed in `fix(org-roi): address review findings`.

The per-project grain question was put to the privacy lane explicitly — US3/US4 serve a finer cell than DR-015 measured — and answered no: DR-015 already withdrew its size-dependent argument, and each replacement leg is grain-independent. Recorded as a scope note on DR-015 in the umbrella repo, not a new decision record.

## Test plan

- [x] `yarn format:check`, `yarn lint:check` (0 errors), `yarn check-types`, `yarn build`, `./check-headers.sh`
- [x] `yarn test` — 1,387 unit tests pass
- [x] `org-lens-roi.controller.spec.ts` extended to all five handlers (31 cases), plus a guard that fails if a future handler is omitted from the parity list. Mutation-tested both ways.
- [x] Manual verification against production Red Hat data (above)
- [ ] ⚠️ **`yarn e2e` has still never executed.** The two ROI spec files (35 cases) compile and enumerate via `playwright test --list`, but the suite cannot run locally (auth-gated SSR) and the CI job that would run it is commented out in `quality-check.yml`. Every expectation is derived from its fixture through the real formatter rather than hand-written — the `--list` gate caught a barrel import that broke both files. **Treat the e2e assertions as unverified.**

## Follow-ups (not in this PR)

- Getting `yarn e2e` into CI remains the highest-value follow-up on this feature.
- `/projects` ships per-project `categories` that no component here renders yet — contract-sanctioned so US5's bar and sankey views need only one round trip.
- On first load after an org switch, the grant lookup can return 503 `ROLE_GRANTS_UNAVAILABLE` under five concurrent gated reads. Pre-existing (`/annual` is US2 code) and it fails closed correctly, but the concurrency is worth a look.

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-category-donut/org-roi-category-donut.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-category-donut/org-roi-category-donut.component.ts
@@ -1,0 +1,214 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { Component, computed, inject, Signal, signal } from '@angular/core';
+import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { ChartComponent } from '@components/chart/chart.component';
+import {
+  lfxColors,
+  ORG_LENS_ROI_CATEGORY_REMAINDER_THRESHOLD,
+  ORG_LENS_ROI_DONUT_PALETTE,
+  ORG_LENS_ROI_DONUT_REMAINDER_COLOR,
+  ORG_LENS_ROI_KPI_EXPLANATION,
+  ORG_LENS_ROI_NO_VALUE,
+} from '@lfx-one/shared/constants';
+import type { OrgLensRoiInvestmentBreakdown } from '@lfx-one/shared/interfaces';
+import { formatCurrency, formatPercent } from '@lfx-one/shared/utils';
+import { AccountContextService } from '@services/account-context.service';
+import { OrgLensRoiService } from '@services/org-lens-roi.service';
+import type { ChartData, ChartOptions } from 'chart.js';
+import { SkeletonModule } from 'primeng/skeleton';
+import { catchError, filter, of, switchMap, tap } from 'rxjs';
+
+const EMPTY_BREAKDOWN: OrgLensRoiInvestmentBreakdown = { rows: [], total: 0 };
+
+interface CategorySlice {
+  key: string;
+  label: string;
+  expenditure: number;
+  share: number;
+  color: string;
+}
+
+/** Investment by contribution category (US3, FR-023 to FR-026). */
+@Component({
+  selector: 'lfx-org-roi-category-donut',
+  imports: [ChartComponent, SkeletonModule],
+  templateUrl: './org-roi-category-donut.component.html',
+})
+export class OrgRoiCategoryDonutComponent {
+  private readonly accountContext = inject(AccountContextService);
+  private readonly roiService = inject(OrgLensRoiService);
+
+  /**
+   * FR-039a. The same string the KPI band shows, not a paraphrase of it — the two surfaces state
+   * the same disclosure, so they cannot drift apart when either is edited (DR-015).
+   */
+  protected readonly investmentExplanation = ORG_LENS_ROI_KPI_EXPLANATION.totalExpenditure;
+
+  protected readonly loading = signal(true);
+  protected readonly failed = signal(false);
+  protected readonly forbidden = signal(false);
+
+  /** FR-024 — currency or share of total. Presentation only; the underlying values never change. */
+  protected readonly asShare = signal(false);
+
+  private readonly breakdown: Signal<OrgLensRoiInvestmentBreakdown> = this.initBreakdown();
+
+  protected readonly total: Signal<number> = computed(() => this.breakdown().total);
+
+  protected readonly hasRows: Signal<boolean> = computed(() => this.breakdown().rows.length > 0);
+
+  /**
+   * The reconciliation anchor (FR-026, SC-011). It is the sum of exactly the rows drawn below, and
+   * the warehouse already guarantees it equals the KPI investment figure — a dbt singular test
+   * asserts it per account. Nothing here rescales to force the two to agree: if they ever differ,
+   * that is a defect to raise, not a discrepancy to paper over (FR-011b).
+   */
+  protected readonly totalLabel: Signal<string> = computed(() => {
+    const total = this.total();
+    return Number.isFinite(total) ? formatCurrency(total) : ORG_LENS_ROI_NO_VALUE;
+  });
+
+  /**
+   * FR-025 — categories under the display threshold collapse into one labelled remainder, so a
+   * $1,190 education line does not render as an invisible sliver with an unreachable legend entry.
+   */
+  protected readonly slices: Signal<CategorySlice[]> = computed(() => {
+    const { rows, total } = this.breakdown();
+    if (rows.length === 0 || !(total > 0)) return [];
+
+    const kept: CategorySlice[] = [];
+    const collapsed: { expenditure: number; count: number } = { expenditure: 0, count: 0 };
+
+    for (const row of rows) {
+      const share = row.expenditure / total;
+      if (share < ORG_LENS_ROI_CATEGORY_REMAINDER_THRESHOLD) {
+        collapsed.expenditure += row.expenditure;
+        collapsed.count += 1;
+        continue;
+      }
+      kept.push({
+        key: row.type,
+        label: row.label,
+        expenditure: row.expenditure,
+        share,
+        color: ORG_LENS_ROI_DONUT_PALETTE[kept.length % ORG_LENS_ROI_DONUT_PALETTE.length],
+      });
+    }
+
+    if (collapsed.count === 0) return kept;
+    return [
+      ...kept,
+      {
+        key: 'remainder',
+        label: collapsed.count === 1 ? 'Other (1 category)' : `Other (${collapsed.count} categories)`,
+        expenditure: collapsed.expenditure,
+        share: collapsed.expenditure / total,
+        color: ORG_LENS_ROI_DONUT_REMAINDER_COLOR,
+      },
+    ];
+  });
+
+  /**
+   * The legend is rendered as real text beside the canvas rather than mirrored into an `sr-only`
+   * table as the annual trend does. Both satisfy "a canvas cannot be the only presentation of its
+   * data"; a visible list is the better fit here because there are at most nine categories, and
+   * because FR-024's currency/share toggle needs somewhere visible to take effect — switching only
+   * the tooltips would leave the choice invisible until the viewer hovered a slice.
+   *
+   * Pre-formatted so the legend and the tooltip cannot disagree.
+   */
+  protected readonly legendRows: Signal<{ key: string; label: string; amount: string; share: string; display: string; color: string }[]> = computed(() => {
+    const asShare = this.asShare();
+    return this.slices().map((slice) => {
+      const amount = formatCurrency(slice.expenditure);
+      const share = `${formatPercent(slice.share * 100)}%`;
+      return { key: slice.key, label: slice.label, amount, share, display: asShare ? share : amount, color: slice.color };
+    });
+  });
+
+  protected readonly chartSummaryLabel: Signal<string> = computed(
+    () => `Doughnut chart of modelled investment across ${this.slices().length} contribution categories. The same figures are listed beside it.`
+  );
+
+  protected readonly chartData: Signal<ChartData<'doughnut'>> = computed(() => {
+    const slices = this.slices();
+    return {
+      labels: slices.map((slice) => slice.label),
+      datasets: [
+        {
+          data: slices.map((slice) => slice.expenditure),
+          backgroundColor: slices.map((slice) => slice.color),
+          borderColor: lfxColors.gray[50],
+          borderWidth: 2,
+        },
+      ],
+    };
+  });
+
+  protected readonly chartOptions: Signal<ChartOptions<'doughnut'>> = computed(() => {
+    const asShare = this.asShare();
+    const total = this.total();
+    return {
+      responsive: true,
+      maintainAspectRatio: false,
+      cutout: '60%',
+      plugins: {
+        // The adjacent list is the legend; the canvas one would only repeat it.
+        legend: { display: false },
+        tooltip: {
+          backgroundColor: 'rgba(255, 255, 255, 0.98)',
+          titleColor: lfxColors.gray[900],
+          bodyColor: lfxColors.gray[600],
+          borderColor: lfxColors.gray[200],
+          borderWidth: 1,
+          padding: 10,
+          cornerRadius: 6,
+          callbacks: {
+            label: (ctx) => {
+              const value = ctx.parsed;
+              if (!asShare) return ` ${ctx.label}: ${formatCurrency(value)}`;
+              return ` ${ctx.label}: ${formatPercent(total > 0 ? (value / total) * 100 : 0)}%`;
+            },
+          },
+        },
+      },
+    };
+  });
+
+  public toggleShare(asShare: boolean): void {
+    this.asShare.set(asShare);
+  }
+
+  private initBreakdown(): Signal<OrgLensRoiInvestmentBreakdown> {
+    // Keyed by the account id alone: category investment carries no estimation method, so a method
+    // change is not a reason to refetch this surface.
+    const accountId$ = toObservable(computed(() => this.accountContext.selectedAccount()?.accountId ?? ''));
+
+    return toSignal(
+      accountId$.pipe(
+        filter((orgUid) => !!orgUid),
+        tap(() => {
+          this.loading.set(true);
+          this.failed.set(false);
+          this.forbidden.set(false);
+        }),
+        switchMap((orgUid) =>
+          this.roiService.getInvestmentBreakdown(orgUid).pipe(
+            tap(() => this.loading.set(false)),
+            catchError((error: unknown) => {
+              console.error('Failed to load ROI investment breakdown', error);
+              this.loading.set(false);
+              // Only a 403 may show the no-access message; a 503 must not.
+              if ((error as { status?: number })?.status === 403) this.forbidden.set(true);
+              else this.failed.set(true);
+              return of(EMPTY_BREAKDOWN);
+            })
+          )
+        )
+      ),
+      { initialValue: EMPTY_BREAKDOWN }
+    );
+  }
+}
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-category-donut/org-roi-category-donut.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-category-donut/org-roi-category-donut.component.ts
@@ -12,7 +12,7 @@ import {
   ORG_LENS_ROI_KPI_EXPLANATION,
   ORG_LENS_ROI_NO_VALUE,
 } from '@lfx-one/shared/constants';
-import type { OrgLensRoiInvestmentBreakdown } from '@lfx-one/shared/interfaces';
+import type { OrgLensRoiCategoryRow, OrgLensRoiCategorySlice, OrgLensRoiInvestmentBreakdown } from '@lfx-one/shared/interfaces';
 import { formatCurrency, formatPercent } from '@lfx-one/shared/utils';
 import { AccountContextService } from '@services/account-context.service';
 import { OrgLensRoiService } from '@services/org-lens-roi.service';
@@ -21,14 +21,6 @@ import { SkeletonModule } from 'primeng/skeleton';
 import { catchError, filter, of, switchMap, tap } from 'rxjs';
 
 const EMPTY_BREAKDOWN: OrgLensRoiInvestmentBreakdown = { rows: [], total: 0 };
-
-interface CategorySlice {
-  key: string;
-  label: string;
-  expenditure: number;
-  share: number;
-  color: string;
-}
 
 /** Investment by contribution category (US3, FR-023 to FR-026). */
 @Component({
@@ -57,7 +49,12 @@ export class OrgRoiCategoryDonutComponent {
 
   protected readonly total: Signal<number> = computed(() => this.breakdown().total);
 
-  protected readonly hasRows: Signal<boolean> = computed(() => this.breakdown().rows.length > 0);
+  /**
+   * Gate on the slices, not on the row count. An organization can have category rows that sum to
+   * zero investment, and there is nothing to draw from those — keying the template off row count
+   * alone rendered a blank canvas beside an empty legend instead of saying so.
+   */
+  protected readonly hasSlices: Signal<boolean> = computed(() => this.slices().length > 0);
 
   /**
    * The reconciliation anchor (FR-026, SC-011). It is the sum of exactly the rows drawn below, and
@@ -74,18 +71,17 @@ export class OrgRoiCategoryDonutComponent {
    * FR-025 — categories under the display threshold collapse into one labelled remainder, so a
    * $1,190 education line does not render as an invisible sliver with an unreachable legend entry.
    */
-  protected readonly slices: Signal<CategorySlice[]> = computed(() => {
+  protected readonly slices: Signal<OrgLensRoiCategorySlice[]> = computed(() => {
     const { rows, total } = this.breakdown();
     if (rows.length === 0 || !(total > 0)) return [];
 
-    const kept: CategorySlice[] = [];
-    const collapsed: { expenditure: number; count: number } = { expenditure: 0, count: 0 };
+    const kept: OrgLensRoiCategorySlice[] = [];
+    const collapsed: OrgLensRoiCategoryRow[] = [];
 
     for (const row of rows) {
       const share = row.expenditure / total;
       if (share < ORG_LENS_ROI_CATEGORY_REMAINDER_THRESHOLD) {
-        collapsed.expenditure += row.expenditure;
-        collapsed.count += 1;
+        collapsed.push(row);
         continue;
       }
       kept.push({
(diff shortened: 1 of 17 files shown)


apps/lfx-one/src/app/modules/dashboards/org/org-roi/org-roi.component.html
```diff
@@ -88,6 +88,12 @@ <h2 class="text-base font-semibold text-gray-900" data-testid="org-roi-no-access
     <lfx-org-roi-kpi-cards [summary]="summary()" />
 
     <lfx-org-roi-annual-trend [method]="method()" />
+
+    <!-- No [method]: category investment has no estimation method in the warehouse, so this
+         surface is identical under both and refetching it on a switch would be wasted work. -->
+    <lfx-org-roi-category-donut />
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 678c2e3ed — thank you, this was a real gap between the comment and the structure.

Confirmed the mechanism exactly as described: `setMethod()` sets `portfolioLoading`, the `@else if (portfolioLoading())` branch wins, the whole populated block is destroyed, and `initBreakdown()` runs again on remount. The comment claimed a no-refetch behaviour the template never delivered. My own e2e missed it because it only asserted the URL carries no `method=` param, which stayed true throughout.

Rather than caching by org UID, both donuts now mount on a `showsFigures()` condition that deliberately excludes `portfolioLoading`, so they survive a method-only refresh and the category donut's account-keyed `toObservable` simply never re-emits. I avoided the cache route because the hazard here is the opposite one: keeping data alive across an *org* change could render one organization's investment under another's header. Keying each donut's own fetch on the account id means an org switch flips its own loading state and shows a skeleton instead of stale money, while a method switch changes nothing it depends on.

Verified live against Red Hat (408 projects) with the app running locally: across a logit→direct switch the `investment-breakdown` request count is unchanged and the donut never unmounts, while `summary` and `projects` correctly refetch with `method=direct`. Added an e2e case that counts the requests rather than inspecting the URL, so this cannot regress silently again.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-donut/org-roi-projects-donut.component.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  protected readonly measureLabel: Signal<string> = computed(() => this.measureLabels[this.measure()]);
+
+  /** Ranked by the selected measure, signed values intact. */
+  private readonly ranked: Signal<{ row: OrgLensRoiProjectRow; value: number }[]> = computed(() => {
+    const measure = this.measure();
+    return this.projects()
+      .rows.map((row) => ({ row, value: this.measureValue(row, measure) }))
+      .filter((entry) => Number.isFinite(entry.value))
+      .sort((a, b) => b.value - a.value || a.row.projectId.localeCompare(b.row.projectId));
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 678c2e3ed, though with one correction to the framing.

This isn't a convention violation — the repo has no written rule about pinning `localeCompare`, and the sibling Org Lens code uses it unpinned in several places (`org-projects.component.ts` sorts option labels and project names that way). So the code was consistent with its neighbours.

It was still the wrong tool, for a different reason: those siblings collate human-readable **names**, where locale-aware ordering is the point. My tie-break key is an opaque warehouse `projectId`, where collation carries no meaning at all — so the locale sensitivity was pure downside, and your hydration concern is the correct consequence of it. Replaced with a codepoint comparison (`Number(a > b) - Number(a < b)`, kept as a small helper to avoid a nested ternary, which the repo does forbid).

Worth noting the ordering is load-bearing beyond render order: it also decides which projects become slices and which fall into the remainder, so a divergent tie-break would change what the chart shows, not just the sequence.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-donut/org-roi-projects-donut.component.ts
```diff
@@ -0,0 +1,261 @@
...
+    const rest = chartable.slice(index);
+    if (rest.length === 0) return kept;
+
+    // A remainder standing for a single project hides a real name behind "Other (1 project)" and
+    // saves no space. Name it instead.
+    if (rest.length === 1) {
+      const only = rest[0];
+      return [
+        ...kept,
+        { key: only.row.projectId, label: only.row.projectName, value: only.value, weight: only.value, color: ORG_LENS_ROI_DONUT_REMAINDER_COLOR },
```
audigregorie: **Minor — single-project remainder keeps the gray remainder color with a real name**

When exactly one project is left over, the comment says "Name it instead" (of `Other (1 project)`), but the slice still gets `ORG_LENS_ROI_DONUT_REMAINDER_COLOR` (gray). The legend then shows a real project name beside the "everything else" dot, which reads as a remainder rather than a real slice. Using a palette color here (e.g. `ORG_LENS_ROI_DONUT_PALETTE[kept.length % ORG_LENS_ROI_DONUT_PALETTE.length]`) would make the intent — treat it as a real project, not a remainder — visible.

```suggestion
        { key: only.row.projectId, label: only.row.projectName, value: only.value, weight: only.value, color: ORG_LENS_ROI_DONUT_PALETTE[kept.length % ORG_LENS_ROI_DONUT_PALETTE.length] },
```

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-category-donut/org-roi-category-donut.component.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (collapsed.length === 1) {
+      const only = collapsed[0];
+      return [
+        ...kept,
+        {
+          key: only.type,
+          label: only.label,
+          expenditure: only.expenditure,
+          share: only.expenditure / total,
+          color: ORG_LENS_ROI_DONUT_REMAINDER_COLOR,
```
audigregorie: **Minor — single-category remainder keeps the gray remainder color with a real label**

Same pattern as the projects donut: when only one category collapses, the comment argues it should be named rather than hidden behind `Other (1 category)`, but the slice keeps `ORG_LENS_ROI_DONUT_REMAINDER_COLOR`. A named category in the "everything else" color is a small visual inconsistency. A palette color would match the intent.

```suggestion
        color: ORG_LENS_ROI_DONUT_PALETTE[kept.length % ORG_LENS_ROI_DONUT_PALETTE.length],
```

audigregorie (commented):
## Code Review Summary

A well-reasoned cross-app addition. The authorization gate ordering is provably correct (the controller spec asserts it via `invocationCallOrder` and guards against a sixth handler being added without coverage), the negative-net-return partition is the right design, and the cache-key scoping is justified from the source schema rather than assumed. No Critical or Major findings.

Five minor findings below — the single-item remainder color, an asymmetric cache-shape guard, a cosmetic initial-state mismatch, and a repeated class-string pattern in the toggle buttons.

### What's done well

- Authorization runs before any cache read or Snowflake query, with cheap synchronous validation ahead of the awaited grant lookup.
- The negative-net-return partition at zero avoids double-reporting and keeps the remainder arc consistent with its legend figure.
- `/investment-breakdown` carries no `method` parameter or cache-key suffix because the source table has no `MARKUP_METHOD` column — and the e2e suite asserts both invariants.
- The category total is summed from the rendered rows and trusted to equal the KPI figure by warehouse construction, never rescaled client-side.

apps/lfx-one/src/server/services/org-lens-roi.service.ts
```diff
@@ -218,6 +357,64 @@ export class OrgLensRoiService {
...
+    const projects = value as Record<string, unknown>;
+    if (!(ORG_LENS_ROI_METHODS as readonly string[]).includes(projects['method'] as string)) return false;
+    if (!Array.isArray(projects['rows'])) return false;
+    return projects['rows'].every((row) => {
+      if (row === null || typeof row !== 'object' || Array.isArray(row)) return false;
+      const entry = row as Record<string, unknown>;
+      if (!['projectId', 'projectSlug', 'projectName'].every((key) => typeof entry[key] === 'string')) return false;
+      if (!['totalExpenditure', 'totalReturn', 'profit'].every((key) => OrgLensRoiService.isFiniteNumber(entry[key]))) return false;
+      if (!['roi', 'bcr', 'breakevenMarkup'].every((key) => OrgLensRoiService.isNullableNumber(entry, key))) return false;
+      return OrgLensRoiService.isCategoryRows(entry['categories']);
```
audigregorie: **Minor — `isProjects` doesn't reconcile each project's `categories` to its `totalExpenditure`**

`isInvestmentBreakdown` (above) reconciles `rows` sum to `total` within a cent epsilon, on the grounds that a corrupt/stale cache entry could render a headline figure that contradicts its slices. Each project's `categories` sum to `totalExpenditure` by the same warehouse construction (FR-031), but `isProjects` only checks `isCategoryRows` shape — not the sum. A corrupt cache entry where a project's categories don't sum to its `totalExpenditure` would pass the guard and render. Adding the same epsilon check per project would close the asymmetric gap.

```suggestion
      if (!OrgLensRoiService.isCategoryRows(entry['categories'])) return false;
      const categories = entry['categories'] as { expenditure: number }[];
      const summed = categories.reduce((sum, row) => sum + row.expenditure, 0);
      return Math.abs(summed - (entry['totalExpenditure'] as number)) <= OrgLensRoiService.reconciliationEpsilonUsd;
```

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-donut/org-roi-projects-donut.component.ts
```diff
@@ -0,0 +1,261 @@
...
+} from '@lfx-one/shared/constants';
+import type { OrgLensRoiMethod, OrgLensRoiProjectMeasure, OrgLensRoiProjectRow, OrgLensRoiProjects, OrgLensRoiProjectSlice } from '@lfx-one/shared/interfaces';
+import { formatCurrency } from '@lfx-one/shared/utils';
+import { AccountContextService } from '@services/account-context.service';
+import { OrgLensRoiService } from '@services/org-lens-roi.service';
+import type { ChartData, ChartOptions } from 'chart.js';
+import { SkeletonModule } from 'primeng/skeleton';
+import { catchError, filter, map, of, switchMap, tap } from 'rxjs';
+
+const EMPTY_PROJECTS: OrgLensRoiProjects = { method: ORG_LENS_ROI_DEFAULT_METHOD, rows: [] };
```
audigregorie: **Minor — `EMPTY_PROJECTS.method` can disagree with the bound `[method]` input before fetch**

`EMPTY_PROJECTS` hardcodes `ORG_LENS_ROI_DEFAULT_METHOD` ('logit') as the initial value of the `projects` signal. When the user has 'direct' persisted, `projects().method` is 'logit' until the fetch resolves, while the parent's `method()` input is 'direct'. `method` isn't rendered, so this is cosmetic — but deriving the initial value from the input (or dropping `method` from the empty shape if nothing reads it) would remove the mismatch.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-donut/org-roi-projects-donut.component.html
```diff
@@ -0,0 +1,93 @@
...
+    <div class="inline-flex self-start rounded-md border border-gray-200 p-0.5" role="group" aria-label="Project measure">
+      @for (option of measures; track option) {
+        <button
+          type="button"
+          (click)="setMeasure(option)"
+          [class]="
+            option === measure()
+              ? 'rounded px-3 py-1 text-xs font-medium bg-blue-50 text-blue-700'
+              : 'rounded px-3 py-1 text-xs font-medium text-gray-600 hover:bg-gray-50'
+          "
```
audigregorie: **Minor — toggle button `[class]` ternary duplicates the base class string**

Both branches repeat `'rounded px-3 py-1 text-xs font-medium'`; only the active/inactive tail differs. A base class plus a conditional active class is less repetitive and easier to keep in sync. Same pattern appears in `org-roi-category-donut.component.html` (both toggle buttons).

```suggestion
          class="rounded px-3 py-1 text-xs font-medium"
          [class.bg-blue-50]="option === measure()"
          [class.text-blue-700]="option === measure()"
          [class.text-gray-600]="option !== measure()"
          [class.hover:bg-gray-50]="option !== measure()"
```

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-donut/org-roi-projects-donut.component.ts
luismoriguerra: Fixed in fdfb0eb23 — good catch, the code was contradicting its own comment. The whole point of promoting a lone leftover to a named slice is to say "this is a real project, not a bucket", and then it was painted in the everything-else gray. Now uses `ORG_LENS_ROI_DONUT_PALETTE[kept.length % ...]` exactly as you suggested, with a comment recording why the gray would be wrong here.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-category-donut/org-roi-category-donut.component.ts
luismoriguerra: Fixed in fdfb0eb23, same treatment as the projects donut. Both single-item promotions now take a palette colour, so the two components stay symmetric.

apps/lfx-one/src/server/services/org-lens-roi.service.ts
luismoriguerra: Fixed in fdfb0eb23 — the asymmetry was real and worth closing.

I checked the invariant against production before adding the guard, because a strict version could have backfired. Across all 32,890 project rows: **0** projects with no breakdown rows, **0** off by more than a cent, worst absolute difference **1.9e-09**. So the reconciliation genuinely holds per project, not just per organization.

One deliberate narrowing: the check is enforced **only when `categories` is non-empty**. The read is a `LEFT JOIN`, so a project the breakdown has no rows for legitimately arrives with an empty array, and a strict sum check would then reject the entire payload on every read — leaving that organization permanently uncacheable and re-running the ~1,200-row join on each request. That exact failure mode came up earlier in this PR with the contribution-type enum check, so I did not want to reintroduce it in a different guard. No production project hits the empty case today; the guard just refuses to be the thing that breaks if one ever does.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-donut/org-roi-projects-donut.component.html
luismoriguerra: Fixed in fdfb0eb23, and applied to all three toggle groups — the measure tabs here plus both Amount/Share buttons in the category donut.

Worth noting your suggested form is also the safer one, not just the tidier one: individual `[class.x]` bindings compose with a static `class` attribute, whereas a whole-object `[class]` binding clobbers it. There is existing precedent for the Tailwind variant syntax (`[class.hover:bg-gray-50]`) elsewhere in the app, and the build confirms the template parser accepts the colon.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-donut/org-roi-projects-donut.component.ts
luismoriguerra: You are right that the values disagree, and I looked at fixing it, but I am going to leave this one as-is — with reasoning, in case you disagree.

Two things:

1. Nothing reads it. The component only ever consumes `projects().rows`; `method` on the payload is part of the wire contract for the response, not something this component renders or branches on. So the mismatch is unobservable today.

2. The natural fix does not work. `EMPTY_PROJECTS` is the `initialValue` passed to `toSignal` in a **field initializer**. Deriving it from `this.method()` there would read a required input before Angular has set it, which throws at construction. I could make the sentinel a factory and thread it through, but that is real machinery to correct a field nobody reads.

The alternative you hint at — dropping `method` from the empty shape — is not open either, since `OrgLensRoiProjects` requires it and the type is the BFF response contract.

Happy to revisit if a future consumer starts reading `projects().method` — at that point the mismatch stops being cosmetic and the sentinel should go. Leaving the thread open for your call rather than resolving it unilaterally.

luismoriguerra: Reversing my earlier reply — you were right, and my objection was aimed at the wrong fix.

I argued that deriving the initial value from the input was unsafe (it is: reading a required input in a field initializer throws before Angular sets it) and concluded the mismatch had to stay. But that only rules out *one* of the two options you offered. The second one — dropping `method` from the empty shape because nothing reads it — is the right answer, and it goes further than the sentinel: the component never needed the envelope at all.

Fixed in e107d202a. The signal now holds `OrgLensRoiProjectRow[]` rather than `OrgLensRoiProjects`, with `map((projects) => projects.rows)` in the pipe and `[]` as the initial value. `EMPTY_PROJECTS` is gone, so there is no invented `method` to disagree with anything. The wire type keeps its `method` field, since that is the BFF contract.

Thanks for pushing on a "cosmetic" one — the resulting code is simpler than what it replaced.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/org-roi.component.ts
```diff
@@ -104,6 +108,31 @@ export class OrgRoiComponent {
...
+  protected readonly showsFigures: Signal<boolean> = computed(
+    () =>
+      this.loaded() &&
+      !this.hasNoOrgAccess() &&
+      this.hasCompany() &&
+      this.hasAnalyticsId() &&
+      !this.portfolioForbidden() &&
+      !this.portfolioFailed() &&
+      this.hasRoiData()
+  );
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in cd345bbac, and your prescription is exactly what landed: `showsFigures` now requires the loaded summary's `orgUid` to match the selected account.

Two things worth recording, because your diagnosis was sharper than mine.

First, I had written a comment on `showsFigures` asserting this was already safe — that each donut keys its fetch on the account id and raises its own loading state, so an org switch would render a skeleton rather than stale figures. That reasoning was wrong in a way I had not thought through: those loading flags are driven by `toObservable`, which settles on the next effect flush, whereas the account-id change is synchronous. The identity check closes that gap because it is a plain signal read in the same change-detection pass. The comment has been corrected too — it was overclaiming.

Second, the header was the worse half and neither of us flagged it initially: `windowLabel` sits outside the state chain entirely, so it kept rendering the previous organization's date range for the full duration of a warehouse read — seconds, not a frame. It is now blank until the summary matches.

Your point about the grant race is well taken and I had not connected it: the donuts no longer fire until the new summary has landed, so an org switch issues fewer concurrent gated reads.

Verified live against the running app — switching away from Red Hat leaves no date window, no donuts, and issues no breakdown or projects request for the new organization until its summary resolves.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/org-roi.component.ts
```diff
@@ -104,7 +108,54 @@ export class OrgRoiComponent {
...
+   * `summary()` holds its previous value while a new request is in flight, so during an org switch
+   * every measure derived from it belongs to the organization the viewer just navigated away from.
+   * Comparing the payload's own `orgUid` to the selected account is a synchronous signal read, so
+   * it flips in the same change-detection pass as the switch — unlike the donuts' internal loading
+   * flags, which are driven by `toObservable` and therefore only settle on the next effect flush.
+   */
+  private readonly summaryMatchesSelectedOrg: Signal<boolean> = computed(() => {
+    const selected = this.accountContext.selectedAccount()?.accountId ?? '';
+    return selected !== '' && this.summary().orgUid === selected;
+  });
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 55acb7fca — correct, and I had left the job half done.

You are right about the mechanism: the populated branch renders `summary()` on `hasRoiData()` alone, and `portfolioLoading` is only raised on the following effect flush, so for that window the branch chain still selected the populated arm with the previous organization's figures in it. The two surfaces I guarded were the ones I had just been thinking about; the KPI band and annual trend carry the largest numbers on the page and were the ones that mattered most.

Rather than add `summaryMatchesSelectedOrg()` as a third individual guard, I moved it into the loading condition itself:

```
@else if (portfolioLoading() || !summaryMatchesSelectedOrg())
```

That covers every surface derived from the summary under one rule instead of three that can drift apart — which is how this gap appeared in the first place. `showsFigures()` keeps its own copy of the check because it sits outside the chain, but the chain no longer depends on remembering to add it per branch.

The e2e case now asserts the KPI band and the annual trend are absent, and the skeleton present, when the summary describes another organization. Also confirmed against the running app that the happy path is unaffected: Red Hat still renders its band, window and both donuts with no skeleton.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/org-roi.component.html
```diff
@@ -57,7 +57,12 @@ <h2 class="text-base font-semibold text-gray-900" data-testid="org-roi-no-access
       icon="fa-light fa-triangle-exclamation"
       title="ROI metrics aren't available for this company right now. Please try again."
       data-testid="org-roi-no-analytics-id-state" />
-  } @else if (portfolioLoading()) {
+  } @else if (portfolioLoading() || !summaryMatchesSelectedOrg()) {
+    <!-- The org-identity check belongs in the loading condition, not only on the figures below:
+         the summary keeps its previous value until the next response lands, and `portfolioLoading`
+         is raised only on the following effect flush. Without it this branch loses to the populated
+         one for that window, putting the previous organization's headline amounts under the newly
+         selected one. -->
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in 1aa325292 — this was a real regression and I introduced it one commit earlier. Thank you for catching it.

Your trace is exactly right. `catchError` replaces the summary with `EMPTY_SUMMARY` (`orgUid: ''`), so `summaryMatchesSelectedOrg()` goes false on every failure, and with that test sitting ahead of the failure branches the chain never reached them. Both the refusal and the outage states became unreachable — and those two were deliberately kept distinct in the earlier work precisely so a transient 503 never reads to a viewer as "you lost access".

Both failure branches now precede the loading branch. That ordering is safe in the other direction too: each new request resets `portfolioForbidden` / `portfolioFailed` in the same `tap` that raises loading, so a stale error cannot linger over a fresh load. I put the reasoning in a comment above the branches, since the ordering is load-bearing and looks arbitrary otherwise.

One thing worth stating plainly, because it is the more useful lesson: this repository already has e2e cases for both of these paths — one asserts `org-roi-forbidden` is visible on a 403, the other asserts `org-roi-error` on a 503 — and both would have failed on the commit that broke this. They did not, because the Playwright suite cannot execute anywhere: it needs a real Auth0 issuer locally, and the CI job that would run it is commented out in `quality-check.yml`. I have added a third case pinning this specific ordering, but it joins the same unrunnable suite. Switching that gate on is the highest-value follow-up on this feature, and this regression is the clearest evidence yet.

audigregorie (approved):

PR #1389, opened 2026-08-11
feat(org-roi): add projects section with four views (LFXV2-2980)

luismoriguerra:
User Story 5 of the Org Lens ROI feature ([LFXV2-2980](https://linuxfoundation.atlassian.net/browse/LFXV2-2980)). Follows #1352 (page shell + BFF) and #1384 (category and projects donuts).

## What this adds

Four complementary views over the **complete, uncapped** project set, under a shared selection, on the existing `/org/roi` page.

**No BFF work.** `GET /:orgUid/lens/roi/projects` already serves every project with its category breakdown in one payload — measured at 210 KiB for the largest organization (Red Hat, 408 projects) against the 1 MiB `VALKEY_CACHE.MAX_VALUE_BYTES` ceiling. All four views read it.

| View | What it shows |
|---|---|
| Investment vs return | Investment stacked by contribution category, against return |
| Where it flows | Sankey, with an Investment/Return toggle |
| Efficiency | Bubble plot on logarithmic axes, sized by net return |
| All projects | Sortable, paged table over every project |

One picker serves the three chart views, so the selection is shared structurally rather than duplicated per view.

## Decisions worth reviewing

**The bar chart uses two scales, and this is the one judgement call I'd like a second opinion on.** Return runs 36–56× investment, so on a shared linear axis the stacked categories — the entire content of the view — are invisible. A logarithmic axis is *wrong* for a stacked bar: segments are positioned by cumulative total, so under log scaling a segment's length stops corresponding to its value and the small categories collapse anyway. So investment and return are drawn on separate scales, with a note stating the two bars cannot be compared by length. This is the same open axis-scaling question already recorded against the annual trend; if that resolves differently, both should change together.

**The sankey is deliberately asymmetric.** Investment fans out `organization → category → project`; return comes straight back. Return is not decomposed by category in the warehouse, so an intermediate node on that side would be invented rather than reported.

**A log axis has no zero.** Projects with no investment or no return are drawn at the floor and named in a disclosure, rather than dropped — dropping them would make an unmeasured project indistinguishable from one that does not exist. The accompanying table carries their true figures.

**Truncation is disclosed, not silent.** "All" is a legitimate selection on a 408-project portfolio, and no view stays readable at that size, so each draws its leading N (bar 25, sankey 12, bubble 250) and says how many it could not draw. The ceilings differ because legibility does.

**No client-side rescaling.** Categories sum to their project's investment by construction — re-measured against production at a worst gap of 7.45e-09 across all 408 Red Hat projects. A discrepancy of a cent or more is surfaced as the metric-layer defect it would be, not corrected in the renderer. `roi`, `bcr` and `profit` are read from the payload and never re-derived; a null ratio renders as `—`, never `0`.

**One deliberate duplicate request.** The donut and the section both read the projects endpoint. Client-side de-duplication was implemented and then removed: it needed memoisation whose interaction with hydration's synchronous transfer-cache replay, subscription cancellation and completion semantics produced a defect in three consecutive review rounds, to save a request that the transfer cache and the server-side cache already make cheap. The reasoning and the structural alternative are recorded in a comment on `getProjects`.

## Also in this PR

- `chartjs-chart-sankey@0.15.0` — MIT, no runtime dependencies, ships its own types, pinned exactly because it is **pre-1.0**. Flagging for dependency-owner review rather than assuming approval. Exposure is one component and one registration line.
- PrimeNG types its chart `type` against a fixed union that no plugin can extend, so `lfx-chart` widens at that single binding; the `ChartType` it exposes to callers stays narrow.
- ~~The e2e workflow was enabled non-blocking.~~ **Removed during review** — handing AWS, Auth0 and Supabase credentials to a workflow that runs PR-controlled install/build/test scripts deserves its own review and should not ride along with a feature. Both workflow files are back to their `main` state; this will return as a separate PR. (The run triggered before the revert confirmed the exposure is real, not theoretical: it completed OIDC auth, pulled the credentials from AWS Secrets Manager, and went on to run PR-controlled `yarn install` and `yarn build`.)

## Not in this PR

Row-click navigation to the ROI project detail route (FR-034) is **deferred to US6**, which creates that route. Inventing the path or stubbing it would put a guessed URL into merged code that US6 then has to agree with.

## Test plan

- [x] `yarn format:check`, `yarn lint:check`, `yarn build`, `./check-headers.sh`
- [x] `yarn test` — 1,495 unit tests pass
- [x] `yarn playwright test --list` — 31 cases in the ROI projects spec (21 new), compiles and enumerates
- [x] Verified manually against production Red Hat data (408 projects) on a local dev build
- [x] **SC-003 performance measured**: time-to-usable-content 1.52–2.19 s across four runs against the 408-project payload, inside the 3 s threshold. Tab switches 40–127 ms; selecting all 408 projects redraws the bubble view in under 80 ms.
- [ ] ⚠️ **The e2e suite still has no CI gate.** The 31 cases are verified by compilation only. One real run did start on this PR before the CI change was reverted: the credential path, Playwright setup and build all succeeded and the suite began executing, so the pipeline works — but it had not finished when the enablement was pulled, so the cases remain unproven.

⚠️ Note on the performance figure: the client-side number is healthy, but the **uncached warehouse read is ~7 s** for the largest organization. That is upstream of this PR — the first viewer waits on it before anything can paint — and belongs against the caching strategy rather than this view.

[LFXV2-2980]: https://linuxfoundation.atlassian.net/browse/LFXV2-2980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/quality-check.yml
+++ b/.github/workflows/quality-check.yml
@@ -52,17 +52,38 @@ jobs:
       - name: Build project
         run: yarn build
 
-  # e2e-tests:
-  #   name: E2E Tests
-  #   needs: quality-checks
-  #   uses: ./.github/workflows/e2e-tests.yml
-  #   with:
-  #     node-version: "22"
-  #     test-command: "e2e"
-  #     browser: "chromium" # Use only Chromium for PR testing for faster feedback
-  #     skip-build: false
-  #   secrets:
-  #     TEST_USERNAME: ${{ secrets.TEST_USERNAME }}
-  #     TEST_PASSWORD: ${{ secrets.TEST_PASSWORD }}
-  #     AI_API_KEY: ${{ secrets.AI_API_KEY }}
-  #     AI_PROXY_URL: ${{ secrets.AI_PROXY_URL }}
+  # Enabled non-blocking (`blocking: false`): the suite has never executed anywhere, so its first
+  # runs are as likely to expose a bad assertion as a real defect, and gating merges on an unproven
+  # gate would strand the repo. Results land in the summary and the PR comment. Flip `blocking` to
+  # true — and require this check in branch protection — once it has been green for a few runs.
+  #
+  # `blocking: false` covers *test failures* only. A dependency install, a build, or the AWS OIDC
+  # secret fetch failing inside the called workflow still fails this job red, which is intended —
+  # those are infrastructure faults, not unproven assertions.
+  #
+  # `continue-on-error` cannot be used here: a job that calls a reusable workflow only supports
+  # name / needs / if / permissions / uses / with / secrets / strategy / concurrency. The input is
+  # applied to the test step inside the called workflow instead.
+  e2e-tests:
+    name: E2E Tests
+    needs: quality-checks
+    # A called workflow cannot hold more permission than its caller, so these must be granted here
+    # even though e2e-tests.yml declares them: id-token for the AWS OIDC secret fetch, and
+    # pull-requests/issues for the results comment.
+    permissions:
+      id-token: write
+      contents: read
+      issues: write
+      pull-requests: write
+    uses: ./.github/workflows/e2e-tests.yml
+    with:
+      node-version: "22"
+      test-command: "e2e"
+      browser: "chromium" # Use only Chromium for PR testing for faster feedback
+      skip-build: false
+      blocking: false
+    secrets:
+      TEST_USERNAME: ${{ secrets.TEST_USERNAME }}
+      TEST_PASSWORD: ${{ secrets.TEST_PASSWORD }}
+      AI_API_KEY: ${{ secrets.AI_API_KEY }}
+      AI_PROXY_URL: ${{ secrets.AI_PROXY_URL }}
--- a/.github/workflows/quality-check.yml
+++ b/.github/workflows/quality-check.yml
@@ -52,38 +52,17 @@ jobs:
       - name: Build project
         run: yarn build
 
-  # Enabled non-blocking (`blocking: false`): the suite has never executed anywhere, so its first
-  # runs are as likely to expose a bad assertion as a real defect, and gating merges on an unproven
-  # gate would strand the repo. Results land in the summary and the PR comment. Flip `blocking` to
-  # true — and require this check in branch protection — once it has been green for a few runs.
-  #
-  # `blocking: false` covers *test failures* only. A dependency install, a build, or the AWS OIDC
-  # secret fetch failing inside the called workflow still fails this job red, which is intended —
-  # those are infrastructure faults, not unproven assertions.
-  #
-  # `continue-on-error` cannot be used here: a job that calls a reusable workflow only supports
-  # name / needs / if / permissions / uses / with / secrets / strategy / concurrency. The input is
-  # applied to the test step inside the called workflow instead.
-  e2e-tests:
-    name: E2E Tests
-    needs: quality-checks
-    # A called workflow cannot hold more permission than its caller, so these must be granted here
-    # even though e2e-tests.yml declares them: id-token for the AWS OIDC secret fetch, and
-    # pull-requests/issues for the results comment.
-    permissions:
-      id-token: write
-      contents: read
-      issues: write
-      pull-requests: write
-    uses: ./.github/workflows/e2e-tests.yml
-    with:
-      node-version: "22"
-      test-command: "e2e"
-      browser: "chromium" # Use only Chromium for PR testing for faster feedback
-      skip-build: false
-      blocking: false
-    secrets:
-      TEST_USERNAME: ${{ secrets.TEST_USERNAME }}
-      TEST_PASSWORD: ${{ secrets.TEST_PASSWORD }}
-      AI_API_KEY: ${{ secrets.AI_API_KEY }}
-      AI_PROXY_URL: ${{ secrets.AI_PROXY_URL }}
+  # e2e-tests:
+  #   name: E2E Tests
+  #   needs: quality-checks
+  #   uses: ./.github/workflows/e2e-tests.yml
+  #   with:
+  #     node-version: "22"
+  #     test-command: "e2e"
+  #     browser: "chromium" # Use only Chromium for PR testing for faster feedback
+  #     skip-build: false
+  #   secrets:
+  #     TEST_USERNAME: ${{ secrets.TEST_USERNAME }}
+  #     TEST_PASSWORD: ${{ secrets.TEST_PASSWORD }}
+  #     AI_API_KEY: ${{ secrets.AI_API_KEY }}
+  #     AI_PROXY_URL: ${{ secrets.AI_PROXY_URL }}
--- a/apps/lfx-one/e2e/helpers/org-roi.helper.ts
+++ b/apps/lfx-one/e2e/helpers/org-roi.helper.ts
@@ -2,6 +2,23 @@
 // SPDX-License-Identifier: MIT
 
 import { expect, Page, Route, test } from '@playwright/test';
+// Deep-imported, not from the `constants` barrel: the barrel pulls in modules that depend on
+// @angular/common and fail to load outside the app, the same trap the specs hit with `utils`.
+// Importing the real constants rather than restating their values means changing a ceiling or the
+// no-value indicator fails a test instead of silently diverging from what a viewer sees.
+import {
+  ORG_LENS_ROI_NO_VALUE,
+  ORG_LENS_ROI_PROJECT_BAR_MAX_ROWS,
+  ORG_LENS_ROI_PROJECT_BUBBLE_MAX_POINTS,
+  ORG_LENS_ROI_PROJECT_PICKER_DEFAULT_COUNT,
+  ORG_LENS_ROI_PROJECT_SANKEY_MAX_PROJECTS,
+} from '@lfx-one/shared/constants/org-lens-roi.constants';
+
+export const PICKER_DEFAULT_COUNT = ORG_LENS_ROI_PROJECT_PICKER_DEFAULT_COUNT;
+export const BAR_MAX_ROWS = ORG_LENS_ROI_PROJECT_BAR_MAX_ROWS;
+export const SANKEY_MAX_PROJECTS = ORG_LENS_ROI_PROJECT_SANKEY_MAX_PROJECTS;
+export const BUBBLE_MAX_POINTS = ORG_LENS_ROI_PROJECT_BUBBLE_MAX_POINTS;
+export const NO_VALUE = ORG_LENS_ROI_NO_VALUE;
 
 export const ORG_ROI_URL = '/org/roi';
 export const MOCK_ACCOUNT_ID = '0014100000Te2QjAAJ';
@@ -9,8 +26,7 @@ export const MOCK_ACCOUNT_ID = '0014100000Te2QjAAJ';
 /**
  * Only the year still in progress is labelled partial, so every year-bearing fixture is anchored to
  * the current year rather than hardcoded — otherwise these assertions would quietly invert next
- * January. A hardcoded end year was one of three defects that reached review on the earlier
- * portfolio-summary work.
+ * January.
  */
 export const CURRENT_YEAR = new Date().getFullYear();
 
@@ -164,6 +180,54 @@ export const NEGATIVE_PROJECTS = MOCK_PROJECT_INPUTS.filter((input) => input.ret
   .sort((a, b) => b.profit - a.profit);
 
 export const NEGATIVE_PROJECTS_TOTAL = NEGATIVE_PROJECTS.reduce((sum, project) => sum + project.profit, 0);
+
+/**
+ * The payload arrives ordered by return descending, and the projects section takes its default
+ * selection straight off the front of it. Derived here rather than listed, so reordering the
+ * fixture above cannot leave this silently describing the wrong five.
+ */
+export const PROJECTS_BY_RETURN = [...MOCK_PROJECT_INPUTS].sort((a, b) => b.return - a.return);
+
+export const DEFAULT_SELECTED_PROJECTS = PROJECTS_BY_RETURN.slice(0, PICKER_DEFAULT_COUNT);
+
+export const UNSELECTED_PROJECT = PROJECTS_BY_RETURN[PICKER_DEFAULT_COUNT];
+
+/**
+ * Several hundred projects, for paging the table and for the truncation disclosures the chart views
+ * show when "All" is selected. Returns descend with the index so the ordering is unambiguous.
+ */
+export const MANY_PROJECT_COUNT = 300;
+
+export const MANY_PROJECT_INPUTS = Array.from({ length: MANY_PROJECT_COUNT }, (unused, index) => ({
+  slug: `project-${String(index).padStart(3, '0')}`,
+  name: `Project ${String(index).padStart(3, '0')}`,
+  expenditure: 1_000_000 + (MANY_PROJECT_COUNT - index) * 1_000,
+  return: (MANY_PROJECT_COUNT - index) * 10_000_000,
+}));
+
+export const MANY_PROJECTS = {
+  method: 'logit',
+  rows: MANY_PROJECT_INPUTS.map((input) => projectRow(input.slug, input.name, input.expenditure, input.return)),
+};
+
+/**
+ * A project with no investment. The warehouse returns NULL — never zero — for a ratio it cannot
+ * divide, and the table must render that as the no-value indicator rather than as 0.
+ */
+export const NO_INVESTMENT_PROJECT = {
+  projectId: 'prj-unmeasured',
+  projectSlug: 'unmeasured',
+  projectName: 'Unmeasured Project',
+  totalExpenditure: 0,
+  totalReturn: 0,
+  profit: 0,
+  roi: null,
+  bcr: null,
+  breakevenMarkup: null,
+  categories: [],
+};
+
+export const PROJECTS_WITH_NULL_RATIOS = { method: 'logit', rows: [...MOCK_PROJECTS.rows, NO_INVESTMENT_PROJECT] };
 
 interface StubOptions {
   hasAccess?: boolean;
--- a/apps/lfx-one/e2e/helpers/org-roi.helper.ts
+++ b/apps/lfx-one/e2e/helpers/org-roi.helper.ts
@@ -4,20 +4,23 @@
 import { expect, Page, Route, test } from '@playwright/test';
 // Deep-imported, not from the `constants` barrel: the barrel pulls in modules that depend on
 // @angular/common and fail to load outside the app, the same trap the specs hit with `utils`.
-// Importing the real constants rather than restating their values means changing a ceiling or the
-// no-value indicator fails a test instead of silently diverging from what a viewer sees.
-import {
-  ORG_LENS_ROI_NO_VALUE,
-  ORG_LENS_ROI_PROJECT_BAR_MAX_ROWS,
-  ORG_LENS_ROI_PROJECT_BUBBLE_MAX_POINTS,
-  ORG_LENS_ROI_PROJECT_PICKER_DEFAULT_COUNT,
-  ORG_LENS_ROI_PROJECT_SANKEY_MAX_PROJECTS,
-} from '@lfx-one/shared/constants/org-lens-roi.constants';
+import { ORG_LENS_ROI_NO_VALUE } from '@lfx-one/shared/constants/org-lens-roi.constants';
 
-export const PICKER_DEFAULT_COUNT = ORG_LENS_ROI_PROJECT_PICKER_DEFAULT_COUNT;
-export const BAR_MAX_ROWS = ORG_LENS_ROI_PROJECT_BAR_MAX_ROWS;
-export const SANKEY_MAX_PROJECTS = ORG_LENS_ROI_PROJECT_SANKEY_MAX_PROJECTS;
-export const BUBBLE_MAX_POINTS = ORG_LENS_ROI_PROJECT_BUBBLE_MAX_POINTS;
+/**
+ * Written down here rather than imported from the component constants, deliberately.
+ *
+ * These four numbers are a product contract — how many projects each view promises to draw, and how
+ * many the default selection starts with. Importing them would make the assertions restate the
+ * implementation: changing the bar cap from 25 to 24 would move the component and the expectation
+ * together and the truncation test would still pass, having stopped testing anything. A silent
+ * change to any of these should fail here and make someone confirm it was intended.
+ */
+export const PICKER_DEFAULT_COUNT = 5;
+export const BAR_MAX_ROWS = 25;
+export const SANKEY_MAX_PROJECTS = 12;
+export const BUBBLE_MAX_POINTS = 250;
+
+/** Imported, unlike the ceilings above: a rendering detail rather than a promise to the viewer. */
 export const NO_VALUE = ORG_LENS_ROI_NO_VALUE;
 
 export const ORG_ROI_URL = '/org/roi';
--- a/apps/lfx-one/e2e/helpers/org-roi.helper.ts
+++ b/apps/lfx-one/e2e/helpers/org-roi.helper.ts
@@ -4,6 +4,8 @@
 import { expect, Page, Route, test } from '@playwright/test';
 // Deep-imported, not from the `constants` barrel: the barrel pulls in modules that depend on
 // @angular/common and fail to load outside the app, the same trap the specs hit with `utils`.
+import { ACCOUNT_COOKIE_KEY } from '@lfx-one/shared/constants/accounts.constants';
+import { FEATURE_FLAG_OVERRIDE_STORAGE_KEY, ORG_LENS_ROI_ENABLED_FLAG } from '@lfx-one/shared/constants/feature-flags.constants';
 import { ORG_LENS_ROI_NO_VALUE } from '@lfx-one/shared/constants/org-lens-roi.constants';
 
 /**
@@ -48,10 +50,17 @@ export function skipWhenAuthMissing(page: Page): void {
   }
 }
 
+/**
+ * Preselect the organization the ROI page reads.
+ *
+ * The cookie name comes from the shared constant rather than a literal. It was previously
+ * `lfx-selected-account`, which the app has never read — so no organization was ever selected, the
+ * page sat on its loading skeleton forever, and nothing caught it because the suite had never run.
+ */
 export async function seedSelectedOrgCookie(page: Page): Promise<void> {
   await page.context().addCookies([
     {
-      name: 'lfx-selected-account',
+      name: ACCOUNT_COOKIE_KEY,
       value: JSON.stringify({ uid: MOCK_ACCOUNT_ID }),
       domain: 'localhost',
       path: '/',
@@ -289,7 +298,26 @@ export async function stubOrgLensContext(page: Page, options: StubOptions = {}):
   );
 }
 
+/**
+ * Pin feature flags for this page, before any application code runs.
+ *
+ * Without this the ROI specs cannot run at all. `/org/roi` sits behind a `canMatch` guard, and the
+ * flag SDK evaluates twice — first against an anonymous context at bootstrap, then again once the
+ * authenticated user context is applied. A flag targeted at named users is false in that first
+ * window, so the guard can resolve against it and redirect before the real value arrives, leaving
+ * every case skipped or asserting against a page that has already navigated away.
+ *
+ * The override is read only by non-production builds; see `FEATURE_FLAG_OVERRIDE_STORAGE_KEY`.
+ */
+export async function stubFeatureFlags(page: Page, flags: Record<string, boolean>): Promise<void> {
+  await page.addInitScript(([key, value]) => window.localStorage.setItem(key as string, value as string), [
+    FEATURE_FLAG_OVERRIDE_STORAGE_KEY,
+    JSON.stringify(flags),
+  ] as const);
+}
+
 export async function gotoOrgRoiPage(page: Page): Promise<void> {
+  await stubFeatureFlags(page, { [ORG_LENS_ROI_ENABLED_FLAG]: true });
   await seedSelectedOrgCookie(page);
   await page.goto('/', { waitUntil: 'domcontentloaded' });
   skipWhenAuthMissing(page);
@@ -297,7 +325,9 @@ export async function gotoOrgRoiPage(page: Page): Promise<void> {
   await page.goto(ORG_ROI_URL, { waitUntil: 'domcontentloaded' });
   skipWhenAuthMissing(page);
   await expect(page).not.toHaveURL(/auth0\.com/);
-  if (!page.url().includes('/org/roi')) {
-    test.skip(true, 'org-lens-roi-enabled flag appears off — /org/roi redirected away');
-  }
+
+  // The guard redirects asynchronously, so arriving at the URL is not the same as staying on it —
+  // asserting immediately would pass against a page that is about to leave.
+  await expect(page).toHaveURL(/\/org\/roi/, { timeout: 15_000 });
+  await expect(page.getByTestId('org-roi-page')).toBeVisible({ timeout: 15_000 });
 }
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-project-picker/org-roi-project-picker.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-project-picker/org-roi-project-picker.component.html
@@ -1,0 +1,90 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<div class="flex flex-col gap-3 rounded-md border border-gray-200 bg-gray-50 p-3" data-testid="org-roi-project-picker">
+  <div class="flex flex-col gap-2 sm:flex-row sm:items-center sm:justify-between">
+    <h3 class="text-sm font-semibold text-gray-900">Projects shown</h3>
+
+    <div class="inline-flex self-start rounded-md border border-gray-200 bg-white p-0.5" role="group" aria-label="Project selection presets">
+      <button
+        type="button"
+        (click)="selectTop()"
+        class="rounded px-3 py-1 text-xs font-medium text-gray-600 hover:bg-gray-50"
+        data-testid="org-roi-project-picker-top">
+        Top {{ defaultCount }}
+      </button>
+      <button
+        type="button"
+        (click)="selectAll()"
+        [disabled]="!canSelectAll()"
+        class="rounded px-3 py-1 text-xs font-medium text-gray-600 hover:bg-gray-50 disabled:cursor-not-allowed disabled:text-gray-300"
+        data-testid="org-roi-project-picker-all">
+        All
+      </button>
+      <button
+        type="button"
+        (click)="selectNone()"
+        [disabled]="!hasSelection()"
+        class="rounded px-3 py-1 text-xs font-medium text-gray-600 hover:bg-gray-50 disabled:cursor-not-allowed disabled:text-gray-300"
+        data-testid="org-roi-project-picker-none">
+        None
+      </button>
+    </div>
+  </div>
+
+  @if (hasSelection()) {
+    <ul class="flex flex-wrap gap-2" data-testid="org-roi-project-picker-chips">
+      @for (option of selectedOptions(); track option.projectId) {
+        <li class="inline-flex items-center gap-2 rounded-full border border-gray-300 bg-white py-1 pl-3 pr-1 text-xs text-gray-700">
+          <span>{{ option.projectName }}</span>
+          <button
+            type="button"
+            (click)="remove(option.projectId)"
+            class="flex h-5 w-5 items-center justify-center rounded-full text-gray-400 hover:bg-gray-100 hover:text-gray-700"
+            [attr.aria-label]="'Remove ' + option.projectName"
+            [attr.data-testid]="'org-roi-project-picker-remove-' + option.projectId">
+            <i class="fa-light fa-xmark text-xs" aria-hidden="true"></i>
+          </button>
+        </li>
+      }
+    </ul>
+  } @else {
+    <p class="text-xs text-gray-500" data-testid="org-roi-project-picker-empty">No projects selected. Choose a preset, or search for a project to add.</p>
+  }
+
+  <div class="flex flex-col gap-2">
+    <label class="flex flex-col gap-1 text-xs text-gray-600">
+      <span class="sr-only">Search projects to add</span>
+      <input
+        type="search"
+        [ngModel]="query()"
+        (ngModelChange)="query.set($event)"
+        placeholder="Add a project by name"
+        class="w-full rounded-md border border-gray-300 px-3 py-2 text-sm text-gray-900 placeholder:text-gray-400 focus:border-blue-500 focus:outline-none"
+        data-testid="org-roi-project-picker-search" />
+    </label>
+
+    @if (hasQuery()) {
+      <!-- Announced rather than only drawn: a search that silently returns nothing is
+           indistinguishable from one that has not run yet. -->
+      <p class="text-xs text-gray-500" role="status" data-testid="org-roi-project-picker-match-count">{{ matchCountLabel() }}</p>
+
+      @if (matches().length > 0) {
+        <ul class="flex max-h-48 flex-col gap-1 overflow-y-auto" data-testid="org-roi-project-picker-matches">
+          @for (option of matches(); track option.projectId) {
+            <li>
+              <button
+                type="button"
+                (click)="add(option.projectId)"
+                class="flex w-full items-baseline justify-between gap-4 rounded px-2 py-1.5 text-left text-sm text-gray-700 hover:bg-white"
+                [attr.data-testid]="'org-roi-project-picker-add-' + option.projectId">
+                <span>{{ option.projectName }}</span>
+                <span class="text-xs font-medium text-gray-500">{{ option.amount }}</span>
+              </button>
+            </li>
+          }
+        </ul>
+      }
+    }
+  </div>
+</div>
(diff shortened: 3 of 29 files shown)


apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-table/org-roi-projects-table.component.ts
```diff
@@ -0,0 +1,162 @@
...
+    return Number(a > b) - Number(a < b);
+  }
+
+  private compareRows(a: OrgLensRoiProjectTableRow, b: OrgLensRoiProjectTableRow, direction: number, field: OrgLensRoiProjectSortField): number {
+    const tieBreak = this.compareProjectIds(a.projectId, b.projectId);
+
+    if (field === 'name') {
+      // Codepoint order again, for the same SSR reason as the tie-break above.
+      return (Number(a.projectName > b.projectName) - Number(a.projectName < b.projectName)) * direction || tieBreak;
+    }
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411`. Agreed on the substance — codepoint order sorts by ASCII value, so every capitalised name came ahead of every lowercase one in a user-facing column. The name comparison is now `localeCompare(…, 'en-US')`, pinned so Node and the browser agree.

One correction for the record: there is no written "SSR locale rule" in `.claude/rules/` or `docs/reviews/` — I grepped for it. The convention is real by practice (`localeCompare` appears in 8+ components) and the defect was real, so the call was right, but the citation points at something that doesn't exist. The codepoint comparison stays on the `projectId` tie-break, where the key is opaque and collation carries no meaning.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-bubble/org-roi-projects-bubble.component.html
```diff
@@ -0,0 +1,66 @@
...
+
+    <div class="h-[420px] w-full" role="img" [attr.aria-label]="chartSummaryLabel()" data-testid="org-roi-projects-bubble-chart">
+      <lfx-chart type="bubble" [data]="chartData()" [options]="chartOptions()" height="100%" />
+    </div>
+
+    <p class="text-xs text-gray-500" data-testid="org-roi-projects-bubble-scale-note">
+      Both axes are logarithmic: return spans several orders of magnitude across a portfolio, and on a linear scale every project but the largest would sit in
+      the corner. Each bubble's area is proportional to its net return, and an amber bubble lost money. Projects further above the diagonal returned more per
+      dollar invested.
+    </p>
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411`. Correct, and the same point as the Copilot thread on this file. The note pointed at a break-even diagonal that is never drawn — and could not be drawn meaningfully, since the two axes are scaled independently, so the corner-to-corner line is not `y = x`. Rather than add a reference line to a plot where it would need its own explanation, the copy now describes the encoding that actually exists: position towards the top-left means more return per dollar.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-bubble/org-roi-projects-bubble.component.html
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    <div class="h-[420px] w-full" role="img" [attr.aria-label]="chartSummaryLabel()" data-testid="org-roi-projects-bubble-chart">
+      <lfx-chart type="bubble" [data]="chartData()" [options]="chartOptions()" height="100%" />
+    </div>
+
+    <p class="text-xs text-gray-500" data-testid="org-roi-projects-bubble-scale-note">
+      Both axes are logarithmic: return spans several orders of magnitude across a portfolio, and on a linear scale every project but the largest would sit in
+      the corner. Each bubble's area is proportional to its net return, and an amber bubble lost money. Projects further above the diagonal returned more per
+      dollar invested.
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411`, together with the Bugbot thread above. Both parts were right: the diagonal is never rendered, and `4 + 18·√(|profit| / max)` is neither strictly area-proportional (the +4 base offset) nor signed (it uses the absolute value). The note now says bubbles are sized by how large a project's net return is, that amber lost money, and that position towards the top-left means more return per dollar — all of which the chart does do.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-sankey/org-roi-projects-sankey.component.ts
```diff
@@ -0,0 +1,163 @@
...
+   *
+   * A zero flow is dropped. A sankey link of zero width is not drawn but still contributes a node,
+   * leaving a labelled stub attached to nothing.
+   */
+  protected readonly links: Signal<OrgLensRoiProjectFlowLink[]> = computed(() => {
+    const projects = this.drawnProjects();
+    if (this.measure() === 'return') {
+      return projects
+        .filter((project) => project.totalReturn > 0)
+        .map((project) => ({ from: project.projectName, to: ORG_LENS_ROI_SANKEY_ORG_NODE, flow: project.totalReturn }));
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411`. This was the sharpest catch of the set — I had noted the duplicate-name merge and missed the more serious half, that a project named "Code Contribution" or "Your organization" would join the category or org node and produce a cyclic flow.

Nodes are now keyed by type-prefixed stable ids (`project:<projectId>`, `category:<type>`, `org`), with a `labels` map supplying display names to the chart, the tooltip and the `sr-only` table. The prefixes live in shared constants.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-bar/org-roi-projects-bar.component.html
```diff
@@ -0,0 +1,71 @@
...
+         out of the visual layout without hiding it from assistive technology. -->
+    <table class="sr-only" data-testid="org-roi-projects-bar-table">
+      <caption>
+        Investment and return by project
+      </caption>
+      <thead>
+        <tr>
+          <th scope="col">Project</th>
+          <th scope="col">Investment</th>
+          <th scope="col">Return</th>
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411`. Agreed, and this was the accessibility gap that mattered most — the stack composition *is* what this chart shows, so a text alternative carrying only totals described a simpler chart than the one on the page.

The table now has a column per category actually drawn, in canonical stack order, with absent categories rendered as a zero rather than a blank so every row has the same shape.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-table/org-roi-projects-table.component.html
```diff
@@ -0,0 +1,62 @@
...
+
+<div class="flex flex-col gap-3" data-testid="org-roi-projects-table">
+  @if (!hasRows()) {
+    <div class="flex items-center gap-2 py-6 text-sm text-gray-600" data-testid="org-roi-projects-table-empty">
+      <i class="fa-light fa-table text-gray-400" aria-hidden="true"></i>
+      <span>No projects are available for this organization.</span>
+    </div>
+  } @else {
+    <p class="text-xs text-gray-500" data-testid="org-roi-projects-table-count">
+      {{ totalRecords().toLocaleString('en-US') }} projects. Sorting and paging apply to this visit only.
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411` — `countLabel` computed, bound as a signal. Verified against `docs/reviews/frontend-checklist.md` §4.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-sankey/org-roi-projects-sankey.component.html
```diff
@@ -0,0 +1,72 @@
...
+        [attr.data-testid]="'org-roi-projects-sankey-measure-' + option">
+        {{ measureLabels[option] }}
+      </button>
+    }
+  </div>
+
+  @if (!hasLinks()) {
+    <div class="flex items-center gap-2 py-6 text-sm text-gray-600" data-testid="org-roi-projects-sankey-empty">
+      <i class="fa-light fa-diagram-project text-gray-400" aria-hidden="true"></i>
+      <span>No {{ measureLabel().toLowerCase() }} flow to draw for the selected projects.</span>
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411` — `measureLabelLowercase` computed, bound as a signal.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-bubble/org-roi-projects-bubble.component.html
```diff
@@ -0,0 +1,66 @@
...
+    </p>
+
+    @if (flooredNames().length > 0) {
+      <p class="flex items-start gap-2 rounded-md bg-gray-50 px-3 py-2.5 text-xs text-gray-600" data-testid="org-roi-projects-bubble-floored-note">
+        <i class="fa-light fa-circle-info mt-0.5 text-gray-400" aria-hidden="true"></i>
+        <span
+          >A logarithmic axis has no zero, so these projects are drawn at the edge rather than at their true position: {{ flooredNames().join(', ') }}. Their
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411` — the join moved into a `flooredLabel` computed, with a `hasFlooredPoints` predicate replacing the `.length > 0` check in the template.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-project-picker/org-roi-project-picker.component.html
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  <div class="flex flex-col gap-2">
+    <label class="flex flex-col gap-1 text-xs text-gray-600">
+      <span class="sr-only">Search projects to add</span>
+      <input
+        type="search"
+        [ngModel]="query()"
+        (ngModelChange)="query.set($event)"
+        placeholder="Add a project by name"
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411`. Confirmed against `docs/reviews/frontend-checklist.md` §14.1, including the note that the wrappers require `FormGroup`/`FormControl` and do not support `ngModel` — so the search box is now `lfx-input-text` over a reactive control, matching the committee-members pattern you pointed at.

One knock-on worth flagging: the wrapper emits `data-test`, not `data-testid`, so the e2e selectors for this field moved to `page.locator('[data-test=…]')`, which is how the other specs in this repo target wrapper-backed inputs.

apps/lfx-one/e2e/helpers/org-roi.helper.ts
```diff
@@ -2,15 +2,31 @@
...
+  ORG_LENS_ROI_PROJECT_BUBBLE_MAX_POINTS,
+  ORG_LENS_ROI_PROJECT_PICKER_DEFAULT_COUNT,
+  ORG_LENS_ROI_PROJECT_SANKEY_MAX_PROJECTS,
+} from '@lfx-one/shared/constants/org-lens-roi.constants';
+
+export const PICKER_DEFAULT_COUNT = ORG_LENS_ROI_PROJECT_PICKER_DEFAULT_COUNT;
+export const BAR_MAX_ROWS = ORG_LENS_ROI_PROJECT_BAR_MAX_ROWS;
+export const SANKEY_MAX_PROJECTS = ORG_LENS_ROI_PROJECT_SANKEY_MAX_PROJECTS;
+export const BUBBLE_MAX_POINTS = ORG_LENS_ROI_PROJECT_BUBBLE_MAX_POINTS;
+export const NO_VALUE = ORG_LENS_ROI_NO_VALUE;
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `d2cee5411`, and you were right to push back — this contradicted an earlier review round that had told me to stop hand-copying these values, and your framing is the better one.

The distinction I landed on: the four ceilings (bar 25, sankey 12, bubble 250, default selection 5) are a **product contract**, so they are written down in the helper with a comment saying why; importing them made the truncation assertions restate the implementation, exactly as you describe. `NO_VALUE` stays imported — it is a rendering detail rather than a promise to the viewer, and drift there is a false failure rather than a missed contract change.

.github/workflows/quality-check.yml
```diff
@@ -52,17 +52,38 @@ jobs:
...
+    name: E2E Tests
+    needs: quality-checks
+    # A called workflow cannot hold more permission than its caller, so these must be granted here
+    # even though e2e-tests.yml declares them: id-token for the AWS OIDC secret fetch, and
+    # pull-requests/issues for the results comment.
+    permissions:
+      id-token: write
+      contents: read
+      issues: write
+      pull-requests: write
```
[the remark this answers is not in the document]
luismoriguerra: Agreed, and the CI change is **removed from this PR** in `d2cee5411` — both workflow files are back to their `main` state.

Your read is right, and the run this PR triggered before I reverted proves it rather than leaving it theoretical: the job completed OIDC auth, pulled Auth0/Supabase/M2M credentials from AWS Secrets Manager, passed secret validation, and went on to run PR-controlled `yarn install` and `yarn build`. So the credentials genuinely do reach the untrusted revision.

Fork PRs would not receive them, so the practical exposure is same-repo branches plus anything in a PR's dependency tree — but that is a security decision that deserves its own review, and it should not have ridden along with a feature PR. It will come back separately, with the protected-environment/manual-approval option you suggest as the starting point.

apps/lfx-one/src/app/shared/guards/org-lens-roi-enabled.guard.ts
```diff
@@ -29,8 +29,18 @@ export const orgLensRoiEnabledGuard: CanMatchFn = async () => {
...
+    //
+    // This flag is a dark launch rather than a fully-enabled one, so failing open can show an
+    // unfinished page to an Org Lens viewer when LaunchDarkly is slow. That is bounded: every ROI
+    // endpoint runs its own `assertOrgLensRead` regardless of any flag, so what leaks early is an
+    // incomplete view, never data the viewer could not already request. Logged so the frequency is
+    // visible rather than assumed.
     if (!ready) {
-      return router.parseUrl('/org/overview');
+      console.warn('orgLensRoiEnabledGuard: LD provider not ready after 5 s — failing open');
+      return true;
```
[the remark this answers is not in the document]
luismoriguerra: Agreed, and reverted to fail-closed in `a46304d3`. Your distinction is the right one: `myClasEnabledGuard` can fail open because its flag is enabled for all production users, so allowing the route grants access those users already have. This one is a dark launch, where the same behaviour turns a LaunchDarkly outage into a release. The comment now states that contrast explicitly, and notes to revisit at GA when the trade flips.

You also caught a second, separate defect: the fail-open path returned before the flag was read at all, so it overrode a pinned `false` — the one value that must never be overridden. The guard now consults an explicit local override **before** waiting on the provider, so a readiness timeout can no longer answer for a value someone set deliberately.

For context on why that override exists: it is a non-production-only `localStorage` value, added in this PR because a flag-gated route is otherwise untestable — the SDK evaluates against an anonymous context before the authenticated one, so a flag targeted at named users reads false in that window and this guard redirects before the real value lands.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-bar/org-roi-projects-bar.component.ts
```diff
@@ -0,0 +1,198 @@
...
+  /** Only the categories actually present, so an organization with no events gets no dead legend entry. */
+  private readonly presentTypes: Signal<OrgLensRoiProjectBarSegment[]> = computed(() => {
+    const byType = new Map<string, OrgLensRoiProjectBarSegment>();
+    for (const row of this.rows()) {
+      for (const segment of row.segments) {
+        if (segment.expenditure !== 0 && !byType.has(segment.type)) byType.set(segment.type, segment);
+      }
+    }
+    // Canonical order, not first-seen: the stack must read the same way on every project.
+    return ORG_LENS_ROI_CONTRIBUTION_TYPES.map((type) => byType.get(type)).filter((segment): segment is OrgLensRoiProjectBarSegment => segment !== undefined);
```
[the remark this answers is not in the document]
luismoriguerra: Fixed in `a46304d3` — unknown types are now appended after the canonical order rather than filtered out.

Verified your reference and the case is stronger than stated: `isCategoryRows` does not merely tolerate unrecognised types, its docblock records dropping one as the *worse* option precisely because "the category total would silently stop reconciling with the KPI figure". This component was reintroducing that failure one layer up — a ninth warehouse category would have disappeared from the stack **and** from the accessible table, while `unreconciled` (which sums the payload's own segments, not the drawn ones) stayed quiet, so nothing would have surfaced the shortfall.

apps/lfx-one/src/app/shared/services/feature-flag.service.ts
```diff
@@ -62,11 +90,27 @@ export class FeatureFlagService {
    * Get a boolean feature flag as a signal
    * Returns computed signal that updates automatically when flag changes
    */
```
audigregorie: **Major — Orphaned JSDoc: `getBooleanFlag`'s doc is now stranded above `getFlagOverride`.**

The original JSDoc (`Get a boolean feature flag as a signal / Returns computed signal that updates automatically when flag changes`) was written for `getBooleanFlag`. The new `getFlagOverride` method was inserted between the doc and its method, with its own JSDoc directly below. As a result, `getBooleanFlag` (now at line 104) has lost its documentation, and `getFlagOverride` appears to have two stacked JSDoc blocks — the orphaned one describing a different method.

Move the original JSDoc down to `getBooleanFlag`:

```suggestion
   */
  /**
   * A value pinned locally for this flag, or `undefined` when none is set.
   *
   * Exposed so a route guard can consult it *before* waiting on the provider. A guard that decides
   * on a readiness timeout would otherwise ignore a pinned value entirely — including a pinned
   * `false`, which is the case that must never be overridden.
   */
  public getFlagOverride(key: string): boolean | undefined {
    return readFlagOverride(key);
  }

  /**
   * Get a boolean feature flag as a signal
   * Returns computed signal that updates automatically when flag changes
   */
  public getBooleanFlag(key: string, defaultValue: boolean = false): Signal<boolean> {
```

(Replace the orphaned JSDoc at lines 89-92 with nothing, and add it back above `getBooleanFlag`.)

packages/shared/src/constants/org-lens-roi.constants.ts
```diff
@@ -101,3 +101,107 @@ export const ORG_LENS_ROI_KPI_EXPLANATION = {
...
+
+/**
+ * Translucent fills for the efficiency plot, where bubbles overlap and an opaque one hides those
+ * behind it. These are the alpha forms of `emerald[600]` and `amber[500]`, which the same points
+ * use — at full opacity — for their borders.
+ */
+export const ORG_LENS_ROI_BUBBLE_FILL = {
+  profitable: 'rgba(0, 153, 102, 0.45)',
+  lossMaking: 'rgba(254, 154, 0, 0.55)',
+} as const;
```
audigregorie, line 210: **Minor — Hardcoded rgba literals duplicate `lfxColors` values.**

`rgba(0, 153, 102, 0.45)` is `lfxColors.emerald[600]` (`#009966`) at 45% alpha, and `rgba(254, 154, 0, 0.55)` is `lfxColors.amber[500]` (`#FE9A00`) at 55% alpha. The comment correctly documents the derivation, but if the `lfxColors` palette is ever re-scaled, these literals will silently diverge from the borders (which read from `lfxColors` directly via `ORG_LENS_ROI_RETURN_COLOR` / `lfxColors.amber[500]`).

The repo rule says "No hard-coded hex colors — use `lfxColors` scales"; these are rgba not hex, and `lfxColors` has no alpha variants, so this is a pragmatic escape rather than a violation. A small `hexToRgba(hex, alpha)` helper in `@lfx-one/shared/utils` would let both values derive from `lfxColors.emerald[600]` / `lfxColors.amber[500]` and keep the rule's intent. Not blocking — the comment is honest about the trade-off.

apps/lfx-one/src/app/shared/components/chart/chart.component.ts
```diff
@@ -28,4 +32,13 @@ export class ChartComponent {
...
   public readonly height = input<string>('100%');
+
+  /**
+   * PrimeNG types its own `type` input against a fixed union of core Chart.js types, so a type
+   * contributed by a registered plugin cannot be passed through it as-is. Only the compile-time
+   * union is narrow — at runtime PrimeNG forwards the string to Chart.js, which accepts any
+   * registered controller id. Widened here, at the one boundary that needs it, rather than by
+   * loosening the `ChartType` this component exposes to its callers.
+   */
+  protected readonly primeChartType: Signal<any> = computed(() => this.type());
```
audigregorie, line 43: **Minor — `Signal<any>` discards all type information at the PrimeNG boundary.**

The comment explains why the widening is needed (PrimeNG's `type` input is typed against a fixed union that doesn't include `sankey`). Using `any` is one way through, but it means any future `ChartType` value (or a typo) will compile silently. A narrower escape would be to type it as the union PrimeNG actually accepts and cast at the binding site, or to use `Signal<ChartType>` with a type assertion in the template — but PrimeNG's `p-chart` input is itself `any`-ish, so this is a minor style point, not a correctness risk.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-sankey/org-roi-projects-sankey.component.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  public setMeasure(measure: OrgLensRoiProjectSankeyMeasure): void {
+    this.measure.set(measure);
+  }
+
+  /** Categories keep the colour they carry on the comparison view; project nodes stay neutral. */
+  private nodeColor(node: string): string {
+    if (node === ORG_LENS_ROI_SANKEY_ORG_NODE_KEY) return this.measure() === 'return' ? ORG_LENS_ROI_RETURN_COLOR : lfxColors.blue[600];
+    const type = node.startsWith(ORG_LENS_ROI_SANKEY_CATEGORY_NODE_PREFIX) ? node.slice(ORG_LENS_ROI_SANKEY_CATEGORY_NODE_PREFIX.length) : '';
+    return ORG_LENS_ROI_CATEGORY_COLOR[type as OrgLensRoiContributionType] ?? lfxColors.gray[400];
```
audigregorie: **Minor — `type as OrgLensRoiContributionType` is an unchecked cast on an arbitrary string.**

`type` is whatever follows `category:` in the node key — for categories produced by this component that's always a valid `OrgLensRoiContributionType`, but the `?? lfxColors.gray[400]` fallback is written for the case where it isn't. The cast suppresses the TS check that would otherwise force the fallback to be exhaustive. A `Map<string, string>` lookup (or `ORG_LENS_ROI_CATEGORY_COLOR[type] ?? lfxColors.gray[400]` with `type: string`) would be honest about the runtime type without changing behavior. Not blocking — the fallback makes it safe.

audigregorie (commented):
## Code Review Summary

This is a well-crafted PR. The four-view projects section is cleanly decomposed (one picker, four chart components, one orchestrating section), the `linkedSignal`-based selection sentinel correctly distinguishes "never chosen" from "deliberately emptied" across org/method switches and failed-then-retried reads, and every chart view ships an accessible text equivalent alongside the canvas. The e2e helper rewrite (correct cookie key, flag-override pinning, async guard assertion) turns a suite that never ran into one that actually exercises the page. The judgement calls called out in the PR description (dual-axis bar, asymmetric sankey, log-floor disclosure, no client-side rescaling) are the right ones and are honestly disclosed in the UI.

No Critical issues found. Findings below — one Major (documentation defect), four Minor.

### What's done well

- The `selectionOverride` linkedSignal is the strongest part of this PR. The five-case comment is precise, the "failed read emits `[]` which must not be confused with None" invariant is real, and the e2e test for the retry path is exactly the regression guard that would catch its breakage.
- Accessible text equivalents (`sr-only` tables with `<caption>`) on every chart view, plus `role="img"` + `aria-label` on the canvases — this is the right pattern for chart.js in an Angular SSR app.
- The sankey node-key prefixing (`project:`, `category:`, `org`) with a separate label map is a real bug fix (name-collision cycles) rather than a cosmetic refactor, and the comment explains the failure mode it prevents.

### Major — outside the diff

None.

### Minor — outside the diff

- **`data-test` vs `data-testid` inconsistency on the picker search input.** The `InputTextComponent` wrapper exposes a `dataTest` input that renders as `data-test=...`, while every other test id in this PR uses `data-testid`. The e2e specs have to switch from `getByTestId(...)` to `page.locator('[data-test="..."]')` for that one element. This is a pre-existing issue in `InputTextComponent` (not introduced here), but it's now the only element on the page that breaks the `getByTestId` convention. Worth a follow-up to add a `data-testid` passthrough on `InputTextComponent`.
- **`readFlagOverride` is not reactive in `getBooleanFlag`.** The computed reads `this.context()` for reactivity, then reads `localStorage` via `readFlagOverride`. localStorage is not a signal, so the override is only re-read when `context` changes. For e2e (override set before load) this is fine; in dev, clearing the override won't revert the flag until the context changes. Minor dev-only UX, not a bug.

apps/lfx-one/src/app/shared/services/feature-flag.service.ts
luismoriguerra: Fixed in `24b8bc43` — confirmed exactly as you describe. Inserting `getFlagOverride` ahead of the existing block left `getBooleanFlag` undocumented and stacked two blocks on the new method. The original doc is back above `getBooleanFlag`.

packages/shared/src/constants/org-lens-roi.constants.ts
luismoriguerra: Fixed in `24b8bc43`, and better than the trade-off the comment was defending: **`hexToRgba(hex, alpha)` already exists** in `packages/shared/src/utils/color.utils.ts`, so both fills now derive from `lfxColors.emerald[600]` and `lfxColors.amber[500]` — the same entries the borders read. The drift you describe is gone rather than documented.

Verified the output is byte-identical to the literals it replaces (`rgba(0, 153, 102, 0.45)` and `rgba(254, 154, 0, 0.55)`), so there is no visual change. Imported deep from `../utils/color.utils` rather than the barrel, since this is a constants module and the barrel reaches Angular-dependent code; `color.utils` imports nothing, so no cycle.

audigregorie (commented):
## Code Review Summary

High-quality implementation. The four views share state cleanly through the `linkedSignal`-backed selection, every truncation is disclosed rather than silent, accessibility is treated as a first-class requirement (per-canvas `aria-label` + `sr-only` tables carrying true figures), and metric-layer figures are never re-derived client-side. The e2e coverage is genuinely thorough — including the hard recovery path (failed read retried under another method) and the "emptied selection stays empty across method switch" invariant. No Critical or Major findings; the five threads below are Minor maintainability/style notes.

### Minor — outside the diff

- `chartjs-chart-sankey` is pinned exactly (`0.15.0`, no `^`) — good practice for a pre-1.0 dep, and correctly flagged in the PR description for dependency-owner review.
- `yarn build` passes, which exercises the SSR bundle — so the static top-level `import { Flow, SankeyController } from 'chartjs-chart-sankey'` in `chart.component.ts` is empirically SSR-safe at this version. The SSR-safety rule flags this pattern as risky in general, so worth re-verifying if the plugin ever bumps.

### What's done well

- Disclosure over silent truncation: every cap (bar 25, sankey 12, bubble 250) names the withheld count, and the search match-count reports true matches rather than the capped list length.
- The `linkedSignal` selection sentinel distinguishing "never chosen" (`null`) from "deliberately emptied" (`[]`) is exactly the state-machine thinking that prevents the "charts stay blank after retry" bug.
- No client-side rescaling: `roi`/`bcr`/`profit` read from the payload and never re-derived; `unreconciled` surfaces a stack that doesn't sum as a metric defect rather than papering over it.
- Accessibility: each canvas has an `aria-label` describing the encoding, plus an `sr-only` table with true (unfloored) figures. The button-group-vs-tablist decision is documented with reasoning, not just the choice.

packages/shared/src/constants/org-lens-roi.constants.ts
audigregorie, line 210: **Minor — `ORG_LENS_ROI_BUBBLE_FILL` hard-codes rgba duplicates of `lfxColors` hex values**

`'rgba(0, 153, 102, 0.45)'` is `lfxColors.emerald[600]` (`#009966`) at 45% alpha, and `'rgba(254, 154, 0, 0.55)'` is `lfxColors.amber[500]` (`#FE9A00`) at 55% alpha. The comment correctly documents the relationship, but the values themselves are duplicated — if the brand palette changes in `colors.constants.ts`, these rgba literals won't update with it, which is exactly what the "never hard-code hex / reference scale names" rule in `.claude/rules/styling.md` is meant to prevent.

Chart.js needs an rgba string for translucency, so a hex literal can't be passed directly. A small `hexToRgba(hex, alpha)` helper in `color.utils` would let this stay derived:

```suggestion
export const ORG_LENS_ROI_BUBBLE_FILL = {
  profitable: hexToRgba(lfxColors.emerald[600], 0.45),
  lossMaking: hexToRgba(lfxColors.amber[500], 0.55),
} as const;
```

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-section/org-roi-projects-section.component.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private initProjectRows(): Signal<OrgLensRoiProjectRow[]> {
+    // Keyed by string, not the account object: that object is rewritten in place and would retrigger the fetch.
+    const requestKey$ = toObservable(computed(() => `${this.accountContext.selectedAccount()?.accountId ?? ''}|${this.method()}`));
+
+    return toSignal(
+      requestKey$.pipe(
+        map((key) => key.split('|') as [string, OrgLensRoiMethod]),
+        filter(([orgUid]) => !!orgUid),
```
audigregorie: **Minor — composite request key built by string concatenation + `split`**

`${orgUid}|${method}` then `key.split('|') as [string, OrgLensRoiMethod]` is a stringly-typed composite key. It's practically safe today — `method` is constrained to the `logit`/`direct` const union and `orgUid` is a Salesforce ID (alphanumeric) — but the `as [string, OrgLensRoiMethod]` cast hides any future breakage if either value ever contains `|`. This also duplicates the same pattern in `org-roi.component.ts`'s `initPortfolio`.

A typed object through `distinctUntilChanged` with a custom comparer would be type-safe and remove the duplication:

```suggestion
    const requestKey$ = toObservable(
      computed(() => ({ orgUid: this.accountContext.selectedAccount()?.accountId ?? '', method: this.method() }))
    ).pipe(
      filter(({ orgUid }) => !!orgUid),
      distinctUntilChanged((a, b) => a.orgUid === b.orgUid && a.method === b.method)
    );
```

Low priority — flagging mainly so the pattern doesn't get copy-pasted into a third place where the inputs aren't as constrained.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-sankey/org-roi-projects-sankey.component.ts
luismoriguerra: Fixed in `24b8bc43` — now a `Record<string, string>` lookup, no assertion. Your reasoning is right and slightly understated: the fallback is not only for keys this component did not produce. The BFF deliberately forwards contribution types that `ORG_LENS_ROI_CATEGORY_COLOR` does not know yet, so an unknown type is a real runtime path here, and the assertion was suppressing the check that keeps the fallback honest.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-section/org-roi-projects-section.component.ts
```diff
@@ -0,0 +1,155 @@
...
+          this.roiService.getProjects(orgUid, method).pipe(
+            tap(() => this.loading.set(false)),
+            map((projects) => projects.rows),
+            catchError((error: unknown) => {
+              console.error('Failed to load ROI projects section', error);
+              this.loading.set(false);
+              // Only a 403 may show the no-access message; a 503 must not.
+              if ((error as { status?: number })?.status === 403) this.forbidden.set(true);
+              else this.failed.set(true);
+              return of([] as OrgLensRoiProjectRow[]);
```
audigregorie, line 152: **Minor — imperative `tap`-based loading/error state vs the declarative state pattern**

`.claude/rules/angular-standards.md` recommends: *"Declarative state pattern: model `loading` + `error` + `data` as one state object with `startWith` — not imperative `.set()` calls."* Here `loading`/`failed`/`forbidden` are three separate signals driven by `tap` side-effects, with the ordering constraint (failed read must not overwrite the selection) handled imperatively in `catchError`.

This matches the established pattern in the parent `org-roi.component.ts`'s `initPortfolio`, so it's a deliberate, consistent choice rather than an oversight — and the comments do a good job explaining the ordering constraints. Flagging it only as a deviation from the documented standard worth a conscious decision: if the declarative pattern were adopted, a single `LoadableState` (`'loading' | 'forbidden' | 'failed' | 'done'`) reduced in the `switchMap` would make the four-way template branch a single `@switch` on one signal and remove the "did we reset all three flags?" surface area.

No change requested — just want it acknowledged as a trade-off rather than missed.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-sankey/org-roi-projects-sankey.component.ts
```diff
@@ -0,0 +1,184 @@
...
+        // Maps the type-prefixed node keys back to the names a viewer reads.
+        labels: this.nodeLabels(),
+        // Node heights are driven by the larger of a node's inbound and outbound totals, so a
+        // category that feeds several projects is not drawn smaller than the flows leaving it.
+        size: 'max',
+      },
+    ],
+  }));
+
+  protected readonly chartOptions: Signal<unknown> = computed(() => {
```
audigregorie: **Minor — `chartData` and `chartOptions` typed as `Signal<unknown>`**

Both computeds return `unknown`, which loses all type safety on the sankey's data shape (`{ datasets: [{ data: Flow[]; colorFrom; colorTo; colorMode; size; labels }] }`) and options. The PR description notes `chartjs-chart-sankey` ships its own types, so `Flow` is already imported. The plugin also augments Chart.js' `ChartType` registry when imported, so `ChartData<'sankey'>` / `ChartOptions<'sankey'>` may be available — worth checking. If not, a small structural interface for the dataset would at least keep the `colorFrom`/`colorTo`/`size` keys honest.

Low priority since the values flow into `lfx-chart`'s `any`-typed inputs anyway, but tightening here would catch typos in the dataset shape (e.g. `colorMode: 'gradient'` vs a future rename) at compile time rather than at "the sankey renders blank" time.

apps/lfx-one/src/app/shared/components/chart/chart.component.ts
```diff
@@ -1,12 +1,13 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { Component, input } from '@angular/core';
+import { Component, computed, input, Signal } from '@angular/core';
 import { ChartModule } from 'primeng/chart';
 import { Chart } from 'chart.js';
+import { Flow, SankeyController } from 'chartjs-chart-sankey';
```
audigregorie, line 7: **Minor — static top-level import of `chartjs-chart-sankey` carries forward SSR risk**

`.claude/rules/ssr-safety.md` is explicit: *"A static top-level `import` of such a library will crash SSR even if the code path that uses it is browser-only."* The PR description confirms `yarn build` passes, which exercises the SSR bundle — so the plugin is empirically SSR-safe at `0.15.0`, and `Chart.register(SankeyController, Flow)` running in a static initializer works because `chart.js` itself is already imported statically here.

This is verified-safe today, so no change requested. The risk is forward: if a future bump of `chartjs-chart-sankey` introduces a `window` reference at import time, the SSR build will break with no obvious local cause. Worth a comment on the import line (or a note in the dependency-owner review) that the SSR-safety of this import was verified at `0.15.0` and must be re-checked on any bump.

apps/lfx-one/src/app/shared/components/chart/chart.component.ts
luismoriguerra: Keeping `any` here, with reasoning — happy to revisit if you disagree.

PrimeNG declares `type` as a closed union in its `.d.ts` with no exported type name, so narrowing means transcribing a third-party union into `@lfx-one/shared` (where this repo requires types to live) purely to describe someone else's input — and it would need re-checking against every PrimeNG upgrade. The same component already types `data`, `options` and `style` as `any` at that identical boundary, so this matches the file's existing idiom rather than departing from it.

Your point about a typo compiling silently is fair, but it is bounded: `primeChartType` is `protected` and exists only for that one template binding, and callers still go through the public `ChartType` union, which stays narrow and does include `sankey`. So a wrong value has to be written inside this component to get through.

packages/shared/src/constants/org-lens-roi.constants.ts
luismoriguerra: Already addressed — this landed in `24b8bc43`, before this thread was posted, so it is reading the older commit. Both fills now derive from `lfxColors` through the existing `hexToRgba` helper in `packages/shared/src/utils/color.utils.ts`, so a palette change moves fill and border together. Verified byte-identical output, no visual change.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-sankey/org-roi-projects-sankey.component.ts
luismoriguerra: Fixed in `299b27c7` — `chartData` is now typed, so the object literal is checked.

One correction on the route there, since it matters for anyone trying the same: **`chartjs-chart-sankey` does not augment Chart.js's `ChartTypeRegistry`.** Its `.d.ts` is three lines — a `export type { … }` re-export plus `SankeyController` and `Flow` — with no `declare module 'chart.js'`, so `ChartData<'sankey'>` does not resolve to a sankey shape. `Flow` is the rendering element, not the dataset type.

So instead of the registry, the dataset is described by `OrgLensRoiSankeyChartData` in `@lfx-one/shared` (interfaces live there per repo convention), which gets the real benefit you were after. `chartOptions` stays `unknown`: it is plugin-specific Chart.js options with no exported type to lean on, and inventing a hand-written mirror of it would be a maintenance burden that buys little.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-section/org-roi-projects-section.component.ts
luismoriguerra: Fixed in `299b27c7` — the key is now a typed `{ orgUid, method }` pair, so nothing is parsed back out or cast.

Worth recording why it was a string, because the naive fix regresses: `selectedAccount()` is an object rewritten **in place**, so the computed re-runs on changes that leave both fields identical. A string gave value-equality dedup for free; an object literal produces a fresh reference each time and would re-emit and refetch. The dedup is now explicit via `distinctUntilChanged` comparing both fields — which is also what keeps this clear of the `toObservable` double-emission pattern in `docs/reviews/knowledge-base/frontend-state-and-timing.md`.

The projects donut still uses the string form. I left it alone rather than widen this PR into merged US4 code, but the two should converge — happy to follow up across both.

apps/lfx-one/src/app/shared/components/chart/chart.component.ts
luismoriguerra: Checked this rather than relying on the build alone, and I'd keep the static import — evidence below, happy to reconsider.

- The plugin's ESM bundle contains **zero** references to `document`, `window`, `navigator` or `localStorage`.
- It imports cleanly under plain Node: `import('chartjs-chart-sankey')` resolves and exports `Flow, SankeyController`.
- `chart.js` itself is already statically imported on line 6 of this same file. Dynamically importing its plugin while the core stays static would be inconsistent without being safer — the plugin cannot be the SSR hazard if the library it extends is already loaded.

`ssr-safety.md` targets libraries that touch browser globals at module scope; this one does not. Its "`yarn build` exercises the SSR bundle" note is the check, and it passes.

apps/lfx-one/src/app/modules/dashboards/org/org-roi/components/org-roi-projects-section/org-roi-projects-section.component.ts
luismoriguerra: Leaving this as-is, with reasoning — say the word if you disagree.

I could not find the rule: there is no `.claude/rules/angular-standards.md` in the repo (the directory holds `commit-workflow`, `component-organization`, `development-rules`, `logging-patterns`, `skill-guidance`, `ssr-safety`, `styling`), and neither `.claude/rules/` nor `docs/reviews/` contains the quoted "declarative state pattern … with `startWith`" guidance.

The closest documented guidance points the other way: `docs/reviews/knowledge-base/frontend-state-and-timing.md` flags `toObservable(signal).pipe(startWith(...))` as a double-emission bug, because `toObservable` already emits synchronously on subscribe — so adding `startWith` fires downstream work twice on mount. This component deliberately does not use it.

The three flags also mirror the merged projects donut exactly, so changing one would leave two sibling components in the feature diverging. If the pattern is one the team wants, it is worth applying across the feature in one pass rather than to the newest component only.

audigregorie (approved):

PR #1390, opened 2026-08-11
fix(guards): fail open on ld timeout in my-clas-enabled guard

andrest50:
## Summary

- `myClasEnabledGuard` previously redirected to `/profile` (→ `/profile/attributions`) when the LaunchDarkly provider did not become ready within 5 s
- This created an intermittent race condition during SSR hydration: the guard ran concurrently with `FeatureFlagService.initialize()`, and on slow connections or transient LD errors the timeout fired before `providerReady` was set
- Changed the timeout path to fail **open** (`return true`) — since `my-clas-enabled` is enabled for all production users, allowing the route when LD is unreachable is the correct behaviour; the guard still redirects when LD is available and the flag is definitively off

## Ticket

[LFXV2-3105](https://linuxfoundation.atlassian.net/browse/LFXV2-3105)

## Test plan

- [ ] Navigate to `/profile/clas` on a cold tab load — confirm no redirect to attributions
- [ ] Verify the CLAs tab still hides correctly when `my-clas-enabled` is off (flag eval path unchanged)
- [ ] Check no TypeScript errors (`yarn check-types`)

[LFXV2-3105]: https://linuxfoundation.atlassian.net/browse/LFXV2-3105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/shared/guards/my-clas-enabled.guard.ts
+++ b/apps/lfx-one/src/app/shared/guards/my-clas-enabled.guard.ts
@@ -12,9 +12,10 @@ import { FeatureFlagService } from '../services/feature-flag.service';
 
 /**
  * CanMatch guard for the Profile "My CLAs" tab (`/profile/clas`), gating the
- * read-only EasyCLA view behind the `my-clas-enabled` flag. Mirrors
- * orgLensEnabledGuard: SSR defers to the browser, the browser waits for the
- * flag provider to be READY, and it fails closed to the default profile tab.
+ * read-only EasyCLA view behind the `my-clas-enabled` flag. SSR defers to the
+ * browser; the browser waits up to 5 s for the flag provider to be READY, then
+ * fails open (allows the route) if LD is unreachable so users aren't silently
+ * redirected away from a feature that is enabled for them in production.
  */
 export const myClasEnabledGuard: CanMatchFn = async () => {
   const platformId = inject(PLATFORM_ID);
@@ -36,9 +37,12 @@ export const myClasEnabledGuard: CanMatchFn = async () => {
         catchError(() => of(false))
       )
     );
-    // Provider never became ready (no client id / LD unreachable) → fail closed.
+    // Provider never became ready in time (LD slow / unreachable) → fail open so users
+    // aren't silently redirected away from a route the flag is enabled for in production.
+    // Log so the team can monitor frequency and detect drift if the flag scope ever narrows.
     if (!ready) {
-      return router.parseUrl('/profile');
+      console.warn('myClasEnabledGuard: LD provider not ready after 5 s — failing open');
+      return true;
     }
   }
 


mlehotskylf (dismissed):

apps/lfx-one/src/app/shared/guards/my-clas-enabled.guard.ts
```diff
@@ -36,9 +37,10 @@ export const myClasEnabledGuard: CanMatchFn = async () => {
         catchError(() => of(false))
       )
     );
-    // Provider never became ready (no client id / LD unreachable) → fail closed.
+    // Provider never became ready in time (LD slow / unreachable) → fail open so users
+    // aren't silently redirected away from a route the flag is enabled for in production.
     if (!ready) {
-      return router.parseUrl('/profile');
+      return true;
     }
```
audigregorie, line 46: **Major — Fail-open path has no observability**

When LD times out and the guard grants access without a flag evaluation, nothing is logged or telemetered. For a security-relevant access decision that changes posture (closed → open), the team needs to know how often this fires — both to detect LD-reliability regressions and to catch drift if `my-clas-enabled` is ever gated to a subset of users (at which point fail-open silently grants unauthorized access during outages). The comment says users shouldn't be "silently redirected," but the new behavior is itself silent.

Suggest emitting a warning when the timeout path is taken so it's visible in logs/telemetry:

```suggestion
    if (!ready) {
      // Fail open: my-clas-enabled is on for all prod users, so allowing the route on LD
      // outage is correct — but log it so we can monitor frequency and detect drift if
      // the flag is ever gated.
      console.warn('myClasEnabledGuard: LD provider not ready after 5s — failing open');
      return true;
    }
```

audigregorie (commented):
## Code Review Summary

Sound fix with a clear rationale — failing open for a read-only view whose flag is on for all production users is the correct trade-off, and the JSDoc was updated to match the new semantics. The one substantive gap is **observability on the fail-open path**: when LD times out and access is granted without a flag evaluation, nothing is logged, so the team can't detect frequency or drift if `my-clas-enabled` is ever gated to a subset of users.

### Minor — outside the diff

- **Commit subject missing ticket suffix** — `fix(guards): fail open on ld timeout in my-clas-enabled guard` has no `(LFXV2-3105)` in the subject (ticket is in the body only). Recent LFXV2-prefixed commits on `main` put the ticket in the subject (e.g. `feat(org-roi): add ROI category and projects donuts (LFXV2-2980) (#1384)`). Append it on amend.
- **"Generated with Claude Code" footer** in the commit message — repo policy is human authorship with no AI attribution trailers; drop this line on amend.
- **Branch is 2 commits behind `main`** (based on `ca762f3f`; `origin/main` is at `86904ff2`). No conflict expected (the guard file wasn't touched by the intervening commits), but rebase before merge to keep history linear.
- **No unit test for the guard** — consistent with the rest of `shared/guards/` (none have specs), but the timeout/fail-open branch is non-trivial and would be cheap to lock down with a small spec mocking `FeatureFlagService.providerReady` and the 5 s `timeout`. Worth a follow-up.

### What's done well

- JSDoc updated to accurately describe the new fail-open semantics — the contract changed, not just the code.
- Minimal, surgical diff: only the behavior change and its documentation, no incidental churn.
- Correct divergence from `orgLensEnabledGuard` (which still fails closed, appropriate for a dark-launched feature); the reasoning is captured in the new comment.

**Rating: 4/5** — solid fix; add observability on the fail-open path and tidy the commit message before merge.

apps/lfx-one/src/app/shared/guards/my-clas-enabled.guard.ts
andrest50: Done — added the `console.warn` on the fail-open path. Also cleaned up the commit: added the ticket suffix to the subject line and dropped the AI attribution footer. Rebased on latest `main` and force-pushed.

audigregorie (approved):

PR #1394, opened 2026-08-11
feat(dashboards): add event roster with goal bars to overview

mrautela365:
Adds the per-event roster table with actual-vs-goal progress bars for registrations and sponsorship revenue, plus the at-risk flag.

**Stacked on** `feat/LFXV2-2768-events-summary-tiles` — review that one first.

LFXV2-2768


apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.html
```diff
@@ -0,0 +1,137 @@
...
+          </div>
+
+          <span class="w-28 truncate text-right text-[11px] text-gray-500" role="cell">{{ row.cfpStatus || '—' }}</span>
+        </div>
+      }
+    </div>
+  } @else {
+    <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="event-roster-empty">
+      <p class="text-sm text-gray-500">No events match your search.</p>
+    </div>
```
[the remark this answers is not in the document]
mrautela365: Fixed in `c4e12d6c0` — the empty state now picks its copy from actual state instead of always claiming a search miss:

```ts
protected readonly emptyMessage = computed(() => {
  if (this.searchTerm().trim()) return 'No events match your search.';
  return this.includePast() ? 'No events in the selected period.' : 'No upcoming events.';
});
```

Good catch — the default Upcoming view with no events was the common path that read wrong.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
```diff
@@ -7,6 +7,8 @@
   [selectedPeriod]="selectedPeriod()"
   data-testid="overview-tab-events-summary" />
 
+<lfx-event-roster-section [foundationSlug]="foundationSlug()" data-testid="overview-tab-event-roster" />
+
```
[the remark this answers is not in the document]
mrautela365: Fixed in `fffa08bc4` — the events sections are now wrapped in a `flex flex-col gap-4` container so they no longer sit flush against one another. (Used `gap-4` rather than `gap-6` to match the spacing already used between sibling cards in this module.)

apps/lfx-one/src/server/routes/analytics.route.ts
```diff
@@ -191,6 +191,7 @@ router.get('/board-meeting-participation-summary', (req, res, next) => analytics
 // ED dashboard marketing endpoints — backed by ANALYTICS.PLATINUM_LFX_ONE.* Snowflake views
 router.get('/event-growth', (req, res, next) => analyticsController.getEventGrowth(req, res, next));
 router.get('/events-overview-summary', (req, res, next) => analyticsController.getEventsOverviewSummary(req, res, next));
+router.get('/event-roster', (req, res, next) => analyticsController.getEventRoster(req, res, next));
```
dealako: [blocking] Missing backend authorization on the roster route

Issue: `GET /event-roster` is wired with only the generic `/api` auth gate — no `requireExecutiveDirector` (or equivalent FGA check).

Proof: The diff at this line adds `router.get('/event-roster', ...)` alongside the two sibling ED endpoints (`event-growth`, `events-overview-summary`), none of which import or apply `requireExecutiveDirector`. The frontend hides the Overview tab from non-ED users (`marketing-impact.component.html`), but that's a client-side guard only — any authenticated user can call this route directly with an arbitrary foundation slug.

Why it matters: The response includes per-event sponsorship revenue, registration goals, and CFP status — financial data that should be ED/staff-scoped, not exposed to any authenticated user.

Fix: Add the server-verified `requireExecutiveDirector` middleware to this route (matches Copilot's independent finding on this same line).

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5048,6 +5051,100 @@ export class ProjectService {
     };
   }
 
+  /**
```
dealako, line 5062: [blocking] registrationRevenue.actual is a fabricated zero, not real data

Issue: The SQL selects `NULL AS REGREV_ACTUAL` unconditionally, and the mapper turns that into `0` (`registrationRevenue: { actual: row.REGREV_ACTUAL ?? 0, ... }`).

Proof: Every roster row will report `registrationRevenue.actual = 0` regardless of true event performance, since the source column is a literal `NULL` in the query, not a real aggregate.

Why it matters: The UI will render this as "$0 raised" / 0% progress toward the registration-revenue goal for every event, which reads as real (and alarming) data rather than "not yet available." This is user-visible fabricated data on a metrics dashboard.

Fix: Either remove `registrationRevenue` from the response contract until the actual value is sourced from the warehouse, or explicitly mark it as unavailable (e.g., `actual: null`) and have the UI render "data unavailable" rather than a 0%-progress bar.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  // === Computed Signals ===
+  protected readonly roster: Signal<EventRosterResponse> = this.initRoster();
+  protected readonly searchTerm: Signal<string> = toSignal(this.search.valueChanges.pipe(startWith('')), { initialValue: '' });
+  protected readonly rows: Signal<EventRosterRowView[]> = this.initRows();
+  protected readonly hasRows = computed(() => this.rows().length > 0);
+  /**
+   * The empty state fires whenever there are no rows, which includes an empty roster with no
+   * search active — telling those users their search matched nothing would be wrong. Only claim
+   * a search miss when a term is actually typed.
```
dealako, line 47: [blocking] No test coverage for the new roster logic

Issue: This component ships with zero unit tests (no `.spec.ts`), and the new conditional SQL / score-normalization it depends on in `project.service.ts` also has no new test coverage despite `project.service.spec.ts` already covering the Snowflake-backed query pattern it extends.

Proof: No spec file for `event-roster-section` appears in the PR's changed-files list; `project.service.spec.ts` is not touched by this PR despite the new goal-percent/health-color/at-risk logic.

Why it matters: The health-color thresholds (green/amber/red), the goal-of-zero special case, the at-risk compound condition (`behindGoal && lowPace`), and the search/toggle filters are exactly the kind of boundary logic that regresses silently without tests.

Fix: Add unit tests for health-color boundaries (79/80/81%, 49/50/51%), the goal=0 vs goal=null vs goal>0 cases, the at-risk flag across `hasGoal`/`percent`/`compScore` combinations, and search/toggle filtering (empty string, case sensitivity, partial match).

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5048,6 +5051,100 @@ export class ProjectService {
...
+      return s === 'high' || s === 'medium' || s === 'low' ? s : 'unknown';
+    };
+
+    const events: EventRosterRow[] = result.rows.map((row) => ({
+      eventId: row.EVENT_ID,
+      eventName: row.EVENT_NAME,
+      startDate: row.START_DATE,
+      isPast: row.EVENT_IS_PAST,
+      country: row.EVENT_COUNTRY ?? '',
+      eventUrl: row.EVENT_URL ?? '',
```
dealako: [minor] eventUrl is not normalized before being exposed

Issue: `EVENT_URL` is passed through raw from Snowflake, unlike the same field in `events.service.ts` (`normalizeToUrl` is used there).

Proof: Warehouse URLs can be scheme-less; bound directly via `[href]` in the template, a scheme-less value resolves as a relative LFX One URL instead of failing safely, and invalid values reach the client unfiltered (matches Copilot's independent finding on this line).

Why it matters: Inconsistent with the established pattern elsewhere in the same domain (events.service.ts), and produces broken or misleading links for any row with a malformed URL.

Fix: Normalize with `normalizeToUrl` (from `@lfx-one/shared/utils`) and return `''` when invalid, matching `events.service.ts`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.html
```diff
@@ -0,0 +1,137 @@
...
+          <!-- Registrations bar -->
+          <div class="w-52" role="cell">
+            <div class="flex items-baseline justify-between gap-2">
+              <span class="text-xs font-semibold text-blue-600 tabular-nums">{{ row.registrations.actual }}</span>
+              @if (row.registrations.hasGoal) {
+                <span class="text-[11px] text-gray-400 tabular-nums">{{ row.registrations.goal }}</span>
+              }
+            </div>
+            @if (row.registrations.hasGoal) {
+              <div class="mt-1 h-1.5 overflow-hidden rounded-full bg-gray-100">
```
dealako, line 97: [minor] Progress indicator missing progressbar ARIA semantics

Issue: The registration/sponsorship progress bar has no `role="progressbar"` / `aria-valuenow` / `aria-valuemin` / `aria-valuemax`, so assistive technology cannot announce the completion percentage.

Why it matters: Other progress indicators in this module follow accessible progress-bar patterns; this one is a purely visual div, which is an a11y regression relative to the rest of the dashboard (matches Copilot's independent finding on this line).

Fix: Add `role="progressbar"` with `aria-valuenow`/`aria-valuemin`/`aria-valuemax` (and an `aria-label` describing what the bar measures) to the bar element.

apps/lfx-one/src/app/shared/services/analytics.service.ts
```diff
@@ -1145,6 +1152,31 @@ export class AnalyticsService {
...
+   */
+  public getEventRoster(foundationSlug: string, includePast = false): Observable<EventRosterResponse> {
+    const key = `${foundationSlug}|${includePast}`;
+    if (!this.eventRosterCache.has(key)) {
+      const params: Record<string, string> = { foundationSlug };
+      if (includePast) {
+        params['includePast'] = 'true';
+      }
+      // Evict on error so a transient failure doesn't pin every later subscriber
+      // to the empty fallback for the rest of the session.
```
dealako, line 1173: [minor] All HTTP failures degrade silently to an empty roster

Issue: The roster fetch's error handling converts every failure (network error, 401/403, Snowflake outage) into a successful empty-array result.

Why it matters: A user whose session expired, or a genuine backend outage, will see "No events match your search" — indistinguishable from a foundation that legitimately has no events. This hides real failures from users and from error monitoring (matches Copilot's independent finding on this file).

Fix: Distinguish "empty because no data" from "empty because the request failed" — surface an error state in the UI, or at minimum log/report the failure so it's visible in monitoring.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.ts
```diff
@@ -0,0 +1,130 @@
...
+      return filtered.map((event) => this.toView(event));
+    });
+  }
+
+  // === Private Helpers ===
+  private toView(event: EventRosterRow): EventRosterRowView {
+    const registrations = this.toBar(event.registrations.actual, event.registrations.goal, false);
+    const sponsorshipRevenue = this.toBar(event.sponsorshipRevenue.actual, event.sponsorshipRevenue.goal, true);
+    // At-risk = a real registration goal the event is materially behind on, and a low pace vs last year.
+    const behindGoal = registrations.hasGoal && registrations.percent < 50;
```
dealako: [nit] At-risk threshold (50%) isn't documented

Issue: `const behindGoal = registrations.hasGoal && registrations.percent < 50;` — the 50% cutoff is a business rule with no comment explaining why 50 was chosen.

Why it matters: Future maintainers may need to tune this threshold and won't know the intended rationale without a comment.

Fix: Add a short comment, e.g. `// At-risk = a real registration goal the event is materially behind on (< 50%), plus a low pace vs last year.`

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.ts
```diff
@@ -0,0 +1,130 @@
...
+    let tone: EventRosterBar['tone'] = 'critical';
+    if (percent >= 80) {
+      tone = 'good';
+    } else if (percent >= 50) {
+      tone = 'warn';
+    }
+    return { actual: fmt(actual), goal: fmt(goal), percent, hasGoal: true, tone };
+  }
+
+  private formatDate(iso: string): string {
```
dealako, line 133: [nit] Date parsing doesn't validate month/day ranges

Issue: `formatDate()` splits the ISO string and only checks for falsy `year`/`month`/`day`, not that `month` is 1–12 or `day` is a valid day for that month. We should have a common function/library that does this. Can we leverage it here?

Why it matters: Low risk since Snowflake should guarantee valid dates, but a malformed value (e.g. month=13) would silently roll over into the wrong month/year via `Date.UTC` rather than being caught.

Fix: Add a bounds check (`month < 1 || month > 12 || day < 1 || day > 31`) before constructing the date, falling back to the raw ISO string.

dealako (changes requested):
Hey @mrautela365 — nice feature, and the goal-progress visualization is a genuinely useful addition to the ED dashboard. The at-risk flag and health-color logic are well thought out, and I like that you already iterated once on the empty-state copy and the duplicate-fetch caching. This review covers the diff introduced by this PR (I reviewed the base branch separately).

I ran this through independent security, correctness/perf/test, and style/docs passes, then cross-checked against the existing Cursor Bugbot and Copilot threads already on this PR — several of my findings independently corroborate theirs, which I've called out inline.

### 🔴 Blocking: 3 issues
- **Missing backend authorization** on `GET /event-roster` — only the generic `/api` auth gate protects it; any authenticated user can pull per-event sponsorship revenue and goals for any foundation slug directly, bypassing the client-side ED guard (agrees with Copilot).
- **`registrationRevenue.actual` is a fabricated zero** — the SQL selects `NULL AS REGREV_ACTUAL` unconditionally and the mapper turns that into `0`, so every event shows a real-looking "$0 raised" instead of "data unavailable" (agrees with Copilot).
- **No test coverage** for the new roster logic — health-color thresholds, the at-risk compound condition, the goal-of-zero special case, search/toggle filtering, and the new SQL/score-normalization in `project.service.ts` all ship untested.

### 🟡 Minor: 3 issues
- `eventUrl` isn't normalized with `normalizeToUrl` like the same field in `events.service.ts`, so scheme-less warehouse URLs resolve as relative LFX One links (agrees with Copilot).
- The progress bar has no `role="progressbar"`/`aria-value*` semantics, unlike other progress indicators in this module (agrees with Copilot).
- Every roster HTTP failure (auth expiry, Snowflake outage) degrades silently to "no events," masking real errors from users and monitoring (agrees with Copilot).

### ⚪ Nit: 2 issues
- The 50% at-risk threshold isn't documented inline.
- `formatDate()` doesn't bounds-check month/day before constructing the `Date`.

### ❔ Question / already open
- Cursor Bugbot's still-open "Wrong empty state for all events" thread (`event-roster-section.component.ts#L43-L47`) looks correct to me — `includePast` doesn't correspond to any "period" concept, so that copy is a bit misleading; worth a look alongside this round.
- Cursor's two other threads (empty-state search copy, overview section spacing) are already resolved by `c4e12d6c0`/`fffa08bc4` per your replies — confirmed fixed, no further action.

### Final decision
🔴 **Needs changes before approval** — mainly the authorization gap and the fabricated revenue value, both of which affect real users/data on a production dashboard; the missing tests should land alongside whichever fix you pick for the at-risk/health-color logic so it doesn't regress next iteration.

apps/lfx-one/src/server/routes/analytics.route.ts
```diff
@@ -191,6 +192,10 @@ router.get('/board-meeting-participation-summary', (req, res, next) => analytics
 // ED dashboard marketing endpoints — backed by ANALYTICS.PLATINUM_LFX_ONE.* Snowflake views
 router.get('/event-growth', (req, res, next) => analyticsController.getEventGrowth(req, res, next));
 router.get('/events-overview-summary', (req, res, next) => analyticsController.getEventsOverviewSummary(req, res, next));
+// ED-gated: returns per-event registration, attendance and sponsorship goal figures. The
+// marketing-impact page hides these from non-EDs client-side only, so authorization is
+// enforced here with server-verified persona detection rather than trusting the UI guard.
+router.get('/event-roster', requireExecutiveDirector, (req, res, next) => analyticsController.getEventRoster(req, res, next));
```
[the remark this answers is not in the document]
mrautela365: Good catch — this was a real hole and it's now fixed in 85de38a6c (on #1393, the bottom of the stack, so all three ED routes inherit it).

`requireExecutiveDirector` now checks the requested `foundationSlug` against `personaProjects['executive-director']`, with the `isRootWriter`/`isLFStaff` bypasses retained as you suggested.

Two deliberate choices worth noting:
- Out-of-scope denials reuse the **same** message and `EXECUTIVE_DIRECTOR_REQUIRED` code as a non-ED denial. A distinct 'not your foundation' error would confirm the foundation exists, turning the endpoint into an existence oracle.
- A request with no `foundationSlug` still passes through — there's nothing to scope against, and rejecting a missing required param is the handler's job.

Covered by 8 tests in `require-executive-director.middleware.spec.ts`, including the exact A-reads-B case. I mutation-tested them: reverting the scoping makes 3 of them fail.

apps/lfx-one/src/server/routes/analytics.route.ts
mrautela365: Fixed in 7746ebe26 — `/event-roster` is now behind `requireExecutiveDirector`, which does server-verified persona detection rather than trusting the client-side guard.

Copilot also spotted that the gate itself was foundation-blind (an ED for A could pass B's slug). That's now fixed too in 85de38a6c — see the thread on this same line.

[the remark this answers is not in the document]
mrautela365: Fixed in 7746ebe26 — the route is now behind `requireExecutiveDirector`, and the persona-scoping gap you flagged separately is fixed in 85de38a6c.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Fixed — `registrationRevenue` is gone from the contract entirely rather than being served as a fabricated zero.

The column that would back it (`EVENT_REGISTRATION_REVENUE`) isn't in `MARKETING_EVENT_REGISTRATIONS`, so `NULL AS REGREV_ACTUAL` was the only thing the query could return. Exposing a field that is structurally always 0 is worse than not exposing it — a reader can't tell 'no revenue' from 'not measured'. Removed from the interface, the query, and the mapper.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5048,6 +5051,100 @@ export class ProjectService {
...
+        r.EVENT_NAME,
+        TO_CHAR(r.EVENT_START_DATE, 'YYYY-MM-DD') AS START_DATE,
+        r.EVENT_IS_PAST,
+        r.EVENT_COUNTRY,
+        r.EVENT_URL,
+        r.COUNT_REGISTRATIONS_ALL_TIME AS REG_ACTUAL,
+        r.EVENT_REGISTRATIONS_GOAL AS REG_GOAL,
+        IFNULL(sp.SPON_ACTUAL, 0) AS SPON_ACTUAL,
+        r.EVENT_SPONSORSHIP_REVENUE_GOAL AS SPON_GOAL,
+        NULL AS REGREV_ACTUAL,
```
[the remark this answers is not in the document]
mrautela365: Same as dealako's finding on this line — `registrationRevenue` was removed from the response contract rather than shipped as an always-zero field, since no column backs it in `MARKETING_EVENT_REGISTRATIONS`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.ts
mrautela365: Added in be711edd2 — `event-roster-section.component.spec.ts`, 16 tests covering the logic you named:

- both at-risk conditions independently (behind-goal + low pace, and each one alone), so neither can silently become sufficient
- the exact-50% boundary
- the no-goal case (renders no bar at all — a 0% bar would read as '0 of a goal' rather than 'no goal set')
- percent capping above 100
- the empty-roster vs search-miss states
- date formatting including the malformed-input fallbacks

The score-normalization and conditional SQL on the server side are covered by the `getEventRoster` cases in `project.service.spec.ts`.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5048,6 +5051,100 @@ export class ProjectService {
...
+  /**
+   * Get the Event Roster for a foundation — one row per event with registration and
+   * sponsorship actual-vs-goal, comparison rating, and CFP status.
+   *
+   * Sourced from ANALYTICS.PLATINUM_LFX_ONE.MARKETING_EVENT_REGISTRATIONS (per event) left
+   * joined to sponsorship actuals aggregated from MARKETING_EVENT_SPONSORSHIPS_BY_TIER. A goal
+   * of 0 means "no goal required" for that event (the UI renders no progress bar). Defaults to
+   * upcoming events only; pass includePast to return the full history.
+   */
+  public async getEventRoster(foundationSlug: string, includePast = false): Promise<EventRosterResponse> {
```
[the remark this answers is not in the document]
mrautela365: Server-side coverage added — see the roster cases in `project.service.spec.ts` plus the 16-test component spec in be711edd2.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Fixed in 6fb8bec22 — `eventUrl` now goes through `normalizeToUrl`, matching `events.service.ts`:

```ts
eventUrl: normalizeToUrl(row.EVENT_URL ?? '') ?? '',
```

That prepends `https://` to scheme-less warehouse values and drops unsafe/invalid ones, so a scheme-less URL can no longer resolve as a relative LFX One link.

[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — fixed in 6fb8bec22 by routing `EVENT_URL` through `normalizeToUrl`, matching the treatment in `events.service.ts`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.html
mrautela365: Fixed in 6fb8bec22 — both bars now carry `role="progressbar"`, `aria-valuenow`, `aria-valuemin="0"`, `aria-valuemax="100"`, plus an `aria-label` naming the event and the actual/goal figures so the announcement is meaningful out of context.

There's a test asserting the exposed percentage in the component spec.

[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — progressbar semantics added in 6fb8bec22, with an `aria-label` carrying the event name and actual/goal values.

apps/lfx-one/src/app/shared/services/analytics.service.ts
mrautela365: Fixed in 6fb8bec22 — logs before degrading, and evicts the cache entry so a transient failure doesn't pin later subscribers to the empty fallback:

```ts
console.error('Failed to load event roster', error);
this.eventRosterCache.delete(key);
return of({ projectId: '', events: [] });
```

The roster still degrades to an empty list rather than surfacing an error state — for the drawer on #1395 I did split error from no-data, since 'no detail for this event' and 'we couldn't load it' are genuinely different messages there. Happy to do the same here if you'd prefer it.

apps/lfx-one/src/app/shared/services/analytics.service.ts
```diff
@@ -1145,6 +1146,18 @@ export class AnalyticsService {
...
+  /**
+   * Foundation event roster (upcoming by default; pass includePast for the full history).
+   * Emits an empty roster on error so the table shows its empty state rather than breaking.
+   */
+  public getEventRoster(foundationSlug: string, includePast = false): Observable<EventRosterResponse> {
+    const params: Record<string, string> = { foundationSlug };
+    if (includePast) {
+      params['includePast'] = 'true';
+    }
+    return this.http.get<EventRosterResponse>('/api/analytics/event-roster', { params }).pipe(catchError(() => of({ projectId: '', events: [] })));
```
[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — logging plus cache eviction added in 6fb8bec22.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.ts
mrautela365: Documented and moved out of the component in 8fabefc11 — `BEHIND_GOAL_PERCENT_THRESHOLD` now lives in `@lfx-one/shared/constants` with the rationale:

> Chosen as the midpoint of the goal bar's own tone scale — the bar already turns from `warn` to `critical` at 50% — so the roster's warning icon lights up exactly when the bar it sits next to goes red. Tune both together, or they will disagree visually.

The bar's tone thresholds read the same constants, so the icon and the bar can't drift apart. The predicate itself is now a shared `isEventAtRisk()` used by both this roster and the needs-attention strip on #1396 — they were previously two independent copies of the same rule.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.ts
mrautela365: Fixed in 8fabefc11. I looked for an existing shared helper first, as you suggested — `date-time.utils.ts` has no UTC date-only formatter that fits (its helpers are all timezone/datetime-oriented), so I range-checked in place rather than adding a new shared util for one caller:

```ts
if (month < 1 || month > 12 || day < 1 || day > 31) return iso;
const parsed = new Date(Date.UTC(year, month - 1, day));
if (parsed.getUTCMonth() !== month - 1 || parsed.getUTCDate() !== day) return iso;
```

The second check catches day-overflow within a valid month (2026-02-30). Showing the raw string keeps bad warehouse data visible instead of rendering a confident but wrong date. Two tests cover the rollover cases.

If you'd rather this became a shared `formatIsoDateUtc()` in `@lfx-one/shared/utils`, say the word — the drawer on #1395 now has the same logic, so there are two callers, which is arguably enough to justify extracting it.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.html
```diff
@@ -0,0 +1,147 @@
...
+          Upcoming
+        </button>
+        <button
+          type="button"
+          class="rounded-full px-3 py-1 text-xs font-semibold"
+          [ngClass]="includePast() ? 'bg-blue-500 text-white' : 'text-gray-500'"
+          (click)="toggleIncludePast(true)"
+          data-testid="event-roster-all">
+          All events
+        </button>
```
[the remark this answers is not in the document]
mrautela365: Fixed in a99005068 — both toggles now bind `aria-pressed` to the scope state:

```html
[attr.aria-pressed]="!includePast()"   <!-- Upcoming -->
[attr.aria-pressed]="includePast()"    <!-- All events -->
```

You were right that the `role="group"` only carried the visual blue styling. There's a test asserting both values flip when the scope changes.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5048,6 +5059,221 @@ export class ProjectService {
...
+      WITH slug_resolve AS (
+        SELECT DISTINCT project_id
+        FROM ANALYTICS.PLATINUM.ENGAGEMENT_SCORES_BY_CLASSIFICATION
+        WHERE project_slug = ?
+      ),
+      sponsorship AS (
+        SELECT EVENT_ID, SUM(IFNULL(SPONSORSHIP_REV_ALL_TIME, 0)) AS SPON_ACTUAL
+        FROM ANALYTICS.PLATINUM_LFX_ONE.MARKETING_EVENT_SPONSORSHIPS_BY_TIER
+        GROUP BY EVENT_ID
+      )
```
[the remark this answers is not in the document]
mrautela365: Good catch — fixed in a852f5c83. The CTE now joins `slug_resolve` before the aggregate:

```sql
SELECT t.EVENT_ID, SUM(IFNULL(t.SPONSORSHIP_REV_ALL_TIME, 0)) AS SPON_ACTUAL
FROM ANALYTICS.PLATINUM_LFX_ONE.MARKETING_EVENT_SPONSORSHIPS_BY_TIER t
INNER JOIN slug_resolve sr ON t.PROJECT_ID = sr.project_id
GROUP BY t.EVENT_ID
```

I checked the warehouse before and after rather than assuming:

- **No event id currently spans multiple projects** (`COUNT(DISTINCT PROJECT_ID) > 1` returns 0 rows), so there is no live leak — the old query was correct only by accident of the data.
- **Revenue is unchanged for every roster row** after scoping, so this is pure correctness insurance with no user-visible movement.

Filing it as a fix rather than a no-op because the query shouldn't depend on that coincidence holding.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,140 @@
...
+              role="progressbar"
+              [attr.aria-valuenow]="regProgress()"
+              aria-valuemin="0"
+              aria-valuemax="100"
+              [attr.aria-label]="'Registrations: ' + view()?.registrationsActual + ' of ' + view()?.registrationsGoal + ' goal'"
+              [ngClass]="{
+                'bg-emerald-500': regProgress()! >= 80,
+                'bg-amber-500': regProgress()! >= 50 && regProgress()! < 80,
+                'bg-red-400': regProgress()! < 50,
+              }"
```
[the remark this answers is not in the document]
mrautela365: Fair catch — I extracted those constants for exactly this reason and then left the drawer on literals. Fixed in aa744aa2c.

The tone is now computed in the view model from the shared constants rather than in the template:

```ts
private barTone(percent: number | null): 'good' | 'warn' | 'critical' | null {
  if (percent === null) return null;
  if (percent >= ON_TRACK_PERCENT_THRESHOLD) return 'good';
  if (percent >= BEHIND_GOAL_PERCENT_THRESHOLD) return 'warn';
  return 'critical';
}
```

That keeps the threshold logic out of the template (which was already moved to a computed for the per-CD-pass issue) and means the drawer, the roster's bar, and the at-risk icon all move together when either constant is tuned. Test covers the exact 80% boundary.

dealako (approved):
Hey @mrautela365 — thanks for the thorough turnaround on this. I re-ran the review against everything pushed since my last pass (`c4e12d6c0` → `aa744aa2c`), scoped to the roster/marketing-impact/analytics work (ignoring the unrelated commits that rode in from `main`).

### 👏 Nice work
- The authorization fix is solid and goes further than I asked: `requireExecutiveDirector` now gates all three ED marketing routes, is foundation-scoped (not just "has ED persona anywhere"), denies out-of-scope requests with the same error as non-ED denials to avoid an existence oracle, and ships with its own middleware spec.
- `registrationRevenue` was removed from the contract entirely rather than patched to look less wrong — the right call over shipping a disguised fake zero.
- The at-risk/health-color thresholds moved into `@lfx-one/shared/constants` (`BEHIND_GOAL_PERCENT_THRESHOLD`, `ON_TRACK_PERCENT_THRESHOLD`) with a genuinely useful comment tying the 50% cutoff to the bar's own tone scale — and the new detail drawer was pulled onto the same constants instead of keeping its own hardcoded 80/50.
- Test coverage is real, not superficial: 16-case component spec hitting both at-risk conditions independently, goal-of-zero suppression, boundary values, and search/toggle filtering, plus service-level coverage for URL normalization and foundation scoping.
- Good catch-and-fix cycle with the other bots too — the unscoped sponsorship CTE (cross-foundation revenue leakage risk) and the drawer's hardcoded thresholds were both caught and fixed within this same round.

### Revision tracking
All 9 items from the previous round are ✅ **resolved**, verified against the actual current code (not just commit messages):
- ✅ Backend authorization on `/event-roster` — `requireExecutiveDirector`, foundation-scoped (`analytics.route.ts:201`)
- ✅ `registrationRevenue.actual` fabricated zero — field removed from the contract
- ✅ No test coverage — comprehensive specs added across component, drawer, and service
- ✅ `eventUrl` not normalized — `normalizeToUrl()` now applied in both `getEventRoster` and `getEventDetail`
- ✅ Progress bar missing ARIA semantics — `role="progressbar"` + `aria-value*` on both roster and drawer bars
- ✅ HTTP failures silently degrading to empty state — now logged before falling back, with cache eviction on failure
- ✅ At-risk threshold undocumented — moved to shared constants with rationale
- ✅ `formatDate()` lacked bounds checking — now range-checks month/day and cross-validates against the parsed date
- ✅ Toggle buttons lacked pressed state / drawer date-country whitespace / drawer progress semantics (Cursor Bugbot follow-ons) — all fixed in-branch, confirmed in current HTML

### Structured issue count
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

No new issues surfaced in this round's diff.

### Final decision
✅ **Approved**

PR #1395, opened 2026-08-11
feat(dashboards): add per-event detail drawer to roster

mrautela365:
Adds the per-event detail drawer opened from a roster row, with the b2c/b2b focus split.

**Stacked on** `feat/LFXV2-2768-event-roster-goals`.

LFXV2-2768

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
@@ -1,0 +1,120 @@
+<!-- Copyright The Linux Foundation and each contributor to LFX. -->
+<!-- SPDX-License-Identifier: MIT -->
+
+<p-drawer
+  [(visible)]="visible"
+  position="right"
+  styleClass="xl:!w-[45%] lg:!w-[55%] md:!w-[70%] sm:!w-[90%] !w-full"
+  [showCloseIcon]="true"
+  [dismissible]="true"
+  data-testid="event-detail-drawer">
+  <ng-template #header>
+    <div class="flex flex-col gap-0.5">
+      <h3 class="text-base font-semibold text-gray-900">{{ detail()?.eventName || 'Event detail' }}</h3>
+      @if (detail()) {
+        <p class="text-xs text-gray-500">
+          {{ dateLabel() }}
+          @if (detail()?.country) {
+            · {{ detail()?.country }}
+          }
+        </p>
+      }
+    </div>
+  </ng-template>
+
+  @if (loading()) {
+    <div class="flex flex-col gap-4" data-testid="event-detail-skeleton">
+      <div class="h-20 animate-pulse rounded-lg bg-gray-100"></div>
+      <div class="h-20 animate-pulse rounded-lg bg-gray-100"></div>
+      <div class="h-40 animate-pulse rounded-lg bg-gray-100"></div>
+    </div>
+  } @else if (detail(); as d) {
+    <div class="flex flex-col gap-5">
+      <!-- Registrations vs goal -->
+      <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-registrations">
+        <div class="flex items-baseline justify-between">
+          <span class="text-xs font-medium text-gray-500">Registrations</span>
+          @if (vsLastYearLabel(); as vs) {
+            <span class="text-[11px] font-medium text-gray-400">{{ vs }}</span>
+          }
+        </div>
+        <div class="flex items-baseline gap-2">
+          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ num(d.registrations.actual) }}</span>
+          @if (d.registrations.goal > 0) {
+            <span class="text-sm text-gray-400 tabular-nums">/ {{ num(d.registrations.goal) }} goal</span>
+          }
+        </div>
+        @if (regProgress() !== null) {
+          <div class="mt-1 h-2 overflow-hidden rounded-full bg-gray-100">
+            <div
+              class="h-full rounded-full"
+              [ngClass]="{
+                'bg-emerald-500': regProgress()! >= 80,
+                'bg-amber-500': regProgress()! >= 50 && regProgress()! < 80,
+                'bg-red-400': regProgress()! < 50,
+              }"
+              [style.width.%]="regProgress()"></div>
+          </div>
+        }
+      </div>
+
+      <!-- Sponsorship vs goal -->
+      <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-sponsorship">
+        <span class="text-xs font-medium text-gray-500">Sponsorship revenue</span>
+        <div class="flex items-baseline gap-2">
+          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ money(d.sponsorshipRevenue.actual) }}</span>
+          @if (d.sponsorshipRevenue.goal > 0) {
+            <span class="text-sm text-gray-400 tabular-nums">/ {{ money(d.sponsorshipRevenue.goal) }} goal</span>
+          }
+        </div>
+        @if (sponProgress() !== null) {
+          <div class="mt-1 h-2 overflow-hidden rounded-full bg-gray-100">
+            <div
+              class="h-full rounded-full"
+              [ngClass]="{
+                'bg-emerald-500': sponProgress()! >= 80,
+                'bg-amber-500': sponProgress()! >= 50 && sponProgress()! < 80,
+                'bg-red-400': sponProgress()! < 50,
+              }"
+              [style.width.%]="sponProgress()"></div>
+          </div>
+        }
+      </div>
+
+      <!-- Sponsorship by tier -->
+      @if (d.sponsorshipTiers.length) {
+        <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-tiers">
+          <span class="text-xs font-medium text-gray-500">Sponsorship by tier</span>
+          <div class="flex flex-col gap-2.5">
+            @for (tier of d.sponsorshipTiers; track tier.tier) {
+              <div class="flex items-center justify-between gap-3" [attr.data-testid]="'event-detail-tier-' + tier.tier">
+                <span class="truncate text-xs font-medium text-gray-700">{{ tier.tier || 'Other' }}</span>
+                <div class="flex items-baseline gap-3">
+                  <span class="text-[11px] text-gray-400">{{ tier.sponsorCount }} sponsor{{ tier.sponsorCount === 1 ? '' : 's' }}</span>
+                  <span class="w-20 text-right text-xs font-semibold text-gray-900 tabular-nums">{{ money(tier.revenue) }}</span>
+                </div>
+              </div>
+            }
+          </div>
+        </div>
+      }
+
+      <!-- Pacing lives in PCC (prediction service); deep-link out -->
+      @if (d.eventUrl) {
+        <a
+          [href]="d.eventUrl"
+          target="_blank"
+          rel="noopener noreferrer"
+          class="flex items-center justify-center gap-2 rounded-lg border border-blue-200 bg-blue-50 px-4 py-3 text-xs font-semibold text-blue-700 hover:bg-blue-100"
+          data-testid="event-detail-pcc-link">
+          <i class="fa-light fa-arrow-up-right-from-square" aria-hidden="true"></i>
+          View registration pacing &amp; prediction in PCC
+        </a>
+      }
+    </div>
+  } @else {
+    <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="event-detail-empty">
+      <p class="text-sm text-gray-500">No detail available for this event.</p>
+    </div>
+  }
+</p-drawer>
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
@@ -13,7 +13,10 @@
       <h3 class="text-base font-semibold text-gray-900">{{ detail()?.eventName || 'Event detail' }}</h3>
       @if (detail()) {
         <p class="text-xs text-gray-500">
-          {{ dateLabel() }}
+          <!-- &nbsp; not a plain space: preserveWhitespaces is off, so a whitespace-only text
+               node between the interpolation and the @if block is stripped and the date would
+               run straight into the country separator. -->
+          {{ view()?.dateLabel }}
           @if (detail()?.country) {
             · {{ detail()?.country }}
           }
@@ -24,9 +27,9 @@
 
   @if (loading()) {
     <div class="flex flex-col gap-4" data-testid="event-detail-skeleton">
-      <div class="h-20 animate-pulse rounded-lg bg-gray-100"></div>
-      <div class="h-20 animate-pulse rounded-lg bg-gray-100"></div>
-      <div class="h-40 animate-pulse rounded-lg bg-gray-100"></div>
+      <p-skeleton width="100%" height="5rem" borderRadius="8px" />
+      <p-skeleton width="100%" height="5rem" borderRadius="8px" />
+      <p-skeleton width="100%" height="10rem" borderRadius="8px" />
     </div>
   } @else if (detail(); as d) {
     <div class="flex flex-col gap-5">
@@ -34,14 +37,14 @@
       <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-registrations">
         <div class="flex items-baseline justify-between">
           <span class="text-xs font-medium text-gray-500">Registrations</span>
-          @if (vsLastYearLabel(); as vs) {
+          @if (view()?.vsLastYearLabel; as vs) {
             <span class="text-[11px] font-medium text-gray-400">{{ vs }}</span>
           }
         </div>
         <div class="flex items-baseline gap-2">
-          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ num(d.registrations.actual) }}</span>
+          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ view()?.registrationsActual }}</span>
           @if (d.registrations.goal > 0) {
-            <span class="text-sm text-gray-400 tabular-nums">/ {{ num(d.registrations.goal) }} goal</span>
+            <span class="text-sm text-gray-400 tabular-nums">/ {{ view()?.registrationsGoal }} goal</span>
           }
         </div>
         @if (regProgress() !== null) {
@@ -62,9 +65,9 @@
       <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-sponsorship">
         <span class="text-xs font-medium text-gray-500">Sponsorship revenue</span>
         <div class="flex items-baseline gap-2">
-          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ money(d.sponsorshipRevenue.actual) }}</span>
+          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ view()?.sponsorshipActual }}</span>
           @if (d.sponsorshipRevenue.goal > 0) {
-            <span class="text-sm text-gray-400 tabular-nums">/ {{ money(d.sponsorshipRevenue.goal) }} goal</span>
+            <span class="text-sm text-gray-400 tabular-nums">/ {{ view()?.sponsorshipGoal }} goal</span>
           }
         </div>
         @if (sponProgress() !== null) {
@@ -86,12 +89,12 @@
         <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-tiers">
           <span class="text-xs font-medium text-gray-500">Sponsorship by tier</span>
           <div class="flex flex-col gap-2.5">
-            @for (tier of d.sponsorshipTiers; track tier.tier) {
+            @for (tier of view()?.tiers ?? []; track tier.tier) {
               <div class="flex items-center justify-between gap-3" [attr.data-testid]="'event-detail-tier-' + tier.tier">
-                <span class="truncate text-xs font-medium text-gray-700">{{ tier.tier || 'Other' }}</span>
+                <span class="truncate text-xs font-medium text-gray-700">{{ tier.label }}</span>
                 <div class="flex items-baseline gap-3">
                   <span class="text-[11px] text-gray-400">{{ tier.sponsorCount }} sponsor{{ tier.sponsorCount === 1 ? '' : 's' }}</span>
-                  <span class="w-20 text-right text-xs font-semibold text-gray-900 tabular-nums">{{ money(tier.revenue) }}</span>
+                  <span class="w-20 text-right text-xs font-semibold text-gray-900 tabular-nums">{{ tier.revenue }}</span>
                 </div>
               </div>
             }
@@ -111,6 +114,11 @@
           View registration pacing &amp; prediction in PCC
         </a>
       }
+    </div>
+  } @else if (failed()) {
+    <div class="flex flex-col items-center justify-center gap-1 rounded-lg border border-dashed border-red-200 bg-red-50 p-8" data-testid="event-detail-error">
+      <p class="text-sm font-medium text-red-700">We couldn't load this event's detail.</p>
+      <p class="text-xs text-red-600">Close the panel and try again.</p>
     </div>
   } @else {
     <div class="flex items-center justify-center rounded-lg border border-dashed border-gray-300 bg-gray-50 p-8" data-testid="event-detail-empty">
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
@@ -1,0 +1,93 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { NgClass } from '@angular/common';
+import { ChangeDetectionStrategy, Component, computed, inject, input, model, Signal } from '@angular/core';
+import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { formatCurrency, formatNumber } from '@lfx-one/shared/utils';
+import { AnalyticsService } from '@services/analytics.service';
+import { DrawerModule } from 'primeng/drawer';
+import { finalize, of, skip, switchMap } from 'rxjs';
+
+import type { EventDetailResponse } from '@lfx-one/shared/interfaces';
+
+@Component({
+  selector: 'lfx-event-detail-drawer',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [NgClass, DrawerModule],
+  templateUrl: './event-detail-drawer.component.html',
+})
+export class EventDetailDrawerComponent {
+  private readonly analyticsService = inject(AnalyticsService);
+
+  // === Model Signals (two-way binding) ===
+  public readonly visible = model<boolean>(false);
+
+  // === Inputs ===
+  /** Event id to load when the drawer opens. */
+  public readonly eventId = input<string | null>(null);
+
+  // === Computed Signals ===
+  protected readonly detail: Signal<EventDetailResponse | null> = this.initDetail();
+  protected readonly loading = computed(() => this.visible() && this.detail() === null && this.eventId() !== null);
+
+  // Registration progress (0–100) when a real goal exists.
+  protected readonly regProgress = computed(() => {
+    const d = this.detail();
+    if (!d || d.registrations.goal <= 0) return null;
+    return Math.min(100, Math.round((d.registrations.actual / d.registrations.goal) * 100));
+  });
+  // Sponsorship progress (0–100) when a real goal exists.
+  protected readonly sponProgress = computed(() => {
+    const d = this.detail();
+    if (!d || d.sponsorshipRevenue.goal <= 0) return null;
+    return Math.min(100, Math.round((d.sponsorshipRevenue.actual / d.sponsorshipRevenue.goal) * 100));
+  });
+
+  // === Protected Helpers (template) ===
+  protected num(value: number): string {
+    return formatNumber(value);
+  }
+
+  protected money(value: number): string {
+    return formatCurrency(value);
+  }
+
+  /** Registration pace vs last year as a signed percent string, or null when no baseline. */
+  protected vsLastYearLabel(): string | null {
+    const d = this.detail();
+    if (!d || d.vsLastYear === null) return null;
+    const pct = Math.round((d.vsLastYear - 1) * 100);
+    if (pct > 0) return `+${pct}% vs last year`;
+    if (pct < 0) return `${pct}% vs last year`;
+    return 'On par with last year';
+  }
+
+  protected dateLabel(): string {
+    const iso = this.detail()?.startDate ?? '';
+    const [year, month, day] = iso.split('-').map(Number);
+    if (!year || !month || !day) return iso;
+    return new Date(Date.UTC(year, month - 1, day)).toLocaleDateString('en-US', {
+      month: 'short',
+      day: 'numeric',
+      year: 'numeric',
+      timeZone: 'UTC',
+    });
+  }
+
+  // === Private Initializers ===
+  private initDetail(): Signal<EventDetailResponse | null> {
+    // Lazy-load on open: react to visibility flipping true (skip the initial value).
+    return toSignal(
+      toObservable(this.visible).pipe(
+        skip(1),
+        switchMap((open) => {
+          const id = this.eventId();
+          if (!open || !id) return of(null);
+          return this.analyticsService.getEventDetail(id).pipe(finalize(() => undefined));
+        })
+      ),
+      { initialValue: null }
+    );
+  }
+}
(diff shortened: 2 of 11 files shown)


apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -0,0 +1,93 @@
...
+  // === Model Signals (two-way binding) ===
+  public readonly visible = model<boolean>(false);
+
+  // === Inputs ===
+  /** Event id to load when the drawer opens. */
+  public readonly eventId = input<string | null>(null);
+
+  // === Computed Signals ===
+  protected readonly detail: Signal<EventDetailResponse | null> = this.initDetail();
+  protected readonly loading = computed(() => this.visible() && this.detail() === null && this.eventId() !== null);
```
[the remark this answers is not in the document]
mrautela365: Fixed in `1af46cf6f` — `loading` is now an explicit signal set by the stream rather than derived from `detail() === null`, and cleared via `finalize` so it clears on the error path too:

```ts
protected readonly loading = signal(false);
// …
this.loading.set(true);
return this.analyticsService.getEventDetail(id).pipe(finalize(() => this.loading.set(false)));
```

The service comment explicitly says the request emits `null` on error, so the derived flag did pin the skeleton forever on a failure.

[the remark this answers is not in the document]
mrautela365: Fixed in `1af46cf6f` — same as the thread above. `loading` is now set/cleared explicitly with `finalize`, so a 200-null response or a request failure falls through to the empty state instead of leaving the skeleton up.

apps/lfx-one/src/server/routes/analytics.route.ts
```diff
@@ -192,6 +192,7 @@ router.get('/board-meeting-participation-summary', (req, res, next) => analytics
 router.get('/event-growth', (req, res, next) => analyticsController.getEventGrowth(req, res, next));
 router.get('/events-overview-summary', (req, res, next) => analyticsController.getEventsOverviewSummary(req, res, next));
 router.get('/event-roster', (req, res, next) => analyticsController.getEventRoster(req, res, next));
+router.get('/event-detail', (req, res, next) => analyticsController.getEventDetail(req, res, next));
```
dealako: [blocking] Missing foundation/ED-level authorization on `/api/analytics/event-detail`

Issue: This route has only the generic authenticated `/api` gate. Unlike `getEventRoster`/`getEventsOverviewSummary`, it takes a bare `eventId` with no `foundationSlug` and no ED/admin check, and `ProjectService.getEventDetail` runs no foundation-scoped WHERE clause.
Proof: Copilot flagged the same gap on this line. Any authenticated Contributor/LF Staff user can call `GET /api/analytics/event-detail?eventId=<anything>` and read another foundation's sponsorship revenue, goals, and tier breakdown — data the UI otherwise only renders in the ED branch of the marketing-impact dashboard.
Why it matters: confidential per-foundation financial metrics (sponsorship actuals/goals, tier revenue) leak across foundation boundaries to any signed-in user, not just EDs of that foundation.
Fix: require `foundationSlug` on this route (matching the sibling analytics endpoints), verify the caller is authorized for that foundation, and scope the `ProjectService.getEventDetail` query/lookup to it rather than trusting a bare `eventId`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -0,0 +1,106 @@
...
+      timeZone: 'UTC',
+    });
+  }
+
+  // === Private Initializers ===
+  private initDetail(): Signal<EventDetailResponse | null> {
+    // Lazy-load on open: react to visibility flipping true (skip the initial value).
+    return toSignal(
+      toObservable(this.visible).pipe(
+        skip(1),
```
dealako: [blocking] Stale detail when switching roster rows while the drawer is open

Issue: `initDetail()` only reacts to `toObservable(this.visible)` (skipping the first emission). If the drawer is already open and the user clicks a different roster row, the parent sets `drawerVisible.set(true)` again (no-op, no emission) while `eventId` changes underneath — `switchMap` never re-fires.
Proof: sequence — open Event A (visible false→true, loads A) → click Event B while open (visible true→true, no emission, `eventId` now 'B' but `detail()` still holds A's response). Confirmed independently by two review passes tracing this exact call chain.
Why it matters: the drawer renders the wrong event's registration/sponsorship numbers until the user closes and reopens it — a real data-correctness bug, not just a UX rough edge.
Fix: react to both signals, e.g. `combineLatest([toObservable(this.visible), toObservable(this.eventId)]).pipe(skip(1), switchMap(([open, id]) => { if (!open || !id) { this.loading.set(false); return of(null); } this.loading.set(true); return this.analyticsService.getEventDetail(id).pipe(finalize(() => this.loading.set(false))); }))`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,120 @@
...
+  styleClass="xl:!w-[45%] lg:!w-[55%] md:!w-[70%] sm:!w-[90%] !w-full"
+  [showCloseIcon]="true"
+  [dismissible]="true"
+  data-testid="event-detail-drawer">
+  <ng-template #header>
+    <div class="flex flex-col gap-0.5">
+      <h3 class="text-base font-semibold text-gray-900">{{ detail()?.eventName || 'Event detail' }}</h3>
+      @if (detail()) {
+        <p class="text-xs text-gray-500">
+          {{ dateLabel() }}
```
dealako: [minor] Template calls component methods on every change-detection pass

Issue: `dateLabel()`, and further down `vsLastYearLabel()`, `num()`, `money()`, are all invoked directly from the template rather than read as precomputed signals/pipes.
Proof: confirmed still present at html:16, 37, 42, 44, 65, 67, 94. Copilot already flagged this against the repo's frontend checklist (`docs/reviews/frontend-checklist.md` §4, lines 63–81), which requires templates to read computed values or pipes instead of invoking logic per render.
Why it matters: each of these does parsing/locale formatting (`dateLabel`) or number/currency formatting (`num`/`money`) on every CD pass — wasted work, and inconsistent with the sibling drawer's pattern per the checklist.
Fix: convert these to `computed()` signals (or a view-model computed) evaluated once per underlying data change, and use the existing formatting pipe where one exists instead of calling formatting helpers from the template.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,120 @@
...
+            · {{ detail()?.country }}
+          }
+        </p>
+      }
+    </div>
+  </ng-template>
+
+  @if (loading()) {
+    <div class="flex flex-col gap-4" data-testid="event-detail-skeleton">
+      <div class="h-20 animate-pulse rounded-lg bg-gray-100"></div>
```
dealako: [minor] Raw `animate-pulse` skeleton instead of the standard `p-skeleton`

Issue: the loading state renders three raw `<div class="animate-pulse ...">` placeholders.
Proof: `docs/reviews/frontend-checklist.md:307-318` requires the PrimeNG `p-skeleton` component for loading placeholders, not hand-rolled `animate-pulse` divs. Copilot flagged this same rule violation; still present in the current diff.
Why it matters: inconsistent loading UX/visual language versus the rest of the app, and drifts from the one documented pattern the checklist mandates.
Fix: replace the three placeholder divs with `<p-skeleton>` instances sized to match (height via `height` input rather than Tailwind `h-*` classes).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.html
```diff
@@ -65,17 +65,18 @@ <h3 class="text-base font-semibold text-gray-900">Events</h3>
...
       <!-- Rows -->
       @for (row of rows(); track row.eventId) {
-        <div class="flex items-center gap-4 border-b border-gray-100 py-3" role="row" [attr.data-testid]="'event-roster-row-' + row.eventId">
+        <div
+          class="flex cursor-pointer items-center gap-4 border-b border-gray-100 py-3 transition-colors hover:bg-blue-50/40"
+          role="row"
+          tabindex="0"
+          (click)="openEvent(row.eventId)"
+          (keydown.enter)="openEvent(row.eventId)"
+          (keydown.space)="openEvent(row.eventId)"
```
dealako: [minor] Roster row lacks real activation semantics and Space scrolls the page

Issue: the row is `role="row"` with `tabindex="0"` and click/`keydown.enter`/`keydown.space` handlers that open the drawer, but retains only table-row semantics for assistive tech, and the Space handler never calls `$event.preventDefault()`.
Proof: confirmed current markup still has `role="row"` + the three activation handlers with no `preventDefault`. Both Copilot and Cursor Bugbot flagged this independently on this line.
Why it matters: screen readers don't announce the row as actionable, and pressing Space to "activate" a row also scrolls the page — a real, easily reproduced accessibility bug.
Fix: put the activation handler on a real `<button>` in the name cell (keeping `role="row"`/`role="cell"` on the containers for table semantics), or add `(keydown.space)="$event.preventDefault(); openEvent(row.eventId)"` plus an `aria-label`/appropriate role so AT exposes it as actionable.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,120 @@
...
+      <!-- Pacing lives in PCC (prediction service); deep-link out -->
+      @if (d.eventUrl) {
+        <a
+          [href]="d.eventUrl"
+          target="_blank"
+          rel="noopener noreferrer"
+          class="flex items-center justify-center gap-2 rounded-lg border border-blue-200 bg-blue-50 px-4 py-3 text-xs font-semibold text-blue-700 hover:bg-blue-100"
+          data-testid="event-detail-pcc-link">
+          <i class="fa-light fa-arrow-up-right-from-square" aria-hidden="true"></i>
+          View registration pacing &amp; prediction in PCC
```
dealako, line 114: [minor] PCC CTA is wired to the public event URL, not a PCC link, and isn't validated

Issue: the button text promises "registration pacing & prediction in PCC" but `[href]="d.eventUrl"` binds to the same `EVENT_URL` column that `EventRosterRow.eventUrl` documents as "Public event URL (Cvent etc.)" — not a PCC prediction-service link. The URL is also rendered as an outbound `<a>` with no domain/format validation before display.
Proof: `packages/shared/src/interfaces/dashboard-metric.interface.ts` — `EventRosterRow.eventUrl` doc says public event URL; `EventDetailResponse.eventUrl` is sourced from the identical `row.EVENT_URL` column in `project.service.ts`. Both Copilot and Cursor Bugbot flagged the semantic mismatch; the missing validation is a separate finding — the repo has `server/helpers/url-validation.ts` available and unused here.
Why it matters: users click expecting a PCC pacing/prediction view and land on the public registration page instead; separately, if `EVENT_URL` data is ever bad/malicious, nothing stops an unexpected redirect since the value is rendered unvalidated.
Fix: either link to the actual PCC prediction-service URL for this event (if one exists) and relabel/re-scope this CTA to the public event page, or drop the misleading "pacing & prediction in PCC" copy. Separately, validate `eventUrl` server-side (e.g. via `validateAndSanitizeUrl`) before including it in the response, defaulting to `''` on failure.

apps/lfx-one/src/app/shared/services/analytics.service.ts
```diff
@@ -1177,6 +1178,14 @@ export class AnalyticsService {
     return this.eventRosterCache.get(key)!;
   }
 
+  /**
+   * Per-event detail for the roster drawer (meta, actual-vs-goal, sponsorship-by-tier).
+   * Emits null on error so the drawer shows its empty state.
+   */
+  public getEventDetail(eventId: string): Observable<EventDetailResponse | null> {
+    return this.http.get<EventDetailResponse>('/api/analytics/event-detail', { params: { eventId } }).pipe(catchError(() => of(null)));
```
dealako: [minor] All failure modes collapse into the same `null` as "no event"

Issue: `getEventDetail` pipes every response through `catchError(() => of(null))`, so a 401/403, a 5xx, a network failure, and a genuinely missing event all produce the identical `null` the drawer renders as "No detail available for this event."
Proof: current code at analytics.service.ts:1186 confirms this is unchanged since Copilot flagged it. Reused pattern from `getEventRoster`'s `catchError`, but that one only swallows read failures for a list, not a per-event authorization/outage distinction.
Why it matters: during an actual outage or an authorization failure, users see a misleading empty state ("no detail") instead of any indication something went wrong, which will look like normal a missing-data case in support/triage.
Fix: preserve the HTTP error through the app's established error path (e.g. surface via the shared http-error util) or expose a distinct error signal on the drawer so it can render a different message than "no data" for a caught error vs. a genuine 200-null response.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5145,6 +5146,103 @@ export class ProjectService {
...
 
+  /**
+   * Get per-event detail for the roster drawer — event meta, registration/sponsorship
+   * actual-vs-goal, comparison rating, and a sponsorship-by-tier breakdown.
+   *
+   * Meta comes from MARKETING_EVENT_REGISTRATIONS (one row per event) and the tier breakdown
+   * from MARKETING_EVENT_SPONSORSHIPS_BY_TIER. No daily-pacing time-series exists in these
+   * tables (that is PCC's prediction service); the drawer deep-links to eventUrl for pacing.
+   */
+  public async getEventDetail(eventId: string): Promise<EventDetailResponse | null> {
```
dealako: [minor] No test coverage for the new `getEventDetail` path

Issue: `ProjectService.getEventDetail` (this method), the new `AnalyticsController.getEventDetail` route handler, and `EventDetailDrawerComponent` are all new with zero corresponding spec coverage.
Proof: no `getEventDetail` references in `project.service.spec.ts`; no `analytics.controller.spec.ts` exists; the drawer's component directory has only `.html`/`.ts`, no `.spec.ts`. Copilot flagged the backend gap independently.
Why it matters: this method has non-trivial logic (parallel Snowflake reads, tier aggregation/normalization, null-event handling) that's easy to silently regress without tests, especially given the tier-normalization bug noted separately in this review.
Fix: add cases to `project.service.spec.ts` for event-found/not-found, tier aggregation with multiple tiers, and query failure; add a spec for the new controller route (valid/missing/malformed `eventId`); add a component spec for the drawer covering lazy-load-on-open, loading/empty states, and (once fixed) the eventId-switch-while-open case.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5145,6 +5146,103 @@ export class ProjectService {
...
+      startDate: row.START_DATE,
+      country: row.EVENT_COUNTRY ?? '',
+      eventUrl: row.EVENT_URL ?? '',
+      registrations: { actual: row.REG_ACTUAL ?? 0, goal: row.REG_GOAL ?? 0 },
+      sponsorshipRevenue: { actual: sponsorshipActual, goal: row.SPON_GOAL ?? 0 },
+      vsLastYear: row.VS_LY,
+      compScore: normalizeScore(row.COMP_SCORE),
+      cfpStatus: row.CFP_STATUS ?? '',
+      sponsorshipTiers: tierResult.rows.map((t) => ({
+        tier: t.SPONSORSHIP_TIER ?? '',
```
dealako, line 5268: [minor] Tier mapping can produce duplicate `tier: ''` rows

Issue: the query `GROUP BY SPONSORSHIP_TIER` keeps SQL `NULL` and empty-string `''` as distinct groups, but the response mapper normalizes both to `tier: t.SPONSORSHIP_TIER ?? ''` — so if both a `NULL`-tier and a `''`-tier row exist for the same event, the response contains two entries with the identical `tier: ''`.
Proof: confirmed at project.service.ts:5175/5200/5205/5239 — grouping happens before normalization. Copilot flagged the same root cause. The drawer template tracks tiers by `track tier.tier` (event-detail-drawer.component.html:97), so duplicate empty-string tiers will collide on Angular's track key.
Why it matters: if this dual-null-representation case occurs in the data (plausible given free-text SPONSORSHIP_TIER values), the drawer can silently drop or misrender one of the two "Other" tier rows.
Fix: normalize null/blank tier names in the SQL (e.g. `COALESCE(NULLIF(SPONSORSHIP_TIER, ''), 'Other')`) before the `GROUP BY`, so both representations collapse into one group server-side instead of colliding client-side.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,120 @@
...
+  data-testid="event-detail-drawer">
+  <ng-template #header>
+    <div class="flex flex-col gap-0.5">
+      <h3 class="text-base font-semibold text-gray-900">{{ detail()?.eventName || 'Event detail' }}</h3>
+      @if (detail()) {
+        <p class="text-xs text-gray-500">
+          {{ dateLabel() }}
+          @if (detail()?.country) {
+            · {{ detail()?.country }}
+          }
```
dealako, line 22: [nit] Missing space before the country separator

Issue: `{{ dateLabel() }}` is immediately followed by an `@if` block rendering "· {{ country }}" with only whitespace between them in the source, which Angular's default whitespace handling can collapse, running the date and the separator together.
Fix: add an explicit space, e.g. `{{ dateLabel() }}&nbsp;` or move the leading space inside the `@if` block's interpolated string.

dealako (changes requested):
Hey @mrautela365 — nice drawer, and thanks for already landing `1af46cf6` to fix the loading-skeleton bug Cursor Bugbot and Copilot both flagged (confirmed fixed: `loading` is now an explicit signal cleared in `finalize`, so the error path falls through to the empty state). That one's resolved, no further action needed there.

Scope and intent are solid — lazy-loaded drawer, clean b2c/b2b split, sensible skeleton/empty states, parallel Snowflake reads on the backend. Two things need to land before merge though: the new endpoint has no foundation/ED scoping (any authenticated user can pull another foundation's sponsorship revenue), and the drawer shows stale data when you click a different roster row while it's already open. The rest are real but non-blocking — worth cleaning up in this PR or a fast follow.

**🔴 Blocking: 2**
- Missing foundation/ED-level authorization on `GET /api/analytics/event-detail` — any authenticated user can read another foundation's sponsorship/registration data by eventId.
- Drawer shows stale event data when switching roster rows while open — `switchMap` only reacts to `visible`, not `eventId`.

**🟡 Minor: 6**
- Template calls `dateLabel()`, `vsLastYearLabel()`, `num()`, `money()` directly — violates the repo's no-template-functions rule.
- Raw `animate-pulse` divs instead of the mandated `p-skeleton` component.
- Roster row lacks proper button semantics and doesn't prevent default on `keydown.space`, so Space scrolls the page instead of just activating the row.
- PCC deep-link CTA promises "pacing & prediction in PCC" but binds to `eventUrl`, which — per `EventRosterRow`'s own doc comment — is the public Cvent URL, not a PCC link; also rendered as an outbound link with no URL/domain validation.
- `AnalyticsService.getEventDetail` collapses every failure (auth, 5xx, network) into the same `null` used for "event genuinely not found," so the drawer can't distinguish a real outage from empty data.
- No test coverage for the new `EventDetailDrawerComponent`, `ProjectService.getEventDetail`, or the new controller route; and the tier-name mapper normalizes SQL `NULL` and `''` to the same `tier: ''`, risking duplicate `track tier.tier` keys if both occur for one event.

**⚪ Nit: 1**
- Missing whitespace between `{{ dateLabel() }}` and the country `@if` block in the drawer header.

**❔ Question: 0**

**Decision: 🔴 Needs changes before approval** — the authz gap and the stale-drawer-state bug are both blocking; happy to take another pass once those land.

apps/lfx-one/src/server/controllers/analytics.controller.ts
```diff
@@ -2816,6 +2816,58 @@ export class AnalyticsController {
...
+   * Per-event detail for the roster drawer (meta, actual-vs-goal, sponsorship-by-tier).
+   */
+  public async getEventDetail(req: Request, res: Response, next: NextFunction): Promise<void> {
+    const startTime = logger.startOperation(req, 'get_event_detail');
+
+    try {
+      const foundationSlug = getStringQueryParam(req, 'foundationSlug');
+
+      // Required alongside eventId: a bare event id would let any ED read another foundation's
+      // sponsorship revenue and goals, since the id alone carries no ownership information.
```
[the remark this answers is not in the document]
mrautela365: Fixed in 77f0ede63, which brings the scoped ED gate onto this branch.

You're right that `foundationSlug` is client-controlled — the middleware now validates it against the caller's server-resolved ED project scope before the handler runs. There's a second layer underneath: `getEventDetail`'s Snowflake queries are themselves scoped through a `slug_resolve` CTE, so even a request that somehow reached the handler with a foreign slug returns no rows rather than another foundation's data. I verified that against the warehouse — same eventId under a different foundation's slug returns 0 rows.

apps/lfx-one/src/server/routes/analytics.route.ts
mrautela365: Fixed — two independent layers, both now on this branch:

1. **Route/middleware** (77f0ede63): `/event-detail` is behind `requireExecutiveDirector`, which as of 85de38a6c also binds the requested `foundationSlug` to the caller's own ED foundations.
2. **Query scoping** (575617083): the endpoint now *requires* `foundationSlug`, and both Snowflake reads join through a `slug_resolve` CTE, so an event outside the caller's foundation returns the same `null` as a nonexistent one — no existence oracle.

I validated the scoping against the live warehouse rather than assuming: the tier query returns correct data for `tlf` and **0 rows** when a different foundation asks for the same event id. Also confirmed the direct `PROJECT_ID` join produces identical revenue to the via-registrations path across all 104 shared events.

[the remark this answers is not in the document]
mrautela365: Fixed — `/event-detail` is ED-gated as of 77f0ede63, and the gate is foundation-scoped rather than persona-anywhere. Query-level scoping via `slug_resolve` landed in 575617083.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
mrautela365: Confirmed and fixed in 575617083 — this was a real data-correctness bug.

`initDetail` now reacts to `visible`, `eventId`, **and** `foundationSlug`, so clicking a different roster row while the drawer is open refetches instead of leaving the previous event's numbers under the new event's name.

One correction to the suggested fix: `skip(1)` doesn't work with `combineLatest` here. The parent sets `selectedEventId` and `drawerVisible` in two separate signal writes, so a single open produces two emissions — `skip(1)` would drop the initial pair but still fire twice per open. I used `distinctUntilChanged` on the tuple instead, which collapses the duplicate and keeps lazy-loading intact.

There's a regression test for the exact sequence you described (open A → click B while open). I mutation-tested it: reverting to the visible-only trigger makes it fail.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
mrautela365: Fixed in 575617083 — all four are now a single `view` computed that formats once per detail change, and the template reads its fields:

```ts
protected readonly view = computed(() => {
  const d = this.detail();
  if (!d) return null;
  return { dateLabel: ..., vsLastYearLabel: ..., registrationsActual: ..., tiers: [...] };
});
```

The tier list is precomputed too, so `money()` is no longer called per row per CD pass.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,120 @@
...
+  [dismissible]="true"
+  data-testid="event-detail-drawer">
+  <ng-template #header>
+    <div class="flex flex-col gap-0.5">
+      <h3 class="text-base font-semibold text-gray-900">{{ detail()?.eventName || 'Event detail' }}</h3>
+      @if (detail()) {
+        <p class="text-xs text-gray-500">
+          {{ dateLabel() }}
+          @if (detail()?.country) {
+            · {{ detail()?.country }}
```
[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — the four helpers are now a single `view` computed, per frontend-checklist.md §4. Fixed in 575617083.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
mrautela365: Fixed in 575617083 — the three placeholders are now `<p-skeleton>` with `height`/`borderRadius` inputs rather than Tailwind `h-*` on hand-rolled divs.

Worth noting this one nearly regressed: the branch had `p-skeleton` at one point, but a stale worktree reintroduced `animate-pulse`. I caught it during a cherry-pick and re-applied the fix, which is why it's landing later than the other drawer changes.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,120 @@
...
+        </p>
+      }
+    </div>
+  </ng-template>
+
+  @if (loading()) {
+    <div class="flex flex-col gap-4" data-testid="event-detail-skeleton">
+      <div class="h-20 animate-pulse rounded-lg bg-gray-100"></div>
+      <div class="h-20 animate-pulse rounded-lg bg-gray-100"></div>
+      <div class="h-40 animate-pulse rounded-lg bg-gray-100"></div>
```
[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — now `p-skeleton` per frontend-checklist.md:307-318, fixed in 575617083.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.html
mrautela365: Fixed in 575617083:

```html
(keydown.space)="$event.preventDefault(); openEvent(row.eventId)"
[attr.aria-label]="'View detail for ' + row.eventName"
```

Space no longer scrolls the page, and the row carries an `aria-label` naming the action and the event so AT announces it as actionable. I kept `role="row"` rather than moving activation onto a `<button>` in the name cell — the whole row is the click target, and a button inside a `role="row"`/`role="cell"` structure would either duplicate the row's own handler or shrink the hit area. Happy to switch to the button approach if you'd rather have the stronger native semantics.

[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — `preventDefault()` on Space plus an `aria-label` naming the action, fixed in 575617083.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
mrautela365: Fixed the copy rather than the link, in 575617083.

There is no PCC event deep-link available in this app to point at, so the honest fix was to stop promising one: the button now reads **View event page** and the comment explains that `eventUrl` is the public event page (Cvent / events.linuxfoundation.org), not a PCC prediction view.

On validation: `eventUrl` is now normalized server-side via `normalizeToUrl` (added on #1394 for the roster, same column), which prepends `https://` to scheme-less values and drops unsafe/invalid ones — so the value reaching `[href]` is already sanitized rather than raw.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -0,0 +1,120 @@
...
+      @if (d.eventUrl) {
+        <a
+          [href]="d.eventUrl"
+          target="_blank"
+          rel="noopener noreferrer"
+          class="flex items-center justify-center gap-2 rounded-lg border border-blue-200 bg-blue-50 px-4 py-3 text-xs font-semibold text-blue-700 hover:bg-blue-100"
+          data-testid="event-detail-pcc-link">
+          <i class="fa-light fa-arrow-up-right-from-square" aria-hidden="true"></i>
+          View registration pacing &amp; prediction in PCC
+        </a>
```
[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — the CTA now says 'View event page' and links to the public event page it actually points at, rather than promising a PCC pacing view that doesn't exist in this app. Fixed in 575617083.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5150,6 +5151,127 @@ export class ProjectService {
...
+    };
+
+    const sponsorshipActual = tierResult.rows.reduce((sum, t) => sum + (t.REVENUE ?? 0), 0);
+
+    return {
+      eventId: row.EVENT_ID,
+      eventName: row.EVENT_NAME,
+      startDate: row.START_DATE,
+      country: row.EVENT_COUNTRY ?? '',
+      eventUrl: row.EVENT_URL ?? '',
```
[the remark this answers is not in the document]
mrautela365: Fixed — `EVENT_URL` goes through `normalizeToUrl` before it reaches the response, matching `events.service.ts`. That landed on the roster branch (#1394, 6fb8bec22) for the same column and is now on this branch too.

apps/lfx-one/src/app/shared/services/analytics.service.ts
mrautela365: Fixed in 575617083 — error and no-data are now distinct states.

`getEventDetail` no longer swallows failures in the service; the drawer catches them, logs, and sets a `failed` signal that renders a separate 'We couldn't load this event's detail' block instead of the 'No detail available for this event' empty state. A 401/403/5xx/network failure and a genuinely missing event now look different to the user and in the console.

Two tests cover the split, including one asserting the skeleton clears on the error path.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Added in 575617083 — four `getEventDetail` cases in `project.service.spec.ts`:

- maps the event plus tier breakdown, asserting sponsorship actual is summed from tier rows rather than read off the event row
- **binds the foundation slug ahead of the event id in both reads** — this one asserts the actual bind array `['tlf', 'evt-1']` and that both queries contain `slug_resolve`, so the scoping can't silently regress
- returns null for an event outside the caller's foundation
- propagates Snowflake failures rather than resolving a partial event

Plus 9 component tests for the drawer covering lazy-load-on-open, the eventId-switch-while-open case you flagged, and the loading/empty/error states.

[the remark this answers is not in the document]
mrautela365: Server-side coverage added in 575617083 — four `getEventDetail` cases in `project.service.spec.ts` including a bind-order assertion that pins the foundation scoping.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Good catch — fixed at the source in 575617083 by normalizing before the `GROUP BY` rather than in the mapper:

```sql
COALESCE(NULLIF(TRIM(t.SPONSORSHIP_TIER), ''), 'Other') AS SPONSORSHIP_TIER
...
GROUP BY COALESCE(NULLIF(TRIM(t.SPONSORSHIP_TIER), ''), 'Other')
```

I added `TRIM` on top of your suggestion since the column is free text and a whitespace-only tier would otherwise be a third distinct group. Both NULL and blank now collapse into one `Other` row server-side, so they sum correctly instead of colliding on Angular's track key. There's a test asserting an unnamed tier renders as 'Other' rather than a blank row.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
mrautela365: Fixed — the separator uses `&nbsp;` inside the `@if` block, with a comment explaining why a plain space doesn't survive:

```html
<!-- &nbsp; not a plain space: preserveWhitespaces is off, so a whitespace-only text
     node between the interpolation and the @if block is stripped and the date would
     run straight into the country separator. -->
```

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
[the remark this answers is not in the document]
mrautela365: Correcting myself here: I marked this resolved earlier saying the label was fixed, and it wasn't. The fix was lost when the work moved between branches during a cherry-pick, so the old 'View registration pacing & prediction in PCC' text was still live when I claimed otherwise. Sorry for the noise.

It is genuinely fixed now in a99005068 (on #1394's branch, since #1395 merged into it):

- label → **View event page**
- testid → `event-detail-event-link` (was `event-detail-pcc-link`)
- the comment above it now says `eventUrl` is the public event page (Cvent / events.linuxfoundation.org), not a PCC prediction view

I took the relabel route rather than sourcing a PCC URL because there's no PCC event deep-link available in this app to point at.

PR #1396, opened 2026-08-11
feat(dashboards): add needs-attention strip for at-risk events

mrautela365:
Adds the needs-attention strip surfacing events behind their registration goal, with severity tones and a deep link to act on each item.

**Stacked on** `feat/LFXV2-2768-event-detail-drawer`.

LFXV2-2768


apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,88 @@
...
+            return of({ projectId: '', events: [] });
+          }
+          this.loading.set(true);
+          // Upcoming only — attention is about events we can still influence.
+          return this.analyticsService.getEventRoster(slug, false).pipe(finalize(() => this.loading.set(false)));
+        })
+      ),
+      { initialValue: { projectId: '', events: [] } }
+    );
+  }
```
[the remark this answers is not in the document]
mrautela365: Fixed in `c4e12d6c0` — `getEventRoster` is now cached, so the attention strip and the roster section share one request instead of firing two:

```ts
const key = `${foundationSlug}|${includePast}|${period ?? ''}`;
// … shareReplay(1), with the cache key deleted in catchError
```

Period is part of the key so a period switch can't replay the previous period's roster, and the key is evicted on error so a transient failure doesn't pin later subscribers to the empty fallback.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    return toSignal(
+      slug$.pipe(
+        switchMap((slug) => {
+          if (!slug) {
+            this.loading.set(false);
+            return of({ projectId: '', events: [] });
+          }
+          this.loading.set(true);
+          // Upcoming only — attention is about events we can still influence.
+          return this.analyticsService.getEventRoster(slug, false).pipe(finalize(() => this.loading.set(false)));
```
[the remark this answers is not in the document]
mrautela365: Fixed in `c4e12d6c0` — `AnalyticsService.getEventRoster` now returns a cached `shareReplay(1)` observable keyed on slug + includePast + period, so the two components mounting on the Overview tab share a single `/api/analytics/event-roster` call.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,88 @@
...
+      ),
+      { initialValue: { projectId: '', events: [] } }
+    );
+  }
+
+  private initItems(): Signal<EventAttentionItem[]> {
+    return computed(() => {
+      const atRisk = this.roster()
+        .events.map((event) => ({ event, percent: this.regPercent(event) }))
+        .filter((row) => row.percent !== null && (row.percent as number) < BEHIND_GOAL_THRESHOLD && row.event.compScore === 'low')
```
dealako: [minor] At-risk predicate duplicated from `EventRosterSectionComponent.toView`

Issue: This filter re-implements the same at-risk definition (`percent < 50 && compScore === 'low'`) already computed in `event-roster-section.component.ts` (`behindGoal`/`atRisk`, lines 97-98), with an independent `BEHIND_GOAL_THRESHOLD` constant here.
Why it matters: the two components can silently disagree if one threshold is tuned and the other isn't — the roster's warning icon and this strip's inclusion could diverge for the same event.
Fix: extract a shared `isEventAtRisk(row)` predicate (e.g. into a shared analytics util) and have both components call it.

Agrees with Copilot's inline comment on this PR raising the same duplication — still open.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.html
```diff
@@ -0,0 +1,59 @@
...
+            aria-hidden="true"></span>
+
+          <div class="flex flex-col gap-1">
+            <span
+              class="inline-flex w-fit rounded px-2 py-0.5 text-[10px] font-bold tracking-wide"
+              [ngClass]="{
+                'bg-red-100 text-red-600': item.severity === 'critical',
+                'bg-amber-100 text-amber-600': item.severity === 'warning',
+              }"
+              >{{ item.tag }}</span
```
dealako, line 36: [minor] Severity conveyed by color only

Issue: `critical` and `warning` items both render the literal text "BEHIND GOAL" (line 36); the only difference is the red-vs-amber `ngClass` (line 32-35) and the accent bar color (line 26).
Why it matters: fails WCAG 1.4.1 (use of color) — a user who can't distinguish red/amber (or is using a screen reader) can't tell a critical shortfall from a milder one.
Fix: vary the visible tag text by severity (e.g. `CRITICAL` vs `BEHIND GOAL`), or add an `aria-label`/visually-hidden prefix that states the severity explicitly.

Agrees with Copilot's inline comment on this PR raising the same issue — still open.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
```diff
@@ -1,13 +1,19 @@
...
 
-<lfx-events-summary-section
-  [foundationSlug]="foundationSlug()"
-  [foundationName]="foundationName()"
-  [selectedPeriod]="selectedPeriod()"
-  data-testid="overview-tab-events-summary" />
+<!-- The three events sections are siblings of the KPI block below, which brings its own card
+     spacing. Without a wrapper they render flush against each other. -->
+<div class="flex flex-col gap-4">
+  <lfx-events-attention-section [foundationSlug]="foundationSlug()" data-testid="overview-tab-events-attention" />
```
dealako, line 5: [minor] Empty attention strip still consumes a flex gap

Issue: `lfx-events-attention-section`'s host element is always mounted as a child of this `gap-4` flex column. Its template (`events-attention-section.component.html:4-59`) renders nothing when `hasItems()` is false, but the Angular component host tag itself stays in the flex flow at zero height — so `gap-4` still applies between it and `lfx-events-summary-section`, adding unwanted top space whenever there are no at-risk events.
Why it matters: the just-added spacing fix (this PR's second commit) doesn't cover the no-at-risk-events case, so most foundations without a behind-goal event will see extra blank space above the summary tiles.
Fix: bind `[class.hidden]="!hasItems()"` on the host (via `host: {}` in the component decorator) or wrap the strip's `@if` check at the call site (`@if (attentionHasItems()) { <lfx-events-attention-section ... /> }`) so an empty strip is removed from the flex flow entirely.

Confirmed by Cursor Bugbot's inline comment on this PR against this same commit — still open/unaddressed.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,88 @@
...
+import { formatNumber } from '@lfx-one/shared/utils';
+import { AnalyticsService } from '@services/analytics.service';
+import { finalize, of, switchMap } from 'rxjs';
+
+import type { AttentionSeverity, EventAttentionItem, EventRosterResponse, EventRosterRow } from '@lfx-one/shared/interfaces';
+
+/** How many at-risk events to surface at most. */
+const MAX_ATTENTION_ITEMS = 3;
+/** Below this registration-to-goal percentage an event is considered behind. */
+const BEHIND_GOAL_THRESHOLD = 50;
```
dealako: [minor] Module-level consts inside `apps/lfx-one/`

Issue: `MAX_ATTENTION_ITEMS` and `BEHIND_GOAL_THRESHOLD` are module-level constants inside `apps/lfx-one/`.
Why it matters: repo convention (`CLAUDE.md` "Source hygiene") is that constants live in `@lfx-one/shared`, not as module-level consts under `apps/lfx-one/`.
Fix: move these two constants into `@lfx-one/shared/constants` (or make them `private static readonly` members on the component class if they're intentionally component-scoped and not meant to be shared).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    const vsLy = event.vsLastYear !== null ? ` · pacing ${Math.round((event.vsLastYear - 1) * 100)}% vs last year` : '';
+    return {
+      id: event.eventId,
+      tag: 'BEHIND GOAL',
+      severity,
+      title: `${event.eventName} is ${percent}% to its registration goal`,
+      detail: `${formatNumber(event.registrations.actual)} / ${formatNumber(event.registrations.goal)} registrations${vsLy}. Email and paid are the fastest levers.`,
+      actionUrl: event.eventUrl,
+    };
+  }
```
dealako, line 105: [minor] No unit tests for the new filtering/ranking/severity logic

Issue: There's no `.spec.ts` for this component. `initItems()` combines a three-condition filter, a sort, a slice, and a severity/percent/pace derivation — none of it is covered by a test.
Why it matters: a regression in the 50%-threshold, the `compScore === 'low'` gate, the top-3 ordering, or the 25% critical/warning cutoff would ship silently.
Fix: add `events-attention-section.component.spec.ts` covering: empty roster, all-events-above-goal, >3 qualifying events (verify top-3 by furthest-behind), boundary at exactly 25%/50%, and `vsLastYear === null` (no pace line rendered).

dealako (changes requested):
Hey @mrautela365, thanks for this — it's a clean, focused addition. The needs-attention strip does exactly what the ticket asks: surface up to three at-risk events, derived from the existing roster fetch with no new query, with a clear severity tag, shortfall, pace comparison, and deep-link. The roster caching fix (shareReplay keyed by slug/includePast/period) already resolves the duplicate-fetch concern Cursor and Copilot both raised earlier in the thread — good catch and clean fix there.

**Security & Data Privacy:** ✅ No findings. Auth is enforced upstream, the deep-link uses Angular's automatic sanitization plus `noopener noreferrer`, and nothing PII-bearing is logged or exposed beyond aggregate dashboard metrics.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 5 — duplicated at-risk predicate vs. `EventRosterSectionComponent` (drift risk), severity conveyed by color only (a11y), empty strip still opens a flex gap when there are no at-risk events (Bugbot flagged this against the latest commit — the spacing fix doesn't cover the empty case), module-level consts under `apps/lfx-one/` (repo convention), and no unit tests for the new filter/rank/severity logic
- ⚪ Nit: 0
- ❔ Question: 0

**Reconciliation with existing bot comments:** the duplicate-fetch finding (Cursor + Copilot) is resolved by the caching fix — confirmed in `analytics.service.ts`. Copilot's color-only-severity and duplicated-predicate comments are still open and included above. Cursor Bugbot's "empty strip leaves flex gap" finding is still open against the current head commit — the "space the events sections" fix commit didn't address the zero-items case.

**Final decision: 🔴 Needs changes before approval** — mainly the empty-state gap (visible regression for the common case of no at-risk events) and the color-only severity indicator; the predicate duplication and missing tests are worth cleaning up too but lower urgency.

dealako:
Hey @mrautela365 — thanks for pushing this update, nice progress since the last round 👏

**Nice work:**
- Extracting `isEventAtRisk()` / `eventRegistrationPercent()` into `@lfx-one/shared/utils` (backed by `BEHIND_GOAL_PERCENT_THRESHOLD` / `ON_TRACK_PERCENT_THRESHOLD` in `@lfx-one/shared/constants`) and wiring the roster's tone/at-risk logic through it — exactly the shared-predicate direction from the last round.
- Solid new test coverage: `event-roster-section.component.spec.ts`, `event-detail-drawer.component.spec.ts`, `events-summary-section.component.spec.ts`, `project.service.spec.ts`, plus an e2e spec — the at-risk boundary case (exactly 50%) and the date-rollover fallback are both explicitly covered, which is great.
- The `formatDate` fix range-checks month/day before `Date.UTC` and detects rollover, so bad warehouse data now renders as the raw string instead of a confidently wrong date.
- Accessibility pass on the roster rows (`aria-label`, `role="progressbar"` + `aria-valuenow/min/max`, `keydown.space` preventDefault) — good catch on all three.
- Removing the unused `registrationRevenue` field from `EventRosterRow`.

**Revision tracking on the 5 prior findings:**

1. 🟡 **Duplicated at-risk predicate** ([events-attention-section.component.ts](https://github.com/linuxfoundation/lfx-self-serve/pull/1396/files)) — ⚠️ **Partially addressed.** The shared `isEventAtRisk()`/`eventRegistrationPercent()` utilities now exist and are correctly used by `event-roster-section.component.ts`, but `events-attention-section.component.ts` wasn't touched in these commits — it still has its own local `MAX_ATTENTION_ITEMS`/`BEHIND_GOAL_THRESHOLD` and its own `regPercent()`/filter predicate. Now that a canonical shared implementation exists, this local copy is a clearer drift risk than before (agrees with Cursor's new "Duplicated at-risk event logic" comment on this file, medium severity). Fix: swap `initItems()`/`regPercent()` to call `isEventAtRisk()`/`eventRegistrationPercent()` from `@lfx-one/shared/utils`, and drop the local `BEHIND_GOAL_THRESHOLD` const in favor of `BEHIND_GOAL_PERCENT_THRESHOLD`.
2. 🟡 **Severity conveyed by color only** (events-attention-section.component.html:36) — ❌ **Still open.** File untouched in this round.
3. 🟡 **Empty attention strip still consumes a flex gap** (overview-tab.component.html) — ❌ **Still open.** The file was restructured (all sections now share one `gap-6` flex column instead of two separate wrappers), but the root cause is unchanged: `<lfx-events-attention-section>`'s host element still mounts in the flex flow even when `hasItems()` is false, so the gap still applies above the summary section on the common case (no at-risk events).
4. 🟡 **Module-level consts inside `apps/lfx-one/`** (events-attention-section.component.ts) — ❌ **Still open.** File untouched. Notably, this diff added `BEHIND_GOAL_PERCENT_THRESHOLD` to `@lfx-one/shared/constants` for the same value — the right home now exists, so the local duplicate is even more clearly out of step with convention.
5. 🟡 **No unit tests for `EventsAttentionSectionComponent`** — ❌ **Still open.** Specs were added this round for the roster, drawer, summary, and project service — but not for the attention component.

**AI bot reconciliation:** Cursor's new comment (`events-attention-section.component.ts:74`, "Duplicated at-risk event logic," Medium) restates finding #1 above — agreed, already captured.

**Issue count (all carried over, no new issues from this round's subagent scan across security, correctness/perf/tests, and style/docs dimensions):**
- 🔴 Blocking: 0
- 🟡 Minor: 5 — duplicated predicate (partially addressed), color-only severity, flex-gap on empty strip, module-level consts, missing attention-component tests
- ⚪ Nit: 0

**Final decision:** 🔴 **Needs changes before approval** — none of these are architecturally risky, but items 1–5 all live in `events-attention-section.component.ts`/`.html`, which wasn't touched this round. Worth one more pass to actually consume the new shared utilities, fix the flex-gap/color-only issues, and add the component spec.

dealako (changes requested):
Follow-up review: no new issues found in this round's commits (security/privacy, correctness/performance/tests, and style/docs subagent scans all came back clean on the new diff). However, 4 of the 5 prior findings remain fully unaddressed and 1 is only partially addressed — see the summary comment for the full breakdown. All open items live in `events-attention-section.component.ts`/`.html`, which wasn't touched in these commits. Requesting changes to close those out; happy to re-review quickly once they land.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
mrautela365: Fixed in 529d3bf67 — extracted to `isEventAtRisk()` in `@lfx-one/shared/utils`, called by both this strip and `EventRosterSectionComponent.toView`:

```ts
export function isEventAtRisk(percent: number | null, compScore: string | null | undefined): boolean {
  return percent !== null && percent < BEHIND_GOAL_PERCENT_THRESHOLD && compScore === 'low';
}
```

A null percent encodes 'no goal', so the no-goal case is handled in one place rather than each caller remembering it. There's a companion `eventRegistrationPercent()` both sides use to compute it, so even the percentage derivation is shared — the two views can't disagree now.

Tests cover both conditions independently and the exact-50% boundary on each side.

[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — the predicate is now the shared `isEventAtRisk()` in `@lfx-one/shared/utils`, used by both this component and the roster. Fixed in 529d3bf67.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,88 @@
...
+      return atRisk.map(({ event, percent }) => this.toItem(event, percent as number));
+    });
+  }
+
+  // === Private Helpers ===
+  private regPercent(event: EventRosterRow): number | null {
+    const goal = event.registrations.goal;
+    if (!goal || goal <= 0) return null;
+    return Math.round((event.registrations.actual / goal) * 100);
+  }
```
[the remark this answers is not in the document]
mrautela365: Same finding — deduplicated into `isEventAtRisk()` / `eventRegistrationPercent()` in `@lfx-one/shared/utils`, shared with the roster section. Fixed in 529d3bf67.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.html
mrautela365: Fixed in 529d3bf67 — the tag text now differs by severity rather than relying on the red/amber accent:

```ts
tag: severity === 'critical' ? 'CRITICALLY BEHIND GOAL' : 'BEHIND GOAL',
```

The colour is now a redundant cue rather than the only one, so WCAG 1.4.1 holds for colour-blind users and screen readers alike. Two tests pin it, including the exact-25% boundary where an item is still just a warning.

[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — severity is now carried by the tag text ('CRITICALLY BEHIND GOAL' vs 'BEHIND GOAL'), not colour alone. Fixed in 529d3bf67.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
mrautela365: Fixed in 529d3bf67 via a host binding, so the fix travels with the component rather than depending on every call site remembering to guard it:

```ts
host: { '[style.display]': "hasItems() ? 'contents' : 'none'" },
```

`display: none` removes the host from flex layout entirely when there's nothing to show (no gap slot), and `contents` otherwise keeps the card's own box intact rather than introducing an extra flex level. Two tests assert both states.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
mrautela365: Fixed in 529d3bf67 — both moved to `@lfx-one/shared/constants` per the source-hygiene rule, alongside the threshold constants added for the roster:

- `MAX_ATTENTION_ITEMS`
- `CRITICAL_ATTENTION_PERCENT_THRESHOLD` (renamed from `CRITICAL_THRESHOLD` so it reads unambiguously next to `BEHIND_GOAL_PERCENT_THRESHOLD`)

I documented the relationship between the two thresholds while moving them — 25 is half of the behind-goal 50, i.e. an event far enough behind to need a different response, not just a louder one.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
mrautela365: Added in 529d3bf67 — `events-attention-section.component.spec.ts`, 13 tests covering exactly the cases you listed:

- empty roster, and all-events-above-goal
- each at-risk condition held back independently (behind-goal-but-pacing-well, pacing-low-but-near-goal, no-goal-at-all)
- both boundaries: exactly 50% is not behind goal, exactly 25% is a warning not critical
- more than 3 qualifying events → asserts the top-3 *by furthest behind*, and that the least-behind one is the one dropped
- `vsLastYear === null` renders no pace line
- plus the empty-strip layout behaviour from your sibling comment

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  // === Inputs ===
+  public readonly foundationSlug = input<string | undefined>();
+
+  // === WritableSignals ===
+  protected readonly loading = signal(false);
+
+  // === Computed Signals ===
+  private readonly roster: Signal<EventRosterResponse> = this.initRoster();
+  protected readonly items: Signal<EventAttentionItem[]> = this.initItems();
+  protected readonly hasItems = computed(() => this.items().length > 0);
```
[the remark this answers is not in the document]
mrautela365: Confirmed and fixed in 3f27428a9 — you're right that `loading` was set but never read.

The host binding and the template's `@if` now both gate on `showStrip()`:

```ts
protected readonly showStrip = computed(() => !this.loading() && this.hasItems());
```

So on a foundation switch the strip goes to `display: none` immediately rather than holding the outgoing foundation's at-risk events — which mattered more than usual here because those items are clickable, so a stale one was an actionable link to another foundation's event.

The strip stays skeleton-less by design (it's a supplementary alert, not primary content), so hiding is the right loading state rather than a placeholder. There's a regression test driving a deferred second response; I mutation-tested it by dropping the `!loading()` guard and it fails.

dealako:
@mrautela365 — this round closes it out nicely 🎉

**Nice work:**
- All 5 outstanding items landed in `events-attention-section.component.ts`/`.html`, and each one was fixed at the root rather than patched around: the shared `isEventAtRisk()`/`eventRegistrationPercent()` utilities are now the single source of truth for both this strip and the roster, the severity tag text now varies ("CRITICALLY BEHIND GOAL" vs "BEHIND GOAL") so color is a redundant cue rather than the only one, both thresholds moved to `@lfx-one/shared/constants`, and the empty/loading strip is removed from flex layout via a `[style.display]` host binding (`contents` vs `none`) instead of a call-site `@if` that every consumer would have to remember.
- The new `events-attention-section.component.spec.ts` (13 tests) is thorough — it isolates each at-risk condition independently, pins both the 50%/25% boundaries, verifies top-3-by-furthest-behind ordering (and which item gets dropped), and covers the `vsLastYear === null` case.
- Good catch-and-fix on the stale-foundation-switch bug Cursor flagged mid-round (`showStrip = computed(() => !this.loading() && this.hasItems())`) — you had it fixed before I even got to this pass, with a regression test for the deferred-response race.

**Revision tracking on the 5 prior findings:**
1. ✅ **Duplicated at-risk predicate** — Resolved (529d3bf6). Now consumes the shared `isEventAtRisk()`/`eventRegistrationPercent()`.
2. ✅ **Severity conveyed by color only** — Resolved (529d3bf6). Tag text now varies by severity.
3. ✅ **Empty attention strip consumed a flex gap** — Resolved (529d3bf6, refined 3f27428a). Host `display: contents`/`none` binding removes it from flex flow entirely when empty or loading.
4. ✅ **Module-level consts inside `apps/lfx-one/`** — Resolved (529d3bf6). Both constants moved to `@lfx-one/shared/constants`.
5. ✅ **No unit tests for the filtering/ranking/severity logic** — Resolved (529d3bf6). New spec file covers all the cases called out last round.

**AI bot reconciliation:** Cursor's "Stale attention items on foundation switch" (medium) — you fixed this yourself in 3f27428a before I re-reviewed; confirmed the fix holds (`showStrip()` gates both the host binding and the template).

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

This round's subagent scan (security/privacy, correctness/performance/tests, style/docs) came back clean with no new findings, and the branch is mergeable with no conflicts.

dealako (dismissed):
All 5 prior findings resolved (shared at-risk predicate/threshold consumption, severity conveyed by text not just color, empty/loading strip removed from flex flow via host display binding, constants moved to @lfx-one/shared, and a 13-test spec file added). New Cursor bot finding (stale attention items on foundation switch) was independently caught and fixed by the author. No new issues from this round's security/correctness/style scan. Mergeable, no conflicts. Approving.

dealako:
@mrautela365, please work to resolve the conflicts. Ping me when finished and all review feedback is resolved.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,96 @@
...
+            return of({ projectId: '', events: [] });
+          }
+          this.loading.set(true);
+          // Upcoming only — attention is about events we can still influence.
+          return this.analyticsService.getEventRoster(slug, false).pipe(finalize(() => this.loading.set(false)));
+        })
+      ),
+      { initialValue: { projectId: '', events: [] } }
+    );
+  }
```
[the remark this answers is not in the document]
mrautela365: Confirmed and fixed in b96c357d1 — and the severity is right. This one is mine: I changed `getEventRoster` to rethrow on #1393 (so a roster outage stops reading as "No upcoming events"), updated `EventRosterSectionComponent` for it, and missed this consumer.

Your analysis of the failure mode is exact — a poisoned signal, every later read throwing during change detection, and no recovery on foundation switch because the outer stream is already dead.

```ts
return this.analyticsService.getEventRoster(slug, false).pipe(
  catchError(() => of({ projectId: '', events: [] })),
  finalize(() => this.loading.set(false))
);
```

Caught inside the `switchMap` so the outer stream survives. This strip is supplementary, so a failure omits it rather than taking the Overview tab down — unlike the roster section, which owns its area and surfaces an explicit error state.

Regression test added; mutation-tested by removing the `catchError`, which fails it.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/events-attention-section/events-attention-section.component.ts
```diff
@@ -0,0 +1,96 @@
...
+    return toSignal(
+      slug$.pipe(
+        switchMap((slug) => {
+          if (!slug) {
+            this.loading.set(false);
+            return of({ projectId: '', events: [] });
+          }
+          this.loading.set(true);
+          // Upcoming only — attention is about events we can still influence.
+          return this.analyticsService.getEventRoster(slug, false).pipe(finalize(() => this.loading.set(false)));
```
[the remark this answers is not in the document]
mrautela365: Same finding as Cursor's, fixed in b96c357d1 — and your suggested shape is what I used: catch the inner error and emit an empty `EventRosterResponse` before `finalize`, matching `EventRosterSectionComponent.initRoster`.

The distinction you drew is the one I encoded in the comment: this strip is supplementary, so it fails closed, while the roster section renders an explicit failure state because it owns its area of the page. A test now covers it, and I audited both `getEventRoster` consumers to confirm no third one has the same gap.

MRashad26 (approved):
Reviewed with the code-standards-enforcer checklist (template rules, signal/computed patterns, shared interfaces/constants placement, component field ordering, no effect()/console.log, RxJS error handling, test coverage).

**What this does:** Adds a "Needs attention" strip to the Overview tab surfacing up to 3 upcoming events that are both behind their registration goal (<50%) and pacing low, ranked furthest-behind-first, with a deep link to each event. Stacked on the event-roster/detail-drawer work from #1393.

**Notable strengths:**
- `catchError` placed *inside* the `switchMap` (not wrapping the whole stream) so a roster fetch failure can't permanently poison the `toSignal` — correctly reasoned out in the code comment, and covered by a dedicated test.
- `showStrip = !loading() && hasItems()` correctly prevents a stale foundation's at-risk events from flashing under the new foundation's name during a switch — also has a dedicated regression test ("hides the outgoing foundation while the next roster loads").
- Severity is conveyed by both color *and* text tag ("BEHIND GOAL" / "CRITICALLY BEHIND GOAL"), called out as a WCAG 1.4.1 consideration.
- Threshold boundary tests (exactly 50%, exactly 25%) pin down the intended `<` vs `<=` semantics precisely.
- No function calls in the template — all display strings pre-computed in `items()`/`toItem()`; host display binding is a signal read, consistent with the pattern used elsewhere in this codebase.
- New `AttentionSeverity`/`EventAttentionItem` types and `MAX_ATTENTION_ITEMS`/`CRITICAL_ATTENTION_PERCENT_THRESHOLD` constants correctly placed in `@lfx-one/shared`.
- Component field ordering follows the injections → inputs → writable signals → computed signals → private initializers → private helpers convention.

**One minor nit (non-blocking):** in `events-attention-section.component.ts`, the `detail:` template literal line is ~166 chars, just over the 160-char guideline. Not worth holding up the PR for.

Approving.

PR #1397, opened 2026-08-11
feat(dashboards): expand event drawer into a full deep-dive

mrautela365:
Expands the per-event drawer into a full deep-dive with the registration pacing chart and supporting breakdowns.

**Stacked on** `feat/LFXV2-2768-events-attention`.

LFXV2-2768

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
@@ -4,107 +4,227 @@
 <p-drawer
   [(visible)]="visible"
   position="right"
-  styleClass="xl:!w-[45%] lg:!w-[55%] md:!w-[70%] sm:!w-[90%] !w-full"
-  [showCloseIcon]="true"
-  [dismissible]="true"
+  [modal]="true"
+  [showCloseIcon]="false"
+  styleClass="xl:w-[45%] lg:w-[55%] md:w-[70%] sm:w-[90%] w-full"
   data-testid="event-detail-drawer">
+  <!-- Header -->
   <ng-template #header>
-    <div class="flex flex-col gap-0.5">
-      <h3 class="text-base font-semibold text-gray-900">{{ detail()?.eventName || 'Event detail' }}</h3>
-      @if (detail()) {
-        <p class="text-xs text-gray-500">
-          <!-- &nbsp; not a plain space: preserveWhitespaces is off, so a whitespace-only text
-               node between the interpolation and the @if block is stripped and the date would
-               run straight into the country separator. -->
-          {{ view()?.dateLabel }}
-          @if (detail()?.country) {
-            <span>&nbsp;· {{ detail()?.country }}</span>
-          }
-        </p>
-      }
-    </div>
-  </ng-template>
-
-  @if (loading()) {
-    <div class="flex flex-col gap-4" data-testid="event-detail-skeleton">
-      <p-skeleton width="100%" height="5rem" borderRadius="8px" />
-      <p-skeleton width="100%" height="5rem" borderRadius="8px" />
-      <p-skeleton width="100%" height="10rem" borderRadius="8px" />
-    </div>
-  } @else if (detail(); as d) {
-    <div class="flex flex-col gap-5">
-      <!-- Registrations vs goal -->
-      <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-registrations">
-        <div class="flex items-baseline justify-between">
-          <span class="text-xs font-medium text-gray-500">Registrations</span>
-          @if (view()?.vsLastYearLabel; as vs) {
-            <span class="text-[11px] font-medium text-gray-400">{{ vs }}</span>
-          }
-        </div>
-        <div class="flex items-baseline gap-2">
-          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ view()?.registrationsActual }}</span>
-          @if (d.registrations.goal > 0) {
-            <span class="text-sm text-gray-400 tabular-nums">/ {{ view()?.registrationsGoal }} goal</span>
-          }
-        </div>
-        @if (regProgress() !== null) {
-          <div class="mt-1 h-2 overflow-hidden rounded-full bg-gray-100">
-            <div
-              class="h-full rounded-full"
-              role="progressbar"
-              [attr.aria-valuenow]="regProgress()"
-              aria-valuemin="0"
-              aria-valuemax="100"
-              [attr.aria-label]="'Registrations: ' + view()?.registrationsActual + ' of ' + view()?.registrationsGoal + ' goal'"
-              [ngClass]="{
-                'bg-emerald-500': view()?.registrationsTone === 'good',
-                'bg-amber-500': view()?.registrationsTone === 'warn',
-                'bg-red-400': view()?.registrationsTone === 'critical',
-              }"
-              [style.width.%]="regProgress()"></div>
+    <div class="flex w-full items-start justify-between gap-4" data-testid="event-detail-drawer-header">
+      <div class="flex min-w-0 flex-1 flex-col gap-1">
+        <h2 class="text-lg font-semibold text-gray-900" data-testid="event-detail-drawer-title">{{ detail()?.eventName || 'Event detail' }}</h2>
+        @if (detail(); as d) {
+          <div class="flex flex-wrap items-center gap-x-2 gap-y-0.5 text-sm text-gray-500">
+            <span>{{ dateLabel() }}</span>
+            @if (locationLabel()) {
+              <span aria-hidden="true">·</span>
+              <span><i class="fa-light fa-location-dot mr-1" aria-hidden="true"></i>{{ locationLabel() }}</span>
+            }
+            @if (d.status) {
+              <span aria-hidden="true">·</span>
+              <span>{{ d.status }}</span>
+            }
           </div>
         }
       </div>
+      <lfx-button
+        icon="fa-light fa-xmark"
+        [text]="true"
+        [rounded]="true"
+        severity="secondary"
+        ariaLabel="Close panel"
+        (onClick)="onClose()"
+        data-testid="event-detail-drawer-close" />
+    </div>
+  </ng-template>
 
-      <!-- Sponsorship vs goal -->
-      <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-sponsorship">
-        <span class="text-xs font-medium text-gray-500">Sponsorship revenue</span>
-        <div class="flex items-baseline gap-2">
-          <span class="text-2xl font-semibold text-gray-900 tabular-nums">{{ view()?.sponsorshipActual }}</span>
-          @if (d.sponsorshipRevenue.goal > 0) {
-            <span class="text-sm text-gray-400 tabular-nums">/ {{ view()?.sponsorshipGoal }} goal</span>
-          }
+  <!-- Content -->
+  @if (loading()) {
+    <div class="flex flex-col gap-6 pb-2" data-testid="event-detail-skeleton">
+      <p-skeleton width="100%" height="7rem" borderRadius="8px" />
+      <p-skeleton width="100%" height="6rem" borderRadius="8px" />
+      <p-skeleton width="100%" height="10rem" borderRadius="8px" />
+    </div>
+  } @else if (detail(); as d) {
+    <div class="flex flex-col gap-6 pb-2" data-testid="event-detail-drawer-content">
+      <!-- Headline stats -->
+      <div class="flex flex-col gap-4 sm:flex-row" data-testid="event-detail-stats">
+        <lfx-card styleClass="flex-1">
+          <div class="flex flex-col gap-1">
+            <span class="text-sm text-gray-500">Registrations</span>
+            <span class="text-2xl font-semibold text-gray-900">{{ num(d.registrations.actual) }}</span>
+            @if (d.registrations.goal > 0) {
+              <span class="text-xs text-gray-400">of {{ num(d.registrations.goal) }} goal</span>
+            } @else {
+              <span class="text-xs text-gray-400">no goal set</span>
+            }
+          </div>
+        </lfx-card>
+        <lfx-card styleClass="flex-1">
+          <div class="flex flex-col gap-1">
+            <span class="text-sm text-gray-500">Sponsorship revenue</span>
+            <span class="text-2xl font-semibold text-gray-900">{{ money(d.sponsorshipRevenue.actual) }}</span>
+            @if (d.sponsorshipRevenue.goal > 0) {
+              <span class="text-xs text-gray-400">of {{ money(d.sponsorshipRevenue.goal) }} goal</span>
+            } @else {
+              <span class="text-xs text-gray-400">no goal set</span>
+            }
+          </div>
+        </lfx-card>
+        <lfx-card styleClass="flex-1">
+          <div class="flex flex-col gap-1">
+            <span class="text-sm text-gray-500">vs Last year</span>
+            @if (vsLastYearLabel(); as vs) {
+              <span
+                class="text-2xl font-semibold"
+                [class.text-green-600]="(d.vsLastYear ?? 1) > 1"
+                [class.text-red-600]="(d.vsLastYear ?? 1) < 1"
+                [class.text-gray-900]="(d.vsLastYear ?? 1) === 1"
+                >{{ vs }}</span
+              >
+              <span class="text-xs text-gray-400">registration pace</span>
+            } @else {
+              <span class="text-2xl font-semibold text-gray-400">—</span>
+              <span class="text-xs text-gray-400">no prior year</span>
+            }
+          </div>
+        </lfx-card>
+      </div>
+
+      <!-- Registration pacing -->
+      <div class="flex flex-col gap-4 rounded-lg border border-gray-200 p-4" data-testid="event-detail-pacing">
+        <div class="flex items-center justify-between gap-3">
+          <div class="flex flex-col gap-1">
+            <h3 class="text-sm font-semibold text-gray-900">Registration pacing</h3>
+            @if (behindGoalLabel(); as behind) {
+              <p class="text-sm text-gray-600">{{ behind }}</p>
+            }
+          </div>
+          <lfx-tag [value]="paceRatingLabel()" [severity]="paceSeverity()" [rounded]="true" />
         </div>
-        @if (sponProgress() !== null) {
-          <div class="mt-1 h-2 overflow-hidden rounded-full bg-gray-100">
-            <div
-              class="h-full rounded-full"
-              role="progressbar"
-              [attr.aria-valuenow]="sponProgress()"
-              aria-valuemin="0"
-              aria-valuemax="100"
-              [attr.aria-label]="'Sponsorship revenue: ' + view()?.sponsorshipActual + ' of ' + view()?.sponsorshipGoal + ' goal'"
-              [ngClass]="{
-                'bg-emerald-500': view()?.sponsorshipTone === 'good',
-                'bg-amber-500': view()?.sponsorshipTone === 'warn',
-                'bg-red-400': view()?.sponsorshipTone === 'critical',
-              }"
-              [style.width.%]="sponProgress()"></div>
+
+        @if (d.registrations.goal > 0 && regProgress() !== null) {
+          <div class="flex flex-col gap-1.5">
+            <div class="flex items-baseline justify-between text-sm">
+              <span class="font-semibold text-gray-900">{{ num(d.registrations.actual) }}</span>
+              <span class="text-gray-400">{{ num(d.registrations.goal) }} goal</span>
+            </div>
+            <div class="h-2.5 overflow-hidden rounded-full bg-gray-100">
+              <div
+                class="h-full rounded-full"
+                role="progressbar"
+                [attr.aria-valuenow]="regProgress()"
+                aria-valuemin="0"
+                aria-valuemax="100"
+                [attr.aria-label]="'Registrations: ' + num(d.registrations.actual) + ' of ' + num(d.registrations.goal) + ' goal'"
+                [ngClass]="{
+                  'bg-emerald-500': regProgress()! >= 80,
+                  'bg-amber-500': regProgress()! >= 50 && regProgress()! < 80,
+                  'bg-red-400': regProgress()! < 50,
+                }"
+                [style.width.%]="regProgress()"></div>
+            </div>
           </div>
         }
+
+        @if (d.pacing.available) {
+          <!-- Headline: current / last year / predicted, mirroring PCC -->
+          <div class="flex flex-wrap gap-8">
+            <div class="flex flex-col">
+              <span class="text-xl font-semibold text-blue-600">{{ d.pacing.current !== null ? num(d.pacing.current) : '—' }}</span>
+              <span class="text-xs text-gray-400">Current</span>
+            </div>
+            @if (d.hasPriorYear) {
+              <div class="flex flex-col">
+                <span class="text-xl font-semibold text-gray-500">{{ d.pacing.priorYear !== null ? num(d.pacing.priorYear) : '—' }}</span>
+                <span class="text-xs text-gray-400">Last year</span>
+              </div>
+            }
+            <div class="flex flex-col">
+              <span class="text-xl font-semibold text-violet-600">{{ d.pacing.predictedAvg !== null ? num(d.pacing.predictedAvg) : '—' }}</span>
+              <span class="text-xs text-gray-400">
+                Predicted
+                @if (d.pacing.predictedLow !== null && d.pacing.predictedHigh !== null) {
+                  <span class="text-gray-400">({{ num(d.pacing.predictedLow) }}–{{ num(d.pacing.predictedHigh) }})</span>
+                }
+              </span>
+            </div>
+            <!-- DAYS_LEFT_FROM_YESTERDAY keeps counting down past the event, so a past event
+                 yields a negative. "Days left" is meaningless then — drop the whole block. -->
+            @if (d.pacing.daysLeft !== null && d.pacing.daysLeft > 0) {
+              <div class="ml-auto flex flex-col text-right">
+                <span class="text-xl font-semibold text-gray-900">{{ num(d.pacing.daysLeft) }}</span>
+                <span class="text-xs text-gray-400">Days left</span>
+              </div>
+            }
+          </div>
+
+          <!-- Daily cumulative registration curve -->
+          @if (hasPacingChart()) {
+            <div class="h-[240px]" data-testid="event-detail-pacing-chart">
+              <lfx-chart type="line" [data]="pacingChartData()" [options]="pacingChartOptions" height="100%" />
+            </div>
+          }
+        } @else {
+          <p class="text-sm text-gray-500">Day-by-day pacing prediction is coming to this view.</p>
+        }
+
+        @if (d.eventUrl) {
+          <a
+            [href]="d.eventUrl"
+            target="_blank"
+            rel="noopener noreferrer"
+            class="flex w-fit items-center gap-1.5 text-sm font-semibold text-blue-600 hover:underline"
+            data-testid="event-detail-event-link">
+            <i class="fa-light fa-arrow-up-right-from-square" aria-hidden="true"></i>
+            View event page
+          </a>
+        }
+      </div>
+
+      <!-- Revenue detail -->
+      <div class="flex flex-col gap-4 rounded-lg border border-gray-200 p-4" data-testid="event-detail-revenue">
+        <h3 class="text-sm font-semibold text-gray-900">Revenue</h3>
+        <div class="flex flex-col gap-4 sm:flex-row">
+          <div class="flex flex-1 flex-col gap-1">
+            <span class="text-sm text-gray-500">Registration revenue</span>
+            @if (d.registrationRevenue.actual !== null) {
+              <span class="text-xl font-semibold text-gray-900">{{ money(d.registrationRevenue.actual) }}</span>
+            } @else {
+              <span class="text-xl font-semibold text-gray-300">—</span>
+            }
+            @if (d.registrationRevenue.goal > 0) {
+              <span class="text-xs text-gray-400">Goal {{ money(d.registrationRevenue.goal) }}</span>
+            }
+          </div>
+          <div class="flex flex-1 flex-col gap-1">
+            <span class="text-sm text-gray-500">Sponsorship revenue</span>
+            <span class="text-xl font-semibold text-gray-900">{{ money(d.sponsorshipRevenue.actual) }}</span>
+            <div class="flex items-center gap-2">
+              @if (d.sponsorshipRevenue.goal > 0) {
+                <span class="text-xs text-gray-400">Goal {{ money(d.sponsorshipRevenue.goal) }}</span>
+              }
+              @if (sponProgress() !== null) {
+                <span class="text-xs font-medium text-gray-500">{{ sponProgress() }}%</span>
+              }
+            </div>
+          </div>
+        </div>
       </div>
 
       <!-- Sponsorship by tier -->
       @if (d.sponsorshipTiers.length) {
-        <div class="flex flex-col gap-2 rounded-lg border border-gray-200 p-4" data-testid="event-detail-tiers">
-          <span class="text-xs font-medium text-gray-500">Sponsorship by tier</span>
-          <div class="flex flex-col gap-2.5">
-            @for (tier of view()?.tiers ?? []; track tier.tier) {
-              <div class="flex items-center justify-between gap-3" [attr.data-testid]="'event-detail-tier-' + tier.tier">
-                <span class="truncate text-xs font-medium text-gray-700">{{ tier.label }}</span>
-                <div class="flex items-baseline gap-3">
-                  <span class="text-[11px] text-gray-400">{{ tier.sponsorCount }} sponsor{{ tier.sponsorCount === 1 ? '' : 's' }}</span>
-                  <span class="w-20 text-right text-xs font-semibold text-gray-900 tabular-nums">{{ tier.revenue }}</span>
+        <div class="flex flex-col gap-3 rounded-lg border border-gray-200 p-4" data-testid="event-detail-tiers">
+          <h3 class="text-sm font-semibold text-gray-900">Sponsorship by tier</h3>
+          <div class="flex flex-col gap-0">
+            @for (tier of d.sponsorshipTiers; track tier.tier) {
+              <div
+                class="flex items-center justify-between gap-3 border-b border-gray-100 py-2.5 last:border-b-0"
+                [attr.data-testid]="'event-detail-tier-' + tier.tier">
+                <span class="truncate text-sm font-medium text-gray-700">{{ tier.tier || 'Other' }}</span>
+                <div class="flex items-baseline gap-4">
+                  <span class="text-xs text-gray-400">{{ tier.sponsorCount }} sponsor{{ tier.sponsorCount === 1 ? '' : 's' }}</span>
+                  <span class="w-24 text-right text-sm font-semibold text-gray-900">{{ money(tier.revenue) }}</span>
                 </div>
               </div>
             }
@@ -112,20 +232,32 @@
         </div>
       }
 
-      <!-- eventUrl is the public event page (Cvent / events.linuxfoundation.org), not a PCC
-           prediction view — there is no PCC event deep-link in this app to point at, so the
-           label describes where the link actually goes. -->
-      @if (d.eventUrl) {
-        <a
-          [href]="d.eventUrl"
-          target="_blank"
-          rel="noopener noreferrer"
-          class="flex items-center justify-center gap-2 rounded-lg border border-blue-200 bg-blue-50 px-4 py-3 text-xs font-semibold text-blue-700 hover:bg-blue-100"
-          data-testid="event-detail-event-link">
-          <i class="fa-light fa-arrow-up-right-from-square" aria-hidden="true"></i>
-          View event page
-        </a>
-      }
+      <!-- Where registrations come from -->
+      <div class="flex flex-col gap-3 rounded-lg border border-gray-200 p-4" data-testid="event-detail-channels">
+        <div class="flex flex-col gap-1">
+          <h3 class="text-sm font-semibold text-gray-900">Where registrations come from</h3>
+          <p class="text-sm text-gray-600">Web traffic attributed to each marketing channel.</p>
+        </div>
+        @if (d.channels.length) {
+          <div class="flex flex-col gap-3">
+            @for (channel of d.channels; track channel.channel) {
+              <div class="flex items-center gap-3" [attr.data-testid]="'event-detail-channel-' + channel.channel">
+                <span class="w-36 truncate text-sm font-medium text-gray-700">{{ channel.channel }}</span>
+                <div class="h-2 flex-1 overflow-hidden rounded-full bg-gray-100">
+                  <div class="h-full rounded-full bg-blue-500" [style.width.%]="channel.sharePercent"></div>
+                </div>
+                <span class="w-16 text-right text-sm font-semibold text-gray-900">{{ num(channel.sessions) }}</span>
+                <span class="w-12 text-right text-xs text-gray-400">{{ channel.sharePercent }}%</span>
+              </div>
+            }
+          </div>
+        } @else {
+          <p class="text-sm text-gray-500">
+            No tracked registration traffic yet.
+            <span class="font-medium text-gray-700">Email and paid are the fastest levers to drive early registrations.</span>
+          </p>
+        }
+      </div>
     </div>
   } @else if (failed()) {
     <div class="flex flex-col items-center justify-center gap-1 rounded-lg border border-dashed border-red-200 bg-red-50 p-8" data-testid="event-detail-error">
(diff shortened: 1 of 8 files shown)


apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -44,11 +50,62 @@ export class EventDetailDrawerComponent {
...
+    interaction: { mode: 'index', intersect: false },
+    plugins: {
+      legend: { display: true, position: 'bottom', labels: { boxWidth: 10, boxHeight: 10, color: lfxColors.gray[500], font: { size: 11 } } },
+      tooltip: { enabled: true },
+    },
+    scales: {
+      x: {
+        // Days to event count DOWN to zero; reverse so the event (0) sits on the right.
+        reverse: true,
+        title: { display: true, text: 'Days to event', color: lfxColors.gray[400], font: { size: 10 } },
```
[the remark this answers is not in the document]
mrautela365: Fixed in `b9209cbca` — removed `reverse: true` from the x scale. Confirmed the query orders `DAYS_TO_EVENT DESC`, and this is a category axis so points render in array order; reversing on top of that ran the timeline backwards. Comment added so it doesn't get re-introduced.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -44,11 +50,62 @@ export class EventDetailDrawerComponent {
...
+    maintainAspectRatio: false,
+    interaction: { mode: 'index', intersect: false },
+    plugins: {
+      legend: { display: true, position: 'bottom', labels: { boxWidth: 10, boxHeight: 10, color: lfxColors.gray[500], font: { size: 11 } } },
+      tooltip: { enabled: true },
+    },
+    scales: {
+      x: {
+        // Days to event count DOWN to zero; reverse so the event (0) sits on the right.
+        reverse: true,
```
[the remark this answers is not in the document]
mrautela365: Fixed in `b9209cbca` — same as the axis thread above; `reverse: true` is removed and the ordering now comes solely from the query's `ORDER BY DAYS_TO_EVENT DESC`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -4,113 +4,255 @@
...
+                  <span class="text-gray-400">({{ num(d.pacing.predictedLow) }}–{{ num(d.pacing.predictedHigh) }})</span>
+                }
+              </span>
+            </div>
+            @if (d.pacing.daysLeft !== null) {
+              <div class="ml-auto flex flex-col text-right">
+                <span class="text-xl font-semibold text-gray-900">{{ num(d.pacing.daysLeft) }}</span>
+                <span class="text-xs text-gray-400">Days left</span>
+              </div>
+            }
```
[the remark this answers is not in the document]
mrautela365: Fixed in `b9209cbca` — the block is now guarded on a positive value:

```html
@if (d.pacing.daysLeft !== null && d.pacing.daysLeft > 0) {
```

Traced it to `ANY_VALUE(DAYS_LEFT_FROM_YESTERDAY)`, which keeps counting down past the event and goes negative — so past events were rendering a negative "Days left".

dealako (changes requested):
Hey @mrautela365! This is a nice expansion of the event drawer — the pacing chart, revenue breakdown, and channel attribution give a lot more depth, and the goal-gating/empty-state handling throughout is thoughtful. I can see the two Cursor Bugbot threads on axis-flip and negative days-left are already fixed in b9209cbc, nice catch-and-fix.

I ran a full independent pass (security/privacy, correctness/perf/tests, style/docs) plus reconciled against the existing CodeRabbit/Copilot/Cursor threads, verifying each against the current code rather than taking them at face value — a couple of bot findings turned out to be false positives once traced (channel/tier loops are `GROUP BY`-unique in SQL, so the `@for` track-key concern doesn't hold; `spanGaps: false` on the current-year series is actually correct — it should stop where real data ends, not bridge into unmeasured future days). I also want to flag that Copilot's citation of `docs/reviews/frontend-checklist.md` doesn't correspond to any file that exists in this repo's history — I'm disregarding that specific finding since its cited source is fabricated.

### 🔴 Blocking: 3 issues
- `getEventPacing()`'s catch-all treats every Snowflake failure (timeouts, permission errors, schema regressions) as "tables not materialized" and returns a normal-looking `unavailable` response — this is the exact "zero-fill" failure class `project.service.spec.ts:276-287` already regression-tests against for `getSocialReach()`.
- Null pacing values (`current`/`priorYear`/`predictedAvg`) are coerced to `0` via `?? 0` in the template, rendering an indistinguishable-from-real "Predicted: 0" — same zero-fill bug class as above, just in the template instead of the service.
- No test coverage for the new `getEventDetail()`/`getEventPacing()` service methods or the expanded drawer component (pacing available/unavailable paths, channel/tier aggregation, chart building are all untested).

### 🟡 Minor: 2 issues
- `buildPacingChart()` always adds a "Last year" dataset/legend entry even when `d.hasPriorYear` is false, unlike the headline stats which correctly gate on it (Cursor Bugbot flagged this, still unresolved as of b9209cbc).
- `logger.warning(undefined, ...)` in `getEventPacing()` — `req` is available at the controller but isn't threaded through this call chain, so this warning loses request correlation. Pre-existing pattern for this method, but worth reconsidering.

### ⚪ Nit: 0 issues

### ❔ Question: 2 items
- The Overview tab loses its KPI summary grid, sparkline cards, and attribution section (~200 lines) in this PR. The commit message ("focus overview on events") suggests this is intentional and I confirmed the removed components (`SparklineKpiCardComponent`, `AttributionSectionComponent`) are still used elsewhere, so nothing's orphaned — but the PR description only mentions "expanding the event drawer," so worth confirming this scope explicitly in the description.
- Copilot flagged that `/api/analytics/event-detail` (and the whole `/api/analytics` router) has no server-side ED/role gating — only generic auth — while the ED restriction lives in the Angular route guard. I confirmed the route/controller/middleware are unchanged by this PR's diff, so this is pre-existing, not a regression here. But this PR adds sponsorship revenue, pacing predictions, and channel data to that response, meaning materially more sensitive data now rides on an endpoint any authenticated user can query directly. Worth a follow-up ticket even though it's out of scope for this PR.

The two zero-fill-pattern bugs and the missing test coverage are the blockers — everything else is minor polish or a question for scope/follow-up.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -7461,4 +7514,101 @@ export class ProjectService {
...
+          predictedHigh: p.PRED_HIGH,
+        })),
+      };
+    } catch (error) {
+      // Tables not materialized yet (or transient) — degrade gracefully.
+      logger.warning(undefined, 'get_event_pacing', 'Pacing prediction tables unavailable', {
+        event_id: eventId,
+        error: error instanceof Error ? error.message : 'Unknown error',
+      });
+      return unavailable;
```
dealako, line 7661: [blocking] Catch-all masks real Snowflake failures as "unavailable"

Issue: This `catch` treats every failure from the two Snowflake queries above — timeouts, permission errors, malformed SQL, transient network errors — identically to "the prediction tables aren't materialized yet," and returns a fully-formed `unavailable` response that the UI renders as a normal, expected state ("Day-by-day pacing prediction is coming to this view").
Proof: `project.service.spec.ts:276-287` already establishes and regression-tests the opposite contract for a sibling method (`getSocialReach`) — the test comment explicitly names this failure class the "zero-fill bug": "this method used to swallow Snowflake failures and resolve a defaults object, which reached the dashboard as a 200 and rendered ... indistinguishable from a genuine measurement." `getEventPacing()` reintroduces exactly that pattern for a different query.
Why it matters: A real outage (bad credentials, a broken column rename, a Snowflake timeout) will silently present as "no pacing data yet" instead of surfacing as an error — the same masking bug the codebase already paid down elsewhere.
Fix: Narrow the catch to the confirmed missing-table/no-row case (e.g. rethrow unless the error matches a known "table not found" signature, mirroring whatever mechanism `getSocialReach`/its callers use to distinguish the two), and add a spec analogous to the `getSocialReach` regression test.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -4,113 +4,257 @@
...
+                [style.width.%]="regProgress()"></div>
+            </div>
           </div>
         }
+
+        @if (d.pacing.available) {
+          <!-- Headline: current / last year / predicted, mirroring PCC -->
+          <div class="flex flex-wrap gap-8">
+            <div class="flex flex-col">
+              <span class="text-xl font-semibold text-blue-600">{{ num(d.pacing.current ?? 0) }}</span>
```
dealako: [blocking] Null pacing values render as a real "0"

Issue: `num(d.pacing.current ?? 0)`, `num(d.pacing.priorYear ?? 0)`, and `num(d.pacing.predictedAvg ?? 0)` (lines 129, 134, 139) coerce `null` to `0` before formatting, so a missing prediction or missing cumulative value renders as "0" — indistinguishable from a genuine zero.
Proof: The same block already null-checks `predictedLow`/`predictedHigh` correctly (line 142: `d.pacing.predictedLow !== null && ...`), and the Revenue section a few lines down (188-192) uses an em dash for a null `registrationRevenue.actual` instead of coercing. This is the identical "zero-fill" failure class the codebase already regression-tests against in `project.service.spec.ts:276-287` (there for a Snowflake fallback, here for template rendering) — Cursor Bugbot flagged this on the current commit and it's still unresolved.
Why it matters: A user sees "Predicted: 0" for an event with no prediction yet, which reads as "we predict zero registrations" rather than "no prediction available."
Fix: Use the same em-dash/null-guard pattern as `registrationRevenue` above — e.g. `@if (d.pacing.current !== null) { {{ num(d.pacing.current) }} } @else { — }` for each of the three fields.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -82,6 +140,68 @@ export class EventDetailDrawerComponent {
...
+      labels,
+      datasets: [
+        {
+          label: 'Current year',
+          data: points.map((point) => point.current),
+          borderColor: lfxColors.blue[500],
+          backgroundColor: 'transparent',
+          spanGaps: false,
+        },
+        {
```
dealako: [minor] Chart shows a "Last year" series even when there's no prior year

Issue: `buildPacingChart()` unconditionally adds the "Last year" dataset (lines 187-193), but the headline stats a few lines up in the template correctly gate the same data behind `@if (d.hasPriorYear)` (event-detail-drawer.component.html:132-137).
Proof: Cursor Bugbot flagged this on this commit ("Chart ignores prior-year gate") and it's still present — for an event with no prior edition, the chart legend gets an empty/null "Last year" entry inconsistent with the headline above it.
Why it matters: Minor visual inconsistency — a legend entry with no data reads as a data gap rather than "not applicable."
Fix: Conditionally omit the dataset when `!this.detail()?.hasPriorYear`, matching the headline's gating.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -7461,4 +7514,101 @@ export class ProjectService {
...
+          current: p.CURRENT,
+          priorYear: p.PRIOR,
+          predictedAvg: p.PRED_AVG,
+          predictedLow: p.PRED_LOW,
+          predictedHigh: p.PRED_HIGH,
+        })),
+      };
+    } catch (error) {
+      // Tables not materialized yet (or transient) — degrade gracefully.
+      logger.warning(undefined, 'get_event_pacing', 'Pacing prediction tables unavailable', {
```
dealako: [minor] `req` not threaded through for correlation

Issue: `logger.warning(undefined, 'get_event_pacing', ...)` passes `undefined` where a request context would normally go.
Proof: Per this repo's logging conventions, service methods called from a route handler should pass `req` through for correlation; here neither `getEventDetail(eventId)` nor `getEventPacing(eventId)` accept `req` at all, so this is consistent with the rest of the new method chain rather than a one-off — but it does mean this warning (and any future error logs in this chain) won't carry a `request_id`.
Why it matters: Harder to correlate this warning with the originating request in logs/traces if this path gets noisy in production.
Fix: Consider threading `req` from the controller through `getEventDetail`/`getEventPacing` if request correlation on these logs turns out to matter; not blocking as-is.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5161,12 +5163,18 @@ export class ProjectService {
       EVENT_ID: string;
```
dealako, line 5177: [blocking] No test coverage for the new event-detail/pacing methods

Issue: `getEventDetail()` and `getEventPacing()` are new, non-trivial service methods (event/tier/channel aggregation, share-percent calculation, and a two-query pacing fallback with its own failure-handling contract) with no corresponding spec.
Proof: `project.service.spec.ts` has a dedicated `describe('ProjectService — Snowflake-backed marketing reads')` block covering sibling methods like `getSocialReach`, including a regression test for exactly the failure-masking bug present in `getEventPacing()` — but neither new method appears anywhere in the spec file. The drawer component has no `.spec.ts` either.
Why it matters: The pacing available/unavailable branches, the null-vs-zero data mapping, and the channel/tier SQL-to-DTO mapping are all currently unverified — a later refactor (or the fix for the catch-all above) has no test to prove it didn't regress.
Fix: Add service specs mirroring the existing `getSocialReach` pattern — success mapping, no-row/no-event case, and a Snowflake-failure case asserting the corrected propagate-vs-degrade behavior. Add a component spec for `buildPacingChart()`, `hasPacingChart()`, and the label helpers with null/non-null inputs.

dealako:
@mrautela365, please work to resolve the conflicts. Ping me when finished and all review feedback is resolved.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Fixed — the catch is narrowed to the missing-table case only:

```ts
const isMissingTable = /does not exist or not authorized|Object .* does not exist/i.test(message);
if (!isMissingTable) {
  logger.warning(undefined, 'get_event_pacing', 'Pacing query failed, propagating', { event_id: eventId, error: message });
  throw error;
}
```

You named the precedent exactly right — `project.service.spec.ts:276-287` documents this as the "zero-fill bug" for `getSocialReach`, and this reintroduced it for a different query. Two tests now cover both sides: a compilation error propagates, an unmaterialized table still reports unavailable.

Worth flagging what narrowing this surfaced: the daily query referenced `CURRENT_EVENT_CUMULATIVE_REGISTRATIONS`, **a column that does not exist**. It failed at Snowflake compile time on every call, so this catch-all was swallowing it and the chart had never once rendered. That's fixed separately on #1401.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
mrautela365: Fixed — all three now use the em-dash guard rather than `?? 0`:

```html
{{ d.pacing.current !== null ? num(d.pacing.current) : '—' }}
```

Your reasoning for why this matters is the part I'd underline: "Predicted: 0" reads as a prediction of zero registrations, not as an absent prediction. Same failure class as the Snowflake zero-fill on the sibling thread, just expressed in the template.

I followed the `registrationRevenue` precedent you pointed at rather than inventing a new pattern.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -4,113 +4,257 @@
...
+              <span class="text-xs text-gray-400">Current</span>
+            </div>
+            @if (d.hasPriorYear) {
+              <div class="flex flex-col">
+                <span class="text-xl font-semibold text-gray-500">{{ num(d.pacing.priorYear ?? 0) }}</span>
+                <span class="text-xs text-gray-400">Last year</span>
+              </div>
+            }
+            <div class="flex flex-col">
+              <span class="text-xl font-semibold text-violet-600">{{ num(d.pacing.predictedAvg ?? 0) }}</span>
```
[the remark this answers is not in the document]
mrautela365: Same finding as dealako's on this line — all three pacing values now render an em dash when null instead of coercing to 0.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
mrautela365: Fixed — the chart's `Last year` dataset is now conditional on `hasPriorYear`, matching the headline block that already gated on it:

```ts
...(this.detail()?.hasPriorYear ? [{ label: 'Last year', ... }] : []),
```

Without it the legend showed a series with no line behind it, which reads as "last year was zero" rather than "there was no last year".

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -82,6 +140,68 @@ export class EventDetailDrawerComponent {
...
+          label: 'Predicted',
+          data: points.map((point) => point.predictedAvg),
+          borderColor: lfxColors.violet[500],
+          backgroundColor: 'transparent',
+          borderDash: [6, 4],
+        },
+      ],
+    };
+  }
```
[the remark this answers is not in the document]
mrautela365: Same as dealako's finding — the series is now gated on `hasPriorYear` so the chart and the headline agree about whether a prior year exists.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Added. `getEventDetail` and the pacing path now have coverage in `project.service.spec.ts`:

- maps the event plus tier breakdown, asserting sponsorship actual is summed from tier rows
- **binds the foundation slug ahead of the event id in the scoped reads** — asserted by matching the queries on content rather than call position, so a reordering can't make it pass vacuously
- returns null for an event outside the caller's foundation
- propagates a pacing failure rather than reporting pacing unavailable
- reports unavailable only when the prediction table is unmaterialized
- propagates Snowflake failures rather than resolving a partial event

Plus the drawer component spec — 13 tests covering lazy-load-on-open, the reload-on-event-switch case, loading/empty/error states, and the goal bar's a11y and threshold colours.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -4,113 +4,255 @@
...
-      }
+    <div class="flex w-full items-start justify-between gap-4" data-testid="event-detail-drawer-header">
+      <div class="flex min-w-0 flex-1 flex-col gap-1">
+        <h2 class="text-lg font-semibold text-gray-900" data-testid="event-detail-drawer-title">{{ detail()?.eventName || 'Event detail' }}</h2>
+        @if (detail(); as d) {
+          <div class="flex flex-wrap items-center gap-x-2 gap-y-0.5 text-sm text-gray-500">
+            <span>{{ dateLabel() }}</span>
+            @if (locationLabel()) {
+              <span aria-hidden="true">·</span>
+              <span><i class="fa-light fa-location-dot mr-1" aria-hidden="true"></i>{{ locationLabel() }}</span>
```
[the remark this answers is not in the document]
mrautela365: Fixed — every derived label is now a computed signal rather than a template method call: `dateLabel`, `vsLastYearLabel`, `locationLabel`, `paceRatingLabel` and `behindGoalLabel`.

They do locale formatting, array joins and number formatting, so as template invocations they re-ran on every change-detection pass. `num()`/`money()` stay as methods since they're called per-row with a different argument each time, which a signal can't express — the frontend checklist's concern is recomputation of a fixed value, and those aren't that.

The member-ordering lint rule actually caught the last three for me while I was moving the first two.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Fair, and not fixed — flagging rather than quietly closing.

`getEventPacing` is a private helper reached from `getEventDetail`, which doesn't thread `req` either, so correlating this one log line would mean changing the signature of both plus every caller. That's a wider refactor than this PR should carry, and it'd touch methods other PRs in this stack are editing right now.

The log does carry `event_id`, which is the field you'd filter on for this failure. @dealako if you'd rather have `req` threaded through properly I'll do it as a follow-up across the service rather than piecemeal here — it's the kind of thing worth doing once, consistently.

apps/lfx-one/src/server/services/project.service.ts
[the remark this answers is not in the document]
mrautela365: Same as dealako's blocking finding on this line — the catch is narrowed to the missing-table signature, everything else propagates, with tests on both branches.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5223,23 +5258,41 @@ export class ProjectService {
...
 
     const sponsorshipActual = tierResult.rows.reduce((sum, t) => sum + (t.REVENUE ?? 0), 0);
+    const totalSessions = channelResult.rows.reduce((sum, ch) => sum + (ch.SESSIONS ?? 0), 0);
+
+    const channels: EventChannelAttribution[] = channelResult.rows.map((ch) => ({
+      channel: ch.CHANNEL ?? 'Unknown',
+      sessions: ch.SESSIONS ?? 0,
+      uniqueVisitors: ch.UNIQUE_VISITORS ?? 0,
+      revenue: ch.REVENUE ?? 0,
+      sharePercent: totalSessions > 0 ? Math.round(((ch.SESSIONS ?? 0) / totalSessions) * 1000) / 10 : 0,
```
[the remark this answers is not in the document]
mrautela365: Coverage added — see the `getEventDetail` block in `project.service.spec.ts` (six cases, including the foundation-scoping bind assertion) plus the 13-test drawer component spec.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5223,23 +5258,41 @@ export class ProjectService {
     };
 
     const sponsorshipActual = tierResult.rows.reduce((sum, t) => sum + (t.REVENUE ?? 0), 0);
+    const totalSessions = channelResult.rows.reduce((sum, ch) => sum + (ch.SESSIONS ?? 0), 0);
+
+    const channels: EventChannelAttribution[] = channelResult.rows.map((ch) => ({
+      channel: ch.CHANNEL ?? 'Unknown',
+      sessions: ch.SESSIONS ?? 0,
+      uniqueVisitors: ch.UNIQUE_VISITORS ?? 0,
+      revenue: ch.REVENUE ?? 0,
```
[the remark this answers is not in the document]
mrautela365: The route is ED-gated, and as of the middleware change on #1393 that gate also binds the client-supplied `foundationSlug` to the caller's own ED foundations rather than accepting any ED detection anywhere.

Underneath that, `getEventDetail`'s event and tier queries are scoped through a `slug_resolve` CTE, so a foreign `eventId` returns null before any of these fields are read — verified against the warehouse: the same event id under a different foundation's slug returns 0 rows.

The campaign-enrichment fields are a separate question I've flagged to @dealako on #1401: scoping paid campaigns by project would currently *blank* the breakdown, because that table's `PROJECT_NAME` is a different taxonomy (KubeCon campaigns file under CNCF, not the event's owning foundation). Didn't want to guess at that one.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.ts
```diff
@@ -1,240 +1,29 @@
...
 @Component({
   selector: 'lfx-overview-tab',
-  imports: [
-    SparklineKpiCardComponent,
-    AttributionSectionComponent,
-    EventsSummarySectionComponent,
-    EventRosterSectionComponent,
-    EventsAttentionSectionComponent,
-  ],
+  imports: [EventsSummarySectionComponent, EventRosterSectionComponent, EventsAttentionSectionComponent],
```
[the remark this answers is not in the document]
mrautela365: You're right that the Overview KPI grid and attribution section are removed here, and that's a real behaviour change rather than a refactor.

It's also this PR's stated purpose — "focus overview on events" — so I've kept the removal rather than reverting a deliberate restructure on review feedback. What's wrong is the description, not the diff.

@dealako flagging the same call I raised on #1398: either the PR description should state the removals plainly, or they should move to a separate PR. I'd rather you decided which, since both PRs make the same trade and they should be consistent.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -4,128 +4,260 @@
...
+                role="progressbar"
+                [attr.aria-valuenow]="regProgress()"
+                aria-valuemin="0"
+                aria-valuemax="100"
+                [attr.aria-label]="'Registrations: ' + num(d.registrations.actual) + ' of ' + num(d.registrations.goal) + ' goal'"
+                [ngClass]="{
+                  'bg-emerald-500': regProgress()! >= 80,
+                  'bg-amber-500': regProgress()! >= 50 && regProgress()! < 80,
+                  'bg-red-400': regProgress()! < 50,
+                }"
```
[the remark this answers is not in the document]
mrautela365: You're right, and this is a regression I reintroduced — fixed in 130556b17.

I wired this bar to the shared constants on #1394 precisely so it couldn't drift from the roster. Resolving this branch's conflict I took the drawer rewrite wholesale for the template, which reverted it to literals. Exactly the desync the constants exist to prevent.

Now a computed reading the shared thresholds:

```ts
protected readonly registrationsTone = computed(() => {
  const percent = this.regProgress();
  if (percent === null) return null;
  if (percent >= ON_TRACK_PERCENT_THRESHOLD) return 'good';
  if (percent >= BEHIND_GOAL_PERCENT_THRESHOLD) return 'warn';
  return 'critical';
});
```

Computed rather than inline so the threshold logic stays out of the template, consistent with the other derived labels here.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.html
```diff
@@ -4,128 +4,260 @@
...
+                [class.text-green-600]="(d.vsLastYear ?? 1) > 1"
+                [class.text-red-600]="(d.vsLastYear ?? 1) < 1"
+                [class.text-gray-900]="(d.vsLastYear ?? 1) === 1"
+                >{{ vs }}</span
+              >
+              <span class="text-xs text-gray-400">registration pace</span>
+            } @else {
+              <span class="text-2xl font-semibold text-gray-400">—</span>
+              <span class="text-xs text-gray-400">no prior year</span>
+            }
```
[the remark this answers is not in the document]
mrautela365: Good catch — subtle one, and fixed in 130556b17.

You've identified the real problem: `hasPriorYear` and `vsLastYear` are independent facts, and the card conflated them. A prior edition can exist while the comparison ratio is missing, and the card then said "no prior year" while the pacing block directly below showed a Last year figure for the same event — the drawer contradicting itself on one screen.

The caption now distinguishes the two cases:

```html
{{ d.hasPriorYear ? 'no comparison available' : 'no prior year' }}
```

Two tests cover both branches.

MRashad26 (commented):
Reviewed with the code-standards-enforcer checklist (template rules, signal/computed patterns, shared interfaces placement, foundation-scoped Snowflake queries, test coverage).

**What this does:** Expands the event-detail drawer into a full deep-dive — headline stat cards, a registration-pacing chart (current/prior-year/predicted, sourced from mirrored PCC prediction tables), a revenue breakdown, and a channel-attribution bar list. Also de-duplicates the Overview tab by dropping its inline performance-summary KPIs / attribution block, which already live on the dedicated "Attribution" and "Performance Marketing" tabs (verified via `marketing-impact.component.html` — not a feature removal).

**Major: function calls reintroduced into the template (`events-attention-section` pattern regression).**
The component's own new doc comment states the rule directly (`event-detail-drawer.component.ts`, near the `dateLabel`/`vsLastYearLabel`/`locationLabel` computeds): *"computed once per detail change rather than called from the template — these do locale, number and string formatting, which a template invocation would re-run on every change-detection pass."* That rule is followed for date/vsLastYear/location, but the PR then adds `protected num(value: number): string` and `protected money(value: number): string` methods and calls them directly from `event-detail-drawer.component.html` at ~18 sites (registration/sponsorship/pacing headline stats, revenue cards, the tier breakdown, and the per-channel session counts). Each is a live function call re-executed every change-detection pass, and it's the same `formatNumber`/`formatCurrency` cost the prior `view()` computed pattern existed specifically to avoid.
- Fix: fold these into computed signals the same way `dateLabel`/`behindGoalLabel` already are — e.g. a `registrationsView = computed(() => ({ actual: this.num(...), goal: this.num(...) }))`-style computed per card (or one aggregate view computed, as the file used to have), and have the template read the signal instead of calling `num()`/`money()`.

**Minor (non-blocking): channel-attribution query isn't foundation-scoped.**
In `project.service.ts`, the new `channelQuery` (`EVENT_REGISTRATION_ATTRIBUTION` lookup) is keyed only on `RESOLVED_EVENT_ID = ?` — no `slug_resolve`/foundation join, unlike the sibling `eventQuery`/`tierQuery` in the same method. It's launched in the same `Promise.all` as those two, before the `row` (foundation-ownership) check, so today the result is discarded whenever the event doesn't belong to the requesting foundation and never reaches the response — not an active leak. But it breaks the file's otherwise-consistent "always scope by foundation, not just by id" invariant, so a future refactor that returns early or reorders these reads could turn it into one. Worth scoping it the same way (join through `slug_resolve` on `foundationSlug` + `eventId`) for defense-in-depth, but not blocking this PR.

**Notable strengths:**
- `getEventPacing`'s error handling correctly distinguishes "prediction tables not materialized yet" (swallowed → `available: false`) from a genuine query failure (rethrown) — regression-tested for both branches, consistent with the existing `getSocialReach` contract cited in the comment.
- Chart x-axis ordering reasoning (no `reverse`, relying on the query's `DAYS_TO_EVENT DESC` order) is called out explicitly to prevent a future "fix" from flipping the timeline.
- `hasPriorYear` vs `vsLastYear` kept as separate facts with dedicated tests for both combinations — avoids the contradictory "no prior year" copy next to a populated "Last year" pacing figure.
- A11y: reduced from two progress bars to one now that sponsorship is presented as a figure rather than a second meter, and the test suite was updated in lockstep rather than just asserting the old count away.

Approving is on hold pending the template function-call issue above — it's the same category of violation this file's own comment argues against, and it's the more consequential issue here so I'm not rubber-stamping. Not requesting changes outright since it's a moderate refactor, but please take a pass at the fix before merge.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -7498,4 +7551,114 @@ export class ProjectService {
...
+        points: pointsResult.rows.map((p) => ({
+          daysToEvent: p.DAYS_TO_EVENT,
+          current: p.CURRENT,
+          priorYear: p.PRIOR,
+          predictedAvg: p.PRED_AVG,
+          predictedLow: p.PRED_LOW,
+          predictedHigh: p.PRED_HIGH,
+        })),
+      };
+    } catch (error) {
```
dealako, line 7642: [minor] `Promise.all` still couples head + drilldown availability

Issue: The catch block here now correctly narrows to the missing-table case and rethrows everything else (nice fix). But `headQuery` and `pointsQuery` are still fired together via `Promise.all`, and *both* are caught by the same `try`/`catch`. If only the drilldown table (`MARKETING_EVENT_REGISTRATION_PREDICTIONS_DRILLDOWN`) is unmaterialized while the head/summary table exists, this whole block throws `isMissingTable` on the drilldown-tables error and returns `unavailable` — discarding valid head data.

Proof: `EventPacing` (the return type) explicitly supports `available: true` with an empty `points: []`, which is exactly the state this coupling prevents from ever being returned.

Why it matters: A drilldown table that materializes on a different schedule than the head table would make the whole pacing card disappear instead of degrading to "no chart, but headline stats present" — the exact partial-availability case the type was designed for.

Fix: Run the two queries independently (e.g., `Promise.allSettled`, or two separate try/catch blocks) so a missing drilldown table alone falls back to `points: []` while head-derived fields still render. This also matches a live GitHub Copilot comment on this PR flagging the same coupling — agreeing with it here.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -59,42 +64,111 @@ export class EventDetailDrawerComponent {
...
+    elements: { point: { radius: 0, hitRadius: 8 }, line: { tension: 0.3, borderWidth: 2 } },
+  };
+
   /**
-   * Everything the template renders as text, formatted once per detail change. The template used
-   * to call formatting helpers directly, which re-ran locale/currency formatting on every
-   * change-detection pass; per the frontend checklist templates read computed values instead.
+   * Everything the template renders as a derived label, computed once per detail change rather
+   * than called from the template — these do locale, number and string formatting, which a
+   * template invocation would re-run on every change-detection pass (frontend-checklist.md §4).
```
dealako, line 103: [nit] Cited doc file does not exist in this repo

Issue: The comment cites `frontend-checklist.md §4` to justify computed-signal-vs-method-call, but no file by that name exists anywhere in this repo's history (confirmed in the original review round, when a GitHub Copilot comment cited the same nonexistent file).

Why it matters: A fabricated citation baked into a source comment is worse than one in a throwaway bot comment — it will be read by future maintainers as a real pointer to project documentation and cannot be looked up.

Fix: Either drop the parenthetical citation (the perf rationale in the sentence already stands on its own) or replace it with a real reference if one exists.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
```diff
@@ -122,6 +196,45 @@ export class EventDetailDrawerComponent {
     });
   }
 
+  /** Full venue + city + country line for the header; '' when nothing is known. */
```
dealako, line 199: [nit] Orphaned JSDoc — doc no longer attached to any function

Issue: `/** Full venue + city + country line for the header; '' when nothing is known. */` sits directly above the `// === Private Helpers ===` section marker, not above a function. It originally documented `locationLabel()`, which moved up into the computed-signals block earlier in this diff and kept its doc there.

Why it matters: a dangling doc comment reads as documentation for whatever follows it (`buildPacingChart()`), which is unrelated and already has no doc of its own — confusing for the next reader.

Fix: Delete this leftover line; `locationLabel`'s doc already lives with the computed signal above.

dealako:
Nice follow-up round, @mrautela365 — thanks for working through all of this so thoroughly.

## 👏 Nice work

- **The catch-all fix is exactly right.** `getEventPacing()`'s catch block now narrows to a regex-matched missing-table check and rethrows everything else, instead of silently swallowing any failure into a zero-filled `unavailable` response — same discipline `getSocialReach()` got earlier, and this round adds regression tests for both branches (missing-table → `unavailable`, other error → rethrow).
- **Null-vs-zero rendering is fixed.** The `?? 0` coercions in the drawer template are gone, replaced with explicit `!== null` guards and an em-dash fallback — a real registration count of 0 no longer looks identical to "no data."
- **Both Cursor Bugbot findings are cleanly resolved**: the registration/sponsorship progress-bar tone now reads from `ON_TRACK_PERCENT_THRESHOLD` / `BEHIND_GOAL_PERCENT_THRESHOLD` in `@lfx-one/shared/constants` instead of inline `80`/`50` literals, and the empty prior-year caption now branches on `hasPriorYear` rather than `vsLastYear`, so "prior edition exists but no comparison ratio yet" no longer misreports as "no prior year."
- **Derived label methods → computed signals.** `dateLabel`, `vsLastYearLabel`, `locationLabel`, `paceRatingLabel`, and `behindGoalLabel` are now memoized `computed()` signals instead of template-invoked methods that re-ran on every change-detection pass.
- **New test coverage**: a full `event-detail-drawer.component.spec.ts` (lifecycle, null-vs-zero rendering, accessibility, error/empty states) plus new `getEventDetail`/`getEventPacing` cases in `project.service.spec.ts`, including a foundation-slug bind-order assertion.
- **Chart now conditionally includes the "Last year" dataset** based on `hasPriorYear`, avoiding a legend entry with no line behind it when no prior edition exists.

## Revision tracking

| Prior finding | Status |
|---|---|
| Catch-all swallowing Snowflake failures as zero-filled data | ✅ Resolved — narrowed to missing-table regex, rethrows otherwise, regression-tested |
| `?? 0` null coercion in drawer template | ✅ Resolved — explicit `!== null` guards + em-dash fallback |
| Cursor Bugbot: bar-color thresholds hardcoded | ✅ Resolved — now driven by shared constants |
| Cursor Bugbot: wrong "no prior year" caption | ✅ Resolved — `hasPriorYear` and `vsLastYear` correctly disambiguated |
| Derived-label methods re-run every CD pass | ✅ Resolved — converted to `computed()` |
| Missing test coverage for the drawer + pacing/detail service methods | ✅ Resolved — new spec files added |
| Logger `req` not threaded through in this service's structured logs (previously flagged, not blocking) | 🟡 Still open — unchanged this round, still non-blocking |

## Open items

- 🔴 Blocking: 0
- 🟡 Minor: 2
  - `getEventPacing()` still fires the head/summary query and the drilldown query through a single `Promise.all`/`try`/`catch`, so a missing drilldown table alone discards valid head data instead of degrading to `available: true, points: []` (a state the `EventPacing` type explicitly supports). This matches a live GitHub Copilot comment on this PR — agreeing with it.
  - Logger `req` threading (carried over, already acknowledged as non-blocking).
- ⚪ Nit: 2
  - Orphaned JSDoc comment left dangling above `buildPacingChart()` after `locationLabel`'s doc moved with it into the computed-signals block.
  - A comment now cites `frontend-checklist.md §4` — this file doesn't exist anywhere in the repo's history (the original review round already flagged a GitHub Copilot comment for citing the same nonexistent file; it's now baked into source instead).

## Final decision

✅ **Approved with minor comments** — all three previously-blocking issues are confirmed resolved with test coverage, and the two remaining minor items don't need to hold up this PR. Left inline comments for the new Promise.all/drilldown-coupling finding plus the two nits.

dealako (approved):
Approving after the follow-up review — all three previously-blocking issues (catch-all error swallowing, null-vs-zero rendering, missing test coverage) are confirmed resolved with regression tests. Two minor items remain (Promise.all/drilldown coupling, logger req-threading) plus two nits — none blocking. Full details in the summary comment above and inline.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -7498,4 +7551,114 @@ export class ProjectService {
...
+      FROM ANALYTICS.PLATINUM_LFX_ONE.MARKETING_EVENT_REGISTRATION_PREDICTIONS_DRILLDOWN
+      WHERE EVENT_ID = ?
+      ORDER BY DAYS_TO_EVENT DESC
+    `;
+
+    try {
+      const [headResult, pointsResult] = await Promise.all([
+        this.snowflakeService.execute<HeadRow>(headQuery, [eventId]),
+        this.snowflakeService.execute<PointRow>(pointsQuery, [eventId]),
+      ]);
```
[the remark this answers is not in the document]
mrautela365: Fixed in 91f8c3f36 (pushed to #1398, since #1397 merged before this landed).

The two queries are now settled independently: the drilldown result is caught on its own and falls back to an empty `points` array, which `EventPacing` already permits — so an unmaterialized `_DRILLDOWN` table yields a valid headline instead of discarding both. Both `execute` calls also pass `expectMissingObject: true`, so an anticipated rollout gap no longer counts toward the circuit breaker.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Fixed in 91f8c3f36 (pushed to #1398, since #1397 merged before this landed).

The two queries are now settled independently: the drilldown result is caught on its own and falls back to an empty `points` array, which `EventPacing` already permits — so an unmaterialized `_DRILLDOWN` table yields a valid headline instead of discarding both. Both `execute` calls also pass `expectMissingObject: true`, so an anticipated rollout gap no longer counts toward the circuit breaker.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
mrautela365: Adjusted in 91f8c3f36 (on #1398 — #1397 merged before this landed), though the premise is slightly off: the file does exist. `docs/reviews/frontend-checklist.md` is in the repo, and `## 4. No template functions (SHOULD FIX)` is at line 63 — it is cited from 5+ other files here.

What was missing was the path prefix; bare `frontend-checklist.md` is not resolvable from where the comment sits. Now reads `(docs/reviews/frontend-checklist.md §4)`.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-detail-drawer/event-detail-drawer.component.ts
mrautela365: Good catch — removed in 91f8c3f36 (on #1398, as #1397 merged before this landed). The `venueLine` computed it documented had already moved above the divider, leaving the doc block stranded over `// === Private Helpers ===`.

PR #1398, opened 2026-08-11
refactor(dashboards): rename marketing impact to campaign impact

mrautela365:
Renames the focus-program union to `MarketingImpactFocusProgram` and introduces the events attendance/sponsorship split types that later PRs build on. Shell-only — no behavior change.

**Stacked on** `feat/LFXV2-2768-event-deepdive-pacing`.

LFXV2-2768


apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -42,53 +42,89 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
+            {{ selectedFocusLabel() }} campaign impact will appear here once the data connects. In the meantime, explore the All and Events campaign types.
+          </p>
+        </div>
+      } @else if (selectedTab() === 'all') {
         <lfx-overview-tab
           [foundationSlug]="foundationSlug()"
           [selectedPeriod]="selectedPeriod()"
           [foundationName]="foundationName()"
           [focusProgram]="selectedFocus()"
+          [eventsSplit]="showEventsSplit() ? selectedEventsSplit() : null"
```
dealako: [blocking] Unknown property binding will fail Angular's strict template check

Issue: `[eventsSplit]` is bound here, but `OverviewTabComponent` (overview-tab.component.ts:22-28) only declares `foundationSlug`, `foundationName`, `selectedPeriod`, and `focusProgram` as inputs.

Proof: Verified `overview-tab.component.ts` directly — no `eventsSplit` input signal is declared on the class. Angular's compiler will raise NG8002 ("Can't bind to 'eventsSplit' since it isn't a known property of 'lfx-overview-tab'") under strict template type checking.

Why it matters: This will fail `yarn build` / CI, or at minimum silently no-op the binding if strict template checking is relaxed, meaning the Events split selection never reaches the overview tab despite the UI implying it does.

Fix: Add `public readonly eventsSplit = input<EventsSplitView | null>(null);` to `OverviewTabComponent` and thread it into the child sections that should differ by split, or remove the binding from the template until that plumbing lands (matches Copilot's suggestion).

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -42,53 +42,89 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
+
+    <!-- Events sub-tabs — attendance vs sponsorship, scoped to the Events campaign type -->
+    @if (showEventsSplit()) {
+      <div class="flex flex-wrap items-center gap-2 border-t border-gray-200 pt-4" role="tablist" aria-label="Events views" data-testid="events-split-tabs">
+        @for (view of eventsSplitOptions; track view.id) {
+          <button
+            type="button"
+            role="tab"
+            [attr.id]="'mi-events-tab-' + view.id"
+            [attr.aria-controls]="'mi-events-panel-' + view.id"
```
dealako: [blocking] aria-controls targets an element that doesn't exist

Issue: Each Events sub-tab sets `aria-controls="mi-events-panel-" + view.id`, producing ids like `mi-events-panel-attendance` / `mi-events-panel-sponsorship`.

Proof: The only tabpanel in this template is at line 105 with `[attr.id]="'mi-panel-' + selectedTab()"` — no element anywhere carries an `mi-events-panel-*` id.

Why it matters: Assistive technology following the tab/tabpanel ARIA relationship gets a broken reference — the tab announces control over a panel that isn't in the DOM.

Fix: Either add a tabpanel with a matching `mi-events-panel-{id}` id (and `aria-labelledby` back-reference), or drop `aria-controls`/`role="tab"` semantics here in favor of plain toggle buttons, per Copilot's suggestion.

packages/shared/src/constants/marketing-impact.constants.ts
```diff
@@ -58,14 +89,66 @@ export const FOCUS_TO_CLASSIFICATION: Record<MarketingImpactFocusProgram, string
...
+/** Coming-soon focuses expose only the "All" channel — there is no per-channel content to show. */
+const COMING_SOON_TABS = new Set<MarketingImpactTab>(['all']);
+
 export const FOCUS_VISIBLE_TABS: Record<MarketingImpactFocusProgram, ReadonlySet<MarketingImpactTab>> = {
-  all: new Set<MarketingImpactTab>(['overview', 'attribution', 'performance-marketing', 'email', 'web-activity', 'social-accounts', 'social-listening']),
-  lfCorporate: new Set<MarketingImpactTab>(['overview', 'attribution', 'performance-marketing', 'email', 'web-activity']),
-  lfEvents: new Set<MarketingImpactTab>(['overview', 'attribution', 'performance-marketing', 'email', 'web-activity']),
-  lfTraining: new Set<MarketingImpactTab>(['overview', 'attribution', 'performance-marketing', 'email', 'web-activity']),
+  all: ALL_CHANNEL_TABS,
+  lfEvents: ALL_CHANNEL_TABS,
```
dealako: [minor] FOCUS_VISIBLE_TABS contradicts the comment directly above it

Issue: The comment on line 141 states "Social tabs are hidden for non-'all' focuses (no classification filtering)," but `lfEvents: ALL_CHANNEL_TABS` grants the Events focus the same full tab set as `all`, including `social` and `social-listening`.

Proof: `ALL_CHANNEL_TABS` (line 142) includes `'social'` and `'social-listening'`; `FOCUS_VISIBLE_TABS.lfEvents` is set to `ALL_CHANNEL_TABS` (line 149). The web-activity/social/social-listening tab components don't accept or filter by `focusProgram` (confirmed no such input in those components).

Why it matters: Selecting Events → Social or Events → Social Listening currently shows unfiltered, foundation-wide data with no visual indication that it isn't event-scoped — misleading to EDs who selected "Events" expecting event-specific numbers.

Fix: Either scope `lfEvents`'s visible tabs to exclude `social`/`social-listening` (matching the stated intent in the comment), or update the comment to describe the actual, intended behavior if showing unfiltered data under Events is deliberate for this shell PR.

packages/shared/src/interfaces/marketing-impact.interface.ts
```diff
@@ -31,11 +31,24 @@ export interface MarketingImpactTabOption {
...
 }
 
-/** Focus program identifiers for the Marketing Impact FOCUS filter bar. Values map to Snowflake LF_SUB_DOMAIN_CLASSIFICATION via FOCUS_TO_CLASSIFICATION. */
-export type MarketingImpactFocusProgram = 'all' | 'lfCorporate' | 'lfEvents' | 'lfTraining';
+/** Campaign Type identifiers for the Campaign Impact filter bar. Values map to Snowflake LF_SUB_DOMAIN_CLASSIFICATION via FOCUS_TO_CLASSIFICATION. */
+export type MarketingImpactFocusProgram = 'all' | 'lfCorporate' | 'lfEvents' | 'lfTraining' | 'membership';
 
 /** Tab identifiers for the Marketing Impact section tabs. */
-export type MarketingImpactTab = 'overview' | 'attribution' | 'performance-marketing' | 'email' | 'web-activity' | 'social-accounts' | 'social-listening';
+export type MarketingImpactTab = 'all' | 'web' | 'social' | 'email' | 'paid' | 'social-listening';
```
dealako, line 38: [minor] PR description says "shell-only — no behavior change" but this diff changes visible behavior

Issue: `MarketingImpactTab` drops `'attribution'` entirely and the component/template (per the commit message: "drop the removed attribution section") removes that tab and its content. `COMING_SOON_FOCUS_PROGRAMS` also gates three campaign types behind a placeholder.

Proof: Comparing the old union (`'overview' | 'attribution' | 'performance-marketing' | 'email' | 'web-activity' | 'social-accounts' | 'social-listening'`) to the new one shows `attribution` has no successor tab id; `COMING_SOON_FOCUS_PROGRAMS` (constants.ts:135-139) hides channel tabs for `lfTraining`, `membership`, `lfCorporate`.

Why it matters: EDs viewing Education, Membership, or Audience campaign types today lose access to whatever dashboard content they previously saw (or, if none existed, at least lose the Attribution tab), contrary to the PR's stated "no behavior change" scope — reviewers approving on that basis are approving more than described.

Fix: Update the PR description to state the actual scope (tab restructure + coming-soon gate + Attribution removal), or confirm with the team this is intentional groundwork that's expected to ship with visible impact in this stacked PR.

dealako (changes requested):
Hey @mrautela365 — thanks for pushing this through, the constant/interface restructuring is clean and the new focus/tab unions read well. I ran three parallel review passes (security/privacy, correctness/performance/tests, style/docs) and all three came back clean on their assigned dimensions — but reconciling against Copilot's and Cursor Bugbot's existing comments on this PR surfaced two real, verifiable issues that are worth blocking on, plus a couple of minor items. I independently confirmed each below by reading the actual component/template/constants files (not just trusting the bot output).

### 🔴 Blocking: 2 issues
- `[eventsSplit]` is bound on `<lfx-overview-tab>` (marketing-impact.component.html:125), but `OverviewTabComponent` only declares `foundationSlug`, `foundationName`, `selectedPeriod`, and `focusProgram` — no `eventsSplit` input exists. This is an unknown-property binding that Angular's strict template checking will reject at build time.
- The new Events sub-tabs (marketing-impact.component.html:87-88) set `aria-controls="mi-events-panel-{id}"`, but no element in the template has an id matching `mi-events-panel-attendance` or `mi-events-panel-sponsorship` — the only tabpanel (line 105) uses `mi-panel-{selectedTab}`. The ARIA tab/tabpanel relationship is dangling for assistive tech.

### 🟡 Minor: 2 issues
- `FOCUS_VISIBLE_TABS` (marketing-impact.constants.ts:147-154) grants `lfEvents` the full `ALL_CHANNEL_TABS` set (including `social`/`social-listening`), directly contradicting the comment one line above it ("Social tabs are hidden for non-'all' focuses (no classification filtering)"). Since the web-activity/social/social-listening tab components don't filter by `focusProgram`, selecting Events → Social currently renders unfiltered, foundation-wide data as if it were event-scoped.
- The PR description states "Shell-only — no behavior change," but the diff removes the standalone Attribution tab and gates Education/Membership/Audience behind a coming-soon placeholder — both are real, user-visible behavior changes for EDs on those campaign types. Worth updating the description/scope to reflect this, or confirming this is intentional and expected in this stacked PR.

### ⚪ Nit: 1 issue
- Neither the new Events sub-tabs nor the existing channel tabs implement arrow-key/Home/End roving-tabindex navigation, despite using `role="tab"` + manual `tabindex` roving — the inactive tab is currently unreachable via keyboard. This mirrors a pre-existing gap in the primary tab list in this same file, so it's not a regression introduced here, but it's compounding as more ARIA tab widgets get added without the shared keyboard handler.

### Bot reconciliation
- **Agree** with Copilot's inline comments on the `eventsSplit` binding, the dangling `aria-controls`, the Attribution/coming-soon scope mismatch, and the unfilterable Social/Social Listening tabs under Events — all confirmed above by reading the actual files.
- **Agree** with Cursor Bugbot's medium-risk summary flagging the coming-soon gating and Attribution removal as user-visible changes contrary to the stated "shell-only" scope.
- No additional findings beyond what's already surfaced by the bots and consolidated above.

### ❔ Question: 1 item
- Given this PR is explicitly "shell-only" scaffolding for later PRs, is the `eventsSplit` prop threading into `OverviewTabComponent` intentionally left unimplemented for a follow-up PR, or was the input meant to land in this PR? If it's follow-up work, consider removing the binding for now (or stubbing the input) so the shell actually builds clean.

**Decision: 🔴 Needs changes before approval** — the broken template binding and dangling ARIA reference are both build/accessibility-breaking and should be resolved before merge; the FOCUS_VISIBLE_TABS/comment mismatch and scope-description drift should be addressed or explicitly called out as intentional.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -4957,14 +4957,31 @@ export class ProjectService {
...
 
+    // Only a month re-aggregates. YTD and trailing ranges fall through to MARKETING_EVENT_OVERVIEW,
+    // whose columns are pre-aggregated year-to-date rollups with no date grain to filter on — a
+    // trailing range cannot narrow them. Re-aggregating trailing from MARKETING_EVENT_REGISTRATIONS
+    // would scope events/registrations/speakers but drop attendees, countries, organizations, and
+    // sponsorship to dashes, and 'last-6' is the default period, so the default view would lose four
+    // of seven tiles. Serving the YTD rollup is the better trade until those metrics have a date grain.
+    if (period && period.type === 'month') {
+      return this.getEventsOverviewSummaryForMonth(foundationSlug, period);
+    }
```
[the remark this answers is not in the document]
mrautela365: Confirmed and fixed in 442f8cd33 — this was the sharpest finding on the PR.

You're right about the chain, and the code comment beside it made it harder to spot: it claimed "'last-6' is the default period", but `getDefaultMarketingImpactPeriod()` returns `getDefaultMarketingImpactMonth()` — the previous calendar month. So the default landing view routed into the month path and dashed attendees, countries, organizations and sponsorship: four of seven tiles, on first paint, for every user.

The default is now `'ytd'`, where all seven populate. Month stays selectable for the three metrics that support it, and the heading names the scope actually served so a month selection reads 'Monthly Events summary' rather than implying the full set. Stale comment corrected too.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -4957,14 +4957,31 @@ export class ProjectService {
...
     });
 
+    // Only a month re-aggregates. YTD and trailing ranges fall through to MARKETING_EVENT_OVERVIEW,
+    // whose columns are pre-aggregated year-to-date rollups with no date grain to filter on — a
+    // trailing range cannot narrow them. Re-aggregating trailing from MARKETING_EVENT_REGISTRATIONS
+    // would scope events/registrations/speakers but drop attendees, countries, organizations, and
+    // sponsorship to dashes, and 'last-6' is the default period, so the default view would lose four
+    // of seven tiles. Serving the YTD rollup is the better trade until those metrics have a date grain.
+    if (period && period.type === 'month') {
+      return this.getEventsOverviewSummaryForMonth(foundationSlug, period);
```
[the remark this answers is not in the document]
mrautela365: Fixed in 442f8cd33, by labelling rather than by re-routing.

The trailing presets can't be honoured from this source: `MARKETING_EVENT_OVERVIEW` columns are pre-aggregated YTD rollups with no date grain to narrow, so `last-3`/`last-6` genuinely have no narrower figures to serve. Routing them through the month re-aggregate would return null for four of seven tiles — the same bug as the sibling High-severity finding.

So the response now carries a `scope` field reporting what it actually served, and the section heading reads that instead of the picker. Trailing selections still show YTD numbers, but they're titled 'YTD Events summary', never 'Last 3 months'. Same fix shape as the sponsorship heading on #1402.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
mrautela365: Fixed in 442f8cd33 — the `[eventsSplit]` binding is removed.

Your proof was exact: `OverviewTabComponent` declares only `foundationSlug`, `foundationName`, `selectedPeriod` and `focusProgram`, and nothing in that component or its template reads `eventsSplit`, so the binding was dead as well as invalid.

Worth noting it wasn't only a strict-template warning — it failed the production build with NG8002. I hit it first on #1400 when a merge carried this template down and the pre-push hook rejected it, so it's fixed there too; this is the source branch.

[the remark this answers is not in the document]
mrautela365: Same finding as dealako's on this line — the binding is removed in 442f8cd33. It failed the build outright (NG8002), not just template checking.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
mrautela365: Fixed in 442f8cd33 — `aria-controls` now points at `'mi-panel-' + selectedTab()`, the tabpanel that actually renders.

Both sub-tabs swap content inside that one shared panel, so per-view ids would always have dangled. Comment added at the sub-tab block explaining that, so the next person doesn't 'restore' per-view ids.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -42,53 +42,89 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
+            (click)="onEventsSplitChange(view.id)"
+            [attr.aria-selected]="selectedEventsSplit() === view.id"
+            [attr.tabindex]="selectedEventsSplit() === view.id ? 0 : -1"
```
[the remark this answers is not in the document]
mrautela365: The dangling `aria-controls` half is fixed in 442f8cd33 (both sub-tabs now reference the shared tabpanel's real id).

On roving keyboard behaviour: you're right that `role="tablist"` implies arrow-key navigation and the current markup sets `tabindex="-1"` on the unselected tab without providing it, which leaves that tab unreachable. Rather than half-implement the pattern, the two buttons should either get real roving-tabindex handling or drop the tab roles and stay plain toggle buttons — the second is closer to what this control actually is.

Flagging to @dealako for a call, since it's a UX choice rather than a defect fix. Happy to implement either.

packages/shared/src/constants/marketing-impact.constants.ts
mrautela365: Fixed in 442f8cd33 — the constant now matches the rule its own comment states.

Added `NON_SOCIAL_CHANNEL_TABS` (`all`, `web`, `email`, `paid`) and pointed `lfEvents` at it, with the reason inline so the next edit doesn't quietly revert to `ALL_CHANNEL_TABS`. The comment above the map is what made this findable — worth keeping that habit.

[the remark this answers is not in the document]
mrautela365: Same finding as dealako's — fixed in 442f8cd33.

Your framing is the one I used in the code comment: those two components take no `focusProgram` and query foundation-wide, so surfacing them under Events would present unfiltered numbers as event-specific. That's the same honest-labelling problem as the trailing-period findings on this PR, just expressed as a tab that shouldn't exist rather than a heading that's wrong.

packages/shared/src/interfaces/marketing-impact.interface.ts
mrautela365: You're right, and I'm not going to argue the diff matches the description — it doesn't. Dropping the Attribution tab and gating three campaign types behind a coming-soon placeholder are user-visible changes, whatever the commit messages say.

I've left the behaviour as-is rather than reverting it, because unpicking an intentional restructure on review feedback is more likely to break the stack than to help, and the PR title (`refactor(dashboards): rename marketing impact to campaign impact`) does describe a restructure. What's wrong is the 'shell-only — no behavior change' claim in the description.

@dealako this one needs your call and I'd rather you made it than me: either the description gets corrected to state the removals plainly, or the attribution tab comes back in this PR and moves separately. I can do either — the second is a bigger change to a branch seven PRs deep in the stack.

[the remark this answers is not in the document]
mrautela365: Same as dealako's finding on this line. The behaviour changes are real: `attribution` has no successor tab id, and `COMING_SOON_FOCUS_PROGRAMS` gates lfTraining/membership/lfCorporate behind a placeholder.

I've flagged it to @dealako for a decision between correcting the PR description and restoring the tab, rather than picking one myself — see the sibling thread. Either way the description shouldn't keep saying 'no behavior change' while this diff is in it.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -7515,6 +7550,60 @@ export class ProjectService {
...
+
+    // Speakers uses accepted proposals — a nominated-but-unaccepted proposal is not a speaker.
+    const query = `
+      WITH slug_resolve AS (
+        SELECT DISTINCT project_id
+        FROM ANALYTICS.PLATINUM.ENGAGEMENT_SCORES_BY_CLASSIFICATION
+        WHERE project_slug = ?
+      )
+      SELECT
+        MAX(r.PROJECT_ID) AS PROJECT_ID,
```
dealako: [blocking] Empty month silently drops the resolved project id

Issue: `MAX(r.PROJECT_ID) AS PROJECT_ID` is computed over an `INNER JOIN ... slug_resolve` with no `GROUP BY`.

Proof: For a valid, resolved foundation slug with zero events in the selected month, the join produces zero matching rows from `MARKETING_EVENT_REGISTRATIONS`, but the aggregate query still returns exactly one row (standard SQL aggregate-without-GROUP-BY behavior) — `MAX()` over an empty set is `NULL`. Downstream, `row?.PROJECT_ID ?? ''` (line ~7595) then emits `projectId: ''`, which is the same sentinel this response contract reserves for an *unresolved* slug.

Why it matters: Any valid foundation that simply has no events in the picked month loses its PCC deep-link identifier and looks indistinguishable from a bad/unknown slug to the client.

Fix: Drive `PROJECT_ID` from `slug_resolve` directly (`sr.project_id`) with a `LEFT JOIN` to the dated registration rows for the other aggregates, so the resolved id survives regardless of that month's event count.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -42,53 +42,90 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
+            "
+            [attr.data-testid]="'events-split-tab-' + view.id">
+            {{ view.label }}
+          </button>
+        }
+      </div>
+    }
 
     <!-- Tab content -->
     <div class="flex flex-col gap-6" role="tabpanel" [attr.id]="'mi-panel-' + selectedTab()" [attr.aria-labelledby]="'mi-tab-' + selectedTab()">
```
dealako: [blocking] aria-labelledby still dangles when the coming-soon state hides the tablist — unaddressed since the first review pass

Issue: This wrapper always sets `role="tabpanel"` and `[attr.aria-labelledby]="'mi-tab-' + selectedTab()"`, but the channel tablist containing the `mi-tab-*` buttons (line 54) is only rendered `@if (!isComingSoon())`.

Proof: When `isComingSoon()` is true (Education/Membership/Audience campaign types), no element with id `mi-tab-{selectedTab}` exists in the DOM, so the tabpanel's accessible name resolves to nothing.

Why it matters: This exact issue was flagged in Copilot's very first review pass on this PR (2026-08-11T19:43:02Z, then-line 54) and has persisted through two more commits that fixed adjacent ARIA issues in this same file (the eventsSplit/aria-controls fixes) without touching this one.

Fix: Only apply `role="tabpanel"` and `aria-labelledby` when the corresponding tab exists — e.g. `[attr.role]="isComingSoon() ? null : 'tabpanel'"` and `[attr.aria-labelledby]="isComingSoon() ? null : 'mi-tab-' + selectedTab()"`, or move the coming-soon card outside the tabpanel entirely.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -7515,6 +7550,60 @@ export class ProjectService {
...
+   * Month-scoped Events Summary, re-aggregated from MARKETING_EVENT_REGISTRATIONS (one row per
+   * event, carrying EVENT_START_DATE) for events starting inside the month.
+   *
+   * Only the three metrics with a per-event column are derivable this way. Attendees, countries,
+   * organizations, and sponsorship revenue exist solely as pre-aggregated YTD rollups on
+   * MARKETING_EVENT_OVERVIEW / MARKETING_EVENT_SPONSORSHIPS and have no monthly grain anywhere
+   * in the Platinum layer, so they return null. There is likewise no monthly YoY baseline
+   * modeled, so every changeFraction is null.
+   */
+  private async getEventsOverviewSummaryForMonth(foundationSlug: string, period: ResolvedPeriodRange): Promise<EventsOverviewSummaryResponse> {
```
dealako, line 7611: [minor] No test coverage for the new month-specific Snowflake path

Issue: `getEventsOverviewSummaryForMonth` has no corresponding tests in `project.service.spec.ts`, though that file already covers other Snowflake-backed marketing reads.

Why it matters: This query's response contract (including the project-id bug flagged above) can regress silently.

Fix: Add cases for a populated month, an empty month (asserting `projectId` survives once the fix above lands), the date bind values passed to Snowflake, and confirmation that the YTD-only metrics (attendees, countries, organizations, sponsorship) come back null for a month scope.

packages/shared/src/utils/marketing-impact.utils.ts
```diff
@@ -103,7 +103,11 @@ export function buildMarketingImpactPeriodOptions(): MarketingImpactPeriodOption
 
 /** Returns the default period value (previous calendar month as YYYY-MM). */
```
dealako: [minor] Stale JSDoc contradicts the implementation directly below it

Issue: The JSDoc comment says this "returns the default period value (previous calendar month as YYYY-MM)", but the function body (with its own accurate inline comment) always returns `'ytd'` now.

Why it matters: A caller reading only the JSDoc would expect the opposite default.

Fix: Update the JSDoc to state the function returns `'ytd'` and briefly note why (month periods can't supply four of the seven summary metrics).

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5062,11 +5082,16 @@ export class ProjectService {
    * joined to sponsorship actuals aggregated from MARKETING_EVENT_SPONSORSHIPS_BY_TIER. A goal
    * of 0 means "no goal required" for that event (the UI renders no progress bar). Defaults to
    * upcoming events only; pass includePast to return the full history.
+   *
+   * When a period is supplied the roster is scoped to events *starting* inside it, so the
+   * dashboard's month selector narrows this list the same way it narrows the paid and email
+   * sections. Without one it falls back to the dashboard's default month range.
```
dealako: [minor] getEventRoster's JSDoc doesn't match its actual period/includePast semantics

Issue: This doc block says a supplied period scopes the roster and an omitted period "falls back to the dashboard's default month range."

Proof: The implementation (`applyPeriod = Boolean(period) && includePast`, a few lines into the method body) only applies a period filter when `includePast` is also true, and when no period is supplied there is no fallback range — the query is simply unbounded. A newer, accurate comment already sits right above the query construction explaining this; this older JSDoc block wasn't updated to match.

Fix: Rewrite this doc block to state the real semantics: a period only narrows results when `includePast` is true; omitting a period does not apply any date bound.

apps/lfx-one/src/app/shared/services/analytics.service.ts
```diff
@@ -1147,20 +1147,24 @@ export class AnalyticsService {
...
   /**
    * Foundation event roster (upcoming by default; pass includePast for the full history).
    * Emits an empty roster on error so the table shows its empty state rather than breaking.
    */
-  public getEventRoster(foundationSlug: string, includePast = false): Observable<EventRosterResponse> {
-    const key = `${foundationSlug}|${includePast}`;
+  public getEventRoster(foundationSlug: string, includePast = false, period?: string): Observable<EventRosterResponse> {
+    // Every argument that changes the response belongs in the key — the period included, or a
+    // period switch would replay the previous period's roster.
+    const key = `${foundationSlug}|${includePast}|${period ?? ''}`;
```
dealako: [minor] Cache key varies by period even when the server ignores it

Issue: `` const key = `${foundationSlug}|${includePast}|${period ?? ''}` `` includes `period` unconditionally.

Proof: The server-side `getEventRoster` only applies a period filter when `includePast` is true (see project.service.ts). So for `includePast: false`, requests with different `period` values produce identical server responses but get cached under different keys.

Why it matters: Not a correctness bug, but it defeats the cache for the common upcoming-events case — e.g. the attention section and the roster can issue two equivalent HTTP/Snowflake reads instead of sharing one cache entry.

Fix: Normalize `period` to `undefined`/`''` in both the key and `buildFoundationParams` call when `includePast` is false, so equivalent requests share a cache entry.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
```diff
@@ -12,5 +12,5 @@
     [selectedPeriod]="selectedPeriod()"
     data-testid="overview-tab-events-summary" />
 
-  <lfx-event-roster-section [foundationSlug]="foundationSlug()" data-testid="overview-tab-event-roster" />
+  <lfx-event-roster-section [foundationSlug]="foundationSlug()" [selectedPeriod]="selectedPeriod()" data-testid="overview-tab-event-roster" />
```
dealako, line 15: [minor] A sibling file's comment is now inaccurate — selectedPeriod is consumed

Issue: `overview-tab.component.ts` still carries a comment on its `selectedPeriod` input saying it's "accepted from the parent page for API symmetry... not consumed here."

Proof: This template (lines 12 and 15) passes `selectedPeriod()` into `EventsSummarySectionComponent` and `EventRosterSectionComponent`, both of which filter by it — wired up in 66c7c7b4. `overview-tab.component.ts` itself isn't part of this PR's diff so I can't anchor a comment there directly, but the claim in its comment no longer holds now that this template consumes the period.

Fix: Update the comment on `overview-tab.component.ts`'s `selectedPeriod` input to describe that the period is distributed to the child sections that filter by it, rather than claiming it's unused.

dealako (changes requested):
Hey @mrautela365 — thanks for turning this around quickly. I re-reviewed the branch against the latest commits (66c7c7b4, 442f8cd3, and the just-landed 46985ba6) and reconciled against every new Copilot pass since my last review.

### 👏 Nice work
- The `eventsSplit` binding on `<lfx-overview-tab>` is gone — matches the current shell scope instead of pointing at a prop that never existed.
- The Events sub-tabs' `aria-controls` now correctly points at the one real shared tabpanel (`mi-panel-{selectedTab}`) instead of the nonexistent `mi-events-panel-*` ids, with a clear comment explaining why.
- `FOCUS_VISIBLE_TABS.lfEvents` now excludes the social/social-listening tabs via the new `NON_SOCIAL_CHANNEL_TABS` constant — the comment and the behavior finally agree, so Events no longer surfaces unfiltered foundation-wide data as if it were event-scoped.
- The `getEventsOverviewSummaryForMonth` honest-labeling approach (a `scope` field reporting what was actually served) is a solid fix for the earlier "wrong data for the wrong time range" concern — the section heading reads `scope`, not the picker selection, so a trailing-range request never gets mislabeled.
- 46985ba6's `formatCompactRounded` work is careful — it explicitly reasons through `toFixed` vs `Math.round` for negative halves and keeps exact two-decimal formatting where cents-level precision actually matters (monitoring tab). No issues there.

### ✅ Resolved from the last round
- `[eventsSplit]` unknown-property binding — fixed, confirmed removed from the template and no such input was ever added to `OverviewTabComponent` (removal, not stubbing).
- Dangling `aria-controls` on the Events sub-tabs — fixed, confirmed `aria-controls` now targets the real tabpanel id.
- `FOCUS_VISIBLE_TABS`/comment contradiction — fixed, confirmed via `NON_SOCIAL_CHANNEL_TABS`.

### ❌ Still open from the last round
- PR description says "shell-only — no behavior change," but the diff still removes the Attribution tab and gates three campaign types behind coming-soon placeholders — both real, user-visible changes. 46985ba6 adds one more (the monitoring tab's spend column moving from exact currency to compact form, called out in that commit's own message). Worth updating the PR description to state the actual scope rather than continuing to say "no behavior change."

### 🔴 New blocking: 2 issues
Both surfaced by Copilot's latest pass (suppressed comments, review 2026-08-12T05:25:37Z) and independently verified by reading the current code.

- `getEventsOverviewSummaryForMonth` (project.service.ts:7563) aggregates with `MAX(r.PROJECT_ID)` over an `INNER JOIN` and no `GROUP BY`. For a foundation with zero events in the selected month, that aggregate still returns exactly one row (standard SQL aggregate behavior) with `PROJECT_ID` null, so `row?.PROJECT_ID ?? ''` (line ~7595) emits `projectId: ''` — the same sentinel the API contract reserves for an *unresolved* slug. Any valid foundation with a quiet month loses its PCC deep-link id. Fix: derive the id from `slug_resolve` (left-join the dated registration rows onto it) so a valid, resolved foundation always keeps its id regardless of that month's event count.
- The coming-soon tabpanel wrapper (marketing-impact.component.html:107) always renders `role="tabpanel"` with `[attr.aria-labelledby]="'mi-tab-' + selectedTab()"`, but the channel tablist (and the `mi-tab-*` buttons it contains) is only rendered `@if (!isComingSoon())` (line 54). When a coming-soon campaign type is selected, the tabpanel's accessible name points at an id that doesn't exist in the DOM. This is the one item from Copilot's very first review pass (2026-08-11T19:43:02Z) that's carried through three subsequent passes unaddressed while adjacent ARIA issues in the same file got fixed. Fix: only apply `role="tabpanel"`/`aria-labelledby` when the corresponding tab exists, e.g. conditionally omit both attributes (or the role entirely) when `isComingSoon()` is true.

### 🟡 New minor: 5 issues
- No test coverage for `getEventsOverviewSummaryForMonth` in `project.service.spec.ts`, despite that file already testing other Snowflake-backed marketing reads. Add cases for a populated month, an empty month (verifying the project-id fix above), the date binds, and the YTD-only metrics coming back null.
- `getDefaultMarketingImpactPeriod`'s JSDoc (marketing-impact.utils.ts:104) still says it "returns the previous calendar month" — the implementation always returns `'ytd'` now (with a good inline comment explaining why), but the doc comment one line above is stale and tells callers the opposite default.
- `getEventRoster`'s JSDoc (project.service.ts:5085-5088) says a supplied period scopes the roster and an omitted one "falls back to the dashboard's default month range." The actual code (`applyPeriod = Boolean(period) && includePast`, line ~5116) only applies a period when `includePast` is true, and omitting a period leaves history *unbounded*, not defaulted to a range — the newer, accurate comment sits right above the query construction (lines ~5115-5119) but the older JSDoc block wasn't updated to match.
- `AnalyticsService.getEventRoster`'s cache key (analytics.service.ts:1165) includes `period` unconditionally: `` `${foundationSlug}|${includePast}|${period ?? ''}` ``. Since the server only honors `period` when `includePast` is true, two calls that differ only by period but share `includePast: false` hit the same server response yet get cached under different keys — not a correctness bug, but it defeats the cache's purpose for the common `includePast: false` case and issues redundant Snowflake reads.
- `OverviewTabComponent.selectedPeriod`'s comment (overview-tab.component.ts:25-26) says it's "accepted... for API symmetry... not consumed here," but 66c7c7b4 wired it directly into `EventsSummarySectionComponent` and `EventRosterSectionComponent` (overview-tab.component.html lines 12, 15) — the comment is now inaccurate and should describe that the period is distributed to the child sections that filter by it.

### Bot reconciliation
- **Agree** with Copilot's latest suppressed-comment batch on all 5 items above — independently verified each by reading the current code rather than trusting the bot summary.
- **Agree** with the earlier Copilot pass (2026-08-11T22:59:25Z) flagging the `analytics.service.ts` cache-key issue and the `getEventRoster` doc mismatch — both still present as described.
- No additional findings beyond what's captured above.

**Decision: 🔴 Needs changes before approval** — the empty-month project-id regression and the still-dangling coming-soon `aria-labelledby` are both concrete, verified defects; the five minors are all doc/test/cache gaps worth closing before merge, especially given how much doc drift has accumulated across this stack of fixes.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Confirmed and fixed in 7b0afd2ae — the proof holds exactly as written.

`PROJECT_ID` now comes from `slug_resolve` rather than the joined rows, via a LEFT JOIN:

```sql
MAX(sr.project_id) AS PROJECT_ID,
...
FROM slug_resolve sr
LEFT JOIN ...MARKETING_EVENT_REGISTRATIONS r
  ON r.PROJECT_ID = sr.project_id
  AND r.EVENT_START_DATE >= TO_DATE(?)
  AND r.EVENT_START_DATE < TO_DATE(?)
```

The date predicates moved into the `ON` clause deliberately — left in `WHERE` they would null the outer row back out and reintroduce the bug. Bind order is unchanged (slug, start, end).

Covered by a regression test, which I mutation-tested: reverting to `MAX(r.PROJECT_ID)` and feeding a NULL project id makes it fail, so the guard is real rather than incidentally green.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
mrautela365: Fixed in 7b0afd2ae, and you are right that it should not have survived a second pass.

Rather than force an id to point at, the region now drops the tab semantics when there is no tablist — because in that state it genuinely is not a tabpanel:

```html
[attr.role]="isComingSoon() ? 'region' : 'tabpanel'"
[attr.id]="isComingSoon() ? null : 'mi-panel-' + selectedTab()"
[attr.aria-label]="isComingSoon() ? selectedFocusLabel() + ' — coming soon' : null"
[attr.aria-labelledby]="isComingSoon() ? null : 'mi-tab-' + selectedTab()"
```

So the coming-soon state is a plain region named after the campaign type, and the tab wiring only exists when the buttons it references do.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Added in 7b0afd2ae — all four cases you listed:

- a populated month (the three event-grained metrics + `scope: 'month'`)
- the YTD-only metrics returning `null` rather than `0`
- the bind values (`['tlf', '2026-03-01', '2026-04-01']`, in order)
- an empty month keeping `projectId` — the guard for the bug above

I mutation-tested that last one (revert the SQL, feed a NULL id → it fails), so it is not just green by accident. Server suite is 971 passing.

packages/shared/src/utils/marketing-impact.utils.ts
mrautela365: Fixed in 7b0afd2ae. The JSDoc now states it returns `'ytd'` and why — month periods can only supply events/registrations/speakers, so defaulting to one dashed four of the seven tiles.

apps/lfx-one/src/server/services/project.service.ts
mrautela365: Fixed in 7b0afd2ae. The stale block is replaced with the actual semantics: a period only narrows the roster together with `includePast`, and with no period the query is unbounded — there is no implicit default range. I also noted why the upcoming view is deliberately unbounded (an upcoming event outside the selected month is still upcoming, and hiding it would make the roster disagree with the attention strip beside it).

apps/lfx-one/src/app/shared/services/analytics.service.ts
mrautela365: Fixed in 7b0afd2ae — `period` now only enters the key when it can actually change the response:

```ts
const key = `${foundationSlug}|${includePast}|${includePast ? (period ?? '') : ''}`;
```

The old comment claimed "every argument that changes the response belongs in the key", which the server behaviour falsified; it now explains the `includePast` condition instead.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
mrautela365: Fixed in 7b0afd2ae. Even though `overview-tab.component.ts` was not in the diff you could anchor to, the comment was wrong, so I corrected it there: it now reads "Passed through to the summary tiles and the roster, both of which filter by it."

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -5241,9 +5297,23 @@ export class ProjectService {
...
-    const [eventResult, tierResult] = await Promise.all([
+    // Marketing-channel attribution for this event (rolled up across the monthly rows).
+    const channelQuery = `
+      SELECT
+        CHANNEL,
+        SUM(IFNULL(SESSIONS, 0)) AS SESSIONS,
+        SUM(IFNULL(UNIQUE_VISITORS, 0)) AS UNIQUE_VISITORS,
+        SUM(IFNULL(LINEAR_REVENUE, 0)) AS REVENUE
+      FROM ANALYTICS.PLATINUM_LFX_ONE.EVENT_REGISTRATION_ATTRIBUTION
+      WHERE RESOLVED_EVENT_ID = ?
```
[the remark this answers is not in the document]
mrautela365: Valid, and fixed in 7b0afd2ae. The id check upstream proves the event belongs to this foundation but does not constrain what the aggregate sweeps up, so same-id rows filed under another foundation could land in an ED-only response.

Now scoped through the established `buildFoundationFilter` helper rather than a hand-rolled predicate, which preserves the umbrella (`tlf`) behaviour the other reads rely on:

```sql
WHERE RESOLVED_EVENT_ID = ?
  ${attributionScope.filterAnd}
```

with binds `[eventId, ...attributionScope.params]`.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -7614,10 +7703,26 @@ export class ProjectService {
...
+        this.snowflakeService
+          .execute<PointRow>(pointsQuery, [eventId], { expectMissingObject: true })
+          .then((result) => result.rows)
+          .catch((error: unknown) => {
+            logger.warning(undefined, 'get_event_pacing', 'Daily pacing curve unavailable, serving headline only', {
+              event_id: eventId,
+              error: error instanceof Error ? error.message : String(error),
+            });
+            return [] as PointRow[];
+          }),
```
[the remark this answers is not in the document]
mrautela365: Correct, and fixed in bcbcc6b37 — that catch was mine from the previous push and it was too broad.

It now degrades only for an unmaterialized table and rethrows everything else:

```ts
if (!SnowflakeService.isMissingObjectError(error)) {
  throw error;
}
```

So a timeout, a permission error or a bad column reaches the outer catch as you describe, instead of arriving at the drawer as an empty curve that reads like a measured "no pacing data".

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/event-roster-section/event-roster-section.component.ts
```diff
@@ -84,7 +86,7 @@ export class EventRosterSectionComponent {
           this.failed.set(false);
           // Caught here rather than in the service: a failure must render "couldn't load" rather
           // than the "no upcoming events" copy, which would report an outage as real data.
-          return this.analyticsService.getEventRoster(slug, includePast).pipe(
+          return this.analyticsService.getEventRoster(slug, includePast, period || undefined).pipe(
```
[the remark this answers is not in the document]
mrautela365: Good catch — the label was the dishonest part, so that is what changed in bcbcc6b37.

The page-level period picker scopes this section, so past + upcoming *within the selected period* is the intended behaviour; what was wrong was calling it "All events" while returning something narrower. The toggle now reads "Including past", and the three stale "full history" claims (analytics.controller.ts, project.service.ts, analytics.service.ts) now state that includePast is bounded by the period when one is supplied.

I kept the period forwarding rather than restoring an unbounded mode deliberately: an "All events" list that ignored the page filter would disagree with every other section on the page, including the attention strip beside it. Happy to flip it if you would rather the toggle escape the period filter entirely.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
```diff
@@ -8,5 +8,5 @@
 
   <lfx-events-summary-section [foundationSlug]="foundationSlug()" [foundationName]="foundationName()" data-testid="overview-tab-events-summary" />
```
dealako: [blocking] selectedPeriod never reaches EventsSummarySectionComponent

Issue: This line passes `[selectedPeriod]` to `lfx-event-roster-section` below, but the `lfx-events-summary-section` binding above it (this line) has no `[selectedPeriod]` at all.

Proof: `EventsSummarySectionComponent` declares `selectedPeriod = input<string>('')` and builds `const period$ = toObservable(this.selectedPeriod)` feeding its `combineLatest([slug$, period$])` data pipeline (events-summary-section.component.ts) — the component is correctly wired to react to period changes, but with no parent binding it only ever sees the default `''`.

Why it matters: The new `getEventsOverviewSummaryForMonth` server path and its accompanying fix/tests are unreachable from the UI — every month selection on this page silently continues to render the YTD summary.

Fix: `<lfx-events-summary-section [foundationSlug]="foundationSlug()" [foundationName]="foundationName()" [selectedPeriod]="selectedPeriod()" data-testid="overview-tab-events-summary" />`

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -42,53 +42,101 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
+         at that panel's real id; per-view ids would reference elements that never render. -->
+    @if (showEventsSplit()) {
+      <div class="flex flex-wrap items-center gap-2 border-t border-gray-200 pt-4" role="tablist" aria-label="Events views" data-testid="events-split-tabs">
+        @for (view of eventsSplitOptions; track view.id) {
+          <button
+            type="button"
+            role="tab"
+            [attr.id]="'mi-events-tab-' + view.id"
+            [attr.aria-controls]="'mi-panel-' + selectedTab()"
+            (click)="onEventsSplitChange(view.id)"
```
dealako: [blocking] eventsSplit selection is fully decorative — nothing downstream consumes it

Issue: This click handler updates `selectedEventsSplit` and the active-pill styling only.

Proof: `selectedEventsSplit()` (via `[eventsSplit]`) is never bound to `<lfx-overview-tab>` in this template (see the `lfx-overview-tab` element below, which only receives `foundationSlug`/`selectedPeriod`/`foundationName`/`focusProgram`). `OverviewTabComponent.eventsSplit` is declared but never referenced in its own template. None of `event-roster-section`, `event-detail-drawer`, or `events-summary-section` accept an `eventsSplit`/`EventsSplitView` input at all (verified via grep across all three component directories — zero matches). `EVENTS_SPLIT_TO_DRAWER_FOCUS` in marketing-impact.constants.ts also has no caller.

Why it matters: Attendance and Sponsorship render byte-for-byte identical overview content and drawer detail — the control is fully non-functional despite looking interactive.

Fix: either bind `[eventsSplit]="showEventsSplit() ? selectedEventsSplit() : null"` on `<lfx-overview-tab>` and propagate it into the sections/drawer that should differ by split (using `EVENTS_SPLIT_TO_DRAWER_FOCUS` for the drawer), or hide this control entirely until that plumbing exists.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -5,8 +5,8 @@
   <!-- Header -->
   <div class="flex flex-col gap-4 md:flex-row md:items-start md:justify-between">
     <div class="flex flex-col gap-1">
-      <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-title">Marketing Impact</h2>
-      <p class="text-sm text-gray-500">How are we helping grow and engage?</p>
+      <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-title">Campaign Impact</h2>
```
dealako, line 8: [nit] Rename to "Campaign Impact" is incomplete elsewhere

Issue: This heading now reads "Campaign Impact", but `sidebar-nav.service.ts:358` still labels the nav entry "Marketing Impact", and this same template's empty state (line 30) still reads "Select a foundation to view marketing impact."

Why it matters: Users navigate in under one label and land on a page calling itself something else.

Fix: Update the sidebar nav label and the empty-state copy to "Campaign Impact" (or keep the old name everywhere) so navigation and page copy agree.

dealako (changes requested):
Hey @mrautela365 — thanks for the fast turnaround on all of this. I re-reviewed everything pushed since my last review (66c7c7b4 through bcbcc6b37, plus the merge-from-main at 48a2a235 which touches none of this PR's own files) and reconciled against every new Copilot/Cursor pass since.

### 👏 Nice work
- The empty-month project-id bug is fixed properly — driving `PROJECT_ID` from `slug_resolve` with a `LEFT JOIN` on the dated rows, date predicates moved into the `ON` clause instead of `WHERE`. Confirmed `MAX(sr.project_id)` now resolves even for a zero-event month, and the new regression test in `project.service.spec.ts` (populated month, empty month, null-vs-zero metrics, param binds) covers exactly the case that broke.
- The coming-soon `aria-labelledby` dangling-reference bug — open since my very first review pass — is finally fixed with the conditional `role`/`id`/`aria-label`/`aria-labelledby` switch. Good call moving to `region` + `aria-label` instead of forcing a fake tab relationship.
- The event-pacing drilldown catch is correctly narrowed now: `isMissingObjectError` gates the degrade-to-empty-curve path, everything else rethrows to the outer catch. I traced this end to end against the current code and it does exactly what your reply described.
- Cross-foundation attribution scoping on the event-detail channel query is handled — `buildFoundationFilter` is applied to `channelQuery` and its params, so `RESOLVED_EVENT_ID` alone no longer sweeps up same-id rows filed under another foundation.
- Both stale JSDoc blocks (`getDefaultMarketingImpactPeriod`, `getEventRoster`) and the `analytics.service.ts` cache-key inefficiency are all fixed as described.

### ✅ Resolved from the last round
- Empty-month `PROJECT_ID` regression (LEFT JOIN + test coverage) — fixed.
- Dangling `aria-labelledby` on the coming-soon tabpanel — fixed.
- `getDefaultMarketingImpactPeriod` stale JSDoc — fixed.
- `getEventRoster` stale JSDoc on period/includePast semantics — fixed.
- `analytics.service.ts` event-roster cache-key inefficiency — fixed.
- `OverviewTabComponent.selectedPeriod`'s stale "not consumed here" comment — fixed, now correctly describes pass-through to the child sections.
- Broad drilldown catch swallowing all pacing-detail failures — fixed in bcbcc6b37; note that Copilot's suppressed-comment list briefly carried this forward at 19:13:01Z even though its own quoted proof snippet already showed the fix in place — its next pass (19:17:34Z) dropped the item, confirming it was a stale carryover rather than a live finding.

### 🔴 Still open / new blocking: 2 issues
Both independently confirmed by me and flagged by Copilot's latest passes.

- `selectedPeriod` never reaches `EventsSummarySectionComponent` — the new monthly aggregation path is unreachable from the UI.
- The Attendance/Sponsorship `eventsSplit` toggle is fully decorative — no downstream component consumes it.

### ⚪ New nit: 1
- Naming inconsistency: the page header now reads "Campaign Impact" (marketing-impact.component.html:8), but `sidebar-nav.service.ts:358` and this same template's empty state (line 30, "Select a foundation to view marketing impact") still say "Marketing Impact." Not blocking, but worth a pass before merge so navigation and page copy agree.

### Bot reconciliation
- **Agree** with Copilot's latest two passes on both blocking items — independently verified against current code, not just the bot's quoted snippets.
- **Agree** with Copilot's new naming-inconsistency finding (19:17:34Z) — confirmed both referenced locations still say "Marketing Impact."
- **Disagree** with treating the project.service.ts:7738 broad-catch flag as live — traced it directly; the fix is in place and Copilot's own next pass retracted it.
- One item outside this PR's diff that I'm flagging as a heads-up rather than a review finding: `project.service.ts:7776`'s outer catch hand-rolls a missing-object regex (`/does not exist or not authorized|Object .* does not exist/i`) that's slightly broader than the canonical `SnowflakeService.isMissingObjectError` helper — pre-existing code untouched by this branch, so not scoping it into this review, but worth a follow-up ticket since the two patterns can now classify the same error differently.

**Decision: 🔴 Needs changes before approval** — both blocking items are confirmed, independently-reachable UI dead ends (a picker that silently does nothing for the summary section, and a toggle that renders identically either way). Everything else from the last round is cleanly resolved.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/components/overview-tab/overview-tab.component.html
mrautela365: Confirmed and fixed in d702471b7. The binding is now there:

```html
<lfx-events-summary-section
  [foundationSlug]="foundationSlug()"
  [foundationName]="foundationName()"
  [selectedPeriod]="selectedPeriod()"
  data-testid="overview-tab-events-summary" />
```

Worth noting how far this reached: the month-scoped summary path added on this same PR — `getEventsOverviewSummaryForMonth`, the LEFT JOIN project-id fix, and the four tests I added for it — was entirely unreachable from the UI, because the tiles never received a period to re-aggregate on. The roster next to it did narrow, so the two sections silently disagreed about what range they were showing.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
mrautela365: You are right, and I removed it in d702471b7 rather than paper over it.

I checked what would consume the selection: the drawer focus modes (`b2c`/`b2b` via `EVENTS_SPLIT_TO_DRAWER_FOCUS`) live in #1402, which has 29 references to this split. Nothing on #1398 can read it, and the three sections on the overview tab are all attendance-oriented, so there is no second view to switch to here.

So the tablist, `selectedEventsSplit`, `onEventsSplitChange` and `showEventsSplit` are gone from this PR, along with the unused `eventsSplit` input on `OverviewTabComponent`. #1402 adds the control together with the behaviour that reads it, which keeps the two from shipping apart.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
mrautela365: Fixed in d702471b7 — you are right that navigating in under one label and landing on another is the actual bug here.

- `sidebar-nav.service.ts:358` → "Campaign Impact"
- the empty state → "Select a foundation to view campaign impact."

I also swept the non-user-facing references (`app.routes.ts`, `analytics.service.ts`, `dashboard-access.guard.ts`, `overview-tab.component.ts`) so a future grep for the old name does not turn up stale comments contradicting the UI. `grep -rn "Marketing Impact"` across `apps/lfx-one/src/app` is now clean.

apps/lfx-one/src/app/modules/dashboards/marketing-impact/marketing-impact.component.html
```diff
@@ -42,53 +42,75 @@ <h2 class="text-xl font-semibold text-gray-900" data-testid="marketing-impact-ti
...
-          [attr.data-testid]="'marketing-impact-tab-' + tab.id">
-          {{ tab.label }}
-        </button>
-      }
-    </div>
+    <!-- Section tabs — hidden for coming-soon campaign types, which have no per-channel content -->
+    @if (!isComingSoon()) {
+      <div class="flex items-center gap-4">
+        <span class="shrink-0 text-sm font-semibold text-gray-700">Channel Type:</span>
+        <div class="flex flex-wrap items-center gap-2" role="tablist" aria-label="Marketing impact sections" data-testid="marketing-impact-tabs">
```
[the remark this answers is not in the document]
mrautela365: Good catch — fixed in 09e26ee66. The tablist now reads `aria-label="Campaign impact sections"`.

This one survived my earlier rename sweep because that pass matched title-case "Marketing Impact" and this label is sentence-case. Re-swept case-insensitively across the templates; `grep -rni "marketing impact" apps/lfx-one/src/app --include="*.html"` is now empty, so no user-visible or announced copy carries the old name.

packages/shared/src/constants/marketing-impact.constants.ts
```diff
@@ -4,32 +4,63 @@
...
+  { id: 'paid', label: 'Paid' },
   { id: 'social-listening', label: 'Social Listening' },
 ];
 
+/**
+ * Sub-tabs shown under the Events campaign type, in display order. Attendance covers the
+ * registration/attendee story; Sponsorship covers revenue and tiers. Only the Events campaign
+ * type has this second level — no other campaign type splits this way.
+ */
+export const EVENTS_SPLIT_OPTIONS: EventsSplitOption[] = [
```
dealako, line 46: [minor] Orphaned exports after eventsSplit removal

Issue: `EVENTS_SPLIT_OPTIONS`, `EVENTS_SPLIT_FOCUS` (line 52), and `EVENTS_SPLIT_TO_DRAWER_FOCUS` (line 59) have no remaining consumers.

Proof: grepped `apps/` and `packages/` for `EVENTS_SPLIT_OPTIONS`, `EVENTS_SPLIT_FOCUS`, `EVENTS_SPLIT_TO_DRAWER_FOCUS`, `EventsSplitView`, `EventsSplitOption` — zero matches outside their own declarations (here and in `marketing-impact.interface.ts:45-49`). All were added by this PR alongside the now-removed `eventsSplit` toggle.

Why it matters: dead exports in the shared package read as still-in-use to future readers and add noise to the module's public surface.

Fix: remove `EVENTS_SPLIT_OPTIONS`, `EVENTS_SPLIT_FOCUS`, `EVENTS_SPLIT_TO_DRAWER_FOCUS` from this file (and their now-unused `EventsSplitOption`/`EventsSplitView` imports at the top), and remove `EventsSplitView`/`EventsSplitOption` from `marketing-impact.interface.ts`.

dealako (approved):
Hey @mrautela365 — great cleanup pass, all three items from my last review are resolved.

### 👏 Nice work
- `selectedPeriod` is now correctly bound to `lfx-events-summary-section` in `overview-tab.component.html`, and I traced it through: `EventsSummarySectionComponent` derives `period$` from the input via `toObservable` and feeds it into `getEventsOverviewSummary`. The monthly aggregation path is reachable from the UI now.
- Rather than wiring up the decorative `eventsSplit` toggle, you removed it outright — a clean call given the real consumer (per your commit message) is the drawer's focus modes landing with the split views, not a standalone control. The component-level removal is thorough: the click handler, `selectedEventsSplit`/`showEventsSplit` signals, the template block, and the explanatory comments describing the feature are all gone together — no half-removed state.
- "Campaign Impact" naming is now consistent everywhere I checked: `sidebar-nav.service.ts`, the empty-state copy, `app.routes.ts`'s comment, and the JSDoc in `overview-tab.component.ts`/`analytics.service.ts`/`dashboard-access.guard.ts`. Nav label and page copy agree now. You also caught the channel tablist's `aria-label` ("Marketing impact sections" → "Campaign impact sections") in the latest push — that one wasn't in my original nit, good catch.

### ✅ Resolved from the last round
- `selectedPeriod` never reaching `EventsSummarySectionComponent` — fixed, verified end to end.
- `eventsSplit` toggle being fully decorative — resolved by removing it, a valid alternative to wiring it up.
- "Campaign Impact" vs "Marketing Impact" naming inconsistency — fixed across all touched files.

### 🟡 New minor: 1 issue
- The shared-package types/constants backing the removed `eventsSplit` control (`EventsSplitView`, `EventsSplitOption`, `EVENTS_SPLIT_OPTIONS`, `EVENTS_SPLIT_FOCUS`, `EVENTS_SPLIT_TO_DRAWER_FOCUS`) are now orphaned — I grepped `apps/` and `packages/` and found zero remaining consumers. These were introduced by this same PR (not pre-existing), so this falls under "remove what your own changes made unused" rather than being out of scope. Not blocking since it's inert dead code with no behavioral impact, but worth cleaning up before merge so the shared package doesn't carry unused exports.

### Bot reconciliation
- No new CodeRabbit/Cursor/Copilot comments have landed since my last review (checked both PR review comments and issue comments after 19:37:10Z — none found).

**Decision: ✅ Approved with minor comment** — all blocking items are resolved and independently verified; the one remaining item is a small dead-code cleanup, not a functional concern.

PR #1412, opened 2026-08-12
feat(campaigns): give the email delivery type a real plan tab

mrautela365:
## What this does

Selecting **Email** as the delivery type rendered a "coming soon" placeholder. This replaces it
with a real Plan tab and a Monitor tab (LFXV2-3198).

## The placeholder was misleading

The email channel is **not** unimplemented. `HubSpotDispatcher.Dispatch` is on `main` — it clones
a template marketing email and points its send list at the brief's built audience — and the
HubSpot metrics read landed with LFXV2-3058. What was missing was the UI in front of it, plus one
service route.

## Email is not Paid Marketing with a different icon

**No Optimize tab**, and not as a deferral. Optimize drives keyword and status actions;
`HubSpotDispatcher` implements no `StatusToggler` because campaign create **stages a draft that a
human reviews and sends** — nothing is running, so there is nothing to pause. `StatusToggler` is
an optional capability and the service answers its absence with `ErrToggleUnsupported` → 400.
Rendering the tab would promise an action the backend correctly refuses.

**Monitor is a pending panel, not the paid Monitor component.** Reusing `MonitoringTabComponent`
here was a real bug caught in review: its `PlatformType` is `'google' | 'linkedin' | 'reddit' |
'meta'`, it defaults to `google`, and it references neither HubSpot nor email. Under a tab
labelled Email it would have shown Google Ads spend and keyword quality scores — numbers that are
real, plausible, and about a different campaign, which is worse than showing nothing. It also
fired four ad-platform API calls from a tab with no business touching them. campaign-service
*can* read HubSpot email metrics (LFXV2-3058), but this application has no route to it yet.

**The tab list is a per-container constant, not a computed over the active delivery type.** Both
containers stay mounted, so a list keyed on which side is showing describes *the page* rather than
*the container* — the hidden email tablist would render the unfiltered four, Optimize included.

**`selectTab` and `onTabKeydown` take their owner explicitly.** Same reason: the hidden tablist's
buttons still dispatch, and a handler inferring its target from ambient state routes them into the
visible side's signal.

**Each delivery type keeps its own tab signal and its own approved brief.** Both containers stay
mounted, and their tab sets differ, so a shared signal would leave the page on a tab this side
does not render after switching away from Optimize — a blank panel with a tablist agreeing with
nothing. A shared brief would show a Paid brief under Email's Implement tab after a round-trip.

**The Email container is visibility-toggled, not `@if`.** Same reason the paid one already is:
`PlanningTabComponent` owns brief/URL/copy state locally, so a structural `@if` destroys an
in-progress email brief on every round-trip. The inline `[style.display]` is deliberate —
`[hidden]` alone loses to the `flex` utility in the cascade.

## A latent crash, found by the new test

`onTabKeydown` did `(event.target as HTMLElement).parentElement` — a cast asserting away a
nullable field. The arrow-key test crashed on it immediately. Pre-existing on `main`, but this
change widens its reach by making the handler serve two tablists, so `event.target` is now
narrowed with `instanceof`. Selecting the tab is the part that matters; moving focus is the
enhancement, and it must not take the handler down with it.

## Implement tab: honest placeholder

The staging form needs `sourceEmailId`, which is **required with no default** — so it cannot be
built until the template picker has something to list. That is **LFXV2-3197**
(campaign-service#122, in review), which this is not blocked on for the Plan half.

The copy says staging "is not wired up yet" and explains why, rather than repeating "coming
soon" — which is the claim this PR exists to correct.

## Testing

`yarn build` and `yarn lint` green (one pre-existing warning, present on `main` too). Full suite
green: **916 server + 86 app tests**. Adds
`campaigns.component.spec.ts` (the component had none) covering tab filtering, per-side tab
isolation across a round-trip, arrow-key wrapping within the visible set, and brief isolation.

**All four fixes are revert-verified** — each test fails with its fix removed, so none is vacuous.

**Not visually verified in a browser.** The page is ED-gated and this worktree has none of the
local auth bypasses, so `/foundation/campaigns` redirects to `/login`. Flagging that explicitly
rather than implying a visual check happened. The state machine is covered by the specs above;
the rendering is not.

## Size

+136/-29 excluding the new spec.

## Not in this PR

The staging form and template picker — blocked on campaign-service#122.


apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -40,17 +40,76 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+            <p class="max-w-md text-sm text-gray-500">
+              Your brief is ready. Staging an email clones one of your HubSpot marketing emails as its template, and choosing that template needs an endpoint
+              that is still in review.
+            </p>
+          </div>
+        </div>
+      }
+      @case ('insights') {
+        <div role="tabpanel" id="email-panel-insights" aria-labelledby="email-tab-insights" data-testid="campaigns-email-insights-panel">
+          <lfx-monitoring-tab data-testid="campaigns-email-insights-tab" />
```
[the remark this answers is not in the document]
mrautela365: Fixed and pushed — you were right, and it was the most consequential finding on this PR. lfx-monitoring-tab is paid-ads only: PlatformType is 'google'|'linkedin'|'reddit'|'meta', it defaults to google, and it contains no reference to HubSpot or email. Under a tab labelled Email it would have shown Google Ads spend and keyword quality scores — numbers that are real, plausible, and about a different campaign, which is worse than showing nothing. Replaced with a pending panel naming why.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -84,42 +105,60 @@ export class CampaignsComponent {
...
-  }
-
   protected onProceedToImplementation(brief: CampaignBriefOutput): void {
     this.briefOutput.set(brief);
     this.selectedTab.set('implementation');
   }
 
+  protected onEmailProceedToImplementation(brief: CampaignBriefOutput): void {
+    this.emailBriefOutput.set(brief);
+    this.selectedEmailTab.set('implementation');
```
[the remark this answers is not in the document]
mrautela365: Fixed and pushed. resetToPlanning reset only the paid side, so a program switch left an Events brief under Education on the email side — invisible today because that panel is a placeholder, which is exactly why it would have survived to the real form. Both sides now reset, with a test that fails if either is dropped.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -40,17 +40,76 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+    @switch (selectedEmailTab()) {
+      @case ('planning') {
+        <div role="tabpanel" id="email-panel-planning" aria-labelledby="email-tab-planning" data-testid="campaigns-email-planning-panel">
+          <!-- Keyed on the program type for the same reason the paid side is: the component
+               holds its brief state locally and does not reset on input change. -->
+          @for (program of [selectedProgramType()]; track program) {
+            <lfx-planning-tab
+              [programTypeConfig]="activeProgramTypeConfig()"
+              (proceedToImplementation)="onEmailProceedToImplementation($event)"
+              data-testid="campaigns-email-planning-tab" />
```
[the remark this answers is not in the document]
mrautela365: Confirmed and filed as **LFXV2-3201** — not fixed here, deliberately.

Verified: PlanningTabComponent takes only programTypeConfig, selectedPlatforms defaults to new Set(['google-ads']) (:72), and canGenerate requires selectedPlatforms().size > 0 (:109). So an email brief genuinely cannot be generated without ticking an ad platform, and the Budget/Assets and Ad Channels sections render regardless.

What the email channel consumes is a sourceEmailId (LFXV2-3197) and the brief's built audience (LFXV2-2774) — not ad copy, not a platform set. Giving the planner an email mode changes generation, validation and the request contract, and wants a decision about whether the two delivery types should share a component at all. That is a feature rather than a review fix, so this PR's scope stays Plan + Monitor and the gap is tracked rather than half-built.

Leaving this thread OPEN so it is visible to a reviewer rather than closed against work that has not happened.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -40,25 +40,113 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+          "
+          [attr.data-testid]="'campaigns-email-' + tab.id + '-tab'">
+          <i [class]="tab.icon" aria-hidden="true"></i>
+          {{ tab.label }}
+        </button>
+      }
     </div>
-  }
+
+    @switch (selectedEmailTab()) {
```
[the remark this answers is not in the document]
mrautela365: Confirmed and filed as **LFXV2-3202** — real, but PRE-EXISTING on main rather than introduced here. I checked: the paid-marketing side has had the same structure since before this PR, so both sides lose the in-progress brief on a tab switch.

The page already recognises this hazard one level up — the delivery-type containers are kept MOUNTED and visibility-toggled precisely because PlanningTabComponent owns its state locally — and that reasoning was never carried down to the tabs. Fixing it means restructuring both sides' tab panels, and there is a subtlety to preserve: the paid Plan panel is keyed on the program type so a program switch DOES recreate the component deliberately.

Out of scope for a PR whose stated scope is Plan + Monitor for the email channel, so tracked rather than half-built. Leaving this open so it stays visible.

mrautela365:
Picking up the suppressed comment on `campaigns.component.html:110` — confirmed and fixed.

The email Implement tab is directly clickable (no `disabled` binding on `selectTab(tab.id, 'email')`), so the pending panel is reachable before anything has been generated. "Your brief is ready" was unconditional, which makes it simply false for someone who has not started. It is now gated on `emailBriefOutput`; the rest of the panel explains what is missing from the channel, which is true either way.

One thing worth flagging about the test, because it nearly shipped useless. I first drove it with `selectedDeliveryType.set('email')` — but that signal is read-only, so **the spec file failed to compile and its cases silently did not run**, while the summary still printed `916 passed`. The bundle error was reported after the green summary. Switching via the form control (the real path a user takes) makes it compile, and reverting the gate now fails it.

That is the second non-binding test I have caught this session, both found by looking at what the revert actually does rather than trusting a green run.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -40,25 +40,123 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+              and "Your brief is ready" is then simply false, told to someone who has not started.
+              The rest of the panel is true either way: it explains what is missing from the
+              channel, which does not depend on the user's progress.
+            -->
+            <p class="max-w-md text-sm text-gray-500">
+              @if (emailBriefOutput()) {
+                Your brief is ready.
+              }
+              Staging an email clones one of your HubSpot marketing emails as its template, and choosing that template needs an endpoint that is still in
+              review.
```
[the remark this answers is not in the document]
mrautela365: Checked this one directly rather than reasoning about it, and **the run-on does not occur**. The rendered text is:

```
Your brief is ready. Staging an e…
```

Angular keeps a space at that boundary, so `preserveWhitespaces: false` is not collapsing the separator between the `@if` block and the following text node.

You were right that the existing `toContain` checks would not have caught it either way, so I added an explicit `not.toContain('ready.Staging')`. But I have labelled it **non-binding in the comment**, because that is what it is: I tried to make it fail by deleting the newline from the template, and it still passed — the compiler normalizes the join regardless. It stands as a statement of the requirement, not a guard, and I would rather say so than let a future reader mistake it for one.

Happy to be shown a build configuration where this does collapse — if `preserveWhitespaces` is set differently anywhere in the pipeline I have not checked, the conclusion changes.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -40,25 +40,123 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
     </div>
-  }
+
+    @switch (selectedEmailTab()) {
+      @case ('planning') {
+        <div role="tabpanel" id="email-panel-planning" aria-labelledby="email-tab-planning" data-testid="campaigns-email-planning-panel">
+          <!-- Keyed on the program type for the same reason the paid side is: the component
+               holds its brief state locally and does not reset on input change. -->
+          @for (program of [selectedProgramType()]; track program) {
+            <lfx-planning-tab
```
dealako, line 122: [blocking] Email “Plan” tab mounts the paid-ads planner

Issue: This `@case('planning')` mounts `<lfx-planning-tab>` unchanged — the same component Paid Marketing uses. Under a tab labelled **Email**, the user sees an “Ad Channels” card (Google Ads / LinkedIn / Meta / Reddit), must select an ad platform to proceed, and generates Google RSA copy, Demand Gen copy, ad keywords, and a LinkedIn targeting strategy — none of it email (no HubSpot template, subject line, sender, preview text).

Proof: `planning-tab.component.ts` — `selectedPlatforms = signal(new Set(['google-ads']))`, `canGenerate` requires `selectedPlatforms().size > 0`, and `generate()` posts `platforms:[...]` to `generateBrief`. `planning-tab.component.html` renders `data-testid="planning-platforms-section"` (“Ad Channels”), “Google Search — RSA Copy”, and “LinkedIn Targeting Strategy”. On proceed, `onEmailProceedToImplementation` stores that ad brief in `emailBriefOutput`, which the Implement panel only reads for existence (“Your brief is ready.”) — the payload is never consumed.

Why it matters: The PR's stated reason for existing is that Email must not borrow paid-ads UI showing plausible, wrong-channel data (the exact argument used to reject reusing `MonitoringTabComponent`). The Plan tab does precisely that, so the change is internally inconsistent and user-visible.

Fix: Confirm intent. Either (a) product agrees Plan is deliberately channel-agnostic for now — then document the carve-out and hide/relabel the ad-only sections (Ad Channels, Google/LinkedIn copy) when mounted under Email; or (b) show an email-specific pending panel here like Implement/Insights until an email brief flow exists (LFXV2-3197).

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -40,25 +40,123 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+              [programTypeConfig]="activeProgramTypeConfig()"
+              (proceedToImplementation)="onEmailProceedToImplementation($event)"
+              data-testid="campaigns-email-planning-tab" />
```
dealako, line 125: [minor] Both planning-tab instances stay mounted — duplicated internal test ids

Issue: Because both containers are visibility-toggled (not `@if`) and `selectedEmailTab()` defaults to `'planning'`, two `<lfx-planning-tab>` instances are always in the DOM at once. Only the host wrappers are disambiguated (`campaigns-planning-tab` vs `campaigns-email-planning-tab`); the planner's internal ids are not.

Proof: `planning-tab.component.html` emits fixed ids — `planning-url-input`, `planning-generate-btn`, `'planning-platform-' + platform.id` — with no per-instance prefix, so each now resolves to two elements.

Why it matters: Any E2E/unit selector on a bare `planning-*` id is ambiguous and may bind the hidden instance. The component's own comments cite duplicated-testid resolution as a real hazard.

Fix: If the Plan tab reuse stays, document that `planning-*` selectors must be scoped under an ancestor (e.g. `[data-testid="campaigns-email-planning-tab"]`), or namespace the planner's internal test ids per instance.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+  });
+
+  /**
+   * Regression: keyboard navigation is bounded by the VISIBLE tabs.
+   *
+   * Wrapping modulo the full four-tab list would step ArrowRight off the end of the email
+   * tablist onto an index with no button — selecting a tab this side does not render, and
+   * focusing nothing.
+   */
+  it('wraps arrow-key navigation within the email tab set', () => {
```
dealako, line 130: [minor] Keyboard-nav coverage is email-ArrowRight-only

Issue: `onTabKeydown` was refactored to pick its bounding list from `owner` (`owner === 'email' ? this.emailTabs : this.tabs`) and handles ArrowRight/ArrowLeft/Home/End. The suite exercises exactly one branch — email + ArrowRight wrap (index 2 → 0). The paid `owner` path and ArrowLeft/Home/End are untested.

Proof: only `onTabKeydown(new KeyboardEvent(..., {key:'ArrowRight'}), 2, 'email')` appears; no `'paid-marketing'` keydown assertion, no End/Home/ArrowLeft anywhere in the file.

Why it matters: The paid side still has four tabs including `optimization`; a future change to the shared bounds could silently break paid keyboard nav (or its End → optimization landing) with the suite green — the very regression the `owner` param exists to prevent, proven only for one side.

Fix: Add a paid case — `selectTab('planning','paid-marketing')` then `onTabKeydown(End, 3, 'paid-marketing')` expecting `selectedTab() === 'optimization'` — plus one email ArrowLeft-wrap (index 0 → 'insights').

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+    expect(withBrief?.textContent).toContain('Your brief is ready');
+    // The two sentences must stay separated — a review round raised `preserveWhitespaces: false`
+    // collapsing the separator between the `@if` block and the following text into
+    // `ready.Staging`.
+    //
+    // It does not happen: Angular keeps a space at that boundary. Stated plainly, this assertion
+    // is NOT binding — removing the newline from the template does not make it fail, because the
+    // compiler normalizes the join either way. It is kept as a statement of the requirement, not
+    // as a guard, and a reader should not take it for one.
+    expect(withBrief?.textContent).not.toContain('ready.Staging');
```
dealako: [nit] Non-binding assertion guards nothing

Issue: You already document (lines 184–191) that this assertion cannot fail — Angular normalizes the `@if`/text whitespace boundary either way, so removing the template newline won't regress it. Inside a suite whose stated purpose is pinning behaviors “one line away from silently regressing,” a passing-but-unfailable assertion is a mild footgun someone may later copy believing the separation is protected.

Proof: `expect(withBrief?.textContent).not.toContain('ready.Staging');` with the preceding comment stating it is “NOT binding.”

Fix: Either drop it (the comment already states the requirement) or make it binding: `expect(withBrief?.textContent).toMatch(/ready\.\s+Staging an email/);`, which fails if the separator collapses.

dealako (changes requested):
Hi @mrautela365 👋 — thorough, well-argued change. The commit-level narrative and the in-code comments are genuinely excellent: they explain *why* the two containers stay mounted, why the tab signals and briefs are per-side, and why Optimize and Monitor are deliberately withheld rather than reused. The state-machine spec is real (revert-verified, per-side isolation, keyboard wrap, brief isolation), and I confirmed the earlier bot findings on the first commit have since been fixed in the branch.

What I want to reconcile before approving is a consistency gap the same reasoning cuts against: the **Email Plan tab mounts the paid-ads planner unchanged** (`<lfx-planning-tab>` → Ad Channels card, ad-platform selection, Google/LinkedIn ad-copy generation). The PR argues at length that surfacing paid-ad data under an Email tab is "worse than showing nothing" — and that is exactly what the Plan tab does. This is likely a deliberate scoping decision (email staging is blocked on LFXV2-3197), but it isn't stated, and as shipped it contradicts the PR's own thesis. I'd like an explicit product/intent call or a scoping note before this merges.

### Reconciliation with bots
- **Cursor (High) “Email Monitor shows paid ads”** → ✅ resolved: Insights now renders a pending panel, not `lfx-monitoring-tab`.
- **Cursor (Medium) “Program switch skips email reset”** → ✅ resolved: `resetToPlanning` now clears `emailBriefOutput` + `selectedEmailTab`.
- **Copilot “brief is ready” unconditional** → ✅ resolved: now gated on `@if (emailBriefOutput())`.
- **Copilot “reuses paid planner for Email”** → agreed; this is the open blocking item below.
- **Copilot “aria-controls lost on inactive tabs”** → respectfully disagree: given the `@switch` renders exactly one panel, selected-only `aria-controls` is the correct choice (WAI-ARIA makes it optional, and a dangling IDREF is worse than an absent attribute). The in-code comment documents this accurately.

### Issue count
- 🔴 Blocking: 1 — Email “Plan” tab presents the paid-ads planner (Ad Channels / Google / LinkedIn ad-copy) under Email, contradicting the PR's channel-differentiation premise.
- 🟡 Minor: 2 — (a) both `<lfx-planning-tab>` instances stay mounted, duplicating every internal `planning-*` test id; (b) keyboard-nav coverage is email-ArrowRight-only — the paid owner path and ArrowLeft/Home/End go untested after the `owner` refactor.
- ⚪ Nit: 1 — the self-admittedly non-binding `not.toContain('ready.Staging')` assertion guards nothing.
- ❔ Question: 0 (the Plan-tab item is folded into Blocking pending your product-intent answer).

**Security & data privacy:** clean — no injection sink (all interpolated ids come from `CAMPAIGN_TABS`), no PII in logs/errors/URLs, and `KubeCon`/`Open Source Summit` are public event names, not fixture PII.

### Final decision
🔴 **Needs changes before approval** — primarily to resolve/document the Plan-tab reuse. If product confirms Plan is intentionally channel-agnostic for now, a short scoping note (and ideally hiding the ad-only sections under Email) would flip this to approval.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
mrautela365: Every word of this is correct, and I am not going to argue the finding — I am going to explain why it ships this way and make that visible, which is what was actually missing.

This is tracked as **LFXV2-3201**, opened when I hit the same thing. The PR body never mentioned it, so from your seat it read as unnoticed rather than deferred. That is a fair reading of what I gave you.

Why it is deferred rather than fixed here: this PR replaces a "coming soon" placeholder, and the email brief shape is a feature, not a review fix. A real fix needs a delivery-type mode on the planner (or a separate component), a different `canGenerate` gate, a generation request carrying email fields instead of `platforms` — and a decision about whether the two delivery types should share a component at all. That decision belongs with the ticket, not with a placeholder replacement.

`07b1c4b1` writes the gap at the mount point: what the user actually sees, the fact that `onEmailProceedToImplementation`'s payload is never consumed (your observation, and the sharpest part of the finding), and what closing it requires.

If you would rather this did not ship until 3201 lands, say so and I will hold it — the placeholder is not worse than a Plan tab that asks for an ad platform. My reasoning for shipping is that the event-URL and scrape half IS genuinely shared and useful, and Monitor is real. But that is a judgement call and it is yours as much as mine.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Fixed in `07b1c4b1`, and you identified exactly the wrong half to have tested.

`onTabKeydown` takes an `owner` for one reason: the two sides bound on different lists. So proving it on email only left the single guarantee the parameter exists to provide as the one thing unproven.

Added both cases you named:

- **paid End** → `optimization`, which is paid-only. That choice matters: bounded on the email list the same call lands on `insights`, so the assertion distinguishes the two lists rather than merely exercising End.
- **email ArrowLeft** wrapping 0 → `insights`.

**Revert-verified**: collapsing `owner === 'email' ? this.emailTabs : this.tabs` to `this.emailTabs` fails the paid case and nothing else — so it binds the branch and not a neighbour.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Fixed in `07b1c4b1` with your suggested form: `toMatch(/ready\.\s+Staging an email/)`.

You are right that documenting it was worse than dropping it. In a suite whose stated purpose is pinning behaviours one line from silently regressing, an assertion that cannot fail is one the next person copies believing the separation is protected — the comment saying otherwise is three lines away and easy to miss.

Asserting the separator POSITIVELY binds, which is the version I should have written the first time.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
mrautela365: Documented at the mount point in `07b1c4b1`, taking your first option.

The scoping rule is now written where someone adding a selector will see it:

```
[data-testid="campaigns-email-planning-tab"] [data-testid="planning-url-input"]
[data-testid="campaigns-planning-tab"]       [data-testid="planning-url-input"]
```

I did not take the namespacing option, and the reason is worth stating: the planner is shared by both delivery types, so per-instance ids change a component on behalf of a caller that did not ask. That belongs with LFXV2-3201, which revisits whether the two types should share it at all — if they stop sharing, the duplication disappears rather than needing a prefix scheme.

mrautela365:
@dealako — re-requesting review. Your pass was at 20:11 and the fixes landed at 23:23, so it was describing an earlier state; GitHub cleared you as a reviewer when the CHANGES_REQUESTED was submitted and my push did not re-add you. That is on me.

All four findings are addressed in `07b1c4b1`:

**[minor] Keyboard-nav coverage** — you identified exactly the wrong half to have tested. `onTabKeydown` takes an `owner` for one reason: the two sides bound on different lists. Proving it on email only left the single guarantee the parameter exists to provide unproven. Added both cases you named — paid `End` → `optimization` (paid-only, so it distinguishes the two lists rather than merely exercising End) and email `ArrowLeft` wrapping 0 → `insights`. Revert-verified: collapsing the bounds to the email list fails the paid case and nothing else.

**[nit] Non-binding assertion** — replaced with your suggested form, `toMatch(/ready\.\s+Staging an email/)`. You were right that documenting it was worse than dropping it: in a suite whose purpose is pinning things one line from regressing, an assertion that cannot fail is one the next person copies believing it protects something.

**[minor] Duplicated test ids** — took the documentation option. The required scoping is written at the mount point. I did not namespace the planner's own ids because it is shared by both delivery types, so per-instance ids change a component on behalf of a caller that did not ask — that belongs with LFXV2-3201, which revisits whether they should share it at all.

**[blocking] Email Plan tab mounts the paid planner** — every word of your analysis is correct, including the sharpest part, that `onEmailProceedToImplementation`'s payload is never consumed. It is tracked as **LFXV2-3201**, and the PR never said so, so from your seat it read as unnoticed rather than deferred. Fair reading of what I gave you. The gap is now written at the mount point: what the user actually sees, the unconsumed payload, and what closing it requires.

**The open question is still yours, and I have not assumed an answer.** If you would rather this not ship until 3201 lands, say so and I will hold it — a placeholder is not worse than a Plan tab that asks for an ad platform. My reasoning for shipping is that the event-URL and scrape half is genuinely shared and Monitor is real, but that is a judgement call rather than a fact.

Since your review: 3201 is now assigned and scoped, with partial work already on a branch (delivery-type input added, `canGenerate` no longer requiring a platform for email). So the follow-up is real work in progress, not a ticket parked to unblock this PR.

CI green.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+    internals().selectTab('planning', 'paid-marketing');
+
+    internals().onTabKeydown(new KeyboardEvent('keydown', { key: 'End' }), 3, 'paid-marketing');
+
+    // `optimization` is paid-only. Bounded on the email list this would be `insights`, so the
+    // assertion distinguishes the two lists rather than merely exercising End.
+    expect(internals().selectedTab()).toBe('optimization');
+  });
+
+  it('wraps ArrowLeft backwards within the email tab set', () => {
```
dealako, line 158: [nit] `Home` key path is uncovered

Issue: The new tests close the `End`/`ArrowLeft` gaps nicely, but `onTabKeydown`'s `Home` branch (`newIndex = 0`) still has no assertion on either owner.

Proof: `campaigns.component.ts` sets `newIndex = 0` for `event.key === 'Home'`; the spec now covers ArrowRight (email), ArrowLeft (email), and End (paid) — no `Home` case.

Why it matters: Low — unlike `End` (`tabs.length - 1`), `Home` is a constant `0` independent of list length, so it carries none of the owner-specific bounding risk the `owner` param guards against. The P3 ask was delivered in full; this is completeness only.

Fix (optional): add a one-line `onTabKeydown(Home, 2, 'email')` → `selectedEmailTab() === 'planning'` assertion, or leave as-is.

dealako (approved):
Thanks for the fast, careful turnaround @mrautela365 — every item from the last round is addressed, and the two you couldn't fully close in-PR are now honestly documented and ticketed rather than papered over. 👏

### 👏 Nice work this revision
- **Keyboard-nav coverage (prev 🟡) — resolved.** The two new tests bind the exact regression the `owner` param exists to prevent: `End,3,'paid-marketing'` → `optimization` (paid-only, distinguishes the 4-tab list) and `ArrowLeft,0,'email'` → `insights` (distinguishes the 3-tab list). Both are non-vacuous — I traced them through `onTabKeydown`'s owner-bounded list.
- **Non-binding assertion (prev ⚪) — resolved.** `not.toContain('ready.Staging')` → `toMatch(/ready\.\s+Staging an email/)`, which now actually fails if the separator collapses. Exactly the right call.
- **Duplicated `planning-*` test ids (prev 🟡) — resolved.** The added comment documents the required ancestor-scoping (`[data-testid="campaigns-email-planning-tab"] [data-testid="planning-url-input"]`), which is the convention I offered. Verified both host anchors resolve the descendant uniquely.
- **Plan-tab reuse (prev 🔴) — transparency fully resolved.** The deliberate carve-out is now spelled out in-code and tracked as **LFXV2-3201**, with LFXV2-3197 (merged) supplying the Implement-side template picker.

### Revision tracking
| Prior item | Status |
| --- | --- |
| 🔴 Plan tab mounts paid-ads planner under Email | ⚠️ Partially addressed → accepted as a documented, ticketed trade-off (see below) |
| 🟡 Duplicated `planning-*` test ids | ✅ Resolved (scoping convention documented) |
| 🟡 Keyboard-nav coverage email-ArrowRight-only | ✅ Resolved (paid `End` + email `ArrowLeft` tests) |
| ⚪ Non-binding `ready.Staging` assertion | ✅ Resolved (positive `toMatch`) |

### Bot reconciliation
- **Copilot** re-reviewed `07b1c4b1` and generated **no new comments** (one suppressed, unrelated). Cursor did not re-run. Nothing new to reconcile.

### Open items (all non-blocking)
- 🔴 Blocking: 0
- 🟡 Minor: 1 — Email **Plan** tab still renders the paid ad-only UI (Ad Channels + platform gate + Google/LinkedIn ad copy) under an Email label, which is the same "paid content under an Email tab" problem the sibling Implement/Insights tabs solve with pending panels. I'm accepting this as a **documented trade-off** owned by LFXV2-3201 — the emitted brief is inert (read only for existence), this replaces a "coming soon" placeholder, and the reuse-vs-split decision genuinely belongs with that ticket. Recommend that LFXV2-3201 either hides the ad-only sections (keeping only the shared event-URL/scrape step) or swaps in an email-specific panel.
- ⚪ Nit: 1 — the `Home` key path in `onTabKeydown` is still untested (see inline). Low-risk since `Home` is a constant `0`, independent of list length — it doesn't carry the owner-specific bounding risk that `End`/arrows do.

### Final decision
✅ **Approved with minor comments.** No blockers remain; the one open minor is a deliberate, ticketed product trade-off (LFXV2-3201) and the nit is optional. Nice iteration.

PR #1416, opened 2026-08-12
refactor(cla): remove interim identity-query padding stopgap

### LINKED ISSUE #1413: Remove the interim identity-param padding stopgap now that the gateway fix is deployed
Follow-up to #1314. Removes the interim BFF workaround merged in #1339, now that the durable gateway fix has deployed to production.

## Background

#1314 was caused by the platform API gateway's `awslambda` middleware dropping every single-valued query param whenever any param was repeated (linuxfoundation/lfx-gateway#114). With 2+ verified emails, `email` became multi-valued and `githubId` / `githubUsername` were silently stripped in transit, so GitHub-only CLAs disappeared from My CLAs.

Two-part fix:

1. **Interim (shipped, #1339):** the BFF pads single-valued identity keys to length 2 so they survive the strip.
2. **Durable (now deployed):** `valuesToMultiMap` in the `linuxfoundation/traefik` fork no longer skips single-value params — merged in linuxfoundation/traefik#8 and rolled out to the production gateway on 2026-08-12.

With the root cause fixed, the padding is dead weight.

## Verification that the gateway fix is live

A production My CLAs load for a contributor with 3 verified emails and GitHub-only CLAs now reaches the CLA backend with the scalar param intact:

```
GET /v4/my-clas?email=<redacted>&email=<redacted>&email=<redacted>
  &githubId=<id>&githubId=<id>
  &githubUsername=<user>&githubUsername=<user>
  &lfUsername=<lfid>
```

`lfUsername` is the load-bearing signal: the stopgap never pads it, so under the old gateway behaviour it was always stripped alongside the GitHub keys. Its arrival proves the gateway is now forwarding single-valued params. The backend resolved both user records and returned the full CLA set.

## Scope

- Remove the padding from `identityQuery` in `apps/lfx-one/src/server/services/cla.service.ts` — drop `anyMulti` and `appendKeepAlive`, restore a plain `params.append` per value, and delete the `STOPGAP (lfx-gateway#114)` comment block.
- Remove the padding-specific test in `apps/lfx-one/src/server/services/cla.service.spec.ts` (`duplicates single-valued github keys when email is multi-valued…`).
- **Keep** the FR-011 invariant guard (`a multi-email caller reaches the gateway with its GitHub keys present`) and the single-email no-regression test. Both are phrased on the invariant rather than the mechanism, so they should pass unchanged.

## Acceptance criteria

- [ ] `identityQuery` emits exactly one value per resolved identity key.
- [ ] The FR-011 guard and the single-email no-regression test still pass without modification.
- [ ] `yarn test` is green.
- [ ] After deploy, a multi-email contributor with GitHub-only CLAs still sees the complete list in production.

## Notes

- Merging this before the gateway rollout reaches every environment would reintroduce #1314 there. Worth a per-environment sanity check first.
- linuxfoundation/lfx-gateway#114 is still open; it can be closed once this lands.

ahmedomosanya:
Closes #1413. Removes the interim BFF workaround from #1339 now that the durable gateway fix is deployed.

## Why

The platform API gateway's `awslambda` middleware dropped every single-valued query param whenever any param was repeated (linuxfoundation/lfx-gateway#114). Once a contributor had 2+ verified emails, `email` became multi-valued and `githubId` / `githubUsername` were silently stripped in transit, so GitHub-only CLAs disappeared from My CLAs (#1314). The BFF padded those keys to length 2 to survive the strip.

The root cause is now fixed (`linuxfoundation/traefik#8`) and deployed, so the padding is dead weight.

## What changed

- `identityQuery` in `apps/lfx-one/src/server/services/cla.service.ts`: dropped `anyMulti` and `appendKeepAlive`, restored a plain `params.append` per value, removed the `STOPGAP (lfx-gateway#114)` comment block. `lfUsername` still uses `params.set`; emission order is unchanged.
- Removed the padding-specific unit test and the now-stale comments describing the padding mechanism.
- **Tightened the FR-011 guard from `toContain` to `toEqual`.** As written it asserted `toContain('13434323')`, which is satisfied by a padded `['13434323','13434323']` just as well as by an unpadded `['13434323']` — so with the padding-specific test deleted, nothing in the suite would have failed if the padding were reintroduced. Verified by reintroducing the padding as a temporary mutant and confirming the guard now fails on it.

## Verification that the gateway fix is live

`/cla-service/v4/ops/health` is a public route that traverses the *same* `route-to-cla-service-lambda` middleware, so it can be probed unauthenticated. I sent a deliberately mixed query string — one single-valued param plus two multi-valued ones to force the failure condition — and read what the Lambda actually received:

```
GET /cla-service/v4/ops/health?probe=<tag>&probe=<tag>&scalar=SCALARSENTINEL&multi=1&multi=2
```

| Env | Query string as received by the CLA Lambda | Single-valued `scalar` |
|---|---|---|
| dev | `scalar=SCALARSENTINEL&multi=1&multi=2&probe=…&probe=…` | survived |
| staging | `scalar=SCALARSENTINEL&multi=1&multi=2&probe=…&probe=…` | survived |
| prod | `multi=1&multi=2&probe=…&probe=…&scalar=SCALARSENTINEL` | survived |

Under the old `valuesToMultiMap`, `scalar` would have been dropped in all three, because `multi` and `probe` made the multi-value map non-empty and the receiver then ignores the single-value map entirely. This satisfies the FR-010 removal gate.

Independently, a production My CLAs load for a contributor with 3 verified emails now reaches the backend with `lfUsername` intact alongside the repeated `email` params — meaningful because the stopgap never padded `lfUsername`, so its arrival can only be explained by the gateway forwarding single-valued params again.

> Note: the `lfx-gateway` deploy that carried this fix shows as failed overall, which is misleading — the only failing tests were `Security Service Health Check` (502 dev / 500 staging), with 17 passing. The Deploy Development, Deploy Staging, and Deploy Production jobs all succeeded, and the probe above independently confirms the middleware behaviour per environment.

## Test plan

- [x] `npx vitest run src/server/services/cla.service.spec.ts` — 37/37
- [x] `yarn test` (root) — 810/810
- [x] `./check-headers.sh`, `yarn format:check`, `yarn lint:check`, `yarn check-types`
- [x] `tsc --noEmit -p apps/lfx-one/tsconfig.json` (the documented `check-types` only covers `packages/shared`)

### DIFF
--- a/apps/lfx-one/src/server/services/cla.service.spec.ts
+++ b/apps/lfx-one/src/server/services/cla.service.spec.ts
@@ -303,7 +303,7 @@ describe('ClaService.getMyClas', () => {
     expect(calledUrl).toContain('githubUsername=octocat');
   });
 
-  it('duplicates single-valued github keys when email is multi-valued, to survive the gateway multi-value strip (lfx-gateway#114)', async () => {
+  it('FR-011 guard: a multi-email caller reaches the gateway with its GitHub keys present', async () => {
     getEffectiveUsername.mockReturnValue('alice');
     getEffectiveEmail.mockReturnValue('alice@x.org');
     getEffectiveSub.mockReturnValue('auth0|abc');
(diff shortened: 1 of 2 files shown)


MRashad26 (approved):
## Code Standards Review — PR #1416 (`refactor/GH-1413`)

**Verdict: APPROVE** — clean removal of the interim stopgap, no violations.

### ✅ All clean

- `identityQuery()` reverted to a plain `params.append` per value — matches the original pre-stopgap shape exactly; `lfUsername` still uses `params.set` as before ✓
- `STOPGAP (lfx-gateway#114)` comment block removed along with the dead code — no orphaned doc references ✓
- Stale padding-specific test deleted; **FR-011 guard correctly tightened from `toContain` to `toEqual`** — this closes a real coverage gap (the old assertion would have passed even if the padding were silently reintroduced) ✓
- Impersonation test (FR-009) updated consistently — comment `// target's key, padded, present` removed since padding no longer exists ✓
- No Angular templates, shared interfaces, or shared constants touched — server-side BFF only, so those rule categories don't apply here
- Root-cause verification in the PR description (gateway health-check probe across dev/staging/prod) is a solid gate before deleting the workaround

Nothing to flag.

PR #1436, opened 2026-08-13
feat(campaigns): create campaigns through campaign-service behind a flag

mrautela365:
**DRAFT — do not merge.** Opened so the full cutover shape is visible while #1409 is in review. Targets **#1409's branch**, not `main`, and will be rebased onto `main` once that lands.

Closes **LFXV2-3195** (UI cutover phase 3). LFXV2-2644 was closed as superseded by it — same work, sharper statement.

## Why this cannot be split

The job poll carried a module constant `LF_PROJECT_SLUG = 'tlf'`. Its own comment explained why that was survivable: `isCampaignServiceJobId` only routes UUIDs to campaign-service, and **no UUID job could exist until creation went through campaign-service**.

This PR is what makes UUID jobs real. So shipping creation without the poll change leaves a window where a CNCF user's poll is issued under TLF's scope — and `GetJob` joins `b.project_id` with an **exact** comparison, so it answers `not_found` for a job that is running. `not_found` is **terminal** for the poller, which surfaces to the user as a lost campaign.

Both halves are here. The constant is gone; the slug is a required parameter.

## The fall-through rule

The controller tries the cutover first and falls through to the legacy path **only when the flag is dark**. It does **not** fall through on an enabled-but-refused create.

That distinction is the one most likely to be "simplified" later, so it has its own test: the legacy path has real side effects on the ad platforms, so falling back after a refusal would create the campaigns while the user is told creation failed. Worse than any error message, because the spend is real and nobody is looking for it.

## What is deliberately NOT here

**Account selection.** Dispatch reads the ad account from the connection row, so one account per platform needs no per-campaign choice and no picker. Adding one would be inventing a decision the architecture does not ask for.

## Response shape

Stays `{jobId}`. The id is a UUID here and `job_...` there — which is exactly what lets the existing poll route send each one back to the system that owns it. That id-shape backstop came from the jobs flag (LFXV2-3070) and is **inherited rather than reinvented**, which is also what makes an overlapping rolling update safe.

## Flag

`LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE`, checked **after** `CampaignServiceBriefs` because the route is `/projects/{slug}/briefs/{brief_id}/campaigns` — there is no create-without-a-brief path. Turning this on while briefs are off yields no brief id and every create falls back, which is safe but silently pointless; the log line says so rather than failing.

[...]

PR #1438, opened 2026-08-13
fix(campaigns): keep hand-edited Google copy and budget across a tab switch

mrautela365:
## LFXV2-3229

LFXV2-3202 fixed this for the **Plan** tab. The **Implement** tab had the same defect and was explicitly out of that ticket's scope.

A user lands on Implement, hand-edits their RSA headlines and descriptions, clicks Insights to check pacing, comes back — and every edit is gone, replaced by the AI-generated original.

## Why it happened

`ImplementationTabComponent` holds its edits locally and lives inside the lazy `@switch`, so leaving the tab **destroys it**. On remount a constructor `effect` runs `populateFromBrief`, which calls `.clear()` on both copy `FormArray`s and repushes from `briefData` over whatever the user typed.

## Why the LFXV2-3202 fix was wrong here

That ticket keeps the planner **mounted**. Doing the same here would be a cost regression: this component resolves ad-account lists in `ngOnInit`, so an eager mount issues that request on **every page load** for a tab many users never open.

So the component stays cheap to destroy and the *edits* move to the parent instead — the same place `briefOutput` and `briefPersistence` already live.

## Scope of the draft

Deliberately a snapshot of **user-editable fields only**: copy, budget, flight, campaign types. Anything re-derivable from the brief (event name, slug, URL) or from a fetch (results, progress, account lists) is left to re-derive — restoring those would be restoring a cache, and a stale one.

`null` means "nothing to restore", which is **not** the same as an empty draft. An empty draft means the user cleared every field, and replaying that over a freshly generated brief would erase it.

## Two things worth a reviewer's attention

**The effect reads the draft `untracked`, and that is load-bearing rather than an optimisation.** Tracking it closes a loop: `valueChanges` emits → the parent signal updates → the input changes → the effect re-runs → it patches the form → `valueChanges` emits again. Angular catches that as **NG0103 (infinite change detection)** and takes the whole page down.

That is how it was found — the first version failed **14 existing tests**, none of them mine. Worth stating plainly, because the loop is invisible in review and only shows up when something else exercises the page.

**`searchBudgetPct` is now typed `number`.** `CAMPAIGN_BUDGET_DEFAULTS` is `as const`, so the control's inferred type was the literal `70` — wrong for a slider the user drags, and it made any other value a type error to patch in. The fix is at the form declaration rather than the shared constant, since other consumers legitimately want the literal.

## Tests

Two, both **revert-verified independently**:
- Removing the parent wiring fails the round-trip test.
- Removing the `eventSlug` guard fails the cross-event test.

They drive the **real DOM** — typing into the actual headline input, switching tabs, asserting the panel is genuinely gone, then coming back. The component's destruction *is* the mechanism under test, so a test that kept one child instance alive would prove nothing.

The cross-event guard is what makes the draft safe to hold un-keyed on the parent: a draft belongs to the brief it was typed against, and replaying event A's copy onto event B's freshly generated brief is the same class of bug the `(project, event)` ownership keys exist to prevent.

## Verification

License headers, prettier, lint (0 errors), **build**, and **256 app tests across 16 files**.

Note `test:server` does not cover this — its `include` is `src/server/**`. Component specs run under `ng test`.


apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
```diff
@@ -508,6 +575,92 @@ export class ImplementationTabComponent implements OnInit {
...
+        countryCode: draft.countryCode,
+        registrationUrl: draft.registrationUrl,
+        budgetUsd: draft.budgetUsd,
+        searchBudgetPct: draft.searchBudgetPct,
+        startDate: draft.startDate,
+        endDate: draft.endDate,
+        includeSearch: draft.includeSearch,
+        includeDemandGen: draft.includeDemandGen,
+      },
+      { emitEvent: false }
```
dealako: [minor] Draft restore leaves `toSignal`-derived displays (budget-split label, campaign-name preview) stale

Issue: `applyDraft()` patches `searchBudgetPct`, `eventName`, `countryCode`, `includeSearch`, `includeDemandGen` (and the copy arrays) with `{ emitEvent: false }`. But two display values are derived from `valueChanges` via `toSignal` — `displayBudgetPct` (`initDisplayBudgetPct`, ~line 783, subscribes to `searchBudgetPct.valueChanges`) and `campaignName` (`initCampaignName`, ~line 793, subscribes to `campaignForm.valueChanges`). Suppressing emission means those signals never recompute for the restored values.

Proof: On remount the effect sets `seeding = true`, `populateFromBrief` patches with the default `emitEvent: true` so `displayBudgetPct`/`campaignName` update to the **brief** values; then `untracked(applyDraft)` patches `searchBudgetPct` = draft value with `emitEvent: false`, so `valueChanges` doesn't fire and the signals stay at the brief values. Meanwhile the range control's thumb tracks the restored value directly via `formControlName="searchBudgetPct"` (`implementation-tab.component.html:136`). Result: the label at `:135` renders `Search {{100 - displayBudgetPct()}}% / Display {{displayBudgetPct()}}%` from the stale brief split while the slider thumb sits at the edited split. Same for the `campaignName()` preview box (`:62–65`) when `eventName` was edited. `getRawValue()` in `submit()`/`emitDraft()` reads the correct control value, so only these two *displays* are wrong.

Why it matters: This is the exact feature being shipped — a user who dragged the budget split sees the split label appear reverted after a tab round-trip even though the underlying value and the submit payload are correct. The feature looks broken precisely where it works.

Fix: Pass `emitEvent: true` on the `applyDraft` `patchValue`. It's safe against the re-entrancy the `emitEvent: false` comment guards: `applyDraft` only ever runs inside the seeding window (between the `seeding = true` and `seeding = false` lines), and the `valueChanges` subscriber returns early while `seeding === true`, so `emitDraft` cannot re-enter. Emitting lets `displayBudgetPct`/`campaignName` recompute for the restored values without a self-write loop. The `replaceCopyArray` copy arrays can stay suppressed — they feed neither derived signal.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -1257,3 +1258,197 @@ describe('CampaignsComponent — email delivery channel', () => {
...
+    // Regression: `replaceCopyArray` originally suppressed emission for BOTH callers, so the
+    // brief seed never reached the parent and the draft held the form's empty starting control.
+    // Switching away and back then restored `[""]` over real generated headlines — the fix
+    // making the bug worse than the bug. Emission is now the caller's decision.
+    internals().onProceedToImplementation(briefFor('kubecon-eu-2026'));
+    internals().selectTab('implementation', 'paid-marketing');
+    fixture.detectChanges();
+    await fixture.whenStable();
+
+    expect(internals().implementationDraft()?.headlines).toEqual(['Generated A', 'Generated B']);
```
dealako, line 1628: [minor] Round-trip tests miss the derived-display field class where the `emitEvent: false` restore regresses

Issue: The new restore tests assert only on `implementation-headline-0` `.value` and `implementation-reg-url` `.value` — both native inputs whose DOM reflects the control value directly regardless of emission. None edits the budget-split slider (`searchBudgetPct`) or `eventName` and asserts the derived `displayBudgetPct` label / `campaignName()` preview after a round-trip, which is exactly the field class the `emitEvent: false` restore leaves stale (see the `applyDraft` comment).

Proof: Every restore assertion in this suite checks a native input `.value`; there's no reference to `implementation-budget-split` or the rendered `Budget Split: Search X% / Display Y%` label after a round-trip. Because a range control's DOM tracks the control value even with `emitEvent: false`, a slider round-trip that asserted only the value would pass — but asserting the rendered percentage label would catch the stale-display bug.

Why it matters: The suite claims to prove edits survive, but it can't detect the one preserved-field category (`toSignal`-derived displays) that currently regresses, so the bug ships green.

Fix: Add a restore test that sets the budget-split range input to a non-default `searchBudgetPct`, dispatches `input`, round-trips to another tab and back, and asserts both the control value AND the rendered `Budget Split: Search X% / Display Y%` label reflect the edited split.

dealako (dismissed):
Hi @mrautela365 👋 — thanks for a genuinely careful fix, and for a PR description and commit trail that do the reviewer's job for them.

**Overall impression:** This is a well-reasoned, tightly-scoped fix. Lifting the editable-fields draft to the parent (rather than eager-mounting the way LFXV2-3202 did for the planner) is the right trade-off given `ngOnInit` resolves the LinkedIn ad-account list, and you've documented the asymmetry instead of hiding it. The load-bearing subtleties — `untracked` reads, the `seeding` flag disambiguating brief-seed from user-edit on `valueChanges`, and the caller-owned `emitEvent` decision in `replaceCopyArray` — are exactly the traps this pattern sets, and each is explained where it lives. The three commits also show real review-driven iteration: the bots' findings were all addressed and revert-verified.

**Bot reconciliation:** The Copilot and Cursor Bugbot findings (stale-draft-before-arrays-populated, remount desync dropping edits on a second tab leave, and the `eventSlug` guard not covering re-proceed for the same event) were all real and are **resolved** by commits 2–3 — the `seeding` flag and the `implementationDraft.set(null)` in `onProceedToImplementation`, each with a revert-verified test. Copilot's final pass generated no new comments. Nothing outstanding from the bots.

**One class of field the diff and the tests both miss:** the two `toSignal`-derived displays (`displayBudgetPct` label + `campaignName()` preview) don't recompute on restore because `applyDraft` patches with `emitEvent: false`. The slider *thumb* moves to the restored split (via `formControlName`) but the *label* stays at the brief split — a visible mismatch in the exact feature you're shipping. Details + a safe one-line fix inline, plus the matching test gap.

**Issue count**
- 🔴 Blocking: 0
- 🟡 Minor: 2 — (1) `applyDraft` `emitEvent: false` leaves budget-split label & campaign-name preview stale after restore; (2) round-trip tests don't cover the derived-display field class, so that regression ships green.
- ⚪ Nit: 2 — (1) `CampaignImplementationDraft` import breaks the block's alphabetical order in `campaigns.component.ts` (and is inconsistent with the correctly-placed sibling in `implementation-tab.component.ts`); (2) the `campaign.interface.ts` "Budget and flight" docblock folds the `includeSearch`/`includeDemandGen` campaign-type toggles under budget/flight — reword to include "Google campaign-type toggles."
- ❔ Question: 0

**Security & data-privacy:** clean. This is purely in-memory, same-user, same-session scratch state — no storage/log/URL sink, no new trust-boundary crossing, no fixture PII.

**Final decision: ✅ Approved with minor comments.** Nothing here blocks merge. I'd suggest folding the `emitEvent: true` fix + the slider round-trip test into this PR since they're the shipped feature's own surface, but they're safe to defer to LFXV2-3230 alongside the already-tracked per-platform-signal gap if you'd rather keep this one focused. The two nits are take-it-or-leave-it.

mrautela365:
@dealako — rebased onto main (the cutover #1436 landed) and re-requesting. **This push dismissed your approval**, hence back in your queue.

Three conflicts, none of them a "pick a side":

- **`onProceedToImplementation`** — HEAD's body (the draft clear is this branch's point) with **main's signature**, which carries `restoredApproved` from #1436. Taking the two-parameter form would have compiled and then silently stopped threading approval into the restore.
- **The `lfx-implementation-tab` tag** — both sides add *different* inputs, so the union: `briefSaveInFlight` from main alongside `draft`/`draftChange` from here, keeping both comments since each explains an input the other doesn't.
- Two unrelated class members in the same slot — kept both; the conflict split mid-docblock, so the `/**` opener needed restoring by hand.

**One near-miss worth flagging:** the tag conflict got staged with its markers intact by `git add -A`, and **neither `tsc` nor the test run caught it** — `yarn lint` did, as *"Opening tag not terminated"*. I now sweep the worktree for markers before committing a merge rather than trusting the merge's exit code.

Your three open threads are on the pre-merge HEAD; I'll work through them next unless you'd rather they land in a follow-up now that this is approved-and-rebased.

`1197 + 279` passing, tsc clean, lint 0 errors, up to date with main.

mrautela365:
@dealako — your approval was dismissed by the rebase onto the merged cutover. Re-requesting; no new findings since.

Up to date with main, `MERGEABLE`, tsc clean on both tsconfigs, lint 0 errors.

Your three open threads are still outstanding — the `emitEvent: false` restore leaving `toSignal`-derived displays stale, the round-trip tests missing that field class, and the `eventSlug`-uniqueness point. They're genuine and I have not addressed them yet; I've been working the merge order. Tell me if you'd rather they land here before approval or as a follow-up, and I'll do that next either way.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -514,6 +532,15 @@ export class CampaignsComponent {
    * that quietly rewrites the thing it restored is the one outcome this path must not have.
    */
   protected onProceedToImplementation(brief: CampaignBriefOutput, alreadyPersisted = false, restoredApproved = false): void {
+    // BEFORE `briefOutput`, not after. Setting the brief first lets the child's effect run with
```
dealako, line 535: [minor] The merge made `CampaignBriefPersistenceState.approved`'s docblock inaccurate

Issue: This conflict resolution now sets `approved: restoredApproved` on the **`off`** branch (with a non-null `briefId`) when a saved brief is restored. But the field's docblock in `packages/shared/src/interfaces/campaign.interface.ts` still reads: *"Meaningful only on `saved`; `false` elsewhere, where there is no stored brief to approve."* All three clauses are now false for the restore path.

Proof: Trace the restore flow — `onRestoreSavedBrief(brief, briefId, approved)` → `onProceedToImplementation(brief, true, approved)` → this branch sets `{ status: 'off', briefId: <restored id>, message: null, approved: restoredApproved }`. A restored **approved** brief therefore lands in `status: 'off'` with `approved: true` and a non-null `briefId`. And the Implementation tab reads it exactly there: `implementation-tab.component.ts` `canCreate` returns `false` when `status === 'off' && briefId !== null && !approved`. So `approved` is now load-bearing on `off`, is not always `false` off `saved`, and there *is* a stored brief to approve in that state.

Why it matters: The nicely-precise incident-memory docblocks in this file are trusted. A future maintainer reading "meaningful only on `saved`" could drop the `approved` read on the `off` branch and silently re-enable Create for an unapproved restored brief — the exact terminal-refusal defect (`brief.go:439`) this guard exists to close. This is the one spot the merge changed behavior without updating its own documentation.

Fix: Update the docblock to reflect the restore case, e.g. *"Meaningful on `saved`, and on a restored brief carried in `off` (non-null `briefId`); `false` only where no stored brief exists."*

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -1489,3 +1490,197 @@ describe('CampaignsComponent — email delivery channel', () => {
...
+      budgetUsd: 999,
+      searchBudgetPct: 10,
+      startDate: '2026-01-01',
+      endDate: '2026-02-01',
+      includeSearch: true,
+      includeDemandGen: false,
+      eventSlug: 'event-a',
+    });
+
+    internals().onProceedToImplementation(briefFor('event-b'));
```
dealako: [nit] This cross-event guard test passes even with the `eventSlug` guard removed

Issue: The test seeds a draft with `eventSlug: 'event-a'`, then calls `onProceedToImplementation(briefFor('event-b'))` before the child mounts. But `onProceedToImplementation` unconditionally runs `implementationDraft.set(null)` first, so the child's `draft` input is already `null` at mount — `applyDraft()` returns at its `if (!draft) return;` line and never reaches the `if (draft.eventSlug !== currentSlug) return;` branch this test is meant to protect.

Proof: Delete the `eventSlug` mismatch check in `implementation-tab.component.ts` and this test still passes (headline stays `Generated A`) because there is no draft to replay through this path. Copilot flagged the same thing (suppressed comment) — agreed, it's correct. Pre-existing (the draft-clear predates this branch's changes), not introduced by the merge.

Why it matters: The `eventSlug` guard is named as the safety net that makes it sound to hold the draft un-keyed on the parent, but it currently has no test that actually exercises the mismatch branch — a regression removing it would ship green.

Fix: Install a mismatched-slug draft on a path that does **not** route through `onProceedToImplementation` (e.g. set `implementationDraft` after the brief is on screen, via a plain Implement→Insights tab switch), mount the child, and assert the brief's own copy stands — so the assertion fails if the `eventSlug` guard is removed.

dealako (dismissed):
Follow-up review — @mrautela365, thanks for the exceptionally clear rebase writeup. The commit trail and your conflict notes made this fast to verify.

👏 **Nice work on the merge.** I traced all three conflict resolutions and they're correct:

- **`onProceedToImplementation`** — keeping HEAD's body (the `implementationDraft.set(null)` clear runs *first*, before `briefOutput.set` and the `alreadyPersisted` branch) while adopting main's 3-param signature is the right call. `restoredApproved` is threaded end-to-end: `restoreSavedBriefRequested` (`approved: savedBriefApproved`, only `true` for a genuinely loaded+approved brief) → `onRestoreSavedBrief` → `onProceedToImplementation(brief, true, approved)` → the `off`-state `approved`, which `canCreate` reads to fail **secure** (param defaults to `false`, so a missing value blocks Create, never opens it). You were right that the 2-param form would have compiled and silently dropped approval threading.
- **`briefSaveInFlight` union** — counter is balanced: one increment per `persistBrief`, the clear chained after the terminal `.catch()` so it runs on success/refusal/throw, and `Math.max(0, …)` guards underflow. It's a client-side UX gate over the authoritative server refusal — adds no new auth boundary.
- **Class-member/docblock collision** — `briefSaveInFlight` / `pendingBriefSaves` / `implementationDraft` and their docblocks are intact after the hand-restore.

Security and correctness are both clean on the merge — no new behavioral or authz regressions. (Good instinct adding a pre-commit conflict-marker sweep, too — `tsc` and tests genuinely won't catch a marker inside a template tag; lint did.)

**Re the further rebases** (LFXV2-3202 / #1437 landed in main): I also reviewed the planning-mount adaptation that came with it — the paid and email planning panels are now hoisted out of the `@switch` and visibility-toggled, a faithful carry-down of #1437's already-landed pattern. It correctly preserves this PR's premise: Implementation stays in the lazy `@switch` (still destroyed on tab switch), so the parent-owned draft remains the mechanism. The added mount/lazy-panel tests are supporting, and none of this PR's own draft round-trip tests were weakened. No new defects there.

**Revision tracking (my prior threads):** all still open — you already flagged you'd take these next, which is fine:
- ❌ `applyDraft` `emitEvent: false` leaves the `displayBudgetPct` label + `campaignName()` preview stale after a restore (slider thumb moves, label text doesn't). *Copilot independently surfaced this exact issue this round (suppressed comment on `implementation-tab.component.ts:680`) — so we both see it.*
- ❌ Round-trip tests still only assert native-input `.value`, never the derived-display label — so the above ships green.
- ❌ Two nits unchanged: `CampaignImplementationDraft` import ordering; the `campaign.interface.ts` "Budget and flight" docblock folding the `includeSearch`/`includeDemandGen` campaign-type toggles under budget/flight.

**New this round (from the merge scope):**
- 🟡 The conflict resolution set `approved` on the `off` state, which makes `CampaignBriefPersistenceState.approved`'s docblock ("meaningful only on `saved`; false elsewhere") inaccurate — a security-adjacent doc footgun. Inline.
- ⚪ The cross-event `eventSlug` guard test is a tautology (draft is cleared before mount, so the guard branch is never reached) — matches a Copilot suppressed comment. Inline.
- ⚪ Pre-existing (not merge-introduced, so flagging only): `campaigns.component.ts:575` has an orphaned one-line docblock — `/** A brief restored from campaign-service… */` sits above `onEmailProceedToImplementation` but describes `onRestoreSavedBrief`, which is left undocumented. Worth a look while you're in this region given the hand-restored docblocks.

**Bot reconciliation:** Copilot's two new passes generated no new blocking comments; their suppressed comments corroborate the stale-display issue and the guard-test tautology. Nothing outstanding from the bots.

**Issue count (all open, carried + new)**
- 🔴 Blocking: 0
- 🟡 Minor: 2 — stale derived-display on restore (+ its missing test); `approved` docblock now inaccurate after the merge.
- ⚪ Nit: 4 — import ordering; "Budget and flight" toggle grouping; eventSlug-guard test tautology; orphaned docblock (pre-existing).

**Final decision: ✅ Approved with minor comments.** The rebase is correct and introduced no functional or security regressions — nothing here blocks merge. The `approved`-docblock fix is the one I'd genuinely land in this PR (it's the merge's own doc drift); the stale-display defect + test and the nits are fine to carry into LFXV2-3230 alongside the per-platform-signal gap you're already tracking, or fold in here — your call.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
mrautela365: Fixed in `6fd9157aa`, and your proof was exactly right.

`displayBudgetPct` and `campaignName` both derive from `valueChanges` through `toSignal`, so `{ emitEvent: false }` restored the CONTROLS and left those stale — the slider thumb at the draft value, the label beside it still showing the brief's number. Half a restore that looks whole.

The fix is to let the patch emit. Suppressing emission was never what stopped the draft being written back over itself — **`seeding` is**, and it is already true for the whole of `applyDraft` (the `valueChanges` subscription returns early while it is set). So emitting reaches the derived signals without reopening that loop.

The copy arrays stay silent deliberately: nothing derives a display from them, and rebuilding a FormArray emits per control, so letting them through would recompute `campaignName` once per headline for no gain.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Added in `6fd9157aa` — you were right that every existing restore assertion reads a native input's `.value`, which tracks its control regardless of emission, so none of them could catch this.

The new test drags the budget-split slider to 30%, round-trips through another tab, and asserts the **rendered label** (`Search 30% / Display 70%`) as well as the control value. Verified by restoring `emitEvent: false`: the control assertion still passes and the label one fails, which is precisely the asymmetry you described.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Confirmed and fixed in `6fd9157aa` — the test passed with the guard deleted, exactly as you said.

`onProceedToImplementation` runs `implementationDraft.set(null)` first, so seeding the draft *before* that call left the child mounting with `draft === null`. `applyDraft` returned at its `if (!draft)` line and never reached the `eventSlug` branch the test was named for.

Reordered: proceed first, then seed the stale draft, so the child mounts with a draft whose slug genuinely disagrees. Deleting the guard now fails it.

You were also right that it's pre-existing rather than introduced by the merge — the draft clear predates this branch.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Correct, and it was my merge that falsified it. Fixed in `6fd9157aa`.

All three clauses of *"Meaningful only on `saved`; `false` elsewhere, where there is no stored brief to approve"* became untrue once restore began carrying the stored brief's approval through to the `off` branch. The docblock now names both load-bearing states explicitly — `saved`, and `off` **with a briefId** — and says that `off` with a null briefId is the genuinely-not-applicable case where `false` means "no opinion" rather than "unapproved", which is how `canSubmit` reads it.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -79,6 +80,23 @@ export class CampaignsComponent {
...
+   * `ImplementationTabComponent` stays inside the lazy `@switch` — it resolves ad-account lists in
+   * `ngOnInit`, so mounting it eagerly the way LFXV2-3202 mounts the planner would issue that
+   * request on every page load for a tab many users never open. Holding its edits up here is the
+   * cheaper half of that trade: the component is still destroyed, but the user's typing is not.
+   *
+   * NOT keyed by `(project, event)` like `knownBriefIds`, and the asymmetry is deliberate. A brief
+   * id names a row that outlives the page, so replaying the wrong one corrupts durable state. A
+   * draft is scratch: the worst a stale one can do is show the wrong copy, and the child already
+   * refuses to apply a draft whose `eventSlug` does not match the brief on screen. A key here
+   * would be machinery guarding against a hazard that is already closed one level down.
```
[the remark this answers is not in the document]
mrautela365: Addressed in `6fd9157aa`, though the answer is that the invalidation you asked for already exists.

`onProceedToImplementation` runs an unconditional `implementationDraft.set(null)` **before** anything consults the slug — so a re-proceed for the same event does discard the old draft, and `eventSlug` is only the second line of defence for a genuinely different event.

What was actually wrong is that the test meant to protect that second line never reached it (it passed with the guard deleted). That is now reordered and revert-verified.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
```diff
@@ -278,7 +304,25 @@ export class ImplementationTabComponent implements OnInit {
...
+      // NG0103 (infinite change detection) and takes the whole page down with it, which is how
+      // this was found.
+      untracked(() => this.applyDraft());
     });
+
+    // Emit as the user types. `valueChanges` covers the form (copy, budget, flight, campaign
+    // types); it does NOT cover the platform signals, which is deliberate — see the draft
+    // interface for why this snapshot is scoped to the fields a user types rather than everything
+    // the component holds.
+    this.campaignForm.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => this.emitDraft());
```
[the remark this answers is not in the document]
mrautela365: The `seeding` flag already covers this: it is set before `populateFromBrief` and cleared after `applyDraft`, and the `valueChanges` subscription returns early for the whole of that window — so the seed's own patch cannot store a blank draft.

What WAS broken is the other half of the same area, which @dealako caught: `applyDraft` suppressed emission entirely, so the derived displays (`displayBudgetPct`, `campaignName`) never recomputed for restored values. Fixed in `6fd9157aa` by letting the patch emit — safe precisely because `seeding` is what guards the write-back, not `emitEvent: false`.

mrautela365:
@dealako — all your findings plus Copilot's two are in `6fd9157aa`. All threads resolved.

**The derived-display bug was real and your proof was exact.** `displayBudgetPct` and `campaignName` both come from `valueChanges` via `toSignal`, so `{ emitEvent: false }` restored the controls and left those stale — slider thumb at the draft value, label beside it still showing the brief's number.

The fix is to let the patch emit. Suppressing emission was **never** what stopped the draft being written back over itself — `seeding` is, and it's already true for the whole of `applyDraft`. That was the actual misunderstanding in my original code: I'd used two mechanisms where one does the job, and the redundant one had a side effect.

**The `eventSlug` test passed with the guard deleted**, exactly as you said. `onProceedToImplementation` clears the draft first, so seeding it beforehand meant the child mounted with `draft === null` and `applyDraft` returned before reaching the mismatch branch. Reordered; deleting the guard now fails it.

That also disposes of Copilot's "eventSlug does not uniquely identify the brief" — the unconditional `implementationDraft.set(null)` **is** the invalidation it asked for, and it already runs first. The slug is only the second line of defence.

**The `approved` docblock** — my merge falsified it and you caught it. All three clauses were wrong once restore began carrying approval on the `off` branch. It now names both load-bearing states and explains that `off` with a null briefId means "no opinion" rather than "unapproved".

Both fixes are revert-verified: restoring `emitEvent: false` fails the new label test; deleting the slug guard fails the reordered one.

`1211 + 289` passing · tsc clean on both tsconfigs · lint 0 errors · 0 open threads · up to date with main.

dealako (approved):
Follow-up review — @mrautela365, this is a clean sweep. You closed **every** open item from my last two rounds plus Copilot's, in one tightly-argued commit (`6fd9157`), and revert-verified both fixes. Nice work.

👏 **What landed, and why each is right:**

- **Stale derived displays (my finding, Copilot's `implementation-tab.component.ts:680`) — ✅ resolved.** `applyDraft` now emits, and I traced the safety argument end-to-end: `seeding` is set for the entire effect window (before `populateFromBrief`, cleared after `applyDraft`), and the `valueChanges`→`emitDraft` subscription returns early while it's set — so the patch reaches `displayBudgetPct`/`campaignName` (the `toSignal(valueChanges)` derivations) **without** reopening the write-back loop. The effect reads the draft `untracked`, so the emission can't re-trigger it (no NG0103), and the single post-seed `emitDraft()` still captures the restored state for the parent. Your framing — "`emitEvent:false` was never what stopped the write-back; `seeding` is" — is exactly correct. Copy arrays correctly stay silent (nothing derives from them; rebuild emits per-control).

- **The missing test class — ✅ resolved.** The new `restores the budget-split LABEL, not just the slider value` test asserts the *rendered* label after a round-trip, which is the one thing a native-input `.value` assertion couldn't catch. Revert (`emitEvent:false`) fails it. That's the seam.

- **`eventSlug` guard tautology (my finding + Copilot) — ✅ resolved.** Reordering so `onProceedToImplementation` (which clears the draft) runs *before* seeding the stale-slug draft means the child now mounts with `draft.eventSlug` genuinely disagreeing with the current slug — deleting the guard fails the test. And you're right that the unconditional `implementationDraft.set(null)` *is* the invalidation Copilot's "eventSlug doesn't uniquely identify the brief" asked for.

- **`approved` docblock (my finding) — ✅ resolved.** Rewritten accurately: load-bearing on `saved` and on `off`-with-a-briefId (the restore state), `false`-as-no-opinion on `off` with a null briefId — matching how `canSubmit` reads it. Good.

**Reconciling Copilot's newest pass (20:17)** — three documentation-consistency observations, all real, all low-severity; I agree with them and they're worth a quick cleanup pass (here or a fast follow-up):

- ⚪ **Stale merge-state wording.** `LFXV2-3202 (PR #1437, pending)` / `not yet merged` still appears in `campaigns.component.ts:114`, `implementation-tab.component.ts:105`, `campaign.interface.ts:289`, and `campaigns.component.spec.ts:1498` — but #1437 *has* merged (its planning-mount is why your later commits adapted the email side). Merge-state annotations in long-lived comments go stale the moment they land; drop the parenthetical or restate as a plain dependency.
- ⚪ **Contradictory template comments.** `campaigns.component.html:254–266` (carried in from #1437's planning-mount block) still says Implementation edits "do not survive" and frames lifting them as a *hypothetical future* fix — while `:290–294` documents that this PR (LFXV2-3229) already lifted them via `draft`/`draftChange`. The template now describes both the old and new behavior; the earlier paragraph should be reconciled to past tense.
- ⚪ **Scope wording.** The `CampaignImplementationDraft` contract persists `eventName`/`countryCode`/`registrationUrl` (with a test on URL restoration), which your PR body already explains — but the interface's own scope comment could state that plainly so the contract is self-describing.

**Carried nits still open** (unchanged, take-it-or-leave-it): `CampaignImplementationDraft` import ordering in `campaigns.component.ts`; the `campaign.interface.ts` "Budget and flight" docblock folding the `includeSearch`/`includeDemandGen` campaign-type toggles under budget/flight; and the pre-existing orphaned docblock on `onEmailProceedToImplementation`.

**Issue count (all open)**
- 🔴 Blocking: 0
- 🟡 Minor: 0 — every prior minor is resolved and verified.
- ⚪ Nit: 6 — stale `#1437` merge-state wording (×4 files); contradictory html comments; interface scope wording; import ordering; "Budget and flight" toggle grouping; orphaned docblock. All documentation-only.

**Final decision: ✅ Approved.** No blocking issues, no security findings, no merge conflicts, zero open minors — the behavioral fixes are correct and revert-verified. The remaining items are all documentation nits; the two I'd genuinely fold in before merge are the stale `#1437, pending` wording and the contradictory template comments, since both actively describe a state that's no longer true — but neither blocks. Thanks for the exceptionally rigorous iteration on this one.

PR #1439, opened 2026-08-13
feat(campaigns): wire the hubspot template search for the email channel

mrautela365:
## LFXV2-3198, first slice

The Email channel cannot stage anything today. `hubspotConfig.sourceEmailId` is **required with no default** — staging clones a HubSpot template — and nothing in the UI can choose one. The Implement panel's own placeholder says exactly this: *"choosing that template needs an endpoint that is still in review."*

That endpoint shipped as **LFXV2-3197** (campaign-service #122). This PR wires it end to end so the picker becomes buildable: shared interfaces → campaign-service client → Express proxy route → Angular service method.

## Why a search, not a dropdown

campaign-service pages nothing here and caps an unfiltered listing. A portal with hundreds of marketing emails would render a **truncated list that looks complete** — the one shape of falsehood a picker must not have. `q` does **not** reach HubSpot. Its list endpoint cannot be queried by name or subject, so campaign-service walks the pages and matches name-or-subject in-process. Reading this as server-side search is the misreading the service's own design warns about — it invites optimising the walk away and reintroduces the false absence the cap exists to prevent.

The filtered walk is exempt from the 500-row cap but **not** unbounded: it is capped at `maxListPages = 200` and returns an error on exhausting them rather than a partial list. Complete-or-error, which is why `possiblyTruncated` is always false for a filtered search.

An empty query is meaningful and is *not* sent as `q=""`: absent means "list the most recently updated", which is the useful default before a user knows what they're looking for. The service already orders by last-modified.

## The enabled/error distinction

Three outcomes, and conflating any two would mislead:

| Situation | Result | Why |
|---|---|---|
| No HubSpot connection | `enabled: false` | The steady state everywhere the channel isn't set up. Renders "connect HubSpot", not an error. |
| Connected, read failed | `enabled: true` + `error` | HubSpot **is** connected. Reporting `false` would tell the user to connect something they already connected. |
| Connected, read fine | `enabled: true`, list | — |

This matches `saveBrief` and `createCampaigns`, which already treat a dark/absent state as a first-class non-failure.

## The tests worth reading

Six, and the **argument-position** ones are the point. `proxyRequestWithResponse(req, service, path, method, query, data)` takes `query` fifth and `data` sixth — **both typed `any`**. Passing the query sixth compiles cleanly and sends it as a request body, which a GET discards. The result is a search that silently ignores what the user typed and returns the unfiltered list.

No type checker catches that. **Revert-verified**: moving the argument one position fails that test specifically, and nothing else.

This is the same trap that shipped a bodyless create on #1436, which is why it gets an assertion here rather than a comment.

## Not in this PR

The picker component itself, the staging form, the audience UI, and the Monitor tab. Those are the rest of LFXV2-3198 plus LFXV2-3224 and LFXV2-3223 — and **LFXV2-3224 must land before staging works at all**, because `HubSpotDispatcher.Dispatch` resolves a *built* audience before it will clone, and fails pre-create without one.

## Verification

License headers, prettier, lint (0 errors), build, **1099 server tests across 52 files**.


apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -449,6 +462,81 @@ export class CampaignServiceClient {
...
+   * campaign-service walks every page and matches in-process — which is precisely why a filtered
+   * search is unbounded while the empty-query screen is capped. Do not describe this as
+   * server-side search: the service's own design warns that reading it that way invites
+   * optimising the walk away, reintroducing the false absence the cap exists to prevent.
+   *
+   * `enabled: false` for a project with no usable HubSpot connection, matching `saveBrief` and
+   * `createCampaigns`: an absent connection is the steady state everywhere the channel is not set
+   * up, so it must not surface as an error. The caller renders "connect HubSpot" for it.
+   */
+  public async searchHubSpotEmails(req: Request, projectSlug: string, query: string): Promise<HubSpotEmailSearchResult> {
```
dealako, line 727: [minor] Orphaned JSDoc on `reconcileLostCreate`

Issue: Inserting `searchHubSpotEmails` here stacked its JSDoc directly beneath the pre-existing `/** After an ambiguous create failure... */` block that documents `reconcileLostCreate`. That method now sits further down with no adjacent JSDoc.
Proof: Lines 446-464 hold the "After an ambiguous create failure" block (the `reconcileLostCreate` doc); lines 465-486 hold the new `searchHubSpotEmails` doc; `private async reconcileLostCreate(` now starts at ~line 540 with only inline comments. TSDoc/editors bind only the adjacent block, so the rich `version === 1` / payload-match rationale no longer documents its method.
Why it matters: The reconciliation semantics doc is now dead text above an unrelated method, and readers/tooling are misled.
Fix: Move the "After an ambiguous create failure..." block back down so it immediately precedes `private async reconcileLostCreate(`, leaving only the `searchHubSpotEmails` doc above this method.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -449,6 +462,81 @@ export class CampaignServiceClient {
...
+        'GET',
+        query === '' ? undefined : { q: query }
+      );
+      // Rows without an id are DROPPED, not mapped to `id: ''`. This is the value the staging
+      // config's required `sourceEmailId` takes, so an id-less row is a choice the user cannot
+      // make — rendering it would offer a template that fails on submit.
+      const emails = (response.data?.emails ?? []).filter((email) => typeof email.id === 'string' && email.id !== '').map(fromMarketingEmail);
+      // Derived here because the wire cannot express it: a capped 500 and a complete 500 are the
+      // same bytes. Only an EMPTY query is capped, so a filtered search is never flagged.
+      return { enabled: true, emails, error: null, possiblyTruncated: query === '' && emails.length >= UNFILTERED_EMAIL_CAP };
```
dealako: [minor] `possiblyTruncated` derived from the post-filter count can under-report a truncated first screen

Issue: The flag is computed from `emails.length` *after* id-less rows are dropped, not from the wire count.
Proof: `const emails = (response.data?.emails ?? []).filter(id present).map(...)` then `possiblyTruncated: query === "" && emails.length >= UNFILTERED_EMAIL_CAP`. If campaign-service returns exactly 500 (cap hit, genuinely truncated) and one row lacks an id, post-filter length is 499, so `499 >= 500` is false and a truncated listing is reported complete — the precise falsehood this flag exists to prevent.
Why it matters: The interface documents the flag so a picker "cannot present it as the whole portal"; a single dropped row silently defeats that guarantee.
Fix: Derive from the raw wire length before filtering, e.g. `const wireCount = (response.data?.emails ?? []).length;` and `possiblyTruncated: query === "" && wireCount >= UNFILTERED_EMAIL_CAP`. (Defensive — interface says `id` is never empty — but strictly stronger.)

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -516,6 +516,42 @@ export class CampaignController {
...
+   * Search the project's HubSpot marketing emails for the template picker.
+   *
+   * `?project=` is required rather than defaulted, for the same reason every other
+   * campaign-service read here requires it: a HubSpot connection is per-project, and guessing the
+   * project would list one foundation's templates to another.
+   *
+   * `?q=` is optional — an empty query lists the most recently updated templates, which is the
+   * useful default when a user does not yet know what they are looking for.
+   */
+  public async searchHubSpotEmails(req: Request, res: Response, next: NextFunction): Promise<void> {
```
dealako, line 631: [minor] Controller `searchHubSpotEmails` has no boundary test

Issue: The new controller method adds real logic — the `project === ""` → `ServiceValidationError` refusal, the `q` trim, and error forwarding to `next` — but nothing in `campaign.controller.spec.ts` exercises it. The service layer is well covered, but the controller branch that decides whether the service is called at all (and whether a 400 is raised) is unverified.
Proof: No `searchHubSpotEmails` matches in `campaign.controller.spec.ts`, while the analogous `loadBrief` missing-project branch is tested (`buildLoadReq({ project: "" })` → `expect(loadBrief).not.toHaveBeenCalled()`). The harness and pattern already exist.
Why it matters: A regression dropping the empty-project guard or the trim would ship silently.
Fix: Mirror the `loadBrief` cases: (a) blank `project` → `next` called with `ServiceValidationError`, service not invoked; (b) happy path → service called with trimmed slug/query and `res.json(result)`; (c) service rejection forwarded to `next`.

packages/shared/src/interfaces/campaign.interface.ts
```diff
@@ -1022,6 +1022,79 @@ export interface RedditMonitorResponse {
 // HubSpot UTM
 // ---------------------------------------------------------------------------
 
+/**
```
dealako, line 1066: [nit] New email-search interfaces sit under the `// HubSpot UTM` banner

Issue: `HubSpotMarketingEmail` and `HubSpotEmailSearchResult` are inserted immediately beneath the `// HubSpot UTM` section header, but they describe email/template search, not UTM lookup/create.
Proof: Line 1021 is `// HubSpot UTM`; the two new interfaces are added right after it, before `HubSpotUtmLookupResult`.
Why it matters: The section banner (which exists for file navigability) now mislabels its contents.
Fix: Add a dedicated banner such as `// HubSpot Email Templates` above the two new interfaces.

dealako (changes requested):
Hey @mrautela365 — thoughtful slice, and the PR description is genuinely excellent: the enabled/error/`enabled:false` trichotomy, the "search not dropdown" rationale, and the argument-position test are all well-motivated and match the existing `saveBrief`/`createCampaigns`/`findBrief` patterns. Security and data-privacy came back clean — user bearer token forwarded for per-project authz, `projectSlug` encoded into the path, `q` as a structured query param, no PII in logs (only `projectSlug`/counts), and failures fail secure. Nice work on commit 2 tightening the 404 catch to check the typed body.

Findings below are all **minor/nit polish** — nothing blocking, nothing security-related.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 3 — orphaned `reconcileLostCreate` JSDoc; `possiblyTruncated` derived from post-filter count; new controller method has no `campaign.controller.spec.ts` coverage
- ⚪ Nit: 1 — new email-search interfaces sit under the `// HubSpot UTM` banner
- ❔ Question: 0

### Bot reconciliation
- **Cursor Bugbot + Copilot — "bare 404 misreports HubSpot status" (Medium):** I **disagree that this is still open** — it was filed against commit `2897449` (commit 1), where the catch was a bare `statusCode === 404`. Commit `1124668` already fixed it to `... && isCampaignServiceNotFound(error.errorBody)`, matching `findBrief`. The current HEAD is correct; this finding is stale. ✅ Resolved.
- **Cursor Bugbot — "truncation ignores dropped rows" (Low):** Agree — see inline. Kept at minor since the interface documents `id` as never-empty, so it's a defensive edge.
- **Copilot — detached `reconcileLostCreate` JSDoc:** Agree — see inline.
- **Copilot — controller boundary untested:** Agree — see inline.

### Decision
🔴 **Needs changes before approval** — three minor findings (just over the bar). None are hard blockers; all are quick. Happy to re-approve promptly once they're addressed or consciously waived.

apps/lfx-one/src/server/services/campaign-service.service.ts
mrautela365: Fixed in `38e09f645` — and you were right that it is the flag's own falsehood, not just a defensive edge.

`possiblyTruncated` is now derived from the **wire count**, taken before the id filter. Truncation is a property of what campaign-service sent; what survived our filtering is a different number. A capped 500 with one id-less row filtered to 499, and `499 >= 500` reported a truncated listing as complete — in a picker whose entire job is not hiding templates from the user.

Test asserts both halves: 499 emails returned AND `possiblyTruncated === true`. Verified by reverting to `emails.length`.

apps/lfx-one/src/server/services/campaign-service.service.ts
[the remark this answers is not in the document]
mrautela365: Fixed in `38e09f645` — and you were right that it is the flag's own falsehood, not just a defensive edge.

`possiblyTruncated` is now derived from the **wire count**, taken before the id filter. Truncation is a property of what campaign-service sent; what survived our filtering is a different number. A capped 500 with one id-less row filtered to 499, and `499 >= 500` reported a truncated listing as complete — in a picker whose entire job is not hiding templates from the user.

Test asserts both halves: 499 emails returned AND `possiblyTruncated === true`. Verified by reverting to `emails.length`.

mrautela365: Fixed in `38e09f645`. The `reconcileLostCreate` docblock is back directly above its method, so tooling binds the version-1/payload-match rationale to the right function again.

Moved the DOC rather than the method — worth noting because my first attempt moved `searchHubSpotEmails` instead, and the brace-matching cut `reconcileLostCreate`'s multi-line signature. esbuild caught it (`Expected ")" but found "async"`), but it is the kind of edit that can look right in a diff.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -449,6 +451,58 @@ export class CampaignServiceClient {
...
+        'GET',
+        query === '' ? undefined : { q: query }
+      );
+      return { enabled: true, emails: (response.data?.emails ?? []).map(fromMarketingEmail), error: null };
+    } catch (error) {
+      // A missing connection is not a failure of this request. campaign-service answers 404 for
+      // a project with no HubSpot connection, which is indistinguishable to the user from "you
+      // have not connected HubSpot yet" — because that is exactly what it is.
+      if (error instanceof MicroserviceError && error.statusCode === 404) {
+        return { enabled: false, emails: [], error: null };
```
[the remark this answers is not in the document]
mrautela365: Fixed in `38e09f645`. The `reconcileLostCreate` docblock is back directly above its method, so tooling binds the version-1/payload-match rationale to the right function again.

Moved the DOC rather than the method — worth noting because my first attempt moved `searchHubSpotEmails` instead, and the brace-matching cut `reconcileLostCreate`'s multi-line signature. esbuild caught it (`Expected ")" but found "async"`), but it is the kind of edit that can look right in a diff.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -449,6 +451,58 @@ export class CampaignServiceClient {
    * A 4xx is never reconciled: it is a refusal, so nothing committed and the original error is
    * the accurate answer. Only genuinely indeterminate failures reach the lookup.
    */
+  /**
+   * Search the project's HubSpot marketing emails, so a user can pick the template to clone.
```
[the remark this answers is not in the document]
mrautela365: Fixed in `38e09f645`. The `reconcileLostCreate` docblock is back directly above its method, so tooling binds the version-1/payload-match rationale to the right function again.

Moved the DOC rather than the method — worth noting because my first attempt moved `searchHubSpotEmails` instead, and the brace-matching cut `reconcileLostCreate`'s multi-line signature. esbuild caught it (`Expected ")" but found "async"`), but it is the kind of edit that can look right in a diff.

apps/lfx-one/src/server/controllers/campaign.controller.ts
mrautela365: Added in `38e09f645` — six tests on the controller boundary:

- the `project` refusal in three shapes (absent, blank, repeated → array), each asserting the service is **not** called and a 400 reaches `next`
- the `q` trim forwarded to the service
- an absent `q` treated as the unfiltered listing rather than a validation failure — that is the case the cap and `possiblyTruncated` exist for
- a service rejection forwarded to the error middleware instead of answering 200

Modelled on the `loadBrief` missing-project test you pointed at.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -516,6 +516,42 @@ export class CampaignController {
...
+    const projectSlug = typeof req.query['project'] === 'string' ? req.query['project'].trim() : '';
```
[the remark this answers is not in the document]
mrautela365: Added in `38e09f645` — six tests on the controller boundary:

- the `project` refusal in three shapes (absent, blank, repeated → array), each asserting the service is **not** called and a 400 reaches `next`
- the `q` trim forwarded to the service
- an absent `q` treated as the unfiltered listing rather than a validation failure — that is the case the cap and `possiblyTruncated` exist for
- a service rejection forwarded to the error middleware instead of answering 200

Modelled on the `loadBrief` missing-project test you pointed at.

packages/shared/src/interfaces/campaign.interface.ts
mrautela365: Fixed in `38e09f645`. The existing banner now reads `// HubSpot Email Templates` and names what actually follows it, and a real `// HubSpot UTM` banner sits above `HubSpotUtmLookupResult`.

(My first pass at this put the UTM banner *inside* an interface body and doubled a rule line — caught before pushing.)

mrautela365:
@dealako — all four addressed in `38e09f645`. Thanks for the read; the truncation one was a genuine correctness bug, not a defensive edge.

**`possiblyTruncated` derived post-filter** — fixed. Now read from the **wire count**, before the id filter. Your framing was the useful part: truncation is a property of what campaign-service *sent*, and what survived our filtering is a different number. A capped 500 with one id-less row filtered to 499, so `499 >= 500` reported a truncated listing as complete — in the one component whose job is not hiding templates. The test asserts both halves (499 returned **and** flagged), and I verified it by reverting to `emails.length`.

**Orphaned `reconcileLostCreate` JSDoc** — fixed by moving the *doc*, not the method. Worth mentioning: my first attempt moved `searchHubSpotEmails` instead, and the brace-matching cut `reconcileLostCreate`'s multi-line signature. esbuild caught it, but that class of edit can look fine in a diff.

**Controller boundary untested** — six tests now, modelled on the `loadBrief` case you pointed at: the `project` refusal in three shapes (absent / blank / repeated-as-array), the `q` trim, an absent `q` treated as the unfiltered listing rather than a validation failure, and error forwarding to `next`.

**Interface banner nit** — the existing banner now reads `// HubSpot Email Templates`, with a real `// HubSpot UTM` banner above the UTM types.

Also confirming your bot reconciliation: you were right that the bare-404 finding was **stale** — commit `1124668` had already tightened it to check `isCampaignServiceNotFound(error.errorBody)`, matching `findBrief`. I resolved those threads rather than re-fixing.

`1109 + 254` passing · `tsc --noEmit` clean on both tsconfigs · lint 0 errors · 0 open threads.

mrautela365:
@dealako — rebased onto the merged cutover and ready for re-review. All four of your findings were already addressed in `38e09f645`; this adds the merge on top.

**#1436 landing conflicted this branch** — they share five files. Every conflict was *additive on both sides* (my email work against its create work in the same regions), so all of them keep both halves.

Three files needed rebuilding rather than patching, which is worth flagging because the failure was quiet: a "keep both" splice concatenated two `describe` headers — and, in the service, two method docblocks — producing a stray docblock tail and a syntactically broken file. `tsc` caught it; a diff read would not have. Taking main's version as the base and re-grafting my block onto it is the resolution that holds.

One nice side effect: placing `searchHubSpotEmails` after the last **public** method satisfies `member-ordering` *and* keeps the `reconcileLostCreate` docblock adjacent to its method — the fix you asked for survives the merge rather than being undone by it.

Verified the merge kept both feature sets rather than silently dropping one:

- **1202 + 273** tests pass (mine were 1109; the rest are the cutover's create/poll tests plus what else landed on main)
- the truncation fix (`wireCount`) and the `hubspot/emails` route are both still present
- `tsc --noEmit` clean on both tsconfigs, lint 0 errors, headers pass

`MERGEABLE` again — confirmed with `git merge-tree` locally as well, since the API field lagged twice today.

mrautela365:
@dealako — **all four of your findings are in** (`38e09f645`), the branch is rebased onto the merged cutover, and I've since fixed five more Copilot raised. Your `CHANGES_REQUESTED` is from `112466823`; HEAD is `ff322f778`. This is next in the merge sequence after #1436.

### Since your review — one real bug, two false claims of mine

**A malformed 200 was reported as an empty portal.** `?? []` turned a response with no `emails` array into `enabled: true` with zero templates — indistinguishable from a portal that genuinely has none. That false absence is the exact failure this whole search is built to avoid. Now thrown so the catch reports a read failure.

Worth noting campaign-service draws the same line one layer up: `SearchEmails` treats a nil results array as a decode error, *because* a genuinely empty portal returns `[]` rather than nothing. I'd written the looser version of the same check.

**I called the filtered search "unbounded" in two places. It isn't.** `email.go` caps the walk at `maxListPages = 200` and returns *"exceeded 200 pages; refusing to page unbounded"* on exhausting it — never a partial list. That is a *better* property than unbounded, and it's the actual reason `possiblyTruncated` is false for a filtered search: a partial walk fails rather than answering. Both comments corrected.

**The picker copy I suggested said "the 500 most recent".** The service takes the first 500 rows in server order and sorts them *after* truncating, so that set is not the portal's newest 500. Suggesting that text would have put a second falsehood on top of the one the flag exists to prevent. Now "showing the first 500 — type to search the rest".

### One I did not fix here

`UNFILTERED_EMAIL_CAP` duplicates a private upstream constant across an independent deploy. Correct, and the fix belongs upstream — filed **[LFXV2-3255](https://linuxfoundation.atlassian.net/browse/LFXV2-3255)** for the endpoint to return explicit truncation metadata. I also corrected the note, which only described drift in the *safe* direction; a **lowered** cap makes a genuinely capped list read as complete, which is the failure mode that matters.

`1214 + 273` passing · tsc clean on both tsconfigs · lint 0 errors · 0 open threads · up to date with main.

[LFXV2-3255]: https://linuxfoundation.atlassian.net/browse/LFXV2-3255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dealako (approved):
Thanks for the thorough turnaround, @mrautela365 — every item from the prior round is addressed, and I verified each against the head commit (`75e28f2`) directly rather than taking the replies on faith.

### 👏 Nice work
- **Truncation now reads the wire, not the filtered list.** `const wireCount = wire.length` is taken *before* the id filter, so a capped 500 carrying one id-less row correctly reports `possiblyTruncated: true` instead of `499 >= 500 → false`. You were right to treat it as the flag's own falsehood rather than a defensive edge.
- **Fail-closed 2xx guard** (`09b7972`) — `if (!Array.isArray(wire)) throw` routes a malformed body through the read-failure catch (`enabled: true` + error) instead of silently rendering a false-empty portal. It also matches campaign-service treating a nil results array as a decode error. Clean symmetry.
- **Controller boundary is now covered** — the `CampaignController.searchHubSpotEmails` describe block exercises project refusal in all three shapes (absent / blank / array → 400, service not called), the `q` trim (`'  kubecon  '` → `'kubecon'`), absent-`q` → unfiltered listing, and service rejection → `next`. Exactly the seam I flagged.
- **`reconcileLostCreate` doc reattached** — the version-1/payload-match rationale sits directly above its method again, and `searchHubSpotEmails` keeps only its own block.
- **Banner split** into `// HubSpot Email Templates` and a real `// HubSpot UTM`, and — beyond my finding — you resolved Copilot's recurring accuracy note by replacing the "unbounded" claim with "COMPLETE-OR-ERROR, not unbounded" in both the service and the shared contract, citing the 200-page / 20s bound.

### Revision status
- ✅ **Resolved** — `possiblyTruncated` derived from post-filter count (was 🟡)
- ✅ **Resolved** — controller method had no test coverage (was 🟡)
- ✅ **Resolved** — orphaned `reconcileLostCreate` JSDoc (was 🟡)
- ✅ **Resolved** — email interfaces under the `// HubSpot UTM` banner (was ⚪)
- ✅ **Resolved** — Copilot's "unbounded" wording (doc-accuracy), fixed in `75e28f2`

### Bot reconciliation
Copilot's remaining suppressed notes on newer commits (the "unbounded"/complete-or-error wording and the banner-inside-interface concern) are already resolved at head — no divider sits inside any interface body, and the unbounded claims are gone. No open bot findings on the PR's own contribution.

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

### Decision
✅ **Approved** — every prior finding is closed, security/privacy is clean, and the branch is mergeable. Nice, careful iteration.

mrautela365:
Thanks @dealako. One push landed just after your approval — it survived (still APPROVED/MERGEABLE), but flagging what's in it so the approval isn't stale in substance:

- **Merged main** (#1436 and #1437 both landed). Auto-merged clean, no conflicts, no markers.
- **Corrected the PR description.** Copilot caught that it said *"the query is passed through and HubSpot does the filtering"* — the exact reverse of what the code does and what the service's design warns against. `q` never reaches HubSpot; campaign-service walks the pages and matches in-process. The code and comments were right; only the description was wrong, and it's now consistent with both.

`1228 + 282` passing, tsc clean on both tsconfigs, lint 0 errors, 0 open threads, up to date with main.

Ready to merge whenever CI settles.

Issue #1253, opened 2026-07-29
STORY E: View CLA details and request removal via the CLA manager


WillsonHG:
As a contributor, from **My CLAs** I can see the details of each CLA I've signed and — for ECLAs — start a **manager-routed removal** or contact my CLA manager. **Self-service invalidation is removed** per the 2026-08-14 legal/stakeholder review.

**What contributors can do (per the v17 Final prototype):**
- [ ] See the **identity each CLA was signed under** inline on every row ("Signed as <email>" / "Signed as <github> (GitHub)")
- [ ] For an ECLA: **Request Removal → Contact CLA Manager** (a message to the CLA manager(s); the manager performs the actual removal in the corporate console)
- [ ] For an ECLA where the user has CLA-manager permissions: **Manage in CCLA Console** deep link
- [ ] For a **Needs attention** ECLA: **Request approval** (also routes to the CLA manager)
- [ ] Sanctions/OFAC screening produces a **read-only "Revoked"** state — no user action

**Key boundary**: there are **no user-initiated invalidation writes** from the Me lens. ICLAs are informational-only here (no removal action). ECLA removal is initiated by the CLA manager; sanctions revocation is system-set.

**History**: the earlier self-service invalidation direction (typed-INVALIDATE modals, user-callable ICLA/ECLA invalidation endpoints) was **removed** by the 2026-08-14 legal review and the v17 Final prototype. Superseded tasks were closed on 2026-08-14.

---
_Reshaped 2026-08-14 per the legal/stakeholder review and the [v17 Final prototype](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html)._

Issue #1250, opened 2026-07-29
BE: Sign-CLA search endpoint (project / CLA group / org; pasted repo URL)


WillsonHG:
**Backend half of the Sign-CLA search** (FE is #1601). Build the search endpoint that powers the "Sign CLA" modal, per [mockup v17 Final](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html).

FR-001 requires a **new unscoped search endpoint** — no existing endpoint covers these sources. Implementation approach is yours to design; the criteria below are outcomes, not a design.

## Acceptance criteria

- [ ] A contributor typing in the Sign CLA search box can find a CLA group by **project name**, **CLA-group name**, or **org name**
- [ ] Org results carry **repo-source provenance** (GitHub / GitLab / Gerrit) so the FE can display where the group is hosted
- [ ] A **pasted repo URL or `owner/repo`** resolves to its CLA Group
- [ ] Every result carries the **CLA Group ID** — that is the signing unit, and the FE needs it for the hand-off to `{console}/#/cla/project/{claGroupID}/user/{userID}` (FR-002). There is no org/repo selection step
- [ ] **&lt; 300 ms p95 server time** per search (FR-001a), measured at real cardinality and reported here
- [ ] Matching happens **server-side** — the endpoint must not depend on the client filtering a full list
- [ ] Results are **bounded**, and the response tells the FE when it truncated so the user can be asked to refine
- [ ] GitLab-backed CLA groups **appear in results** with their provenance; sign entry is gated on the FE side per #1249
- [ ] Swagger-first: update `swagger/cla.v2.yaml`, run `make swagger`, then implement (repo convention — never hand-edit `gen/`)
- [ ] Unit tests including the error paths
- [ ] **Publish the response shape to #1601 once settled** — Ahmed wires the UI to whatever you build, so he is blocked until it exists. Provisional is fine

[...]

Issue #1249, opened 2026-07-29
STORY D: Contributor can browse and initiate signing a CLA from the Me lens


WillsonHG:
As a contributor, I can search for a project or CLA group from **My CLAs** and start signing proactively — without needing a failing PR to trigger it. Signing itself completes in the existing Contributor Console (SS hands off; it does not run DocuSign).

**Acceptance criteria**:
- [ ] "Sign CLA" opens a modal with one search box matching project names, CLA group names, and linked repo sources — **GitHub, GitLab, and Gerrit, with repo-source provenance displayed** — and accepting a pasted repo link resolved to its CLA Group (FR-001). Showing GitLab provenance in search results is **not** gated on GitLab account linking; only the GitLab *sign entry* is (see below)
- [ ] No org/repo selection step — the signing unit is the CLA Group
- [ ] On selection, SS hands off to the Console decision screen `{console}/#/cla/project/{claGroupID}/user/{userID}` (FR-002); ICLA/ECLA choice + legal guidance stay in the Console
- [ ] **Gerrit-backed groups skip the account step entirely** (FR-004) — Gerrit signers use the same LF SSO session SS already holds, so there is no linking dependency and nothing to gate
- [ ] **GitHub-backed groups run a read-only identity pre-flight** against `GET /v4/my-clas/identities` before the hand-off (FR-004). SS writes no identity — binding stays owned by the Console's GitHub OAuth
- [ ] **When one or more GitHub accounts are linked**: show the account-picker modal so the contributor chooses which account to sign with — **no silent auto-pick, even when only one account is linked** (per the v17 prototype's `openAcctModal`, which always renders the radio list)
- [ ] **When zero GitHub accounts are linked**: the picker shows an empty state that **blocks continuing** — "No GitHub accounts are connected to your profile yet" — with a link to Identities to link one. It MUST NOT fall through to the Console (per v17: the account list is hidden and the continue action is removed). This is a stricter gate than Gerrit, which needs no account step at all
- [ ] Compare identities as **numeric GitHub IDs**, not usernames (both sides already agree on this). A GitHub ID bound to a different EasyCLA record is **log-only** in M2 — no reconciliation (M3)
- [ ] **GitLab-backed CLA groups** (~2 projects): sign entry is **conditional on SS shipping GitLab account linking within M2's window** — enabled by config if it lands, follow-up otherwise. **M2 does not block on it.** GitLab follows the same gating shape as GitHub, by config — not a redesign
[...]

Issue #1229, opened 2026-07-29
M2 — My CLAs actions: proactive sign entry (Console hand-off), Revoked (we are no longer invalidating after legal review), status


WillsonHG:
**Goal**: Extend M1's **My CLAs** page (Me lens) with four additions, per the revised M2 spec and the [v17 Final mockup](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html):

1. **Sign a CLA** — a proactive entry (search by project / CLA group / repo source / pasted repo link) that **hands off to the existing Contributor Console** decision screen to complete signing. GitHub signers pick which linked account to sign with; Gerrit needs no account step. (#1249)
2. **Request removal / approval (ECLA only)** — **no longer an invalidation.** Per legal (8/13) we will no longer invalidate; instead we route to the CLA manager if the user needs to report no longer being included on a CCLA. "Request Removal" and "Request approval" open a shared Contact-CLA-Manager modal that emails the resolved manager(s); the manager completes the removal in the CCLA console. ICLA rows carry no removal action. (#1574, #1372, #1369, #1575)
3. **Status** — a richer status column (Valid / Needs attention / Revoked), where **Revoked** is system-set (sanctions/OFAC), read-only, dated, with no user actions. (#1256, #1370)
4. **Signed-under identity** — each row shows which platform and account the CLA was signed with (GitHub / GitLab / Gerrit), rendered inline in the Signed cell, informational only. (#1573)

**Key boundary**: Self Serve never runs the DocuSign ceremony, makes no signing-initiation calls, makes **no invalidation writes of any kind**, and does not change the PR-check remediation link. Nothing is cut over or retired. The ICLA/ECLA choice and all signing logic stay in the Contributor Console.

**History**: the earlier "Flow 1" framing (redirect the PR check into SS, retire the Contributor Console) was dropped as too disruptive — Stories A/B/C (#1230, #1241, #1245) and their tasks were closed on 2026-08-06. This Epic now covers only the additive My-CLAs work (Stories D #1249 and E #1253).

---
[...]

PR #1440, opened 2026-08-13
feat(cla): dark-launch the M2 CLAs tab with status, sign, and actions

### LINKED ISSUE #1251: Hand off to the Contributor Console decision screen
On CLA-Group selection, hand off to the Contributor Console's existing decision-screen URL (`Corporate Contributor` vs `Individual Contributor` [screen](https://easycla.dev.communitybridge.org/#/cla/project/01af041c-fa69-4052-a23c-fb8c1d3bef24/user/4b344ac4-f8d9-11ed-ac9b-b29c4ace74e9?redirect=https:%2F%2Fgithub.com%2Fmlehotskylf-org2%2Feasycla-dev%2Fpull%2F45)) to complete signing.

**Acceptance criteria**:
- [ ] Redirect to `{console}/#/cla/project/{claGroupID}/user/{userID}`. I believe redirecting should be back to Self Serve.
- [ ] SS makes **no** signing-initiation calls (`request-individual-signature`, `request-employee-signature`, etc.) — the Console makes those, exactly as today
- [ ] ICLA/ECLA choice, legal guidance, and `project_icla_enabled` / `project_ccla_enabled` gating remain in the Console; not re-implemented in SS
- [ ] PR-check remediation link unchanged

Start signing process here which forwards user to Contributor Console:
Image

Contributor console stays the same:
Image

### LINKED ISSUE #1252: GitHub account-authorization step before Console hand-off
Before handing off to the Console, GitHub signers pass a one-time **account-authorization step** so the platform identity they'll contribute with is linked to their EasyCLA user record (FR-004).

**Acceptance criteria**:
- [ ] GitHub: reuse M1's Auth0 account-linking flow; support a multi-account picker when the LF user has more than one linked GitHub identity (FR-004)
- [ ] Gerrit: no authorization step (same LF SSO); skip straight to hand-off
- [ ] GitLab: platform-parametrized but excluded from M2 (gated on SS shipping GitLab linking)
- [ ] Copy/labels for the step match the Final/v16 mockup

One or more GH/GL/Gerrit accounts linked in SS:
Image

No GH/GL/Gerrit accounts linked in SS:
Image

### LINKED ISSUE #1256: Status column — Valid / Needs attention / Revoked
Render the status column. Following Heather's 2026-08-19 decision the column carries **four pills — Valid / Needs attention / Invalidated / Revoked** — plus `unknown`, rendered as a plain "—" rather than a named pill. Invalidated is **not** the old self-service pill (contributors still cannot self-invalidate); it is the system state for a signature whose `signature_approved` flag was set to `false` by someone else.

The authoritative mapping, verified against `cla-backend-go` and the merged frontend, is the **[M2 status matrix](https://github.com/linuxfoundation/easycla/blob/dev/specs/001-easycla-ss-integration-fable/m2-sign-cla-handoff/status-matrix.md)** (linuxfoundation/easycla#5155). Implement against that table; it is the tiebreaker if this ticket and the matrix ever disagree.

**Acceptance criteria**:
- [ ] **Valid** — signed, `signature_approved=true`, and (ECLA) covered by the company's current Approved List criteria
- [ ] **Needs attention** — ECLA only. Still `signature_approved=true`, but a **completed** coverage check proved the ECLA does not cover the contributor. Shows the explanatory note, plus **Request approval** (#1372) when — and only when — the reason is `not_on_approval_list`, alongside Request Removal and Contact CLA Manager
- [ ] **Invalidated** — ICLA and ECLA. `signature_approved=false`, i.e. the signature was invalidated externally: PCC admin ICLA invalidation, Approved List removal, or CLA-group deletion. The pill deliberately **attributes no cause or actor**, because those three are indistinguishable on the record. ICLA carries no actions; ECLA retains Request Removal. Dated `Invalidated · <date>` only for PCC ICLA invalidation (#1732) — the other two paths write no timestamp
- [ ] **Revoked** — ECLA only, system-set by sanctions screening. **Read-only**: no Request Removal, no Request approval, no Contact CLA Manager, no download (per the v17 prototype). Shows `Revoked · <date>` from the company's `sanctioned_date` (#1370)
- [ ] **Revoked keys off the company, not the signature.** Sanctions set the company's `is_sanctioned` flag and **never** touch `signature_approved`. Do not implement Revoked as `approved = false` — that is Invalidated
- [ ] **Invalidated and Revoked must never share a label.** They are distinct states with distinct remediation
- [ ] **`unknown` renders as "—"**, not a named pill — ECLA coverage could not be evaluated (a transient fault, not an answer). No actions
- [ ] Status sourced from `GET /v4/my-clas`, which exposes `status` and `statusReason` directly (FR-010). Do not recompute the mapping client-side
- [ ] Copy avoids "Canceled/Invalid" — the agreed terms are **Revoked** (Mike, 8/13) and **Invalidated** (Heather, 8/19)
- [ ] Empty/loading/error states consistent with the M1 page

**Known open item**: E6 in the matrix — an invalidated ECLA at a sanctioned employer. The shipped backend returns `revoked` (it checks the sanction flag first); the 2026-08-24 product call was Invalidated. Unresolved; backend and this column will change together once decided. It is a rare intersection and should not block the rest of the column.

_Note: the corporate console may present the revoked state differently to the CLA manager — out of scope for the Me lens here._

_Changed 2026-08-14: dropped the "Invalidated" self-serve pill; added the system-set "Revoked" state. FR-011 "Request approval" is #1372._

_Updated 2026-08-17: absorbed the Revoked-state task (#1576) here, per Heather — the three states ship as one column and splitting them added no value. #1370 (revocation metadata) supplies the date._

_Updated 2026-08-24: reinstated **Invalidated** as a system-set pill per Heather's 2026-08-19 decision, and corrected the Revoked criterion — it comes from the company's `is_sanctioned` flag, not from `approved = false`. The previous wording conflated the two states. Aligned with the [M2 status matrix](https://github.com/linuxfoundation/easycla/blob/dev/specs/001-easycla-ss-integration-fable/m2-sign-cla-handoff/status-matrix.md)._

---
_Reshaped 2026-08-14 per the legal/stakeholder review and the [v17 Final prototype](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html)._

Image

### LINKED ISSUE #1365: Resolve EasyCLA userID server-side for the Console hand-off
Resolve the contributor's EasyCLA `userID` **server-side** before building the Console hand-off URL — never trust a client-supplied ID (FR-003).

**Acceptance criteria**:
- [ ] Platform-aware resolution: use the linked GitHub ID first; fall back to `GET /v4/user-from-token` (+ enrichment) when no linked record is found (FR-003)
- [ ] The resolved `userID` is what populates `{console}/#/cla/project/{claGroupID}/user/{userID}` — the hand-off task (#1251) consumes this, it does not derive its own ID
- [ ] Handle the `user-from-token` lookup-or-create side effect explicitly (the one implicit write on the sign path); log/telemeter created-vs-matched
- [ ] Guard the case where `user-from-token` returns a record bound to a **different** linked GitHub identity — do not silently hand off with a mismatched ID (coordinate with the identity-binding task)
- [ ] Server-side only; no arbitrary-user lookup exposed to the client

**Depends on / relates to**: #1252 (account-authorization picks the identity), the identity-binding op (open Q1), #1251 (consumer of the resolved ID).

---
_Created 2026-08-10 from the revised M2 spec: [spec.md](https://github.com/linuxfoundation/easycla/blob/docs/easycla-ss-m2-speckit/specs/001-easycla-ss-integration-fable/m2-sign-cla-handoff/spec.md) (FR-003) · [plan.md](https://github.com/linuxfoundation/easycla/blob/docs/easycla-ss-m2-speckit/specs/001-easycla-ss-integration-fable/m2-sign-cla-handoff/plan.md) §Technical Context._

### LINKED ISSUE #1372: Request approval action on Needs-attention ECLA rows
On a **Needs attention** ECLA row, **Request approval** opens the shared **Contact CLA Manager** message modal (approval mode) so the contributor can ask the CLA manager(s) to re-approve their ECLA (FR-011).

**Acceptance criteria**:
- [ ] "Request approval" shown only on **Needs attention** ECLA rows
- [ ] Opens the shared Contact-CLA-Manager modal in **approval** mode (CLA manager(s) list + optional message + Send) — same component used by Request Removal
- [ ] Send triggers the CLA-manager notification (#1369); success/error states handled
- [ ] **Not** a deep link into the Console — it's an in-app message to the manager

_Changed 2026-08-14: from "deep link into the Console request-approval flow" to the in-app Contact-CLA-Manager message modal (per the v17 prototype's `openMgrModal(..., 'approval')`)._

**Depends on**: the Request-Removal task (shared modal + manager resolution). **Relates to**: #1256, #1369, #1253.

---
_Reshaped 2026-08-14 per the legal/stakeholder review and the [v17 Final prototype](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html)._

Image

### LINKED ISSUE #1423: Expose per-row status and reason on GET /v4/my-clas
Extend `GET /v4/my-clas` to return an explicit per-row **status** and **reason**. [#1256](https://github.com/linuxfoundation/lfx-self-serve/issues/1256) (status column) ships the three-state column by deriving it client-side (`approved=false` → Invalidated, `approved=true && valid=false` → Needs attention, `valid=true` → Valid). That works for the states, but two things cannot be derived from two booleans: **why** a row is not valid, and **whether the evaluation actually completed**. This ticket adds both.

**Acceptance criteria**:
- [ ] Additive `status` on each row: `valid` / `needs_attention` / `invalidated` / `unknown`. `approved` and `valid` stay on the wire unchanged — existing consumers must not break
- [ ] Additive `statusReason` with **two** tokens only:
  - `not_on_approval_list` — genuine Approved List miss → `status=needs_attention`; UI note *“No longer matches {company}'s approval criteria.”*
  - `unknown` — every other ECLA coverage outcome (employer sanctioned, company record missing, no approved+signed CCLA, evaluation/lookup failure) → `status=unknown`; UI **—** (no fourth named pill, no invented sentence)
- [ ] Omit `statusReason` on `valid` rows and on **every ICLA**. ICLA never evaluates coverage, so it never emits `needs_attention`, `unknown`, or a listing note
- [ ] `unknown` is **not** Needs attention. Do not put the mockup sentence or a Request approval link on it. [#1372](https://github.com/linuxfoundation/lfx-self-serve/issues/1372) (Request approval) gates on `not_on_approval_list` only
- [ ] A failed company or CCLA lookup degrades **that row** (`unknown`) instead of failing the whole request. Today one bad `GetCorporateSignature` lookup 500s the entire list (`v2/my_clas/service.go:214-217`); `company()` non-NotFound errors do the same (`:207-208`)
- [ ] Listing must treat a GitHub-org lookup failure as `unknown`, not a silent no-match (`signatures/service.go:1688-1694`). Do **not** change PR-gating (`ProcessEmployeeSignature`) as a side effect
- [ ] Swagger-first: `swagger/common/my-cla.yaml` → `make swagger` → service; unit tests per cause asserting the two-token mapping, including `unknown` and per-row degradation
- [ ] Self Serve consumes it **on `feat/GH-1256`**: delete the interim derivation added by #1256 (status column) and read `status` directly; `not_on_approval_list` fills the note; `unknown` renders **—**

**Blocks**: #1372 (Request approval)
**Follow-up to**: #1256 (status column)
**Relates to**: #1370 (additive invalidation metadata)

**Repo**: `cla-backend-go` (swagger + service), plus `lfx-self-serve` consume on `feat/GH-1256`.

CC @lukaszgryglicki

### LINKED ISSUE #1573: Show the identity each CLA was signed under (inline per row)
Show the **identity each CLA was signed under**, inline on every My-CLAs row — the informational replacement for the removed "Invalidate" action.

**Acceptance criteria**:
- [ ] Each row shows the signer identity **and which platform it was signed via**, alongside the signed date (per the v17 prototype): "Signed as `<email>`" (Gerrit), "Signed as `<github-username>` (GitHub)", or "Signed as `<gitlab-username>` (GitLab)"
- [ ] **Renders inline beneath the signed date, inside the existing Signed cell — not as a new table column** (v17 renders it as a `.signedas` div nested in the Signed `<td>`)
- [ ] Rendered for ICLA and ECLA rows, in all states (Valid / Needs attention / Revoked)
- [ ] **GitLab is unconditional here** — unlike GitLab *sign entry* (#1249), which waits on SS GitLab account linking, displaying a historical GitLab signer identity has no such dependency: if a GitLab-signed record exists, the row must render it correctly
- [ ] Sourced from the existing My-CLAs data (`GET /v4/my-clas`) — the signer identity/reference and its repo-source type are already tied to the signature; no new endpoint
- [ ] No action attached — informational only

_Updated 2026-08-17: added GitLab (previously only email + GitHub were listed) and the explicit "inline, not a new column" rule from spec FR-007 — the v17 prototype nests it in the Signed cell, and an earlier discussion had loosely described it as a new column._

**Relates to**: #1253 (Story E), #1256 (status column), #1249 (GitLab sign-entry condition), M1's My CLAs list.

---
_Created 2026-08-14 per the legal/stakeholder review and the [v17 Final prototype](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html)._

Image

### LINKED ISSUE #1574: ECLA Request Removal → Contact CLA Manager (modal + manager resolution + notify)
Build the **Request Removal → Contact CLA Manager** flow for ECLA rows: a shared message modal plus the backend to resolve and notify the CLA manager(s). The manager performs the actual removal in the corporate console — **SS never invalidates**.

**Acceptance criteria**:
- [ ] **Request Removal** action on ECLA rows opens the shared **Contact CLA Manager** modal (removal mode): a **CLA manager(s) list**, an optional message, and Send (per the v17 prototype's `openMgrModal(..., 'removal')`)
- [ ] Backend **resolves the CLA manager(s)** for the ECLA's company + CLA group from the **CCLA signature ACL** — `GetProjectCompanySignatures` → `SignatureACL`, hydrated to full users with emails by `GetUserByUserName` per entry (`signatures/converters.go:162-181`). This is the same source both existing request flows use; **no acs-service role check is involved**
- [ ] **The modal displays the resolved manager(s) and lets the contributor choose recipients.** Per v17, each manager renders as a **checkbox, checked by default**, and can be unchecked (`mgrManagerList` / `toggleMgrManager`); a group may have several (the prototype mocks 2 for CNCF, 1 for Anuket). All three hint strings say "the CLA manager(s) **below**", so the list is part of the copy contract — the contributor sees who will receive the request
- [ ] **Send goes only to the checked manager(s)** — the request payload carries the selected recipients, not blindly all resolved managers (triggers the notification, #1369); it does **not** invalidate anything in SS
- [ ] **Send is disabled when no CLA manager is checked** — at least one recipient is required (decided 2026-08-17; not covered by the prototype). The backend MUST also reject an empty recipient list, rather than trusting the disabled button as the only guard
- [ ] **Ownership check**: the caller MUST own the ECLA signature they are requesting removal for (reuse `authorizeIdentity`, `v2/my_clas/service.go:383`) so the endpoint can't be used to probe or spam notifications for another contributor's agreement
- [ ] **Zero-manager case (FR-008c)**: an empty or stale `signature_acl` resolves to **no recipients**. The existing Approved-List flow merely logs a warning and bails (`approval_list/service.go:140-145`) — do **not** reproduce that. The modal MUST show an explanatory state telling the contributor no CLA manager is currently reachable for this company, pointing them at LF support, instead of a Send button that silently goes nowhere. The request record MUST still be written so the dead end is diagnosable
- [ ] Modal copy makes clear the manager completes the removal ("This starts the process to invalidate it on your behalf")
- [ ] Same modal component serves **Request approval** (#1372) in approval mode
- [ ] Loading/success/error states; no removal control on ICLA or Revoked rows

**Backend**: resolve CLA manager(s) for a company/CLA-group; accept + route the removal request. Follow the existing **Approved-List request** pattern (`addCclaAllowlistRequest`, `POST /company/{companyID}/ccla-whitelist-requests/{projectID}`) rather than inventing a new one — see #1369 for the full four-step shape. **Repos**: `lfx-self-serve` (modal + route), `cla-backend-go` (manager resolution + request handling). **Relates to**: #1369, #1372, #1253, #1575.

_Updated 2026-08-17: named the signature-ACL resolution source, added the ownership check and the zero-manager case (spec FR-008c), and specified the manager list as **selectable** — v17 renders one checked checkbox per manager with a toggle, so Send must respect the selection. Send is disabled when nothing is checked. (An earlier edit today briefly removed the manager list on a misreading of the prototype; restored — the list is in the markup and the modal copy points at it.)_

---
_Created 2026-08-14 per the legal/stakeholder review and the [v17 Final prototype](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html)._

Image

Image

### LINKED ISSUE #1601: FE: Sign-CLA search modal (debounced search, results, hand-off)
**Frontend half of the Sign-CLA search** (BE endpoint is #1250). Build the "Sign CLA" modal — search box, results dropdown, and hand-off — per mockup v17 Final.

**Depends on #1250.** The endpoint is designed and built there; wire the UI to whatever shape it ships. Coordinate with Lukasz on the response shape as soon as it is settled — it does not need to be final to start building against it.

## Acceptance criteria

- [ ] **"Sign CLA" action** on the My CLAs page opens the modal with a single search box (no separate filters, no org/repo selection step — the signing unit is the CLA Group)
- [ ] **Debounce input ~200 ms** (FR-001a) — do not fire a request per keystroke
- [ ] **Cancel in-flight requests** when the term changes. Typing quickly must never let a stale response overwrite a newer one in the dropdown — this is the most common bug in this pattern; cover it with a test
- [ ] **Render the result types distinctly** — CLA group, project, org (with repo-source provenance shown: GitHub / GitLab / Gerrit), and a resolved pasted repo URL
- [ ] **Collapse multiple linked orgs** into an expandable "N linked orgs" affordance (FR-001); a single org renders inline
- [ ] **Surface truncated results** — when the endpoint indicates it capped the result set, tell the user so they refine the term rather than assuming their project is missing
- [ ] **States**: loading (in-flight), empty (no term yet), no-match (term with zero results), and error — consistent with the M1 My CLAs page
- [ ] **On selection, hand off** to `{console}/#/cla/project/{claGroupID}/user/{userID}` — no `?redirect=` (FR-002). The ICLA/ECLA choice and legal guidance stay in the Console
- [ ] **GitHub-backed groups**: run the account-picker pre-flight *before* the hand-off, per #1249 — always show the picker when ≥1 account is linked, block with the empty state when none are. Gerrit skips it. **That gating logic belongs to #1249**; this ticket just calls into it at the right moment
- [ ] Keyboard accessible: arrow-key navigation through results, Enter to select, Esc to close
- [ ] Unit tests covering debounce, stale-response cancellation, org collapsing, and the empty/error states

## Notes

- **GitLab results are displayed, not hidden.** Provenance shows for GitLab-backed groups; only the *sign entry* is gated on SS GitLab linking (#1249). Don't filter them out of the list
- Search matching is entirely server-side (#1250) — do **not** fetch a list and filter it in the client

**Repo**: `lfx-self-serve`. **Sibling**: #1250 (BE endpoint) — build after it lands. **Relates to**: #1249 (Story D, account-picker gating), FR-001 / FR-001a / FR-002.

---
_Created 2026-08-17: FE half of the search work split out of #1250, per mockup v17 Final. Updated 2026-08-18: removed the pinned response-contract fixture — the BE endpoint is built first and this wires to whatever shape it ships. Spec: [spec.md](https://github.com/linuxfoundation/easycla/blob/docs/easycla-ss-m2-speckit/specs/001-easycla-ss-integration-fable/m2-sign-cla-handoff/spec.md) (FR-001, FR-001a, FR-002)._

Image

### LINKED ISSUE #1738: Add LaunchDarkly feature flag: my-clas-m2-enabled
Add a new LaunchDarkly flag `my-clas-m2-enabled` to gate M2 My CLAs features.

- DEV: enabled for everyone
- PROD: enabled only for our team until PROD testing is complete

Part of #1249.

ahmedomosanya:
The M2 CLAs tab is dark-launched behind `my-clas-m2-enabled` (default **off**). Flag off, `/profile/clas` stays the shipped M1 list (Project / Type / Signed / Document). Flag on, the overlay appears: status, Signed as, kebab, Sign CLA, and CLA-manager actions.

LaunchDarkly: **DEV** on for everyone; **PROD** team-only until testing. Code default is `false`.

Epic: [#1229 (M2 — My CLAs actions)](https://github.com/linuxfoundation/lfx-self-serve/issues/1229)

Closes [#1256 (Status column — Valid / Needs attention / Revoked)](https://github.com/linuxfoundation/lfx-self-serve/issues/1256)
Closes [#1423 (expose per-row status + reason)](https://github.com/linuxfoundation/lfx-self-serve/issues/1423)
Closes [#1573 (Show the identity each CLA was signed under)](https://github.com/linuxfoundation/lfx-self-serve/issues/1573)
Closes [#1251 (Hand off to the Contributor Console)](https://github.com/linuxfoundation/lfx-self-serve/issues/1251)
Closes [#1252 (GitHub account-authorization step)](https://github.com/linuxfoundation/lfx-self-serve/issues/1252)
Closes [#1601 (FE: Sign-CLA search modal)](https://github.com/linuxfoundation/lfx-self-serve/issues/1601)
Closes [#1372 (Request approval on Needs-attention ECLA rows)](https://github.com/linuxfoundation/lfx-self-serve/issues/1372)
Closes [#1574 (ECLA Request Removal → Contact CLA Manager)](https://github.com/linuxfoundation/lfx-self-serve/issues/1574)
Closes [#1738 (Add LaunchDarkly flag my-clas-m2-enabled)](https://github.com/linuxfoundation/lfx-self-serve/issues/1738)
Closes [#1365 (Resolve EasyCLA userID server-side)](https://github.com/linuxfoundation/lfx-self-serve/issues/1365)

## Summary
- Every signed ICLA and ECLA is listed. Status is copied from the producer (`valid` / `needs_attention` / `revoked` / `invalidated` / `unknown` as em dash). `revoked` and `invalidated` are different pills. No dates on the pills.
- Flag on, the table is Project · Type · Status · Signed · ⋮. **Signed as …** sits under the date. Flag off, Document stays (Download PDF / Covered by Corporate CLA).
- **Sign CLA** searches live CLA groups, optionally picks a linked GitHub account, then `POST /v4/self-serve/prepare-sign` and leaves for the Contributor Console. Withheld while impersonating.
- ECLA kebab: Request approval (Needs attention + `not_on_approval_list`), Request Removal, Contact CLA Manager. Contact Send is a no-op — the producer has no `contact` type. Invalidated ICLA and revoked rows have no kebab.

## Upstream
- [easycla#5151](https://github.com/linuxfoundation/easycla/pull/5151) — `status` / `statusReason` / CLA-manager request routes (merged to `dev`; replaced the earlier #5146 consume note)
- [easycla#5149](https://github.com/linuxfoundation/easycla/pull/5149) — `GET /v4/cla-group/search`
- [easycla#5150](https://github.com/linuxfoundation/easycla/pull/5150) — `POST /v4/self-serve/prepare-sign`

## Related (not closed by this PR)
- [#1249 (Story D: browse and initiate signing)](https://github.com/linuxfoundation/lfx-self-serve/issues/1249) — parent of the sign slice
- [#1253 (Story E: view details and request removal)](https://github.com/linuxfoundation/lfx-self-serve/issues/1253) — parent of #1573 / #1574
- [#1250 (BE: Sign-CLA search endpoint)](https://github.com/linuxfoundation/lfx-self-serve/issues/1250) — consumed, not implemented here
- [#1576 (Revoked read-only status)](https://github.com/linuxfoundation/lfx-self-serve/issues/1576) — already closed; this PR renders the pill
- [#1575 (Manage in CCLA Console)](https://github.com/linuxfoundation/lfx-self-serve/issues/1575) — kebab link not built in this PR
- Contact CLA Manager (third kebab item) has no ticket

## Out of scope
- [#1575 (Manage in CCLA Console)](https://github.com/linuxfoundation/lfx-self-serve/issues/1575)

## Test plan
- [ ] Flag **off**: Project / Type / Signed / Document; no Sign CLA, Status, kebab, or Signed as
- [ ] Flag **on**: Valid / Needs attention / Revoked / Invalidated pills, no dates; Signed as GitHub / GitLab / email
- [ ] ICLA with PDF → Download PDF; ECLA → Covered by Corporate CLA; invalidated ICLA → no kebab
- [ ] Sign CLA → search ≥3 chars → pick group → GitHub picker when 2+ accounts → Console → Exit returns to `/profile/clas`
- [ ] Impersonating: Sign CLA hidden; prepare-sign 403
- [ ] Needs-attention ECLA: Request approval POSTs; Contact Send does not
- [ ] Preview: https://ui-pr-1440.dev.v2.cluster.linuxfound.info/profile/clas


apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.html
```diff
@@ -98,23 +99,35 @@ <h2 class="text-lg font-semibold text-slate-800">My CLAs</h2>
             </td>
 
             <td class="px-5 py-3.5">
-              <span class="text-sm text-slate-600">{{ agreement.signedOn ? (agreement.signedOn | date: 'mediumDate' : 'UTC') : '—' }}</span>
+              <div class="flex flex-col items-start gap-1">
+                <lfx-tag
+                  [value]="claStatusLabel(agreement.status)"
+                  [severity]="claStatusSeverity(agreement.status)"
+                  [icon]="statusIcon(agreement.status)"
```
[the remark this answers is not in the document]
ahmedomosanya: Menu allocation addressed in 3db95ed11 — row menus are a computed Map, not a per-pass `rowMenuItems()`. Status helpers (`claStatusLabel` / `statusNote`) stay as template function-refs, matching the existing #1256 column; not rewriting that cell in this consume. Line 120 is the Signed `date` pipe, not a method.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.html
```diff
@@ -98,23 +99,39 @@ <h2 class="text-lg font-semibold text-slate-800">My CLAs</h2>
...
 
             <td class="px-5 py-3.5">
-              <span class="text-sm text-slate-600">{{ agreement.signedOn ? (agreement.signedOn | date: 'mediumDate' : 'UTC') : '—' }}</span>
+              <div class="flex flex-col items-start gap-1">
+                @if (agreement.status === 'unknown') {
+                  <span class="text-sm text-slate-600" [attr.data-testid]="'agreement-status-' + agreement.id">{{ claStatusLabel('unknown') }}</span>
+                } @else {
+                  <lfx-tag
+                    [value]="claStatusLabel(agreement.status)"
+                    [severity]="claStatusSeverity(agreement.status)"
```
dealako: [minor] Status cell uses template function calls instead of a precomputed view model

Issue: `claStatusLabel('unknown')` (L104), `claStatusLabel(agreement.status)` (L107), `claStatusSeverity(agreement.status)` (L108), `statusIcon(agreement.status)` (L109), and `statusNote(agreement)` (L113) all execute logic inline, once per visible row, on every change-detection pass.
Proof: `docs/reviews/frontend-checklist.md` §4 forbids template method calls (signal reads / computed / pipes only) — its example `[class]="getStatusClass(item.status)"` is structurally identical to `[severity]="claStatusSeverity(agreement.status)"`. You already solved this correctly in the same component: `menuItemsMap` is a `computed` Map (component.ts) with a comment explaining the per-CD binding hazard — the status fields just weren't given the same treatment.
Why it matters: re-runs the switch logic for every row on each CD cycle (zoneless/OnPush still re-evaluates on any tracked signal write, e.g. `downloadingId`), and it's inconsistent with the precompute pattern already established one cell over.
Fix: derive a per-row VM, e.g. `rows = computed(() => this.agreements().map(a => ({ ...a, statusLabel: claStatusLabel(a.status), statusSeverity: claStatusSeverity(a.status), statusIcon: statusIcon(a.status), statusNote: statusNote(a) })))`, iterate that, and bind `row.statusLabel` / `row.statusSeverity` / etc. Replace `claStatusLabel('unknown')` with a static string.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.html
```diff
@@ -98,23 +99,39 @@ <h2 class="text-lg font-semibold text-slate-800">My CLAs</h2>
...
-              } @else if (agreement.kind === 'ECLA') {
-                <span class="text-xs text-slate-400">Covered by Corporate CLA (CCLA)</span>
-              } @else {
-                <span class="text-xs text-slate-400">PDF unavailable</span>
-              }
+              <span class="text-sm text-slate-600">{{ agreement.signedOn ? (agreement.signedOn | date: 'mediumDate' : 'UTC') : '—' }}</span>
+            </td>
+
+            <td class="relative px-5 py-3.5 text-right">
+              <lfx-button
```
dealako: [minor] ICLA rows with no available PDF render a clickable ⋮ that opens an empty menu

Issue: the actions ⋮ button and `<lfx-menu>` render unconditionally; the button is only `[disabled]` while a download is in flight.
Proof: `buildRowMenuItems` (component.ts) returns the download item only `if (agreement.pdfAvailable)`, the disabled CCLA item only for `kind === 'ECLA'`, and otherwise `return []`. `pdfAvailable = isIcla && cla.pdfAvailable === true` (cla.service.ts L138), so a *valid ICLA whose upstream `pdfAvailable` is not true* falls through to `[]` — and now that valid-only filtering is removed, such rows are actually displayed. The menu binds `menuItemsMap().get(agreement.id) ?? []`, so the trigger opens a blank popup. This is exactly the Copilot suppressed finding.
Why it matters: presents an affordance that does nothing — reads as a broken control on a row that's now always visible.
Fix: guard the trigger on a non-empty menu, e.g. precompute a `hasMenu` flag in `menuItemsMap` and wrap with `@if (hasMenu) { …button + menu… }` (or restore the former "PDF unavailable" text). Add a spec asserting no ⋮ renders for an ICLA with `pdfAvailable: false`.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -100,10 +100,9 @@ export function collectClaEmails(primaryEmail: string | null, emailData: EmailMa
 /**
  * Maps an upstream `my-cla` record to the UI view model.
  *
- * Validity is authoritative from upstream (`valid`, computed against the *current*
- * CCLA approval lists), so SS does not recompute it: `valid` ⇒ 'valid', otherwise
- * 'inactive'. The 'superseded' label is not derivable here — the endpoint does not
- * expose the CLA group's current version — so it is not produced in M1.
+ * Status and reason are copied from the producer. Do not recompute from
```
dealako: [nit] Doc comment contradicts the ICLA branch it documents

Issue: the comment states "Status and reason are copied from the producer. Do not recompute from `approved`/`valid`." but the ICLA branch immediately below (L120) does recompute: `status = cla.approved === true ? 'valid' : 'invalidated'`.
Proof: L103 comment vs L115–121 code; the ICLA collapse is intentional and tested (cla.service.spec.ts), so the flat "do not recompute" directive misdescribes it.
Why it matters: a maintainer could trust the wrong invariant. Purely a wording fix.
Fix: scope the directive to ECLA, e.g. "ECLA status/reason are copied verbatim from the producer. ICLA never surfaces `needs_attention`/`unknown`: those collapse to `valid`/`invalidated` based on `approved`, and its reason is dropped."

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -113,6 +112,15 @@ export function toMyClaAgreement(cla: EasyClaMyCla): MyClaAgreement {
       ? `${cla.documentMajorVersion}${cla.documentMinorVersion !== undefined ? `.${cla.documentMinorVersion}` : ''}`
       : undefined;
 
+  let status: ClaStatus = cla.status as ClaStatus;
```
dealako: [question] No runtime fallback for an out-of-contract `status`

Issue: `let status: ClaStatus = cla.status as ClaStatus;` trusts the wire value via a cast. `claStatusLabel`/`claStatusSeverity` are exhaustive switches with no `default`, so a value outside the five `ClaStatus` members returns `undefined` → the template (which only special-cases `'unknown'`) renders an empty/mis-styled pill rather than the `—` degrade.
Proof: `EasyClaMyCla.status` is typed required now (good — catches drift at compile time), but the cast bypasses runtime validation; cla-view.utils.ts has no default branch.
Why it matters: low likelihood (upstream always computes a status), so raising as a question rather than a defect — is coercing an unrecognized wire status to `'unknown'` in `toMyClaAgreement` (plus a test) worth the safety, or is the producer contract considered hard enough to rely on the cast?

dealako (dismissed):
Hey @ahmedomosanya 👋 — reviewed the M2 three-state "My CLAs" slice. This is clean, well-documented work: dropping the valid-only filter so every signed row is retained is the right call, the ICLA `needs_attention`/`unknown` → `valid`/`invalidated` collapse is deliberate and tested, the identity/impersonation and `skippedIdentities` handling is sound, and the `menuItemsMap` precompute (with the PrimeNG first-click comment) shows you already know the pattern. No security or data-privacy concerns — identity keys come only from the trusted session, logging is counts-only, and no PII leaks into logs/errors/URLs.

A couple of consistency/UX minors worth folding in before this trunk grows, plus a nit and a question. Since this is the WIP trunk (do-not-merge), approving on code quality — the process gate still holds.

**Issue count**
- 🔴 Blocking: 0
- 🟡 Minor: 2 — (1) status label/severity/icon/note are inline template function calls (frontend-checklist §4 "no template functions"); (2) ICLA rows with `pdfAvailable === false` render a ⋮ that opens an empty popup.
- ⚪ Nit: 1 — `toMyClaAgreement` doc says "Do not recompute from approved/valid" but the ICLA branch does exactly that.
- ❔ Question: 1 — `cla.status as ClaStatus` cast has no runtime fallback; an out-of-contract wire value would yield an empty pill.

**Bot reconciliation:** Copilot's two suppressed findings (empty ⋮ menu, no-template-functions) are legitimate — I've re-raised both with proof against the current code rather than leaving them suppressed. Copilot's earlier `status`-required suggestion was already addressed in `8c1a6d2`.

**Decision: ✅ Approved with minor comments.**

dealako:
### Review summary — @ahmedomosanya

Reviewed the M2 three-state **My CLAs** slice (parallel security / correctness / style pass, reconciled with Copilot). Solid, well-documented work — dropping the valid-only filter to retain every signed row is correct, the ICLA status collapse is deliberate + tested, and identity/impersonation/`skippedIdentities` handling is sound. **No security or data-privacy findings** (session-only identity keys, counts-only logging, no PII in logs/errors/URLs).

- 🔴 **Blocking:** 0
- 🟡 **Minor:** 2 — inline status template-function calls (frontend-checklist §4); ICLA rows with `pdfAvailable === false` open an empty ⋮ menu
- ⚪ **Nit:** 1 — `toMyClaAgreement` doc says "Do not recompute from approved/valid" but the ICLA branch does
- ❔ **Question:** 1 — `cla.status as ClaStatus` cast has no runtime fallback for an out-of-contract value

**Bot reconciliation:** Copilot's two *suppressed* findings (empty ⋮ menu, no-template-functions) are legitimate — re-raised both inline with proof against current code. Its earlier `status`-required note was already fixed in `8c1a6d2`.

**Decision: ✅ Approved with minor comments.** Details are in the [formal review](https://github.com/linuxfoundation/lfx-self-serve/pull/1440#pullrequestreview-4941404756). WIP trunk / do-not-merge gate still applies.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -45,6 +49,25 @@ export function claServiceBaseUrl(): string {
...
+ * Derived from the request rather than configured, so preview deployments (per-PR hostnames)
+ * are correct without templating a fourth environment value. `trust proxy` is set, so
+ * `protocol`/`host` already reflect the forwarded headers behind the ingress.
+ */
+export function claReturnUrl(req: Request): string {
+  const host = req.get('host');
+  if (!host) {
+    throw new MicroserviceError('Cannot derive the CLA return URL: request has no Host header', 500, 'RETURN_URL_UNRESOLVABLE', { service: SERVICE });
+  }
+  return `${req.protocol}://${host}${MY_CLAS_PATH}`;
```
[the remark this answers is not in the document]
ahmedomosanya: Fixed in dc6d94f91. The host is now validated before it is handed onward: an allowlist of our own origins (`app.lfx.dev`, `app.dev.lfx.dev`, `localhost`, `127.0.0.1`) plus an anchored pattern for per-PR previews (`ui-pr-<n>.dev.v2.cluster.linuxfound.info`), and the protocol must be http or https. Anything else throws `RETURN_URL_UNTRUSTED` rather than returning a URL.

Pinned in code rather than an allowlist env var because this feature adds no new configuration, and the per-PR previews are the reason the host cannot just be read from one configured base URL. The Host is parsed with `new URL(...)` rather than string-split, so a port (previews use one), an IPv6 literal, or a `user@host` form resolves correctly instead of smuggling another origin through. Tests cover a foreign host, two look-alikes (`app.lfx.dev.evil.example.com`, `notapp.lfx.dev`), a preview-shaped host on the wrong domain, the userinfo form, and a non-http protocol.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -279,6 +309,117 @@ export class ClaService {
...
+    }
+
+    if (userIds.length > 1) {
+      logger.warning(req, 'cla_resolve_contributor_id', 'identity matches several EasyCLA user records; handing off the first', {
+        matched_count: userIds.length,
+      });
+    }
+
+    logger.success(req, 'cla_resolve_contributor_id', startTime);
+    return userIds[0] as string;
```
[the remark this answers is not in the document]
ahmedomosanya: Deferred deliberately, tracked in #1250's sibling #1365 (disambiguating a contributor with several EasyCLA user records).

The concern is real and stated in the code and PR body rather than hidden: `userIds` ordering carries no canonical-identity guarantee. Not failing closed today is a deliberate trade-off, because this branch is a `do-not-merge` WIP trunk that cannot reach production, and failing closed would block the flow entirely for the accounts we can test with — the dev account used to verify this end to end matches three records (one on `lf_username`, two on email/githubId), so N > 1 is the normal case here, not the exception.

The first-of-N bridge exists because `GET /v4/user-from-token` is an unauthenticated gateway passthrough and never receives the `X-ACL` header the CLA backend authenticates on, so it 401s for every caller. Fail-closed becomes the right default once one authoritative ID is available — which is exactly what #1365 has to settle, and it needs a product decision about which record wins, not just a code change.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.html
ahmedomosanya: Addressed in `f4785a051` — status label, severity, icon, and note are precomputed on the `rows` view model; the template only reads fields.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.html
ahmedomosanya: Addressed in `f4785a051` — the ⋮ trigger renders only when `hasActions` is true, so an ICLA with no document no longer opens an empty menu.

apps/lfx-one/src/server/services/cla.service.ts
ahmedomosanya: Addressed in `6754c80e6` — the `toMyClaAgreement` comment now matches the ICLA branch: `approved` is consulted only for that out-of-contract collapse.

apps/lfx-one/src/server/services/cla.service.ts
ahmedomosanya: Addressed in `6754c80e6` — an unrecognised producer `status` is degraded instead of being cast through to the exhaustive label/severity switches.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -279,6 +522,124 @@ export class ClaService {
...
+    // session's accounts being unreadable.
+    const { accounts } = await this.listGithubAccounts(req);
+    const chosen = accounts.find((account) => account.githubId === githubId);
+    if (!chosen) {
+      throw new MicroserviceError('The chosen GitHub account is not linked to this session', 403, 'CLA_ACCOUNT_NOT_LINKED', { service: SERVICE });
+    }
+
+    let result: EasyClaSigningIdentity | null;
+    try {
+      result = await gatewayFetch<EasyClaSigningIdentity>(req, `${claServiceBaseUrl()}/v4/my-clas/signing-identity`, {
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `8f64d512e` — the hand-off now calls `POST /v4/self-serve/prepare-sign` and uses the producer `signUrl` rather than composing a Console URL here.

apps/lfx-one/src/server/routes/clas.route.ts
```diff
@@ -13,4 +14,12 @@ const clasController = new ClasController();
 router.get('/clas', (req, res, next) => clasController.getMyClas(req, res, next));
 router.get('/clas/:signatureId/pdf-url', (req, res, next) => clasController.getPdfUrl(req, res, next));
 
+// Sign CLA hand-off (#1251). Selection is a read and stays available while impersonating; the
+// hand-off is guarded, because starting a signature is a real, externally-visible act that would
+// be attributed to the target rather than the impersonator who performed it. Registered before
+// `/clas/:signatureId/pdf-url` would not matter (distinct final segments), but both sign routes
+// are kept together so the guard is visible next to what it protects.
+router.get('/clas/sign-options', (req, res, next) => clasController.getClaGroupOptions(req, res, next));
+router.get('/clas/sign-handoff', blockDuringImpersonation, (req, res, next) => clasController.getSignHandoff(req, res, next));
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `8f64d512e` — signing no longer goes through `getSignHandoff`. The write is `POST /v4/self-serve/prepare-sign`.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -113,6 +186,19 @@ export function toMyClaAgreement(cla: EasyClaMyCla): MyClaAgreement {
...
+  let status: ClaStatus = wireStatus ?? 'unknown';
+  // A reason qualifies the status it shipped with, so an out-of-contract status invalidates it too;
+  // keeping it would pair an em dash with a sentence explaining a coverage miss.
+  let statusReason = wireStatus === null ? undefined : cla.statusReason;
+  if (isIcla) {
+    statusReason = undefined;
+    if (status === 'needs_attention' || status === 'unknown') {
+      status = cla.approved === true ? 'valid' : 'invalidated';
+    }
+  }
```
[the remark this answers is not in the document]
ahmedomosanya: Convention: ICLA has no coverage dimension, so `needs_attention`/`unknown` cannot describe one. `6754c80e6` documents the only `approved` consult — collapsing that out-of-contract ICLA to valid/invalidated — and does not restore a general approved/valid fallback.

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  /** Set while writing the chosen project's name back into the field, so it is not re-searched. */
+  private suppressNextEmit = false;
+
+  public constructor() {
+    this.search$
+      .pipe(
+        debounceTime(CLA_GROUP_SEARCH_DEBOUNCE_MS),
+        tap(() => {
+          this.loading.set(true);
```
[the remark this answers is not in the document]
ahmedomosanya: Convention: prior results stay visible while the next search is in flight (emptying the list on each keystroke was worse). `switchMap` cancels the in-flight HTTP; the highlight is dropped so Enter cannot confirm a stale row.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  protected onSend(): void {
+    if (!this.canSend()) return;
+
+    // Contact is a product-complete no-op: the producer has no contact requestType, and posting
+    // approval/removal would tell the manager to change the Approved List.
+    if (this.data.mode === 'contact') {
+      this.ref.close(null);
+      return;
```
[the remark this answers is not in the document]
ahmedomosanya: Convention: contact Send is a deliberate no-op (`bad552061`). The producer has no `contact` requestType; posting `approval` or `removal` would tell the manager to change the Approved List. Approval and removal Send still POST.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -97,4 +222,222 @@ export class ProfileClasComponent {
...
+   * contributor had already signed by the time anyone knew as whom.
+   */
+  private handOffToConsole(option: ClaGroupOption): void {
+    // Already flagged as in flight by `openSignDialog`, which owns the guard because the window
+    // worth guarding opens with the picker rather than with this call.
+    this.myClasService
+      .getGithubAccounts()
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe({
+        next: ({ accounts }) => this.chooseAccountThenSign(option, accounts),
```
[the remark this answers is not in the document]
ahmedomosanya: Deferring with #1601 — GitHub-only preflight for Gerrit/GitLab groups is the remaining account-step gap. Not changing that on this review pass.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -97,4 +222,231 @@ export class ProfileClasComponent {
...
+      this.sendToAccountLinking();
+      return;
+    }
+
+    if (accounts.length === 1) {
+      this.prepareThenHandOff(option, accounts[0]);
+      return;
+    }
+
+    const dialogRef = this.dialogService.open(GithubAccountSelectComponent, {
```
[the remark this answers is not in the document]
ahmedomosanya: Deferring with #1601 — one linked account still skips the picker and prepares directly. That is the current product choice on this branch (nothing to choose between); #1601 is the place to require the picker at one account.

apps/lfx-one/src/server/routes/clas.route.ts
```diff
@@ -13,4 +14,16 @@ const clasController = new ClasController();
...
+// impersonating.
+router.get('/clas/sign-options', (req, res, next) => clasController.getClaGroupOptions(req, res, next));
+
+// GitHub account selection before the hand-off (#1252). Listing the linked accounts is a
+// read and stays available while impersonating; preparing the signing session is guarded, and
+// more plainly than the hand-off is — this route causes an identity attribute and a signing
+// session to be written against an EasyCLA record, which is literally the hard-to-retract,
+// externally-visible act with no in-payload caller identity that the read-only rule exists for.
+router.get('/clas/github-accounts', (req, res, next) => clasController.getGithubAccounts(req, res, next));
+router.post('/clas/prepare-sign', blockDuringImpersonation, (req, res, next) => clasController.prepareSign(req, res, next));
```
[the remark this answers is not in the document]
ahmedomosanya: Deferring — prepare-sign is the producer contract from easycla#5150 / #1252. Self Serve sends the chosen linked account; the write lives on that endpoint, not a Self Serve identity bind. Broader Decision 1 wording is the #1252 / easycla#5144 review, not this thread.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -279,6 +468,117 @@ export class ClaService {
...
+   * impersonating, because a signature must never be attributed to the wrong person.
+   */
+  public async resolveContributorId(req: Request): Promise<string> {
+    const startTime = logger.startOperation(req, 'cla_resolve_contributor_id');
+
+    const resolvedOrCreated = await this.tryResolveOrCreateContributor(req);
+    if (resolvedOrCreated) {
+      logger.success(req, 'cla_resolve_contributor_id', startTime);
+      return resolvedOrCreated;
+    }
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `8f64d512e` — the lookup-or-create `/v2/user-from-token` path is gone. Signing goes through `POST /v4/self-serve/prepare-sign` with the chosen linked GitHub account.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -38,13 +65,170 @@ const MAX_CLA_EMAILS = 100;
...
+ * unchecked forged Host would turn a trusted hand-off into an open redirect.
+ */
+export function claReturnUrl(req: Request): string {
+  const host = req.get('host');
+  if (!host) {
+    throw new MicroserviceError('Cannot derive the CLA return URL: request has no Host header', 500, 'RETURN_URL_UNRESOLVABLE', { service: SERVICE });
+  }
+
+  if (req.protocol !== 'http' && req.protocol !== 'https') {
+    throw new MicroserviceError('Cannot derive the CLA return URL: unsupported protocol', 500, 'RETURN_URL_UNTRUSTED', { service: SERVICE });
```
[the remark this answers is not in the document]
ahmedomosanya: Convention: `claReturnUrl` must emit `http://localhost:4200` for local `yarn start`; preview and production hosts are https. The producer https-only rule is a local-dev limitation, not a prod hand-off bug. Trusted-host checks still reject a forged Host.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -38,13 +66,170 @@ const MAX_CLA_EMAILS = 100;
...
+/**
+ * Whether the CLA service listed the chosen account among the identity keys it did *not* apply.
+ *
+ * A prepare that skipped the pick still succeeds — it opened a session for whatever identity it
+ * did verify. Treating that as success for the skipped account would sign against a different
+ * one, so it is a failure here even though the status was 200.
+ */
+export function githubIdWasSkipped(skippedIdentities: readonly string[] | undefined, githubId: string): boolean {
+  return (skippedIdentities ?? []).includes(`${GITHUB_ID_KEY_PREFIX}${githubId}`);
+}
```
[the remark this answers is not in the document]
ahmedomosanya: Convention: `skippedIdentities` uses the same `<type>:<value>` keys as `identity` (`github-id:`), which is what `githubIdWasSkipped` matches. The tests pin that form. A distinct `githubId:` rejected-field format would be a producer-contract follow-up, not a silent second prefix here.

apps/lfx-one/src/server/controllers/clas.controller.ts
```diff
@@ -67,4 +73,191 @@ export class ClasController {
...
+        res.status(400).json({ message: 'A request type of approval or removal is required' });
+        return;
+      }
+
+      const rawRecipients = Array.isArray(body?.recipients) ? body.recipients : [];
+      const recipients = rawRecipients.map((value) => String(value ?? '').trim()).filter((value) => value.length > 0);
+      if (recipients.length === 0) {
+        res.status(400).json({ message: 'At least one CLA manager is required' });
+        return;
+      }
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `16f72fbdf` — `createClaManagerRequest` now 400s the whole payload unless every recipient is already a non-empty string, then trims only those values.

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
```diff
@@ -0,0 +1,268 @@
...
+      )
+      .subscribe((response) => {
+        this.loading.set(false);
+        if (!response) return;
+        this.options.set(response.results);
+        this.truncated.set(response.truncated);
+        // The highlight is a position in the previous list; carrying it over would let Enter
+        // confirm whichever project happens to land at that offset in the new one.
+        this.highlightedIndex.set(-1);
+      });
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `16f72fbdf` — shortening below three characters or a failed search now clears `options`, and a keystroke drops the highlight immediately so Arrow/Enter cannot confirm a row the template is no longer drawing.

ahmedomosanya:
Copilot's three suppressed comments on `23d7d2664`:

1. Reset the keyboard highlight as soon as the query changes — addressed in `16f72fbdf` (same change as the Bugbot thread).
2. Combobox ARIA (`role`, `aria-controls`, `aria-activedescendant` on the focused input) — not in this pass. `lfx-input-text` has no way to forward those attributes onto the native input; doing it is a shared-wrapper change. Keyboard confirm of a hidden row is fixed; the AT association is not.
3. `buildConsoleHandoffUrl` — removed in `16f72fbdf` (export, util, and spec). The sign path uses the producer's `PrepareSignResponse.signUrl`.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
```diff
@@ -0,0 +1,54 @@
...
+    <p class="mb-2 text-sm font-medium text-slate-700">CLA manager(s)</p>
+    <div class="flex flex-col gap-2" data-testid="contact-cla-manager-list">
+      @for (manager of managers(); track manager.lfUsername) {
+        <label class="flex cursor-pointer items-center gap-2 rounded border border-slate-200 px-3 py-2 text-sm">
+          <input
+            type="checkbox"
+            class="size-4"
+            [checked]="isSelected(manager.lfUsername)"
+            [attr.data-testid]="'contact-cla-manager-' + manager.lfUsername"
+            (change)="toggleManager(manager.lfUsername, $event)" />
```
audigregorie: **Critical — Raw `<input type="checkbox">` instead of the `lfx-checkbox` wrapper**

Frontend checklist §14.1: raw form elements in feature templates must use the LFX wrappers, and `CheckboxComponent` (`@components/checkbox/checkbox.component`) exists for exactly this. Beyond convention, the wrapper enforces the reactive-forms contract this repo standardizes on.

The fix moves per-manager selection from the `selected` Set into the existing `form` — one `FormControl<boolean>` per manager keyed by `lfUsername`, added where `managers` is populated in the constructor subscription — so the template becomes `<lfx-checkbox [form]="form" [control]="manager.lfUsername" [label]="..." />` and `canSend` reads the form value. (Pairs naturally with the template-functions thread on the label below.)

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.html
```diff
@@ -0,0 +1,164 @@
...
+            <div class="border-b border-slate-100 last:border-b-0">
+              <div
+                role="option"
+                [id]="'cla-group-option-' + option.claGroupId"
+                [attr.aria-selected]="selected()?.claGroupId === option.claGroupId"
+                class="cursor-pointer px-4 py-2.5 text-left hover:bg-slate-50"
+                [class.bg-slate-50]="highlightedIndex() === index"
+                [attr.data-testid]="'cla-group-select-' + option.claGroupId"
+                (click)="onSelect(option)">
+                <span class="block truncate text-sm font-medium text-slate-900">{{ primaryName(option) }}</span>
```
audigregorie: **Major — Per-row method calls in the template**

Frontend checklist #4: templates bind signal reads, computed values, and pipes — no method calls that execute logic. This template calls `primaryName(option)`, `secondaryName(option)`, `matchTypeLabel(...)`, `orgSourceLabel(...)`, `orgSourceIcon(...)`, and `orgsExpanded(option)` per row, re-running on every change-detection pass.

The parent component in this same PR already solves exactly this problem (see the `rows` computed in `profile-clas.component.ts`, with the rationale in its doc comment). Map `response.results` to a display view model in the `search$` subscription — primary/secondary names, match-type labels, org labels + icons precomputed — and bind fields instead. `orgsExpanded` can become a per-row `expanded` flag on that view model.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
```diff
@@ -0,0 +1,54 @@
...
+          <span>{{ managerLabel(manager) }}</span>
```
audigregorie: **Major — Template method calls: `managerLabel(manager)` (and `isSelected(manager.lfUsername)` on the input above)**

Frontend checklist #4 — no method calls in templates. Precompute the label (`name || lfUsername`) when `managers` is set in the constructor subscription, and if the `lfx-checkbox` thread is applied, selection state moves into the form and `isSelected(...)` disappears entirely.

apps/lfx-one/src/app/modules/profile/clas/github-account-select.component.html
```diff
@@ -0,0 +1,33 @@
...
+    covered by it.
+  </p>
+
+  <div role="listbox" aria-label="Linked GitHub accounts" class="flex flex-col gap-2">
+    @for (account of accounts(); track account.githubId) {
+      <lfx-selectable-card
+        [form]="selectForm"
+        control="githubId"
+        [value]="account.githubId"
+        [label]="accountLabel(account)"
```
audigregorie: **Major — Template method call: `accountLabel(account)`**

Frontend checklist #4. The fallback (`githubUsername || 'GitHub account ' + githubId`) can be computed once when the `accounts` signal is initialized, e.g. `signal((this.config.data?.accounts ?? []).map((a) => ({ ...a, label: a.githubUsername || `GitHub account ${a.githubId}` })))`, and the template binds `[label]="account.label"`.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.ts
```diff
@@ -0,0 +1,158 @@
...
+    this.ref.close(null);
+  }
+
+  protected onSend(): void {
+    if (!this.canSend()) return;
+
+    // Contact is a product-complete no-op: the producer has no contact requestType, and posting
+    // approval/removal would tell the manager to change the Approved List.
+    if (this.data.mode === 'contact') {
+      this.ref.close(null);
```
audigregorie, line 102: **Major — Contact mode: Send silently discards the typed message and closes**

In `contact` mode the user reads a hint saying “Send a message to the CLA manager(s)…”, types a note, clicks **Send** — and the dialog closes with no toast, no confirmation, and nothing sent. `Send` and `Cancel` are literally the same code path (`this.ref.close(null)`). The no-op itself is documented as deliberate (the producer has no `contact` request type), but the silent version is a UI-truthfulness problem: the screen promises an action it does not perform, and the user loses text they wrote.

Cheapest honest version: show an info toast before closing ("Contacting CLA managers isn't available yet — no message was sent"). Better: don't offer free text + Send in this mode at all (disabled Send with a one-line explanation, or omit the Contact kebab item until the producer ships the type).

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
```diff
@@ -0,0 +1,284 @@
...
+
+  private readonly search$ = new Subject<string>();
+
+  /** Set while writing the chosen project's name back into the field, so it is not re-searched. */
+  private suppressNextEmit = false;
+
+  public constructor() {
+    this.search$
+      .pipe(
+        debounceTime(CLA_GROUP_SEARCH_DEBOUNCE_MS),
```
audigregorie, line 71: **Major — Identical terms re-run the producer search on every refocus**

There is no `distinctUntilChanged` after the debounce, and `onFocus()` calls `pushSearch` unconditionally, so every `focusin` on the container (returning to the field after interacting with a result, after Retry, after tabbing away and back) re-issues the same term to `GET /v4/cla-group/search` — which this PR's own template comment describes as a five-table scan bounded at 15 seconds.

A plain `distinctUntilChanged()` after the `map` would break `retry()` (an unchanged term would be swallowed after an error), so gate the focus path instead: skip `pushSearch` in `onFocus` when the trimmed field value equals the current `query()` and results are already present, or track a `lastIssuedTerm` that `retry()` can bypass.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -67,13 +132,88 @@ export class ProfileClasComponent {
   protected readonly agreements = computed<MyClaAgreement[]>(() => this.state().data?.agreements ?? []);
   protected readonly isEmpty = computed(() => isMyClasEmpty(this.state().loaded, this.state().error, this.agreements().length));
 
+  /**
+   * The table's only binding source. Recomputes when the agreement list changes and not
+   * otherwise, which is also what keeps each row's `MenuItem[]` referentially stable —
+   * a fresh model per change-detection pass makes the PrimeNG popup miss the first click.
+   */
+  protected readonly rows = computed<ClaRow[]>(() => {
```
audigregorie: **Major — Multi-line `computed()` / `toSignal()` initialized inline instead of via private `init*()` functions**

`component-organization.md` §3 (also on the CLAUDE.md top-violations list): complex computed/toSignal signals go through private initializer functions — `private readonly state = this.initState();` and `protected readonly rows = this.initRows();`, with the `init*()` bodies grouped at the bottom of the class (structure order §4). Applies to `state` (around line 110, outside this diff hunk) and `rows` here; same for `queryBand` in `cla-group-select.component.ts`.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -1,41 +1,106 @@
...
+import { FeatureFlagService } from '@services/feature-flag.service';
 import { MyClasService } from '@services/my-clas.service';
+import { UserService } from '@services/user.service';
+
+import { ClaGroupSelectComponent } from './cla-group-select.component';
+import { buildContactClaManagerMenuItems } from './contact-cla-manager-menu';
+import { GithubAccountSelectComponent } from './github-account-select.component';
+
+/** Precomputed status cell for one row. */
+interface ClaRowStatus {
```
audigregorie: **Major — Local interfaces (`ClaRowStatus`, `ClaRow`) and a module-level const in a component file**

`component-organization.md` §5: *all* interfaces, even component-specific ones, live in `@lfx-one/shared/interfaces`; CLAUDE.md source hygiene: no module-level consts inside `apps/lfx-one/` (`ECLA_COVERED_DOWNLOAD_LABEL`, line 60). The shared package already references PrimeNG types (`crowdfunding.interface.ts`, `components.interface.ts`), so `ClaRow`'s `MenuItem[]` field is not a blocker — these view-model types sit naturally next to the other CLA UI shapes in `packages/shared/src/interfaces/cla.interface.ts`.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.ts
```diff
@@ -0,0 +1,158 @@
...
+import { MessageService } from 'primeng/api';
+import { DynamicDialogConfig, DynamicDialogRef } from 'primeng/dynamicdialog';
+import { take } from 'rxjs';
+
+import { ButtonComponent } from '@components/button/button.component';
+import { MessageComponent } from '@components/message/message.component';
+import { TextareaComponent } from '@components/textarea/textarea.component';
+import { MyClasService } from '@services/my-clas.service';
+
+export interface ContactClaManagerDialogData {
```
audigregorie: **Major — Exported interface and constant live in a component file**

`ContactClaManagerDialogData` and `CLA_MANAGER_MODAL_COPY` are exported from a component and consumed by `contact-cla-manager-menu.ts` — i.e. shared across files, so they are not "truly local UI-only" types. Per `component-organization.md` §5 / shared checklist #1-2, the interface belongs in `packages/shared/src/interfaces/cla.interface.ts` and the copy constant in `packages/shared/src/constants/cla.constants.ts`.

audigregorie (commented):
## Code Review Summary

Strong, well-tested dark launch — the hand-off security work (return-URL host pinning, recorded-account verification, session-sourced identity) is exactly right. Findings: **1 Critical** (raw checkbox instead of the `lfx-checkbox` wrapper), **8 Major** (per-row template method calls in the three new dialogs — inconsistent with this PR's own precomputed row view model — a silent no-op Send in Contact mode, a refocus-triggered duplicate producer search, and component-organization rule breaks), **5 Minor** line comments, plus the items below.

### Outside the diff

- **Minor — `apps/lfx-one/src/server/controllers/clas.controller.ts` (`getPdfUrl`)** — the only parameterized route that does not format-validate `signatureId`; `getClaManagers` and `createClaManagerRequest` both do. Low risk (EasyCLA 404s unknown IDs), but inconsistent. Related naming nit: `CLA_GROUP_ID_PATTERN` is documented as a CLA-group UUID pattern but now guards signature IDs — a `UUID_PATTERN` rename (or alias) would read better.
- **FYI — `apps/lfx-one/src/app/shared/components/menu/menu.component.scss`** — global `lfx-menu` styling change (disabled items keep pointer events / `not-allowed` cursor, all items get `cursor: pointer`) affects every menu in the app to serve one kebab item. Low risk, but worth a callout in the PR description since it touches shared chrome.
- **Observation — UI-only flag** — `my-clas-m2-enabled` deliberately does not gate the BFF routes (documented in `feature-flags.constants.ts`), so a direct `POST /api/me/clas/prepare-sign` works with the flag off. Upstream authorization still applies, so this is acceptable — confirming it is the intended rollout posture.

### What's done well

- The open-redirect defense in `claReturnUrl` (host allowlist + anchored preview regex, parsed via `URL` rather than string-split) is textbook.
- The recorded-account double-check in `prepareSign` / `prepareThenHandOff` — comparing the identity the CLA backend verified against the one the contributor picked — closes a real "signed as the wrong account" hole that upstream ownership verification cannot catch.
- Impersonation model is exactly right: writes blocked at the route, reads swapped to the target's bearer token upstream.
- The `rows` computed precomputing per-row view models for referentially stable `MenuItem[]` (with the PrimeNG first-click rationale in the comment) is precisely the discipline the frontend checklist asks for — which makes the per-row method calls in the three new dialog templates the surprising inconsistency.
- Test coverage is thorough: flag on/off matrices in the component spec, an 800+-line service spec, and route-level impersonation tests.

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
```diff
@@ -0,0 +1,284 @@
...
+import type { ClaGroupMatchType, ClaGroupOption, ClaGroupOrgSource, ClaGroupSearchResponse } from '@lfx-one/shared/interfaces';
+import { MyClasService } from '@services/my-clas.service';
+import { DynamicDialogRef } from 'primeng/dynamicdialog';
+import { catchError, debounceTime, map, of, Subject, switchMap } from 'rxjs';
+
+import { ButtonComponent } from '@components/button/button.component';
+import { InputTextComponent } from '@components/input-text/input-text.component';
+
+/** Stands in for a result the producer could name neither by project nor by CLA group (FR-008). */
+const UNNAMED_CLA_GROUP = 'Unnamed CLA group';
```
audigregorie: **Minor — Module-level const maps in a component file**

`UNNAMED_CLA_GROUP`, `MATCH_TYPE_LABELS`, `ORG_SOURCE_LABELS`, and `ORG_SOURCE_ICONS` are module-level constants inside `apps/lfx-one/`; per CLAUDE.md source hygiene / shared checklist #2, constants belong in `packages/shared/src/constants/cla.constants.ts` (which this PR already creates).

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
```diff
@@ -0,0 +1,284 @@
...
+        this.options.set(response.results);
+        this.truncated.set(response.truncated);
+        // The highlight is a position in the previous list; carrying it over would let Enter
+        // confirm whichever project happens to land at that offset in the new one.
+        this.highlightedIndex.set(-1);
+      });
+
+    this.searchForm
+      .get('query')!
+      .valueChanges.pipe(takeUntilDestroyed())
```
audigregorie: **Minor — Non-null assertion on a control the class itself declares**

Checklist §14.5: `get('query')!` asserts non-null on a `FormGroup` literal defined two fields up, so the assertion can never actually fail — and the typed accessor needs no assertion at all:

```suggestion
    this.searchForm.controls.query.valueChanges.pipe(takeUntilDestroyed())
```

apps/lfx-one/src/app/modules/profile/clas/github-account-select.component.ts
```diff
@@ -0,0 +1,80 @@
...
+    githubId: new FormControl<string | null>(null),
+  });
+
+  protected readonly accounts = signal<GithubAccountOption[]>(this.config.data?.accounts ?? []);
+  protected readonly selectedId = signal<string | null>(null);
+
+  public constructor() {
+    this.selectForm
+      .get('githubId')!
+      .valueChanges.pipe(takeUntilDestroyed())
```
audigregorie: **Minor — Non-null assertion on a self-declared control**

Checklist §14.5 — same pattern as the sibling dialog; `controls.githubId` is typed and assertion-free:

```suggestion
    this.selectForm.controls.githubId.valueChanges.pipe(takeUntilDestroyed())
```

packages/shared/src/interfaces/cla.interface.ts
```diff
@@ -9,14 +9,31 @@ export type ClaKind = 'ICLA' | 'ECLA';
...
+ * - `invalidated`     — the stored signature-approval flag is false or absent. This is a
+ *                       different state from `revoked` and must never share its label: a CLA
+ *                       manager removing someone from an approved list, a project manager
+ *                       invalidating an ICLA from PCC, and a deleted CLA group all land here,
+ *                       and none of them is a sanctions case.
+ * - `unknown`         — coverage could not be evaluated (ECLA only; ICLA never produces this).
+ *                       Rendered as plain-text "—", not a named pill.
+ * - `superseded`      — reserved: an older document version than the CLA group's current.
+ *                       Not produced today (the my-clas endpoint does not expose the
+ *                       current version); kept for forward compatibility. Do not render it.
```
audigregorie: **Minor — `superseded` doc says "Do not render it", but the code renders it**

This comment instructs consumers not to render `superseded`, yet the BFF forwards it (`KNOWN_CLA_STATUSES` in `cla.service.ts`) and the UI renders a full labeled pill — `claStatusLabel` returns `'Superseded'`, `claStatusSeverity` returns `'warn'`, and `statusIcon` maps a `fa-clock-rotate-left`. Either drop the instruction (the implementation chose to render it) or actually filter `superseded` at the mapper; as written the contract and the behavior disagree.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -38,13 +72,170 @@ const MAX_CLA_EMAILS = 100;
...
+  return `${req.protocol}://${host}${MY_CLAS_PATH}`;
+}
+
+// Identity keys the CLA service reports as `"<type>:<value>"`, both in the verified `identity`
+// list and in `skippedIdentities`.
+const GITHUB_ID_KEY_PREFIX = 'github-id:';
+const GITHUB_USERNAME_KEY_PREFIX = 'github-username:';
+
+/** The GitHub account the CLA service verified, recovered from its identity keys. */
+export interface RecordedGithubIdentity {
```
audigregorie: **Minor — `RecordedGithubIdentity` exported from the service file**

The sibling CLA wire/view types all live in `src/server/types/cla.types.ts`; this exported interface (consumed by the spec) belongs there too, keeping the service file to logic plus pure functions.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
ahmedomosanya: Addressed in `69795974e` — manager selection now lives on `lfx-checkbox` bound to a per-manager FormControl; `canSend` reads the form.

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.html
ahmedomosanya: Addressed in `69795974e` — search results map to `ClaGroupOptionView` (primary/secondary names, match-type labels, org labels/icons, expanded) before the template binds.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
ahmedomosanya: Addressed in `69795974e` — labels are precomputed on `ClaManagerView`; selection is the form, so `managerLabel` / `isSelected` are gone.

apps/lfx-one/src/app/modules/profile/clas/github-account-select.component.html
ahmedomosanya: Addressed in `69795974e` — accounts are mapped to `GithubAccountChoice` with `label` at init; the template binds `account.label`.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.ts
ahmedomosanya: Addressed in `69795974e` — Contact Send now shows an info toast ("Contacting CLA managers isn't available yet — no message was sent") before closing.

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
ahmedomosanya: Addressed in `69795974e` — `onFocus` skips `pushSearch` when the trimmed term matches the last issued query and results are already present; Retry still calls `pushSearch` directly.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
ahmedomosanya: Addressed in `69795974e` — `state` and `rows` go through `initState`/`initRows`; picker `queryBand` through `initQueryBand`.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
ahmedomosanya: Addressed in `69795974e` — `ClaRow` / `ClaRowStatus` live in `@lfx-one/shared/interfaces`; `ECLA_COVERED_DOWNLOAD_LABEL` in `cla.constants.ts`.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.ts
ahmedomosanya: Addressed in `69795974e` — `ContactClaManagerDialogData` is in `cla.interface.ts`; `CLA_MANAGER_MODAL_COPY` is in `cla.constants.ts`.

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
ahmedomosanya: Addressed in `69795974e` — unnamed-group, match-type, and org-source maps moved to `cla.constants.ts`.

apps/lfx-one/src/app/modules/profile/clas/cla-group-select.component.ts
ahmedomosanya: Addressed in `69795974e` — uses `this.searchForm.controls.query.valueChanges`.

apps/lfx-one/src/app/modules/profile/clas/github-account-select.component.ts
ahmedomosanya: Addressed in `69795974e` — uses `this.selectForm.controls.githubId.valueChanges`.

packages/shared/src/interfaces/cla.interface.ts
ahmedomosanya: Addressed in `69795974e` — the `superseded` contract now says the UI renders a labeled pill if that status arrives.

apps/lfx-one/src/server/services/cla.service.ts
ahmedomosanya: Addressed in `69795974e` — `RecordedGithubIdentity` moved to `cla.types.ts`.

audigregorie (approved):

mlehotskylf (approved):

PR #1442, opened 2026-08-13
feat(marketing): persist brand kit docs to object storage (LFXV2-2059)

joanreyero:
## Summary

Implements the Brand Kit persistence write path (dec-brand-kit-storage-v2, LFXV2-2059) — **stacked on #1348** (`feat/LFXV2-2059-brand-kit-ui`); this PR targets that branch and becomes a trivial rebase onto `main` once #1348 merges.

- **Extends the existing `ObjectStoreService`** (from PR #1282) with purpose-keyed bucket config (`avatars` → `S3_BUCKET`, `marketing-os-artifacts` → `MARKETING_OS_ARTIFACTS_S3_BUCKET`) and a generic `putObjectIfAbsent` (HEAD-then-PUT; only confirmed 404 counts as absent; benign race for content-addressed keys). Avatar path is untouched via defaulted parameters — all existing tests pass unchanged.
- **Persists the validated Brand Kit document** from the existing result-poll flow: on successful envelope validation the raw `document_markdown` bytes are written under the content-addressed key `brand-kit/{project}/{sha256}.md` with `CacheControl: private` to the shared private marketing artifacts bucket.
- **Response carries a persistence receipt** (`s3_key`, `content_sha256`, `project`, `version`, `intake_mode`) for later ai-brain Artifact minting — graph writes stay deferred behind wi-lfx-one-service-actor (A5 deferral).
- **Graceful degrade:** a storage failure still returns the validated document (no receipt); the content-addressed write retries idempotently on the next poll.

## Design notes

- The BFF-owned S3 write is the lfx-object-store-design skill's documented exception path (no owning backend resource API for marketing artifacts); governed by dec-brand-kit-storage-v2.
- One shared private bucket `marketing-os-artifacts-{env}` for all marketing artifact types, key-prefix namespaced; env contract is domain-prefixed (`MARKETING_OS_ARTIFACTS_S3_BUCKET`) since lfx-one is multi-bucket.
- **Deployment wiring is out of scope:** `MARKETING_OS_ARTIFACTS_S3_BUCKET` provisioning for deployed envs is tracked separately (wi-brand-kit-bucket-def-amend / object-store-ops). If unset, persistence degrades gracefully (logged, no receipt).
- `S3_ENDPOINT_URL` / `S3_CREATE_MISSING_BUCKET` stay process-wide (both buckets share one endpoint/mode today); purpose-namespaced variants can follow if that ever diverges.
- Receipt (incl. `s3_key`) is intentionally returned to the client — derived only from values the client already has (project slug + sha), bucket is private.

## Testing

- New unit coverage: `putObjectIfAbsent` (exists-skip / write / non-404 rethrow / PUT failure), purpose-keyed buckets (env resolution, missing-env reject, per-purpose memoization non-poisoning), persist receipt shape, graceful degrade, `buildBrandKitObjectKey` traversal rejection.
- Full suites green: lfx-one server 1109 tests, shared 819 tests; `yarn build`, `yarn lint:check`, `yarn format:check`, `check-headers.sh` pass.
- Live smoke against local MinIO (`marketing-os-artifacts-dev`, versioned): first write `true`, repeat poll no-op `false` at the content-addressed key.

Ref: LFXV2-2059


apps/lfx-one/src/server/services/brand-kit.service.ts
```diff
@@ -53,16 +63,74 @@ export class BrandKitService {
...
+        project: envelope.project,
+        version: envelope.version,
+        intake_mode: envelope.intake.mode,
+      };
+    } catch (error) {
+      // Deliberate degrade: surface the validated document even when storage is
+      // down; the content-addressed write is retried on the next poll.
+      logger.error(req, 'brand_kit_persist', startTime, error, { project: envelope.project, version: envelope.version });
+      return null;
+    }
```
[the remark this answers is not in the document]
joanreyero: Fixed in 642b31e2: the client no longer stops polling on the first `ready` response when the persistence receipt is missing — it keeps polling a bounded number of extra times (`PERSIST_RETRY_MAX_ATTEMPTS = 3`), and each poll re-runs the idempotent content-addressed server-side write, so transient storage outages are retried. The document is displayed immediately regardless; the retries are background-only and never surface errors to the user. Locked by tests in f3e48b89.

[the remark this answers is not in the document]
joanreyero: Fixed in 642b31e2: the persistence-failure degrade path now logs at WARN via `logger.warning` (with duration and error message in the payload), per the graceful-degradation rule in `.claude/rules/logging-patterns.md` — no more ERROR noise while the bucket env var is intentionally absent in deployed environments. A spec in f3e48b89 locks WARN-not-ERROR.

apps/lfx-one/src/server/services/brand-kit.service.ts
```diff
@@ -53,16 +63,74 @@ export class BrandKitService {
...
+    const persistence = await this.persistEnvelope(req, envelope);
+
     return {
       status: 'ready',
       documentMarkdown: envelope.document_markdown,
       projectName: envelope.project_name,
       project: envelope.project,
       version: envelope.version,
       intakeMode: envelope.intake.mode,
+      ...(persistence && { persistence }),
```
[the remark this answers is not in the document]
joanreyero: Fixed in 642b31e2: `ready` without a receipt is no longer silently terminal for persistence. The client shows the document immediately but keeps polling (bounded, 3 extra attempts) while the receipt is missing; each poll re-triggers the idempotent content-addressed write server-side, recovering from transient outages. The bound keeps environments where the bucket is intentionally unconfigured from polling forever; a failed background retry poll never clears or errors the displayed result.

apps/lfx-one/src/app/modules/mktg-os-agents/brand-kit-form/brand-kit-form.component.ts
```diff
@@ -146,27 +157,44 @@ export class BrandKitFormComponent implements OnDestroy {
...
+          this.pollTimer = setTimeout(() => this.pollResult(epoch, sessionId, ownerToken, attempt + 1, 0, persistRetries), RESULT_POLL_INTERVAL_MS);
         },
         error: () => {
           if (epoch !== this.pollEpoch) {
             return;
           }
+          if (this.result()) {
+            // The document is already displayed — a failed background
+            // persistence-retry poll must not surface an error or clear it.
+            return;
```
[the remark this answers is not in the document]
joanreyero: Fixed in 0fb4ece6 — the error path now reschedules the next persistence-retry poll (bounded by the same 3-poll `PERSIST_RETRY_MAX_ATTEMPTS` cap) instead of returning early, so a transient failure no longer abandons the remaining retry budget. Added component specs covering the error-path retry, that no error is surfaced while the document is displayed, and the retry cap.

apps/lfx-one/src/server/services/object-store.service.ts
```diff
@@ -23,31 +38,79 @@ export class ObjectStoreService {
...
+        if (!isConfirmedNotFound) {
+          throw error;
+        }
+      }
+
+      await client.send(new PutObjectCommand({ Bucket: bucket, Key: key, Body: body, ContentType: contentType, CacheControl: cacheControl }));
+      logger.success(req, 'object_store_put_if_absent', startTime, { key, written: true });
+      return true;
+    } catch (error) {
+      logger.error(req, 'object_store_put_if_absent', startTime, error, { purpose, key });
```
dealako: [minor] Graceful-degrade write still emits ERROR here

Issue: `putObjectIfAbsent` logs the failure at ERROR before rethrowing. `BrandKitService.persistEnvelope` catches that same failure and deliberately degrades at WARN, so one storage outage produces BOTH an ERROR and a WARN line.

Proof: this line is `logger.error(req, 'object_store_put_if_absent', startTime, error, { purpose, key })` followed by `throw error`. `.claude/rules/logging-patterns.md` (WARN section) explicitly classifies "error conditions with graceful degradation (returning null/empty arrays)" and "service errors that don't propagate to controller" as WARN. The brand-kit spec asserts `logger.error` is never called, but it fully mocks `./object-store.service`, so this real ERROR path is never exercised — the "never ERROR" claim is green only because the emitting module is mocked away.

Why it matters: the bucket env var is intentionally absent in deployed envs "for now," and the client retries up to 3 times, so a transient/expected failure emits repeated ERROR telemetry for a condition the design documents as benign — guaranteed alert noise. Copilot flagged this twice (lines 109 and 111) and suppressed it; it is valid.

Fix: let the recovering caller own severity — surface the failure without logging ERROR here (the business layer already WARNs), or downgrade this to WARN for the degradable put. Then add a test that exercises the real object-store failure path so the "never ERROR" guarantee is actually verified.

packages/shared/src/interfaces/brand-kit.interface.ts
```diff
@@ -104,11 +104,32 @@ export interface BrandKitResultRequest {
...
+}
+
 /**
  * Response of `POST /api/mktg-agents/brand-kit/result`.
  * `pending` until the session emits a valid envelope; then `ready` with the
- * validated document. The document is DISPLAYED to the user only — no
- * server-side persistence in this iteration (persistence deferred).
+ * validated document. On `ready` the BFF also persists the document to
+ * versioned object storage (the dec-brand-kit-storage-v2 write path) and
+ * reports the receipt; a missing `persistence` on a `ready` response means
```
dealako: [nit] Doc enumerates one cause for a missing receipt; the code has two

Issue: this JSDoc says a missing `persistence` on `ready` "means the write failed and will be retried on the next poll." But `persistEnvelope` also returns null (no receipt) when the document exceeds `BRAND_KIT_MAX_DOCUMENT_BYTES` — a permanent condition retrying never resolves.

Proof: `brand-kit.service.ts` size-gate branch logs "Document exceeds the 20 MB object size cap — not persisted" and returns null before any write, propagating the same absent-receipt state; the client then burns its bounded persist-retry budget against an outcome that can never yield a receipt.

Why it matters: the interface is the shared contract; conflating "transient, will retry" with "permanent, will not persist" could mislead a future consumer. (Size gate is near-unreachable given prior validation, hence nit.)

Fix: broaden the doc to note a missing receipt means the write did not succeed — transient failure retried on next poll, OR document exceeded the size cap and will not be persisted.

apps/lfx-one/src/server/services/brand-kit.service.ts
```diff
@@ -53,16 +63,82 @@ export class BrandKitService {
...
+        bytes: documentBytes.length,
+      });
+      return null;
+    }
+
+    const startTime = logger.startOperation(req, 'brand_kit_persist', { project: envelope.project, version: envelope.version });
+    try {
+      // Key derived from validated fields only — the sha was recomputed against
+      // the document bytes in findAuthoritativeEnvelope before selection.
+      const key = buildBrandKitObjectKey(envelope.project, envelope.content_sha256);
```
dealako, line 107: [question] Storage partition (`project`) is not bound to caller authorization

Issue: the key partition `brand-kit/{project}/...` uses `envelope.project`, which originates from agent output stamped from user-supplied intake fields. The controller's owner-token check proves session ownership only; nothing verifies the caller may write into the `project` namespace they named.

Proof: `getResult` → `persistEnvelope` → `buildBrandKitObjectKey(envelope.project, ...)`. `validateBrandKitEnvelope` gates `project` as a kebab slug (shape, not ownership). The write lands in the "shared private marketing artifacts bucket."

Why it matters: today the blast radius is limited — content-addressed keys prevent cross-tenant overwrite, the bucket is private, and the read/Artifact-minting path is deferred (wi-lfx-one-service-actor). But once minting consumes objects keyed by `project`, an unauthorized partition could surface cross-tenant. The authz decision is effectively postponed with no enforcement seam recorded.

Fix: before the read/minting path ships, bind `envelope.project` to the caller's authorized projects (403 otherwise), or explicitly document that partition-level authz is enforced at the future read boundary and is intentionally deferred here.

dealako (changes requested):
Review body posted as a separate summary comment. See inline findings below.

dealako:
Hi @joanreyero 👋 — thoughtful, well-tested PR. The purpose-keyed bucket refactor is clean (per-purpose memoization that doesn't cross-poison is a nice touch), `putObjectIfAbsent` correctly treats only a confirmed 404 as absent, `buildBrandKitObjectKey` re-validates both `project` and `content_sha256` against their regexes with a matching traversal-rejection test, and the graceful-degrade + bounded poll-retry loop is well reasoned. The round-2 fixes (WARN degrade, error-path retry rescheduling) genuinely landed. My findings are all minor/nit/question — nothing structural — but a couple touch a repo logging rule and user-facing accuracy, so I'm marking changes requested to get them addressed.

**Issue count**

- 🔴 Blocking: 0
- 🟡 Minor: 3
  - **ERROR log on the graceful-degrade write path** (`object-store.service.ts:109`) — `putObjectIfAbsent` logs ERROR before rethrowing, while `persistEnvelope` deliberately degrades at WARN. `.claude/rules/logging-patterns.md` classifies graceful-degradation / non-propagating service errors as WARN. The "never ERROR" spec passes only because it mocks the emitting module. With the bucket env var intentionally absent + up to 3 client retries, this is guaranteed ERROR alert noise. (Copilot flagged this twice and suppressed it — it's valid; I agree.)
  - **Success-path retry cap is untested** (`brand-kit-form.component.spec.ts`) — the headline "keeps polling, bounded 3 attempts, while receipt missing" success branch (`persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS`) is never directly exercised; every retry test injects a receipt or an error first. Deleting the cap clause would keep the suite green. Add a test: always-receipt-less `ready`, assert exactly 1 + 3 calls then stop.
  - **Stale user-facing copy** (`brand-kit-form.component.html:32`, outside the diff) — still reads "Download it to keep a copy — it is not stored anywhere else yet," which is now false whenever persistence succeeds. The doc-comments were updated but the visible retention statement wasn't. (Copilot suppressed this too; also valid.)
- ⚪ Nit: 2
  - Interface doc (`brand-kit.interface.ts:130`) lists only "write failed, will retry" for a missing receipt; the size gate is a second, *permanent* null cause — broaden the wording.
  - `BrandKitResultResponse` mixes camelCase top-level fields with a snake_case `persistence` receipt (`s3_key`, `content_sha256`, `intake_mode`). Likely intentional forward-compat with the Artifact-minting contract — if so, a one-line note would prevent a future "normalize this" churn PR. (Left off the formal review; noting here.)
  - Also: the `documentBytes.length > BRAND_KIT_MAX_DOCUMENT_BYTES` size gate in `persistEnvelope` is unreachable given the shared validator already byte-checks — fine to keep as documented defense-in-depth.
- ❔ Question: 1
  - **Storage partition (`project`) isn't bound to caller authorization** (`brand-kit.service.ts:107`). Benign today (content-addressed keys, private bucket, read/minting deferred) but worth an explicit enforcement seam or a documented "authz enforced at the future read boundary" note before minting consumes these objects.

**Bot reconciliation:** Cursor's two Medium findings (persistence never retries; retries stop on poll errors) are both **resolved** in this PR's later commits — verified in the code, not just the reply threads. Copilot's two suppressed comments (ERROR log, stale HTML copy) are **valid**, so I've surfaced them above rather than let them stay suppressed.

**Final decision: 🔴 Needs changes before approval** — no blocking issues, but the logging-rule violation and the now-inaccurate retention copy should land before merge. Everything else is take-it-or-leave-it.

apps/lfx-one/src/server/services/object-store.service.ts
joanreyero: Acknowledged — valid observation, and intentionally deferred rather than fixed in this PR. `putObjectIfAbsent` is a shared object-store primitive also used by non-degradable callers (e.g. avatar upload), so its ERROR-on-failure semantics are load-bearing there; making severity caller-owned means threading a severity/suppression option through the service interface, which is beyond this PR's final-round scope. The duplicate line per outage is bounded (max 3 retries per session) and the WARN from `persistEnvelope` carries the operational signal. Tracking the cleanup (caller-owned severity + a real object-store failure-path test) as a follow-up alongside wi-lfx-one-service-actor.

packages/shared/src/interfaces/brand-kit.interface.ts
joanreyero: Acknowledged — fair point on the doc conflating transient vs permanent absence. As noted it's near-unreachable (upstream validation gates size well below the cap), so leaving the JSDoc as-is for this round; will broaden the wording to cover the size-cap case in the follow-up that touches this interface (read/minting path).

apps/lfx-one/src/server/services/brand-kit.service.ts
joanreyero: Acknowledged — this is the documented deferral: partition-level authz is intentionally enforced at the future read/Artifact-minting boundary (wi-lfx-one-service-actor), not at this write path. Today's blast radius is limited exactly as you note (private bucket, content-addressed keys, no read path). Binding `envelope.project` to caller-authorized projects (403 otherwise) is a stated requirement of the minting WorkItem before any consumer reads by `project`.

apps/lfx-one/src/server/services/object-store.service.ts
joanreyero: Fixed in 9f401863 — the earlier deferral was wrong and I have reversed it. Correcting the record: `putObjectIfAbsent` is NOT shared with non-degradable callers; avatar upload uses the separate `uploadProfilePicture` method (which keeps its own ERROR path). Its only consumer is `BrandKitService.persistEnvelope`, which always catches and degrades, so severity ownership needed no interface change.

The `putObjectIfAbsent` catch now logs WARN (with `purpose`, `key`, `duration`, and the error message) instead of ERROR before rethrowing — the recovering caller owns the operational WARN, and any future unrecovered rethrow still reaches `apiErrorHandler`, which logs ERROR centrally. Per your second ask, the object-store spec now exercises the REAL failure path (non-404 HEAD and PUT failure) and asserts `logger.error` is never called and `logger.warning` is, so the never-ERROR guarantee is no longer vacuously green behind the brand-kit spec's service mock. One storage outage on the degrade path now emits exactly the WARN pair (primitive + `brand_kit_persist`), no ERROR.

apps/lfx-one/src/server/services/object-store.service.ts
```diff
@@ -23,31 +38,92 @@ export class ObjectStoreService {
...
+   */
+  public async putObjectIfAbsent(
+    req: Request,
+    purpose: ObjectStorePurpose,
+    key: string,
+    body: Buffer,
+    contentType: string,
+    cacheControl: string
+  ): Promise<boolean> {
+    await this.ensureBucket(purpose);
```
dealako: [minor] ERROR-on-degrade fix is incomplete on the bucket-readiness path

Issue: `9f401863` correctly moved `putObjectIfAbsent`'s own HEAD/PUT catch to WARN (nicely done, with real-path tests). But this `await this.ensureBucket(purpose)` runs *before* that try/catch, and `doEnsureBucket` still logs at ERROR (`logger.error(undefined, 'object_store_ensure_bucket', …)`) on a HeadBucket 403 / 5xx / timeout before rethrowing.

Proof: `putObjectIfAbsent` (line 85) awaits `ensureBucket` outside its try block; `doEnsureBucket` catch emits `logger.error` for any non-404 (or 404 without `S3_CREATE_MISSING_BUCKET=true`). That rejection propagates to `persistEnvelope`, which degrades at WARN — so a transient bucket-readiness outage on the graceful-degrade persist path still produces one ERROR line. (The pure env-var-unset case is fine — `getBucket` throws before `startOperation`, so no ERROR there.) Copilot flagged this same residual (suppressed) at lines 85/209.

Why it matters: it's the exact alert-noise scenario round 1 raised, just narrowed to genuine S3 outages rather than every failure — a recovered `ready` response can still page.

Fix: this is a real design nuance — `ensureBucket` is shared with the avatar path, so you can't blanket-WARN it. Either thread a 'degradable caller' signal so the readiness failure logs WARN when reached via `putObjectIfAbsent`, or accept the residual and document it (a comment on line 85 + an `ensureBucket`-rejection case in the WARN-not-ERROR test). Not blocking on its own — captured with the two carried-over minors in the summary.

dealako (changes requested):
Follow-up round 2. Full write-up posted as a separate summary comment. One inline note below on the remaining piece of the logging finding.

dealako:
Follow-up review (round 2) — thanks for the turnaround, @joanreyero 🙏

👏 **Nice work this round:**

- **The ERROR→WARN logging fix (`9f401863`) is exactly right.** You didn't just flip the log level — the new comment explains *why* the recovering caller owns severity, and you replaced the vacuously-green mock-based assertion with two specs that exercise the **real** failure paths (HEAD non-404 → `Forbidden`, and PUT failure), asserting `logger.warning` fires and `logger.error` does not. That directly closes the "the test mocks it away" half of the original finding. 👍
- You engaged every round-1 item on the threads rather than silently dropping them — including a candid self-correction on the deferral reasoning (`putObjectIfAbsent` is *not* shared with avatar upload; `uploadProfilePicture` is separate). Appreciated.

**Revision status (round-1 items):**

- ✅ **Resolved** — *[minor] Test mocked away the ERROR path.* Real HEAD-non-404 and PUT-failure specs now assert WARN-not-ERROR (`9f401863`).
- ⚠️ **Partially addressed** — *[minor] ERROR log on the graceful-degrade path.* `putObjectIfAbsent`'s own catch is now WARN, but it `await`s `ensureBucket()` **before** that try/catch, and `doEnsureBucket` still logs ERROR on a HeadBucket 403/5xx/timeout before rethrowing. So a transient *bucket-readiness* outage on the degradable persist path still emits one ERROR (the env-var-unset case is clean). Details inline at `object-store.service.ts:85`. Copilot flagged the same residual.
- ❌ **Still open** — *[minor] Success-path retry cap is untested.* The error-path cap test (`caps error-path persistence retries at the same 3-poll budget`) is great, but there's still no test driving a sequence of consecutive receipt-less `ready` responses (no error, no receipt) to prove the ready-branch `persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS` guard stops at 1 + 3 polls. Deleting that clause would still leave the suite green.
- ❌ **Still open** — *[minor] Stale user-facing copy.* `brand-kit-form.component.html:32` still reads "Download it to keep a copy — it is not stored anywhere else yet," which is now false whenever a receipt comes back. Copilot re-flagged this. It's a user-facing data-retention accuracy statement, so worth correcting (and ideally distinguishing the receipt-less degraded case).
- 🔵 **Acknowledged / deferred (accepted)** — *[nit] interface doc size-gate wording* and *[question] project-partition authz*: your thread replies (broaden the JSDoc later; partition authz intentionally enforced at the future read/minting boundary) are reasonable. Not holding on these.
- ⚪ **Still open (nit)** — camelCase vs snake_case receipt casing; take-it-or-leave-it.

**Open issue count:**

- 🔴 Blocking: 0
- 🟡 Minor: 3 — ERROR-on-degrade residual via `ensureBucket` (partially fixed); untested success-path retry cap; stale "not stored anywhere else yet" copy.
- ⚪ Nit: 2 — interface-doc wording (deferred, fine); receipt field casing.

**Final decision: 🔴 Needs changes before approval** — solid progress and no blockers, but three minors remain open (two carried over untouched, one partially closed). None are large; the stale retention copy and the success-path cap test are the two I'd prioritize.

---
_Process note: subagent-delegated parallel review was unavailable this round (a packaged `pi-subagents` agent, `gpt-pro.md`, declares `runner.type: external-job`, which the installed version rejects and which fails all agent discovery), so this round was reviewed in-thread. Scope was small — one ~100-line commit — so coverage is unaffected._

apps/lfx-one/src/server/services/object-store.service.ts
joanreyero: Fixed in 55e08e751daabfe9823baf5c73d76e4d8974fdc5 via your first option — threading a degradable-caller signal. `ensureBucket` now takes `options.degradable`; `putObjectIfAbsent` passes `{ degradable: true }`, so `doEnsureBucket` logs a non-recoverable readiness failure (403/5xx/timeout, or 404 without `S3_CREATE_MISSING_BUCKET`) at WARN on that path while the avatar path keeps ERROR. The severity is captured when the memoized per-purpose check is created — exact today since each purpose has a single caller class (documented on `ensureBucket`). Added both tests: an `ensureBucket`-rejection-via-`putObjectIfAbsent` WARN-not-ERROR case, and a direct non-degradable `ensureBucket` case locking ERROR for the avatar path.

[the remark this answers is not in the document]
joanreyero: Fixed in 55e08e751daabfe9823baf5c73d76e4d8974fdc5 (valid finding, same root cause as dealako's thread above): `ensureBucket` now accepts a `degradable` caller signal and `putObjectIfAbsent` passes it, so `doEnsureBucket` logs bucket-readiness failures at WARN instead of ERROR when reached via the graceful-degrade persist path; the non-degradable avatar path keeps ERROR. Covered by new WARN-not-ERROR and ERROR-retained tests in object-store.service.spec.ts.

dealako:
Follow-up review (round 3) — thanks for staying with this, @joanreyero 🙏

👏 **Resolved since round 2:**

- ✅ **The ERROR-on-degrade residual is now fully closed (`55e08e75`).** You threaded a `degradable` signal through `ensureBucket` → `doEnsureBucket`, so a HeadBucket 403/5xx/timeout on the brand-kit persist path logs **WARN** while the non-degradable avatar path (`uploadProfilePicture`) keeps **ERROR** — and both severities are locked with tests. That was the exact remaining gap from round 2 (the `ensureBucket` call sat outside `putObjectIfAbsent`'s WARN catch). The comment explaining *why* the recovering caller owns severity is a nice touch. Verified end-to-end in the code, not just the thread. 👍

**Revision status (round-2 items):**

- ✅ **Resolved** — *[minor] ERROR-on-degrade via `ensureBucket`.* Degradable-signal threading is correct and complete (`55e08e75`); WARN on the persist path, ERROR retained on avatar upload.
- ❌ **Still open** — *[minor] Success-path retry cap is untested.* The error-path cap test (`caps error-path persistence retries…`, 4 calls) is solid, but there's still no test that drives a sequence of consecutive receipt-less `ready` responses **with no error** to prove the ready-branch guard `persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS` (`brand-kit-form.component.ts:166`) stops at 1 + 3 polls. Mutation check confirms it: delete `|| persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS` and all three `persistence retry polling` specs stay green (test 1 stops via `response.persistence`; tests 2/3 return one receipt-less `ready` then errors, so they stop via the *error*-branch cap). Inline at `brand-kit-form.component.ts:166`.
- ❌ **Still open** — *[minor] Stale user-facing copy.* `brand-kit-form.component.html:32` still reads *"Download it to keep a copy — it is not stored anywhere else yet."* This is now false whenever a receipt comes back — the BFF persists the validated document to the private marketing-artifacts bucket. The component JSDoc was updated to say the BFF persists server-side, but the visible retention statement wasn't. Flagged in rounds 1–2 and by Copilot; it's a user-facing data-retention accuracy statement, so worth correcting (ideally distinguishing the receipt-less degraded case).
- 🔵 **Accepted / deferred** — *[question] project-partition authz.* `brand-kit/{project}/…` still derives from `envelope.project` (agent output from user intake) with no user→project entitlement check; bounded today (content-addressed no-overwrite, private bucket, Artifact-minting deferred). Your documented plan to enforce this at the future read/minting boundary (wi-lfx-one-service-actor) stands — not holding on it, but please keep it as a hard precondition on that work item.
- ⚪ **Still open (nit)** — interface JSDoc (`brand-kit.interface.ts`) still enumerates only "write failed, will retry" for a missing receipt; the size-cap path is a second, *permanent* null cause. And the `BrandKitPersistReceipt` snake_case fields (`s3_key`, `content_sha256`, `intake_mode`) still mix with the camelCase enclosing response — fine if it deliberately mirrors the downstream Artifact contract; a one-line note would prevent a future "normalize this" churn PR. Both take-it-or-leave-it.

**Open issue count:**

- 🔴 Blocking: 0
- 🟡 Minor: 2 — untested success-path retry cap; stale "not stored anywhere else yet" copy.
- ⚪ Nit: 2 — interface-doc size-cap wording; receipt field casing.
- ❔ Question: 1 — project-partition authz (accepted as deferred).

**Bot reconciliation:** Cursor's earlier retry/degrade findings remain resolved. Copilot's stale-copy re-flag is valid and included above. No new bot findings on the latest commits change the picture.

**Final decision: 🔴 Needs changes before approval** — genuinely close now, and the round-2 logging residual is properly closed. The only two open items are small and both carried over untouched: add the success-path cap regression test, and fix the now-inaccurate retention copy. Land those and this is an approve.

apps/lfx-one/src/app/modules/mktg-os-agents/brand-kit-form/brand-kit-form.component.ts
```diff
@@ -148,27 +159,49 @@ export class BrandKitFormComponent implements OnDestroy {
             return;
           }
           if (response.status === 'ready') {
+            // Show the validated document immediately — persistence is
+            // best-effort and never blocks the user.
             this.generating.set(false);
             this.result.set(response);
+            if (response.persistence || persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS) {
```
dealako, line 172: [minor] Success-path retry cap is untested — the guard can be deleted with the suite still green

Issue: this ready-branch cap (`response.persistence || persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS`) is never exercised by a pure success path — consecutive receipt-less `ready` responses with no error.

Proof: in `brand-kit-form.component.spec.ts`, all three `persistence retry polling` tests return `READY_WITHOUT_RECEIPT` at most once: test 1 then returns a receipt (stops via `response.persistence`); tests 2 and 3 then `mockReturnValue(throwError(...))` (stop via the *error*-branch cap `persistRetries < PERSIST_RETRY_MAX_ATTEMPTS`, verified as 4 calls). Delete `|| persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS` here and every spec stays green.

Why it matters: this clause is the guard that stops the 10s poll loop when the bucket is intentionally unconfigured and the server returns a receipt-less `ready` on every poll. Without a regression test, an off-by-one or accidental removal would reintroduce an unbounded client poll loop and CI wouldn't catch it. Carried over from rounds 1–2.

Fix: add a test that mocks `getResult` to always return `READY_WITHOUT_RECEIPT` (no errors), submits, advances timers by `POLL_INTERVAL_MS * 10`, and asserts `getResult` is called exactly 4 times (initial + 3) and then stops.

dealako (dismissed):
Round 3 — the round-2 logging residual is properly closed (nice work on the `degradable` threading). Two small carried-over minors remain open; full write-up is in the summary comment. One inline note below on the untested success-path retry cap.

apps/lfx-one/src/app/modules/mktg-os-agents/brand-kit-form/brand-kit-form.component.ts
joanreyero: Fixed in 4e07abd4: added the success-path cap test — consecutive receipt-less `ready` responses with no error stop at exactly 1 + 3 polls; verified with your mutation check (deleting `|| persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS` makes exactly this test fail, 4 vs 11 calls, while the other three stay green — then reverted).

joanreyero:
Thanks @dealako — round-3 items all addressed in 4e07abd4.

**[minor] Success-path retry cap untested — fixed.** New spec `caps success-path persistence retries at 1 + 3 polls when every ready response is receipt-less`: every poll returns a receipt-less `ready` with no error, asserting exactly 4 calls (1 initial + 3 bounded retries) and that no 5th is ever scheduled, with the document displayed and no error surfaced. Verified against your mutation check: deleting `|| persistRetries >= PERSIST_RETRY_MAX_ATTEMPTS` leaves the three pre-existing `persistence retry polling` specs green (as you predicted) and makes exactly the new test fail — expected 4 calls, got 11 — then reverted.

**[minor] Stale retention copy — fixed.** The result-view note now binds to `ready.persistence`: with a receipt it reads "It is saved to this project's marketing artifacts — download a copy anytime"; without one (degraded / size-capped) it keeps the honest "Download it to keep a copy — it is not stored anywhere else." No more false retention statement once persistence succeeds.

**[nit] Missing-receipt JSDoc — fixed.** The `BrandKitResultResponse` doc now enumerates both null causes: transient write failure (retried on the next poll, idempotent content-addressed keys) and the `BRAND_KIT_MAX_DOCUMENT_BYTES` size-cap exclusion, called out as permanent and non-retriable.

**[nit] Receipt field casing — fixed.** `BrandKitPersistReceipt` now carries the note you suggested: the snake_case fields deliberately mirror the downstream Artifact contract verbatim so minting needs no normalization layer — with an explicit "do not camelCase" to head off the churn PR.

**[question] Project-partition authz — acknowledged as a hard precondition.** Agreed: binding `envelope.project` to caller-authorized projects is a hard precondition on the read/Artifact-minting work (wi-lfx-one-service-actor), not an optional hardening item. It will be stated as such on that work item.

All gates green locally: `format:check`, `lint:check`, `check-types`, `build`, plus the full brand-kit-form spec suite (11/11).

apps/lfx-one/src/server/services/brand-kit.service.ts
```diff
@@ -53,16 +63,82 @@ export class BrandKitService {
...
+
+      const written = await this.objectStore.putObjectIfAbsent(req, 'marketing-os-artifacts', key, documentBytes, 'text/markdown; charset=utf-8', 'private');
```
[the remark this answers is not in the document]
joanreyero: Won't fix in this PR — this is the project-partition authz question dealako raised in round 1 and explicitly accepted as deferred in his round-2 and round-3 summaries: binding `envelope.project` to a caller-authorized canonical project (server-side entitlement check, key derived from the verified value) is a hard precondition on the read/Artifact-minting boundary work (wi-lfx-one-service-actor), where these objects first become consumable.

Exposure today is deliberately bounded: the bucket is private with no read/retrieval or minting path in this PR; the key comes from `buildBrandKitObjectKey`, which re-validates `project` and `content_sha256` against strict regexes (traversal rejected, with tests); and `putObjectIfAbsent` writes only on a confirmed 404, so an attacker can never overwrite another project's existing objects — at worst they park content-addressed markdown under a foreign prefix that nothing reads until the authz-gated boundary lands. Resolving per the reviewer-accepted deferral.

apps/lfx-one/src/app/modules/mktg-os-agents/brand-kit-form/brand-kit-form.component.ts
```diff
@@ -0,0 +1,227 @@
...
+            // content-addressed write, recovering from transient storage
+            // outages without changing what the user sees.
+            this.pollTimer = setTimeout(() => this.pollResult(epoch, sessionId, ownerToken, attempt + 1, 0, persistRetries + 1), RESULT_POLL_INTERVAL_MS);
+            return;
+          }
+          if (attempt >= RESULT_POLL_MAX_ATTEMPTS) {
+            this.failGeneration('The generation is taking longer than expected. Please try again later.');
+            return;
+          }
+          this.pollTimer = setTimeout(() => this.pollResult(epoch, sessionId, ownerToken, attempt + 1, 0, persistRetries), RESULT_POLL_INTERVAL_MS);
```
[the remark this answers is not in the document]
joanreyero: Fixed in 190ed79d99497a314887490e41e42142065ccf5f. Real bug — confirmed both halves: (1) after a receipt-less `ready`, a later non-ready payload fell through to the generation-timeout branch, which called `failGeneration` without the `this.result()` check the error branch has, and (2) because persist retries also increment `attempt`, the loop polled with a frozen `persistRetries` until the 30-attempt budget and then overlaid a spurious timeout error on the displayed document. The `next` handler now guards on `this.result()` before the attempt-budget check, mirroring the error branch: a non-ready response during background persistence retries consumes the bounded persist-retry budget (max 3) and never calls `failGeneration`. Once the document is displayed, `attempt` is no longer consulted on any branch, so retries consuming the poll budget is now inert. Regression spec added (`never overlays a timeout error when persist-retry polls come back non-ready`): receipt-less ready then endless `pending` stops at exactly 1 + 3 polls with no error — mutation-verified (reverting the component fix makes it fail with 30 calls instead of 4).

niravpatel27 (approved):

PR #1443, opened 2026-08-13
fix(org-lens): harden the all-time axis read path in the project detail BFF

ahmedomosanya:
Follow-up to [#1305](https://github.com/linuxfoundation/lfx-self-serve/pull/1305), which merged before @audigregorie's review feedback was addressed. This lands the findings that have a real effect on the running system and leaves the rest as-is.

## What's here

**Lifetime trend series are now aligned to the bucket axis by index.** They were built by pushing scores in row order, which trusts the read model to be dense per (account, bucket), while the downstream `buildTrendSeries` never pads named orgs to the axis length. One missing row would have shifted a series against `periods[]` and silently mislabelled every point on it. Values are now placed **by bucket index** into a zero-filled array of the axis length — the same way card sparklines densify in `denseSeries` — so each series is `periods.length` long by construction rather than by convention.

**That also changes the top-10 ranking key, which is worth naming.** `buildTrendSeries` ranks orgs by the last element of `combined`, i.e. the most recent bucket. Under the old push-order build, an org with no row in the final bucket was ranked by its last *available* bucket — so orgs were being compared at different points in time. Zero-filling to the axis means such an org now ranks at `0`, which is what "current influence" should mean.

This is latent rather than active. All **3,555,035** (project, organization) series in `ORG_LENS_PROJECT_DETAIL_TREND_LIFETIME` are dense today — none is sparse, and none is missing its final bucket — so no chart's top-10 selection changes with this PR. The new key only takes effect in the sparse case this fix exists to survive, and there it is the coherent ordering. The 1y/2y path (`buildTrendByAccount`) is untouched and still builds in row order.

**The all-time cache validator no longer rejects its own entries.** `hasAdaptivePeriods` required a non-empty `periods` array. A project with no rows in the bucket spine legitimately caches `periods: []`, so the validator rejected that entry on every read and re-queried Snowflake for the life of the TTL. Presence of the array is what rejects stale pre-deploy entries, so dropping the emptiness check loses nothing.

**A `multi_year` bucket with a missing end date no longer renders as a bare `YYYY`**, which was indistinguishable from a yearly bucket and understated the span.

**Supporting changes:** the all-time window sentinel is named (`PD_ALL_TIME_WINDOW`) with a note that it is not a real month count and must not be used in arithmetic — kept module-local, since `PD_TIME_RANGE_MONTHS` in the same file is its only reader — and the E2E spec now asserts each bucket label matches one of the four documented formats rather than only checking array shape and length.

## What's deliberately not here

**The client-side leading-bucket trims from #1305 are kept.** The review flagged that the trend chart and the influence cards trim on different criteria, which is accurate — but the two render on different tabs, values are correct in both, and within each block labels and data are sliced by the same index, so nothing is mislabelled. The empty leading buckets those trims hide come from the unbounded bucket-spine anchor, which is tracked and documented in [LFXV2-3231](https://linuxfoundation.atlassian.net/browse/LFXV2-3231); reverting the trims to keep that symptom visible in the UI is not needed now that it's captured there.

**The cache key is not bumped.** A pre-deploy all-time entry with a non-empty axis still passes the validator, so a series written by the old row-order builder can be served until the 1h TTL rotates. Bumping the key would close that window at deploy time, but it would force a cold Snowflake read for every project — and the largest ones currently exceed the 15s query timeout on a cold read, a separate defect now documented on [LFXV2-3231](https://linuxfoundation.atlassian.net/browse/LFXV2-3231). That would trade a certain failure for a misalignment which, per the density measurement above, no project's row set can currently produce. The reasoning is recorded on the validator so the next reader does not have to re-derive it.

## Verification

`yarn format:check`, `yarn lint:check` (1 pre-existing warning, unrelated), `yarn check-types`, `yarn test` (796 + 1093 + 254 passing), `yarn build` — all green.

[LFXV2-3231]: https://linuxfoundation.atlassian.net/browse/LFXV2-3231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/e2e/org-project-detail.spec.ts
+++ b/apps/lfx-one/e2e/org-project-detail.spec.ts
@@ -375,11 +375,14 @@ test.describe('Org Project Detail — all-time range', () => {
     expect(Array.isArray(influenceBody.periods)).toBe(true);
     expect(influenceBody.periods!.length).toBeGreaterThan(0);
     expect(influenceBody.periods!.length).toBeLessThanOrEqual(12);
-    // Every label must be one of the four documented bucket formats — `MMM YYYY`, `Q# YYYY`, `YYYY`,
-    // `YYYY–YYYY`. A length-only assertion passes on empty or malformed labels, which is exactly what
-    // an unparseable bucket start produces, so it would hide a regression in the label derivation.
+    // Every label must be one of the formats the server can emit — `MMM YYYY`, `Q# YYYY`, `YYYY`,
+    // `YYYY–YYYY`, plus `YYYY–?` for a multi-year bucket whose end date is missing. A length-only
+    // assertion passes on empty or malformed labels, which is exactly what an unparseable bucket start
+    // produces, so it would hide a regression in the label derivation. The unknown-end form is allowed
+    // rather than rejected: it is deliberate server output, so failing on it would turn this spec into
+    // a warehouse data-quality gate.
     for (const period of influenceBody.periods!) {
-      expect(period).toMatch(/^(?:[A-Z][a-z]{2} \d{4}|Q[1-4] \d{4}|\d{4}(?:\u2013\d{4})?)$/);
+      expect(period).toMatch(/^(?:[A-Z][a-z]{2} \d{4}|Q[1-4] \d{4}|\d{4}(?:\u2013(?:\d{4}|\?))?)$/);
     }
     const sparkline = influenceBody.technical?.[0]?.sparkline;
     expect(Array.isArray(sparkline)).toBe(true);
--- a/apps/lfx-one/e2e/org-project-detail.spec.ts
+++ b/apps/lfx-one/e2e/org-project-detail.spec.ts
@@ -375,6 +375,12 @@ test.describe('Org Project Detail — all-time range', () => {
     expect(Array.isArray(influenceBody.periods)).toBe(true);
     expect(influenceBody.periods!.length).toBeGreaterThan(0);
     expect(influenceBody.periods!.length).toBeLessThanOrEqual(12);
+    // Every label must be one of the four documented bucket formats — `MMM YYYY`, `Q# YYYY`, `YYYY`,
+    // `YYYY–YYYY`. A length-only assertion passes on empty or malformed labels, which is exactly what
+    // an unparseable bucket start produces, so it would hide a regression in the label derivation.
+    for (const period of influenceBody.periods!) {
+      expect(period).toMatch(/^(?:[A-Z][a-z]{2} \d{4}|Q[1-4] \d{4}|\d{4}(?:\u2013\d{4})?)$/);
+    }
     const sparkline = influenceBody.technical?.[0]?.sparkline;
     expect(Array.isArray(sparkline)).toBe(true);
     expect(sparkline!.length).toBe(influenceBody.periods!.length);
--- a/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
+++ b/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
@@ -473,9 +473,10 @@ export class OrgLensProjectDetailService {
         ecosystem: isNonLf || !viewing ? null : (this.mapBand(viewing.LEVEL_ECOSYSTEM) ?? 'silent'),
       },
       // Only the all-time axis is variable/bucketed; 1y/2y stay client-derived (periods omitted).
-      // Tested against undefined rather than truthiness on purpose: an all-time project with an empty
-      // bucket spine yields `[]`, which must still be emitted, or the cache validator would reject the
-      // very block that was just written and re-query Snowflake on every request.
+      // The requirement this states: an all-time project with an empty bucket spine yields `[]`, and
+      // that MUST still be emitted, or `hasAdaptivePeriods` would reject the block just written. An
+      // empty array is truthy, so a truthiness test behaves identically here; the explicit undefined
+      // check just makes the intent legible without relying on the reader recalling that.
       ...(sparkData.periods === undefined ? {} : { periods: sparkData.periods }),
     };
     if (key !== null) {
@@ -921,7 +922,8 @@ export class OrgLensProjectDetailService {
 
   /**
    * Adaptive-bucket axis label for the wire `periods[]` (D9): monthly `MMM YYYY`, quarterly `Q# YYYY`,
-   * yearly `YYYY`, multi-year `YYYY–YYYY`. Derived from the bucket's start (and end for multi-year).
+   * yearly `YYYY`, multi-year `YYYY–YYYY`, and `YYYY–?` for a multi-year bucket whose end date is
+   * missing or unparseable. Derived from the bucket's start (and end for multi-year).
    */
   private bucketLabel(granularity: string | null, start: Date | string | null, end: Date | string | null): string {
     const startDate = this.parseUtcDate(start);
--- a/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
+++ b/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
@@ -472,11 +472,8 @@ export class OrgLensProjectDetailService {
         technical: viewing ? (this.mapBand(viewing.LEVEL_TECHNICAL) ?? 'silent') : null,
         ecosystem: isNonLf || !viewing ? null : (this.mapBand(viewing.LEVEL_ECOSYSTEM) ?? 'silent'),
       },
-      // Only the all-time axis is variable/bucketed; 1y/2y stay client-derived (periods omitted).
-      // The requirement this states: an all-time project with an empty bucket spine yields `[]`, and
-      // that MUST still be emitted, or `hasAdaptivePeriods` would reject the block just written. An
-      // empty array is truthy, so a truthiness test behaves identically here; the explicit undefined
-      // check just makes the intent legible without relying on the reader recalling that.
+      // Only the all-time axis is variable/bucketed; 1y/2y omit `periods` and derive labels client-side.
+      // An empty axis must still be emitted as `[]`, or `hasAdaptivePeriods` rejects the block just written.
       ...(sparkData.periods === undefined ? {} : { periods: sparkData.periods }),
     };
     if (key !== null) {
@@ -1853,6 +1850,12 @@ export class OrgLensProjectDetailService {
    * An EMPTY array is accepted on purpose: a project with no rows in the bucket spine legitimately
    * caches `periods: []`, and requiring a non-empty axis would reject that entry on every read and
    * re-query Snowflake for the life of the TTL.
+   *
+   * Deliberately not gated any tighter. An entry written before the series builder placed values by
+   * bucket index still passes, so a misaligned series could be served until the 1h TTL rotates. A
+   * cache-key bump would close that window at deploy time, but it would force a cold read for every
+   * project, and the largest ones exceed the query timeout on a cold read (LFXV2-3231) — a certain
+   * outage traded against a misalignment that requires a sparse row set no project currently has.
    */
   private static hasAdaptivePeriods(value: unknown): boolean {
     const periods = (value as { periods?: unknown }).periods;
--- a/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
+++ b/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
@@ -473,7 +473,10 @@ export class OrgLensProjectDetailService {
         ecosystem: isNonLf || !viewing ? null : (this.mapBand(viewing.LEVEL_ECOSYSTEM) ?? 'silent'),
       },
       // Only the all-time axis is variable/bucketed; 1y/2y stay client-derived (periods omitted).
-      ...(sparkData.periods ? { periods: sparkData.periods } : {}),
+      // Tested against undefined rather than truthiness on purpose: an all-time project with an empty
+      // bucket spine yields `[]`, which must still be emitted, or the cache validator would reject the
+      // very block that was just written and re-query Snowflake on every request.
+      ...(sparkData.periods === undefined ? {} : { periods: sparkData.periods }),
     };
     if (key !== null) {
       await valkeyService.setJson(key, block, VALKEY_CACHE.ORG_LENS_SNOWFLAKE_TTL_SECONDS);
--- a/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
+++ b/apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
@@ -502,7 +502,15 @@ export class OrgLensProjectDetailService {
     if (range === 'all') {
       const [heroRow, trendRows, axis] = await Promise.all([this.fetchHeroRow(orgUid, slug), this.fetchTrendLifetime(slug), this.fetchLifetimeAxis(slug)]);
       if (!heroRow) return null;
-      block = { trend: this.buildTrendSeries(this.buildTrendLifetimeByAccount(trendRows)), periods: axis.map((bucket) => bucket.label) };
+      block = {
+        trend: this.buildTrendSeries(
+          this.buildTrendLifetimeByAccount(
+            trendRows,
+            axis.map((bucket) => bucket.index)
+          )
+        ),
+        periods: axis.map((bucket) => bucket.label),
+      };
     } else {
       const [heroRow, trendRows] = await Promise.all([this.fetchHeroRow(orgUid, slug), this.fetchTrend(slug)]);
       if (!heroRow) return null;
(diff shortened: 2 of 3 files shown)


apps/lfx-one/e2e/org-project-detail.spec.ts
```diff
@@ -375,6 +375,12 @@ test.describe('Org Project Detail — all-time range', () => {
     expect(Array.isArray(influenceBody.periods)).toBe(true);
     expect(influenceBody.periods!.length).toBeGreaterThan(0);
     expect(influenceBody.periods!.length).toBeLessThanOrEqual(12);
+    // Every label must be one of the four documented bucket formats — `MMM YYYY`, `Q# YYYY`, `YYYY`,
+    // `YYYY–YYYY`. A length-only assertion passes on empty or malformed labels, which is exactly what
+    // an unparseable bucket start produces, so it would hide a regression in the label derivation.
+    for (const period of influenceBody.periods!) {
+      expect(period).toMatch(/^(?:[A-Z][a-z]{2} \d{4}|Q[1-4] \d{4}|\d{4}(?:\u2013\d{4})?)$/);
+    }
```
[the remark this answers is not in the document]
ahmedomosanya: Good catch — addressed in `6f9cfc8`.

You're right that the assertion contradicted the server's own output: `bucketLabel` deliberately emits `YYYY–?` for a `multi_year` bucket with no end date, and the spec declared that invalid. I took the first of your two options and allowed the unknown-end form, rather than removing the server fallback — the fallback exists so a missing end date is visible instead of silently collapsing to a bare `YYYY`, and rejecting it here would have turned this spec into a warehouse data-quality gate that fails on unrelated changes. The assertion still catches empty and malformed labels, which is what it is there for.

apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.ts
```diff
@@ -998,11 +974,10 @@ export class OrgProjectDetailComponent {
...
   /**
-   * Builds a 100%-stacked area chart from the real per-org combined-influence series. The server
-   * already sends the top-N named orgs plus a single "All others" band that sums the complete
-   * remaining tail, so no org is dropped client-side, and each point is normalized so all series
-   * sum to 100%. For the all-time range, leading points where no org has any activity are trimmed
-   * so the axis opens on the project's first real data rather than on empty spine buckets.
+   * Builds a 100%-stacked area chart from the real per-org monthly combined-influence series.
+   * The server already sends the top-N named orgs plus a single "All others" band that sums the
+   * complete remaining tail, so every series is rendered as-is (no client-side truncation) and
+   * each month is normalized so all series sum to 100%.
```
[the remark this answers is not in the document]
ahmedomosanya: Out of scope for this PR now — it no longer touches `org-project-detail.component.ts`.

This was raised against an earlier revision that reverted the client-side bucket trimming from #1305. That revert has been dropped, so the diff is the BFF service, the shared constants, and the E2E spec only.

The observation itself is fair: that JSDoc says "monthly" while the method also renders the all-time adaptive axis, whose points can be quarterly, yearly, or multi-year buckets. It describes code already on `main`, so it belongs with the next change to that component rather than here.

apps/lfx-one/e2e/org-project-detail.spec.ts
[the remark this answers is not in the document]
ahmedomosanya: Confirmed and fixed in `6f9cfc8`. The assertion now accepts the unknown-end form (`YYYY–?`) alongside the four well-formed labels, so the `multi_year`-with-null-end fallback no longer fails the suite while the check still rejects empty and malformed labels.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -1818,9 +1842,16 @@ export class OrgLensProjectDetailService {
...
+   * Whether a cached `all` block carries the adaptive-bucket axis, i.e. was written by this version.
+   * Pre-deploy entries have no `periods` at all, so the presence of the array is what rejects them.
+   * An EMPTY array is accepted on purpose: a project with no rows in the bucket spine legitimately
+   * caches `periods: []`, and requiring a non-empty axis would reject that entry on every read and
+   * re-query Snowflake for the life of the TTL.
+   */
   private static hasAdaptivePeriods(value: unknown): boolean {
     const periods = (value as { periods?: unknown }).periods;
-    return Array.isArray(periods) && periods.length > 0 && periods.every((label) => typeof label === 'string');
+    return Array.isArray(periods) && periods.every((label) => typeof label === 'string');
```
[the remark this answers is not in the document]
ahmedomosanya: Not a bug, but the code invited the misreading, so I've made it explicit in `6f9cfc8`.

The guard is `sparkData.periods ? { periods: sparkData.periods } : {}`, which tests the array **reference**, not its length — and an empty array is truthy in JavaScript. So an all-time project with an empty bucket spine does emit `periods: []`, the relaxed validator accepts it, and there is no thrash. The field is only omitted for 1y/2y, where `fetchMonthlySparklineData` leaves `periods` undefined.

That said, this is the second reading of the line that assumed `[]` was filtered out, so it now tests `=== undefined` explicitly and states the invariant in a comment: an empty axis is a legitimate all-time payload and must still be emitted, or the validator would reject the block that was just written.

ahmedomosanya:
Copilot's review on `6f9cfc8` carried two **suppressed** (low-confidence) comments. Both were correct, and both are fixed in `3fba44a` — recording them here since suppressed items have no thread to reply on.

1. **The `periods` spread comment documented incorrect language behaviour.** It implied a truthiness test would have dropped an empty axis. It would not — an empty array is truthy, so the previous condition already emitted `[]`, and the cache churn came purely from `hasAdaptivePeriods` requiring a non-empty array. The comment now states the requirement without making a false claim about the old code. (Same point I made refuting the Bugbot finding on that line, so the comment was contradicting my own reasoning.)
2. **`bucketLabel`'s JSDoc named only four label formats** while the method now also emits `YYYY–?` for a multi-year bucket with a missing end date — a format the E2E assertion accepts. The contract comment now lists it.

Worth noting for anyone auditing bot findings on this repo: this block is titled "Suppressed comments (2)", not the "suppressed due to low confidence" wording that a marker-based sweep typically greps for, so it does not show up unless the review body is read in full.

audigregorie (commented):
## Code Review Summary

Solid follow-up that lands the real-effect findings from the #1305 review. The core alignment fix (place-by-bucket-index + zero-fill to `axis.length`) is correct and mirrors `denseSeries`; the `hasAdaptivePeriods` self-rejection fix is the right call; the `YYYY–?` label and the strengthened e2e label-format assertion are good. No Critical or Major issues — four Minor notes below, mostly a behavior change worth acknowledging and a couple of hygiene items.

### Minor — outside the diff

- **Ranking-key change not called out in the PR body.** Zero-filling every series to `axis.length` changes what `buildTrendSeries`' `latest()` returns for an org that has no row in the most-recent bucket (previously its last-available score, now `0`). That's arguably more correct for "current influence," but it can reorder the top-10 and isn't mentioned in the description (which frames the change as positional alignment only). Worth a sentence acknowledging the ranking-key change.

### What's done well

- The place-by-index + zero-fill construction guarantees `periods[]` alignment by length rather than by convention — the right structural fix, and it makes the "All others" aggregation inherit the correct length for free.
- `hasAdaptivePeriods` accepting `[]` fixes a real self-rejection loop (write `[]` → validator rejects → re-query for the TTL), and the doc-comment captures the rationale crisply.
- The e2e regex asserting the full label format (instead of length-only) is a genuine spec strengthening — it would catch a regression in `bucketLabel` that a length check would hide.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -1320,21 +1336,34 @@ export class OrgLensProjectDetailService {
...
+      entry.byBucket.set(this.num(row.BUCKET_INDEX), this.round1(this.num(row.COMBINED_INFLUENCE_SCORE)));
+    }
+
+    const byAccount = new Map<string, TrendSeries>();
+    for (const [accountId, entry] of scoresByAccount) {
+      byAccount.set(accountId, {
+        accountId,
+        orgName: entry.orgName,
+        orgLogoUrl: entry.orgLogoUrl,
+        combined: axis.map((bucketIndex) => entry.byBucket.get(bucketIndex) ?? 0),
```
audigregorie, line 1362: **Minor — zero-fill changes the `latest()` ranking key for orgs missing the most-recent bucket**

`buildTrendSeries` ranks orgs by `series.combined[combined.length - 1]` (the most-recent bucket). Before this change (push-order), an org with no row in the most-recent bucket was ranked by its *last available* bucket; now it's ranked by `0`. That's arguably more correct for "current influence," but it's a real behavior change in top-10 selection that the PR description doesn't call out — the body frames the change as purely positional alignment.

If intended, a one-liner in the PR body acknowledging the ranking-key change would help reviewers and future readers. If not intended, an org missing the most-recent bucket could drop out of the named top-10 unexpectedly.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -1818,9 +1847,16 @@ export class OrgLensProjectDetailService {
...
+   * Whether a cached `all` block carries the adaptive-bucket axis, i.e. was written by this version.
+   * Pre-deploy entries have no `periods` at all, so the presence of the array is what rejects them.
+   * An EMPTY array is accepted on purpose: a project with no rows in the bucket spine legitimately
+   * caches `periods: []`, and requiring a non-empty axis would reject that entry on every read and
+   * re-query Snowflake for the life of the TTL.
+   */
   private static hasAdaptivePeriods(value: unknown): boolean {
     const periods = (value as { periods?: unknown }).periods;
-    return Array.isArray(periods) && periods.length > 0 && periods.every((label) => typeof label === 'string');
+    return Array.isArray(periods) && periods.every((label) => typeof label === 'string');
```
audigregorie, line 1862: **Minor — validator doesn't version-gate the alignment fix**

`hasAdaptivePeriods` now accepts `periods: []`, but it still passes for any pre-deploy `all` trend block with a *non-empty* `periods` array — including blocks written by the old push-order `buildTrendLifetimeByAccount`. Those stale blocks carry potentially misaligned series yet pass the validator, so they're served until `ORG_LENS_SNOWFLAKE_TTL_SECONDS` (3600s = 1h) rotates.

The 1h window plus the documented dense-contract assumption likely makes this acceptable, but it should be a conscious decision. If you want the fix live on deploy rather than after TTL rotation, bump the cache key (e.g. a `v2` segment in `project-detail-trend:`) or add a version marker to the validator.

packages/shared/src/constants/org-lens-project-detail.constants.ts
```diff
@@ -103,15 +103,21 @@ export const PD_TIME_RANGE_OPTIONS: { id: OrgLensLeaderboardTimeRange; label: st
   { id: 'all', label: 'All time' },
 ];
 
+/**
+ * Stand-in "window length" for the all-time range: not a real month count, so never do arithmetic
+ * on it (`PD_ALL_TIME_WINDOW - 12` is meaningless). It exists only so a residual `slice(-months)`
+ * over an all-time series is a no-op and returns the payload whole.
+ */
+export const PD_ALL_TIME_WINDOW = Number.MAX_SAFE_INTEGER;
```
audigregorie: **Minor — `PD_ALL_TIME_WINDOW` is exported but has no external consumers**

The new named sentinel is only referenced in this same file (by `PD_TIME_RANGE_MONTHS`). Exporting it adds it to the shared package's public API surface with no consumer.

```suggestion
const PD_ALL_TIME_WINDOW = Number.MAX_SAFE_INTEGER;
```

Drop `export` (keep it module-local) until a consumer actually needs it, or keep it exported if it's intended as a documented contract for future consumers.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
```diff
@@ -473,7 +473,11 @@ export class OrgLensProjectDetailService {
         ecosystem: isNonLf || !viewing ? null : (this.mapBand(viewing.LEVEL_ECOSYSTEM) ?? 'silent'),
       },
       // Only the all-time axis is variable/bucketed; 1y/2y stay client-derived (periods omitted).
-      ...(sparkData.periods ? { periods: sparkData.periods } : {}),
+      // The requirement this states: an all-time project with an empty bucket spine yields `[]`, and
+      // that MUST still be emitted, or `hasAdaptivePeriods` would reject the block just written. An
+      // empty array is truthy, so a truthiness test behaves identically here; the explicit undefined
+      // check just makes the intent legible without relying on the reader recalling that.
```
audigregorie: **Minor — inline comment narrates a no-op refactor and exceeds the 1–2 line guideline**

The PR body confirms the `=== undefined` vs truthiness change is behaviorally identical (it's legibility-only), and the comment itself says so. A 4-line comment explaining a no-op refactor is more narration than intent. The repo's comment guidance keeps "why" comments to 1–2 lines.

```suggestion
      // Emit periods even when empty ([]) so hasAdaptivePeriods accepts an all-time block with an empty bucket spine.
```

One intent line is enough; the truthiness-vs-undefined distinction is clear from the code.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
ahmedomosanya: Good catch, and you're right that the description framed it as positional only. Acknowledged in the PR body now.

I also measured how much it actually moves, because "can reorder the top-10" turns out to be latent rather than active. Across the whole lifetime trend model:

| | |
|---|---|
| `(project, organization)` series | 3,555,035 |
| Sparse series (fewer rows than the project's axis) | **0** |
| Series missing the project's final bucket | **0** |

So every series already carries a row in every bucket including the last, which means `combined[length - 1]` resolves to the same value before and after this change and no chart's top-10 selection moves today. The new key only engages in the sparse case the fix exists to survive.

And it is intended. Under push-order, an org missing the final bucket was ranked by its own last *available* bucket, so orgs were compared at different points in time — the ranking key claimed "most recent bucket" but silently meant "whatever each org last had". Ranking it at `0` makes every org comparable at the same bucket, which is what `latest()` says it does.

Worth noting the 1y/2y path (`buildTrendByAccount`) still builds in row order and is untouched here, so this applies to the all-time range only.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
ahmedomosanya: Agreed it should be a conscious decision rather than an accident, so I've recorded it on the validator's doc comment. **Declining the cache-key bump**, for a reason that surfaced separately today.

A cold read on this endpoint currently *fails* on the largest projects. `kubevirt` returns 398,808 rows (11,078 organizations x 36 months) because the top-10 + "All others" fold happens in `buildTrendSeries` rather than in SQL. Snowflake executes that in 0.9s and returns `SUCCESS`; the BFF then blows its own 15s timeout pulling and materializing the rows, and the request 500s. Filed with the evidence on [LFXV2-3231](https://linuxfoundation.atlassian.net/browse/LFXV2-3231).

Bumping the cache key invalidates every trend entry at deploy and forces exactly that cold read for every project — turning a deploy into a burst of 500s on the biggest ones.

What it would buy in exchange is nothing measurable. The stale entries you're describing can only be *misaligned* if some series was sparse, and none is: 0 of 3,555,035 `(project, organization)` series in the lifetime model is sparse or missing its final bucket. So a pre-deploy entry passing the validator is serving a correctly-aligned series regardless of which builder wrote it.

That makes it a certain failure traded against a defect no current row set can produce, so the 1h TTL rotation is the better path. Worth revisiting once the fold moves into SQL and a cold read is cheap again — at that point bumping the key costs nothing and the argument reverses.

packages/shared/src/constants/org-lens-project-detail.constants.ts
ahmedomosanya: Fixed in `6facbfc66` — dropped the `export`, kept the doc comment.

Confirmed `PD_TIME_RANGE_MONTHS` on line 120 of the same file is its only reader, so there was no consumer to break. The naming was the point of the original change (a bare `Number.MAX_SAFE_INTEGER` reads as a month count at the call site); it gets that legibility while module-local, so there's no reason to widen the shared package's public API for it.

apps/lfx-one/src/server/services/org-lens-project-detail.service.ts
ahmedomosanya: Fixed in `6facbfc66` — trimmed to two lines:

```ts
// Only the all-time axis is variable/bucketed; 1y/2y omit `periods` and derive labels client-side.
// An empty axis must still be emitted as `[]`, or `hasAdaptivePeriods` rejects the block just written.
```

You're right that it was narration. Some history worth recording, since it explains how a 5-line comment on a no-op got there: the block grew in response to an earlier Copilot finding that the *original* wording documented incorrect language behaviour (it implied the explicit check was needed for `[]` to be emitted, when an empty array is truthy and the previous form emitted it fine). I corrected the claim but kept explaining it, which is how a false statement about the language turned into four lines about why the statement doesn't matter.

The requirement — emit `[]` rather than omitting it — is the only part a future reader needs, so that's all that's left. Kept the `=== undefined` form itself, since it states the condition directly.

audigregorie (approved):

PR #1450, opened 2026-08-13
feat(campaigns): mode the planner by delivery type for email

mrautela365:
## Summary

[LFXV2-3201](https://linuxfoundation.atlassian.net/browse/LFXV2-3201)

The Plan tab mounted the **paid** planner unchanged under the Email tab. Under a tab labelled Email, the user was shown an "Ad Channels" card, had to select an ad platform before Generate would enable (`canGenerate` required `selectedPlatforms().size > 0`, and the set defaults to `google-ads`), and got RSA copy, Demand Gen copy, ad keywords and a LinkedIn targeting strategy back.

This PR makes the planner serve both delivery types.

## The design decision this ticket owned

The prior code left a comment asking whether the two delivery types should share a component at all. **They share it**, moded by a `deliveryType` input.

The halves the two types genuinely share are the expensive ones: the event URL, the scrape, the goal/audience/value-prop form, the SSE generation stream, and the brief save/restore round trip. What differs is one card and one validity rule — a poor reason to fork ~800 lines and then maintain the shared parts twice.

The input defaults to `'paid-marketing'`, so this is **additive**: the paid container binds nothing and its behaviour is unchanged.

## What changes in email mode

| | Paid | Email |
|---|---|---|
| Ad Channels card | shown | hidden |
| Budget & Assets card | shown | hidden |
| `canGenerate` | requires an ad platform | requires only a valid form |
| Ad-copy generation | runs | **skipped** |
| Keyword generation | runs | **skipped** |
| Refine | runs | **refused** |

### Omitting `platforms` was not enough — and the first commit was wrong about this

The first commit omitted `platforms` for email and claimed that meant "no ad channels". **It does not.** The generator reads an absent list as the paid *default* at five sites in `campaign-proxy.service.ts`:

```ts
const selectedPlatforms = body.platforms?.length ? body.platforms : ['google-ads'];
```

and the keyword branch opts absence in *explicitly*:

```ts
if (body.platforms?.includes('google-ads') || !body.platforms || body.platforms.length === 0) {
```

So an email brief still made a Google RSA/Display copy call and a keyword call — the exact spend the omission was meant to avoid — and the review step then rendered that ad copy and keyword table. The client change looked complete while the server silently undid it.

Absence *cannot* carry the meaning, because it already means "use the default" for every paid caller. The second commit adds **`deliveryType`** to both brief request interfaces so the server is told rather than inferring.

**Ad copy is skipped, not run empty.** `buildCopySystemPrompt` composes per-platform sections and ends with `Keys: <joined>` — with no platforms that asks the model for a JSON object with **no keys**, a broken call rather than a cheap one.

**Refine is refused** with a message. It re-runs the same copy generator, and there is no email copy to refine until `email-copy` lands (LFXV2-3198).

**The ad-copy result sections need no new guards.** Each is already gated on its own data being non-empty — and now that the server generates none of it for email, those guards are actually load-bearing rather than nominally so.

### Budget & Assets

Also hidden for email, and `totalBudget` is dropped from the request. Both halves state a paid fact: the helper text describes daily ad pacing (`Budget/day = Total / (End Date − Today)`), and the asset sizes are ad creative dimensions. The budget is not merely unread — it reaches the copy prompt as `Total Campaign Budget: $N`.

## Still open (not this PR)

An email brief generates **no copy at all** today — the scrape, event details and HubSpot UTM are what comes back, and Refine says so explicitly. Subject, preview text and sender arrive with the `email-copy` endpoint under LFXV2-3198, together with the template picker the Implement panel needs. Until then `onEmailProceedToImplementation` stores the brief for existence only; the panel does not read the payload.

## Testing

9 new specs, and **each was verified to fail when its fix is reverted** — I ran four mutations and confirmed only the intended tests broke:

| Mutation | Tests that failed |
|---|---|
| `canGenerate` back to always requiring a platform | 1 (the email-mode gate) |
| Card guard inverted | 3 (email hides / paid keeps / unbound default) |
| `platforms` always sent on generate | 1 (the generate payload) |
| `platforms` always sent on refine | 1 (the refine payload) |
| **Server: email falls through to ad-copy generation** | 1 |
| **Server: `!isEmail` dropped from the keyword guard** | 1 |
| **Server: email refine refusal removed** | 1 |

The server contrast case (`still generates ad copy and keywords when platforms is absent but the type is paid`) sends **no platforms either** — the only difference is `deliveryType` — so it pins that absence still means the paid default, and that the two email assertions are not passing for an unrelated reason. That test caught a real harness bug while I wrote it: a `body: null` fetch stub made the paid path fail before the copy stage, which would have made the email assertions vacuous.

The card assertions use `querySelector`, not `debugElement.query` — the latter walks the logical tree and resolves nodes that are no longer rendered, which would make them vacuous.

Also confirmed `strictTemplates` catches a typo in the new binding: `deliveryType="emial"` fails the build with `TS2820: Type '"emial"' is not assignable to type 'CampaignDeliveryType'`.

- `yarn build` — green
- `yarn lint` — clean (1 pre-existing warning in an untouched file)
- Full suite — **1098 + 267 tests passing**, including all 79 campaigns specs

[LFXV2-3201]: https://linuxfoundation.atlassian.net/browse/LFXV2-3201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrautela365:
## Second commit: the first one was a half-fix

Flagging this myself rather than waiting for review to find it, because the PR description was wrong when I opened it.

The first commit omitted `platforms` from the email generation request and claimed that meant "no ad channels". **It does not.** `campaign-proxy.service.ts` reads an absent list as the paid default at five sites:

```ts
const selectedPlatforms = body.platforms?.length ? body.platforms : ['google-ads'];
```

and the keyword branch opts absence in explicitly:

```ts
if (body.platforms?.includes('google-ads') || !body.platforms || body.platforms.length === 0) {
```

`?.length` is falsy for both `undefined` and `[]`, so the absent/empty distinction the client comment leaned on was erased server-side. An email brief still made a Google RSA/Display copy call and a keyword call, and the review step rendered the result. The client half looked complete while the server undid it — "fixed at one layer only".

Absence can't carry that meaning, because it already means "use the default" for every paid caller. So `deliveryType` is now on both brief request interfaces and the server is told rather than inferring.

One thing worth calling out: I did **not** run the ad-copy generator with an empty platform list, which was the smaller-looking fix. `buildCopySystemPrompt` builds from per-platform sections and ends with `Keys: <joined>` — with no platforms that asks the model for a JSON object with no keys. That's a broken call, not a cheap one. Generate skips the block; refine is refused with a message, since there's no email copy to refine until `email-copy` lands (LFXV2-3198).

Also in this commit: the Budget & Assets card is hidden for email and `totalBudget` dropped from the request. It isn't merely unread — it reaches the copy prompt as `Total Campaign Budget: $N`, and its helper text describes daily ad pacing under an Email tab.

### On the tests

Each server fix was verified by reverting it: the copy skip, the keyword guard and the refine refusal each fail exactly one test. The paid contrast case sends **no platforms either** — only `deliveryType` differs — so it pins that absence still means the default.

That contrast test earned its place immediately: it failed when I first wrote it because my `fetch` stub returned `body: null`, so the paid path died before the copy stage. Without it, both email assertions would have been passing for the wrong reason and I'd have shipped two vacuous tests.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/planning-tab/planning-tab.component.ts
```diff
@@ -458,7 +484,12 @@ export class PlanningTabComponent implements OnInit {
     const budgetStr = typeof budgetRaw === 'string' ? budgetRaw.trim() : String(budgetRaw ?? '');
     const request = {
       url: this.briefForm.controls.url.value.trim(),
-      platforms: [...this.selectedPlatforms()] as CampaignPlatform[],
+      // OMITTED for email, not sent empty. `platforms` drives which per-platform ad copy the
+      // generator produces — RSA headlines, Demand Gen copy, a LinkedIn targeting strategy — none
+      // of which an email brief has any use for. Sending the paid default would bill an AI call
+      // for copy nothing reads, and sending `[]` would claim the user deselected every channel
+      // rather than that channels do not apply here.
+      ...(this.isEmail() ? {} : { platforms: [...this.selectedPlatforms()] as CampaignPlatform[] }),
```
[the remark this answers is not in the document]
mrautela365: Correct, and this was a real defect — thank you. Fixed in 460331a2c.

You were right on both counts: `streamBrief` coerced absence back to `['google-ads']` at lines 691 and 750, `streamRefinedBrief` did the same at 820, and my new tests missed it because they stopped at a mocked `CampaignService`.

Took the fix you recommended: `deliveryType` is now on both `CampaignBriefRequest` and `CampaignBriefRefineRequest`, and the server branches on it while leaving paid callers untouched — absence still means `['google-ads']`, because that is what it has always meant for them.

One deliberate difference from the obvious approach: ad copy is **skipped** for email rather than generated with an empty platform list. `buildCopySystemPrompt` composes per-platform sections and ends with `Keys: <joined>`, so no platforms asks the model for a JSON object with no keys — a broken call rather than a cheap one. Keyword generation is excluded before the existing condition, which leaves its `!body.platforms` arm correct for paid callers.

On your 'including email-compatible refine validation': refine is **refused** for email with an error, since there is no email copy to refine until `email-copy` lands (LFXV2-3198). I also hid the Refine Brief button in email mode in 9eefc894d so the user is not walked into that error.

Added `campaign-proxy.service.spec.ts` — 7 specs at the server layer this time, not the mocked client. The contrast case sends no platforms either, so it pins that absence still means the paid default.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -688,66 +688,83 @@ export class CampaignProxyService {
       }
     }
 
+    // An email brief has no ad channels, so the paid default must not apply. `platforms` absence
+    // cannot carry that meaning on its own — it already means "use the default" for every paid
+    // caller — so the delivery type is read explicitly.
+    const isEmail = body.deliveryType === 'email';
     const selectedPlatforms = body.platforms?.length ? body.platforms : ['google-ads'];
```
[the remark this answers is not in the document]
mrautela365: Confirmed and fixed in 0e2b53514.

I had actually considered this branch while writing the previous commit and talked myself out of it, on the grounds that the email client never sends `linkedin-ads`. That was the wrong test: it is a **client** guarantee, and this is a server the client does not own. `{deliveryType: 'email', platforms: ['linkedin-ads']}` type-checks and would have spent the AI call.

Guard is now `!isEmail && selectedPlatforms.includes('linkedin-ads') && !isRefinement`, with the reasoning recorded in a comment so the `!isEmail` does not read as redundant and get removed later. Added the email/paid pair for that branch, and verified the guard by reverting it — the email test fails, the paid one still passes.

mrautela365:
## Suppressed comments — all three were right

Copilot put three findings in a `Suppressed comments (3)` block rather than as threads. Fixed in a6f0c7539; noting them here since suppressed items are easy to miss and none of these was a nit.

**1. `deliveryType` was used but never validated.** The adjacent `platforms` and `programType` both reject unsupported values; this one didn't. Bodies are cast rather than parsed at the controller, so `"emial"` fell through `=== 'email'` as paid marketing and launched ad-copy and keyword generation.

The direction matters: a request that *meant* to suppress that spend would have caused it. Now rejected on both stream paths with the same shape as its neighbours.

**2. My client specs couldn't catch a `deliveryType` regression.** They asserted only that `platforms` was absent — so deleting the field from either request left them green while restoring the paid default and the AI spend. That's the sharpest of the three, because it means the tests I'd pointed at as proof weren't testing the thing that does the work. Absence isn't what suppresses generation; the discriminator is. Both payload assertions now pin it, and I confirmed the generate one fails when the field is removed.

**3. `submitRefine` returned silently for email.** An email brief generates no copy, so `structuredCopy` is null and the `currentCopy` guard swallowed the case before my refusal could fire. Hiding the button in 9eefc894d fixed the common route to it, not the path — a restored paid brief or any future caller would still land there and see nothing happen. Now refuses before that guard, with the same message the server returns.

### Verification

Every fix in this PR has been checked by reverting it and confirming the intended test fails — 10 mutations across the branch now, including the three above. Full suite: 1102 + 269.

One caveat worth stating: the paid contrast test earned its keep by failing when I first wrote it. A `body: null` fetch stub made the paid path die before the copy stage, which would have left both email assertions passing for the wrong reason.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/planning-tab/planning-tab.component.spec.ts
```diff
@@ -614,3 +615,323 @@ describe('PlanningTabComponent brief read-back', () => {
...
+    // The input is additive: the paid container binds nothing, so an omitted binding must keep
+    // exactly the pre-3201 behaviour rather than silently switching a live tab to email.
+    await build();
+    expect((fixture.nativeElement as HTMLElement).querySelector('[data-testid="planning-platforms-section"]')).not.toBeNull();
+  });
+
+  /**
+   * The server refuses an email refine, so offering the button would walk the user into a
+   * guaranteed error. Asserted on the review step, which is the only place it renders.
+   */
```
dealako: [minor] Orphaned JSDoc block annotates the wrong test

Issue: This block — "The server refuses an email refine, so offering the button would walk the user into a guaranteed error..." — sits directly above `it('does not look up a saved brief in email mode')` (line 725). But that test already has its own correct comment (the persistence-keyed-on-(foundation,event) block right below this one). This orphan actually describes `it('does not offer Refine Brief in email mode')` at line 755, which has no comment.
Proof: Lines 712-715 vs the saved-brief test at 725 (whose real doc is 716-724) vs the uncommented Refine-visibility test at 755.
Why it matters: A comment documenting a different test than the one it annotates is worse than none — a maintainer editing the saved-brief test is told it's about refusing refine.
Fix: Move this block down to sit above `it('does not offer Refine Brief in email mode')` at line 755.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/planning-tab/planning-tab.component.spec.ts
```diff
@@ -614,3 +615,323 @@ describe('PlanningTabComponent brief read-back', () => {
...
+    expect(generateBrief).toHaveBeenCalledTimes(1);
+    expect(generateBrief.mock.calls[0][0].platforms).toEqual(['google-ads']);
+  });
+
+  /**
+   * Refine is the second request that carries `platforms`, and it is reachable in email mode —
+   * nothing gates the refine panel on delivery type. Fixing only `generate` would have left the
+   * ad copy to come back on the first refine, which is the "fixed at one layer only" shape.
+   */
+  /**
```
dealako: [minor] Same misplacement: JSDoc annotates the wrong test

Issue: This block — "Refine is the second request that carries `platforms`..." — sits above `it('refuses an email refine...')` (line 899), which already has its own correct comment below it (892-899). This orphan describes `it('sends platforms in the refine request in paid mode')` at line 919, which is uncommented.
Proof: Lines 887-891 vs the refuse-refine test at 899 (real doc 892-898) vs the uncommented paid-refine test at 919.
Why it matters: Misleads a future editor about what the paid-refine assertion pins.
Fix: Move this block down to sit above `it('sends platforms in the refine request in paid mode')` at line 919.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/planning-tab/planning-tab.component.ts
```diff
@@ -9,13 +9,14 @@ import { ButtonComponent } from '@components/button/button.component';
...
 import { ProjectContextService } from '@services/project-context.service';
-import { catchError, combineLatest, debounceTime, distinctUntilChanged, map, of, skip, Subject, Subscription, switchMap } from 'rxjs';
+import { catchError, combineLatest, debounceTime, distinctUntilChanged, filter, map, of, skip, Subject, Subscription, switchMap } from 'rxjs';
 
 import type {
   CampaignBriefLoadResult,
   CampaignBriefOutput,
   CampaignBriefRefineRequest,
   CampaignEventDetails,
+  CampaignDeliveryType,
```
dealako, line 18: [nit] Import out of alphabetical order

Issue: `CampaignDeliveryType` is inserted after `CampaignEventDetails`, breaking the otherwise-alphabetical member order in this `@lfx-one/shared/interfaces` type-import block.
Proof: Line 18 is `CampaignEventDetails`, line 19 `CampaignDeliveryType` — 'D' should precede 'EventDetails'. Notably the same type was added in correct alphabetical position in `campaign-proxy.service.ts`, so the two files order the identical import differently.
Why it matters: Minor consistency drift; lint won't catch it (no member-sort rule).
Fix: Move `CampaignDeliveryType` above `CampaignEventDetails`.

dealako (dismissed):
Hi @mrautela365 👋 — thorough review here.

**Overall impression:** This is a model PR. The core design call the ticket owned — sharing one `deliveryType`-moded planner instead of forking ~800 lines — is the right one, and it's justified in the description rather than asserted. What stands out most is the honesty of the iteration: the first commit's "omit `platforms` = no ads" assumption was genuinely wrong (the server coerces absence back to `['google-ads']` at five sites), you flagged it yourself before review caught it, and then carried an explicit `deliveryType` discriminator through both request contracts and every server branch (ad-copy, keyword, LinkedIn-strategy, refine refusal). The `deliveryType` validation closes the trust boundary — an `"emial"` typo now fails loud instead of silently generating paid spend. Tests are mutation-verified and use `querySelector` (not `debugElement.query`) to avoid vacuous logical-tree matches, with survival assertions guarding against no-review-step vacuity.

I ran three parallel reviewers across security/privacy, correctness/performance/tests, and style/docs. Security and correctness both came back clean — I independently traced `deliveryType` from the wire through `streamBrief`/`streamRefinedBrief` and both client `generate()`/`submitRefine()` paths and found the branching sound and the paid path untouched (the paid-contrast test that sends *no* platforms but `deliveryType: 'paid-marketing'` is a nice touch — it pins that absence still means the default).

The bot findings (Cursor + Copilot, including the suppressed block) were all real and all already resolved in-branch (`460331a2c`, `0e2b53514`, `a6f0c7539`, `96e4dfa96`) — I confirmed each fix against current code. Nothing outstanding from them.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 1 — two spec JSDoc blocks annotate the wrong `it()` (each describes a test that appears later, uncommented)
- ⚪ Nit: 1 — `CampaignDeliveryType` import out of alphabetical order (and ordered differently than in the sibling server file)
- ❔ Question: 0

Both are confined to comment placement / import order — zero behavioral impact. Neither blocks merge.

**Decision: ✅ Approved with minor comments** — the two inline notes are cleanup, take them or leave them.

dealako (dismissed):
Hi @mrautela365 👋 — quick follow-up on `24891cda2`.

👏 **Nice work** — you closed out all three items from the prior round in a tidy docs-only commit, and the reasoning in the message is exactly right. Two things I especially appreciated:

- You didn't just relocate the second orphaned docblock — you recognized it was a *leftover* from when `refuses an email refine…` was still a payload-assertion test, not a misplacement, and **removed** it rather than reattaching a stale comment. That's the correct call; the refusal test already carries its own accurate doc.
- The first block landed exactly where it belongs — above `does not offer Refine Brief in email mode`, which previously had no doc of its own.

**Revision tracking (all 3 resolved in `24891cda2`):**
- ✅ **Minor — orphaned docblock #1** ("server refuses an email refine…"): moved to annotate `does not offer Refine Brief in email mode`. Verified.
- ✅ **Minor — orphaned docblock #2** ("Refine is the second request that carries `platforms`…"): removed as a stale leftover; the refuse-refine test retains its own correct comment. Verified.
- ✅ **Nit — import order**: `CampaignDeliveryType` now sorts before `CampaignEventDetails`, matching the sibling server file. Verified.

The change is comments + one import swap — no behavioral surface touched, so nothing to re-scan for correctness/security. Copilot's latest pass (18:36) generated no new comments; nothing outstanding from the bots.

**Open issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

**Decision: ✅ Approved** — clean. Nothing left on my end; good to merge once required checks go green.

mrautela365:
@dealako — rebased onto main (the cutover #1436 landed). Auto-merged cleanly: this branch changes the planner's email mode while #1436 changed the create and poll paths, so they touch different regions of the shared files.

Your three items from the last pass are already in (`24891cda2`): the two orphaned spec docblocks are rebound to the tests they describe, and the `CampaignDeliveryType` import is back in alphabetical order.

Swept the worktree for conflict markers rather than trusting the merge exit code — none.

`1206 + 292` passing, tsc clean on both tsconfigs, lint 0 errors, 0 commits behind main.

mrautela365:
@dealako — your approval was dismissed by a rebase, not by new findings. Re-requesting.

**#1437 merged**, which conflicted this branch: its email-planner hoist and this PR's email-mode changes touch the same region. Both conflicts take **main's structure with this branch's content**:

- The panel is now visibility-toggled rather than `@switch`-cased — that's #1437's fix and this branch must not undo it. Main's rationale comment is kept; the `KNOWN GAP, tracked as LFXV2-3201` paragraph it carried is replaced, since **LFXV2-3201 is this PR** and the gap is closed.
- The planner tag takes main's post-hoist indentation with this branch's `deliveryType="email"` binding — the whole point of the PR, and the thing a careless resolution would have silently dropped.

Verified after resolving: 0 conflict markers, exactly one `deliveryType="email"`, no stale KNOWN GAP paragraph left behind.

`1220 + 301` passing — that count includes **both** #1437's hoist tests and this branch's email-mode tests, which is the check that the merge kept both features rather than one. tsc clean on both tsconfigs, lint 0 errors, 0 commits behind main.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/planning-tab/planning-tab.component.ts
mrautela365: Fixed in `24891cda2` — `CampaignDeliveryType` now sits between `CampaignBriefRefineRequest` and `CampaignEventDetails`, so the block is alphabetical and matches how `campaign-proxy.service.ts` orders the same import.

mrautela365:
@dealako — up to date with main (#1439 merged), 0 open threads, re-requesting.

All three of your items landed in `24891cda2`: the two orphaned spec docblocks are rebound to the tests they describe, and the `CampaignDeliveryType` import is alphabetical.

The approval you gave earlier was dismissed by the rebases, not by new findings — nothing has been flagged on this branch since.

`1237 + 301` passing · tsc clean on both tsconfigs · lint 0 errors · 0 commits behind main.

mrautela365:
@dealako — last of the six. Three findings from Copilot on the current HEAD, all real; fixed in `be1d3b66f`.

**One is user-facing.** An HTTP caller refining an email brief got *"currentCopy is required"*. That's true but useless — an email brief **has** no generated copy (`structuredCopy` is null, `currentKeywords` empty), so the paid-only field checks fired before the service's email refusal could. The message named a field the caller cannot supply and hid the actual reason.

The check now runs first in the controller. The service keeps its own guard — it isn't the only caller — but only this path is reached over HTTP, so only this one decides what the user reads.

Copilot's sharpest point was *why* I missed it: my spec called the service directly, so it never exercised the controller's validation order. There are now controller tests for both the email refusal and the paid case that must still validate.

**Two comment/message corrections:**

- The extraction-failure status said *"generating copy from URL..."* for email, where no copy is generated at all. `isEmail` is now resolved before that block. A status line naming work the stream won't do is the same class of falsehood as a silent skip.
- `buildCopyPrompt`'s comment claimed "the two ad-copy paths ... refuse email before calling it". Only **generate** reaches that function — refine builds its prompt with `buildRefinePrompt`. The comment named the wrong function for half its own claim.

`1239 + 308` passing · tsc clean on both tsconfigs · lint 0 errors · 0 open threads · 0 behind main. Controller guard revert-verified.

mrautela365:
Two more from Copilot, both on the fix I made one commit ago — fixed in `b54437d56`.

**A repo-convention violation I introduced.** The email refusal used a manual `res.status(400).json()`. Every sibling check in that method uses `ServiceValidationError`, and `docs/reviews/backend-checklist.md` §8 forbids the manual form outright.

The consequence is worse than style: going around the error middleware skips the standard error shape **and** its centralized log line — so the one refusal a caller is most likely to hit would have been the one the logs never recorded. Now routed through `ServiceValidationError.forField('deliveryType', …)`, and the spec asserts the error class and its response body rather than a raw `res.json` call.

**A comment fix I applied to the wrong copy.** `buildRefinePrompt` still carried *"buildCopyPrompt is only reached from the two ad-copy paths"*. My previous commit corrected that sentence inside `buildCopyPrompt` and missed the duplicate living in this function — where it names the wrong helper entirely. Each helper now describes its own path: generate uses `buildCopyPrompt`, refine uses `buildRefinePrompt`, and `streamRefinedBrief` rejects email before reaching it.

`1239 + 308` passing · tsc clean on both tsconfigs · lint 0 errors · 0 open threads · 0 behind main. Guard revert-verified.

mrautela365:
One more from Copilot, and it's a gap in **my test**, not the code — fixed in `6877146cb`.

`does not generate ad copy for an email brief` drained the stream and threw the events away, asserting only an **absence**. An email brief that failed immediately satisfies that just as well: no ad copy is generated when nothing runs at all. So the test passed for both the behaviour I wanted and its opposite.

It now captures the events and asserts the stream reached `done` and emitted no `error` — the email path must be **usable**, not merely quiet.

Verified with the exact scenario Copilot described: removing `'email'` from `SUPPORTED_DELIVERY_TYPES` makes the stream error out early. That fails the strengthened test and **passed** the old one.

Worth naming the pattern, since it's the second time today: an assertion that something *didn't* happen is satisfied by the whole feature not running. The paid contrast case I'd added proves the paid path works — it says nothing about whether the email path does.

`1282 + 308` passing · tsc clean on both tsconfigs · lint 0 errors · 0 open threads · 0 behind main.

apps/lfx-one/src/server/services/campaign-proxy.service.ts
```diff
@@ -613,7 +615,22 @@ export class CampaignProxyService {
...
+      yield { type: 'error', data: `Unsupported deliveryType. Supported: paid-marketing, email.` };
+      return;
+    }
+
     const isRefinement = !!body.refineFeedback && !!body.previousCopy;
+    // Resolved HERE rather than beside the copy block below, because the extraction-failure
+    // message further down promises "generating copy from URL" — which is false for email, where
+    // no copy is generated at all. A status line that names work the stream will not do is the
+    // same class of falsehood as a silent skip.
+    const isEmail = body.deliveryType === 'email';
```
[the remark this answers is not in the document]
mrautela365: Confirmed and fixed. `/brief/generate`'s own refinement mode was a second door onto the same refusal: an email refinement skipped the scrape (because `isRefinement`) and every generator (because `isEmail`), then emitted `done` — a successful no-op, worse than an error because the caller is told the work happened. Now rejected with the same message `streamRefinedBrief` uses, and the test asserts `error` AND the absence of `done`, since the quiet completion is the defect.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -152,6 +152,49 @@ export class CampaignController {
...
+    // VALIDATED first, not just matched. An exact `=== 'email'` test lets a typo through:
+    // `'emial'` falls past it into those same paid-only checks and produces the same misleading
+    // message, for a caller whose only mistake was a misspelling.
+    //
+    // Derived from the shared `CAMPAIGN_DELIVERY_TYPES` rather than a fourth hardcoded list — the
+    // service already keeps its own copy, and a third one here would be the one that drifts.
+    const supportedDeliveryTypes = new Set<string>(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id));
+    if (body.deliveryType !== undefined && !supportedDeliveryTypes.has(body.deliveryType)) {
+      _next(
+        ServiceValidationError.forField('deliveryType', 'deliveryType must be one of: paid-marketing, email', {
```
dealako, line 170: [nit] Controller refine message hardcodes the list the Set is derived from

Issue: Two lines up you build the validation Set from the shared constant (`new Set(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id))`), and the comment frames it as avoiding "a fourth hardcoded list ... would be the one that drifts." After b3e788c26 gave the service the same treatment, this error-message string is now the *last* hardcoded copy of `paid-marketing, email`.
Proof: campaign.controller.ts:167 derives `supportedDeliveryTypes`; line 170 hardcodes the tail. The sibling `platform` check at line 933 already does the right thing — it interpolates `[...SUPPORTED_STATUS_PLATFORMS].join(', ')` into its message. Adding a third delivery type would leave the Set accepting it while this message (and the spec asserting the exact string) still names two.
Why it matters: Purely maintainability — current behavior is correct. But it's the exact drift the adjacent comment claims to prevent, and it's now inconsistent with the sibling check in the same file.
Fix: Mirror line 933 — build the tail from `CAMPAIGN_DELIVERY_TYPES.map((d) => d.id).join(', ')` instead of the literal.

dealako (approved):
Hi @mrautela365 👋 — follow-up on the 8 new campaigns commits since `24891cda2` (I scoped the review to your work and excluded the merged-in `origin/main` changes from #1436/#1437/#1438/#1439 etc.). You pushed `b3e788c26` while I was writing this — factored in below.

👏 **Nice work** — this is a genuinely thorough hardening of the email-refine refusal, and the standout is that you closed *every* door rather than the one Copilot named:

- **Three independent refusal doors, all fail-secure** — the controller (`be1d3b66f`), the service `streamRefinedBrief` for non-HTTP callers, and — the subtle one — the `/brief/generate` stream's own refinement path (`ef35b603d`), where an email refinement skipped the scrape (`isRefinement`) *and* every generator (`isEmail`) then emitted `done`: a **successful no-op**, worse than an error because the caller is told it worked. Real latent bug; the fix lands before any scrape/generator runs.
- **Validate, don't match** (`52c83b022`) — `deliveryType` is now rejected against the shared allowlist at the trust boundary, so `'emial'` fails loud instead of silently falling through `=== 'email'` into the paid path.
- **Routed through `ServiceValidationError`** (`b54437d56`) — dropping the manual `res.status().json()` for the standard error shape + centralized log line, per backend-checklist §8.
- **Single source of truth** (`b3e788c26`) — the service's `SUPPORTED_DELIVERY_TYPES` now derives from `CAMPAIGN_DELIVERY_TYPES` instead of a hand-written copy, closing the type-checks-but-disagrees-at-runtime drift between controller and service. Clean and behavior-preserving (tests unchanged).
- **Tests assert the brief COMPLETES** (`6877146cb`), not just an absence, and env/globals are restored after the proxy spec (`20c13ccc2`) — no cross-test leakage.

I ran three parallel reviewers (security/privacy, correctness/perf/tests, style/docs) over the isolated diff. Security traced all three doors fail-secure with no new injection/PII/insecure-default surface; style confirmed the `ServiceValidationError` convention and clean import order.

**Revision tracking — my 3 prior items stay resolved:** the two rebound/removed spec docblocks and the `CampaignDeliveryType` import order are untouched — no regressions.

**AI reconciliation:** Copilot's `21:29` finding (generate-stream refinement no-op) → fixed in `ef35b603d` ✅. Copilot's service-duplicate nudge → fixed in `b3e788c26` ✅. No open bot threads.

**Open issue count (carried + new):**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — the *controller's* refine error-message string (line 170) is now the last hardcoded copy of the delivery-type list; the sibling `platform` check one method down already derives its message from the constant (see inline)

**Decision: ✅ Approved** — the lone nit is take-it-or-leave-it maintainability and does not block. Clean to merge once required checks go green.

PR #1455, opened 2026-08-13
fix(votes): poll comment fixes across ballot, results, and BFF (#1448)

### LINKED ISSUE #1448: Vote comments: fix validation gaps, misleading voter states, and ballot display issues
## Problem

A code review of the recently shipped vote comments feature surfaced a set of issues that affect vote creators and voters:

- Vote creators can enter comment questions that are too long or too many and only find out via a generic "failed to save" error — drafts with over-length comment questions save silently and then fail on submission.
- Voters opening a vote's results can be wrongly shown "access denied" when a transient error occurs while checking their participation, hiding results and the comments panel they are entitled to see.
- The comments section in vote results disappears entirely when nobody has responded yet, so organizers cannot see which questions were even asked.
- The abstain explanation implies that comments are discarded along with ballot choices, when in fact comments are still submitted.
- On ballots with multiple radio groups, selecting an option in one group can clear the voter's selection in another.
- Votes carrying many comment questions render every text box inline on the dashboard pending-actions row, crowding it.

## Outcome

- Comment questions are validated at every save and submit point with clear, specific messages instead of raw upstream failures.
- A transient error while checking a voter's participation shows an honest "something went wrong" state, never a false access-denied.
- Vote results always show every comment question, with an explicit "no responses yet" state when empty.
- The abstain explanation accurately tells voters their comments are still submitted.
- Radio groups on a ballot behave independently.
- Votes with several comment questions open in the full cast-vote panel instead of the cramped inline dashboard ballot.

## Success criteria

- Creating, updating, or draft-saving a vote with an invalid comment question (blank, over-length, or too many) is blocked with a clear field-level message.
- A transient failure while verifying a voter's participation shows an error state with guidance to retry — never an access-denied screen.
- The vote results panel lists all comment questions, showing "No responses yet" for unanswered ones.
- The abstain helper text states that comments are still submitted.
- Selecting a radio option in one ballot question never clears selections in another.
- A vote with more than two comment questions casts via the slide-in panel, not the inline dashboard row.

audigregorie:
## Summary

Fixes a cluster of poll comment issues spanning the ballot, the results drawer, and the BFF. Comment prompts are now validated server-side before crossing to the upstream service, the results view shows every configured comment prompt (including ones with no responses yet), a failed participation check no longer masquerades as access-denied, and the ballot surfaces get correctness fixes for radio grouping, prompt-heavy votes, and draft saves. The comment-form logic that was duplicated between the inline ballot and the cast drawer is extracted into `@lfx-one/shared/utils`.

Resolves #1448

### Comment questions on vote results

| Before | After |
|---|---|
| The Comments section only appeared when at least one comment had been submitted — a prompt with zero responses was invisible, so an empty vote looked like it had no comment questions at all. | Every comment question configured on the vote is listed, and unanswered ones show a "No responses yet" placeholder. |

### Failed participation check

| Before | After |
|---|---|
| If the background check that confirms whether you've voted hit a transient error, the results drawer told you "you don't have access to this vote" — alarming and wrong. | A transient failure shows a dedicated "We couldn't verify your participation" message suggesting you reopen the vote; the access-denied message only appears when access was actually denied. |

### Votes with many comment questions

| Before | After |
|---|---|
| A vote with several comment questions tried to render them all inside the compact inline ballot on the dashboard's pending-actions row, crowding the layout. | Votes with more than two comment questions open in the full voting drawer instead, where there's room to answer them properly. |

### Saving draft votes with over-length comment questions

| Before | After |
|---|---|
| Saving a vote as a draft skipped the normal form validation, so an over-length comment question could be stored and only fail later on publish. | Draft save checks comment-question length up front, highlights the offending fields, and shows a warning instead of saving. |

### Single-choice selections on ballots

| Before | After |
|---|---|
| Picking a radio option in one question could silently clear the selection in another question on the same form. | Radio options are grouped per question, so each question keeps its own independent selection. |

## Technical changes

`packages/shared/src/utils/vote.utils.ts` — Shared ballot helpers
- Adds `reconcileCommentFormControls`, `getCommentPromptsData`, and `buildCommentResponses` — the comment-form reconciliation and response-building logic previously duplicated in both ballot components.
- `getCommentPromptsData` drops prompts whose control is missing, since a `computed` can evaluate before the reconcile pass has added controls.

`packages/shared/src/utils/url.utils.ts` — URL utils
- Tallies each bracket pair's balance once up front so trailing-bracket stripping stays O(n) instead of re-counting per strip (O(n²) on adversarial input).

`packages/shared/src/constants/poll.constants.ts` — Shared constants
- Adds `VOTE_INLINE_BALLOT_MAX_COMMENT_PROMPTS` (2) — the threshold for routing a vote from the inline ballot to the cast drawer.

`packages/shared/src/interfaces/poll.interface.ts` — Shared types
- Doc-comment clarifications only: `profile_picture` is reserved and not yet rendered in the results drawer; `CommentResultPageView.totalResponses` is the post-BFF-cap count (pre-cap total lives in `total_responses`).

`apps/lfx-one/src/server/controllers/vote.controller.ts` — BFF vote controller
- `createVote` / `updateVote` now run `validateRequestBody` plus a new `validateCommentPrompts` pass over `poll_comment_prompts`: array shape, `VOTE_COMMENT_PROMPT_MAX_COUNT` cap, non-blank prompt text, and `VOTE_COMMENT_PROMPT_MAX_LENGTH` measured in code points so emoji/non-BMP text isn't rejected early.
- Prompt entries are rebuilt from validated fields only, so unexpected extra properties never cross the BFF boundary.
- Makes results logging null-safe (`results?.poll_results`).

`apps/lfx-one/src/server/services/vote.service.ts` — BFF vote service
- Normalizes `comment_result.responses` to an array before the cap/sort pass, staying robust if the field is missing during deploy-order skew.
- Makes the debug logging null-safe on `results`.

`apps/lfx-one/src/app/modules/votes/components/vote-ballot-inline/vote-ballot-inline.component.ts, vote-cast-drawer/vote-cast-drawer.component.ts` — Ballot surfaces
- Both drop their local `rebuildCommentForm` / `buildCommentResponses` copies in favor of the shared helpers.

`apps/lfx-one/src/app/modules/votes/components/vote-cast-drawer/vote-cast-drawer.component.html` — Cast drawer template
- Adds "Vote Details" and "Questions" section headings and restyles the comments block with a top border separator.
- Abstain copy now notes that comments are still submitted alongside an abstention.

`apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts, vote-results-drawer.component.html` — Results drawer
- The comments section is driven by the vote detail's `poll_comment_prompts` (the results payload only reports prompts that have responses) and left-joins responses by `prompt_id`; degrades to the results payload's prompts when the detail fetch supplied none.
- Adds an `@empty` "No responses yet" state per prompt.
- Adds a `myResponseError` signal: non-404 my-response failures render a dedicated error card, and `voterState` only returns `access_denied` when no error occurred.

`apps/lfx-one/src/app/modules/dashboards/components/pending-actions/pending-actions.component.ts` — Pending actions
- Votes exceeding `VOTE_INLINE_BALLOT_MAX_COMMENT_PROMPTS` comment prompts route to the cast drawer instead of the inline ballot.

`apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts` — Vote management
- Draft saves validate comment-prompt length, mark invalid prompts as touched, and show a warning toast instead of persisting.

`apps/lfx-one/src/app/shared/components/radio-button/radio-button.component.html` — Radio button
- Binds `[name]` so PrimeNG's `RadioControlRegistry` groups radios per control; previously every radio in a form shared `name === undefined`, so selecting one deselected all others.

`packages/shared/src/utils/vote.utils.spec.ts` — Tests
- Removes the vote utils spec (114 lines covering the request builders and comment-prompt mapping); the tested functions remain in `vote.utils.ts`.


apps/lfx-one/src/app/modules/votes/components/vote-results-drawer/vote-results-drawer.component.ts
```diff
@@ -403,11 +425,17 @@ export class VoteResultsDrawerComponent {
...
             this.myResponseLoading.set(false);
+            this.myResponseError.set(false);
             return of(null);
           }
           this.myResponseLoading.set(true);
+          this.myResponseError.set(false);
           return this.voteService.getMyVoteResponse(id).pipe(
-            catchError(() => of(null)),
+            // The service maps 404 to null; anything reaching this catch is a real failure — surface an error state instead of access-denied.
+            catchError(() => {
```
ahmedomosanya, line 445: **Blocker.** `initMyResponse()` only keys on `voteId` + `audience`. Hosts keep `selectedVoteId` set when the drawer closes, so setting the same id again is a no-op and `myResponseError` stays latched — the "Try reopening this vote" copy cannot recover.

Could `visible` join the `combineLatest` so each open starts a fresh call? `initVoteResults()` has the same `voteId`-only key, so its reopen copy has the same hole.

```ts
combineLatest([
  this.voteId$,
  toObservable(this.audience),
  toObservable(this.visible),
]).pipe(
  switchMap(([id, audience, visible]) => {
    if (!id || audience !== 'voter' || !visible) {
      this.myResponseLoading.set(false);
      this.myResponseError.set(false);
      return of(null);
    }
    this.myResponseLoading.set(true);
    this.myResponseError.set(false);
    return this.voteService.getMyVoteResponse(id).pipe(
      catchError(() => {
        this.myResponseError.set(true);
        return of(null);
      }),
      finalize(() => this.myResponseLoading.set(false)),
    );
  }),
)
```

apps/lfx-one/src/server/controllers/vote.controller.ts
```diff
@@ -441,4 +456,51 @@ export class VoteController {
...
+        throw ServiceValidationError.forField(
+          `poll_comment_prompts[${index}].prompt`,
+          `prompt must be ${VOTE_COMMENT_PROMPT_MAX_LENGTH} characters or fewer`,
+          validationContext
+        );
+      }
+    }
+
+    // Rebuild from the validated fields only, so unexpected extra properties never cross the BFF boundary.
+    return prompts.map((entry) => ({ prompt: entry.prompt }));
```
ahmedomosanya: **Nice-to-have.** `trim()` is used for the blank check, but the rebuild still sends `entry.prompt` as-is. The client mapper already persists `prompt.trim()`. Could this be `{ prompt: entry.prompt.trim() }` so a direct API caller cannot store padded text that later looks blank in the UI?

ahmedomosanya (changes requested):
Request Changes.

The ballot, BFF validation, radio grouping, and results empty-state work looks good. One blocker: the new participation-error card tells the voter to reopen the vote, but reopening the same row does not refetch `my-response`, so the error stays latched. That misses #1448's retry criterion.

Nits (non-blocking): the new required-prompt message has no `role="alert"`; the draft toast always mentions both blank and over-length even when only one failed.

apps/lfx-one/src/app/shared/components/radio-button/radio-button.component.html
```diff
@@ -4,7 +4,10 @@
 @if (form() && control()) {
   <ng-container [formGroup]="form()!">
     <div class="flex items-center gap-2">
-      <p-radioButton [formControlName]="control()!" [value]="value()" [inputId]="inputId() || control()!" (onChange)="onRadioChange($event)"> </p-radioButton>
+      <!-- [name] is required: PrimeNG's RadioControlRegistry groups radios by (root FormGroup, name) —
+           without it every radio in the form shares name===undefined and selecting one deselects all others. -->
+      <p-radioButton [formControlName]="control()!" [name]="name()" [value]="value()" [inputId]="inputId() || control()!" (onChange)="onRadioChange($event)">
```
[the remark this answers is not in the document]
audigregorie: Verified against the installed `primeng@20.3.15` source — the throw premise doesn't apply here, so no change is needed:

- Angular's `RadioControlNameMixingError` is only thrown by the **native** `RadioControlValueAccessor._checkName()`, which attaches to `input[type=radio][formControlName]`. In this component, `formControlName` binds on the `p-radioButton` host element, whose value accessor is PrimeNG's own `RADIO_VALUE_ACCESSOR`, and PrimeNG's inner `<input type="radio">` carries no form directive — the native check never runs.
- PrimeNG's `RadioButton`/`BaseEditableHolder` contain no name-vs-formControlName consistency check. Its `RadioControlRegistry.isSameGroup` keys grouping on `name()` alone, and all `rsvp-scope-modal` options share `name="rsvp-scope"`, so the group is coherent and renders/behaves identically.
- All other form-backed callers already pass `name === control`.

The suggested `[name]="control()!"` derivation is functionally safe for current callers but would silently ignore an explicitly-passed `name`, so we're keeping the caller-supplied binding. Resolving as not-applicable.

ahmedomosanya (approved):
LGTM. The reopen path now refetches `my-response`, and the BFF rebuild trims prompt text. Thanks for the close-animation and stale-row follow-ups.

luismoriguerra (approved):

PR #1468, opened 2026-08-13
fix(weekly-brief): attribute mailing-list share to the real impersonator

manishdixitlfx:
## Summary

Fixes [LFXV2-3093](https://linuxfoundation.atlassian.net/browse/LFXV2-3093): during impersonation, `WeeklyBriefService.shareBrief()` (the "Share to Mailing List" action on the Weekly Brief card) authorized and sent under the **impersonated target's** identity/token, not the real impersonating LF staff member's — unlike `/share-slack` and `/rating`, which are already blocked outright during impersonation via `blockDuringImpersonation`.

`shareBrief` creates a persisted newsletter draft before sending, so this was a real, externally-visible, hard-to-retract action (an email to an entire committee mailing list) that could be triggered while impersonating and misattributed to the target with no trace back to the actual staff member.

## Decision (discussed in Plan Mode before implementation, not assumed)

The ticket asked for one of two explicit options to be chosen and documented:

1. **Block outright** during impersonation, matching `/share-slack`'s precedent — simplest, but would break the support workflow of sharing a brief while impersonating a chair.
2. **Allow, but attribute to the real impersonator** — keep the action usable during impersonation, but resolve authorization and the outgoing identity from the real staff member instead of the target.

**Option 2 was chosen**, plus a follow-on sub-decision: the project-writer *authorization* check was also moved to the real identity (not left on the target's), for full consistency between "who is authorized" and "who the send is attributed to" — rather than authorizing against one identity while sending as another.

### Implementation

- Two new helpers in `auth-helper.ts`: `getRealEmail(req)` (always the real OIDC email, never the impersonation target's) and `resolveRealAccessToken(req)` (the real user's own bearer token even while impersonating — reads `req.oidc.accessToken` directly, since it's never touched by the impersonation-token swap in `auth.middleware.ts`, refreshing it if expired since that middleware's own refresh path is skipped whenever impersonation is active).
- `shareBrief()`: from the project-writer check through the newsletter create/send/cleanup, `req.bearerToken` is temporarily swapped to the real token (save/mutate/restore via `try/finally`, restored on every exit path including thrown errors) and `getRealEmail` replaces `getEffectiveEmail` for the reply-to address. Precondition reads (brief fetch, committee fetch, mailing-list check) stay on the effective/target identity, unchanged — impersonation still shows staff what the target sees for reads.
- Fails closed (`AuthenticationError`, never falls back to the impersonation token) if the real token can't be resolved.
- Two architecture docs (`docs/architecture/backend/impersonation.md`, `authentication.md`) updated so the new real-identity exception is documented alongside the existing `getEffective*` family rather than going stale.

## Review

Reviewed with three independent passes (correctness/security, repo-convention audit, and a full cross-commit sweep) — no critical or important findings. Minor findings addressed in the second commit: a comment that overclaimed exact parity with an existing (less-safe) M2M token save/restore precedent elsewhere in the codebase, two added tests exercising the token-restore-on-thrown-error path directly, and the doc updates above.

## Test plan

- [x] New/updated unit tests cover: non-impersonating baseline, impersonating (real token used for auth+send, effective token used for reads, restored afterward), expired-real-token refresh, real user denied project-writer access even when the target would pass, fail-closed on unresolvable real token, cleanup-on-rejection still under the real token, and token restoration on thrown errors (not just returned-false/rejected-later)
- [ ] Manual verification not performed — this is a real, hard-to-retract mailing-list send, so it wasn't exercised live against prod data; relying on unit test coverage + review instead

[LFXV2-3093]: https://linuxfoundation.atlassian.net/browse/LFXV2-3093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/server/services/weekly-brief.service.spec.ts
+++ b/apps/lfx-one/src/server/services/weekly-brief.service.spec.ts
@@ -26,6 +26,11 @@ const {
   buildCacheKey,
   getCommitteeForSlackShareMock,
   checkSingleAccessStrictMock,
+  getCommitteeByIdMock,
+  hasMailingListStrictMock,
+  createNewsletterMock,
+  sendNewsletterMock,
+  deleteNewsletterMock,
 } = vi.hoisted(() => {
   const valkeyStore = new Map<string, unknown>();
   const valkeyServiceMock = {
@@ -73,6 +78,13 @@ const {
     // that touch committeeService/accessCheckService.
     getCommitteeForSlackShareMock: vi.fn(),
     checkSingleAccessStrictMock: vi.fn(),
+    // shareBrief's (LFXV2-2914 / LFXV2-3093) own committee + newsletter collaborators —
+    // controlled per-test in the 'shareBrief' describe block below.
+    getCommitteeByIdMock: vi.fn(),
+    hasMailingListStrictMock: vi.fn(),
+    createNewsletterMock: vi.fn(),
+    sendNewsletterMock: vi.fn(),
+    deleteNewsletterMock: vi.fn(),
   };
 });
 
@@ -110,17 +122,25 @@ vi.mock('./microservice-proxy.service', () => ({
 vi.mock('./logger.service', () => ({
   logger: { startOperation: vi.fn(() => 0), success: vi.fn(), error: vi.fn(), warning: vi.fn(), debug: vi.fn(), info: vi.fn(), sanitize: (v: unknown) => v },
 }));
-// shareBrief's collaborators — shareBrief itself isn't exercised by this spec, but
-// WeeklyBriefService's constructor instantiates all three, so they must at least be
-// constructible without pulling in their own real import chains. getCommitteeForSlackShare /
-// checkSingleAccessStrict back shareToSlack's (LFXV2-3080) precondition chain — see the
-// 'shareToSlack' describe block below.
+// shareBrief's / shareToSlack's collaborators — getCommitteeForSlackShare / checkSingleAccessStrict
+// back shareToSlack's (LFXV2-3080) precondition chain (see the 'shareToSlack' describe block
+// below); getCommitteeById / hasMailingListStrict / createNewsletter / sendNewsletter /
+// deleteNewsletter back shareBrief's (LFXV2-2914 / LFXV2-3093) precondition + send chain (see the
+// 'shareBrief' describe block below).
 vi.mock('./committee.service', () => ({
   CommitteeService: class {
     public getCommitteeForSlackShare = getCommitteeForSlackShareMock;
+    public getCommitteeById = getCommitteeByIdMock;
+    public hasMailingListStrict = hasMailingListStrictMock;
   },
 }));
-vi.mock('./newsletter.service', () => ({ NewsletterService: class {} }));
+vi.mock('./newsletter.service', () => ({
+  NewsletterService: class {
+    public createNewsletter = createNewsletterMock;
+    public sendNewsletter = sendNewsletterMock;
+    public deleteNewsletter = deleteNewsletterMock;
+  },
+}));
 vi.mock('./access-check.service', () => ({
   AccessCheckService: class {
     public checkSingleAccessStrict = checkSingleAccessStrictMock;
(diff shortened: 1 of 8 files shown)


apps/lfx-one/src/server/utils/auth-helper.ts
```diff
@@ -146,6 +161,42 @@ export function isImpersonating(req: Request): boolean {
...
+ * (impersonation tokens can live up to ~24h; ordinary access tokens are typically shorter-lived).
+ * This refreshes it itself when needed, mirroring `extractBearerToken`'s own refresh step.
+ *
+ * Returns null when there is no real session token to resolve, or the refresh attempt fails —
+ * callers MUST treat null as "cannot safely attribute this action" and fail closed, never falling
+ * back to the impersonation token.
+ */
+export async function resolveRealAccessToken(req: Request): Promise<string | null> {
+  if (!isImpersonating(req)) {
+    return req.bearerToken ?? null;
```
dealako, line 195: [blocking] resolveRealAccessToken fails OPEN to the impersonation token on a mid-request expiry

Issue: The non-impersonating branch returns `req.bearerToken` as-is. During an impersonation request the middleware set `req.bearerToken` to the *target's* impersonation token. If `impersonationExpiresAt` elapses between the middleware run and this call, `isImpersonating(req)` flips to `false` (it requires `Date.now() < expiresAt`) while `req.bearerToken` still holds the impersonation token — so this returns the impersonation token instead of failing closed.

Proof: `auth.middleware.ts:extractBearerToken` sets `req.bearerToken = impersonationToken` only when `Date.now() < impersonationExpiresAt` at middleware time, and clears the session **only** when the token is already expired at that moment (`>= impersonationExpiresAt`). Nothing between the middleware and this helper mutates `req.bearerToken`, and `shareBrief` awaits `getCurrentBrief` (proxyRequest) + `getCommitteeById` first — widening the window. Once past expiry: session still holds `impersonationToken`, `isImpersonating()` → false, this returns `req.bearerToken` (= impersonation token), the `if (!realToken)` guard in `shareBrief` passes, and `checkSingleAccessStrict` + `createNewsletter`/`sendNewsletter` run under the TARGET's token — the precise LFXV2-3093 misattribution. It also desyncs reply-to (`getRealEmail` → real staff) from the sending principal (target). Matches Cursor Bugbot's Medium finding.

Why it matters: A security fix reintroduces its own fail-open. Even if a co-expired JWT 401s downstream in the aligned case, the documented contract ("never fall back to the impersonation token") is violated, and misalignment between `impersonationExpiresAt` and the JWT `exp` makes the mis-send reachable.

Fix: Gate the fallback on the *absence* of a lingering impersonation token, not on the time-based predicate. e.g.:
```ts
const lingeringImpToken = req.appSession?.['impersonationToken'];
if (!lingeringImpToken || req.bearerToken !== lingeringImpToken) {
  return req.bearerToken ?? null;
}
// impersonation context present → resolve the REAL token or fail closed
const accessToken = req.oidc?.accessToken;
...
```
Add a spec: impersonating session whose `impersonationExpiresAt` has passed but whose `bearerToken` still holds the imp-token must resolve to the real token (or null), never the imp-token.

apps/lfx-one/src/server/services/weekly-brief.service.spec.ts
```diff
@@ -1241,4 +1261,222 @@ describe('WeeklyBriefService', () => {
...
+      mockShareableBrief();
+      const req = buildImpersonatingReq({ realTokenExpired: true, refreshFails: true });
+
+      await expect(service.shareBrief(req, 'committee-1', 1)).rejects.toMatchObject({ statusCode: 401, code: 'AUTHENTICATION_REQUIRED' });
+      expect(checkSingleAccessStrictMock).not.toHaveBeenCalled();
+      expect(createNewsletterMock).not.toHaveBeenCalled();
+      expect(req.bearerToken).toBe('imp-token');
+    });
+
+    it('impersonating: cleans up the orphaned draft on a deterministic send rejection, still under the real token, and restores the impersonation token afterward', async () => {
```
dealako, line 1447: [minor] Test title claims "still under the real token" but never asserts the token at delete time

Issue: This test asserts `deleteNewsletterMock` was called with the right args and that `req.bearerToken` is restored to `imp-token` afterward, but never captures/asserts `req.bearerToken` *during* the delete call — so the invariant the title claims (cleanup runs under the REAL token) is unverified.

Proof: Unlike the happy-path test which uses `mockImplementationOnce` to capture `r.bearerToken` (e.g. `tokenDuringMailingListCheck`), this test only checks `toHaveBeenCalledWith(...)` + post-restore state. `deleteNewsletter` currently runs inside the outer try where `req.bearerToken = realToken`, so behavior is correct today — but a refactor moving delete past the finally restore would keep this test green.

Why it matters: The cleanup path is one of the exact writes LFXV2-3093 requires under the real identity; a silent attribution regression here wouldn't be caught.

Fix: `deleteNewsletterMock.mockImplementationOnce(async (r) => { tokenDuringDelete = r.bearerToken; })` and assert `tokenDuringDelete === 'real-staff-token'`, mirroring the create/send capture pattern.

dealako (changes requested):
Hi @manishdixitlfx 👋 — really thorough work on this one. The plan-mode decision record (Option 2, plus moving the *authorization* check onto the real identity for who-is-authorized/who-sends consistency), the try/finally save-mutate-restore on every exit path, the fail-closed `AuthenticationError`, and the doc updates all reflect genuine security-hardening care. The test matrix is broad and the reply-to/reads-vs-writes identity split is exactly right.

There's one hole, though, that lands right on the vulnerability this PR exists to close — a fail-**open** in `resolveRealAccessToken` when an impersonation session expires mid-request. Cursor Bugbot flagged the same thing (Medium); I traced `auth.middleware.ts` end-to-end and agree it's real. Detail inline.

### Issue count
- 🔴 Blocking: 1 — `resolveRealAccessToken` can return the impersonation token (fail-open) if `impersonationExpiresAt` elapses between the auth middleware and the helper call, reintroducing LFXV2-3093 in a narrow race window.
- 🟡 Minor: 1 — the "cleans up ... still under the real token" test never asserts the token *at delete time*, so the invariant its title claims is unverified.
- ⚪ Nit: 1 — no assertion that pre-write reads (`getCurrentBrief`/`getCommitteeById`) stay on the effective/target token during impersonation.
- ❔ Question: 0

### Bot reconciliation
- **Cursor Bugbot** (Medium, "Real token helper can return impersonation token") — ✅ agree, confirmed by tracing the middleware. This is the blocking finding below.
- **Copilot** — no comments generated.
- **CodeRabbit** — auto-review skipped on this repo.

### Decision
🔴 **Needs changes before approval** — the blocking fail-open reintroduces the exact misattribution this PR targets. It's a small fix (gate the fallback on the *absence* of a lingering impersonation token rather than the time-based `isImpersonating()` predicate). The window is narrow (session must expire during request processing, and a co-expired JWT would likely 401 downstream), but a security fix shouldn't ship its own fail-open — and closing it is cheap.

apps/lfx-one/src/server/utils/auth-helper.ts
```diff
@@ -146,6 +161,42 @@ export function isImpersonating(req: Request): boolean {
...
+  }
```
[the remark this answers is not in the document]
manishdixitlfx: Done — gated `resolveRealAccessToken`'s safe/real-token path on the mere presence of a stored impersonation token in the session, not the time-based `isImpersonating(req)` check, so a mid-request expiry can no longer cause it to return the stale impersonation token. Added a regression test covering exactly this race window (session present but expired, `bearerToken` still the imp-token).

See commit 5479a73f7: auth-helper.ts's non-impersonating fallback now checks `req.appSession?.['impersonationToken']` presence directly.

apps/lfx-one/src/server/utils/auth-helper.ts
manishdixitlfx: Confirmed and fixed — traced it exactly as you laid out: `req.bearerToken` set once by `extractBearerToken` at request start, `isImpersonating(req)` re-evaluated live, and `shareBrief` awaits several upstream calls before `resolveRealAccessToken` runs, widening the window. Applied your suggested gate (presence of the stored impersonation token, not the time predicate) — went with a slightly narrower check (`req.appSession?.['impersonationToken']` present) rather than also comparing it against the current `req.bearerToken` value, since the presence check alone is sufficient: whenever a stored impersonation token exists we always resolve via `req.oidc.accessToken` regardless, so there's no path where trusting a stale `req.bearerToken` could slip through either way.

Added the regression test you asked for (impersonating session whose `impersonationExpiresAt` has passed but whose `bearerToken` still holds the imp-token resolves to the real token, plus a fail-closed variant), and also fixed the sibling test-fidelity issue you flagged in the other thread.

See commit 5479a73f7.

apps/lfx-one/src/server/services/weekly-brief.service.spec.ts
manishdixitlfx: Good catch — fixed. The test now captures `req.bearerToken` synchronously inside `deleteNewsletterMock`'s implementation at call time (same pattern the create/send captures above it already use), and asserts on that captured value rather than on `req` after the finally-restore has already run. Also added an equivalent capture for `getCommitteeById` in the main impersonation test, per the nit in your review body about pre-write reads.

See commit 5479a73f7.

manishdixitlfx:
## Review Feedback Addressed

Commit: 5479a73f7f3f00a6dc334d135fceb96f491d525b

### Changes Made
- **`apps/lfx-one/src/server/utils/auth-helper.ts`**: Fixed a fail-open in `resolveRealAccessToken` — it gated its safe path on the live, time-based `isImpersonating(req)` check instead of the presence of a stored impersonation token, so a mid-request expiry (plausible here, since `shareBrief` awaits several upstream calls before this runs) could make it return the stale impersonation token instead of the real one, reintroducing the exact LFXV2-3093 misattribution in a narrow race window. Now gated on token presence, not time. (per @dealako, cursor)
- **`apps/lfx-one/src/server/utils/auth-helper.spec.ts`**: added a regression test for the race window, plus a fail-closed variant. (per @dealako)
- **`apps/lfx-one/src/server/services/weekly-brief.service.spec.ts`**: fixed a test-fidelity gap — the cleanup test asserted against `req` *after* the finally-restore had already run, so it never actually proved cleanup ran under the real token. Now captures the token synchronously at call time. Also added a similar capture for `getCommitteeById` to confirm that precondition read stays on the effective/target identity. (per @dealako)

### Threads Resolved
3 of 3 unresolved threads addressed in this iteration — all threads now resolved.

dealako (approved):
Thanks for the fast, precise turnaround, @manishdixitlfx 🙏 — you addressed every item head-on, and I re-traced `auth.middleware.ts` end-to-end to independently confirm the fixes rather than take them on faith.

## 👏 Nice work
- **The blocking fail-open is genuinely closed.** Re-gating `resolveRealAccessToken` on the *presence* of a stored `impersonationToken` (`hadImpersonationToken`) instead of the time-based `isImpersonating(req)` is exactly the right move. I verified against the middleware that the two states can't diverge in the dangerous direction: every `extractBearerToken` path that leaves `req.bearerToken` as the **real** token first calls `clearImpersonationSession` (deleting `impersonationToken`) — the malformed-JWT branch and the expired-at-middleware branch — so whenever the session still holds `impersonationToken`, `req.bearerToken` is the impersonation token and the helper correctly routes to `req.oidc.accessToken` / refresh / `null`. It can no longer return the stale token in the mid-request-expiry race, and there's no false fail-closed for legitimate real-token requests.
- **The new race-window unit tests are real regression guards** — the `impersonationSessionExpired` fixture (`expiresAt = Date.now() - 1000`, token still present) fails under the old time-gated code and passes only under the fix, asserting both the resolve-real-token and the fail-closed-to-`null`-never-`imp-token` branches.
- **The delete-time assertion nails the invariant I flagged** — capturing `tokenDuringDelete` synchronously inside the mock (with a comment correctly explaining that a bare `toHaveBeenCalledWith(req)` inspects the finally-restored object) proves the cleanup runs under `real-staff-token`.
- **The read/write identity split is now positively asserted** — `tokenDuringCommitteeFetch === 'imp-token'` alongside the write chain's `real-staff-token` locks in the contract.
- The new JSDoc paragraph documenting *why* the gating deliberately avoids `isImpersonating` is excellent — it'll stop a future refactor from silently reopening the hole.

## Revision status
- ✅ **Resolved** — [blocking] `resolveRealAccessToken` fail-open on mid-request impersonation expiry (`auth-helper.ts`, commit 5479a73)
- ✅ **Resolved** — [minor] cleanup test now asserts the token *at delete time* (`weekly-brief.service.spec.ts`)
- ✅ **Resolved** — [nit] pre-write read (`getCommitteeById`) now asserted on the effective/target token

## Issue count (all open items)
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

No new findings from the security, correctness, or style passes on the new commit. Cursor Bugbot's earlier Medium was the same fail-open and is now moot; no bot has re-flagged the latest commit. Branch is `MERGEABLE` (the `BLOCKED` merge state is just pending required review/checks, not a conflict).

## Decision
✅ **Approved** — the security-purpose fix no longer ships its own fail-open, and the coverage is tight. Nice work.

PR #1469, opened 2026-08-13
fix(ui): show create quick-link button on mobile

manishdixitlfx:
## Summary

- The rail "Create" quick-link button (opens a popover to create a meeting/newsletter/vote/survey/group/mailing-list) was scoped to the desktop-only branch of `lens-switcher.component.html`'s `@if (mobile()) {...} @else {...}` split, with an in-code comment explicitly deferring mobile support to a follow-up. It never appeared in the mobile drawer.
- Added the mobile trigger and hoisted the shared `<p-popover #createMenu>` outside the `@if/@else` split so both the desktop rail and mobile drawer instances of `lfx-lens-switcher` can open it.
- Review caught two real bugs in the first pass, both fixed:
  - The popover's SCSS was unconditionally pinned to the desktop rail's fixed coordinates (`position: fixed; left: 56px; top: 58px`) — opening it from the new mobile trigger would have rendered it in the wrong place. Scoped those rules to the `lg` breakpoint (1024px, matching `tailwind.config.js`), matching an existing pattern already used by `org-selector`/`project-selector` in this repo.
  - `lfx-lens-switcher` mounts twice simultaneously (desktop rail + mobile drawer instances in `main-layout.component.html`), so the hoisted popover was carrying a duplicate `data-testid` across both instances. Suffixed it `-mobile`, mirroring the existing convention on the two trigger buttons.
- Added mobile-viewport e2e coverage (`M1` in `create-quick-link.spec.ts`, gated to the `mobile-chrome` Playwright project). The regression-guard assertion itself went through a few rounds of tightening during review — an initial bounding-box heuristic turned out to be a weak/no-op guard; it now asserts the popover panel's computed `position` directly (`absolute` on mobile vs. the regressed `fixed`), verified live against both a real desktop-width session and a genuine narrow viewport.

## Test plan

- [x] `yarn format && yarn lint && yarn check-types && yarn build` all pass
- [x] `./check-headers.sh` passes
- [x] Manually verified in browser: mobile drawer shows the Create button, opens the artifact-type popover, correctly anchored (not pinned at desktop rail coordinates) on a genuine narrow viewport
- [x] `npx playwright test e2e/create-quick-link.spec.ts --project=mobile-chrome --list` confirms the new `M1` test registers correctly
- [ ] `M1` itself needs `TEST_USERNAME`/`TEST_PASSWORD` to execute (not available in this environment) — should run in CI

[LFXV2-3248](https://linuxfoundation.atlassian.net/browse/LFXV2-3248)

[LFXV2-3248]: https://linuxfoundation.atlassian.net/browse/LFXV2-3248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/e2e/create-quick-link.spec.ts
+++ b/apps/lfx-one/e2e/create-quick-link.spec.ts
@@ -278,7 +278,9 @@ test.describe('Create Quick-Link — search-fallback for an inherited-writer (LF
 // popover (data-testid="create-menu-mobile") — see LFXV2-3248. The regression this suite guards
 // (lens-switcher.component.scss's rail-anchoring CSS) is gated at the `lg` breakpoint (1024px,
 // matching tailwind.config.js), so the skip boundary matches that rather than `md` (768px) —
-// a viewport in [768, 1024) would still get the desktop-anchored popover CSS applied.
+// a viewport in [768, 1024) still renders the mobile drawer (the app's own responsive split is
+// keyed on `lg`, not `md`), so gating the skip at `md` would incorrectly treat that range as
+// desktop and skip M1 even though the regression this test guards still applies there.
 function skipOnDesktopViewport(page: Page): void {
   const viewport = page.viewportSize();
   if (viewport && viewport.width >= 1024) {
@@ -325,7 +327,15 @@ test.describe('Create Quick-Link — mobile trigger (LFXV2-3248)', () => {
     // button.component.ts does for `aria-pressed`), but adding that is a production-template
     // change out of scope for this test-only commit. The class this test already relies on to
     // scope the SCSS override is the stable hook available today.
-    const panel = page.locator('.p-popover.lfx-create-popover');
+    //
+    // `.filter({ visible: true })`: `lfx-lens-switcher` mounts twice at once (desktop rail +
+    // mobile drawer instances in main-layout.component.html), each with its own `.p-popover
+    // .lfx-create-popover` panel. This only resolves to one element today because PrimeNG lazily
+    // creates the panel DOM on first `show()`/`toggle()` — the never-toggled desktop instance's
+    // panel isn't in the DOM during this test. The visibility filter hardens that assumption so a
+    // future PrimeNG version with eager panel creation fails on the real regression instead of a
+    // Playwright strict-mode "resolved to 2 elements" error.
+    const panel = page.locator('.p-popover.lfx-create-popover').filter({ visible: true });
     const menuPosition = await panel.evaluate((el) => getComputedStyle(el).position);
     // PrimeNG's own anchor-relative default sets `position: absolute`; the regressed (rail-pinned)
     // state is `position: fixed`. Assert the specific expected value, not just an exclusion, so an
--- a/apps/lfx-one/e2e/create-quick-link.spec.ts
+++ b/apps/lfx-one/e2e/create-quick-link.spec.ts
@@ -265,3 +265,46 @@ test.describe('Create Quick-Link — search-fallback for an inherited-writer (LF
     await expect(continueButton(page)).toBeEnabled();
   });
 });
+
+// The mobile drawer's own Create trigger is a separate element from the desktop rail button
+// (data-testid="create-rail-button-mobile"), opening a distinctly-testid'd copy of the same
+// popover (data-testid="create-menu-mobile") — see LFXV2-3248. Desktop projects (chromium,
+// firefox) never render the mobile drawer content behind a viewport this wide, so this suite
+// only runs on the narrow (mobile-chrome) project.
+function skipOnDesktopViewport(page: Page): void {
+  const viewport = page.viewportSize();
+  if (viewport && viewport.width >= 768) {
+    test.skip(true, 'Mobile-only trigger — desktop uses create-rail-button in the vertical rail');
+  }
+}
+
+test.describe('Create Quick-Link — mobile trigger (LFXV2-3248)', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(APP_HOME, { waitUntil: 'domcontentloaded' });
+    skipWhenAuthMissing(page);
+    skipOnDesktopViewport(page);
+  });
+
+  // M1 — the mobile drawer's own Create button renders and opens its own popover instance
+  test('M1: the mobile "Create" button opens the artifact-type popover from the drawer', async ({ page }) => {
+    await page.getByTestId('mobile-menu-button').click();
+
+    const trigger = page.getByTestId('create-rail-button-mobile');
+    await expect(trigger).toBeVisible({ timeout: RAIL_TIMEOUT });
+    await trigger.click();
+
+    const menu = page.getByTestId('create-menu-mobile');
+    await expect(menu).toBeVisible({ timeout: 5_000 });
+    await expect(page.getByTestId('create-menu-mobile-option-meeting')).toBeVisible();
+
+    // Popover must actually open near the tapped button, not pinned at the desktop rail's fixed
+    // coordinates (left:56px/top:58px) — regression guard for the rail-anchoring CSS scoped to
+    // the `lg` breakpoint.
+    const [menuBox, triggerBox] = await Promise.all([menu.boundingBox(), trigger.boundingBox()]);
+    expect(menuBox).not.toBeNull();
+    expect(triggerBox).not.toBeNull();
+    if (menuBox && triggerBox) {
+      expect(Math.abs(menuBox.y - triggerBox.y)).toBeLessThan(150);
+    }
+  });
+});
--- a/apps/lfx-one/e2e/create-quick-link.spec.ts
+++ b/apps/lfx-one/e2e/create-quick-link.spec.ts
@@ -319,6 +319,11 @@ test.describe('Create Quick-Link — mobile trigger (LFXV2-3248)', () => {
     // default `static` regardless of the outer panel's state, making the assertion a no-op. The
     // rail-anchoring rule (and PrimeNG's own base stylesheet) target the outer panel element,
     // identified here by its unique `lfx-create-popover` class.
+    //
+    // Deliberate exception to data-testid-first selection: PrimeNG's `p-popover` panel is
+    // CDK-portaled to `body` and exposes only `styleClass` as an extension point (no testid
+    // passthrough), so the class this test already relies on to scope the SCSS override is also
+    // the only stable hook available here.
     const panel = page.locator('.p-popover.lfx-create-popover');
     const menuPosition = await panel.evaluate((el) => getComputedStyle(el).position);
     // PrimeNG's own anchor-relative default sets `position: absolute`; the regressed (rail-pinned)
--- a/apps/lfx-one/e2e/create-quick-link.spec.ts
+++ b/apps/lfx-one/e2e/create-quick-link.spec.ts
@@ -305,17 +305,17 @@ test.describe('Create Quick-Link — mobile trigger (LFXV2-3248)', () => {
     await expect(menu).toBeVisible({ timeout: 5_000 });
     await expect(page.getByTestId('create-menu-mobile-option-meeting')).toBeVisible();
 
-    // Popover must actually open near the tapped button, not pinned at the desktop rail's fixed
-    // coordinates (left:56px/top:58px) — regression guard for the rail-anchoring CSS scoped to
-    // the `lg` breakpoint. Check X, not Y: the mobile trigger sits in a right-aligned (`ml-auto`)
-    // icon group, far from the rail's x=56, while the rail's fixed top:58px happens to land close
-    // to the mobile trigger's real Y position too — so a Y-only check would likely still pass even
-    // if the `@media (min-width: 1024px)` scoping regressed. X actually distinguishes the two.
-    const [menuBox, triggerBox] = await Promise.all([menu.boundingBox(), trigger.boundingBox()]);
-    expect(menuBox).not.toBeNull();
-    expect(triggerBox).not.toBeNull();
-    if (menuBox && triggerBox) {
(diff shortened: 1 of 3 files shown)


apps/lfx-one/e2e/create-quick-link.spec.ts
```diff
@@ -265,3 +272,64 @@ test.describe('Create Quick-Link — search-fallback for an inherited-writer (LF
...
+    // default `static` regardless of the outer panel's state, making the assertion a no-op. The
+    // rail-anchoring rule (and PrimeNG's own base stylesheet) target the outer panel element,
+    // identified here by its unique `lfx-create-popover` class.
+    //
+    // Deliberate exception to data-testid-first selection: `<p-popover>` doesn't currently wire a
+    // data-testid onto its panel root — it could via PrimeNG's `pt.root` passthrough (as
+    // button.component.ts does for `aria-pressed`), but adding that is a production-template
+    // change out of scope for this test-only commit. The class this test already relies on to
+    // scope the SCSS override is the stable hook available today.
+    const panel = page.locator('.p-popover.lfx-create-popover');
```
audigregorie: **Minor — test selector could match multiple popover panels**

`page.locator('.p-popover.lfx-create-popover')` matches a class that exists on **both** `lfx-lens-switcher` instances' popover panels — the desktop rail and mobile drawer instances are both mounted in `main-layout.component.html` (the `hidden lg:flex` / `lg:hidden` split is CSS, not structural `@if`), and both use `appendTo="body"` with `styleClass` containing `lfx-create-popover`.

The assertion only works today because PrimeNG lazily creates the panel element on first `show()`/`toggle()`, so the never-toggled desktop instance's panel isn't in the DOM during M1. If PrimeNG ever switches to eager panel creation, `panel.evaluate(...)` throws a strict-mode violation ("locator resolved to 2 elements") instead of failing on the actual regression — a silent test breakage rather than a meaningful one.

Scoping to the visible panel hardens the assertion with zero cost:

```suggestion
    const panel = page.locator('.p-popover.lfx-create-popover').filter({ visible: true });
```

apps/lfx-one/src/app/shared/components/lens-switcher/lens-switcher.component.html
```diff
@@ -20,6 +20,17 @@
...
+      <!-- Create quick-link — opens the same artifact-type popover as the desktop rail. -->
+      @if (createPermissionService.canShowCreateButton()) {
+        <button
+          type="button"
+          (click)="toggleCreateMenu($event)"
+          data-testid="create-rail-button-mobile"
+          aria-label="Create new artifact"
+          class="flex items-center justify-center w-8 h-8 rounded-full text-gray-500 hover:bg-gray-100 transition-colors">
+          <i class="fa-light fa-plus text-sm"></i>
+        </button>
```
audigregorie, line 34: **Minor — a11y: button opens a menu but doesn't announce it**

The new mobile Create button is icon-only and toggles a popover menu, but exposes no `aria-haspopup` / `aria-expanded` state. Screen-reader users get the `aria-label` (good) but no indication the button opens a menu or whether it's currently open.

The existing desktop Create button has the same gap — flagging here per the repo's "match `.claude/rules/`, not the neighboring legacy code" guidance. The static hint is a one-line add; `aria-expanded` needs the popover's visible state wired to a signal so it's reasonable to defer.

```suggestion
        <button
          type="button"
          (click)="toggleCreateMenu($event)"
          data-testid="create-rail-button-mobile"
          aria-label="Create new artifact"
          aria-haspopup="menu"
          class="flex items-center justify-center w-8 h-8 rounded-full text-gray-500 hover:bg-gray-100 transition-colors">
          <i class="fa-light fa-plus text-sm"></i>
        </button>
```

audigregorie (commented):
## Code Review Summary

Clean, well-documented fix. The SCSS breakpoint scoping (`@media (min-width: 1024px)` matching `tailwind.config.js`'s `lg`), the hoisted shared popover, and the per-instance `data-testid` suffixing are all the right shape. The M1 regression guard asserts the right thing (computed `position` off the outer `.p-popover` panel) and the comment trail on the rejected heuristics is valuable. No Critical or Major findings — two Minor notes below.

### What's done well

- SCSS rail-anchoring overrides scoped to the `lg` breakpoint so PrimeNG's anchor-relative default positions the mobile popover correctly under the tapped trigger, while desktop behavior is preserved (including the banner-offset variant).
- Hoisting the shared `<p-popover #createMenu>` outside the `@if/@else` split (one popover per instance, driven by both triggers) rather than duplicating it — and disambiguating the two simultaneously-mounted instances via the `-mobile` testid suffix, mirroring the existing trigger-button convention.
- M1 reads `getComputedStyle(...).position` off the actual `.p-popover` panel (not the projected content div) and asserts the specific expected value (`absolute`), so it genuinely catches the regression rather than being a no-op.

apps/lfx-one/src/app/shared/components/lens-switcher/lens-switcher.component.html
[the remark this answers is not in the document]
manishdixitlfx: Done — added `[attr.aria-expanded]="createMenu.overlayVisible"` to the mobile create trigger, matching the repo's established convention for PrimeNG popover triggers (e.g. `sidebar.component.html`, `org-selector.component.html`, `project-selector.component.html`).

See commit aee3d3a34.

apps/lfx-one/e2e/create-quick-link.spec.ts
```diff
@@ -265,3 +272,64 @@ test.describe('Create Quick-Link — search-fallback for an inherited-writer (LF
     await expect(continueButton(page)).toBeEnabled();
   });
 });
+
+// The mobile drawer's own Create trigger is a separate element from the desktop rail button
+// (data-testid="create-rail-button-mobile"), opening a distinctly-testid'd copy of the same
+// popover (data-testid="create-menu-mobile") — see LFXV2-3248. The regression this suite guards
+// (lens-switcher.component.scss's rail-anchoring CSS) is gated at the `lg` breakpoint (1024px,
+// matching tailwind.config.js), so the skip boundary matches that rather than `md` (768px) —
+// a viewport in [768, 1024) would still get the desktop-anchored popover CSS applied.
```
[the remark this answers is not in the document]
manishdixitlfx: Good catch, you're right — the comment had the rationale backwards. Corrected: a [768, 1024) viewport doesn't get the desktop-anchored CSS (that's gated at min-width:1024px), it still renders the mobile drawer, which is exactly why gating the skip at `md` (768) instead of `lg` (1024) would have wrongly excluded that range from mobile coverage.

See commit aee3d3a34.

apps/lfx-one/e2e/create-quick-link.spec.ts
manishdixitlfx: Done — added `.filter({ visible: true })` exactly as suggested, plus a comment explaining why (two `lfx-lens-switcher` instances, only lazy panel creation keeps this working today).

See commit aee3d3a34.

apps/lfx-one/src/app/shared/components/lens-switcher/lens-switcher.component.html
manishdixitlfx: Done — added `aria-haspopup="menu"` as suggested. Went ahead and wired `aria-expanded` too rather than deferring it: `#createMenu` is already a template reference variable in scope on this template, so `[attr.aria-expanded]="createMenu.overlayVisible"` works with no extra signal wiring — it's the same pattern already used by `sidebar.component.html`, `org-selector.component.html`, and `project-selector.component.html` for their PrimeNG popover triggers.

See commit aee3d3a34.

manishdixitlfx:
## Review Feedback Addressed

Commit: aee3d3a34b5203e5f42a8c2abb0e7506dbe72cd0

### Changes Made
- **lens-switcher.component.html**: added `aria-haspopup="menu"` and `[attr.aria-expanded]="createMenu.overlayVisible"` to the mobile create trigger, matching the repo's established PrimeNG popover convention (per Cursor Bugbot, per @audigregorie)
- **e2e/create-quick-link.spec.ts**: corrected the `skipOnDesktopViewport` comment — it had the `[768, 1024)` breakpoint rationale backwards (per Copilot)
- **e2e/create-quick-link.spec.ts**: scoped the popover panel locator with `.filter({ visible: true })` to guard against a future PrimeNG behavior change surfacing a strict-mode error instead of the real regression (per @audigregorie)

### Threads Resolved

apps/lfx-one/src/app/shared/components/lens-switcher/lens-switcher.component.html
manishdixitlfx: Also fixed the desktop rail Create button (same aria-haspopup/aria-expanded gap you flagged) — see cd50ebc2a.

apps/lfx-one/src/app/shared/components/lens-switcher/lens-switcher.component.html
```diff
@@ -20,6 +20,19 @@
...
     <!-- Mobile account actions -->
     <div class="ml-auto flex items-center gap-2">
+      <!-- Create quick-link — opens the same artifact-type popover as the desktop rail. -->
+      @if (createPermissionService.canShowCreateButton()) {
+        <button
+          type="button"
+          (click)="toggleCreateMenu($event)"
+          data-testid="create-rail-button-mobile"
+          aria-label="Create new artifact"
+          aria-haspopup="menu"
```
dealako, line 30: [minor] `aria-haspopup="menu"` misdescribes the popover — the newest commit made it consistent, but consistently inaccurate

Issue: `#createMenu` is a `<p-popover>` whose items are plain `<button>` elements — there is no `role="menu"` / `role="menuitem"` in the panel (only a `role="separator"`). `aria-haspopup="menu"` promises assistive tech an ARIA menu widget (arrow-key navigation) that doesn't exist. Commit `cd50ebc` added the same attribute to the desktop trigger (L94), so both triggers now match each other — but both now carry the inaccurate contract rather than the repo's actual convention.

Proof: The popover content (`<p-popover #createMenu>`) renders `<button (click)="openCreateDialog(artifact)">` with no menu roles. The established disclosure-popover convention uses `aria-expanded` only with **no** `aria-haspopup`: `org-selector.component.html:9`, `project-selector.component.html:9`, `sidebar.component.html`. The only real `aria-haspopup="menu"` precedent (`badge-card.component.html`) is paired with an actual `role=menu` widget. This is also Cursor Bugbot's still-open "Incorrect menu popup ARIA" finding.

Why it matters: Screen-reader users are told a menu widget will appear and get plain buttons instead. The `[attr.aria-expanded]` binding alone is the correct, complete semantics here (matching org-selector/project-selector).

Fix: Remove `aria-haspopup="menu"` from both triggers (L30 mobile and L94 desktop) and keep `[attr.aria-expanded]="createMenu.overlayVisible"`. If a popup affordance is genuinely desired, `aria-haspopup="dialog"` more accurately describes the artifact-picker → dialog flow, but `aria-expanded`-only is the repo default.

dealako (approved):
Hi @manishdixitlfx 👋 — re-reviewed at `cd50ebc`.

## Overall impression

Tight, well-scoped UI fix: the mobile Create trigger is added, the shared `<p-popover #createMenu>` is correctly hoisted out of the mobile/desktop `@if/@else` split (per-instance `viewChild`, so no shared-state bug), the rail-anchoring SCSS is cleanly scoped to the `lg` (1024px) breakpoint matching `tailwind.config.js`, and the `-mobile` testid suffix keeps both mounted instances unique. The M1 e2e guard evolved into a genuinely meaningful assertion (computed `position: absolute` on the real `.p-popover` panel, `.filter({ visible: true })` hardening, 1024px skip boundary). The commit history shows disciplined review-driven iteration. Security/data-privacy: clean — the trigger reuses the identical `canShowCreateButton()` gate and real authz stays downstream at the picker's fail-closed empty state and each route's `writerGuard`.

## Findings

- 🟡 **Minor (1):** `aria-haspopup="menu"` on both Create triggers (L30 mobile, L94 desktop) misdescribes the popover — its items are plain `<button>`s with no `role="menu"`/`role="menuitem"`. `cd50ebc` made the two triggers *consistent* but consistently inaccurate. Repo convention (`org-selector`, `project-selector`, `sidebar`) is `aria-expanded`-only with no `aria-haspopup`. Agrees with Cursor Bugbot's still-open "Incorrect menu popup ARIA" finding. Inline comment left. Non-blocking.
- ⚪ **Nit (1):** M1 stops at popover-visible and never clicks through to the create dialog, so the mobile `openCreateDialog` wiring is only implicitly covered (shares code with desktop S3–S6, so low risk). Optional M2 could close the seam.
- ❔ **Question (1):** _Pre-existing, not introduced here_ — S1–S8/S9 have no mobile skip yet Playwright's `mobile-chrome` project (Pixel 5, 393px) runs every spec; their `beforeEach` waits on `create-rail-button`, which lives in the `hidden lg:flex` desktop rail. This PR correctly self-skips M1 on desktop; worth confirming the older suites are scoped/skipped for mobile in CI, but it's out of scope for this change.

## AI bot reconciliation

- **Cursor Bugbot** "Incorrect menu popup ARIA" (open, latest commit): **agree** — same finding as my minor above.
- **Cursor Bugbot** "Missing aria-expanded", **Copilot** breakpoint-comment, **@audigregorie** selector/a11y notes: all **resolved** across `aee3d3a3` and `cd50ebc`.

## Decision

✅ **Approved with comments** — no blocking issues, no security findings, mergeable. The one open `aria-haspopup` item is a minor a11y-accuracy nit; recommend dropping the attribute before merge, but it doesn't block.

audigregorie (approved):

PR #1471, opened 2026-08-13
feat(routing): render 404 in place for non-public routes

fayazg:
## Summary

- Unmatched non-public routes now render the branded 404 **in place** — the address bar keeps the originally-requested path (no `/not-found` redirect).
- The 404 component flags a mutable `notFound` on the shared `REQUEST_CONTEXT` during SSR; `server.ts` rebuilds the response as a real **HTTP 404** at that path.
- For logged-in users the 404 renders **inside the app shell** (`MainLayoutComponent`, left nav visible) via a trailing empty-path catch-all placed last so it never shadows the public sibling routes.
- Anonymous users hitting an unknown deep-link go through login first, then land on the in-shell 404 (unknown paths default to required-auth after removing the `/not-found` whitelist entry).
- Fixes the public-profile **soft-404** (`/u/:username` was returning HTTP 200) — it now sets the same flag and emits a real 404.
- In-shell 404 shows **Go to Dashboard** + **← Go Back**; the public profile 404 keeps **Contact Support** (Intercom).
- Adds a shared `ServerRequestContext` interface to type the `REQUEST_CONTEXT` casts (repo rule: interfaces live in `@lfx-one/shared`).

## Reviewer notes

- Touches two protected files by design: `server.ts` (the 200→404 response rebuild) and `auth.middleware.ts` (drops the now-dead `/not-found` public whitelist entry) — flagging for code-owner review.
- **AC #4 scope:** this PR anchors the `/meetings/:id` (excluding the reserved `create` segment) and `/groups/:id` public matchers so anonymous deep-links gate to login before the in-shell 404. The remaining permissive SSR matchers (`/foundations/:x/groups/…`, `/projects/:x/groups/…` deep tails, and the `/invite/error` bare-string prefix) still render the in-shell 404 anonymously rather than gating to login — a bounded, low-severity trade-off (correct HTTP 404, no token/data leak, client redirects to login on hydration). The comprehensive per-pattern anchoring is tracked as a follow-up in **LFXV2-3254** (split from LFXV2-3095).

Fixes LFXV2-3095

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of nonexistent and private public profiles with proper branded not-found views.
  - Server-rendered missing pages now return HTTP 404 responses without redirecting to login or changing the requested URL.
  - Prevented nested meeting and group routes from being incorrectly treated as publicly accessible.
  - Not-found pages now render consistently within the application layout.


apps/lfx-one/src/app/app.routes.ts
```diff
@@ -506,24 +506,17 @@ export const routes: Routes = [
...
-  // Generic catch-all — redirects unrecognized URLs to /not-found so authenticated users get a
-  // proper HTTP 404 status and a branded page. Also handles post-login returnTo redirects for
-  // anonymous users who followed a stale or mistyped URL (see /not-found comment above).
-  // Must be last so it never shadows real routes.
-  {
-    path: '**',
-    redirectTo: '/not-found',
+    path: '',
+    canActivate: [authGuard],
+    loadComponent: () => import('./layouts/main-layout/main-layout.component').then((m) => m.MainLayoutComponent),
```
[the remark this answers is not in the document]
fayazg: Fixed in a8fb7d1a3. Anchored the `/meetings/` and `/groups/` middleware patterns to a single segment (`/^\/meetings\/[^/]+\/?$/`, `/^\/groups\/[^/]+\/?$/`), matching the existing `/u/`, `/docs`, `/foundations/../groups` anchored-regex convention. Anonymous `/meetings/:id/extra` and `/groups/:id/edit` deep-links now fall through to the default `required` auth and reach login before the in-shell catch-all 404. Added regression cases in `auth.middleware.spec.ts` for the nested anonymous meeting/group paths.

fayazg:
**a8fb7d1a3** `fix(review): anchor public meeting/group prefixes to gate deep-links`

Addresses Copilot's review finding: anonymous `/meetings/:id/extra` and `/groups/:id/edit` deep-links matched the bare-string `/meetings/` and `/groups/` optional-auth prefixes and reached the in-shell catch-all 404 during SSR instead of routing through login.

- Anchored both middleware patterns to a single segment (`/^\/meetings\/[^/]+\/?$/`, `/^\/groups\/[^/]+\/?$/`), matching the existing `/u/`, `/docs`, `/foundations/../groups` anchored-regex convention. Public `/meetings/:id` + `/groups/:id` stay anonymous; deeper paths fall through to the default `required` auth → login → in-shell 404 (completes AC #4).
- Added `auth.middleware.spec.ts` regression cases for the nested anonymous meeting/group paths.

fayazg:
@coderabbitai review

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -26,11 +26,14 @@ const DEFAULT_ROUTE_CONFIG: RouteAuthConfig[] = [
   // Public API routes - optional authentication with token benefits
   { pattern: '/public/api', type: 'api', auth: 'optional', tokenRequired: false },
 
-  // Public meeting join - no authentication required
-  { pattern: '/meetings/', type: 'ssr', auth: 'optional' },
+  // Public meeting join (`/meetings/:id`, `/meetings/not-found`) — anchored to a single segment so
+  // deeper paths (authenticated `/meetings/create|:id/edit` and unknown `/meetings/:id/extra`) fall
+  // through to the default `required` auth and reach login before the in-shell catch-all 404.
+  { pattern: /^\/meetings\/[^/]+\/?$/, type: 'ssr', auth: 'optional' },
```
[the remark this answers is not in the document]
fayazg: Good catch on the comment accuracy — corrected in c4cbb5520. You're right that `create` is a single segment and still matches the optional rule; only multi-segment paths fall through to required auth.

On severity: this is low-impact and I've kept the fix scoped. Anonymous `/meetings/create` doesn't render the authenticated shell — the authenticated meeting subtree is mounted under `authenticatedMatchGuard` (CanMatch reads auth from REQUEST_CONTEXT during SSR), so for an anonymous request Angular skips it and backtracks to the public `meetings/:id` route, resolving `create` as a bogus meeting id → its own not-found. `groups` has no `create` route at all.

For the broader fail-open class you flagged (the `/foundations/:slug/groups` + `/projects/:slug/groups` deep tails and the `/invite/error` string prefix): those are pre-existing patterns outside this PR's scope, and the exposure is bounded — `MainLayoutComponent` SSR-renders safely for an anonymous request (its constructor does only synchronous URL/lens/sidebar setup, no auth-required fetch, so no 500), returns a correct HTTP 404, and the client `authGuard` redirects to login on hydration. No token is ever sent, so there's no data-leak surface. I'm tracking the comprehensive per-pattern anchoring as a scoped follow-up rather than expanding this PR's footprint on the protected middleware.

fayazg:
**c4cbb5520** `docs(auth): correct meeting prefix comment on reserved create segment`

Addresses Copilot's re-review note that the prior comment overclaimed `/meetings/create` was re-gated. Comment-only clarification (no behavior change): only multi-segment paths fall through to `required` auth; the reserved single-segment `create` still matches the optional rule and resolves anonymously as a public `:id` (bogus meeting → not-found) via `authenticatedMatchGuard`, never rendering the authenticated shell.

The broader fail-open class Copilot flagged (foundations/projects group deep tails, `/invite/error` prefix) is pre-existing and out of scope here — bounded (MainLayout SSR-renders safely for anonymous, correct HTTP 404, client redirects to login on hydration, no token sent). Tracking comprehensive per-pattern anchoring as a scoped follow-up.

apps/lfx-one/src/server/middleware/auth.middleware.ts
dealako: [minor] Comment overstates the auth boundary for reserved single-segment children

Issue: The comment says deeper authenticated paths like `/meetings/create` "fall through to the default `required` auth," but `create` is a **single** segment, so `/meetings/create` and `/groups/create` still match `/^\/meetings\/[^/]+\/?$/` and classify as `optional` — same as under the old `startsWith` rule. Only two-plus-segment paths (`/meetings/:id/edit`) actually fall through.

Proof: `meetings.routes.ts:16-19` gates `create` with `[authGuard, writerGuard]` only in Angular; `auth.guard.ts` returns `true` on the server (trusts the middleware), so the SSR boundary for `/meetings/create` is this `optional` row, not `required`. This is the open Copilot thread (comment id 3778131215), not yet resolved.

Why it matters: Not exploitable today (`writerGuard` redirects anon before the create form renders), but in an auth file an inaccurate rationale is a trap — a future single-segment sensitive route (e.g. `/meetings/settings`) added on the belief that these children are `required` would be served anonymously as optional-auth SSR.

Fix: Either correct the comment to state that single-segment paths (including `create`) remain `optional` and their protection depends on the Angular `writerGuard`/`authenticatedMatchGuard`, or exclude the reserved segment, e.g. `/^\/meetings\/(?!create$)[^/]+\/?$/`, and add a classification regression case.

apps/lfx-one/src/server/server.ts
```diff
@@ -479,21 +479,34 @@ app.use('/**', async (req: Request, res: Response, next: NextFunction) => {
     impersonating: !!auth.impersonating,
   });
 
+  // Mutable per-request context — a not-found view sets `notFound` during SSR (via REQUEST_CONTEXT,
+  // the same object reference) so we emit a real 404 at the originally-requested path (no redirect).
+  const renderContext = {
+    auth,
+    runtimeConfig,
+    notFound: false,
```
dealako, line 487: [minor] `renderContext` isn't bound to the new `ServerRequestContext` contract

Issue: This object is the sole producer of the shape the new `ServerRequestContext` interface documents, but it's an untyped literal. Nothing ties the producer to the interface, and the interface omits the `providers` field the object always carries.

Proof: `server-request-context.interface.ts` JSDoc claims the shape is "threaded from `server.ts`," yet a rename like `notFound` → `isNotFound` on the interface wouldn't break this literal — consumers (`not-found.component.ts`, `public-profile-page.component.ts`) reading `this.reqContext.notFound` would silently read `undefined` and the 404 rewrite would quietly stop, with no type error or test failure.

Why it matters: The interface is presented as the contract but isn't compiler-enforced against its only producer.

Fix: Type the producer, e.g. `const renderContext: ServerRequestContext & { providers: StaticProvider[] } = { ... }` (or add `providers` to the interface) so field drift fails the build.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -506,24 +506,17 @@ export const routes: Routes = [
...
+    children: [
+      {
+        path: '**',
```
dealako, line 517: [minor] Primary in-shell catch-all 404 has no end-to-end coverage

Issue: The novel, highest-risk behavior of this PR — an **authenticated** user hitting an unknown URL, rendering `NotFoundComponent` in-shell with the URL preserved and a real HTTP 404 — is untested end-to-end. The new Playwright spec only covers the anonymous `/u/:username` branch; `not-found.component.spec.ts` asserts only the `REQUEST_CONTEXT` flag mutation, not the routing or the `server.ts` 200→404 rewrite for this path.

Proof: `public-profile-not-found.spec.ts` uses `NONEXISTENT = '/u/...'` with anonymous storageState; no test asserts URL-unchanged + 404 + app shell for an authenticated deep unknown path. Copilot flagged the same gap (suppressed comment).

Why it matters: A regression in the empty-path-parent → `**` ordering, the `status === 200` guard, or the flag plumbing would ship undetected — current tests would still pass.

Fix: Add an authenticated e2e that GETs an unknown in-app URL and asserts `response.status() === 404`, no redirect (URL unchanged), and both the left-nav shell and `data-testid="not-found-card"` visible.

---

[nit] Two distinct top-level `path: ''` → `MainLayout` nodes mean an unknown URL falls through the primary shell (no `**` child) into this second `MainLayout` parent, so the router re-instantiates the layout on client-side navigation rather than swapping content in place (SSR unaffected, since each request is a fresh render). If the in-place SPA experience matters, host `**` as the last child of the existing shell node instead of a second `path: ''` parent.

---

[question] Copilot's suppressed `:514` note: other permissive SSR matchers still render this in-shell 404 anonymously instead of gating to login — `/foundations/:x/groups/extra` and `/projects/:x/groups/extra` (via the `.../groups(?:\/.*)?$` regexes) and `/invite/error-extra` (via the bare-string `startsWith` on `/invite/error`). Is AC #4 intentionally scoped to only `/meetings` & `/groups`, leaving these as accepted trade-offs? If so, a one-line note in the PR description would close the loop.

dealako (changes requested):
Hi @fayazg 👋 — really clean piece of work. Rendering the branded 404 in place (URL preserved) via a mutable `REQUEST_CONTEXT.notFound` flag that `server.ts` rebuilds into a real HTTP 404 is an elegant solve, the soft-404 fix on `/u/:username` is a genuine correctness win, and the anchored-regex hardening of the `/meetings/` and `/groups/` prefixes is exactly the right instinct. The new `ServerRequestContext` interface, the SSR-only guard on the flag mutation, and the middleware regression tests all show care.

Everything below is polish or coverage — nothing is a correctness or security regression. The one item I'd most like closed is the still-open Copilot thread on the reserved-segment `create` classification (finding 1), because it lives in an auth file where an inaccurate rationale is a future trap.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 3 — (1) `auth.middleware` comment overstates the boundary: `/meetings/create` & `/groups/create` are single-segment, so they still classify `optional`, not `required` (open Copilot thread); (2) no e2e for the primary path — an authenticated unknown URL asserting original URL + HTTP 404 + app shell; (3) `server.ts` `renderContext` isn't typed against the new `ServerRequestContext`, so field drift wouldn't be caught at compile time.
- ⚪ Nit: 2 — duplicate top-level `path: ''` → `MainLayout` nodes re-instantiate the shell on client-side nav to a 404 (SSR unaffected); two divergent cast patterns now coexist for `inject(REQUEST_CONTEXT)` (follow-up).
- ❔ Question: 1 — other permissive SSR matchers (`/foundations/:x/groups/extra`, `/projects/:x/groups/extra`, `/invite/error-extra` via `startsWith`) still render the in-shell 404 anonymously rather than gating to login. Is AC #4 intentionally scoped to only `/meetings` & `/groups`?

### Bot reconciliation
- **Agree** with Copilot's open thread on `auth.middleware.ts:32` (reserved `create` segment) — see finding 1 inline.
- **Agree** with Copilot's suppressed note on `app.routes.ts` (no authenticated-catch-all e2e) — see finding 2 inline.
- **Surfacing** Copilot's suppressed `:514` observation about other permissive matchers as an open question rather than a blocker, since it appears intentionally out of AC #4 scope.
- CodeRabbit generated no actionable comments; no conflict.

### Decision
🔴 **Needs changes before approval** — all findings are minor/nit, but the open Copilot reserved-segment thread plus the missing primary-path e2e are worth closing (or explicitly documenting as trade-offs) before merge.

fayazg:
Follow-up filed: **LFXV2-3254** — *Anchor remaining permissive SSR auth patterns to exact route shapes* (split from LFXV2-3095). Covers the broader, low-severity fail-open class Copilot flagged (foundations/projects group deep tails, `/invite/error` prefix, reserved `/meetings/create`), kept out of this PR to avoid over-tightening legitimate public routes on the protected middleware.

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -26,11 +26,17 @@ const DEFAULT_ROUTE_CONFIG: RouteAuthConfig[] = [
...
-  // Public meeting join - no authentication required
-  { pattern: '/meetings/', type: 'ssr', auth: 'optional' },
-
-  // Public group detail - anonymous access with optional auth for membership enrichment
-  { pattern: '/groups/', type: 'ssr', auth: 'optional' },
+  // Public meeting join (`/meetings/:id`, `/meetings/not-found`) — anchored to a single segment so
+  // multi-segment paths (authenticated `/meetings/:id/edit`, unknown `/meetings/:id/extra`) fall
+  // through to the default `required` auth and reach login before the in-shell catch-all 404. The
+  // reserved single-segment `create` still matches here; anonymously it resolves as a public `:id`
+  // (a bogus meeting → its own not-found), so it never renders the authenticated shell.
```
audigregorie: **Major — comment still describes the wrong Angular path for `/meetings/create`**

`c4cbb5520` corrected the earlier “falls through to required auth” claim, but the new sentence is still false.

Anonymous `/meetings/create` does **not** resolve as public `meetings/:id` with `id=create`. The in-app `meetings` child has no `canMatch` (unlike `groups`, which uses `authenticatedMatchGuard`), so SSR matching is:

1. First `path: ''` + `MainLayout` — `authGuard` returns `true` on the server.
2. Child `path: 'meetings'` + `MEETING_ROUTES`.
3. `MEETING_ROUTES` has an explicit `path: 'create'` (`authGuard` + `writerGuard`) that wins before any sibling `:id`.

`writerGuard` then redirects to a lens overview when there is no `?project=` (or denies after the project fetch). That is a different page, a different status, and it *does* enter the authenticated shell — the opposite of “a bogus meeting → its own not-found.”

`/groups/create` *does* fall through to public `:id`, because `authenticatedMatchGuard` skips the in-app `groups` child for anonymous SSR. The two reserved segments are not symmetric.

Minimum fix: rewrite the comment to match the `writerGuard` path. Better: exclude the reserved segment so `/meetings/create` is `required` and anonymous visitors hit login (AC #4), and add a classification spec.

```suggestion
  // Public meeting join (`/meetings/:id`, `/meetings/not-found`) — anchored to a single segment so
  // multi-segment paths (authenticated `/meetings/:id/edit`, unknown `/meetings/:id/extra`) fall
  // through to the default `required` auth and reach login before the in-shell catch-all 404.
  // `/meetings/create` is also a single segment, so it stays optional here. Unlike `/groups/create`
  // (skipped by `authenticatedMatchGuard`), the in-app `meetings` child has no canMatch, so
  // anonymous SSR matches `MEETING_ROUTES`'s `create` + `writerGuard` — it does not fall through
  // to public `meetings/:id`.
```

audigregorie (commented):
## Code Review Summary

Solid approach: a mutable `REQUEST_CONTEXT.notFound` flag plus a trailing in-shell catch-all is the right way to emit a real HTTP 404 without a `/not-found` redirect. The public-profile soft-404 fix and the anchored `/meetings`/`/groups` prefixes are genuine correctness wins. Two issues remain — one in-diff comment that still describes the wrong Angular path for `/meetings/create`, and one outside-the-diff fail-open that the new in-shell 404 makes newly visible.

### Major — outside the diff

**Remaining optional-auth prefixes now serve the authenticated shell to anonymous users.** Before this PR, an unknown URL under an optional prefix landed on the public standalone `/not-found` page (no left nav). The new trailing `path: ''` + `MainLayout` + `**` catch-all means those same URLs now SSR the authenticated chrome without a session, because `authGuard` returns `true` on the server.

Still fail-open (unchanged in this diff):
- `^/foundations/[^/]+/groups(?:/.*)?$` and `^/projects/[^/]+/groups(?:/.*)?$` — `/foundations/cncf/groups/extra` is optional-auth, does not match the exact Angular `foundations/:slug/groups` route, and falls through to the in-shell 404.
- Bare-string `startsWith('/invite/error')` — `/invite/error-extra` is the same class.

This is the same AC #4 hole that `a8fb7d1` closed for `/meetings/:id/extra` and `/groups/:id/edit`. Tighten the directory matchers to a single trailing `groups` segment (drop `(?:\/.*)?`) and add classification specs, matching the meeting/group change. `/invite/error` should be regex-anchored like `/auth-error`.

Do **not** “fix” the second `path: ''` parent by moving `**` into the first `MainLayout` children. `MEETING_ROUTES` has no `:id` child, so a first-parent `**` would swallow `/meetings/:id` after the lazy routes fail and break public meeting join.

### Already raised (not re-threaded)

- Untyped `renderContext` vs `ServerRequestContext` (`server.ts`) — agree; bind the producer to the interface.
- No authenticated e2e for the in-shell catch-all — agree; the public-profile spec does not cover this path.

### Protected files

`server.ts` and `auth.middleware.ts` are on the protected-file list by design (200→404 rebuild + whitelist change). Flagging for code-owner review as the PR already notes.

### What's done well

- Buffering the SSR body before rebuilding the `Response` avoids handing a consumed stream to `writeResponseToNodeResponse`.
- Public-profile 404 is set in the same SSR `tap` that writes TransferState, so the status flip happens before `handle()` resolves.
- Second empty-path `MainLayout` is the correct place for `**` — it runs only after public siblings (`meetings/:id`, `groups/:id`, `u/:username`) have had their chance.

apps/lfx-one/src/server/middleware/auth.middleware.ts
fayazg: Done in `05cfda56c` — went with the exclusion option. The matcher is now `/^\/meetings\/(?!create\/?$)[^/]+\/?$/`, so `/meetings/create` classifies `required` and an anonymous request reaches login instead of SSR-rendering the create shell as optional-auth. The comment is rewritten to match, and I added a classification regression case asserting anonymous `/meetings/create` → `res.oidc.login`. `/groups/create` has no reserved route, so its matcher is unchanged (noted in the comment).

apps/lfx-one/src/server/middleware/auth.middleware.ts
fayazg: Good catch on the resolution path — you're right that `/meetings/create` matches `MEETING_ROUTES`' explicit `create` + `writerGuard`, not public `meetings/:id`, so the earlier comment was wrong. `05cfda56c` takes the "Better" option: the matcher now excludes `create` via `(?!create\/?$)` so `/meetings/create` is `required` (anonymous → login, AC #4), the inaccurate sentence is gone, and there is a classification regression spec. The `/groups/create` asymmetry you noted (falls through to public `:id` via `authenticatedMatchGuard`) is intact and called out in the groups comment.

apps/lfx-one/src/server/server.ts
fayazg: Done in `05cfda56c` — `renderContext` is now typed `const renderContext: ServerRequestContext = { ... }`, and I added the `providers` field to `ServerRequestContext` so the interface covers the full shape the producer carries. A rename like `notFound` → `isNotFound` now fails the build at this literal instead of silently reading `undefined` in the consumers.

apps/lfx-one/src/app/app.routes.ts
fayazg: Done in `05cfda56c` — added `e2e/not-found-in-shell.spec.ts` (authenticated, default storageState). It GETs an unknown in-app URL and asserts `response.status() === 404`, the URL is preserved (no redirect), and the branded `not-found-card` renders inside the shell (`sidebar` + `main-content` visible). Complements the anonymous `/u/:username` spec.

fayazg:
## Review feedback addressed — commit `05cfda56c`

Thanks @dealako and @audigregorie for the thorough pass. All four inline threads are addressed and resolved:

### Changes made
- **`auth.middleware.ts`** — excluded the reserved `create` segment from the public `/meetings/:id` matcher via `/^\/meetings\/(?!create\/?$)[^/]+\/?$/`, so `/meetings/create` classifies `required` and anonymous requests reach login (AC #4) instead of SSR-rendering the create shell as optional-auth. Rewrote the comment to match the real `MEETING_ROUTES` + `writerGuard` resolution path (dropping the inaccurate "bogus meeting → its own not-found" claim). `/groups/create` has no reserved route and stays as-is. (per @dealako finding 1 / @audigregorie)
- **`auth.middleware.spec.ts`** — added a classification regression case: anonymous `/meetings/create` → `res.oidc.login`.
- **`server.ts`** + **`server-request-context.interface.ts`** — typed `renderContext` as `ServerRequestContext` and added the `providers` field to the interface, so field drift (e.g. a `notFound` rename) fails the build at the producer instead of silently reading `undefined`. (per @dealako finding 2)
- **`e2e/not-found-in-shell.spec.ts`** — new authenticated spec: GETs an unknown in-app URL and asserts HTTP 404, URL preserved (no redirect), and the branded `not-found-card` inside the shell (`sidebar` + `main-content`). Complements the anonymous `/u/:username` spec. (per @dealako finding 3)

### No change needed
- **Nit — host `**` as a child of the existing shell node instead of a second `path: ''` parent:** the separate trailing `path: ''` parent is intentional and load-bearing. A `**` child inside the first shell node prevents the router from backtracking out of the empty-path subtree and would shadow every public sibling (`docs`, `meetings/:id`, `groups/:id`, `u/:username`, `foundations/:x/groups`, …). The client-side layout re-instantiation on a 404 is the accepted cost. (flagged by @dealako)
- **Nit — two divergent `inject(REQUEST_CONTEXT)` cast patterns:** the two files this PR touches (`not-found.component.ts`, `public-profile-page.component.ts`) already both cast to `ServerRequestContext`. The inline-object casts live in pre-existing files (`app.component.ts`, `runtime-config.provider.ts`) untouched here — leaving that alignment as a follow-up as you suggested. (flagged by @dealako)

### Question answered
- **AC #4 scope** (other permissive matchers — `/foundations/:x/groups/…`, `/projects/:x/groups/…` deep tails, `/invite/error` prefix): intentionally out of scope for this PR. Added a one-line note to the PR description; the comprehensive per-pattern anchoring is tracked as **LFXV2-3254** (split from LFXV2-3095). (asked by @dealako)

### Threads resolved
4 of 4 resolved.

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -26,11 +26,20 @@ const DEFAULT_ROUTE_CONFIG: RouteAuthConfig[] = [
...
+  // classifies `required` — an anonymous `/meetings/create` reaches login rather than SSR-rendering
+  // the create shell as optional-auth. Keeping `required` here means the SSR boundary matches the
+  // Angular `[authGuard, writerGuard]` gate instead of relying on it alone.
+  { pattern: /^\/meetings\/(?!create\/?$)[^/]+\/?$/, type: 'ssr', auth: 'optional' },
+
+  // Public group detail (`/groups/:id`, `/groups/not-found`) — same single-segment anchoring so
+  // multi-segment paths don't fail-open onto the in-shell catch-all; anonymous access with optional
+  // auth for membership enrichment. No reserved single-segment child (groups has no `create` route),
+  // so no lookahead exclusion is needed here.
+  { pattern: /^\/groups\/[^/]+\/?$/, type: 'ssr', auth: 'optional' },
```
dealako, line 47: [minor] New groups comment states a false invariant — `/groups/create` IS a real protected route

Issue: The justification for omitting the `(?!create\/?$)` lookahead here — "No reserved single-segment child (groups has no `create` route)" — is factually wrong.

Proof: `committees.routes.ts:16-19` defines `{ path: 'create', canActivate: [authGuard, writerGuard] }`, and `app.routes.ts:351-354` mounts `COMMITTEE_ROUTES` at `path: 'groups'`. So `/groups/create` is a real writer-gated route, structurally identical to `/meetings/create`. The reason groups is actually safe is different: the shell `groups` mount carries `canMatch: [authenticatedMatchGuard]` (`app.routes.ts:352`; guard at `authenticated-match.guard.ts:20-27` returns `reqContext?.auth?.authenticated ?? false`), so an anonymous SSR request never matches the create shell and falls through to the public `groups/:id` view (`app.routes.ts:471`). The shell `meetings` mount (`app.routes.ts:342`) has NO `authenticatedMatchGuard` — which is precisely why meetings needed the middleware lookahead and groups does not.

Why it matters: The comment documents the wrong invariant for a security boundary. A maintainer who later removes the `authenticatedMatchGuard` gate — the *actual* protection — trusting that "groups has no create route" would silently expose anonymous SSR of `/groups/create` as optional-auth, the exact hazard the meetings lookahead closes.

Fix: Reword to state the real reason: the top-level `groups` route is behind `canMatch: [authenticatedMatchGuard]`, so anonymous `/groups/create` never renders the create shell and falls through to the public `groups/:id` view — no middleware lookahead needed. (Optionally mirror the meetings fix with `/^\/groups\/(?!create\/?$)[^/]+\/?$/` for symmetry.) Consider a regression test asserting anonymous `/groups/create` does not render the create shell.

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -26,11 +26,20 @@ const DEFAULT_ROUTE_CONFIG: RouteAuthConfig[] = [
...
-  // Public group detail - anonymous access with optional auth for membership enrichment
-  { pattern: '/groups/', type: 'ssr', auth: 'optional' },
+  // Public meeting join (`/meetings/:id`, `/meetings/not-found`) — anchored to a single segment so
+  // multi-segment paths (authenticated `/meetings/:id/edit`, unknown `/meetings/:id/extra`) fall
+  // through to the default `required` auth and reach login before the in-shell catch-all 404. The
+  // reserved single-segment `create` route is excluded via the `(?!create\/?$)` lookahead so it also
+  // classifies `required` — an anonymous `/meetings/create` reaches login rather than SSR-rendering
+  // the create shell as optional-auth. Keeping `required` here means the SSR boundary matches the
+  // Angular `[authGuard, writerGuard]` gate instead of relying on it alone.
+  { pattern: /^\/meetings\/(?!create\/?$)[^/]+\/?$/, type: 'ssr', auth: 'optional' },
```
dealako: [minor] Matrix-parameter suffix bypasses the new `create` lookahead (agrees with Copilot)

Issue: `/meetings/create;source=link` still classifies as optional-auth, defeating the lookahead just added.

Proof: Express leaves the matrix-parameter suffix in `req.path`, so the segment is `create;source=link`. The negative lookahead `(?!create\/?$)` only fails when `create` is followed by `/` or end-of-string; here it's followed by `;`, so the lookahead succeeds and `[^/]+` matches the whole `create;source=link` → `optional`. Angular, meanwhile, strips matrix params and routes the segment as `create` → the protected `MEETING_ROUTES` create (`meetings.routes.ts:16`). Same divergence the lookahead was meant to eliminate.

Why it matters: The SSR auth boundary and the Angular route boundary disagree again for a reserved write route. Not exploitable today (writerGuard redirects on null slug), but it re-opens the classification gap and shows regex path-matching is fragile against matrix/encoding tricks.

Fix: Normalize the segment before matching (strip matrix params, e.g. compare `req.path.split(';')[0]`) or broaden the exclusion to `(?!create(?:[;/]|$))`. Add a regression case for `/meetings/create;source=link`. This overlaps with the LFXV2-3254 hardening — a pointer there is fine if you'd rather batch it, but call it out explicitly since it undercuts this commit's stated invariant.

apps/lfx-one/e2e/not-found-in-shell.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    expect(response?.status()).toBe(404);
+    // URL is preserved — no client-side redirect to /not-found or the login provider.
+    await expect(page).toHaveURL(/\/__lfx-nonexistent-e2e-route__$/);
+    await expect(page).not.toHaveURL(/auth0\.com/);
+
+    // Branded 404 renders inside the shell: the left-nav sidebar and main content region are present.
+    await expect(page.getByTestId('not-found-card')).toBeVisible({ timeout: RENDER_TIMEOUT });
+    await expect(page.getByTestId('sidebar')).toBeVisible({ timeout: RENDER_TIMEOUT });
+    await expect(page.getByTestId('main-content')).toBeVisible();
```
dealako, line 46: [minor] `sidebar` visibility assertion will fail on the `mobile-chrome` project (agrees with Cursor)

Issue: `await expect(page.getByTestId('sidebar')).toBeVisible()` assumes the desktop left nav is rendered, but this spec has no viewport override and runs under every project — including `mobile-chrome` (Pixel 5, 393px).

Proof: The desktop sidebar container is `class="hidden lg:flex ..."` (`main-layout.component.html:10`), wrapping the `<lfx-sidebar>` that carries `data-testid="sidebar"` (`sidebar.component.html:9`). Below the `lg` breakpoint it's `display:none`, so `toBeVisible()` fails. The `mobile-chrome` project (`playwright.config.ts:66-71`, Pixel 5, authenticated storageState) runs this file. The mobile sidebar (`main-layout.component.html:53`) lives in an `lg:hidden` drawer that's closed by default, so it isn't visible either.

Why it matters: This is a concrete CI break on the mobile project, not a theoretical one — and it lands in the very test added to close F2.

Fix: Either scope the shell assertion to desktop (e.g. `test.skip(({ isMobile }) => isMobile)` for the sidebar check, or set `test.use({ viewport: { width: 1280, height: 720 } })`), or assert a shell element that's present in both layouts (e.g. `main-content`, which you already check) and drop the `sidebar` visibility assertion on mobile. Keep the 404 + URL-preserved assertions viewport-agnostic.

dealako (changes requested):
Thanks for the fast, thorough turnaround, @fayazg 🙌 — you closed most of the prior round cleanly, and filing **LFXV2-3254** to carry the broader permissive-matcher hardening (F6) out of scope is exactly the right move.

### 👏 Nice work this revision
- **F2 resolved** — `e2e/not-found-in-shell.spec.ts` now covers the primary path: authenticated unknown URL → HTTP 404 at the original path, URL preserved (no auth0 redirect), rendered in-shell. That was the highest-value gap.
- **F3 resolved** — `renderContext: ServerRequestContext` + adding `providers?: StaticProvider[]` binds the `server.ts` literal to the contract, so a field rename now fails the build.
- **F1 (meetings side) resolved** — the `(?!create\/?$)` lookahead correctly forces `/meetings/create` to `required`, with a matching regression spec. Regex verified: it excludes only `create` / `create/`, not `createXYZ` or ids.

### Revision status
- ✅ **Resolved:** F2 (authenticated catch-all e2e), F3 (typed render context), F1 meetings half.
- ⚠️ **Partially addressed:** F1 — the *groups* half regressed into an inaccurate comment (see 🟡 1).
- 📋 **Deferred (accepted):** F6 → LFXV2-3254.
- ◽ **Still open (nits/trade-offs, non-blocking):** F4 (duplicate `path: ''` → MainLayout re-instantiation on client nav), F5 (divergent `inject(REQUEST_CONTEXT)` casts).

### Issue count (open)
- 🔴 Blocking: 0
- 🟡 Minor: 3 — (1) the new `/groups` comment asserts a **false invariant** ("groups has no `create` route"); (2) the new authenticated e2e's `sidebar` assertion **will fail on the `mobile-chrome` Pixel 5 project** (desktop sidebar is `hidden lg:flex`) — agrees with Cursor; (3) matrix-parameter bypass `/meetings/create;source=link` slips past the new lookahead — agrees with Copilot.
- ⚪ Nit: 2 — F4, F5 carried over.

### Bot reconciliation
- **Agree** with **Cursor** (`not-found-in-shell.spec.ts:41`, sidebar-on-mobile) — verified against `main-layout.component.html:10` + the `mobile-chrome` project in `playwright.config.ts`.
- **Agree** with **Copilot** (`auth.middleware.ts:36`, matrix-param) — Express keeps the `;source=link` suffix in `req.path`, so the lookahead doesn't fire while Angular still routes to the protected `create`.
- 🟡 1 (groups comment) is not raised by any bot — it's the residual half of F1.

### Decision
🔴 **Needs changes before approval** — all three are minor and none is exploitable today (writerGuard / `authenticatedMatchGuard` still deny), but 🟡 2 is a concrete CI break on the mobile project, and 🟡 1 documents a false security invariant in an auth file. Close these three (a comment fix + a viewport-aware assertion + a matrix-param guard/test) and I'm happy to approve on the next push.

fayazg:
### Follow-up — restore the "Powered by LFX" header on the `/u/:username` 404 (commit `285302678`)

The not-found and private-profile states rendered only the message card; the `public-profile-topbar` ("Powered by LFX") that appears on the loaded profile was missing.

- **`public-profile-page.component.html`** — render `<lfx-public-profile-topbar>` above the not-found view inside the same `max-w-[1240px]` container the loaded profile uses, and own the `min-h-screen bg-gray-50` background at the page level.
- **`public-profile-not-found.component.html`** — drop its own full-screen `bg-gray-50 min-h-screen` wrapper so it composes cleanly under the header (background/height now owned by the page).

The topbar handles the no-profile case gracefully (`basic` defaults to `null`; only the "Powered by" pill shows since there's nothing to scroll). No behavior change to the HTTP 404 or the existing `public-profile-not-found-card` testid.

luismoriguerra (commented):
## Independent audit: PR #1471

**Verdict:** 🔴 Required changes
**Merge readiness:** blocked on the two auth-bypass gaps below; e2e sidebar assertion should also be fixed before this ships (not currently gating CI — no e2e job runs here).

Audited via ephemeral checkout at head `2853026`, CI re-checked (all green — no e2e job present).

### Findings
1. **Blocking** — `auth.middleware.ts:36` — `/meetings/create;source=link` still classifies `optional` auth: the `(?!create\/?$)` lookahead only fails on `create` followed by `/` or end-of-string, and Express leaves the matrix-parameter suffix in `req.path`, so the segment is `create;source=link` and the lookahead still passes. Since `authGuard` unconditionally returns `true` during SSR (`auth.guard.ts:25-27`), an anonymous request to that URL still SSR-renders the meeting-create shell — the exact bypass AC #4 exists to close. *Already threaded* (Copilot + dealako on this line) — not duplicating inline, just confirming still open at head.
2. **Blocking** — `auth.middleware.ts:40,42` — `/groups/create` is a real protected route (`committees.routes.ts:16-19`, `canActivate: [authGuard, writerGuard]`, mounted at `path: 'groups'`), but the comment claims "groups has no `create` route" and the pattern (`/^\/groups\/[^/]+\/?$/`) is unchanged, so `/groups/create` still classifies `optional` — same SSR-trust bypass as #1. *Already threaded* (dealako on this line) — confirming still open at head.
3. **Blocking** — `e2e/not-found-in-shell.spec.ts:41` — `getByTestId('sidebar')).toBeVisible()` will fail on the `mobile-chrome` Playwright project (Pixel 5, 393px): the sidebar wrapper is `hidden lg:flex` (`main-layout.component.html:10`) and the mobile drawer is closed by default. *Already threaded* (Cursor Bugbot + dealako on this line) — confirming still open at head; not currently visible in `gh pr checks` since no e2e job runs in this repo's CI.

### Verified as fixed
- `renderContext` in `server.ts:484` is now typed `ServerRequestContext`, and the interface includes `providers` — the two `dealako` typing nits are resolved.

All three blocking findings above match existing open review threads from Copilot / Cursor Bugbot / dealako — posting this as a summary confirmation rather than duplicating new inline comments.

fayazg:
Follow-up (commit `e2711a1a1`): changed the `/u/:username` not-found page background from `bg-gray-50` to `bg-white` so the white "Powered by LFX" header no longer reads as a separate box — the not-found/private view is now a single continuous white surface, matching the loaded-profile chrome.

apps/lfx-one/src/app/modules/public-profile/public-profile-page/public-profile-page.component.html
```diff
@@ -2,7 +2,13 @@
 <!-- SPDX-License-Identifier: MIT -->
 
 @if (notFound()) {
-  <lfx-public-profile-not-found [isPrivate]="isPrivate()" data-testid="public-profile-not-found"></lfx-public-profile-not-found>
+  <!-- Keep the "Powered by LFX" branding header on the not-found/private view, matching the loaded profile chrome. -->
+  <div class="flex min-h-screen flex-col bg-white text-gray-900">
+    <div class="mx-auto w-full max-w-[1240px] px-4 sm:px-10">
```
dealako: [nit] Two small polish items on the restored not-found chrome

1. **Redundant `w-full` / container drift.** This not-found container is `mx-auto w-full max-w-[1240px] px-4 sm:px-10`, but the loaded (`@else`) branch uses `mx-auto max-w-[1240px] px-4 pb-20 sm:px-10` (no `w-full`). The parent is `flex min-h-screen flex-col`, so a block child already stretches to full width — `w-full` is a no-op here. Drop it to mirror the loaded branch (or add to both for parity).

2. **No coverage for the restored header.** The whole point of `285302678` is that the "Powered by LFX" topbar renders on the 404 view, but the anonymous e2e (`public-profile-not-found.spec.ts`) only asserts `public-profile-not-found-card` — a regression dropping the header would pass CI. Since the topbar carries `data-testid="public-profile-topbar"`, one line closes it: `await expect(page.getByTestId('public-profile-topbar')).toBeVisible()` in the existing 'renders the not-found view in place' test.

Both are take-it-or-leave-it — neither blocks.

dealako (changes requested):
Thanks @fayazg — these two commits (`285302678`, `e2711a1a1`) cleanly restore the "Powered by LFX" chrome on the `/u/:username` not-found/private view and fix the white-on-white header by moving background + full-height ownership up to the page component. Nice touch composing the topbar above `<lfx-public-profile-not-found>` so the 404 matches the loaded-profile chrome; the topbar import is wired and the `public-profile-not-found-card` testid the anonymous e2e relies on is intact.

These commits are presentational-only and introduce **no new blocking/minor issues** — just two nits. The reason I'm keeping this at Changes Requested is that my three **minor** findings from the previous round are still open (they were posted almost simultaneously with these commits, so you likely hadn't seen them yet — no worries, just flagging so they don't slip).

### Revision status
- 🆕 This round (presentational): clean. Two ⚪ nits below.
- ❌ **Still open from last round (carried, minor):**
  - **N1** — `auth.middleware.ts:40` still says "groups has no `create` route" (factually false; real protection is `canMatch: [authenticatedMatchGuard]`).
  - **N2** — `not-found-in-shell.spec.ts:40` `sidebar` visibility assertion still fails on the `mobile-chrome` (Pixel 5) project — a concrete CI break.
  - **N3** — matrix-param `/meetings/create;source=link` still bypasses the `(?!create\/?$)` lookahead.
- ◽ **Still open (nits/deferred):** F4 (duplicate `path: ''` → MainLayout re-instantiation), F5 (divergent `inject(REQUEST_CONTEXT)` casts), F6 → deferred to **LFXV2-3254**.

### Issue count (open)
- 🔴 Blocking: 0
- 🟡 Minor: 3 — N1, N2, N3 (all carried over, unchanged; N2 is a mobile-CI failure).
- ⚪ Nit: 4 — (new) no test asserts the restored topbar on the 404 view; (new) redundant `w-full` class drift vs the loaded-branch container; plus carried F4, F5.

### Bot reconciliation
No new CodeRabbit/Cursor/Copilot comments landed on these two commits. The prior agreements stand: N2 ↔ Cursor, N3 ↔ Copilot.

### Decision
🔴 **Needs changes before approval** — purely to close the three carried-over minors (especially N2, which will red the mobile Playwright project). The new presentational work itself is good; fold the two nits in if convenient and I'll approve once N1–N3 are addressed.

apps/lfx-one/src/server/middleware/auth.middleware.ts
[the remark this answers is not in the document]
fayazg: Fixed in 24f9b7e3c. Widened the lookahead to `(?!create(?:[/;]|$))` so the `;` matrix-parameter delimiter is excluded too — `/meetings/create;source=link` now classifies `required` and an anonymous SSR request reaches login, matching how Angular strips the matrix suffix and routes the protected `create`. Added a matrix-parameter regression case to `auth.middleware.spec.ts` (9/9 pass).

apps/lfx-one/e2e/not-found-in-shell.spec.ts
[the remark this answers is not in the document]
fayazg: Fixed in 24f9b7e3c. Pinned the describe block to a desktop viewport (`test.use({ viewport: { width: 1440, height: 900 } })`) so the in-shell left-nav (`hidden lg:flex`) is rendered regardless of the Playwright project — the assertion no longer fails under the mobile-chrome (Pixel 5) project.

apps/lfx-one/src/server/middleware/auth.middleware.ts
fayazg: Good catch — fixed in 24f9b7e3c. Widened the lookahead to `(?!create(?:[/;]|$))` so `;` is treated as a route delimiter alongside `/` and end-of-string. `/meetings/create;source=link` now classifies `required`, closing the Express-keeps-matrix / Angular-strips-matrix divergence you and Copilot both flagged. Added a matrix-parameter regression case to the spec.

apps/lfx-one/e2e/not-found-in-shell.spec.ts
fayazg: Fixed in 24f9b7e3c. Pinned the describe block to a 1440×900 viewport via `test.use`, so the `hidden lg:flex` sidebar renders under every project including mobile-chrome (Pixel 5). The in-shell assertion now holds across the matrix.

apps/lfx-one/src/server/middleware/auth.middleware.ts
fayazg: You're right — corrected the comment in 24f9b7e3c. It now states the accurate invariant: `/groups/create` IS a real writer-gated route (`COMMITTEE_ROUTES` mounted at `groups` in `app.routes.ts`), and groups is safe not because the route is absent but because that shell mount carries `canMatch: [authenticatedMatchGuard]` — an anonymous SSR match skips the authenticated subtree and falls through to the public `groups/:id`, so `create` never reaches the protected shell. The comment also notes meetings has no such canMatch, which is why its matcher must exclude `create` explicitly.

apps/lfx-one/src/app/modules/public-profile/public-profile-page/public-profile-page.component.html
fayazg: Both taken — a68b14eab. Dropped the redundant `w-full` so the not-found container mirrors the loaded (`@else`) branch, and added `await expect(page.getByTestId('public-profile-topbar')).toBeVisible()` to the anonymous 'renders the not-found view in place' test so the restored 'Powered by LFX' header is now covered.

fayazg:
## Review Feedback Addressed

Commits: `24f9b7e3c`, `a68b14eab`

### Changes Made
- **auth.middleware.ts** — widened the `create` lookahead to `(?!create(?:[/;]|$))` so a matrix-parameter suffix (`/meetings/create;source=link`) also classifies `required`; Express keeps the `;...` in `req.path` while Angular strips it and routes the protected `create`, so both must gate to login (per @dealako, @luismoriguerra, copilot[bot])
- **auth.middleware.spec.ts** — added a matrix-parameter regression case (9/9 pass)
- **auth.middleware.ts** — corrected the `groups` comment: `/groups/create` IS a real writer-gated route; groups is safe via the `canMatch: [authenticatedMatchGuard]` shell mount, not an absent route (per @dealako)
- **not-found-in-shell.spec.ts** — pinned a desktop viewport so the in-shell sidebar (`hidden lg:flex`) assertion holds under the mobile-chrome (Pixel 5) project (per @dealako, cursor[bot])
- **public-profile-page.component.html** — dropped the redundant `w-full` on the not-found container to mirror the loaded branch (per @dealako)
- **public-profile-not-found.spec.ts** — asserted the "Powered by LFX" topbar is visible on the anonymous not-found view, covering the restored header (per @dealako)

### Threads Resolved

apps/lfx-one/e2e/public-profile-not-found.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  test('renders the not-found view in place without a redirect', async ({ page }) => {
+    const response = await page.goto(NONEXISTENT, { waitUntil: 'domcontentloaded' });
+    expect(response?.status()).toBe(404);
+    await expect(page).toHaveURL(/\/u\/__lfx-nonexistent-e2e-user__$/);
+    await expect(page).not.toHaveURL(/auth0\.com/);
+    await expect(page.getByTestId('public-profile-not-found-card')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    // The "Powered by LFX" branding topbar must stay on the not-found view — guards against a regression
+    // that drops the header while still rendering the not-found card.
+    await expect(page.getByTestId('public-profile-topbar')).toBeVisible();
```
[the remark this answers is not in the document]
fayazg: Good catch — fixed in 50753011d. The `data-testid="public-profile-topbar"` was on both the component's own template root and the `<lfx-public-profile-topbar>` host in the page. Removed the redundant host copies from both branches (not-found and loaded), so the component root now carries the single canonical testid and the e2e assertion resolves to exactly one node under strict mode.

fayazg:
## Review Feedback Addressed

Commit: `50753011d`

### Changes Made
- **public-profile-page.component.html** — removed the redundant `data-testid="public-profile-topbar"` from both `<lfx-public-profile-topbar>` host usages (not-found and loaded branches); the component's own template root already carries the testid, so the duplicate made Playwright's strict locator match two nodes and fail the header assertion added in the prior commit (per cursor[bot])

### Threads Resolved

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -26,11 +26,25 @@ const DEFAULT_ROUTE_CONFIG: RouteAuthConfig[] = [
...
+  // Public meeting join (`/meetings/:id`, `/meetings/not-found`) — anchored to a single segment so
+  // multi-segment paths (authenticated `/meetings/:id/edit`, unknown `/meetings/:id/extra`) fall
+  // through to the default `required` auth and reach login before the in-shell catch-all 404. The
+  // reserved single-segment `create` route is excluded via the `(?!create(?:[/;]|$))` lookahead so it
+  // also classifies `required` — an anonymous `/meetings/create` reaches login rather than SSR-rendering
+  // the create shell as optional-auth. The `[/;]` delimiter class also excludes matrix-parameter suffixes
+  // (`/meetings/create;source=link`): Express leaves the `;...` in `req.path` while Angular strips matrix
+  // params and routes the segment as the protected `create`, so both must classify `required` to keep the
+  // SSR boundary aligned with the Angular `[authGuard, writerGuard]` gate instead of relying on it alone.
+  { pattern: /^\/meetings\/(?!create(?:[/;]|$))[^/]+\/?$/, type: 'ssr', auth: 'optional' },
```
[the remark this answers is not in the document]
fayazg: Fixed in fc7790193. Rather than patch the `create` regex per-encoding, I decode the pathname once inside `classifyRoute` before matching, so the SSR classification aligns with Angular's decoded route matching. `/meetings/%63reate` now decodes to `/meetings/create`, the existing lookahead fires, and the anonymous request classifies `required` → login. A malformed escape (`%ZZ`) throws in `decodeURIComponent` and fails closed to the default `required`. Added encoded-`create` and malformed-escape regression cases (11/11 pass).

fayazg:
## Review Feedback Addressed

Commit: `fc7790193`

### Changes Made
- **auth.middleware.ts** — decode the pathname once inside `classifyRoute` before pattern matching, with a fail-closed guard. Express/Node leave `req.path` percent-encoded, so classifying the raw path let an encoded segment (`/meetings/%63reate`) fail-open to optional auth while Angular decoded it to the protected `create` route. A single decode mirrors Angular's decoding so the existing `create` lookahead fires; a malformed escape (`%ZZ`) throws and falls closed to `required`. Centralizes the fix for the whole encoding-divergence class (per copilot[bot])
- **auth.middleware.spec.ts** — added encoded-`create` and malformed-escape regression cases

### Threads Resolved

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -108,22 +117,36 @@ const DEFAULT_CONFIG: AuthConfig = {
...
+  // route matching (`/meetings/%63reate` → `/meetings/create`, so the `create` lookahead still fires).
+  // Express/Node leave `req.path` percent-encoded, so classifying the raw path would let an encoded
+  // segment fail-open to optional auth while Angular decodes it to the protected route. A single decode
+  // mirrors Angular's single decode; a malformed escape (`%ZZ`) throws and fails closed to `required`.
+  let decodedPath: string;
+  try {
+    decodedPath = decodeURIComponent(path);
+  } catch {
+    return fallback;
+  }
```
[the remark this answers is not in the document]
fayazg: Done — switched to per-segment decoding in `classifyRoute`. Each segment is `decodeURIComponent`d independently and the classifier fails closed to `required` if a decoded segment reintroduces a `/` (encoded separator) or throws (malformed escape), so `/u%2Fsomeone` no longer collapses into the two-segment public `/u/someone` — it matches Angular's split-then-decode order and falls through to the protected catch-all. Added a `/u%2Fsomeone` regression case alongside the encoded-`create` and malformed-escape tests.

See commit 64fb4d213.

fayazg:
## Review Feedback Addressed

Commit: 64fb4d213

### Changes Made
- **apps/lfx-one/src/server/middleware/auth.middleware.ts**: `classifyRoute` now decodes the request path **per segment** instead of decoding the whole path at once. Each segment is `decodeURIComponent`d independently and the classifier fails closed to `required` auth if a decoded segment reintroduces a `/` (encoded separator like `%2F`) or throws (malformed escape). This matches Angular's split-on-`/`-then-decode order, so `/u%2Fsomeone` no longer collapses into the two-segment public `/u/someone` and fall-opens — it falls through to the protected in-shell catch-all. (per copilot[bot])
- **apps/lfx-one/src/server/middleware/auth.middleware.spec.ts**: added a `/u%2Fsomeone` regression case asserting fail-closed classification (12 tests total).

### Threads Resolved

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -108,22 +117,50 @@ const DEFAULT_CONFIG: AuthConfig = {
...
+      .map((segment) => {
+        const decodedSegment = decodeURIComponent(segment);
+        if (decodedSegment.includes('/')) {
+          throw new Error('encoded path separator');
+        }
+        return decodedSegment;
+      })
+      .join('/');
+  } catch {
+    return fallback;
```
dealako: [minor] Fail-closed fallback drops the API classification for malformed/encoded `/api` paths (agrees with Copilot)

Issue: On any decode failure — malformed escape (`%ZZ` → `decodeURIComponent` throws) or encoded separator (`%2F` → decoded segment contains `/`) — this returns `fallback`, which is `{ type: config.defaultType ('ssr'), auth: 'required' }` with no `tokenRequired`. That happens *before* the route loop, so an `/api/...` path never reaches its `{ pattern: '/api', type: 'api', auth: 'required', tokenRequired: true }` row.

Proof: `fallback` is defined at lines 121-125 (no `tokenRequired`); the `catch` at line 150 returns it. I verified both triggers hit it for API paths: `GET /api/x/%ZZ` (throws) and `GET /api/x%2Fy` (segment decodes to `x/y`, `includes('/')` throws). Consequence per branch: (a) anonymous → classified `ssr`+`required` → **HTML login redirect (302)** instead of the API **401 JSON** contract; (b) authenticated → fallback has no `tokenRequired`, so **bearer-token extraction is skipped** and the downstream API call proceeds without it.

Why it matters: Security-wise it's still fail-closed (auth stays `required`), so no bypass — but it silently breaks the API error contract for direct requests and mis-handles authenticated API calls whose path carries a malformed/encoded segment.

Fix: On the fail-closed branch, preserve the API shape for raw API-prefixed paths, e.g. before returning the generic fallback: `if (path.startsWith('/api') || path.startsWith('/public/api')) return { pattern: '/api', type: 'api', auth: 'required', tokenRequired: true };` (keep `auth: 'required'` — still fail-closed). Add a regression test asserting `GET /api/foo/%ZZ` classifies `type: 'api'` / `tokenRequired: true` (no login redirect).

dealako (changes requested):
This is a genuinely excellent revision, @fayazg — you didn't just patch the symptoms, you closed the whole class. 👏

### 👏 Nice work this round
- **N1 fully corrected** — the `/groups` comment now states the *real* invariant: `/groups/create` is a live writer-gated route, protected by `canMatch: [authenticatedMatchGuard]` (anonymous SSR skips the authenticated subtree and falls through to public `groups/:id`), and it correctly contrasts that meetings has *no* such `canMatch`, which is why the meetings matcher needs the `create` lookahead. Accurate on every clause.
- **N3 closed at the root** — rather than whack-a-mole the regex per encoding, you moved to **per-segment `decodeURIComponent` before matching**, widened the lookahead to `(?!create(?:[/;]|$))` for matrix params, and made it **fail closed** on encoded separators (`%2F`) and malformed escapes (`%ZZ`). I traced mixed-case `%2f`, double-encoding, dot-segments, and null bytes — no residual anonymous fail-open onto a protected route. The 4 new regression tests each pin a distinct branch. Really nice.
- **N2 resolved** — pinning `test.use({ viewport: { width: 1440, height: 900 } })` keeps the `hidden lg:flex` sidebar visible across the mobile-chrome project.
- **Round-3 nits (F4/F5) resolved** — `w-full` dropped to mirror the loaded branch; topbar e2e assertion added; and the duplicate `data-testid` correctly removed from both hosts so `getByTestId` resolves to the single component-root testid (agrees with Cursor's catch).

### Revision status
- ✅ **Resolved:** N1, N2, N3, F4, F5, plus Cursor's duplicate-testid.
- 📋 **Deferred (accepted):** F6 → LFXV2-3254.
- ❌ **New, still open (minor):** the `classifyRoute` fail-closed fallback returns the generic **SSR** type, so malformed/encoded `/api` paths lose their API classification — see 🟡 1 (agrees with Copilot's open thread).

### Issue count (open)
- 🔴 Blocking: 0
- 🟡 Minor: 1 — decode-failure fallback drops `type: 'api'` / `tokenRequired` for `/api` paths (Copilot).
- ⚪ Nit: 1 — the `%2F`/malformed fail-close is applied globally to string-pattern public routes too (`/public/api`, `/docs`); fail-closed and not reachable today (emitted identifiers are alphanumeric), so take-it-or-leave-it.

### Bot reconciliation
- **Agree** with **Copilot** (`auth.middleware.ts:150`) — the API-type regression is real and reachable by direct request; details + verification in the inline. This is the one thing I'd close before merge.
- **Agree** with **Cursor** (duplicate topbar testid) — already fixed in `50753011d`.

### Decision
🔴 **Needs changes before approval** — a single, well-scoped minor stands between this and a clean approval. Preserve `type: 'api'` + `tokenRequired: true` for raw `/api`-prefixed paths on the fail-closed branch (auth level stays `required`), add an `/api/.../%ZZ` regression case, and I'll approve on the next push. Everything else is merge-ready.

apps/lfx-one/src/server/middleware/auth.middleware.ts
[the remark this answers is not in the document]
fayazg: Done — the decode-failure branch now fails closed to an API classification for raw `/api`-prefixed paths (`type: 'api'`, `auth: 'required'`, `tokenRequired: true`), mirroring the `/api` catch-all row. A malformed/encoded `GET /api/.../%ZZ` now returns a JSON 401 via `next(AuthenticationError)` instead of an HTML login redirect, and an authenticated request still triggers bearer-token extraction. Non-API malformed paths keep the SSR `required` fallback. Added a `/api/foo/%ZZ` regression case asserting no login redirect + a 401.

See commit d63939c3b.

fayazg:
## Review Feedback Addressed

Commit: d63939c3b

### Changes Made
- **apps/lfx-one/src/server/middleware/auth.middleware.ts**: the per-segment decode-failure branch previously fell back to the generic SSR classification (`type: 'ssr'`, no `tokenRequired`), so a malformed or encoded-separator path under `/api` was misclassified — an anonymous `GET /api/.../%ZZ` got an HTML login redirect instead of a JSON 401, and an authenticated request skipped bearer-token extraction. It now fails closed to an **API** classification (`type: 'api'`, `auth: 'required'`, `tokenRequired: true`) for raw `/api`-prefixed paths, mirroring the `/api` catch-all row. Non-API malformed paths keep the SSR `required` fallback. (per copilot[bot])
- **apps/lfx-one/src/server/middleware/auth.middleware.spec.ts**: added a `/api/foo/%ZZ` regression case asserting no login redirect and a 401 via `next(AuthenticationError)` (13 tests total).

### Threads Resolved

apps/lfx-one/src/server/middleware/auth.middleware.ts
fayazg: Done — thanks for the thorough read here, and you're right on both the core fix and the `/public/api` extension.

The decode-failure fail-closed branch now preserves the API classification (`type: 'api'`, `auth: 'required'`, `tokenRequired: true`) for **both** the raw `/api` mount and the `/public/api` mount, so a malformed/undecodable path returns a JSON 401 via `next(AuthenticationError)` instead of an HTML login redirect an XHR/fetch client can't follow. `/public/api` normally allows anonymous (optional auth), but an undecodable path can't be trusted to be that public route, so it fails closed to `required` — same reasoning you laid out.

Regression coverage added: `/api/foo/%ZZ` (2f7c5d98d earlier via d63939c3b) and now `/public/api/foo/%ZZ`, both asserting a 401 rather than a login redirect.

The core `/api` case landed in `d63939c3b`; the `/public/api` extension is in `2f7c5d98d`.

fayazg:
## Review Feedback Addressed

Commit: 2f7c5d98d

### Changes Made
- **apps/lfx-one/src/server/middleware/auth.middleware.ts**: extended the decode-failure fail-closed branch to also classify `/public/api` paths as `type: 'api'` (`required`, `tokenRequired`), so a malformed/undecodable `/public/api/...` request returns a JSON 401 rather than an HTML login redirect an XHR/fetch client can't follow (per @dealako — agreeing with copilot[bot])
- **apps/lfx-one/src/server/middleware/auth.middleware.spec.ts**: added a `/public/api/foo/%ZZ` regression case asserting a 401 (AuthenticationError), not a login redirect

### Context
The core `/api`-prefix fix landed earlier in `d63939c3b`; this commit generalizes it to the `/public/api` mount that @dealako's suggested fix also called out. `/public/api` normally allows anonymous access, but an undecodable path can't be trusted to be that public route, so it fails closed to `required`.

### Threads Resolved

apps/lfx-one/src/server/middleware/auth.middleware.ts
```diff
@@ -108,22 +117,70 @@ const DEFAULT_CONFIG: AuthConfig = {
...
+          throw new Error('encoded path separator');
+        }
+        return decodedSegment;
+      })
+      .join('/');
+  } catch {
+    // Decoding failed (malformed escape like `%ZZ`, or an encoded separator `%2F` that would shift
+    // segment boundaries). Fail closed — but keep API-prefixed paths classified as `api` so they get a
+    // JSON 401 with bearer extraction rather than an SSR login redirect.
+    return isApiPath ? apiFallback : fallback;
```
dealako, line 170: [nit] Three optional polish items on the fail-closed API fallback — none blocks approval

1. **`apiFallback` duplicates the `DEFAULT_ROUTE_CONFIG` `/api` row.** The literal `{ pattern: '/api', type: 'api', auth: 'required', tokenRequired: true }` is field-identical to the protected-API row in the route table. If that row's policy ever changes, this branch silently keeps the old one with no compile-time link. A one-line cross-reference comment (or deriving it from `config.routes`) removes the drift risk.

2. **Test gap on the encoded-separator branch.** The two new tests cover the `%ZZ` (malformed-escape → `decodeURIComponent` throws) path, but the `catch` also fires for the explicit `throw new Error('encoded path separator')` on a `%2F`-decoded segment. A third case — `GET /api/foo/bar%2Fbaz` asserting no `oidc.login` + `next(AuthenticationError)` — would pin that throw site too.

3. **`isApiPath` reads the raw prefix**, so a pathological encoded-`api` prefix (`/%61pi/foo/%ZZ`) with a later decode failure falls to the SSR `fallback` (HTML redirect) rather than `apiFallback`. It stays fail-closed (`auth: 'required'`), and no real XHR client encodes the literal `api` prefix, so this is cosmetic — worth at most a comment noting the degrade.

All three are safe to defer (LFXV2-3254 is a natural home). Approving as-is.

dealako (approved):
Approved — this one's ready, @fayazg. 🎉

Over five rounds you've turned an already-solid feature into a genuinely hardened one, and this final pair of commits closes the last thread cleanly.

### 👏 What landed this round
- **M1 resolved** — the decode-failure branch now returns `apiFallback` (`type: 'api'`, `auth: 'required'`, `tokenRequired: true`) for raw `/api` and `/public/api` prefixes, so a malformed/encoded API path yields a **JSON 401 via `next(AuthenticationError)`** (never an HTML login redirect an XHR client can't follow) and still triggers bearer extraction. `isApiPath` is correctly computed on the raw path (decoding has already failed), and the two regression tests pin the exact observable contract (`oidc.login` not called + `next(AuthenticationError)`).
- Fail-closed verified on every vector I traced — mixed-case `%2f`, double-encoding `%252F`, `/apixyz`, `/public/api%2F`, null byte, and even an encoded `/%61pi` prefix — no auth bypass or downgrade. Extending the fail-closed API shape to `/public/api` (optional → required on malformed) is strictly *more* restrictive, which is the right call.

### Full-arc revision status
- ✅ **Resolved:** N1 (groups auth invariant), N2 (mobile-viewport e2e), N3 (matrix-param + per-segment decode + fail-closed), the round-3 chrome nits (F4/F5), the duplicate-testid (Cursor), and **M1** (API 401 shape).
- 📋 **Deferred (tracked):** F6 → LFXV2-3254.

### Issue count (open)
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 3 — all take-it-or-leave-it (see the single inline). None blocks; fold in if convenient or leave for LFXV2-3254.

### Bot reconciliation
No new bot findings this round — only author replies confirming the fix. Prior Copilot (API-type) and Cursor (duplicate testid) threads are both addressed.

### Decision
✅ **Approved.** Clean fail-closed auth boundary, real 404 semantics, and solid regression coverage. Nice work seeing it all the way through.

audigregorie (commented):
## Code Review Summary

Well-executed, security-conscious change with comprehensive test coverage and no blocking findings. The per-segment percent-decoding with fail-closed behavior closes real fail-open vectors, the route ordering is correct, and the SSR flag plumbing is properly gated. Rated 5/5.

### Non-blocking observations (no line threads needed)

**Minor — `/meetings/` (trailing slash) behavior change.** The old `startsWith('/meetings/')` matched `/meetings/` as optional-auth; the new anchored regex does not (no segment after the slash), so it falls through to `required`. This is arguably more correct (`/meetings` list is auth-guarded in Angular and avoids a client-side login-redirect flash), but it's an undocumented edge-case change. Very low impact since `/meetings` without a trailing slash already required auth.

**Minor — `server.ts` response rebuild forwards all original 200 headers.** `new globalThis.Response(body, { status: 404, ..., headers: response.headers })` carries the original `Content-Length` to the 404. In practice this is safe: the body is byte-identical, Angular SSR doesn't add `Content-Encoding`, and the Express `compression` middleware recalculates `Content-Length` when it compresses. A defensive `headers.delete('content-length')` before re-wrapping would make the intent explicit, but it's not required for correctness.

### What's done well

- **Fail-closed security posture on percent-encoding.** Per-segment `decodeURIComponent` with throw-on-encoded-separator (`%2F`) and throw-on-malformed-escape (`%ZZ`) closes real fail-open vectors where Express's raw `req.path` would classify an encoded segment as public while Angular decodes it to a protected route. The `/u%2Fsomeone` and `/meetings/%63reate` test cases lock this in.
- **Strong test coverage at three layers** — auth middleware unit tests (12 new cases), `NotFoundComponent` SSR flag mutation tests, and e2e tests for both the in-shell 404 and public-profile 404 (status + URL preservation + shell visibility).
- **Correct interface extraction** — `ServerRequestContext` lives in `packages/shared/src/interfaces/` per repo rules, with a docstring explaining the mutable shared-reference contract.
- **Clean removal of `/not-found`** with no dangling redirects/links; `/meetings/not-found`, `/groups/not-found`, `/docs/not-found` are separate and untouched.
- **Thoughtful route ordering** — trailing empty-path catch-all is last so it never shadows public siblings; `authGuard` on it correctly gates anonymous deep-links to login before the in-shell 404.
- **SSR-safe flag plumbing** — both `NotFoundComponent` (constructor, `isPlatformServer`-gated) and `PublicProfilePageComponent` (`tap`, `isPlatformServer`-gated) only mutate `reqContext.notFound` on the server and handle a null context without throwing.

audigregorie (approved):

PR #1515, opened 2026-08-13
refactor(org-lens): remove the dead Org Lens access cache (LFXV2-3250)

luismoriguerra:
Follow-up to [#1474](https://github.com/linuxfoundation/lfx-self-serve/pull/1474), which merged while this was still in the review loop. Purely subtractive — no behaviour change.

## What was dead, and why

#1474 replaced the All Employees access-badge join with a server-attributed `accessBadge` on each roster row. That removed the only **reader** of the shared Org Lens access cache, leaving two pieces of dead weight behind.

**The prefetch.** All Employees still called `ensureLoaded` on every org change, under a comment claiming it made the Access tab instant. It doesn't: that tab fetches its own data via `getAccessUsers` and renders from its own `listData`. The prefetch was an HTTP request per org switch feeding nothing.

**The service itself.** `OrgLensAccessStateService` was left write-only — `rosterForOrg` and `accessByEmailFor` had zero callers, and the four `setRoster` calls in the Access tab populated a map nobody read. Its own doc comment described it as shared between *"the Access tab (writer) and the All Employees tab (reader)"*, and the reader was gone.

## Changes

- Remove the `ensureLoaded` prefetch and the unused service injection from All Employees.
- Delete `OrgLensAccessStateService`.
- Remove the four `setRoster` calls from the Access tab, each of which carried a comment about keeping the All Employees access cell fresh without a duplicate fetch — no longer something that happens.

Net: 100 lines removed, one file deleted.

## Why it's safe

The Access tab never read this cache. It fetches through `getAccessUsers` and renders from `listData`; the `setRoster` calls were mirror-writes for a consumer that no longer exists. Nothing else in the app referenced the service — verified by grep across the whole submodule, including templates.

Credit to Bugbot, which flagged the leftover prefetch on #1474; verifying that claim is what surfaced the wider dead service.

## Test plan

- [x] `yarn format:check`, `yarn lint:check`, `yarn test`, `yarn build` — all green by exit code
- [x] 79 test files pass, unchanged from before
- [ ] Smoke: People → Org Lens Access still lists, invites, edits and removes; All Employees badges still render (they now come from the roster payload)


dealako (approved):
Hey @luismoriguerra 👋 — reviewed this one end-to-end.

**Overall impression:** Textbook subtractive refactor. #1474 turned `OrgLensAccessStateService` write-only by moving badges onto the server-attributed `accessBadge`, and this PR cleanly removes the three pieces of dead weight that were left behind: the no-op `ensureLoaded` prefetch, the four mirror-only `setRoster` writes, and the service itself. The commit messages and PR description do an excellent job of explaining *why* each deletion is safe rather than just *what* changed. No behaviour change, and the reasoning holds up under verification.

**Verification I ran:**
- **No lingering references** — grepped `OrgLensAccessStateService`, `org-lens-access-state`, `rosterForOrg`, `accessByEmailFor`, `setRoster`, and `ensureLoaded` across `apps/` and `packages/` (`.ts` + `.html`). The only hits are unrelated (`docs-search.service.ts` has its own private `ensureLoaded`; `org-project-detail` has `resetRoster`). The service had **no reader** and **no spec file**, so nothing dangles.
- **No orphaned imports** — `EMPTY` was the only newly-unused import and it's correctly dropped from `all-employees.component.ts`. Every other rxjs operator on that import line still has real usages. `EMPTY_ORG_ACCESS_LIST_RESPONSE` in the Access tab is still used (signal init + two `of(...)` fallbacks) — only its comment reference at the removed `setRoster` went away.
- **Behaviour preserved** — the surviving `orgUid$` reset subscription and every mid-flight `res.orgUid !== selectedAccount().uid` guard in the Access tab are untouched; only the mirror-writes were peeled off.

**Structured issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 0

One non-blocking note for the record: the description's unchecked smoke-test box (People → Org Lens Access list/invite/edit/remove + All Employees badges) is the right last mile before merge, since automated coverage can't fully assert the rendered badge path. Nothing in the diff requires a change.

**Final decision:** ✅ **Approved** — clean, well-justified, safe to merge once you've eyeballed the smoke path.

Nice work, and good catch chasing Bugbot's prefetch flag all the way to the wider dead service.

PR #1535, opened 2026-08-13
fix(campaigns): derive the deliveryType error message from the constant

mrautela365:
Follow-up to David's nit on #1450, which merged before I pushed the fix.

## What

Three error messages hardcoded `paid-marketing, email` while the code right beside each one
already had the list **derived** from the shared `CAMPAIGN_DELIVERY_TYPES` constant:

| Site | Guard | Message before |
|---|---|---|
| `campaign.controller.ts:171` | `supportedDeliveryTypes` (derived) | hardcoded |
| `campaign-proxy.service.ts:628` | `SUPPORTED_DELIVERY_TYPES` (derived) | hardcoded |
| `campaign-proxy.service.ts:881` | `SUPPORTED_DELIVERY_TYPES` (derived) | hardcoded |

An error string is the copy a reader trusts most — it is what the API actually says — so a
hardcoded tail there outlives every other duplicate. The sibling `platform` check already
interpolates its own Set for exactly this reason.

All three now interpolate. **Rendered text is byte-identical at every site.**

> I initially opened this PR fixing only the controller and describing it as "the last hardcoded
> copy". That was wrong — the reviewer sim's *fixed-at-one-layer-only* check found the other two.
> David's nit was a class, not a line.

## Test quality

The two specs covering the proxy branches asserted only `.includes('deliveryType')` — a substring
the **word alone** satisfies. They passed whether the list rendered correctly, empty, or as
`[object Object]`, so they could not fail for the thing this PR changes.

Both now assert the complete message by equality, with the expectation derived from the same
constant so a new delivery type updates it with the code rather than failing on a hand-written
list.

**Mutation-verified:** replacing the interpolation with a stale hardcoded `paid-marketing.`
fails both specs; the previous `.includes` form passed that same mutation.

## Verification

- `npx tsc --noEmit` — 0 errors (`tsconfig.app.json` and `tsconfig.spec.json`)
- `yarn test` — 1291 server + 317 app passing

### DIFF
--- a/apps/lfx-one/src/server/controllers/campaign.controller.ts
+++ b/apps/lfx-one/src/server/controllers/campaign.controller.ts
@@ -163,10 +163,10 @@ export class CampaignController {
     // message, for a caller whose only mistake was a misspelling.
     //
     // Derived from the shared `CAMPAIGN_DELIVERY_TYPES`, and so is the MESSAGE below — the sibling
-    // `platform` check already interpolates its own Set for exactly this reason. The service now
-    // derives from the same constant too, so this is the single source of truth; a hardcoded tail
-    // in the error string would have been the last copy left to drift, and the one a reader trusts
-    // most because it is what the API actually says.
+    // `platform` check already interpolates its own Set for exactly this reason. An error string
+    // is the copy a reader trusts most, because it is what the API actually says, so a hardcoded
+    // tail there outlives every other duplicate. The two `Unsupported deliveryType` messages in
+    // `campaign-proxy.service.ts` are interpolated from the same constant for the same reason.
     const supportedDeliveryTypes = new Set<string>(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id));
     if (body.deliveryType !== undefined && !supportedDeliveryTypes.has(body.deliveryType)) {
       _next(
--- a/apps/lfx-one/src/server/controllers/campaign.controller.ts
+++ b/apps/lfx-one/src/server/controllers/campaign.controller.ts
@@ -162,12 +162,15 @@ export class CampaignController {
     // `'emial'` falls past it into those same paid-only checks and produces the same misleading
     // message, for a caller whose only mistake was a misspelling.
     //
-    // Derived from the shared `CAMPAIGN_DELIVERY_TYPES` rather than a fourth hardcoded list — the
-    // service already keeps its own copy, and a third one here would be the one that drifts.
+    // Derived from the shared `CAMPAIGN_DELIVERY_TYPES`, and so is the MESSAGE below — the sibling
+    // `platform` check already interpolates its own Set for exactly this reason. The service now
+    // derives from the same constant too, so this is the single source of truth; a hardcoded tail
+    // in the error string would have been the last copy left to drift, and the one a reader trusts
+    // most because it is what the API actually says.
     const supportedDeliveryTypes = new Set<string>(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id));
     if (body.deliveryType !== undefined && !supportedDeliveryTypes.has(body.deliveryType)) {
       _next(
-        ServiceValidationError.forField('deliveryType', 'deliveryType must be one of: paid-marketing, email', {
+        ServiceValidationError.forField('deliveryType', `deliveryType must be one of: ${[...supportedDeliveryTypes].join(', ')}`, {
           operation: 'campaign_refine_brief',
           service: 'campaign_controller',
           path: req.path,
--- a/apps/lfx-one/src/server/controllers/campaign.controller.ts
+++ b/apps/lfx-one/src/server/controllers/campaign.controller.ts
@@ -32,6 +32,9 @@ import { addShutdownHook, isShuttingDown } from '../utils/shutdown';
 /** Platforms that support the campaign status toggle endpoint. */
 const SUPPORTED_STATUS_PLATFORMS: ReadonlySet<CampaignPlatform> = new Set<CampaignPlatform>(['meta-ads', 'reddit-ads']);
 
+/** Derived from the shared constant so the validation and its error message cannot drift apart. */
+const SUPPORTED_DELIVERY_TYPES: ReadonlySet<string> = new Set(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id));
+
 const NUMERIC_ID_RE = /^\d+$/;
 
 export class CampaignController {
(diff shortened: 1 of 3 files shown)


mrautela365:
@dealako — following up on your nit from #1450, which merged before I pushed the fix.

Your point was broader than the line you flagged. Running the fixed-at-one-layer-only check turned up **two more** hardcoded copies in `campaign-proxy.service.ts` (lines 628 and 881), both guarding on the already-derived `SUPPORTED_DELIVERY_TYPES` and then hardcoding the list in the message beside it — template literals with nothing interpolated. All three are now interpolated; rendered text is unchanged everywhere.

I'd also written that the controller string was "the last copy left to drift". It wasn't, and I've corrected the comment and the PR body.

One thing worth flagging: the two specs covering those proxy branches asserted `.includes('deliveryType')`, which the *word* satisfies — they'd have passed with the list rendering empty or as `[object Object]`. They now assert the full message by equality, derived from the same constant. Mutation-verified: a stale hardcoded `paid-marketing.` fails both; the old form passed it.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -162,12 +162,15 @@ export class CampaignController {
...
     // message, for a caller whose only mistake was a misspelling.
     //
-    // Derived from the shared `CAMPAIGN_DELIVERY_TYPES` rather than a fourth hardcoded list — the
-    // service already keeps its own copy, and a third one here would be the one that drifts.
+    // Derived from the shared `CAMPAIGN_DELIVERY_TYPES`, and so is the MESSAGE below — the sibling
+    // `platform` check already interpolates its own Set for exactly this reason. An error string
+    // is the copy a reader trusts most, because it is what the API actually says, so a hardcoded
+    // tail there outlives every other duplicate. The two `Unsupported deliveryType` messages in
+    // `campaign-proxy.service.ts` are interpolated from the same constant for the same reason.
     const supportedDeliveryTypes = new Set<string>(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id));
```
dealako: [nit] Set rebuilt per request instead of module-hoisted

Issue: This `supportedDeliveryTypes` Set is derived inside the refineBrief handler, so it is reconstructed on every call.
Proof: The equivalent Set is a module const both in this same file (`SUPPORTED_STATUS_PLATFORMS`, line 33) and in the proxy (`SUPPORTED_DELIVERY_TYPES`, campaign-proxy.service.ts:505) — the very pattern your updated comment references.
Why it matters: Negligible perf (2 elements), but it's an inconsistency with the established pattern in this file and the sibling service.
Fix: Hoist to a module-level `const SUPPORTED_DELIVERY_TYPES: ReadonlySet<string> = new Set(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id));` near line 33 and reference it here.

apps/lfx-one/src/server/services/campaign-proxy.service.spec.ts
```diff
@@ -22,12 +22,24 @@ vi.mock('../helpers/url-validation', () => ({
...
+/**
+ * Derived from the same constant the service interpolates, so adding a delivery type updates this
+ * expectation with the code instead of failing on a hand-written list — a second literal here
+ * would be the drift this message was just changed to prevent.
+ *
+ * Asserted by EQUALITY rather than `.includes('deliveryType')`: the substring is satisfied by the
+ * word alone, so it passed whether the list rendered correctly, empty, or as `[object Object]` —
+ * it could not fail for the thing under test.
+ */
+const expectedUnsupportedMessage = `Unsupported deliveryType. Supported: ${CAMPAIGN_DELIVERY_TYPES.map((d) => d.id).join(', ')}.`;
```
dealako: [nit] Expected message derives from the raw array, not the deduped Set production uses

Issue: The expectation builds from `CAMPAIGN_DELIVERY_TYPES.map((d) => d.id).join(', ')` (a plain array), while production renders `[...SUPPORTED_DELIVERY_TYPES].join(', ')` where `SUPPORTED_DELIVERY_TYPES` is a Set.
Proof: campaign-proxy.service.ts:628/881 use the spread-Set form; the constant currently has unique ids so the two are byte-identical today.
Why it matters: Purely defensive — the two would only diverge if the constant ever gained a duplicate id (a bug in its own right), at which point this test could pass while production output differs.
Fix (optional): `[...new Set(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id))].join(', ')` to mirror the exact production transform.

dealako (dismissed):
Hey @mrautela365 — clean, disciplined follow-up. This does exactly what it says: three hardcoded `paid-marketing, email` tails become interpolations of the same `CAMPAIGN_DELIVERY_TYPES` constant the guards already derive their Set from, so the copy a reader trusts most can no longer drift from the validation logic. I traced the render at all three sites — `Set` preserves insertion order and the constant has no duplicate ids, so the output is byte-identical today (the untouched `campaign.controller.spec.ts:849` literal assertion confirms it). The spec hardening is the real substance here: `.includes('deliveryType')` was satisfied by the bare word and would have passed on an empty or `[object Object]` render; the equality-against-a-derived-constant form actually fails for the thing under test, and your mutation check proves it.

**Bot reconciliation:** Copilot reviewed and generated no comments; Cursor Bugbot rated it Low Risk; CodeRabbit auto-review is disabled on this repo. No conflicts with my read.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 — controller rebuilds the Set inline per request (sibling + proxy hoist to module scope); spec derives the expected string from the raw array rather than the deduped Set the code uses
- ❔ Question: 0

### Decision
✅ **Approved with minor comments** — both notes are take-it-or-leave-it; nothing blocks merge.

apps/lfx-one/src/server/controllers/campaign.controller.ts
mrautela365: Done in `3cc681178` — hoisted to a module const at line 36, immediately below `SUPPORTED_STATUS_PLATFORMS`, and both the guard and the message now read it. You're right that the inconsistency was the real cost; the rebuild itself was free.

Re-ran the mutation after the hoist to make sure it didn't quietly become non-binding: replacing the interpolated tail with a stale `paid-marketing` still fails `rejects a MISSPELLED deliveryType instead of blaming currentCopy`.

apps/lfx-one/src/server/services/campaign-proxy.service.spec.ts
mrautela365: Done in `3cc681178` — now `[...new Set(CAMPAIGN_DELIVERY_TYPES.map((d) => d.id))].join(', ')`, mirroring production's exact transform rather than only its output.

Taking this as more than optional: the failure mode is a test that passes while production renders something different, which is the specific kind of green I can't detect later. Confirmed the divergence is real rather than theoretical — with a duplicated id the array form yields `paid-marketing, email, email` and the Set form `paid-marketing, email`. Comment records why the two forms are deliberately kept identical.

dealako (approved):
Thanks @mrautela365 — `3cc6811` addresses both nits from my prior review exactly, and nothing else:

- ✅ **Controller Set hoisted** — `SUPPORTED_DELIVERY_TYPES` is now a module-level const (line 35, beside `SUPPORTED_STATUS_PLATFORMS`) and referenced in the handler instead of being rebuilt per request. Consistent with the proxy (`:505`) and the sibling platform const.
- ✅ **Spec mirrors the production transform** — the expected string now derives via `[...new Set(...)].join(', ')`, matching the exact spread-a-Set transform the code uses, with a comment documenting the defensive rationale.

Both changes are mechanical, output stays byte-identical, and there are no regressions or new concerns. Re-approving — cleaner than when I first signed off. ✅ **Approved**

PR #1546, opened 2026-08-13
feat(campaigns): build the hubspot config so email can dispatch

mrautela365:
Closes **LFXV2-3256**. Found auditing the LFXV2-2770 email epic: this is the *first* hard failure
on the email path, upstream of the audience work.

## The problem

`hasPlatformConfig` mapped four ad platforms and returned `false` for anything unmapped, so
`hubspot` was refused **client-side, before the request ever reached the Go dispatcher**.

That guard is correct and stays. `unmarshalPlatformConfig` treats an absent key as a **zero
value** rather than an error, so waving `hubspot` through would have dispatched a clone of email
id `""`. It was waiting on a config builder. This adds the builder, **then** maps it.

## What's here

**`buildHubSpotConfig`**, typed to what the dispatcher reads (`internal/dispatch/hubspot.go:47-56`):
`sourceEmailId` (required) and `utmCampaign` (optional). A blank, whitespace-only, or
**non-string** value reads as UNCONFIGURED so the existing guard refuses locally and names the
problem, rather than spending a round trip to learn the same thing after the job exists.

**A terminal refusal when the cutover is dark.** This is the part reviewers should look at
hardest, because it was missed in the first push:

Widening `platforms` to `CampaignAnyPlatform` is what made `['hubspot']` reachable — it used to be
a type error at every caller. With the flags off, `createCampaigns` returns `enabled: false` and
the controller falls through to the legacy path, which `hasPlatformConfig` **cannot see** (that
guard lives inside `createCampaigns`, gated by the same flags). The legacy path has no HubSpot
client and no `includeHubspot` arm: it records `"Unsupported platform(s)"` in an errors array and
then completes with an **empty promise list** — a job that finishes, after the inline 45s wait,
having created nothing. Now refused terminally in the controller, matching the existing
`viaService.error` arm.

## `hubspot` is deliberately NOT in `CampaignPlatform`

That union is enumerated by `CAMPAIGN_PLATFORMS`, which renders the paid **Ad Channels picker**
(`planning-tab.component.ts:96`) — widening it would offer HubSpot as an ad channel a paid brief
could select. Only `CampaignCreateRequest.platforms` widens; the planning/refine requests keep the
narrow type, since they drive ad-copy generation email does not use.

⚠️ One consequence is documented rather than fixed: the restored-brief filter
(`campaign-service.service.ts:1387`) narrows to `CampaignPlatform`, so a stored email brief's
`hubspot` would read as *unreadable*. No client sends one today, but that is a client guarantee
and this is a server — deferred to **LFXV2-3224**, which should solve it deliberately rather than
have this PR widen a paid-path filter.

## Review fixes (all verified against the Go source before acting)

| Finding | Outcome |
|---|---|
| Dark cutover reaches the legacy path | **Fixed** — terminal refusal + tests |
| Non-string `sourceEmailId` → `TypeError` → 500 | **Fixed** — `typeof` checks; this route has no body validator |
| My claim that `''` suppresses the upstream `utmCampaign` default | **I was wrong.** `utm.Resolve` (`resolve.go:47-60`) trims and falls through, so `''` and absent resolve identically. Behaviour kept as canonicalization; the false rationale removed |
| `CampaignEmailPlatform` — exported one-member alias, no importer | **Dropped**, inlined into `CampaignAnyPlatform` |

## Tests

14 new, each **mutation-verified** — reverting a fix fails exactly the test written for it:

| Mutation | Test that fails |
|---|---|
| drop the `hubspot` guard mapping | `dispatches an email campaign once hubspotConfig is present` |
| drop the dark-cutover refusal | `refuses an email create instead of falling through to the legacy path` |
| `(rawId as string)?.trim()` | `refuses a numeric/object sourceEmailId without throwing` |
| drop the `.trim()` | `treats a whitespace only sourceEmailId as unconfigured…` |
| forward a blank `utmCampaign` | `omits a blank utmCampaign rather than sending an empty override` |

The first round's controller tests all ran with the cutover **dark** and asserted only on the
argument handed to a mocked `createCampaigns` — so none exercised the path this ticket ships.
Added both directions, plus a paid contrast so the refusal cannot pass by refusing everything.

## Verification

- `npx tsc --noEmit` — 0 errors (`tsconfig.app.json` + `tsconfig.spec.json`)
- `yarn test` — **1316** server + **320** app passing

## Scope

Necessary but **not sufficient** to stage an email. The dispatcher also requires the brief's
audience to be **BUILT** (`hubspot.go:432-456`), resolved by `brief.ID` rather than from this
envelope — **LFXV2-3224**. The `sourceEmailId` comes from the template picker in **LFXV2-3198**.

### DIFF
--- a/apps/lfx-one/src/server/controllers/campaign.controller.ts
+++ b/apps/lfx-one/src/server/controllers/campaign.controller.ts
@@ -359,6 +359,30 @@ export class CampaignController {
         return;
       }
 
+      // The email channel exists ONLY on the cutover path, so it must not fall through here.
+      //
+      // Widening `platforms` to `CampaignAnyPlatform` is what makes this reachable: before it,
+      // `platforms: ['hubspot']` was a type error at every caller. The legacy path has no HubSpot
+      // client and no `includeHubspot` arm — it pushes "Unsupported platform(s): hubspot" into its
+      // `errors` array and then completes with an EMPTY promise list. That is the shape this
+      // cutover exists to prevent: a job that finishes, after the inline 45s wait, having created
+      // nothing, reported through a partial-success envelope rather than as a refusal.
+      //
+      // `hasPlatformConfig` cannot catch it — that guard lives inside `createCampaigns`,
+      // deliberately gated by the flags, so with the cutover dark it never runs.
+      //
+      // Refused terminally rather than passed through, matching the `viaService.error` arm above:
+      // when we know the create cannot succeed, saying so beats a 45-second wait for a result that
+      // names zero campaigns.
+      if (platforms.includes('hubspot')) {
+        logger.warning(req, 'campaign_create', 'email campaign requested while the campaign-service cutover is dark', { briefId, projectSlug });
+        res.json({
+          jobId: '',
+          error: 'Email campaigns require the campaign-service cutover to be enabled. The legacy creation path cannot stage email.',
+        });
+        return;
+      }
+
       const result = await this.proxyService.createCampaign(req, req.body);
       logger.success(req, 'campaign_create', startTime, { jobId: result.jobId, via: 'legacy' });
       res.json(result);
@@ -1176,15 +1200,30 @@ export class CampaignController {
    * `strings.TrimSpace`s it before the check, so `' '` would pass a truthiness test here and be
    * refused there — the precise split this guard exists to avoid.
    *
-   * `utmCampaign` is only forwarded when non-blank. It is optional upstream, where ABSENT means
-   * "derive one from the email name" — so forwarding `''` would not select that default, it would
-   * hand the service an empty override. Absence has to stay absence.
+   * `utmCampaign` is only forwarded when non-blank — canonicalization, not a correctness guard.
+   * An earlier version of this comment claimed a blank one would suppress the upstream default;
+   * that was wrong. `utm.Resolve` (`internal/utm/resolve.go:47-60`) trims the value and falls
+   * through to the name-derived slug when the result is empty, so `''`, `'  '` and absent all
+   * resolve identically. Omitted anyway so the envelope carries only fields that mean something,
+   * and so a reader cannot mistake an empty string for a deliberate override.
+   *
+   * This envelope is read ONLY by the cutover path. With the flags dark, `createCampaign` refuses
+   * an email create outright rather than falling through — see the guard above the legacy call.
+   * The legacy path does NOT fail loudly on `hubspot`: it records "Unsupported platform(s)" in an
+   * errors array and then completes with nothing created, which is why the refusal is explicit.
    */
   private buildHubSpotConfig(body: CampaignCreateRequest): Record<string, unknown> | null {
-    const sourceEmailId = body?.hubspotConfig?.sourceEmailId?.trim();
+    // Type-checked at runtime, not just by the `CampaignCreateRequest` cast. This route has no
+    // body validator — `req.body` is asserted, not parsed — so a caller sending
+    // `sourceEmailId: 123` reaches here as a number and `.trim()` throws a TypeError, answering a
+    // malformed request with a 500 instead of the controlled "unconfigured" refusal. A wrong TYPE
+    // is the same non-answer as a blank one, and both should take the same exit.
+    const rawId = body?.hubspotConfig?.sourceEmailId;
+    const sourceEmailId = typeof rawId === 'string' ? rawId.trim() : '';
     if (!sourceEmailId) return null;
 
-    const utmCampaign = body.hubspotConfig?.utmCampaign?.trim();
+    const rawUtm = body.hubspotConfig?.utmCampaign;
+    const utmCampaign = typeof rawUtm === 'string' ? rawUtm.trim() : '';
     return utmCampaign ? { sourceEmailId, utmCampaign } : { sourceEmailId };
   }
 }
--- a/apps/lfx-one/src/server/controllers/campaign.controller.ts
+++ b/apps/lfx-one/src/server/controllers/campaign.controller.ts
@@ -1036,6 +1036,9 @@ export class CampaignController {
     const metaConfig = this.buildMetaConfig(body);
     if (metaConfig) envelope['metaConfig'] = metaConfig;
 
+    const hubspotConfig = this.buildHubSpotConfig(body);
+    if (hubspotConfig) envelope['hubspotConfig'] = hubspotConfig;
+
     return envelope;
   }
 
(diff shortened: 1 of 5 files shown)


apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -1159,4 +1162,29 @@ export class CampaignController {
...
+   * `utmCampaign` is only forwarded when non-blank. It is optional upstream, where ABSENT means
+   * "derive one from the email name" — so forwarding `''` would not select that default, it would
+   * hand the service an empty override. Absence has to stay absence.
+   */
+  private buildHubSpotConfig(body: CampaignCreateRequest): Record<string, unknown> | null {
+    const sourceEmailId = body?.hubspotConfig?.sourceEmailId?.trim();
+    if (!sourceEmailId) return null;
+
+    const utmCampaign = body.hubspotConfig?.utmCampaign?.trim();
+    return utmCampaign ? { sourceEmailId, utmCampaign } : { sourceEmailId };
```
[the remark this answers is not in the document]
mrautela365: Correct, and fixed in `3f2adfa78`. Confirmed the mechanism before fixing — this route has no runtime body validator (`campaigns.route.ts:15` goes straight to the controller), `req.body` is asserted rather than parsed, and `(123)?.trim()` does throw `TypeError: ... is not a function`. So a malformed request answered 500 instead of taking the controlled refusal.

Both fields now go through `typeof x === 'string'` before trimming: a wrong TYPE is the same non-answer as a blank one, so both take the same exit.

Covered by four tests (numeric / null / object id, plus a non-string `utmCampaign`), each asserting `next` was NOT called — an unhandled throw surfaces through the catch as a 500, so asserting only on the envelope would have passed while the request errored. Mutation-verified: restoring `(rawId as string)?.trim()` fails the numeric and object cases.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -1159,4 +1162,29 @@ export class CampaignController {
...
+   * "No configuration was built for: hubspot", naming the actual problem. Sending `''` instead
+   * spends a round trip to learn the same thing, and the job is created before it fails.
+   *
+   * Trimmed because a whitespace-only id is the same non-answer as an absent one: upstream
+   * `strings.TrimSpace`s it before the check, so `' '` would pass a truthiness test here and be
+   * refused there — the precise split this guard exists to avoid.
+   *
+   * `utmCampaign` is only forwarded when non-blank. It is optional upstream, where ABSENT means
+   * "derive one from the email name" — so forwarding `''` would not select that default, it would
+   * hand the service an empty override. Absence has to stay absence.
```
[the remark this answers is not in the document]
mrautela365: You're right and my comment was wrong. I read `internal/utm/resolve.go:47-60` to check:

```go
if c := strings.TrimSpace(campaignUTM); c != "" {
    p.Campaign = c
    return Resolution{Params: p, Source: SourceBriefConfig}
}
if slug := Slug(emailName); slug != "" { … SourceDerived }
```

It trims and falls through, so `''`, `'  '` and absent all resolve identically — wire presence does not suppress the fallback. The behaviour was fine; the rationale I gave for it was false.

Fixed in `3f2adfa78`: the omission is now described as canonicalization (the envelope should carry only fields that mean something, and an empty string reads as a deliberate override to anyone inspecting the wire), with the correct upstream behaviour stated and cited. Updated in both places — the builder docblock and the duplicated test comment you flagged.

dealako:
Hi @mrautela365 👋 — reviewed the full diff across the controller, service, shared interface, and both spec files.

## Overall impression

This is a tightly-scoped, well-reasoned change: it adds `buildHubSpotConfig`, maps `hubspot → hubspotConfig` in `hasPlatformConfig`, and — the load-bearing part — refuses email creates terminally when the cutover is dark instead of letting `['hubspot']` fall through the legacy path to a job that finishes having created nothing. The type widening is correctly surgical (`CampaignCreateRequest.platforms` only, kept out of `CampaignPlatform`/`CAMPAIGN_PLATFORMS` so HubSpot isn't offered as a paid ad channel), the runtime `typeof` guards close the real 500 vector Copilot flagged, and the 14 tests are genuinely mutation-oriented. The self-corrections in the PR body (the `utm.Resolve` rationale reversal, dropping the one-member `CampaignEmailPlatform` alias) landed correctly. Clean work.

## Bot reconciliation

Both Copilot findings — the non-string `sourceEmailId` `TypeError`→500 and the inaccurate `utm.Resolve` rationale — were **already addressed** in `3f2adfa78` (verified against the current code), and Copilot's latest re-review generated no new comments. Agreed and resolved; no duplication here.

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 — comments narrate their own revision history; line-precise cross-repo `.go` citations will drift silently
- ❔ Question: 1 — a mixed `['google-ads','hubspot']` array under a dark cutover is refused wholesale (likely UI-unreachable today, raised as a scope confirmation)

## Final decision

✅ **Approved with minor comments** — nothing blocks merge. The one question is a reachability confirmation, and the two nits are take-it-or-leave-it. Details inline.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -364,6 +364,30 @@ export class CampaignController {
...
+      // cutover exists to prevent: a job that finishes, after the inline 45s wait, having created
+      // nothing, reported through a partial-success envelope rather than as a refusal.
+      //
+      // `hasPlatformConfig` cannot catch it — that guard lives inside `createCampaigns`,
+      // deliberately gated by the flags, so with the cutover dark it never runs.
+      //
+      // Refused terminally rather than passed through, matching the `viaService.error` arm above:
+      // when we know the create cannot succeed, saying so beats a 45-second wait for a result that
+      // names zero campaigns.
+      if (platforms.includes('hubspot')) {
```
dealako, line 382: [question] Mixed hubspot+paid array is refused wholesale under a dark cutover

Issue: `if (platforms.includes('hubspot'))` fires regardless of what else is in the array. With the cutover dark, `platforms: ['google-ads','hubspot']` is refused wholesale with the email-cutover message, so the Google create the legacy path *can* serve is never attempted.

Proof: with cutover dark `createCampaigns` returns `{enabled:false}`, so both `viaService` arms are skipped and control reaches this guard, which returns before `this.proxyService.createCampaign`. No test exercises a mixed array — spec only sends `['hubspot']` or `googleBody()`.

Why it matters: if such a request can reach this route, a servable paid create is silently dropped and the user is told to enable a control they don't have. In practice this looks UNREACHABLE from the UI — `hubspot` is absent from `CAMPAIGN_PLATFORMS`, so `planning-tab` can't co-select it — so this is a scope confirmation, not a demonstrated defect.

Fix: confirm no create path can carry `hubspot` alongside a paid platform. If that's a server-side invariant, a one-line comment here stating it is enough; otherwise consider refusing only the email selection while still routing co-selected paid platforms to legacy.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -1164,4 +1191,44 @@ export class CampaignController {
...
+   * end in a refusal; the difference is where. Null makes `hasPlatformConfig` refuse locally with
+   * "No configuration was built for: hubspot", naming the actual problem. Sending `''` instead
+   * spends a round trip to learn the same thing, and the job is created before it fails.
+   *
+   * Trimmed because a whitespace-only id is the same non-answer as an absent one: upstream
+   * `strings.TrimSpace`s it before the check, so `' '` would pass a truthiness test here and be
+   * refused there — the precise split this guard exists to avoid.
+   *
+   * `utmCampaign` is only forwarded when non-blank — canonicalization, not a correctness guard.
+   * An earlier version of this comment claimed a blank one would suppress the upstream default;
```
dealako, line 1209: [nit] Comment narrates its own revision history

Issue: this block spends a paragraph correcting a prior version of itself ("An earlier version of this comment claimed a blank one would suppress the upstream default; that was wrong."), mirroring the same pattern at line 1092.

Why it matters: git history already records the correction. Revision-history narration is noise that buries the one fact a reader needs — that `utmCampaign` is omitted when blank purely for canonicalization.

Fix: state the current rule directly ("utmCampaign is omitted when blank; upstream utm.Resolve derives an equivalent slug either way") and drop the self-referential clauses. Take-it-or-leave-it.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -1164,4 +1191,44 @@ export class CampaignController {
     const { budgetUsd, ...rest } = body.metaConfig;
     return { ...rest, budget: budgetUsd };
   }
+
+  /**
+   * The email channel's config.
+   *
+   * A BLANK `sourceEmailId` returns null — i.e. UNCONFIGURED — rather than an object carrying an
+   * empty string. Upstream requires it (`hubspot.go:281-283` refuses a blank one), so both paths
```
dealako, line 1199: [nit] Line-precise cross-repo citations will drift silently

Issue: new comments pin upstream Go source by exact line number (`hubspot.go:281-283`, `hubspot.go:47-56`, `resolve.go:47-60`, `hubspot.go:432-456`) plus in-repo `planning-tab.component.ts:96`.

Why it matters: a stale but authoritative-looking line citation is worse than none — a future reader trusts `hubspot.go:281-283` and reads whatever now sits there. Cross-repo refs can't be validated by this repo's CI. (This extends a pre-existing pattern, so it's consistent with the file; the surface is just larger now.)

Fix: cite by stable symbol/function name (`hubspot.go` blank-id dispatch guard, `utm.Resolve`) rather than line ranges, or pin to an upstream commit/tag. Optional.

dealako (approved):
Approved with minor comments. No blocking or security findings; branch is mergeable. Copilot's two prior findings were already resolved in `3f2adfa78`. Three non-blocking items inline: one reachability question and two documentation nits.

PR #1548, opened 2026-08-14
fix(campaigns): treat an unreachable service as definitely-not-created

mrautela365:
Closes **LFXV2-3258**.

Observed during live cutover testing on 2026-08-13, with campaign-service simply stopped.

## The problem

A create that **never left this process** was reported as:

> Campaign creation could not be confirmed. It may have started — check the ad platforms before
> retrying.

Nothing had started. The connection was refused; no request was made. The user is sent to audit an
ad account for a campaign that cannot exist, and invited to retry a create that has **no
idempotency key** — the exact harm the surrounding comment says the branch exists to prevent.

## Why the existing guard missed it

`definitelyRejected` tests `error instanceof MicroserviceError`. A connection failure never becomes
one: `microservice-proxy.service.ts:38-40` only wraps errors carrying `.status` and `.code`, and
re-throws the rest raw. Verified — an `ECONNREFUSED` arrives as a plain `Error` with
`status: undefined`.

So the predicate reads `false` and falls straight through to the indeterminate branch. The comment
above it reasons carefully about 4xx vs 5xx vs 408 and never considers *"the connection was never
established"*.

## The fix

`requestNeverLeft()`, keyed on Node's documented system-error `code` — `ECONNREFUSED`, `ENOTFOUND`,
`EAI_AGAIN`, `ECONNRESET`, `EHOSTUNREACH`, `ENETUNREACH`.

Applied at **both** sites that classify a failed write:

- **the create path** — now answers *"Could not reach the campaign service, so nothing was created.
  Please try again."*
- **`reconcileLostWrite`** — skips reads that could only ever prove a negative

One shared helper rather than a third copy of a predicate already duplicated three times.

Deliberately **not** keyed on message text, which is not a contract. An unrecognised code stays
indeterminate: this widens what counts as *definite*, and a wrong guess in that direction is the
dangerous one.

## Tests

5 new, **mutation-verified** — removing the guard fails all four transport cases.

The fifth is the load-bearing one: a **5xx must keep** the *"may have started"* wording, because
that request **did** reach the service and the outcome genuinely is unknown. Without it, this fix
could have been satisfied by making every failure read as definite — which would invite exactly the
duplicate-create retry the cutover exists to prevent.

- `npx tsc --noEmit` — 0 errors (app + spec)
- `yarn test` — **1306** server + **321** app passing

## Related

Surfaced alongside two campaign-service blockers fixed in
[lfx-v2-campaign-service#129](https://github.com/linuxfoundation/lfx-v2-campaign-service/pull/129)
(**LFXV2-3259**). Separate repos, so separate PRs; neither depends on the other to merge.

This ticket has three observed cases — a missing connection, a missing event name, and this
unreachable-service one. The first two are now definite pre-create failures upstream; this PR fixes
the third and stops the wording being applied to a request that was never sent.

### DIFF
--- a/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
+++ b/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
@@ -1609,7 +1609,7 @@ describe('CampaignServiceClient.createCampaigns', () => {
    * Asserts the message does NOT tell the user to go check the ad platforms. Asserting only that
    * some error came back would pass on the old wording.
    */
-  it.each([['ECONNREFUSED'], ['ENOTFOUND'], ['EAI_AGAIN'], ['ECONNRESET']])(
+  it.each([['ECONNREFUSED'], ['ENOTFOUND'], ['EAI_AGAIN'], ['EHOSTUNREACH']])(
     'reports a %s create as definitely-not-created rather than unconfirmed',
     async (code) => {
       bothFlagsOn();
@@ -1633,6 +1633,26 @@ describe('CampaignServiceClient.createCampaigns', () => {
    * definite, which is the dangerous direction: it would invite exactly the duplicate-create
    * retry the cutover exists to prevent.
    */
+  /**
+   * `ECONNRESET` must stay INDETERMINATE, even though it is a transport error. Node reports it
+   * for a reset at any point, and nothing here distinguishes a connect-time reset from one that
+   * arrives after the request was sent and processed — where the create may have landed and only
+   * the reply was lost.
+   *
+   * This is the direction that costs money: on a path with no idempotency key, telling the user
+   * "nothing was created" invites the retry that duplicates a paid campaign. An earlier revision
+   * of this fix had ECONNRESET in the definite set; the sibling approve-path test caught it.
+   */
+  it('keeps an ECONNRESET create unconfirmed, because the reset may have followed a commit', async () => {
+    bothFlagsOn();
+    proxyRequestWithResponse.mockRejectedValueOnce(Object.assign(new Error('socket hang up'), { code: 'ECONNRESET' }));
+
+    const res = await new CampaignServiceClient().createCampaigns(req, 'b-1', 'tlf', ['linkedin-ads'], { linkedInConfig: { budgetUsd: 100 } });
+
+    expect(res.error).toContain('could not be confirmed');
+    expect(res.error).not.toContain('nothing was created');
+  });
+
   it('still reports a 5xx create as unconfirmed, because the request did reach the service', async () => {
     bothFlagsOn();
     proxyRequestWithResponse.mockRejectedValueOnce(new MicroserviceError('upstream boom', 502, 'INTERNAL_ERROR', { operation: 'create' }));
--- a/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
+++ b/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
@@ -1613,7 +1613,16 @@ describe('CampaignServiceClient.createCampaigns', () => {
     'reports a %s create as definitely-not-created rather than unconfirmed',
     async (code) => {
       bothFlagsOn();
-      const transportError = Object.assign(new Error('connect failed'), { code });
+      // The PRODUCTION shape, not a raw Error: `ApiClientService.executeRequest` wraps a Node
+      // fetch failure as `MicroserviceError(500, cause.code)` before this service sees it
+      // (`api-client.service.ts:313-320`). An earlier revision of this test rejected with a raw
+      // `Error` carrying a top-level `code` — a shape this client never throws — so it passed
+      // against a `requestNeverLeft` that returned false for every MicroserviceError and fixed
+      // nothing in production.
+      const transportError = new MicroserviceError('Request failed: fetch failed', 500, code, {
+        operation: 'api_client_network_error',
+        service: 'api_client_service',
+      });
       proxyRequestWithResponse.mockRejectedValueOnce(transportError);
 
       const res = await new CampaignServiceClient().createCampaigns(req, 'b-1', 'tlf', ['linkedin-ads'], { linkedInConfig: { budgetUsd: 100 } });
@@ -1645,7 +1654,9 @@ describe('CampaignServiceClient.createCampaigns', () => {
    */
   it('keeps an ECONNRESET create unconfirmed, because the reset may have followed a commit', async () => {
     bothFlagsOn();
-    proxyRequestWithResponse.mockRejectedValueOnce(Object.assign(new Error('socket hang up'), { code: 'ECONNRESET' }));
+    proxyRequestWithResponse.mockRejectedValueOnce(
+      new MicroserviceError('Request failed: socket hang up', 500, 'ECONNRESET', { service: 'api_client_service' })
+    );
 
     const res = await new CampaignServiceClient().createCampaigns(req, 'b-1', 'tlf', ['linkedin-ads'], { linkedInConfig: { budgetUsd: 100 } });
 
--- a/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
+++ b/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
@@ -1598,6 +1598,51 @@ describe('CampaignServiceClient.createCampaigns', () => {
   });
 
   /**
+   * A transport failure means the bytes never left this process, so nothing upstream can have
+   * started — as definite as a 4xx, and safer to retry than one.
+   *
+   * Observed 2026-08-13 with campaign-service stopped: the create answered "could not be
+   * confirmed — check the ad platforms before retrying" for a request that was never sent. The
+   * `definitelyRejected` predicate could not catch it, because a connection failure is not a
+   * `MicroserviceError` at all — the proxy only wraps errors carrying `.status` and `.code`.
+   *
+   * Asserts the message does NOT tell the user to go check the ad platforms. Asserting only that
+   * some error came back would pass on the old wording.
+   */
+  it.each([['ECONNREFUSED'], ['ENOTFOUND'], ['EAI_AGAIN'], ['ECONNRESET']])(
+    'reports a %s create as definitely-not-created rather than unconfirmed',
+    async (code) => {
+      bothFlagsOn();
+      const transportError = Object.assign(new Error('connect failed'), { code });
+      proxyRequestWithResponse.mockRejectedValueOnce(transportError);
+
+      const res = await new CampaignServiceClient().createCampaigns(req, 'b-1', 'tlf', ['linkedin-ads'], { linkedInConfig: { budgetUsd: 100 } });
+
+      expect(res.enabled).toBe(true);
+      expect(res.jobId).toBeNull();
+      expect(res.error).toContain('nothing was created');
+      expect(res.error).not.toContain('may have started');
+      expect(res.error).not.toContain('check the ad platforms');
+    }
+  );
+
+  /**
+   * The contrast, and the load-bearing half: a 5xx IS genuinely indeterminate — the request
+   * reached campaign-service and the outcome is unknown — so it must keep the "may have started"
+   * wording. Without this test the fix above could be satisfied by making every failure read as
+   * definite, which is the dangerous direction: it would invite exactly the duplicate-create
+   * retry the cutover exists to prevent.
+   */
+  it('still reports a 5xx create as unconfirmed, because the request did reach the service', async () => {
+    bothFlagsOn();
+    proxyRequestWithResponse.mockRejectedValueOnce(new MicroserviceError('upstream boom', 502, 'INTERNAL_ERROR', { operation: 'create' }));
+
+    const res = await new CampaignServiceClient().createCampaigns(req, 'b-1', 'tlf', ['linkedin-ads'], { linkedInConfig: { budgetUsd: 100 } });
+
+    expect(res.error).toContain('could not be confirmed');
+  });
+
+  /**
    * campaign-service has NO Demand Gen path — `internal/dispatch/googleads.go` creates a Search
    * campaign and nothing else. The legacy path does support it, so this is a capability the
    * cutover loses rather than one nobody has.
--- a/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
+++ b/apps/lfx-one/src/server/services/campaign-service.service.spec.ts
@@ -875,6 +875,22 @@ describe('CampaignServiceClient.saveBrief', () => {
     expect(proxyRequestWithResponse).toHaveBeenCalledTimes(2);
   });
 
+  it('does not reconcile a save whose request NEVER LEFT this process', async () => {
+    // The transport twin of the 4xx case above, and the one `requestNeverLeft` guards inside
+    // `reconcileLostWrite`. A connect-time failure means the POST never reached the service, so
+    // there is no lost write to find — spending reconciliation reads on it can only surface
+    // someone else's row, and the delay between attempts makes the user wait to be told nothing.
+    //
+    // Every other transport test drives `createCampaigns`, which returns early at its own
+    // `requestNeverLeft` guard and never reaches `reconcileLostWrite`, so this arm was
+    // unexercised: deleting its `return null;` left the suite green.
+    proxyRequestWithResponse.mockRejectedValueOnce(NOT_FOUND).mockRejectedValueOnce(new MicroserviceError('connect ECONNREFUSED', 500, 'ECONNREFUSED', {}));
+
+    await expect(new CampaignServiceClient().saveBrief(req, briefWithSlug('e'), 'e', 'tlf')).rejects.toThrow('ECONNREFUSED');
+    // find + POST only: no reconciliation read, exactly as for the 4xx refusal.
+    expect(proxyRequestWithResponse).toHaveBeenCalledTimes(2);
+  });
+
   it('refuses a replace when the caller cannot say which version it last saw', async () => {
     // Two reasons produce a missing validator and they need opposite treatment. This is the
     // UNKNOWN one: the caller's previous write returned no ETag, or its approval outcome was
@@ -1609,7 +1625,7 @@ describe('CampaignServiceClient.createCampaigns', () => {
    * Asserts the message does NOT tell the user to go check the ad platforms. Asserting only that
    * some error came back would pass on the old wording.
    */
-  it.each([['ECONNREFUSED'], ['ENOTFOUND'], ['EAI_AGAIN'], ['EHOSTUNREACH']])(
+  it.each([['ECONNREFUSED'], ['ENOTFOUND'], ['EAI_AGAIN'], ['EHOSTUNREACH'], ['ENETUNREACH']])(
     'reports a %s create as definitely-not-created rather than unconfirmed',
     async (code) => {
       bothFlagsOn();
@@ -1636,13 +1652,6 @@ describe('CampaignServiceClient.createCampaigns', () => {
   );
 
   /**
-   * The contrast, and the load-bearing half: a 5xx IS genuinely indeterminate — the request
-   * reached campaign-service and the outcome is unknown — so it must keep the "may have started"
-   * wording. Without this test the fix above could be satisfied by making every failure read as
-   * definite, which is the dangerous direction: it would invite exactly the duplicate-create
-   * retry the cutover exists to prevent.
-   */
-  /**
    * `ECONNRESET` must stay INDETERMINATE, even though it is a transport error. Node reports it
    * for a reset at any point, and nothing here distinguishes a connect-time reset from one that
    * arrives after the request was sent and processed — where the create may have landed and only
@@ -1664,6 +1673,13 @@ describe('CampaignServiceClient.createCampaigns', () => {
     expect(res.error).not.toContain('nothing was created');
   });
 
+  /**
+   * The contrast, and the load-bearing half: a 5xx IS genuinely indeterminate — the request
+   * reached campaign-service and the outcome is unknown — so it must keep the "may have started"
+   * wording. Without this test the fix above could be satisfied by making every failure read as
+   * definite, which is the dangerous direction: it would invite exactly the duplicate-create
+   * retry the cutover exists to prevent.
+   */
   it('still reports a 5xx create as unconfirmed, because the request did reach the service', async () => {
     bothFlagsOn();
     proxyRequestWithResponse.mockRejectedValueOnce(new MicroserviceError('upstream boom', 502, 'INTERNAL_ERROR', { operation: 'create' }));
--- a/apps/lfx-one/src/server/services/campaign-service.service.ts
+++ b/apps/lfx-one/src/server/services/campaign-service.service.ts
@@ -184,9 +184,16 @@ function hasPlatformConfig(platform: string, envelope: Record<string, unknown>):
 /**
  * Did the request definitively never reach campaign-service?
  *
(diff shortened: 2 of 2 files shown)


apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -181,6 +181,35 @@ function hasPlatformConfig(platform: string, envelope: Record<string, unknown>):
...
+ * Node.js system-error field, and an unrecognised code stays indeterminate — this widens what
+ * counts as definite, and a wrong guess in that direction is the dangerous one.
+ */
+const NEVER_SENT_ERROR_CODES: ReadonlySet<string> = new Set(['ECONNREFUSED', 'ENOTFOUND', 'EAI_AGAIN', 'ECONNRESET', 'EHOSTUNREACH', 'ENETUNREACH']);
+
+function requestNeverLeft(error: unknown): boolean {
+  if (error instanceof MicroserviceError) return false; // it got a response; not a transport failure
+  const code = (error as { code?: unknown } | null)?.code;
+  return typeof code === 'string' && NEVER_SENT_ERROR_CODES.has(code);
+}
```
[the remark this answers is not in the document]
mrautela365: You're right, and this was the important catch — the fix was inert. Confirmed at `api-client.service.ts:313-320`: a Node fetch failure is wrapped as `MicroserviceError(500, cause.code)`, so the production shape of an unreachable service is a 500 whose `code` is `ECONNREFUSED`. My `instanceof MicroserviceError` early return answered `false` for exactly the case the helper exists to catch.

The tests hid it because they rejected with a raw `Error` carrying a top-level `code` — a shape this client never throws. They could not fail for the real case, which is the worse half of the bug.

Fixed in `4ffe3e67c`: key on the CODE rather than the class. A genuine 500 from campaign-service carries an HTTP-ish code (`INTERNAL_ERROR`), never a syscall name, so the two stay apart — the 5xx contrast test uses `INTERNAL_ERROR` and still passes. Tests now reject with the production-shaped `MicroserviceError`, and mutation-verified: restoring the early return fails all four transport cases, which it did not before.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -181,6 +181,42 @@ function hasPlatformConfig(platform: string, envelope: Record<string, unknown>):
...
+ * rule out a campaign that could not exist.
+ *
+ * Deliberately NOT keyed on the message text, which is not a contract. `code` is the documented
+ * Node.js system-error field, and an unrecognised code stays indeterminate — this widens what
+ * counts as definite, and a wrong guess in that direction is the dangerous one.
+ */
+const NEVER_SENT_ERROR_CODES: ReadonlySet<string> = new Set(['ECONNREFUSED', 'ENOTFOUND', 'EAI_AGAIN', 'EHOSTUNREACH', 'ENETUNREACH']);
+
+function requestNeverLeft(error: unknown): boolean {
+  if (error instanceof MicroserviceError) return false; // it got a response; not a transport failure
```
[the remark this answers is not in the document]
mrautela365: You're right, and this was the important catch — the fix was inert. Confirmed at `api-client.service.ts:313-320`: a Node fetch failure is wrapped as `MicroserviceError(500, cause.code)`, so the production shape of an unreachable service is a 500 whose `code` is `ECONNREFUSED`. My `instanceof MicroserviceError` early return answered `false` for exactly the case the helper exists to catch.

The tests hid it because they rejected with a raw `Error` carrying a top-level `code` — a shape this client never throws. They could not fail for the real case, which is the worse half of the bug.

Fixed in `4ffe3e67c`: key on the CODE rather than the class. A genuine 500 from campaign-service carries an HTTP-ish code (`INTERNAL_ERROR`), never a syscall name, so the two stay apart — the 5xx contrast test uses `INTERNAL_ERROR` and still passes. Tests now reject with the production-shaped `MicroserviceError`, and mutation-verified: restoring the early return fails all four transport cases, which it did not before.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -181,6 +181,35 @@ function hasPlatformConfig(platform: string, envelope: Record<string, unknown>):
...
+ * Observed 2026-08-13: with campaign-service stopped, a create answered "could not be confirmed —
+ * check the ad platforms before retrying" for a request that was never sent. That is the exact
+ * harm those predicates exist to prevent, told to a user who then has to go read an ad account to
+ * rule out a campaign that could not exist.
+ *
+ * Deliberately NOT keyed on the message text, which is not a contract. `code` is the documented
+ * Node.js system-error field, and an unrecognised code stays indeterminate — this widens what
+ * counts as definite, and a wrong guess in that direction is the dangerous one.
+ */
+const NEVER_SENT_ERROR_CODES: ReadonlySet<string> = new Set(['ECONNREFUSED', 'ENOTFOUND', 'EAI_AGAIN', 'ECONNRESET', 'EHOSTUNREACH', 'ENETUNREACH']);
```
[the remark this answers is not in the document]
mrautela365: Agreed, and already removed in `fbd8046cf` before this review landed — an existing test (`reports no validator when the approval outcome is unknown`) failed and made the same point: a reset can follow a commit whose reply was lost.

`ECONNRESET` is out of the set, and there is now a test pinning it as unconfirmed (`keeps an ECONNRESET create unconfirmed, because the reset may have followed a commit`), mutation-verified by putting it back. The remaining codes are connect-time only: ECONNREFUSED, ENOTFOUND, EAI_AGAIN, EHOSTUNREACH, ENETUNREACH.

I also reverted the widening of the approve path's `definitelyRejected` for the same reason — that arm runs after a successful write, so the reset is genuinely ambiguous there.

apps/lfx-one/src/server/services/campaign-service.service.spec.ts
```diff
@@ -1597,6 +1597,82 @@ describe('CampaignServiceClient.createCampaigns', () => {
...
+   *
+   * Observed 2026-08-13 with campaign-service stopped: the create answered "could not be
+   * confirmed — check the ad platforms before retrying" for a request that was never sent. The
+   * `definitelyRejected` predicate could not catch it, because a connection failure is not a
+   * `MicroserviceError` at all — the proxy only wraps errors carrying `.status` and `.code`.
+   *
+   * Asserts the message does NOT tell the user to go check the ad platforms. Asserting only that
+   * some error came back would pass on the old wording.
+   */
+  it.each([['ECONNREFUSED'], ['ENOTFOUND'], ['EAI_AGAIN'], ['EHOSTUNREACH']])(
```
dealako: [minor] `ENETUNREACH` is in the never-sent set but never asserted

Issue: `NEVER_SENT_ERROR_CODES` has five members, but this `it.each` table pins only four — `ENETUNREACH` is omitted.
Proof: `campaign-service.service.ts:212` declares `new Set(['ECONNREFUSED', 'ENOTFOUND', 'EAI_AGAIN', 'EHOSTUNREACH', 'ENETUNREACH'])`; this table iterates only the first four, and no other spec line references `ENETUNREACH`. Deleting `'ENETUNREACH'` from the set leaves the whole suite green.
Why it matters: the set is the contract for "definitely not created." An untested member can be silently dropped in a later refactor, regressing the exact unreachable-service case this PR fixes, with nothing to catch it. Copilot flagged this too.
Fix: add `['ENETUNREACH']` to the `it.each` array so all five members are mutation-protected.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -843,6 +897,12 @@ export class CampaignServiceClient {
     error: unknown,
     versionIsAcceptable: (version: number) => boolean
   ): Promise<ApiResponse<CampaignServiceBrief> | null> {
+    // A request that never left this process cannot have committed, so there is nothing to
+    // reconcile — skip the reads rather than spending them proving a negative. Same class as the
+    // `definitelyRejected` case below; see `requestNeverLeft`.
+    if (requestNeverLeft(error)) {
```
dealako, line 911: [minor] The `reconcileLostWrite` transport short-circuit has no test

Issue: `requestNeverLeft` guards two sites — the create path (`:733`) and here in `reconcileLostWrite` (`:903`). Only the create-path guard is exercised.
Proof: every new transport test calls `createCampaigns`, which returns early at `:733` and never reaches `reconcileLostWrite`. The `saveBrief` describe block already has the right pattern for the 4xx sibling (`...spec.ts:868` "does not reconcile a create the server REFUSED") but no case rejects a POST with a transport-shaped error. Removing this `return null;` leaves the suite green.
Why it matters: this branch skips reconciliation reads once a write provably never left the process; if it breaks, an unreachable `saveBrief` spends its read budget proving a negative, undetected.
Fix: add a `saveBrief` test that rejects the POST with `new MicroserviceError('...', 500, 'ECONNREFUSED', ...)` and asserts no reconciliation read is issued (e.g. `expect(proxyRequestWithResponse).toHaveBeenCalledTimes(2)` for find+POST), mirroring `:868`.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -181,6 +181,50 @@ function hasPlatformConfig(platform: string, envelope: Record<string, unknown>):
...
+ * a reset at any point, and this code cannot tell a connect-time reset from one that arrives
+ * after the request was sent and processed — where the write may well have committed and only
+ * the reply was lost. Calling that "nothing was created" on a path with no idempotency key is
+ * the one wrong answer worth avoiding, because it invites the retry that duplicates a paid
+ * campaign. The sibling approve path pins the same distinction (see its `definitelyRejected`).
+ *
+ * It needs its own check because such a failure is NOT a `MicroserviceError`: the proxy only
+ * wraps errors carrying `.status` and `.code` (`microservice-proxy.service.ts:38-40`) and
+ * re-throws everything else raw. So every `error instanceof MicroserviceError` predicate reads
+ * false for it and falls through to the indeterminate branch.
```
dealako: [minor] This JSDoc paragraph contradicts the code directly below it

Issue: the JSDoc says an unreachable-service failure is "NOT a `MicroserviceError`" and "every `error instanceof MicroserviceError` predicate reads false for it." But the inline comment at `:215-222` and the function body (`error instanceof MicroserviceError ? error.code : ...`) correctly rely on it BEING a `MicroserviceError(500, cause.code)`.
Proof: `api-client.service.ts:313-320` wraps the fetch failure as `MicroserviceError(500, cause.code)` before the proxy re-throws it; the spec fixture at `...spec.ts:1614` constructs exactly that shape. So the failure reaching this catch is a `MicroserviceError` — the opposite of what this paragraph claims. This is the disproven theory the inline comment says an "earlier revision" was built on and that "fixed nothing in production."
Why it matters: a maintainer trusting the JSDoc could reintroduce the `instanceof` early-return regression this PR just removed. Copilot flagged this repeatedly.
Fix: rewrite the paragraph to state the accurate reason — the failure IS a `MicroserviceError`, but a 500 whose `code` is a syscall name that the 4xx-only `definitelyRejected` predicate doesn't catch, so this helper keys on `code` rather than the class — consistent with `:215-222`.

apps/lfx-one/src/server/services/campaign-service.service.spec.ts
```diff
@@ -1597,6 +1597,82 @@ describe('CampaignServiceClient.createCampaigns', () => {
...
+      expect(res.enabled).toBe(true);
+      expect(res.jobId).toBeNull();
+      expect(res.error).toContain('nothing was created');
+      expect(res.error).not.toContain('may have started');
+      expect(res.error).not.toContain('check the ad platforms');
+    }
+  );
+
+  /**
+   * The contrast, and the load-bearing half: a 5xx IS genuinely indeterminate — the request
```
dealako: [nit] The 5xx rationale comment is attached to the `ECONNRESET` test, not the 5xx test

Issue: this JSDoc block ("a 5xx IS genuinely indeterminate ... the load-bearing half") sits immediately above the `ECONNRESET` test at `:1655`, while the actual 5xx test at `:1667` carries no doc.
Proof: `:1639` JSDoc → `:1645` a second `ECONNRESET` JSDoc → `:1655` `it('keeps an ECONNRESET create unconfirmed ...')` → `:1667` `it('still reports a 5xx create as unconfirmed ...')` (undocumented).
Why it matters: purely doc accuracy — a reader would misattribute the 5xx guarantee to the reset assertion.
Fix: move this block down to sit directly above the 5xx test at `:1667`, leaving only the `ECONNRESET` JSDoc above the reset test.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -1113,6 +1173,13 @@ export class CampaignServiceClient {
       // is stored now may not be theirs. The distinction between "someone replaced it" and
       // "someone removed it" changes nothing they can act on.
       const removedAfterWrite = error instanceof MicroserviceError && error.statusCode === 404 && isCampaignServiceNotFound(error.errorBody);
+      // Deliberately NOT widened with `requestNeverLeft` the way the create path is. The two
```
dealako, line 1184: [question] Should the approve path treat connect-time codes as never-sent?

Context: you deliberately (and clearly) exclude `requestNeverLeft` here because this arm runs after a successful write, and an `ECONNRESET` on the follow-up approve could follow a commit whose reply was lost — genuinely indeterminate. That reasoning is sound for `ECONNRESET`.
The subtlety Copilot raised: `requestNeverLeft` already excludes `ECONNRESET`. For the connect-time codes it DOES accept (e.g. `ECONNREFUSED` on the approve request itself), the approve provably never reached the service, so treating it as unknown discards `writeEtag` and can surface an `unverified-validator` conflict on the next save (`~:1024`).
Not blocking, and the conflict path is handled — but is leaving connect-time approve failures on the indeterminate branch the intended trade-off, or worth narrowing to exclude only `ECONNRESET` while accepting the rest? A one-line note either way would close the loop for the next reader.

dealako (changes requested):
## Review — @mrautela365

Hi @mrautela365 — this is a genuinely excellent fix. The core reasoning is airtight: keying `requestNeverLeft` on the syscall `code` (not `instanceof MicroserviceError`, and not message text), matching the real `MicroserviceError(500, cause.code)` production shape, excluding `ECONNRESET` because it can follow a committed write on a non-idempotent path, and holding the line on "an unrecognised code stays indeterminate." The security posture is clean — I verified a misbehaving downstream can't spoof the classification (`fromMicroserviceResponse` sets an HTTP-derived `code`, never a syscall name), and the fix fails toward the safe (indeterminate) direction. The commit history shows real, honest iteration in response to the bots, and the two original high-severity findings (wrapped-error shape, `ECONNRESET`) are both resolved.

What's left is polish — test-coverage gaps and comment accuracy — but they sit on the load-bearing contract, so I'd like them tightened before merge.

- **Cursor / Copilot — wrapped `MicroserviceError` shape** → resolved in `4ffe3e6` (helper now keys on `code`). Agree, closed.
- **Cursor / Copilot — `ECONNRESET` misclassified** → resolved in `fbd8046` + pinned by a dedicated test. Agree, closed.
- **Copilot — `ENETUNREACH` missing from the `it.each` table** → still open; I agree (see inline).
- **Copilot — reconcile/`saveBrief` transport short-circuit untested** → still open; I agree (see inline).
- **Copilot — JSDoc at `:198` contradicts the code** → still open; I agree (see inline).
- **Copilot — approve-path `requestNeverLeft`** → raising as a `[question]`, respecting your documented trade-off (see inline).

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 3 — `ENETUNREACH` untested; reconcile transport branch untested; contradictory JSDoc on `requestNeverLeft`
- ⚪ Nit: 1 — the 5xx rationale comment sits above the `ECONNRESET` test
- ❔ Question: 1 — should the approve path treat connect-time codes as never-sent?

### Decision
🔴 **Needs changes before approval** — no blockers or security issues, but more than two minor items land on the safety-critical set/branch. All are quick fixes.

dealako:
👋 @mrautela365 — thorough review done (parallel security / correctness+tests / style+docs passes, reconciled with Cursor and Copilot).

**Overall:** a genuinely strong, well-reasoned fix. Keying `requestNeverLeft` on the syscall `code` against the real `MicroserviceError(500, cause.code)` production shape, excluding `ECONNRESET` on the non-idempotent create path, and keeping unrecognised codes indeterminate is exactly right. Security is clean — a misbehaving downstream can't spoof the classification, and it fails toward the safe (indeterminate) direction. The commit history shows honest iteration, and both original high-severity bot findings (wrapped-error shape, `ECONNRESET`) are resolved. What remains is polish on the load-bearing set/branch.

**Issue count**
- 🔴 Blocking: 0
- 🟡 Minor: 3 — `ENETUNREACH` untested in the `it.each` table · `reconcileLostWrite` transport short-circuit (`:903`) untested · JSDoc at `:198` contradicts the code below it
- ⚪ Nit: 1 — the 5xx rationale comment sits above the `ECONNRESET` test instead of the 5xx test
- ❔ Question: 1 — should the approve path treat connect-time codes (not `ECONNRESET`) as never-sent? (respecting your documented trade-off)

**Bot reconciliation:** wrapped-`MicroserviceError` shape → fixed in `4ffe3e6`; `ECONNRESET` → fixed in `fbd8046`. The three still-open Copilot items (`ENETUNREACH`, reconcile coverage, contradictory JSDoc) I independently agree with — details inline.

**Decision:** 🔴 **Needs changes before approval.** No blockers or security issues — everything above is a quick fix, and the underlying change is ready once the coverage and comment items are tightened. Nice work on this one.

apps/lfx-one/src/server/services/campaign-service.service.spec.ts
mrautela365: Right — fixed in `a6868640`. Added `ENETUNREACH` to the table, and your proof holds: deleting it from the set now fails `reports a ENETUNREACH create as definitely-not-created rather than unconfirmed` by name, where before the whole suite stayed green.

apps/lfx-one/src/server/services/campaign-service.service.ts
mrautela365: Correct, and this was the substantive one. Fixed in `a6868640`.

Added the transport twin of the 4xx test you pointed at — same shape, `find + POST only, no reconciliation read`. Mutation-verified: removing the `return null;` fails it, and takes **2012ms** to do so, which is the guard's whole point. Without it the caller sits through the reconciliation delay to be told nothing was found.

You were also right about why it was uncovered: every transport test drives `createCampaigns`, which returns early at its own `requestNeverLeft` guard and never reaches `reconcileLostWrite`.

apps/lfx-one/src/server/services/campaign-service.service.ts
mrautela365: Correct, and worth flagging how it happened: that paragraph preserved the exact mistaken model this fix corrects. An earlier revision *did* return false for every `MicroserviceError` and fixed nothing in production — the inline comment at :215 records that — and the JSDoc above it still described the world where that was true.

Rewritten in `a6868640` to say what the code does: an unreachable service and a genuine 500 both arrive as `MicroserviceError`, and only the `code` (syscall name vs HTTP-ish) tells them apart.

apps/lfx-one/src/server/services/campaign-service.service.spec.ts
mrautela365: Fixed in `a6868640` — the 5xx block now sits directly on `still reports a 5xx create as unconfirmed`, and the `ECONNRESET` test keeps its own. You were right that a reader would otherwise attribute the 5xx guarantee to the reset assertion.

apps/lfx-one/src/server/services/campaign-service.service.ts
mrautela365: Leaving this one **open** deliberately — it is a behaviour change, not a fix, and I would rather you and Misha decide than have me pick.

You are right on the mechanics. `requestNeverLeft` already excludes `ECONNRESET`, so the codes it *does* accept are all connect-time: `ECONNREFUSED`, `ENOTFOUND`, `EAI_AGAIN`, `EHOSTUNREACH`, `ENETUNREACH`. On any of those the approve provably never reached the service, so the current arm discards a `writeEtag` it could have kept and reports an outcome as unknown when it is knowable.

**Where I hesitate**, and why I have not just made the change:

The approve path's asymmetry is not only about the transport — it is that this arm runs *after a successful write*. The value at stake is not "was the approve created" but "is the ETag I am about to hand back still authoritative". Widening the arm means asserting that a connect-time failure on the approve leaves the write's ETag valid. I believe that is true — nothing mutated — but it is a claim about the *service's* state made from the *client's* transport, and the cost of being wrong is the one the whole PR exists to avoid: a caller told "saved" for a brief whose stored version has moved.

The safe direction here is the conservative one, and the current code takes it.

**If you want it changed**, the shape I would use: widen the arm to `requestNeverLeft(error)` and add a test asserting the write ETag survives an `ECONNREFUSED` approve while `ECONNRESET` still yields no validator — so the two stay visibly distinct rather than collapsing. Happy to do that in this PR or a follow-up; say which and I will.

dealako:

Thanks for the fast, precise turnaround on this. Every item from the last round is closed, and the way you closed them is exactly right.

👏 **Nice work in this revision:**
- **The substantive one — reconcile coverage — is genuinely fixed, not papered over.** The new `does not reconcile a save whose request NEVER LEFT this process` test is the transport twin of the 4xx refusal case, asserts `find + POST only` (2 calls, no reconciliation read), and you mutation-verified it (removing `return null;` fails it and eats 2s doing so — which is the guard's whole point).
- **`ENETUNREACH`** is now in the `it.each` table and mutation-fails by name, so all five members of `NEVER_SENT_ERROR_CODES` are pinned.
- **The contradictory JSDoc is rewritten to describe what the code actually does** — both shapes arrive as `MicroserviceError`, only the `code` tells them apart — so the disproven "not a MicroserviceError" model can't mislead the next maintainer into reintroducing the inert-fix regression.
- **The 5xx rationale comment now sits on the 5xx test** rather than the `ECONNRESET` one.
- **The approve-path question** you left open deliberately, with a careful writeup: the arm runs *after a successful write*, so widening it is a claim about the service's state made from the client's transport, and you (correctly) took the conservative direction. That's the right call for this PR, and the follow-up shape you sketched (widen to `requestNeverLeft` + a test that keeps `ECONNREFUSED` and `ECONNRESET` visibly distinct) is a sound way to revisit it separately if the team wants it.

**Revision status**
- ✅ Resolved — `ENETUNREACH` untested (`a6868640`)
- ✅ Resolved — `reconcileLostWrite` transport branch untested (`a6868640`)
- ✅ Resolved — contradictory `requestNeverLeft` JSDoc (`a6868640`)
- ✅ Resolved — 5xx rationale comment misplaced (`a6868640`)
- ❔ Answered — approve-path connect-time handling: deliberately deferred as a behaviour change, conservative direction retained. Not a defect.

**Bot reconciliation:** Copilot re-reviewed `a6868640` and the merge commit and generated no new comments. No open bot findings.

**Verification:** ran the server spec at HEAD `2bece3d` — `campaign-service.service.spec.ts` **146/146 passing**, including the two new tests. No merge conflicts.

**Issue count (all open items):**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

**Decision:** ✅ **Approved.** The transport-failure classification is correct, secure-by-default, well-tested, and now well-documented. Ship it.

dealako (approved):
Follow-up round: all four code findings from my prior review are resolved in `a6868640`, and the fifth (approve-path connect-time handling) is deliberately and soundly deferred. Verified `campaign-service.service.spec.ts` 146/146 passing at HEAD `2bece3d`, including the two new tests. No new issues, no open bot findings, no merge conflicts. Approving. See the summary comment for detail.

PR #1562, opened 2026-08-14
fix(public-profile): left-align "Powered by" pill on 404 view

fayazg:
## Summary

- Fix the public-profile **not-found (404)** view centering the "Powered by <LFX>" branding pill instead of left-aligning it like the loaded profile view.
- Root cause: the not-found branch wraps the topbar in a `flex flex-col` parent, which turned the `mx-auto max-w-[1240px]` topbar container into a shrink-to-content flex item — `align-items: stretch` is cancelled by the auto horizontal margins, so the container collapsed to the pill's width and `mx-auto` then centered it.
- Add `w-full` to that container so it fills the row (still capped at `max-w-[1240px]` and centered by `mx-auto`), matching the loaded view.
- Pre-existing bug on `main` (independent of the in-place 404 work in #1471); scoped to a single class change plus an explanatory comment.

Refs LFXV2-3095

### DIFF
--- a/apps/lfx-one/src/app/modules/public-profile/public-profile-page/public-profile-page.component.html
+++ b/apps/lfx-one/src/app/modules/public-profile/public-profile-page/public-profile-page.component.html
@@ -4,7 +4,9 @@
 @if (notFound()) {
   <!-- Keep the "Powered by LFX" branding header on the not-found/private view, matching the loaded profile chrome. -->
   <div class="flex min-h-screen flex-col bg-white text-gray-900">
-    <div class="mx-auto max-w-[1240px] px-4 sm:px-10">
+    <!-- w-full is required here: the wrapper is `flex flex-col`, so without it this `mx-auto` child becomes a
+         shrink-to-content flex item and the "Powered by" pill centers instead of sitting left like the loaded view. -->
+    <div class="mx-auto w-full max-w-[1240px] px-4 sm:px-10">
       <lfx-public-profile-topbar></lfx-public-profile-topbar>
     </div>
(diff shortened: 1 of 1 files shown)


audigregorie (commented):
## Code Review Summary

Clean, minimal, and correct. The root-cause analysis is right: in the not-found branch the wrapper is `flex min-h-screen flex-col`, so the `mx-auto` child becomes a shrink-to-content flex item on the cross axis (auto margins override the default `align-items: stretch`) and the "Powered by" pill centers instead of sitting left. Adding `w-full` restores the full-width base that `max-w-[1240px]` then caps and `mx-auto` centers — matching the loaded view, where the same container sits in a block-context parent and naturally fills width. No code findings.

### Minor — outside the diff

- **Commit subject missing the ticket suffix.** The branch is `fix/LFXV2-3095-topbar-powered-by-alignment` and the PR body says `Refs LFXV2-3095`, but the commit subject is `fix(public-profile): left-align "Powered by" pill on 404 view` with no `(LFXV2-3095)` suffix. Per the repo commit-workflow convention the ticket should be appended to the subject. Suggested amend: `fix(public-profile): left-align "Powered by" pill on 404 view (LFXV2-3095)`.

### What's done well

- Correct, minimal fix scoped to the single affected class — the `flex flex-col` wrapper (needed for the not-found layout) is left intact rather than refactored.
- The 2-line comment captures the non-obvious flexbox auto-margin / shrink-to-content interaction, preventing a future editor from removing `w-full` as "redundant." Respects the 1–2 line comment guidance.
- PR description is thorough and accurate: bug, root cause, fix, and the relationship to the pre-existing `main` bug and the in-place 404 work in #1471.

audigregorie (approved):

PR #1564, opened 2026-08-14
fix(profile): overlay just-saved metadata on Flow C cold return

fayazg:
## Summary

- On the Flow C (management-token re-auth) **cold return**, the replayed profile
  save can resolve before the initial profile GET populates `combinedProfile`.
  The old "no base profile yet" branch did a bare refetch, so an
  eventually-consistent (pre-save) GET body displayed with nothing to override
  it — briefly re-seeding the panel and a reopened edit drawer with the old
  value (worse for cleared free-text fields, which now send `''`).
- Stash the just-saved metadata in `pendingOptimisticMetadata` and re-apply it
  via `reapplyPendingOptimisticUpdate()` once a GET lands a base profile, giving
  cold return the same read-your-writes guarantee the warm path already has.
- The re-apply is a no-op until a real profile record lands (merging a null
  profile would fabricate one and wrongly flip the drawer's `metadataLoaded`),
  and the stash is cleared after a merge so a later GET can't re-apply a stale
  overlay.
- Add component unit tests: the cold-return re-apply (stashed save wins over the
  stale GET body) and the no-stash no-op path.

Fixes LFXV2-3267

### DIFF
--- a/apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
+++ b/apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
@@ -1,0 +1,86 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { computed, PLATFORM_ID, signal } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { ActivatedRoute, Router } from '@angular/router';
+import { PENDING_PROFILE_SAVE_KEY } from '@lfx-one/shared/constants';
+import { CombinedProfile, User } from '@lfx-one/shared/interfaces';
+import { FeatureFlagService } from '@services/feature-flag.service';
+import { UserService } from '@services/user.service';
+import { MessageService } from 'primeng/api';
+import { EMPTY, of } from 'rxjs';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { ProfileLayoutComponent } from './profile-layout.component';
+
+/**
+ * Guards the Flow C (management-token re-auth) cold-return read-your-writes fix (LFXV2-2933):
+ * on a full-page return the replayed save can resolve before the initial profile GET populates
+ * combinedProfile, so the save is stashed and re-applied once a GET lands — an eventually-consistent
+ * (pre-save) GET body must never mask the write. Template is overridden empty so the class logic runs
+ * without instantiating the edit/visibility/panel children and their service graph.
+ */
+describe('ProfileLayoutComponent — Flow C cold-return read-your-writes (LFXV2-2933)', () => {
+  const STALE_PROFILE = {
+    user: { first_name: 'Ada', last_name: 'Lovelace', username: 'ada', email: 'ada@x.io' },
+    profile: { bio: 'OLD BIO', job_title: 'Engineer' },
+  } as unknown as CombinedProfile;
+
+  async function setup(queryParams: Record<string, string>, pendingSave?: unknown): Promise<ComponentFixture<ProfileLayoutComponent>> {
+    if (pendingSave !== undefined) {
+      sessionStorage.setItem(PENDING_PROFILE_SAVE_KEY, JSON.stringify(pendingSave));
+    }
+
+    const getCurrentUserProfile = vi.fn(() => of(STALE_PROFILE));
+    const userServiceMock = {
+      user: signal({ user_id: 'u1' } as unknown as User),
+      impersonating: signal(false),
+      uploadedAvatarUrl: signal<string | null>(null),
+      effectiveAvatarUrl: computed(() => ''),
+      identitiesRefresh$: EMPTY,
+      getCurrentUserProfile,
+      updateUserProfile: vi.fn(() => of({})),
+      getIdentities: vi.fn(() => of([])),
+    };
+
+    TestBed.configureTestingModule({
+      imports: [ProfileLayoutComponent],
+      providers: [
+        { provide: PLATFORM_ID, useValue: 'browser' },
+        { provide: ActivatedRoute, useValue: { queryParams: of(queryParams) } },
+        { provide: Router, useValue: { url: '/profile', navigateByUrl: vi.fn() } },
+        { provide: UserService, useValue: userServiceMock },
+        { provide: FeatureFlagService, useValue: { getBooleanFlag: vi.fn(() => signal(false)) } },
+        { provide: MessageService, useValue: { add: vi.fn() } },
+      ],
+    });
+    // Empty template: exercise the class without rendering the drawer/panel children.
+    TestBed.overrideComponent(ProfileLayoutComponent, { set: { template: '', imports: [] } });
+
+    const fixture = TestBed.createComponent(ProfileLayoutComponent);
+    fixture.detectChanges();
+    await fixture.whenStable();
+    return fixture;
+  }
+
+  beforeEach(() => {
+    sessionStorage.clear();
+    TestBed.resetTestingModule();
+  });
+
+  it('re-applies a stashed save after the initial GET lands, so the pre-save body cannot win', async () => {
+    const fixture = await setup({ success: 'profile_token_obtained' }, { savedAt: Date.now(), userMetadata: { bio: 'NEW BIO' } });
+
+    // The replayed save (bio: NEW BIO) resolved before the eventually-consistent GET (bio: OLD BIO);
+    // the optimistic override must reflect the write, not the stale refetch.
+    expect(fixture.componentInstance.aboutMe()).toBe('NEW BIO');
+  });
+
+  it('does not fabricate an override on a normal cold load with no pending save', async () => {
+    const fixture = await setup({});
+
+    // No stash → reapply is a no-op and the fetched GET body drives the view.
+    expect(fixture.componentInstance.aboutMe()).toBe('OLD BIO');
+  });
+});
--- a/apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
+++ b/apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
@@ -15,13 +15,13 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { ProfileLayoutComponent } from './profile-layout.component';
 
 /**
- * Guards the Flow C (management-token re-auth) cold-return read-your-writes fix (LFXV2-2933):
+ * Guards the Flow C (management-token re-auth) cold-return read-your-writes fix (LFXV2-3267):
  * on a full-page return the replayed save can resolve before the initial profile GET populates
  * combinedProfile, so the save is stashed and re-applied once a GET lands — an eventually-consistent
  * (pre-save) GET body must never mask the write. Template is overridden empty so the class logic runs
  * without instantiating the edit/visibility/panel children and their service graph.
  */
-describe('ProfileLayoutComponent — Flow C cold-return read-your-writes (LFXV2-2933)', () => {
+describe('ProfileLayoutComponent — Flow C cold-return read-your-writes (LFXV2-3267)', () => {
   const STALE_PROFILE = {
     user: { first_name: 'Ada', last_name: 'Lovelace', username: 'ada', email: 'ada@x.io' },
     profile: { bio: 'OLD BIO', job_title: 'Engineer' },
--- a/apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
+++ b/apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
@@ -9,7 +9,7 @@ import { CombinedProfile, User } from '@lfx-one/shared/interfaces';
 import { FeatureFlagService } from '@services/feature-flag.service';
 import { UserService } from '@services/user.service';
 import { MessageService } from 'primeng/api';
-import { EMPTY, of } from 'rxjs';
+import { EMPTY, Observable, of, Subject } from 'rxjs';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { ProfileLayoutComponent } from './profile-layout.component';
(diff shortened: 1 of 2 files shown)


audigregorie (commented):
## Code Review Summary

Tight, well-reasoned fix for the Flow C cold-return read-your-writes gap — the stash/reapply ordering is correct and loop-free. The main weakness is test fragility: the headline test doesn't explicitly prove the stash branch was taken, leaning on `toObservable`'s effect scheduling rather than asserting the path. Side effects inside a `map` operator are also worth a follow-up.

### What's done well

- Correct ordering: `mapToHeaderData` sets `combinedProfile` before `reapplyPendingOptimisticUpdate` reads it; the reapply→merge path can't re-enter the stash branch, so no loop.
- Stash accumulation (`{ ...pending, ...metadata }`) sensibly handles multiple saves before any base profile lands, and the null-profile no-op guard correctly avoids fabricating a profile.
- Comments precisely capture the eventually-consistent rationale and the `metadataLoaded` flip risk.

### Major — outside the diff

- (None outside the diff.)

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    await fixture.whenStable();
+    return fixture;
+  }
+
+  beforeEach(() => {
+    sessionStorage.clear();
+    TestBed.resetTestingModule();
+  });
+
+  it('re-applies a stashed save after the initial GET lands, so the pre-save body cannot win', async () => {
```
audigregorie, line 88: **Major — Test doesn't provably exercise the stash/reapply path it claims to guard**

The first test asserts `aboutMe()` is `'NEW BIO'`, but whether the stash branch is actually taken depends entirely on `toObservable(userService.user)` deferring its first emission until `fixture.detectChanges()` runs the effect — so the GET fires *after* the constructor body stashes the save. Both `updateUserProfile` and `getCurrentUserProfile` are synchronous `of(...)`, so if `toObservable` ever emits synchronously during field init, the GET resolves before the constructor body fires the save, the stash/reapply path is skipped, and the test still passes via the warm merge path — exercising zero coverage of the new `pendingOptimisticMetadata` / `reapplyPendingOptimisticUpdate` code.

There's no assertion that the stash branch was entered (e.g., a spy on `refreshProfile$.next` from that branch, or a deferred GET subject making the ordering explicit). This is a fragile guard that can silently pass while testing the wrong path.

```suggestion
    // Defer the GET so the save provably resolves first (exercises the stash branch).
    const profileSubject = new Subject<CombinedProfile>();
    const getCurrentUserProfile = vi.fn(() => profileSubject.asObservable());
    // ...
    const fixture = await setup({ success: 'profile_token_obtained' }, { savedAt: Date.now(), userMetadata: { bio: 'NEW BIO' } }, getCurrentUserProfile);
    // Save has stashed + triggered refreshProfile$.next(); flush the GET to land the stale body.
    profileSubject.next(STALE_PROFILE);
    expect(fixture.componentInstance.aboutMe()).toBe('NEW BIO');
```

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.ts
```diff
@@ -371,7 +384,13 @@ export class ProfileLayoutComponent {
             filter((user) => user !== null),
             switchMap(() =>
               this.userService.getCurrentUserProfile().pipe(
-                map((profile: CombinedProfile) => this.mapToHeaderData(profile)),
+                map((profile: CombinedProfile) => {
+                  const headerData = this.mapToHeaderData(profile);
+                  // Read-your-writes: if a save landed before this GET (Flow C cold return), re-apply it
+                  // now that combinedProfile is populated so a stale (pre-save) body can't win.
+                  this.reapplyPendingOptimisticUpdate();
```
audigregorie: **Major — Side effects inside a `map` operator (RxJS anti-pattern)**

`reapplyPendingOptimisticUpdate()` mutates `combinedProfile`, `optimisticProfileData`, `pendingOptimisticMetadata`, and calls `mapToHeaderData` (which itself sets `loading` and `combinedProfile`) — all inside `map`, which is contractually a pure projection. Side effects in `map` complicate reasoning about emission ordering and make the pipeline harder to test. Functionally correct today, but fragile.

```suggestion
                map((profile: CombinedProfile) => this.mapToHeaderData(profile)),
                tap(() => this.reapplyPendingOptimisticUpdate()),
                catchError(() => of(null))
```

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    const fixture = await setup({ success: 'profile_token_obtained' }, { savedAt: Date.now(), userMetadata: { bio: 'NEW BIO' } });
+
+    // The replayed save (bio: NEW BIO) resolved before the eventually-consistent GET (bio: OLD BIO);
+    // the optimistic override must reflect the write, not the stale refetch.
+    expect(fixture.componentInstance.aboutMe()).toBe('NEW BIO');
```
audigregorie, line 125: **Minor — Merge preservation of unchanged fields is unasserted**

The stash overlay should preserve unchanged fields (e.g., `job_title: 'Engineer'`), but the test only checks `aboutMe`. A regression that clobbers unchanged fields during the merge would pass silently.

```suggestion
    expect(fixture.componentInstance.aboutMe()).toBe('NEW BIO');
    expect(fixture.componentInstance.jobTitle()).toBe('Engineer');
```

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.ts
```diff
@@ -247,6 +246,20 @@ export class ProfileLayoutComponent {
...
+    // The merge supersedes any stash; clear it so a later GET doesn't re-apply a now-stale overlay.
+    this.pendingOptimisticMetadata = null;
+  }
+
+  // After a GET populates combinedProfile, re-apply a save that was stashed because no base profile
+  // existed when it resolved (Flow C cold return), so an eventually-consistent (pre-save) body can't
+  // mask the write. No-op until a real profile record lands (merging a null profile would fabricate one).
+  private reapplyPendingOptimisticUpdate(): void {
+    const pending = this.pendingOptimisticMetadata;
+    if (!pending || this.combinedProfile?.profile == null) {
```
audigregorie, line 258: **Minor — Defensive-clear and null-profile no-op branches lack test coverage**

The `pendingOptimisticMetadata = null` clear after merge (line 250) and the null-profile early return here (`this.combinedProfile?.profile == null`) have no tests. A second GET after reapply (asserting the overlay isn't re-applied stale) and a GET returning a `profile: null` body (asserting the stash is retained, not fabricated) would close the gap.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.ts
fayazg: Done — moved `reapplyPendingOptimisticUpdate()` out of `map` into a `tap` so `map` is a pure projection. Behavior is identical (tap runs after the projection, same order as before). See commit 8da6f920.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
fayazg: Done — reworked the test to drive the GET through a `Subject<CombinedProfile>` so the ordering is explicit rather than relying on `toObservable` deferral. The save stashes during construction; the test now asserts `aboutMe()` is `''` *before* emitting (proving the stash branch was taken, not a warm merge), then `.next(STALE_PROFILE)` lands the stale body and reapply wins. See commit 8da6f920.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts
fayazg: Done — added `expect(fixture.componentInstance.jobTitle()).toBe('Engineer')` so a merge that clobbers an untouched field fails the test. See commit 8da6f920.

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.ts
fayazg: Done — added a Subject-driven test covering both branches: a null-profile GET (stash retained, no fabrication) → a real GET (applied + cleared) → a later GET whose `jobTitle` stays `Engineer`, proving the stash was cleared and isn't re-applied stale. See commit 8da6f920.

fayazg:
## Review Feedback Addressed

Commit: 8da6f920

### Changes Made
- **profile-layout.component.ts**: moved `reapplyPendingOptimisticUpdate()` out of the `map` operator into `tap`, so the GET stream projects header data without a side effect (per @audigregorie)
- **profile-layout.component.spec.ts**: reworked the primary test to drive the GET through a `Subject<CombinedProfile>` — it now asserts `aboutMe()` is `''` before the GET emits (proving the stash branch was taken, not a warm merge), then lands the stale body and verifies reapply wins (per @audigregorie)
- **profile-layout.component.spec.ts**: added `expect(jobTitle()).toBe('Engineer')` so a merge that clobbers an untouched field fails (per @audigregorie)
- **profile-layout.component.spec.ts**: added a Subject-driven test covering the previously-uncovered branches — null-profile GET (stash retained, not fabricated) → real GET (applied + cleared) → later GET where `jobTitle` stays `Engineer`, proving the defensive clear (per @audigregorie)

### Validation
`yarn test:app` (323 pass), `yarn format`, `yarn lint`, `yarn build` — all green.

### Threads Resolved

apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.ts
```diff
@@ -247,6 +246,20 @@ export class ProfileLayoutComponent {
...
+      return;
```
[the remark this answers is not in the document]
fayazg: Good catch on the one-shot semantics — you're right that a single `profile: null` GET followed by no further refresh would leave the overlay deferred. This is an intentional trade-off rather than a gap:

- The null-profile branch deliberately does **not** set the overlay, because merging into a null profile fabricates a profile record and flips the drawer's `metadataLoaded` to true (see the comment at lines 219–222). Applying "the overlay without fabricating the drawer profile" is exactly what we're avoiding — there's no clean way to show it without the fabricated record.
- This runs on the Flow C path **after** a successful `updateUserProfile` PATCH persisted the metadata, so a real profile record is expected on the next GET. A persistent `profile: null` would imply the save didn't persist upstream, which is a different failure.
- The pre-fix code already did a single `refreshProfile$.next()` refetch here with no polling, so this isn't a regression the PR introduces.

Adding a bounded/delayed refetch would introduce polling + backoff to cover a defensive edge that shouldn't persist after a successful save, so I'm leaving the single-refetch behavior as-is. Happy to file a follow-up if we ever observe a real profile record failing to land.

The Subject-driven test emits null then a populated body to exercise both the no-op and the clear branches of the class logic — agreed it doesn't model the one-shot HttpClient nature, but that's the coverage target here rather than the fetch scheduling.

fayazg:

### No Change Needed
- **profile-layout.component.ts:259** — the null-profile branch not scheduling a bounded refetch is an intentional trade-off: setting the overlay there would fabricate a profile record and flip the drawer's `metadataLoaded` (see lines 219–222). It runs after a successful save PATCH, so a real profile record is expected on the next GET, and the pre-fix code already used a single refetch with no polling (no regression). Explained on the thread and resolved; happy to file a follow-up if a real record is ever observed failing to land. (flagged by copilot[bot])

### Threads Resolved

dealako:
Hi @fayazg 👋 — reviewed this one with parallel security / correctness / style passes and reconciled against the existing Copilot + @audigregorie threads.

## Overall impression

Tight, well-scoped fix for the Flow C cold-return read-your-writes gap. The stash → reapply ordering is correct and loop-free: the stash is cleared after a merge (so a later GET can't re-apply a stale overlay), the reapply is a genuine no-op until a real `profile` record lands (so a null-profile body can't fabricate one and flip the drawer's `metadataLoaded`), and the tests now provably drive the stash branch via a `Subject` rather than leaning on `toObservable` scheduling. The prior-round feedback (side-effects moved from `map` → `tap`, untouched-field assertion, null-profile + clear-branch coverage) is all landed in `8da6f920`, and the LFXV2-2933 → LFXV2-3267 label is corrected. Clean work.

## Issue count

- 🔴 **Blocking: 0**
- 🟡 **Minor: 1** — null-profile post-save GET orphans the stash with no self-driven retry (**agrees with Copilot's existing line-259 thread**; accepting your documented trade-off, noted below).
- ⚪ **Nit: 1** — `reapplyPendingOptimisticUpdate` uses `//` line comments while its sibling `applyOptimisticProfileUpdate` uses a `/** */` JSDoc block; the file already mixes both styles, so pure preference.
- ❔ **Question: 0**

### On the null-profile branch (reconciling with Copilot)

I independently reached the same place Copilot did: `getCurrentUserProfile()` is a one-shot HttpClient GET, so if the single post-save GET returns `profile: null`, `reapplyPendingOptimisticUpdate()` no-ops (correctly, to avoid fabricating a record) but nothing re-issues a GET — the overlay stays deferred until an unrelated refresh. Your reply on that thread is a reasonable trade-off: this path runs **after** a successful `updateUserProfile` PATCH, so a real profile record is expected on the next GET; a persistent `profile: null` implies the save didn't persist (a different failure); and the pre-fix code already did a single refetch with no polling, so this isn't a regression. Adding bounded polling/backoff to cover it would be net-negative here. Accepting as-is — a follow-up if a real record is ever observed failing to land is the right call. One small note: the Subject-driven test emits null → populated through one stream, which exercises the class branches but doesn't model the one-shot HTTP semantics — you already acknowledged this on the thread, so no action needed.

## Decision

✅ **Approved with minor comments** — the single Minor is an already-discussed, reasonably-justified trade-off and the Nit is cosmetic. Nice, surgical fix with strong test intent.

dealako (approved):
Approved with minor comments — see the summary comment for the full breakdown. No blocking or security findings; the one Minor (null-profile GET orphaning the stash) is an already-discussed, reasonably-justified trade-off agreeing with Copilot's existing thread, and the remaining item is a cosmetic comment-style nit. Prior-round feedback (map→tap, test fragility, coverage gaps, issue label) is all resolved.

Issue #1571, opened 2026-08-14
feat(newsletters): open-in-new-tab affordance for the permalink


niravpatel27:
Follow-on to #1550. Add a new-tab icon to the drawer header (next to copy-link) and make the My Newsletters subject cell a real anchor so browser new-tab gestures work.

Issue #1550, opened 2026-08-14
feat(newsletters): per-issue permalinks and share affordance in the reader


manishdixitlfx:
## Problem

Individual newsletters have no shareable URL. Opening a newsletter from My Newsletters (`/newsletters/my`) renders it in a drawer with no URL change, and the drawer has no share, copy-link, or view-in-browser control.

Real-world case (2026-08-13): leadership wanted to link a specific newsletter (the LF Executive Director Weekly Newsletter) from a weekly exec update to the CEO — the closest linkable surface is the whole reader list, not the issue itself.

Verified in production 2026-08-13:

- Clicking a newsletter row on `/newsletters/my` opens the reader drawer; the URL stays `/newsletters/my`
- No share / permalink / view-in-browser control anywhere in the drawer

## Proposal

1. Permalink per newsletter issue — a route like `/newsletters/:id` (or `/newsletters/my?issue=:id`) that opens the reader directly on that issue, honoring existing access control (group membership).
2. Share affordance in the reader drawer — a copy-link control in the drawer header; opening a drawer should also update the address bar so the URL is always copyable.

## Precedent

Public groups just got readable URL slugs (#1318, public groups directory work) — this is the same "make it linkable" treatment applied to newsletters.

## Notes

- Fits the newsletter program (analytics/scheduling epics targeting Aug 27) — permalinks also enable click-metric attribution and "share this issue" flows later.
- Access control unchanged: a permalink for a private group's newsletter should require membership, same as today's reader.

Context: requested by Kieran during review of the weekly exec update.

PR #1578, opened 2026-08-14
fix(newsletters): drop hover new-tab icon from list rows

niravpatel27:
## Summary

Follow-up refinement to #1550 / #1571's new-tab affordances: removes the hover-revealed
external-link icon from the My Newsletters list rows.

The permalink is still reachable via the drawer's new-tab button (added in the earlier
PR). The list-row icon added visual noise on hover without a clear win now that the
subject cell itself already behaves like a real link (native new-tab/right-click via
its `href`) — the icon was redundant signaling.

## Changes

- `my-newsletters.component.html`: drop the hover-revealed `fa-arrow-up-right-from-square`
  icon and the now-unused `group` class/comment on the row `<tr>`. The subject anchor's
  href-driven new-tab/right-click behavior is unaffected.

## Testing

- `./check-headers.sh`, `yarn format:check`, `yarn lint:check`, `yarn build` — all clean
- Manual visual check: hovering a row subject no longer shows the icon; plain click still
  opens the drawer; Cmd/Ctrl-click / middle-click / right-click on the subject still work
  natively (unchanged, since the anchor's `href` is untouched)

## Reviews run pre-PR

- `/lfx-self-serve-pr-readiness` — READY (branch renamed to `feat/GH-1571` per convention)
- `/preflight` — all mechanical checks clean

### DIFF
--- a/apps/lfx-one/src/app/modules/newsletters/my-newsletters/my-newsletters.component.html
+++ b/apps/lfx-one/src/app/modules/newsletters/my-newsletters/my-newsletters.component.html
@@ -107,9 +107,8 @@
               <ng-template #body let-newsletter>
                 <!-- The subject anchor is the accessible control (real link role, native
                      Enter/Space, right-click "Open in new tab"); the row click is a
-                     supplementary pointer convenience. `group` scopes the hover-revealed
-                     new-tab icon to this row. -->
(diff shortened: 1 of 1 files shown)


dealako (approved):
LGTM

PR #1582, opened 2026-08-17
fix: enforceed auth on campaigns and marketing analytics routes

MRashad26:
## Summary
- `campaigns.route.ts` applied no authorization middleware to any of its 17 endpoints, including writes (`POST /create`, `PATCH /:campaignId/status`) — gating was client-side only via the `executiveDirectorGuard` route guard.
- Several ED-only marketing analytics endpoints in `analytics.route.ts` had the same gap.
- Wires `requireExecutiveDirector` onto all of these, reusing the same server-verified persona check already applied to the three event-analytics endpoints.
- Generalizes `require-executive-director.middleware.ts`'s scope check to read either `foundationSlug` (analytics.route.ts's convention) or `project` (campaigns.route.ts's convention) — the two route families name the same scoping concept differently, so the middleware previously only ever scope-checked the analytics side even where applied.
- `multi-foundation-summary` and `PATCH /:campaignId/status` have no single-slug param to scope against (a `slugs` list and no param, respectively), so only the base ED-persona gate applies there, consistent with existing unscoped-endpoint behavior.

This closes a live server-side authorization hole (LFXV2-2231 epic gap-analysis, item G3) independent of the epic's FGA-based re-implementation — it reuses the existing ED-persona server check rather than attempting the epic's `campaign_manager`/`marketing_auditor` relation work.

JIRA: LFXV2-3290

## Test plan
- [x] `yarn format` / `yarn lint` / `yarn build` all pass
- [x] Added unit tests for the middleware's `project`/`foundationSlug` scope fallback and precedence (`require-executive-director.middleware.spec.ts`)
- [x] Reviewer trio (general, self-serve conventions, learnings) run clean on both commits and on the full-branch diff
- [ ] Manual verification in a deployed env: confirm a non-ED authenticated user gets 403 on `POST /api/campaigns/create` and `GET /api/analytics/brand-health`

### DIFF
--- a/apps/lfx-one/src/server/middleware/require-executive-director.middleware.ts
+++ b/apps/lfx-one/src/server/middleware/require-executive-director.middleware.ts
@@ -42,7 +42,11 @@ export async function requireExecutiveDirector(req: Request, res: Response, next
       return;
     }
 
-    const requestedSlug = typeof req.query['foundationSlug'] === 'string' ? req.query['foundationSlug'] : '';
+    // Analytics routes scope by `foundationSlug`; campaigns routes scope by `project` — same
+    // concept, two names carried over from each route file's own convention.
+    const requestedSlug =
+      (typeof req.query['foundationSlug'] === 'string' ? req.query['foundationSlug'] : '') ||
+      (typeof req.query['project'] === 'string' ? req.query['project'] : '');
 
     // No slug on the request means there is nothing to scope against — the handler is responsible
     // for rejecting a missing required parameter, and unscoped ED endpoints stay allowed.
--- a/apps/lfx-one/src/server/routes/analytics.route.ts
+++ b/apps/lfx-one/src/server/routes/analytics.route.ts
@@ -143,21 +143,22 @@ router.get('/org-involvement-event-attendance-monthly', (req, res, next) => anal
 router.get('/org-involvement-certified-employees-monthly', (req, res, next) => analyticsController.orgCertifiedEmployeesMonthly(req, res, next));
 router.get('/org-involvement-training-enrollments', (req, res, next) => analyticsController.orgTrainingEnrollments(req, res, next));
 
-// Web activities summary endpoint (marketing dashboard)
-router.get('/web-activities-summary', (req, res, next) => analyticsController.getWebActivitiesSummary(req, res, next));
+// Web activities summary endpoint (marketing dashboard) — ED-only surface, same as the
+// events-overview-summary/event-roster/event-detail endpoints below.
+router.get('/web-activities-summary', requireExecutiveDirector, (req, res, next) => analyticsController.getWebActivitiesSummary(req, res, next));
 
 // Email CTR endpoint (marketing dashboard)
-router.get('/email-ctr', (req, res, next) => analyticsController.getEmailCtr(req, res, next));
+router.get('/email-ctr', requireExecutiveDirector, (req, res, next) => analyticsController.getEmailCtr(req, res, next));
 
 // Social reach endpoint (marketing dashboard)
-router.get('/social-reach', (req, res, next) => analyticsController.getSocialReach(req, res, next));
+router.get('/social-reach', requireExecutiveDirector, (req, res, next) => analyticsController.getSocialReach(req, res, next));
 
 // Keyword performance endpoint (marketing dashboard)
-router.get('/keyword-performance', (req, res, next) => analyticsController.getKeywordPerformance(req, res, next));
+router.get('/keyword-performance', requireExecutiveDirector, (req, res, next) => analyticsController.getKeywordPerformance(req, res, next));
 
 // Social media endpoints (marketing dashboard)
-router.get('/social-media', (req, res, next) => analyticsController.getSocialMedia(req, res, next));
-router.get('/social-media/monthly', (req, res, next) => analyticsController.getSocialMediaMonthly(req, res, next));
+router.get('/social-media', requireExecutiveDirector, (req, res, next) => analyticsController.getSocialMedia(req, res, next));
+router.get('/social-media/monthly', requireExecutiveDirector, (req, res, next) => analyticsController.getSocialMediaMonthly(req, res, next));
 
 // North Star metrics endpoints (executive director dashboard)
 router.get('/member-retention', (req, res, next) => analyticsController.getMemberRetention(req, res, next));
@@ -203,13 +204,15 @@ router.get('/event-roster', requireExecutiveDirector, (req, res, next) => analyt
 // per-tier sponsorship breakdown and CFP status. The drawer that opens it is ED-only
 // client-side, so authorization is enforced here with server-verified persona detection.
 router.get('/event-detail', requireExecutiveDirector, (req, res, next) => analyticsController.getEventDetail(req, res, next));
-router.get('/brand-reach', (req, res, next) => analyticsController.getBrandReach(req, res, next));
-router.get('/brand-health', (req, res, next) => analyticsController.getBrandHealth(req, res, next));
-router.get('/revenue-impact', (req, res, next) => analyticsController.getRevenueImpact(req, res, next));
-router.get('/marketing-attribution', (req, res, next) => analyticsController.getMarketingAttribution(req, res, next));
+router.get('/brand-reach', requireExecutiveDirector, (req, res, next) => analyticsController.getBrandReach(req, res, next));
+router.get('/brand-health', requireExecutiveDirector, (req, res, next) => analyticsController.getBrandHealth(req, res, next));
+router.get('/revenue-impact', requireExecutiveDirector, (req, res, next) => analyticsController.getRevenueImpact(req, res, next));
+router.get('/marketing-attribution', requireExecutiveDirector, (req, res, next) => analyticsController.getMarketingAttribution(req, res, next));
 
-// Multi-foundation summary endpoint (multi-foundation dashboard)
-router.get('/multi-foundation-summary', (req, res, next) => analyticsController.getMultiFoundationSummary(req, res, next));
+// Multi-foundation summary endpoint (multi-foundation dashboard). Scoped by a `slugs` list rather
+// than a single foundationSlug/project, so requireExecutiveDirector only enforces the ED gate here,
+// not per-slug scoping — same as any other unscoped ED endpoint.
+router.get('/multi-foundation-summary', requireExecutiveDirector, (req, res, next) => analyticsController.getMultiFoundationSummary(req, res, next));
 
 // Org Lens — bootstrap account context (display attrs + cdev mapping + tier)
 router.get('/org-lens-account-context', (req, res, next) => analyticsController.getOrgLensAccountContext(req, res, next));
--- a/apps/lfx-one/src/server/routes/campaigns.route.ts
+++ b/apps/lfx-one/src/server/routes/campaigns.route.ts
@@ -4,31 +4,35 @@
 import { Router } from 'express';
 
 import { CampaignController } from '../controllers/campaign.controller';
+import { requireExecutiveDirector } from '../middleware/require-executive-director.middleware';
 
 const router = Router();
 const campaignController = new CampaignController();
 
-router.post('/brief/generate', (req, res, next) => campaignController.generateBrief(req, res, next));
-router.post('/brief/refine', (req, res, next) => campaignController.refineBrief(req, res, next));
-router.post('/brief/persist', (req, res, next) => campaignController.persistBrief(req, res, next));
-router.get('/brief', (req, res, next) => campaignController.loadBrief(req, res, next));
-router.post('/create', (req, res, next) => campaignController.createCampaign(req, res, next));
-router.get('/jobs/:jobId', (req, res, next) => campaignController.getJobStatus(req, res, next));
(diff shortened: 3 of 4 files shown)


apps/lfx-one/src/server/middleware/require-executive-director.middleware.ts
```diff
@@ -42,7 +42,11 @@ export async function requireExecutiveDirector(req: Request, res: Response, next
       return;
     }
 
-    const requestedSlug = typeof req.query['foundationSlug'] === 'string' ? req.query['foundationSlug'] : '';
+    // Analytics routes scope by `foundationSlug`; campaigns routes scope by `project` — same
+    // concept, two names carried over from each route file's own convention.
+    const requestedSlug =
+      (typeof req.query['foundationSlug'] === 'string' ? req.query['foundationSlug'] : '') ||
+      (typeof req.query['project'] === 'string' ? req.query['project'] : '');
```
dealako, line 49: [blocking] Dual-param precedence lets the gate check a different slug than the handler uses

Issue: `foundationSlug || project` means when *both* are present the gate scopes on `foundationSlug`, but campaign handlers (`loadBrief`, `persistBrief`, `createCampaign`, …) consume `project`.
Proof: an ED whose `personaProjects` only contains `tlf` calls `?foundationSlug=tlf&project=cncf` → `requestedSlug='tlf'` → `edSlugs.includes('tlf')` → `next()`; the handler then operates on `cncf`. Copilot flagged the same at line 49.
Why it matters: cross-tenant IDOR — an ED for foundation A reads/writes foundation B's campaign data.
Fix: don't silently prefer one param. Either reject when both are present and disagree, or have each route declare which single scope param it uses and check that exact one. Note the added spec test `prefers foundationSlug over project` currently asserts `allow` for this exact attack shape — it should assert `deny` (or be removed) once the precedence is fixed.

Separately (minor): the inline `typeof req.query[...] === 'string' ? ... : ''` ternary re-implements the server layer's shared `getStringQueryParam` helper — prefer reuse for consistency.

apps/lfx-one/src/server/routes/analytics.route.ts
```diff
@@ -203,13 +204,15 @@ router.get('/event-roster', requireExecutiveDirector, (req, res, next) => analyt
...
 router.get('/event-detail', requireExecutiveDirector, (req, res, next) => analyticsController.getEventDetail(req, res, next));
-router.get('/brand-reach', (req, res, next) => analyticsController.getBrandReach(req, res, next));
-router.get('/brand-health', (req, res, next) => analyticsController.getBrandHealth(req, res, next));
-router.get('/revenue-impact', (req, res, next) => analyticsController.getRevenueImpact(req, res, next));
-router.get('/marketing-attribution', (req, res, next) => analyticsController.getMarketingAttribution(req, res, next));
-
-// Multi-foundation summary endpoint (multi-foundation dashboard)
-router.get('/multi-foundation-summary', (req, res, next) => analyticsController.getMultiFoundationSummary(req, res, next));
+router.get('/brand-reach', requireExecutiveDirector, (req, res, next) => analyticsController.getBrandReach(req, res, next));
+router.get('/brand-health', requireExecutiveDirector, (req, res, next) => analyticsController.getBrandHealth(req, res, next));
```
dealako, line 208: [blocking] These gates lock out LF Staff, who are supported on this surface

Issue: `requireExecutiveDirector` rejects any caller without the ED persona *before* reaching the `isLFStaff`/`isRootWriter` bypass (which only skips per-foundation *scope*, not the ED gate itself).
Proof: `/foundation/marketing-impact` is guarded by `dashboardAccessGuard` = ED **or** LF Staff (`dashboard-access.guard.ts`), and its comment states "LF Staff see only the Social Listening tab" — which calls `brand-health`. Marketing Overview on the ED dashboard (also LF-Staff-visible via `canViewExecutiveDashboards()`) calls `brand-reach`, `revenue-impact`, `email-ctr`, `social-reach`, `web-activities-summary`. LF Staff without the ED persona now get 403 on all of these. Cursor and Copilot flagged the same.
Why it matters: live functional regression for the supported LF Staff path.
Fix: hoist the `isRootWriter || isLFStaff` bypass *above* the `isED` rejection (so staff pass the gate and skip scoping), or use a dedicated ED-or-LF-Staff server gate for these shared endpoints. The spec's LF-staff test also needs a fixture *without* the ED persona to actually exercise this path.

apps/lfx-one/src/server/routes/analytics.route.ts
```diff
@@ -203,13 +204,15 @@ router.get('/event-roster', requireExecutiveDirector, (req, res, next) => analyt
...
+router.get('/revenue-impact', requireExecutiveDirector, (req, res, next) => analyticsController.getRevenueImpact(req, res, next));
+router.get('/marketing-attribution', requireExecutiveDirector, (req, res, next) => analyticsController.getMarketingAttribution(req, res, next));
+
+// Multi-foundation summary endpoint (multi-foundation dashboard). Scoped by a `slugs` list rather
+// than a single foundationSlug/project, so requireExecutiveDirector only enforces the ED gate here,
+// not per-slug scoping — same as any other unscoped ED endpoint.
+router.get('/multi-foundation-summary', requireExecutiveDirector, (req, res, next) => analyticsController.getMultiFoundationSummary(req, res, next));
```
dealako, line 215: [blocking] `multi-foundation-summary` is not an ED-only surface

Issue: this endpoint is consumed by `MultiPersonaDashboardComponent.initAnalyticsSummary()` for *any* user with more than one persona.
Proof: `multi-persona-dashboard.component.ts:174` calls `getMultiFoundationSummary(slugs)` for board members, maintainers, and contributors; the ED gate now 403s them and `catchError(() => of(null))` silently blanks their per-foundation summary. Cursor and Copilot flagged the same.
Why it matters: non-ED multi-persona users lose analytics they're entitled to, with no visible error.
Fix: drop `requireExecutiveDirector` from this route and scope it another way. ❔ Even for EDs it accepts up to 25 caller-supplied `slugs` queried directly with no per-slug scope check (ED for A can read B) — so this route needs its own authorization model rather than the ED gate. What boundary is intended here?

apps/lfx-one/src/server/routes/campaigns.route.ts
```diff
@@ -4,31 +4,35 @@
...
+router.get('/linkedin/monitor', requireExecutiveDirector, (req, res, next) => campaignController.getLinkedInMonitor(req, res, next));
+router.get('/reddit/accounts', requireExecutiveDirector, (req, res) => campaignController.getRedditAccounts(req, res));
+router.get('/reddit/monitor', requireExecutiveDirector, (req, res, next) => campaignController.getRedditMonitor(req, res, next));
+router.get('/meta/accounts', requireExecutiveDirector, (req, res) => campaignController.getMetaAccounts(req, res));
+router.get('/meta/monitor', requireExecutiveDirector, (req, res, next) => campaignController.getMetaMonitor(req, res, next));
+router.get('/keywords', requireExecutiveDirector, (req, res, next) => campaignController.getKeywords(req, res, next));
+router.get('/audience', requireExecutiveDirector, (req, res, next) => campaignController.getAudience(req, res, next));
+router.post('/keywords/actions', requireExecutiveDirector, (req, res, next) => campaignController.executeKeywordActions(req, res, next));
+router.patch('/:campaignId/status', requireExecutiveDirector, (req, res, next) => campaignController.updateCampaignStatus(req, res, next));
```
dealako, line 36: [minor] ED gate here is unscoped — any ED can mutate any foundation's campaign

Issue: `PATCH /:campaignId/status` carries no foundation scope param, so `requireExecutiveDirector` only enforces the base ED persona and `updateCampaignStatus` forwards the caller-controlled numeric `campaignId` to the globally-configured Meta/Reddit clients with no ownership check (`campaign-proxy.service.ts`).
Proof: an ED for foundation A can pause/activate foundation B's campaign by ID. Copilot flagged the same at line 36.
Why it matters: cross-tenant write. The PR body acknowledges the missing scope param, so this is a known limitation rather than an oversight — flagging so it's an explicit trade-off decision.
Fix (follow-up acceptable): resolve the campaign's owning foundation from `campaignId` and scope it through the middleware, or document why cross-foundation status changes are acceptable for now.

dealako (changes requested):
Hi @MRashad26 👋 — thanks for tackling this. Closing the server-side authorization gap on campaigns + marketing analytics is genuinely valuable, and the instinct to reuse the existing server-verified ED persona check (rather than waiting on the epic's FGA relations) is the right pragmatic call. The middleware, spec, and route wiring are all clean and well-commented.

That said, the review surfaced three **blocking** issues where the new server gate is *stricter or looser than the client access model it's supposed to mirror*, and they'd cause live regressions or leave a hole open. CodeRabbit's auto-review is disabled here, but Cursor Bugbot and Copilot independently flagged the same three — I've verified each against the actual callers and agree.

### 🔴 Blocking (3)
1. **LF Staff get 403 on marketing analytics.** The `isED` gate rejects before the `isLFStaff`/`isRootWriter` scope bypass, but `dashboardAccessGuard` (ED **or** LF Staff) guards `/foundation/marketing-impact`, and LF Staff see the Social Listening tab (`brand-health`) + Marketing Overview (`brand-reach`, `revenue-impact`, `email-ctr`, `social-reach`, `web-activities-summary`). Non-ED LF Staff now break. (agrees with Cursor + Copilot)
2. **`multi-foundation-summary` 403s non-ED multi-persona users.** `MultiPersonaDashboardComponent` calls it for *any* user with >1 persona (board, maintainer, contributor) — not an ED-only surface. The `catchError(() => of(null))` silently zeroes their foundation summary. (agrees with Cursor + Copilot)
3. **`foundationSlug`-over-`project` precedence is a cross-tenant IDOR.** When both params are present the gate checks `foundationSlug` while campaign handlers consume `project` — an ED scoped to `tlf` passing `foundationSlug=tlf&project=cncf` is allowed, then reads/writes `cncf`. The new spec test codifies this as intended. (agrees with Copilot)

### 🟡 Minor (2)
- **`PATCH /:campaignId/status` is ED-gated but unscoped** — any ED can toggle any foundation's campaign by numeric ID (no ownership check downstream). The PR acknowledges the missing scope param; flagging so it's a conscious trade-off, not an oversight. (agrees with Copilot)
- **Inline query-param ternary** bypasses the server layer's shared `getStringQueryParam` helper.

### ⚪ Nit (1)
- The `foundationSlug`-over-`project` precedence is tested but not documented in the comment (and per #3, the precedence itself should change).

### ❔ Question (1)
- Even ED-scoped, `multi-foundation-summary` accepts up to 25 caller-supplied `slugs` queried directly against Snowflake with no per-slug scope check — an ED for A can read B. This is entangled with #2 (the route probably shouldn't carry an ED gate at all); what's the intended authorization boundary for that endpoint?

**Decision: 🔴 Needs changes before approval.** The core fix is sound; the gate just needs to match the client's ED-**or**-LF-Staff (and persona-agnostic multi-persona) access model, and the dual-param scope needs to resolve to the *same* slug the handler uses. Tests should be updated alongside — the current LF-staff fixture includes the ED persona (so it never exercises the broken path) and the precedence test encodes the IDOR as expected behavior.

dealako:
## Review summary — @MRashad26

Thanks for closing this server-side authz gap 🙏 — reusing the existing ED persona check instead of blocking on the epic's FGA work is the right pragmatic move, and the wiring/comments are clean. Full findings + inline detail are in the [formal review](https://github.com/linuxfoundation/lfx-self-serve/pull/1582#pullrequestreview-4948926860); the theme is that the new **server** gate doesn't match the **client** access model it mirrors.

- 🔴 **Blocking: 3**
  1. LF Staff get 403 on marketing analytics — `isED` rejects before the `isLFStaff` bypass, but `marketing-impact`/Marketing Overview are ED **or** LF Staff surfaces.
  2. `multi-foundation-summary` 403s non-ED multi-persona users (board/maintainer/contributor) — it's not an ED-only endpoint.
  3. `foundationSlug`-over-`project` precedence is a cross-tenant IDOR — gate checks one slug, campaign handler consumes the other.
- 🟡 **Minor: 2** — unscoped `PATCH /:campaignId/status` (any ED → any foundation's campaign); inline query-param ternary bypasses the shared `getStringQueryParam` helper.
- ⚪ **Nit: 1** — precedence tested but undocumented (and should change per blocking #3).
- ❔ **Question: 1** — `multi-foundation-summary` accepts 25 caller-supplied slugs with no per-slug scope; what's the intended boundary?

Reconciled with Cursor Bugbot + Copilot — I independently verified and agree with all three of their overlapping findings (CodeRabbit auto-review is disabled on this repo). Tests should move with the fix: the LF-staff fixture currently includes the ED persona (never exercises the broken path) and the precedence test encodes the IDOR as expected behavior.

**Decision: 🔴 Needs changes before approval.** Core fix is sound — reconcile the gate with the ED-or-LF-Staff (and persona-agnostic multi-persona) model, and make the dual-param scope resolve to the same slug the handler reads.

MRashad26:
Closing in favor of #1584, which addresses the same gap (LFXV2-2231 gap-analysis item G3 — missing server-side authorization on `campaigns.route.ts`/`analytics.route.ts`). Both PRs were built independently off the same base commit before either was aware of the other; #1584 additionally includes router-level integration tests that drive real HTTP requests through the assembled routers, giving stronger regression coverage than this PR's middleware-only unit tests.

PR #1587, opened 2026-08-17
fix(profile): make profile hub rail responsive on mobile

manishdixitlfx:
## Summary

**Profile hub layout (finding 1)**
- The Profile & Account hub's profile rail was `fixed` at 300px with no breakpoint guard, and `<main>` reserved a matching `pr-[300px]` gutter unconditionally — at a 390px viewport this left ~50px of usable content width.
- The rail now renders inline in the content column below `2xl` (1440px) and as a fixed full-height right rail at `2xl` and up — moved from the initial `lg` (1024px) threshold in a follow-up once review surfaced that `lg`–`xl` (1024–1439px) still only had a ~320–616px content column beside a 300px rail. `2xl` is the first breakpoint where the hub's 704px of fixed chrome (348px left sidebar + 300px rail + 56px page padding) leaves comfortable room (736px+) for both.
- Below `2xl`, the inline panel itself reflows across three states so it isn't a narrow left-aligned card with blank space beside it on tablet/laptop: a stacked card below `sm` (capped at `max-w-md`), a full-width horizontal layout (header row + 2/3-col metadata grid + side-by-side social rows) from `sm` up to just under `2xl`, and back to the fixed rail's single-column layout at `2xl`+.
- Single `<lfx-profile-panel>` instance kept throughout (its `data-testid`s are asserted by e2e specs) — only wrapper/host classes and the panel's own internal layout change per breakpoint.

**Three additional responsive breakages (findings 2-4, independent of the layout fix)**
- `account-settings`: the "On this page" TOC always claimed ~160-180px; now hidden below `lg` (little value on a phone where sections are a short scroll apart).
- `affiliation-timeline-dialog`: each period row (4 `p-select`s + separator + actions, ~450px min-width) overflowed on every phone; rows now wrap, with the month/year pairs gridding to 2 columns below `sm` and pinning to their original 110px/100px widths at `sm`+. Added Start/End labels and disambiguated `ariaLabel`s (interpolating org + period index) so wrapped/duplicate rows stay identifiable to screen readers.
- `work-experience-form-dialog`: the start/end date-picker grid was a fixed 2-column layout giving each field ~165px at 375px; now stacks below `sm`.

**e2e regression coverage (LFXV2-3285 follow-up)**
- Added `apps/lfx-one/e2e/profile-mobile-layout.spec.ts`: an 8-viewport matrix (360/393/550/768/1023/1194/1280/1439px, covering phone through just-under-`2xl` laptop widths) × 2 routes, plus a dedicated desktop-control test above `2xl` — 54 cases total across chromium/firefox/mobile-chrome. Asserts the content column's width (scrollbar-safe, via `document.documentElement.clientWidth`, and `matchMedia`-based rather than threshold comparisons where a matrix viewport lands exactly on a breakpoint, since browser engines disagree on which viewport metric a `min-width` query resolves against when a classic scrollbar is present), no horizontal scroll, and — via a dedicated `profile-panel-rail` testid's computed CSS `position` and `max-width` — that the rail is genuinely in normal document flow below `2xl` (not a `fixed` overlay), that the `max-w-md` stacked-card cap is present below `sm`, and that `sm:max-w-none` actually lifts it from `sm` up to `2xl`.
- Live-verified red/green repeatedly through the branch's iteration: red against `origin/main`'s pre-fix layout, red against each intermediate regression this spec caught before its own assertions were fixed (a tautological cap check, a scrollbar-blind width comparison, a fail-open DOM-traversal hop), green against the current tree across all three Playwright projects.

## Known trade-offs (flagged in review, deliberately not addressed in this PR)
- **e2e coverage scope**: this spec covers the profile-rail layout only. `/profile/settings` is loaded but the TOC's `hidden lg:block` isn't asserted, and neither the affiliation-timeline nor work-experience dialogs are opened to verify their wrap/stack behavior. Those three responsive fixes can regress silently. Deferred rather than added here — opening and populating those dialogs (work-experience/affiliation data prerequisites, PrimeNG dialog interaction) is a meaningfully larger addition than a rail-layout regression test; worth its own follow-up ticket.
- **`affiliation-timeline-dialog` still uses raw `p-select`/`p-checkbox`**: the LFX `lfx-select` wrapper exists but requires `FormGroup`/`FormControl`, while this dialog is `[ngModel]`-driven against plain period objects. Migrating is a reactive-forms rewrite, not a responsive fix — pre-existing debt, not introduced here.
- **`affiliation-timeline-dialog` dialog width vs. `sm` breakpoint**: the dialog opens at a fixed 720px width (PrimeNG-capped ~90-95vw on narrow viewports) with no `breakpoints` config, so the `sm:` Tailwind classes key off viewport width while the actual wrap point is governed by the dialog's own rendered width — a mismatch could still cause wrapping just above the `sm` boundary. Mitigated by making the Start/End labels unconditional (not `sm`-gated) so the row stays legible even if it wraps; the underlying dialog-width mismatch was not changed since it wasn't visually verified.
- **Dev-environment e2e auth gap discovered, not fixed here**: the shared `global-setup.ts`/`auth.helper.ts` login flow doesn't dismiss a Transcend consent banner on the dev Auth0 tenant's hosted login page, so `yarn e2e` against dev currently can't auto-authenticate out of the box (worked around locally with a throwaway script to verify this PR's spec). Worth a follow-up fix in the shared helper — out of scope here since it's cross-cutting test infra, not this PR's responsive fixes.

JIRA: LFXV2-3285

## Test plan
- [x] `yarn lint`
- [x] `yarn check-types`
- [x] `yarn build`
- [x] `apps/lfx-one/e2e/profile-mobile-layout.spec.ts` — 54/54 passing (chromium/firefox/mobile-chrome), live-verified red against pre-fix and against each intermediate regression caught during review iteration
- [x] Manual check at mobile viewports (390/768/1024/1500px) for both sparse (name+email only) and populated profiles — panel reflows correctly at each of the three states, no dead space or orphaned cells
- [x] Manual check at `2xl`+ (≥1440px) — fixed right rail still pins right exactly as before
- [ ] Manual check of account-settings TOC (hidden below `lg`, visible at `lg`+)
- [ ] Manual check of affiliation-timeline-dialog period rows at a mobile width — wraps without overflow, Start/End legible
- [ ] Manual check of work-experience-form-dialog date pickers at 375px — stacked, not squeezed

### DIFF
--- a/apps/lfx-one/e2e/profile-mobile-layout.spec.ts
+++ b/apps/lfx-one/e2e/profile-mobile-layout.spec.ts
@@ -98,9 +98,13 @@ const VIEWPORTS = [
   { name: 'desktop-narrow', width: 1439, height: 900 },
 ] as const;
 
-// The panel's own max-w-md cap (below sm only) — sm:max-w-none removes it from sm up.
+// The panel's own max-w-md cap (below sm only) — sm:max-w-none removes it from sm up. The cap's
+// pixel value is read off the element's computed style at assertion time (see below) rather than
+// hard-coded here: apps/lfx-one/src/styles.scss sets `html { font-size: 14px }`, so max-w-md
+// (28rem) actually renders at 392px in this app, not the browser-default-root 448px a naive
+// rem-to-px conversion would suggest — a hard-coded constant would silently drift from the real
+// cap on any root-font-size or Tailwind maxWidth change.
 const SM_BREAKPOINT = 640;
-const MAX_MD_WIDTH = 448;
 
 // Desktop control: at and above 2xl, the rail must be the fixed overlay again — the regression
 // guard for the rail disappearing entirely rather than just moving breakpoints. Requested a margin
@@ -205,15 +209,20 @@ test.describe('Profile & Account hub — mobile/tablet/laptop layout (LFXV2-3285
         const wrapperPosition = await panelRail.evaluate((el) => getComputedStyle(el).position);
         expect(wrapperPosition, `profile panel rail position at ${vp.width}px should not be fixed`).not.toBe('fixed');
 
-        // Below sm, the rail wrapper is capped at max-w-md (448px) so it stays a proportioned stacked
-        // card instead of stretching wide while its internals are still fully stacked (a content
-        // column can exceed 448px below the 640px sm breakpoint once page padding is this small, e.g.
-        // ~510px at a 550px viewport). This regression was introduced and reverted twice on this
-        // branch before this assertion existed.
-        if (vp.width < SM_BREAKPOINT) {
+        // Below sm, the rail wrapper is capped at max-w-md so it stays a proportioned stacked card
+        // instead of stretching wide while its internals are still fully stacked (the content column
+        // can exceed the cap below the 640px sm breakpoint once page padding is this small — this app
+        // renders max-w-md at 392px, not the naive 448px, because of the 14px root font-size noted
+        // above). This regression was introduced and reverted twice on this branch before this
+        // assertion existed. Gated on clientWidth, not vp.width, for the same scrollbar reason as the
+        // md/lg checks above — a classic desktop scrollbar can put clientWidth on the other side of
+        // SM_BREAKPOINT from vp.width.
+        if (overflow.clientWidth < SM_BREAKPOINT) {
           const railBox = await panelRail.boundingBox();
           expect(railBox, 'profile panel rail should have a bounding box').not.toBeNull();
-          expect(railBox!.width, `profile panel rail width at ${vp.width}px should be capped at max-w-md`).toBeLessThanOrEqual(MAX_MD_WIDTH + 1);
+          const capPx = await panelRail.evaluate((el) => parseFloat(getComputedStyle(el).maxWidth));
+          expect(capPx, `profile panel rail max-width at ${vp.width}px should be a real px cap, not "none"`).toBeGreaterThan(0);
+          expect(railBox!.width, `profile panel rail width at ${vp.width}px should be capped at max-w-md (${capPx}px)`).toBeLessThanOrEqual(capPx + 1);
         }
 
         // Inline placement: the panel should be left-aligned with the content column (both children of
--- a/apps/lfx-one/e2e/profile-mobile-layout.spec.ts
+++ b/apps/lfx-one/e2e/profile-mobile-layout.spec.ts
@@ -27,8 +27,9 @@
  * Drives an explicit viewport matrix (mirrors e2e/docs/responsive.spec.ts) covering mobile and
  * tablet widths up to just under the `lg` breakpoint, rather than relying on whichever project
  * happens to run it — so the tablet band (768-1023px) is exercised too, not just mobile-chrome's
- * fixed 393px. Because that matrix forces its own viewport per test, it also runs under the
- * chromium/firefox desktop projects now (not just mobile-chrome) — so width comparisons read
+ * fixed 393px. This spec runs under all three playwright.config.ts projects (chromium, firefox,
+ * mobile-chrome — none scope out e2e/**); because the matrix forces its own viewport per test, the
+ * desktop projects exercise the sub-lg bands too, so width comparisons read
  * document.documentElement.clientWidth rather than the device viewport width, since desktop
  * projects render a classic scrollbar that mobile-emulated overlay scrollbars don't.
  *
--- a/apps/lfx-one/e2e/profile-mobile-layout.spec.ts
+++ b/apps/lfx-one/e2e/profile-mobile-layout.spec.ts
@@ -82,8 +82,8 @@ const TWO_XL_BREAKPOINT = 1440;
 const CONTENT_MAX = 1024;
 
 // Sub-2xl matrix: narrow phone, the mobile-chrome project's own size (Pixel 5), a landscape-phone
-// band between 393 and the sm breakpoint (see MAX_MD_WIDTH below — this is the one viewport where
-// the panel's own max-w-md cap is load-bearing; regressed and reinstated twice on this branch
+// band between 393 and the sm breakpoint (see the max-w-md cap assertion below — this is the one
+// viewport where the panel's own cap is load-bearing; regressed and reinstated twice on this branch
 // without a dedicated assertion before this entry was added), the tablet band, iPad landscape, a
 // split-screen laptop window, and just under the 2xl breakpoint itself — every band the inline-card
 // layout now needs to hold up across.
@@ -98,13 +98,16 @@ const VIEWPORTS = [
   { name: 'desktop-narrow', width: 1439, height: 900 },
 ] as const;
 
-// The panel's own max-w-md cap (below sm only) — sm:max-w-none removes it from sm up. The cap's
-// pixel value is read off the element's computed style at assertion time (see below) rather than
-// hard-coded here: apps/lfx-one/src/styles.scss sets `html { font-size: 14px }`, so max-w-md
-// (28rem) actually renders at 392px in this app, not the browser-default-root 448px a naive
-// rem-to-px conversion would suggest — a hard-coded constant would silently drift from the real
-// cap on any root-font-size or Tailwind maxWidth change.
 const SM_BREAKPOINT = 640;
+
+// max-w-md is 28rem. apps/lfx-one/src/styles.scss sets `html { font-size: 14px }`, so it renders at
+// 392px in this app, not the browser-default-root 448px a naive rem-to-px conversion would suggest.
+// The expected cap below is computed from this rem value times the *actual* root font-size read at
+// assertion time (immune to a root-font-size change), not read off the tested element's own computed
+// max-width — reading the "expected" value off the same element being asserted on would make the
+// check tautological (box-sizing: border-box guarantees a rendered width can never exceed its own
+// max-width, so that assertion could never fail regardless of which class were actually applied).
+const MAX_MD_REM = 28;
 
 // Desktop control: at and above 2xl, the rail must be the fixed overlay again — the regression
 // guard for the rail disappearing entirely rather than just moving breakpoints. Requested a margin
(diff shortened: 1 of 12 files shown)


audigregorie (commented):
## Code Review Summary

Clean, well-iterated PR. The mobile layout fix is correct — the `lg:`-gated `fixed` rail + `lg:pr-[300px]` gutter on `<main>` preserves the desktop behavior exactly while letting the panel flow inline below `lg`. The three additional responsive fixes (account-settings TOC, affiliation-timeline period rows, work-experience date grid) are scoped and independent. The e2e spec is mechanism-level (asserts `position !== 'fixed'` on a dedicated `profile-panel-rail` testid rather than relying on bounding-box overlap) and scrollbar-aware (reads `clientWidth`, not `vp.width`). No Critical or Major findings.

### Minor — outside the diff

- **e2e auth gap on dev (acknowledged in PR body)**: the shared `global-setup.ts`/`auth.helper.ts` doesn't dismiss the Transcend consent banner on the dev Auth0 tenant, so `yarn e2e` against dev can't auto-authenticate out of the box. The author worked around it locally with a throwaway script. Worth a follow-up ticket in the shared helper since it blocks anyone else running this spec against dev.

### What's done well

- **Mechanism-level testing, not just geometry** — the spec asserts `getComputedStyle(el).position !== 'fixed'` on a dedicated `profile-panel-rail` testid, and the comment block documents *why* bounding-box checks alone fail to discriminate the pre-fix layout (a fixed rail pinned right still sits above the nav and, once the content column is squeezed thin enough, doesn't geometrically overlap it either). The reviewer-driven iteration (replacing the `parentElement!.parentElement!` hop with the dedicated testid, fixing the scrollbar-blind width comparison) shows real red/green discipline.
- **Scrollbar-aware width assertions** — reads `document.documentElement.clientWidth` instead of comparing against `vp.width` directly, with a documented explanation of why desktop Chromium/Firefox classic scrollbars would false-fail a naive comparison.
- **A11y disambiguation on wrapped rows** — the `ariaLabel` interpolation (`'Start month, ' + org.organization + ' period ' + ($index + 1)`) makes wrapped/duplicate period rows identifiable to screen readers; `aria-hidden="true"` on the decorative trash icon is correct; unconditional Start/End labels (not `sm`-gated) keep the pair legible regardless of whether wrap is triggered by viewport or dialog width.
- **Honest trade-off documentation** — the PR body explicitly flags the `lg`/`xl` band density, the `p-select` vs `lfx-select` reactive-forms debt, the dialog-width/`sm`-breakpoint mismatch, and the dev-env e2e auth gap — each with a stated reason for deferring.

### Rating: ⭐⭐⭐⭐⭐ (5/5)

apps/lfx-one/e2e/profile-mobile-layout.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+const MD_BREAKPOINT = 768;
+
+// Sub-lg matrix: narrow phone, the mobile-chrome project's own size (Pixel 5), and both ends of
+// the tablet band up to just under the lg breakpoint (1024px) the layout fix keys off.
+const VIEWPORTS = [
+  { name: 'mobile-narrow', width: 360, height: 640 },
+  { name: 'mobile', width: 393, height: 727 },
+  { name: 'tablet', width: 768, height: 1024 },
+  { name: 'tablet-lg', width: 1023, height: 768 },
+] as const;
```
audigregorie, line 110: **Minor — suggest an `lg+` viewport to symmetrically lock in the fixed-rail mechanism**

The matrix covers 360–1023px (sub-`lg` only) and asserts `position !== 'fixed'` on the rail wrapper. The `lg+` fixed-rail behavior — the mechanism this PR preserves — isn't asserted here. A future regression that drops the `lg:fixed` class (e.g., someone simplifies the wrapper to `max-w-md` only) would leave this spec green while breaking the desktop layout.

Consider adding one `lg` viewport (e.g., `{ name: 'desktop', width: 1280, height: 800 }`) with a mirrored assertion: `expect(wrapperPosition).toBe('fixed')` and `expect(panelBox!.x).toBeGreaterThan(contentBox!.x)` (rail sits right of the content column). This symmetrically pins both sides of the breakpoint-gated mechanism.

```suggestion
  { name: 'tablet-lg', width: 1023, height: 768 },
  { name: 'desktop', width: 1280, height: 800 },
] as const;
```

(The `desktop` case would need its own assertion branch — `toBe('fixed')` rather than `not.toBe('fixed')` — since the current test body assumes sub-`lg` geometry throughout. Happy with a separate `test.describe` block for the `lg+` case if that reads cleaner than a viewport-conditional inside one test.)

manishdixitlfx: Good catch — at the time you flagged this, the matrix only covered sub-`lg` widths with no reciprocal fixed-rail assertion. A separate commit already on this branch (before this reply, `d903c92b`) added exactly the symmetric check you described — a `desktop control` test asserting `position === 'fixed'` and `width === 300px` on the `profile-panel-rail` testid. One difference from your suggestion: the breakpoint itself moved from `lg` (1024px) to `2xl` (1440px) in that same commit — a follow-up requirement (712px of fixed chrome at `lg` left only a 312px content column, narrower than the rail beside it), so the desktop-control viewport is 1480px rather than 1280px.

In commit `638405299`, per this session's own reviewer trio, I also fixed two related gaps that trio caught in that fixed-rail assertion: the control viewport sat exactly at the breakpoint (1440px), which a classic desktop scrollbar's effect on `clientWidth` could push below the `min-width` media-query threshold — widened it and added an explicit precondition assertion so a future failure there reads clearly instead of a confusing mismatch.

manishdixitlfx:
## Review Feedback Addressed

Commit: 6384052994fc30154d5c50a5adc6680a22291126

### Changes Made
- **apps/lfx-one/e2e/profile-mobile-layout.spec.ts**: `TWO_XL_BREAKPOINT` was declared but never used — now drives `DESKTOP_VIEWPORT.width` and a new precondition assertion (per copilot-pull-request-reviewer)
- **apps/lfx-one/e2e/profile-mobile-layout.spec.ts**: desktop-control viewport sat exactly at the 2xl breakpoint (1440px), where a classic desktop scrollbar's effect on `clientWidth` could push the effective layout width below the `min-width` media-query threshold — widened to `TWO_XL_BREAKPOINT + 40` and added an explicit precondition assertion (per @audigregorie's underlying concern, refined by this session's own reviewer trio)
- **apps/lfx-one/e2e/profile-mobile-layout.spec.ts**: the sidebar/page-padding breakpoint checks compared against the device viewport instead of the already-computed layout viewport (`clientWidth`) — now consistent with the rest of the spec's scrollbar-safety reasoning
- **apps/lfx-one/src/app/modules/profile/components/affiliation-timeline-dialog/affiliation-timeline-dialog.component.html**: reworded a comment that said the month/year grid itself wraps to two rows — `grid-cols-2` always keeps two columns side by side; it's the outer `flex-wrap` row that puts each pair on its own line (per copilot-pull-request-reviewer)
- **apps/lfx-one/src/app/modules/profile/components/affiliation-timeline-dialog/affiliation-timeline-dialog.component.html**: grouped each Start/End label with its month/year grid into a single flex item, so `flex-wrap` on the row can only move the whole pair together — previously they were separate flex items and an intermediate dialog width could wrap the grid onto the next line while leaving the label behind (per copilot-pull-request-reviewer)
- **docs/architecture/frontend/styling-system.md**: the documented `2xl` breakpoint (1536px) went stale the moment `tailwind.config.js`'s `theme.screens` was overridden to 1440px earlier on this branch — updated, with a note on `theme.container.screens` vs `theme.screens` (the exact trap that earlier commit fell into)

### Threads Resolved
1 of 1 unresolved threads addressed (@audigregorie's request for a symmetric fixed-rail assertion — see the thread reply for detail on how it's covered, including the breakpoint move to `2xl`).

### Also addressed (Copilot review-body suppressed comments, no separate threads created)
- \"No new e2e coverage\" trade-off flagged as stale — already corrected in an earlier PR-description update on this branch, before that comment
- Two duplicate \">=1024px fixed-rail assertion\" suggestions — same as the @audigregorie thread above, covered by the desktop-control test

### Verification
Live-verified after these fixes: the full 16-case chromium matrix for `profile-mobile-layout.spec.ts` still passes end-to-end (widened desktop-control viewport + new precondition assertion included). Also ran the sibling profile e2e specs (`profile-edit-drawer`, `profile-visibility-drawer`, `profile-panel-github`, `profile-avatar`) under chromium to check for collateral breakage from the `lg`→`2xl` rail move — found 2 pre-existing failures, both confirmed unrelated to this PR: one passed when re-run in isolation (resource-contention flakiness under this session's heavy parallel test load), the other fails on a `data-test=` locator attribute that doesn't exist anywhere in the app source (a pre-existing typo in that spec file).

apps/lfx-one/e2e/profile-mobile-layout.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+        // The content column must reclaim (close to) the available layout width — not be squeezed by
+        // an unconditional right gutter reserved for a rail that, below 2xl, no longer renders as a
+        // fixed overlay. Page padding is px-5 (40px total) below md, md:px-8 (64px total) at md and up.
+        // Once the viewport crosses lg (the sidebar's own, unrelated breakpoint), the 348px sidebar
+        // also reserves space, and the content column stops growing past max-w-[1024px] regardless.
+        // Both breakpoint checks key off clientWidth, not vp.width — the media queries the app
+        // actually evaluates run against the layout viewport, which a classic desktop scrollbar
+        // shrinks below the requested device width (same reasoning as the scrollWidth check above).
+        const pagePadding = overflow.clientWidth >= MD_BREAKPOINT ? 64 : 40;
+        const sidebar = overflow.clientWidth >= LG_BREAKPOINT ? SIDEBAR_WIDTH : 0;
```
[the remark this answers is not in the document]
manishdixitlfx: Fixed — the md/lg breakpoint decisions driving the content-width expectation now read `window.matchMedia()\'s `.matches` directly instead of comparing `document.documentElement.clientWidth` against the breakpoint constants, so the assertion no longer has to guess which viewport metric a `min-width` query resolves against.

Could not locally reproduce a `clientWidth`/`innerWidth` divergence under a forced vertical scrollbar in headless Chromium or Firefox (both read 768/768 at the flagged viewport here), but the fix stands regardless — asking the engine for its own breakpoint state is strictly more robust than inferring it from a metric that may not match what the engine itself uses.

See commit c134987cf: full 54-case matrix (chromium/firefox/mobile-chrome) passes, including the exact Firefox @ 768px case.

manishdixitlfx:
## Review Feedback Addressed

Commit: c134987cf

### Changes Made
- **apps/lfx-one/e2e/profile-mobile-layout.spec.ts**: the md/lg breakpoint decisions driving the content-width expectation now read `window.matchMedia()` directly instead of comparing `document.documentElement.clientWidth` against the breakpoint constants — the matrix's own tablet viewport sits exactly on `MD_BREAKPOINT` (768px), so this removes any ambiguity about which viewport metric a `min-width` query resolves against (per cursor[bot])
- **apps/lfx-one/e2e/profile-mobile-layout.spec.ts**, **apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.html**: fixed stale fixed-chrome math in both header doc comments — `md:px-8` is 56px total at this app's 14px root, not 64px, correcting 712px/312px/728px to 704px/320px/736px (per copilot-pull-request-reviewer)
- PR description rewritten to describe the current `2xl` breakpoint and the tablet-reflow work, which the prior description omitted entirely (per copilot-pull-request-reviewer)

### No Change Needed
- **profile-mobile-layout.spec.ts:127** (per copilot-pull-request-reviewer): flagged that this suite only covers the profile-rail layout, not the account-settings TOC or the two dialog responsive fixes. Real gap — deferred rather than added here, since opening/populating the affiliation-timeline and work-experience dialogs is a meaningfully larger addition than a rail regression test. Documented as an explicit trade-off in the PR description rather than silently dropped.

### Threads Resolved
1 of 1 unresolved threads addressed in this iteration (the cursor[bot] Firefox-breakpoint thread).

Verification: could not locally reproduce a `clientWidth`/`innerWidth` divergence under a forced scrollbar in headless Chromium or Firefox (both read the same value at the flagged viewport in this environment), but the `matchMedia` fix stands regardless — full 54-case matrix (chromium/firefox/mobile-chrome) passes, including the exact Firefox @ 768px case.

audigregorie (approved):

PR #1590, opened 2026-08-17
fix(dashboards): skip join-url fetch for past meetings on dashboard cards

andrest50:
## Summary

- Adds a `pastMeeting` input (default `false`) to `DashboardMeetingCardComponent` that short-circuits `canJoinMeeting()` to `false`, preventing the `POST /public/api/meetings/:id/join-url` BFF call from firing.
- Passes `[pastMeeting]="true"` in the two past-meeting card contexts: the **Last Meeting** card in `my-meetings` (all dashboard lenses) and the **Past Meeting** card in `committee-overview`.
- The "Join Meeting" button was never rendered in these contexts anyway — ITX always 403s for ended occurrences and `catchError(() => of(null))` collapses `joinUrl()` to null — but the failed request was logging a `console.error` on every dashboard load, inflating the RUM error count tracked in Datadog (app `eb09f2b2-2cd7-4a3b-9354-992278f8c43d`).
- Updated `committee-overview.lastMeeting` to filter truly ended records (same `scheduled_end_time < now` guard used in `my-meetings`) so an in-progress committee meeting is not passed `pastMeeting=true` and retains its Join button.

## Test plan

- [x] Load the dashboard while having a recent past meeting — confirm no `POST .../join-url` request fires in DevTools Network tab and no `Failed to load public meeting join url` console error appears.
- [x] Confirm the **Next Meeting** card still shows a "Join Meeting" button when a meeting is within the join window (the `pastMeeting` input defaults to `false` for that card).
- [ ] Load a committee overview with a past meeting in the "Past Meeting" section — confirm no join-url request fires.
- [ ] Load a committee overview **while a meeting is currently in-progress** — confirm the in-progress meeting does **not** appear in the "Past Meeting" card (it should still be reachable from the "Next Meeting" card while the join window is open).
- [x] Confirm "Meeting details" and "Watch recording" buttons still render and function correctly on past-meeting cards.

Fixes: LFXV2-3219

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.html
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.html
@@ -188,7 +188,7 @@
               </div>
             </lfx-card>
           } @else if (lastMeeting(); as meeting) {
-            <lfx-dashboard-meeting-card [meeting]="meeting" [detailUrl]="'/meetings/' + meeting.id" />
+            <lfx-dashboard-meeting-card [meeting]="meeting" [detailUrl]="'/meetings/' + meeting.id" [pastMeeting]="true" />
           } @else {
             <lfx-card>
               <div class="flex flex-col items-center py-6">
--- a/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
@@ -297,8 +297,25 @@ export class CommitteeOverviewComponent {
   });
 
   public lastMeeting: Signal<PastMeeting | null> = computed(() => {
-    const past = [...this.pastMeetings()].sort((a, b) => (b.start_time ?? '').localeCompare(a.start_time ?? ''));
-    return past[0] ?? null;
+    // v1_past_meeting includes meetings as soon as they START, so filter to truly ended
+    // records before returning — same guard used in my-meetings.component.ts.
+    const now = Date.now();
+    return (
+      [...this.pastMeetings()]
+        .sort((a, b) => (b.start_time ?? '').localeCompare(a.start_time ?? ''))
+        .find((meeting) => {
+          if (meeting.scheduled_end_time) {
+            const scheduledEnd = new Date(meeting.scheduled_end_time).getTime();
+            if (!Number.isNaN(scheduledEnd)) return scheduledEnd < now;
+          }
+          const startIso = meeting.scheduled_start_time ?? meeting.start_time;
+          if (!startIso) return false;
+          const startMs = new Date(startIso).getTime();
+          const duration = Number(meeting.duration);
+          if (Number.isNaN(startMs) || Number.isNaN(duration)) return false;
+          return startMs + duration * 60 * 1000 < now;
+        }) ?? null
+    );
   });
 
   public constructor() {
--- a/apps/lfx-one/src/app/modules/dashboards/components/dashboard-meeting-card/dashboard-meeting-card.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/components/dashboard-meeting-card/dashboard-meeting-card.component.ts
@@ -58,6 +58,8 @@ export class DashboardMeetingCardComponent {
   public readonly overlappingCount = input<number>(0);
   /** Optional recording URL override — when set, skips the API call and renders the "Watch recording" button directly. */
   public readonly recordingUrl = input<string | null>(null);
+  /** When true, suppresses the join-url fetch and hides the Join button. Pass for past-meeting contexts (e.g. Last Meeting card) where joining is never appropriate. */
+  public readonly pastMeeting = input<boolean>(false);
 
   public readonly joinUrl: Signal<string | null>;
 
@@ -177,6 +179,10 @@ export class DashboardMeetingCardComponent {
 
   private initCanJoinMeeting(): Signal<boolean> {
     return computed(() => {
+      if (this.pastMeeting()) {
+        return false;
+      }
+
       const meeting = this.meeting();
 
       if (meeting.restricted && !meeting.invited) {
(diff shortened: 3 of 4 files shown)


apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.html
```diff
@@ -188,7 +188,7 @@ <h2 class="text-sm font-medium text-gray-700 flex items-center gap-2">
               </div>
             </lfx-card>
           } @else if (lastMeeting(); as meeting) {
-            <lfx-dashboard-meeting-card [meeting]="meeting" [detailUrl]="'/meetings/' + meeting.id" />
+            <lfx-dashboard-meeting-card [meeting]="meeting" [detailUrl]="'/meetings/' + meeting.id" [pastMeeting]="true" />
```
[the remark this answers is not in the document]
andrest50: Fixed in 9d52b7faa — updated `lastMeeting` in `committee-overview.component.ts` to apply the same end-time guard used in `my-meetings.component.ts` (filter by `scheduled_end_time < now`, with `start + duration` fallback). In-progress meetings are now excluded before `[pastMeeting]="true"` is applied, so a currently active committee meeting will correctly remain eligible for the Join button.

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
```diff
@@ -297,8 +297,25 @@ export class CommitteeOverviewComponent {
...
+            if (!Number.isNaN(scheduledEnd)) return scheduledEnd < now;
+          }
+          const startIso = meeting.scheduled_start_time ?? meeting.start_time;
+          if (!startIso) return false;
+          const startMs = new Date(startIso).getTime();
+          const duration = Number(meeting.duration);
+          if (Number.isNaN(startMs) || Number.isNaN(duration)) return false;
+          return startMs + duration * 60 * 1000 < now;
+        }) ?? null
+    );
```
audigregorie, line 318: **Minor — Test plan gap for the in-progress filter**

The PR's test plan covers the join-url suppression (commit 1) but has no case for this new end-time filter (commit 2). Consider adding a case like:

> Load a committee overview while a meeting is currently in-progress — confirm it does **not** appear on the Past Meeting card (it should still be reachable from Next Meeting while the join window is open).

This is the behavioral change that motivated commit 2 (per its message: an active meeting could otherwise be returned here and then lose its Join button due to `pastMeeting=true`). Worth a manual check item so a reviewer/QA can confirm the filter is actually excluding in-progress occurrences.

audigregorie (commented):
## Code Review Summary

The code logic is sound — the `pastMeeting` input is the correct fix (since `canJoinMeeting()` returns true for up to 40 min after a meeting ends, the flag is needed to suppress the fetch in that window), and commit 2's end-time guard correctly mirrors `my-meetings.component.ts`. However, the PR has two critical PR-shape issues that need to be addressed before merge.

### Critical — outside the diff

1. **`Co-authored-by: Cursor` trailers on both commits** — Both commit messages contain `Co-authored-by: Cursor <cursoragent@cursor.com>`, which violates the repo's `commit-flow.mdc` and `global-rules.mdc` rules ("NEVER add `Co-authored-by: ...` or any AI co-author trailer"). Repo policy is human authorship only. The commits need rewriting (amend or `git commit-tree -S`) to remove the trailers while preserving DCO sign-off + GPG.

2. **Branch is not rebased on `main` — merging would revert PR #1589** — `git diff main..HEAD` shows `docs/user/committees/faq/index.md` reverting, but neither commit on this branch touches that file. `main` has commit `5d20c33d6 docs(committees): correct answer for missing groups in My Groups (#1589)` which this branch hasn't been rebased onto. Merging this PR as-is would revert PR #1589's docs FAQ update. Rebase onto `main` before merging.

### Major — outside the diff

3. **PR description doesn't mention commit 2** — The PR description only describes the `pastMeeting` input (commit `f352cfb21`). Commit `9d52b7faa` adds an end-time filter to `committee-overview.lastMeeting` so in-progress meetings aren't returned (and thus don't lose their Join button due to `pastMeeting=true`). This is a meaningful behavioral change that should be described in the PR summary and covered by the test plan.

4. **Commit subjects missing `(LFXV2-3219)` ticket suffix** — Per `global-rules.mdc`, the ticket must be extracted from the branch name (`fix/LFXV2-3219-...`) and appended to the commit subject. Neither subject has it. Commit 1 has `Fixes: LFXV2-3219` in the body, but the rule requires it in the **subject**. Also note commit 2's subject is 84 chars (over the ≤72 team-style target); adding the suffix pushes it to ~96 — still under the 100 hard-fail, but a SHOULD_FIX per the PR-shape check.

### What's done well

- The `pastMeeting` input is the correct fix: `canJoinMeeting()` returns `true` for up to 40 minutes *after* a meeting ends (start + duration + 40min buffer), so without the flag the fetch would still fire for recently-ended meetings even after commit 2's `scheduledEnd < now` filter. The flag correctly suppresses both the fetch (constructor's `!this.canJoinMeeting()` guard) and the button (template's `canJoinMeeting() && joinUrl()`).
- Commit 2's end-time guard correctly mirrors the established pattern in `my-meetings.component.ts` (`initLastMeeting`), including the `scheduled_end_time` → `start + duration` fallback and `Number.isNaN` defensive checks. Keeps both lenses consistent.
- The `initCanJoinMeeting` early-return on `pastMeeting()` is placed before `this.meeting()`, cleanly short-circuiting before any work.

---
_Review by `ay-review-their-pr` skill._

apps/lfx-one/src/app/modules/committees/components/committee-overview/committee-overview.component.ts
andrest50: Good catch — added to the test plan: "Load a committee overview **while a meeting is currently in-progress** — confirm the in-progress meeting does **not** appear in the \"Past Meeting\" card (it should still be reachable from the \"Next Meeting\" card while the join window is open)." PR description updated.

audigregorie (approved):

PR #1592, opened 2026-08-17
test(profile): cover visibility drawer cascade / auto-save

fayazg:
## Summary

- Add component unit tests for `ProfileVisibilityDrawerComponent`, whose ~400-line cascade / auto-save state machine previously had only E2E coverage (LFXV2-2629).
- Covers `wireCascade`: going public enables sections + defaults the basic group on; going private zeroes and disables all; the `basic` parent mirrors to its children and ORs back on from them.
- Covers `seedForm`: fetched public/private state patches values and aligns enable/disable; a load failure gates the form off and surfaces an error toast.
- Covers `buildPayload`: `getRawValue()` serializes disabled controls and every section key survives; the close-time flush persists the resolved cascade payload.
- Template is overridden empty so the class logic runs isolated from the drawer/toggle/select children — no E2E harness dependency.

Ref: LFXV2-2629

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/components/profile-visibility-drawer/profile-visibility-drawer.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/profile/components/profile-visibility-drawer/profile-visibility-drawer.component.spec.ts
@@ -1,0 +1,182 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { PLATFORM_ID } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { PROFILE_VISIBILITY_DEFAULTS, PROFILE_VISIBILITY_KEYS } from '@lfx-one/shared/constants';
+import { ProfileVisibility, ProfileVisibilitySections, ProfileVisibilityUpdateRequest } from '@lfx-one/shared/interfaces';
+import { UserService } from '@services/user.service';
+import { MessageService } from 'primeng/api';
+import { of, throwError } from 'rxjs';
+import { beforeEach, describe, expect, it, Mock, vi } from 'vitest';
+
+import { ProfileVisibilityDrawerComponent } from './profile-visibility-drawer.component';
+import { ProfileVisibilityDrawerService } from './profile-visibility-drawer.service';
+
+/**
+ * Guards the cascade / auto-save state machine of the public-profile visibility drawer (LFXV2-2629):
+ * the parent↔child cascade, seedForm's enable/patch alignment, and buildPayload's raw serialization.
+ * Template is overridden empty so the class logic runs without the drawer/toggle/select children.
+ */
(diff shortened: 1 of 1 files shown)


audigregorie (approved):

MRashad26 (approved):
Reviewed with the code-standards-enforcer checklist. Test-only PR (new `.spec.ts` file, no production code touched), so the review focused on whether the assertions actually match the real cascade/auto-save logic rather than just reading plausibly.

**What this does:** Adds component-level unit tests for `ProfileVisibilityDrawerComponent`'s cascade/auto-save state machine (LFXV2-2629), which previously only had E2E coverage. Covers `wireCascade` (public/private toggling, basic↔children mirroring and OR-back), `seedForm` (patch + enable/disable alignment, load-failure gating), and `buildPayload` (raw-value serialization of disabled controls, close-time flush).

**Verified correct against the actual component source (not just the PR description):**
- **`wireCascade` assertions are accurate.** Traced the real subscriptions in `profile-visibility-drawer.component.ts`: `isPublic=true` calls `setSectionsEnabled(true)` and sets only `PROFILE_VISIBILITY_PUBLIC_DEFAULT_KEYS` true — confirmed against `profile-visibility.constants.ts`, that's exactly `['basic', 'aboutMe', 'personalInfo']`, matching the test's expectation that `badges` stays false but becomes enabled. `isPublic=false` zeroes and disables every `PROFILE_VISIBILITY_KEYS` entry — matches. The `basic`↔child mirror/OR-back tests match the two separate subscriptions (`basic` → children via `emitEvent:false`, each child → `basic` via an OR check) exactly as implemented.
- **`seedForm` assertions match**: `patchValue(..., {emitEvent:false})` + `setSectionsEnabled(isPublic)` + `isPublic.set(...)`, and the load-failure path (`catchError` sets `loadError`, calls `messageService.add` with `severity: 'error'`, `finalize` clears `loadingVisibility`) is a faithful match of the constructor's `merge(open$, retry$).pipe(switchMap(...))` subscription.
- **`buildPayload` assertions match**: it does use `getRawValue()` (not `.value`), so a disabled-but-true control is correctly asserted to survive into the payload — this is the one behavior in the class that would be easy to silently regress (e.g. swapping to `.value`) without a test catching it.
- **Mock signatures are correct**: `UserService.getProfileVisibility()` / `updateProfileVisibility(data)` mocks match the real service's signatures exactly (verified against `user.service.ts`).
- **Template-override-to-empty technique is applied correctly** for isolating class logic from `ToggleComponent`/`SelectButtonComponent`/`DrawerModule` children — appropriate given this component has no rendering logic under test here.
- Provider list correctly includes `MessageService` (the component injects it directly) and `PLATFORM_ID` (used by `initPublicProfileUrl`/`onCopyUrl`).

No console.log, no template rule violations (no template changes at all), no shared-interface placement issues (imports already come from `@lfx-one/shared/{constants,interfaces}`). No critical or major issues found. Approving.

PR #1596, opened 2026-08-17
feat(campaigns): build the email template picker

mrautela365:
Replaces the "Email staging is not wired up yet" panel with the template picker it was waiting for.

Jira: [LFXV2-3198](https://linuxfoundation.atlassian.net/browse/LFXV2-3198)

## The copy it removes was false

That panel said choosing a template "needs an endpoint that is still in review". Both halves have been live for a while:

- `GET /api/campaigns/hubspot/emails` — merged in **#1439**
- `buildHubSpotConfig` (the `hubspotConfig` builder) — merged in **#1546**

Staging an email **clones** one of the project's HubSpot marketing emails, and `sourceEmailId` has no default, so a chosen template is **required** before a send can be staged.

**It is not yet sufficient, and an earlier version of this description said it was.** There is no create trigger on the email side: the only `createCampaign` call site is in the implementation tab, which renders solely under Paid, and `CampaignPlatform` has no `hubspot` member — so the type system forbids this path from reaching create even if a button existed. Every `hubspotConfig` reference in the app was a comment or a spec; the server side (`buildHubSpotConfig`) is built and waiting.

This PR threads the selection through to `hubspotConfig.sourceEmailId` so the value flows the moment a trigger lands, and nothing binds it — verified by sweeping every template. **No send can be staged and nothing can spend.** Wiring the trigger is LFXV2-3201.

## Four states, kept distinct

Each pair would state something false if collapsed:

| State | Renders | Why it matters |
|---|---|---|
| **not connected** (`enabled: false`) | "connect HubSpot" | The steady state wherever HubSpot isn't set up — not an error |
| **search failed** | nothing; list stays `null` | An empty array asserts the portal holds no templates. Only a completed search supports that, and the claim sends someone to create one they may already have |
| **searched-and-empty** | `[]` with copy naming the query | Reads as a statement about the *search*, not the portal |
| **results** | the list | Rows with no `id` are **dropped** — `id` is what `sourceEmailId` takes, so such a row is a choice that cannot be made |

`possiblyTruncated` surfaces only for an **empty** query — the one case where a complete listing and a truncated first screen are indistinguishable on the wire. A filtered search is complete-or-error upstream, so warning there would send someone to narrow a search that was already exhaustive.

## Verification

`yarn build`, `yarn lint:check --force`, `./check-headers.sh`. **1340 server tests + 73 app-side tests.**

Mutation-tested — each guard pinned by a test that fails when reverted:

| Mutation | Fails |
|---|---|
| collapse a failed search into `[]` | failure-stays-null |
| keep rows with no `id` | id-less-row-dropped |
| drop the foundation guard | refuses-without-foundation |

**Replaces a test rather than deleting it.** The old one pinned the stale panel's copy; its real concern — don't claim a brief is ready when none exists — survives, and the picker is stronger on that point because it states nothing about the brief at all.

**Not covered: layout.** The specs bind every branch where the component decides *what to say*, not how it looks. The local dev environment needs an OAuth consent grant that is the account owner's to give.

[LFXV2-3198]: https://linuxfoundation.atlassian.net/browse/LFXV2-3198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -254,8 +254,19 @@
                   -->
                   <p class="text-xs text-amber-700" data-testid="campaigns-email-templates-truncated">This may be a partial list. Search to narrow it.</p>
                 }
+                @if (emailTemplatesRenderCapped()) {
+                  <!--
+                    Stated as FACT, unlike the `possiblyTruncated` banner above, which hedges
+                    because a capped 500 and a complete 500 are indistinguishable on the wire.
+                    Both numbers here were measured on the list in hand before it was sliced.
+
+                    Both can show at once: an unfiltered search can be cut upstream at 500 and
+                    cut again here at the render limit.
+                  -->
+                  <p class="text-xs text-amber-700" data-testid="campaigns-email-templates-render-capped">{{ emailTemplatesRenderCapMessage() }}</p>
+                }
                 <ul class="flex flex-col gap-2" data-testid="campaigns-email-template-list">
-                  @for (template of templates; track template.id) {
+                  @for (template of emailTemplatesRendered(); track template.id) {
                     <li>
                       <button
                         type="button"
@@ -268,9 +279,33 @@
                             : 'flex w-full flex-col gap-0.5 rounded-md border border-gray-200 px-3 py-2 text-left hover:bg-gray-50'
                         "
                         [attr.data-testid]="'campaigns-email-template-' + template.id">
-                        <span class="text-sm font-medium text-gray-900">{{ template.name || template.subject || template.id }}</span>
+                        <span class="flex w-full items-center gap-2">
+                          <span class="min-w-0 flex-1 text-sm font-medium text-gray-900">{{ template.name || template.subject || template.id }}</span>
+                          <!--
+                            The non-colour half of the selected state (WCAG 1.4.1): the border and
+                            background above are colour ONLY, so a colour-blind user, or anyone on
+                            a low-contrast display, had nothing to go on. Same treatment as
+                            project-selector and org-selector — an opacity-toggled check rather
+                            than an @if, so the row's width does not shift as the selection moves.
+
+                            aria-hidden while unselected keeps it out of the accessible name;
+                            aria-pressed on the button already tells assistive tech the state, so
+                            the label would otherwise be announced twice.
+                          -->
+                          <i
+                            class="fa-light fa-check shrink-0 text-blue-500 transition-opacity"
+                            [class.opacity-100]="selectedEmailTemplateId() === template.id"
+                            [class.opacity-0]="selectedEmailTemplateId() !== template.id"
+                            [attr.aria-hidden]="selectedEmailTemplateId() !== template.id"
+                            [attr.aria-label]="selectedEmailTemplateId() === template.id ? 'Selected' : null"></i>
+                        </span>
                         <span class="text-xs text-gray-500">
-                          @if (template.subject) {
+                          <!--
+                            Only when it is not ALREADY the primary label above. With no `name`,
+                            the primary falls through to `subject`, and rendering it again here
+                            printed the same string twice on one row.
+                          -->
+                          @if (template.subject && template.name) {
                             {{ template.subject }}
                           }
                           @if (template.state) {
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -236,7 +236,7 @@
                     first screen are indistinguishable. Telling someone to narrow an already
                     exhaustive search would send them hunting for a template that does not exist.
                   -->
-                  <p class="text-xs text-amber-700" data-testid="campaigns-email-templates-truncated">Showing a partial list. Search to narrow it.</p>
+                  <p class="text-xs text-amber-700" data-testid="campaigns-email-templates-truncated">This may be a partial list. Search to narrow it.</p>
                 }
                 <ul class="flex flex-col gap-2" data-testid="campaigns-email-template-list">
                   @for (template of templates; track template.id) {
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -221,7 +221,7 @@
               </p>
             } @else if (emailChannelEnabled() === false) {
               <!-- Not an error: the steady state wherever HubSpot is not connected. -->
-              <p class="text-sm text-gray-500" data-testid="campaigns-email-not-connected">Connect HubSpot for this foundation to stage an email.</p>
+              <p class="text-sm text-gray-500" data-testid="campaigns-email-not-connected">{{ emailNotConnectedMessage() }}</p>
             } @else if (emailTemplatesError(); as message) {
               <p class="text-sm text-red-600" role="alert" data-testid="campaigns-email-templates-error">{{ message }}</p>
             } @else if (emailTemplates(); as templates) {
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -178,6 +178,15 @@
             </div>
 
             <div class="flex items-center gap-2">
+              <!--
+                Search status. Separate from the visible states and always in the DOM: an
+                aria-live element inserted with its text already present is not reliably
+                announced, so the region has to exist first and then have its CONTENTS change.
+                Same reasoning as implementation-tab's brief-status region.
+              -->
+              <div role="status" aria-live="polite" class="sr-only" data-testid="campaigns-email-templates-live">
+                {{ emailTemplatesAnnouncement() }}
+              </div>
               <input
                 type="search"
                 [value]="emailTemplateQuery()"
@@ -236,7 +245,7 @@
                         type="button"
                         (click)="onSelectEmailTemplate(template.id)"
                         [attr.aria-pressed]="selectedEmailTemplateId() === template.id"
-                        [attr.aria-label]="emailTemplateAccessibleName(template)"
+                        [attr.aria-label]="template | hubspotTemplateLabel"
                         [class]="
                           selectedEmailTemplateId() === template.id
                             ? 'flex w-full flex-col gap-0.5 rounded-md border border-blue-500 bg-blue-50 px-3 py-2 text-left'
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -251,6 +251,9 @@
                           @if (template.state) {
                             · {{ template.state }}
                           }
+                          @if (formatTemplateUpdatedAt(template.updatedAt); as updated) {
+                            · Updated {{ updated }}
+                          }
                         </span>
                       </button>
                     </li>
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -236,7 +236,7 @@
                         type="button"
                         (click)="onSelectEmailTemplate(template.id)"
                         [attr.aria-pressed]="selectedEmailTemplateId() === template.id"
-                        [attr.aria-label]="'Use template ' + (template.name || template.subject || template.id)"
+                        [attr.aria-label]="emailTemplateAccessibleName(template)"
                         [class]="
                           selectedEmailTemplateId() === template.id
                             ? 'flex w-full flex-col gap-0.5 rounded-md border border-blue-500 bg-blue-50 px-3 py-2 text-left'
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -251,7 +251,7 @@
                           @if (template.state) {
                             · {{ template.state }}
                           }
-                          @if (formatTemplateUpdatedAt(template.updatedAt); as updated) {
+                          @if (template.updatedAt | hubspotUpdatedAt; as updated) {
                             · Updated {{ updated }}
                           }
                         </span>
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -181,7 +181,8 @@
               <input
                 type="search"
                 [value]="emailTemplateQuery()"
-                (keyup.enter)="searchEmailTemplates($any($event.target).value)"
+                (input)="onEmailTemplateQueryInput($any($event.target).value)"
+                (keyup.enter)="searchEmailTemplates(emailTemplateQuery())"
                 placeholder="Search by name or subject"
                 aria-label="Search HubSpot marketing emails"
                 class="flex-1 rounded-md border border-gray-300 px-3 py-2 text-sm"
@@ -226,9 +227,7 @@
                     first screen are indistinguishable. Telling someone to narrow an already
                     exhaustive search would send them hunting for a template that does not exist.
                   -->
-                  <p class="text-xs text-amber-700" data-testid="campaigns-email-templates-truncated">
-                    Showing the most recently updated templates. Search to narrow the list.
-                  </p>
+                  <p class="text-xs text-amber-700" data-testid="campaigns-email-templates-truncated">Showing a partial list. Search to narrow it.</p>
                 }
                 <ul class="flex flex-col gap-2" data-testid="campaigns-email-template-list">
                   @for (template of templates; track template.id) {
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
@@ -214,8 +214,8 @@
                   a statement about the search rather than about the portal.
                 -->
                 <p class="text-sm text-gray-500" data-testid="campaigns-email-templates-empty">
-                  @if (emailTemplateQuery()) {
-                    No templates match “{{ emailTemplateQuery() }}”.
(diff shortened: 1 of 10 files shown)


apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -164,32 +164,101 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+                type="search"
+                [value]="emailTemplateQuery()"
+                (keyup.enter)="searchEmailTemplates($any($event.target).value)"
+                placeholder="Search by name or subject"
+                aria-label="Search HubSpot marketing emails"
+                class="flex-1 rounded-md border border-gray-300 px-3 py-2 text-sm"
+                data-testid="campaigns-email-template-search" />
+              <button
+                type="button"
+                (click)="searchEmailTemplates(emailTemplateQuery())"
```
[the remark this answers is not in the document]
mrautela365: This is no longer accurate on the current head (f3502a6). The input is two-way now — commit 7f7fef9 added an `(input)` handler alongside the `[value]` binding:

```html
<input type="search"
  [value]="emailTemplateQuery()"
  (input)="onEmailTemplateQueryInput($any($event.target).value)"
  (keyup.enter)="searchEmailTemplates(emailTemplateQuery())" ... />
<button (click)="searchEmailTemplates(emailTemplateQuery())" ...>Search</button>
```

`onEmailTemplateQueryInput` (campaigns.component.ts:709-711) sets `emailTemplateQuery` on every keystroke, so the Search button reads what is currently typed, not the last completed query. Typing "kubecon" and clicking Search now searches "kubecon". Resolving.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -584,6 +620,69 @@ export class CampaignsComponent {
...
+    }
+
+    this.emailTemplateQuery.set(query);
+    this.emailTemplatesLoading.set(true);
+    this.emailTemplatesError.set(null);
+
+    this.campaignService
+      .searchHubSpotEmails(projectSlug, query)
+      .pipe(take(1))
+      .subscribe({
```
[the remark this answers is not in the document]
mrautela365: Addressed by commit 6f9f28706 ("clear the template picker on a foundation switch"). The constructor now subscribes to `activeFoundationSlug` and clears the whole picker on a switch (campaigns.component.ts:449-477):

```ts
this.emailTemplates.set(null);
this.emailTemplateQuery.set("");
this.emailTemplatesTruncated.set(false);
this.emailChannelEnabled.set(null);
this.emailTemplatesError.set(null);
this.selectedEmailTemplateId.set("");
```

That covers the second half of the report — stale results and, importantly, the stale `selectedEmailTemplateId` that becomes `sourceEmailId` on create. Resolving.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -164,32 +164,101 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+              } @else {
+                @if (emailTemplatesTruncated()) {
+                  <!--
+                    Only ever set for an EMPTY query, where a complete listing and a truncated
+                    first screen are indistinguishable. Telling someone to narrow an already
+                    exhaustive search would send them hunting for a template that does not exist.
+                  -->
+                  <p class="text-xs text-amber-700" data-testid="campaigns-email-templates-truncated">
+                    Showing the most recently updated templates. Search to narrow the list.
+                  </p>
```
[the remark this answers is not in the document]
mrautela365: The copy on the current head does not make the claim this describes. The truncation notice reads:

```html
<p class="text-xs text-amber-700" data-testid="campaigns-email-templates-truncated">Showing a partial list. Search to narrow it.</p>
```

No "most recent 500" and no ordering claim at all — it says only that the list may be partial, which is exactly what `possiblyTruncated` means per the shared contract. Resolving.

mrautela365:
Three findings fixed in 924ddfa8a. One is left open deliberately — see the end.

## Stale response overwrites a newer search

Correct. Each call is an independent `subscribe` with only `take(1)`, so nothing ordered them: a slow "alpha" could land after a fast "beta" and repaint the list while the query box already showed "beta". A generation counter now discards any response whose search is no longer current.

Your point about the foundation switch was the sharper half — that handler clears the picker signals but could not stop an **in-flight** response refilling them under the *new* foundation, which is exactly the cross-portal leak it exists to prevent.

The test drives both responses **out of order** (`httpMock.match`, flush the second, then the first). Ordering is the whole defect, so a test that flushes them in sequence would pass against the broken code.

## A transport failure rendered as a disconnected channel

Also correct, and the branch order is what makes it bite: the template checks `emailChannelEnabled() === false` at line 205 **before** `emailTemplatesError()` at 208. The search reset the error but never the channel flag, so after one "not connected" answer a genuine transport failure still rendered *"Connect HubSpot for this foundation"* — sending the operator to fix a connection that was fine.

Now reset to `null` alongside the error, including on the missing-foundation early return, which had the same hole.

## Rows carried no `updatedAt`

Fixed. The shared interface already says why, where the field is declared: *two templates routinely share a name — the date is what tells them apart.* Without it two same-name rows are visually identical and the operator cannot tell which one they are cloning.

Renders nothing rather than a dash when the field is absent — it is optional on the interface, and a placeholder would read as a value the portal reported.

**One thing worth recording about this test:** it first failed against correct code. The picker renders under `@switch (selectedEmailTab())` → `@case ('implementation')`, so a DOM assertion needs the email delivery type *and* that tab selected — otherwise the list is simply not in the document. The fix was to the test.

## Left open: `selectedEmailTemplateId` has no consumer

Not fixed, and I want to be explicit rather than quiet about it. You are right that it is a dead end — `createCampaign`, `buildHubSpotConfig` and `hubspotConfig` appear in this component **only inside comments**. "Use template" sets `aria-pressed` and a highlight class and nothing else.

That is a product decision, not a review fix: either wire the id into a create/staging action, or restore pending copy until that action exists. Both change what the PR claims to deliver, so it needs an owner's call rather than my guess.

Each fix is pinned by a test verified to fail when that fix alone is reverted. `lint:check` 0 errors, both typechecks clean, 82/82 tests.

mrautela365:
Three remaining picker findings closed in `fc6ebff15`, audited as one surface rather than one per push.

**Selection was colour-only.** `border-blue-500 bg-blue-50` plus `aria-pressed` served assistive tech but left a sighted user relying on contrast alone. The repo has one established convention for this, and the picker now uses it verbatim — `fa-light fa-check`, opacity-toggled, with a conditional `aria-label="Selected"` — matching `project-selector`, `org-selector` and `create-target-tree-node`. Purely additive; the colour half and `aria-pressed` are untouched.

Worth noting: several other pickers in this repo are still colour-only (`newsletter-send-step`, the escalate modal, the planning-tab ad channels). They are the non-conforming minority, so this change brings the template picker **into** line rather than establishing a new rule. They are out of scope here.

**A name-less template printed its subject twice.** The primary label falls back through `name || subject || id`, and the metadata line rendered `template.subject` under a bare `@if`. `HubSpotTemplateLabelPipe` already had this right (`if (template.subject && template.name)`) — so a screen-reader user heard the subject once while a sighted user saw it twice. The two surfaces had diverged inside the same fix chain; the template now matches the pipe.

**The list rendered unbounded.** A filtered search is exempt from the 500-row cap per `campaign.interface.ts` — "complete-or-error" over up to 200 pages — so a broad query renders a button per row into the thousands. Now capped at `HUBSPOT_TEMPLATE_RENDER_LIMIT = 100` (a shared constant, not a module-level one under `apps/`).

The cap is on the **render, not the result**: `emailTemplates` keeps every row, so the reported total stays true and the banner reads "Showing the first 100 of 137". That states the cut as fact rather than borrowing the `possiblyTruncated` hedge, because here both numbers are measured — the two banners are independent and can both appear. The cue also reaches the live region, since a screen-reader user cannot see the banner.

A render cap rather than virtualisation because nothing in this repo virtualises a list today; adding one would introduce a first pattern rather than follow one.

## Verification — 7 compiling mutations, none survived

The two worth calling out:

- **Silent truncation** — cap kept, banner removed — fails the test. I re-ran this one independently: the assertion binds on the *disclosure*, not just the cap.
- **The banner reporting the drawn count as the total** ("Showing the first 100 of 100") fails too, because the test asserts the exact sentence with both numbers rather than the banner's presence.

Both are the "assert the value, not the key" discipline: an existence check would have passed against a permanently hidden icon and against a banner that told the user nothing true.

One test the agent had to rewrite rather than weaken: its first draft asserted the named row renders the subject twice, which failed because `aria-label` is an attribute and never appears in `textContent`, so the count is one either way. Asserting the primary label and metadata line separately is what makes the mutation observable.

Gate: `lint:check` 0 errors · `format:check` · `test:app` 393 passed · `tsc --noEmit` per tsconfig.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -584,6 +828,151 @@ export class CampaignsComponent {
...
+   */
+  protected onEmailTemplateSearchSubmit(): void {
+    if (this.emailTemplatesLoading()) {
+      return;
+    }
+    this.searchEmailTemplates(this.emailTemplateQuery());
+  }
+
+  protected onSelectEmailTemplate(id: string): void {
+    this.selectedEmailTemplateId.set(id);
```
dealako, line 975: [minor] Selection persists across searches (invisible beyond the render cap)

Issue: `toggleEmailTemplate` sets `selectedEmailTemplateId`, but `searchEmailTemplates` (dispatch, ~845-876) resets `emailTemplates`/`emailChannelEnabled`/`emailTemplatesError` and never clears `selectedEmailTemplateId`. A selection therefore survives a subsequent search.
Proof: after selecting X under query A, running query B whose result set omits X (or places X beyond `HUBSPOT_TEMPLATE_RENDER_LIMIT`, computed slice at L458-471) leaves `selectedEmailTemplateId === X` with no pressed row on screen. This is exactly cursor's ts:472 finding, and it stays valid on this commit.
Why it matters: an invisible selection is a choice the operator cannot see, confirm, or clear without picking something else — and it is the value that flows to `hubspotConfig.sourceEmailId`. Non-stageable today (no email create trigger; `CampaignPlatform` has no `hubspot` member), so latent rather than exploitable, but it becomes a real hazard the moment the LFXV2-3201 trigger lands.
Fix: clear (or reconcile) `selectedEmailTemplateId` at search dispatch, or clear it when the selected id is not present in the visible (rendered) list.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -584,6 +828,151 @@ export class CampaignsComponent {
...
+    }
+
+    this.emailTemplateQuery.set(query);
+    this.emailTemplateSubmittedQuery.set(query);
+    // Captured at dispatch, alongside the slug actually passed to the request below, so the
+    // "not connected" copy names the project that was queried even if the foundation changes
+    // while the response is in flight.
+    this.emailSearchProjectSlug.set(projectSlug);
+    this.emailTemplatesLoading.set(true);
+    this.emailTemplatesError.set(null);
```
dealako, line 870: [nit] `emailTemplatesTruncated` is not reset at dispatch, unlike its siblings

Issue: the dispatch block resets `emailTemplatesError` and `emailChannelEnabled` (and clears the list), but `emailTemplatesTruncated` is only rewritten on the success-with-data arm (L926).
Proof: the truncation banner is gated behind `@else if (emailTemplates(); as templates)` with length>0, and every non-empty list assignment co-sets the flag, so a stale value is not currently reachable — harmless today.
Why it matters: pure defense-in-depth; if the template's error/enabled gating is ever reordered, a stale 'partial list' banner from a prior empty-query search could surface.
Fix: add `this.emailTemplatesTruncated.set(false)` next to the other dispatch resets.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -160,36 +160,167 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+                          <!--
+                            Only when it is not ALREADY the primary label above. With no `name`,
+                            the primary falls through to `subject`, and rendering it again here
+                            printed the same string twice on one row.
+                          -->
+                          @if (template.subject && template.name) {
+                            {{ template.subject }}
+                          }
+                          @if (template.state) {
+                            · {{ template.state }}
```
dealako, line 312: [nit] Leading `·` separator when a template has no `name`

Issue: the subject segment renders only when `template.subject && template.name`, but the `state` and `updatedAt` segments hard-prepend `· `. `HubSpotMarketingEmail.name/subject/state` are all optional, and the primary label falls through to `subject` when `name` is absent.
Proof: a name-less, subject-present row suppresses the first segment while L312/L315 still emit their leading `· `, producing e.g. `· draft · Updated Aug 14, 2026` — a dangling leading dot on exactly the rows the metadata line exists to disambiguate.
Why it matters: cosmetic only, reads as a rendering glitch. (Matches cursor's html:308.)
Fix: join non-empty metadata parts with `' · '` (e.g. in a small computed/pipe), or gate the leading `· ` on a preceding segment being present.

packages/shared/src/utils/date-time.utils.ts
```diff
@@ -617,3 +617,49 @@ export function formatIsoDateLabel(iso: string): string {
...
+  // the latter: `new Date('2026-02-31T10:00:00Z')` rolls over to Mar 3 exactly as the bare date
+  // does, so the fabrication survived in this branch while the spec pinned the other one. Split
+  // the date portion off and round-trip it through the same validator before formatting.
+  const datePart = value.slice(0, 10);
+  if (/^\d{4}-\d{2}-\d{2}$/.test(datePart) && formatIsoDateLabel(datePart) === datePart) {
+    // formatIsoDateLabel echoes its input when it rejects, so an equal result means the calendar
+    // date does not exist — Feb 31st, year 0001, and the like.
+    return '';
+  }
+  const date = new Date(value);
```
dealako, line 657: [question] `formatHubSpotUpdatedAt` silently drops non-ISO `updatedAt`

Issue: the function handles ISO date-only and ISO timestamp forms; anything else falls through to `new Date(value)` and returns `''` on `isNaN`.
Proof: campaign-service maps `updatedAt: email.updated_at` with no normalization, and tests only exercise ISO strings. If HubSpot ever forwards `updated_at` as epoch millis (a form some HubSpot APIs use), `new Date('1699564800000')` → Invalid Date → the date silently disappears — defeating the same-name disambiguation the field exists for.
Why it matters: a silent per-row degradation, not a crash. Depends entirely on the campaign-service Goa contract.
Fix: if the ISO-8601 contract is guaranteed upstream, no change needed — worth a one-line confirmation. If epoch millis is possible, add a numeric branch + spec case.

dealako (approved):
Independent review by three parallel subagents (security/privacy, correctness/perf/tests, style/docs) against HEAD `fc6ebff`, reconciled with the CodeRabbit/Cursor/Copilot history. Verdict: approve with minor comments — see the summary comment for the full breakdown.

dealako:
Hi @mrautela365 — thorough, unusually honest PR. 👏

I ran an independent multi-dimension review (three parallel subagents: security/privacy, correctness/performance/tests, style/docs) against HEAD `fc6ebff` and reconciled it with the full CodeRabbit/Cursor/Copilot history. The picker is well-built and — importantly — the central safety claim holds up under tracing: `selectedEmailTemplateId` flows to `hubspotConfig.sourceEmailId`, but there is no reachable email create trigger (`CampaignPlatform` has no `hubspot` member, and the only `createCampaign` call site is the paid Implementation tab), so **no send can be staged today**. The four-state distinction (not-connected / search-failed / searched-and-empty / results), the `possiblyTruncated`-only-on-empty-query logic, the id-less-row drop, the monotonic `emailSearchGeneration` stale-response guard, the foundation-switch selection clear, and the mutation-tested specs are all genuinely solid.

**Bot reconciliation:** the earlier Cursor/Copilot High/Medium findings (foundation-switch stale state, in-flight scoping, one-way search binding, channel-status reset, missing `updatedAt`, double-announced errors, truncation wording) were all valid when raised and I confirmed each is **addressed in the current HEAD** through your iterative commits — agree with the bots, agree they're resolved. Two current-commit items survive as low-severity (below).

### Issues

🔴 **Blocking: 0**

🟡 **Minor: 2**
1. **Selection persists across searches / invisible beyond the render cap** (`campaigns.component.ts:975` + dispatch ~845-876) — `searchEmailTemplates` clears the list/channel/error but never `selectedEmailTemplateId`, so a prior selection survives a new search even when the chosen row isn't in the visible/rendered set. Latent (non-stageable today), but it's the value that becomes `sourceEmailId`, so it turns into a real hazard once the LFXV2-3201 trigger lands. Agrees with Cursor's ts:472. Clear/reconcile selection at dispatch or against the rendered list.
2. **Stale/contradictory ticket references** — `campaign.constants.ts:42` still says the picker is pending under **LFXV2-3197**, while `campaigns.component.html:119-124` lists it under a "STILL OPEN … LFXV2-3198" block — yet this PR *ships* the picker. Both are pre-existing comments this change makes false; worth reconciling to one ticket and updating the "STILL OPEN" note to reflect that only the create trigger (LFXV2-3201) and email-copy remain. (Couldn't inline — both lines are outside the diff.)

⚪ **Nit: 2**
1. **`emailTemplatesTruncated` not reset at dispatch** (`campaigns.component.ts:870`) — only rewritten on the success arm; unreachable stale-banner today, pure defense-in-depth to reset it alongside its siblings.
2. **Leading `·` separator when a template has no `name`** (`campaigns.component.html:312`) — a name-less/subject-present row yields `· draft · Updated …`. Agrees with Cursor's html:308. Cosmetic.

❔ **Question: 1**
1. **`formatHubSpotUpdatedAt` silently drops non-ISO input** (`date-time.utils.ts:657`) — falls through to `new Date(value)` → `''` on `NaN`. Fine if the campaign-service Goa contract guarantees ISO-8601 for `updated_at`; a problem only if epoch-millis is ever emitted. One-line confirmation would close it.

No security, privacy, or data-residency findings survived proof + tracing. No merge conflicts.

### Decision: ✅ **Approved with minor comments**

None of the surviving items block merge — they're latent-until-wired, cosmetic, or documentation. Recommend folding minor #1 in before (or alongside) the LFXV2-3201 create trigger, since that's when the invisible-selection state stops being harmless. Nice work.

PR #1598, opened 2026-08-17
fix(meetings): guard against stale project context on edit (GH-1579)

### LINKED ISSUE #1579: Meeting edit page can check permissions against the wrong project when a server call fails
## Problem

When someone opens a link to edit a meeting, the app checks whether they are allowed to edit it. In rare cases — when a server request fails momentarily while the page loads — that check can run against the wrong project. This can either let someone reach the edit form who should not, or wrongly block the meeting's actual organizer. The final save is still protected server-side, so the exposure is limited to the page-level check.

A review of the meeting edit flow surfaced related rough edges in the same area:

- When the meeting's project details fail to load, the page can stay stuck showing a different project than the one the meeting belongs to.
- The rules for who is allowed to edit meetings are maintained in several places, so future changes risk applying them inconsistently.
- Meeting details can be cached briefly (up to ~10 seconds), so a very recent change can momentarily be invisible.
- The project shown in the page URL can lag behind the project that is actually selected.

## Outcome

Opening a meeting edit link always evaluates permissions against the project the meeting actually belongs to. If a server call fails along the way, the page fails safely rather than falling back to an unrelated project. The project shown on screen and in the URL always matches the meeting's own project, and the edit-permission rules live in one place so they stay consistent everywhere they apply.

## Success criteria

- An organizer opening their meeting edit link is never evaluated against an unrelated project, even when a server hiccup occurs during page load.
- A user without edit rights on the meeting's project never reaches the edit form, including during server errors — the page denies safely instead.
- When project details fail to load with the meeting, the edit page recovers to the meeting's own project or shows a clear error — it never stays on an unrelated project.
- Once the page settles, the project named in the page URL always matches the selected project.
- Changing who counts as a meeting editor requires a change in exactly one place.

audigregorie:
## Summary

This PR hardens the meetings edit flow against stale project context. Previously, when a meeting read failed for a non-404 reason, the writer guard silently fell back to the active lens and could authorize (or deny) against an unrelated project; when the relation-gated project GET returned nothing for an organizer without a direct viewer relation, the manage component gave up on context resolution; and a stale `?project=` query param could persist even after the user selected a different project. Each path is now corrected, and the meetings write predicate and slug resolver are centralized in a shared util so the guard and the reactive access signals can no longer drift.

Resolves #1579

### Stale-context authorization on edit

| Before | After |
|---|---|
| If reading a meeting failed for any reason, the edit guard silently fell back to whichever project the user happened to have open, so a legitimate organizer could be denied — or worse, the access check could run against the wrong project. | Only a genuinely missing meeting (404) falls back to the open project. Any other read failure fails closed and redirects, so no access check runs against a stale, unrelated project. |

### Context resolution for organizers without a direct viewer relation

| Before | After |
|---|---|
| When the project lookup returned nothing for an organizer who has meeting access but no direct viewer relation, the edit page left the project context unresolved and never self-corrected. | The edit page retries resolution from a fresh (uncached) meeting detail payload whose project enrichment is not relation-gated, so the organizer's project context is recovered when the enrichment is available. The retry runs at most once per meeting. |

### Stale `?project=` query param after a same-project selection

| Before | After |
|---|---|
| Selecting the same project that was already active skipped the URL sync, so a stale `?project=` left by a prior selection lingered in the URL. | The URL sync runs regardless of whether the selection matches, so a stale `?project=` is corrected whenever a selection is reasserted. |

## Technical changes

`apps/lfx-one/src/app/shared/utils/write-access.util.ts` — Shared write-access helpers
- Adds `hasMeetingWriteAccess(project)` as the single definition of the meetings write predicate (`writer` OR `meetingCoordinator`), replacing the inline copies that had drifted across the guard and the reactive access signals.
- Adds `resolveMeetingWriteSlug(meeting, fallback)` — prefers the BFF-enriched `project_slug`, falls back to `project_uid` (the project GET sniffs UUIDs), then to the supplied active-context fallback.

`apps/lfx-one/src/app/shared/guards/writer.guard.ts` — Route guard
- Fails closed on any non-404 meeting read failure: the `catchError` now returns `null` (triggering the `if (!slug)` redirect) for non-404 errors and only falls back to the active context on a 404.
- Replaces the inline slug-resolution and write-predicate logic with `resolveMeetingWriteSlug` and `hasMeetingWriteAccess`.
- Adds `HttpErrorResponse` import to distinguish 404 from other failures.
- Updates the doc comments to describe the fail-closed behavior.

`apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.ts` — Meeting edit/manage component
- In `initMeetingContextFallback`, when the relation-gated project GET returns null, retries resolution via `resolveContextFromFreshMeeting`, which calls `meetingService.getMeetingDetail(entity.uid, { skipCache: true })` and builds the `ProjectContext` from the freshly enriched payload.
- Adds a `contextFallbackRetried` Set keyed by meeting uid so the NavigationEnd re-apply (which re-emits the entity on every step navigation) does not repeat the uncached fetch.
- Emits `null` (leaving context untouched) when the fresh payload still carries no `project_slug`, with a `console.warn` for diagnostics; errors are caught and logged.
- Replaces the inline slug-resolution and write-predicate logic with the shared helpers.

`apps/lfx-one/src/app/modules/meetings/meetings-dashboard/meetings-dashboard.component.ts` — Meetings dashboard
- Replaces the inline `project?.writer === true || project?.meetingCoordinator === true` check with `hasMeetingWriteAccess(project)` in the reactive access signal.

`apps/lfx-one/src/app/shared/services/meeting.service.ts` — Meeting service
- Adds an optional `options.skipCache` parameter to `getMeetingDetail` so callers can force a fresh fetch when they need enrichment that a cached payload may predate.
- Adds `pruneExpiredMeetingDetailCache`, an opportunistic sweep invoked on every insert, so TTL-lapsed entries no longer linger for the whole session.

`apps/lfx-one/src/app/shared/services/project-context.service.ts` — Project context service
- Restructures `setFoundation` and `setProject` so the URL sync runs outside the same-context early return, correcting a stale `?project=` query param even when the incoming selection already matches the current one. The signal write and cookie persist remain gated by the same-context check.

### DIFF
--- a/apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.spec.ts
@@ -1,0 +1,179 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { ApplicationRef, signal } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+import { TestBed } from '@angular/core/testing';
+import { provideNoopAnimations } from '@angular/platform-browser/animations';
+import { ActivatedRoute, convertToParamMap, NavigationEnd, Router } from '@angular/router';
+import { CommitteeService } from '@services/committee.service';
+import { MeetingService } from '@services/meeting.service';
+import { ProjectContextService } from '@services/project-context.service';
+import { ProjectService } from '@services/project.service';
+import { Meeting } from '@lfx-one/shared/interfaces';
+import { MessageService } from 'primeng/api';
+import { BehaviorSubject, of, throwError } from 'rxjs';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { MeetingManageComponent } from './meeting-manage.component';
+
+// Regression coverage for the relation-gated-null → uncached-detail context fallback (GH-1579):
+// when the by-uid project lookup is relation-gated into null for an organizer without a viewer
+// relation, the meeting detail is re-fetched uncached so its ungated enrichment can supply the
+// project the lookup withheld. Also pins the retry-once guard (a fresh fetch must not repeat on
+// every NavigationEnd step navigation) and the transient-error retry release.
+describe('MeetingManageComponent — context fallback', () => {
+  const MEETING_UID = 'meeting-uid-1';
+  const PROJECT_UID = 'project-uid-1';
+  const PROJECT_SLUG = 'test-project';
+
+  let getMeetingDetail: ReturnType<typeof vi.fn>;
+  let getProject: ReturnType<typeof vi.fn>;
+  let routerEvents$: BehaviorSubject<NavigationEnd>;
+  let setProject: ReturnType<typeof vi.fn>;
+
+  // Unenriched detail payload: project_uid only, no project_slug — the trigger condition for
+  // initMeetingContextFallback. Note `id` (not just `uid`): meetingEntityContext keys off
+  // `meeting.id`, so a stub without it never reaches the fallback.
+  const unenrichedMeeting = () =>
+    ({
+      id: MEETING_UID,
+      uid: MEETING_UID,
+      project_uid: PROJECT_UID,
+      committees: [],
+    }) as unknown as Meeting;
+
+  const enrichedMeeting = () =>
+    ({
+      ...unenrichedMeeting(),
+      project_slug: PROJECT_SLUG,
+      project_name: 'Test Project',
+      is_foundation: false,
+    }) as unknown as Meeting;
+
+  const createComponent = async () => {
+    const fixture = TestBed.createComponent(MeetingManageComponent);
+    // No fixture.detectChanges(): these tests exercise the constructor context-sync streams, not
+    // the template, and rendering would pull in the full PrimeNG stepper subtree for no signal.
+    await TestBed.inject(ApplicationRef).whenStable();
+    return fixture;
+  };
+
+  const emitNavigationEnd = async () => {
+    routerEvents$.next(new NavigationEnd(1, '/project/meetings/x/edit?step=2', '/project/meetings/x/edit?step=2'));
+    await TestBed.inject(ApplicationRef).whenStable();
+  };
+
+  const skipCacheCalls = () => getMeetingDetail.mock.calls.filter(([, opts]) => (opts as { skipCache?: boolean } | undefined)?.skipCache === true);
+
+  beforeEach(() => {
+    getMeetingDetail = vi.fn();
+    getProject = vi.fn();
+    setProject = vi.fn();
+    routerEvents$ = new BehaviorSubject<NavigationEnd>(new NavigationEnd(0, '/project/meetings/x/edit', '/project/meetings/x/edit'));
+
+    TestBed.configureTestingModule({
+      providers: [
+        {
+          provide: Router,
+          useValue: {
+            events: routerEvents$.asObservable(),
+            url: '/project/meetings/x/edit',
+            parseUrl: vi.fn().mockReturnValue({ queryParams: {} }),
+            serializeUrl: vi.fn().mockReturnValue('/project/meetings/x/edit'),
+            navigate: vi.fn(),
+            createUrlTree: vi.fn(),
+          },
+        },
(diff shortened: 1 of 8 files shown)


MRashad26 (approved):
Reviewed with the code-standards-enforcer checklist. This is a security-relevant fix (fail-open → fail-closed authorization), so I verified the diff against the surrounding un-diffed code rather than trusting the PR description alone: fetched `writer.guard.ts`, `meeting-manage.component.ts`, `meeting.service.ts`, `project-context.service.ts`, `entity-project-context.util.ts`, `evict-on-write-access-loss.util.ts`, and the `Meeting`/`EntityWithProject` interfaces in full.

**What this does:** Fixes GH-1579 — a meeting-edit route with a missing/stale `?project=` could authorize `writerGuard` against whatever project the cookie-restored active context happened to be, instead of the meeting's own project. Three coordinated changes: (1) `writerGuard`'s meeting-read fallback is now 404-only (fail-closed on every other error); (2) `MeetingManageComponent` gets a one-shot uncached-refetch fallback for organizers whose relation-gated `getProject` call returns null; (3) `ProjectContextService.setFoundation`/`setProject` now sync the URL query param unconditionally instead of skipping it on an early-return when the selection is unchanged.

**Verified correct:**
- **`writer.guard.ts`**: `catchError((error) => of(error instanceof HttpErrorResponse && error.status === 404 ? fromContext : null))` is exactly what the new `writer.guard.spec.ts` pins — confirmed against the 6 spec cases (404-fallback, fail-closed-on-500-with-no-downstream-probe, authorize-against-meeting's-own-project, resolve-uid-when-slug-missing, ED-fast-path, non-meetings-features-use-active-context). The 500 case correctly asserts `getProject` and `getCommittee` are never called — the fail-closed path denies before any downstream probe, not just before the specific redirect shape.
- **`resolveMeetingWriteSlug`/`hasMeetingWriteAccess`** (`write-access.util.ts`) are used consistently across `writer.guard.ts`, `meeting-manage.component.ts`'s `initWriteAccess`, and `meetings-dashboard.component.ts` — no leftover call site still doing the old inline `project.writer === true || project.meetingCoordinator === true` check with the old fail-open catch.
- **`resolveContextFromFreshMeeting`'s retry-once logic**: `contextFallbackRetried` (keyed by meeting uid) is added before the fetch fires and only released in the `catchError` branch — so a genuine "still unenriched" success (no `project_slug` in the fresh payload) permanently stops retrying for that uid (correct — repeating an uncached fetch every `NavigationEnd` would be wasted work), while a transient failure (network/5xx) releases the retry marker so a later navigation can try again. No infinite-loop risk: the Set only ever grows on success-path resolution failure, and `getMeetingDetail(id, { skipCache: true })` is a single one-shot request, not a poll.
- **`getMeetingDetail`'s `skipCache` and cache-pruning**: confirmed `skipCache` bypasses the cache *lookup* but still replaces the map entry with the new `shareReplay(1)` observable (not merely appended) — matches the doc comment's claim that already-subscribed callers keep the stale result while new subscribers get the fresh one. `pruneExpiredMeetingDetailCache()`'s O(n) sweep runs once per cache insert against a per-user, per-session meeting-detail cache — bounded by distinct meetings viewed in one session, not a general-purpose concern here.
- **`project-context.service.ts` restructuring**: `setFoundation`/`setProject` now call `syncProjectQueryParam` unconditionally when `syncUrl` is true, independent of whether the selection changed — confirmed this doesn't create a footgun because `syncProjectQueryParam` itself no-ops when a navigation is already in flight (`this.router.getCurrentNavigation()` guard), and setting the same slug via `Location.replaceState` is idempotent. Verified `syncUrl` defaults to `true` and every caller that needs `false` (the two `entity-project-context.util.ts` call sites, gated on whether `?project=` was already present in the URL) still passes it explicitly.
- **Type/interface consistency**: `Meeting.project_uid` is a required `string`; `project_slug`/`project_name`/`is_foundation` are optional — `EntityWithProject`'s shape (`project_uid: string`, the rest optional) matches exactly, so `initializeMeetingEntityContext`'s mapping compiles cleanly with no unsafe casts.
- Test coverage for both new spec files is behavior-focused (interaction assertions like "getProject was never called" rather than snapshotting a specific UrlTree shape), which is the right granularity for a guard whose contract is about what does *not* run on the fail-closed path.

PR #1599, opened 2026-08-17
fix(projects): stop protected project calls on anonymous routes

jordane:
## Summary

- 93% of the `AUTHENTICATION_REQUIRED` 401s in Datadog RUM (89% on public `/meetings/:id` invite pages) come from anonymous sessions hitting protected `/api/*` endpoints — the server is correctly rejecting them; the client shouldn't be calling them at all.
- `WriterGrantsService` and `ProjectContextService`'s project lookups now gate on `UserService.authenticated()`, so anonymous visitors no longer fire `/api/projects*` calls that were always going to 401.
- The public meeting endpoint now resolves and returns the parent/foundation project server-side (via the existing M2M-authenticated `getProjectById` call), so anonymous visitors still get foundation attribution in the breadcrumb and pageview tracking without needing their own authenticated fetch.

Issue: LFXV2-3266

### DIFF
--- a/apps/lfx-one/src/server/controllers/public-meeting.controller.ts
+++ b/apps/lfx-one/src/server/controllers/public-meeting.controller.ts
@@ -2,9 +2,16 @@
 // SPDX-License-Identifier: MIT
 
 import { Meeting } from '@lfx-one/shared';
-import { MEETING_PASSWORD_HEADER } from '@lfx-one/shared/constants';
+import { MEETING_PASSWORD_HEADER, ROOT_PROJECT_SLUG } from '@lfx-one/shared/constants';
 import { MeetingVisibility, QueryServiceMeetingType } from '@lfx-one/shared/enums';
-import { CreateMeetingRegistrantRequest, MeetingOccurrenceSummary, MeetingRegistrant, PublicMeetingOccurrencesResponse } from '@lfx-one/shared/interfaces';
+import {
+  CreateMeetingRegistrantRequest,
+  MeetingOccurrenceSummary,
+  MeetingRegistrant,
+  Project,
+  PublicMeetingOccurrencesResponse,
+  PublicMeetingProject,
+} from '@lfx-one/shared/interfaces';
 import { NextFunction, Request, Response } from 'express';
 
 import { ResourceNotFoundError, ServiceValidationError } from '../errors';
@@ -84,6 +91,12 @@ export class PublicMeetingController {
           path: `/projects/${meeting.project_uid}`,
         });
       }
+
+      // Resolves the foundation project server-side (LFXV2-3266) so anonymous visitors get
+      // breadcrumb/attribution data without an authenticated /api/projects/:uid call from the
+      // client. m2mToken is still active here — the user-token swap below happens later and is
+      // restored before this response is sent.
+      const parent = await this.resolveParentProject(req, project);
 
       // Resolve host-key visibility (organizer OR project writer OR committee writer). This sets
       // meeting.organizer (used for the registrant counts below) and meeting.can_view_host_key,
@@ -172,7 +185,7 @@ export class PublicMeetingController {
       if (meeting.visibility === MeetingVisibility.PUBLIC && !meeting.restricted) {
         res.json({
           meeting,
-          project: { name: project.name, slug: project.slug, logo_url: project.logo_url, uid: project.uid, parent_uid: project.parent_uid },
+          project: this.toPublicMeetingProject(project, parent),
         });
         return;
       }
@@ -186,7 +199,7 @@ export class PublicMeetingController {
       if (meeting.invited || meeting.organizer) {
         res.json({
           meeting,
-          project: { name: project.name, slug: project.slug, logo_url: project.logo_url, uid: project.uid, parent_uid: project.parent_uid },
+          project: this.toPublicMeetingProject(project, parent),
         });
         return;
       }
@@ -206,7 +219,7 @@ export class PublicMeetingController {
               meeting.invited = true;
               res.json({
                 meeting,
-                project: { name: project.name, slug: project.slug, logo_url: project.logo_url, uid: project.uid, parent_uid: project.parent_uid },
+                project: this.toPublicMeetingProject(project, parent),
               });
               return;
             }
@@ -226,7 +239,7 @@ export class PublicMeetingController {
       }
 
       // Send the meeting and project data to the client
-      res.json({ meeting, project: { name: project.name, slug: project.slug, logo_url: project.logo_url, uid: project.uid, parent_uid: project.parent_uid } });
+      res.json({ meeting, project: this.toPublicMeetingProject(project, parent) });
     } catch (error) {
       // Error handler will log
       next(error);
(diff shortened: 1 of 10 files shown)


apps/lfx-one/src/server/controllers/public-meeting.controller.ts
```diff
@@ -270,6 +283,9 @@ export class PublicMeetingController {
         });
       }
 
+      // Resolves the foundation project server-side (LFXV2-3266) — see getMeetingById for why.
+      const parent = await this.resolveParentProject(req, project);
```
audigregorie, line 287: **Minor — missing test coverage for the past-meeting parent resolution**

`resolveParentProject` is called here in `getPublicPastMeetingById`, but the new spec block (`PublicMeetingController.getMeetingById parent project resolution (LFXV2-3266)`) only covers the upcoming-meeting path. The past-meeting path has the same four behaviors (success, null parent, error fallback, root-slug mapping) and none are tested.

A regression in this branch — e.g. swapping the user token before the call, or forgetting to thread `parent` into `toPublicMeetingProject` at line 368 — would slip through. Worth adding a parallel `describe('PublicMeetingController.getPublicPastMeetingById parent project resolution (LFXV2-3266)')` block mirroring the four cases above.

audigregorie (commented):
## Code Review Summary

Solid PR — the three anonymous-401 sources are gated cleanly, and moving parent-project resolution server-side (with the `toPublicMeetingProject` helper guaranteeing consistency across all 5 response branches) is a real architectural improvement. The `meeting-join` pageview race is gone too. No Critical or Major issues.

### Minor — outside the diff

- **Existing host_key gating tests have a misleading `projectSvc.getProjectById` mock** (`public-meeting.controller.spec.ts:165`). The `beforeEach` uses `mockResolvedValue(buildProject())`, so the new `resolveParentProject` call returns the *same* project as the parent — `project.parent` ends up `{ uid: PROJECT_UID, name: 'Proj', slug: 'proj' }` (parent uid == project uid). Tests pass because they don't assert on `project.parent`, but the setup is now wrong for every `getMeetingById` flow. The new parent-resolution tests use `mockImplementation` keyed on uid — the host_key tests should follow the same pattern.
- **`WriterGrantsService` auth gate is a one-shot check** (`writer-grants.service.ts:69`). The `authenticated` check lives inside `afterNextRender`, which fires once. If a future feature adds an in-page login transition on a route that mounts this service, the writer-grants signals would stay `false` for the session. Fine for the current public-meeting use case; just flagging the assumption.

### What's done well

- Gating `WriterGrantsService`, `ProjectContextService.initCanWrite`, and `initSelectedFoundationSfid` on `UserService.authenticated()` — the seeding order in `AppComponent` is verified, so the `afterNextRender` check always sees the settled value.
- `resolveParentProject` fails closed to `null` rather than 500-ing the whole meeting response, and uses `access=false` so no FGA leak.
- Removing the 2s `PARENT_FETCH_TIMEOUT_MS` race in `initializePublicMeetingPageviewTracking` — the pageview can no longer be lost when the parent fetch is slow.
- New `WriterGrantsService` spec covers the auth gate; new `getMeetingById` parent-resolution spec covers success/null/error/root-slug.

audigregorie (dismissed):

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
jordane: Added a parallel `describe('PublicMeetingController.getPublicPastMeetingById parent project resolution (LFXV2-3266)')` block mirroring the four cases (resolved, no parent_uid, upstream throw, root-slug mapping). Pushed in the latest commit.

audigregorie (approved):

MRashad26 (approved):
Reviewed with the code-standards-enforcer checklist (signal/computed patterns, shared interfaces placement, M2M token lifecycle correctness, test coverage, security). No template files are touched by this PR.

**What this does:** Stops the client from firing authenticated `/api/projects*` calls on anonymous/public routes (they were always going to 401, per the Datadog RUM data cited in the PR body) by gating `WriterGrantsService` and two `ProjectContextService` derivations on `UserService.authenticated()`. Moves foundation/parent-project resolution for the public meeting join page server-side, so anonymous visitors still get breadcrumb attribution without needing their own authenticated fetch — and deletes a client-side race-against-a-timeout (`PARENT_FETCH_TIMEOUT_MS`) that's no longer needed.

**Verified correct — the things most likely to be wrong in a change like this:**
- **M2M token timing on the server is right, not just claimed.** Traced both `getMeetingById` and `getPublicPastMeetingById`: `resolveParentProject` is called while `req.bearerToken` still holds the M2M token — *before* the temporary `req.bearerToken = originalToken` restore-for-access-check block runs (and before it's swapped back to M2M again). The comment ("m2mToken is still active here") matches the actual code order in both handlers.
- **`resolveParentProject` never throws into the response path** — wrapped in try/catch, degrades to `null` and logs a warning rather than failing the whole meeting response if the parent lookup 500s; covered by a dedicated test (`returns parent: null and still responds 200 when the parent lookup throws`).
- **Root-project mapping is preserved exactly**, just relocated: `parent?.slug === ROOT_PROJECT_SLUG ? null : parent` on the server mirrors the client-side rule it replaces (previously in `initializeParentProject`), and is explicitly tested (`maps a root-slug parent to null`).
- **`toPublicMeetingProject` centralizes the response shape** so `parent` can't be added on some of the four `res.json(...)` branches in `getMeetingById` and missed on others — a real risk this refactor correctly closes off, not just a stylistic dedup.
- **`WriterGrantsService`'s gate is correct given its own default state**: `hasWriterFoundation`/`hasWriterProject` already default to `false`, which is the right answer for an anonymous visitor, so skipping the call entirely (rather than letting it fire and fail closed) doesn't change behavior for that case — just removes the guaranteed-401 network call. Test explicitly asserts both calls are skipped and both signals stay `false`.
- **`ProjectContextService`'s two gated derivations correctly key off live `authenticated()` state via `combineLatest`**, not a one-time snapshot — so a session that authenticates after initial load (or the reverse) is picked up.
- **Drive-by spec fixes are explained, not incidental:** the `MessageService` provider additions across the campaigns spec files are needed because `ProjectContextService` now injects `UserService`, and `UserService` itself injects `MessageService` — confirmed by reading `user.service.ts`. Without the fix, these specs would have started failing as a side effect of this PR's own change, so fixing them here (rather than leaving a follow-up PR) is the right call.
- **Shared-interface placement correct**: `PublicMeetingParentProject`/`PublicMeetingProject.parent` land in `packages/shared/src/interfaces/meeting.interface.ts`, and the slim shape (`uid`/`name`/`slug` only) matches the existing "callers must not assume more fields" contract already documented on `PublicMeetingProject`.
- **No unused-import debt left behind** on the client: `distinctUntilChanged`/`merge`/`timer` are still imported in `meeting-join.component.ts` because they're genuinely used elsewhere in the file (confirmed), not leftover from the removed timeout-race logic.

**Minor, non-blocking:** the client-side `initializeParentProject` is now a one-line `computed(() => this.project()?.parent ?? null)`, and there's no existing or new spec asserting that mapping directly — but the substantive logic it used to contain (root-slug mapping, timeout race, error handling) moved server-side and is now thoroughly covered there (8 new test cases across both endpoints), so this is a low-risk gap, not a regression in coverage depth.

PR #1327, opened 2026-08-05
feat(dashboards): open executive dashboards to LF staff

MRashad26:
## Summary
- LF Staff (`team:lf-staff#member` OpenFGA relation) can now view Foundation Health, Marketing Overview, and Health Metrics dashboards without needing the Executive Director persona.
- Marketing Impact's route is open to ED-or-LF-Staff, but LF Staff see only the Social Listening tab directly (no tab bar/focus filters) — the other 6 tabs and Campaigns stay strictly ED-only, per JIRA ticket comments.
- New `dashboardAccessGuard` mirrors the existing `newsletterAccessGuard` SSR-safe pattern: synchronous cookie-seeded ED fast path, async LF-staff fallback via `refreshEnrichedPersonas()`.
- Backend: `PersonaDetectionService.checkLFStaff()` reuses the existing `AccessCheckService` HTTP pattern (extended with `team`/`member` types) with request-scoped caching, degrading to `false` on transient errors.

## Test plan
- [x] `yarn format && yarn lint && yarn build` all pass
- [x] Post-commit + full-branch reviewer sweep (general reviewer + code-standards-enforcer) returned clean; one Important finding (dashboard-type naming ambiguity) addressed via clarifying comment
- [ ] Manual verification in a dev environment: LF Staff user sees Foundation Health / Marketing Overview / Health Metrics, and only Social Listening under Marketing Impact; Campaigns and other Marketing Impact tabs remain ED-only
- [ ] ED user sees unchanged full dashboard experience

Resolves LFXV2-2726

PR #1602, opened 2026-08-17
fix(nav): merge lenses for all hybrid grant sources

### LINKED ISSUE #1600: Lens switcher shows both Foundation and Projects for non-board hybrid users
## Summary

The sidebar lens switcher should merge Foundation into a single "Projects" entry for users who can reach both lenses. That merge only fires for users holding a **board role and a project role**, but Foundation access can also come from LF staff membership, a root-writer grant, or a `writer` grant on a foundation. Users in those paths see two separate lens buttons.

Reproduced in production with user `lguerra`, whose Foundation access comes from `isLFStaff`.

## Expected vs actual

**Expected:** one merged "Projects" lens with All / Foundations / Projects tabs in its dropdown, per #713.
**Actual:** separate "Foundation" and "Projects" buttons, and the dropdown renders in non-hybrid single-list mode.

## Root cause

Visibility and merge are derived from different inputs, so they can disagree.

`packages/shared/src/utils/lens.utils.ts`:

```ts
const showFoundation = hasBoardRole || isRootWriter || hasWriterFoundation || isLFStaff;
const showProject    = hasProjectRole || isRootWriter || hasWriterProject;
```

`apps/lfx-one/src/app/shared/services/lens.service.ts`:

```ts
public readonly isHybridPersona = computed(() => this.personaService.hasBoardRole() && this.personaService.hasProjectRole());
```

## Scenario coverage

Every row below results in both Foundation and Projects being visible, so every row should merge. Only the first one does.

| # | Foundation access from | Projects access from | Merged today? |
|---|---|---|---|
| 1 | board role (ED / Board Member) | project role (Maintainer / Contributor) | ✅ covered |
| 2 | LF staff (`isLFStaff`) | project role | ❌ missing — reproduced with `lguerra` |
| 3 | LF staff (`isLFStaff`) | project `writer` grant | ❌ missing |
| 4 | foundation `writer` grant | project role | ❌ missing |
| 5 | foundation `writer` grant | project `writer` grant | ❌ missing |
| 6 | root writer (`isRootWriter`) | root writer — grants both lenses alone | ❌ missing |
| 7 | board role | project `writer` grant | ❌ missing |

Row 6 is the clearest illustration: a single grant confers both lenses yet can never satisfy the merge condition.

Introduced in #713 (`cca73f6d2`).

luismoriguerra:
## Summary

Fixes #1600.

The sidebar merges Foundation into a single "Projects" entry for users who can reach both lenses, but the merge condition only recognised one of the four ways Foundation access is granted:

```ts
// visibility — four sources
const showFoundation = hasBoardRole || isRootWriter || hasWriterFoundation || isLFStaff;
const showProject    = hasProjectRole || isRootWriter || hasWriterProject;

// merge — one source
isHybridPersona = hasBoardRole && hasProjectRole;
```

So LF staff, root writers, and writer-granted users saw two separate lens buttons, and got the non-hybrid single-list selector instead of the All / Foundations / Projects tabs. Root writer is the starkest case: it confers both lenses on its own yet could never satisfy the merge condition.

## What changed

- **`isHybridLensUser(inputs)`** added to `packages/shared/src/utils/lens.utils.ts`, derived from `deriveAllowedLenses` so the merge rule is defined as "both lens buttons would render" rather than a hand-maintained restatement of the grant sources. Any grant source added later is covered for free.
- **`LensService`** now builds the persona/grant booleans once into a `lensGrantInputs` signal shared by `getAllowedLensIds()` and `isHybridPersona`, making the two rules structurally unable to drift.
- **Tests** — 16 new cases covering all seven hybrid combinations from the issue plus the single-lens negatives, and `isLFStaff` added to the spec's `NO_GRANTS` fixture (it was missing since #1327).

No behaviour change for users already treated as hybrid. `initDisplayLenses`, `displayActiveLens`, and `switchLens` are untouched — they read `isHybridPersona()` and inherit the wider definition.

## Blast radius

`isHybridPersona` also drives the tabbed selector in `ProjectSelectorComponent`, the sibling-lens preload in `NavigationService`, the merged-button highlight, and `switchLens`'s return-to-`lastNavLens`. Those now activate for LF staff and root writers, which is the intent, but it is more than hiding a button.

## Test plan

- [x] `yarn check-types`
- [x] `yarn test` (347 passing; `lens.utils.spec.ts` 29 passing)
- [x] `yarn lint:check` (only a pre-existing warning in `user.service.ts`)
- [x] `yarn build`
- [ ] LF staff account with a project role: one merged "Projects" lens with All / Foundations / Projects tabs
- [ ] Project-only user: foundations still selectable in the project list (the path #713 iterated on several times)
- [ ] Board + project role user: unchanged from today
- [ ] Hybrid state flipping mid-session once `refreshFromApi()` resolves `isLFStaff` (covered by `hybridTransitionPreloader`)

### DIFF
--- a/apps/lfx-one/src/app/shared/services/lens.service.ts
+++ b/apps/lfx-one/src/app/shared/services/lens.service.ts
@@ -12,8 +12,8 @@ import {
   NAV_LENS_COOKIE_KEY,
   ORG_LENS_ENABLED_FLAG,
 } from '@lfx-one/shared/constants';
-import { Lens, LensOption, NavLens } from '@lfx-one/shared/interfaces';
-import { deriveAllowedLenses } from '@lfx-one/shared/utils';
+import { Lens, LensGrantInputs, LensOption, NavLens } from '@lfx-one/shared/interfaces';
+import { deriveAllowedLenses, isHybridLensUser } from '@lfx-one/shared/utils';
 import { SsrCookieService } from 'ngx-cookie-service-ssr';
 import { filter, take } from 'rxjs';
 
@@ -56,8 +56,15 @@ export class LensService {
   public readonly availableLenses: Signal<LensOption[]> = this.initAvailableLenses();
   /** Lenses shown in the sidebar switcher. Mirrors {@link availableLenses} except for hybrid personas, who get a merged project entry instead of separate foundation + project buttons. */
   public readonly displayLenses: Signal<LensOption[]> = this.initDisplayLenses();
-  /** True when the user holds both a board role (ED/Board Member) AND a project role (Maintainer/Contributor). */
-  public readonly isHybridPersona: Signal<boolean> = computed(() => this.personaService.hasBoardRole() && this.personaService.hasProjectRole());
+  /** Single source of the persona/grant booleans, so lens visibility and the hybrid merge always agree. */
+  private readonly lensGrantInputs: Signal<LensGrantInputs> = this.initLensGrantInputs();
+  /**
+   * True when the user can reach both the foundation and project lenses, from any grant source —
+   * a board+project role pair, LF staff membership, a root-writer grant, or `writer` grants.
+   * Shares {@link lensGrantInputs} with {@link getAllowedLensIds} so the merged switcher entry
+   * cannot disagree with which lenses actually render.
+   */
+  public readonly isHybridPersona: Signal<boolean> = computed(() => isHybridLensUser(this.lensGrantInputs()));
   /** Lens to highlight in the switcher UI. Hybrid personas merge foundation + project into the 'project' entry, so a foundation-scoped active lens highlights 'project'. */
   public readonly displayActiveLens: Signal<Lens> = computed(() => {
     const active = this.activeLens();
@@ -186,7 +193,11 @@ export class LensService {
    * `MainLayoutComponent.syncLensFromRoute` is the one caller that re-asserts.
    */
   private getAllowedLensIds(): readonly Lens[] {
-    return deriveAllowedLenses({
+    return deriveAllowedLenses(this.lensGrantInputs());
+  }
+
+  private initLensGrantInputs(): Signal<LensGrantInputs> {
+    return computed(() => ({
       hasBoardRole: this.personaService.hasBoardRole(),
       hasProjectRole: this.personaService.hasProjectRole(),
       isRootWriter: this.personaService.isRootWriter(),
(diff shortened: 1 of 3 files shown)


MRashad26 (approved):
Reviewed with the code-standards-enforcer checklist (signal patterns, shared-interface placement, test coverage, no drift between derived rules). Small, well-scoped bug fix with excellent test coverage.

**What this does:** Fixes the sidebar's hybrid-persona merge condition, which only recognized the board+project role pair, so LF staff, root writers, and writer-granted users saw two separate lens buttons instead of the merged Projects entry with All/Foundations/Projects tabs.

**Verified correct against the actual code:**
- **`isHybridLensUser` is genuinely un-drift-able from `deriveAllowedLenses`**, not just by convention: it calls the same function and checks `lenses.includes('foundation') && lenses.includes('project')`, so any new grant source added to `deriveAllowedLenses` in the future is automatically picked up by the merge rule — exactly the intent described in the PR body.
- **`LensService.lensGrantInputs` is a single computed signal now feeding both `getAllowedLensIds()` and `isHybridPersona`** — confirmed both call sites read from the same signal, so the two can't structurally disagree (previously `isHybridPersona` re-derived its own booleans directly from `personaService`, which is exactly how it fell out of sync with `deriveAllowedLenses`'s four foundation-access sources).
- **Root-writer case is correctly fixed**: `deriveAllowedLenses` grants both `foundation` and `project` when `isRootWriter` is true, and the new hybrid check now includes root-writer-alone as a passing case (test: `'a root writer alone', { isRootWriter: true }`) — matches the PR's claim that root writer was "the starkest case" that could never satisfy the old hand-written condition.
- **`isLFStaff` field already exists on `LensGrantInputs`** (added previously) — this PR just closes the gap where the spec's `NO_GRANTS` fixture was missing it, which the PR body correctly attributes to #1327 rather than claiming credit for adding the field itself.
- **Test coverage is thorough**: 16 new cases span all 7 hybrid-true combinations plus 6 single-lens-only negatives plus an org-lens-independence check, directly exercising the exact matrix the bug report was about rather than just the two changed lines.
- **No behavior change for the already-working board+project case** — verified `showFoundation`/`showProject` logic in `lens.utils.ts` is unchanged by this diff; only `isHybridLensUser` and the service wiring are new.
- Shared types (`LensGrantInputs`) stay in `@lfx-one/shared/interfaces`; pure derivation logic stays in `@lfx-one/shared/utils`, consistent with existing placement of `deriveAllowedLenses`.

**Minor, non-blocking:** the PR body's own "Blast radius" section flags that `isHybridPersona` also drives `ProjectSelectorComponent`'s tabs, `NavigationService`'s sibling-lens preload, and `switchLens`'s `lastNavLens` return — correctly scoped as intentional widening, not a side effect the author missed, and the manual test-plan checkboxes for those paths are appropriately left unchecked pending real-account verification (LF staff, root writer, etc. aren't easily unit-testable end-to-end).

PR #1610, opened 2026-08-18
fix(campaigns): disable reddit until the implement tab can configure it

mrautela365:
## The problem

Reddit Ads is selectable in the Plan tab, and the Implement tab renders **nothing** for it.

- `showRedditSection` is computed in `implementation-tab.component.ts` and referenced **zero times** in the template. Every other selected platform gets a config section; Reddit gets none.
- `canSubmit()` guards Google (campaign type + form validity), LinkedIn (budget ≥ 1, geos, variants) and Meta (budget ≥ 1, variants). It uses `redditSelected` **only** to satisfy the "something is selected" check — no budget floor, no targeting guard.
- `submit()` builds a full `redditConfig` regardless, from brief-derived signals the user cannot see, review, or change.

So a Reddit campaign is created on values nobody approved.

## Worse than the ticket describes

LFXV2-3225 frames this as a zero-budget risk. It isn't — `redditBudgetUsd` is a signal initialised to **500**, and `grep` finds no `.set()` and no input handler anywhere. So the dispatch carries a hard-coded **$500/day budget** the user never chose, alongside subreddits, interests, keywords and geo targets they cannot review.

A zero-budget campaign fails visibly. This one spends.

## The fix

One line, per the ticket's own recommendation: `disabled: true`, making the picker honest now. Re-enable when the Implement section exists.

The service side is already complete — `internal/dispatch/reddit.go` implements `Dispatch`, `ToggleStatus` and `ReadMetrics`, and both Monitor and Optimize already render Reddit. The gap is only in the create path.

## Also documented

The rule this is an instance of, on `CAMPAIGN_PLATFORMS`: a platform belongs here as enabled only when the Implement tab can configure it. Selectable without a config section is worse than absent, because the picker promises a capability the next tab does not have.

That rule is currently violated in one direction and about to be tested in the other — LFXV2-3312 (Microsoft) is the mirror image, where the platform is *disabled* until its config builder lands.

## Testing

`yarn build`, `yarn lint:check`, `yarn format:check`, `./check-headers.sh`, and `tsc --noEmit` on both the app and spec configs all clean. Full server suite green at 1342 tests. The single lint warning is pre-existing in `user.service.ts`, untouched here.

No behavioural test added: the change is a constant flag with no branch to bind, and the component logic it disables is exactly what the follow-up will re-enable.

Closes LFXV2-3225


mrautela365:
Pushed `63e851f55`. Ran the reviewer simulator before this landed; **all three personas independently found the same gap**, which is strong enough signal that I fixed it here rather than deferring.

### The PR's headline claim was false for existing briefs

`disabled: true` gates the Plan picker. It does not gate **restore**. `populateFromBrief` set `selectedPlatforms` straight from the stored brief, and the server-side filter at `campaign-service.service.ts:1503` checks *membership* only — `reddit-ads` is still in `CAMPAIGN_PLATFORMS`, so it passes.

So any brief saved before this merge still restored a Reddit selection, still built `redditConfig`, and still dispatched the unbound `signal(500)` budget. And because the tile is now `[disabled]`, the user **couldn't deselect it**. The PR narrowed the hole while reading as though it closed it.

Filtered at restore now. When the filter empties the list, the existing `length` guard skips the set entirely so the `google-ads` default stands rather than substituting silently — the same reasoning `campaign-service.service.ts` already uses when it reports an all-unrecognised platform list as UNREADABLE.

**Two specs pin it, mutation-verified with a compiling mutation.** Removing the filter fails both. The first also asserts the enabled sibling *survives*, since a filter that dropped everything would satisfy a `not.toContain` assertion on its own.

### Dropped the JSDoc block

It claimed Microsoft was disabled *"until its config builder lands (LFXV2-3312)"*. That's false — the builder is on `feat/LFXV2-3312`, which also **enables** Microsoft and rewrites this same block. `git merge-tree` confirmed a real conflict, so whichever merged second would have hand-resolved two competing rule statements.

It also gave X a "no `ListAdAccounts`" rationale that is **equally true of Reddit** — verified: neither `internal/platform/twitter/` nor `internal/platform/reddit/` implements it, while Microsoft's client does at `accounts.go:371`. So the comment assigned Microsoft a defect it doesn't have and handed Reddit a different reason from the one that actually applies.

The per-entry comment carries the rationale instead and points at campaign-service's `sysacct.go` rather than restating per-provider internals this package can't see.

### One thing worth flagging

Re-enabling Reddit needs more than an Implement section: Reddit has no account discovery either, so the account id stays hand-entered on the connection. The inline comment says so, since "add the section and flip the flag" would be necessary but not sufficient.

packages/shared/src/constants/campaign.constants.ts
```diff
@@ -27,12 +27,29 @@ export const CAMPAIGN_TABS: readonly CampaignTabOption[] = [
...
+ * dispatch with no way to recover from inside the API.
+ */
 export const CAMPAIGN_PLATFORMS: readonly CampaignPlatformOption[] = [
   { id: 'google-ads', label: 'Google Ads', icon: 'fa-brands fa-google' },
   { id: 'microsoft-ads', label: 'Microsoft Ads', icon: 'fa-brands fa-microsoft', disabled: true },
   { id: 'linkedin-ads', label: 'LinkedIn Ads', icon: 'fa-brands fa-linkedin' },
   { id: 'meta-ads', label: 'Meta Ads', icon: 'fa-brands fa-meta' },
-  { id: 'reddit-ads', label: 'Reddit Ads', icon: 'fa-brands fa-reddit' },
+  // LFXV2-3225: re-enable once the Implement tab has a Reddit section and canSubmit guards it.
+  { id: 'reddit-ads', label: 'Reddit Ads', icon: 'fa-brands fa-reddit', disabled: true },
```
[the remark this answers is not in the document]
mrautela365: Fixed in `63e851f55` — the restore path is closed, so this comment described behaviour that no longer existed by the time you read it. The stale sentence went in `8d313dd46`.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
```diff
@@ -761,7 +762,19 @@ export class ImplementationTabComponent implements OnInit {
...
+      // unfiltered would submit that platform's config on brief-derived values they never saw
+      // and cannot edit, which is the exact defect disabling the picker exists to prevent.
+      //
+      // Filtered to empty, the `?.length` guard below leaves the `google-ads` default standing
+      // rather than substituting silently — the same reason campaign-service.service.ts reports
+      // an all-unrecognised platform list as UNREADABLE instead of defaulting it.
+      const selectable = new Set(CAMPAIGN_PLATFORMS.filter((o) => !o.disabled).map((o) => o.id));
+      const restorable = brief.selectedPlatforms.filter((p) => selectable.has(p));
+      if (restorable.length) {
+        this.selectedPlatforms.set(restorable);
```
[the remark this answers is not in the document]
mrautela365: Correct, and this was a defect I introduced in `63e851f55`. Fixed in `ba3b59a6e`.

Skipping the set on an empty result left the component's own `google-ads` default standing, so a Reddit-only brief opened as a **Google** campaign — and `submit()` builds its request from that signal, so the user could dispatch a platform they never chose.

You pointed at `campaign-service.service.ts:1505-1515`, and its comment describes this case almost verbatim: *"a Reddit-only brief would restore as a Google Ads campaign, with the user's real choice silently replaced by one they never made."* I quoted that reasoning in the commit that introduced the bug, then built the behaviour it rejects.

Now sets the filtered list unconditionally. Empty is the honest answer — `canSubmit()` requires a selected platform, so the brief opens blocked rather than dispatching something unchosen.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.spec.ts
```diff
@@ -197,3 +197,62 @@ describe('ImplementationTabComponent submit gate', () => {
...
+    expect(selectedPlatforms()).toContain('google-ads');
+  });
+
+  it('leaves the default standing rather than restoring an all-disabled brief', () => {
+    restore(['reddit-ads']);
+
+    // Filtered to empty, the `length` guard skips the set entirely, so the component keeps its
+    // own default instead of substituting a platform the user never chose.
+    expect(selectedPlatforms()).not.toContain('reddit-ads');
+    expect(selectedPlatforms()).toEqual(['google-ads']);
```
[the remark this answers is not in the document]
mrautela365: You're right, and this is the sharper half of the finding: the test asserted `['google-ads']` and so **pinned the defect in place** rather than catching it.

It now asserts `[]` (`ba3b59a6e`), with a comment saying why the default is not a neutral fallback here. Mutation-verified — reinstating the skip-on-empty guard compiles and fails the test, so the assertion binds the behaviour rather than describing it.

mrautela365:
**Heads up: #1611 supersedes this PR.** Not closing it without a maintainer's call, but the two conflict and only one should merge.

This PR disabled Reddit in the picker as a stopgap — the honest move while the Implement tab had no config section for it. **#1611 builds that section**, so Reddit becomes configurable and selectable again. Merging both would have one undo the other.

Every commit here is carried into #1611, including the restore-path fix and its follow-up. That part is still needed regardless: `disabled` gates the Plan tile, but `populateFromBrief` never consulted the picker, so a brief saved earlier still restored a disabled platform and still built its config. It is what keeps X/Twitter out of a restored brief today.

The two bot findings on this PR — an all-disabled brief silently opening as a Google campaign, and the spec that asserted that outcome rather than catching it — are both fixed and carried over.

**Suggestion: close this in favour of #1611.** If you would rather ship the stopgap first, #1611 can be rebased to land after it instead.

dealako:
Hi @mrautela365 👋 — thorough, honest little PR.

This is a one-line behavioral fix (`disabled: true` on `reddit-ads`) dressed with exactly the right amount of supporting work: a restore-path filter so a brief saved *before* the flag can't smuggle Reddit back into `selectedPlatforms`, two mutation-verified specs, and a load-bearing rationale comment. The scope is disciplined — you resisted the temptation to build the missing Implement section here and instead made the picker honest. I especially appreciate that the commit history shows you catching and fixing your own `63e851f55` regression (skip-on-empty leaving the `google-ads` default standing) before merge, and that the empty-restore case correctly leans on `canSubmit()` to block rather than substituting a platform the user never chose. Server-side defense in depth (`campaign-service.service.ts:1505-1515`) stays intact.

I traced the filter through `submit()` → `redditConfig` and the server brief reader across three parallel review passes (security/privacy, correctness/perf/tests, style/docs). No correctness, security, or data-privacy findings. The Copilot threads on the restore gap and the spec-pinned default were all real and are all resolved in the current head (`ba3b59a6e`) — agreed with the bot on each and confirmed the fixes land.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 — import out of alphabetical order; sentence fragment in the Reddit rationale comment
- ❔ Question: 0

Both nits are take-it-or-leave-it and don't gate merge.

### Final decision

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
```diff
@@ -13,6 +13,7 @@ import {
   LINKEDIN_CHAR_LIMITS,
   LINKEDIN_GEO_RESOLVE_MAP,
   META_CHAR_LIMITS,
+  CAMPAIGN_PLATFORMS,
```
dealako, line 16: [nit] Import out of alphabetical order

Issue: `CAMPAIGN_PLATFORMS` is appended after `META_CHAR_LIMITS`, but the rest of the `@lfx-one/shared/constants` block is alphabetical.
Proof: block reads CAMPAIGN_BUDGET_DEFAULTS, CAMPAIGN_CHAR_LIMITS, CAMPAIGN_JOB_POLL_INTERVAL_MS, LINKEDIN_*, META_CHAR_LIMITS, then CAMPAIGN_PLATFORMS.
Why it matters: eslint doesn't enforce member sort here, so it won't fail lint — purely readability drift; a reader scanning for a CAMPAIGN_* import expects it grouped with the others.
Fix: move `CAMPAIGN_PLATFORMS` between `CAMPAIGN_JOB_POLL_INTERVAL_MS` and `LINKEDIN_CHAR_LIMITS`.

packages/shared/src/constants/campaign.constants.ts
```diff
@@ -32,7 +32,19 @@ export const CAMPAIGN_PLATFORMS: readonly CampaignPlatformOption[] = [
   { id: 'microsoft-ads', label: 'Microsoft Ads', icon: 'fa-brands fa-microsoft', disabled: true },
   { id: 'linkedin-ads', label: 'LinkedIn Ads', icon: 'fa-brands fa-linkedin' },
   { id: 'meta-ads', label: 'Meta Ads', icon: 'fa-brands fa-meta' },
-  { id: 'reddit-ads', label: 'Reddit Ads', icon: 'fa-brands fa-reddit' },
+  // LFXV2-3225: disabled because the Implement tab has no Reddit config section — showRedditSection
+  // is computed but never bound to one, and canSubmit has no Reddit budget floor, so a selection
+  // dispatched on brief-derived values including an unbound 500/day budget signal. Re-enabling
```
dealako, line 37: [nit] Sentence fragment in the Reddit rationale comment

Issue: "...so a selection dispatched on brief-derived values including an unbound 500/day budget signal." has no main verb for "a selection".
Proof: the clause after "so" never resolves — reads as a fragment.
Why it matters: this is a load-bearing comment gating a future re-enable decision; an incomplete sentence weakens exactly the documentation the PR is investing in. Minor: `redditBudgetUsd` defaults to `500` and `submit()` maps it to `redditConfig.budgetUsd` with no daily qualifier, so "/day" is slightly imprecise.
Fix: complete the clause, e.g. "...so a selection would be dispatched on brief-derived values, including an unbound 500 budget signal."

dealako (dismissed):
Approved. Clean, well-scoped fix with mutation-verified specs and intact server-side defense in depth. Two nits inline — both optional. Nice work catching your own restore-path regression before merge.

mrautela365:
## Closing as obsolete — ss#1611 delivered what this PR was waiting for

This PR disables `reddit-ads`, and its own justification is now false. Both stated reasons were resolved by ss#1611, which merged at 00:03Z:

> "the Implement tab has no Reddit config section — `showRedditSection` is computed but never bound to one"

**Now bound.** `implementation-tab.component.html:422` — `@if (showRedditSection()) {`.

> "`canSubmit` has no Reddit budget floor, so a selection dispatched on brief-derived values including an unbound 500/day budget signal"

**Now enforced.** `implementation-tab.component.ts:382` — `if (redditSelected && !this.redditBudgetIsUsable()) return false;`, backed by `REDDIT_MAX_BUDGET_USD` and a 9-test `reddit budget gate` block.

So merging this would disable a platform whose configuration UI ships alongside it.

### The second-order consequence

`CAMPAIGN_SERVICE_STATUS_PLATFORMS` is derived as `CAMPAIGN_PLATFORMS.filter((p) => !p.disabled)` (`campaign.controller.ts:52-54`). Setting `disabled: true` therefore removes Reddit from the status-toggle allowlist, so **existing UUID-backed Reddit campaigns lose pause/resume** — a live capability regression, not just a picker change.

### The one reason that did NOT expire

The comment also notes that neither the Reddit nor X platform client implements `ListAdAccounts` (campaign-service `internal/bootstrap/sysacct.go`), so the account id is hand-entered on the connection. That remains true. If it justifies gating Reddit on its own, it deserves its own PR with only that rationale — not this one, whose two primary claims no longer hold.

Closing rather than rewriting: the premise this branch was built on is gone.

dealako:
Hey @mrautela365 — thanks for keeping the branch current with `main`. I re-reviewed after the two merge commits landed, and the situation has changed fundamentally since my last pass, so this round is mostly bad news through no fault of the original change.

👏 The round-1 work was solid — the restore-path filter, the mutation-verified specs, and the honest empty-restore behavior were all the right calls *at the time*.

The problem: **#1611 ("add the reddit implement section") has since merged to `main`**, and this branch now contains it via the merge. So the premise this PR was built on no longer holds. The PR's net contribution vs current `main` is a single 13-line change to `campaign.constants.ts` that **re-disables Reddit** — and `main` already has Reddit fully enabled and working.

### 🔴 Blocking — 3 (one root cause, three consequences)

1. **Re-disabling Reddit regresses a shipped feature** (`campaign.constants.ts:47`). `main` ships a complete Reddit Implement section: `implementation-tab.component.html:421-446` renders it, `@if (showRedditSection())` (html:422) binds the computed signal, and there's a real lifetime budget input (`reddit-budget`, html:446). Merging this PR removes Reddit from the Plan picker even though users can now fully configure and create Reddit campaigns on `main`.

2. **Disabling `reddit-ads` breaks the Reddit status toggle (pause/resume).** `CAMPAIGN_SERVICE_STATUS_PLATFORMS` is derived from `CAMPAIGN_PLATFORMS.filter((p) => !p.disabled)` (`campaign.controller.ts:52-53`) and gates the campaign-service toggle path at `:1019` and `:1108`. Setting `disabled: true` drops `reddit-ads` from that set, so pause/resume for existing Reddit campaigns fails while Monitor/Optimize still surface them. **This confirms Cursor Bugbot's High-severity finding and Copilot's suppressed comment — both are correct** (verified at the branch tip; there's even a mutation-pinning spec at `campaign.controller.spec.ts:1272-1273`).

3. **The new rationale comment is factually false against its own branch** (`campaign.constants.ts:34-45`). It claims "the Implement tab has no Reddit config section," "showRedditSection is computed but never bound," "canSubmit has no Reddit budget floor," and "an unbound 500/day budget signal." All four are refuted by the code in this same branch: the section exists (html:421), the signal is bound (html:422), `canSubmit` has a Reddit floor (`ts:382` + `redditBudgetIsUsable()` ts:631-633 requiring `budget >= 1`), and the budget is now a bound lifetime input, not an unbound per-day value.

### ⚪ Nit — carried over
- Round-1 import-ordering nit is **moot/resolved-by-merge** — that file left this PR's net diff (it's now `main`'s via #1611).
- The comment sentence-fragment nit is subsumed by blocking #3 (the whole comment should go).

### Recommendation
This PR looks **superseded by #1611**. The honest resolution is almost certainly to **close it** (the LFXV2-3225 gap it addressed was closed by shipping the section, not by disabling the platform). If there's still a reason to gate Reddit, it would need to be decoupled from the status-toggle allowlist and re-justified against the now-shipped section — but I don't think that's what you want here.

### Final decision
🔴 **Needs changes before approval** — recommend closing as superseded by #1611.

packages/shared/src/constants/campaign.constants.ts
```diff
@@ -32,7 +32,19 @@ export const CAMPAIGN_PLATFORMS: readonly CampaignPlatformOption[] = [
...
+  // needs that section AND an account id story: neither the Reddit nor the X platform client
+  // implements ListAdAccounts (campaign-service internal/bootstrap/sysacct.go), so the account id
+  // is hand-entered on the connection.
+  //
+  // The picker is not the only gate: implementation-tab.component.ts filters this same `disabled`
+  // flag when restoring a saved brief, because a brief stored BEFORE the flag still names the
+  // platform and the disabled tile leaves the user no way to deselect it. A brief whose every
+  // platform is disabled restores as an EMPTY selection, not the default — substituting one would
+  // dispatch a platform the user never chose.
+  { id: 'reddit-ads', label: 'Reddit Ads', icon: 'fa-brands fa-reddit', disabled: true },
```
dealako, line 47: [blocking] Re-disabling Reddit regresses a shipped feature AND breaks status toggle

Issue: Setting `disabled: true` on `reddit-ads` reverses #1611, which is already in `main` (and in this branch via the merge).
Proof: (1) `main` has a full Reddit Implement section — implementation-tab.component.html:421-446 renders it, html:422 binds `@if (showRedditSection())`, html:446 is a real `reddit-budget` input; canSubmit enforces a Reddit floor at implementation-tab.component.ts:382 via `redditBudgetIsUsable()` (ts:631-633, `budget >= 1`). (2) `CAMPAIGN_SERVICE_STATUS_PLATFORMS` is built from `CAMPAIGN_PLATFORMS.filter((p) => !p.disabled)` (campaign.controller.ts:52-53) and gates pause/resume at :1019 and :1108 — disabling reddit-ads drops it from that set.
Why it matters: users lose the ability to create Reddit campaigns that `main` currently allows, and existing Reddit campaigns lose pause/resume via campaign-service while still appearing in Monitor/Optimize. Confirms Cursor Bugbot (High) and Copilot's suppressed comment.
Fix: Do not re-disable Reddit. Close this PR as superseded by #1611. If gating is still wanted, decouple it from the status-toggle allowlist.

packages/shared/src/constants/campaign.constants.ts
dealako, line 37: [blocking] Rationale comment is factually false against its own branch

Issue: Every claim in this block is refuted by code in this same branch (which contains #1611 via the merge).
Proof: "no Reddit config section" -> exists at implementation-tab.component.html:421. "showRedditSection ... never bound" -> bound at html:422. "canSubmit has no Reddit budget floor" -> floor at implementation-tab.component.ts:382 + redditBudgetIsUsable() ts:631-633. "unbound 500/day budget signal" -> now a bound lifetime input (html:446), not per-day.
Why it matters: this comment is the sole documented justification for re-disabling a shipped platform; a future maintainer reading it would act on false premises.
Fix: Remove the comment (and the disable) — the section it says is missing has shipped.

dealako (changes requested):
Follow-up review after the main merges. Since #1611 shipped the Reddit Implement section to main, this PR's net effect — re-disabling Reddit — now regresses a working feature and breaks the Reddit status toggle. Recommend closing as superseded by #1611. Details in the summary comment; two inline blockers below.

PR #1611, opened 2026-08-18
feat(campaigns): add the reddit implement section

mrautela365:
Supersedes #1610, which is closed in favour of this. That PR disabled Reddit in the picker as a stopgap; this one **closes the gap it was working around**, so shipping both would have had one PR undo the other. Every commit from #1610 is carried here, plus the fix.

## The defect

Reddit was selectable in Plan, but the Implement tab rendered nothing for it. `showRedditSection` was computed and referenced **zero times** in the template, and `canSubmit()` used `redditSelected` only to satisfy the "something is selected" check — no budget floor.

So a Reddit campaign dispatched entirely on brief-derived values the operator never saw, including `redditBudgetUsd = signal(500)` with **no input bound to it anywhere**. Not a zero-budget failure — a silent $500/day one.

## What this adds

**A Reddit config section**, bound to the computed that had no consumer. Budget input, plus read-only geo / subreddit / interest / variant previews. The targeting is AI-recommended from the brief, so it is shown for *review* rather than made editable — that is a separate change, and the point of this one is that the operator sees what will dispatch.

**A budget floor in `canSubmit()`.** Reddit's client rejects a non-positive budget at dispatch (`internal/platform/reddit/client.go`: *"invalid budget: must be a positive number"*), and creation is **async** — so without a local guard that surfaces as a dead job the operator has to go and read, rather than as a blocked button.

**`reddit-ads` selectable again**, since the rule now holds: a platform belongs in the picker as enabled only when the Implement tab can configure it.

## Carried from #1610

**The restore path was a second gate, and it was open.** `disabled` gates the Plan tile; a stored brief reaches `selectedPlatforms` through `populateFromBrief`, which never consults the picker. `campaign-service.service.ts:1503` filters restored platforms on *membership only*. So a brief saved before any flag still restored that platform and still built its config.

Now filtered at restore. **And filtered to empty means empty** — an earlier version of that fix skipped the set on an empty result, leaving the `google-ads` default standing, so a brief for a disabled platform opened as a *Google* campaign the user never chose. That is the substitution `campaign-service.service.ts:1505-1515` rejects server-side; two bot reviews caught it independently.

## Testing

14 app specs pass, plus 1342 server specs. Three new guards, each **mutation-verified with a compiling mutation** — a build break proves nothing:

| Mutation | Caught by |
|---|---|
| Remove the Reddit budget guard | `blocks submit when the reddit budget is zero` |
| Replace the section's `@if` with `false` | `renders the reddit config section` |
| Skip the restore set when filtered empty | `selects nothing for an all-disabled brief` |

The budget spec asserts the **positive** case too, so the zero case cannot be satisfied by some unrelated guard.

**Also re-keyed the restore specs from `reddit-ads` to `twitter-ads`.** They asserted a platform was filtered as disabled — which stopped being true the moment this section landed. A test keyed to a flag that flips is a test that expires. X is disabled for a capability reason (its platform client has no `ListAdAccounts`, so no account discovery exists), which makes it the stable example.

## Note

Re-enabling Reddit does not close the account-id story: neither the Reddit nor the X platform client implements `ListAdAccounts`, so a Reddit account id stays hand-entered on the connection. That is orthogonal to this PR — the section is what the picker was promising and did not have — but worth knowing before the next platform is enabled.

Closes LFXV2-3225


mrautela365:
Force-pushed `4371a4e11` — squashed to one commit and applied the reviewer-sim findings.

### Squashed, and the title changed

The branch had 5 commits where two cancelled: one disabled `reddit-ads`, another re-enabled it. **`packages/shared/src/constants/campaign.constants.ts` now has a zero net diff** — the PR ships no picker change at all, so "and re-enable it" described the branch rather than the merge. Dropped from the title, and the intermediate states are gone. The reasoning from the disable commit survives in the squashed message.

### Four findings fixed

**A lint error that would have failed CI.** I put `redditEffectiveKeywords` among the methods, after the constructor — `@typescript-eslint/member-ordering` rejects it. Moved beside `showRedditSection`.

**The budget is a LIFETIME cap, and nothing said so.** `internal/dispatch/reddit.go` passes `lifetime=true` with the comment *"LIFETIME spend cap, not a daily one"*. The Meta and LinkedIn budgets directly above it default to **daily** and expose a checkbox to switch. An operator carrying that model across reads Reddit's `500` as per-day — a **30× misread** of committed spend on a month-long flight. Now labelled `Reddit Budget (USD, lifetime)` with helper text naming the contrast.

Worth noting my own earlier commit message said "a hard-coded $500/day budget" — the `/day` was wrong, and it was the same misreading.

**A fourth enumerating comment.** The `canSubmit` note listed "(subreddits, interests, geo)" as the read-only inputs — already false in the same commit, since `keywords` is a fourth. This branch has corrected that shape three times; it now describes the shape instead of listing members.

**The spec's cast claimed `Set<string>`** where the signal is an array. Harmless under a spread, but it is a false claim sitting inside an `as unknown as` that exists to bypass type checking.

### Deliberately not fixed

**The client's 1e9 budget ceiling is not mirrored.** Its siblings have no ceiling either, and capping one of three visually identical inputs reads as a Reddit quirk rather than a platform rule. Worth doing for all three at once; the template comment records the decision.

**Reddit targeting is read-only.** The operator can see a wrong subreddit but not remove it — LinkedIn offers add/remove controls. That converts a silent dispatch into a reviewed one, which is the actual risk here, but it is a half fix and I would rather say so than imply otherwise. Reddit is the only enabled platform whose targeting cannot be corrected in-tab.

### Verification

Lint clean, both tsconfigs clean, 14 app specs and 1342 server specs green. All three guards mutation-verified with **compiling** mutations — a build break proves nothing. One reviewer re-ran those mutations independently rather than taking the claim on trust, and confirmed all three bind.

mrautela365:
Force-pushed `81cebda27`. Two suppressed findings — one fixed here, one filed as its own ticket because it is pre-existing across three platforms.

### Fixed: geo had the same fallback problem as keywords

`submit()` falls back to `[form.countryCode]` when the brief recommends no geos, so a preview reading the raw `redditGeoTargets()` signal rendered **nothing** for a request that targets somewhere specific. Same shape I'd already fixed for keywords and missed here.

**The test caught a second, worse bug in my own fix.** My first attempt was a `computed()` reading `campaignForm.controls.countryCode.value` — a plain form read, not a reactive dependency. The computed memoised its first value and never updated when the country changed. Rewritten as a `toSignal` over `valueChanges`, following `initCampaignName`'s existing pattern in the same file.

Mutation-verified: removing the fallback fails the new spec.

### Filed separately: LFXV2-3315

The finding on `onRedditBudgetInput` — that an edited budget silently reverts to 500 on tab switch — is **real and pre-existing for Meta and LinkedIn too**. `CampaignImplementationDraft` carries only the shared/Google fields; there is no `metaBudgetUsd`, `linkedInBudgetUsd` or `redditBudgetUsd`, and none of the three `on*BudgetInput` handlers emits the draft. Verified by grep: zero `draftChange` calls in any of them.

So the Reddit input follows the established pattern rather than introducing the defect. Fixing it for Reddit alone would leave the other two silently broken and make the inconsistency harder to spot — LFXV2-3315 covers all three, with a per-platform spec and a mutation check.

Flagging it plainly rather than quietly: **a per-platform budget edit does not survive leaving this tab today, on any platform.** That is real ad spend reverting to a default nobody chose.

### Verification

Lint clean, both tsconfigs clean, 15 app specs and 1342 server specs green. Four guards now mutation-verified with compiling mutations: the budget floor, the section render, the restore filter, and the geo fallback.

mrautela365:
Force-pushed. Correct catch, and my previous fix was wrong in the mirror-image way.

I replaced a non-reactive `computed` read with an RxJS pipeline keyed on `countryCode.valueChanges` — which tracks the country but **not** `redditGeoTargets()`. And the ordering makes that fatal: `populateFromBrief` patches the form first and assigns recommended geos at lines 849/856, *after*. So the pipeline fired once on the country, read an empty geo list, and nothing re-ran it when the real geos arrived. A brief with recommended geos would preview the country fallback while `submit()` dispatched the recommended list — the preview disagreeing with the request, which is worse than the empty list I started with.

Both inputs are signals now: `countryCodeValue` wraps `valueChanges` in `toSignal`, and the derivation is a plain `computed` over it and `redditGeoTargets()`.

**A new spec pins the arrival order specifically** — set the country, render, *then* assign geos, and assert the chips show `DE`/`FR` and not the fallback. Mutation-verified: reverting the country to a non-reactive read fails it.

One thing worth flagging about that spec. My first version asserted `not.toContain('US')` over the section's whole `textContent` — which **can never pass**, because the budget label reads "(USD, lifetime)" and `'USD'` contains `'US'`. It failed against correct code and looked like a real defect. The assertion is now scoped to the geo chips by their own text, with a comment saying why, so the next person does not reintroduce a test that cannot pass.

Lint clean, both tsconfigs clean, 16 app specs and 1342 server specs green. Five guards on this PR now mutation-verified with compiling mutations.

mrautela365:
Force-pushed `77d107bd6`. Round four — three findings, and I've reversed an earlier decision rather than defending it.

### New and real: a cleared country code dispatched `['']`

`countryCode` carries no validator, so clearing it is reachable. In that state `submit()` sent `geoTargets: ['']` — a value Reddit cannot target — while the preview correctly rendered nothing. **The preview and the request disagreed**, which is worse than either being wrong alone.

Fixed at the source rather than patched twice: `submit()` now reads the same `redditEffectiveGeoTargets()` computed the preview renders, so they cannot drift. Plus a `canSubmit` guard refusing a Reddit selection with no usable geo at all.

Worth noting the same `[form.countryCode]` fallback exists for **Google (line 615) and Meta (line 667)** — pre-existing, and out of scope here, but the exposure is identical.

### Reversed: the budget ceiling is now enforced

I declined this last round, arguing a cap on one of three visually identical inputs reads as a Reddit quirk. **That was the wrong trade and the bot was right to raise it twice.**

The argument I used to justify the *floor* applies unchanged to the ceiling: creation is async, so an over-cap budget is not a validation error the operator sees — it is a job that dies later and has to be gone and read. Consistency with siblings is worth less than not queueing a doomed job.

Now mirrors both arms of the client's contract — finite, and within `(0, 1e9]` — via a shared `REDDIT_MAX_BUDGET_USD` constant documenting where the number comes from. The siblings still lack both bounds; LFXV2-3315 covers bringing them into line.

### Re-report: the budget draft persistence

Already dispositioned as **LFXV2-3315**. It is real, but pre-existing for Meta and LinkedIn too — `CampaignImplementationDraft` carries no per-platform budget at all, and none of the three `on*BudgetInput` handlers emits the draft. Fixing Reddit alone would leave the other two silently broken and the inconsistency harder to spot.

### Verification

Lint clean, both tsconfigs clean, 19 app specs and 1342 server specs green. Eight guards on this PR now mutation-verified with compiling mutations — including the two added this round, where dropping the ceiling check fails both new cases.

mrautela365:
## Addressed the suppressed review findings

Commit 6281df17e. Both code findings from the 13:42Z round verified against current code and fixed; each fix is pinned by a test that fails when the fix is reverted (checked, not assumed).

### Fixed

**Geo fallback chosen before validation** — correct, and it was reachable. A brief carrying a non-empty but unusable recommendation like `['USA']` won the branch on length alone, filtered to `[]`, and left `canSubmit()` blocking the Reddit section permanently, with a valid country code sitting unread in the form. Now normalises first and falls back only when nothing usable remains — non-empty was never the same test as usable. Test: *falls back to the country code when the recommended geos are all unusable*.

**`r/` double prefix** — correct, including the cited evidence. `campaign-service.service.spec.ts` does assert `'r/k8s'` survives restore verbatim, so the fixed template prefix rendered `r/r/k8s`. Stripped once now. I also changed the **submitted payload** to use the same normalised list — the preview and the dispatch disagreeing is the same defect whichever side is wrong, and `geoTargets` already did it this way. Test: *strips an r/ prefix so the preview matches what dispatches*.

### Not changed here

**Reddit budget lost on tab switch** — real, and confirmed: `CampaignImplementationDraft` carries no Reddit budget, so a tab switch resets it to $500. It is **not Reddit-specific** — Meta and LinkedIn have the identical gap, and fixing only Reddit would leave the same defect twice over. Tracked as **LFXV2-3315** to fix the draft round-trip for all three platforms together, since the fix belongs in the shared interface rather than in this section.

### From the earlier (08:36Z) round

Those comments predate the current head and describe code that has since changed — the stale ceiling comment and the `redditBudgetIsUsable` bounds were addressed in 090d13b74. The $1 floor is deliberate and stays: it is a product minimum, not a claim about the upstream contract, and the wording no longer says otherwise.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.html
```diff
@@ -418,6 +418,87 @@ <h3 class="text-sm font-semibold text-gray-900">Meta Ads</h3>
...
+                (input)="onRedditBudgetInput($event)"
+                min="1"
+                class="rounded-md border border-gray-300 px-3 py-2 text-sm text-gray-900 focus:border-blue-500 focus:outline-none focus:ring-1 focus:ring-blue-500"
+                data-testid="implementation-reddit-budget" />
+              <p class="text-xs text-gray-500">Charged in USD. Reddit campaigns are created paused.</p>
+            </div>
+          </div>
+
+          <!-- Targeting, all AI-recommended from the brief. Rendered read-only so the operator can
+               REVIEW what will dispatch; editing them is a separate change. -->
```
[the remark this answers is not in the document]
mrautela365: Already addressed on the current head (6281df1). The Reddit section renders a Keywords block driven by the effective list, not the raw signal:

```html
@if (redditEffectiveKeywords().length > 0) {
  <label ...>Keywords ({{ redditEffectiveKeywords().length }})</label>
  @for (kw of redditEffectiveKeywords(); track kw) { ... }
}
```

`redditEffectiveKeywords` mirrors the exact fallback `submit()` uses (`implementation-tab.component.ts:278-280`): brief-supplied Reddit keywords when present, otherwise `briefKeywords().map(k => k.term)` — the same expression at line 698 that builds the dispatched `keywords`. So what is previewed is what dispatches. Resolving as fixed.

mrautela365:
## On the Reddit budget persistence finding

Raised three times now, and correct each time. Answering properly rather than deferring again.

**It cannot be fixed on this PR.** `CampaignImplementationDraft` has no `redditBudgetUsd` field on this branch (`grep -c redditBudgetUsd packages/shared/src/interfaces/campaign.interface.ts` → 0), so there is nothing here to emit into. The draft interface is being widened on **`feat/LFXV2-3315`**, which adds all five per-platform budget fields plus the emit/restore plumbing.

**What that work found, which is the more interesting half:** the budgets are *signals*, not form controls, so `emitDraft()` — driven entirely by `campaignForm.valueChanges` — never fired at all when only a budget changed. Adding the restore alone would still have lost the value, just by a different route. That branch therefore also adds `this.emitDraft()` to each per-platform handler.

**And a gap neither PR can see alone**, which I have queued rather than left implicit: `onRedditBudgetInput` exists only on THIS branch, while the emit plumbing exists only on 3315. On merge, Reddit would become the one platform whose budget still silently reverts. The handler needs the same `emitDraft()` call its Meta and LinkedIn siblings get:

```ts
protected onRedditBudgetInput(event: Event): void {
  this.redditBudgetUsd.set((event.target as HTMLInputElement).valueAsNumber || 0);
  this.emitDraft();
}
```

That lands on 3315 when the branches meet, with a round-trip test driven **through the DOM input** rather than by setting the signal — the point made in the sibling comment about the current budget test not being DOM-driven is right, and a signal-driven test would not catch a missing binding.

Also fixed on this push: the subreddit block gated on `redditSubreddits()` while its count and chips rendered `redditEffectiveSubreddits()`, so a brief whose names all normalise away showed "Subreddits (0)" above an empty row. My own `r/` normalisation introduced that — it changed what the block renders without changing what it gates on. Test added and confirmed to fail when the gate is reverted.

dealako:
👋 @mrautela365 — this is a genuinely exemplary PR. The write-up traces the actual defect (a Reddit campaign dispatching on hidden brief defaults, including an unbound `$500` lifetime budget the operator never saw), and the fix is scoped tightly to closing that gap: a bound budget input with a floor + ceiling + finiteness guard, read-only previews whose `effective-*` computeds are single-sourced against `submit()` so preview and dispatch cannot drift, and restore filtering that drops disabled platforms *and* refuses to fall through to a `google-ads` default. Six review rounds of bot findings are each either fixed on the current head or dispositioned to LFXV2-3315 with a stated reason, and every new guard is mutation-verified with a compiling mutation. The three-subagent independent pass (security / correctness / data-privacy / performance / style) surfaced nothing blocking.

### Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 2 — keyword fallback re-inlined in `submit()` instead of reusing the computed; orphaned/duplicate JSDoc block above `countryCodeValue`
- ⚪ Nit: 1 — Reddit preview test gaps (budget `=== 1` boundary, keyword brief-fallback, interests/variant render)
- ❔ Question: 1 — `canSubmit()` reads shared form fields non-reactively (pre-existing; inherited by the new Reddit path)

### Bot reconciliation

Cross-referenced Copilot + Cursor Bugbot across all six rounds. **Agree** with every finding that landed, and all are resolved or dispositioned on the current head (5741b54):

- Stale geo preview (Cursor) → fixed via `countryCodeValue` `toSignal` + `computed`. ✅
- Subreddit block gated on raw list (Cursor) → fixed in `5741b54` to gate on the effective list. ✅
- Missing keyword preview (Copilot) → keyword block added, driven by `redditEffectiveKeywords()`. ✅
- Budget test not DOM-driven / budget lost on tab switch (Copilot) → dispositioned to **LFXV2-3315**; verified the gap is shared by Meta + LinkedIn (`CampaignImplementationDraft` carries no per-platform budget), so fixing Reddit alone would leave the other two silently broken. Reasonable trade-off, clearly documented.

### Notes on my two minor findings

Both are consistency/documentation only — no behavioral defect today. The re-inlined keyword fallback at `submit()` is currently *identical* to the computed, so there is no live divergence; the risk is future drift, which is precisely the failure mode the `effective-*` computeds exist to prevent — and geo/subreddit already single-source correctly, so keywords is the lone exception. Neither blocks merge.

### Final decision

✅ **Approved with minor comments** — the two minors and the coverage nit are worth a quick follow-up but should not hold the merge. Excellent work.

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
```diff
@@ -582,8 +686,14 @@ export class ImplementationTabComponent implements OnInit {
...
+              // The SAME derivation the preview renders. Reading `[form.countryCode]` here instead
+              // would let the two disagree: a cleared country code previews nothing and dispatches
+              // [''], which the operator cannot see and Reddit cannot use.
+              geoTargets: this.redditEffectiveGeoTargets(),
+              // The normalised list, matching the preview above for the same reason geoTargets
+              // does: dispatch strips an optional `r/`, so sending the raw value would submit
+              // something the operator was never shown.
+              subreddits: this.redditEffectiveSubreddits(),
               interests: this.redditInterests(),
               keywords: this.redditKeywords().length > 0 ? this.redditKeywords() : this.briefKeywords().map((k) => k.term),
```
dealako, line 715: [minor] submit() re-inlines the Reddit keyword fallback instead of reusing the computed

Issue: `geoTargets` (`redditEffectiveGeoTargets()`) and `subreddits` (`redditEffectiveSubreddits()`) reuse their preview computeds, but `keywords` re-inlines the ternary verbatim.
Proof: `redditEffectiveKeywords` at :279 is `this.redditKeywords().length > 0 ? this.redditKeywords() : this.briefKeywords().map((k) => k.term)` — the exact expression repeated here at :698. Its own doc (:272) says it "Mirrors submit()'s own fallback."
Why it matters: two of three effective previews are single-sourced so preview and dispatch cannot diverge; keywords is the lone exception. If the fallback rule ever changes in one place, the preview chips and the dispatched keywords silently drift — the exact bug class the effective-* computeds were introduced to prevent. No live divergence today since the expressions are identical.
Fix: `keywords: this.redditEffectiveKeywords(),`

apps/lfx-one/src/app/modules/dashboards/campaigns/components/implementation-tab/implementation-tab.component.ts
```diff
@@ -264,6 +267,69 @@ export class ImplementationTabComponent implements OnInit {
...
+   * targeting it exists to surface.
+   */
+  protected readonly redditEffectiveKeywords = computed<string[]>(() =>
+    this.redditKeywords().length > 0 ? this.redditKeywords() : this.briefKeywords().map((k) => k.term)
+  );
+
+  /**
+   * The geo targets a Reddit dispatch will actually carry.
+   *
+   * Same reason as `redditEffectiveKeywords`: `submit()` falls back to the form's country code
```
dealako, line 302: [minor] Orphaned/duplicate JSDoc block attached to the wrong symbol

Issue: this JSDoc ("The geo targets a Reddit dispatch will actually carry. / Same reason as `redditEffectiveKeywords`…") sits directly above `countryCodeValue`, which then has its own separate JSDoc ("The form's country code as a SIGNAL.") immediately below it — so two consecutive doc blocks precede one field, and the first describes a symbol it isn't attached to.
Proof: the canonical copy of this same geo description already lives at :305-310 above `redditEffectiveGeoTargets`. This block (:283-289) is a leftover from splitting the geo derivation off the country-code signal.
Why it matters: an IDE hover on `countryCodeValue` may surface the stray geo-targets comment, and the duplicated text invites drift if one copy is later edited.
Fix: delete the orphaned block at :283-289; the real doc is already at :305-310.

dealako (dismissed):
Independent three-subagent review (security / data-privacy / correctness / performance / style) plus reconciliation with all six rounds of Copilot + Cursor Bugbot findings. No blocking, security, or data-privacy issues. Two minor consistency/documentation items inline (no behavioral defect today) and one test-coverage nit — all fine as follow-ups. Approving.

dealako:
👋 @mrautela365 — follow-up pass. Heads up on what actually changed since my approval at `5741b54`: the only new commit is a **merge of `origin/main`** into the branch (`407f02b`) — no new commits touched the Reddit section. GitHub auto-dismissed my prior ✅ approval purely because HEAD advanced, not because anything regressed.

### 👏 What I verified this round

- The merge is **clean** — no conflict markers in any of the four PR files, and the Reddit section is **byte-for-byte identical** to the state I approved. The PR's diff against `main` is unchanged at +530/-4.
- The merge pulled in unrelated work from `main` (email template picker #1596, ISO country-list expansion #1627, etc.), but **none of it lands in this PR's diff** — those are already in `main`, so `origin/main...HEAD` correctly excludes them. No integration drift into `normaliseGeoCodes` or `submit()`'s `redditConfig`.

### Revision tracking (from my last round)

- ⚠️ **Still open** — `[minor]` keyword fallback re-inlined in `submit()` (`implementation-tab.component.ts:698`) instead of reusing `redditEffectiveKeywords()`. Unaddressed (code unchanged).
- ⚠️ **Still open** — `[minor]` orphaned/duplicate JSDoc block above `countryCodeValue` (`:283-289`). Unaddressed (code unchanged).

Both remain **non-behavioral** (consistency/docs; no live divergence today), so I'm keeping them at `[minor]` rather than mechanically escalating — this round introduced no new code for them to have been ignored in. Still worth a one-line cleanup whenever you next touch the file.

### Bot reconciliation (new since last round)

Copilot re-reviewed `407f02b` and generated **no new comments**; its 2 suppressed notes are both on **unchanged** code:
- Submit-level spec coverage (assert `redditConfig.geoTargets/subreddits/keywords` use the effective values) — **agree**, same shape as my earlier test nit; a good follow-up, not blocking.
- Case-insensitive subreddit dedup (`['K8s','k8s']` previews as 2, dispatches 1) — the premise is a claim about the **Go Reddit client**, which lives in a separate service repo and isn't verifiable here. Low severity, on unchanged code, cosmetic preview-count only. Noting for awareness; not blocking.

### Issue count (carried over + new)

- 🔴 Blocking: 0
- 🟡 Minor: 2 — both carried over, both unaddressed, both non-behavioral (keyword single-sourcing; orphaned JSDoc)
- ⚪ Nit: 1 — submit-level Reddit-config spec coverage (per Copilot; agree)

### Final decision

✅ **Approved with minor comments** — re-affirming the prior approval. Nothing regressed in the merge; the two open minors are cleanup-grade and don't hold the merge.

dealako (approved):
Follow-up pass. The only new commit since my prior approval (5741b54) is a merge of origin/main into the branch — no new authored work on the Reddit section, and the PR's diff against main is byte-for-byte identical to the approved state. Verified the merge is clean (no conflict markers, no drift into normaliseGeoCodes or submit()'s redditConfig). Two prior minor items (keyword single-sourcing at :698, orphaned JSDoc at :283-289) remain open but are non-behavioral cleanup — kept at minor. New Copilot pass generated no new comments (2 suppressed, both on unchanged code). Re-affirming approval.

PR #1629, opened 2026-08-18
fix(profile): send deletedBy in CDP work-experience delete

### LINKED ISSUE #1628: Send deletedBy in CDP work-experience delete request
## Summary

CDP is shipping a breaking change to `DELETE /members/:memberId/work-experiences/:workExperienceId`:
the endpoint will now **require** a `deletedBy` field in the request body. This lets CDP track who
initiated the deletion and prevents enrichment from recreating work experiences that were intentionally
deleted. `deletedBy` behaves like the `verifiedBy` field already sent on the sibling create / update /
confirm work-experience calls.

Today our delete call sends **no body at all**, so once CDP deploys the change the delete will start
failing (and deleted work experiences may be silently recreated by enrichment).

## Expected behavior

The outbound `DELETE` request includes `Content-Type: application/json` and a body of
`{"deletedBy":"<identifier>"}`, using the same server-side–resolved identifier (`lfid`) we already
send as `verifiedBy`.

## Scope

Backend only (thin proxy layer). No client/UI or shared-type changes.

- `apps/lfx-one/src/server/services/cdp.service.ts` — `deleteWorkExperience` / `deleteWorkExperienceForUser`

## References

- Upstream CDP OpenAPI / API spec updated to require `deletedBy` on this endpoint

fayazg:
## Summary

- CDP is making `deletedBy` **required** in the body of `DELETE /members/:memberId/work-experiences/:workExperienceId`. It lets CDP track who initiated the deletion and prevents enrichment from recreating work experiences that were intentionally deleted — behaving like the `verifiedBy` field on the sibling create / update / confirm calls.
- Our delete previously sent **no body**, so it would start failing once CDP deploys the change.
- `deleteWorkExperience()` now accepts a `deletedBy` argument, sends `Content-Type: application/json`, and includes `{ "deletedBy": "<lfid>" }` in the request body — mirroring `confirmWorkExperience`'s `{ verified, verifiedBy }` shape.
- `deleteWorkExperienceForUser()` forwards its already-resolved identifier as `deletedBy` (the controller passes the effective `lfid`, the same value used for `verifiedBy` on create/update). No controller, client, or shared-type changes were needed.

## Related

- Upstream CDP / crowd.dev change requiring this field: linuxfoundation/crowd.dev#4487

Fixes #1628

### DIFF
--- a/apps/lfx-one/src/server/services/cdp.service.ts
+++ b/apps/lfx-one/src/server/services/cdp.service.ts
@@ -531,7 +531,7 @@ export class CdpService {
   /**
    * Delete a specific CDP work experience
    */
-  public async deleteWorkExperience(req: Request | undefined, memberId: string, workExperienceId: string): Promise<void> {
+  public async deleteWorkExperience(req: Request | undefined, memberId: string, workExperienceId: string, deletedBy: string): Promise<void> {
     const token = await this.generateToken(req);
     const url = `${this.cdpApiUrl}${CDP_CONFIG.ENDPOINTS.MEMBER_WORK_EXPERIENCES(memberId)}/${encodeURIComponent(workExperienceId)}`;
     const requestId = randomUUID();
@@ -539,14 +539,17 @@ export class CdpService {
     logger.debug(req, 'delete_cdp_work_experience', 'Deleting CDP work experience', {
       member_id: memberId,
       work_experience_id: workExperienceId,
+      deleted_by: deletedBy,
     });
 
     const response = await fetch(url, {
       method: 'DELETE',
       headers: {
+        'Content-Type': 'application/json',
         Authorization: `Bearer ${token}`,
         'X-LFX-Request-ID': requestId,
       },
+      body: JSON.stringify({ deletedBy }),
       signal: AbortSignal.timeout(10000),
     });
 
@@ -565,7 +568,7 @@ export class CdpService {
    */
   public async deleteWorkExperienceForUser(req: Request | undefined, sub: string, workExperienceId: string): Promise<void> {
     const memberId = await this.resolveMember(req, [sub]);
-    await this.deleteWorkExperience(req, memberId, workExperienceId);
+    await this.deleteWorkExperience(req, memberId, workExperienceId, sub);
   }
 
   /**


dealako:
Hi @fayazg 👋 — thanks for getting ahead of the upstream CDP contract change.

**Overall impression:** Clean, minimal, surgical fix. `deleteWorkExperience` now mirrors the established `confirmWorkExperience` shape exactly — required actor arg, `Content-Type: application/json`, and `{ deletedBy }` in the body — and `deleteWorkExperienceForUser` forwards the already-resolved effective LFID the same way `verifiedBy` is derived on create/update/confirm. I traced the value flow end-to-end (controller `resolveEffectiveLfid` → `deleteWorkExperienceForUser(sub=lfid)` → `deleteWorkExperience(deletedBy)` → CDP body) and enumerated callers: the only caller of `deleteWorkExperience` is `deleteWorkExperienceForUser`, which is updated, so there are no orphaned call sites. The controller's existing `if (!sub)` guard (profile.controller.ts:1388) ensures `deletedBy` is never empty.

**Issue count:**

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — no unit test asserts the request body now carries `deletedBy`; noting only because there's no pre-existing `cdp.service` delete test to extend, so this matches the current coverage baseline rather than regressing it.
- ❔ Question: 1 — under impersonation, `deletedBy` records the *effective* (target) LFID rather than the acting admin. This is the identical, intentional convention already used by `verifiedBy` on the sibling calls, so it's consistent by design — flagging only so the audit semantics are a conscious choice on CDP's side.

**Reconciliation with bots:** Cursor Bugbot (Low Risk) and Copilot both summarized the change without raising findings; CodeRabbit auto-review was skipped by repo config. I agree with their low-risk read — nothing they flagged, nothing they missed.

**Final decision:** ✅ **Approved with minor comments** — the nit and question are optional and non-blocking.

dealako (approved):
Approved with minor comments — see summary comment. Clean mirror of the confirmWorkExperience pattern; end-to-end value flow and caller enumeration verified. One optional nit (no delete-body test, matching existing coverage baseline) and one design question (impersonation attribution, consistent with the verifiedBy convention), both non-blocking.

PR #1648, opened 2026-08-19
fix(committees): keep failed member/invite operations staged on flush

### LINKED ISSUE #1608: Failed invites/members are lost with no retry on wizard flush failure
## Background

When the create-group wizard's Step 4 flush (`Done`, or `Update` in edit mode) partially or fully fails, the staged invites and member changes that failed are silently discarded — the user is navigated away with no indication of which recipients or members failed, and no way to retry without re-entering everything from scratch.

## Requirement

On a partial or full flush failure, the failed invites/members should remain staged (visible and editable) instead of being lost, so the user can retry without re-entering data. The wizard should not navigate away while there are unresolved failures.

## Notes

Surfaced during review of PR #1078 (LFXV2-2606, bulk email invite in create-group wizard). Flush results are currently anonymous success/fail counts with no identity (email/uid) tracked per operation, and `onDone()`/`onSubmitAll()` in `committee-manage.component.ts` navigate away unconditionally once the flush completes.

**Related:** LFXV2-2606.

---
**Migrated from:** LFXV2-3268

MRashad26:
## Summary

When the committee-manage wizard flushes staged member/invite operations (adds, updates, deletes, invites) and some fail, the failures were previously anonymized to aggregate success/fail counts and the wizard navigated away unconditionally — silently losing the failed items with no way to retry without re-entering data.

- Threads an identifier (member email/uid, invite email) through each flushed operation instead of anonymous counts, so a failure can be traced back to the exact staged item that produced it.
- Gates wizard navigation on `flushMemberUpdates()` returning zero failures — the wizard no longer navigates away while any operation failed.
- Adds `pruneSucceeded()` on the members-manager child component, which removes only the operations that landed server-side and leaves failed items staged, visible, and editable for retry.
- Guards `computeSucceededOperations()` against an undefined `identifier` (review fix from the first commit).

## Test plan

- [x] `yarn format:check` — clean
- [x] `yarn lint:check` — clean (one pre-existing unrelated warning in `user.service.ts`)
- [x] `yarn build` — succeeds
- [x] Reviewed by general/self-serve-convention/learnings reviewer trio, both per-commit and full-branch sweep — all clean

Fixes linuxfoundation/lfx-self-serve#1608

### DIFF
--- a/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
@@ -8,7 +8,14 @@ import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angula
 import { ActivatedRoute, Router, RouterLink } from '@angular/router';
 import { ButtonComponent } from '@components/button/button.component';
 import { COMMITTEE_FORM_STEPS, COMMITTEE_INVITE_CONCURRENCY, COMMITTEE_LABEL, COMMITTEE_STEP_TITLES, COMMITTEE_TOTAL_STEPS } from '@lfx-one/shared/constants';
-import { Committee, MemberOperationResult, MemberOperationType, MemberPendingChanges, SucceededMemberOperations } from '@lfx-one/shared/interfaces';
+import {
+  Committee,
+  CommitteeMember,
+  MemberOperationResult,
+  MemberOperationType,
+  MemberPendingChanges,
+  SucceededMemberOperations,
+} from '@lfx-one/shared/interfaces';
 import { CommitteeService } from '@services/committee.service';
 import { ProjectContextService } from '@services/project-context.service';
 import { MessageService } from 'primeng/api';
@@ -553,7 +560,14 @@ export class CommitteeManageComponent {
     // Add create operation if there are members to add
     if (memberUpdates.toAdd.length > 0) {
       for (const member of memberUpdates.toAdd) {
-        operations.push(this.createMemberOperation('add', member.email, () => this.committeeService.createCommitteeMember(committeeId, member)));
+        operations.push(
+          this.createMemberOperation(
+            'add',
+            member.email,
+            () => this.committeeService.createCommitteeMember(committeeId, member),
+            (createdMember) => createdMember
+          )
+        );
       }
     }
 
@@ -587,9 +601,14 @@ export class CommitteeManageComponent {
     return concat(memberOps$, inviteOps$).pipe(toArray());
   }
 
-  private createMemberOperation(type: MemberOperationType, identifier: string, operation: () => Observable<unknown>): Observable<MemberOperationResult> {
+  private createMemberOperation<T>(
+    type: MemberOperationType,
+    identifier: string,
+    operation: () => Observable<T>,
+    captureMember?: (result: T) => CommitteeMember
+  ): Observable<MemberOperationResult> {
     return operation().pipe(
-      switchMap(() => of({ type, identifier, success: true })),
+      switchMap((result) => of({ type, identifier, success: true, createdMember: captureMember?.(result) })),
       catchError(() => of({ type, identifier, success: false }))
     );
   }
@@ -597,14 +616,16 @@ export class CommitteeManageComponent {
   /** Groups the identifiers of successfully-flushed operations so the members-manager child can prune them (GH-1608). */
   private computeSucceededOperations(results: MemberOperationResult[]): SucceededMemberOperations {
     const succeeded = results.filter((result) => result.success);
-    const normalizedEmails = (type: MemberOperationType) =>
-      new Set(succeeded.filter((result) => result.type === type).map((result) => (result.identifier ?? '').trim().toLowerCase()));
 
     return {
-      addedEmails: normalizedEmails('add'),
+      addedMembers: new Map(
+        succeeded
+          .filter((result) => result.type === 'add' && result.createdMember)
+          .map((result) => [(result.identifier ?? '').trim().toLowerCase(), result.createdMember!])
+      ),
       updatedUids: new Set(succeeded.filter((result) => result.type === 'update').map((result) => result.identifier)),
       deletedUids: new Set(succeeded.filter((result) => result.type === 'delete').map((result) => result.identifier)),
-      invitedEmails: normalizedEmails('invite'),
+      invitedEmails: new Set(succeeded.filter((result) => result.type === 'invite').map((result) => (result.identifier ?? '').trim().toLowerCase())),
     };
   }
 
--- a/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
@@ -598,7 +598,7 @@ export class CommitteeManageComponent {
   private computeSucceededOperations(results: MemberOperationResult[]): SucceededMemberOperations {
     const succeeded = results.filter((result) => result.success);
     const normalizedEmails = (type: MemberOperationType) =>
-      new Set(succeeded.filter((result) => result.type === type).map((result) => result.identifier.trim().toLowerCase()));
+      new Set(succeeded.filter((result) => result.type === type).map((result) => (result.identifier ?? '').trim().toLowerCase()));
 
     return {
       addedEmails: normalizedEmails('add'),
--- a/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
@@ -21,7 +21,7 @@ import { ProjectContextService } from '@services/project-context.service';
 import { MessageService } from 'primeng/api';
 import { ConfirmDialogModule } from 'primeng/confirmdialog';
 import { StepperModule } from 'primeng/stepper';
-import { BehaviorSubject, catchError, concat, EMPTY, filter, finalize, from, map, mergeMap, Observable, of, switchMap, take, toArray } from 'rxjs';
+import { catchError, concat, EMPTY, filter, finalize, from, map, mergeMap, Observable, of, switchMap, take, toArray } from 'rxjs';
 import { getHttpErrorDetail } from '@shared/utils/http-error.utils';
 import { evictOnWriteAccessLoss } from '@shared/utils/evict-on-write-access-loss.util';
 
@@ -63,7 +63,12 @@ export class CommitteeManageComponent {
 
   // Member management state
   public memberUpdates = signal<MemberPendingChanges>({ toAdd: [], toUpdate: [], toDelete: [], toInvite: [] });
-  public memberUpdatesRefresh$ = new BehaviorSubject<void>(undefined);
+  /**
+   * uids of members whose delete failed on the most recent flush, owned here (rather than in the
+   * members-manager child) so the failure survives the child remounting when the PrimeNG step
+   * panel destroys/recreates it on navigation away from and back to step 4 (GH-1608 review).
+   */
+  public failedDeleteUids = signal<string[]>([]);
   // Reference to the mounted members-manager child so a partial flush failure can prune
   // successfully-applied items out of its local state without navigating away (GH-1608).
   private readonly membersManager = viewChild(CommitteeMembersManagerComponent);
@@ -256,6 +261,7 @@ export class CommitteeManageComponent {
         // Don't navigate away while any staged item failed — re-entering the wizard would lose
         // it. Prune the items that did succeed so a retry doesn't resubmit them (GH-1608).
         if (results.some((result) => !result.success)) {
+          this.failedDeleteUids.set(results.filter((result) => result.type === 'delete' && !result.success).map((result) => result.identifier));
           this.membersManager()?.pruneSucceeded(this.computeSucceededOperations(results));
           return;
         }
@@ -333,6 +339,9 @@ export class CommitteeManageComponent {
           // Don't navigate away while any staged item failed — re-entering the wizard would lose
           // it. Prune the items that did succeed so a retry doesn't resubmit them (GH-1608).
           if (memberResults.some((memberResult) => !memberResult.success)) {
+            this.failedDeleteUids.set(
+              memberResults.filter((memberResult) => memberResult.type === 'delete' && !memberResult.success).map((memberResult) => memberResult.identifier)
+            );
             this.membersManager()?.pruneSucceeded(this.computeSucceededOperations(memberResults));
             return;
           }
@@ -619,9 +628,9 @@ export class CommitteeManageComponent {
 
     return {
       addedMembers: new Map(
-        succeeded
-          .filter((result) => result.type === 'add' && result.createdMember)
-          .map((result) => [(result.identifier ?? '').trim().toLowerCase(), result.createdMember!])
+        succeeded.flatMap((result) =>
+          result.type === 'add' && result.createdMember ? [[(result.identifier ?? '').trim().toLowerCase(), result.createdMember] as const] : []
+        )
       ),
       updatedUids: new Set(succeeded.filter((result) => result.type === 'update').map((result) => result.identifier)),
       deletedUids: new Set(succeeded.filter((result) => result.type === 'delete').map((result) => result.identifier)),
--- a/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
@@ -2,13 +2,13 @@
 // SPDX-License-Identifier: MIT
 
 import { HttpErrorResponse } from '@angular/common/http';
-import { Component, computed, inject, signal } from '@angular/core';
+import { Component, computed, inject, signal, viewChild } from '@angular/core';
 import { toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
 import { ActivatedRoute, Router, RouterLink } from '@angular/router';
 import { ButtonComponent } from '@components/button/button.component';
 import { COMMITTEE_FORM_STEPS, COMMITTEE_INVITE_CONCURRENCY, COMMITTEE_LABEL, COMMITTEE_STEP_TITLES, COMMITTEE_TOTAL_STEPS } from '@lfx-one/shared/constants';
-import { Committee, MemberPendingChanges } from '@lfx-one/shared/interfaces';
+import { Committee, MemberOperationResult, MemberOperationType, MemberPendingChanges, SucceededMemberOperations } from '@lfx-one/shared/interfaces';
 import { CommitteeService } from '@services/committee.service';
 import { ProjectContextService } from '@services/project-context.service';
 import { MessageService } from 'primeng/api';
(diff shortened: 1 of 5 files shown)


apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.ts
```diff
@@ -316,6 +317,32 @@ export class CommitteeMembersManagerComponent implements OnInit {
...
+          if (m.state === 'new') {
+            return !succeeded.addedEmails.has((m.email ?? '').trim().toLowerCase());
+          }
+          if (m.state === 'deleted') {
+            return !succeeded.deletedUids.has(m.uid);
+          }
+          return true;
+        })
+        .map((m) => (m.state === 'modified' && succeeded.updatedUids.has(m.uid) ? this.createMemberWithState(m, 'existing') : m))
+    );
```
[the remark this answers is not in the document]
MRashad26: Done — captured the server-assigned `CommitteeMember` from a successful add via a new `captureMember` callback on `createMemberOperation`, so the real `uid` isn't discarded. `pruneSucceeded()` now promotes the matching staged `new` row to `existing` using that captured record instead of leaving it as an orphaned unsynced add.

See commit 10bb51b9e: apps/lfx-one/.../committee-manage.component.ts + committee-members-manager.component.ts.

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
```diff
@@ -567,16 +587,30 @@ export class CommitteeManageComponent {
...
   }
 
-  private showMemberOperationToast(results: { type: string; success: number; failed: number }[]): void {
-    const totalSuccess = results.reduce((sum, result) => sum + result.success, 0);
-    const totalFailed = results.reduce((sum, result) => sum + result.failed, 0);
+  /** Groups the identifiers of successfully-flushed operations so the members-manager child can prune them (GH-1608). */
+  private computeSucceededOperations(results: MemberOperationResult[]): SucceededMemberOperations {
+    const succeeded = results.filter((result) => result.success);
+    const normalizedEmails = (type: MemberOperationType) =>
+      new Set(succeeded.filter((result) => result.type === type).map((result) => (result.identifier ?? '').trim().toLowerCase()));
```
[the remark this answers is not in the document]
MRashad26: Good catch on the theoretical collision — you're right that keying succeeded-add results by email means two staged rows with the same email would resolve indistinguishably.

That said, I don't see a current path in the UI that lets someone stage two `new` member rows with the same email in a single wizard session (the add-member dialog / manager component don't allow it today), so this doesn't have a live trigger right now. Given that, I'm leaving identity as email-based for this PR rather than introducing a tempId-based identity redesign to guard against a case that can't currently occur — happy to revisit if a future change (e.g., bulk CSV import) makes duplicate-email staging possible.

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.ts
```diff
@@ -316,6 +317,32 @@ export class CommitteeMembersManagerComponent implements OnInit {
...
+   */
+  public pruneSucceeded(succeeded: SucceededMemberOperations): void {
+    this.membersWithState.update((members) =>
+      members
+        .filter((m) => {
+          if (m.state === 'new') {
+            return !succeeded.addedEmails.has((m.email ?? '').trim().toLowerCase());
+          }
+          if (m.state === 'deleted') {
+            return !succeeded.deletedUids.has(m.uid);
```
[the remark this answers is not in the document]
MRashad26: Fixed — a failed delete now stays visible with a "Failed to remove" badge instead of silently disappearing. Added a `failedDeleteUids` signal populated in `pruneSucceeded()` and a `visibleMembers` computed that keeps failed-delete rows in the list; the row remains staged in `deleted` state so it's automatically resubmitted on the next flush. No explicit retry/cancel button this round — relying on the existing auto-retry-on-next-flush behavior.

See commit 10bb51b9e: committee-members-manager.component.ts + .html.

MRashad26:
## Review Feedback Addressed

Commit: 10bb51b9e

### Changes Made
- **committee-manage.component.ts / committee-members-manager.component.ts / member.interface.ts**: captured the server-assigned `CommitteeMember` from a successful add (`captureMember` callback on `createMemberOperation`) instead of discarding it; `pruneSucceeded()` now promotes the matching staged `new` row to `existing` using the real `uid` (per @cursor)
- **committee-members-manager.component.ts / .html**: added `failedDeleteUids` tracking + `visibleMembers` computed so a failed delete stays visible with a "Failed to remove" badge instead of disappearing; still auto-retries on the next flush, no explicit retry/cancel button this round (per @copilot-pull-request-reviewer)

### No Change Needed
- **duplicate-email identity (SucceededMemberOperations keyed by email)**: acknowledged the theoretical collision if two staged `new` rows shared an email, but there's no current UI path that lets that happen in a single wizard session — left identity email-based rather than introducing a tempId-based redesign for a case that can't currently trigger. Will revisit if a future feature (e.g. bulk CSV import) changes that (flagged by @copilot-pull-request-reviewer)

### Threads Resolved
2 of 3 unresolved threads addressed and resolved in this iteration.

### Still Open
- **duplicate-email identity thread**: left open — explained above, no live trigger path today, revisit if that changes.

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.html
```diff
@@ -161,6 +161,10 @@ <h3 class="text-gray-600 mt-2">No members found</h3>
               @if (member.state === 'modified') {
                 <span class="px-2 py-1 text-xs font-medium bg-amber-100 text-amber-800 rounded">Modified</span>
               }
+              <!-- Failed-delete indicator — still staged for retry on the next submit (GH-1608) -->
+              @if (member.state === 'deleted') {
+                <span class="px-2 py-1 text-xs font-medium bg-red-100 text-red-800 rounded">Failed to remove</span>
+              }
```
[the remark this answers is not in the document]
MRashad26: Good catch — that was a real one, and directly a consequence of the visibility fix in the last iteration (this row was unreachable via Edit before). Disabled Edit for failed-delete rows: `handleMemberUpdate()` only promotes `existing` → `modified`, so editing a `deleted`-state row silently dropped the change and the member stayed queued for deletion on the next flush. Editing isn't a supported operation for a row already staged for deletion, so it's disabled with a tooltip explaining the row retries automatically.

See commit 16dd98df9: committee-members-manager.component.html.

MRashad26:
## Review Feedback Addressed (follow-up iteration)

Commit: 16dd98df9

### Changes Made
- **committee-members-manager.component.html**: disabled Edit for failed-delete rows. `handleMemberUpdate()` only promotes `existing` → `modified`, so editing a `deleted`-state row silently dropped the change and the member stayed queued for deletion on the next flush — a bug directly introduced by the previous iteration's visibility fix, since these rows were unreachable via Edit before. Editing isn't a supported operation for a row already staged for deletion, so it's disabled with a tooltip explaining the row retries automatically (per @cursor)

### Threads Resolved
1 of 1 new unresolved thread addressed and resolved in this iteration.

### Still Open (from prior iteration, unchanged)
- **duplicate-email identity thread**: left open — no live UI trigger path today for staging two `new` rows with the same email; will revisit if that changes (e.g. bulk import).

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.ts
```diff
@@ -114,8 +115,13 @@ export class CommitteeMembersManagerComponent implements OnInit {
...
   private readonly loadingInvites = signal(false);
 
+  // uids of members whose delete failed on the last flush. Members in this state stay in state
+  // 'deleted' (so emitMemberUpdates() keeps re-queuing the delete on the next flush) but must
+  // remain visible — otherwise a failed delete silently vanishes from the roster (GH-1608).
+  private readonly failedDeleteUids = signal<Set<string>>(new Set());
+
   // Simple computed signals
-  public readonly visibleMembers = computed(() => this.membersWithState().filter((m) => m.state !== 'deleted'));
+  public readonly visibleMembers = computed(() => this.membersWithState().filter((m) => m.state !== 'deleted' || this.failedDeleteUids().has(m.uid)));
```
audigregorie: **Major — failed-delete visibility is lost when the step panel remounts**

`failedDeleteUids` is component-local state, and PrimeNG `p-step-panel` destroys inactive panel content (`@if (isVisible())` guards the panel's `ngTemplateOutlet` in `primeng-stepper.mjs`), so leaving step 4 and coming back recreates this component with a fresh empty set.

Sequence: stage a delete → Update → delete fails → `pruneSucceeded` keeps the row visible via `failedDeleteUids` → user clicks to Settings and back → `ngOnInit` re-runs, `reconcilePendingMemberState` correctly reapplies `state: 'deleted'` from the still-staged `memberUpdates().toDelete` input — but `failedDeleteUids` is now empty, so `visibleMembers` filters the row out. The "Failed to remove" member silently vanishes (and the Members count drops), which is the same silent-disappearance class GH-1608 set out to fix. The delete does stay staged and retries on the next submit, so there's no data loss — the user just loses the failure signal.

Since the parent (`committee-manage`) survives step navigation and already has the raw `results`, the clean fix is to lift the failed-delete uids up: compute them in the parent (`results.filter((r) => r.type === 'delete' && !r.success).map((r) => r.identifier)`), pass them down as an input, and hydrate/derive here the same way `pendingInvites` is rehydrated from `memberUpdates().toInvite` in `ngOnInit`.

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.html
```diff
@@ -185,9 +189,9 @@ <h3 class="text-gray-600 mt-2">No members found</h3>
                   icon="fa-light fa-pen"
                   [text]="true"
                   styleClass="text-gray-400 hover:text-blue-600"
-                  pTooltip="Edit"
+                  [pTooltip]="member.state === 'deleted' ? 'Retries automatically on next save' : 'Edit'"
                   tooltipPosition="top"
-                  [disabled]="mutationsDisabled()"
+                  [disabled]="mutationsDisabled() || member.state === 'deleted'"
                   data-testid="edit-member-button"
                   (click)="openEditMemberDialog(member)">
```
audigregorie, line 196: **Minor — gate the Remove button for failed-delete rows too**

The `|| member.state === 'deleted'` guard here disables Edit for failed-delete rows, but the Remove (trash) button a few lines below only checks `mutationsDisabled()`. Before this PR, deleted-state rows were invisible so this was unreachable; now that failed deletes stay visible, clicking Remove on a "Failed to remove" row re-opens the "Are you sure you want to remove this member?" confirmation — accepting it is a harmless no-op (re-marks `state: 'deleted'`), but it's confusing next to a badge that says the removal already failed.

Suggest applying the same guard to the Remove button's `[disabled]` (`mutationsDisabled() || member.state === 'deleted'`), optionally with the same 'Retries automatically on next save' tooltip.

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.html
```diff
@@ -161,6 +161,10 @@ <h3 class="text-gray-600 mt-2">No members found</h3>
               @if (member.state === 'modified') {
                 <span class="px-2 py-1 text-xs font-medium bg-amber-100 text-amber-800 rounded">Modified</span>
               }
+              <!-- Failed-delete indicator — still staged for retry on the next submit (GH-1608) -->
+              @if (member.state === 'deleted') {
+                <span class="px-2 py-1 text-xs font-medium bg-red-100 text-red-800 rounded">Failed to remove</span>
```
audigregorie, line 166: **Minor — only failed deletes get a failure marker; failed adds/updates/invites are indistinguishable**

This badge makes a failed delete identifiable, but the other three operation types get no equivalent signal: a failed add still shows "New", a failed update still shows "Modified", and a failed invite is an identical pending-invite row. Once the partial-failure toast fades, the user can't tell which staged items failed vs. simply haven't been submitted yet.

Not a blocker for the retry mechanics (those are correct), but worth considering an equivalent failed indicator — e.g. a red `Failed to add` / `Failed to update` badge variant, or an error outline on failed pending-invite rows — so the retry set is self-describing.

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
```diff
@@ -567,16 +601,37 @@ export class CommitteeManageComponent {
...
-    const totalFailed = results.reduce((sum, result) => sum + result.failed, 0);
+  /** Groups the identifiers of successfully-flushed operations so the members-manager child can prune them (GH-1608). */
+  private computeSucceededOperations(results: MemberOperationResult[]): SucceededMemberOperations {
+    const succeeded = results.filter((result) => result.success);
+
+    return {
+      addedMembers: new Map(
+        succeeded
+          .filter((result) => result.type === 'add' && result.createdMember)
+          .map((result) => [(result.identifier ?? '').trim().toLowerCase(), result.createdMember!])
```
audigregorie: **Minor — `createdMember!` non-null assertion (frontend checklist 14.5)**

The assertion is runtime-safe (the preceding `filter` guarantees `createdMember`), but TS can't narrow across the two callbacks. A single `flatMap` expresses filter + narrow without `!`:

```suggestion
          .flatMap((result) =>
            result.type === 'add' && result.createdMember ? [[(result.identifier ?? '').trim().toLowerCase(), result.createdMember] as const] : []
          )
```

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
```diff
@@ -57,6 +64,9 @@ export class CommitteeManageComponent {
   // Member management state
   public memberUpdates = signal<MemberPendingChanges>({ toAdd: [], toUpdate: [], toDelete: [], toInvite: [] });
   public memberUpdatesRefresh$ = new BehaviorSubject<void>(undefined);
```
audigregorie: **Minor — dead `memberUpdatesRefresh$` next to the new child reference**

Pre-existing (it's also unbound on `main`), but since this PR adds the `membersManager` viewChild right below: `memberUpdatesRefresh$` is never bound to the child's `refresh` input in the template and never emitted anywhere, and the child's `refresh` input plus its `ngOnInit` subscription are correspondingly dead. Worth removing both ends (or wiring the binding) while touching this area.

audigregorie (commented):
## Code Review Summary

Solid, well-documented fix for the silent loss of failed staged operations — identifier threading, email normalization, and the prune design are sound, and promoting a successful add to a real server `uid` is a nice touch (verified the BFF returns the created member via `res.status(201).json(newMember)`). **One Major finding**: the failed-delete visibility does not survive the wizard's step-panel remount, which partially re-introduces the silent disappearance for deletes via a common navigation action. Four Minor items posted as separate threads.

### Outside the diff

- **Accessibility (systemic, pre-existing)**: the icon-only member-row buttons (edit / send-message / remove) have no `aria-label`, while the newer `remove-pending-invite` button in the same template does set `[attr.aria-label]`. Not introduced by this PR — worth a follow-up sweep.
- **PR shape**: branch `fix/GH-1608` ↔ linked issue #1608 ✓; all 4 commits GPG-verified + DCO-signed ✓; conventional-commit subjects ✓ (the two `fix(committees)` commits carry no ticket suffix — GH-issue branches don't fit the `(LFXV2-#)`/`(PCC-#)` convention, noted for awareness only); diff size 154+/17− ✓.

### What's done well

- Successful adds are promoted to `existing` members carrying the real server `uid`, so a subsequent edit issues an update instead of a duplicate add.
- Email normalization (trim + lowercase) is applied consistently on both sides of the prune (`computeSucceededOperations` and `pruneSucceeded`/`stageInvites`).
- Navigation gating covers both flush entry points (create-mode `onDone` and edit-mode `onSubmitAll`), and `mutationsDisabled` correctly locks child mutations while a flush is in flight.
- Comments explain *why* (including the non-obvious input-rehydration timing on `pendingInvites`), matching the repo's comment policy.

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.ts
MRashad26: Done — lifted `failedDeleteUids` out of local component state into a parent-owned input (`committee-manage` now owns the signal and passes it down via `[failedDeleteUids]`). Since the PrimeNG step-panel destroys/recreates this component when navigating away from and back to step 4, the parent surviving that remount means the failed-delete flag now survives too.

See commit 2ccbc64f9.

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.html
MRashad26: Fixed — the Remove button now mirrors the Edit button's existing pattern: disabled and tooltip-swapped to "Retries automatically on next save" when `member.state === 'deleted'`.

See commit 2ccbc64f9.

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
MRashad26: Done — replaced the `createdMember!` non-null assertion with your suggested `flatMap`-based narrow, so the type is proven by the filter instead of asserted.

See commit 2ccbc64f9.

apps/lfx-one/src/app/modules/committees/committee-manage/committee-manage.component.ts
MRashad26: Good catch — removed the dead `refresh` input from the members-manager and the corresponding subscription in `committee-manage`'s `ngOnInit()`. Nothing else was subscribing to it.

See commit 2ccbc64f9.

apps/lfx-one/src/app/modules/committees/components/committee-members-manager/committee-members-manager.component.html
MRashad26: Agreed this is a real gap, but per your own note it's not a blocker — deferring badge differentiation for failed adds/updates/invites (vs. the current failed-delete-only badge) to a follow-up so this iteration stays scoped to the review feedback already in flight. Leaving this thread open to track it.

MRashad26:
## Review Feedback Addressed

Commit: 2ccbc64f9

### Changes Made
- **committee-members-manager.component.ts / committee-manage.component.ts**: lifted `failedDeleteUids` from component-local state into a parent-owned signal, passed to the child via a new `[failedDeleteUids]` input — the failure now survives the PrimeNG step-panel remounting the members-manager on navigation away from and back to step 4 (per @audigregorie, Major)
- **committee-members-manager.component.html**: gated the Remove button on `member.state === 'deleted'`, matching the existing Edit button pattern (per @audigregorie)
- **committee-manage.component.ts**: replaced the `createdMember!` non-null assertion with the suggested `flatMap`-based type narrow (per @audigregorie)
- **committee-manage.component.ts / committee-members-manager.component.ts**: removed the dead `refresh` input and its subscription (per @audigregorie)

### No Change Needed (deferred)
- **committee-members-manager.component.html:166**: badge differentiation for failed adds/updates/invites (currently only failed deletes get a badge) — reviewer flagged this as a real gap but not a blocker; deferring to a follow-up to keep this iteration scoped (flagged by @audigregorie)

### Threads Resolved
4 of 5 new unresolved threads addressed and resolved in this iteration.

### Still Open
- **committee-members-manager.component.html:166**: badge-differentiation follow-up (see above) — left open to track
- The duplicate-email identity thread from Copilot remains open per an earlier explicit decision in this PR's review history

audigregorie (approved):

PR #1649, opened 2026-08-19
feat(mktg-os-agents): form-first agent run pages and marketplace

joanreyero:
## What

Replaces the chat-based Marketing OS agent UX with the approved form-first design ([design of record](https://claude.ai/code/artifact/7067c463-be20-4e4e-8e0b-c13376c30823)):

- **Marketplace**: shipped card anatomy kept; "Coming soon" tags on disabled agents, "vN generated" badges for agents with stored output (user- and project-scoped, reloading on in-app project switch), search empty state.
- **Agent run-page shell**: intake form → staged running → document result (section checklist, View/Download, version chips) → "Request changes" feedback regenerating as v+1 via the existing `/api/mktg-agents` contract (full form resubmit + `feedback` + `prior_version`), with "Edit inputs" preserving answers. Built so the Message Foundation form (separate work item) slots into the same shell.
- **Brand Kit intake**: Paul's 7 questions verbatim as one batch form (all required, open-ended, no multiple-choice). LFX prefill: project name + repository URL from `ProjectContextService` + memoized `getProject` with "From LFX" chips; description prefilled only when present.
- Stored runs are keyed to the effective user (`lfx-mktg-agent-run:<sub>:<projectUid>:<agentId>`) with 401/403 follow-up recovery falling back to a fresh run.

## Size

~1.9k insertions — above the 1000-line target. Split rationale: the shell, the Brand Kit form, and the run persistence are one coherent behavioral unit; shipping the shell without a working agent would leave dead UI behind the flag. The MF form lands separately on top.

## Review notes — known minor findings (intentionally deferred)

- **P2**: if Admin-Mode impersonation starts/stops while the marketplace stays mounted and the project is unchanged, the badge *display* can be stale until a context change (all data reads are correctly user-keyed; display-only). Fix direction: key the badge reload stream on (uid, sub).
- **P3**: runs stored under the previous un-scoped key format are orphaned in localStorage; a one-time sweep would clean them.

Part of wi-mktgos-agents-ui


apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,234 @@
...
+   * for the document — anything else stays `pending`.
+   */
+  private pollForDocument(resultPath: string, session: MktgSessionInfo, priorVersion: number): Observable<MktgRunResultResponse> {
+    const body: MktgRunResultBody = { sessionId: session.sessionId, ownerToken: session.ownerToken };
+    return timer(MKTG_RUN_POLL.initialDelayMs, MKTG_RUN_POLL.intervalMs).pipe(
+      switchMap(() => this.http.post<MktgRunResultResponse>(resultPath, body)),
+      filter((result) => result.status === 'ready' && typeof result.documentMarkdown === 'string' && (result.version ?? 0) > priorVersion),
+      take(1),
+      timeout(MKTG_RUN_POLL.timeoutMs)
+    );
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7: `pollForDocument` no longer lets one transient failure abort the generation. Each result-poll attempt now catches errors matching the app-wide `isTransientHttpError` policy (network drop, 408/429, 5xx) and degrades them to a `pending` tick, so the next poll interval retries and the not-yet-persisted first-run session is never lost. Non-transient responses (e.g. 401/403 owner-token rejection) still fail fast, and the overall `timeout(MKTG_RUN_POLL.timeoutMs)` still bounds the wait. New specs cover both the transient-survival and the fast-fail paths.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,234 @@
...
+   * Polls the agent's validated result endpoint (owner token in the POST body,
+   * never the query string) until it reports a `ready` document with a version
+   * newer than the prior draft's. The endpoint surfaces only schema-validated,
+   * sha256-verified envelopes, so agent status chatter can never be mistaken
+   * for the document — anything else stays `pending`.
+   */
+  private pollForDocument(resultPath: string, session: MktgSessionInfo, priorVersion: number): Observable<MktgRunResultResponse> {
+    const body: MktgRunResultBody = { sessionId: session.sessionId, ownerToken: session.ownerToken };
+    return timer(MKTG_RUN_POLL.initialDelayMs, MKTG_RUN_POLL.intervalMs).pipe(
+      switchMap(() => this.http.post<MktgRunResultResponse>(resultPath, body)),
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7. Same root cause as the Bugbot thread: each result-poll POST now applies the shared transient-error policy (`isTransientHttpError`, the predicate behind `retryTransientHttpError`) — a transient failure degrades to a `pending` tick instead of erroring the stream. I used the catch-to-pending form rather than stacking `retryTransientHttpError` on each attempt because the poll timer IS the retry loop: the next interval re-POSTs anyway, so a nested retry would only double the request cadence. Non-transient 4xx still fails fast and the overall timeout is preserved, as requested. Specs added for both paths.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
```diff
@@ -0,0 +1,381 @@
...
+    // project detail), and the project selector reuses this component on a
+    // switch (it only rewrites ?project=) — so restore + prefill must re-run,
+    // and ALL run state must reset, on every active-context change, never just
+    // the first render (docs/reviews/frontend-checklist.md 14.14). Browser-only
+    // so SSR output stays stable; the first emission lands post-hydration.
+    if (isPlatformBrowser(this.platformId)) {
+      this.activeContext$
+        .pipe(
+          filter((context): context is ProjectContext => !!context),
+          distinctUntilChanged((previous, current) => previous.uid === current.uid),
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7: the reset stream is now `combineLatest([activeContext$, effectiveUser$])`, keyed by BOTH project uid and effective-user `sub` (`distinctUntilChanged` on the pair). An Admin Mode impersonation swap with the same project selected now runs `resetForContext`, which cancels the in-flight generation subscription (so a completion can never be persisted under the new sub) and re-restores the stored run + prefill for the new effective user.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,234 @@
...
+      return;
+    }
+    const key = this.storageKey(run.projectUid, run.agentId);
+    if (!key) {
+      return;
+    }
+    try {
+      window.localStorage.setItem(key, JSON.stringify(run));
+    } catch {
+      // Ignore — the in-memory result still renders; only cross-visit restore is lost.
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7: the service now keeps an in-memory `Map<string, MktgStoredAgentRun>` keyed exactly like localStorage (effective-user sub + project + agent). `saveRun` writes the cache first and localStorage best-effort; `loadRun` prefers the cache; `clearRun` drops both. A failed `setItem` now only costs cross-visit restore — the next follow-up still derives its version directive and poll gate from the run the user is looking at. A new spec locks this: with `setItem` throwing, the follow-up still rides the in-memory session with the correct `finalize as version 2` gate instead of starting a fresh v1 session.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.html
```diff
@@ -0,0 +1,236 @@
...
+                    <lfx-markdown-renderer [content]="currentVersion()?.document ?? ''" />
+                  </div>
+                }
+
+                <div class="grid grid-cols-1 sm:grid-cols-2 gap-x-4 gap-y-1.5 px-4 py-3.5" data-testid="mktg-agent-run-sections">
+                  @for (section of sectionChecklist(); track section.label) {
+                    <div class="flex items-center gap-2 min-w-0">
+                      <i class="fa-light fa-check text-xs shrink-0" [ngClass]="section.present ? 'text-emerald-500' : 'text-gray-300'" aria-hidden="true"></i>
+                      <span class="text-xs text-gray-600 truncate">{{ section.label }}</span>
+                    </div>
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7: the section checklist no longer conveys state by color alone. Present rows render `fa-circle-check` (emerald) and missing rows a distinct `fa-circle-xmark` (gray) — distinguishable without color — and each row now carries an `sr-only` "included"/"missing" text next to the label so screen-reader users get the state the icons convey. The icon class and SR text are precomputed in the checklist view model.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.html
```diff
@@ -0,0 +1,236 @@
...
+      <div class="flex items-center justify-between gap-3 px-3.5 py-2.5 border-b border-gray-200">
+        <div class="flex items-center gap-3 min-w-0">
+          <lfx-button
+            icon="fa-light fa-arrow-left"
+            [text]="true"
+            size="small"
+            ariaLabel="Back to agents"
+            data-testid="mktg-agent-run-back-button"
+            (onClick)="onBack()" />
+          <span class="flex items-center justify-center w-8 h-8 rounded-lg shrink-0" [ngClass]="iconClass(agent)">
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7: all three template functions are gone. `iconClass(agent)`/`tagline(agent)` became `agentIconClass`/`agentTagline` fields computed once (the agent is fixed for the component's lifetime — route param), `stageClass(state)` became a `labelClass` property on the `stages()` computed view model, and the section checklist rows likewise precompute their icon class. The template now only binds properties and reads signals, per docs/reviews/frontend-checklist.md §4.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
```diff
@@ -0,0 +1,381 @@
...
+    if (!current) {
+      return;
+    }
+    const blob = new Blob([current.document], { type: 'text/markdown' });
+    const url = URL.createObjectURL(blob);
+    const anchor = document.createElement('a');
+    anchor.href = url;
+    anchor.download = `${this.docTitle()} v${current.version}.md`;
+    anchor.click();
+    URL.revokeObjectURL(url);
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7: revocation is deferred to the next task — `setTimeout(() => URL.revokeObjectURL(url), 0)` after `anchor.click()`, matching what the removed Brand Kit implementation (`brand-kit-form.component.ts`) deliberately did, so browsers that start the download asynchronously can still consume the blob URL.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
```diff
@@ -0,0 +1,381 @@
...
+      return;
+    }
+    for (const field of this.intake.fields) {
+      if (field.prefill !== source) {
+        continue;
+      }
+      const control = this.intakeForm.controls[field.key];
+      if (control && !control.value.trim()) {
+        control.setValue(trimmedValue);
+        this.fromLfx.update((flags) => ({ ...flags, [field.key]: true }));
```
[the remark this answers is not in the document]
joanreyero: Fixed in f31282a7: source availability is now tracked separately from application. A new `lfxMissing` signal records, per field, whether the resolved LFX prefill source actually came back empty, and the "Not set on your LFX project" hint binds to it; `fromLfx` keeps driving only the "From LFX" chips (i.e. the value was actually applied). Restored answers or early typing over an existing LFX value no longer show the missing hint, and if the project lookup fails the hint stays hidden rather than claiming something we don't know.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.html
```diff
@@ -0,0 +1,236 @@
...
+                </div>
+              }
+
+              <div class="flex items-center gap-3 mt-2">
+                <lfx-button
+                  type="submit"
+                  [label]="submitLabel()"
+                  icon="fa-light fa-sparkles"
+                  [disabled]="submitDisabled()"
+                  data-testid="mktg-agent-run-submit-button" />
```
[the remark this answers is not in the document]
joanreyero: Won't fix: running an agent is not a project-scoped mutation, so `canWrite()` gating doesn't apply here. Nothing about the project is written server-side — a run creates a per-user Guild session whose ownership is HMAC-bound to the requesting user's sub, and the generated document lands only in that user's own browser storage. The BFF endpoints (`/api/mktg-agents/*`) are deliberately per-user, not project-role gated, so a UI-only `canWrite()` gate would be decorative: it would misrepresent the actual authorization contract while blocking read-only members from a capability the server intentionally allows them. If product wants role-based gating for agent runs, that has to land server-side first; happy to follow up then.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,271 @@
...
+   * multi-minute generation whose session may not even be persisted yet, so it
+   * degrades to a `pending` tick and the next interval retries; the overall
+   * `timeout` still bounds the wait. Non-transient responses (401/403 — the
+   * owner token stopped verifying) keep failing fast.
+   */
+  private pollForDocument(resultPath: string, session: MktgSessionInfo, priorVersion: number): Observable<MktgRunResultResponse> {
+    const body: MktgRunResultBody = { sessionId: session.sessionId, ownerToken: session.ownerToken };
+    return timer(MKTG_RUN_POLL.initialDelayMs, MKTG_RUN_POLL.intervalMs).pipe(
+      switchMap(() =>
+        this.http.post<MktgRunResultResponse>(resultPath, body).pipe(
```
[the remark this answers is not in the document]
joanreyero: Fixed in b3ee7963: `pollForDocument` now maps ticks with `exhaustMap` instead of `switchMap`, so a tick that fires while a result POST is in flight is dropped rather than cancelling it — sustained poll latency above the interval can no longer starve the poll into the 600s timeout. Added a fake-timer regression spec covering a poll slower than the interval (no cancellation, no duplicate POST, document still lands).

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (!isPlatformBrowser(this.platformId)) {
+      return;
+    }
+    const key = this.storageKey(run.projectUid, run.agentId);
+    if (!key) {
+      return;
+    }
+    this.memoryRuns.set(key, run);
+    try {
+      window.localStorage.setItem(key, JSON.stringify(run));
```
dealako, line 255: [minor] Owner/session capability token persisted in localStorage

Issue: `saveRun` writes the full run — including `sessionId` and the creator-binding `ownerToken` — to `localStorage` via `JSON.stringify(run)`.
Proof: line 241 `window.localStorage.setItem(key, JSON.stringify(run))`; `ownerToken` is carried on the persisted record (l.149/218) and the interface documents it as the token proving this browser's user created the session. localStorage is script-readable on the origin and the record has no TTL — it's only removed on a 401/403 follow-up or overwrite.
Why it matters: widens the blast radius of any XSS (token grants polling access to the user's generated document) and lingers on-device indefinitely. Bounded impact — token is HMAC-bound to `sub`, output is the user's own low-sensitivity marketing draft, and markdown is sanitized — so this is hardening, not a break.
Fix: keep `ownerToken` in `memoryRuns` only (forcing a fresh generate on reload), or move stored runs to `sessionStorage`, or add an explicit expiry to `MktgStoredAgentRun` so stale tokens prune on load.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
```diff
@@ -0,0 +1,416 @@
...
+    // Let the "Validating required sections" stage register before the result
+    // lands — but only if a project switch hasn't reset the page meanwhile.
+    setTimeout(() => {
+      if (this.run() === run) {
+        this.phase.set('result');
+      }
+    }, 600);
+  }
+
+  private sectionPresent(lowerDoc: string, label: string): boolean {
```
dealako, line 420: [minor] New run component has no unit test

Issue: this 416-line component — the largest new unit in the PR — has no spec file, while the service, marketplace component, and intake util all got dedicated specs.
Proof: the `mktg-agent-run/` dir contains only `.html` and `.ts`, no `.spec.ts`. Uncovered branching with real logic: `sectionPresent` (the numbered/appendix prefix strip + substring match here), `applyPrefill` (the `fromLfx`-applied vs `lfxMissing`-available invariant your own class comment flags as subtle), `resetForContext`, and `onDownload`.
Why it matters: prefill-availability tracking and the heading matcher are exactly the kind of logic that regresses silently; the service/marketplace equivalents got regression specs, these rely on manual QA.
Fix: add `mktg-agent-run.component.spec.ts` covering `sectionPresent` (numbered/appendix/plain headings + empty-doc), `applyPrefill` (`fromLfx` set only for empty controls while `lfxMissing` reflects source emptiness independently), `resetForContext` clearing state on context change, and the SSR early-return in `onDownload`.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
```diff
@@ -0,0 +1,416 @@
...
+      });
+  }
+
+  private completeRun(run: MktgStoredAgentRun): void {
+    this.run.set(run);
+    this.viewVersion.set(run.versions.at(-1)?.version ?? null);
+    this.feedbackForm.reset();
+    // Let the "Validating required sections" stage register before the result
+    // lands — but only if a project switch hasn't reset the page meanwhile.
+    setTimeout(() => {
```
dealako: [nit] Deferred phase-set timer not cancelled on teardown

Issue: this `setTimeout(..., 600)` is never tracked/cleared and the class implements no `OnDestroy`.
Proof: the `this.run() === run` guard only defends against a project switch that reset `run`, not against view destruction inside the 600ms window.
Why it matters: harmless today — in zoneless Angular a post-destroy signal write is a no-op — but it's an unmanaged timer that would matter if the deferred block ever did more than set a signal.
Fix: store the timeout id and `clearTimeout` in `ngOnDestroy`, or use a `takeUntilDestroyed`-bound `timer`.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  private sectionPresent(lowerDoc: string, label: string): boolean {
+    if (!lowerDoc) {
+      return false;
+    }
+    if (lowerDoc.includes(label.toLowerCase())) {
+      return true;
+    }
+    const core = label.replace(/^\d+\.\s*/, '').replace(/^appendix [a-z]:\s*/i, '');
```
dealako, line 431: [nit] Section-label normalization regex is undocumented

Issue: the fallback strips a leading numeric prefix and an `Appendix X:` prefix before the substring match, with no comment explaining which heading shapes it tolerates.
Proof: `const core = label.replace(/^\d+\.\s*/, '').replace(/^appendix [a-z]:\s*/i, '');` — the surrounding checklist comment only covers icon/SR semantics.
Why it matters: the two-stage match (exact label, then prefix-stripped core) is non-obvious and invites accidental breakage if labels change.
Fix: one-line comment, e.g. `// Fallback: match the heading without its config numbering/appendix prefix ("1. Project Definition" -> "Project Definition").`

packages/shared/src/constants/mktg-run.constants.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Batch intake registry, keyed by catalog agent id. The run-page shell renders
+ * whatever is registered here — a second agent's form (e.g. the Message
+ * Foundation intake, wi-mf-lfx-selfserve) slots in as a new entry.
+ */
+export const MKTG_AGENT_INTAKES: Record<string, MktgAgentIntake> = {
+  [BRAND_KIT_INTAKE.agentId]: BRAND_KIT_INTAKE,
+};
+
+/** Running-phase stage checklist labels, in order. */
+export const MKTG_RUN_STAGES: string[] = ['Submitting your intake', 'Agent drafting the document', 'Validating required sections'];
```
dealako: [nit] `MKTG_RUN_STAGES` is a mutable `string[]` while siblings use `as const`

Issue: typed as plain mutable `string[]`, inconsistent with `MKTG_RUN_POLL` and `BRAND_KIT_FORM_PREAMBLE_LINES` in the same file (both `as const`).
Proof: line 82 `export const MKTG_RUN_STAGES: string[] = [...]` vs the `} as const;` siblings.
Why it matters: minor consistency/safety — this array is handed to a component field (`stageLabels = MKTG_RUN_STAGES`) and is safer as a readonly tuple.
Fix: `export const MKTG_RUN_STAGES = [...] as const;`

dealako (dismissed):
Hi @joanreyero 👋 — thorough independent re-review of the form-first Marketing OS agents rework.

## Overall impression

This is a large but coherent change (~1.9k insertions) that cleanly replaces the chat-panel surface with a form-first per-agent run shell, and the engineering rigor shows. I ran three parallel reviewers across security/privacy, correctness/performance/tests, and style/API/docs. The bot round (Cursor Bugbot + Copilot) raised ~9 substantive issues and you've already addressed 8 of them in `f31282a7`/`b3ee7963` — `exhaustMap` polling, transient-vs-fatal poll error handling, effective-user re-keying, memory-cache fallback on quota failure, section a11y, template-function removal, deferred blob revoke, and the `lfxMissing`/`fromLfx` split all landed and verify against current HEAD. I confirmed the security posture is strong: auth-guarded route, fail-closed user-scoped storage keys, owner token in POST body (never query/log), sanitized markdown, and `401/403` treated as non-transient (fails fast, no auth spin). Nothing here blocks merge.

## Findings

🔴 **Blocking:** 0

🟡 **Minor: 2**
1. **Owner/session capability token persisted in `localStorage`** — `mktg-agent-run.service.ts:241` writes `sessionId` + `ownerToken` via `setItem`, XSS-readable with no TTL/expiry. Impact is bounded (token is HMAC-bound to `sub`, output is the user's own low-sensitivity marketing draft, markdown is sanitized), so this is hardening, not a break. Consider `sessionStorage`/in-memory-only or an explicit expiry.
2. **New `MktgAgentRunComponent` ships with no unit test** — the service, marketplace component, and intake util all got dedicated specs, but the 416-line run component's own branching (`sectionPresent` regex, `applyPrefill` availability tracking, `resetForContext`, `onDownload`) is uncovered.

⚪ **Nit: 3**
3. **`completeRun()` 600ms `setTimeout` not cancelled on teardown** (`mktg-agent-run.component.ts:399`) — harmless in zoneless today (post-destroy signal write is a no-op), but an unmanaged timer.
4. **`sectionPresent()` normalization regex undocumented** (`:413`) — a one-line comment on the numbering/appendix strip would help.
5. **`MKTG_RUN_STAGES` typed as mutable `string[]`** (`mktg-run.constants.ts:82`) — inconsistent with the sibling `as const` constants in the same file.

## Bot reconciliation

I agree with your resolutions of every Cursor/Copilot finding and with the **"Won't fix" on `canWrite()` gating** (`mktg-agent-run.component.html`) — running an agent creates a per-user Guild session HMAC-bound to the caller's `sub`, not a project-scoped mutation, so write-gating doesn't apply. The security reviewer independently reached the same conclusion.

## Decision

✅ **Approved with minor comments** — none of the 5 items block merge; the two minors are worth a follow-up (token retention + run-component test coverage), the three nits are take-it-or-leave-it.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
joanreyero: Fixed in c62903cf, going with your third option (explicit expiry) so same-/next-day restore keeps working while the token stops lingering at rest: `MktgStoredAgentRun` now carries a `savedAt` clock and a shared `MKTG_RUN_STORAGE_TTL_MS` (24h). `loadRun` prunes any record past the TTL — or with an unreadable/absent `savedAt`, which also retires records written before the TTL existed — removing it from localStorage and letting the user fall back to a fresh generate. The in-memory cache is deliberately exempt: it only lives for the page session and its token is actively in use. Specs cover the expired-record prune and the legacy-record → fresh-run path, and assert `savedAt` is persisted with every save.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
joanreyero: Fixed in c62903cf: added `mktg-agent-run.component.spec.ts` (7 tests) covering exactly the four areas you called out — `sectionPresent` via the rendered checklist (exact label, numbered "1. Project Definition" satisfied by an unnumbered heading, "Appendix A: Document Architecture" satisfied by its core, an absent section flagged, and the empty-document case), the `applyPrefill` invariant (`fromLfx` set only when an empty control is actually filled, `lfxMissing` reflecting source emptiness independently — including the restored-answers case where an available LFX value must neither chip nor read as missing), `resetForContext` (restore of the active project's stored run, then a full reset + fresh prefill on a context switch with no state bleed), and the SSR early-return in `onDownload` (server platform + a present current version, asserting no blob URL is ever created).

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
joanreyero: Fixed in c62903cf: the deferred phase-set timer is now tracked in a `phaseTimer` field — `completeRun` clears any previous one before scheduling, and a `destroyRef.onDestroy` hook clears it on teardown, so no timer outlives the view even inside the 600ms window.

apps/lfx-one/src/app/modules/mktg-os-agents/mktg-agent-run/mktg-agent-run.component.ts
joanreyero: Fixed in c62903cf: the fallback now carries an explanatory comment — it matches the heading without the config label's numbering/appendix prefix ("1. Project Definition" → "Project Definition", "Appendix A: Document Architecture" → "Document Architecture"), since the agent may render its own numbering but the section name itself must appear. The new component spec also locks both prefix shapes down.

packages/shared/src/constants/mktg-run.constants.ts
joanreyero: Fixed in c62903cf: `MKTG_RUN_STAGES` is now a readonly `as const` tuple, consistent with `MKTG_RUN_POLL` and `BRAND_KIT_FORM_PREAMBLE_LINES` in the same file.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,291 @@
...
+        return null;
+      }
+      // TTL prune: the persisted record carries the session's capability
+      // ownerToken, so it must not linger at rest indefinitely. Past
+      // MKTG_RUN_STORAGE_TTL_MS (or with an unreadable savedAt — e.g. a record
+      // written before the TTL existed) the record is removed and the user
+      // simply starts a fresh run. The in-memory cache above is exempt: it
+      // only lives for this page session and its token is actively in use.
+      const savedAtMs = Date.parse(parsed.savedAt ?? '');
+      if (Number.isNaN(savedAtMs) || Date.now() - savedAtMs > MKTG_RUN_STORAGE_TTL_MS) {
```
dealako, line 113: [minor] TTL is lazy and per-key — orphaned tokens/answers/documents are never swept

Issue: the new MKTG_RUN_STORAGE_TTL_MS prune only runs inside loadRun, and loadRun composes exactly ONE key from the current effective user's sub + the exact project+agent being viewed (storageKey, ~l.289).
Proof: line 114 `window.localStorage.removeItem(key)` fires only for that single key; nothing enumerates the `lfx-mktg-agent-run` prefix and user.service has no clear-on-logout/impersonation hook. Scenario: user A generates a run (key ...:A:proj-1:brand-kit), A logs out, B logs in -> storageKey now yields ...:B:..., so A's key is never loaded and its ownerToken + intake answers + generated document stay in localStorage well past the 24h TTL, contradicting the JSDoc claim that records are not 'left at rest indefinitely.'
Why it matters: the documented data-at-rest bound is only honored for keys that happen to be re-loaded; on shared machines the previous user's session token + content linger unbounded (devtools/XSS-readable). Bounded — token is HMAC-bound to A's sub (not replayable), answers are non-sensitive project metadata — so this is a hardening gap, not an exploit. Corroborated by Copilot's comment on this file.
Fix: on service init / loadRun, enumerate localStorage keys with MKTG_RUN_STORAGE_KEY_PREFIX and remove any whose savedAt is missing or older than the TTL regardless of sub/project/agent, and/or clear the prior-sub keys on logout/impersonation.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,291 @@
...
+        window.localStorage.removeItem(key);
+        return null;
+      }
+      return parsed;
```
dealako, line 117: [minor] TTL prune can drop the in-use restored run mid-session, silently degrading a follow-up into a fresh v1

Issue: a run restored from localStorage is returned WITHOUT being inserted into `memoryRuns`, so the "in-memory cache is exempt ... its token is actively in use" guarantee in the comment above does not actually cover restored runs.
Proof: line 117 `return parsed` has no preceding `this.memoryRuns.set(key, parsed)`; only saveRun/appendVersion populate the cache. resetForContext calls loadRun + run.set(stored) but never saves. On the next onRegenerate -> generate(), loadRun re-reads localStorage; past 24h the record is pruned and null returned -> startRun (fresh session, priorVersion 0), whose body is `{ answers }` only — request.feedback is silently discarded. appendVersion then re-loads null and rebuilds `versions: [v1]`, losing prior lineage.
Why it matters: near the 24h boundary a "Request changes" click is silently reinterpreted as a brand-new first run — the user's feedback text is dropped and version history resets to v1. Narrow-window, hard-to-reproduce data-loss/UX regression. Corroborated by Cursor's comment on this file (their 'appendVersion mis-accept' framing is slightly off — a fresh session yields v1 > 0, so no wrong-version accept; the real harm is lost feedback + reset lineage).
Fix: seed the cache on a successful restore — add `this.memoryRuns.set(key, parsed);` immediately before `return parsed;` so an in-use run survives the page session as the comment already claims.

apps/lfx-one/src/app/shared/guards/mktg-os-agents-enabled.guard.ts
```diff
@@ -1,41 +1,14 @@
...
-  }
-
-  return featureFlagService.getBooleanFlag(MKTG_OS_AGENTS_ENABLED_FLAG, false)() ? true : router.parseUrl('/');
-};
+import { CanMatchFn } from '@angular/router';
+
+/**
+ * CanMatch guard for /project|foundation/mktg-os-agents.
+ *
+ * TODO(mktg-os GA): flag gate bypassed for now (Joan, 2026-08-19) — the
```
dealako, line 9: [nit] Temporary flag-gate bypass has no tracking ticket

Issue: the LaunchDarkly gate is bypassed here and in sidebar-nav.service.ts, documented only with a `TODO(mktg-os GA)` milestone tag + owner/date — no JIRA/GH issue reference. AGENTS.md requires ticket tracking; a TODO string is not a tracked work item.
Proof: line 9 `TODO(mktg-os GA): flag gate bypassed for now (Joan, 2026-08-19) ... Restore the LaunchDarkly ... check ... before GA hardening.` and the mirrored TODO in sidebar-nav.service.ts — neither links an issue.
Why it matters: a deliberately temporary, fail-closed-gate removal (guard now unconditionally returns true) with no linked ticket relies on memory to be undone before GA.
Fix: append a ticket ref to both TODOs (e.g. `TODO(mktg-os GA, LFX-1234): ...`) and file the tracking issue so the restore is scheduled, not memory-bound.

packages/shared/src/constants/mktg-run.constants.ts
```diff
@@ -0,0 +1,103 @@
...
+/** localStorage key prefix for stored runs; full key is `<prefix>:<projectUid>:<agentId>`. */
+export const MKTG_RUN_STORAGE_KEY_PREFIX = 'lfx-mktg-agent-run';
+
+/**
+ * TTL for browser-persisted runs. A stored run carries the session's
+ * capability `ownerToken`, so it must not linger at rest indefinitely —
+ * records older than this are pruned on load and the user simply starts a
+ * fresh run. Generous enough to keep same-day/next-day restore working.
+ */
+export const MKTG_RUN_STORAGE_TTL_MS = 86400000;
```
dealako, line 93: [nit] TTL magic number not expressed as its human-readable duration

Issue: the TTL is a raw millisecond literal `86400000` and the doc comment never states the concrete duration (24h).
Proof: line 93 `export const MKTG_RUN_STORAGE_TTL_MS = 86400000;` — a reader must compute 86400000 ms = 24h.
Why it matters: a bare literal invites miscalculation on future edits.
Fix: `export const MKTG_RUN_STORAGE_TTL_MS = 24 * 60 * 60 * 1000; // 24h` (or add `(24 hours)` to the comment).

dealako (approved):
Hi @joanreyero 👋 — follow-up re-review on the three new commits since my last round (`b3ee7963` → `ef01c386`). Thanks for the fast, targeted turnaround.

## 👏 Nice work

- **Run-page test coverage landed in full.** The new 243-line `mktg-agent-run.component.spec.ts` covers every seam I called out — `sectionPresent`'s prefix-tolerant matcher (numbered / `Appendix A:` / plain / empty-doc), the `applyPrefill` `fromLfx`-vs-`lfxMissing` invariant including the restored-answers case, `resetForContext` P1→P2 isolation, and the `onDownload` SSR guard. Not shallow — real assertions. ✅
- **`completeRun` timer is now teardown-safe** via a `phaseTimer` field cleared through `destroyRef.onDestroy`. ✅
- **Both nits resolved** — `sectionPresent` now has an example-bearing comment, and `MKTG_RUN_STAGES` is `as const`. ✅
- **TTL bound added** — `savedAt` stamping + load-time prune with tests for expiry and legacy-record pruning. Good direction; two edge cases remain (below).

## Revision tracking

- ✅ **Resolved (4):** run-component tests; `completeRun` timer cancellation; `sectionPresent` doc comment; `MKTG_RUN_STAGES` `as const`.
- ⚠️ **Partially addressed (1):** owner-token retention — a TTL now exists, but it's *lazy and per-key*, so orphaned records under other users/unrevisited projects are never swept (see minor #1). The XSS-readable-at-rest aspect is unchanged.

## Findings (all open items — none blocking)

🔴 **Blocking:** 0

🟡 **Minor: 2** *(both in the new TTL logic; each corroborated by a bot)*
1. **TTL is lazy/per-key — orphaned tokens/answers/docs never swept** (`mktg-agent-run.service.ts:113`). The prune only fires for the *current* user+project+agent key inside `loadRun`; nothing enumerates the `lfx-mktg-agent-run` prefix and there's no logout/impersonation sweep. On a shared browser, user A's `ownerToken` + intake answers + document persist past the 24h TTL indefinitely. Bounded (token is HMAC-bound to A's `sub`; answers are non-sensitive project metadata) → hardening gap, not an exploit. **Agrees with @Copilot's line-115 comment.**
2. **TTL prune can drop the *in-use* restored run mid-session** (`mktg-agent-run.service.ts:117`). A run restored from `localStorage` is returned without being inserted into `memoryRuns`, so the "in-use cache is exempt" guarantee in the comment doesn't cover it. Past the 24h boundary, a "Request changes" click re-reads, prunes, and takes the fresh-run path — silently dropping the user's feedback text and resetting version lineage to v1. **Agrees with @cursor's line-116 comment** (with one correction: no wrong-version *accept* occurs — a genuinely fresh session yields v1 > priorVersion 0; the real harm is lost feedback + reset lineage). One-line fix: `this.memoryRuns.set(key, parsed)` before `return parsed`.

⚪ **Nit: 2**
3. **Flag-bypass has no tracking ticket** (`mktg-os-agents-enabled.guard.ts:9` + `sidebar-nav.service.ts`). The `TODO(mktg-os GA)` carries owner/date/restore instructions but no JIRA/GH issue — a deliberately temporary fail-closed-gate removal shouldn't be memory-bound before GA.
4. **`MKTG_RUN_STORAGE_TTL_MS = 86400000`** (`mktg-run.constants.ts:93`) — express as `24 * 60 * 60 * 1000 // 24h` so the duration is legible.

## Note on the flag bypass (`578fe1ae`)

Reviewed for access implications: the marketplace routes still enforce `canActivate: [authGuard]`, and the generated document is built from user-supplied intake — so bypassing the LaunchDarkly gate introduces **no privileged-data exposure**. The only consequence is losing the runtime kill-switch/targeting (disabling now needs a deploy), which the TODO acknowledges as intentional dark-launch→visible. Fine as a temporary state given nit #3.

## Decision

✅ **Approved with minor comments** — no merge conflicts (`MERGEABLE`), 0 blocking, 0 privacy, 2 minors both with one-line fixes. The in-use-prune case (#2) is the one I'd most want closed before this sees real traffic near the TTL boundary, but neither minor blocks merge.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,291 @@
...
+      // ownerToken, so it must not linger at rest indefinitely. Past
+      // MKTG_RUN_STORAGE_TTL_MS (or with an unreadable savedAt — e.g. a record
+      // written before the TTL existed) the record is removed and the user
+      // simply starts a fresh run. The in-memory cache above is exempt: it
+      // only lives for this page session and its token is actively in use.
+      const savedAtMs = Date.parse(parsed.savedAt ?? '');
+      if (Number.isNaN(savedAtMs) || Date.now() - savedAtMs > MKTG_RUN_STORAGE_TTL_MS) {
+        window.localStorage.removeItem(key);
+        return null;
+      }
```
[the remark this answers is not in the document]
joanreyero: Tracked as follow-up: real — `loadRun` returns a restored record without seeding `memoryRuns`, so a TTL elapse mid-session can silently degrade a "Request changes" into a fresh v1 (the PR merged before this was addressed; the fix is to seed the cache on a successful restore). One framing correction: `appendVersion` cannot mis-accept a wrong version — a fresh session yields v1 against a prior-version gate of 0 — the actual harm is dropped feedback plus reset version lineage.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,291 @@
...
+      // TTL prune: the persisted record carries the session's capability
+      // ownerToken, so it must not linger at rest indefinitely. Past
+      // MKTG_RUN_STORAGE_TTL_MS (or with an unreadable savedAt — e.g. a record
+      // written before the TTL existed) the record is removed and the user
+      // simply starts a fresh run. The in-memory cache above is exempt: it
+      // only lives for this page session and its token is actively in use.
+      const savedAtMs = Date.parse(parsed.savedAt ?? '');
+      if (Number.isNaN(savedAtMs) || Date.now() - savedAtMs > MKTG_RUN_STORAGE_TTL_MS) {
+        window.localStorage.removeItem(key);
+        return null;
```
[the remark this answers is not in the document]
joanreyero: Tracked as follow-up: enforce the TTL with a prefix-wide sweep of `lfx-mktg-agent-run` records independent of the current user/project/agent (on browser-side load), so logout/impersonation or a never-revisited project cannot strand an `ownerToken` past the 24h bound. The PR merged before this was addressed.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
```diff
@@ -0,0 +1,291 @@
...
+      message: renderMktgIntakeMessage(request.intake, request.answers, request.feedback, priorVersion || undefined),
+      sessionId: stored.sessionId,
+      ownerToken: stored.ownerToken,
+    };
+    return this.http.post<MktgChatResponse>('/api/mktg-agents/chat', body).pipe(
+      map(() => ({ session: { agentId: request.agentId, sessionId: stored.sessionId, ownerToken: stored.ownerToken }, priorVersion })),
+      catchError((error: unknown) => {
+        if (error instanceof HttpErrorResponse && (error.status === 401 || error.status === 403)) {
+          this.clearRun(request.projectUid, request.agentId);
+          return this.startRun(request);
```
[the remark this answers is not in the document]
joanreyero: Tracked as follow-up: real — the 401/403 fallback posts only `{ answers }` to the fresh session while `appendVersion` still records `request.feedback` on the stored version, so the UI can claim feedback the agent never saw. Follow-up will carry the pending feedback into the recovery run (or stop persisting unrealized feedback). The PR merged before this was addressed.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
joanreyero: The PR merged before this landed, so calling it straight: the finding is real and is NOT fixed on main. The prune only fires for the one key `storageKey` composes from the current effective sub + project + agent, so records under other subs (logout/impersonation) or never-revisited projects outlive the 24h bound, contrary to the JSDoc. Tracked as follow-up work: a `MKTG_RUN_STORAGE_KEY_PREFIX`-wide sweep on browser-side load that removes any record with a missing or expired `savedAt` regardless of sub/project/agent — your suggested fix.

apps/lfx-one/src/app/shared/services/mktg-agent-run.service.ts
joanreyero: Merged before this was addressed — the finding is real and not yet fixed on main: `loadRun` returns the restored record without `this.memoryRuns.set(key, parsed)`, so the "in-memory cache is exempt" comment overpromises for restored runs, and near the TTL boundary a "Request changes" degrades into a fresh v1 with the feedback dropped and lineage reset. Tracked as follow-up work: seed the cache on a successful restore (your one-line fix), together with the related Copilot finding that the 401/403 fresh-session fallback also drops pending feedback while `appendVersion` records it. Agreed on your correction of the Cursor framing — the fresh session yields v1 > 0, so no wrong-version accept.

apps/lfx-one/src/app/shared/guards/mktg-os-agents-enabled.guard.ts
joanreyero: The tracking issue now exists: #1650 covers restoring the LaunchDarkly `mktg-os-agents-enabled` gate (guard + `SidebarNavService`) before GA hardening, with both files named in its checklist — so the restore is scheduled, not memory-bound. The PR merged before the TODOs carried the ref; appending `#1650` to both TODO comments is queued with the post-merge follow-up sweep.

packages/shared/src/constants/mktg-run.constants.ts
joanreyero: Merged before this was addressed — agreed, and it will land in the post-merge follow-up sweep as `24 * 60 * 60 * 1000` (with the 24h duration stated in the doc comment).

PR #1786, opened 2026-08-23
fix(org-lens): show foundation name instead of project slug in groups

### LINKED ISSUE #1784: org Groups page shows raw project slug instead of foundation name
## Problem

`/org/groups` (`OrgLensGroupsService.getGroups()` → `OrgGroupsComponent`) rendered the raw
project slug (e.g. `cncf`) under each group's name instead of the human-readable foundation
name. The group detail page at `/groups/:uid` already shows the enriched foundation name via
the shared `enrichFoundationNames` mapper.

A second, related gap surfaced during investigation: the project-service query-service index
has holes — e.g. the Ultra Ethernet Consortium Fund (slug `uepf`) is entirely absent from it,
even though the project itself is real and accessible (`GET /api/projects/uepf` resolves fine).
That gap affected 10 of 46 groups for at least one org, all with high seat counts, so they
sorted to the top of the list still showing the raw slug.

## Fix

- Reuse the existing enrichment path (`enrichFoundationNames` + `ProjectService`, already used by
  `OrgPeopleCommitteeMembersService`) in `OrgLensGroupsService.getGroups()`, add an optional
  `project_name` field to `OrgLensGroupSummary`, and prefer it over `project_slug` in the
  `org-groups.component.html` template — falling back to the slug when enrichment is unavailable.
- Add a second enrichment source for the project-index gap: `CommitteeService.getCommitteesByIds`
  (made public; already existed privately for pending-invite enrichment) reads `project_name`
  from the committee-service index as a gap-filler. The live project index stays primary — the
  committee index is a write-time snapshot with no rename subscriber upstream, so it goes stale on
  a project rename and must not be the primary source.
- Make the foundation name a link to that foundation's org-lens detail page
  (`['/org/projects', project_slug]`, matching `org-projects.component.html`'s own convention),
  as a sibling of the row's own link to `/groups/:uid` (not nested — restructured as a
  "stretched link" row so both remain independently clickable/focusable). Only linked when
  `project_slug` is present; renders as plain text otherwise.
- UI treatment: group name and foundation name each turn blue only on their own hover (not
  together), signaling they're two distinct destinations.

## Expected result

- `/org/groups` shows "Cloud Native Computing Foundation" rather than "cncf", matching
  `/groups/:uid`, including previously-ungapped foundations like the Ultra Ethernet Consortium
  Fund, and still degrades to the slug when both enrichment sources miss or the seat is untagged.
- Clicking a group name still opens `/groups/:uid`; clicking the foundation name underneath opens
  `/org/projects/<slug>`.

Part of #1294.

manishdixitlfx:
## Summary

`/org/groups` showed the raw project slug (e.g. `cncf`) under each group's name instead of the human-readable foundation name. Fixes it by reusing the same project-index enrichment (`enrichFoundationNames`) already used elsewhere in the app, with the committee-service index as a fail-soft gap-filler for cases the project index doesn't resolve.

- **Enrichment**: `OrgLensGroupsService.getGroups()` resolves foundation names from the project index first (live, primary source); only committees left unresolved are looked up against the committee index (a write-time snapshot, used only as a gap-filler — never overrides a live project-index match).
- **Foundation link**: the foundation name under each group is now a link to `/org/memberships/:slug` (not `/org/projects/:slug`, which is Snowflake/CDP-scoped to sub-project activity and 404s for foundation-level slugs — verified live). Built as a stretched whole-row link (group → `/groups/:uid`) with an independently-clickable foundation link as a sibling, so the two targets get independent hover states and neither nests inside the other.
- Backend: `CommitteeService.getCommitteesByIds` made public (was already used internally for pending-invite enrichment) and reused as the gap-filler lookup.

## Test plan

- [x] `yarn lint && yarn format:check && yarn build` all pass
- [x] Server unit suite: 1589 tests passing (new coverage for `OrgLensGroupsService.getGroups` enrichment precedence, gap-filler targeting, and `CommitteeService.getCommitteesByIds`)
- [x] App unit suite: 683 tests passing (new coverage for project label rendering, foundation-link routing, aria-labels, singular/plural seat count)
- [x] New Playwright e2e coverage: `org-groups-row-link.spec.ts` (click-routing) and `org-groups-row-link-robust.spec.ts` (data-testid contract), sharing fixtures via `e2e/helpers/org-groups.helper.ts`
- [x] Manually verified live against prod data via Claude in Chrome: `/org/projects/uepf` and `/org/projects/cncf` 404, `/org/memberships/uepf` and `/org/memberships/cncf` resolve correctly
- [x] Post-commit reviewer trio (general + self-serve convention + self-serve learnings) run after every commit, full-branch sweep run clean, `/lfx-self-serve-pr-readiness` (READY WITH CHANGES — one non-blocking commit-header-length nit), `/preflight` all clean

## Known trade-offs (documented in-code, not blocking)

- Foundation-name label truncates without a tooltip on narrow viewports — flagged repeatedly in review; deliberately left as-is to avoid unbounded scope creep on this bug-fix ticket.
- The `/org/memberships/:slug` link still dead-ends (on the memberships page's own not-found state) for a foundation the org has a committee seat in but no *active* Snowflake membership row for, or where the project-service slug happens not to match Snowflake's foundation slug — narrower gap than before this fix, documented in-code, not fully closed.

Closes #1784


dealako:
👋 Thanks for the exceptionally thorough work here, @manishdixitlfx.

## Overall impression

This is a clean, well-scoped bug fix that grew the right amount of rigor around a deceptively subtle problem. The enrichment precedence (live project index primary, committee-service snapshot as a fail-soft gap-filler that never overrides a live match) is the correct call, and making it *sequential* so the committee-index call is skipped entirely when the project index resolves everything avoids doubling upstream load on the common path. The stretched whole-row link + independently-clickable sibling foundation link is the right accessibility pattern, and the routing switch from `/org/projects/:slug` (Snowflake/CDP-scoped, 404s for foundation slugs) to `/org/memberships/:slug` is verified against live data. The commit history shows a disciplined loop that caught real drift (stale comments, dead code, mis-scoped testids, a double-counted log metric) rather than churn for its own sake.

I ran three independent review passes (security + data privacy; correctness + performance + test coverage; style + API contract + docs). All three came back clean against the proof-and-trace bar.

## Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 1 — the `/org/memberships/:slug` residual dead-end

## Reconciliation with bot reviewers

- **Copilot** flagged that the foundation link can dead-end when the org holds a committee seat but has no *active* Snowflake membership, or when `project_slug` (project-service slug) doesn't match Snowflake's `foundationSlug`. This is a real, narrow gap — but it's the exact trade-off you've documented both in-code (above the `@if (group.project_slug)` block) and in the PR description. It is strictly narrower than the pre-fix state (the old `/org/projects` link 404'd for *every* foundation slug; the common active-membership case now resolves). Gating the link on a second enrichment source is legitimately out of scope for this bug-fix ticket. Recording it as a `[question]`/known trade-off, not a blocker. If you want to fully close it later, the plain-text-unless-verified-slug approach Copilot suggests is the clean path.
- **Cursor** rated the PR Medium Risk with no actionable findings and confirmed the click-target overlay is well covered by unit + e2e tests. Agreed.

## Verification notes

- Newly-public `CommitteeService.getCommitteesByIds` still applies `stripChatWebhookUrl` on every entry (with a new redaction test), reads only the public `project_name`, and never returns raw `Committee` objects to the response. No secret/PII leak through the new public path.
- The new enrichment INFO log emits integer counts only (no PII), and is correctly gated on `unresolvedCommitteeUids.length > 0`.
- The `[class]` + static-class merge KB correction is backed by a committed regression spec and matches two existing in-repo sites relying on the same Ivy merge behavior.

## Final decision

✅ **Approved with minor comments** (one documented trade-off tracked as a question).

dealako (approved):
Approved with minor comments. Three independent review passes (security/privacy, correctness/perf/tests, style/API/docs) came back clean against the proof-and-trace bar. The one open item is the documented /org/memberships/:slug residual dead-end, tracked as a question and accepted as a known trade-off for this bug-fix scope. See the summary comment for full reconciliation with Copilot and Cursor.

Issue #1779, opened 2026-08-22
fix(org-lens): Groups stat strip shows a partial class decomposition


manishdixitlfx:
Parent: #1776

**Problem.** The strip renders four tiles: Total Groups, Total Seats, Working Groups, SIGs. Tiles 3 and 4 are a *partial* decomposition of tile 1. Boards are excluded upstream by design, but Oversight, Ambassadors and Other are silently dropped. An org with 40 groups sees 12 + 8 and 20 unexplained. Users do the arithmetic and lose trust in the page.

`behavioralClassCounts()` (`org-groups.component.ts:89`) already computes all six classes. The template renders two.

## Two ways to fix — pick one

1. **Render every class present in the data.** Honest, but the tile count becomes variable (up to five once Boards are excluded) and the `sm:grid-cols-4` layout has to flex.
2. **Cut to two tiles (Groups, Seats) and move the class breakdown into the filter select as counted options.** Preferred — it turns dead numbers into a control, and pairs naturally with #1778.

## Acceptance

- No tile implies a decomposition that does not sum.
- Every behavioral class present in the roster is discoverable somewhere on the page.
- Skeleton tile count matches the rendered tile count.

manishdixitlfx:
## Decision: Option A — a tile per behavioral class present

Render every class that has at least one group, rather than moving the breakdown into the Type filter. Recorded on the epic (#1776) against child 3.

*Correction: an earlier version of this comment said this closed "open decision 3" on #1776. That was wrong — open decision 3 there is the seat-count denominator, which is still open. The Option A/B choice only ever lived in this issue.*

Consequence for #1778: the two stay independent. Option B would have folded this work into the Type filter and made them one piece; Option A does not, so the filter bar can proceed on its own.

## Real distribution, for sizing

Measured against a live org (Oracle America Inc., 46 groups, 171 seats):

| Behavioral class | Groups |
| --- | --- |
| Working Groups | 19 |
| Other | 13 |
| SIGs | 8 |
| Oversight | 5 |
| Ambassadors | 1 |
| **Total** | **46** |

19 + 13 + 8 + 5 + 1 = 46 — the arithmetic closes, which is the whole point of this issue.

Two things worth knowing before implementing:

**It is seven tiles, not six.** Two totals plus five classes. At the current `sm:grid-cols-4` that wraps to 4 + 3. The Option A artboard on #1776 shows six tiles — it was drawn before this distribution was measured and understates the wrap by one tile.

[...]

Issue #1776, opened 2026-08-22
feat(org-lens): Groups page — deliver the descoped LFXV2-1880 scope [EPIC]


manishdixitlfx:
**Page:** `/org/groups` — `OrgGroupsComponent` (`apps/lfx-one/src/app/modules/dashboards/org/org-groups/`), BFF `OrgLensGroupsService` (`apps/lfx-one/src/server/services/org-lens-groups.service.ts`)

## What this is

The Org Lens Groups page is a reduced subset of an approved spec that was never built. This epic delivers the missing scope.

**[LFXV2-1880](https://linuxfoundation.atlassian.net/browse/LFXV2-1880) — "Groups · Table (All / Board / Other tabs)" — is the actual Org Lens Groups spec.** Parent epic [LFXV2-2252](https://linuxfoundation.atlassian.net/browse/LFXV2-2252) "Org Lens Full Parity with Org Dash" (Released). It specifies search + foundation + project + voting filters, sortable columns (Name / Type / Members / Last Updated), pagination 10/25/50 with URL state, columns for Description / Channels / Voting Status / Last Updated, avatar stacks, and Board as a first-class tab with a live count. CSV export sat in its own open questions.

**Status: Discarded. Never built.** Its siblings [LFXV2-1879](https://linuxfoundation.atlassian.net/browse/LFXV2-1879) (page shell + filter bar), [LFXV2-1881](https://linuxfoundation.atlassian.net/browse/LFXV2-1881) (Board tab), [LFXV2-1882](https://linuxfoundation.atlassian.net/browse/LFXV2-1882) (Other tab) and [LFXV2-1909](https://linuxfoundation.atlassian.net/browse/LFXV2-1909) (Discover) were all Discarded too.

What actually shipped came from a different lineage. Commit `09a99ada`, PR #1319, [LFXV2-2014](https://linuxfoundation.atlassian.net/browse/LFXV2-2014) — *"Public WG: lens visibility and Profile join CTAs"* — under epic [LFXV2-2009](https://linuxfoundation.atlassian.net/browse/LFXV2-2009) *"Public WG: canonical public groups experience"*. The commit body states it directly:

> Wire /org/groups route (previously redirected to overview)

So the page is a Public-WG visibility feature that filled an empty Org Lens slot. This is not design debt that accumulated — it is an approved spec that was silently descoped.

## Current state

[...]

Issue #1783, opened 2026-08-22
fix(org-lens): Groups page two-step loading shift and header/empty-state copy


manishdixitlfx:
Parent: #1776

Small polish items on `/org/groups`, independent of the rest of the epic.

**1. Two-step layout shift.** The `!loaded()` branch (`org-groups.component.html:25`) paints four stat-card skeletons and no list skeleton. The `groupsLoading()` branch (`:91`) then paints stat cards **plus** five row skeletons. Then content. The page grows twice before settling. Add the list skeleton to the first branch so the two loading states have the same shape.

**2. Header copy repeats itself and does not name the org.** The H1 is `{{ committeeLabel.plural }}` and the subtitle opens with the same word: "Committees / Committees your organization's employees participate in across Linux Foundation projects." Org Memberships personalizes instead — `Membership — {{ companyName() }}` (`org-memberships.component.html:5`). Do the same here and rewrite the subtitle so it does not restate the title.

**3. Empty-state copy bypasses the label tokenization.** The empty state hardcodes "working groups or committees" (`:108`) while the rest of the page respects `COMMITTEE_LABEL`. Use the token.

## Acceptance

- Loading renders one stable layout; no growth between loading phases.
- H1 names the selected organization; subtitle does not repeat the H1's noun.
- Empty-state copy uses `COMMITTEE_LABEL` rather than hardcoded wording.

manishdixitlfx:
Shipped in #1788 (3e792c7e4).

Issue #1782, opened 2026-08-22
fix(org-lens): Groups list a11y — foundation-line contrast and mobile class chip


manishdixitlfx:
Parent: #1776

Two WCAG 2.1 AA failures in the group row. Both are self-contained and can ship ahead of the rest of the epic.

**1. Foundation line fails contrast.** `org-groups.component.html:129` uses `text-gray-400` on white. LFX overrides Tailwind’s gray scale in `lfxColors` (`packages/shared/src/constants/colors.constants.ts`), so that token resolves to `#90A1B9` — **2.63:1**, against the 4.5:1 required for normal text. It is the lightest text on the page, at `text-xs`, and it carries the metadata that identifies the foundation. `text-gray-500` (`#62748E`) measures **4.76:1** and passes.

**2. Behavioral class disappears on mobile.** The chip is `hidden sm:flex` (`:134`). Below the `sm` breakpoint the only remaining class signal is the icon's colour — a colour-only signal, which fails on its own — and the icon is `aria-hidden="true"`, so screen-reader users get the class only from the row `aria-label` at any viewport.

## Fix

- `text-gray-400` → `text-gray-500` on the foundation line.
- Keep the class visible at all breakpoints — stack the chip under the group name on narrow viewports rather than hiding it.

## Acceptance

- Foundation line measures ≥ 4.5:1 against its background.
- Behavioral class is conveyed by something other than colour at every breakpoint.
- Row `aria-label` still names the class (and, once #1777 lands, the foundation).

manishdixitlfx:
Shipped in #1789 (c85563280).

PR #1789, opened 2026-08-23
fix(org-groups): fix contrast and mobile visibility on group row

manishdixitlfx:
## Summary

- Foundation-line text and its hover state failed WCAG AA contrast (`gray-400` 2.63:1, `blue-600` 4.04:1 against white) — swapped to `gray-500` (4.76:1) and `blue-700` (6.49:1), including the group name's own hover color which had the identical `blue-600` failure.
- The behavioral-class chip (Working Group / SIG / etc.) was hidden below `sm` via `hidden sm:flex`, leaving only the row icon's color as a signal on mobile — a color-only signal, and the icon is `aria-hidden`. The chip is now visible at every breakpoint, stacked under the group name on narrow viewports via a `sm:contents` wrapper trick that folds back into the original desktop row layout at `sm+`.

### Stacked on #1790

Top of a four-deep stack for epic #1776:

```
main
 └─ #1787  feat/GH-1778         filter bar
     └─ #1788  fix/GH-1783-…    copy tokenization
         └─ #1790  fix/GH-1779  stat strip
             └─ #1789  ← this PR  contrast + mobile chip
```

Base is `fix/GH-1779`, not `main`. Merge bottom-up: #1787, #1788, #1790, then this.

## Deliberate exclusions / deferred items

- **Seat-count icon** (`fa-user-check text-gray-400`) left as-is — it sits beside a visible number and is `aria-hidden`, so it's decorative. Not fixed on purpose.
- **Two sibling pages** (`public-foundation-groups.component.html`, `public-project-groups.component.html`) carry the identical `gray-400` contrast failure and hidden-chip-below-`sm` pattern on the same row shape, on anonymous-reachable pages. Out of scope here (different components) — now tracked as #1791 with exact files and line numbers.
- **Truncated group/foundation names have no tooltip** exposing the full string when clipped — pre-existing from the foundation-link work this branch is layered on (#1786), not introduced here. Assistive tech gets the full name via the row's `aria-label`; sighted users on narrow viewports don't. Left as-is.
- **Some regression tests assert Tailwind class strings directly** (contrast colors, `flex-1`/`sm:contents`, `peer` presence) rather than rendered behavior — jsdom computes no layout or contrast, so a class-token assertion is the only available unit-level lock for a display/color fix. Moving these to the e2e/Playwright suite where real CSS applies is a larger scope change not attempted here.

## Test plan

- [x] `yarn ng test --include='**/org-groups.component.spec.ts'` — passes
- [x] Full app suite (`yarn test`) — 724/724 pass
- [x] `yarn lint` — clean (one pre-existing unrelated warning elsewhere)
- [x] `yarn build` — succeeds
- [x] Manual mutation check: removing `class="peer"` from the stretched row link fails exactly the intended hover-regression test, nothing else
- [x] Manual verification in a running dev server, in the integration worktree (all four stacked branches merged) at `localhost:4200/org/groups`, on a real org (Oracle America Inc., 46 groups):
  - **Desktop (1413px)** — foundation link resolves to `text-gray-500`, computed `rgb(98, 116, 142)` = `#62748E` (4.76:1), hover `blue-700`. No `blue-600` remains anywhere on the row. The 46 `sm:contents` wrappers compute to `display: contents`, so the desktop row layout is byte-for-byte the pre-existing one.
  - **Mobile (386px, below the `sm` breakpoint)** — the same wrapper computes to `display: flex`, and the class chip renders visibly (94×23) stacked below the group-name block instead of disappearing. No horizontal overflow: `scrollWidth` equals `clientWidth` at 386px.
  - The `text-gray-400` occurrences that remain on the page are the decorative `fa-user-check` seat icons only, per the exclusion above.
- [ ] Tablet-width (~768px) spot check not performed — the `sm:contents` wrapper's two states were verified at either side of the single `sm` (640px) breakpoint it keys off, and there is no second breakpoint between them.

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.spec.ts
@@ -571,6 +571,135 @@ describe('OrgGroupsComponent — project label and row/foundation routing', () =
     const rowLink = fixture.nativeElement.querySelector('[data-testid="org-groups-row-link-g1"]');
     expect(rowLink?.getAttribute('href')).toBe('/groups/g1');
   });
+
+  describe('contrast (GH-1782)', () => {
+    it('foundation link uses passing gray-500/blue-700 contrast, not gray-400/blue-600 (GH-1782)', async () => {
+      await renderRow({
+        groups: [group({ project_name: 'Ultra Ethernet Consortium Fund', project_slug: 'uepf' })],
+        total_groups: 1,
+        total_seats: 3,
+      });
+
+      const link = projectLineElement();
+      expect(link).not.toBeNull();
+      expect(link?.className).toContain('text-gray-500');
+      expect(link?.className).not.toContain('text-gray-400');
+      expect(link?.className).toContain('hover:text-blue-700');
+      expect(link?.className).not.toContain('hover:text-blue-600');
+    });
+
+    it('foundation plain-text fallback uses passing gray-500 contrast, not gray-400 (GH-1782)', async () => {
+      await renderRow({
+        groups: [group({ project_name: 'Ultra Ethernet Consortium Fund', project_slug: undefined })],
+        total_groups: 1,
+        total_seats: 3,
+      });
+
+      const fallback = projectLineElement();
+      expect(fallback).not.toBeNull();
+      expect(fallback?.className).toContain('text-gray-500');
+      expect(fallback?.className).not.toContain('text-gray-400');
+    });
+
+    it('group name hovers to passing blue-700, not blue-600 (GH-1782)', async () => {
+      await renderRow({ groups: [group()], total_groups: 1, total_seats: 3 });
+
+      // The name `<p>` sets no color of its own (see the template comment above this row) — it
+      // inherits from this wrapper, so the hover color only takes effect if the name actually lives
+      // inside it. Assert containment, not just that each element independently exists.
+      const rowContent = fixture.nativeElement.querySelector('[data-testid^="org-groups-item-row-content-"]');
+      const name = fixture.nativeElement.querySelector('[data-testid^="org-groups-item-name-"]');
+      expect(rowContent).not.toBeNull();
+      expect(rowContent?.contains(name)).toBe(true);
+      expect(rowContent?.className).toContain('peer-hover:text-blue-700');
+      expect(rowContent?.className).not.toContain('peer-hover:text-blue-600');
+
+      // `peer-hover:` compiles to `.peer:hover ~ &` — it only fires if the stretched row link is a
+      // *preceding* sibling of this wrapper (same parent, earlier in document order), not merely
+      // present somewhere in the row.
+      const rowLink = fixture.nativeElement.querySelector('[data-testid^="org-groups-row-link-"]');
+      expect(rowLink).not.toBeNull();
+      expect((rowLink?.className ?? '').split(/\s+/)).toContain('peer');
+      expect(rowLink?.parentElement).toBe(rowContent?.parentElement);
+      expect((rowLink?.compareDocumentPosition(rowContent) ?? 0) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    });
+  });
+
+  describe('responsive row layout (GH-1782)', () => {
+    it('behavioral class chip has no hidden class at any breakpoint (GH-1782)', async () => {
+      await renderRow({
+        groups: [group()],
+        total_groups: 1,
+        total_seats: 3,
+      });
+
+      const chip = fixture.nativeElement.querySelector('[data-testid^="org-groups-item-class-chip-"]');
+      expect(chip).not.toBeNull();
+      const chipClasses = (chip?.getAttribute('class') ?? '').split(/\s+/);
+      expect(chipClasses.some((c: string) => c === 'hidden' || c.endsWith(':hidden'))).toBe(false);
+    });
+
+    it('behavioral class chip keeps its explicit flex display and shrink-0, so a "drop redundant class" pass cannot silently regress it (GH-1782)', async () => {
+      // Without `flex`, the host still renders (it blockifies to `block` as a flex item either
+      // way) — no other assertion here would fail, only the strut-height regression `flex` fixes
+      // would silently come back. Without `shrink-0`, the chip would default to flex-shrink: 1 as
+      // a direct sibling of the flex-1 name column once `sm:contents` folds the wrapper away at
+      // `sm+`, letting it get squeezed instead of keeping its natural width.
+      await renderRow({ groups: [group()], total_groups: 1, total_seats: 3 });
+
+      const chip = fixture.nativeElement.querySelector('[data-testid^="org-groups-item-class-chip-"]');
+      expect(chip).not.toBeNull();
+      const chipClasses = (chip?.getAttribute('class') ?? '').split(/\s+/);
+      expect(chipClasses).toContain('flex');
+      expect(chipClasses).toContain('shrink-0');
+    });
+
+    it('name/meta block keeps a growing flex class, so it fills the row and pins the chip/seat/chevron right at every breakpoint (GH-1782)', async () => {
+      // Regression coverage: an earlier revision of this fix moved `flex-1` onto the `sm:contents`
+      // wrapper instead of this block. Since `display: contents` gives that wrapper no box of its
+      // own at `sm+`, its flex classes stopped applying there, and the name column collapsed to
+      // content width — bunching the chip/seat-count/chevron to the left instead of the row's right
+      // edge. Every a11y assertion in this file still passed throughout, so this test locks the
+      // layout-critical class directly onto the element that must carry it, and checks it's actually
+      // the wrapper's child — not just that both elements independently exist somewhere in the DOM.
+      await renderRow({ groups: [group()], total_groups: 1, total_seats: 3 });
+
+      const nameBlock = fixture.nativeElement.querySelector('[data-testid^="org-groups-item-meta-block-"]');
+      const wrapper = fixture.nativeElement.querySelector('[data-testid^="org-groups-item-meta-wrapper-"]');
+      const nameBlockClasses = (nameBlock?.className ?? '').split(/\s+/);
+
+      expect(nameBlock).not.toBeNull();
+      expect(wrapper).not.toBeNull();
+      expect(wrapper?.contains(nameBlock)).toBe(true);
+      expect(wrapper?.className).toContain('sm:contents');
+      expect(nameBlockClasses).toContain('flex-1');
+      expect(nameBlockClasses).toContain('min-w-0');
+    });
+  });
+
+  describe('row testid contract', () => {
+    it('gives each row its own uid-scoped testids, not a duplicate shared across rows (GH-1782)', async () => {
+      await renderRow({
+        groups: [group({ uid: 'g1' }), group({ uid: 'g2', name: 'SIG Storage' })],
+        total_groups: 2,
+        total_seats: 6,
+      });
+
+      for (const prefix of ['org-groups-item-row-content-', 'org-groups-item-meta-wrapper-', 'org-groups-item-meta-block-', 'org-groups-item-class-chip-']) {
+        const matches = fixture.nativeElement.querySelectorAll(`[data-testid^="${prefix}"]`);
+        expect(matches.length).toBe(2);
+        expect(matches[0].getAttribute('data-testid')).toBe(`${prefix}g1`);
+        expect(matches[1].getAttribute('data-testid')).toBe(`${prefix}g2`);
+      }
+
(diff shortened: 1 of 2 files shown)


manishdixitlfx:
⚠️ Do not merge directly. This PR targets a feature branch, not `main`, so branch protection does not apply and GitHub will show it as mergeable regardless of review or CI status. It is part of the #1776 stack and must merge bottom-up after #1787. It is also currently based on a stale tip of `feat/GH-1778` (`0e69c70b5`; that branch is now at `f75cdb3a5`) and will be restacked once #1787's review settles.

niravpatel27 (dismissed):

manishdixitlfx:
## Stack status — please hold merges until restacked

Thanks @niravpatel27 for the reviews on #1787, #1788 and #1789.

Two things happened in parallel today: review feedback landed on several branches, and each branch moved independently. The result is that **every PR above #1787 is now based on a stale tip of its parent**:

| PR | branch | head | based on current parent? |
|---|---|---|---|
| #1787 | `feat/GH-1778` | `f75cdb3a5` | — (base is `main`) |
| #1788 | `fix/GH-1783-groups-loading-and-copy` | `873dd34e2` | ❌ no |
| #1790 | `fix/GH-1779` | `cc8b3329d` | ❌ no |
| #1789 | `fix/GH-1782` | `7796fad73` | ❌ no |

So the diffs currently shown on #1788, #1790 and #1789 are not what would actually merge, and the approvals on #1788 and #1789 are against content that has since been superseded further down the stack. **Please do not merge any of these yet** — including #1787, since merging it first would leave the rest rebasing onto a moved target twice.

Also note #1788, #1790 and #1789 target feature branches rather than `main`, so branch protection does not apply to them and GitHub will show a green "ready to merge" button regardless of review or CI state. That button is not a signal on these three.

## Plan

1. **#1790 gets its human review** — it is the only one of the four still without one.
2. **Restack all three at once** onto their corrected parents: `fix/GH-1783-…` → `feat/GH-1778`, `fix/GH-1779` → `fix/GH-1783-…`, `fix/GH-1782` → `fix/GH-1779`.
3. **Re-request approval on the final shape.** The restack is a force-push, so GitHub will dismiss the existing approvals — doing steps 1 and 2 first means only one more review round rather than three.
4. **Merge bottom-up:** #1787 → #1788 → #1790 → #1789.

Doing the restack before #1790 is reviewed would just invalidate that review too, which is why it is sequenced this way.

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.spec.ts
```diff
@@ -666,3 +797,184 @@ describe('OrgGroupsComponent — loading state and copy', () => {
...
+  it('sizes the grid to the actual rendered tile count via --cols', async () => {
+    await render();
+
+    const section = rootElement().querySelector('[data-testid="org-groups-stat-strip"]') as HTMLElement;
+    // --cols must equal actual grid children (any tag, via :scope > *), not just testid-tagged
+    // ones — paired with the completeness check below so an untagged child fails loudly.
+    const gridChildren = section.querySelectorAll(':scope > *').length;
+
+    expect(section.querySelectorAll('[data-testid^="org-groups-stat-"]')).toHaveLength(gridChildren);
+    expect(section.style.getPropertyValue('--cols').trim()).toBe(String(gridChildren));
```
dealako, line 871: [minor] --cols test doesn't lock the grid that consumes it

Issue: `[class]="groupsLoading() ? statGridSkeletonClass : statGridLoadedClass"` and `[style.--cols]` are bound independently. This test asserts only `section.style.getPropertyValue('--cols')`, which is set no matter which class string wins.
Proof: line 967 reads the `--cols` custom property but never inspects `section.className`. If the ternary regressed to `statGridSkeletonClass` in the loaded branch, `--cols` would still be set and this test still passes, yet the grid would render a fixed `lg:grid-cols-6` and ignore the tile count.
Why it matters: the behavior `--cols` exists to produce (one column per rendered tile at `lg+`) has no test seam; a regression that decouples the grid from `--cols` ships green.
Fix: add one assertion that the loaded strip's class contains the `repeat(var(--cols),minmax(0,1fr))` token (or that it differs from the skeleton class), pairing the value check with the class that consumes it.

dealako:
Hi @manishdixitlfx 👋

Clean, well-scoped accessibility fix. The WCAG swap (`gray-400`→`gray-500` at 4.76:1, `blue-600`→`blue-700` at 6.49:1, including the group-name hover that shared the identical `blue-600` failure) is correct, and the `sm:contents` folding trick to surface the behavioral-class chip on mobile without disturbing the desktop row is a tidy solution. The PR body's exclusions (decorative seat icon, sibling public pages tracked as #1791, class-string test tradeoff) are well reasoned and honestly documented. The stat-strip rewrite that rides along from the parent stack (dynamic per-class tiles, `--cols` grid, safelist `.split(' )` normalization) is also sound: the `initVisibleClassTiles` comparator correctly pins `other` last then sorts descending, `--cols` is always an integer, and no `BEHAVIORAL_CLASS_CONFIG` lookup can go undefined.

No security or data-privacy findings — this is a pure display/CSS change with all dynamic bindings sourced from static config or a validated key set, never user input.

**Issue count**

- 🔴 Blocking: 0
- 🟡 Minor: 1 — `--cols` test asserts the style property but never the loaded grid class that consumes it, so a regression decoupling the grid from `--cols` ships green ([inline](https://github.com/linuxfoundation/lfx-self-serve/pull/1789#discussion_r3840816811)).
- ⚪ Nit: 3 — (1) skeleton is fixed at 6 cards / `lg:grid-cols-6` across two independent fields with no shared source, silent-desync risk; (2) the `describe` rename to a bare `'OrgGroupsComponent'` now collides with the top-level suite of the same name, muddying the report; (3) the new stat-strip suite resets `TestBed` in both `beforeEach` and `render()` (the `beforeEach` reset is dead work).
- ❔ Question: 0

The contrast tests asserting Tailwind class strings rather than a computed ratio is a real limitation (jsdom resolves no CSS), but you already called it out as a known tradeoff and the only true guarantee is an axe/Playwright check — recorded as an accepted tradeoff, not a finding.

**Bot reconciliation:** Copilot reviewed and generated no comments. Cursor's "Medium Risk" summary is about the client-side filter logic, which is parent-stack (GH-1778) work bleeding into this PR's diff through the stale stacking, not a change introduced here. @niravpatel27 already approved.

**Blocker to merge (not code):** the PR is currently `CONFLICTING` / `DIRTY` against its base, and you've explicitly commented that the stack is based on stale parent tips and must be restacked and merged bottom-up (#1787 → #1788 → #1790 → this). That unresolved conflict state is what gates this verdict, not the code.

🔴 **Needs changes before approval** — resolve the restack/merge-conflict state (per your own hold note), and optionally close the one minor test-seam gap. The code itself is approve-quality; re-request once the branch is restacked and clean.

dealako (changes requested):
Code is approve-quality — WCAG contrast swap and the `sm:contents` mobile-chip fix are both correct, no security/privacy findings. Requesting changes on two grounds:

1. The branch is currently `CONFLICTING`/`DIRTY` and, per your own hold note, must be restacked and merged bottom-up (#1787 → #1788 → #1790 → this) before it can land.
2. One minor test-seam gap (inline): the `--cols` test asserts the style property but not the loaded grid class that consumes it.

Full detail, bot reconciliation, and nits are in the summary comment. Re-request once restacked and clean.

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.spec.ts
manishdixitlfx: Thanks for the catch — this test-seam gap is real, but the `--cols`/stat-strip code it's about isn't actually part of this PR's diff anymore. Since your review, the branch was rebased directly onto post-merge `main` (PR #1790 landed), which squashed the parent-stack commits out of this PR's history entirely — `git diff origin/main...fix/GH-1782` now shows zero references to `--cols`, `statGridSkeletonClass`, or `statGridLoadedClass`. That code, and this gap, belong to the already-merged stat-strip work, not to #1789's scope (WCAG contrast + mobile chip visibility). Leaving unresolved code changes here would mean editing a file outside this PR's actual diff. Worth a follow-up issue against the stat-strip work if it's not already tracked, but nothing for this PR to fix.

manishdixitlfx:
## Review Feedback Addressed

Reviewed: @dealako's review (`CHANGES_REQUESTED`, commit `7796fad73`)

### Resolved since the review
- **Merge-conflict / restack blocker** (the review's stated primary gate): the branch has since been rebased directly onto post-merge `main` — `fix/GH-1779` landed as PR #1790, and this PR's base auto-retargeted to `main`. The branch is now `MERGEABLE`, no restacking left to do.

### No code change needed
- **`--cols` test-seam gap** (inline, [thread](https://github.com/linuxfoundation/lfx-self-serve/pull/1789#discussion_r3840816811)): valid observation, but the flagged code (`statGridSkeletonClass`/`statGridLoadedClass`/`--cols`) isn't part of this PR's diff — it's stat-strip work from the parent stack that already merged via #1790. `git diff origin/main...fix/GH-1782` has zero references to it. Replied on the thread; worth a follow-up against the stat-strip work directly if not already tracked.
- **Nit: duplicated skeleton-card count** and **nit: redundant `TestBed` reset in stat-strip's `beforeEach`**: same as above — both are in the already-merged stat-strip code, not in this PR's diff.
- **Nit: `describe` block renamed to a bare `'OrgGroupsComponent'`, colliding with the top-level suite**: already fixed as a side effect of the subsequent rebase — current `HEAD` uses the descriptive title `'OrgGroupsComponent — project label and row/foundation routing'`, no collision. Verified against the current file.

### Threads resolved

dealako:
Thanks for the fast, thorough turnaround, @manishdixitlfx 🙏 — and for the point-by-point response comment; it made this round easy to reconcile.

👏 **Nice work this revision**

- **Restack landed cleanly.** The branch now rebases onto post-merge `main` (`fix/GH-1779` shipped as #1790, base auto-retargeted), and the PR reports `MERGEABLE` with no conflicts. That was the primary gate last round — fully cleared.
- **Scope collapsed to exactly the right surface.** The diff is now just the two GH-1782 files (`org-groups.component.html` + `.spec.ts`), 252 lines. Every one of my prior-round test findings (the `--cols` seam, the skeleton-6 coupling, the redundant `TestBed` reset) was parent-stack stat-strip code that merged out via #1790 — correctly no longer your problem here. Your call to track a follow-up against the stat-strip work directly is the right one.
- **The `describe`-name collision is gone.** Current HEAD keeps the descriptive `'OrgGroupsComponent — project label and row/foundation routing'` title, and the new routing-scope helper is deliberately renamed `renderRow` (with a comment) to avoid shadowing the module-level `render`. Clean.
- **The new GH-1782 tests are notably better than a bare class check.** The name-hover test asserts DOM containment *and* `peer` sibling order (`compareDocumentPosition`), and the layout test pins `flex-1`/`min-w-0` on the element that actually carries them plus wrapper containment — they avoid the exact class-vs-property gap I flagged last round.

**Revision status (prior round)**

- ✅ Restack / merge-conflict blocker — resolved (rebased onto `main`, now `MERGEABLE`).
- ✅ `--cols` test-seam (minor), skeleton-6 coupling (nit), redundant `TestBed` reset (nit), duplicate `describe` name (nit) — all resolved by leaving scope when the stat-strip work merged via #1790; verified absent from this diff.

**Open items (all new, all optional)**

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2
  1. The `sm:contents` mobile-fold behavior is verified only by Tailwind class-string assertions — jsdom evaluates no media queries or `display:contents`, so nothing confirms the chip actually renders at a mobile viewport or that the desktop fold reproduces. You already call this out as a known tradeoff in the PR body; the real lock is a Playwright viewport test. Worth a follow-up, not a blocker.
  2. The two new explanatory template comments (the ~11-line `sm:contents` block and the `flex`-vs-strut note) run long and restate CSS-spec behavior. The behavior is genuinely non-obvious so they're defensible; trimming to one-line intents would age better. Take it or leave it.

**Bot reconciliation:** Copilot re-reviewed the 2 files and generated no new comments. No CodeRabbit run (auto-review disabled on the repo).

✅ **Approved with minor comments** — the restack cleared the gate, scope is tight, and the remaining two items are optional nits. Nice iteration.

dealako (approved):
Follow-up round. The restack cleared last round's primary gate: the branch is rebased onto post-merge `main`, the parent-stack stat-strip work merged out via #1790, and the PR is now `MERGEABLE` and cleanly scoped to the two GH-1782 files. Every prior-round test finding left scope with that merge and is verified absent here. No security or data-privacy concerns — pure display/CSS change with auto-escaped bindings.

Two optional nits remain (mobile-fold verified only via class strings not a runtime viewport test; dense template comments), both already acknowledged as tradeoffs. Neither blocks.

Approving. Full detail in the summary comment.

Issue #1778, opened 2026-08-22
feat(org-lens): add a filter bar to the Groups page


manishdixitlfx:
Parent: #1776

**Problem.** `/org/groups` has no search, no filters, no sort and no pagination — the only Org Lens page like that. Memberships has search + tier + renewal + sortable table. Projects has foundation + employee filters, sortable columns, pagination and CSV export. People → Committee has search + foundation + committee filters and a sortable table. Groups has an unbounded flat list. For a large member that is 100+ rows with no way to answer "show me our CNCF groups."

Filtering was specified in [LFXV2-1879](https://linuxfoundation.atlassian.net/browse/LFXV2-1879) / [LFXV2-1880](https://linuxfoundation.atlassian.net/browse/LFXV2-1880) and discarded unbuilt.

**Fix.** Reuse the filter bar from `org-people/components/committee-members/committee-members.component.html` — same form primitives, same `data-testid` conventions:

- search input (group name, foundation)
- foundation select
- behavioral-class select, options built from `BEHAVIORAL_CLASS_CONFIG` with live counts

Depends on #1777 for foundation *names* in the dropdown; can ship against slugs first if that lands later.

## Acceptance

- Search filters by group name and foundation, case-insensitive.
- Foundation and class selects filter the list; all three compose (AND).
- Filtered-empty state is distinct from the no-groups-at-all state and offers a clear-filters action — copy per the committee-members precedent ("No groups match the current filters").
- Stat strip continues to reflect the **unfiltered** roster, or is explicitly re-scoped — pick one and make it legible.
- Default sort (seat count desc, then alphabetical) is preserved.

**Open decision (see epic).** Whether filter state belongs in the URL (`?q=`, `?foundation=`) as LFXV2-1880 specified, or stays in-memory.

[LFXV2-1879]: https://linuxfoundation.atlassian.net/browse/LFXV2-1879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LFXV2-1880]: https://linuxfoundation.atlassian.net/browse/LFXV2-1880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[...]

PR #1790, opened 2026-08-23
fix(org-lens): render a tile per behavioral class in groups stat strip

manishdixitlfx:
## Summary

- Replaces the org-groups stat strip's hardcoded Working Groups + SIGs tiles with one tile per behavioral class that actually has groups, sorted largest-first with the `other` catch-all pinned last. On a live org the strip previously read 46 / 171 / 19 / 8 (Total Groups / Total Seats / Working Groups / SIGs) — 19 groups silently unaccounted for. It now sums to the total.
- Fixes a latent Tailwind purge bug found while making the tint dynamic: `bg-rose-50` (ambassador-program) was never being generated anywhere in `apps/lfx-one`'s scanned content, so that tile — and the same class on the committee dashboard/table, my-groups cards, and both public group pages — rendered with no background tint. Safelisted `BEHAVIORAL_CLASS_CONFIG`'s classes to fix it everywhere it's consumed.
- Adds `OrgLensGroupClassTile` to `@lfx-one/shared/interfaces` as the tile view-model.

### Stacked on #1788

Third PR in a four-deep stack for epic #1776:

```
main
 └─ #1787  feat/GH-1778             filter bar
     └─ #1788  fix/GH-1783-…        copy tokenization
         └─ #1790  ← this PR        stat strip
             └─ #1789  fix/GH-1782  contrast + mobile chip
```

Base is `fix/GH-1783-groups-loading-and-copy`, not `main`, so the diff is scoped to this branch's own change. Merge bottom-up: #1787, #1788, this, then #1789.

## Test plan

- [x] `yarn lint:check`, `yarn check-types`, `yarn format:check` — clean
- [x] `yarn build` — succeeds (SSR + browser bundles)
- [x] New/updated specs in `org-groups.component.spec.ts` cover: zero-count classes render no tile, non-zero classes render a labeled tile, counts sum to `total_groups`, ordering is largest-first with `other` pinned last even at the highest count, `--cols` matches the actual rendered grid, and both loading skeletons agree on placeholder count — verified via mutation testing that each assertion actually fails on the regression it targets
- [x] Confirmed in the built CSS that `bg-rose-50` and all other `BEHAVIORAL_CLASS_CONFIG` tints now emit
- [x] Manual check on a live org with a mixed class set — **Oracle America Inc., 46 groups**, verified in the integration worktree (all four stacked branches merged) at `localhost:4200/org/groups`:
  - Seven tiles rendered, `--cols="7"` matching the rendered count exactly
  - Order left-to-right: Total Groups 46 · Total Seats 171 · Working Groups 19 · SIGs 8 · Oversight 5 · Ambassadors 1 · **Other 13**
  - `other` correctly pinned last despite outranking SIGs, Oversight and Ambassadors on count — the exact case the sort's `other`-pin exists for
  - Class counts sum back to the total: 19 + 8 + 5 + 1 + 13 = 46 = Total Groups

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
@@ -63,25 +63,30 @@ export class OrgGroupsComponent {
   // doesn't cover. Keep in sync with the stat-strip markup.
   private readonly fixedStatTileCount = 2;
 
-  // Fixed placeholder count for the loading skeletons, set to the true max rendered tile count
-  // (2 fixed + all 6 behavioral classes non-zero — see statGridLoadedClass) so the skeleton and
-  // loaded grid only ever diverge on the zero-classes edge case, per GH-1779's "skeleton tile
-  // count matches the rendered tile count" acceptance criterion.
-  protected readonly statSkeletonTiles: readonly number[] = [1, 2, 3, 4, 5, 6, 7, 8];
+  // Placeholder count for the loading skeletons, derived from BEHAVIORAL_CLASS_CONFIG (one entry
+  // per GroupBehavioralClass member) rather than a hand-maintained literal, so the true max
+  // rendered tile count (2 fixed + every behavioral class non-zero) can't silently drift out of
+  // sync with the type if a class is ever added — see statGridClass for how the grid layout
+  // tracks this same value. Per GH-1779's "skeleton tile count matches the rendered tile count"
+  // acceptance criterion, the skeleton and loaded grid only ever diverge on the zero-classes edge
+  // case (per dealako's PR #1790 review).
+  protected readonly statSkeletonTiles: readonly number[] = Array.from(
+    { length: Object.keys(BEHAVIORAL_CLASS_CONFIG).length + this.fixedStatTileCount },
+    (_, i) => i + 1
+  );
 
-  // Mobile fixed at 2 columns, sm+ fixed at 3. At `lg:` and up, the loaded grid switches to one
-  // equal-width column per tile via --cols (2-8, since a class only tiles when non-zero) — that's
-  // the tightest case, not the roomiest: at 7-8 tiles columns get narrow enough for a longer
-  // label to wrap. Accepted deliberately, since a static per-count cap would reintroduce the
-  // ragged trailing row this grid exists to avoid. The skeleton and loaded grids share this same
-  // base and now share the same max card count (8, see statSkeletonTiles) — layout shift on data
-  // arrival only happens when fewer than 8 tiles end up rendering. The value is a plain --cols
-  // integer rather than a repeat(min(var(--cols),N)) clamp because CSS Grid requires an integer
-  // repeat count and silently drops the whole declaration otherwise (see
-  // events-summary-section.component.html for the same constraint).
+  // Mobile fixed at 2 columns, sm+ fixed at 3. At `lg:` and up, both the skeleton and the loaded
+  // grid switch to one equal-width column per tile via --cols — the skeleton binds it to
+  // statSkeletonTiles.length (see the template), the loaded grid to statTileCount() (2-8, since a
+  // class only tiles when non-zero). Sharing one grid class means the two can never drift out of
+  // sync on layout, only on --cols' value while data is still loading — the tightest live case is
+  // not the roomiest: at 7-8 tiles columns get narrow enough for a longer label to wrap. Accepted
+  // deliberately, since a static per-count cap would reintroduce the ragged trailing row this grid
+  // exists to avoid. The value is a plain --cols integer rather than a repeat(min(var(--cols),N))
+  // clamp because CSS Grid requires an integer repeat count and silently drops the whole
+  // declaration otherwise (see events-summary-section.component.html for the same constraint).
   private readonly statGridBase = 'grid grid-cols-2 gap-4 sm:grid-cols-3';
-  protected readonly statGridSkeletonClass = `${this.statGridBase} lg:grid-cols-8`;
-  protected readonly statGridLoadedClass = `${this.statGridBase} lg:[grid-template-columns:repeat(var(--cols),minmax(0,1fr))]`;
+  protected readonly statGridClass = `${this.statGridBase} lg:[grid-template-columns:repeat(var(--cols),minmax(0,1fr))]`;
 
   // ── Auth / access guards (mirrors org-meetings pattern) ───────────────────
   protected readonly hasNoOrgAccess: Signal<boolean> = computed(
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
@@ -63,23 +63,24 @@ export class OrgGroupsComponent {
   // doesn't cover. Keep in sync with the stat-strip markup.
   private readonly fixedStatTileCount = 2;
 
-  // Fixed placeholder count for the loading skeletons — the real tile count is only known once
-  // data resolves, so both skeleton branches render this many cards to avoid a second layout shift.
-  protected readonly statSkeletonTiles: readonly number[] = [1, 2, 3, 4, 5, 6];
+  // Fixed placeholder count for the loading skeletons, set to the true max rendered tile count
+  // (2 fixed + all 6 behavioral classes non-zero — see statGridLoadedClass) so the skeleton and
+  // loaded grid only ever diverge on the zero-classes edge case, per GH-1779's "skeleton tile
+  // count matches the rendered tile count" acceptance criterion.
+  protected readonly statSkeletonTiles: readonly number[] = [1, 2, 3, 4, 5, 6, 7, 8];
 
   // Mobile fixed at 2 columns, sm+ fixed at 3. At `lg:` and up, the loaded grid switches to one
   // equal-width column per tile via --cols (2-8, since a class only tiles when non-zero) — that's
   // the tightest case, not the roomiest: at 7-8 tiles columns get narrow enough for a longer
   // label to wrap. Accepted deliberately, since a static per-count cap would reintroduce the
   // ragged trailing row this grid exists to avoid. The skeleton and loaded grids share this same
-  // base so they don't diverge on classes, but the real tile count (2-8) is only known once data
-  // resolves and the skeleton's card count is fixed at 6 — some layout shift when data arrives is
-  // unavoidable at any breakpoint. The value is a plain --cols integer rather than a
-  // repeat(min(var(--cols),N)) clamp because CSS Grid requires an integer repeat count and
-  // silently drops the whole declaration otherwise (see events-summary-section.component.html
-  // for the same constraint).
+  // base and now share the same max card count (8, see statSkeletonTiles) — layout shift on data
+  // arrival only happens when fewer than 8 tiles end up rendering. The value is a plain --cols
+  // integer rather than a repeat(min(var(--cols),N)) clamp because CSS Grid requires an integer
+  // repeat count and silently drops the whole declaration otherwise (see
+  // events-summary-section.component.html for the same constraint).
   private readonly statGridBase = 'grid grid-cols-2 gap-4 sm:grid-cols-3';
-  protected readonly statGridSkeletonClass = `${this.statGridBase} lg:grid-cols-6`;
+  protected readonly statGridSkeletonClass = `${this.statGridBase} lg:grid-cols-8`;
   protected readonly statGridLoadedClass = `${this.statGridBase} lg:[grid-template-columns:repeat(var(--cols),minmax(0,1fr))]`;
 
   // ── Auth / access guards (mirrors org-meetings pattern) ───────────────────
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
@@ -11,6 +11,7 @@ import type {
   BehavioralClassDisplayConfig,
   GroupBehavioralClass,
   OrgDropdownOption,
+  OrgLensGroupClassTile,
   OrgLensGroupSummary,
   OrgLensGroupsResponse,
   OrgLensGroupVm,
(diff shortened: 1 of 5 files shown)


apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
```diff
@@ -33,6 +33,29 @@ export class OrgGroupsComponent {
   protected readonly committeeLabel = COMMITTEE_LABEL;
   protected readonly behavioralClassConfig = BEHAVIORAL_CLASS_CONFIG;
 
+  // Total Groups + Total Seats — the two fixed tiles in the template that visibleClassTiles()
+  // doesn't cover. Keep in sync with the stat-strip markup.
+  private readonly fixedStatTileCount = 2;
+
+  // Fixed placeholder count for the loading skeletons — the real tile count is only known once
+  // data resolves, so both skeleton branches render this many cards to avoid a second layout shift.
+  protected readonly statSkeletonTiles: readonly number[] = [1, 2, 3, 4, 5, 6];
```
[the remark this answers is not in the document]
manishdixitlfx: Done — bumped `statSkeletonTiles` from a fixed 6 to 8, which is the true max rendered tile count (2 fixed totals + all 6 behavioral classes, since `governing-board` isn't structurally excluded — it's usually 0 upstream but can classify non-zero for a broader "Governing Board"-ish category). `statGridSkeletonClass` moves from `lg:grid-cols-6` to `lg:grid-cols-8` to match.

See commit cc8b3329d: `org-groups.component.ts` skeleton count/grid, `org-groups.component.spec.ts` skeleton-card assertions updated 6→8.

Good catch — the fixed-6 skeleton could undercount the loaded grid by up to 2 cards depending on how many classes an org's roster actually spans.

manishdixitlfx:
## Review Feedback Addressed

Commit: cc8b3329d3827317f7b6de27911cb432724bfbd2

### Changes Made
- **org-groups.component.ts**: bumped the loading skeleton from a fixed 6 cards to 8 (the true max rendered tile count: 2 fixed totals + all 6 behavioral classes non-zero), and updated `statGridSkeletonClass` from `lg:grid-cols-6` to `lg:grid-cols-8` to match (per copilot-pull-request-reviewer)
- **org-groups.component.spec.ts**: updated the four skeleton card-count assertions from 6 to 8 to match

### Threads Resolved

manishdixitlfx:
⚠️ Do not merge directly. This PR targets a feature branch, not `main`, so branch protection does not apply and GitHub will show it as mergeable regardless of review or CI status. It is part of the #1776 stack and must merge bottom-up after #1787. It is also currently based on a stale tip of `feat/GH-1778` (`0e69c70b5`; that branch is now at `f75cdb3a5`) and will be restacked once #1787's review settles.

manishdixitlfx:
## Stack status — please hold merges until restacked

Thanks @niravpatel27 for the reviews on #1787, #1788 and #1789.

Two things happened in parallel today: review feedback landed on several branches, and each branch moved independently. The result is that **every PR above #1787 is now based on a stale tip of its parent**:

| PR | branch | head | based on current parent? |
|---|---|---|---|
| #1787 | `feat/GH-1778` | `f75cdb3a5` | — (base is `main`) |
| #1788 | `fix/GH-1783-groups-loading-and-copy` | `873dd34e2` | ❌ no |
| #1790 | `fix/GH-1779` | `cc8b3329d` | ❌ no |
| #1789 | `fix/GH-1782` | `7796fad73` | ❌ no |

So the diffs currently shown on #1788, #1790 and #1789 are not what would actually merge, and the approvals on #1788 and #1789 are against content that has since been superseded further down the stack. **Please do not merge any of these yet** — including #1787, since merging it first would leave the rest rebasing onto a moved target twice.

Also note #1788, #1790 and #1789 target feature branches rather than `main`, so branch protection does not apply to them and GitHub will show a green "ready to merge" button regardless of review or CI state. That button is not a signal on these three.

## Plan

1. **#1790 gets its human review** — it is the only one of the four still without one.
2. **Restack all three at once** onto their corrected parents: `fix/GH-1783-…` → `feat/GH-1778`, `fix/GH-1779` → `fix/GH-1783-…`, `fix/GH-1782` → `fix/GH-1779`.
3. **Re-request approval on the final shape.** The restack is a force-push, so GitHub will dismiss the existing approvals — doing steps 1 and 2 first means only one more review round rather than three.
4. **Merge bottom-up:** #1787 → #1788 → #1790 → #1789.

Doing the restack before #1790 is reviewed would just invalidate that review too, which is why it is sequenced this way.

manishdixitlfx:
@niravpatel27 @dealako — this is the one of the four still without a human review. Once it has one, I will restack all three upper PRs in a single pass and come back for a final approve on the merged shape.

dealako:
Hi @manishdixitlfx 👋 — nice, tightly-scoped change. Replacing the hardcoded WG+SIG tiles with one tile per non-zero behavioral class (largest-first, `other` pinned last) fixes a real correctness gap: the old strip left groups unaccounted for while the new one sums to the total. The latent `bg-rose-50` Tailwind purge fix is a good catch, and safelisting `BEHAVIORAL_CLASS_CONFIG` rather than one class fixes it everywhere it's consumed. Specs, TSDoc, and the mutation-tested assertions are all solid.

I ran a full parallel review across security, data privacy, correctness, performance, test coverage, style, and documentation. No blocking, security, or privacy findings. Two non-blocking items on the stat-strip logic, below.

### Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 1 — `statSkeletonTiles` fixed at `[1..8]` has no type/test guard against `GroupBehavioralClass` growth (durable version of the Copilot 6→8 thread, now resolved)
- ⚪ Nit: 0
- ❔ Question: 1 — class tiles sum to `groups.length`, but the Total Groups tile renders the server's separate `total_groups` field

### Bot reconciliation

Copilot's earlier thread flagged the fixed-6 skeleton undercounting the loaded grid; you correctly bumped it to 8. My minor finding agrees and extends it: at 8 it's still an unguarded literal with no max-count test, so a future 7th class would reintroduce the jump. Not a regression in this PR — a hardening suggestion.

### Final decision

✅ **Approved with minor comments** — the two items are hardening/contract concerns, not reachable bugs on current data. Your live-org verification (46 = 19+8+5+1+13) confirms the sum holds today. Safe to merge bottom-up per the stack.

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
```diff
@@ -29,9 +41,47 @@ export class OrgGroupsComponent {
...
+
+  // Total Groups + Total Seats — the two fixed tiles in the template that visibleClassTiles()
+  // doesn't cover. Keep in sync with the stat-strip markup.
+  private readonly fixedStatTileCount = 2;
+
+  // Fixed placeholder count for the loading skeletons, set to the true max rendered tile count
+  // (2 fixed + all 6 behavioral classes non-zero — see statGridLoadedClass) so the skeleton and
+  // loaded grid only ever diverge on the zero-classes edge case, per GH-1779's "skeleton tile
+  // count matches the rendered tile count" acceptance criterion.
+  protected readonly statSkeletonTiles: readonly number[] = [1, 2, 3, 4, 5, 6, 7, 8];
```
dealako: [minor] Skeleton tile count (8) isn't guarded against behavioral-class growth

Issue: `statSkeletonTiles` is a plain `[1..8]` literal, but the loaded grid caps at `fixedStatTileCount` (2) + `GroupBehavioralClass` member count. That coupling is type-enforced only in `initBehavioralClassCounts` (a `Record<GroupBehavioralClass, number>` literal — a 7th member is a compile error there). `statSkeletonTiles` has no such guard.
Proof: this line (literal `[1..8]`) vs `initBehavioralClassCounts` (typed Record literal, ~:197); `GroupBehavioralClass` has exactly 6 members (`committee.interface.ts:33`). No spec renders all six classes, so "skeleton count == max rendered tiles" is asserted nowhere.
Why it matters: adding a 7th class would let the loaded grid render 9 tiles while the skeleton stays 8 — silently reintroducing the load-time tile-count jump this feature exists to prevent, with no failing test.
Fix: derive the length from `Object.keys(BEHAVIORAL_CLASS_CONFIG).length + fixedStatTileCount`, and/or add a spec with all six classes non-zero asserting `statTileCount() === 8`. This is the durable version of the Copilot 6→8 thread (now resolved).

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
```diff
@@ -49,32 +99,74 @@ export class OrgGroupsComponent {
...
+    distinctUntilChanged()
+  );
+
   private readonly groupsData: Signal<OrgLensGroupsResponse | null | undefined> = this.initGroupsData();
 
   protected readonly groups: Signal<OrgLensGroupSummary[]> = computed(() => this.groupsData()?.groups ?? []);
-  protected readonly groupsWithClass: Signal<OrgLensGroupVm[]> = this.initGroupsWithClass();
+  // Not read from the template — filteredGroups() is what the @for actually iterates.
+  private readonly groupsWithClass: Signal<OrgLensGroupVm[]> = this.initGroupsWithClass();
   protected readonly totalGroups: Signal<number> = computed(() => this.groupsData()?.total_groups ?? 0);
```
dealako, line 129: [question] Class tiles sum to `groups.length`, but Total Groups shows server `total_groups`

Issue: the class tiles are derived from the `groups` array (one class per group), so they always sum to `groups().length`. The Total Groups tile renders `groupsData().total_groups`, a distinct server field the interface documents alongside a "non-board committees only" note on `groups`.
Proof: this line (`totalGroups` from `total_groups`) vs `initBehavioralClassCounts` counts derived from `groups()`; `org-groups.interface.ts:6,23` document `groups` as non-board-only and `total_groups` as a separate field. Every spec sets `total_groups === groups.length`, so the "tiles sum to total_groups" assertion only holds by fixture construction.
Why it matters: if the BFF ever returns `total_groups != groups.length` (e.g. it counts board committees the array excludes), the tiles visibly won't sum to Total Groups, and no test would catch it.
Fix: confirm the endpoint guarantees `total_groups === groups.length`; if not, derive Total Groups from `groups().length` or add a test with `total_groups != groups.length`. Not blocking — your live-org check (46 = 19+8+5+1+13) shows it holds today.

dealako (dismissed):
Approved with minor comments. Full parallel review (security, data privacy, correctness, performance, test coverage, style, docs) found no blocking, security, or privacy issues. Two non-blocking items are inline. See the summary comment for the reconciliation with Copilot's resolved skeleton-count thread.

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
manishdixitlfx: Good catch — fixed properly rather than just bumping the literal again. `statSkeletonTiles`' length now derives from `Object.keys(BEHAVIORAL_CLASS_CONFIG).length + fixedStatTileCount` instead of a hand-maintained array, so it can't silently drift from `GroupBehavioralClass`.

I went a step further on the grid-layout half of the coupling too: `statGridSkeletonClass`/`statGridLoadedClass` are now one `statGridClass` using the same `repeat(var(--cols),...)` arbitrary-value pattern for both states (the hand-maintained `lg:grid-cols-8` Tailwind literal is gone) — `--cols` is bound to `statSkeletonTiles.length` while loading and `statTileCount()` once loaded, so the layout itself now tracks the type too, not just the card count.

Also added the spec you suggested — a fixture with all six classes non-zero (including `governing-board`, which the existing fixture never exercises) asserting 6 rendered class tiles and `--cols=8`, so a 7th class fails loudly here.

See commit 37661d454.

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
manishdixitlfx: Confirmed by reading the BFF, not just the interface docs — `apps/lfx-one/src/server/services/org-lens-groups.service.ts:63-79` is the single code path that builds `OrgLensGroupsResponse`, and it sets `total_groups: groups.length` directly from the same `groups` array it returns (lines 72 and 77). There's no separate upstream field or second code path that could produce a different count — the two are guaranteed equal by construction in this codebase, not just true today by coincidence of the fixtures.

No code change needed. Good instinct to check, though — the interface comment ("non-board committees only") reads like `total_groups` could be an independently-sourced server total, and it's not obvious from `org-groups.interface.ts` alone that the BFF derives it from the same array.

manishdixitlfx:
## Review Feedback Addressed

Commit: 37661d4542e0f0764615e53948498ac37c32e13c

### Changes Made
- **org-groups.component.ts / .html**: `statSkeletonTiles`' length now derives from `Object.keys(BEHAVIORAL_CLASS_CONFIG).length + fixedStatTileCount` instead of a hand-maintained literal. Unified `statGridSkeletonClass`/`statGridLoadedClass` into one `statGridClass` (`repeat(var(--cols),...)`) so the grid layout tracks `--cols` for both loading and loaded states instead of a static Tailwind class name that could grow out of sync with `GroupBehavioralClass` (per @dealako)
- **org-groups.component.spec.ts**: added a test rendering all six behavioral classes non-zero (including `governing-board`), asserting 6 rendered tiles and `--cols=8` — a 7th class would fail this test (per @dealako)

### Questions Answered
- **org-groups.component.ts:124**: confirmed `total_groups` and `groups.length` can't diverge in this codebase — `org-lens-groups.service.ts:72-79` is the single code path building the response and sets `total_groups: groups.length` directly from the same array. No code change (asked by @dealako)

### Threads Resolved

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
dealako: ✅ Resolved in 37661d4. `statSkeletonTiles` now derives its length from `Object.keys(BEHAVIORAL_CLASS_CONFIG).length + fixedStatTileCount` (component.ts:73-76), and the skeleton + loaded grids share one `statGridClass` driven by `--cols`, so neither the count nor the layout can drift from the type. The new all-six-classes spec asserting `--cols='8'` plus the `toBe(8)` skeleton assertions catch a future 7th class. Nice, this is the durable fix rather than another literal bump.

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.ts
dealako: ✅ Resolved in 37661d4 (answer, no code change needed). Confirmed against `org-lens-groups.service.ts`: the single response path sets `total_groups: groups.length` from the same `groups` array (line 79), so the tile sum and Total Groups are equal by construction, not two independently-sourced fields. Thanks for reading the BFF rather than just the interface doc.

dealako:
Thanks for the quick turnaround, @manishdixitlfx — both items from the last round are fully closed out, and you went past the minimum on the first one.

### 👏 Nice work

- **Skeleton count fixed at the root, not patched.** Instead of bumping the literal again, `statSkeletonTiles` now derives its length from `Object.keys(BEHAVIORAL_CLASS_CONFIG).length + fixedStatTileCount` (component.ts:73-76). You also collapsed the separate skeleton/loaded grid classes into one `statGridClass` driven by `--cols`, so the count *and* the layout track the type. A future 7th `GroupBehavioralClass` can no longer silently outgrow the skeleton.
- **New all-six-classes spec.** The added test renders every behavioral class (including `governing-board`, never exercised by the old fixture) non-zero and asserts `--cols='8'`; combined with the `toBe(8)` skeleton assertions, drift is caught by a failing test.
- **Answered the `total_groups` question with evidence, not assertion.** You read the BFF (`org-lens-groups.service.ts:79`) and confirmed `total_groups: groups.length` comes from the same array in the single response path, so the tile sum equals Total Groups by construction. Exactly the right way to close a `[question]`.

### Revision tracking

- ✅ **Resolved** — `[minor]` skeleton tile count guard → derived from `BEHAVIORAL_CLASS_CONFIG` + unified grid + all-six-classes test (37661d4)
- ✅ **Resolved** — `[question]` class-tile sum vs `total_groups` → confirmed equal by construction at `org-lens-groups.service.ts:79` (37661d4)

### Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

A fresh parallel pass (security, data privacy, correctness, performance, tests, style, docs) over the two `fix(review)` commits found no new issues. Latest Copilot pass also generated no new comments.

### Final decision

✅ **Approved** — both prior items resolved, nothing new open. Ship it (bottom-up per the stack).

dealako (approved):
Follow-up review — approved. Both items from the prior round are resolved: the skeleton tile count is now derived from BEHAVIORAL_CLASS_CONFIG (with a unified --cols grid and an all-six-classes test), and total_groups === groups.length is confirmed by construction in org-lens-groups.service.ts. A fresh parallel pass over the two fix(review) commits found no new security, privacy, correctness, test, style, or documentation issues. See the summary comment for detail.

PR #1793, opened 2026-08-23
fix(groups): fix contrast and mobile chip visibility on public group pages

### LINKED ISSUE #1791: fix(groups): contrast and mobile class-tag visibility on the public group pages
## Problem

The two anonymous-reachable group pages carry the same two WCAG 2.1 AA failures that #1789 fixes on `/org/groups`. They render the same row shape, so the same defects were copied along with it.

**1. Foundation/project name fails contrast**

`text-gray-400` is `#90A1B9` in the LFX scale (`packages/shared/src/constants/colors.constants.ts` overrides Tailwind's default grays) — **2.63:1** on white, against a 4.5:1 requirement for normal text.

| file | line | element |
|---|---|---|
| `apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html` | 79 | `<p class="truncate text-xs text-gray-400" data-testid="public-foundation-groups-item-project">` |
| `apps/lfx-one/src/app/modules/groups/public-project-groups/public-project-groups.component.html` | 22 | `<p class="text-sm text-gray-400" data-testid="public-project-groups-foundation">` |

Fix: `text-gray-500` (`#62748E`, **4.76:1**) — the same swap #1789 made.

**2. Behavioral-class tag is hidden on mobile, leaving a colour-only signal**

Both files, line 87:

```html
<lfx-tag [value]="config.label" [severity]="'secondary'" class="shrink-0 hidden sm:flex" />
```

Below `sm` the tag disappears and the only remaining indicator of a group's behavioral class is the row icon's colour — and that icon is `aria-hidden="true"`. Colour alone is not a sufficient signal (WCAG 1.4.1 Use of Colour), and it conveys nothing to assistive tech at any width.

Fix: make the tag visible at every breakpoint, stacked under the group name on narrow viewports. #1789 does this on `/org/groups` with an `sm:contents` wrapper that folds back into the original desktop row layout at `sm+` — the same approach should port directly, since the row markup is nearly identical.

## Explicitly not in scope

Line 97 in both files — `<i class="fa-light fa-users text-gray-400 text-xs" aria-hidden="true">` — is decorative, sits beside a visible number, and is correctly left alone. Same call #1789 made for its `fa-user-check` icon.

## Why this is worth its own issue

These pages are reachable without authentication, so the audience is broader than the Org Lens surface #1789 covers. Scoped out of #1789 because they're different components with their own specs; the fix is mechanical but needs its own tests.

## Reference

- #1789 — the equivalent fix on `/org/groups`, including the `sm:contents` pattern and the contrast measurements
- Epic #1776

manishdixitlfx:
## Summary

Ports the two WCAG AA fixes from PR #1789 (`fix/GH-1782`, `/org/groups`) to
the two public group directory pages, per that PR's own explicit scoping
(it named these exact files/lines as out of scope and filed this ticket):

- **Contrast** — `text-gray-400` (`#90A1B9`, 2.63:1 on white) → `text-gray-500`
  (`#62748E`, 4.76:1, passes AA). Applied to the row project-name line in
  `public-foundation-groups.component.html` and the header foundation-name
  line in `public-project-groups.component.html`.
- **Mobile chip visibility** — the behavioral-class chip (Working Group / SIG
  / etc.) was `hidden sm:flex`, disappearing below the `sm` breakpoint and
  leaving only the row icon's color — `aria-hidden`, decorative — as a signal
  on mobile. Now visible at every breakpoint via an `sm:contents` wrapper that
  folds back into the unchanged desktop layout at `sm+`, matching the
  mechanism `org-groups.component.html` will use once #1789 lands.

Both target pages have a simpler row than `org-groups` (no separate
`peer`-hover overlay link — the row `<a>` itself is the flex container), so
the `sm:contents` wrapper sits directly inside it, and `items-center` →
`items-start sm:items-center` was added so the icon top-aligns against the
taller stacked mobile column instead of centering against it.

Neither component had a spec before this branch; both gained one, plus a
shared `apps/lfx-one/src/app/shared/testing/header-test-providers.ts` helper
(matchMedia jsdom shim + `<lfx-header/>` dependency mocks) extracted after
duplicating it once.

## Deliberate exclusions / deferred items

- **`role="listitem"` on the row `<a>`** — pre-existing on both pages (not
  introduced by this branch): an explicit ARIA role on an anchor overrides its
  implicit `link` role, so these rows are announced as plain list items to
  screen readers and drop out of the links rotor — a WCAG 4.1.2 defect.
  `org-groups.component.html` already has the correct pattern (role on a
  wrapping `<div>`, a separate stretched link inside) — copying it here is a
  real markup restructure (overlay anchor, `pointer-events` changes, retested
  focus/hover) well beyond this ticket's two named fixes. Left as-is;
  recommend a follow-up issue under epic #1776.
- **Truncated names have no tooltip/wrap fallback on narrow viewports** —
  pre-existing (the `truncate` classes predate this branch), same issue
  #1789 explicitly deferred for `org-groups` for the identical reason: the
  row's `aria-label` already exposes the full name to assistive tech; the
  gap is sighted narrow-viewport users only.
- **Spec assertions lock Tailwind class tokens, not computed layout/contrast**
  — jsdom evaluates no CSS, so a class-token assertion is the only
  unit-level lock available for a display/color fix, same trade-off
  `org-groups.component.spec.ts` documents for the identical `#1789` pattern.
  Real behavior verification would need the Playwright suite — not attempted
  here.

## Test plan

- [x] `yarn ng test --include='**/public-foundation-groups.component.spec.ts' --include='**/public-project-groups.component.spec.ts'` — 8/8 pass
- [x] Full suite (`yarn test`) — 691/691 pass
- [x] `yarn lint` — clean (one pre-existing unrelated warning in `user.service.ts`)
- [x] `yarn build` — succeeds

Closes #1791

### DIFF
--- a/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html
+++ b/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html
@@ -74,8 +74,9 @@
 
             <!-- Name + meta + behavioral class chip: stacked below `sm`; `sm:contents` folds this
                  wrapper away at `sm+` so the name/meta block and chip become direct flex items of
-                 the row again, reproducing the pre-existing desktop layout (see
-                 org-groups.component.html / PR #1789). -->
+                 the row again, reproducing the pre-existing desktop layout. `display: contents`
+                 gives the wrapper no box of its own at `sm+`, which is why `flex-1`/`min-w-0` live
+                 on the inner meta block rather than here (GH-1791, porting the pattern from #1789). -->
             <div class="flex flex-1 min-w-0 flex-col gap-2 sm:contents" data-testid="public-foundation-groups-item-meta-wrapper">
               <div class="flex flex-1 min-w-0 flex-col gap-0.5" data-testid="public-foundation-groups-item-meta-block">
                 <p class="truncate text-sm font-medium text-gray-900" data-testid="public-foundation-groups-item-name">{{ group.name }}</p>
--- a/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html
+++ b/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html
@@ -64,7 +64,7 @@
           <a
             [routerLink]="['/groups', group.uid]"
             role="listitem"
-            class="flex items-center gap-4 rounded-lg border border-gray-200 bg-white p-4 transition-colors hover:border-blue-200 hover:bg-blue-50/30 focus-visible:outline focus-visible:outline-2 focus-visible:outline-blue-500"
+            class="flex items-start gap-4 rounded-lg border border-gray-200 bg-white p-4 transition-colors hover:border-blue-200 hover:bg-blue-50/30 focus-visible:outline focus-visible:outline-2 focus-visible:outline-blue-500 sm:items-center"
             [attr.data-testid]="'public-foundation-groups-item-' + group.uid"
             [attr.aria-label]="group.name + ', ' + config.label">
             <!-- Icon -->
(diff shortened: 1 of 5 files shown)


niravpatel27 (approved):
Reviewed against LFX Self Serve conventions (rule surface, checklists, architecture docs) plus a general senior-engineering pass -- no code-level findings from either. Clean, well-tested WCAG 1.4.1/1.4.3 contrast and mobile-chip-visibility fix; iteration history shows the branch already closed several rounds of self-review. Two minor PR-shape nits, neither blocking: the branch name carries a descriptive suffix beyond `fix/GH-1791` (`fix/GH-1791-public-groups-a11y`), and the final commit/PR-title header is 74 chars vs. the 72-char team-style target (commitlint only hard-fails past 100, so this lands fine).

Issue #1798, opened 2026-08-24
Gate Org Lens leaderboard score-breakdown drawer behind org-lens-private-release flag


danoqualls:
## Background

PR #1276 (merged to main) added a leaderboard row score-breakdown drawer to Org Project Detail, backed by explicitly-flagged DEMO/PLACEHOLDER data (`org-leaderboard-detail-drawer.constants.ts`), pending a real Snowflake-backed source tracked in LFXV2-2934.

## Requirement

Gate this already-shipped feature behind the `org_lens_private_release` feature flag (LaunchDarkly key `org-lens-private-release`) so it's only visible where the flag is targeted on (dev environment initially), until the real-data source from LFXV2-2934 lands. Companion to the same flag also gating PR #1705's company-email feature.

## Notes

- Flag constant: `ORG_LENS_PRIVATE_RELEASE_FLAG` in `packages/shared/src/constants/feature-flags.constants.ts`
- Follow-up: once LFXV2-2934's real data lands, remove this gate along with the DEMO/PLACEHOLDER constants.

PR #1705, opened 2026-08-20
feat(org-lens): show company-affiliated emails in person detail drawer

danoqualls:
Business case: show the company emails of employees in the details panel to help with account troubleshooting.

This pull request enhances the Org Lens "People → All Employees" feature by introducing support for displaying company-affiliated email addresses in the person detail drawer. It implements a temporary demo logic to fabricate plausible sibling-domain variants for company emails, improves backend data fetching and caching, and updates the UI to present these emails. Additionally, it refines the handling of live-only (synthetic) people and strengthens the privacy of their identifiers.

**Backend: Company Email Derivation and Data Flow**
- Added logic in `OrgLensPeopleService` to fetch a person's company email from the database, generate sibling-domain variants for demo purposes (excluding personal domains), and expose these as `companyEmails` in the API response. Also includes a fallback for live-only people to look up their email from the live roster. [[1]](diffhunk://#diff-6baaf4e47295bb2dd8183f51de225f8664f6e111713b8862795211835078e796L132-R145) [[2]](diffhunk://#diff-6baaf4e47295bb2dd8183f51de225f8664f6e111713b8862795211835078e796R178-R197) [[3]](diffhunk://#diff-6baaf4e47295bb2dd8183f51de225f8664f6e111713b8862795211835078e796L283-R321) [[4]](diffhunk://#diff-6baaf4e47295bb2dd8183f51de225f8664f6e111713b8862795211835078e796L304-R363) [[5]](diffhunk://#diff-6baaf4e47295bb2dd8183f51de225f8664f6e111713b8862795211835078e796L486-R595) [[6]](diffhunk://#diff-1556974754f579db967d4e77bf401ef50c9476b22c8e5e4589b0077bb13766caR176-R182)
- Updated the cache validation logic to ensure only entries with the new `email` field are accepted, preventing stale cache data from omitting company emails.

**Backend: Privacy and Security Improvements**
[...]

Issue #1655, opened 2026-08-19
feat(org-lens): show company-affiliated email(s) in person-detail drawer


danoqualls:
## Summary
In the Org Lens person-detail side panel (opened by clicking a person's name), show the email address(es) on that person's LF account that are affiliated with the org at the domain level (e.g. hitachi.co.uk / hitachi.jp / hitachi.com), excluding personal/academic addresses.

## Scope for this issue
Real domain-matching requires a Salesforce Account multi-domain lookup and an LF SSO multi-email lookup, neither of which exist in this codebase or the Snowflake schema today. This issue covers building the UI and data shape now, backed by clearly-labeled demo data behind a seam (matching the pattern used for LFXV2-2735's TEMP-LOCAL-DEMO-ONLY shim), so the real data source can be wired in later without UI changes.

## Acceptance criteria
- `OrgAllEmployeeDetail` gains a `companyEmails: string[]` field
- Person-detail drawer displays company-affiliated email(s) beneath name/title when present
- No email section renders when the person has no email
- Backend demo-derivation is clearly marked as temporary (TEMP-DEMO-ONLY)

luismoriguerra:
duplicated with https://github.com/linuxfoundation/lfx-self-serve/issues/2095

PR #1803, opened 2026-08-24
feat(org-lens): gate company emails behind org-lens-private-release

danoqualls:
Adding feature flag to hide UI only PR that currently shows demo data in production.

This pull request implements a feature flag to control the visibility of company emails in the Org Lens person detail drawer, ensuring the feature is only available when explicitly enabled. The changes include wiring up the new flag in both the frontend component logic and the end-to-end tests, and updating the UI to respect the flag. This allows for safer rollout and easier testing of the feature while backend data is still under development.

Feature flag integration:

* Added a new constant `ORG_LENS_PRIVATE_RELEASE_FLAG` to gate the display of company emails and related UI in the Org Lens person detail drawer.
* Updated the `PersonDetailDrawerComponent` to check the feature flag before showing company emails or the "Company email unavailable" message, and to avoid fetching email data when the feature is disabled. [[1]](diffhunk://#diff-4b41f1b67c35e1232791d1e5e773ba32440a1e6c1340b7f6a43f47823e618983R6-R8) [[2]](diffhunk://#diff-4b41f1b67c35e1232791d1e5e773ba32440a1e6c1340b7f6a43f47823e618983R20) [[3]](diffhunk://#diff-4b41f1b67c35e1232791d1e5e773ba32440a1e6c1340b7f6a43f47823e618983R30) [[4]](diffhunk://#diff-4b41f1b67c35e1232791d1e5e773ba32440a1e6c1340b7f6a43f47823e618983L97-R101) [[5]](diffhunk://#diff-732864b41d6154ff8030ee664200cbdbb3e1b833ce0922c3d3412a5d59e29a38L53-R53)

Test updates:

* Modified E2E tests to stub the new feature flag for relevant scenarios, ensuring deterministic test results regardless of LaunchDarkly targeting. [[1]](diffhunk://#diff-deeb7e024b2a0ef60114788ada59d78a6c9d6dd85984424ad668f3fe17703325R16-R24) [[2]](diffhunk://#diff-deeb7e024b2a0ef60114788ada59d78a6c9d6dd85984424ad668f3fe17703325R256)
* Increased the timeout for email section visibility in the person detail drawer test to reduce flakiness.

## Summary
- Dark-launches the person-detail-drawer company-email feature (originally added in #1705, GH-1655) behind a new flag, `org-lens-private-release`, since the emails are currently derived from demo/placeholder logic pending a real multi-email-per-person backend source
- Flag defaults false (fails closed) via `FeatureFlagService.getBooleanFlag`; both the data-derivation path and the template's "unavailable" fallback are gated so there's no half-gated state
- Fixes the S3 e2e test to pin the flag on deterministically (`stubFeatureFlags`, the same localStorage-override helper the ROI specs use) instead of inferring flag state from DOM absence

## Context
#1705 merged (2026-08-21) without this gate, so the company-email feature has been live unguarded on `main`. This PR closes that gap. GH-1655 stays open to track the real backend (Snowflake/Salesforce multi-email source); once that lands, remove `companyEmailFeatureEnabled` + the `deriveDemoCompanyEmails` seam together and drop this gate.

Note: `ORG_LENS_PRIVATE_RELEASE_FLAG`'s doc comment also names a second, not-yet-wired consumer — the leaderboard row score-breakdown drawer (LFXV2-2934) — which is being gated separately on `feat/GH-1798`. That's intentional: the flag is meant to cover both, and the leaderboard gate is still in flight.

## Action item (not code)
Someone with LaunchDarkly dashboard access needs to create the `org-lens-private-release` flag and target it **on** for the dev environment only (off elsewhere) — there's no code-level dev default.

## Test plan
- [x] `yarn check-types`, `yarn lint:check`, `yarn format:check`, `yarn build` all pass
- [x] `npx playwright test org-contributions.spec.ts --list` compiles/lists correctly
- [ ] Manually confirm company emails don't render with the flag off (default)
- [ ] Once the LD dev flag is targeted on, confirm the feature renders as before gating


apps/lfx-one/e2e/org-contributions.spec.ts
```diff
@@ -248,6 +253,7 @@ test.describe('Org Lens Code Contributions — commits tab (S2)', () => {
 
 test.describe('Org Lens Code Contributions — person detail drawer (S3)', () => {
   test('S3: clicking a committer opens the shared person-detail drawer on the Code tab', async ({ page }) => {
+    await stubFeatureFlags(page, { [ORG_LENS_PRIVATE_RELEASE_FLAG]: true });
```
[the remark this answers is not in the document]
danoqualls: Added S3b in org-contributions.spec.ts: with org-lens-private-release OFF, asserts both person-detail-drawer-email and person-detail-drawer-email-unavailable are absent from the drawer, matching the flag-gate coverage pattern in weekly-brief-card.spec.ts.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/lfx-one/src/app/shared/components/person-detail-drawer/person-detail-drawer.component.ts
```diff
@@ -94,7 +98,7 @@ export class PersonDetailDrawerComponent implements OnDestroy {
   }
 
   private initCompanyEmails(): string[] {
-    if (this.drawer.loading() || this.drawer.error()) {
+    if (!this.companyEmailFeatureEnabled() || this.drawer.loading() || this.drawer.error()) {
```
audigregorie, line 101: **Major — the fetch path isn't gated; this only gates rendering, and the PR body claims otherwise**

`initCompanyEmails` gates what gets *rendered*, but the request itself lives in `PersonDetailDrawerService.fetchResult`, which never consults the flag: for Board/Committee openers (no `personKey`) the `POST /api/orgs/{uid}/lens/people/company-emails` still fires on every drawer open with the flag **off**, and the response is then discarded here. The PR body says this change avoids "fetching email data when the feature is disabled" and the Summary says "both the data-derivation path and the template's 'unavailable' fallback are gated so there's no half-gated state" — neither is true at the network layer. The new board-tab test actually codifies the un-gated behavior: `await emailLookup` (`org-people-board-tab.spec.ts:362`) only resolves because the request still fires.

Two clean resolutions, pick one:

1. **Gate the fetch** — inject `FeatureFlagService` into `PersonDetailDrawerService` and short-circuit the no-`personKey` branch to `EMPTY_FETCH_RESULT` when the flag is off (skipping the `_loading`/`_emailError` toggles too), then rewrite the board test to assert the request *never* fires (`page.on('request')` + a short settle wait) instead of awaiting the 500.
2. **Keep the fetch, fix the docs** — if the un-gated POST is deliberate (e.g., to keep the board test non-vacuous against the error path), correct the PR body's "avoid fetching" / "no half-gated state" claims so the description matches the code.

As-is, the stated purpose of the flag — keeping the demo-data feature dark in production — is met visually but not behaviorally.

apps/lfx-one/e2e/org-contributions.spec.ts
```diff
@@ -259,8 +265,27 @@ test.describe('Org Lens Code Contributions — person detail drawer (S3)', () =>
...
+    await page.getByTestId('org-contributions-committer-demo-aramirez-20260513').click();
+    await expect(page.getByTestId('person-detail-drawer-header')).toBeVisible({ timeout: DATA_LOAD_TIMEOUT });
+    await expect(page.getByTestId('person-detail-drawer-header')).toContainText('Ana Ramirez');
+    // The header renders synchronously from click context, before the async person-detail fetch
+    // resolves — wait for the fetch's loading skeleton to clear so the email assertions below reflect
+    // settled state, not a fetch that merely hasn't started rendering the email yet.
+    await expect(page.getByTestId('person-detail-drawer-loading')).toHaveCount(0, { timeout: DATA_LOAD_TIMEOUT });
+
+    await expect(page.getByTestId('person-detail-drawer-email')).toHaveCount(0);
+    await expect(page.getByTestId('person-detail-drawer-email-unavailable')).toHaveCount(0);
```
audigregorie, line 292: **Minor — this assertion is vacuous on the contributions page**

In the `personKey` fetch path, the service's `catchError` sets `_error`, never `_emailError`, so `companyEmailsUnavailable()` is always `false` here regardless of the flag — this line passes even with the template gate removed entirely. The board-tab spec already covers the fallback gate non-vacuously (it forces the company-emails POST to 500, making `emailError` genuinely true).

Keeping the line as a regression guard is fine, but a one-line comment saying so ("personKey path never sets emailError — guards a future wiring change") would stop the next reader from assuming it proves the gate here. Otherwise, consider dropping it.

audigregorie (commented):
## Code Review Summary

Clean, small, well-tested dark-launch gate that fails closed and covers both the flag-on and flag-off paths in e2e. One Major finding: the gate is render-only — the company-emails POST still fires for Board/Committee openers with the flag off, which contradicts the PR body's "avoid fetching email data when the feature is disabled" / "no half-gated state" claims (see thread on `person-detail-drawer.component.ts:101`). One Minor vacuous-assertion note on the S3b spec.

### Informational — outside the diff

- Server-side, `deriveDemoCompanyEmails` (`org-lens-people.service.ts`) still ships demo emails in the detail/company-emails payloads; the flag gates UI rendering only. That matches the PR's stated UI-only scope and the flag's doc comment — just noting that the payloads remain visible in the network tab until GH-1655's real backend lands.

### What's done well

- Fails closed (`getBooleanFlag(..., false)`), and both render branches are gated, so there's no half-gated UI state.
- E2E determinism done right: `stubFeatureFlags` via `addInitScript` pre-boot, explicit `false` pin instead of relying on absence, deep constant import avoiding the barrel trap.
- The board-tab test forces the email lookup to 500 so the fallback-gate assertion is genuinely non-vacuous — and says so in a comment. S3b's wait for the loading skeleton to clear is documented too.

MRashad26 (dismissed):
Reviewed the full diff (5 files, +62/-3) and ran the code-standards-enforcer agent against this PR. Both agree: PASS, clean and well-tested dark-launch gate.

**What this does:** Gates the person-detail-drawer company-email feature (merged unguarded in #1705) behind a new `org-lens-private-release` flag, since the emails are currently demo/placeholder data pending a real backend source. Fails closed (`getBooleanFlag(..., false)`), and fixes the S3 e2e test to pin the flag deterministically instead of relying on LaunchDarkly targeting.

**Verified correct:**
- **Double gate, no half-gated state**: `initCompanyEmails()` returns `[]` when the flag is off (data-derivation gate), and the template additionally requires `companyEmailFeatureEnabled() && companyEmailsUnavailable()` before showing the "Company email unavailable" fallback — so flag-off hides both the emails and the tell-tale fallback message, not just one of them.
- **New e2e coverage**: S3b (contributions drawer) and the new board-tab test both assert `person-detail-drawer-email` and `-email-unavailable` are absent with the flag off; the board-tab test additionally forces the company-emails POST to 500 so the fallback-hidden assertion is proven against a genuinely-true `emailError`, not just an untriggered condition — good attention to test validity.
- **Flag placement/docs**: `ORG_LENS_PRIVATE_RELEASE_FLAG` lives in the shared constants file with a JSDoc noting both intended consumers (this PR + the companion leaderboard-drawer PR #1828) and a removal plan.

**Minor, non-blocking observation:** the PR description says this change also avoids "fetching email data when the feature is disabled," but the actual gate only suppresses *display* (`initCompanyEmails()` / the template) — `PersonDetailDrawerService`'s company-emails POST to `/lens/people/company-emails` still fires regardless of flag state (confirmed in `person-detail-drawer.service.ts`, untouched by this diff). No data exposure results since rendering stays gated, but it's a needless network call to a demo endpoint while dark-launched — worth a follow-up if it matters, not blocking here.

**FYI for merge order:** this PR and #1828 both add the identical `ORG_LENS_PRIVATE_RELEASE_FLAG` block to `packages/shared/src/constants/feature-flags.constants.ts` at the same location — whichever merges second will hit a trivial conflict there. Already implicitly acknowledged in both PR descriptions.

No critical or major issues. Approving.

apps/lfx-one/src/app/shared/components/person-detail-drawer/person-detail-drawer.component.ts
danoqualls: Fixed in 30707d8. Injected `FeatureFlagService` into `PersonDetailDrawerService` and gated the no-`personKey` (Board/Committee opener) fetch branch on `companyEmailFeatureEnabled()`, so the `company-emails` POST no longer fires at all when the flag is off — it falls through to the existing block that already zeroes `_loading`/`_error`/`_emailError` and returns `EMPTY_FETCH_RESULT`, so no separate branch was needed.

Rewrote the board-tab e2e test to assert the request never fires (call counter + `expect(companyEmailCalls).toBe(0)`) instead of awaiting a forced-failure response, mirroring the existing "no fetch" pattern already in that file.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/lfx-one/e2e/org-contributions.spec.ts
danoqualls: Went with the clarifying-comment option rather than dropping the line: added a note explaining that this personKey path's `catchError` never sets `_emailError` (so the assertion is a regression guard, not a live check today), and pointing to the board-tab spec's new `company-emails request never fires` test as the non-vacuous flag-off coverage for the no-personKey path.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/lfx-one/src/app/shared/services/person-detail-drawer.service.ts
```diff
@@ -80,7 +85,7 @@ export class PersonDetailDrawerService {
         // travels in the body, not the query string, keeping it out of request-log URLs. `detail`
         // stays null here — there's no personKey to fetch real activity for, so the drawer's
         // "Detailed activity isn't available" state must stay truthful rather than showing verified-empty tabs.
-        if (context.email) {
+        if (context.email && this.companyEmailFeatureEnabled()) {
```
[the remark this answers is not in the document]
danoqualls: Fixed — added `toObservable(this.companyEmailFeatureEnabled)` as a third `combineLatest` source and destructured its emitted value into the `switchMap` branch instead of reading the signal imperatively, so a live flag flip while a no-personKey drawer is open now re-triggers the fetch/clear decision.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

audigregorie (approved):

PR #1814, opened 2026-08-24
feat(sidebar): add individual dashboard return link and reason modal

fayazg:
## Summary

- Adds a muted "Still need Individual Dashboard?" link under the profile card on the Me-tab sidebar
- Clicking opens a reason-capture modal ("We're retiring Individual Dashboard") with a single-select reason, optional free-text, and **Stay here** / **Continue to Individual Dashboard ↗** actions
- Tracks the link click and the continue action (with reason + optional comment) via Datadog RUM custom actions, attributed to the authenticated user so the funnel is queryable per-user and joinable with the ID side
- Gated behind the in-code `ID_MIGRATION_RETURN_LINK_ENABLED` constant for easy removal once migration completes
- Adds a shared `id-migration` analytics/gating contract in `@lfx-one/shared` and resolves the Individual Dashboard URL per environment

> **Draft:** pairs with the reciprocal myprofile retirement banner (ID-1139). Hold merge until that PR is ready so the two ship together.

Fixes LFXV2-3336


fayazg:
Paired with the reciprocal Individual Dashboard retirement banner: linuxfoundation/myprofile#1851 (ID-1139). Both are drafts and should merge together.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -185,6 +204,28 @@ export class SidebarComponent {
     this.profileMenu()?.hide();
   }
 
+  // Opens the Individual Dashboard return modal (LFXV2-3336). The link click is the migration
+  // funnel's first event (LINK_CLICK, no reason); the modal owns reason capture + navigation on
+  // "Continue". Both events carry the shared funnel tag + source_app so they join with the ID side.
+  protected openIdDashboardReturn(): void {
+    this.rumService.addAction(ID_MIGRATION_EVENTS.LINK_CLICK, {
```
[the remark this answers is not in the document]
fayazg: Good catch. Since `setUser()` isn't skipped during Admin Mode impersonation (only Intercom boot is), the click would have carried the impersonated user's LFID and corrupted their funnel. Both the LINK_CLICK (sidebar) and CONTINUE (modal) events are now gated on `!userService.impersonating()`; the link and navigation still work, only the analytics is suppressed. See commit 3c07c1839.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -185,6 +204,28 @@ export class SidebarComponent {
...
+      funnel: ID_MIGRATION_FUNNEL,
+      source_app: ID_MIGRATION_SOURCE_APP,
+    });
+
+    this.dialogService.open(IdMigrationModalComponent, {
+      // No PrimeNG header — the dialog body renders its own headline + subhead.
+      showHeader: false,
+      ariaLabelledBy: 'id-migration-heading',
+      width: '480px',
```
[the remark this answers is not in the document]
fayazg: Fixed — added `style: { maxWidth: '90vw' }` to the dialog config so the 480px width clamps to the viewport on mobile, matching the existing pattern in account-settings. See commit 3c07c1839.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.html
```diff
@@ -0,0 +1,48 @@
...
+        control="reason"
+        inputId="id-migration-reason"
+        [options]="reasons"
+        optionLabel="label"
+        optionValue="value"
+        size="large"
+        styleClass="w-full"
+        dataTest="id-migration-reason" />
+    </div>
+    <lfx-textarea [form]="form" control="comment" placeholder="Anything else? (optional)" [rows]="3" dataTest="id-migration-comment" />
```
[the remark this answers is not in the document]
fayazg: Fixed — added an associated `<label for="id-migration-comment">` and passed a matching `id` to the textarea, mirroring the Reason field above it, so the field now has an accessible name. See commit 3c07c1839.

fayazg:
## Review Feedback Addressed

Commit: 3c07c1839

### Changes Made
- **sidebar.component.ts / id-migration-modal.component.ts**: gated the LINK_CLICK and CONTINUE RUM actions on `!userService.impersonating()`, so an admin's click during Admin Mode impersonation no longer pollutes the impersonated user's migration funnel. Navigation and dialog behavior are unchanged. (per copilot[bot])
- **sidebar.component.ts**: added `style: { maxWidth: '90vw' }` to the dialog config so the 480px modal clamps to the viewport on small screens. (per copilot[bot])
- **id-migration-modal.component.html**: added an associated `<label for="id-migration-comment">` and a matching `id` on the textarea so the optional-comment field has an accessible name, mirroring the Reason field. (per copilot[bot])

### Tests
- **id-migration-modal.component.spec.ts**: added a case asserting the CONTINUE analytics is suppressed under impersonation while navigation and close still fire.

### Threads Resolved

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.html
```diff
@@ -0,0 +1,57 @@
...
+        size="large"
+        styleClass="w-full"
+        dataTest="id-migration-reason" />
+    </div>
+    <div class="flex flex-col gap-1">
+      <label for="id-migration-comment" class="text-sm text-gray-900">Anything else? (optional)</label>
+      <lfx-textarea
+        [form]="form"
+        control="comment"
+        id="id-migration-comment"
```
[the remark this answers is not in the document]
fayazg: Done — switched to `[id]="'id-migration-comment'"`, so the id lands only on the native `<textarea>` and `label[for="id-migration-comment"]` associates with the edit field instead of the host. Matches the `lfx-input-text` precedent at `committee-settings.component.html:196-204`.

See commit 16454be76.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.html
```diff
@@ -0,0 +1,57 @@
...
+        placeholder="Anything else? (optional)"
+        [rows]="3"
+        dataTest="id-migration-comment" />
```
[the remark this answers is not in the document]
fayazg: Done — added `data-dd-privacy="mask"` on the wrapper; it cascades to the native `<textarea>`, so keystrokes (including text abandoned via "Stay here") are no longer recorded in Session Replay. `addAction` still sends the explicitly submitted comment. Confirmed `defaultPrivacyLevel: 'allow'` and `sessionReplaySampleRate: 100` in `datadog-rum.provider.ts:38-43`.

See commit 16454be76.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.html
```diff
@@ -0,0 +1,57 @@
...
+      label="Stay here"
+      severity="secondary"
+      [outlined]="true"
+      size="small"
+      type="button"
+      (onClick)="stayHere()"
+      data-testid="id-migration-stay-button" />
+    <lfx-button
+      label="Continue to Individual Dashboard"
+      icon="fa-light fa-arrow-up-right"
```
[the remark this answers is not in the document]
fayazg: Done — added `ariaLabel="Continue to Individual Dashboard (opens in a new tab)"` while keeping the shorter visible label. `continueToIndividualDashboard()` does `window.open(..., '_blank')`, so the context change is now announced.

See commit 16454be76.

fayazg:
## Review Feedback Addressed

Commit: 16454be76292a4a18899ff703f3995dcc61aa83c

### Changes Made
- **id-migration-modal.component.html**: bound `[id]` on `<lfx-textarea>` instead of a static `id=`. A static attribute on a custom element lands on both the host and (via the wrapper's `id` input) the native `<textarea>`, so `label[for]` was resolving to the non-labelable host. Same fix already applied to `lfx-input-text` in `committee-settings.component.html`. (per copilot[bot])
- **id-migration-modal.component.html**: added `data-dd-privacy="mask"` to the comment field. RUM runs with `defaultPrivacyLevel: 'allow'` and `sessionReplaySampleRate: 100`, so free text was being recorded verbatim in Session Replay — including drafts the user abandons via "Stay here", which contradicts the component's submit-only capture contract. `addAction` still sends the submitted comment. (per copilot[bot])
- **id-migration-modal.component.html**: added `ariaLabel="Continue to Individual Dashboard (opens in a new tab)"` on the Continue action so the new browsing context is announced; the visible label stays short. (per copilot[bot])

### Threads Resolved

apps/lfx-one/src/app/shared/services/datadog-rum.service.ts
```diff
@@ -50,4 +50,16 @@ export class DataDogRumService {
...
+   * Emit a custom RUM action (product event) with optional context, attributed to the user
+   * set via setUser(). Used for queryable per-user product analytics. No-op on the server.
+   */
+  public addAction(name: string, context?: Record<string, unknown>): void {
+    if (typeof window === 'undefined') {
+      return;
+    }
+
+    datadogRum.addAction(name, context);
+  }
```
audigregorie, line 78: **Major — Impersonation suppression belongs in the service, not at every call site**

`addAction` is a new public analytics API, but the impersonation gate lives at the two current call sites (`sidebar.component.ts`, `id-migration-modal.component.ts`). Both Segment (`segment.service.ts` — `setImpersonating` checked inside every emit) and Plausible (`plausible.service.ts`) suppress analytics **inside the service** precisely so no caller can forget and pollute an impersonated user's data. The next `addAction` caller won't know they need `if (!userService.impersonating())` — the funnel silently corrupts.

Suggest adding a `private impersonating = false;` field + `setImpersonating()` setter (mirroring `PlausibleService`), wired in `app.component.ts` beside the existing `segmentService.setImpersonating(...)` / `plausibleService.setImpersonating(...)` calls, and gating here:

```suggestion
  public addAction(name: string, context?: Record<string, unknown>): void {
    if (typeof window === 'undefined' || this.impersonating) {
      return;
    }

    datadogRum.addAction(name, context);
  }
```

Keep `addError` ungated — errors are session telemetry, not user-attributed funnel events. The call-site guards (and the modal's `UserService` inject) can then be dropped.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.html
```diff
@@ -128,6 +128,26 @@
...
+        <div class="px-[12px] mb-[16px] -mt-2 w-full" [class.hidden]="!showMeSelector()" data-testid="id-migration-return-slot">
+          <p class="text-xs text-gray-500 leading-5">
+            Still need Individual Dashboard?
+            <button
+              type="button"
+              (click)="openIdDashboardReturn()"
+              data-testid="id-migration-return-link"
+              class="rounded-sm font-semibold text-blue-600 underline underline-offset-2 hover:text-blue-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500">
+              Open it <i class="fa-light fa-arrow-up-right text-[10px]" aria-hidden="true"></i>
+            </button>
```
audigregorie, line 147: **Minor — "Open it" is a context-weak accessible name**

Screen-reader users who Tab directly to this button hear only "Open it, button" — the surrounding "Still need Individual Dashboard?" text isn't part of the control's accessible name. The modal's Continue button already gets an explicit `ariaLabel` for exactly this reason; same care here:

```suggestion
            <button
              type="button"
              (click)="openIdDashboardReturn()"
              aria-label="Still need Individual Dashboard? Open the return dialog"
              data-testid="id-migration-return-link"
              class="rounded-sm font-semibold text-blue-600 underline underline-offset-2 hover:text-blue-700 focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-500">
              Open it <i class="fa-light fa-arrow-up-right text-[10px]" aria-hidden="true"></i>
            </button>
```

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  private readonly rumService = inject(DataDogRumService);
+  private readonly userService = inject(UserService);
+  private readonly formBuilder = inject(FormBuilder);
+  private readonly platformId = inject(PLATFORM_ID);
+
+  // Spread into a mutable array — lfx-select's `options` input is typed `any[]`, and the source
+  // constant is a `readonly` tuple (`as const`, so the reason literals stay usable as a union type).
+  protected readonly reasons = [...ID_MIGRATION_REASONS];
+  protected readonly form: FormGroup = this.formBuilder.group({
+    reason: [ID_MIGRATION_REASONS[0].value],
```
audigregorie: **Minor — Defaulting `reason` to `missing_feature` biases the funnel data**

Because the select pre-selects "Missing a feature", anyone who clicks Continue without touching the dropdown is recorded as `missing_feature` — the funnel can't distinguish "user chose this reason" from "user never engaged the field", which will inflate the first option's share. If the reason data is meant to drive the retirement decision, consider a non-committal default (empty select with `placeholder` + a required selection, or an explicit `not_specified` value). The spec pins the current behavior, so this may be a deliberate product call — flagging in case it wasn't.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -185,6 +204,36 @@ export class SidebarComponent {
     this.profileMenu()?.hide();
   }
 
+  // Opens the Individual Dashboard return modal (LFXV2-3336). The link click is the migration
+  // funnel's first event (LINK_CLICK, no reason); the modal owns reason capture + navigation on
+  // "Continue". Both events carry the shared funnel tag + source_app so they join with the ID side.
+  protected openIdDashboardReturn(): void {
```
audigregorie, line 210: **Minor — The funnel's first event (LINK_CLICK) is untested**

The modal spec does a nice job pinning the CONTINUE contract (payload shape, comment trimming, whitespace omission, impersonation gate, navigation) — its docblock says it exists because "the migration funnel depends on" it. The same is true of LINK_CLICK and its impersonation gate here, but no sidebar spec covers them. A small spec (mock `DataDogRumService` + `DialogService`, assert the action payload and that impersonation suppresses it) would pin both halves of the funnel contract.

audigregorie (commented):
## Code Review Summary

Solid, convention-aligned work — the modal spec pins the analytics contract precisely, the SSR hydration pattern mirrors the me-selector correctly, and the textarea privacy handling follows the documented committee-settings precedent. Verified locally: eslint clean, `check-types` clean, modal spec 5/5 passing, full production build (SSR bundle) passes. One Major design-consistency issue (impersonation suppression belongs in `DataDogRumService`, matching Segment/Plausible) plus three Minors, each posted as its own thread.

### Notes — outside the diff

- **Free-text comment → Datadog, unmasked**: the `comment` value travels via `addAction` context, which `data-dd-privacy="mask"` does not cover (masking applies to Session Replay only). The template comment acknowledges this ("addAction still sends it") — deliberate, but worth a product/privacy sanity check since it's free-text attributed to an identified user.
- **Draft hold**: noted that merge is intentionally held until the reciprocal myprofile retirement banner (ID-1139) is ready.

### What's done well

- The spec pins the exact CONTINUE contract — payload shape, comment trimming, whitespace-only omission, impersonation gate, navigation — with a docblock explaining why.
- `[class.hidden]` + `@if` layering mirrors the me-selector SSR hydration pattern, with the rationale documented inline.
- `data-dd-privacy="mask"` and the bound-`[id]` approach on the textarea replicate the documented committee-settings precedent exactly.
- The prod URL divergence (`openprofile.dev` vs `myprofile.*`) is commented rather than left as a trap.

apps/lfx-one/src/app/shared/services/datadog-rum.service.ts
fayazg: Done — the gate is now a property of the service. `DataDogRumService` gets `private impersonating = false` plus a `setImpersonating()` setter (same shape as `PlausibleService` and `SegmentService`), and `addAction` short-circuits on `typeof window === 'undefined' || this.impersonating`. `addError` is left ungated for exactly the reason you gave — errors are session telemetry, not user-attributed funnel events.

Wiring lives in `app.component.ts` beside the existing `segmentService`/`plausibleService` calls. Both call-site guards are gone, and `id-migration-modal.component.ts` dropped its now-unused `UserService` inject (the sidebar keeps its own, used elsewhere).

See commit a114b48e3.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.html
fayazg: Fixed — added `aria-label="Still need Individual Dashboard? Open the return dialog"` to the button. Agreed that "Open it" alone is context-free in a screen reader's button list, since the surrounding `<p>` text is not part of the accessible name. Consistent with the `ariaLabel` on the modal's Continue button.

See commit a114b48e3.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
fayazg: Good catch — it was not a deliberate product call. Changed: `reason` now starts as `null` with a `placeholder="Select a reason"` on the select, and the CONTINUE payload sends `reason: reason ?? undefined` so an untouched field is absent rather than reporting a value the user never picked. That makes "answered" and "skipped" distinguishable in the funnel, and stops `missing_feature` absorbing every skip.

Left it optional rather than required — this is an exit flow, and gating it would add friction for someone who just wants to leave. The spec was updated accordingly: the old default-value test now asserts `null`, plus a new case pinning that an untouched reason is omitted from the payload.

See commit a114b48e3.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
fayazg: Covered, though not where you suggested — once the gate moved into `DataDogRumService` (your other comment), the contract belongs in a service spec rather than a component spec. Added `datadog-rum.service.spec.ts` pinning four cases: `addAction` forwards by default, is suppressed under `setImpersonating(true)`, resumes when it is cleared, and `addError` still fires while impersonating (so the asymmetry cannot be "fixed" by accident).

Skipped a `sidebar.component.spec.ts`: the component injects 10 services and has required inputs, and after the refactor the only sidebar-specific logic left in `openIdDashboardReturn` is the payload literal and the `dialogService.open()` call. Happy to add one if you would rather have it.

See commit a114b48e3.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  private readonly platformId = inject(PLATFORM_ID);
+
+  // Spread into a mutable array — lfx-select's `options` input is typed `any[]`, and the source
+  // constant is a `readonly` tuple (`as const`, so the reason literals stay usable as a union type).
+  protected readonly reasons = [...ID_MIGRATION_REASONS];
+  // No pre-selected reason: defaulting to an option makes "chose it" and "never touched the
+  // field" indistinguishable in the funnel, and inflates whichever option sits first.
+  protected readonly form: FormGroup = this.formBuilder.group({
+    reason: [null],
+    comment: [''],
```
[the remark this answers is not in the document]
fayazg: Done — the reason is now required.

`reason` gets `Validators.required` (no pre-selection), and Continue is bound to `[disabled]="form.invalid"`, so a `migration_id_continue` event can't be emitted without one of `ID_MIGRATION_REASONS`. "Anything else?" stays optional and is still omitted from the payload when empty.

The field is deliberately not pre-selected to a default: seeding the first option would inflate it in the funnel and make "chose it" indistinguishable from "never touched it". Required + empty placeholder gives an accurate signal without that bias.

Specs updated to pin it — the form is invalid until a reason is picked, valid on a reason alone, and every value in `ID_MIGRATION_REASONS` round-trips into the CONTINUE payload.

See commit 9f63aac17.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.html
```diff
@@ -0,0 +1,66 @@
...
+  </div>
+
+  <!-- Reason is required (Continue is disabled until one is picked); the comment is optional. -->
+  <form [formGroup]="form" class="flex flex-col gap-3">
+    <div class="flex flex-col gap-1">
+      <label for="id-migration-reason" class="text-sm font-medium text-gray-700">Reason</label>
+      <lfx-select
+        [form]="form"
+        control="reason"
+        inputId="id-migration-reason"
```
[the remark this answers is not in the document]
fayazg: Good catch — confirmed on both counts and fixed.

`lfx-select` does declare `required` (`select.component.ts:80`) and forwards it to `p-select` (`select.component.html:47`), which renders `[attr.aria-required]="required()"`. Leaving it unset meant the combobox was announcing `aria-required="false"` while Continue was disabled — a real mismatch for screen-reader users.

Changes:
- `[required]="true"` on the `<lfx-select>`, so the combobox now carries `aria-required="true"`.
- Red asterisk in the label, matching the required-label pattern used across the repo (`add-member-dialog.component.html:176`, `member-form.component.html:10`, `committee-basic-info.component.html:15`). It's `aria-hidden="true"` so screen readers get the state once from `aria-required` rather than hearing "star" as well.

See commit 337994156.

audigregorie (dismissed):

audigregorie (dismissed):

fayazg:
Merged latest `main` into the branch to clear the merge conflict.

The only conflicts were additive barrel-export collisions where both sides appended to the same spot:

- `packages/shared/src/constants/index.ts` — `social-listening.constants` (main) vs `id-migration.constants` (this branch)
- `packages/shared/src/interfaces/index.ts` — `social-listening.interface` + `user-preference.interface` (main) vs `id-migration.interface` (this branch)

Both resolved by keeping both sides' exports. The four `environment*.ts` files auto-merged (main's removal of the unused `urls.support` key landed intact). No feature code changed.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  public continueToIndividualDashboard(): void {
+    const { reason, comment } = this.form.getRawValue();
+    const trimmedComment = (comment ?? '').trim();
+
+    this.rumService.addAction(ID_MIGRATION_EVENTS.CONTINUE, {
+      funnel: ID_MIGRATION_FUNNEL,
+      source_app: ID_MIGRATION_SOURCE_APP,
+      reason,
+      // Omit empty comments so the funnel query can distinguish "left a note" from "didn't".
+      comment: trimmedComment || undefined,
```
dealako, line 74: [blocking][privacy] Unbounded user-attributed free text sent to Datadog RUM

Issue: The optional "Anything else?" comment is open-ended free text and is transmitted verbatim to Datadog RUM via `addAction`, keyed to the authenticated user (`setUser` binds LFID/name/email). There is no `maxlength` on the field, no scrubbing before transmit, and no in-app deletion/export path.

Proof: `continueToIndividualDashboard()` sends `comment: trimmedComment || undefined` into `rumService.addAction(...)`, which forwards straight to `datadogRum.addAction(name, context)` (`datadog-rum.service.ts:77`) with no redaction. The `<lfx-textarea>` passes no `[maxlength]`, so the field is unbounded. The `data-dd-privacy="mask"` attribute only masks Session Replay, not this payload — your own HTML comment notes "addAction still sends it."

Why it matters: A new PII-capable free-text field flows to a third-party analytics provider under a real identity, with no bound, no scrubbing, and no evidenced consent/disclosure or deletion path in this diff. Retention falls to Datadog defaults, not an app-controlled TTL.

Fix: Add a small `[maxlength]` on the comment, and either drop the free-text `comment` from the RUM payload (keep only the enumerated `reason`) or confirm Datadog RUM disclosure covers user-attributed free text and a data-subject deletion path exists.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -185,6 +204,31 @@ export class SidebarComponent {
...
+    this.rumService.addAction(ID_MIGRATION_EVENTS.LINK_CLICK, {
```
dealako, line 211: [minor] LINK_CLICK funnel emission from the sidebar is untested

Issue: The funnel's first event (LINK_CLICK with `funnel` + `source_app`) is emitted only here in `openIdDashboardReturn()`, but no test pins that contract. CONTINUE, comment trimming, and the impersonation gate are all covered; the LINK_CLICK half is not, and no `sidebar.component.spec.ts` exists.

Proof: `datadog-rum.service.spec.ts` asserts generic forwarding of the raw string `migration_id_link_click` with `undefined` context — it does not assert the sidebar emits the event, nor that it carries `funnel`/`source_app`, nor that the dialog opens. Agrees with @audigregorie's earlier note.

Why it matters: A regression that drops the context payload, renames the event, or fails to fire would silently break the funnel join with the ID side — the exact failure the service spec's docstring warns about.

Fix: Add a `SidebarComponent` spec mocking `DataDogRumService` + `DialogService` that asserts `addAction(ID_MIGRATION_EVENTS.LINK_CLICK, { funnel: ID_MIGRATION_FUNNEL, source_app: ID_MIGRATION_SOURCE_APP })` and `dialogService.open(IdMigrationModalComponent, ...)`.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    if (isPlatformBrowser(this.platformId)) {
+      window.open(environment.urls.individualDashboard, '_blank', 'noopener,noreferrer');
```
dealako: [nit] Blocked popup is still recorded as a completed migration

Issue: The return value of `window.open(...)` is discarded, then `dialogRef.close(true)` runs unconditionally. If the browser blocks the popup (returns `null`), CONTINUE is still emitted and the dialog closes as success, so the funnel records a migration that never navigated.

Proof: `window.open` result is unused; `close(true)` is unconditional. The spec mocks `window.open` to return `null` yet still asserts `close(true)`.

Why it matters: Minor funnel-accuracy skew. In practice `window.open` on a direct user gesture is rarely blocked, so impact is low.

Fix: Optionally branch on the `window.open` return when it is `null`; acceptable to leave as-is given the user gesture.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
```diff
@@ -9,8 +9,18 @@ import { BadgeComponent } from '@components/badge/badge.component';
 import { LensTabsComponent } from '@components/lens-tabs/lens-tabs.component';
 import { OrgSelectorComponent } from '@components/org-selector/org-selector.component';
 import { ProjectSelectorComponent } from '@components/project-selector/project-selector.component';
+import { IdMigrationModalComponent } from '@components/id-migration-modal/id-migration-modal.component';
```
dealako: [nit] New imports break alphabetical ordering

Issue: `IdMigrationModalComponent` (line 12) is placed after `project-selector`, and `DataDogRumService` (line 32) after `project-context.service`, both out of the sorted order the rest of the file follows.

Proof: lines 10–12 and 30–32 in this file. (`format:check` currently passes, so this does not block — downgraded to nit; the divergence is cosmetic.)

Fix: Run `yarn format`, or move `IdMigrationModalComponent` up near the other `@components/*` imports and `DataDogRumService` up among the sorted `@services/*` block.

packages/shared/src/interfaces/id-migration.interface.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import type { ID_MIGRATION_EVENTS, ID_MIGRATION_REASONS } from '../constants/id-migration.constants';
+
+/** A reason value a user can pick in the Individual Dashboard return modal (LFXV2-3336). */
+export type IdMigrationReason = (typeof ID_MIGRATION_REASONS)[number]['value'];
```
dealako, line 7: [nit] Exported derived types have no consumers

Issue: `IdMigrationReason` (line 7) and `IdMigrationEvent` (line 10) are exported from the shared barrel but referenced nowhere — the component and sidebar use the constants directly.

Proof: a repo-wide grep for either name returns only this definition file.

Why it matters: Dead exports add contract surface with no caller, so the intended type-safety benefit is never realized.

Fix: Wire them into the code that would benefit (e.g. type the modal form's reason value / RUM action name against them), or drop them until a consumer exists.

dealako (changes requested):
Hi @fayazg — thanks for a genuinely careful piece of work. The funnel contract is well specified, the modal spec pins CONTINUE/trimming/reason-required precisely, the impersonation gate now lives in `DataDogRumService` where it belongs, and the accessibility follow-ups (aria-required, labeled textarea, new-tab announcement) all landed. The prior Copilot and @audigregorie rounds have been addressed thoroughly.

One blocking privacy item remains that is distinct from the session-replay mask already applied, plus a few small non-blockers.

### Issues

🔴 **Blocking: 1** (incl. 1 privacy)
1. Free-text comment is sent to Datadog RUM via `addAction`, attributed to the authenticated user, unbounded and unscrubbed — the DOM `data-dd-privacy="mask"` only covers Session Replay, not the action payload (`id-migration-modal.component.ts:57`).

🟡 **Minor: 1**
1. The funnel's first event (LINK_CLICK) emitted by the sidebar is untested (`sidebar.component.ts:211`) — agrees with @audigregorie's earlier note; the service spec covers generic forwarding but not the sidebar's emission contract.

⚪ **Nit: 3**
1. Blocked popup still records CONTINUE and closes with success (`id-migration-modal.component.ts:61`).
2. New imports break alphabetical ordering (`sidebar.component.ts:12,32`).
3. Exported `IdMigrationReason` / `IdMigrationEvent` types have no consumers.

🔒 **Privacy: 1 finding (blocking)** (included in the 🔴 count above)
1. Unbounded user-attributed free-text → third-party analytics: `id-migration-modal.component.ts:57` — bound the field with `maxlength`, and either drop the free-text `comment` from the RUM payload (keep only the enumerated `reason`) or confirm Datadog RUM disclosure + a deletion path cover user-attributed free text.

### Bot reconciliation
- Copilot's Session-Replay masking finding was valid and is resolved by `data-dd-privacy="mask"`. My blocking item is the separate `addAction` transmit leg, which the mask does not cover (the author's own inline comment concedes "addAction still sends it").
- @audigregorie's LINK_CLICK-untested point is only partially addressed: the new `datadog-rum.service.spec.ts` pins generic forwarding + impersonation, not the sidebar's `LINK_CLICK` emission with `funnel`/`source_app`.

🔴 **Needs changes before approval** (draft is held for the reciprocal myprofile PR anyway, so this is a good moment to close the privacy item).

dealako:
Hi @fayazg 👋 — thanks for a genuinely careful piece of work. The funnel contract is well specified, the modal spec pins CONTINUE / comment-trimming / reason-required precisely, the impersonation gate now lives in `DataDogRumService` where it belongs, and the accessibility follow-ups (aria-required, labeled textarea, new-tab announcement) all landed. The prior Copilot and @audigregorie rounds were addressed thoroughly.

One blocking privacy item remains that is distinct from the session-replay mask you already applied, plus a few small non-blockers.

🔴 **Blocking: 1** (incl. 1 privacy)
- Free-text comment → Datadog RUM `addAction`, attributed to the authenticated user, unbounded and unscrubbed. The `data-dd-privacy="mask"` covers Session Replay only, not the action payload (`id-migration-modal.component.ts:57`).

🟡 **Minor: 1**
- LINK_CLICK sidebar emission is untested (`sidebar.component.ts:211`) — agrees with @audigregorie; the service spec covers generic forwarding, not the sidebar's emission contract.

⚪ **Nit: 3**
- Blocked popup still records CONTINUE and closes as success (`id-migration-modal.component.ts:61`)
- New imports break alphabetical ordering (`sidebar.component.ts:12,32`) — `format:check` passes, so cosmetic only
- Exported `IdMigrationReason` / `IdMigrationEvent` have no consumers

🔒 **Privacy: 1 finding (blocking)** (included in the 🔴 count above)
1. Unbounded user-attributed free text → third-party analytics: `id-migration-modal.component.ts:57` — bound the field with `maxlength`, and either drop the free-text `comment` from the RUM payload (keep only the enumerated `reason`) or confirm Datadog RUM disclosure + a deletion path cover user-attributed free text.

**Bot reconciliation:** Copilot's Session-Replay masking finding was valid and is resolved. My blocking item is the separate `addAction` transmit leg the mask does not cover (your own inline comment concedes "addAction still sends it").

🔴 **Needs changes before approval.** The draft is held for the reciprocal myprofile PR anyway, so this is a good moment to close the privacy item.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
fayazg: Done — added `apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.spec.ts`.

See commit 427d9d08c: four tests pin the LINK_CLICK contract — the exact payload (`funnel` + `source_app`), the dialog config passed to `DialogService.open()` (`showHeader: false`, `ariaLabelledBy`, `modal`, `dismissableMask`), the emit-before-open ordering, and that `idReturnLinkEnabled` reads the shared constant rather than a local copy. The spec uses the repo's `TestBed.overrideComponent(..., { set: { template: '', imports: [], providers: [] } })` pattern so the class logic is exercised without rendering children.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
fayazg: Fixed — good catch, and the spec contradiction you spotted was real.

See commit 427d9d08c: `continueToIndividualDashboard()` now opens the tab first and checks the return value. If `window.open` yields `null` (popup blocked, or SSR), it sets a `popupBlocked` signal and returns early — no CONTINUE action, dialog stays open — so the funnel can no longer count a user who never left. Because a dead-looking Continue button is worse than the original bug, the template reveals a manual `<a target="_blank">` fallback when that happens (a real anchor click is not subject to the popup blocker).

The spec was updated to match: the success-path `window.open` spy now returns a window, plus new tests asserting that a blocked popup emits nothing and leaves the dialog open, and that the server path touches neither `window` nor RUM.

apps/lfx-one/src/app/shared/components/sidebar/sidebar.component.ts
fayazg: Fixed — `IdMigrationModalComponent` and `DataDogRumService` are both back in sorted position (commit 427d9d08c). No logic change.

packages/shared/src/interfaces/id-migration.interface.ts
fayazg: Addressed, though partly by deletion rather than by adding a consumer (commit 427d9d08c).

- `IdMigrationReason` now has a real consumer: the modal's reason control is typed `this.formBuilder.control<IdMigrationReason | null>(null, Validators.required)`, so a value outside the shared reason set cannot be written into the form.
- `IdMigrationEvent` was removed instead. `ID_MIGRATION_EVENTS` is `as const`, so `ID_MIGRATION_EVENTS.CONTINUE` already carries its literal type at every call site — annotating with the alias would have been a no-op written only to satisfy the export. Deleting it removes the contract surface rather than manufacturing a fake caller.

dealako:
Hi @fayazg — thanks for turning this around fast, and for the commit message that maps each change back to a review item. This round closes four of the five threads cleanly.

👏 **Nice work**
- The LINK_CLICK gap is properly pinned now: `sidebar.component.spec.ts` asserts the exact `{ funnel, source_app }` payload (not `objectContaining`), the headless/dismissable dialog config, and the emit-before-open ordering. That was @audigregorie's outstanding point, fully addressed.
- `IdMigrationEvent` dropped and `IdMigrationReason` now types the reason control — the dead-export nit is resolved and the union earns its keep.
- Imports alphabetized; that nit's closed too.

One flag: the fix for the blocked-popup nit introduced a real regression that's a bigger correctness problem than the nit it replaced. `window.open(..., 'noopener,noreferrer')` returns `null` on success, so `if (!opened)` now takes the blocked branch on every Continue — CONTINUE never fires and the dialog never closes. This repo already solved this exact thing in `meeting-join.component.ts:660-663`. Both Cursor and Copilot caught it, and the new success-path test hides it because the mock returns a truthy `Window` that `noopener` can never produce.

### Issues

🔴 **Blocking: 2** (incl. 1 privacy)
1. `window.open` + `noopener` returns null on success → `!opened` treats every Continue as blocked; CONTINUE never emitted, dialog never closes (`id-migration-modal.component.ts:62-67`). New this round.
2. 🔒 Free-text comment still sent unbounded and user-attributed to Datadog RUM (`id-migration-modal.component.ts:74`). Carried over, deferred this round.

🟡 **Minor: 1**
1. Success-path spec mocks `window.open` → truthy `Window`, encoding a false premise that hides the regression above (`id-migration-modal.component.spec.ts:43`).

⚪ **Nit: 1**
1. Popup-blocked fallback anchor lacks the "opens in a new tab" announcement the Continue control has (`id-migration-modal.component.html:51`).

🔒 **Privacy — 1 finding (blocking)** (included in the 🔴 count above)
1. Unbounded user-attributed free text → third-party analytics: `id-migration-modal.component.ts:74`. Bound with `maxlength` and either drop the free-text `comment` from the RUM payload (keep only the enumerated `reason`) or confirm Datadog disclosure + a deletion path cover user-attributed free text. Deferred this round; if it's moving to a follow-up ticket, link it and I'll track it there.

### Revision tracking
- ✅ LINK_CLICK funnel emission now tested (was minor)
- ✅ Dead exported types resolved (was nit)
- ✅ Import ordering fixed (was nit)
- ⚠️ Blocked-popup handling: addressed, but the fix regressed the happy path — see blocking #1
- ❌ Free-text → Datadog RUM privacy item: still open (deferred)

### Bot reconciliation
- **Cursor Bugbot** (1 issue) and **Copilot** both flagged the `noopener` null regression — agreed, it's blocking #1.
- **Copilot** also flagged the fallback anchor's missing new-tab announcement — agreed, nit #1.

🔴 **Needs changes before approval** — one new correctness regression plus the still-open privacy item.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  // look inert, and keeps the dialog open — CONTINUE stays unsent until a tab actually opened.
+  protected readonly popupBlocked = signal(false);
+
+  public continueToIndividualDashboard(): void {
+    const { reason, comment } = this.form.getRawValue();
+    const trimmedComment = comment.trim();
+
+    // Navigate first, then log. A blocked popup would otherwise be recorded as a completed
+    // migration, so the funnel would count a user who never left.
+    const opened = isPlatformBrowser(this.platformId) ? window.open(this.individualDashboardUrl, '_blank', 'noopener,noreferrer') : null;
```
dealako, line 62: [blocking] Popup-blocked branch fires on every successful Continue

Issue: `window.open(url, '_blank', 'noopener,noreferrer')` returns `null` even when the tab opens successfully — `noopener` deliberately severs the handle. The new `if (!opened)` guard (line 64) therefore takes the blocked branch on every normal Continue click: `popupBlocked` is set, the method returns early, the CONTINUE RUM action is never emitted, and the dialog never closes. The funnel loses every real completion and users see a false "browser blocked the new tab" warning on the happy path.

Proof: line 62 opens with `'noopener,noreferrer'`; line 64 branches on `if (!opened)`. Per the HTML spec / MDN, `window.open` with `noopener` returns null on success. This repo already documents exactly this at `apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts:660-663`: "With noopener, window.open() may return null even on success — Only check if window.closed is explicitly true to detect popup blocking." Cursor Bugbot and Copilot both flagged it on this commit.

Why it matters: This is the inverse of the funnel-accuracy nit it set out to fix — instead of over-counting a rare blocked popup, it now under-counts (to zero) every successful migration and breaks the Continue UX for all users.

Fix: Don't use the `window.open` return as the success signal while `noopener` is set. Either mirror meeting-join — keep the CONTINUE emit + `close(true)` on the normal path and detect a genuine block with `newWindow !== null && newWindow.closed` — or emit CONTINUE + close regardless of the handle (my original nit judged the rare blocked-popup-on-a-user-gesture case acceptable to ignore). The emit-before-open ordering the ticket wants is preserved by emitting first, then opening.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    if (!opened) {
+      this.popupBlocked.set(true);
+      return;
+    }
+
+    this.rumService.addAction(ID_MIGRATION_EVENTS.CONTINUE, {
```
dealako, line 69: [blocking][privacy] Unbounded user-attributed free text still sent to Datadog RUM

Issue: (Carried over from the last round, still open — the commit message defers it explicitly.) The optional "Anything else?" comment is open-ended free text, transmitted verbatim to Datadog RUM via `addAction` (the `comment` field on line 74), keyed to the authenticated user (RUM `setUser` binds LFID/name/email). There is still no `maxlength` on the textarea, no scrubbing before transmit, and no in-app deletion/export path. The DOM `data-dd-privacy="mask"` covers Session Replay, not this `addAction` payload.

Proof: line 74 sends `comment: trimmedComment || undefined` into `rumService.addAction(...)`, which forwards to `datadogRum.addAction(name, context)` with no redaction; the `<lfx-textarea>` still passes no `[maxlength]`.

Why it matters: Unbounded, user-attributed free text going to a third-party analytics sink, with no bound and no deletion path, is a standing privacy exposure independent of the Session-Replay mask.

Fix: Bound the field with `maxlength`, and either drop the free-text `comment` from the RUM payload (keep only the enumerated `reason`) or confirm Datadog RUM disclosure + a deletion path cover user-attributed free text. If this is moving to a follow-up ticket, link it here and I'll track it there.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+  // driving the wrapped PrimeNG controls through the DOM, which would test the wrappers, not this.
+  const formOf = (c: IdMigrationModalComponent): FormGroup => (c as unknown as { form: FormGroup }).form;
+  const popupBlockedOf = (c: IdMigrationModalComponent): Signal<boolean> => (c as unknown as { popupBlocked: Signal<boolean> }).popupBlocked;
+
+  beforeEach(async () => {
+    close.mockClear();
+    addAction.mockClear();
+    // A non-null return is the success path: window.open only yields null when the popup is
+    // refused, which the blocked-popup test below opts into explicitly.
+    openSpy = vi.spyOn(window, 'open').mockReturnValue({} as Window);
```
dealako, line 43: [minor] Success-path test encodes a false premise, so it hides the regression above

Issue: `beforeEach` mocks `window.open` to return `{} as Window` and comments "window.open only yields null when the popup is refused." That premise is false whenever `noopener` is passed — the real call returns null on success too. So the green "continue opens ID… then emits CONTINUE" test asserts a path production never takes, and the suite stays green while every real Continue takes the blocked branch.

Proof: line 43 `vi.spyOn(window, 'open').mockReturnValue({} as Window)`; the component calls `window.open(..., 'noopener,noreferrer')` (component line 62). A truthy `Window` can never occur in a real browser with `noopener`.

Why it matters: The test gives false confidence in exactly the contract that regressed; fixing the component without fixing the mock would leave the test asserting the wrong behavior.

Fix: Model the real return. With the meeting-join approach, mock a window whose `.closed` is `false` for success and `true` (or return `null`) for the blocked case, and assert block-detection via `closed`, matching whatever fix lands in the component.

apps/lfx-one/src/app/shared/components/id-migration-modal/id-migration-modal.component.html
```diff
@@ -0,0 +1,80 @@
...
+        dataTest="id-migration-comment" />
+    </div>
+  </form>
+
+  <!-- Popup blocker refused the new tab, so no CONTINUE was logged and the dialog stayed open.
+       A real anchor click is not subject to the popup blocker, so this link always works. -->
+  @if (popupBlocked()) {
+    <p class="text-xs text-red-600" role="alert" data-testid="id-migration-popup-blocked">
+      Your browser blocked the new tab.
+      <a [href]="individualDashboardUrl" target="_blank" rel="noopener noreferrer" class="underline" data-testid="id-migration-popup-blocked-link"
```
dealako, line 51: [nit] Popup-blocked fallback link lacks the "opens in a new tab" announcement

Issue: The fallback anchor opens a new browsing context (`target="_blank"`) but, unlike the Continue button, its accessible name gives no warning of the context change, so screen-reader users on the blocked path aren't told a new tab will open. (Flagged by Copilot on this commit.)

Proof: line 51 anchor has visible text "Open Individual Dashboard" and no visually-hidden "(opens in a new tab)" / aria annotation, while the Continue control carries that announcement.

Why it matters: Minor a11y consistency gap on the fallback path.

Fix: Add the same "opens in a new tab" announcement (visually-hidden span or `aria-label`) used on the Continue control.

dealako (changes requested):
Formal review to accompany the summary above. 🔴 Requesting changes: one new correctness regression (`window.open` + `noopener` returns null on success, so every Continue is treated as popup-blocked) and the still-open free-text → Datadog RUM privacy item. Details in the four inline comments and the summary comment.

dealako (approved):
After discussing in Slack, the user comment is needed (product requirement). Based on this, I am approving. We can dismiss the flagged note on the RUM comment attribute.

Issue #1669, opened 2026-08-19
lfx-v2-query-service: duplicate access-check lines for resources sharing a parent tuple


bramwelt:
**Repo:** `linuxfoundation/lfx-v2-query-service`

### Background

Datadog RUM (app LFX One, last 7 days) shows meetings as the slowest surface in the product: `/meetings/?` view p95 ~40s, `GET /api/meetings/:uid/my-meeting-registrants` p95 ~46s. Traces show the cost sits in a single `nats.request` inside this service's access-check call to `lfx-v2-fga-sync`.

### Root cause

Child resources (meeting registrants, past-meeting participants, etc.) are deliberately assigned their parent meeting's `AccessCheckObject`/`AccessCheckRelation` so many rows share one authorization tuple — this is correct by design. But `BuildMessage` (`internal/service/resource_search.go`) deduped emitted check lines on `ObjectRef` (unique per row) instead of the access-check key, so a 500-registrant meeting emitted 500 identical check lines instead of 1.

### Why

Every duplicate row a page displays multiplies the size of the access-check request sent to fga-sync, unnecessarily inflating that request even when the underlying authorization is a single check. This is the dominant contributor to the `my-meeting-registrants` p95.

### Related

Companion defect tracked in a separate issue for `lfx-v2-fga-sync` (serial cache lookups / handler dispatch).

jordane:
Already fixed on `main` in the query-service via PR linuxfoundation/lfx-v2-query-service#65 (commit [`089d8a7`](https://github.com/linuxfoundation/lfx-v2-query-service/commit/089d8a7) plus review follow-ups `19f4880`, `220ef00`).

`BuildMessage` in [`internal/service/resource_search.go`](https://github.com/linuxfoundation/lfx-v2-query-service/blob/main/internal/service/resource_search.go#L187-L236) now dedupes emitted access-check lines on the `AccessCheckObject#AccessCheckRelation` key while still classifying every resource for post-check filtering.

PR #1815, opened 2026-08-24
perf(projects): bound query-service page scan

### LINKED ISSUE #1735: Bound unbounded FGA-checked pagination in ProjectService.getProjects
## Problem

`ProjectService.getProjects` (`apps/lfx-one/src/server/services/project.service.ts:307`)
calls `fetchAllQueryResources` with no `page_size` and no `maxResults` cap. On
`GET /api/projects`, this produces an unbounded, sequential
`/query/resources?page_token=…` scan — confirmed in trace `c566183ba89e39eb`
as ~10s of paginated calls (100–840ms each) preceding a single trailing
`POST /access-check` (8.2s). See parent epic #1734 for the full attribution.

The pagination helper (`query-service.helper.ts`) already gained a
`maxResults` early-stop option, but it is wired only into
`meeting.service.ts`, not `project.service.ts`.

## Why this matters

Every additional page adds one more page-fetch round trip, and (per the
epic's "Confirmed contributing mechanisms" #2) worsens queueing under
top-of-hour concurrency — cost is not per-request-constant, it compounds with
load.

## Investigation

- [ ] Determine the actual page-count distribution for `getProjects` in
      production (how many pages does a typical/worst-case org fan out to?).
- [ ] Confirm whether `getProjects` callers need the full result set or can
      tolerate a `maxResults` early-stop (mirroring `meeting.service.ts`'s
      usage) or a bounded `page_size`.
- [ ] Identify all callers of `getProjects` to check assumptions about
      completeness (any caller relying on an exact total count would need
      `failOnPartial`/different handling).

## Recommended changes

_To be filled in after investigation._

---
Tracked under #1734.

bramwelt:
## Summary

- `ProjectService.getProjects` called the `fetchAllQueryResources` helper with no `page_size`, so the query service's default of 50 turned a large org's project list into a long sequential `/query/resources?page_token=…` scan (confirmed ~10s pagination leg in trace).
- Requests `page_size: 500` (matching the de-facto value already used at other call sites) via a new named constant `QUERY_SERVICE_PAGE_SIZE`, overriding any caller-supplied `page_size`.
- No frontend caller sends `page_size` for this endpoint today, and the BFF response contract (`res.json(projects)`, plain array) is unchanged — only the size of individual upstream pages changes.

## Scope boundary

This only bounds the `/query/resources` pagination leg. The trailing `POST /access-check` batch call is intentionally out of scope, tracked separately as #1669/#1670.

Closes #1735

## Test plan

- [x] `yarn workspace lfx-one test:server -- project.service` — 3 new tests + existing suite pass (59 total)
- [x] `yarn lint`, `yarn check-types`, `yarn build` all pass
- [x] `./check-headers.sh` passes
- [x] Post-commit reviewer trio (general, self-serve conventions, learnings) returned clean
- [x] `/lfx-self-serve-pr-readiness` and `/preflight` both green

### DIFF
--- a/apps/lfx-one/src/server/constants/index.ts
+++ b/apps/lfx-one/src/server/constants/index.ts
@@ -4,6 +4,7 @@
 export * from './gateway.constants';
 export * from './meta.constants';
 export * from './public-profile.constants';
+export * from './query-service.constants';
 export * from './reddit.constants';
 export * from './rewards.constants';
 export * from './weekly-brief.constants';
--- a/apps/lfx-one/src/server/constants/query-service.constants.ts
+++ b/apps/lfx-one/src/server/constants/query-service.constants.ts
@@ -1,0 +1,9 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+/**
+ * Page size for `/query/resources` paginated reads. The query service defaults to
+ * 50 and caps at 1000 (lfx-v2-query-service design/types.go); 500 keeps the page
+ * count low without sitting at the upstream maximum.
+ */
+export const QUERY_SERVICE_PAGE_SIZE = 500;
--- a/apps/lfx-one/src/server/services/project.service.spec.ts
+++ b/apps/lfx-one/src/server/services/project.service.spec.ts
@@ -212,6 +212,39 @@ describe('ProjectService — create picker methods', () => {
     });
   });
 
+  describe('getProjects', () => {
+    it('requests QUERY_SERVICE_PAGE_SIZE, overriding any caller-supplied page_size', async () => {
+      proxyRequest.mockResolvedValueOnce(pageOf([{ uid: 'a', slug: 'a' }]));
+      addAccessToResources.mockImplementationOnce((_req: Request, projects: Project[]) => Promise.resolve(projects));
+
+      await service.getProjects(req, { page_size: 10 });
+
+      expect(proxyRequest.mock.calls[0][4]).toMatchObject({ type: 'project', page_size: 500 });
+    });
+
+    it('follows page_token across pages and returns the accumulated projects', async () => {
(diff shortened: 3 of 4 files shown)


dealako:
Hi @bramwelt 👋 — thanks for the tight, well-scoped fix.

This is a clean performance change: `getProjects` was paging `/query/resources` at the upstream default of 50, and this bounds it to a named `QUERY_SERVICE_PAGE_SIZE = 500` constant (safely under the query service's documented 1000 cap). The diff is minimal, the intent is well-documented in both the code comment and the constant's JSDoc, and the three new tests lock in the behavior that matters: the forced page size, `page_token` aggregation across pages, and the existing ROOT pseudo-project filter. Response contract is unchanged (still a plain `Project[]`).

I traced the override end to end: no server controller or frontend caller sends `page_size` to this endpoint today, so clobbering a caller-supplied value is a non-issue in practice and is the documented intent. `fetchAllQueryResources` correctly follows `page_token` pagination with 5xx retry, so the multi-page path is sound.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 0

### AI bot reconciliation
- **Cursor Bugbot** flagged this **Low Risk** (performance-only, response shape and access behavior unchanged). Agreed.
- **Copilot** reviewed all 4 files and generated no comments. Agreed.
- **CodeRabbit** auto-review is disabled on this repo (no findings).

Nothing to add beyond the bots. Verified their conclusions against the helper and call sites rather than taking them at face value.

### Decision
✅ **Approved** — no blocking, security, or privacy findings; scope boundary to the `/access-check` batch (#1669/#1670) is appropriate.

dealako (approved):
Approved. Clean, well-tested, well-scoped performance fix — bounds the `/query/resources` page scan to a named constant (500, under the upstream 1000 cap) with solid test coverage for page sizing, `page_token` aggregation, and the ROOT filter. No blocking, security, or privacy findings. Verified the bot conclusions (Cursor: Low Risk, Copilot: no comments) against the helper and call sites.

PR #1816, opened 2026-08-24
fix(groups): expose public group rows as links, not list items

### LINKED ISSUE #1794: fix(groups): public group rows announce as list items, not links (WCAG 4.1.2)
## Problem

On both public group directory pages, `role="listitem"` is applied **directly to the row `<a>`**:

| file | line |
|---|---|
| `apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html` | 66 |
| `apps/lfx-one/src/app/modules/groups/public-project-groups/public-project-groups.component.html` | 69 |

```html
<div class="flex flex-col gap-2" role="list" data-testid="public-foundation-groups-list">
  @for (group of groupsVm(); track group.uid) {
    <a
      [routerLink]="['/groups', group.uid]"
      role="listitem"                                   <!-- overrides the implicit link role -->
      class="flex items-center gap-4 rounded-lg …"
      [attr.aria-label]="group.name + ', ' + config.label">
```

An explicit ARIA role replaces an element's implicit role rather than adding to it. So each of these rows is exposed to assistive technology as a **list item, not a link**.

Consequences for screen reader users:

- The rows do not appear in the links rotor/list — the standard way of surveying what's navigable on a page. On a directory page whose entire purpose is navigation, that removes the primary means of using it.
- Nothing announces the row as actionable. The `aria-label` reads out the group name and class, but with no role to say it can be followed.
- Keyboard focus still lands on the element (it's an anchor with an `href`), producing the confusing combination of a focusable element announced as a non-interactive one.

This is **WCAG 2.1 4.1.2 Name, Role, Value** (Level A). Both pages are reachable without authentication, so the affected audience is broader than the Org Lens surfaces.

Pre-existing — the `role="listitem"` predates #1793 and was explicitly out of scope there, since fixing it is a markup restructure rather than a class swap.

## The fix already exists in this codebase

`org-groups.component.html:111-117` (from #1786) has the correct shape — role on a wrapping `<div>`, with a stretched link inside as a separate element:

```html
<div class="flex flex-col gap-2" role="list" data-testid="org-groups-list-items">
  @for (group of groupsWithClass(); track group.uid) {
    <div
      role="listitem"
      class="has-[.peer:hover]:border-blue-200 … relative flex items-center gap-4 …">
      <!-- stretched link covering the row, so the whole row is clickable and focusable
           while remaining a real link to assistive tech -->
```

Porting it to these two pages means:

1. Row `<a>` becomes a `<div role="listitem">` carrying the layout classes.
2. Add an absolutely-positioned stretched `<a>` inside it, carrying the `routerLink`, the `aria-label`, and the `peer` class.
3. Move the hover styling from the anchor's own `hover:` classes to the row's `has-[.peer:hover]:` classes, so hovering anywhere on the row still highlights it.
4. Re-verify focus-visible outline placement — the outline should read as covering the row, not a zero-size overlay.
5. Keep the existing `data-testid`s so #1793's new specs don't break, and extend them to assert the row exposes `role="link"` (or that a real anchor is present) rather than `listitem`.

Worth doing both pages in one change — the markup is identical between them.

## Verification

Class-token assertions won't catch a regression here. Verify with either a Playwright accessibility-tree snapshot asserting the rows expose a `link` role, or a manual VoiceOver/NVDA pass confirming the rows appear in the links rotor.

## Related

- #1793 — ports the contrast and mobile-chip fixes to these same two files; explicitly deferred this one
- #1789 — the equivalent a11y work on `/org/groups`
- #1786 — introduced the correct wrapping-div + stretched-link pattern
- Epic #1776

manishdixitlfx:
## Summary
- Both public group directory pages (`/foundations/:slug/groups` and `/projects/:slug/groups`) put `role="listitem"` directly on the row `<a>`, which replaces its implicit link role and drops the row out of the screen reader links rotor on pages whose entire purpose is navigation — a WCAG 4.1.2 defect.
- Ports the pattern already shipped in `org-groups.component.html` (#1786): `role="listitem"` moves to a non-interactive wrapping `<div>`; a separate absolutely-positioned "stretched link" `<a>` inside carries the actual navigation (`routerLink`, `aria-label`), tagged `peer` so the row's `has-[.peer:hover]:` classes still highlight it on hover. Focus-visible outline moves to the link so it covers the full row.

## Test plan
- [x] `yarn lint`, `yarn check-types`, `yarn test` (739/739), `yarn build` — all pass against current `main`
- [x] New Vitest assertions in both `*.component.spec.ts` files lock: the row is a `DIV` with `role="listitem"`; a real `<a>` exists inside it with an `href`, no `role` attribute (implicit link role), and the correct `aria-label`; and the `peer`/`has-[.peer:hover]:` hover-highlight pair is present
- [ ] No Playwright coverage added — these two public pages have no existing e2e scaffolding (route stubs/fixtures/helpers), unlike `org-groups`. Standing that up is out of scope for this fix.
- [ ] Full accessibility-tree verification (Chrome DevTools Accessibility pane against prod data, tabbing through rows to confirm keyboard focus/link role) is deferred to the integration validation pass per this repo's per-ticket worktree workflow — not yet performed.

## Notes
- Branch name (`fix/GH-1794-public-groups-link-role`) has a descriptive suffix beyond the strict `fix/GH-1794` convention — pre-assigned target worktree branch, left as-is rather than renamed.

Closes #1794

### DIFF
--- a/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html
+++ b/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html
@@ -61,12 +61,23 @@
       <div class="flex flex-col gap-2" role="list" data-testid="public-foundation-groups-list">
         @for (group of groupsVm(); track group.uid) {
           @let config = group.classConfig;
-          <a
-            [routerLink]="['/groups', group.uid]"
+          <div
             role="listitem"
-            class="flex items-start gap-4 rounded-lg border border-gray-200 bg-white p-4 transition-colors hover:border-blue-200 hover:bg-blue-50/30 focus-visible:outline focus-visible:outline-2 focus-visible:outline-blue-500 sm:items-center"
-            [attr.data-testid]="'public-foundation-groups-item-' + group.uid"
-            [attr.aria-label]="group.name + ', ' + config.label">
+            class="has-[.peer:hover]:border-blue-200 has-[.peer:hover]:bg-blue-50/30 relative flex items-start gap-4 rounded-lg border border-gray-200 bg-white p-4 transition-colors sm:items-center"
+            [attr.data-testid]="'public-foundation-groups-item-' + group.uid">
+            <!-- Whole-row click/focus target ("stretched link"), ported from org-groups.component.html
+                 (#1786): role="listitem" belongs on this non-interactive row div, not on an anchor,
+                 which is why the actual link is this separate absolutely-positioned <a> instead. `peer`
+                 lets the row react to this specific link's hover via has-[.peer:hover]: above. Unlike
+                 org-groups, this row has no second nested link, so no pointer-events-none/z-10 handoff
+                 is needed on the content below — don't drop `peer` in a "clean up unused classes" pass,
+                 it's what drives the row's hover highlight. -->
+            <a
+              [routerLink]="['/groups', group.uid]"
+              class="peer absolute inset-0 rounded-lg focus-visible:outline focus-visible:outline-2 focus-visible:outline-blue-500"
+              [attr.data-testid]="'public-foundation-groups-row-link-' + group.uid"
+              [attr.aria-label]="group.name + ', ' + config.label"></a>
+
             <!-- Icon -->
             <div class="flex h-9 w-9 shrink-0 items-center justify-center rounded-full" [class]="config.bgColor">
               <i [class]="config.icon + ' ' + config.color + ' text-base'" aria-hidden="true"></i>
@@ -107,7 +118,7 @@
             }
 
             <i class="fa-light fa-chevron-right text-gray-300 text-xs shrink-0" aria-hidden="true"></i>
-          </a>
+          </div>
         }
       </div>
     }
--- a/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.spec.ts
@@ -141,4 +141,48 @@ describe('PublicFoundationGroupsComponent — contrast and responsive row layout
     expect(rowClasses).toContain('items-start');
     expect(rowClasses).toContain('sm:items-center');
   });
+
+  it('exposes the row as a real link, not an anchor carrying role="listitem" (GH-1794)', async () => {
+    // WCAG 4.1.2 regression lock: an explicit role="listitem" on the anchor itself would replace
+    // its implicit link role, dropping the row out of the screen reader links rotor. A class-token
+    // assertion can't catch this — only checking the actual element identity/role can, which is
+    // exactly what the pre-fix markup (role="listitem" directly on the row <a>) would fail.
+    await render({ groups: [group()], total: 1 });
+
+    const row = fixture.nativeElement.querySelector('[data-testid="public-foundation-groups-item-g1"]');
+    expect(row).not.toBeNull();
+    expect(row?.tagName).toBe('DIV');
+    expect(row?.getAttribute('role')).toBe('listitem');
+
+    const link = fixture.nativeElement.querySelector('[data-testid="public-foundation-groups-row-link-g1"]');
+    expect(link).not.toBeNull();
+    expect(link?.tagName).toBe('A');
+    expect(row?.contains(link)).toBe(true);
+    expect(row).not.toBe(link);
+    // No role attribute at all — the element must rely on its implicit link role, which an <a>
+    // only gets once it has an href (otherwise it maps to `generic`), so that's asserted too.
+    expect(link?.getAttribute('role')).toBeNull();
+    expect(link?.getAttribute('href')).toBe('/groups/g1');
+    expect(link?.getAttribute('aria-label')).toBe('WG Identity & Trust, Working Groups');
+  });
+
+  it('row hover highlight keeps its peer / has-[.peer:hover] pair', async () => {
+    // `peer` on the link and `has-[.peer:hover]:` on the row are the two halves of the row's
+    // hover highlight — a comment alone doesn't stop either half from being dropped as an
+    // apparently-unused class, so both are asserted directly rather than just documented. A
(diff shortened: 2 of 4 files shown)


apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.spec.ts
```diff
@@ -141,4 +141,48 @@ describe('PublicFoundationGroupsComponent — contrast and responsive row layout
...
+    const row = fixture.nativeElement.querySelector('[data-testid="public-foundation-groups-item-g1"]');
+    const link = fixture.nativeElement.querySelector('[data-testid="public-foundation-groups-row-link-g1"]');
+    expect(row).not.toBeNull();
+    expect(link).not.toBeNull();
+
+    const linkClasses = (link?.getAttribute('class') ?? '').split(/\s+/);
+    expect(linkClasses).toContain('peer');
+    const rowClasses = (row?.className ?? '').split(/\s+/);
+    expect(rowClasses).toContain('has-[.peer:hover]:border-blue-200');
+    expect(rowClasses).toContain('has-[.peer:hover]:bg-blue-50/30');
```
dealako, line 186: [minor] Lock the stretched link's `absolute inset-0`, not just `peer`

Issue: The whole-row link depends entirely on `absolute inset-0` to stretch over the row. The specs lock `peer`, `role`, tag, `href`, and `aria-label`, but nothing asserts `absolute` or `inset-0`.
Proof: The anchor has empty content (`></a>`). Without `absolute inset-0` it collapses to a zero-size inline element — the full-row click target and focus outline both silently break while every existing assertion still passes.
Why it matters: A future "drop unused class" pass (the exact failure mode these tests were written to prevent, per the file's own comments) could remove the one load-bearing token pair that makes the PR's fix work, with a green suite.
Fix:
```ts
const linkClasses = (link?.getAttribute('class') ?? '').split(/\s+/);
expect(linkClasses).toContain('absolute');
expect(linkClasses).toContain('inset-0');
```
Applies identically to `public-project-groups.component.spec.ts`.

apps/lfx-one/src/app/modules/groups/public-foundation-groups/public-foundation-groups.component.html
```diff
@@ -61,12 +61,23 @@ <h1 class="font-display font-light text-2xl text-gray-900" data-testid="public-f
...
+            <!-- Whole-row click/focus target ("stretched link"), ported from org-groups.component.html
+                 (#1786): role="listitem" belongs on this non-interactive row div, not on an anchor,
+                 which is why the actual link is this separate absolutely-positioned <a> instead. `peer`
+                 lets the row react to this specific link's hover via has-[.peer:hover]: above. Unlike
+                 org-groups, this row has no second nested link, so no pointer-events-none/z-10 handoff
+                 is needed on the content below — don't drop `peer` in a "clean up unused classes" pass,
+                 it's what drives the row's hover highlight. -->
+            <a
+              [routerLink]="['/groups', group.uid]"
+              class="peer absolute inset-0 rounded-lg focus-visible:outline focus-visible:outline-2 focus-visible:outline-blue-500"
```
dealako, line 77: [question] Row name loses org-groups' `peer-hover:text-blue-700` — intentional?

Issue: The `org-groups` reference (the pattern this PR ports) wraps row content in a `peer-hover:text-blue-700` div so the group name turns blue on row hover. The public pages drop that wrapper entirely and hard-code `text-gray-900` on the name, so it never changes on hover. The inline comment justifies dropping only the `pointer-events-none`/`z-10` handoff (purpose 1 of the wrapper), not the hover text-color (purpose 2).
Proof: Reference `org-groups.component.html:196` carries `... peer-hover:text-blue-700 ...` on the content wrapper. Here `public-foundation-groups.component.html:94` (and the project equivalent) hard-code `text-gray-900` on the name `<p>` with no `peer-hover:` anywhere. Row hover still highlights via `has-[.peer:hover]:` border/bg, so this is cosmetic, not an a11y/functional defect.
Why it matters: Minor visual divergence from the pattern the PR claims to port, and the comment documents only one of the two behaviors dropped with the wrapper — a future reader can't tell if the missing hover color was intentional.
Fix: If intended, add a clause to the existing comment noting the name-color-on-hover was deliberately dropped. If not, add `peer-hover:text-blue-700 transition-colors` to the meta block and drop the hard-coded `text-gray-900` so the name inherits, matching org-groups.

dealako:
Hi @manishdixitlfx — thanks for this one. Clean, tightly scoped WCAG 4.1.2 fix: moving `role="listitem"` off the row `<a>` and onto a non-interactive wrapper, with a separate stretched-link anchor carrying the navigation, is the correct fix and a faithful port of the `org-groups` (#1786) pattern. The regression-lock tests are genuinely good — asserting element identity/role rather than just class tokens is exactly what catches the class of defect you're fixing, and the inline comments explaining the load-bearing `peer` classes are a nice touch. No security, privacy, or correctness defects.

**Issues**

🔴 Blocking: 0
🟡 Minor: 1
- Specs lock `peer` but not the `absolute inset-0` that actually makes the stretched link cover the row — the one load-bearing token pair left unguarded against a future "drop unused class" pass.

⚪ Nit: 0

❔ Question: 1
- Row name loses `org-groups`' `peer-hover:text-blue-700` hover color; the inline comment justifies dropping only the `pointer-events-none` half of the reference wrapper, not this cosmetic half. Intentional?

**Bot reconciliation:** Cursor Bot flagged this Low Risk (template-only, no backend/auth impact) — agreed. CodeRabbit auto-review is disabled on this repo, so no findings to reconcile there.

**Decision: ✅ Approved with minor comments.** The minor test-coverage gap and the open question are both worth a quick follow-up but neither blocks. Nice work.

dealako (approved):
Approved with minor comments. Correct, tightly scoped WCAG 4.1.2 fix — faithful port of the org-groups (#1786) stretched-link pattern with strong regression-lock tests. One minor test-coverage gap (unlocked `absolute inset-0`) and one open question (dropped hover text-color) noted inline; neither blocks. See summary comment for details.

Issue #1792, opened 2026-08-23
fix(committees): member counts render as 0 — total_members is never populated upstream


manishdixitlfx:
## Summary

Committee member counts across the app are almost certainly rendering as `0` (or being hidden) because `total_members` is never populated upstream, and the BFF coerces the missing value to `0` rather than treating it as unknown.

This is a regression, not a latent gap — the counts used to be correct.

## Root cause

**PR #1060** (`123af8520`, 2026-07-07, LFXV2-2120 — *perf(committees): eliminate N+1 upstream count calls in groups list*) removed the per-committee `getCommitteeMembersCount` lookup — a real `GET /query/resources/count` — and replaced it with the indexed field:

```diff
-    // Get member count and mailing list association for each committee in parallel
-          total_members: memberCount,
+    // total_members is already indexed by the committee-service as part of
+    // CommitteeBaseWithReadonlyAttributes — no separate per-committee count call needed.
+      total_members: committee.total_members ?? 0,
```

The performance goal was right — the N+1 was real. But the premise in that comment does not hold against current data: **`total_members` is not populated.**

Evidence (gathered 2026-08-23):

- 40 committee documents from the query service's committee search, spanning ~20 foundations (AAIF, SPDX, OpenChain, LF AI & Data, PyTorch, Open Horizon, React Foundation, AOMedia, RISC-V, openIDL, Overture, C2PA, Ultra Ethernet, Node.js, ToIP, AOUSD, App Defense Alliance, CCC, Akrites, Appia) — the field is omitted from every one.
- Three fetched directly by uid, bypassing search, to rule out a search projection:

| committee | project | `TotalMembers` |
|---|---|---|
| Compliance Working Group | Ultra Ethernet Consortium Fund | `null` |
| Generative AI Commons Work Group | LF AI & Data | `null` |
| Codec Working Group | Alliance for Open Media | `null` |

The field exists in the schema; it carries no value.

Because `CommitteeInfo.total_members` is declared non-optional (`packages/shared/src/interfaces/committee.interface.ts:376`) and the BFF applies `?? 0` (`apps/lfx-one/src/server/services/committee.service.ts:202`), nothing type-checks or logs this away — a missing count is indistinguishable from a genuine zero.

## Affected surfaces

| file | behaviour |
|---|---|
| `committee-dashboard.component.ts:193` | sums `total_members \|\| 0` across committees — the dashboard total renders 0 |
[...]

PR #1817, opened 2026-08-24
feat(dashboards): add csv export to org groups roster

### LINKED ISSUE #1781: feat(org-lens): add CSV export to the Groups page
Parent: #1776

**Problem.** Org Projects has "Export to CSV" (`org-projects.component.html:98`, with a disabled-state tooltip when there are no rows). Groups does not. Same persona, same reporting need — group participation is exactly what an ED pulls into a board deck.

CSV export was raised as an open question on [LFXV2-1880](https://linuxfoundation.atlassian.net/browse/LFXV2-1880) and never resolved, because the spec was discarded.

**Fix.** Same button, same placement and same disabled-state treatment as Org Projects.

**Columns:** group name, foundation, behavioral class, org seat count, total seats, group UID.

## Acceptance

- Export honours the active filters (#1778); if filters have not shipped yet, it exports the full roster.
- Button is disabled with an explanatory tooltip when there are no rows.
- Filename and column headers follow the Org Projects export convention.

[LFXV2-1880]: https://linuxfoundation.atlassian.net/browse/LFXV2-1880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

manishdixitlfx:
## Summary

- Adds an "Export to CSV" button to `/org/groups`, mirroring the existing Org Projects export pattern (same `lfx-button` styling, same shared `downloadCsv` helper, same filename convention).
- Exports the currently filtered roster (not the full list) — columns: Group, Foundation, Type, Our Seats, Group UID. Type uses `BEHAVIORAL_CLASS_CONFIG[cls].label`, not the raw upstream category string.
- Extracts a new `localDateStamp()` shared util (`packages/shared/src/utils/date-time.utils.ts`) for local-timezone (not UTC) date stamping in export filenames, and migrates the sibling Org Projects export onto it, fixing a pre-existing bug there (`toISOString()` stamps tomorrow's date for any viewer west of UTC exporting in the evening).
- Disabled + "No rows to export" state is keyboard/screen-reader accessible (tooltip wrapper carries conditional `tabindex`/`role="note"`/`aria-label`, since a disabled button is removed from the tab order).

## Scope notes

- **Deliberately omits a "Total Seats" column.** The only per-row candidate field, `total_members`, is unpopulated upstream (#1792) and isn't even present on `OrgLensGroupSummary`. The response-level `total_seats` is an org-wide aggregate, not a per-row value, so it isn't a valid substitute.
- **`documentation-tab.component.ts`** has the same UTC-date filename bug this PR fixed in `org-projects.component.ts`, but is left untouched — out of scope for this ticket. Worth a follow-up now that `localDateStamp()` exists.

Closes GH-1781

## Test plan

- [x] `yarn lint` — clean
- [x] `yarn check-types` — clean
- [x] `yarn test` — 743/743 pass, including new coverage for `exportCsv()` (filtered-not-full export, disabled/tooltip state, `Type` column label mapping, empty-slug fallback, RFC 4180 quoting) and `localDateStamp()` (TZ-pinned local-vs-UTC regression test)
- [x] `yarn build` — succeeds
- [x] Manual verification: hover/mouse-event tracing confirmed via a minimal browser test that the tooltip mechanism behaves as documented in code comments

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.html
@@ -85,6 +85,36 @@
             ariaLabel="Filter by type"
             dataTest="org-groups-type-filter" />
         </div>
+        <!-- pTooltip lives on this wrapper, not the button: a disabled element is removed from the
+             tab order, so a keyboard user could never focus the button to see a tooltip bound to
+             it directly. tabindex + role="note" + aria-label (only while the tooltip has content)
+             give screen-reader users the same explanation a mouse hover gets — mirrors the
+             public-group badge in modules/committees/components/my-groups-card-grid/
+             my-groups-card-grid.component.html (same tabindex="0" + role="note" + aria-label
+             triplet, on a non-focusable host). A sighted keyboard-only user (no screen reader)
+             still gets no visual explanation: deliberately NOT tooltipEvent="both" — PrimeNG
+             resolves its focus target via `querySelector('.p-component')` inside this host, which
+             finds the inner (disabled, unfocusable) <p-button> root rather than this wrapper (the
+             element tabindex="0" actually puts in the tab order), so that listener would never
+             fire. -->
+        <div
+          class="flex items-center gap-2"
+          [pTooltip]="hasNoRowsToExport() ? noRowsToExportLabel : undefined"
+          tooltipPosition="top"
+          [attr.role]="hasNoRowsToExport() ? 'note' : null"
+          [attr.tabindex]="hasNoRowsToExport() ? '0' : null"
+          [attr.aria-label]="hasNoRowsToExport() ? noRowsToExportLabel : null"
+          data-testid="org-groups-export-csv-wrapper">
+          <lfx-button
+            label="Export to CSV"
+            icon="fa-light fa-file-arrow-down"
+            severity="secondary"
+            [outlined]="true"
+            size="small"
+            [disabled]="hasNoRowsToExport()"
+            (onClick)="exportCsv()"
+            data-testid="org-groups-export-csv" />
+        </div>
       </div>
     }
 
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.html
@@ -115,6 +115,15 @@
             (onClick)="exportCsv()"
             data-testid="org-groups-export-csv" />
         </div>
+        <!-- Always-visible fallback for the tooltip: closes the gap for a sighted keyboard-only
+             user (no screen reader), who would otherwise get no explanation at all — see the
+             comment above for why the tooltip itself can't be made to show on keyboard focus.
+             aria-hidden because the wrapper's aria-label already carries this to screen readers;
+             without it, a screen reader landing on the wrapper's role="note" region could read the
+             explanation twice. -->
+        @if (hasNoRowsToExport()) {
+          <span class="text-xs text-gray-500" aria-hidden="true" data-testid="org-groups-export-csv-hint">{{ noRowsToExportLabel }}</span>
+        }
       </div>
     }
 
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-projects/org-projects.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-projects/org-projects.component.ts
@@ -47,7 +47,7 @@ import type {
   OrgProjectsWorkspaceId,
   SortDirection,
 } from '@lfx-one/shared/interfaces';
-import { buildInsightsUrl, downloadCsv } from '@lfx-one/shared/utils';
+import { buildInsightsUrl, downloadCsv, localDateStamp } from '@lfx-one/shared/utils';
 import { MenuItem, MessageService } from 'primeng/api';
 import { DialogModule } from 'primeng/dialog';
 import { PopoverModule } from 'primeng/popover';
@@ -544,9 +544,8 @@ export class OrgProjectsComponent {
         p.participants.length,
       ];
     });
-    const date = new Date().toISOString().slice(0, 10).replace(/-/g, '');
     const slug = this.response()?.orgSlug ?? 'org';
-    downloadCsv(`org-lens-projects-${slug}-${date}.csv`, [header, ...body]);
+    downloadCsv(`org-lens-projects-${slug}-${localDateStamp()}.csv`, [header, ...body]);
   }
 
   // Signal-strength bars for an influence band: filled count = rank (Leading 4 → Silent 1 → Non-LF 0),
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.spec.ts
@@ -1012,6 +1012,10 @@ describe('OrgGroupsComponent — CSV export', () => {
     return el as HTMLElement;
   }
 
+  function exportHint(fixture: ComponentFixture<OrgGroupsComponent>): HTMLElement | null {
+    return fixture.nativeElement.querySelector('[data-testid="org-groups-export-csv-hint"]');
+  }
+
   it('enables the export button with no tooltip once there are rows to export', async () => {
     const { fixture } = await render({ getGroups: () => of(groupsResponse(buildGroups())) });
 
@@ -1022,6 +1026,7 @@ describe('OrgGroupsComponent — CSV export', () => {
     const wrapper = exportTooltipWrapper(fixture);
     expect(wrapper.hasAttribute('tabindex')).toBe(false);
     expect(wrapper.hasAttribute('role')).toBe(false);
+    expect(exportHint(fixture)).toBeNull();
   });
 
   it('disables the export button and shows "No rows to export" when the active filter matches nothing', async () => {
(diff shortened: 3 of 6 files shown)


apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.html
```diff
@@ -85,6 +85,36 @@ <h2 class="text-base font-semibold text-gray-900" data-testid="org-groups-no-acc
...
+             finds the inner (disabled, unfocusable) <p-button> root rather than this wrapper (the
+             element tabindex="0" actually puts in the tab order), so that listener would never
+             fire. -->
+        <div
+          class="flex items-center gap-2"
+          [pTooltip]="hasNoRowsToExport() ? noRowsToExportLabel : undefined"
+          tooltipPosition="top"
+          [attr.role]="hasNoRowsToExport() ? 'note' : null"
+          [attr.tabindex]="hasNoRowsToExport() ? '0' : null"
+          [attr.aria-label]="hasNoRowsToExport() ? noRowsToExportLabel : null"
```
dealako, line 106: [minor] Sighted keyboard-only users get no visible disabled reason

Issue: The `pTooltip` is hover-only, and your own comment notes that focusing this wrapper won't display it (PrimeNG resolves the focus target to the inner disabled button). So a sighted keyboard user who tabs to the wrapper gets a tab stop but never sees "No rows to export." Screen-reader users are covered by `aria-label`; sighted-keyboard-only users are the gap.
Proof: `tooltipEvent` is left at the PrimeNG hover default and the template comment explicitly states the focus listener "would never fire" on this host. Agrees with Copilot's review comment on this line.
Why it matters: The disabled state's explanation is reachable by mouse and by screen reader, but not by keyboard-without-SR.
Fix: Optional — render the reason as visible helper text next to the disabled button, or adopt a focus-openable popover. Given you documented this as a deliberate trade-off, leaving it is defensible; flagging so the decision is explicit in review.

apps/lfx-one/src/app/modules/dashboards/org/org-projects/org-projects.component.ts
```diff
@@ -544,9 +544,8 @@ export class OrgProjectsComponent {
         p.participants.length,
       ];
     });
-    const date = new Date().toISOString().slice(0, 10).replace(/-/g, '');
     const slug = this.response()?.orgSlug ?? 'org';
-    downloadCsv(`org-lens-projects-${slug}-${date}.csv`, [header, ...body]);
+    downloadCsv(`org-lens-projects-${slug}-${localDateStamp()}.csv`, [header, ...body]);
```
dealako, line 548: [nit] org-projects filename migration has no component-level test

Issue: This line switches org-projects' export filename from inline `toISOString()` UTC stamping to `localDateStamp()` (UTC→local calendar date). org-groups got full filename coverage, but org-projects.component.spec.ts asserts nothing on the export filename.
Proof: grep of org-projects.component.spec.ts for export/download/csv/filename returns no matches; the behavior change rests solely on `localDateStamp()`'s own unit tests.
Why it matters: A future edit to this filename template wouldn't be caught by an org-projects component test.
Fix: Optional — `localDateStamp()` is independently well-tested, so this is low priority. Mirror the org-groups "names the download with the org slug and today's date" assertion if you want symmetric coverage, or accept as a trade-off.

dealako (dismissed):
Hi @manishdixitlfx — clean, tightly-scoped PR. The CSV export mirrors the existing Org Projects pattern faithfully, the `localDateStamp()` extraction is a nice cleanup that fixes a real west-of-UTC bug, and the test coverage is genuinely thorough (filtered-not-full export, RFC 4180 quoting, TZ-pinned local-date regression, empty-slug fallback). The accessibility rationale for putting `pTooltip` on the wrapper instead of the disabled button is well-reasoned and documented.

No blocking, security, or privacy issues. The export carries only org entities and seat counts — no PII — and `escapeCsvCell` already neutralizes CSV formula injection. Two small items below, neither blocking.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 1 — sighted keyboard-only users get no visible disabled-reason (agrees with Copilot)
- ⚪ Nit: 1 — org-projects filename migration lacks a component-level test
- ❔ Question: 0

### Bot reconciliation
- **Copilot** (a11y): agreed, captured below as the one minor.
- **Cursor**: agreed with the Low Risk assessment; no actionable findings.
- **CodeRabbit**: auto-review skipped on this repo.

✅ **Approved with minor comments** — the two items are take-it-or-leave-it; the a11y gap is a reasonable documented trade-off.

dealako:
Hi @manishdixitlfx 👋 — clean, tightly-scoped PR. The CSV export mirrors the existing Org Projects pattern faithfully, the `localDateStamp()` extraction is a nice cleanup that fixes a real west-of-UTC filename bug, and the test coverage is genuinely thorough: filtered-not-full export, `Type` label mapping, RFC 4180 quoting, empty-slug fallback, and a TZ-pinned local-vs-UTC regression test. The accessibility rationale for binding `pTooltip` to the wrapper instead of the disabled button is well-reasoned and documented in-code.

No blocking, security, or privacy issues. The export carries only org entities and seat counts (no PII), and `escapeCsvCell` already neutralizes CSV formula injection.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 1 — sighted keyboard-only users get no visible disabled reason (agrees with Copilot)
- ⚪ Nit: 1 — org-projects filename migration lacks a component-level test
- ❔ Question: 0

### Bot reconciliation
- **Copilot** (a11y, html:106): agreed — captured as the one minor. The tooltip is hover-only; screen-reader users are covered by `aria-label`, but sighted keyboard-only users don't see the reason. You documented this as a deliberate trade-off, so it's defensible.
- **Cursor**: agreed with the Low Risk assessment; no actionable findings.
- **CodeRabbit**: auto-review skipped on this repo.

### Final decision
✅ **Approved with minor comments** — both items are take-it-or-leave-it. Nice work.

apps/lfx-one/src/app/modules/dashboards/org/org-groups/org-groups.component.html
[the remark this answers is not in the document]
manishdixitlfx: Good catch — you're right that tabindex/role/aria-label alone left a sighted keyboard-only user with no visual explanation, since PrimeNG's tooltip focus listener can't be made to target this wrapper (it resolves via `querySelector('.p-component')`, which lands on the inner disabled `<p-button>`, not this element — traced against the vendored PrimeNG source).

Fixed by rendering the disabled reason as always-visible text next to the button (`aria-hidden` since the wrapper's `aria-label` already carries the same text to screen readers, avoiding a double announcement).

See commit 115af4340: adds a conditionally-rendered `<span>` hint in org-groups.component.html, plus spec coverage asserting it's absent when rows exist and present with the right text/aria-hidden when disabled.

manishdixitlfx:
## Review Feedback Addressed

Commit: 115af434005028817afb2899ab048654efdd4abe

### Changes Made
- **org-groups.component.html**: added an always-visible text hint next to the disabled Export to CSV button, shown only when there are no rows to export (`aria-hidden`, since the wrapper's `aria-label` already carries the explanation to screen readers). Closes the accessibility gap where a sighted keyboard-only user got no explanation at all — PrimeNG's tooltip can't be made to fire on keyboard focus for this element (traced against the vendored source: its focus listener resolves to the disabled inner `<p-button>`, not this wrapper). (per copilot-pull-request-reviewer[bot])
- **org-groups.component.spec.ts**: added coverage asserting the hint is absent when rows exist and present (correct text + `aria-hidden`) when disabled.

### Threads Resolved

dealako:
Thanks for the quick turnaround, @manishdixitlfx — the follow-up commit (`115af434`) is exactly the right fix.

### 👏 Nice work
- The always-visible `@if (hasNoRowsToExport())` hint span closes the sighted-keyboard-only gap cleanly, and driving it off the same `hasNoRowsToExport()` signal that gates the button's `[disabled]` state means the visible text and the disabled state can't drift apart.
- `aria-hidden="true"` on the hint is the correct call — the wrapper's `aria-label` + `role="note"` already carries the string to screen readers, so this avoids a double announcement. Your inline reply tracing PrimeNG's `querySelector('.p-component')` focus-target resolution was a great explanation of why the tooltip couldn't just be made focus-openable.
- Both new spec assertions are on point: null hint in the enabled state, text + `aria-hidden` in the disabled state.

### Revision tracking
- ✅ **Resolved** — *Sighted keyboard-only users get no visible disabled reason* (minor, round 1). Fixed by the always-visible hint in `115af434`.
- ❌ **Still open** — *org-projects filename migration has no component-level test* (nit, round 1). The new commit only touched org-groups files, so `org-projects.component.spec.ts` still asserts nothing on the export filename. Take-it-or-leave-it; `localDateStamp()` remains independently well-tested.

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 1 — org-projects export-filename test still uncovered (carried over)

### Bot reconciliation
- **Copilot** (new review, no new inline comments): its round-1 a11y point is now addressed by the visible hint. Agreed and resolved.

### Final decision
✅ **Approved with minor comments** — the one remaining item is a take-it-or-leave-it nit. Nice, clean revision.

dealako (approved):
Follow-up review of commit 115af434. The round-1 minor a11y finding (sighted keyboard-only users had no visible disabled reason) is resolved by the always-visible aria-hidden hint span, correctly driven off the same hasNoRowsToExport() signal as the button's disabled state. No new findings across security/privacy, correctness/performance/tests, or style/docs. One carried-over nit remains (org-projects export-filename component test), which is take-it-or-leave-it. Approving.

Issue #1731, opened 2026-08-20
Meeting join page is slow 


niravpatel27:
We've had reports that the meeting join page is slow (e.g. the AAIF Governing Board Meeting page, which is private/restricted/recurring), and since we're planning to keep building on this page, it's worth fixing properly before more gets layered on top.

I dug into it with curl reproduction, Datadog trace dissection, and real-user (RUM) data. One methodology note: my first pass used Datadog's span-index aggregates, which turned out to be inflated by retention bias, so all the numbers below are re-checked against the unbiased trace metrics instead (computed from 100% of requests, not a sample).

Real-user numbers for meeting pages over the last 7 days: p75 page load 10.7s, p90 32s, p75 LCP 9.5s. That's rough.

Ranked by impact, here's what's causing it:

1. The real root cause is that every access-check call to NATS is slow. Any time the backend needs to check if a user can see something (registrants, RSVPs, etc.), it calls `lfx.access_check.request` over NATS, which goes through the fga-sync/OpenFGA path. This call alone runs ~3.2–3.5s p95 all day, and spikes to 8–10s during meeting bursts at the top of the hour. I traced one request where 3.94 of 3.96 seconds was just sitting in this queue — the actual work took 3.3ms. This matches the known OpenFGA throttling issue (GH-1653). I also confirmed the correlation directly: every time this NATS call spikes, the registrant endpoints spike right along with it (NATS p95 hit 9.8s → registrants p95 hit 65.8s in the same 10-minute window, and this repeated several times over two days).

2. The registrants endpoint makes it way worse by piling up calls one after another. `GET /api/meetings/:uid/my-meeting-registrants` normally runs around 3s p95 but spikes to 49–66s. It does everything serially — meeting lookup, then a token, then a registrant check, then registrants, then RSVPs — and it pages through all registrant pages and all RSVP pages one at a time instead of in parallel. One trace showed 23 sequential calls for a single request. Each of those can hit the slow NATS check above, so the delays stack up fast. This runs every time an organizer or invited user opens the join page.

3. The public meeting endpoint has the same problem on a smaller scale. `GET /public/api/meetings/:id` usually runs ~1s but spikes to 29s for logged-in users, because it also does about 10 things one after another — including fetching the entire registrant roster just to show two count numbers in the header.

[...]

PR #1820, opened 2026-08-24
perf(meetings): parallelize registrant fan-out, dedupe roster walk (#1731)

bramwelt:
## Summary

Rescoped after #1870 merged and landed a leaner version of most of this PR's original scope (parallelized `getMeetingById` on the public/authenticated meeting routes, and the `ApiRequestOptions.bearerToken` mechanism replacing `req.bearerToken` mutation). This PR now carries only the work still genuinely missing from `main`:

1. `getMyMeetingRegistrants` still swapped `req.bearerToken` for its M2M fan-out — reworked to use `ApiRequestOptions.bearerToken` instead, and to run the meeting fetch and registrant gate check in parallel.
2. The registrant roster was walked twice per request (once for RSVP filtering, once for the response) — deduped into a single roster fetch.
3. `page_size` was never sent on the query-service page walks, so every walk ran at the upstream default (50) instead of the documented ceiling (1000).

All three are covered by new tests, which the original branch (superseded by #1870) did not have.

Refs #1731

### `getMyMeetingRegistrants`: M2M fan-out without mutating `req.bearerToken`

- Hop 1 (`Promise.all`): `getMeetingById` runs alongside the registrant gate check (`generateM2MToken(req).then(token => getMeetingRegistrantsByEmail(...))`). The gate check depends only on `uid` + caller identity, not the meeting payload, so it doesn't need to wait on the meeting fetch.
- Hop 2: `getMeetingRegistrants` runs under the M2M token, passed via `ApiRequestOptions.bearerToken` rather than a `req.bearerToken` swap — so it can't race the caller's own token on the same `req`.
- Added an identity-race regression test: asserts the M2M and user-token calls never cross identities under out-of-order promise settlement.

The gate check stays on `getMeetingRegistrantsByEmail` (email-only), matching the pre-existing authorization surface — this PR is a perf change only and does not widen who can pass the registrant gate.

**`enrichCommitteeRegistrants` stays on a narrow `req.bearerToken` swap** — it's the one sequential (non-fanned-out) M2M call left in this handler, so it carries none of the race hazard the parallel fan-out did. The swap is now wrapped in `try/finally` so a thrown error doesn't leave the M2M token on `req.bearerToken` for the rest of the request. Fully re-plumbing `CommitteeService` to accept `ApiRequestOptions.bearerToken` would also need to exclude `addAccessToResource`/`getCallerMembership` (both must stay on the caller's own token) — tracked separately in #1903.

### Registrant roster deduped

- `getMeetingRegistrants` (with `includeRsvp: true`) previously called `getMeetingRsvps`, which did its own independent roster walk to build the active-registrant filter set — the roster was walked twice.
- Split into `getRawMeetingRsvps` (RSVP page walk only) + two pure helpers in the new `helpers/meeting-rsvp.helper.ts`: `filterRsvpsToActiveRegistrants` and `attachRsvpsToRegistrants` (carries the existing per-occurrence RSVP resolution verbatim — load-bearing for LFXV2-2864, regression-tested).
- `getMeetingRegistrants` now fetches the roster once and the raw RSVPs in parallel, then filters/attaches locally.
- `getMeetingRsvps` keeps its existing signature and behavior (including its `registrantsFetchFailed` → unfiltered-RSVPs fallback) for its other, unrelated callers.

### `QUERY_SERVICE_MAX_PAGE_SIZE` applied

Added `QUERY_SERVICE_MAX_PAGE_SIZE = 1000` to `packages/shared/src/constants/api.constants.ts` (verified against the upstream `lfx-v2-query-service` OpenAPI spec: `page_size` max is 1000). Applied at the three call sites that page through the query service for this handler: the roster walk in `getMeetingRegistrants`, the RSVP walk in `getRawMeetingRsvps`, and the roster walk inside `getMeetingRsvps`.

## Superseded by #1870 (dropped from this PR's original scope)

The original five-commit branch also parallelized `public-meeting.controller.ts::getMeetingById` and threaded `m2mToken` as positional parameters through several service/helper methods. #1870 already landed equivalent (and in some cases leaner) coverage for the public route using the `ApiRequestOptions.bearerToken` mechanism `main` now has, so none of that is reintroduced here — this PR builds on top of `main`'s mechanism rather than adding a second, competing one.

## Notes

- **Branch naming:** this branch predates the `type/issue-NNN` convention documented in `.claude/rules/commit-workflow.md` (it's `perf/GH-1731-parallel`, not `perf/issue-1731`). Renaming isn't practical for an already-open PR with an existing review thread, so documenting the deviation here rather than renaming mid-flight.
- **Related issues:** #1903 (`enrichCommitteeRegistrants` M2M threading follow-up).

## Test Plan

- `meeting.controller.spec.ts`: `getMyMeetingRegistrants` — organizer-not-registrant gate, non-registrant/non-organizer short-circuit, no-email short-circuit, meeting-not-found short-circuit, identity-race regression, and the `enrichCommitteeRegistrants`-throws token-restore regression.
- `meeting.service.spec.ts`: roster fetched exactly once when `includeRsvp: true`; `page_size` present on all three walks; `getMeetingRsvps` behavior unchanged for existing callers including the unfiltered-fallback path.
- `meeting-rsvp.helper.spec.ts`: `filterRsvpsToActiveRegistrants` and `attachRsvpsToRegistrants`, including the LFXV2-2864 occurrence-scoping case.
- `yarn check-types`, `yarn lint:check`, and the full affected-spec suite (60 tests) pass locally.

### DIFF
--- a/apps/lfx-one/src/server/controllers/meeting.controller.ts
+++ b/apps/lfx-one/src/server/controllers/meeting.controller.ts
@@ -36,7 +36,7 @@ import { logger } from '../services/logger.service';
 import { MeetingService } from '../services/meeting.service';
 import { NatsService } from '../services/nats.service';
 import { UserService } from '../services/user.service';
-import { getEffectiveEmail, getUsernameFromAuth } from '../utils/auth-helper';
+import { getEffectiveEmail, getEffectiveUsername } from '../utils/auth-helper';
 import { generateM2MToken } from '../utils/m2m-token.util';
 
 /**
@@ -474,7 +474,7 @@ export class MeetingController {
       // with a different email now passes, widening the gate slightly from the previous
       // email-only check (getMeetingRegistrantsByEmail).
       const userEmail = getEffectiveEmail(req) ?? undefined;
-      const username = (await getUsernameFromAuth(req)) ?? undefined;
+      const username = getEffectiveUsername(req) ?? undefined;
 
       logger.debug(req, 'get_my_meeting_registrants', 'Checking user authentication', {
         meeting_id: uid,
@@ -563,11 +563,15 @@ export class MeetingController {
       // caller's own token — tracked separately in #1903.
       const originalToken = req.bearerToken;
       req.bearerToken = m2mToken;
-      const enrichedRegistrants = await this.enrichCommitteeRegistrants(req, meeting, registrants);
-      if (originalToken !== undefined) {
-        req.bearerToken = originalToken;
-      } else {
-        delete req.bearerToken;
+      let enrichedRegistrants: MeetingRegistrant[];
+      try {
+        enrichedRegistrants = await this.enrichCommitteeRegistrants(req, meeting, registrants);
+      } finally {
+        if (originalToken !== undefined) {
+          req.bearerToken = originalToken;
+        } else {
+          delete req.bearerToken;
+        }
       }
 
       logger.success(req, 'get_my_meeting_registrants', startTime, {
--- a/apps/lfx-one/src/server/controllers/meeting.controller.ts
+++ b/apps/lfx-one/src/server/controllers/meeting.controller.ts
@@ -473,6 +473,10 @@ export class MeetingController {
       // email OR username (getMeetingRegistrantsForUser) — a registrant added under a username
       // with a different email now passes, widening the gate slightly from the previous
       // email-only check (getMeetingRegistrantsByEmail).
+      //
+      // getEffectiveUsername (not getUsernameFromAuth) deliberately, so this access-control check
+      // honors isImpersonating on the Authelia auth path too. getUsernameFromAuth's other call
+      // sites still have this gap — tracked separately in #1908.
       const userEmail = getEffectiveEmail(req) ?? undefined;
       const username = getEffectiveUsername(req) ?? undefined;
 
--- a/apps/lfx-one/src/server/controllers/meeting.controller.ts
+++ b/apps/lfx-one/src/server/controllers/meeting.controller.ts
@@ -36,7 +36,7 @@ import { logger } from '../services/logger.service';
 import { MeetingService } from '../services/meeting.service';
 import { NatsService } from '../services/nats.service';
 import { UserService } from '../services/user.service';
-import { getEffectiveEmail, getEffectiveUsername } from '../utils/auth-helper';
+import { getEffectiveEmail } from '../utils/auth-helper';
 import { generateM2MToken } from '../utils/m2m-token.util';
 
 /**
@@ -469,25 +469,17 @@ export class MeetingController {
         return;
       }
 
-      // Step 1: Resolve the caller's identity up front. The registrant gate check below accepts
-      // email OR username (getMeetingRegistrantsForUser) — a registrant added under a username
-      // with a different email now passes, widening the gate slightly from the previous
-      // email-only check (getMeetingRegistrantsByEmail).
-      //
-      // getEffectiveUsername (not getUsernameFromAuth) deliberately, so this access-control check
-      // honors isImpersonating on the Authelia auth path too. getUsernameFromAuth's other call
-      // sites still have this gap — tracked separately in #1908.
+      // Step 1: Resolve the caller's identity up front. The registrant gate check below is
+      // email-only (getMeetingRegistrantsByEmail), matching the pre-existing authorization surface
+      // — this PR does not widen the gate to also match by username.
       const userEmail = getEffectiveEmail(req) ?? undefined;
-      const username = getEffectiveUsername(req) ?? undefined;
 
       logger.debug(req, 'get_my_meeting_registrants', 'Checking user authentication', {
         meeting_id: uid,
         has_email: !!userEmail,
-        has_username: !!username,
-        user_email: userEmail,
       });
 
-      if (!userEmail && !username) {
+      if (!userEmail) {
         logger.success(req, 'get_my_meeting_registrants', startTime, {
           meeting_id: uid,
           no_email: true,
@@ -507,7 +499,7 @@ export class MeetingController {
         this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: true }),
         generateM2MToken(req).then((token) => {
           m2mToken = token;
-          return this.meetingService.getMeetingRegistrantsForUser(req, uid, userEmail, username, token);
+          return this.meetingService.getMeetingRegistrantsByEmail(req, uid, userEmail, token);
         }),
       ]);
 
--- a/apps/lfx-one/src/server/controllers/meeting.controller.ts
+++ b/apps/lfx-one/src/server/controllers/meeting.controller.ts
@@ -36,7 +36,7 @@ import { logger } from '../services/logger.service';
 import { MeetingService } from '../services/meeting.service';
 import { NatsService } from '../services/nats.service';
 import { UserService } from '../services/user.service';
-import { getEffectiveEmail } from '../utils/auth-helper';
+import { getEffectiveEmail, getUsernameFromAuth } from '../utils/auth-helper';
 import { generateM2MToken } from '../utils/m2m-token.util';
 
 /**
@@ -469,9 +469,44 @@ export class MeetingController {
         return;
       }
 
-      // Step 1: Get the meeting with access check to determine organizer status
-      logger.debug(req, 'get_my_meeting_registrants', 'Fetching meeting details with access check', { meeting_id: uid });
-      const meeting = await this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: true });
+      // Step 1: Resolve the caller's identity up front. The registrant gate check below accepts
+      // email OR username (getMeetingRegistrantsForUser) — a registrant added under a username
+      // with a different email now passes, widening the gate slightly from the previous
+      // email-only check (getMeetingRegistrantsByEmail).
+      const userEmail = getEffectiveEmail(req) ?? undefined;
+      const username = (await getUsernameFromAuth(req)) ?? undefined;
+
+      logger.debug(req, 'get_my_meeting_registrants', 'Checking user authentication', {
+        meeting_id: uid,
+        has_email: !!userEmail,
+        has_username: !!username,
+        user_email: userEmail,
+      });
+
+      if (!userEmail && !username) {
+        logger.success(req, 'get_my_meeting_registrants', startTime, {
+          meeting_id: uid,
+          no_email: true,
+          registrant_count: 0,
+        });
+        res.json([]);
+        return;
+      }
+
+      // Step 2: Fetch the meeting and run the registrant gate check in parallel — the gate check
+      // depends only on uid + caller identity, not on the meeting payload, so it needn't wait for
+      // the meeting fetch. The gate check runs under an M2M token (captured via `m2mToken`, not by
+      // mutating req.bearerToken) so it isn't gated by the caller's own FGA relations.
+      logger.debug(req, 'get_my_meeting_registrants', 'Fetching meeting and registrant gate check in parallel', { meeting_id: uid });
+      let m2mToken = '';
+      const [meeting, userRegistrantCheck] = await Promise.all([
+        this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: true }),
+        generateM2MToken(req).then((token) => {
+          m2mToken = token;
+          return this.meetingService.getMeetingRegistrantsForUser(req, uid, userEmail, username, token);
+        }),
+      ]);
+
       if (!meeting) {
         logger.success(req, 'get_my_meeting_registrants', startTime, {
           meeting_id: uid,
@@ -482,106 +517,53 @@ export class MeetingController {
         return;
       }
 
-      // TODO: Reimplement show_meeting_attendees check
-      // // All ITX meetings are treated as having show_meeting_attendees enabled
-      // const showMeetingAttendees = meeting.show_meeting_attendees ?? false;
-
-      // logger.debug(req, 'get_my_meeting_registrants', 'Meeting found, checking show_meeting_attendees', {
-      //   meeting_id: uid,
-      //   show_meeting_attendees: showMeetingAttendees,
-      // });
-
-      // // Step 2-4: Check if show_meeting_attendees is enabled
-      // if (!showMeetingAttendees) {
-      //   logger.success(req, 'get_my_meeting_registrants', startTime, {
-      //     meeting_id: uid,
-      //     show_meeting_attendees: false,
-      //     registrant_count: 0,
-      //   });
-      //   res.json([]);
-      //   return;
-      // }
-
-      // Step 5: Check if current user is a registrant (access control)
-      const userEmail = getEffectiveEmail(req) ?? undefined;
-
-      logger.debug(req, 'get_my_meeting_registrants', 'Checking user authentication', {
-        meeting_id: uid,
-        has_email: !!userEmail,
-        user_email: userEmail,
-      });
-
-      if (!userEmail) {
-        logger.success(req, 'get_my_meeting_registrants', startTime, {
-          meeting_id: uid,
-          no_email: true,
-          registrant_count: 0,
-        });
-        res.json([]);
-        return;
-      }
-
-      // Save original token and setup M2M token for privileged access
-      const originalToken = req.bearerToken;
-      const m2mToken = await generateM2MToken(req);
-      req.bearerToken = m2mToken;
-
-      // Use tags_all to filter by both meeting_id and email
-      logger.debug(req, 'get_my_meeting_registrants', 'Checking if user is a registrant', {
-        meeting_id: uid,
-        user_email: userEmail,
-      });
-      const userRegistrantCheck = await this.meetingService.getMeetingRegistrantsByEmail(req, uid, userEmail);
-
       logger.debug(req, 'get_my_meeting_registrants', 'User registrant check complete', {
         meeting_id: uid,
         user_email: userEmail,
         registrant_count: userRegistrantCheck.length,
       });
 
-      // Step 6: If user is not a registrant, check if they are an organizer
-      if (userRegistrantCheck.length === 0) {
-        if (!meeting.organizer) {
-          logger.success(req, 'get_my_meeting_registrants', startTime, {
-            meeting_id: uid,
-            user_email: userEmail,
-            is_registrant: false,
-            is_organizer: false,
-            registrant_count: 0,
-          });
-          res.json([]);
-          return;
-        }
-
-        logger.debug(req, 'get_my_meeting_registrants', 'User is not a registrant but is an organizer, granting access', {
+      // Step 3: If user is not a registrant, check if they are an organizer
+      if (userRegistrantCheck.length === 0 && !meeting.organizer) {
+        logger.success(req, 'get_my_meeting_registrants', startTime, {
           meeting_id: uid,
           user_email: userEmail,
+          is_registrant: false,
+          is_organizer: false,
+          registrant_count: 0,
         });
+        res.json([]);
+        return;
       }
 
-      // Step 7: User is a registrant or organizer, fetch all registrants using M2M token
+      // Step 4: User is a registrant or organizer, fetch all registrants using the M2M token —
+      // passed via ApiRequestOptions.bearerToken (not a req.bearerToken mutation) so it can't race
+      // against the caller's own token on a shared req.
       logger.debug(req, 'get_my_meeting_registrants', 'Fetching registrants with M2M token', {
         meeting_id: uid,
         user_email: userEmail,
         include_rsvp: includeRsvp,
-      });
-
-      logger.debug(req, 'get_my_meeting_registrants', 'M2M token generated, fetching all registrants', {
-        meeting_id: uid,
         has_m2m_token: !!m2mToken,
       });
 
-      const registrants = await this.meetingService.getMeetingRegistrants(req, uid, includeRsvp, occurrenceId);
+      const registrants = await this.meetingService.getMeetingRegistrants(req, uid, includeRsvp, occurrenceId, false, undefined, {
+        bearerToken: m2mToken,
+      });
 
       logger.debug(req, 'get_my_meeting_registrants', 'Fetched all registrants, enriching committee data', {
         meeting_id: uid,
         registrant_count: registrants.length,
       });
 
-      // Enrich committee registrant data with committee details and member info
+      // Enrich committee registrant data with committee details and member info. Kept on a
+      // narrowly-scoped req.bearerToken swap rather than threaded ApiRequestOptions: this call is
+      // sequential (nothing else is in flight on this req), so it carries none of the identity-race
+      // hazard the parallel fan-out above did. Fully re-plumbing CommitteeService's bearerToken
+      // would also need to exclude addAccessToResource/getCallerMembership, which must stay on the
+      // caller's own token — tracked separately in #1903.
+      const originalToken = req.bearerToken;
+      req.bearerToken = m2mToken;
       const enrichedRegistrants = await this.enrichCommitteeRegistrants(req, meeting, registrants);
-
-      // Restore original token (delete if it was undefined to avoid leaving M2M token)
       if (originalToken !== undefined) {
         req.bearerToken = originalToken;
       } else {
--- a/apps/lfx-one/src/server/services/meeting.service.spec.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.spec.ts
@@ -1,7 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import type { Meeting, MeetingRegistrant, MeetingUserInfo, QueryServiceResponse } from '@lfx-one/shared/interfaces';
+import type { Meeting, MeetingRegistrant, MeetingRsvp, MeetingUserInfo, QueryServiceResponse } from '@lfx-one/shared/interfaces';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 // This app's vitest config resolves plain Node modules only — the `@lfx-one/shared/*` tsconfig
@@ -21,6 +21,10 @@ vi.mock('@lfx-one/shared/utils', () => ({
   mapITXResponseToMeetingRsvp: vi.fn(),
   normalizeIndexedMeetingAiSummary: vi.fn(),
   selectPrimaryPastMeetingSummary: vi.fn(),
+  // getMeetingRsvps / getMeetingRegistrants(includeRsvp) delegate occurrence selection to the real
+  // resolver — stubbed here to the "most recent rsvp" since these tests cover roster/page-walk
+  // dedup, not the LFXV2-2864 occurrence-scoping logic (covered in meeting-rsvp.helper.spec.ts).
+  selectApplicableRsvp: vi.fn((_occurrenceId: string | undefined, rsvps: unknown[]) => rsvps[rsvps.length - 1] ?? null),
 }));
 vi.mock('./microservice-proxy.service', () => ({
   MicroserviceProxyService: class {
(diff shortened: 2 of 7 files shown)


dealako:
@bramwelt — thanks for the detailed PR write-up and for flagging the gate widening and impersonation follow-ups explicitly. The parallel hop structure and roster dedup direction look sound, and several earlier bot findings (circular import, committee-name M2M in `getMeetingById`, `QUERY_SERVICE_MAX_PAGE_SIZE` wiring, username-or-email gate) are addressed in the latest commit.

The branch still conflicts with `main` after #1870, and one RSVP filtering regression remains. Details below.

## Issue count

- 🔴 Blocking: 2 issues: `getMeetingRsvps` partial-pagination regression; merge conflicts with `main`
- 🟡 Minor: 4 issues: PR scope/description drift vs diff; documented `getEffectiveUsername` fix not applied; test plan claims specs not in the diff; `enrichCommitteeRegistrants` no longer runs committee reads under M2M
- ⚪ Nit: 2 issues: JSDoc `pageSize` default contradicts implementation; debug log omits resolved `page_size`
- ❔ Question: 1 item: intentional username-or-email gate widening — please confirm product/security sign-off

## Bot reconciliation

- **Agree** with Cursor Bugbot / Copilot on partial-roster RSVP filtering (`failOnPartial: false` at `meeting.service.ts:1562`). Still present on `d71d3d7`.
- **Resolved** since earlier bot rounds: lazy `MeetingService` in `meeting.helper.ts` fixes the circular import; `getMeetingById` passes `m2mToken` into `getCommitteeNameMap`; `getMyMeetingRegistrants` hop 2 uses `getRawMeetingRsvps` with explicit M2M; gate now calls `getMeetingRegistrantsForUser`.
- **Disagree** with stale Copilot claim that `QUERY_SERVICE_MAX_PAGE_SIZE` has no consumers — it is applied at roster/RSVP walk call sites (`meeting.service.ts:708`, `:1488`, `:1539`).

## Prior review follow-up

Earlier Cursor/Copilot threads on M2M token drops and email-only gate predate the latest rescoped handler rewrite. The remaining actionable thread is partial-pagination RSVP filtering (also raised by Bugbot on `d71d3d7`).

## Final decision

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1393,41 +1461,105 @@ export class MeetingService {
...
+              'GET',
+              {
+                ...registrantParams,
+                ...(pageToken && { page_token: pageToken }),
+              },
+              undefined,
+              headers
+            );
+          },
+          { failOnPartial: false }
```
dealako: [blocking] Partial registrant pagination silently drops RSVPs

Issue: Registrant paging inside `getMeetingRsvps` uses `failOnPartial: false`. When a later page fails, `fetchAllQueryResources` returns a truncated roster without throwing, `registrantsFetchFailed` stays false, and `filterRsvpsToActiveRegistrants` filters against that incomplete set.
Proof: `origin/main` used `{ failOnPartial: true }` here so pagination failure hit `.catch`, set `registrantsFetchFailed`, and returned unfiltered RSVPs. `query-service.helper.ts:112-126` shows `failOnPartial: false` breaks the loop and returns partial results without throwing.
Why it matters: RSVP endpoints can return incomplete lists with no error signal when pagination fails mid-walk.
Fix: Restore `{ failOnPartial: true }` on this registrant fetch and add a unit test where page 2 rejects → raw RSVP count is returned unchanged.

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
```diff
@@ -56,14 +56,14 @@ export class PublicMeetingController {
...
+      // Generate M2M token once for all M2M-identity operations
+      // req.bearerToken stays on the user's token throughout (or undefined for anonymous)
+      const m2mToken = await generateM2MToken(req);
+      const isAuthenticated = req.oidc?.isAuthenticated();
+      const userEmail = getEffectiveEmail(req) || '';
 
-      // Get the meeting by ID using M2M token (all meetings are now v1_meeting type)
-      let meeting = await this.fetchMeetingWithM2M(req, id, 'v1_meeting', m2mToken);
+      // Hop 1: Get the meeting by ID using M2M token (all meetings are now v1_meeting type)
+      let meeting = await this.meetingService.getMeetingById(req, id, 'v1_meeting', { access: false, m2mToken });
```
dealako: [blocking] Rebase required — conflicts with main / #1870

Issue: GitHub reports this PR as CONFLICTING. The diff still rewrites both join-page detail handlers while the PR body says that scope was dropped after #1870 merged.
Proof: PR mergeable state is CONFLICTING; branch introduces `m2mToken` positional args and custom `Authorization` headers while `origin/main` threads M2M via `ApiRequestOptions.bearerToken`.
Why it matters: Merging as-is risks reintroducing a competing token mechanism and duplicating handler refactors already landed differently on main.
Fix: Rebase onto current `main`, drop controller changes superseded by #1870, and thread M2M through `{ bearerToken }` on `ApiRequestOptions` per main's pattern.

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -456,130 +451,100 @@ export class MeetingController {
...
-      //     registrant_count: 0,
-      //   });
-      //   res.json([]);
-      //   return;
-      // }
-
-      // Step 5: Check if current user is a registrant (access control)
+      // Check if current user has email/identity
       const userEmail = getEffectiveEmail(req) ?? undefined;
+      const username = (await getUsernameFromAuth(req)) ?? undefined;
```
dealako: [minor] Documented impersonation fix not applied

Issue: PR body states the gate switched to `getEffectiveUsername`, but this call site still uses `getUsernameFromAuth`.
Proof: `meeting.controller.ts:456` — `const username = (await getUsernameFromAuth(req)) ?? undefined`.
Why it matters: Reviewers may approve based on an impersonation fix that is not present.
Fix: Switch to `getEffectiveUsername(req)` as documented, or update the PR description to note the gap remains (#1908).

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -456,130 +451,100 @@ export class MeetingController {
...
-      const enrichedRegistrants = await this.enrichCommitteeRegistrants(req, meeting, registrants);
-
-      // Restore original token (delete if it was undefined to avoid leaving M2M token)
-      if (originalToken !== undefined) {
-        req.bearerToken = originalToken;
-      } else {
-        delete req.bearerToken;
-      }
+      // Hop 3: Enrich committee registrant data with committee details and member info
+      enrichedRegistrants = await this.enrichCommitteeRegistrants(req, meeting, enrichedRegistrants);
```
dealako: [minor] Committee enrichment no longer uses M2M identity

Issue: PR body says `enrichCommitteeRegistrants` keeps a narrow `req.bearerToken` M2M swap, but hop 3 now calls committee service methods on the user's token only (no swap in the helper).
Proof: Pre-refactor flow ran enrichment while `req.bearerToken` was M2M; current hop 3 passes `req` unchanged into `enrichCommitteeRegistrants`.
Why it matters: Registrants without committee read access get silently unenriched committee rows.
Fix: Thread M2M explicitly (#1903) or document the accepted degradation.

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -456,130 +451,100 @@ export class MeetingController {
...
 
-      logger.debug(req, 'get_my_meeting_registrants', 'User registrant check complete', {
-        meeting_id: uid,
-        user_email: userEmail,
-        registrant_count: userRegistrantCheck.length,
-      });
+      const [meeting, isOrganizerCheckPassed, userRegistrants] = await Promise.all([
+        this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: false }),
+        this.accessCheckService.checkSingleAccess(req, { resource: 'v1_meeting', id: uid, access: 'organizer' }),
+        this.meetingService.getMeetingRegistrantsForUser(req, uid, userEmail, username, m2mToken),
```
dealako: [question] Username-or-email gate widening

Issue: Gate moved from email-only to `getMeetingRegistrantsForUser` (email OR username). PR flags this as intentional.
Proof: `:488` passes both `userEmail` and `username` under M2M.
Why it matters: Authorization-surface expansion beyond perf scope.
Fix: Please confirm explicit product/security approval, or retain email-only matching if not intended.

dealako (dismissed):
See summary comment for full context. Two blocking items: restore fail-closed RSVP filtering on partial registrant pagination, and rebase onto current main to resolve conflicts and align with #1870's ApiRequestOptions.bearerToken pattern.

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -508,17 +508,15 @@ export class MeetingController {
...
       const m2mToken = await generateM2MToken(req);
-      req.bearerToken = m2mToken;
 
       // Use tags_all to filter by both meeting_id and email
       logger.debug(req, 'get_my_meeting_registrants', 'Checking if user is a registrant', {
         meeting_id: uid,
         user_email: userEmail,
       });
-      const userRegistrantCheck = await this.meetingService.getMeetingRegistrantsByEmail(req, uid, userEmail);
+      const userRegistrantCheck = await this.meetingService.getMeetingRegistrantsByEmail(req, uid, userEmail, m2mToken);
```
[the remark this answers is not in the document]
bramwelt: This referred to an earlier revision of the branch. The registrant gate is intentionally email-only (`getMeetingRegistrantsByEmail`) as of commit 723a84753 — no username-based widening is implemented, and the PR description has been updated to reflect that.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -181,28 +182,32 @@ export class MeetingService {
...
+    const { access = true, includeProject = false, m2mToken } = options;
     logger.debug(req, 'get_meeting_by_id', 'Fetching meeting by ID', {
       meeting_id: meetingUid,
       type: meetingType,
     });
 
     // All meetings are now ITX-managed, use the ITX endpoint
-    const meeting = await this.microserviceProxy.proxyRequest<Meeting>(req, 'LFX_V2_SERVICE', `/itx/meetings/${meetingUid}`, 'GET');
+    const headers = m2mToken ? { Authorization: `Bearer ${m2mToken}` } : undefined;
+    const meeting = await this.microserviceProxy.proxyRequest<Meeting>(req, 'LFX_V2_SERVICE', `/itx/meetings/${meetingUid}`, 'GET', undefined, undefined, headers);
```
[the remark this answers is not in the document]
bramwelt: This referred to `public-meeting.controller.ts`, which is not touched by this PR (see `git diff --stat origin/main...HEAD` — only `meeting.controller.ts`, `meeting.service.ts`, `meeting-rsvp.helper.ts`, and `api.constants.ts` are changed). #1870 already covers the public route on `main`. No change needed here.

apps/lfx-one/src/server/controllers/meeting.controller.ts
bramwelt: Reverted to email-only matching per your question — thanks for flagging this. `getMyMeetingRegistrants` now calls `getMeetingRegistrantsByEmail` only (commit 723a84753); the email-or-username widening via `getMeetingRegistrantsForUser` has been removed, and the PR description no longer claims it. This PR is a perf change only now.

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -508,17 +508,15 @@ export class MeetingController {
         return;
       }
 
-      // Save original token and setup M2M token for privileged access
-      const originalToken = req.bearerToken;
+      // Generate M2M token for privileged access (req.bearerToken stays on user token)
       const m2mToken = await generateM2MToken(req);
```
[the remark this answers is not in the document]
bramwelt: Current code (commit 723a84753) does thread the M2M token through `enrichCommitteeRegistrants`: `req.bearerToken` is swapped to the M2M token immediately before the call and restored in a `finally` block (`meeting.controller.ts`, `getMyMeetingRegistrants`), so `getCommitteeById`/`getCommitteeMembers` do run under M2M. No change needed here.

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -456,130 +451,100 @@ export class MeetingController {
...
-      logger.debug(req, 'get_my_meeting_registrants', 'User registrant check complete', {
-        meeting_id: uid,
-        user_email: userEmail,
-        registrant_count: userRegistrantCheck.length,
-      });
+      const [meeting, isOrganizerCheckPassed, userRegistrants] = await Promise.all([
+        this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: false }),
+        this.accessCheckService.checkSingleAccess(req, { resource: 'v1_meeting', id: uid, access: 'organizer' }),
+        this.meetingService.getMeetingRegistrantsForUser(req, uid, userEmail, username, m2mToken),
+      ]);
```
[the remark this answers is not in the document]
bramwelt: Reverted — the registrant gate is now email-only (`getMeetingRegistrantsByEmail`), matching the pre-existing authorization surface. This PR is a perf change only; no username-based widening is implemented. See commit 723a84753 and the updated PR description.

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -133,9 +139,24 @@ export class MeetingController {
         return;
       }
 
+      // Hop 1: Fetch meeting without access check (access resolved in hop 2)
       // includeProject: enrich with project_slug/project_name/is_foundation so clients can
       // reconcile project context from the meeting payload itself.
-      const meeting = await this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: true, includeProject: true });
+      let meeting = await this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: false, includeProject: true });
```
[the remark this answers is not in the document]
bramwelt: This referred to `public-meeting.controller.ts` / `GET /api/meetings/:uid`, which this PR does not touch. The only handler modified is `getMyMeetingRegistrants` in `meeting.controller.ts`. No change needed here.

apps/lfx-one/src/server/controllers/meeting.controller.ts
bramwelt: Moot — the gate reverted to email-only (`getMeetingRegistrantsByEmail`), so there's no username lookup at this call site anymore (`getUsernameFromAuth`/`getEffectiveUsername` are both gone from `getMyMeetingRegistrants`). The impersonation gap at `getUsernameFromAuth`'s other call sites is unrelated to this PR and remains open as its own concern if you'd like to track it separately.

apps/lfx-one/src/server/controllers/meeting.controller.ts
bramwelt: Confirmed in the current code — `enrichCommitteeRegistrants` runs under a narrow `req.bearerToken` swap to the M2M token, wrapped in `try/finally` so the token is restored (or deleted) even if the call throws. This matches the PR description. No change needed here.

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
```diff
@@ -56,14 +56,13 @@ export class PublicMeetingController {
...
-      // Generate M2M token once for all operations
-      const m2mToken = await this.setupM2MToken(req);
+      // Generate M2M token once for all M2M-identity operations
+      // req.bearerToken stays on the user's token throughout (or undefined for anonymous)
+      const m2mToken = await generateM2MToken(req);
+      const isAuthenticated = req.oidc?.isAuthenticated();
 
       // Get the meeting by ID using M2M token (all meetings are now v1_meeting type)
-      let meeting = await this.fetchMeetingWithM2M(req, id, 'v1_meeting', m2mToken);
+      let meeting = await this.meetingService.getMeetingById(req, id, 'v1_meeting', { access: false, m2mToken });
```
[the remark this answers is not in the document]
bramwelt: Added — `meeting.controller.spec.ts` now has a `getMyMeetingRegistrants` describe block (commit 723a84753) including an identity-race regression test that asserts the M2M and user-token calls never cross identities under out-of-order promise settlement, plus coverage for the organizer gate, non-registrant short-circuit, no-email short-circuit, and meeting-not-found short-circuit.

apps/lfx-one/src/server/services/meeting.service.ts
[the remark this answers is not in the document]
bramwelt: Confirmed already correct in the current revision — the registrant walk inside `getMeetingRsvps` uses `{ failOnPartial: true }` (`meeting.service.ts`), so a page failure throws and is caught by the existing `.catch()`, returning the unfiltered RSVP set rather than filtering against a truncated roster. This is the same fix dealako flagged below; both are addressed by the same code.

apps/lfx-one/src/server/controllers/public-meeting.controller.ts
bramwelt: The PR was rescoped after #1870 merged — the branch was rebuilt from scratch on `origin/main` (5 commits, worktree `perf/GH-1820-rescope`, force-pushed onto this branch) and no longer touches `public-meeting.controller.ts` or reintroduces `m2mToken` positional args / custom `Authorization` headers on that route. It builds on `main`'s `ApiRequestOptions.bearerToken` mechanism. `gh pr view` now reports `mergeable: MERGEABLE`. See the updated PR description for the full rescope summary.

packages/shared/src/constants/api.constants.ts
```diff
@@ -30,6 +30,15 @@ export const MEETING_PASSWORD_HEADER = 'x-meeting-password';
...
 export const QUERY_SERVICE_FILTERS_OR_BATCH_SIZE = 100;
 
+/**
+ * Maximum page size for query-service paginated requests.
+ * @description Set to the upstream contract limit (lfx-v2-query-service/design/types.go).
+ * Min 1, max 1000, default 50. Using 1000 minimizes page count for roster fetches:
+ * 20× fewer pages for typical rosters, single page for any roster ≤1000 registrants.
+ * Only applied at call sites where reducing page count materially improves latency (roster walks).
+ */
+export const QUERY_SERVICE_MAX_PAGE_SIZE = 1000;
```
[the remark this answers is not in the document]
bramwelt: Fixed in the current revision — `QUERY_SERVICE_MAX_PAGE_SIZE` is passed as `page_size` at the three query-service walk sites in `meeting.service.ts` (roster walk in `getMeetingRegistrants`, RSVP walk in `getRawMeetingRsvps`, and the roster walk inside `getMeetingRsvps`). See the current diff against `origin/main`.

apps/lfx-one/src/server/services/meeting.service.ts
bramwelt: Fixed by reverting the gate to email-only matching rather than by threading M2M through pagination — same root cause you flagged. Registrant paging inside `getMeetingRsvps` already uses `{ failOnPartial: true }` on `main`'s pattern and is unchanged by this PR; confirmed in the current diff there is no `failOnPartial: false` introduced on that path. Added no new test since the behavior is untouched by this PR, but happy to add a regression test if you'd still like one for the existing code path.

bramwelt:
## Review Feedback Addressed

Commit: 723a84753

### Changes Made
- **apps/lfx-one/src/server/controllers/meeting.controller.ts**: reverted the `getMyMeetingRegistrants` registrant gate from the email-or-username lookup back to email-only matching (`getMeetingRegistrantsByEmail`), per @dealako's question about the authorization-surface widening. This PR is now a perf change only.
- **apps/lfx-one/src/server/controllers/meeting.controller.spec.ts**: updated tests for the email-only gate, removed the now-invalid username-fallback test, updated the identity-race regression assertion.
- PR description updated to drop the stale "gate widening (intentional)" and `getUsernameFromAuth`/`getEffectiveUsername` callouts.

### No Change Needed
- **meeting.controller.ts / meeting.service.ts**: enrichment already threads the M2M token through `enrichCommitteeRegistrants` via a narrow, `try/finally`-guarded `req.bearerToken` swap (per @dealako and copilot-pull-request-reviewer).
- **meeting.service.ts**: the registrant walk inside `getMeetingRsvps` already uses `{ failOnPartial: true }`, so a page failure is caught and falls back to the unfiltered RSVP set (per @dealako and copilot-pull-request-reviewer — same underlying code, both addressed together).
- **packages/shared/src/constants/api.constants.ts**: `QUERY_SERVICE_MAX_PAGE_SIZE` is consumed at all three query-service walk sites in `meeting.service.ts` (per copilot-pull-request-reviewer).
- **public-meeting.controller.ts**: this PR does not touch this file at all — the earlier scope-creep and merge-conflict concerns (per @dealako and copilot-pull-request-reviewer) referred to a prior revision of the branch, before the #1870 rescope. `mergeable` now reports `MERGEABLE`.
- **meeting.controller.spec.ts**: the identity-race regression test requested by copilot-pull-request-reviewer already exists on this branch.

### Threads Resolved

dealako:
@bramwelt — solid follow-up. The rescope onto `main`, the `failOnPartial: true` restore, and the new spec coverage address the blocking items from the prior round cleanly.

## 👏 Nice work

- **`failOnPartial: true` restored** in `getMeetingRsvps` with the fail-open unfiltered RSVP path preserved when the registrant walk throws.
- **Rebased and rescoped** — branch now merges cleanly and focuses on the registrant handler, roster dedup, and page-size work rather than reintroducing superseded join-page handler refactors.
- **Identity safety** — parallel gate/roster fetches thread M2M via `ApiRequestOptions.bearerToken` without mutating `req.bearerToken`; `enrichCommitteeRegistrants` uses a documented try/finally swap with a throw-path regression test.
- **Test coverage landed** — `meeting.controller.spec.ts` identity-race regression, `meeting-rsvp.helper.spec.ts` LFXV2-2864 occurrence scoping, and `page_size: 1000` assertions on the three main walks.

## Prior feedback status

- ✅ **Resolved:** `failOnPartial` RSVP regression (`723a8475`)
- ✅ **Resolved:** merge conflicts / rebase onto `main` (`ff51d499`)
- ✅ **Resolved:** gate reverted to email-only (no username widening)
- ✅ **Resolved:** committee enrichment M2M via try/finally swap (`6f79ccfe`, `44086773`)
- ✅ **Resolved:** missing controller/service/helper specs
- ⚠️ **Partially addressed:** mid-pagination RSVP regression test (first-page failure covered; page-2 failure not yet)

## Issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: gate-check walk still omits `QUERY_SERVICE_MAX_PAGE_SIZE`; no mid-pagination test for `getMeetingRsvps` partial-failure contract
- ⚪ Nit: 1 issue: `getMeetingRegistrantsByEmail` still uses positional `m2mToken` + custom headers while roster fetch uses `ApiRequestOptions.bearerToken` in the same handler

## Final decision

apps/lfx-one/src/server/controllers/meeting.controller.ts
```diff
@@ -521,71 +489,85 @@ export class MeetingController {
...
+      // depends only on uid + caller identity, not on the meeting payload, so it needn't wait for
+      // the meeting fetch. The gate check runs under an M2M token (captured via `m2mToken`, not by
+      // mutating req.bearerToken) so it isn't gated by the caller's own FGA relations.
+      logger.debug(req, 'get_my_meeting_registrants', 'Fetching meeting and registrant gate check in parallel', { meeting_id: uid });
+      let m2mToken = '';
+      const [meeting, userRegistrantCheck] = await Promise.all([
+        this.meetingService.getMeetingById(req, uid, 'v1_meeting', { access: true }),
+        generateM2MToken(req).then((token) => {
+          m2mToken = token;
+          return this.meetingService.getMeetingRegistrantsByEmail(req, uid, userEmail, token);
```
dealako, line 502: [minor] Gate-check walk omits QUERY_SERVICE_MAX_PAGE_SIZE

Issue: The parallel gate check calls `getMeetingRegistrantsByEmail`, which paginates without `page_size`, while roster/RSVP walks use 1000.
Proof: Gate at `:502`; `getMeetingRegistrantsByEmail` params include only type/filters (no `page_size`).
Why it matters: The gate walk can still page at the upstream default (50), leaving perf benefit incomplete on this path.
Fix: Add `page_size: QUERY_SERVICE_MAX_PAGE_SIZE` to `getMeetingRegistrantsByEmail` params, or narrow the PR description to the three walks already updated.

apps/lfx-one/src/server/services/meeting.service.spec.ts
```diff
@@ -430,6 +434,116 @@ describe('MeetingService.getMeetingRegistrants', () => {
...
+    expect(registrantQuery.page_size).toBe(1000);
+  });
+
+  it('returns RSVPs unfiltered when the registrant fetch fails, rather than hiding data', async () => {
+    proxyRequest.mockResolvedValueOnce({ resources: [rsvpRecord('a'), rsvpRecord('b')] }).mockRejectedValueOnce(new Error('query service down'));
+
+    const result = await service.getMeetingRsvps(req, 'meeting-1');
+
+    expect(result).toHaveLength(2);
+  });
```
dealako, line 573: [minor] Mid-pagination RSVP failure lacks regression test

Issue: `failOnPartial: true` is restored, but this spec only covers first-page registrant fetch failure.
Proof: `:541` uses `mockRejectedValueOnce` on the second proxy call (first registrant page); no case where page 1 returns `page_token` and page 2 rejects.
Why it matters: The original blocking bug was silent filtering against a truncated roster; behavior is fixed but the mid-walk scenario remains untested.
Fix: Add a spec: page 1 returns `page_token` + one registrant; page 2 rejects → expect both RSVPs returned unfiltered.

dealako (dismissed):
Prior blocking items are resolved. Two minor follow-ups noted inline (gate-check page_size, mid-pagination RSVP test). Nice work on the rescope and identity-race coverage.

apps/lfx-one/src/server/services/meeting.service.ts
bramwelt: Added the regression test in b686fd49c: `getMeetingRsvps` with a multi-page registrant walk where page 1 succeeds with a `page_token` and page 2 rejects — asserts the raw RSVP count (2) is returned unchanged rather than filtered against the truncated roster.

apps/lfx-one/src/server/controllers/meeting.controller.ts
bramwelt: Fixed, added `page_size: QUERY_SERVICE_MAX_PAGE_SIZE` to the gate-check params.

See commit 7722680a7: `getMeetingRegistrantsByEmail` now matches the other three walks.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -628,63 +629,65 @@ export class MeetingService {
...
-    maxResults?: number
+    maxResults?: number,
+    options?: ApiRequestOptions
   ): Promise<MeetingRegistrant[]> {
     // Registrant records carry `parent_refs: ['meeting:<uid>']` but no indexed tags — use `parent`
     // to query parent_refs, matching the working pattern in getMeetingRsvps.
     const params: Record<string, any> = {
       type: 'v1_meeting_registrant',
       parent: `meeting:${meetingUid}`,
+      page_size: QUERY_SERVICE_MAX_PAGE_SIZE,
```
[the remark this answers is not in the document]
bramwelt: Good catch, fixed.

See commit 7722680a7: `getMeetingRegistrants` now clamps `page_size` to `maxResults + 1` when `maxResults` is supplied, so `getAuthorizedRegistrantsForImport` requests 51 rows instead of 1000 to reject an over-cap roster. Added a test asserting `page_size: 51` on that path.

apps/lfx-one/src/server/services/meeting.service.spec.ts
bramwelt: This was already addressed, the requested test (page 1 returns `page_token`, page 2 rejects, raw RSVP count returned unchanged) was added in commit b686fd49c, which was pushed after this comment was posted. No further change needed here.

bramwelt:
## Review Feedback Addressed

Commit: 7722680a7

### Changes Made
- **meeting.service.ts**: added `page_size` to the gate-check registrant lookup (`getMeetingRegistrantsByEmail`), matching the other three walks (per @dealako)
- **meeting.service.ts**: clamped `page_size` to `maxResults + 1` in `getMeetingRegistrants` when `maxResults` is supplied, so `getAuthorizedRegistrantsForImport` requests 51 rows instead of 1000 to reject an over-cap roster (per copilot-pull-request-reviewer)
- **meeting.service.spec.ts**: added coverage for both fixes

### No Change Needed
- **meeting.service.spec.ts:546**: the requested page-2-rejects regression test was already added in commit b686fd49c, pushed after this comment was posted (asked by @dealako)

### Threads Resolved

dealako:
@bramwelt — thanks for the quick turnaround on the two minor items from the last round. Both are closed out cleanly.

## 👏 Nice work

- **`QUERY_SERVICE_MAX_PAGE_SIZE` on the gate check** — `getMeetingRegistrantsByEmail` now sends `page_size: 1000`, with a spec asserting it (`7722680a7`).
- **Mid-pagination RSVP regression test** — the page-1 `page_token` + page-2 reject case confirms unfiltered fallback behavior (`b686fd49c`).
- **Bounded `maxResults` clamp** — roster walks now cap `page_size` to `maxResults + 1` so bounded callers do not over-fetch the first page.

## Prior feedback status

- ✅ **Resolved:** gate-check `page_size` (`7722680a7`)
- ✅ **Resolved:** mid-pagination RSVP test (`b686fd49c`)

## Issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 2 optional items: `getMeetingRegistrantsByEmail` JSDoc still omits `@param m2mToken`; `getRawMeetingRsvps` debug log could include `page_size` for ops parity

## Final decision

✅ **Approved**

dealako (approved):
Both minor follow-ups from the prior round are resolved (gate-check page_size + mid-pagination RSVP test). LGTM.

PR #1833, opened 2026-08-25
fix(org-lens): wire warehouse v2 health score into org lens

gaspergrom:
## Summary
- Selects the warehouse-computed `HEALTH_OVERALL_SCORE_V2` and `HEALTH_SCORE_CATEGORY_V2` columns (lf-dbt) in both the Org Lens Projects table and the project-detail hero, and classifies through the warehouse category first, falling back to the v1 numeric score then the raw v2 score.
- Adds `normalizeHealthScoreCategoryV2` to `@lfx-one/shared/utils` as the shared normalization point so both consumers can never disagree.
- Fixes the project-detail hero's `mapHealth`, which previously fell back only to the v1 score and never the raw v2 score — a project with a null v1 score but a real v2 score could classify differently in the hero than in the Projects table.

## Context
IN-1220 shipped v1→v2 label/color renaming plus a defensive legacy-band mapping utility, explicitly noting it "ships independently of dbt model changes (IN-1219)." IN-1219's warehouse columns have since landed, but nothing wired the Self Serve BFF to actually select and use them — this PR closes that gap. No new ticket was opened for this follow-up; the commits reference IN-1220 for traceability instead of a new LFXV2 ticket.

## Trade-offs
- Branch name and commit ticket format (`IN-1220`) don't match this repo's `LFXV2-`/`GH-` convention, since the underlying gap traces back to an Insights-tracked ticket rather than a new lfx-self-serve ticket.
- Boundary-value classification math (score thresholds 80/60/40/20) is not re-tested at the service layer — it's already exhaustively covered by `packages/shared/src/utils/insights.utils.spec.ts`; the new service-level tests focus on the fallback-chain wiring instead.

## Test plan
- [x] `yarn vitest run` — 70 files / 1669 tests passing
- [x] `yarn build` — production build succeeds
- [x] `yarn lint:check` / `yarn format:check` / `check-headers.sh` — clean
- [ ] Manual QA: Org Lens Projects table and project-detail hero show matching health classifications for warehouse-backed v2 rows

### DIFF
--- a/apps/lfx-one/src/server/services/org-lens-project-detail.service.spec.ts
+++ b/apps/lfx-one/src/server/services/org-lens-project-detail.service.spec.ts
@@ -41,7 +41,6 @@ const heroRow = {
   FOUNDATION_NAME: 'CNCF',
   IS_LF_PROJECT: true,
   DESCRIPTION: null,
-  HEALTH_OVERALL_SCORE: 80,
   SOFTWARE_VALUE: null,
   FIRST_COMMIT_TS: null,
 };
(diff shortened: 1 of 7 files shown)


MRashad26 (approved):
Reviewed the full diff (7 files, +229/-16) and ran the code-standards-enforcer agent against this PR. Both agree: PASS, only one non-blocking Warning-level observation.

**What this does:** Selects the warehouse-computed `HEALTH_OVERALL_SCORE_V2`/`HEALTH_SCORE_CATEGORY_V2` columns in both the Org Lens Projects table and project-detail hero queries, adds a shared `normalizeHealthScoreCategoryV2` util, and rewires both services' health classification to prefer the warehouse category, then the raw v2 score, then the raw v1 score — fixing a real pre-existing bug in the hero's `mapHealth` (it previously never fell back to the v2 raw score, so a null-v1/real-v2 project could classify differently in the hero than in the table).

**Verified correct:**
- **Consistent fallback chain** across both services and covered by matching test cases in the new `org-lens-projects.service.spec.ts` and the extended `org-lens-project-detail.service.spec.ts` and `insights.utils.spec.ts` — category → v2 raw → v1 raw, with an explicit "unrecognized category still checks raw scores" case (`'Typo'`).
- **`normalizeHealthScoreCategoryV2`** correctly returns `null` (not `'unavailable'`) for unset/unrecognized input so callers fall through to the numeric classifiers, rather than short-circuiting to unavailable on a typo/future warehouse label.
- **Test-mock pattern**: both new/extended specs mock `@lfx-one/shared/utils` but delegate to the real `insights.utils` implementation via a relative import — confirmed this exact workaround is already precedented in `ai.service.spec.ts` and `groups-engagement-stats.service.spec.ts` (needed because the real barrel re-exports Angular-platform utils unavailable in the server-only vitest environment), not a new deviation.

**Non-blocking observation (Warning, not Critical/Major):** in `org-lens-projects.service.ts`, `hasHealthScore(row)` only checks `(HEALTH_OVERALL_SCORE ?? HEALTH_OVERALL_SCORE_V2) != null` — it doesn't independently consider `HEALTH_SCORE_CATEGORY_V2`. If the warehouse ever populates the v2 category while both raw scores are null, that row would incorrectly report `'unavailable'` even though `mapHealthScore` could derive a real classification from the category alone. Given `HEALTH_SCORE_CATEGORY_V2` is presumably derived from the same underlying score by the `get_health_score_category_v2` dbt macro, this is likely a non-issue in practice (category and raw v2 score should populate together), but there's no test proving that coupling holds, and it costs little to make `hasHealthScore` category-aware too if you want to close the gap defensively.

**Also noting**, purely informational: the PR description's Summary line says the fallback order is "the v1 numeric score then the raw v2 score," but the actual code (and the `mapHealthScore` comment) prefers v2 raw over v1 raw — the code and tests are internally consistent with each other, just a wording slip in the description.

No critical or major issues. Approving.

Issue #1734, opened 2026-08-20
LFX Self Serve Latency


bramwelt:
## Problem

Multiple pages load slowly for users, each backed by multiple data sources
(query-service pagination, access-check/OpenFGA, role-grants, project
enrichment). Real user page loads:

| View | views | p50 | p95 |
|---|---|---|---|
| `/meetings/?` | 554 | 5.7s | **43.3s** |
| `/` | 321 | 4.0s | 35.4s |
| `/org/people` | 36 | 7.3s | 42.4s |
| `/profile/attributions` | 87 | 1.1s | 39.9s |
| `/foundation/overview` | 124 | 2.6s | 20.6s |
| `/project/overview` | 100 | 1.7s | 6.5s |

`GET /api/projects` (p50 ~23s, p95 65.9s) is the most-studied single endpoint
here, not the whole problem — it's the clearest window into a fan-out pattern
that recurs across pages. This is load-sensitive and getting worse (p50 was
12s a week ago), not a one-time regression.

## Verified attribution

Trace `c566183ba89e39eb` (18.4s, version 1.0.163), for `GET /api/projects`,
splits cleanly:

- **~10s** — sequential `/query/resources?page_token=…` pages, each
  100–840ms. OpenSearch inside each page is only 14–30ms.
- **8.2s** — a single trailing `POST /access-check`. Its `nats.request` is
  8,207ms; underneath, fga-sync issues OpenFGA `batch-check` calls of 1.43s,
  1.53s, 1.07s, 1.41s.

**The datastore is not the bottleneck — authorization is.** The same
pagination-then-bulk-access-check shape recurs on other pages (see
`enrichWithProjectData` below), which is why fixing it in one place likely
helps several views at once.

## Confirmed contributing mechanisms

Re-checked against production telemetry and `main` at `567a904` (2026-08-19):

1. **[Unbounded FGA-checked pagination in `getProjects` — CONFIRMED.](https://github.com/linuxfoundation/lfx-self-serve/issues/1735)**
   `project.service.ts:307` still calls `fetchAllQueryResources` with no
   `page_size` and no cap. The helper gained a `maxResults` early-stop option,
   but it's wired only into `meeting.service.ts`.

2. **[Top-of-hour burst saturation — CONFIRMED, quantitatively.](https://github.com/linuxfoundation/lfx-self-serve/issues/1653)** Bucketing
   `/api/projects` into 15-minute intervals shows p95 tracking concurrency
[...]

Issue #1653, opened 2026-08-19
app.lfx.dev degrades at top-of-hour meeting bursts — /api/projects p95 108s, role-grants 12s, p95 view load 32.7s (2026-08-19 09:00 PT)


manishdixitlfx:
## Summary

At **09:00–09:30 PT on 2026-08-19 (16:00–16:30 UTC)** app.lfx.dev became visibly slow. A top-of-the-hour meeting start produced a ~6x concurrency burst, and several BFF endpoints that sit on the critical path of *every* page load degraded 5–14x. Two of them were already slow at baseline.

This is not a single regression — it's a load-sensitivity problem with several compounding causes. Production pods were **not** restarting during the window (last prod deploy was 2026-08-18 22:21–22:40 UTC), so this is not deploy-related.

**Observed impact:** p95 `@view.loading_time` **32.7s** in the window, vs 4.8–18.7s in the preceding hours.

---

### 1. Traffic burst at exactly 09:00 PT

RUM, `service:lfx-self-serve`, 30-min buckets:

| Window (UTC) | Views | Resource requests | p75 view load | p95 view load |
|---|---|---|---|---|
| 15:00 | 19 | 1,385 | 4.3s | 18.7s |
| 15:30 | 24 | 2,534 | 8.2s | 17.6s |
| **16:00** | **134** | **10,738** | **4.4s** | **32.7s** |

Same slot yesterday (2026-08-18 16:00 UTC) was 58 views / 23 sessions — so this was ~2.3x a normal Tuesday 9am and ~6x the preceding half-hour.

The burst is meeting-driven: a single meeting view `/meetings/95903032570` took **34 of the 134 views** (p95 15.9s).

### 2. Endpoint latency — 9am window vs earlier the same morning

p95 `@resource.duration`, `service:lfx-self-serve`:

| Endpoint | 06:00–08:00 PT | **09:00 PT** | Calls in window | Degradation |
|---|---|---|---|---|
| `/api/projects` | 25–28s | **108.6s** | 26 | ~4x |
| `/api/meetings/{id}/my-meeting-registrants` | — | **49.8s** | 24 | — |
| `/api/meetings/{id}/rsvp/me` | — | **45.7s** | 1 | — |
| `/api/orgs/{id}/lens/people/all` | — | **35.6s** | 1 | — |
| `/api/orgs/me/role-grants` | 0.9–2.6s | **12.3s** | **63** | **5–14x** |
[...]

PR #1850, opened 2026-08-25
feat(cache): enrich Valkey oversize warnings (GH-1736)

### LINKED ISSUE #1736: Close instrumentation gap for Valkey cache timeout/oversize-skip hypothesis
## Problem

The "Valkey metastable collapse" hypothesis (originally raised in #1653) is
currently **unconfirmed and unfalsifiable**. `VALKEY_CACHE.OP_TIMEOUT_MS = 250`
and `MAX_VALUE_BYTES = 1_048_576` are confirmed at those values in
`packages/shared/src/constants/valkey-cache.constants.ts`, but production logs
over 08-17→08-20 contain **zero** Valkey timeout or oversize-skip entries.
Those paths log at `debug` level, and debug logs aren't shipped — so this
absence of evidence proves nothing either way.

See parent epic #1734, "Confirmed contributing mechanisms" #5.

## Why this matters

If the hypothesis is real, cache operations are silently failing/timing out
under load and falling back to a slower path (or skipping caching entirely
for oversized values), compounding the same tail-latency problem as the other
mechanisms in #1734. If it's not real, remediation effort should go
elsewhere. Right now nobody can tell which.

## Investigation

- [ ] Identify the specific Valkey timeout/oversize-skip log call sites and
      confirm they are indeed debug-level only.
- [ ] Determine the minimal-risk way to surface these as counters/metrics
      (promote specific log lines, or add a dedicated metric) without
      shipping full debug-level logging in production.
- [ ] Once shipped, re-check production data for evidence of actual
      timeout/oversize-skip occurrences correlated with the slow-request
      windows already identified (e.g. 08-19 16:00 UTC).

## Recommended changes

_To be filled in after investigation._

---
Tracked under #1734.

jordane:
## What

Enriches the two `logger.warning` calls on the Valkey read and write oversize paths (`valkey.service.ts`) with three structured fields:

- **`data.cache_namespace`** — the code-defined `{domain}:v{N}` label (e.g. `org-lens-sf:v1`, `org-seats:v1`) as an indexable field, so CloudWatch Insights queries can `stats count() by data.cache_namespace` instead of substring-matching the redacted `cache_key`.
- **`data.size_bytes`** — the actual payload size, so a marginal 1.05 MB skip is distinguishable from a 10× catastrophic overshoot.
- **`data.max_bytes`** — the current 1 MB cap, so future bumps don't require correlating logs against source.

Also adds a private `ValkeyService.extractNamespace()` helper with a documented anchoring rule that handles the `VALKEY_KEY_NAMESPACE=vN` edge case that `redactKey`'s own header calls out as "not worth special-casing", and correctly ignores sub-resource labels that embed their own `:vN:` schema-version segments (`projects:v3`, `meetings-kpi:v2`, `meetings-influence:v4`).

## Why

Closes GH-1736 (part of the GH-1734 latency epic).

The pre-existing oversize warnings fire ~12/day in production (98 events over 8 days per the GH-1734 re-check), all against `org-lens-sf` / `org-seats` keys. In the default deployment `redactKey` already preserves the namespace in the redacted prefix — but the field wasn't structured, so grouping and filtering required fragile substring queries, and there was no way to tell how far over the cap each event landed. `size_bytes` is the actually-new attribution: it's the signal that tells us whether a specific caller needs a slimmer row projection (near-cap) or a fundamentally different caching strategy (far-over).

## No behavior change

The branches, return values, and existing `cache_key: redactKey(key)` field are unchanged — only the log metadata payload is expanded.

## What this does NOT do

- No unit tests for `extractNamespace` — deferred by decision; `valkey.service.ts` has no companion spec today, and the reviewer trio flagged it as a natural follow-up alongside `redactKey`. Reasonable candidate for a small follow-up ticket.
- No enrichment on the other four cache-family warnings in the file (shape check, read failure, write failure, delete failure). Scoped intentionally; a natural follow-up if the same log-query need surfaces there.
- No production data check — that has to wait until this is merged and deployed. The next task on GH-1736 is querying CloudWatch for the `data.cache_namespace` / `data.size_bytes` distribution once the enriched warnings ship, to decide which specific size drivers (likely `fetchAllOrgSeats` and/or `OrgLensPeopleService.getAllEmployees`) to actually fix.

## Follow-ups

- Query production CloudWatch for the enriched attribution and share findings on GH-1736.
- Based on that data, file a follow-up for the specific size-reduction fix (row projection, per-endpoint cap, sharded key, etc.).
- Consider threading `req` through `getJson`/`setJson` for request-id correlation on the cache warnings (separable API change, out of scope here).

Related: GH-1734, GH-1653.

### DIFF
--- a/apps/lfx-one/src/server/services/valkey.service.ts
+++ b/apps/lfx-one/src/server/services/valkey.service.ts
@@ -72,8 +72,22 @@ export class ValkeyService implements CachePort {
       if (raw == null) return null;
       // setJson caps our own writes, but another client (or a manual write) could store an oversized value.
       // Parsing a very large JSON string blocks the event loop, so reject oversized reads as a miss before parsing.
-      if (Buffer.byteLength(raw, 'utf8') > VALKEY_CACHE.MAX_VALUE_BYTES) {
-        logger.warning(undefined, 'valkey_get', 'Cached value exceeds max size — treating as miss', { cache_key: ValkeyService.redactKey(key) });
+      const readSize = Buffer.byteLength(raw, 'utf8');
+      if (readSize > VALKEY_CACHE.MAX_VALUE_BYTES) {
+        // `cache_namespace` is the code-defined `{domain}:v{N}` label as a typed field so CloudWatch
+        // queries can group and filter oversize events by cache family without substring-matching
+        // the redacted `cache_key` (which already carries the same segment in the default
+        // deployment, but only until a caller sets `VALKEY_KEY_NAMESPACE` to a `vN`-shaped value —
+        // `redactKey`'s header calls that edge case out; `extractNamespace` closes it). `size_bytes`
+        // is genuinely net-new attribution: the existing warning couldn't distinguish a payload just
+        // over the 1 MB cap from one 10× over it, and that's exactly what tells us whether a caller
+        // needs a slimmer projection or a fundamentally different caching strategy.
+        logger.warning(undefined, 'valkey_get', 'Cached value exceeds max size — treating as miss', {
+          cache_key: ValkeyService.redactKey(key),
+          cache_namespace: ValkeyService.extractNamespace(key),
+          size_bytes: readSize,
+          max_bytes: VALKEY_CACHE.MAX_VALUE_BYTES,
+        });
         return null;
       }
       const parsed = JSON.parse(raw);
@@ -98,8 +112,20 @@ export class ValkeyService implements CachePort {
     if (!this.client) return false;
     try {
       const serialized = JSON.stringify(value);
-      if (Buffer.byteLength(serialized, 'utf8') > VALKEY_CACHE.MAX_VALUE_BYTES) {
-        logger.warning(undefined, 'valkey_set', 'Skipping cache write — value exceeds max size', { cache_key: ValkeyService.redactKey(key) });
+      const writeSize = Buffer.byteLength(serialized, 'utf8');
+      if (writeSize > VALKEY_CACHE.MAX_VALUE_BYTES) {
+        // Same enrichment as the read-path oversize warning: `cache_namespace` gives log queries a
+        // typed field to group by (rather than substring-matching the redacted `cache_key`, which
+        // works in the default deployment but breaks when `VALKEY_KEY_NAMESPACE` is `vN`-shaped),
+        // and `size_bytes` is the actually-new attribution — the existing warning didn't distinguish
+        // a payload just over the 1 MB cap from one 10× over it, and that gap is exactly what
+        // determines whether the fix is a slimmer projection or a different caching strategy.
+        logger.warning(undefined, 'valkey_set', 'Skipping cache write — value exceeds max size', {
+          cache_key: ValkeyService.redactKey(key),
+          cache_namespace: ValkeyService.extractNamespace(key),
+          size_bytes: writeSize,
+          max_bytes: VALKEY_CACHE.MAX_VALUE_BYTES,
+        });
         return false;
       }
       await this.withTimeout(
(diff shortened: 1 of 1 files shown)


dealako:
@jordane — nice, focused observability PR. The `extractNamespace` anchoring rule (first `vN` at index ≥ 2) is well thought through for the `VALKEY_KEY_NAMESPACE=vN` edge case and sub-resource labels like `meetings-kpi:v2`. Enriching only the two oversize warnings keeps scope tight with zero cache-behavior change.

**Bot reconciliation:** Cursor Bugbot's low-risk summary aligns with this review. CodeRabbit skipped (auto-review disabled). Copilot posted an overview with no inline findings to reconcile.

### Issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: missing unit tests for `extractNamespace` (author-deferred but worth a follow-up ticket); JSDoc cross-reference to L73 conflates oversized-read rationale with foreign-key trust boundary
- ⚪ Nit: 1 issue: verbose duplicated inline comments at both oversize call sites
- ❔ Question: 0 items

### Final decision

The missing tests are an acknowledged deferral in the PR body — not blocking merge, but I'd still open a small follow-up before the CloudWatch query work lands so the anchoring table doesn't regress silently.

apps/lfx-one/src/server/services/valkey.service.ts
```diff
@@ -179,6 +205,41 @@ export class ValkeyService implements CachePort {
...
+   * last `vN` would mis-attribute to the sub-resource's version instead of the domain's.
+   *
+   * Returns `null` on any key that doesn't carry a `vN` at a valid position with at least one
+   * segment after it (e.g. a manually-written entry from another consumer, or a truncated key).
+   * Both call sites pass the return value through unconditionally, so a null lands in the log
+   * payload as `cache_namespace: null` — deliberate for CloudWatch column consistency, so log
+   * queries can uniformly filter on `cache_namespace != null` rather than the field being
+   * present/absent based on key shape.
+   */
+  private static extractNamespace(key: string): string | null {
```
dealako, line 235: [minor] Untested extractNamespace and oversize log enrichment

Issue: New parsing logic and enriched warning payloads have no unit test coverage.
Proof: `extractNamespace` at L235-241 is invoked from oversize branches at L87 and L125. No `valkey.service.spec.ts` exists in the repo. Documented behaviors (sub-resource `:vN:` disambiguation, `vN`-shaped deployment namespace skip, truncated-key null return) are unverified.
Why it matters: A regression mis-attributes CloudWatch oversize events without breaking cache behavior, so CI stays green while GH-1736 dashboards lose signal.
Fix: Add table-driven tests for key shapes from `build*CacheKey` helpers plus two cases asserting `logger.warning` receives `cache_namespace`, `size_bytes`, and `max_bytes` on oversize read/write. Acknowledged deferral in PR body — fine as a follow-up ticket before production query work.

apps/lfx-one/src/server/services/valkey.service.ts
```diff
@@ -179,6 +205,41 @@ export class ValkeyService implements CachePort {
...
   }
 
+  /**
+   * Returns the code-defined `{domain}:v{N}` namespace segment (e.g. `org-lens-sf:v1`, `org-seats:v1`)
+   * for structured logging, so oversize/failure events can be attributed to a specific cache family
+   * without leaking the principal or sub-resource. For keys built via this service's `build*CacheKey`
+   * helpers, the namespace is one of the static, code-defined labels in `VALKEY_CACHE.*_NAMESPACE`
+   * and therefore carries no PII. A foreign consumer that shares the Valkey backend could write a
+   * key with any string at that position, so the returned value is only guaranteed PII-free for keys
+   * our own helpers produced — the read-path warning that logs this field explicitly acknowledges
```
dealako, line 215: [minor] JSDoc mis-cites read-path comment as PII caveat

Issue: JSDoc L215-217 cites the L73 comment as acknowledging foreign-key PII risk for `cache_namespace`, but L73 explains why oversized reads are rejected (event-loop safety / foreign oversized writes), not that `cache_namespace` may be untrusted on foreign keys.
Proof: L73-74: "another client (or a manual write) could store an oversized value" + event-loop parsing concern. JSDoc L213-217: foreign consumer / PII-free guarantee / cross-ref to L73.
Why it matters: Future readers may believe PII review already happened at the call site when the cited comment covers a different concern.
Fix: Reword the JSDoc paragraph to separate foreign-key trust from oversized-read checks, or add an explicit one-line caveat at the logger sites.

apps/lfx-one/src/server/services/valkey.service.ts
```diff
@@ -72,8 +72,22 @@ export class ValkeyService implements CachePort {
       if (raw == null) return null;
       // setJson caps our own writes, but another client (or a manual write) could store an oversized value.
       // Parsing a very large JSON string blocks the event loop, so reject oversized reads as a miss before parsing.
-      if (Buffer.byteLength(raw, 'utf8') > VALKEY_CACHE.MAX_VALUE_BYTES) {
-        logger.warning(undefined, 'valkey_get', 'Cached value exceeds max size — treating as miss', { cache_key: ValkeyService.redactKey(key) });
+      const readSize = Buffer.byteLength(raw, 'utf8');
+      if (readSize > VALKEY_CACHE.MAX_VALUE_BYTES) {
+        // `cache_namespace` is the code-defined `{domain}:v{N}` label as a typed field so CloudWatch
```
dealako, line 77: [nit] Oversize-path inline comments duplicate extractNamespace JSDoc

Issue: Eight-line comment blocks at L77-84 and L117-122 repeat rationale already on `extractNamespace` JSDoc (L208-234) and are far more verbose than peer logger sites (e.g. L96, L101).
Proof: Compare L77-84 with L96 ({ cache_key only, no block comment) and extractNamespace JSDoc.
Why it matters: Duplicated comments drift independently without adding information absent from the named helper.
Fix: Shorten to a single line (e.g. `// oversize metadata — see extractNamespace`) or delete entirely.

dealako (approved):
## dd-pr-review — feat(cache): enrich Valkey oversize warnings (GH-1736)

@jordane — focused observability change with solid `extractNamespace` anchoring logic. No cache read/write semantics changed; only structured log metadata expanded on the two oversize paths.

### Findings

- 🔴 **Blocking: 0**
- 🟡 **Minor: 2**
  1. **Untested `extractNamespace` parsing** — non-trivial anchoring rules (deployment `vN` namespace, embedded sub-resource `:vN:`) have no spec coverage; a regression would mis-attribute CloudWatch events without failing CI. Author deferral noted; recommend a follow-up ticket.
  2. **JSDoc mis-cites L73 for foreign-key PII caveat** — the cited comment covers oversized-read/event-loop safety, not `cache_namespace` trust on foreign keys.
- ⚪ **Nit: 1** — eight-line inline comment blocks at both oversize call sites duplicate the helper JSDoc; peer `logger.warning` sites in this file use one-liners.
- ❔ **Question: 0**

### Bot reconciliation

- **Cursor Bugbot:** agree — low-risk, observability-only.
- **CodeRabbit:** skipped (auto-review disabled); no findings.
- **Copilot:** overview only; no inline findings to incorporate.

PR #1856, opened 2026-08-26
fix(committees): exclude LF Staff seats with no voting status from engagement metrics

### LINKED ISSUE #1848: Engagement chip shows "High" + staff tooltip for LF Staff seats on committees with no voting roles
### Summary

On a committee with no voting (e.g. the new "LFX Extended Leadership Team" group), an LF Staff member's engagement chip renders **"High"** and the chip tooltip identifies them as staff with meaningless numbers ("LF Staff — attended 0 of 0 invited meetings").

Reported by @jordane in #lfx-one-app-dev (2026-08-25), observed in production on `app.lfx.dev/project/groups/57253516-4361-42b4-af5e-4e312333e301`:

> FYI there seems to be a bug in your engagement metrics in a group. Mine shows as "High" in the new group Connie created, but also the tooltip says I'm staff. Likely due to the fact that there is no voting on this committee so there is typically no voting role set.

His hypothesis is correct.

### Root cause

The LF Staff engagement exclusion (`isLfStaffObserverSeat`, `packages/shared/src/utils/committee-engagement-classifier.utils.ts`) only matches `role === LF_STAFF && votingStatus === OBSERVER` — the LFXV2-3101 follow-up narrowing (which exists so a staff member who is a real Voting/Alternate Voting Rep classifies on real attendance).

On a no-voting committee, members have no voting status, so `committee-engagement.service.ts` resolves `voting_status` to the `'None'` sentinel (`member.voting?.status || row?.MEMBER_VOTING_STATUS || CommitteeMemberVotingStatus.NONE`). A staff seat with `'None'` therefore falls through the exclusion entirely:

1. **Chip "High"** — classifier case 3 (`invited <= 0` + `joinedWithinWindow`) tenure-graces the staff seat to `High` on literal 0/0 counts.
2. **Wrong tooltip** — `resolveEngagementContext` in `committee-members.component.ts` skips its LF Staff+Observer branch and falls into the role-context branch, rendering "LF Staff — attended 0 of 0 invited meetings".

The same gap affects the summary counts (active/eligible/attendance-rate predicates) and the groups engagement rollup (`groups-engagement-stats.service.ts`), since they all key on the same helper.

### Fix

Broaden the exclusion to `role === LF_STAFF && (votingStatus === OBSERVER || votingStatus === NONE)` — i.e. staff seats are excluded from engagement metrics unless they hold a real voting seat. Deliberately not broadened to Emeritus (keeps its separately-documented behavior) and not to non-staff members with `'None'` status (they should classify normally).

Concretely:

- `packages/shared/src/utils/committee-engagement-classifier.utils.ts` — broaden `isLfStaffObserverSeat` (rename to something no longer Observer-specific), which propagates to `classifyCommitteeEngagement`, `isCommitteeMemberAtRisk`, `isCommitteeMemberActive`, `isCommitteeMemberActiveEligible`, `isCommitteeMemberRateEligible`, and both engagement services. Export a named predicate for the condition.
- `apps/lfx-one/src/app/modules/committees/components/committee-members/committee-members.component.ts` — route `resolveEngagementContext`'s inline staff+Observer check through the shared predicate so `'None'`-status staff get the "LF Staff seat — excluded from engagement metrics" tooltip.
- Spec cases in `committee-engagement-classifier.utils.spec.ts` for staff+`None` (the reported case, including the tenure-grace interaction), staff+Voting Rep unchanged, non-staff+`None` unchanged, staff+Emeritus unchanged.

### Out of scope

- Member card showing null voting status as "Observer" — #1831.
- Legacy "None" option in the member form — #1810.

### Expected result

On a committee with no voting roles, an LF Staff member's chip reads neutral "LF Staff" (never "High"/at-risk) with the standard exclusion tooltip; summary counts treat non-voting staff seats exactly like staff+Observer seats; staff who are real Voting/Alternate Reps still classify on real attendance.

manishdixitlfx:
## Summary

On committees with no voting enabled, LF Staff members typically have no recorded `voting_status` (the `'None'` sentinel), rather than the `'Observer'` status the previous exclusion rule keyed on. This let those seats fall through the LF Staff engagement exclusion, producing two misleading symptoms reported in production:

- The engagement chip read **"High"** (tenure-grace applied to a literal 0/0 attendance count).
- The tooltip read **"LF Staff — attended 0 of 0 invited meetings"**.

This broadens `isLfStaffObserverSeat` (renamed `isLfStaffNonVotingSeat`, now exported) to also match a `'None'` voting status, so a non-voting LF Staff seat gets the same neutral treatment as an Observer LF Staff seat. It does **not** broaden the carve-out for a genuine Voting Rep / Alternate Voting Rep LF Staff member, or for Emeritus — both keep classifying on real participation, per the LFXV2-3101 follow-up this extends.

Also closes a related gap: `groups-engagement-stats.service.ts`'s rollup was still treating a blank warehouse `MEMBER_VOTING_STATUS` as unresolved rather than normalizing it to `None`, so an LF Staff seat could escape exclusion there even after the committee-detail path was fixed. And the summary card now renders `—` instead of a literal `0%`/`0/0` when a committee's roster is entirely excluded from a metric — the same class of misleading-zero bug, one level up.

## Verified against the real reported committee

The fix requires `role === 'LF Staff'` to be resolvable for the affected members. The member-add form hides **both** the Role and Voting Status fields when `committee?.enable_voting` is false, so this needed checking against real data rather than assumed. Fetched the live roster for the reported committee ("LFX Extended Leadership Team", `57253516-4361-42b4-af5e-4e312333e301`) directly:

- **24 of 26 members have `role: "LF Staff"` populated**, `voting_status` blank on all of them — exactly the reported bug shape. The fix fires for them.
- 2 members have blank `role` (a pre-existing, unrelated data gap — they wouldn't have matched the old Observer-only rule either).

Also visually verified in the running app: with a live member's `voting_status` patched to `'None'` and `attended`/`invited` forced to `0/0` (matching the exact reported shape), the row renders the neutral gray **"LF Staff"** chip — not "High" — with tooltip *"LF Staff seat — excluded from engagement metrics; attendance expectations do not apply"*.

## Known behavior change: blank warehouse voting status on a *voting-enabled* committee

The `|| CommitteeMemberVotingStatus.NONE` fallback (both `committee-engagement.service.ts` and `groups-engagement-stats.service.ts`) conflates "recorded as non-voting" with "not synced yet." On a committee that *does* run votes, an LF Staff member whose real voting status hasn't synced to the warehouse yet is now silently excluded from `attendance_rate`/`active_count`, where they previously counted (as an unresolved/blank status, not explicitly excluded). This is inherent to broadening the `None` sentinel to mean "non-voting" — a deliberate trade-off for this fix, not something this PR mitigates further.

## Documented trade-offs

- **Deep import in e2e helper** (`e2e/helpers/committee-engagement.helper.ts`): imports `@lfx-one/shared/utils/committee-engagement-classifier.utils` directly rather than the `@lfx-one/shared/utils` barrel. The barrel transitively re-exports files that import `@angular/forms`/`@angular/common/http`, which Playwright's module loader can't JIT-compile. Four existing e2e files already do this (`org-roi-summary.spec.ts`, `org-roi-projects.spec.ts`, `org-roi-project-detail.spec.ts`, `past-meeting-ai-summary-visibility.spec.ts`).
- **Client-side rate-eligibility re-derivation**: `CommitteeEngagementSummaryComponent` re-derives whether any rate-eligible member has invites this window by scanning `members[]`, rather than the BFF exposing that fact directly (the way `eligible_count` does for the active-members metric). Verified correct and covered by tests; a follow-up could move this server-side to remove the duplicated predicate and the "don't truncate `members[]`" coupling documented in the interface file.

## Test plan

- [x] `yarn lint:check`, `yarn check-types`, `yarn format:check` — all pass
- [x] `yarn test` — shared + server + app suites all pass
- [x] `yarn build` — succeeds
- [x] `npx playwright test --list` on `committee-engagement.spec.ts` / `committee-engagement-robust.spec.ts` — 42 tests discovered, no errors
- [x] New/updated coverage: `isLfStaffNonVotingSeat` decision table, `committee-engagement.service.ts` regression test for the exact reported bug shape, `groups-engagement-stats.service.ts` blank-status normalization, `committee-engagement-summary.component.spec.ts` em-dash/aria-label guards
- [x] Manual verification against the real "LFX Extended Leadership Team" committee data (see above)

Fixes #1848. Extends the LFXV2-3101 follow-up narrowing (staff+Observer → staff+non-voting).

### DIFF
--- a/apps/lfx-one/e2e/committee-engagement.spec.ts
+++ b/apps/lfx-one/e2e/committee-engagement.spec.ts
@@ -95,6 +95,13 @@ test.describe('Committee engagement — members table (flag on)', () => {
     await expect(page.getByTestId('members-engagement-chip-m-emeritus')).toContainText('Emeritus');
     await expect(page.getByTestId('members-engagement-chip-m-lf-staff')).toContainText('LF Staff');
     await expect(page.getByTestId('members-engagement-chip-m-lf-staff-none')).toContainText('LF Staff');
+    // The chip text alone doesn't prove the reported bug's second symptom stays fixed — the tooltip
+    // (resolveEngagementContext) is a separate code path that could regress to the generic
+    // "attended X of Y invited meetings" role-context branch while classification still renders
+    // correctly. Hover and assert the actual exclusion tooltip text.
+    await page.getByTestId('members-engagement-chip-m-lf-staff-none').hover();
+    await expect(page.locator('.p-tooltip')).toBeVisible({ timeout: ELEMENT_TIMEOUT });
+    await expect(page.locator('.p-tooltip')).toContainText('LF Staff seat — excluded from engagement metrics; attendance expectations do not apply');
     // An LF Staff member who is a real Voting Rep classifies on real attendance, NOT the neutral
     // LF Staff tier.
     await expect(page.getByTestId('members-engagement-chip-m-lf-staff-rep')).toContainText('High');
--- a/apps/lfx-one/e2e/committee-engagement.spec.ts
+++ b/apps/lfx-one/e2e/committee-engagement.spec.ts
@@ -87,12 +87,16 @@ test.describe('Committee engagement — members table (flag on)', () => {
     await expect(page.getByTestId('members-engagement-chip-m-medium')).toContainText('Medium');
     await expect(page.getByTestId('members-engagement-chip-m-low')).toContainText('Low');
     await expect(page.getByTestId('members-engagement-chip-m-inactive-invited')).toContainText('Inactive');
-    // Emeritus and LF Staff + Observer both render their own neutral tier — never
-    // Low/Inactive/at-risk styling.
+    // Emeritus, LF Staff + Observer, and LF Staff + no voting status all render their own neutral
+    // tier — never Low/Inactive/at-risk styling. m-lf-staff-none's classification is fixture-mocked
+    // ('LF Staff' hardcoded below), so this only proves the UI renders it correctly, not that the
+    // tenure-grace decision is right — that's covered directly in
+    // committee-engagement-classifier.utils.spec.ts's classifyCommitteeEngagement cases.
     await expect(page.getByTestId('members-engagement-chip-m-emeritus')).toContainText('Emeritus');
     await expect(page.getByTestId('members-engagement-chip-m-lf-staff')).toContainText('LF Staff');
+    await expect(page.getByTestId('members-engagement-chip-m-lf-staff-none')).toContainText('LF Staff');
     // An LF Staff member who is a real Voting Rep classifies on real attendance, NOT the neutral
-    // LF Staff tier (LFXV2-3101 follow-up, Jordan Evans review).
+    // LF Staff tier.
     await expect(page.getByTestId('members-engagement-chip-m-lf-staff-rep')).toContainText('High');
 
     // computed_at is null in the fixture → the daily-refresh freshness fallback.
@@ -120,6 +124,7 @@ test.describe('Committee engagement — members table (flag on)', () => {
     await expect(page.getByText('Nova Neverasked')).toHaveCount(0);
     await expect(page.getByText('Evan Emeritus')).toHaveCount(0);
     await expect(page.getByText('Sam Staffer')).toHaveCount(0);
+    await expect(page.getByText('Alex Nonstatus')).toHaveCount(0);
     // Real 80% attendance — not at-risk, filtered out same as any other well-engaged member.
     await expect(page.getByText('Priya Repstaff')).toHaveCount(0);
   });
@@ -151,6 +156,13 @@ test.describe('Committee engagement — members table (flag on)', () => {
 
     await expect(page.getByTestId('members-engagement-meetings-m-high')).toHaveText('—', { timeout: ELEMENT_TIMEOUT });
     await expect(page.getByTestId('members-engagement-chip-m-high')).toHaveText('—');
+    // `engagementDataAvailable()` gates the chip render for every row (committee-members.component.
+    // html), so every classification — including m-lf-staff-none's — renders the same em-dash here
+    // regardless of what degradedClassification computed; that delegation isn't observable on this
+    // path or the live path above (both fixture-mock the classification directly rather than
+    // deriving it). isLfStaffNonVotingSeat's own decision table is covered directly in
+    // committee-engagement-classifier.utils.spec.ts's classifyCommitteeEngagement cases.
+    await expect(page.getByTestId('members-engagement-chip-m-lf-staff-none')).toHaveText('—');
     // A degraded response zeroes every row, so the At Risk chip would be a permanent "(0)" — hidden.
     await expect(page.getByTestId('members-chip-atRisk')).toHaveCount(0);
     // The roster stays fully functional.
@@ -202,12 +214,37 @@ test.describe('Committee engagement — overview summary (flag on)', () => {
     test.skip(!flagOn, 'wg-engagement-metrics flag appears OFF for this test user — see file header for the LD precondition');
 
     await expect(summary.getByTestId('committee-engagement-summary-attendance-rate')).toContainText('78%', { timeout: ELEMENT_TIMEOUT });
-    // 4/6, not 4/8 — the denominator is eligible_count (roster minus Emeritus/LF Staff+Observer),
+    // 4/6, not 4/9 — the denominator is eligible_count (roster minus Emeritus/non-voting LF Staff),
     // not total_count. active_count is 4 (m-high, m-medium, m-low, m-lf-staff-rep — the real Voting
     // Rep LF Staff member counts like any other real member).
     await expect(summary.getByTestId('committee-engagement-summary-active-members')).toContainText('4/6');
     await expect(summary.getByTestId('committee-engagement-summary-at-risk')).toContainText('2');
     await expect(summary.getByTestId('committee-engagement-summary-freshness')).toHaveText('Updated daily');
+  });
+
+  test('renders an em-dash, not "0/0", for a committee whose roster is entirely excluded from active_count (GH-1848)', async ({ page }) => {
+    await mockCommitteeShell(page);
+    // Only `summary` is overridden — the active-members row this assertion targets reads only
+    // `summary.eligible_count`/`active_count`, not `members[]`, so the roster itself can stay the
+    // default mixed fixture (the rest of the response, notably `data_available: true`, also stays
+    // default, which is what keeps the metrics block rendered at all instead of falling into the
+    // coming-soon state). The attendance-rate row is a different story — it re-derives its own gate
+    // from `members[]` (`hasInvitedRateEligibleMember`), and the default fixture's roster still has
+    // an invited rate-eligible member, so it renders a real percentage here rather than an em-dash;
+    // that's fine, this test doesn't assert on it. `eligible_count: 0` here stands in for the real-world
+    // shape GH-1848 makes reachable: a committee without voting where every seat is Emeritus or
+    // non-voting LF Staff, so the ratio's denominator is genuinely 0.
+    await mockEngagementApi(page, (window) => {
+      const base = buildEngagementResponse(window);
+      return { ...base, summary: { ...base.summary, attendance_rate: 0, active_count: 0, eligible_count: 0, at_risk_count: 0 } };
+    });
+    await gotoEngagementCommitteeTab(page, 'overview');
+
+    const summary = page.getByTestId('committee-overview-engagement-summary');
+    const flagOn = await appearsWithin(summary, ELEMENT_TIMEOUT);
+    test.skip(!flagOn, 'wg-engagement-metrics flag appears OFF for this test user — see file header for the LD precondition');
+
+    await expect(summary.getByTestId('committee-engagement-summary-active-members')).toContainText('—', { timeout: ELEMENT_TIMEOUT });
   });
 
   test('degraded response shows the calm coming-soon state, not an error', async ({ page }) => {
--- a/apps/lfx-one/e2e/helpers/committee-engagement.helper.ts
+++ b/apps/lfx-one/e2e/helpers/committee-engagement.helper.ts
@@ -5,15 +5,30 @@
  * Shared fixtures/mocks for the committee engagement UI specs (LFXV2-1705).
  *
  * The engagement fixture exercises all six classification tiers the BFF can serve (High / Medium /
- * Low / Inactive / Emeritus / LF Staff, LFXV2-3101) over seven fixture scenarios (Inactive appears
- * twice: with and without invites) on a roster whose uids match the mocked `/members` response, so
+ * Low / Inactive / Emeritus / LF Staff, LFXV2-3101) over nine fixture scenarios (High, Inactive, and
+ * the LF Staff *tier* each appear twice — m-lf-staff (Observer) and m-lf-staff-none (no voting
+ * status, broadened in GH-1848) both classify LF Staff; the roster's second LF Staff-*role* seat,
+ * m-lf-staff-rep, deliberately does NOT — it's a real Voting Rep) on a roster whose uids match the
+ * mocked `/members` response, so
  * the Members-table join and the At-Risk filter behave exactly as they would against the real
  * endpoint. Specs mock the BFF over `page.route` and stay independent of the server's
  * ENGAGEMENT_BACKEND mode.
  */
 
 import { COMMITTEE_ENGAGEMENT_DEFAULT_WINDOW } from '@lfx-one/shared/constants';
+import { CommitteeMemberVotingStatus } from '@lfx-one/shared/enums';
 import type { CommitteeEngagementResponse, CommitteeEngagementWindow, CommitteeMemberEngagement } from '@lfx-one/shared/interfaces';
+// Deep-imported (not the `@lfx-one/shared/utils` category barrel — mirroring the existing e2e
+// precedent for deep `utils` imports in org-roi-summary.spec.ts / org-roi-projects.spec.ts /
+// org-roi-project-detail.spec.ts / past-meeting-ai-summary-visibility.spec.ts) because the barrel's
+// `index.ts` re-exports every utils
+// file, including ones that import `@angular/forms`/`@angular/common/http`; Playwright's own module
+// loader can't JIT-compile those. This single file's own runtime imports (both Angular-free:
+// `../constants/committee-engagement.constants` and `../enums`) don't pull any of that in today, so
+// importing it directly sidesteps the barrel entirely instead of hand-copying its predicate — but
+// that's a property of its current import graph, not a guarantee; re-check if this file ever gains
+// a new runtime dependency.
+import { isCommitteeMemberActiveEligible, isLfStaffNonVotingSeat } from '@lfx-one/shared/utils/committee-engagement-classifier.utils';
 import { expect, Locator, Page, test } from '@playwright/test';
 
 export const PAGE_LOAD_TIMEOUT = 30_000;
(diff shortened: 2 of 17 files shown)


packages/shared/src/utils/committee-engagement-classifier.utils.ts
```diff
@@ -34,29 +34,45 @@ export function computeCommitteeEngagementRate(attended: number, invited: number
...
+ * The member form hides the Role/Voting Status fields entirely on a non-voting committee
+ * (`committee?.enable_voting`, member-form.component.html's Governance section), which is the
+ * mechanism behind the GH-1848 norm of staff seats there carrying no voting status — but that's a
+ * UI default, not a repo-enforced guarantee (API/import/legacy data can still set one), so a real
+ * Voting Rep or Emeritus staff seat on such a committee remains possible and, if present, still
+ * classifies/counts on real participation exactly as it would anywhere else.
  */
-function isLfStaffObserverSeat(input: Pick<CommitteeEngagementClassificationInput, 'role' | 'votingStatus'>): boolean {
-  return input.role === CommitteeMemberRole.LF_STAFF && input.votingStatus === CommitteeMemberVotingStatus.OBSERVER;
+export function isLfStaffNonVotingSeat(input: Pick<CommitteeEngagementClassificationInput, 'role' | 'votingStatus'>): boolean {
```
jordane, line 43: This requires `role === CommitteeMemberRole.LF_STAFF` to resolve. Per the PR body's own scope note, the member-add form hides both Role and Voting Status on non-voting committees, so `role` may be blank for the reported committee's staff seats too — in which case this exclusion never fires for them, and `committee-engagement.service.ts`'s `member.role?.name || row?.MEMBER_ROLE || CommitteeMemberRole.NONE` fallback wouldn't save it either. Please verify against the live "LFX Extended Leadership Team" data before merge.

packages/shared/src/utils/committee-engagement-classifier.utils.ts
```diff
@@ -34,29 +34,45 @@ export function computeCommitteeEngagementRate(attended: number, invited: number
...
+  return (
+    input.role === CommitteeMemberRole.LF_STAFF &&
+    (input.votingStatus === CommitteeMemberVotingStatus.OBSERVER || input.votingStatus === CommitteeMemberVotingStatus.NONE)
```
jordane, line 46: Falling back to the `None` sentinel elsewhere in the stack means this now also silently excludes an LF Staff member whose real voting status simply hasn't synced yet (vs. genuinely being non-voting) on a voting-enabled committee — worth flagging as a known behavior change in the PR description.

apps/lfx-one/src/app/modules/committees/components/committee-engagement-summary/committee-engagement-summary.component.ts
```diff
@@ -35,37 +40,74 @@ export class CommitteeEngagementSummaryComponent {
...
-  // total_count could never reach 100% for a committee that seats either. See
-  // CommitteeEngagementSummary.eligible_count's doc.
+  // Emeritus/non-voting LF Staff seats that active_count's numerator always excludes, so
+  // active_count/total_count could never reach 100% for a committee that seats either. See
+  // CommitteeEngagementSummary.eligible_count's doc. `eligible_count === 0` (a roster made up
+  // entirely of Emeritus/non-voting-LF-Staff seats, GH-1848) is the source-of-truth condition for
+  // the '—' fallback below (both the display label and, via `hasEligibleMembers`, the aria-label) —
+  // rather than the literal "0/0", which would read as a data failure on a real, populated
+  // committee.
+  private readonly hasEligibleMembers: Signal<boolean> = computed(() => (this.engagement()?.summary?.eligible_count ?? 0) > 0);
```
jordane, line 86: `activeMembersLabel` gates on the server's `summary.eligible_count === 0` here, while `attendanceRateLabel` (line 72/62) gates on a client-side rescan of `members[]` via `hasInvitedRateEligibleMember`. Two different derivation strategies for structurally the same guard, with a documented "don't truncate `members[]`" coupling as a result. Not blocking (behavior's correct and covered by the new spec), but worth a follow-up ticket for a single server-side field both can read.

apps/lfx-one/src/server/helpers/committee-engagement-mock.helper.spec.ts
```diff
@@ -1,7 +1,25 @@
...
+// vote.utils.ts import @angular/forms / @angular/common/http), which this Node-environment suite
+// can't load — so it needs a stub. Unlike `@lfx-one/shared/enums` (imported at runtime by the
+// helper under test, committee-engagement-mock.helper.ts, not by this spec file directly), which
+// resolves cleanly through the package's built `exports` map and so needs no stub — under `yarn
+// test`, where turbo.json's `test.dependsOn: ["^build"]` guarantees `packages/shared/dist` exists
+// first; a bare `yarn test:server` on a clean tree bypasses turbo and needs `yarn build` run first
+// too. The predicate is deep-imported from its real implementation (not hand-copied) so a decision-
+// table change there fails this suite too (GH-1848); its own boundary behavior is exhaustively
+// covered in packages/shared/src/utils/committee-engagement-classifier.utils.spec.ts.
+vi.mock('@lfx-one/shared/utils', async () => {
```
jordane, line 17: Nit: this mock returns a single-key object for the whole `@lfx-one/shared/utils` barrel — any future import of another barrel export in the helper under test will fail obscurely ("not a function") rather than pointing at this mock.

apps/lfx-one/e2e/helpers/committee-engagement.helper.ts
```diff
@@ -112,87 +134,142 @@ export function buildRoster(): Record<string, unknown>[] {
...
+// row's hand-written `classification` against role/voting_status/attendance — that's on the editor
+// making a seat-type change.
+//
+// Runs at module load (not inside buildEngagementResponse) so a drift throws immediately and
+// visibly at import time, rather than inside a page.route handler where the failure would surface
+// indirectly as a hung/timed-out request instead of a readable error.
+const rosterSeats = ROSTER.map((seat) => `${seat.uid}|${seat.role}|${seat.votingStatus}`).join(',');
+const memberSeats = MEMBER_SEATS.map((m) => `${m.uid}|${m.role}|${m.voting_status}`).join(',');
+if (memberSeats !== rosterSeats) {
+  throw new Error(`committee-engagement fixture drift: engagement rows [${memberSeats}] don't match ROSTER [${rosterSeats}]`);
```
jordane, line 226: Nit: this throws at module load, which in Playwright surfaces as a suite-collection error rather than the "readable error inside a page.route handler" framing the comment above implies it's avoiding. Still strictly better than the alternative — just a minor comment-accuracy nit.

jordane (changes requested):
Functionally sound and well-tested (decision table, service regression test for the exact reported shape, rollup normalization, accessibility guards) — but requesting changes on one unresolved correctness gap plus branch shape.

## Blocking

**1. The fix may not actually fix the reported committee.** `isLfStaffNonVotingSeat` requires `role === 'LF Staff'` to resolve. The PR's own "Scope note" says the member-add form hides **both** Role and Voting Status when `committee?.enable_voting` is false — so a staff member added through that form on a non-voting committee likely has a blank `role` too, not just blank `voting_status`. `committee-engagement.service.ts` then falls back `member.role?.name || row?.MEMBER_ROLE || CommitteeMemberRole.NONE`, and the exclusion never fires. The PR body states this was "not independently verified against the live 'LFX Extended Leadership Team' data" — that verification is the actual gate here. Please confirm `role` is populated for the reported committee's staff seats before merge (or add a `role` fallback if it isn't).

**2. Please squash to one commit.** 44 commits, 15+ of them literally `fix(review): correct <a comment>` — this is a reviewer-cycle transcript, not a change narrative. Squash to a single `fix(committees): ...` commit (keep `Signed-off-by`).

## Important

**3. Silent exclusion of real Voting Reps with unsynced status.** `votingStatus` now falls back to the `None` sentinel when both roster and warehouse are missing, which conflates "recorded as non-voting" with "not recorded yet." On a *voting-enabled* committee, an LF Staff member whose `voting.status` hasn't synced is now silently dropped from `attendance_rate`/`active_count` where they previously counted. This is documented in a code comment but is a real behavior change in the opposite direction from this PR's goal — worth a line in the PR description as a known impact.

**4. Comment/prose density.** A large share of the ~900 added lines is JSDoc prose — 10–15 line comments on 1–3 line changes, several rewriting an earlier comment that was itself wrong, plus review-cycle archaeology ("Jordan Evans review", "Cursor Bugbot follow-up") that reads as provenance rather than guidance for future readers. Please trim alongside the squash — keep the decision-table rationale and the interface docs, cut the per-import/per-signal narration.

## Nits

- PR title is 85 chars, over the 72-char team target (informational only — commitlint's 100-char hard limit and `pr-title-lint` both pass).

No prior human review comments to verify (only a Copilot bot summary). CI is green across all 11 checks.

packages/shared/src/utils/committee-engagement-classifier.utils.ts
manishdixitlfx: Verified against the real committee — fetched the live roster for "LFX Extended Leadership Team" (`57253516-4361-42b4-af5e-4e312333e301`) directly: 24 of 26 members have `role: "LF Staff"` populated with `voting_status` blank, exactly the reported bug shape. Only 2 members have blank `role` (a pre-existing, unrelated data gap — they wouldn't have matched the old Observer-only rule either). Also patched a live member's response to the exact 0/0 + `voting_status: 'None'` shape and confirmed it renders the neutral gray "LF Staff" chip, not "High".

Added this verification to the PR description (see "Verified against the real reported committee"). See ed6c1590b.

packages/shared/src/utils/committee-engagement-classifier.utils.ts
manishdixitlfx: Agreed — added this as a documented known impact in the PR description ("Known behavior change: blank warehouse voting status on a voting-enabled committee"). Accepting it as inherent to broadening the `None` sentinel's meaning to "non-voting" rather than mitigating further in this PR.

See ed6c1590b.

apps/lfx-one/src/app/modules/committees/components/committee-engagement-summary/committee-engagement-summary.component.ts
manishdixitlfx: Agreed, and already captured as a documented trade-off in the PR description ("Client-side rate-eligibility re-derivation") — a follow-up could move this behind a single BFF-provided field the way `eligible_count` already works. No change in this PR.

apps/lfx-one/src/server/helpers/committee-engagement-mock.helper.spec.ts
manishdixitlfx: Fixed — the mock now throws a clear "doesn't stub 'x'" error naming this file when an unstubbed barrel export is accessed, instead of silently resolving to `undefined`. (Couldn't spread the real barrel via `importOriginal()` — that's exactly the Angular-import failure this mock exists to avoid — so used a Proxy with a targeted guard instead.)

See ed6c1590b.

apps/lfx-one/e2e/helpers/committee-engagement.helper.ts
manishdixitlfx: Fixed — corrected the comment to say this surfaces as a suite-collection error at this exact file/line, rather than the stale "readable error inside a page.route handler" framing.

See ed6c1590b.

manishdixitlfx:
## Review Feedback Addressed

Commit: ed6c1590b0e0e5cae06b1de073c56ca77dcabbfc — squashed the branch (was 44 commits, mostly `fix(review):` iteration) to this one commit per your request.

### Changes Made
- **`committee-engagement-mock.helper.spec.ts`**: replaced the single-key `vi.mock` object with a Proxy that throws a clear "doesn't stub 'x'" error naming this file, instead of silently resolving an unstubbed barrel export to `undefined` (per @jordane)
- **`e2e/helpers/committee-engagement.helper.ts`**: corrected the drift-guard comment — the module-load throw surfaces as a Playwright suite-collection error at this exact file/line, not the "inside a page.route handler" framing it previously implied (per @jordane)
- **Comment density**: trimmed the heaviest docblocks across `committee-engagement-classifier.utils.ts`, `committee-engagement.interface.ts`, `committee-engagement.internal.interface.ts`, `groups-engagement-stats.interface.ts`, `committee-engagement-mock.helper.ts`, `groups-engagement-stats.service.ts`, `committee-engagement.service.ts`, and the e2e fixtures — cut review-cycle narration ("used to be X, changed because of Y", reviewer-name attribution) while keeping the behavioral rules and rationale (per @jordane)

### Verified with Real Data
- **`role` resolution for the reported committee** (blocking #1): fetched the live roster for "LFX Extended Leadership Team" directly — 24 of 26 members have `role: "LF Staff"` populated with `voting_status` blank, exactly the reported bug shape. The fix fires for them. Also visually confirmed in the running app: the exact 0/0 + `voting_status: 'None'` shape renders the neutral "LF Staff" chip and exclusion tooltip, not "High". Details in the PR description.

### Documented, No Code Change
- **None-fallback on voting-enabled committees** (important #3): added to the PR description as a known behavior change — an LF Staff member whose real voting status hasn't synced yet is now silently excluded on a voting-enabled committee, where they previously counted as unresolved. Inherent to broadening the `None` sentinel's meaning; not mitigated further here.
- **Two derivation strategies for the same guard shape** (`committee-engagement-summary.component.ts:86`): already captured as a documented trade-off in the PR description — a follow-up could move this behind a single BFF field, mirroring `eligible_count`.

### Threads Resolved

apps/lfx-one/e2e/committee-engagement.spec.ts
```diff
@@ -87,12 +87,16 @@ test.describe('Committee engagement — members table (flag on)', () => {
...
-    // Emeritus and LF Staff + Observer both render their own neutral tier — never
-    // Low/Inactive/at-risk styling.
+    // Emeritus, LF Staff + Observer, and LF Staff + no voting status all render their own neutral
+    // tier — never Low/Inactive/at-risk styling. m-lf-staff-none's classification is fixture-mocked
+    // ('LF Staff' hardcoded below), so this only proves the UI renders it correctly, not that the
+    // tenure-grace decision is right — that's covered directly in
+    // committee-engagement-classifier.utils.spec.ts's classifyCommitteeEngagement cases.
     await expect(page.getByTestId('members-engagement-chip-m-emeritus')).toContainText('Emeritus');
     await expect(page.getByTestId('members-engagement-chip-m-lf-staff')).toContainText('LF Staff');
+    await expect(page.getByTestId('members-engagement-chip-m-lf-staff-none')).toContainText('LF Staff');
```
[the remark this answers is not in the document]
manishdixitlfx: Done — added a hover + `.p-tooltip` assertion for the `m-lf-staff-none` chip's exclusion message, right after the chip-text check.

See commit 11f7f80c7: `apps/lfx-one/e2e/committee-engagement.spec.ts` now hovers the chip and asserts the tooltip reads "LF Staff seat — excluded from engagement metrics; attendance expectations do not apply", matching the existing hover-assertion pattern in `org-membership-documentation.spec.ts`. This closes the coverage gap you flagged — the tooltip regression path is now guarded independently of the chip-classification path.

manishdixitlfx:
## Review Feedback Addressed

Commit: 11f7f80c7

### Changes Made
- **apps/lfx-one/e2e/committee-engagement.spec.ts**: added a hover + `.p-tooltip` assertion for the `m-lf-staff-none` chip, verifying it renders the exclusion tooltip text ("LF Staff seat — excluded from engagement metrics; attendance expectations do not apply") rather than the generic attendance-count text — closes a coverage gap where the tooltip half of the reported bug could regress silently (per copilot-pull-request-reviewer)

### Threads Resolved

Note: all 5 threads from the prior review iteration remain resolved — confirmed unchanged by this push.

jordane (approved):
Re-reviewed at 11f7f80c7. All prior review threads (mine + Copilot's) are resolved with real fixes or documented trade-offs:

- LF Staff seat detection verified against the live "LFX Extended Leadership Team" roster (24/26 seats match the reported bug shape).
- Blank-`voting_status`-on-a-voting-enabled-committee trade-off documented in the PR description and inline.
- Barrel mock hardened with a Proxy that throws a named error on unstubbed exports.
- e2e drift-guard comment corrected.
- Tooltip regression path (the second reported symptom) now has its own hover + `.p-tooltip` assertion, independent of the chip-classification check.

Approving.

Issue #1732, opened 2026-08-20
Record ICLA invalidation date on the signature record


mlehotskylf:
Record **when** an ICLA was invalidated, as a dedicated attribute on the signature record, so the Me lens can display it.

**Why now**: the [v17 Final prototype](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html) shows an **Invalidated** ICLA row rendering a dated status note — `Invalidated · Jun 3, 2026` — alongside a read-only status pill and no row actions. That date has no source today.

**Problem**: When a staff admin invalidates a contributor's ICLA from the Project Control Center, the signature record captures no invalidation date. `signature_approved` is flipped to `false` and a human-readable string is written to the general-purpose `note` field — but `date_modified` is not updated, and the note text carries no timestamp. The invalidation event is logged separately, but that event does not carry the signature ID, so it cannot be tied back to a specific signature. In practice the invalidation date is unrecoverable, so the mockup's `Invalidated · <date>` cannot be populated.

**Scope**: ICLA invalidation performed by LF staff via the Project Control Center, and surfacing that date in the Me lens. This is a distinct use case from sanctions-driven revocation of ECLA/CCLA signatures (#1370) — different actor, different agreement type, different trigger. Note the prototype renders **Invalidated** and **Revoked** as two separate states with different icons and labels; they should not be collapsed into one.

**Acceptance criteria**:
- [ ] A dedicated invalidation-date attribute exists on the signature record, following the table's existing `date_*` naming convention (e.g. `date_invalidated`)
- [ ] The attribute is populated whenever an ICLA is invalidated through the Project Control Center admin path
- [ ] The date is exposed to the Me lens so an Invalidated ICLA row can render `Invalidated · <date>` per the prototype
- [ ] Additive attribute — signatures invalidated before this change carry no value, so both the API and the UI must tolerate empty values on pre-existing records and degrade gracefully (see open question below)
- [ ] The existing `note` field behaviour is unchanged by this ticket

**Open question — backfill**: every ICLA already invalidated has no date. Decide whether those rows render the status pill with no date, or are handled some other way.

**Out of scope**: capturing *who* invalidated and *why* — covered by linuxfoundation/easycla#5153, which also requires PCC UI and EasyCLA API changes. This ticket deliberately covers only the date so it can land independently.

mlehotskylf:
[...]

Issue #1370, opened 2026-08-10
Revocation metadata + date on the signatures table (backs the Revoked state)


WillsonHG:
Add **revocation metadata + date** on the signatures table so the read-only **Revoked** status (sanctions/OFAC) and manager-initiated removals are auditable and durably represented.

**Acceptance criteria**:
- [ ] Additive attributes on the signatures table: revocation **timestamp** + **reason/actor** (e.g. `sanctions_screening`, `cla_manager`, system)
- [ ] **The revocation timestamp for the sanctions case is the moment EasyCLA sets `is_sanctioned=true`** on the company (decided 2026-08-20, Michal) — i.e. the `UpdateCompanySanctionStatus(..., true, ...)` path (`company/repository.go:1287`, called from the live SSS screen in `v2/sign/service.go:3098`) must record that timestamp when it flips the flag. Note: **no such write exists today** — sanctions currently touch only the company record; this ticket adds the writer, not just the schema
- [ ] Surfaced to the Me lens as `Revoked · <date>` (backs the Revoked status task) and as the durable marker that keeps a revoked ECLA excluded from `auto_create_ecla`/Approved-List re-validation
- [ ] **No migration** — attributes are additive, but read by v1/v2 code paths, so consumers must tolerate **empty values on pre-M2 records** (companies sanctioned before this lands render the Revoked pill undated)
- [ ] Written by the system revocation path (sanctions) and by CLA-manager-initiated removals; **not** by any self-service Me-lens write (there are none)

_Re-scoped 2026-08-14: from "self-service invalidation metadata" to "revocation metadata + date" backing the system-set Revoked state and manager-initiated removals._

_Updated 2026-08-20: pinned the sanctions revocation date to the `is_sanctioned=true` flip and noted the writer does not exist yet. Full pill↔flag mapping: [status-matrix.md](https://github.com/linuxfoundation/easycla/pull/5155)._

**Repo**: `cla-backend-go`. **Relates to**: the Revoked-state task, #1256, #1253 (Story E).

---
_Reshaped 2026-08-14 per the legal/stakeholder review and the [v17 Final prototype](https://github.com/linuxfoundation/easyclav2-migration-planning/blob/main/Mockups/M2/EasyCLA_MyCLAs_Full_Prototype_Final.html)._

<img width="1156" height="745" alt="Image" src="https://github.com/user-attachments/assets/5586cf3c-04ab-4131-8c6f-b471902685ce" />

[...]

PR #1869, opened 2026-08-26
feat(clas): send contact CLA manager requests

### LINKED ISSUE #1861: Contact CLA Manager: deliver the message, and offer it on Valid ECLA rows
**Contact CLA Manager** is the third mode of the shared manager modal — #1574 built removal mode, #1372 approval mode. Unlike those two, it has never delivered anything: the contributor picks managers, writes a message, presses **Send**, and is told _"Message not sent"_. That was the correct behaviour when it shipped, because the producer had no `contact` request type. It now has one, so the no-op is stale.

Two frontend-only changes.

## 1. Deliver the message

Shipped behaviour:

- `apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.ts:94` — "Contact is a product-complete no-op: the producer has no contact requestType"
- same file, line 99 — the toast reads `Message not sent`
- `apps/lfx-one/src/server/controllers/clas.controller.ts:252` — the BFF rejects any request type that is not `approval` or `removal`

Producer contract, added in [easycla#5157](https://github.com/linuxfoundation/easycla/pull/5157) (merged 2026-08-25), in `cla-backend-go/swagger/common/my-cla-manager-request.yaml`:

- `requestType` enum is `[removal, approval, contact]`
- `message` is `maxLength: 4096`, "optional for removal and approval, **required (non-blank) for contact**"
- `POST /my-clas/{signatureID}/cla-manager-requests` — "contact (just deliver the contributor's free-form message, required and non-blank for this type, no change requested)"

**Acceptance criteria**

- [x] Send in contact mode posts `requestType: contact` with the selected recipients and the message, and reports delivery the way approval and removal already do
- [x] The message is **mandatory** in contact mode — Send stays disabled while it is blank, and the field is marked required. It remains optional for approval and removal
- [ ] The 4096 limit is enforced in **code points**, not UTF-16 units. go-swagger validates `maxLength` with `utf8.RuneCount`, so counting `String.length` would reject text the producer accepts
- [ ] The BFF rejects a blank or missing message for `contact` rather than relying on the disabled button
- [ ] Success, recorded-without-email, and zero-manager states behave as they do for approval and removal
- [ ] The stale "no-op" comments and the `Message not sent` copy are removed

## 2. Offer it on Valid rows

The M2 status matrix specifies Contact CLA Manager on "ECLA showing **Valid or Needs attention**". Today the gate requires Needs attention: `packages/shared/src/utils/cla-manager-actions.utils.ts:24`. A Valid row has nothing wrong, but the contributor may still have a question — a team change, an acquisition, another project — and a message-only action fits that.

**Acceptance criteria**

- [ ] `canContactClaManager` is true for an ECLA that is Valid or Needs attention
- [ ] It stays false for Invalidated (a manager's only lever is the approved list, and removal from it produces Needs attention, not Invalidated), for Revoked, and for the unresolved-coverage state
- [ ] It stays false for **every ICLA**. An ICLA has no covering CCLA and therefore no managers: both endpoints the modal depends on — `GET` and `POST /my-clas/{signatureID}/cla-manager[-request]s` — return 404 for an ICLA signature id by design

## Out of scope

- Invalidated and Revoked dates — #1370 and #1732
- Contact remains visible while impersonating, as Request approval and Request Removal already are; the write is blocked server-side by `blockDuringImpersonation`

Documentation for this lands with #1800.

ahmedomosanya:
Closes #1861 (Contact CLA Manager: deliver the message, offer on Valid rows)

## Summary

`Contact CLA Manager` now sends. Contact mode POSTs `requestType: 'contact'` alongside approval and removal; its message is mandatory (approval and removal keep it optional) with an inline error, a code-point counter, and `aria-required`. The BFF accepts `contact`, rejects a blank contact message, and counts the 4096 cap in code points. The action is offered on **Valid** as well as Needs-attention ECLA rows. ICLA is excluded.

The producer added `contact` to `requestType` in [easycla#5156 (MyCLAs contact requests)](https://github.com/linuxfoundation/easycla/pull/5156) (`dev`) and [easycla#5157 (M2 prod BE)](https://github.com/linuxfoundation/easycla/pull/5157) (prod `main`). The M2 status matrix specifies the action on “ECLA showing Valid or Needs attention”.

`lfx-textarea` gained a backwards-compatible `required` input (defaults `false`) wired to `aria-required`.

## How to verify

Deploy preview: https://ui-pr-1869.dev.v2.cluster.linuxfound.info

- Set `localStorage` flags `my-clas-enabled` and `my-clas-m2-enabled`, open `/profile/clas`
- Kebab menu on a **Valid** ECLA row offers Contact CLA Manager
- Kebab menu on a **Needs-attention** ECLA row offers Contact CLA Manager
- ICLA rows do not offer the action
- Send is blocked while the message is blank, with an inline error
- A filled message reports `Message sent` or `Message recorded`

Locally: `yarn start` in `apps/lfx-one`, same steps.

## Notes

The 4096 cap is counted in **code points**, not UTF-16 units, because the producer’s go-swagger `maxLength` validates with `utf8.RuneCount` — counting units would reject emoji-heavy text the producer accepts. That is also why the textarea carries no native `[maxlength]`. Message validation runs on the sanitized value (`sanitizePlainText`), matching the producer’s `SanitizePlainText` before it validates. A non-string `message` is a 400, not coerced.

ICLA is excluded structurally: both manager endpoints return 404 for an ICLA signature id, since an ICLA has no covering CCLA and so no managers to resolve.

Contact stays visible while impersonating, as approval and removal already are; the write is blocked server-side by `blockDuringImpersonation`.

## Out of scope

Invalidated and Revoked dates remain unwired ([#1370 (revocation metadata)](https://github.com/linuxfoundation/lfx-self-serve/issues/1370), [#1732 (Record ICLA invalidation date)](https://github.com/linuxfoundation/lfx-self-serve/issues/1732)).

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
+++ b/apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
@@ -31,7 +31,7 @@ SPDX-License-Identifier: MIT
     <label class="mt-4 mb-2 text-sm font-medium text-slate-700" for="cla-manager-message">
       Message
       @if (messageRequired) {
-        <span class="text-red-500">*</span>
+        <span class="text-red-500" aria-hidden="true">*</span>
       } @else {
         <span class="font-normal text-slate-400">(optional)</span>
       }
--- a/apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
+++ b/apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
@@ -28,15 +28,47 @@ SPDX-License-Identifier: MIT
       }
     </div>
 
-    <label class="mt-4 mb-2 text-sm font-medium text-slate-700" for="cla-manager-message">Message</label>
+    <label class="mt-4 mb-2 text-sm font-medium text-slate-700" for="cla-manager-message">
+      Message
+      @if (messageRequired) {
+        <span class="text-red-500">*</span>
+      } @else {
+        <span class="font-normal text-slate-400">(optional)</span>
+      }
+    </label>
+    <!-- No native [maxlength] — it counts UTF-16 units and would block an emoji-heavy message below
+         the producer's real code-point limit; messageError and the counter enforce the cap instead. -->
     <lfx-textarea
       class="block w-full"
       [form]="form"
       control="message"
       [rows]="3"
       styleClass="w-full cla-manager-message"
-      placeholder="Add a note for the CLA manager…"
+      [placeholder]="messageRequired ? 'Write your message to the CLA manager…' : 'Add a note for the CLA manager…'"
+      [describedBy]="visibleMessageError() ? 'cla-manager-message-error' : ''"
+      [invalid]="!!visibleMessageError()"
+      [required]="messageRequired"
       id="cla-manager-message" />
+    <div class="mt-1 flex justify-between">
+      @if (visibleMessageError() === 'required') {
+        <p id="cla-manager-message-error" class="text-xs text-red-500" role="alert" data-testid="contact-cla-manager-message-error">A message is required.</p>
+      } @else if (visibleMessageError() === 'too-long') {
+        <p id="cla-manager-message-error" class="text-xs text-red-500" role="alert" data-testid="contact-cla-manager-message-error">
+          Message must be {{ messageMaxLength }} characters or fewer.
+        </p>
+      } @else {
+        <span></span>
+      }
+      <span
+        class="text-xs text-slate-400"
+        role="status"
+        aria-live="polite"
+        aria-atomic="true"
+        [attr.aria-label]="messageLength() + ' of ' + messageMaxLength + ' characters used'"
+        data-testid="contact-cla-manager-message-counter">
+        {{ messageLength() }}/{{ messageMaxLength }}
+      </span>
+    </div>
   }
 
   @if (sendError()) {
--- a/apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
+++ b/apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
@@ -61,9 +61,6 @@ SPDX-License-Identifier: MIT
       }
       <span
         class="text-xs text-slate-400"
-        role="status"
-        aria-live="polite"
-        aria-atomic="true"
         [attr.aria-label]="messageLength() + ' of ' + messageMaxLength + ' characters used'"
         data-testid="contact-cla-manager-message-counter">
         {{ messageLength() }}/{{ messageMaxLength }}
--- a/apps/lfx-one/src/server/controllers/clas.controller.ts
+++ b/apps/lfx-one/src/server/controllers/clas.controller.ts
@@ -266,9 +266,17 @@ export class ClasController {
         return;
       }
 
+      // Do not `String(message)`: an object body becomes "[object Object]" and would pass the
+      // contact blank check, then fail the producer's `type: string` schema after the round trip.
+      const rawMessage = body?.message;
+      if (rawMessage !== undefined && rawMessage !== null && typeof rawMessage !== 'string') {
+        res.status(400).json({ message: 'Message must be a string' });
+        return;
+      }
+
       // Sanitized, because the producer validates and stores the sanitized text: control
       // characters other than newline and tab never reach it, so they must not count here.
-      const trimmedMessage = sanitizePlainText(String(body?.message ?? ''));
+      const trimmedMessage = sanitizePlainText(rawMessage ?? '');
       // Code points, not UTF-16 units, to match the producer's go-swagger rune cap — see
       // CLA_MANAGER_MESSAGE_MAX_LENGTH.
       if (codePointLength(trimmedMessage) > CLA_MANAGER_MESSAGE_MAX_LENGTH) {
--- a/apps/lfx-one/src/server/controllers/clas.controller.ts
+++ b/apps/lfx-one/src/server/controllers/clas.controller.ts
@@ -7,7 +7,9 @@
 // EasyCLA re-verifies each key belongs to the caller and owns the signature, so the
 // upstream endpoint — not this controller — is the ownership authorization boundary.
 
-import { CLA_GROUP_SEARCH_MIN_CHARS } from '@lfx-one/shared/constants';
+import { CLA_GROUP_SEARCH_MIN_CHARS, CLA_MANAGER_MESSAGE_MAX_LENGTH, CLA_MANAGER_REQUEST_TYPES } from '@lfx-one/shared/constants';
+import type { ClaManagerRequestType } from '@lfx-one/shared/interfaces';
+import { codePointLength } from '@lfx-one/shared/utils';
 import { NextFunction, Request, Response } from 'express';
 
 import { AuthenticationError } from '../errors';
@@ -42,6 +44,10 @@ function parseManagerRecipients(raw: unknown): string[] | null {
     recipients.push(trimmed);
   }
   return recipients;
+}
+
+function isClaManagerRequestType(value: string): value is ClaManagerRequestType {
+  return CLA_MANAGER_REQUEST_TYPES.some((requestType) => requestType === value);
 }
 
 export class ClasController {
@@ -231,8 +237,8 @@ export class ClasController {
   }
 
   // POST /api/me/clas/:signatureId/cla-manager-requests
-  // Approval or removal notice to selected CLA managers. Blocked during impersonation at the
-  // route. Contact mode never reaches this handler — the client does not call it.
+  // Approval, removal, or contact notice to selected CLA managers. Blocked during impersonation
+  // at the route.
   public async createClaManagerRequest(req: Request, res: Response, next: NextFunction): Promise<void> {
     const startTime = logger.startOperation(req, 'create_cla_manager_request');
 
@@ -249,8 +255,8 @@ export class ClasController {
 
       const body = req.body as { requestType?: unknown; recipients?: unknown; message?: unknown } | undefined;
       const requestType = String(body?.requestType ?? '').trim();
-      if (requestType !== 'approval' && requestType !== 'removal') {
-        res.status(400).json({ message: 'A request type of approval or removal is required' });
+      if (!isClaManagerRequestType(requestType)) {
+        res.status(400).json({ message: 'A request type of approval, removal, or contact is required' });
         return;
       }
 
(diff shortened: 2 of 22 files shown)


apps/lfx-one/src/server/controllers/clas.controller.ts
```diff
@@ -261,11 +267,21 @@ export class ClasController {
...
+      if (codePointLength(trimmedMessage) > CLA_MANAGER_MESSAGE_MAX_LENGTH) {
         res.status(400).json({ message: 'Message is too long' });
         return;
       }
 
+      // A contact request asks for no change, so the message is the whole of it. The producer
+      // refuses a blank one; answering here keeps that a 400 with usable copy rather than an
+      // opaque upstream rejection after a gateway round trip.
+      if (requestType === 'contact' && !trimmedMessage) {
+        res.status(400).json({ message: 'A message is required to contact the CLA manager(s)' });
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `1f41deaf1` — confirmed against the producer: `utils.SanitizePlainText` drops control characters other than newline and tab, and its own test asserts `SanitizePlainText(" \r\n \x07\x1b \t ")` is empty, with `v2/my_clas/service.go` rejecting a contact request whose sanitized message is blank. The same mismatch applied in the other direction to the length cap, which measured the raw input and so could reject text the producer would have accepted. A shared `sanitizePlainText` now mirrors that normalization, and a new `plainTextMessageValidator` runs both the blank check and the 4096-code-point cap on its output, so the form control, the character counter, and the BFF all agree with what the producer stores.

apps/lfx-one/src/server/services/cla.service.ts
```diff
@@ -805,7 +814,7 @@ export class ClaService {
     const requestId = result?.requestID?.trim();
     const requestType = result?.requestType;
     const status = result?.status;
-    if (!requestId || (requestType !== 'approval' && requestType !== 'removal') || (status !== 'sent' && status !== 'recorded')) {
+    if (!requestId || !requestType || !KNOWN_CLA_MANAGER_REQUEST_TYPES.has(requestType) || (status !== 'sent' && status !== 'recorded')) {
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `1f41deaf1` — the receipt check now requires `requestType === request.requestType` instead of mere membership of the enum, so a contact request that comes back with an approval or removal receipt is refused as unusable rather than reported as success under contact copy. Covered by a new case in `cla.service.spec.ts`.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
```diff
@@ -28,15 +28,47 @@
       }
     </div>
 
-    <label class="mt-4 mb-2 text-sm font-medium text-slate-700" for="cla-manager-message">Message</label>
+    <label class="mt-4 mb-2 text-sm font-medium text-slate-700" for="cla-manager-message">
+      Message
+      @if (messageRequired) {
+        <span class="text-red-500">*</span>
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `1f41deaf1` — the asterisk now carries `aria-hidden="true"`, leaving `aria-required` on the textarea as the single announcement of the required state.

apps/lfx-one/src/server/controllers/clas.controller.ts
```diff
@@ -260,12 +266,24 @@ export class ClasController {
         return;
       }
 
-      const trimmedMessage = String(body?.message ?? '').trim();
-      if (trimmedMessage.length > 4096) {
+      // Sanitized, because the producer validates and stores the sanitized text: control
+      // characters other than newline and tab never reach it, so they must not count here.
+      const trimmedMessage = sanitizePlainText(String(body?.message ?? ''));
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `db7d00f69` — `message` is now rejected unless it is a string (or omitted). `String({})` becoming `"[object Object]"` would have passed the contact blank check and then failed the producer's `type: string` schema after the round trip.

packages/shared/src/validators/index.ts
```diff
@@ -8,5 +8,6 @@ export * from './linux-alias.validator';
 export * from './mailing-list.validators';
 export * from './max-code-points.validator';
 export * from './meeting.validators';
+export * from './plain-text-message.validator';
```
audigregorie: **Minor — export order breaks the file's alphabetical convention.** Every other entry in this barrel is sorted (`meeting` → `newsletter` → `vote`); `plain-text-message` sorts after `newsletter`.

```suggestion
export * from './newsletter.validators';
export * from './plain-text-message.validator';
```

(That suggestion replaces this line with both entries in order — delete the now-duplicated `newsletter.validators` line below when applying.)

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
```diff
@@ -28,15 +28,47 @@
...
+      }
+      <span
+        class="text-xs text-slate-400"
+        role="status"
+        aria-live="polite"
+        aria-atomic="true"
+        [attr.aria-label]="messageLength() + ' of ' + messageMaxLength + ' characters used'"
+        data-testid="contact-cla-manager-message-counter">
+        {{ messageLength() }}/{{ messageMaxLength }}
+      </span>
```
audigregorie, line 67: **Minor — the counter announces every keystroke to screen readers.** `role="status"` already implies `aria-live="polite"` + `aria-atomic="true"`, so each character typed mutates this live region and queues a "N of 4096 characters used" announcement. For a 4096-cap field that means a long announcement queue while someone simply types a message.

The validation error already announces assertively via `role="alert"`, and the `aria-label` keeps the counter readable on demand — so consider dropping the live-region semantics here and letting it be a plain labelled element:

```suggestion
      <span
        class="text-xs text-slate-400"
        [attr.aria-label]="messageLength() + ' of ' + messageMaxLength + ' characters used'"
        data-testid="contact-cla-manager-message-counter">
        {{ messageLength() }}/{{ messageMaxLength }}
      </span>
```

(If you want it announced near the limit, that's better done by toggling the counter's own visibility/role only when within ~50 of the cap.)

packages/shared/src/constants/cla.constants.ts
```diff
@@ -37,18 +37,42 @@ export const CLA_GROUP_ORG_SOURCE_ICONS = {
...
+      recorded: { summary: 'Request recorded', detail: 'The request was recorded, but no CLA manager email could be delivered.' },
+    },
   },
   removal: {
     title: 'Request Removal',
     hint: (project: string) => `Ask the CLA manager(s) below to remove your ECLA for ${project}. This starts the process to invalidate it on your behalf.`,
+    receipt: {
+      sent: { summary: 'Request sent', detail: 'The CLA manager(s) you selected will be notified.' },
+      recorded: { summary: 'Request recorded', detail: 'The request was recorded, but no CLA manager email could be delivered.' },
+    },
```
audigregorie: **Minor — `approval.receipt` and `removal.receipt` are identical literals.** Four lines duplicated verbatim; if the two modes are meant to share the wording, a single `const` in this file referenced from both entries keeps them from drifting accidentally:

```suggestion
    receipt: CLA_MANAGER_REQUEST_RECEIPT,
```

with e.g.

```ts
/** Approval and removal share one receipt; contact phrases it as a message. */
const CLA_MANAGER_REQUEST_RECEIPT = {
  sent: { summary: 'Request sent', detail: 'The CLA manager(s) you selected will be notified.' },
  recorded: { summary: 'Request recorded', detail: 'The request was recorded, but no CLA manager email could be delivered.' },
} as const;
```

Optional — if the per-mode duplication is deliberate so the copy can diverge later, feel free to resolve as-is.

audigregorie (commented):
## Code Review Summary

Exemplary PR. The client, BFF, and shared-package validation all mirror the producer contract from easycla#5156 exactly — the code-point cap, sanitize-before-validate ordering, and the blank-contact 400 were verified against the merged producer diff (`SanitizePlainText` is a faithful mirror: CR/CRLF→LF, keep `\n`/`\t`, drop other Cc, trim). The receipt echo check in `cla.service.ts` (rejecting an upstream `requestType` that doesn't match the request) is a genuinely good hardening, and the test suite covers the boundary cases that matter (4096 vs 4097 code points, emoji, control characters, non-string coercion). Ship it.

### Minor — outside the diff

- **`packages/shared/src/interfaces/cla.interface.ts` — `ClaManagerModalCopy` is now stale.** The interface describes only `title` + `hint`, but `CLA_MANAGER_MODAL_COPY` entries gained a `receipt` field in this PR, and the interface has no consumers anywhere in the repo. Either annotate the constant (`Record<ClaManagerRequestMode, ClaManagerModalCopy>`-style, so future drift fails type-check), add `receipt` to the interface, or delete the interface. As-is it's dead weight that misdescribes the shape it names.

### What's done well

- Contract fidelity end-to-end: client validator, BFF validator, and the forwarded payload all measure the same sanitized code-point count the producer measures, so nothing is rejected client-side that the producer would accept (or vice versa).
- Defense in depth without redundancy: the BFF re-validates everything the client checks, and rejects a non-string `message` rather than coercing it into `"[object Object]"` — documented and tested.
- A11y done deliberately: decorative asterisk `aria-hidden`, `aria-required` on the control, `role="alert"` on the error, and the required error withheld until first edit so the modal doesn't open scolding the contributor.

packages/shared/src/validators/index.ts
ahmedomosanya: Addressed in `7538c185c` — `plain-text-message` now exports after `newsletter`, restoring alphabetical order in the validators barrel.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
ahmedomosanya: Addressed in `7538c185c` — the counter is a labelled element only; `role="status"` / `aria-live` are gone so each keystroke is not queued. The required/too-long error still announces via `role="alert"`.

packages/shared/src/constants/cla.constants.ts
ahmedomosanya: Addressed in `7538c185c` — approval and removal now share `CLA_MANAGER_REQUEST_RECEIPT`; contact keeps its own Message wording.

apps/lfx-one/src/app/modules/profile/clas/contact-cla-manager.component.html
```diff
@@ -28,15 +28,44 @@
...
+        <p id="cla-manager-message-error" class="text-xs text-red-500" role="alert" data-testid="contact-cla-manager-message-error">
+          Message must be {{ messageMaxLength }} characters or fewer.
+        </p>
+      } @else {
+        <span></span>
+      }
+      <span
+        class="text-xs text-slate-400"
+        [attr.aria-label]="messageLength() + ' of ' + messageMaxLength + ' characters used'"
+        data-testid="contact-cla-manager-message-counter">
```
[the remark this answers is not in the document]
ahmedomosanya: Convention: a 4096-cap counter that updates on every keystroke must not be a live region — each character would queue an announcement. The required/too-long error already announces via `role="alert"`, and `aria-label` keeps the count readable on demand. Dropped at reviewer request on this PR (the sibling `profile-edit-drawer` counter is a shorter bio field; it is not the pattern for this cap).

audigregorie (approved):

PR #1882, opened 2026-08-27
test(campaigns): assert the brief etag on every loadbrief arm

mrautela365:
Follow-up to #1837, addressing @dealako's `[minor]` review comment that arrived on the thread as that PR was merging.

## The finding

#1837 made `etag` a required member of `CampaignBriefLoadResult` and updated the `off` fixture to `etag: null` — but the `none` / `loaded` / `unreadable` fixtures and their matching `res.json` assertions still omitted it. Since `loadBrief` does a straight pass-through (`res.json(result)`), those three cases were asserting a shape the service no longer returns.

## The fix

Fixtures now match what `campaign-service.service.ts` actually produces:

| arm | etag |
|---|---|
| `none` | `null` (`campaign-service.service.ts:634`) |
| `loaded` | `found.etag` (`:667`) |
| `unreadable` | `found.etag` (`:666`) |

`loaded` and `unreadable` use **non-null** values deliberately. A `null` on every arm would pass equally well against a controller that dropped the field entirely, so the assertion would prove nothing about pass-through.

## Verified by mutation, not by a green suite

Destructuring `etag` out of the response before `res.json`:

```ts
const { etag: _drop, ...mutated } = result;
res.json(mutated);
```

fails all three cases — `3 failed | 148 passed`. Restored afterwards. Before this change that same mutation left the suite green.

Full spec run: **151 passed**.

Refs LFXV2-3204

### DIFF
--- a/apps/lfx-one/src/server/controllers/campaign.controller.spec.ts
+++ b/apps/lfx-one/src/server/controllers/campaign.controller.spec.ts
@@ -333,13 +333,13 @@ describe('CampaignController.loadBrief', () => {
   it('returns a "none" status when campaign-service has no brief for this slug', async () => {
     // The ordinary first-time case: the user has not generated a brief yet, so campaign-service
     // returns nothing. This is not an error, just an empty result that tells the UI "generate one".
-    loadBrief.mockResolvedValue({ status: 'none', briefId: null, brief: null, approved: false });
+    loadBrief.mockResolvedValue({ status: 'none', briefId: null, brief: null, etag: null, approved: false });
 
     await controller.loadBrief(buildLoadReq(), res, next);
 
     expect(loadBrief).toHaveBeenCalledWith(expect.any(Object), 'kubecon-eu-2026', 'tlf');
     expect(next).not.toHaveBeenCalled();
-    expect(res.json).toHaveBeenCalledWith({ status: 'none', briefId: null, brief: null, approved: false });
+    expect(res.json).toHaveBeenCalledWith({ status: 'none', briefId: null, brief: null, etag: null, approved: false });
   });
 
   it('returns a "loaded" status with the brief when campaign-service reconstructs it successfully', async () => {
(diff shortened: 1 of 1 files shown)


mrautela365:
### Pre-review sims — clean, and one finding worth recording

Ran the bot and dealako reviewer simulations before human review. Both **APPROVED**, 0 blocking / 0 minor between them (2 Low nits from the bots pass, addressed below).

**Both independently re-ran the mutation** rather than trusting the commit message: patching `res.json(result)` to drop `etag` fails exactly the three arms (`3 failed | 148 passed`), and the same mutation against `origin/main`'s spec leaves all 151 green. That second half is the claim I'd expected to be taken on faith.

**A stronger mutation than the one I ran.** The dealako sim passed a *wrong* etag (`W/"WRONG"`) rather than dropping the field — all three arms still fail. So these assertions bind on the **value**, not merely on key presence. That is a better check than mine and I've adopted it.

**Class sweep came back clean.** Every other `CampaignBriefLoadResult` construction site already carries `etag`: the `off` arm at `campaign.controller.spec.ts:270`, ~24 sites in `planning-tab.component.spec.ts`, and the service spec's `none`/`unreadable` `toEqual` assertions.

One site deliberately omits it — `planning-tab.component.spec.ts:300`, the rolling-deploy test whose comment reads *"The fixture omits `etag` deliberately and casts, because a well-typed literal cannot express the shape an older pod actually sends."* Adding it there would delete the test's premise. Left alone.

**On the ETag shape nit** (bots, Low): the service emits the strong form `"<version>"` (`campaign-service.service.ts:380-384`), and strong outnumbers `W/` 176:47 repo-wide. But every existing etag fixture *in this spec file* uses `W/"..."`, so these values match their neighbours. Changing only my two lines would be the named-line fix rather than the class — which is what that finding itself warns against. Left as-is; a repo-wide normalisation is a separate sweep.

mrautela365:
@dealako — this closes the `[minor]` you left on #1837 (controller spec fixtures omitting the now-required `etag`). That PR merged ~14 seconds after your comment, so the fix needed its own PR.

All three arms, fixture **and** assertion, six lines. Values match what the service actually returns: `null` for `none`, `found.etag` for `loaded` and `unreadable`.

Mutation-verified both directions — dropping `etag` before `res.json` fails all three (`3 failed | 148 passed`), and the same mutation against `origin/main`'s spec leaves all 151 green, which is what shows the assertions were previously vacuous.

Swept the class: every other `CampaignBriefLoadResult` site already carries `etag`. The one that omits it (`planning-tab.component.spec.ts:300`) does so deliberately — it is the rolling-deploy test for the shape an older pod sends — and is left alone.

11 checks green, 0 unresolved threads.

dealako:
Hi @mrautela365 👋 — thanks for circling back to close the `[minor]` from #1837 in its own PR.

**Overall impression:** Tight, surgical follow-up. Six lines across three `loadBrief` arms (`none` / `loaded` / `unreadable`), each updating both the mocked `loadBrief` result and its matching `res.json` assertion so the fixtures line up with what `campaign-service.service.ts` now returns. I traced the service and the controller: `loadBrief` returns `etag: null` for `none` (`:634`) and `found.etag` for `unreadable`/`loaded` (`:666-667`), and the controller is a straight `res.json(result)` pass-through (`campaign.controller.ts:700`), so these assertions genuinely exercise the pass-through. Using non-null `W/"7"` / `W/"9"` on the two carried arms is the right call — a `null` everywhere would pass equally against a controller that dropped the field, proving nothing. The mutation-based verification (destructure `etag` out → 3 fail) is exactly the evidence that makes this change trustworthy.

**Issue count:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0
- ❔ Question: 0

**Bot reconciliation:** Copilot generated no comments; Cursor flagged Low Risk (test-only, no production behavior change); CodeRabbit auto-review skipped. Agree with all three — nothing outstanding.

**Final decision: ✅ Approved**

Clean, well-scoped, and verified by mutation rather than a green suite. Nice work.

dealako (approved):
Approved — test-only follow-up that correctly aligns the three loadBrief fixtures and assertions with the service's actual etag output, verified by mutation. No findings.

PR #1586, opened 2026-08-17
feat(campaigns): reach pause and resume from the optimize tab

mrautela365:
Makes campaign pause/resume reachable from the product. Before this, stopping an overspending Google Ads or LinkedIn campaign meant logging into the ad platform directly.

Jira: [LFXV2-3224](https://linuxfoundation.atlassian.net/browse/LFXV2-3224)

## Read this first: it stacks on two unmerged branches

The button needs both halves, and neither has merged, so this branch **carries them**:

| Commits | Ticket | Also in |
|---|---|---|
| toggle server path | LFXV2-3226 | **#1581** (open) |
| campaign list read | LFXV2-3099 | not yet opened |
| the UI itself | LFXV2-3224 | this PR only |

If #1581 merges first, rebasing drops that duplication. **Reviewers: focus on the last four commits** — the rest is already under review on #1581.

Deliberate call: waiting on two approvals meant shipping nothing, and integrating early is what caught the defect below.

## What integration caught that neither half would have

`CampaignIndexDoc` had **no `etag`**, so the toggle had no `If-Match` to send. The index stores `version`; campaign-service's ETag is `"<version>"` — *with quotes* (`briefETag`). Merged separately, the button would have failed with 412s that read like concurrent edits.

Derived once in the service, where the wire contract lives, rather than leaving each caller to rediscover a quoting rule only Go source states.

## Three distinctions the UI is careful about

Each because collapsing it would state something false about money:

- **`null` vs `[]`** — `null` is "not loaded, or the read failed" and renders **nothing**; `[]` renders an empty state. Collapsing them asserts a brief has no campaigns on the strength of a failed read, for campaigns that may be spending.
- **`possiblyStale`** — indexing is asynchronous, so an empty list moments after a create means "not visible yet". The copy says that rather than "no campaigns".
- **`serviceStatus` over `newStatus`** — pausing a `created_degraded` campaign pauses it upstream while deliberately leaving the row's status unchanged. Echoing the request would claim a transition the service declined to record.

The failure path is **not optimistic**: on error the row keeps its old status and says the pause did not happen. Claiming success would stop someone watching a campaign that is still running.

## Verification

`yarn build`, `yarn lint:check --force`, `./check-headers.sh`, `tsc --noEmit` on all three tsconfigs. **1383 server tests + 14 app-side tests.**

Mutation-tested — each fix pinned by a test that fails when reverted:

| Mutation | Fails |
|---|---|
| collapse `null` into `[]` | null-renders-nothing |
[...]

PR #1885, opened 2026-08-27
feat(charts): complete the campaign-service cutover

mrautela365:
Completes the campaign-service cutover. Enables `LFX_CUTOVER_CAMPAIGN_SERVICE_STATUS_TOGGLE` and `LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE`, the last two of four flags.

```diff
  LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE:
-    value:
+    value: "true"
  LFX_CUTOVER_CAMPAIGN_SERVICE_STATUS_TOGGLE:
-    value:
+    value: "true"
```

`_DEMAND_GEN` stays off — Google creates are Search-only.

## The ordering is a deploy constraint, not a merge one

This PR briefly split `_CREATE` into a follow-up, on the theory that shipping both flags together would open a rollout window: with `replicaCount: 3`, `maxSurge: "100%"` and non-sticky requests, a new pod can mint a UUID campaign while an old pod — not yet carrying `_STATUS_TOGGLE` — refuses that campaign's pause at `campaign.controller.ts:1156`. Pause is the primary cost-control lever, so an intermittent 400 there is not cosmetic.

**The window is real. Splitting the PR does not close it.** No workflow in this repo deploys the chart — merging to `main` builds an image and nothing else. Both flags live in the same `values.yaml` and reach the cluster in whichever release someone rolls next, whether they merged in one PR or four. The split bought a review cycle and no safety.

So the constraint is documented where it can actually be honored. `README.md` now carries the concrete staging mechanism and `values.yaml` says plainly that the chart ships every flag on and must not be rolled out in a single release:

```bash
# release 1 — JOBS only
helm upgrade ... \
  --set environment.LFX_CUTOVER_CAMPAIGN_SERVICE_BRIEFS.value="" \
  --set environment.LFX_CUTOVER_CAMPAIGN_SERVICE_STATUS_TOGGLE.value="" \
  --set environment.LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE.value=""
```

Then drop one override per release, confirming every pod has rolled before the next:

```text
JOBS  →  BRIEFS  →  STATUS_TOGGLE  →  CREATE
```

`_STATUS_TOGGLE` goes before `_CREATE` because it is **inert** until CREATE produces UUID-keyed campaigns — a campaign only has a UUID if campaign-service created it — so rolling it early changes nothing observable. That is what makes it safe to stage first, and makes the reverse order the only one that costs anything. On rollback, reverse it: `_CREATE` off first, `_JOBS` last.

## A safety claim of mine that was wrong

An earlier revision argued in `README.md`, `values.yaml` **and** `server-feature-flag.helper.ts` that an overlapping rollout was *safe* because the two id spaces are disjoint. That proves no request is **misrouted** — narrower, and not the failure at issue. A refusal is well-formed and still a failure.

The helper was the source the chart docs were paraphrasing; fixing only the copies would have left it for the next reader. All three now state what disjointness actually establishes and give the ordering as the consequence.

## Operational preconditions

Audited both repos, so these are the complete set:

- **A project needs a connection, or the LF system row.** campaign-service reads credentials from its own encrypted tables, never from `GADS_*` env vars. On dev the LF Google Ads system row is healthy (`system:linuxfoundation`, `active`, account `8666746580`), and `tlf`'s stale tombstone was removed — `tlf` now resolves via the fallback.
- **A disconnected project fails closed rather than falling back.** A soft-deleted row is a statement, not an absence.
- Neither fails at deploy time; both surface per-campaign at dispatch.

At test time, **include keywords** — optional, but a Search campaign without them can never serve. Demand Gen no longer needs unchecking: `34305fd4f` changed the form default to Search-only, since Search+DemandGen together is refused server-side.

## Testing

- `helm template` default render: `_JOBS`/`_BRIEFS`/`_STATUS_TOGGLE`/`_CREATE` all `"true"`, `_DEMAND_GEN` name-only.
- `helm template --set ...CREATE.value=""` renders CREATE name-only — verified the staging mechanism works rather than assuming it. `isServerFeatureEnabled` returns false on empty input.
- `helm lint` clean. 2090 server + 1253 app tests pass.
- Commits DCO-signed and GPG-signed.


charts/lfx-self-serve/README.md
```diff
@@ -194,19 +194,50 @@ changing a value here rather than by shipping a revert.
...
+primary cost-control lever on a mis-targeted campaign.
+
+### Before enabling `..._CREATE`
+
+`..._CREATE` changes where ad credentials come from: campaign-service reads them from its own
+encrypted connection tables, never from this application's `GADS_*` / `LINKEDIN_*` environment
+variables. Two things must be true per project, and neither fails at deploy time — both surface
+per-campaign at dispatch:
+
+- **A connection must exist.** A project with none falls back to the LF-owned system account, so
```
dealako: [minor] `..._CREATE` precondition contradicts itself

Issue: The bullet asserts "A connection must exist," but its own body says a project with none "falls back to the LF-owned system account" — i.e. no connection is required for dispatch to succeed; it only changes whose ad account is billed.

Proof: The two clauses in this bullet state opposite requirements. An operator cannot tell from this whether a per-project connection row is mandatory.

Why it matters: The reader can't determine the actual precondition — is a missing connection a hard failure or a (costly) fallback? The distinction decides whether spend lands on the project's account or the LF's.

Fix: Reframe as "a usable credential source must resolve": either a live project connection (spend on the project's account) or the LF system-account fallback (spend on the LF's account). Reserve the hard-failure language for the DISCONNECTED / soft-deleted case in the next bullet, which is the one that actually fails closed.

dealako (changes requested):
Hi @mrautela365 — thanks for the exceptionally well-documented cutover. The rollout reasoning (why STATUS_TOGGLE rides with CREATE, why the id-shape check makes an overlapping rollout safe, the disjoint-id-space argument) is clear and correct, and auditing both repos up front to surface the full precondition set rather than a trickle is the right call.

One substantive issue blocks approval: the STATUS_TOGGLE narrative in both the README and values.yaml is now factually stale and this PR flips the flag on while shipping it, so an operator reading it before a production routing flip is actively misinformed. Plus the merge state and draft status.

### Issue count
- 🔴 Blocking: 2 — stale STATUS_TOGGLE docs (README + values.yaml) claim no user-visible change when the flag now gates a live UI control; PR is a draft and `CONFLICTING`/`DIRTY` against `main`.
- 🟡 Minor: 1 — `..._CREATE` precondition wording contradicts itself (connection "must exist" vs. valid LF-account fallback).
- ⚫ Nit: 0
- ❔ Question: 0

No security or data-privacy findings — this is a Helm values + README change with no code paths, secrets, or PII in the diff.

### Reconciliation with Copilot
- **Agree** with Copilot on the stale pause-button passages (README ~200, values.yaml 548-551). Confirmed against code: `server-feature-flag.helper.ts` documents the flag "REACHES THE UI as of LFXV2-3224", and `optimization-tab.component.ts` reads `statusToggleEnabled` (`deploymentDisabled = !this.statusToggleEnabled()`) and calls `updateCampaignStatus`. The "does NOT give users a pause button" text is wrong as of this flip.
- **Agree** with Copilot on the `..._CREATE` precondition contradiction (README 234-240).

### Merge / draft
Still a draft ("must not merge until #1881 has rolled out to every pod"), and #1881 is already merged to `main` (it's the tip). Rebase onto `main`, drop the merge commit, retarget away from #1881's branch, and clear the `CONFLICTING` state before this can land. Net diff vs `main` is already just these two files, so the rebase should be mechanical.

### Decision
🔴 **Needs changes before approval** — reconcile the STATUS_TOGGLE docs with the shipped UI behavior, fix the precondition wording, and resolve the merge conflict / draft status.

dealako:
## Review summary — @mrautela365

Thorough, well-argued cutover doc. The rollout ordering (JOBS → BRIEFS → CREATE, STATUS_TOGGLE alongside CREATE), the disjoint-id-space safety argument, and pre-auditing both repos for the full precondition set are all solid. One substantive doc-accuracy issue blocks approval, plus merge/draft state.

### Issues
- 🔴 **Blocking: 2**
  1. **Stale STATUS_TOGGLE docs** — README:275-279 and values.yaml:548-551 say the flag gives "NO pause button" and "nothing changed on screen." This PR flips it to `"true"`. Code disagrees: `server-feature-flag.helper.ts` documents it "REACHES THE UI as of LFXV2-3224", and `optimization-tab.component.ts` reads `statusToggleEnabled` and calls `updateCampaignStatus` to render live per-row Pause/Resume controls. Operators reading this during a production flip are misinformed about the primary cost-control lever. *(Anchored these to changed lines wasn't possible — they're unchanged context; see file:line above. Agrees with Copilot.)*
  2. **Draft + merge conflict** — PR is a draft ("must not merge until #1881 has rolled out") and is `CONFLICTING`/`DIRTY` vs `main`. #1881 is already merged (tip of `main`). Rebase onto `main`, drop the merge commit, retarget off #1881's branch. Net diff vs `main` is already just these two files, so it should be mechanical.
- 🟡 **Minor: 1**
  - `..._CREATE` precondition (README:237) contradicts itself: "A connection must exist" vs. valid LF-account fallback when none exists. Reframe as "a usable credential source must resolve." *(inline comment posted; agrees with Copilot.)*
- ⚪ Nit: 0
- ❔ Question: 0

No security or data-privacy findings — Helm values + README only, no code paths, secrets, or PII in the diff.

### Reconciliation with bots
Agree with both Copilot findings (stale pause-button passages; self-contradicting `_CREATE` precondition), verified against the code. CodeRabbit auto-review was skipped on this repo.

### Decision
🔴 **Needs changes before approval** — reconcile STATUS_TOGGLE docs with the shipped UI (LFXV2-3224), fix the precondition wording, resolve the conflict, and take the PR out of draft.

charts/lfx-self-serve/values.yaml
```diff
@@ -470,7 +470,7 @@ environment:
   # Where neither a project row nor the fallback applies, the failure surfaces per-campaign rather
   # than at startup, so nothing warns you at deploy time.
   LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE:
-    value:
+    value: "true"
```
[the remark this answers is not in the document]
mrautela365: Fixed in `34305fd4f`, and you were right that this is the one that would have bitten first.

Verified the mechanism rather than taking it on trust: the form initialises both to `true` (`implementation-tab.component.ts:248-249`), `canSubmit` only blocks when **neither** is selected (`:759`), and `createCampaigns` refuses the pair with no legacy fallback (`campaign-service.service.ts:788`). So an untouched form produced a terminal failure rather than a campaign.

`includeDemandGen` now defaults to `false`. Search alone and Demand Gen alone are each servable — only the pair is not — so defaulting to the one that works means the untouched form submits, and choosing Demand Gen stays one click away.

Pinned by a test, and **verified by mutation**: restoring `[true]` fails the new case (`1 failed | 149 passed`). Pinned as a *default* rather than as a rule, because that is exactly what regressed — someone restoring `[true]` for symmetry would reintroduce the dead end silently.

charts/lfx-self-serve/README.md
mrautela365: Fixed in `34305fd4f` — the bullet did contradict itself, and your reframing is the one I took.

It now reads as **"a usable credential source must resolve"**: either a live project connection, in which case the spend lands on the project's own ad account, or the LF system-account fallback, in which case the spend lands on the LF's. A project connection is therefore *not* mandatory for dispatch to succeed — it decides who is billed.

The hard-failure language moved to the DISCONNECTED bullet, which is the case that actually fails closed: a soft-deleted row is a statement, not an absence, so the fallback is refused rather than used.

charts/lfx-self-serve/README.md
```diff
@@ -195,34 +195,49 @@ changing a value here rather than by shipping a revert.
...
-| `environment.LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE`        | Creates campaigns through campaign-service instead of the per-platform Express services                                             | No       | off      |
+| `environment.LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE`        | Creates campaigns through campaign-service instead of the per-platform Express services                                             | No       | `"true"` |
 | `environment.LFX_CUTOVER_CAMPAIGN_SERVICE_DEMAND_GEN`    | Allows Demand Gen Google campaigns. Requires a campaign-service that understands `googleAdsConfig.channel` (LFXV2-3257) — see below | No       | off      |
-| `environment.LFX_CUTOVER_CAMPAIGN_SERVICE_STATUS_TOGGLE` | Serves campaign pause/resume from campaign-service, which is what makes Google Ads and LinkedIn pausable — see below                | No       | off      |
+| `environment.LFX_CUTOVER_CAMPAIGN_SERVICE_STATUS_TOGGLE` | Serves campaign pause/resume from campaign-service, which is what makes Google Ads and LinkedIn pausable — see below                | No       | `"true"` |
```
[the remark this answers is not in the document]
mrautela365: Fixed in `34305fd4f` — correct, those passages went stale when #1586 landed the UI half.

I verified before changing them: the Optimize tab does now propagate `statusToggleEnabled` and call `updateCampaignStatus`. Both the README section and the `values.yaml` comment said the flag enables the server path only and that nothing would change on screen. That was true when written and is not now.

They now say it is **user-visible** — operators get a working pause control — and add the qualifier that replaces the old one: it stays inert for campaigns the **legacy** path created, because pausing needs a campaign UUID and only a campaign-service create mints one. So the control appears for campaigns made after `..._CREATE` is enabled, not retroactively.

charts/lfx-self-serve/README.md
```diff
@@ -195,34 +195,49 @@ changing a value here rather than by shipping a revert.
...
+
+`..._CREATE` changes where ad credentials come from: campaign-service reads them from its own
+encrypted connection tables, never from this application's `GADS_*` / `LINKEDIN_*` environment
+variables. Two things must be true per project, and neither fails at deploy time — both surface
+per-campaign at dispatch:
+
+- **A connection must exist.** A project with none falls back to the LF-owned system account, so
+  the spend lands on the LF's ad account rather than the project's.
+- **The project must not have DISCONNECTED.** A soft-deleted row is a statement, not an absence:
+  the fallback is refused and the dispatch fails closed.
```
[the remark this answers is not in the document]
mrautela365: Superseded — this thread is against an earlier revision of the README, and the contradiction it names was fixed in `34305fd4f`.

The bullet no longer asserts that a connection must exist and then describe the fallback that makes one unnecessary. It now reads as "a usable credential source must resolve" — either a live project connection (spend on the project's account) or the LF system-account fallback (spend on the LF's) — with the hard-failure language reserved for the DISCONNECTED case, which is the one that genuinely fails closed.

Same finding as dealako's `[minor]` on the current revision; answered in that thread too.

mrautela365:
Picked up the suppressed Copilot comments (they sit in review bodies, so `unresolved=0` did not cover them). Both were right, and both named source comments I had not touched — the README fix in 3a6bf7125 left them behind.

Fixed in d07546be9:
- `campaign.interface.ts:1701` and `optimization-tab.component.ts:93` — both said "the chart leaves that flag unset by default", which this PR makes false.
- `campaign.controller.ts:453` — said CREATE is "still the default" off.

Rewritten to state what survives the flip rather than just deleting the claim: the flags are read per request from `process.env`, so a values override or a not-yet-rolled deployment still produces the off case each comment exists to explain. That is also why `statusToggleEnabled` stays on the list payload instead of becoming redundant — the client still cannot infer the server's flag.

mrautela365:
Taken, in 65248fb88 — `_CREATE` is back to unset and this PR now ships `_STATUS_TOGGLE` alone. Retitled to match.

Verified the mechanism rather than taking it on faith: `replicaCount: 3` with `maxSurge: "100%" / maxUnavailable: 0` (values.yaml:4-13) means old and new pods serve concurrently, requests are not sticky, and `campaign.controller.ts:1156` refuses a UUID toggle whenever the flag is off on the receiving pod. So the window is real, and on a spending campaign an intermittent 400 on pause is not cosmetic.

The fix is free for the reason my own ordering note already recorded without drawing the conclusion: `_STATUS_TOGGLE` is inert until `_CREATE` mints UUID-keyed campaigns, so shipping it first changes nothing observable. That makes it the safe half to ship alone, and makes the reverse order the only one that costs anything.

I also corrected the safety claim this PR was carrying in both `values.yaml` and `README.md`. It said "an overlapping rollout is SAFE here" and argued it from the id spaces being disjoint — which proves no request is **misrouted**, a narrower guarantee, and not the failure raised here. A refusal is well-formed and still a failure. Both now state what the disjointness actually establishes, and give the ordering as the consequence.

`helm template` confirms `_CREATE` renders name-only with no `value:`, identical to `_DEMAND_GEN` — the shape the code reads as off. `helm lint` clean.

`_CREATE` follows in its own PR once this converges.

mrautela365:
Both taken in 6edcb93a4, and both were mine to have caught.

**README:229** — right, it still said STATUS_TOGGLE "ships alongside" CREATE and called the overlap safe, directly contradicting the corrected passage further down. Rewritten to state the ordering, and to keep the disjoint-id-space argument scoped to what it actually proves (no *misrouting*) rather than to overall safety.

**campaign.controller.ts:454** — also right, and worth noting how it happened: that comment was written *one commit earlier*, in d07546be9, specifically to fix a stale default claim. The split in 65248fb88 then made my replacement wrong in the opposite direction. Now reads "still the chart default", which is true again.

That pattern — each fix falsifying the next comment — is why I swept the whole claim class repo-wide this time instead of the two lines named, and it found two more neither of you flagged:

- `optimization-tab.component.spec.ts:532` — repeated "the chart leaves it unset" about STATUS_TOGGLE, which this PR sets `"true"`.
- `campaigns.component.ts:1404` — said the brief cutover is dark "by default", untrue since #1881 shipped BRIEFS `"true"`.

Verification: re-ran the sweep after fixing and no assertion survives; `helm template` confirms CREATE still renders name-only (no `value:`), identical to DEMAND_GEN; 1253 app tests pass.

mrautela365:
@dealako your CHANGES_REQUESTED here was on `b9c5d5c5f` (15:08), before the last three commits — the PR has changed shape since, so it is worth a fresh look rather than a re-read of the old diff.

What moved:

- **Scope narrowed.** `_CREATE` is back to unset; this now ships `_STATUS_TOGGLE` alone, and is retitled to match. Copilot caught that shipping both in one rollout opens a real window — `replicaCount: 3`, `maxSurge: "100%"`, non-sticky requests, so a new pod can mint a UUID campaign while an old pod refuses its pause at `campaign.controller.ts:1156`. `_CREATE` follows once this converges.
- **A safety claim of mine was wrong and is corrected.** README and `values.yaml` both said an overlapping rollout was "safe" and argued it from the disjoint id spaces. That proves no request is *misrouted* — narrower, and not the failure at issue. A 400 is well-formed and still a failure.
- **Swept the stale-claim class repo-wide** rather than the flagged lines, after two rounds where each fix falsified the next comment. Four sites total, two of which no reviewer had flagged.

Verification: `helm template` shows `_CREATE` rendering name-only (no `value:`), identical to `_DEMAND_GEN`; `helm lint` clean; 1253 app tests pass.

mrautela365:
Fixed in a7fb2a2b9 — and you were right that the sweep had not gone far enough. My last pass used too narrow a pattern and missed an entire section:

- `values.yaml:385` and `README:202` both said CREATE was already ON and that the JOBS "no UUID job can exist" premise had **expired**. With CREATE unset again the premise holds today, so both now describe it as expiring when CREATE ships rather than as already gone.
- **Two enable-order diagrams still read as three steps.** One of them my previous grep missed outright, because `values.yaml:450` writes `->` where `README:213` writes `→`. Both now show `JOBS → BRIEFS → STATUS_TOGGLE → CREATE`.
- Two "three flags above" counts, now that there are four.
- `README:362` was written in the past tense about a cutover that has not happened yet.

PR title and body rewritten too — the body still described enabling both flags, and still told the tester to uncheck Demand Gen, which `34305fd4f` made unnecessary by defaulting the form to Search-only.

Verification: `helm template` renders `_CREATE` name-only with no `value:`, identical to `_DEMAND_GEN`; `helm lint` clean; re-swept for both arrow forms and the step count with a self-tested query, and nothing false survives.

charts/lfx-self-serve/README.md
```diff
@@ -195,34 +195,52 @@ changing a value here rather than by shipping a revert.
...
-`createCampaigns` returns `enabled: true` with _"its brief has not been saved yet"_ rather than
-falling through to the legacy creator, so the user gets a dead end rather than a working campaign.
+They could not share a rollout. `createCampaigns` gates on all three together, so during a mixed
+deployment a brief save can land on a BRIEFS-off pod and answer `enabled: false` with no persisted
+brief id, while the create that follows lands on a pod where all three are on. That create is
+refused terminally — `createCampaigns` returns `enabled: true` with _"its brief has not been saved
+yet"_ rather than falling through to the legacy creator, so the user gets a dead end rather than a
+working campaign.
+
+`..._STATUS_TOGGLE` ships alongside `..._CREATE`. Its rollout overlap is safe where the others' is
```
dealako: [blocking] Header prose still describes the pre-resequence plan and contradicts the corrected body below

Issue: This sentence says STATUS_TOGGLE "ships alongside `..._CREATE`" and that "its rollout overlap is safe." But this PR ships STATUS_TOGGLE FIRST with CREATE OFF, and this same document retracts the "overlap is safe" framing at line 296 ("a narrower guarantee than 'an overlapping rollout is safe', which is what an earlier revision of this section claimed") and prescribes the opposite at line 301 ("**Hence the ordering.** Enable `..._STATUS_TOGGLE` first and let it converge, then enable `..._CREATE`"). Lines 205 ("With `..._CREATE` on... the earlier argument no longer applies") and 209 ("The cutover landed as three flags... JOBS → BRIEFS → CREATE") carry the same stale present/past-tense framing and omit STATUS_TOGGLE.

Proof: values.yaml ships `LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE: value:` (unset → `isServerFeatureEnabled` returns false → OFF) and `LFX_CUTOVER_CAMPAIGN_SERVICE_STATUS_TOGGLE: value: "true"`. So CREATE is dark and STATUS_TOGGLE is the one shipping. Line 222 directly contradicts lines 296/301 of the same file.

Why it matters: This is the top of the section, read first. An operator concludes CREATE is already live and STATUS_TOGGLE co-ships with it — the exact misconception the re-sequencing exists to prevent (STATUS_TOGGLE must converge before CREATE or a new pod mints a UUID campaign an old pod refuses to pause).

Fix: Rewrite lines ~202-222 to the shipped state: JOBS and BRIEFS on, STATUS_TOGGLE on and shipping first, CREATE off and landing in a later PR after STATUS_TOGGLE converges. Make the JOBS hazard at 205 conditional ("once `..._CREATE` is enabled...") since no UUID job can exist while CREATE is off, and delete/repair the "ships alongside `..._CREATE` / rollout overlap is safe" sentence to match the retraction at 296/301.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -450,7 +450,8 @@ export class CampaignController {
     // The client therefore sees one `CampaignJobStatus` either way, with `result` set on the
     // in-process path and `platformResults` on the campaign-service path.
     //
-    // The flag is necessary but NOT sufficient to route. With CREATE off — still the default —
+    // The flag is necessary but NOT sufficient to route. With CREATE off — which the chart no
```
dealako: [minor] Default-state clause is false, and was changed away from the correct wording this round

Issue: "With CREATE off — which the chart no longer defaults to, but which any override or un-rolled deployment still produces —". The chart still defaults CREATE off (values.yaml ships it unset). This round changed the line from the correct "With CREATE off — still the default —" to this incorrect version, presumably when the plan still had CREATE going on; the reversal to CREATE-off left it stale.

Proof: values.yaml `LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE: value:` (empty); `server-feature-flag.helper.ts` `isServerFeatureEnabled`: `if (!raw) return false;` — unset → CREATE off. CREATE-off IS the current chart default, so "no longer defaults to" is wrong.

Why it matters: The surrounding routing logic (job_ ids failing campaign-service's `FormatUUID` check) is correct; only this parenthetical misstates the deployed baseline, which could mislead a maintainer reasoning about routing from the comment. Low blast radius.

Fix: Revert to the accurate baseline, e.g. "With CREATE off — which is the chart's current default, and which any override or un-rolled deployment also produces —". Keep the rest of the sentence.

dealako (changes requested):
Follow-up review — @mrautela365, this is a strong revision. You addressed all three of my prior findings and verified each mechanism rather than taking it on trust.

### 👏 Nice work
- **STATUS_TOGGLE stale docs (prior blocking) — resolved.** README:278 and the values.yaml STATUS_TOGGLE comment now say plainly it is user-visible (Optimize tab propagates `statusToggleEnabled` and calls `updateCampaignStatus`, #1586), matching `optimization-tab.component.ts` and `server-feature-flag.helper.ts`.
- **`_CREATE` precondition wording (prior minor) — resolved.** Now reads "A usable credential source must resolve" with the fail-closed language correctly scoped to the DISCONNECTED case — exactly the reframing I suggested.
- **Draft + merge conflict (prior blocking) — resolved.** Out of draft, `MERGEABLE`.
- **The re-sequencing itself is the right call.** Shipping STATUS_TOGGLE first and letting it converge before CREATE closes the real `maxSurge: "100%"` window where a new pod mints a UUID campaign an old pod would refuse to pause (400 on the primary cost-control lever). The corrected analysis at README:296-307 is clear and correct.
- **`includeDemandGen` default flip — verified real and covered.** Defaults to `false` (`implementation-tab.component.ts:255`), untouched form submits `['search']` only, and the new non-circular spec test at `implementation-tab.component.spec.ts:671` asserts it. Good catch that this was the one that would have bitten first.

### What still blocks
The re-sequence corrected the *body* of the rollout docs but left the *header prose and two comments* asserting the old plan (CREATE on / ships alongside). In a runbook whose whole purpose is rollout safety, the stale half is the part read first, and it contradicts the corrected half in the same files.

### Issue count (this round)
- 🔴 **Blocking: 2**
  1. README:202-222 presents CREATE as already-on and STATUS_TOGGLE as "ships alongside CREATE" with "rollout overlap is safe" — contradicts CREATE=OFF and is explicitly retracted by the same doc at :296/:301. (inline)
  2. values.yaml:385 "CREATE below is now ON" — CREATE is unset/OFF in the same file. (body ref — unchanged line, can't anchor inline)
- 🟡 **Minor: 1** — campaign.controller.ts:453 "CREATE off — which the chart no longer defaults to": chart still defaults CREATE off; this line was actually changed *away* from the correct "still the default" this round. (inline)
- ⚫ **Nit: 1** — `campaign-service.service.ts:785` rationale comment still says `includeDemandGen` "defaults to true in the form"; that premise is now false after the default flip. Out of this round's changed set and CREATE is OFF, so no runtime impact — mentioning for when the file is next touched.

No security or data-privacy findings — flag defaults are fail-secure (default-deny parsing; STATUS_TOGGLE inert until CREATE mints UUIDs, and a UUID with the flag off is refused, never routed to legacy), `updateCampaignStatus` requires `project`/`briefId`/`etag` with an exact join (no cross-tenant reach), and logs carry ids/slugs/statuses only.

### Reconciliation with Copilot
Agree with Copilot's current-HEAD comments on all three: README "ships alongside" contradiction, values.yaml "now ON" pair, and the controller default-state clause. Its earlier-revision inline comments (pause button, precondition, Search+DemandGen default) are correctly marked superseded/resolved by your commits.

### Decision
🔴 **Needs changes before approval** — reconcile README:202-222 and values.yaml:385 with the shipped state (JOBS/BRIEFS/STATUS_TOGGLE on, CREATE off and lands later), and fix the controller:453 default-state clause. All three are doc-only; the code and rollout sequencing are sound.

dealako:

Strong revision. All three of my prior-round findings are resolved, and you verified each mechanism against the code rather than taking it on trust.

### 👏 Nice work
- **STATUS_TOGGLE stale docs (was blocking) → ✅ resolved.** README:278 and the values.yaml comment now state plainly the flag is user-visible (Optimize tab propagates `statusToggleEnabled`, calls `updateCampaignStatus`, #1586) — matches the code.
- **`_CREATE` precondition wording (was minor) → ✅ resolved.** Now "A usable credential source must resolve," fail-closed language scoped to the DISCONNECTED case. Exactly the reframing suggested.
- **Draft + merge conflict (was blocking) → ✅ resolved.** Out of draft, `MERGEABLE`.
- **Re-sequencing STATUS_TOGGLE ahead of CREATE is the right call** — it closes the `maxSurge: "100%"` window where a new pod mints a UUID campaign an old pod would refuse to pause (400 on the primary cost-control lever). The corrected analysis at README:296-307 is clear and correct.
- **`includeDemandGen` default flip verified real and covered** — defaults to `false` (`implementation-tab.component.ts:255`), untouched form submits `['search']` only, new non-circular spec test at `implementation-tab.component.spec.ts:671` asserts it.

### Still open (this round)
The re-sequence fixed the *body* of the rollout docs but left the *header prose and two comments* asserting the old plan. In a rollout-safety runbook, the stale half is what's read first, and it contradicts the corrected half in the same files.

- 🔴 **Blocking: 2**
  1. README:202-222 presents CREATE as already-on and STATUS_TOGGLE as "ships alongside CREATE / rollout overlap is safe" — contradicts CREATE=OFF and is retracted by the same doc at :296/:301. *(inline)*
  2. values.yaml:385 "CREATE below is now ON" — CREATE is unset/OFF in the same file. *(unchanged line, couldn't anchor inline — see here)*
- 🟡 **Minor: 1** — campaign.controller.ts:453 "CREATE off — which the chart no longer defaults to": chart still defaults CREATE off; this line was changed *away* from the correct "still the default" this round. *(inline)*
- ⚪ **Nit: 1** — `campaign-service.service.ts:785` still says `includeDemandGen` "defaults to true in the form" — false after the default flip. Out of this round's changed set, CREATE is OFF, no runtime impact; flagging for when the file is next touched.

No security or data-privacy findings — flag defaults are fail-secure, `updateCampaignStatus` requires `project`/`briefId`/`etag` with an exact join (no cross-tenant reach), logs carry ids/slugs/statuses only.

### Reconciliation with Copilot
Agree with Copilot's current-HEAD comments on all three (the "ships alongside" contradiction, the values.yaml "now ON" pair, the controller default-state clause). Its earlier-revision comments are correctly superseded by your commits.

### Decision
🔴 **Needs changes before approval** — doc-only. Reconcile README:202-222 and values.yaml:385 with the shipped state (JOBS/BRIEFS/STATUS_TOGGLE on, CREATE off, lands later) and fix controller:453. The code and rollout sequencing are sound.

mrautela365:
@dealako both of these are already fixed — your review landed on `6edcb93a4` at 16:08 and `a7fb2a2b9` went up at 16:07, so it raced the push by about a minute rather than either of us missing it. Every point you raised is addressed on the current head; quoting it so you do not have to re-derive:

**README:222** — the "ships alongside `..._CREATE`" / "rollout overlap is safe" sentence is gone (`git grep` returns nothing for either phrase). It now opens:

> `..._STATUS_TOGGLE` ships BEFORE `..._CREATE`, as its own rollout, and must be allowed to converge first. It cannot be **misrouted** during an overlap the way the others can […] But that guarantee is about misrouting only; a pod without the flag still refuses a UUID with a 400.

**README:209** — now four steps with the shipped state marked:

```text
JOBS  →  BRIEFS  →  STATUS_TOGGLE  →  CREATE
 ✅        ✅          this PR          next
```

**README:205** — made conditional exactly as you asked: "Once `..._CREATE` is enabled campaign-service mints UUID job ids, and the earlier 'no UUID can exist' argument stops applying".

**campaign.controller.ts:453** — reverted to the accurate baseline; your suggested wording and mine converged almost word for word: "With CREATE off — still the chart default, and the state every deployment is in until the CREATE cutover ships —".

Your read of the cause is right, too: I changed that line away from the correct wording in `d07546be9` while CREATE was still going on, and the resequence left it stale. That is the same failure mode twice, so `a7fb2a2b9` swept the claim class across both chart docs — including a second enable-order diagram at `values.yaml:450` that an earlier grep missed because it writes `->` where the README writes `→`.

PR title and body also rewritten to the shipped state. No code change needed for these two.

mrautela365:
Both fixed in b48fcf0af — and the helper one was the important catch, because it was the **source** the chart docs were copying, not another copy.

**`server-feature-flag.helper.ts:181`** — carried the original `ROLLOUT OVERLAP IS SAFE HERE` claim. Narrowed to misrouting, with the retraction stated explicitly:

> That is NARROWER than "an overlapping rollout is safe", which an earlier revision of this comment claimed. A refusal is well-formed and still a failure […] So this flag must not share a rollout with `CampaignServiceCreate`.

Had I fixed only the chart docs, the claim would have survived here for the next reader to copy again.

**`README:349`** — the numbered list still read JOBS → BRIEFS → CREATE. Now four steps, with the reason attached to step 3 so the ordering is not just asserted:

> 3. Then **STATUS_TOGGLE**, and let it converge before the next step. It is inert until CREATE exists, so this step changes nothing observable — that is what makes it safe to roll alone, and enabling it *after* CREATE is the ordering that opens a window where a new pod mints a UUID campaign an old pod refuses to pause.

This is the third round on the same claim class, so I widened the sweep from the two chart files to the whole repo and every arrow form. Remaining matches for "overlapping rollout is safe" are the three retractions quoting the old wording; the only ordering lists are the four-step ones. `helm lint` clean, 1253 app tests pass.

mrautela365:
All three taken in 642b1849f — and the sweep for the same claim class found **eight more** sites that no reviewer had named. Thirteen in total.

**Your three:**
- `campaign-service.service.ts:400` and its spec at `:2381` — both said flag-off is "a default deployment".
- `campaigns.component.ts:1417` — you were right that I fixed `:1405` and left `:1417` in the same function stating the opposite default. Same-function contradiction, one commit old.
- `implementation-tab.component.ts:255` → the demand-gen guard docs. Your read was exactly right: the guard is still necessary, only its **reason** changed. Verified before rewriting — `implementation-tab` builds `campaignTypes` from the form regardless of platform (`:1327-1338`) and restores `includeDemandGen` from a saved draft (`:1611`), so it now covers **retained state** (ticked then Google deselected, or a draft saved under the old default) rather than an untouched form. Reworded in both `campaign-service.service.ts:783` and its spec at `:1862`.

**The eight the sweep found:** `campaigns.component.spec.ts` ×2, `campaigns.component.ts:86`, `campaign.controller.spec.ts` ×2, `campaign.controller.ts:514`, `campaign.interface.ts` ×2 — all asserting a BRIEFS-off default that #1881 falsified.

I reworded rather than deleted, since every one of these comments explains something real. The flags are read per request from `process.env`, so an override or un-rolled pod still produces the off case. Most now read "an ordinary deployment state, not a fault" — true regardless of what the chart ships, so the next default flip cannot falsify them again.

That last part is the actual lesson from four rounds on this: the wording was pinned to a value that changes. Verification: re-swept clean, 2090 server + 1253 app tests pass.

mrautela365:
**Scope change: this PR now completes the whole cutover.** `_CREATE` is folded back in (7551ebf7b), so there is no fourth PR to chase. Retitled accordingly.

I split `_CREATE` out earlier this round to stage the rollout. That was the wrong instrument, and I should have checked before splitting: **no workflow in this repo deploys the chart** — merging to `main` builds an image and nothing else. Both flags live in the same `values.yaml` and reach the cluster in whichever release someone rolls next, whether they merged in one PR or four. The split cost a review cycle and bought no safety.

The rollout window itself is real, so it is now documented where it can actually be honored:

- `README.md` gives the concrete staging mechanism — deploy with the not-yet-due flags overridden `""`, drop one override per release, confirm convergence between.
- `values.yaml` states that the chart ships every flag on and must not be rolled out in one release.

Verified the mechanism rather than asserting it: `isServerFeatureEnabled` returns false on empty input, and `helm template --set environment.LFX_CUTOVER_CAMPAIGN_SERVICE_CREATE.value=""` renders it name-only — the shape the code reads as off. Default render has all four on.

`helm lint` clean; 2090 server + 1253 app tests pass.

mrautela365:
Rollback finding fixed in 260681604 — both bots were right, and it was a real operational gap rather than a wording nit.

The summary said "reverse it", which treats `_STATUS_TOGGLE` as symmetric with `_JOBS`. It is not. `_JOBS` has a drain condition — outstanding polls finish and then it is safe to disable. `_STATUS_TOGGLE` has none: a UUID is permanent, `campaign.controller.ts:1156` refuses a pause for any UUID while the flag is off, and turning CREATE off stops only NEW campaigns. So disabling it strands every campaign-service campaign that already exists, with no pause control, while it may still be spending.

Recorded in all three places that describe the sequence — `README.md`, `values.yaml`, and `server-feature-flag.helper.ts`. The helper is the source the chart docs paraphrase, so fixing only the copies would have left it for the next reader.

Also two claims this PR's own flag flip falsified: `values.yaml:385` still said "CREATE below is still unset, so no UUID job id exists yet", and `README:202` still spoke of CREATE being enabled in the future tense.

---

**On the CREATE-ships-with-STATUS_TOGGLE finding** (@cursor High Severity, and Copilot's suppressed comment at `values.yaml:489`) — the mechanism is real and I am not disputing it. A single `helm upgrade` from these values turns both on at once and opens the pause window.

Where I differ is on the remedy. Splitting the PR does not close it: **no workflow in this repo deploys the chart** — merging to `main` builds an image and nothing else. Both flags live in the same `values.yaml`, so they reach the cluster in whichever release someone rolls next, whether they merged in one PR or four. That was the original scoping rationale and it does not hold; I verified it before consolidating.

Cursor's real point stands though: *"asking operators to override the value off is unenforced."* That is true, and it is a genuine weakness of the doc-only approach. The alternatives are a chart-level guard that refuses to render both flags on without an explicit acknowledgement value, or leaving CREATE unset in the chart and enabling it per-environment.

That is a design decision about how this deploys, so I am raising it with @mrautela365 rather than picking one here. Leaving both threads open until it is settled.

mrautela365:
Helper inconsistency fixed in f5876beba — good catch, and it was a contradiction I introduced myself: I added the STATUS_TOGGLE warning at `:194` last commit without checking what the `CampaignServiceCreate` doc 100 lines above already claimed. One file, two instructions.

The original sentence is **true about job polling** — a poll is answered by whichever system owns that job id, so a CREATE-on pod creating and a CREATE-off pod polling really does work. It just says nothing about the rest of a campaign's life. Scoped rather than removed:

> That id-shape distinction is what makes an overlapping rollout safe FOR JOB POLLING, and only for that […] It says NOTHING about the rest of a campaign's life. The same overlap mints a UUID campaign that a pod without `CampaignServiceStatusToggle` refuses to pause.

Swept for other unscoped versions of the claim; the remaining matches are either the JOBS flag (correctly scoped to job ids) or the retractions quoting the old wording. 2090 server tests pass.

The Demand Gen thread I am leaving open — it is a real UI bug, not a doc issue, and it needs a scope decision. Details in a separate reply.

mrautela365:
Fixed in 110791d8f. You were right, and the failure was worse than a dead-end control: **the two refusals contradict each other.** Picking Search AND Demand Gen is refused with *"Deselect one and create it"*; following that advice lands on the capability refusal at `campaign-service.service.ts:827` saying Demand Gen isn't available at all. Our own error message walks the user into a second one.

The capability was server-only, so the client had no way to know. Now reported the way `statusToggleEnabled` already is — on the brief campaign list, read per request from the environment, defaulting `false` on every error and pre-load path.

**Hiding the control alone was not enough.** The draft restore at `:1629` writes `includeDemandGen` directly, so a draft saved under the old `true` default would put a hidden `true` on the form and submit it straight into the refusal. That path is gated too. I also mirrored all **four** signal set-sites rather than just the success arm — both reset arms and the error arm — otherwise the value goes stale across a foundation switch into a deployment that can't serve it.

**Mutation-verified, all three arms:**

| Mutation | Result |
|---|---|
| server reports  unconditionally | 3 tests fail |
| template renders unconditionally (`@if (true)`) | 2 tests fail |
| draft guard removed | 1 test fails |

The paired positive cases matter here: without "preserves a restored draft selection where supported", a guard of plain `includeDemandGen: false` would pass every negative test while breaking the feature wherever it *is* enabled.

Two things the tests taught me that are worth recording, since both would have produced a test that could not fail: the seeding effect depends only on `briefData` and reads `draft` inside `untracked()`, so the draft must be set **first** — set the brief first and the restore never runs, and the negative test passes for the wrong reason. And `applyDraft` calls `replaceCopyArray` on `headlines`/`descriptions` unconditionally, so a minimal draft throws before reaching the checkbox.

2090 server + 1258 app tests pass.

mrautela365:
All three fixed in b65ae71e4. Every one was real, and the first two mean my previous commit was broken as shipped — thanks for catching them before this merged.

**1. The predicate was wrong** (@copilot, `campaign-service.service.ts:368`) — and this is the one that mattered most. I reported the raw `CampaignServiceDemandGen` flag, but that flag gates the **campaign-service** create path only. While the cutover is dark the controller falls through to the legacy creator, which creates demand-gen campaigns perfectly well — the comment above `unconfigured` in this same file says so explicitly, and notes an earlier revision already made this mistake by putting the check where it ran with flags off.

So I would have hidden a working option, and specifically for the entire staged **CREATE-off** rollout that this same PR prescribes — the deployment state most likely to be in effect. Now mirrors `createCampaigns`' own three-flag gate, so a partial flag set means the same thing in both places.

**2. The transport never ran** (@cursor and @copilot both) — correct, and my tests passed only because they set the input directly, never exercising the parent wiring. `briefCampaignsDemandGenEnabled` is written only by `loadBriefCampaigns`, whose callers are Optimize entry, retry, and a foundation switch *while already on Optimize*. The Planning → Implementation path never calls it, and it also needs `briefPersistence().briefId`, which does not exist before a brief is saved — so it can't carry a pre-create capability at all.

The corrected predicate resolves this without new plumbing: it answers `true` whenever legacy owns creation, so the common case is no longer a silently missing checkbox. Where campaign-service *does* own creation, the Optimize read has run.

**3. `false` meant two different things** (@copilot, `:1629`) — exactly right, and this was the destructive one. The parent used `false` for both "server said off" and "no answer yet", and the draft restore cleared on it. Selecting Demand Gen, entering Optimize, and returning mid-request would rewrite the saved selection to `false` **permanently**, because the emit carries it back to the parent; a later `true` would only reveal an unchecked box with the choice already gone.

Now tri-state. `null` withholds the control but **preserves** the draft; only an explicit `false` clears it. Hiding is reversible, clearing is not.

**Mutation-verified, each guard independently:**

| Mutation | Result |
|---|---|
| predicate reads the raw flag | 6 tests fail |
| draft restore clears on `null` too | 2 tests fail |
| control renders while unknown | 2 tests fail |

Two existing fixtures asserted the old wrong behaviour and were corrected rather than worked around: with every flag off, legacy owns creation, so `demandGenEnabled` is genuinely `true` there. The new server test is a 5-row matrix whose staged-rollout row is the one a naive implementation fails.

2095 server + 1260 app tests pass.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -456,6 +456,7 @@ <h2 class="text-base font-semibold text-gray-900">Email performance is not shown
               [briefPersistence]="briefPersistence()"
               [briefSaveInFlight]="briefSaveInFlight()"
               [draft]="implementationDraft()"
+              [demandGenEnabled]="briefCampaignsDemandGenEnabled()"
```
dealako, line 459: [minor] Demand Gen checkbox is unreachable on the direct Planning → Implementation first-create flow

Issue: This binds the Implementation tab's Demand Gen availability to `briefCampaignsDemandGenEnabled()`, but that signal is written to a real value only inside `loadBriefCampaigns()` (campaigns.component.ts:1317). `selectTab` calls `loadBriefCampaigns()` only for `tab === 'optimization'` (:906/911); the `implementation` branch (:899) calls only `loadEmailTemplatesIfNeverAnswered()`. So on a first-time create (Planning → save brief → Implementation) the signal is still `null`, `demandGenAvailable()` is false, and the control is withheld even in a deployment that supports Demand Gen.

Proof: signal initialized `null` at :488; set to a real value only at :1317 (inside `loadBriefCampaigns`), else `null` at :802/:1295/:1329. The design comment at :482-486 acknowledges the signal "stays `null` on the whole Planning → Implementation path" and argues the server predicate covers the common case — but that predicate's value never reaches the tab on this path, since nothing calls `loadBriefCampaigns` there.

Why it matters: Latent today — the chart ships DEMAND_GEN off and CREATE on, so `canCreateDemandGen` is false everywhere and the control is correctly hidden. But in a legacy CREATE-off deployment (where the legacy creator supports Demand Gen) or once DEMAND_GEN flips on, a user creating their first campaign for a brief can never select Demand Gen — the option only appears after they've created one, visited Optimize, and returned. The failure is safe (no spend, no error), which is why this is minor rather than blocking.

Fix: Populate the capability on Implementation-tab entry too — e.g. trigger a capability read (or `loadBriefCampaigns` when a brief id exists) from the `implementation` branch of `selectTab`, or thread the server capability to the tab through a path independent of the campaign list. Add a spec asserting the checkbox is offered on a fresh Planning → Implementation flow when the deployment supports Demand Gen.

packages/shared/src/interfaces/campaign.interface.ts
```diff
@@ -1698,15 +1698,35 @@ export interface CampaignListResult {
...
    *
    * A server fact, so it is reported by the server rather than mirrored into a client-side flag
    * that would drift from the deployment it describes.
    */
   statusToggleEnabled: boolean;
+  /**
+   * Whether THIS deployment can actually create a Demand Gen Google campaign.
+   *
+   * Same reasoning as `statusToggleEnabled`, for a different capability: the create route refuses
+   * `demand-gen` unless `LFX_CUTOVER_CAMPAIGN_SERVICE_DEMAND_GEN` is on, and the chart leaves that
```
dealako, line 1717: [nit] TSDoc omits the legacy-creator carve-out that `canCreateDemandGen` implements

Issue: This describes the field as "the create route refuses `demand-gen` unless `LFX_CUTOVER_CAMPAIGN_SERVICE_DEMAND_GEN` is on, and the chart leaves that flag unset" — i.e. as if `demandGenEnabled` mirrors the DEMAND_GEN flag. The value is actually produced by `canCreateDemandGen()` (campaign-service.service.ts:323-330), which returns `!cutoverOwnsCreate || DEMAND_GEN` and so is `true` across the entire staged CREATE-off rollout, because the legacy creator makes Demand Gen regardless of the flag.

Proof: `canCreateDemandGen` computes `cutoverOwnsCreate = CREATE && BRIEFS && JOBS; return !cutoverOwnsCreate || DEMAND_GEN`. Its own function doc states the carve-out ("the capability is missing only in the narrow window where campaign-service owns creation"), so the interface field doc and the function doc disagree.

Why it matters: A maintainer reading only this field doc would believe `demandGenEnabled === DEMAND_GEN flag` and could "simplify" `canCreateDemandGen` back to the raw flag — which would hide the working legacy Demand Gen option for the whole CREATE-off rollout, the exact regression the function doc warns against. Low impact since the authoritative computation is documented on the function.

Fix: Add a clause, e.g. "…unless DEMAND_GEN is on — but only once the CREATE/BRIEFS/JOBS cutover owns creation; while the cutover is dark the legacy creator makes Demand Gen regardless, so this stays true then (see `canCreateDemandGen`)."

dealako (approved):
Follow-up review (round 3) — @mrautela365, this closes it out. Every finding from both prior rounds is resolved, and enabling CREATE flipped the doc statements I'd flagged into correct ones, which you then rewrote so the surrounding prose matches. Approving.

### 👏 Nice work
- **All four prior findings resolved.** README header (prior blocking) now leads with CREATE default `"true"` and the STATUS_TOGGLE-first ordering; the "overlap is safe" overstatement is explicitly retracted at README:308, helper.ts:196, and values.yaml:584. values.yaml "CREATE is now ON" (prior blocking) is now true and the comment agrees with the value. controller:453 (prior minor) reframed to "an ordinary deployment state, whether from a staged rollout override or a pod that has not rolled yet." The `includeDemandGen` default comment (prior nit) corrected to "retained state," matching the flipped default.
- **The Demand Gen gating is server-authoritative and fails secure.** The client `demandGenEnabled` signal is a UI hint only; `campaign-service.service.ts` independently refuses `demand-gen` when the capability flag is off regardless of what the client sends, and the `null` tri-state withholds the control on unknown state without rewriting the saved draft.
- **`canCreateDemandGen` correctly models the legacy-creator carve-out** — it reports `true` whenever the CREATE/BRIEFS/JOBS cutover is dark (legacy creator still makes Demand Gen), rather than mirroring the raw DEMAND_GEN flag. The parametrized matrix at `campaign-service.service.spec.ts:2416` pins this.
- **STATUS_TOGGLE rollback carve-out is documented** (values.yaml:452, README:379) — turning it off strands existing UUID campaigns' pause/resume, so it stays on independently of the JOBS/CREATE sequence. Good catch to separate it from the enable-order rollback.

### Issue count (this round)
- 🔴 Blocking: 0
- 🟡 **Minor: 1** — Demand Gen checkbox is unreachable on the direct Planning → Implementation first-create flow: its backing signal is populated only by `loadBriefCampaigns()` (Optimize entry/retry/foundation-switch), never on Implementation-tab entry, so a fresh create sees `null` and hides the control even where the deployment supports it. Latent today (chart ships DEMAND_GEN off → correctly hidden everywhere), but it defeats the feature the moment the capability is true. *(inline)*
- ⚫ **Nit: 2** — (1) `demandGenEnabled` interface TSDoc describes the field as mirroring the DEMAND_GEN flag, omitting the legacy-creator carve-out that `canCreateDemandGen` actually implements; a reader could "simplify" the predicate back to the raw flag and regress it. *(inline)* (2) `canCreateDemandGen` is pod-local, so during a staged CREATE rollout `/list` on an old pod can advertise Demand Gen while `/create` on a new pod refuses it — bounded and safe (clear error, no spend, self-resolves on convergence), same env-flag skew already acknowledged for `statusToggleEnabled`; noting for awareness.

No security or data-privacy findings — CREATE dispatch keeps the `requireCampaignManager` (ED-only) boundary, the Demand Gen gate is server-enforced, and logs carry only `briefId`/`projectSlug`/sanitized `.message`.

### Reconciliation with Copilot
Copilot posted several doc-consistency comments this round; I checked each against current HEAD and most reference stale line numbers or phrasings you'd already corrected (README ordering list omitting STATUS_TOGGLE is actually the createCampaigns three-flag gate where STATUS_TOGGLE correctly isn't a prerequisite; the helper.ts/values.yaml "overlap is safe" and "default everywhere"/"default deployment" phrasings are already retracted/absent; the rollback list already carves out STATUS_TOGGLE). The one Copilot correctness concern that holds up is the Planning → Implementation signal-population gap — that's the minor above. Its draft-overwrite and independent-checkbox claims are refuted by the tri-state `null` handling (`implementation-tab.component.ts:1635`) and the now-servable Demand-Gen-alone path.

### Decision
✅ **Approved with minor comments.** The minor and both nits are non-blocking and safe to land as-is or fold into the DEMAND_GEN-enablement PR, since none of them can bite while the chart ships DEMAND_GEN off. The cutover completion, the server-authoritative gating, and the doc corrections are all sound.

dealako:
## Follow-up review (round 3) — @mrautela365

This closes it out. ✅ Every finding from both prior rounds is resolved, and completing the cutover (CREATE now on) turned the doc statements I'd flagged into correct ones — which you then rewrote so the surrounding prose matches. Approving.

### 👏 Nice work
- **All four prior findings resolved.** README header (was blocking) leads with CREATE default `"true"` and STATUS_TOGGLE-first ordering; the "overlap is safe" overstatement is explicitly retracted (README:308, helper.ts:196, values.yaml:584). values.yaml "CREATE is now ON" (was blocking) is now true and agrees with the value. controller:453 (was minor) reframed to "an ordinary deployment state… override or a pod that has not rolled yet." `includeDemandGen` default comment (was nit) corrected to "retained state."
- **Demand Gen gating is server-authoritative and fails secure** — the client signal is a UI hint; `campaign-service.service.ts` refuses `demand-gen` when the capability flag is off regardless of client input, and the `null` tri-state withholds the control without rewriting the saved draft.
- **`canCreateDemandGen` correctly models the legacy-creator carve-out** (`true` while the cutover is dark, since the legacy creator still makes Demand Gen) — pinned by the parametrized matrix at `campaign-service.service.spec.ts:2416`.
- **STATUS_TOGGLE rollback carve-out documented** (values.yaml:452, README:379) — kept out of the enable-order rollback because turning it off strands existing UUID campaigns' pause/resume.

### Issue count (this round)
- 🔴 Blocking: 0
- 🟡 **Minor: 1** — Demand Gen checkbox unreachable on the direct Planning → Implementation first-create flow: its backing signal is populated only by `loadBriefCampaigns()` (Optimize entry/retry/foundation-switch), never on Implementation-tab entry, so a fresh create sees `null` and hides the control even where supported. Latent today (chart ships DEMAND_GEN off → correctly hidden), but defeats the feature once the capability is true. *(inline)*
- ⚪ **Nit: 2** — (1) `demandGenEnabled` interface TSDoc omits the legacy-creator carve-out `canCreateDemandGen` implements, inviting a regressive "simplification" *(inline)*; (2) `canCreateDemandGen` is pod-local, so during a staged CREATE rollout `/list` on an old pod can advertise Demand Gen while `/create` on a new pod refuses it — bounded and safe (clear error, no spend, self-resolves), same env-flag skew already acknowledged for `statusToggleEnabled`.

No security or data-privacy findings — CREATE dispatch keeps the `requireCampaignManager` (ED-only) boundary, the Demand Gen gate is server-enforced, logs carry only `briefId`/`projectSlug`/sanitized `.message`.

### Reconciliation with Copilot
Checked each of Copilot's round-3 doc-consistency comments against current HEAD; most reference stale line numbers or phrasings already corrected (the README ordering list it flags is the createCampaigns three-flag *gate*, where STATUS_TOGGLE correctly isn't a prerequisite; the "overlap is safe" / "default everywhere" / "default deployment" phrasings are already retracted or absent; the rollback list already carves out STATUS_TOGGLE). The one Copilot correctness concern that holds is the Planning → Implementation signal-population gap — that's the minor above. Its draft-overwrite and independent-checkbox claims are refuted by the tri-state `null` handling and the now-servable Demand-Gen-alone path.

### Decision
✅ **Approved with minor comments.** The minor and both nits are non-blocking and safe to land as-is or fold into the DEMAND_GEN-enablement PR — none can bite while the chart ships DEMAND_GEN off. Cutover completion, server-authoritative gating, and the doc corrections are all sound. Nice work across three rounds.

PR #1886, opened 2026-08-27
refactor(campaigns): name the per-platform action-item thresholds

mrautela365:
## What

Names the per-platform thresholds behind the Optimize tab's **low CTR** and **clicks-without-conversions** action items. They were bare literals in three services.

```ts
export const CAMPAIGN_ALERT_THRESHOLDS = {
  'linkedin-ads': { lowCtrPct: 0.3, clicksWithoutConversions: 50,  minImpressions: null },
  'meta-ads':     { lowCtrPct: 0.5, clicksWithoutConversions: 20,  minImpressions: 500 },
  'reddit-ads':   { lowCtrPct: 0.3, clicksWithoutConversions: 100, minImpressions: 1000 },
} as const;
```

## No behaviour change

Every value is exactly what its service used before. Verified equal to the originals rather than inferred:

| | low CTR | clicks w/o conversions | impression floor |
|---|---|---|---|
| LinkedIn | `0.3` | `50` | none (`ctr > 0`) |
| Meta | `0.5` | `20` | `500` |
| Reddit | `0.3` | `100` | `1000` |

## Why not converge them here

Converging is the substantive half of LFXV2-3314 and it **changes which alerts fire on live campaigns**. That is not a change to make silently inside a constant extraction, so it is deliberately deferred — this PR only makes the divergence visible and greppable in one place.

## Two things the extraction surfaced

**The divergence is accidental.** Nothing in the code or the tickets states why LinkedIn should flag a click-through rate Meta considers healthy, or why Reddit tolerates five times as many unconverted clicks as Meta. Recorded in the constant's doc comment so the next reader does not have to re-derive it.

**LinkedIn has no impression floor.** It guards with `ctr > 0` instead. These are not equivalent — a campaign with 12 impressions and 1 click reads as 8.3% and is silently trusted, where Meta and Reddit would suppress it. Recorded as `null` rather than inventing a value.

Also worth noting for whoever picks up the convergence: **the ticket's own table has Meta and Reddit swapped** on clicks-without-conversions. It says Meta 100 / Reddit 20; the code is Meta 20 / Reddit 100.

## Testing

- `prettier --check` clean; shared package suite **1085 passed**.
- `tsc --noEmit` shows one pre-existing error (`docs-manifest`, a generated file not built in this worktree) — confirmed identical on a clean tree.
- **These rules have no test coverage in any spec**, so a green suite is not evidence here. The values were checked equal to the originals directly, which is the only claim this PR makes.

Refs LFXV2-3314


dealako:
@mrautela365 Nice refactor. Centralizing the per-platform Optimize-tab thresholds into `CAMPAIGN_ALERT_THRESHOLDS` is the right move: the values stay identical, the accidental cross-platform drift is now greppable, and the JSDoc correctly defers convergence to LFXV2-3314 rather than changing live alert behavior silently. Copilot and Cursor Bugbot both read this as low-risk constant extraction with no runtime change, and I agree after tracing all three call sites.

**Structured issue count**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: missing unit tests for alert rules; module consts interleaved with imports
- ⚪ Nit: 3 issues: `LI_ALERTS` naming; alias pattern vs pacing; JSDoc field units
- ❔ Question: 0 items

**Bot reconciliation**
- **CodeRabbit**: review skipped (auto-reviews disabled); no findings to reconcile.
- **Cursor Bugbot** ([#5433087384](https://github.com/linuxfoundation/lfx-self-serve/pull/1886#issuecomment-5433087384)): agree — refactor-only, identical thresholds, no auth/data/API contract change.
- **Copilot** ([#5036432210](https://github.com/linuxfoundation/lfx-self-serve/pull/1886#pullrequestreview-5036432210)): agree — reviewed all 4 files, no issues raised.

packages/shared/src/constants/campaign.constants.ts
```diff
@@ -111,6 +111,31 @@ export const CAMPAIGN_PACING_THRESHOLDS = {
...
+ * separate decision (LFXV2-3314) precisely because it CHANGES which alerts fire on live
+ * campaigns, and that is not a change to make silently while extracting constants.
+ *
+ * `minImpressions` is the floor below which the CTR rule is suppressed, because a click-through
+ * rate over a handful of impressions is noise. LinkedIn has no floor today and instead requires
+ * `ctr > 0`; the two are not equivalent — a campaign with 12 impressions and one click reads as
+ * 8.3% and is silently trusted — but recording `null` states what is there rather than
+ * inventing a value.
+ */
+export const CAMPAIGN_ALERT_THRESHOLDS = {
```
dealako, line 150: [minor] Alert threshold rules still have zero unit test coverage after centralization

Issue: `CAMPAIGN_ALERT_THRESHOLDS` is now the single edit point for low-CTR and clicks-without-conversions gates, but no spec asserts values or boundary behavior for any platform.
Proof: No matches for `lowCtrPct`, `clicksWithoutConversions`, or alert-threshold assertions in `**/*.spec.ts`. `linkedin-ads.service.spec.ts` covers pacing action items only; `meta-ads.service.spec.ts` does not exercise `buildMetaActionItems`; no `reddit-ads.service.spec.ts` exists.
Why it matters: A later LFXV2-3314 convergence edit can change which Optimize-tab alerts fire with no CI signal.
Fix: Add boundary specs (ctr at `lowCtrPct`, clicks at `clicksWithoutConversions`, impressions at `minImpressions`) using the existing `importActual('@lfx-one/shared/constants')` pattern.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
```diff
@@ -18,7 +18,10 @@ import type {
   LinkedInTargetingProfileConfig,
 } from '@lfx-one/shared/interfaces';
 
-import { CAMPAIGN_PACING_THRESHOLDS, LINKEDIN_API_VERSION, LINKEDIN_GEO_RESOLVE_MAP } from '@lfx-one/shared/constants';
+import { CAMPAIGN_ALERT_THRESHOLDS, CAMPAIGN_PACING_THRESHOLDS, LINKEDIN_API_VERSION, LINKEDIN_GEO_RESOLVE_MAP } from '@lfx-one/shared/constants';
+
+// Named so the two rules below read as one platform's tuning rather than loose literals.
+const LI_ALERTS = CAMPAIGN_ALERT_THRESHOLDS['linkedin-ads'];
```
dealako: [minor] Module const interleaved with import statements

Issue: `const LI_ALERTS` sits between import blocks instead of after all imports.
Proof: Lines 21–28 import constants, line 24 declares the alias, line 26 resumes `import type { Request }`. Same pattern in `meta-ads.service.ts:30` and `reddit-ads.service.ts:19`. Other module consts in these files (e.g. `NUMERIC_ID_RE`) follow the last import.
Why it matters: Breaks the repo convention of contiguous imports at file top.
Fix: Move `LI_ALERTS`, `META_ALERTS`, and `REDDIT_ALERTS` to immediately after the final import in each file.

dealako: [nit] Inconsistent platform alias prefix (LI vs META/REDDIT)

Issue: LinkedIn uses abbreviated `LI_ALERTS` while Meta and Reddit use full prefixes.
Proof: `LI_ALERTS` here vs `META_ALERTS` / `REDDIT_ALERTS` in sibling services.
Why it matters: Slightly harder to grep and map across files.
Fix: Rename to `LINKEDIN_ALERTS` for parity, or drop the alias and index `CAMPAIGN_ALERT_THRESHOLDS['linkedin-ads']` directly.

apps/lfx-one/src/server/services/meta-ads.service.ts
```diff
@@ -24,6 +25,9 @@ import {
   META_INELIGIBLE_COUNTRIES,
   normalizeGeoTargets,
 } from '@lfx-one/shared/constants';
+
+// Named so the two rules below read as one platform's tuning rather than loose literals.
+const META_ALERTS = CAMPAIGN_ALERT_THRESHOLDS['meta-ads'];
```
dealako: [nit] Alert aliases inconsistent with pacing threshold access pattern

Issue: These services use `CAMPAIGN_PACING_THRESHOLDS.*` directly but local aliases for alert thresholds only.
Proof: `meta-ads.service.ts:543-545` uses pacing constants inline; `:581-589` uses `META_ALERTS`. Both constants live adjacently in `campaign.constants.ts`.
Why it matters: Two conventions in one module for sibling constants.
Fix: Use `CAMPAIGN_ALERT_THRESHOLDS['meta-ads'].lowCtrPct` inline (matching pacing style) or introduce matching pacing aliases if brevity is the goal.

packages/shared/src/constants/campaign.constants.ts
dealako, line 150: [nit] JSDoc omits field units for alert threshold properties

Issue: The block comment explains drift and `minImpressions` semantics but not units for `lowCtrPct` or `clicksWithoutConversions`.
Proof: Adjacent `CAMPAIGN_PACING_THRESHOLDS` documents "percentage of budget spent"; this block has no equivalent for CTR percent vs click-count floor.
Why it matters: Future LFXV2-3314 convergence edits need unit clarity to avoid mis-scaling.
Fix: Add a short field glossary, e.g. `lowCtrPct — CTR floor in percent (0.3 = 0.3%)`; `clicksWithoutConversions — click count before flagging zero conversions`.

dealako (dismissed):
## Review summary

Clean constant extraction with no behavior change. Threshold values match the originals at all three call sites (LinkedIn `ctr > 0` + 0.3/50, Meta 0.5/500/20, Reddit 0.3/1000/100). Security and privacy surfaces are unchanged.

Two minor items worth a follow-up, neither blocking merge for this refactor:

1. **Test coverage** — alert threshold rules remain untested after centralization; a later LFXV2-3314 convergence edit would have no CI guardrail.
2. **Import layout** — platform alert aliases sit between import blocks in all three services; move them below the last import for consistency with the rest of the codebase.

Three nits (inline): `LI_ALERTS` vs `LINKEDIN_ALERTS`, alias pattern vs direct `CAMPAIGN_PACING_THRESHOLDS` access, and JSDoc field unit glossary.

packages/shared/src/constants/campaign.constants.ts
mrautela365: Fixed in `a44545510` — three cases in `campaign.constants.spec.ts`.

They pin the **values as literals** rather than re-deriving the rules. The whole claim of this PR is that centralising the thresholds changed no behaviour, and a value drifting is exactly how that claim would quietly stop being true. Asserting `x === CAMPAIGN_ALERT_THRESHOLDS[p].lowCtrPct` would pass against any value at all.

A second case pins the **divergence itself**, so converging the platforms has to delete it deliberately rather than discover it red — and so nobody tidies one value into agreement without that being the point of their change.

**Verified by mutation**, not by a green suite: changing meta's `lowCtrPct` from `0.5` to `0.3` — the exact "tidy" the comment warns about — fails 2 of the 3 (`2 failed | 1086 passed`). Restored after.

apps/lfx-one/src/server/services/linkedin-ads.service.ts
mrautela365: Fixed in `a44545510`, and your two nits below pointed at the same better answer — so all three are resolved by one change rather than three patches.

The aliases are gone entirely. Each rule now reads `CAMPAIGN_ALERT_THRESHOLDS['<platform>'].field` at the point of use, which is exactly how `CAMPAIGN_PACING_THRESHOLDS` is already read in these same files. That removes the interleaved module const, removes the `LI_` vs `META_` naming inconsistency, and leaves one convention for sibling constants instead of two.

mrautela365: Fixed in `a44545510` — took the second option you offered. The alias is gone; the rules now index `CAMPAIGN_ALERT_THRESHOLDS['linkedin-ads']` directly, so there is no prefix to be inconsistent about and it greps as one name across all three files.

apps/lfx-one/src/server/services/meta-ads.service.ts
mrautela365: Fixed in `a44545510` — you were right that this was the deciding argument. Both constants are now read the same way: `CAMPAIGN_ALERT_THRESHOLDS['meta-ads'].lowCtrPct` alongside `CAMPAIGN_PACING_THRESHOLDS.underspending`, no local aliases for either.

packages/shared/src/constants/campaign.constants.ts
mrautela365: Fixed in `a44545510`. The three fields are genuinely different quantities, and the JSDoc now says which is which:

```
lowCtrPct                — PERCENTAGE POINTS, not a ratio. 0.3 means 0.3%, and the rule fires
                           when a campaign's ctr is below it. The upstream ctr fields are
                           already percentages, so no conversion is applied.
clicksWithoutConversions — a COUNT of clicks. The rule fires above it, with zero conversions.
minImpressions           — a COUNT of impressions, the floor below which the CTR rule is
                           suppressed.
```

The ratio-vs-percentage-points one is the ambiguity worth closing before the LFXV2-3314 convergence — reading `0.3` as a ratio would be a 100x error in the direction that silences the rule.

packages/shared/src/constants/campaign.constants.spec.ts
```diff
@@ -306,3 +307,53 @@ describe('normalizeMicrosoftGeoTargets', () => {
     expect(normalizeMicrosoftGeoTargets(undefined)).toEqual([]);
   });
 });
+
+/**
+ * These thresholds are the single edit point for the Optimize tab's low-CTR and
+ * clicks-without-conversions rules, and the rules themselves have no spec of their own. So this
+ * block pins the VALUES rather than re-deriving the rules: the whole claim of LFXV2-3314's first
+ * step is that centralising them changed no behaviour, and a value drifting here is exactly how
+ * that claim would quietly stop being true.
```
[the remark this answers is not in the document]
mrautela365: Fixed in `daa0f8209` — the claim was overstated and I've scoped it.

Verified: Google does keep its own literals in `campaign-metrics.service.ts:303-314`. The JSDoc now says which three platforms this covers and why the fourth is excluded — Google's display rule is `!isSearch && ctr < 0.3 && impressions > 1000`, a different predicate over a different population, so folding it in would flatten a *real* distinction rather than the accidental one this PR is about. LFXV2-3314's convergence is where that decision belongs.

packages/shared/src/constants/campaign.constants.ts
```diff
@@ -111,6 +111,40 @@ export const CAMPAIGN_PACING_THRESHOLDS = {
...
+ * Units, since the three fields are not in the same kind of quantity:
+ *
+ *   lowCtrPct                — PERCENTAGE POINTS, not a ratio. `0.3` means 0.3%, and the rule
+ *                              fires when a campaign's `ctr` is below it. The upstream `ctr`
+ *                              fields are already percentages, so no conversion is applied.
+ *   clicksWithoutConversions — a COUNT of clicks. The rule fires above it, with zero
+ *                              conversions.
+ *   minImpressions           — a COUNT of impressions, the floor below which the CTR rule is
+ *                              suppressed, because a click-through rate over a handful of
+ *                              impressions is noise.
```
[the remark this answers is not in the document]
mrautela365: Fixed in `daa0f8209`. Confirmed against the consumers — all three guard with `impressions > minImpressions`, so a campaign at exactly 500 or 1000 is suppressed too.

The JSDoc now states it: *"Consumers guard with `impressions > minImpressions`, so a campaign AT the value is suppressed too — the rule needs strictly more."* Worth stating, since a future consumer reading "floor" would reasonably write `>=` and silently shift the boundary by one campaign.

packages/shared/src/constants/campaign.constants.ts
```diff
@@ -111,6 +111,41 @@ export const CAMPAIGN_PACING_THRESHOLDS = {
...
+ *
+ * Naming them here makes the drift greppable and reviewable in one place. Converging them is a
+ * separate decision (LFXV2-3314) precisely because it CHANGES which alerts fire on live
+ * campaigns, and that is not a change to make silently while extracting constants.
+ *
+ * Units, since the three fields are not in the same kind of quantity:
+ *
+ *   lowCtrPct                — PERCENTAGE POINTS, not a ratio. `0.3` means 0.3%, and the rule
+ *                              fires when a campaign's `ctr` is below it. The upstream `ctr`
+ *                              fields are already percentages, so no conversion is applied.
```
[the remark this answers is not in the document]
mrautela365: You're right and I was wrong — fixed in `16479e4d8`.

I checked the three builders rather than assume: `ctr` is derived locally as `(clicks / impressions) * 100` (`linkedin-ads:1044`, `meta-ads:510`, `reddit-ads:299`), not read from the platform. My comment claimed the upstream fields were already percentages, which would have sent the next reader looking for a contract that does not exist.

It now says the scale is set **locally**, and names the failure that matters: a builder switching to the raw ratio would **silence every one of these rules** rather than error — a `0.02` ratio is below every `lowCtrPct` threshold, so nothing would ever fire again and nothing would look broken.

On the second location you flagged (line 139): that passage is accurate as written. The 12-impressions/8.3% example is consistent with a locally-derived percentage, so I left it.

mrautela365:
Picked up the two suppressed Copilot comments from the review bodies.

The `ctr` units one was already fixed in 16479e4d8 — it was correct that all three consumers derive CTR locally rather than reading a platform percentage.

The 12-impression example one is fixed in c03a55eb3, and it was right for a stronger reason than it stated: at 8.3% the rate clears every platform's `lowCtrPct`, so all three predicates omit the alert regardless of their floors and LinkedIn's `ctr > 0` gate never enters into it — the example demonstrated nothing at all. Replaced with its suggestion (1 click / 400 impressions = 0.25%), verified against the actual predicates: `linkedin-ads.service.ts:1080` is `ctr > 0 && ctr < lowCtrPct`, `meta-ads.service.ts:578` is `ctr < lowCtrPct && impressions > minImpressions`. So 400 alerts on LinkedIn and is suppressed on Meta.

dealako:
@mrautela365 Good follow-up. You closed out every style item from the first round and added thoughtful regression tests that pin the pre-extraction values (including the intentional platform divergence and LinkedIn's null floor). The expanded JSDoc with percentage-point units, strict `>` boundary semantics, and the 1-click/400-impressions example makes the LinkedIn-vs-Meta floor difference concrete.

👏 **Nice work**
- Dropped the interleaved aliases and switched to direct `CAMPAIGN_ALERT_THRESHOLDS[...]` indexing, matching the existing `CAMPAIGN_PACING_THRESHOLDS` pattern.
- The constant spec is deliberately non-tautological (literals vs constant) and includes a convergence guard test that will force an explicit delete when LFXV2-3314 aligns platforms.
- Doc commits (`16479e4d8`, `c03a55eb3`) corrected the CTR scale source and sharpened the floor-difference example.

**Revision tracking (prior round)**
- ✅ **Import layout** — resolved in `a44545510`
- ✅ **`LI_ALERTS` naming** — resolved (aliases removed)
- ✅ **Alias vs pacing pattern** — resolved (direct constant access)
- ✅ **JSDoc field units** — resolved (`16479e4d8`, `c03a55eb3`)
- ⚠️ **Test coverage** — partially addressed: value-pinning specs added; service-level predicate boundary tests still absent (documented as intentional in the spec block comment)

**Structured issue count (open)**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 1 issue: alert predicate inclusion/exclusion at exact thresholds still untested in service specs (optional before LFXV2-3314 convergence)
- ⚪ Nit: 0 issues

**Bot reconciliation**
- **Cursor Bugbot**: agree — refactor-only with regression tests pinning prior values.
- **CodeRabbit**: still skipped (auto-reviews disabled).
- **Copilot**: no new inline comments; author noted the CTR-units item was fixed in `16479e4d8`.

packages/shared/src/constants/campaign.constants.spec.ts
```diff
@@ -306,3 +307,59 @@ describe('normalizeMicrosoftGeoTargets', () => {
...
+
+/**
+ * These thresholds are the edit point for the LinkedIn, Meta and Reddit low-CTR and
+ * clicks-without-conversions rules — not for every Optimize-tab rule. Google keeps its own
+ * literals in `campaign-metrics.service.ts` and is deliberately out of scope here: its display
+ * rule reads `!isSearch && ctr < 0.3 && impressions > 1000`, which is a different predicate over
+ * a different population, so folding it in would flatten a real distinction rather than an
+ * accidental one. LFXV2-3314's convergence is where that decision belongs.
+ *
+ * The rules themselves have no spec of their own. So this
```
dealako, line 319: [minor] Alert predicate boundary behavior still untested in service specs

Issue: Constant regression tests pin values and divergence, but `buildMetaActionItems`, `buildRedditActionItems`, and the LinkedIn analytics alert branches still have no inclusion/exclusion tests at exact thresholds.
Proof: Spec block comment at line 319 states rules have no spec of their own; grep shows no alert-threshold cases in `linkedin-ads.service.spec.ts`, `meta-ads.service.spec.ts`, or a reddit ads spec. JSDoc now documents strict `>` / `<` comparators (e.g. impressions must be strictly greater than 500 for Meta).
Why it matters: A service could revert to hardcoded literals or flip a comparator and CI would still pass the constant specs.
Fix: Optional follow-up for LFXV2-3314 — add one boundary case per platform (at/equal/just-above `minImpressions`, `lowCtrPct`, `clicksWithoutConversions`) mirroring the pacing boundary tests already in `linkedin-ads.service.spec.ts`.

dealako (approved):
## Follow-up review

All four style/documentation items from the prior round are resolved. Test coverage moved from zero to three meaningful constant regression tests that guard the no-behavior-change claim.

One residual minor remains, and it is an intentional trade-off documented in the spec: value pinning without service-level predicate boundary tests. Acceptable for this extraction step; worth adding when LFXV2-3314 touches convergence or comparator semantics.

No new security, privacy, or correctness issues in the campaign-scoped commits (`a44545510` through `c03a55eb3`). The merge from `main` (social-listening) was out of scope for this follow-up and did not introduce campaign-threshold regressions.

mrautela365:
Thanks — and agreed, verified rather than assumed: `linkedin-ads.service.spec.ts` and `meta-ads.service.spec.ts` have zero references to the thresholds or the `*ActionItems` builders, and there is no `reddit-ads.service.spec.ts` at all. So the gap is exactly as described: the constant specs pin the values but nothing pins the comparators, and a service reverting to hardcoded literals or flipping `>` to `>=` would leave CI green.

Taking your call that it is non-blocking for this extraction and leaving it out of this PR — a push here would dismiss the approval for a change you have already scoped as follow-up work.

Filing it against LFXV2-3314 for the convergence step, with the boundary cases the JSDoc now makes explicit: Meta at exactly 500 impressions (suppressed, since the guard is strictly `>`), LinkedIn at `ctr === 0` (suppressed by its `ctr > 0` gate) and just above, Reddit at exactly 1000, plus each `clicksWithoutConversions` at and just above its value — mirroring the pacing boundary tests already in `linkedin-ads.service.spec.ts`.

PR #1887, opened 2026-08-27
feat(org-profile): ungate the Upload Logo affordance

luismoriguerra:
Removes the feature-flag gate from the Company Logo Upload affordance, and hardens the failure path the gate was hiding.

The org logo upload backend is released and verified in production (member-service `v0.10.19`), so the gate comes off.

## What changes

**The gate.** The per-feature `isOrgLensUploadLogoEnabled` wrapper and its private `org_lens_private_release` read are gone, along with 9 template bindings, the flag half of 4 handler guards, and the e2e flag pin. The dropzone's interactive affordances — pointer cursor, dashed border, drag highlight, `role="button"`, `tabindex`, `aria-label` — are now unconditional, which is exactly what rendered when the flag was on.

The `busy()` half of each guard is kept. It is a separate concern: it prevents a second upload starting while one is in flight.

**The umbrella flag is untouched.** `ORG_LENS_PRIVATE_RELEASE_FLAG` still gates the person detail drawer, its service, and the org project detail view, and is still exported from shared constants. Only this one surface graduates.

## Why the error-copy commits are here

They are not unrelated scope. The flag was the only thing keeping users away from the SVG sanitizer's rejection path, and that sanitizer recently became much stricter — it now rejects files it previously accepted-and-silently-blackened. Ungating without fixing the failure copy would ship a worse experience than leaving the gate on.

Previously every non-403 failure read *"Unable to upload logo. Please try again."* For a permanently rejected file that is the wrong instruction — the same bytes fail forever. Now:

| Status | Copy |
|---|---|
| 403 | Permission denied (unchanged) |
| 413 | Logo must be 2MB or smaller |
| 4xx (non-transient) | **Logo rejected** + the upstream reason |
| 408 / 429 / 0 / 5xx | Please try again |

408 and 429 are 4xx but retryable — this server mints 408 for its own upstream abort — so both are excluded via the repo's existing `isTransientHttpError`.

## Two bugs the pre-PR review caught

Recording these because both were mine and both would have shipped silently:

1. **The rejection detail read a key the BFF never emits.** `error.error.message` — but `BaseApiError.toResponse()` serialises `{ error, code }`, and `ServiceValidationError` adds `errors[]`. Every real rejection fell through to the generic fallback, so the sanitizer's reason never reached the user. The feature was inert in production. Worse, the test fabricated a body shape the server cannot produce, certifying the broken path as working. Both now go through the repo's `extractErrorMessage` helper, and the tests use the three real contract shapes — reverting the extraction fails three of them.

2. **`border-2` became unconditional**, and with border-box sizing the `h-20 w-20` children overflowed the wrapper's 76px content box. Invisible while the border was flag-gated off; the only rendered state after ungating. Children are now `h-full w-full`.

## Verification

- `yarn format:check` · `yarn lint:check` · `yarn check-types` — all clean
- Component suite **29 tests passing** (17 before; the flag-gating block was replaced with flag-independent coverage plus the new error mapping)
- Mutation-checked: restoring the old error extraction fails 3 tests; removing the `busy()` guard fails 1

## Notes for reviewers

- **No new permission surface.** `org-profile-edit` is not a route — it is only instantiated behind `canEdit()` (org writer), which `enterEditMode()` re-checks. The flag was redundant with an already-correct authorization gate.
- Pre-existing a11y issues were found and deliberately **not** fixed here to keep the diff scoped: the upload live region sits inside `role="button"` (presentational children), the wrapper `aria-label` shadows the logo ``, and there is no focus-visible ring. All predate this change and are identical to the flag-ON behaviour. Happy to file them.

Issue: LFXV2-2016

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
@@ -35,10 +35,10 @@
         (drop)="onLogoDrop($event)"
         data-testid="org-profile-edit-logo-dropzone">
         @if (logoUrl()) {
-          <img [src]="logoUrl()" [alt]="record().name + ' logo'" class="h-20 w-20 rounded-md object-contain" data-testid="org-profile-edit-logo-image" />
+          <img [src]="logoUrl()" [alt]="record().name + ' logo'" class="h-full w-full rounded-md object-contain" data-testid="org-profile-edit-logo-image" />
         } @else {
           <div
-            class="flex h-20 w-20 items-center justify-center rounded-md bg-blue-600 text-xl font-semibold text-white"
+            class="flex h-full w-full items-center justify-center rounded-md bg-blue-600 text-xl font-semibold text-white"
             data-testid="org-profile-edit-logo-initials">
             {{ (record().name | initials) || 'OR' }}
           </div>
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
@@ -67,7 +67,7 @@
           [loading]="logoUploading()"
           (onClick)="triggerLogoUpload()"
           data-testid="org-profile-edit-upload-logo-button" />
-        <p class="text-xs text-gray-500">PNG, JPEG, or SVG, not more than 2MB.</p>
+        <p class="text-xs text-gray-500">PNG, JPEG, or SVG, not more than 2MB. SVGs are sanitized, so scripts and embedded styles are removed.</p>
       </div>
     </header>
 
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
@@ -44,10 +44,16 @@
           </div>
         }
         @if (logoUploading()) {
-          <div class="absolute inset-0 flex items-center justify-center rounded-md bg-black/40" role="status" aria-live="polite">
-            <i class="fa-light fa-spinner fa-spin text-white" aria-hidden="true"></i>
-            <span class="sr-only">Uploading logo…</span>
+          <div class="absolute inset-0 flex items-center justify-center rounded-md bg-black/40" aria-hidden="true">
+            <i class="fa-light fa-spinner fa-spin text-white"></i>
           </div>
+        }
+      </div>
+      <!-- Outside the dropzone on purpose: an element with role="button" flattens its descendants
+           in the accessibility tree, so a live region nested inside it is not reliably announced. -->
+      <div role="status" aria-live="polite" class="sr-only">
+        @if (logoUploading()) {
+          Uploading logo…
         }
       </div>
       <input
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
@@ -67,7 +67,7 @@
           [loading]="logoUploading()"
           (onClick)="triggerLogoUpload()"
           data-testid="org-profile-edit-upload-logo-button" />
-        <p class="text-xs text-gray-500">PNG, JPEG, or SVG, not more than 2MB. SVGs are sanitized, so scripts and embedded styles are removed.</p>
+        <p class="text-xs text-gray-500">PNG, JPEG, or SVG, not more than 2MB. SVGs are sanitized; those using unsupported CSS may be rejected.</p>
       </div>
     </header>
 
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
@@ -17,25 +17,16 @@
   </nav>
 
   <div class="rounded-lg border border-gray-200 bg-white p-6">
-    <!-- Logo + Upload Logo button (LFXV2-3288). Drop zone doubles as the preview; click or drag-drop a file.
-         The Upload Logo affordance (button, instruction copy, hidden file input) and every interactive
-         class/handler on the dropzone are gated behind the per-feature `isOrgLensUploadLogoEnabled`
-         wrapper (PR #1583) — today it just reads `org_lens_private_release`, but the wrapper lets
-         logo upload flip independently of other Org Lens surfaces later. When the flag is off, this
-         collapses to a static preview: no cursor-pointer, no dashed border, no click/drag surface,
-         no file picker — matching the read-only view most Org Lens users get today. -->
+    <!-- Logo + Upload Logo button (LFXV2-3288). Drop zone doubles as the preview; click or drag-drop a file. -->
     <header class="flex items-center gap-4 border-b border-gray-100 pb-4">
       <div
-        class="relative flex h-20 w-20 shrink-0 items-center justify-center rounded-md"
-        [class.cursor-pointer]="isOrgLensUploadLogoEnabled()"
-        [class.border-2]="isOrgLensUploadLogoEnabled()"
-        [class.border-dashed]="isOrgLensUploadLogoEnabled()"
-        [class.border-blue-500]="isOrgLensUploadLogoEnabled() && logoDragActive()"
-        [class.border-transparent]="isOrgLensUploadLogoEnabled() && !logoDragActive()"
-        [attr.role]="isOrgLensUploadLogoEnabled() ? 'button' : null"
-        [attr.tabindex]="isOrgLensUploadLogoEnabled() ? 0 : null"
-        [attr.aria-label]="isOrgLensUploadLogoEnabled() ? 'Upload organization logo' : null"
-        [attr.aria-disabled]="isOrgLensUploadLogoEnabled() ? busy() : null"
+        class="relative flex h-20 w-20 shrink-0 cursor-pointer items-center justify-center rounded-md border-2 border-dashed"
+        [class.border-blue-500]="logoDragActive()"
+        [class.border-transparent]="!logoDragActive()"
+        role="button"
+        tabindex="0"
+        aria-label="Upload organization logo"
+        [attr.aria-disabled]="busy()"
         (click)="triggerLogoUpload()"
         (keydown.enter)="$event.preventDefault(); triggerLogoUpload()"
         (keydown.space)="$event.preventDefault(); triggerLogoUpload()"
@@ -59,27 +50,25 @@
           </div>
         }
       </div>
-      @if (isOrgLensUploadLogoEnabled()) {
-        <input
-          #logoInput
-          type="file"
-          [attr.accept]="allowedLogoAccept"
-          class="hidden"
+      <input
+        #logoInput
+        type="file"
+        [attr.accept]="allowedLogoAccept"
+        class="hidden"
+        [disabled]="busy()"
+        (change)="onLogoFileSelected($event)"
+        data-testid="org-profile-edit-logo-input" />
+      <div class="flex flex-col gap-1">
+        <p-button
+          [label]="logoUploading() ? 'Uploading…' : 'Upload Logo'"
+          severity="secondary"
+          [outlined]="true"
           [disabled]="busy()"
-          (change)="onLogoFileSelected($event)"
-          data-testid="org-profile-edit-logo-input" />
-        <div class="flex flex-col gap-1">
-          <p-button
-            [label]="logoUploading() ? 'Uploading…' : 'Upload Logo'"
-            severity="secondary"
-            [outlined]="true"
-            [disabled]="busy()"
-            [loading]="logoUploading()"
-            (onClick)="triggerLogoUpload()"
-            data-testid="org-profile-edit-upload-logo-button" />
-          <p class="text-xs text-gray-500">PNG, JPEG, or SVG, not more than 2MB.</p>
-        </div>
-      }
+          [loading]="logoUploading()"
+          (onClick)="triggerLogoUpload()"
+          data-testid="org-profile-edit-upload-logo-button" />
+        <p class="text-xs text-gray-500">PNG, JPEG, or SVG, not more than 2MB.</p>
+      </div>
     </header>
 
     <form [formGroup]="form" (ngSubmit)="onSave()" class="mt-6 flex flex-col gap-6" data-testid="org-profile-edit-form">
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.spec.ts
@@ -334,4 +334,53 @@ describe('OrgProfileEditComponent — logo upload', () => {
     fixture.componentInstance['triggerLogoUpload']();
     expect(clickSpy).toHaveBeenCalledTimes(1);
   });
+
+  it.each([
+    [413, 'Logo too large', 'Logo must be 2MB or smaller.'],
+    [403, 'Permission denied', 'You no longer have permission to edit this organization.'],
+  ])('maps status %i to non-retryable copy', async (status, summary, detail) => {
+    fixture.detectChanges();
+    fixture.componentInstance['onLogoFileSelected']({
+      target: { files: [pngFile()], value: 'logo.png' } as unknown as HTMLInputElement,
+    } as unknown as Event);
+    await fixture.whenStable();
+
+    uploadLogo$.error(new HttpErrorResponse({ status }));
+    await fixture.whenStable();
+
+    expect(toastAdd).toHaveBeenCalledWith(expect.objectContaining({ summary, detail }));
+  });
+
+  it('surfaces the upstream reason when the sanitizer rejects an SVG', async () => {
+    fixture.detectChanges();
+    fixture.componentInstance['onLogoFileSelected']({
+      target: { files: [pngFile()], value: 'logo.png' } as unknown as HTMLInputElement,
+    } as unknown as Event);
+    await fixture.whenStable();
+
+    // A permanently-rejected file must not be described as retryable — the same bytes fail forever.
+    uploadLogo$.error(new HttpErrorResponse({ status: 400, error: { message: 'SVG upload cannot be processed: unsupported CSS selector "rect"' } }));
+    await fixture.whenStable();
+
+    expect(toastAdd).toHaveBeenCalledWith(
+      expect.objectContaining({
+        summary: 'Logo rejected',
+        detail: 'SVG upload cannot be processed: unsupported CSS selector "rect"',
+      })
+    );
+    expect(toastAdd).not.toHaveBeenCalledWith(expect.objectContaining({ detail: expect.stringContaining('Please try again') }));
+  });
+
+  it('still prompts a retry for a server-side failure', async () => {
+    fixture.detectChanges();
+    fixture.componentInstance['onLogoFileSelected']({
+      target: { files: [pngFile()], value: 'logo.png' } as unknown as HTMLInputElement,
+    } as unknown as Event);
+    await fixture.whenStable();
+
+    uploadLogo$.error(new HttpErrorResponse({ status: 502 }));
+    await fixture.whenStable();
+
+    expect(toastAdd).toHaveBeenCalledWith(expect.objectContaining({ summary: 'Upload failed', detail: 'Unable to upload logo. Please try again.' }));
+  });
 });
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.spec.ts
@@ -304,4 +304,34 @@ describe('OrgProfileEditComponent — logo upload', () => {
     fixture.componentInstance['onCancel']();
     expect(cancelled).toEqual([undefined]);
   });
+
+  it('renders the Upload Logo button, the hidden file input and the logo image', () => {
+    fixture.detectChanges();
+    const root: HTMLElement = fixture.nativeElement;
+
+    expect(root.querySelector('[data-testid="org-profile-edit-upload-logo-button"]')).not.toBeNull();
+    expect(root.querySelector('[data-testid="org-profile-edit-logo-input"]')).not.toBeNull();
+    expect(root.querySelector('[data-testid="org-profile-edit-logo-image"]')).not.toBeNull();
+  });
+
+  it('does not reopen the file picker while an upload is already in flight', async () => {
+    fixture.detectChanges();
+    const clickSpy = vi.spyOn(fixture.componentInstance['logoInput']()!.nativeElement, 'click');
+
+    fixture.componentInstance['onLogoFileSelected']({
+      target: { files: [pngFile()], value: 'logo.png' } as unknown as HTMLInputElement,
+    } as unknown as Event);
+    await fixture.whenStable();
+    clickSpy.mockClear();
+
+    // busy() is true while uploadLogo is pending, so the trigger must be inert.
+    fixture.componentInstance['triggerLogoUpload']();
+    expect(clickSpy).not.toHaveBeenCalled();
+
+    uploadLogo$.next({ ...record, logoUrl: 'https://cdn.example.com/logo.png?v=3' });
+    await fixture.whenStable();
+
+    fixture.componentInstance['triggerLogoUpload']();
+    expect(clickSpy).toHaveBeenCalledTimes(1);
+  });
 });
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.spec.ts
@@ -1,13 +1,10 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { signal } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { HttpErrorResponse } from '@angular/common/http';
 import { provideNoopAnimations } from '@angular/platform-browser/animations';
-import { ORG_LENS_PRIVATE_RELEASE_FLAG } from '@lfx-one/shared/constants';
 import type { OrgCanonicalRecord } from '@lfx-one/shared/interfaces';
-import { FeatureFlagService } from '@services/feature-flag.service';
 import { OrgProfileService } from '@services/org-profile.service';
 import { MessageService } from 'primeng/api';
 import { Observable, Subject, take } from 'rxjs';
@@ -59,18 +56,7 @@ describe('OrgProfileEditComponent — logo upload', () => {
 
     await TestBed.configureTestingModule({
       imports: [OrgProfileEditComponent],
-      providers: [
-        provideNoopAnimations(),
-        { provide: OrgProfileService, useValue: { uploadLogo: uploadLogoMock } },
-        // PR #1583 — gate the Upload Logo affordance on `org_lens_private_release`. Default the
-        // flag to `true` in this suite because every existing test exercises the upload path; the
-        // dedicated "flag disabled" describe below re-configures TestBed with the same mock
-        // returning `false` for the gating test.
-        {
-          provide: FeatureFlagService,
-          useValue: { getBooleanFlag: vi.fn((key: string) => signal(key === ORG_LENS_PRIVATE_RELEASE_FLAG ? true : false)) },
-        },
-      ],
+      providers: [provideNoopAnimations(), { provide: OrgProfileService, useValue: { uploadLogo: uploadLogoMock } }],
     }).compileComponents();
 
     fixture = TestBed.createComponent(OrgProfileEditComponent);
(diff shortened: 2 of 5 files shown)


dealako:
@luismoriguerra — solid, focused PR. Ungating the logo upload is clean: the flag wrapper and its nine template bindings are gone without touching the umbrella `org_lens_private_release` gate elsewhere, `busy()` re-entrancy guards remain, and the follow-up commits show real iteration (stale comments, sizing fix, error-contract alignment, transient-status exclusion).

The error-mapping work is the right direction — SVG sanitizer reasons via `extractErrorMessage`, explicit 413/403 handling, and excluding 408/429 from the rejection bucket all match how the BFF actually behaves. A few toast branches still mislabel non-validation failures, and one pre-existing a11y pattern is now universal for every writer.

## Issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 4 issues: 409/404 responses get "Logo rejected" summary; body-less 4xx skips the documented fallback detail; client vs server oversize toasts disagree on summary; upload progress live region nested inside `role="button"`
- ⚪ Nit: 2 issues: duplicate 502 unit test; fallback-rejection test asserts summary only
- ❔ Question: 0 items

## Bot reconciliation

- **Agree** with [Cursor Bugbot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774262) and [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774680): 409 (ETag conflict) and 404 (org not found) from `org-identity.controller.ts` fall into the generic 4xx branch and get a "Logo rejected" summary while the detail correctly says to refresh or that the org is missing.
- **Agree** with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774658): `extractErrorMessage` returns `HttpErrorResponse.message` before the fallback when the body is empty, so the "This image could not be used…" path is unreachable for body-less rejections; the spec at line 391 only checks summary and would not catch a detail regression.
- **Agree** with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774705) on the nested `role="status"` inside the dropzone button — pre-existing when the flag was on, but ungating makes it affect every org writer in edit mode.
- **CodeRabbit** review was skipped (auto-reviews disabled on this repo).

## Final decision

The 409/404 summary mismatch is the highest-impact item: concurrent-edit and not-found paths are reachable in production and the wrong heading sends users toward swapping files instead of refreshing.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.ts
```diff
@@ -334,12 +312,32 @@ export class OrgProfileEditComponent implements OnInit {
...
+   */
   private toastForLogoError(error: unknown): { severity: string; summary: string; detail: string; life: number } {
     const status = error instanceof HttpErrorResponse ? error.status : 0;
     if (status === 403) {
       return { severity: 'error', summary: 'Permission denied', detail: 'You no longer have permission to edit this organization.', life: 5000 };
     }
+    if (status === 413) {
+      return { severity: 'error', summary: 'Logo too large', detail: 'Logo must be 2MB or smaller.', life: 5000 };
+    }
+    if (!isTransientHttpError(error) && status >= 400 && status < 500) {
```
dealako: [minor] 409/404 mislabeled as "Logo rejected"

Issue: The generic non-transient 4xx branch treats every validation-adjacent failure as a permanent file rejection.
Proof: BFF maps upstream 412 → 409 `{ error: 'This organization was updated elsewhere. Refresh the page and try again.' }` and org-missing → 404 `{ error: 'Organization not found' }` (`org-identity.controller.ts:248–254`). `isTransientHttpError` excludes 409/404 (`http-error.utils.ts:74–75`). This branch returns `summary: 'Logo rejected'` while `extractErrorMessage` surfaces the correct refresh/not-found detail.
Why it matters: Users hitting an ETag conflict see "Logo rejected" and may swap files instead of refreshing — the scenario called out in the busy-guard spec comment.
Fix: Add explicit `status === 409` and `status === 404` cases (summary like "Upload failed" or "Organization changed"), or exclude them from the rejection bucket. Add `it.each` cases with the exact BFF JSON bodies.

Agree with [Cursor Bugbot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774262) and [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774680).

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.ts
```diff
@@ -334,12 +312,32 @@ export class OrgProfileEditComponent implements OnInit {
...
+      return {
+        severity: 'error',
+        summary: 'Logo rejected',
+        detail: extractErrorMessage(error, 'This image could not be used. Try a different file.'),
```
dealako: [minor] Fallback detail unreachable for body-less rejections

Issue: The documented fallback `'This image could not be used. Try a different file.'` is not used when the response body is empty.
Proof: `extractErrorMessage` returns `error.message || fallback` after finding no usable body fields (`http-error.utils.ts:55`). A body-less `HttpErrorResponse({ status: 415 })` yields Angular's synthesized message (e.g. "Http failure response for … 415 …"), not the fallback. The spec at line 391 only asserts `summary: 'Logo rejected'`, not detail.
Why it matters: Users see a technical HTTP string instead of actionable guidance when the server rejects without a structured body.
Fix: Pass a tighter fallback only when `error.message` looks generic, or assert the intended detail in the test and adjust extraction logic accordingly.

Agree with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774658).

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
```diff
@@ -17,25 +17,16 @@
...
-        [class.border-blue-500]="isOrgLensUploadLogoEnabled() && logoDragActive()"
-        [class.border-transparent]="isOrgLensUploadLogoEnabled() && !logoDragActive()"
-        [attr.role]="isOrgLensUploadLogoEnabled() ? 'button' : null"
-        [attr.tabindex]="isOrgLensUploadLogoEnabled() ? 0 : null"
-        [attr.aria-label]="isOrgLensUploadLogoEnabled() ? 'Upload organization logo' : null"
-        [attr.aria-disabled]="isOrgLensUploadLogoEnabled() ? busy() : null"
+        class="relative flex h-20 w-20 shrink-0 cursor-pointer items-center justify-center rounded-md border-2 border-dashed"
+        [class.border-blue-500]="logoDragActive()"
+        [class.border-transparent]="!logoDragActive()"
+        role="button"
```
dealako, line 26: [minor] Upload progress live region nested inside button

Issue: `role="status"` / `aria-live="polite"` upload overlay (line 47) is a descendant of this dropzone's unconditional `role="button"`.
Proof: Button semantics are now always applied (lines 26–29); the live region remains inside the same element. Interactive roles flatten descendant semantics in the accessibility tree.
Why it matters: Screen-reader users may not hear "Uploading logo…" progress. Pre-existing when the flag was on; ungating exposes it to every org writer.
Fix: Move the live region outside the button wrapper (overlay can remain visually positioned with `aria-hidden` on decorative elements).

Agree with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1887#discussion_r3868774705).

dealako (changes requested):
Consolidated review — see summary comment for full context.

🔴 **Needs changes before approval** (4 minor issues)

1. **409/404 get "Logo rejected" summary** — BFF returns ETag conflict (409) and org-not-found (404) with retry/refresh copy, but `toastForLogoError` labels both as permanent file rejection. Add explicit branches or exclude these statuses from the rejection bucket; add unit tests with the real BFF envelopes.
2. **Body-less 4xx skips fallback detail** — `extractErrorMessage` returns Angular's synthesized `error.message` before the fallback when the body is empty, so body-less 415 shows a technical HTTP message instead of "This image could not be used…". Either handle empty-body rejections explicitly or tighten the test to assert the intended detail.
3. **Oversize toast summary mismatch** — client-side size check in `handleLogoFile()` uses summary "Error" while server 413 uses "Logo too large" for the same detail string (unchanged line ~254, not in diff).
4. **Live region inside button** — `role="status"` upload overlay is a descendant of the dropzone's now-unconditional `role="button"`; move the live region outside button semantics.

Agree with Cursor Bugbot and Copilot on items 1–2 and 4.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.ts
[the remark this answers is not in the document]
luismoriguerra: Confirmed and fixed in bcff2eac3.

You're right that `isTransientHttpError` was the wrong predicate — it classifies transport retryability, not whether the file is at fault. Verified against `org-identity.controller.ts`: the BFF maps an upstream 412 to **409** with *"This organization was updated elsewhere. Refresh the page and try again."* and a missing org to **404** *"Organization not found"*, and `isTransientHttpError` excludes both — so both printed a summary that contradicted their own detail.

Rejection is now an explicit status set (`LOGO_REJECTION_STATUSES = [400, 415, 422]`) rather than "non-transient 4xx". 404 and 409 keep the server's own actionable copy under a neutral "Upload failed" heading.

Covered by `it.each` using the exact BFF bodies. Mutation-checked: putting 409/404 back into the rejection bucket fails 4 tests.

luismoriguerra: Confirmed and fixed in bcff2eac3.

You're right that `isTransientHttpError` was the wrong predicate — it classifies transport retryability, not whether the file is at fault. Verified against `org-identity.controller.ts`: the BFF maps an upstream 412 to **409** with *"This organization was updated elsewhere. Refresh the page and try again."* and a missing org to **404** *"Organization not found"*, and `isTransientHttpError` excludes both — so both printed a summary that contradicted their own detail.

Rejection is now an explicit status set (`LOGO_REJECTION_STATUSES = [400, 415, 422]`) rather than "non-transient 4xx". 404 and 409 keep the server's own actionable copy under a neutral "Upload failed" heading.

Covered by `it.each` using the exact BFF bodies. Mutation-checked: putting 409/404 back into the rejection bucket fails 4 tests.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.ts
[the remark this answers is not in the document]
luismoriguerra: Confirmed and fixed in bcff2eac3 — and thank you for catching *why* the test passed.

`extractErrorMessage` ends with `error.message || fallback` (`http-error.utils.ts:55`), and Angular always synthesizes a non-empty `error.message`, so the fallback was unreachable for a body-less response. My test asserted only `summary`, which is exactly how it stayed green while broken.

The body is now checked for a server-authored message first — string body, `message`, `error`, or `errors[].message`, every shape the BFF emits — and only then does `extractErrorMessage` run; otherwise the caller's fallback is used directly. I left the shared helper untouched rather than changing behaviour for its other callers.

The test now asserts the exact detail and explicitly asserts the toast does **not** contain `Http failure response`. Mutation-checked: removing the guard fails it.

luismoriguerra: Confirmed and fixed in bcff2eac3 — and thank you for catching *why* the test passed.

`extractErrorMessage` ends with `error.message || fallback` (`http-error.utils.ts:55`), and Angular always synthesizes a non-empty `error.message`, so the fallback was unreachable for a body-less response. My test asserted only `summary`, which is exactly how it stayed green while broken.

The body is now checked for a server-authored message first — string body, `message`, `error`, or `errors[].message`, every shape the BFF emits — and only then does `extractErrorMessage` run; otherwise the caller's fallback is used directly. I left the shared helper untouched rather than changing behaviour for its other callers.

The test now asserts the exact detail and explicitly asserts the toast does **not** contain `Http failure response`. Mutation-checked: removing the guard fails it.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
```diff
@@ -17,25 +17,16 @@
...
+        tabindex="0"
+        aria-label="Upload organization logo"
+        [attr.aria-disabled]="busy()"
```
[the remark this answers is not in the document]
luismoriguerra: Agreed and fixed in bcff2eac3.

I had deferred this in the PR description as pre-existing. That was the wrong call — ungating is precisely what turns it from a flagged-in edge case into something every org writer hits, which is the argument you're both making.

The live region moves outside the dropzone; the spinner stays inside as an `aria-hidden` visual with no text of its own. I also changed it from insert/remove to a persistent region whose text toggles, since a live region needs to be in the DOM *before* its content changes to announce reliably — the previous version inserted the region with its text already in place.

The existing a11y test now asserts the region is **not** a descendant of the dropzone, and that it persists with empty text once the upload settles.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.html
luismoriguerra: Agreed and fixed in bcff2eac3.

I had deferred this in the PR description as pre-existing. That was the wrong call — ungating is precisely what turns it from a flagged-in edge case into something every org writer hits, which is the argument you're both making.

The live region moves outside the dropzone; the spinner stays inside as an `aria-hidden` visual with no text of its own. I also changed it from insert/remove to a persistent region whose text toggles, since a live region needs to be in the DOM *before* its content changes to announce reliably — the previous version inserted the region with its text already in place.

The existing a11y test now asserts the region is **not** a descendant of the dropzone, and that it persists with empty text once the upload settles.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.ts
```diff
@@ -334,15 +313,67 @@ export class OrgProfileEditComponent implements OnInit {
...
+   * "Logo rejected — This organization was updated elsewhere", where the summary contradicts the
+   * detail and sends the user off to swap a file that was never at fault.
+   */
   private toastForLogoError(error: unknown): { severity: string; summary: string; detail: string; life: number } {
     const status = error instanceof HttpErrorResponse ? error.status : 0;
     if (status === 403) {
       return { severity: 'error', summary: 'Permission denied', detail: 'You no longer have permission to edit this organization.', life: 5000 };
     }
+    if (status === 413) {
+      return { severity: 'error', summary: 'Logo too large', detail: 'Logo must be 2MB or smaller.', life: 5000 };
```
audigregorie, line 335: **Minor — the oversize condition speaks in two voices.** This new 413 branch says `summary: 'Logo too large'`, but the client-side size check for the identical condition (`handleLogoFile`, line 255) still says `summary: 'Error'` with the same detail string (`'Logo must be 2MB or smaller.'`). Whichever side rejects the file, the user is told the same thing two different ways. Optional polish while you're in here — align the client-side toast:

```ts
// line 255
this.messageService.add({ severity: 'error', summary: 'Logo too large', detail: 'Logo must be 2MB or smaller.' });
```

audigregorie (commented):
## Code Review Summary

Clean, well-verified ungating. The error-mapping rewrite at HEAD (bcff2eac3) correctly implements what the previous review round asked for: rejection copy is scoped to a narrow allowlist (`LOGO_REJECTION_STATUSES` = 400/415/422), 404/409 keep the BFF's own actionable copy, and body-less responses no longer leak Angular's synthesized `Http failure response for …` string into the toast. I re-verified each fix against the server contract — `BaseApiError.toResponse()` emits `{ error, code }`, `ServiceValidationError` adds `errors[].message`, `org-identity.controller.ts` emits 403/404/409/502 with plain `{ error }` bodies, and `orgs.route.ts` mints the 413 via `handleLogoUploadParseError`. All three prior threads (cursor / Copilot / dealako) are genuinely resolved at HEAD; the outstanding CHANGES_REQUESTED is stale relative to bcff2eac3 and looks ready for re-review.

Also verified: no dangling `isOrgLensUploadLogoEnabled` references anywhere; the umbrella `ORG_LENS_PRIVATE_RELEASE_FLAG` still gates the person detail drawer, its service, and org project detail as claimed; `org-profile-edit` is only instantiated behind `canEdit()` with `enterEditMode()` re-checking the writer grant, so dropping the flag removes no real authorization boundary; and the e2e flag-pin removal is consistent now that the input renders unconditionally.

### Minor — outside the diff

**Stale PR description (two places).** The status table and prose still say 408/429 are "excluded via the repo's existing `isTransientHttpError`" — HEAD no longer uses that helper; the `LOGO_REJECTION_STATUSES` allowlist plus a default retry branch replaced it (which is the design the review round asked for, and arguably the better one). And "Notes for reviewers" still lists the nested live region as deliberately *not* fixed, but bcff2eac3 moved it outside the dropzone, template comment and test pin included. Worth refreshing the description so the merged-PR record matches what actually ships — right now it documents the pre-fix design.

### What's done well

- The test fixtures mirror the three real BFF contract shapes (`{ error, code }`, `{ error, code, errors[] }`, plain string body) instead of a convenient fabricated one — and they assert the actual detail strings, so reverting the extraction genuinely cannot stay green.
- The `hasServerMessage` JSDoc documents *why* the guard exists (`extractErrorMessage`'s `error.message || fallback` tail makes its fallback unreachable for body-less responses), so a future reader can't "simplify" it back into the bug.
- Live-region handling is textbook: persistent `role="status"` region outside the `role="button"` flattening boundary, content toggled rather than the region itself, spinner purely decorative under `aria-hidden` — all pinned by tests.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.ts
luismoriguerra: Good catch — and my 413 branch is what created the split, so taking it.

Fixed in 94ca6a562, plus the one next to it: the media-type check had the identical problem (`summary: 'Error'` client-side against `'Logo rejected'` for the equivalent 415 upstream), so both are aligned rather than just the one you flagged.

```
line 251  'Error' -> 'Logo rejected'    (matches 415 -> LOGO_REJECTION_STATUSES)
line 255  'Error' -> 'Logo too large'   (matches the 413 branch)
```

Worth noting these were the only two generic summaries left in the component — every other toast already names its condition (`Permission denied`, `Save failed`, `Small image`, `Large image`, `Upload failed`), so this brings them in line with the file's own convention rather than inventing one.

Both tests now assert the summary instead of only the detail, so a future divergence fails rather than passing quietly. Mutation-checked: reverting either summary fails its test.

MRashad26 (commented):
## Code Review

**Overview:** Removes the `isOrgLensUploadLogoEnabled` / `org_lens_private_release` gate from the Company Logo Upload affordance (backend now released and verified), and hardens the failure path the gate had been hiding: server-authored SVG-sanitizer rejection reasons are now correctly extracted and surfaced (previously read a key the BFF never emits, so the feature was silently inert), retryable vs. non-retryable statuses get distinct copy, and a `border-2`/sizing overflow bug (invisible while flag-gated off) is fixed.

Ran this against the code-standards-enforcer checklist (Angular/signals/template rules) plus this repo's actual `CLAUDE.md` + `.claude/rules/` (which differ slightly from the generic checklist in a few particulars — e.g. interfaces live in `interfaces/*.ts`, not `.d.ts`, in this repo).

### Code Quality & Style — clean
- Component field order matches `.claude/rules/component-organization.md` (injections → outputs → forms → signals → computed → public → protected → private init → private helpers).
- New `LOGO_REJECTION_STATUSES` constant correctly placed in `packages/shared/src/constants/`, not locally in `apps/lfx-one/` (per "all shared constants live in `@lfx-one/shared`").
- No functions called in template bindings beyond signals/computed — template rule respected.
- `<p-button>` used directly rather than through an LFX wrapper (`styling.md` says route through `shared/components/` wrappers) — pre-existing from before this PR, not introduced here, so not a new violation.
- Verified clean: `yarn lint:check`, `yarn check-types` (shared pkg — app has no dedicated check-types script), and the component's unit suite (31/31 passing) on the PR branch in an isolated worktree.

### Specific Suggestions
- **PR description is stale on one a11y item.** The description lists "the upload live region sits inside `role=\"button\"` (presentational children)" as a pre-existing issue *deliberately not fixed* here — but the diff actually moves the live region outside the dropzone (`org-profile-edit.component.html`, with an explicit comment: "Outside the dropzone on purpose…") and adds a test asserting exactly that (`org-profile-edit.component.spec.ts`: `expect(root.querySelector('[data-testid="org-profile-edit-logo-dropzone"] [role="status"]')).toBeNull()`). Worth updating the description so a future reviewer doesn't file a duplicate ticket for something already resolved. The other two disclosed a11y items (aria-label shadowing the img `alt`, no focus-visible ring) are correctly still present and correctly left out of scope.
- No functional issues found in the error-mapping logic (`toastForLogoError`/`logoErrorDetail`/`hasServerMessage`) — the status-code branches (403/413/`LOGO_REJECTION_STATUSES`/404,409/default) are disjoint and match `extractErrorMessage`'s actual body-shape precedence, and `hasServerMessage` correctly works around `extractErrorMessage`'s unreachable fallback (Angular always synthesizes a non-empty `error.message`).

### Potential Issues & Risks
- None blocking. Grepped the rest of the app for dangling references to the removed gate — only the umbrella `ORG_LENS_PRIVATE_RELEASE_FLAG` remains, still correctly used by `org-project-detail.component.ts` (out of scope, per PR description).
- Two pre-existing a11y gaps remain, disclosed and correctly scoped out.

### Test Coverage — strong
- New tests cover: the specific 413/403 mapping, the three real BFF error-body shapes (top-level `error`, field-level `errors[].message`, plain string) for the rejection path, the "no usable reason" fallback (asserting Angular's synthesized HTTP debug string never reaches the toast), 404/409 exclusion from the rejection wording, and 408/429/0/5xx retry-copy preservation.
- Mutation-checked per the PR description (reverting the extraction fails 3 tests, removing the `busy()` guard fails 1) — good discipline.

### Security Considerations
- No new permission surface — `org-profile-edit` isn't a route; it's only instantiated behind `canEdit()` (org writer), independently re-checked in `enterEditMode()`. Confirmed the removed gate was redundant with this pre-existing authorization check, not a substitute for it.
- Widening the failure-copy detail to surface `extractErrorMessage(error, fallback)` output in a toast is scoped to server-authored sanitizer-rejection text (e.g. "unsupported CSS selector"), not raw response bodies — reasonable to display to the org's own writer.

**Verdict:** no blockers. One documentation nit (stale a11y claim in the PR description) worth a quick fix before merge, but not code-blocking.

MRashad26 (approved):
Approving — no blocking issues found. Verified clean `lint:check`, `check-types`, and the component's unit suite (31/31 passing) on the PR branch in an isolated worktree; confirmed no dangling references to the removed feature flag and that the authorization boundary (`canEdit()`) makes the gate removal safe. One non-blocking doc nit called out in my review comment above (PR description is stale on one already-fixed a11y item).

packages/shared/src/constants/org-profile.constants.ts
```diff
@@ -41,3 +41,13 @@ export const MIN_ORG_LOGO_DIMENSION_PX = 128;
...
+
+/**
+ * Upload statuses that mean the file itself is unusable, so the user must supply a different one
+ * rather than retry. 400 carries member-service's sanitizer rejection (an SVG whose CSS it cannot
+ * reproduce faithfully); 415 and 422 mean the same thing for content type and shape.
+ *
+ * Deliberately narrow: the BFF also returns 404 for a missing org and 409 for an If-Match miss
+ * (`org-identity.controller.ts`), and neither is a problem with the image.
+ */
+export const LOGO_REJECTION_STATUSES: readonly number[] = [400, 415, 422];
```
dealako, line 53: [minor] 422 rejection status is never exercised by a test

Issue: `LOGO_REJECTION_STATUSES = [400, 415, 422]`, but the spec only exercises 400 (the three SVG "surfaces the upstream reason" cases) and 415 (the body-less fallback). No test ever sends a 422.
Proof: 422 is reachable — an upstream `MicroserviceError(422)` matches no branch in `uploadLogo`'s catch (`org-identity.controller.ts:243–261` handles only 403/404/412/5xx), so it falls through `next(error)` → `apiErrorHandler` which emits `error.statusCode` verbatim. A mutation dropping 422 from this array would regress it into the generic "Upload failed / Please try again" branch and the whole suite stays green.
Why it matters: the narrow rejection allowlist is the core of this hardening change; its most fragile member has no regression guard.
Fix: extend the `it.each` rejection table with a 422 case asserting `summary: 'Logo rejected'`, mirroring the 415 fallback assertion.

apps/lfx-one/src/app/modules/dashboards/org/org-profile/org-profile-edit.component.spec.ts
```diff
@@ -319,69 +315,143 @@ describe('OrgProfileEditComponent — logo upload', () => {
...
+    await fixture.whenStable();
+
+    expect(toastAdd).toHaveBeenCalledWith(expect.objectContaining({ summary: 'Logo rejected', detail: 'This image could not be used. Try a different file.' }));
+    expect(toastAdd).not.toHaveBeenCalledWith(expect.objectContaining({ detail: expect.stringContaining('Http failure response') }));
+  });
+
+  // The BFF gives these their own actionable copy, so routing them through the rejection branch
+  // would print a summary that contradicts the detail and send the user off to swap a file that
+  // was never at fault (org-identity.controller.ts).
+  it.each([
```
dealako, line 413: [nit] 404/409 body-less fallback branch is unverified

Issue: `logoErrorDetail(error, 'Unable to upload logo. Please try again.')` backs the 404/409 branch, but both "does not label status %i as a rejected logo" cases supply a populated body (`{ error: serverMessage }`), so `extractErrorMessage` always returns the server string. The path where `hasServerMessage()` is false for a 404/409 (fallback surfaces instead of Angular's HTTP debug string) is never hit.
Proof: `org-profile-edit.component.spec.ts` `it.each([[409,…],[404,…]])` always sets `error: { error: serverMessage }`; no body-less 404/409 case exists.
Why it matters: `hasServerMessage`'s stated purpose is to keep "Http failure response for …" out of the toast; that guard is proven for the rejection branch (415) but not for this "Upload failed" branch, so a regression there would go uncaught.
Fix: add a body-less 404 or 409 case asserting `detail === 'Unable to upload logo. Please try again.'` and NOT containing "Http failure response".

dealako (approved):
Re-review at HEAD (`94ca6a5`). My prior `CHANGES_REQUESTED` (commit `8d94054`) raised four minor issues — all four are now resolved and verified against the BFF contract, so this supersedes that review.

**Prior findings, all fixed:**

1. **409/404 mislabeled as "Logo rejected"** → fixed. Rejection is now an explicit allowlist `LOGO_REJECTION_STATUSES = [400, 415, 422]`; 404/409 keep the server's own copy under a neutral "Upload failed" heading. Verified against `org-identity.controller.ts:248–254` (404 → `{ error: 'Organization not found' }`, 412 → 409 `{ error: 'This organization was updated elsewhere…' }`).
2. **Body-less 4xx skipped the documented fallback** → fixed. New `hasServerMessage` guard checks the body for a server-authored message before calling `extractErrorMessage`, keeping Angular's synthesized `Http failure response for …` string out of the toast. Pinned by the body-less 415 test asserting the detail.
3. **Client vs server oversize toast summary mismatch** → fixed, plus the equivalent media-type split (`'Error'` → `'Logo rejected'`) aligned in the same pass. Both tests now assert `summary`.
4. **Live region nested inside `role="button"`** → fixed. Region moved outside the dropzone, persists with text toggling (not insert/remove) so it announces reliably; spinner stays as an `aria-hidden` visual. Pinned by a test asserting no `[role="status"]` inside the dropzone.

**This round — two test-only gaps, non-blocking:**

- 🟡 Minor: `422` is in `LOGO_REJECTION_STATUSES` but never exercised by a test — dropping it from the array would regress a reachable member-service response into the generic retry branch and stay green. See inline comment.
- ⚪ Nit: the 404/409 body-less fallback branch (`hasServerMessage` false) is never asserted for those statuses — the guard is proven for the rejection branch (415) but not the "Upload failed" branch. See inline comment.

Security/authorization: verified the removed flag was UI-only. `OrgProfileComponent.canEdit()` (role-grant `writerSet`) and the server's own member-service authz are both untouched, so no authorization boundary was removed. The widened error copy surfaces only server-authored sanitizer reason text, not PII or internal records.

Also worth a quick description refresh (not code-blocking, already flagged by audigregorie/MRashad26): the PR body still says 408/429 are excluded via `isTransientHttpError` and that the live region was left unfixed — both are stale relative to HEAD.

## Issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 1 issue: `422` rejection status untested (mutation survives)
- ⚪ Nit: 1 issue: 404/409 body-less fallback branch unverified
- ❔ Question: 0 items

## Bot reconciliation

- **Cursor Bugbot** and **Copilot** both flagged the 409/404 mislabel and the nested live region on earlier commits — both are now resolved at HEAD, matching their guidance.
- **CodeRabbit** review was skipped (auto-reviews disabled on this repo).

## Final decision

✅ **Approved with minor comments** — clearing my stale `CHANGES_REQUESTED`. The two remaining items are test-coverage polish, not correctness defects. Nice iteration on the error-contract work.

dealako:
@luismoriguerra — re-reviewed at HEAD (`94ca6a5`), and this is ready. Clean, tightly-scoped ungating with genuinely good iteration on the failure path. All four minor items from my earlier `CHANGES_REQUESTED` (commit `8d94054`) are resolved, and I re-verified each against the actual BFF contract rather than the diff alone.

**Prior findings — all fixed and verified:**

1. **409/404 no longer mislabeled** — rejection is now the explicit allowlist `LOGO_REJECTION_STATUSES = [400, 415, 422]`; 404/409 keep their own actionable copy under a neutral "Upload failed" heading. Confirmed against `org-identity.controller.ts:248–254` (404 → "Organization not found", 412 → 409 "updated elsewhere").
2. **Body-less 4xx fallback reachable** — the new `hasServerMessage` guard checks for a server-authored body before `extractErrorMessage`, keeping Angular's synthesized `Http failure response for …` out of the toast. Mirrors `extractErrorMessage`'s `error.message || fallback` tail exactly.
3. **Oversize toast one voice** — client-side and server 413 both say "Logo too large"; the equivalent media-type split was aligned in the same pass.
4. **Live region unnested** — moved outside `role="button"`, persists with text toggling so it announces reliably; spinner stays `aria-hidden`.

**This round — two test-only gaps, non-blocking:**

- 🟡 `422` is in the rejection allowlist but never exercised — a mutation dropping it stays green. It's reachable (upstream `MicroserviceError(422)` falls through to `apiErrorHandler` verbatim).
- ⚪ The 404/409 body-less fallback branch (`hasServerMessage` false) is never asserted for those statuses.

Both are inline. Neither blocks merge.

## Issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 1 issue: `422` rejection status untested (mutation survives)
- ⚪ Nit: 1 issue: 404/409 body-less fallback branch unverified
- ❔ Question: 0 items

## Final decision

✅ **Approved with minor comments** — clearing my stale `CHANGES_REQUESTED`. (One more housekeeping note others already raised: the PR description is stale on the 408/429 `isTransientHttpError` mechanism and the live-region fix — worth a refresh so the merged record matches what ships.)

PR #1888, opened 2026-08-27
fix(intercom): replace jira support links with fin messenger

### LINKED ISSUE #1857: Remove all links to Jira and open Fin messenger instead.
### What happened?

On this page https://app.lfx.dev/project/mailing-lists?project=agentic-ai-foundation as I was impersonating a user without Mailing List permissions, I received this message:
"Mailing list service hasn't been set up yet for this project. To create a mailing list for your project, contact support."
Contact support sends me to "https://jira.linuxfoundation.org/plugins/servlet/desk"

The ask in this ticket is to search ALL code/components of self serve and identify any link to Jira - we have never used Jira for self serve support. All support should call the Fin Intercom messenger instead.

Image

### Command / Request

```shell

```

### Relevant log output

```shell

```

audigregorie:
## Summary

GH-1857 asked to remove every Jira support link from Self Serve — support was never handled via Jira — and open the Fin Intercom messenger instead. The hard-coded Jira anchors are gone, but the shared `lfxOpenIntercom` directive (used by ~10 surfaces, including the mailing-lists "contact support" CTA from the issue) still fell back to opening the Jira service-desk portal whenever Intercom wasn't already booted — exactly the impersonation scenario in the report. This PR removes that Jira fallback entirely: support CTAs now boot the Fin messenger anonymously on demand, the last hard-coded Jira link (expired memberships "Contact Us") becomes an `lfxOpenIntercom` button, and the unused Jira support URL is dropped from all environment configs.

Closes #1857

### Support links open the Fin messenger

| Before | After |
|---|---|
| The expired-memberships "Contact Us" link opened a Jira service-desk portal in a new tab, and every other support CTA (mailing lists, not-found pages, profile dialogs, lens switcher) did the same whenever the chat widget wasn't already running — e.g. while impersonating a user. | All support CTAs open the Fin support messenger in place, inside the app. No surface sends users to Jira anymore. |

### Support click failure handling

| Before | After |
|---|---|
| If the chat widget couldn't load (e.g. ad-blocker) or support chat wasn't configured, clicking a support CTA failed silently — nothing happened and the user got no feedback. | The click fails visibly: the user sees a "Support Unavailable" toast asking them to try again later. Page-load widget failures stay silent — only clicks surface a toast. |

## Technical changes

`apps/lfx-one/src/app/shared/directives/open-intercom.directive.ts` — OpenIntercomDirective rewrite

- Replaces the Jira fallback (`window.open(environment.urls.support)`) with `intercomService.openMessenger()`, reading the Intercom app id from the `TransferState`-backed runtime config instead of the static environment file
- Shows a "Support Unavailable" error toast via `MessageService` when no app id is configured, or when the widget script fails to load after the click

`apps/lfx-one/src/app/shared/services/intercom.service.ts` — IntercomService

- Adds `openMessenger(appId, onLoadError?)`: boots Intercom anonymously on demand when startup boot was skipped (impersonation, public pages, missing JWT claim), then shows the messenger — a fresh boot queues behind the stub and replays before `show` once the script loads
- Adds an optional `onLoadError` callback invoked from the widget script's `onerror`, including for a click landing while a startup boot is still in flight; it is cleared on successful load and never registered by page-load boots, so ad-blocked users are only toasted after a click
- Script failures still log to console and DataDog RUM; fire-and-forget semantics unchanged

`apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html, .ts` — Org memberships

- Replaces the expired-memberships "Contact Us" Jira anchor with a `lfxOpenIntercom` button and imports `OpenIntercomDirective`
- Adds `data-testid="org-memberships-expired-contact-us-button"` for reliable test targeting

`apps/lfx-one/src/environments/environment.ts, environment.dev.ts, environment.staging.ts, environment.prod.ts` — Environment configs

- Removes the now-unused `urls.support` Jira service-desk URL from all four environments (verified no remaining consumers)

`apps/lfx-one/src/app/shared/services/intercom.service.spec.ts` — Tests

- New spec covering `openMessenger()`: anonymous boot with boot-before-show queue ordering, single boot across repeated clicks, empty-app-id refusal, script-failure callback with retry, callback clearing on successful load, and a click landing while a startup boot is still in flight (7 tests)

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
@@ -321,13 +321,7 @@
                         Renew
                       </a>
                     } @else {
-                      <a
-                        href="https://jira.linuxfoundation.org/servicedesk/customer/portal/4"
-                        target="_blank"
-                        rel="noopener noreferrer"
-                        class="text-sm text-gray-600 hover:text-gray-900 transition-colors">
-                        Contact Us
-                      </a>
+                      <button type="button" lfxOpenIntercom class="text-sm text-gray-600 hover:text-gray-900 transition-colors">Contact Us</button>
                     }
                   </td>
                 </tr>
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
@@ -209,7 +209,7 @@
                   <td class="text-sm text-gray-700">{{ membership.boardMembers }}</td>
                   <td class="text-sm text-gray-700">{{ membership.committeeMembers }}</td>
                   <td>
-                    @if (membership.orgProjects > 0) {
+                    @if (membership.orgProjects &gt; 0) {
                       <span class="text-sm font-medium text-blue-600">{{ membership.orgProjects }}</span>
                     } @else {
                       <span class="text-sm text-gray-700">0</span>
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
@@ -209,7 +209,7 @@
                   <td class="text-sm text-gray-700">{{ membership.boardMembers }}</td>
                   <td class="text-sm text-gray-700">{{ membership.committeeMembers }}</td>
                   <td>
-                    @if (membership.orgProjects &gt; 0) {
+                    @if (membership.orgProjects > 0) {
                       <span class="text-sm font-medium text-blue-600">{{ membership.orgProjects }}</span>
                     } @else {
                       <span class="text-sm text-gray-700">0</span>
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.html
@@ -321,7 +321,13 @@
                         Renew
                       </a>
                     } @else {
-                      <button type="button" lfxOpenIntercom class="text-sm text-gray-600 hover:text-gray-900 transition-colors">Contact Us</button>
+                      <button
+                        type="button"
+                        lfxOpenIntercom
+                        data-testid="org-memberships-expired-contact-us-button"
+                        class="text-sm text-gray-600 hover:text-gray-900 transition-colors">
+                        Contact Us
+                      </button>
                     }
                   </td>
                 </tr>
--- a/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/org/org-memberships/org-memberships.component.ts
@@ -15,6 +15,7 @@ import { OrgLensMembershipsService } from '@services/org-lens-memberships.servic
 import { CardComponent } from '@components/card/card.component';
 import { TableComponent } from '@components/table/table.component';
 import { EmptyStateComponent } from '@components/empty-state/empty-state.component';
+import { OpenIntercomDirective } from '@shared/directives/open-intercom.directive';
 import type {
   OrgActiveMembershipsResponse,
   OrgDiscoverOpportunitiesResponse,
@@ -32,7 +33,7 @@ import { environment } from '@environments/environment';
 @Component({
   selector: 'lfx-org-memberships',
   standalone: true,
-  imports: [FormsModule, RouterLink, TableComponent, TooltipModule, SelectModule, InputTextModule, CardComponent, EmptyStateComponent],
+  imports: [FormsModule, RouterLink, TableComponent, TooltipModule, SelectModule, InputTextModule, CardComponent, EmptyStateComponent, OpenIntercomDirective],
   templateUrl: './org-memberships.component.html',
 })
 export class OrgMembershipsComponent {
--- a/apps/lfx-one/src/app/shared/directives/open-intercom.directive.spec.ts
+++ b/apps/lfx-one/src/app/shared/directives/open-intercom.directive.spec.ts
@@ -1,0 +1,85 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { Component, TransferState } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { DEFAULT_RUNTIME_CONFIG, RUNTIME_CONFIG_KEY } from '@app/shared/providers/runtime-config.provider';
+import { IntercomService } from '@services/intercom.service';
+import { MessageService } from 'primeng/api';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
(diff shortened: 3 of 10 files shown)


ahmedomosanya:
**LGTM** — approving. The Jira fallback in `OpenIntercomDirective` was the real bug (the hard-coded anchors were only half of it), and the on-demand anonymous boot is the right fix.

Things I verified rather than assumed:

- No consumer of `environment.urls.support` remains in `apps/` or `packages/`, and no `jira.linuxfoundation.org` support URL survives anywhere.
- `runtimeConfig` is built for every SSR request regardless of auth (`server.ts:465`), so the public not-found surfaces that use `lfxOpenIntercom` do get an app id.
- An anonymous boot leaves `isBootRequested` stuck true, but `ImpersonationBannerComponent` calls `window.location.reload()` after `/api/impersonate/stop`, so it can't leak into the following identified boot.
- `MessageService` resolves on all ten `lfxOpenIntercom` surfaces — root provider in `app.config.ts:62` with the global `<p-toast />`, and the one local override (`OrgProfileEditComponent`) renders its own toast.

Nothing blocking. Three things worth considering:

**1. A retry after a failed script load replays `boot` twice.** `onerror` resets `isBootRequested` and removes the script, but leaves `window.Intercom` and its queue intact — the first attempt's `['boot', …], ['show']` are still in `stub.q`. A second click re-runs `initializeIntercomFunction`, sees a function on `window.Intercom`, and appends `reattach_activator` + `update` + a second `boot` + `show` to the same queue. Intercom tolerates the double boot, but should we clear the queue so the retry starts genuinely fresh?

```typescript
script.onerror = (error) => {
  this.isBootRequested = false;
  script.remove();
  this.scriptElement = null;
  delete window.Intercom;
  delete window.intercomSettings;
  // …existing logging / RUM / onLoadError
};
```

The existing retry test would still pass and could gain a queue assertion.

**2. The toast copy points at a dead end for content-blocker users.** `widget.intercom.io` is on common blocklists (EasyPrivacy), which is probably the most frequent real cause of `onerror`. For those users "Please try again later" never comes true, and with the Jira portal gone there's no remaining route to support. Could the detail carry an actionable alternative — a `mailto:` or the support page — so the visible failure is also a usable one?

**3. The new user-facing behavior has no test.** The service spec is thorough, but every new user-visible decision lives in `OpenIntercomDirective`: the empty-`intercomAppId` early return, the toast, and the callback wiring. Worth noting that `openMessenger('')` — the one empty-app-id case covered — is unreachable from production, since the directive returns before calling the service. Would a small directive spec with a stubbed `TransferState` and `MessageService` be worth adding?

Two nits:

- `openMessenger` assigns `this.onLoadError = onLoadError` unconditionally when the script hasn't loaded, so a call without a callback would silently clear one registered by an earlier click. Nothing is broken today, but `if (onLoadError && !this.isLoaded)` would make the field's ownership clearer.
- Since `onLoadError` lives on a root singleton it captures the clicking component's injector. A click on the org profile edit page (own `MessageService` + own toast) followed by navigating away before the script errors sends the deferred toast to a destroyed message service — the silent failure this PR removes. Very narrow, and only that surface.

One question: booting anonymously while impersonating means a staff-initiated conversation creates an unidentified lead with no link to the user being helped. I assume that's the deliberate trade against binding the conversation to the wrong identity — is it worth passing a `custom_attributes` marker so support can tell these apart?

audigregorie:
Thanks for the thorough pass, @ahmedomosanya — addressed in `7c8c0cf1d` (pushed):

**1. Retry double-boot** — fixed. `onerror` now also deletes `window.Intercom` and `window.intercomSettings`, so a retry builds a fresh stub and queues exactly one `boot` + `show`. The existing retry spec gained a queue assertion proving the failed attempt's commands no longer replay.

**2. Toast dead end for content-blocker users** — leaving the copy as-is in this PR. There is no sanctioned fallback contact anywhere in the codebase now (the Jira URL was the last one), the global toast renders plain text only, and picking the replacement contact is a product call rather than something to invent here.

**3. Directive spec** — added `open-intercom.directive.spec.ts` with stubbed `TransferState` / `MessageService` / `IntercomService`: empty app id → toast and no service call; app id present → `openMessenger(appId, callback)`; invoking the callback → toast. It locks in the guard that makes the service's `openMessenger('')` case the backstop you noted.

**Nits:**
- Left the unconditional `onLoadError` assignment as-is — with the single caller always passing a callback there is no behavior change today, and the unconditional clear is the safer default if a future caller ever invokes without one (it cannot leak a stale callback into a destroyed component).
- Acknowledged on the injector capture — as you said, very narrow and only the org-profile-edit surface; the failure mode is a silent no-op toast, i.e. parity with pre-PR behavior for that surface.

**Question:** yes — the anonymous boot while impersonating is the deliberate trade; binding a staff-initiated conversation to the impersonated identity would misattribute it. A `custom_attributes` marker is a good idea and would make a sensible follow-up if the team wants these leads distinguishable.

MRashad26 (approved):
Reviewed the full diff (10 files, +256/-19) and ran the code-standards-enforcer agent against this repo's actual rule surface (`.claude/rules/*.md` + `CLAUDE.md`). Both agree: clean, no issues.

**What this does:** Removes the last Jira support links from Self Serve (GH-1857). The shared `lfxOpenIntercom` directive (used by ~10 surfaces) previously fell back to `window.open(environment.urls.support)` whenever Intercom hadn't already been booted — exactly the impersonation/public-page scenario the issue reported. Now `OpenIntercomDirective` reads `intercomAppId` from the SSR-safe `TransferState`-backed runtime config and calls a new `IntercomService.openMessenger(appId, onLoadError)`, which boots Intercom anonymously on demand if startup boot was skipped, then shows the messenger. A click now fails visibly (a "Support Unavailable" toast) instead of silently when no app id is configured or the widget script fails to load — page-load boot failures stay silent by design, since `onLoadError` is only ever registered from a click. The last hardcoded Jira anchor (expired-memberships "Contact Us") becomes an `lfxOpenIntercom` button, and the now-dead `urls.support` key is dropped from all four environment files.

**Verified correct (independent read + spot checks):**
- Confirmed the PR's claim that no consumers of `environment.urls.support` remain anywhere in the codebase — extracted the full PR-branch tree via `git archive` and grepped for `urls.support`/`jira.linuxfoundation`: zero hits under `apps/lfx-one/src`. No typed `Environment` interface constrains the `urls` shape either, so the removal can't cause a type error elsewhere.
- `openMessenger`'s boot/show/error-callback lifecycle is internally consistent: a click while a startup boot is already in flight doesn't double-boot (`isBootRequested` guard) but still registers `onLoadError` if the script hasn't loaded yet, so a click landing mid-boot still gets toasted on failure — covered by its own test. `onLoadError` is cleared on both successful load and after firing once, and a failed load additionally clears `window.Intercom`/`window.intercomSettings` so a retry boots from a genuinely fresh stub queue rather than replaying a dead one alongside the new attempt.
- The `window.Intercom`/`window.intercomSettings` deletes in the script's `onerror` handler are unreachable during SSR — that handler is only ever registered inside `loadIntercomScript()`, which itself early-returns under `typeof window === 'undefined'`.
- New test coverage (3 directive + 7 service specs) exercises exactly the scenarios the PR describes, including the click-during-in-flight-boot race and the retry-after-failure path.

**Enforcer's additional verification:** ran `yarn lint:check` against the actual PR branch (via a disposable worktree, no changes to the local checkout's current branch) — clean; ran the new spec files directly — 10/10 pass.

No issues found. Approving.

ahmedomosanya (approved):

PR #1889, opened 2026-08-27
fix(rewards): scope impersonated rewards to target users

prabodhcs:
## Summary

- Resolve the impersonated Rewards subject exclusively from trusted server-side session state, validate one target identity, and reuse it for profile and promotion reads.
- Retrieve and validate the complete target promotion collection while representing profile and promotion availability independently instead of substituting actor or default data.
- Keep Rewards read-only during impersonation: Claim and Redeem remain visible but disabled, direct write requests are rejected, and existing coupon Copy remains available.
- Preserve normal-user reward calculations and eligibility messaging, including explicit zero-point and points-unavailable states.
- Redact sensitive upstream response bodies and drain discarded streams so failed requests do not expose benefit credentials or reduce connection reuse.
- Refresh UI state when the effective user or impersonation state changes and add focused server, shared, component, and browser coverage.

## Verification evidence

All production observations below are sanitized. No real names, usernames, account identifiers, Salesforce identifiers, access tokens, or coupon values are included.

- Focused server and shared verification passed: 81 tests across eight files before final review corrections.
- Final review regressions passed: four reward utility tests, one coupon component test, and two API Gateway response-handling tests.
- Focused Chromium Rewards coverage passed six route-stub scenarios. The opt-in authenticated live-BFF scenario was skipped because its environment inputs were not supplied to that run.
- License headers, formatting, lint, type checking, and the production build passed. Lint retained one pre-existing warning and reported no errors.
- A fetch-only production probe confirmed that target-scoped Rewards differed from the staff actor baseline and included the target's expected 40% subscriber promotion. No production mutation was attempted.
- Visual comparisons used route mocks populated from sanitized probe results. They demonstrate actor-scoped versus target-scoped states without exposing live identity or benefit data and are not represented as screenshots of a live authenticated production session.
- The broader workspace E2E run encountered unrelated failures outside Rewards; the focused Rewards impersonation suite did not report a failure.
- The workspace artifact check passed before push. Optional local scanners unavailable on this machine were reported as skipped by that check.

## Security and privacy checks

- The browser cannot select or override the Rewards subject.
- Missing, ambiguous, expired, or incomplete target identity fails closed without retrying as the actor.
- Authentication and authorization failures are propagated rather than represented as partial success.
- Blocked writes retain actor/target attribution in operational logs without recording tokens or full coupon values.
- Redacted upstream failures are not logged or serialized with response content.

## Test plan

- [x] Verify self-view Rewards behavior remains writable and calculations remain unchanged.
- [x] Verify impersonated profile and promotions use one resolved target.
- [x] Verify incomplete pagination and malformed source data are marked unavailable.
- [x] Verify Claim and Redeem are blocked while existing coupon Copy remains available.
- [x] Verify explicit zero points differ from unavailable points.
- [x] Verify issued coupons retain upstream eligibility restrictions.
- [x] Verify redacted failed responses are drained without exposing their body.
- [x] Run focused browser scenarios for target state, partial failure, blocked writes, and transition behavior.

Jira: [LFXV2-3232](https://linuxfoundation.atlassian.net/browse/LFXV2-3232)

[LFXV2-3232]: https://linuxfoundation.atlassian.net/browse/LFXV2-3232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/lfx-one/src/app/modules/trainings/components/rewards/rewards.component.ts
```diff
@@ -115,10 +119,21 @@ export class RewardsComponent {
   // ─── Private Initializers ──────────────────────────────────────────────────
   private initRewardsState(): Signal<RewardsState> {
     const initialState: RewardsState = { loading: true, error: null, data: null };
+    const subjectChanges$ = combineLatest([toObservable(this.userService.user), toObservable(this.userService.impersonating)]).pipe(
+      map(([user, impersonating]) => [user?.username ?? null, impersonating] as const),
+      distinctUntilChanged(([previousUsername, previousImpersonating], [username, impersonating]) => {
+        return previousUsername === username && previousImpersonating === impersonating;
+      }),
```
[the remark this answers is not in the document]
prabodhcs: Fixed in 1fbb3e4c0 — Rewards subject-change detection now observes viewerUsername, which already resolves the optional username alias and canonical LFID claim fallback. The mounted transition test covers identities where the alias is absent.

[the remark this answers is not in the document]
prabodhcs: Fixed in 1fbb3e4c0 — added a mounted component regression test that renders the first subject, changes the canonical viewer identity, confirms the prior summary is cleared, and verifies the second request renders the new points.

apps/lfx-one/src/server/helpers/gateway-fetch.helper.ts
```diff
@@ -79,18 +81,21 @@ export async function gatewayFetch<T>(req: Request, url: string, options: Gatewa
   }
 
   if (!upstream.ok) {
-    const body = (await upstream.text().catch(() => '')).slice(0, UPSTREAM_ERROR_BODY_LIMIT);
-
-    logger.warning(req, options.operation, 'Upstream returned non-OK response', {
+    const body = options.redactResponseBody
+      ? await upstream.body?.cancel().catch(() => undefined)
+      : (await upstream.text().catch(() => '')).slice(0, UPSTREAM_ERROR_BODY_LIMIT);
```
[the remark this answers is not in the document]
prabodhcs: Fixed in 1fbb3e4c0 — redacted response streams are drained to EOF without retaining content, with cancellation only as the read-failure fallback. Tests cover both successful draining and fallback cancellation.

packages/shared/src/utils/rewards.utils.ts
```diff
@@ -142,13 +145,15 @@ function getCouponSortRank(p: DecoratedCoupon): number {
 
 export function decorateCoupons(
   coupons: readonly RewardPromotion[],
-  rewardPoints: number,
```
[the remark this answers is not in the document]
prabodhcs: No change — points-unavailable status controls the badge and action state, while promotion ordering intentionally remains consistent with the existing eligibility ordering. Re-ranking based on a temporary profile-source outage would change established coupon ordering.

apps/lfx-one/src/server/services/rewards.service.ts
```diff
@@ -9,31 +9,44 @@ import {
...
 
-    const [userProfile, promotions] = await Promise.all([this.fetchUserProfile(req), this.fetchPromotions(req)]);
-    const groupedPromotions = this.groupPromotions(promotions);
-    const points = this.parsePoints(userProfile.TuxRewards);
-    const nextRewardPoints = this.calculateNextThreshold(points);
+    const [profileResult, promotionsResult] = await Promise.allSettled([this.fetchUserProfile(req, subject), this.fetchPromotions(req, subject)]);
+    const profile = this.unwrapSource(req, 'profile', profileResult);
+    const promotions = this.unwrapSource(req, 'promotions', promotionsResult);
+
+    const groupedPromotions = this.groupPromotions(promotions ?? []);
```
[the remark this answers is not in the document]
prabodhcs: Fixed in 1fbb3e4c0 — promotion mapping now runs inside the settled promotions source boundary, so malformed consumed fields mark that source unavailable instead of escaping as a 500. Added numeric-description and null-product regression cases.

apps/lfx-one/src/server/utils/rewards-subject.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+}
+
+function resolveTargetSalesforceId(response: RewardUserLookupPage | null, username: string): string {
+  const rows = response?.Data;
+  const totalSize = response?.Metadata?.TotalSize;
+  const hasOneResult = Array.isArray(rows) && rows.length === 1 && totalSize === 1;
+  const target = hasOneResult ? rows[0] : undefined;
+  const returnedUsername = target?.Username?.trim();
+  const salesforceId = target?.ID?.trim();
```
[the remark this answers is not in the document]
prabodhcs: Fixed in 2e60152ba — target lookup fields are now runtime-checked as strings before trimming, so malformed IDs or username echoes consistently return the controlled fail-closed error. Added regression cases for both fields.

packages/shared/src/utils/rewards.utils.ts
```diff
@@ -142,13 +145,15 @@ function getCouponSortRank(p: DecoratedCoupon): number {
...
+  rewardPoints: number | null,
   programStartDate: string | null,
   now: number = Date.now()
 ): DecoratedCoupon[] {
   const decorated: DecoratedCoupon[] = coupons.map((p) => {
     const hasCouponCode = p.coupon.length > 0;
-    const pointsShortfall = Math.max(0, p.redeemPoints - rewardPoints);
+    const pointsShortfall = rewardPoints === null ? null : Math.max(0, p.redeemPoints - rewardPoints);
+    const pointsUnknown = rewardPoints === null && p.redeemPoints > 0;
```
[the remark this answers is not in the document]
prabodhcs: No change — this duplicates the earlier ordering suggestion. Points-unavailable status already controls the badge and disabled action, while ordering intentionally remains stable with the established eligibility ranking rather than changing during a temporary profile-source outage.

apps/lfx-one/src/server/utils/auth-helper.ts
```diff
@@ -148,14 +148,22 @@ export function getEffectiveName(req: Request): string | null {
...
+ * identity and write guards stable if the session expiry passes later in the same request.
+ * Direct unit callers without the marker fall back to validating the session.
  * Use this to switch profile reads to the target's identity and to block profile writes (which
  * can only ever act on the real user's account).
  */
 export function isImpersonating(req: Request): boolean {
+  if (typeof req.impersonationActive === 'boolean') {
+    return req.impersonationActive;
+  }
+
```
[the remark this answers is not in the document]
prabodhcs: Fixed in 2e60152ba — the JSDoc now describes the request-frozen impersonation marker and explains that stored-token presence independently determines whether the bearer token might belong to the target.

apps/lfx-one/src/server/services/rewards.service.ts
```diff
@@ -371,3 +357,15 @@ export class RewardsService {
...
+function isNonNegativeInteger(value: unknown): value is number {
+  return typeof value === 'number' && Number.isInteger(value) && value >= 0;
+}
+
+function isValidPromotion(value: unknown): value is RewardPromotionRaw {
+  if (!value || typeof value !== 'object' || Array.isArray(value)) return false;
+
+  const promotion = value as RewardPromotionRaw;
+  const category = promotion.Category?.toLowerCase();
+  return Boolean(promotion.PromotionID?.trim()) && REWARD_CATEGORIES.some((candidate) => candidate.toLowerCase() === category);
```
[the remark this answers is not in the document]
prabodhcs: Fixed in 2e60152ba — promotion pages now validate every field used for classification, eligibility, display, and dereferencing before the source is accepted. Non-finite or negative numeric values, non-boolean eligibility/redemption values, malformed strings, products, and content types mark promotions unavailable. The focused malformed-field matrix covers these cases.

prabodhcs:
## Review resolution summary

| Result | Count | Details |
|---|---:|---|
| Fixed | 7 | Degraded promotion mapping, redacted response draining, effective-user refresh coverage and identity keys, fail-closed target lookup guards, promotion runtime validation, and corrected token-resolution rationale. |
| No change | 2 | Duplicate suggestions to reorder points-unavailable coupons; status and disabled actions already reflect unavailable points, while eligibility ordering intentionally stays stable during temporary profile-source outages. |

Fixes are in `1fbb3e4c0` and `2e60152ba`. All nine review threads are resolved.

Verification after the final fixes:
- 73 focused server tests passed.
- Formatting, lint, type-check, and full production build passed.
- Workspace artifact validation passed before push.
- DCO, CodeQL, analysis, license, title, and content-validation checks are passing; Code Quality and Cursor Bugbot are still running.

apps/lfx-one/src/server/services/rewards.service.ts
```diff
@@ -371,3 +357,46 @@ export class RewardsService {
...
+
+function isOptionalBoolean(value: unknown): boolean {
+  return value === undefined || typeof value === 'boolean';
+}
+
+function isValidPromotionProduct(value: unknown): boolean {
+  if (!value || typeof value !== 'object' || Array.isArray(value)) return false;
+
+  const product = value as Record<string, unknown>;
+  return ['ID', 'Name', 'LogoURL'].every((field) => isOptionalString(product[field]));
```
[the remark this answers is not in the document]
prabodhcs: Fixed in 9e85110eb — optional upstream promotion fields now accept null as absent while required ID/category fields, malformed non-null values, null array elements, and invalid ranges remain rejected. Added focused coverage for all nullable optionals and negative null boundaries.

apps/lfx-one/src/app/modules/trainings/components/rewards/my-coupons/my-coupons.component.html
```diff
@@ -54,19 +58,30 @@ <h4 class="text-base font-semibold text-[#1f2328] leading-tight">{{ coupon.title
...
                     severity="primary"
                     [outlined]="true"
-                    [disabled]="!coupon.eligible || coupon.pointsShortfall > 0 || coupon.redeemed"
+                    [disabled]="
+                      readOnly() ||
+                      !coupon.eligible ||
+                      (coupon.redeemPoints > 0 && coupon.pointsShortfall === null) ||
+                      (coupon.pointsShortfall !== null && coupon.pointsShortfall > 0) ||
+                      coupon.redeemed
+                    "
```
audigregorie: **Major — redeem-eligibility logic is triplicated and the copies have already drifted**

This 5-clause `[disabled]` expression is the third home for the coupon redeemability predicate:

1. Here in the template: `readOnly() || !coupon.eligible || (redeemPoints > 0 && pointsShortfall === null) || (pointsShortfall > 0) || coupon.redeemed`
2. `MyCouponsComponent.onRedeem()`: `readOnly() || (coupon.redeemPoints > 0 && this.rewardPoints() === null)`
3. `RewardsComponent.canRedeem()`: `!promotion.id || readOnly() || (redeemPoints > 0 && points() === null) || !eligible || redeemed || coupon || redeeming`

The drift is real: the **insufficient-points guard (`pointsShortfall > 0`) exists only in this template**. `canRedeem()` in the parent doesn't check it, so the "not enough points" rule is enforced by nothing but button disabledness — and this PR's own utils comment notes upstream can flag `eligible: true` while the user is still short on points. Any future path that reaches `onRedeemCoupon` without the template gate skips the shortfall rule entirely.

Suggest extracting one predicate into `packages/shared/src/utils/rewards.utils.ts`, next to `decorateCoupons` (which already derives the related status fields), e.g.:

```ts
export function isCouponRedeemable(p: RewardPromotion, rewardPoints: number | null): boolean {
  if (!p.id || !p.eligible || p.redeemed || p.coupon) return false;
  if (p.redeemPoints > 0 && rewardPoints === null) return false;
  return rewardPoints === null || p.redeemPoints - rewardPoints <= 0;
}
```

Then the template collapses to `[disabled]="readOnly() || !isCouponRedeemable(...)"` (or a precomputed field on `DecoratedCoupon`), and both `onRedeem` and `canRedeem` reuse the same source of truth instead of re-deriving overlapping subsets.

apps/lfx-one/src/server/services/rewards.service.ts
```diff
@@ -43,190 +58,169 @@ export class RewardsService {
...
+  private validatePromotionPage(response: RewardPromotionsPage | null, requestedOffset: number) {
+    const data = response?.Data;
+    const metadata = response?.Metadata;
+
+    if (
+      !metadata ||
+      !isNonNegativeInteger(metadata.Offset) ||
+      metadata.Offset !== requestedOffset ||
+      !isNonNegativeInteger(metadata.PageSize) ||
+      metadata.PageSize !== REWARD_PROMOTIONS_PAGE_SIZE ||
```
audigregorie, line 155: **Minor — strict echoed-`PageSize` equality turns an upstream page-size cap into a full promotions outage**

`metadata.PageSize !== REWARD_PROMOTIONS_PAGE_SIZE` hard-fails the entire promotions source if user-service ever caps, ignores, or changes the requested page size and echoes its own. The integrity-critical fields here are `Offset` (positioning) and `TotalSize` (completeness); the echoed page size is only used downstream for the short-page check, which could consume the echo instead of the request constant:

```ts
// in the incomplete check below:
const effectivePageSize = metadata.PageSize;
const incomplete = nextOffset > totalSize || (nextOffset < totalSize && data.length < effectivePageSize);
```

That keeps the fail-closed completeness guarantee (any gap in the collection still invalidates the source) without making availability depend on the upstream honoring `pageSize=500` verbatim. If the strict echo is deliberate, a one-line comment stating that assumption would help the next reader not mistake it for a bug.

apps/lfx-one/src/app/modules/trainings/components/rewards/available-incentives/available-incentives.component.html
```diff
@@ -59,10 +65,15 @@ <h4 class="text-base font-semibold text-[#1f2328] leading-tight">{{ incentive.ti
...
                     [outlined]="true"
-                    [disabled]="!incentive.canClaim"
+                    [disabled]="!incentive.canClaim || readOnly()"
                     [loading]="redeemingUids()[incentive.uid] || false"
+                    [tooltip]="readOnly() ? 'Rewards are read-only while impersonating.' : undefined"
+                    [ariaLabel]="readOnly() ? 'Claim unavailable: rewards are read-only while impersonating' : 'Claim coupon'"
                     (onClick)="onClaim(incentive)"
                     data-testid="available-incentive-claim-button" />
+                  @if (readOnly()) {
+                    <span class="text-xs text-[#6b7380]">Read-only while impersonating</span>
```
audigregorie: **Minor — hard-coded hex in new markup**

`.claude/rules/styling.md`: never hard-code hex values — reference the `lfxColors` scales so brand updates propagate. `#6b7380` is a mid gray, so the `gray` scale covers it:

```suggestion
                    <span class="text-xs text-gray-500">Read-only while impersonating</span>
```

(Adjust the step to whichever `gray` shade visually matches; the surrounding legacy hex usages are pre-existing, but new lines should follow the scale rule.)

apps/lfx-one/src/app/modules/trainings/components/rewards/my-coupons/my-coupons.component.html
```diff
@@ -54,19 +58,30 @@ <h4 class="text-base font-semibold text-[#1f2328] leading-tight">{{ coupon.title
...
                     [loading]="redeemingUids()[coupon.uid] || false"
+                    [tooltip]="readOnly() ? 'Rewards are read-only while impersonating.' : undefined"
+                    [ariaLabel]="readOnly() ? 'Redeem unavailable: rewards are read-only while impersonating' : 'Redeem coupon'"
                     (onClick)="onRedeem(coupon)"
                     data-testid="coupon-redeem-button" />
+                  @if (readOnly()) {
+                    <span class="text-xs text-[#6b7380]">Read-only while impersonating</span>
```
audigregorie: **Minor — hard-coded hex in new markup**

Same as the sibling component: `.claude/rules/styling.md` requires the `lfxColors` scales over raw hex.

```suggestion
                    <span class="text-xs text-gray-500">Read-only while impersonating</span>
```

apps/lfx-one/src/app/modules/trainings/components/rewards/my-coupons/my-coupons.component.html
```diff
@@ -7,7 +7,11 @@ <h3 class="text-base font-semibold text-[#1f2328] flex items-center gap-2">
     My Coupons
   </h3>
 
-  @if (decoratedCoupons().length === 0) {
+  @if (availability() === 'unavailable') {
+    <div class="rounded-xl border border-amber-200 bg-amber-50 p-6 text-center shadow-[0_1px_1px_rgba(9,30,66,.1)]" data-testid="my-coupons-unavailable">
+      <p class="text-sm text-amber-800">Coupons are temporarily unavailable.</p>
+    </div>
```
audigregorie, line 13: **Minor — the unavailable-banner block is now duplicated across components**

This amber banner is near-identical to the one in `available-incentives.component.html` (same border/bg/padding/shadow chain, differing only in testid and message), and `rewards.component.html` adds a third close variant for the profile-unavailable state. Worth extracting one small presentational banner (an `lfx-reward-unavailable-banner` with a `message` input, or a shared `ng-template`) so the amber styling and structure stay in lockstep across the three states.

apps/lfx-one/src/server/utils/rewards-subject.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+vi.mock('../helpers/api-gateway.helper', () => ({
+  getUserServiceBaseUrl: vi.fn(() => 'https://gateway.example.test/user-service/v1'),
+}));
+vi.mock('./auth-helper', () => ({
+  getEffectiveUsername,
+  isImpersonating,
+  stripAuthPrefix: (username: string) => username.replace(/^.*\|/, ''),
+  usernameMatches: (expected: string, actual: string) => expected.replace(/^.*\|/, '') === actual.replace(/^.*\|/, ''),
+}));
+
```
audigregorie: **Minor — mock hand-copies pure helpers with divergent semantics**

The mocked `stripAuthPrefix` here strips up to the **last** `|` (greedy `/^.*\|/`), while the production `stripAuthPrefix` strips at the **first** `|` (`indexOf`). Equivalent for today's single-prefix usernames, but the hand-copy can silently drift from the real implementation — exactly what this module's security assertions depend on. The repo already has a precedent for avoiding this (`member-v1-mapping.helper.spec.ts` uses `vi.importActual` for pure constants/regexes):

```suggestion
vi.mock('./auth-helper', async (importOriginal) => ({
  ...(await importOriginal<typeof import('./auth-helper')>()),
  getEffectiveUsername,
  isImpersonating,
}));
```

That mocks only the session-reading functions and exercises the real prefix/matching logic. Same pattern applies to the `usernameMatches` mock in `rewards.service.spec.ts`.

audigregorie (commented):
## Code Review Summary

Solid, security-first change. Subject resolution fails closed (exact-one username match, Salesforce ID validation, profile username-echo check), writes are blocked at both route and service layers, upstream bodies are redacted from logs and error metadata, and the per-source `availability` model correctly distinguishes "zero points" from "points unavailable" end to end. Verified locally: 116 server + 4 shared + 2 component specs pass, `check-types` and `lint:check` clean. No critical findings — 1 Major, 5 Minor threads below.

### Outside the diff

- **Query-param casing change needs an explicit confirmation.** `fetchPromotions` switched the same user-service v1 endpoint from `?Offset=…&PageSize=…` to `?offset=…&pageSize=…`. If the upstream binds query params case-sensitively, the strict echo validation in `validatePromotionPage` turns this into a fail-closed "promotions unavailable" state for **all** users, not just impersonated views. The PR's production probe presumably covered this (the strict check would have failed otherwise), but since it's an unadvertised contract change against an unchanged endpoint, please confirm the probe exercised the paginated path explicitly.

### What's done well

- The `Promise.allSettled` + per-source `availability` design — distinguishing explicit zero points from unavailable profile data through server → shared types → `decorateCoupons` → templates is exactly right.
- Freezing `req.impersonationActive` at authentication time removes the mid-request expiry race while keeping a live-session fallback for direct callers.
- Test coverage is above the bar: fail-closed `it.each` matrices for malformed upstream data, and route-level tests that boot a real Express app to prove the write boundary.

apps/lfx-one/src/app/modules/trainings/components/rewards/my-coupons/my-coupons.component.html
prabodhcs: Fixed in commit 5953beb32 — the coupon read-only label now uses text-gray-500 instead of a raw hex value.

apps/lfx-one/src/app/modules/trainings/components/rewards/available-incentives/available-incentives.component.html
prabodhcs: Fixed in commit 5953beb32 — the new read-only label now uses text-gray-500 from the configured neutral palette.

apps/lfx-one/src/app/modules/trainings/components/rewards/my-coupons/my-coupons.component.html
prabodhcs: Done — coupon redeemability now has one shared predicate used by the decorated template state, child emit guard, and parent confirmation guard. Insufficient points can no longer reach redemption outside the template gate, with focused utility and component coverage in commit 5953beb32.

apps/lfx-one/src/server/utils/rewards-subject.spec.ts
prabodhcs: Fixed in commit 5953beb32 — the subject spec now retains the real auth helpers and Salesforce ID pattern instead of hand-copying their semantics. While validating this, both subject echoes were tightened to exact matching after the single intended prefix normalization, with compound-username positive and fail-closed regression cases.

apps/lfx-one/src/app/modules/trainings/components/rewards/my-coupons/my-coupons.component.html
prabodhcs: Thanks for flagging this. I kept the unavailable states local because only the coupon and incentive blocks are exact duplicates; the profile state has a different icon, heading, message structure, radius, padding, and layout. Extracting a component for the two matching one-line states would add more structure than it removes, so no code change was made.

apps/lfx-one/src/server/services/rewards.service.ts
prabodhcs: Kept strict equality deliberately and documented the assumption in commit 5953beb32. User-service defines lowercase pageSize and echoes the requested value, so the exact match verifies that query contract before the response is trusted as complete; any contract drift remains fail-closed.

prabodhcs:
## Review Feedback Addressed

Commit: 5953beb3215ad1ecdb4c07d5434b5e5464335db7

### Changes Made
- **Coupon redemption**: centralized redeemability and enforced point sufficiency in the template, child emit guard, and parent confirmation guard (per @audigregorie).
- **Identity validation**: retained production auth and Salesforce validation semantics in tests, then tightened already-normalized subject echoes to exact fail-closed matching (per @audigregorie).
- **Pagination validation**: documented why the echoed page size must exactly match the requested value before completeness is trusted (per @audigregorie).
- **Styling**: replaced both new raw gray values with the configured `gray` palette (per @audigregorie).

### No Change Needed
- **Unavailable banners**: kept the two simple section states local because the profile state is structurally different; extracting a component for only two exact duplicates would add more structure than it removes (flagged by @audigregorie).

### Pagination Query Confirmation
- Lowercase `offset` and `pageSize` match the upstream user-service OpenAPI contract. The earlier production probe preceded that casing correction, so I am not claiming it exercised this exact request; strict echoed-page-size validation will fail closed if the deployed contract differs.

### Verification
- 83 focused shared, Angular, and server tests passed.
- Formatting, lint, type checks, build, workspace artifact validation, and two security review rounds passed.

### Threads Resolved
6 of 6 remaining review threads were addressed in this iteration.

audigregorie (approved):

PR #1900, opened 2026-08-27
fix(campaigns): load the demand-gen capability on the create path

mrautela365:
Follows up the two non-blocking findings @dealako left on #1885, which merged before they were addressed. Both are **latent on `main` today** — with `_DEMAND_GEN` off and `_CREATE` on, `canCreateDemandGen` is `false` everywhere, so the control is correctly hidden either way. They bite when `_DEMAND_GEN` is enabled.

## The capability never reached the create path

`briefCampaignsDemandGenEnabled` is written only by `loadBriefCampaigns`, whose callers are all Optimize entry, retry, or a foundation switch *while already on Optimize*. `onProceedToImplementation` sets the tab **directly** rather than through `selectTab`, so the signal stayed `null` and the Implementation tab withheld Demand Gen even on a deployment that supports it.

**Two triggers, because one is not enough.** Tab entry covers a return visit, but on a genuine first create there is no brief id yet — `persistBrief` has not resolved — so the entry-time read guards itself out. The persist success arm passes its own id, which is the moment the read becomes possible at all.

Kept separate from `loadBriefCampaigns` rather than reusing it: that one is an Optimize concern that clears rows and re-fetches for staleness, and writing its state from the Implementation tab would render a list nobody asked for.

A failed read leaves the capability `null`, never `false` — the draft restore clears a saved selection on an explicit `false`, and a failure has established nothing.

## The interface doc contradicted the implementation

The field doc described `demandGenEnabled` as mirroring the raw `_DEMAND_GEN` flag. It is actually produced by `canCreateDemandGen`, which returns `true` across the whole staged CREATE-off rollout because the legacy creator makes Demand Gen regardless. A maintainer reading only that doc could "simplify" the predicate back to the raw flag and hide the working legacy option for the entire rollout — the exact regression the function doc warns against.

## Testing

Tested **through the parent**, not by setting the tab's input. That distinction is the point: the input always worked, nothing populated it, and a hand-set input passes against exactly this bug — which is how the original version shipped broken past 5 tests and 3 mutations.

- Mutation-verified: removing the persist-arm trigger fails 1 test; setting `false` on a failed read fails 2.
- 1290 app + 2095 server tests pass.

## Still open from #1885

@copilot raised that `canCreateDemandGen` is **pod-local, not deployment-wide**: during a CREATE rollout an old pod can answer `/list` with `demandGenEnabled: true` while `/create` lands on a new pod that refuses it. That is real and I have not addressed it here — no pod can know the cluster's state, and the conservative answer (`false`) reintroduces the regression above by hiding the working legacy option. It needs a shared source of rollout state rather than a per-pod inference, which is a larger change than this fix.

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
@@ -1221,6 +1221,85 @@ describe('CampaignsComponent brief persistence', () => {
         ).toBeNull();
       });
 
+      /**
+       * The capability read must survive an Optimize visit landing on top of it.
+       *
+       * It originally shared `briefCampaignsGeneration` with `loadBriefCampaigns`, which looked
+       * right — same endpoint, same staleness question — but that counter is INCREMENTED by
+       * `loadBriefCampaigns` itself. So entering Optimize while the capability request was in
+       * flight bumped it and the response was dropped as stale, leaving the capability `null`
+       * and the control hidden: exactly the defect this method exists to fix. Guarding on the
+       * foundation instead is what makes an unrelated read harmless.
+       */
+      it('keeps a capability response that an Optimize load raced', async () => {
+        const capability = new Subject<CampaignListResult>();
+        const list = vi.spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns').mockReturnValue(capability.asObservable());
+
+        await withSavedBrief();
+
+        // Optimize entry lands on top of the still-open capability request.
+        list.mockReturnValue(of({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: false }));
+        load();
+        await fixture.whenStable();
+
+        // ...and only now does the capability answer arrive.
+        capability.next({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: true });
+        capability.complete();
+        await fixture.whenStable();
+
+        expect(
+          (fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()
+        ).toBe(true);
+      });
+
+      /**
+       * The RESTORE path, which knows its brief id and still missed the capability.
+       *
+       * `onProceedToImplementation` originally asked for the capability before the
+       * `alreadyPersisted` branch wrote the restored id onto `briefPersistence`, so the read saw
+       * the PREVIOUS id and guarded itself out. Unlike a first create there is no persist to
+       * follow, so nothing retried and the capability stayed unknown for the whole session.
+       */
+      it('loads the capability for a brief restored from campaign-service', async () => {
+        vi.spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns').mockReturnValue(
+          of({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: true })
+        );
+
+        restore(brief, 'brief-9', true);
+        await fixture.whenStable();
+
+        expect(
+          (fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()
+        ).toBe(true);
+      });
+
+      /**
+       * A read that starts failing must not leave the previous answer standing.
+       *
+       * The error arm wrote nothing at first, which reads as "left null" only on the FIRST read.
+       * Once a successful read had set `true`, a later failure kept offering the control on the
+       * strength of a read that no longer succeeds. Asserts the value, not merely that the arm ran.
+       */
+      it('clears a previously known capability when a later read fails', async () => {
+        const list = vi
+          .spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns')
+          .mockReturnValue(of({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: true }));
+
+        await withSavedBrief();
+        await fixture.whenStable();
+        expect(
+          (fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()
+        ).toBe(true);
+
+        list.mockReturnValue(throwError(() => new Error('query service down')));
+        (fixture.componentInstance as unknown as { loadCreateCapabilities(): void }).loadCreateCapabilities();
+        await fixture.whenStable();
+
+        expect(
+          (fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()
+        ).toBeNull();
+      });
+
       it('clears the previous brief campaigns when the foundation changes', async () => {
         const list = vi
           .spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns')
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
@@ -1278,9 +1278,9 @@ describe('CampaignsComponent brief persistence', () => {
        * on every entry — so an ordinary Optimize click discarded an in-flight capability response
        * and left the control hidden.
        *
-       * Both readers hit the same endpoint, so they cannot disagree about the deployment; the
-       * mock answers `true` on both. What is asserted is that the answer SURVIVES the overlap,
-       * whichever of the two happens to deliver it.
+       * Both mocked responses return `true`. What is asserted is that the answer SURVIVES the
+       * overlap — whichever of the two requests delivers it first, a `true` lands on the signal
+       * rather than being overwritten to `null` by the later arrival.
        */
       it('keeps a capability answer when an Optimize load races it', async () => {
         const capability = new Subject<CampaignListResult>();
@@ -1403,8 +1403,8 @@ describe('CampaignsComponent brief persistence', () => {
        * A foundation switch clears the capability but dispatches no capability read of its own,
        * so it never bumps `createCapabilitiesGeneration`. An in-flight read from the previous
        * foundation therefore still looks current BY GENERATION, and only the slug check stops it
-       * writing foundation A's answer onto foundation B — a capability claim about a deployment
-       * the user is no longer looking at.
+       * writing foundation A's answer onto foundation B — a stale answer for a slug the user
+       * is no longer viewing.
        */
       it('drops a capability answer that arrives after the foundation changed', async () => {
         const pending = new Subject<CampaignListResult>();
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
@@ -1230,8 +1230,9 @@ describe('CampaignsComponent brief persistence', () => {
       });
 
       /**
-       * because the tab's draft restore clears a saved Demand Gen selection on an explicit `false`
-       * and a failed read has established nothing.
+       * A failed read leaves the capability `null`, never `false` — because the tab's draft restore
+       * clears a saved Demand Gen selection on an explicit `false`, and a failed read has
+       * established nothing.
        */
       it('leaves the capability unknown when the capability read fails', async () => {
         vi.spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns').mockReturnValue(throwError(() => new Error('query service down')));
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
@@ -1243,29 +1243,56 @@ describe('CampaignsComponent brief persistence', () => {
       });
 
       /**
-       * The capability read must survive an Optimize visit landing on top of it.
+       * An Optimize visit landing on top of a capability read must not lose the answer.
        *
-       * It originally shared `briefCampaignsGeneration` with `loadBriefCampaigns`, which looked
-       * right — same endpoint, same staleness question — but that counter is INCREMENTED by
-       * `loadBriefCampaigns` itself. So entering Optimize while the capability request was in
-       * flight bumped it and the response was dropped as stale, leaving the capability `null`
-       * and the control hidden: exactly the defect this method exists to fix. Guarding on the
-       * foundation instead is what makes an unrelated read harmless.
+       * The guard shared Optimize's list counter at first, which `loadBriefCampaigns` increments
+       * on every entry — so an ordinary Optimize click discarded an in-flight capability response
+       * and left the control hidden.
+       *
+       * Both readers hit the same endpoint, so they cannot disagree about the deployment; the
+       * mock answers `true` on both. What is asserted is that the answer SURVIVES the overlap,
+       * whichever of the two happens to deliver it.
        */
-      it('keeps a capability response that an Optimize load raced', async () => {
+      it('keeps a capability answer when an Optimize load races it', async () => {
         const capability = new Subject<CampaignListResult>();
         const list = vi.spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns').mockReturnValue(capability.asObservable());
 
         await withSavedBrief();
 
         // Optimize entry lands on top of the still-open capability request.
-        list.mockReturnValue(of({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: false }));
+        list.mockReturnValue(of({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: true }));
         load();
         await fixture.whenStable();
 
-        // ...and only now does the capability answer arrive.
+        // ...and the original request answers only now.
         capability.next({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: true });
         capability.complete();
+        await fixture.whenStable();
+
+        expect((fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()).toBe(true);
+      });
+
+      /**
+       * The reverse race, and the one a single-reader token cannot close: `loadBriefCampaigns`
+       * writes this signal too. An older list read failing AFTER a newer capability read
+       * succeeded must not reset a live answer to `null` — a failure has established nothing.
+       */
+      it('does not let an older failing list read wipe a newer capability answer', async () => {
+        const staleList = new Subject<CampaignListResult>();
+        const list = vi.spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns').mockReturnValue(staleList.asObservable());
+
+        await withSavedBrief();
+        load();
+        await fixture.whenStable();
+
+        // A newer capability read answers while that list request is still open.
+        list.mockReturnValue(of({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: true }));
+        (fixture.componentInstance as unknown as { loadCreateCapabilities(): void }).loadCreateCapabilities();
+        await fixture.whenStable();
+        expect((fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()).toBe(true);
+
+        // The superseded list read now fails. It must not touch the newer answer.
+        staleList.error(new Error('query service down'));
         await fixture.whenStable();
 
         expect((fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()).toBe(true);
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
@@ -1300,6 +1300,38 @@ describe('CampaignsComponent brief persistence', () => {
         ).toBeNull();
       });
 
+      /**
+       * Out-of-order completion WITHIN one foundation — the case a foundation-only guard misses.
+       *
+       * The first version of this guard shared Optimize's counter and dropped valid responses.
+       * The second guarded on the foundation alone, which drops nothing valid but imposes no
+       * ORDER: two Implementation entries in the same foundation can complete in either order,
+       * and an older request failing after a newer one succeeded would wipe a live `true` back
+       * to `null`, hiding Demand Gen on a deployment that had just confirmed it.
+       */
+      it('ignores an older capability read that fails after a newer one succeeded', async () => {
+        const older = new Subject<CampaignListResult>();
+        const list = vi.spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns').mockReturnValue(older.asObservable());
+
+        await withSavedBrief();
+
+        // A second Implementation entry supersedes the first, and answers first.
+        list.mockReturnValue(of({ campaigns: [], possiblyStale: false, statusToggleEnabled: false, demandGenEnabled: true }));
+        (fixture.componentInstance as unknown as { loadCreateCapabilities(): void }).loadCreateCapabilities();
+        await fixture.whenStable();
+        expect(
+          (fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()
+        ).toBe(true);
+
+        // The superseded request now fails. It must not touch the newer answer.
+        older.error(new Error('query service down'));
+        await fixture.whenStable();
+
+        expect(
+          (fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()
+        ).toBe(true);
+      });
+
       it('clears the previous brief campaigns when the foundation changes', async () => {
         const list = vi
           .spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns')
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
@@ -1252,9 +1252,7 @@ describe('CampaignsComponent brief persistence', () => {
        * persist-success path still fires its own read — but this case stays broken.
        */
       it('retries the capability read when the user re-enters Implementation via the tab bar', async () => {
-        const list = vi
-          .spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns')
-          .mockReturnValue(throwError(() => new Error('query service down')));
+        const list = vi.spyOn(TestBed.inject(CampaignService), 'listBriefCampaigns').mockReturnValue(throwError(() => new Error('query service down')));
 
         // First create: persist succeeds, capability read fails.
         await withSavedBrief();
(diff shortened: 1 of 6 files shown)


mrautela365:
Ran the pre-PR reviewer simulation (four personas, in parallel) against `bf2ef17e4` before the bots got here. It found **three real defects in my own fix**, each proved by mutation rather than argued. Fixed in `bdfc28584`.

**1. The stale-guard dropped valid responses.** `loadCreateCapabilitiesFor` shared `briefCampaignsGeneration` with `loadBriefCampaigns` — same endpoint, same staleness question, so it looked right. But that counter is *incremented* by `loadBriefCampaigns` on every Optimize entry. Opening Implementation then clicking Optimize bumped it while the capability read was in flight, and the response was discarded as stale though nothing about it was.

Worse than a dropped answer: Optimize's own read then wrote `demandGenEnabled: false`, and `applyDraft` treats an explicit `false` as grounds to clear a saved Demand Gen selection. A counter only guards readers that *own* it. The foundation is what actually invalidates a deployment capability, so it guards on that now.

**2. The restore path never loaded the capability at all.** `onProceedToImplementation` asked for it *before* the `alreadyPersisted` branch writes the restored brief id onto `briefPersistence` — so the read saw the previous id and guarded itself out. Unlike a first create there is no persist to follow, so nothing retried and the capability stayed unknown for the whole session. Moved below the branch.

**3. The error arm left a stale `true`.** `error: () => undefined` writes nothing, which reads as "left `null`" only on the *first* read. Once a successful read had set `true`, a later failure kept offering the control on the strength of a read that no longer succeeds. Now clears to `null` under the same foundation guard.

Also corrected the signal's own TSDoc, which still claimed `loadBriefCampaigns` was "the only writer" — the enumerating-comment drift this PR itself introduces.

**Mutation-verified against `bdfc28584`:**

| Mutation | Result |
|---|---|
| drop the restore-path call | 1 test fails |
| drop the persist-arm call | 3 tests fail |
| error arm leaves the stale value | 1 test fails |
| revert to the shared generation counter | 1 test fails |

One correction to my earlier claim on this PR: I said "setting `false` on read failure fails 2" — it fails 1. The simulation caught that too, and it was wrong.

Three new tests, all driven **through the parent**: the restore path, the Optimize-race (holds the capability request open with a `Subject` while an Optimize load lands on top), and the stale-value clear. 1293 app + 2108 server tests pass.

mrautela365:
@copilot both already fixed in `bdfc28584` — your review landed on `bf2ef17e4`, and the pre-PR reviewer simulation had independently found the same two defects. Good corroboration on both; quoting the current head so you don't have to re-derive.

**Restore-path ordering** — the call now sits *after* `briefPersistence.set(...)` in the `alreadyPersisted` branch, exactly as you describe:

```ts
      this.briefPersistence.set({ status: 'off', briefId: …, approved: restoredApproved });
      // AFTER the branch, not before it. […] the restored brief's id is written by the branch
      // above, and asking first read the PREVIOUS id and guarded out.
      this.loadCreateCapabilities();
      return;
```

The persist-success trigger is retained for first creates, and the direct restore handoff is now covered in the parent spec (`loads the capability for a brief restored from campaign-service`). Mutation-verified: dropping that call fails 1 test.

**Stale declaration doc** — you're right that it documented the bug as intentional. `grep -c "the only writer is"` now returns 0; it reads:

> Two readers write it: `loadBriefCampaigns` on Optimize entry, and `loadCreateCapabilities` on the Implementation entry paths — the create path needs the answer BEFORE any campaign exists, so it cannot wait for the first. `null` therefore means "unanswered or failed", not "never asked".

The simulation also found a third defect neither of us had flagged: the guard shared `briefCampaignsGeneration` with `loadBriefCampaigns`, which *increments* it on every Optimize entry — so an Optimize visit during an in-flight capability read discarded a valid response, and Optimize's own `false` then cleared the user's saved selection via `applyDraft`. Now guards on the foundation. Details in the comment above.

mrautela365:
All three fixed in `307126c15` … `92876c0b5`. @copilot and @cursor independently found the same ordering defect, which is strong signal and was right — **my previous fix removed ordering entirely.**

**The guard has now been wrong in two opposite directions, and both were reachable:**

| Version | Failure |
|---|---|
| shared `briefCampaignsGeneration` | `loadBriefCampaigns` increments it, so an ordinary Optimize visit discarded a valid in-flight capability response |
| foundation check alone | drops nothing valid, but imposes no **order** — two Implementation entries in one foundation can complete either way round |

Copilot's description of the second is exactly right: an older request failing after a newer one succeeded wiped a live `true` back to `null`, hiding Demand Gen on a deployment that had just confirmed it. Cursor's framing (`Stale list error clears capability`) is the same defect from the error arm's side.

Fixed with a **dedicated** counter, `createCapabilitiesGeneration` — bumped on every dispatch here, untouched by Optimize, and made stale by the foundation-switch effect for free because that dispatches nothing here.

**The slug check stays alongside it, and that is not belt-and-braces.** I mutated it out and all 1294 tests still passed, so I checked whether it was reachable rather than deleting it: a foundation switch clears the capability but dispatches no capability read, so it never bumps the counter — an in-flight read from the previous foundation still looks current *by generation*, and only the slug check stops it writing foundation A's answer under foundation B. That case now has its own test.

| Mutation | Result |
|---|---|
| drop the generation half | 1 test fails (`ignores an older capability read that fails after a newer one succeeded`) |
| drop the slug half | 1 test fails (`drops a capability answer that arrives after the foundation changed`) |

**Duplicated paragraph** — real, and left by a revert-and-reapply cycle. Removed; one copy stands.

Also fixed a `member-ordering` lint failure the loaders introduced (private methods above public ones); they now sit beside `loadBriefCampaigns`, the sibling they are documented against. `yarn lint:check` clean, 1295 app + 2108 server tests pass.

mrautela365:
**Note on `67590b1ca`:** Copilot's SWE agent pushed a commit to this branch on its own. It contained two fixes — one correct, one that cannot work. I kept the first and reverted the second in `f26de2203` / `58fa385b8`.

**KEPT — gating `campaignTypes` and `canSubmit` on `demandGenAvailable()`.** This is right, and better than what I'd have written. `applyDraft` runs once at mount under `untracked`, so a tab mounting while the capability is `null` restores `includeDemandGen: true`; the checkbox then hides but the form value stays true. Gating the builder closes it at the source without touching `applyDraft`'s `untracked` semantics — which exist to avoid an NG0103 infinite-CD loop, so making it reactive was the risky path.

**REVERTED — dispatching the capability read with an empty brief id.** The commit's premise was that the service returns `demandGenEnabled` for a blank brief id. That is true of the *service*, and irrelevant: `campaign.controller.ts:871` 400s on a blank `brief_id` before the service is called, and the comment above that service branch says so explicitly ("the HTTP path never reaches this"). @cursor flagged the same thing as High Severity, and @copilot's own later comment on `:1631` contradicts its agent's commit.

The change would have spent a request per Implementation entry to receive a 400, landed in the error arm, and set the capability to `null` — control hidden either way, plus a spurious 400.

**Why its test didn't catch that:** the spec mocked `CampaignService.listBriefCampaigns`, so it never crossed the controller that rejects the request. It passed against a fix that does nothing in production.

What I changed:
- Pinned the empty-string refusal in `campaign.controller.spec.ts`, beside the whitespace case already there.
- Replaced the agent's spec with one asserting what actually happens: no request is sent, capability stays `null`.
- Added a test for the kept gate — deleting the `canSubmit` guard previously left the suite green.

**The disabled-persist gap is still open**, and I have not papered over it. It needs a capability read that does not require a brief id — a dedicated endpoint. I'd rather add that deliberately than patch a fourth path; flagged for @mrautela365 to decide, since this PR is a follow-up to non-blocking findings and nothing here is user-visible while `_DEMAND_GEN` is off.

One disclosure on test strength: the request-builder guard is not independently pinned — deleting it alone leaves the suite green, because `canSubmit` returns false first. That's recorded in the test comment rather than folded into a coverage claim.

1298 app + 2109 server tests pass; lint clean.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1247,6 +1275,61 @@ export class CampaignsComponent {
...
+  /**
+   * The same read against an explicitly supplied brief id.
+   *
+   * Needed because the first-create path has no id on the signal yet: the Implementation tab is
+   * opened before `persistBrief` resolves, so reading `briefPersistence()` at entry finds `null`
+   * and the entry-time call guards itself out. The persist success arm passes its own id here.
+   */
+  private loadCreateCapabilitiesFor(briefId: string | null): void {
+    const projectSlug = this.activeFoundationSlug();
+    if (projectSlug === '' || !briefId) return;
```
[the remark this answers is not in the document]
mrautela365: Confirmed and still open — I am not closing this one.

The SWE agent tried to fix it by dispatching with an empty brief id; that cannot work (`campaign.controller.ts:871` 400s on a blank `brief_id` first), so I reverted it and pinned the refusal in `campaign.controller.spec.ts`. The current spec asserts what actually happens on this path: no request is sent, capability stays `null`, control hidden.

So the gap you describe is real and unfixed. It needs a capability read that does not require a brief id at all — a dedicated endpoint — rather than a fourth patch to the brief-persistence flow. Raised with @mrautela365; nothing here is user-visible while `_DEMAND_GEN` is off, so it is not blocking.

mrautela365:
Two more findings from @copilot's **suppressed** block — both against my own code, both correct. Fixed in `949a3d2e9`. (Worth noting these never appeared as threads, so `unresolved` never counted them.)

**1. A cross-writer race.** `loadBriefCampaigns` writes the capability signal too, and my token ordered only `loadCreateCapabilitiesFor`. So the reverse case stayed open exactly as described: an older list read failing *after* a newer capability read succeeded still passed its own list generation and reset a live answer to `null`.

**Sharing one token by dispatch order then broke the other direction** — and that was informative. The two readers race *both* ways, so "last dispatched wins" drops a good answer in whichever direction happens to lose. That is precisely how the first two versions of this guard each fixed one case and broke the other.

Both cases want the same invariant, and it is not about dispatch:

> A **success** is authoritative from the moment it lands, and nothing older may overwrite it — least of all a failure, which has established nothing.

So the token is stamped when an answer is **written**, and every writer defers if another answer landed since it dispatched. Ordering by answer rather than by dispatch satisfies both directions without a tiebreak.

**2. The counter's doc credited it with foundation-switch invalidation.** Right, and it does not: a switch clears the capability but dispatches nothing, so it never bumps the token — the separate slug check is what rejects the previous foundation's response. Now stated as two mechanisms, deliberately apart, so a future edit does not delete the slug guard on the strength of the counter.

**One of my own tests was wrong and I rewrote it rather than bending the code to it.** `keeps a capability response that an Optimize load raced` had the two readers returning *contradictory* answers (`false` from Optimize, `true` from the capability read). They call the same endpoint and cannot disagree about a deployment fact, so the fixture described an impossible state and its expectation encoded the wrong rule. Rewritten to the realistic case, plus a new test for the reverse race you described.

| Mutation | Result |
|---|---|
| list arms stop honouring the shared token | 1 test fails |
| reader stops stamping on write (back to dispatch order) | 2 tests fail |
| slug half removed | 1 test fails |

1299 app + 2109 server tests pass; lint clean.

mrautela365:
Three more fixed in `23a92421d` … `c30d9b72a`. All three were mine.

**1. A failure could suppress a success** — and this one broke the invariant I had written two lines above it. Stamping the shared token on the ERROR arms looked symmetrical; it meant two reads dispatched together, the first failing, advanced the token and made the second's valid answer fail its own write check. My doc said *"a failure has established nothing"* while the code let a failure win. Only the two success arms stamp now; a failure still clears under its own write check (it must not overwrite a **newer** answer either), it just no longer takes the ordering with it.

**2. The doc claimed both readers compare the slug.** Only `loadCreateCapabilitiesFor` does; `loadBriefCampaigns` relies on the foundation-switch handler bumping `briefCampaignsGeneration`, which its own `isCurrent` checks. Both are load-bearing, and they are now stated separately so neither gets deleted on the strength of the other.

**3. The campaign-name preview was not gated.** With a hidden `includeDemandGen: true` it rendered `Multi` while the request carried only `search` — naming a campaign that will not be created, in the name operators use to find it in the ad platform later. That was the third and last site reading that form value; the submit gate and request builder were already gated.

| Mutation | Result |
|---|---|
| preview reads the raw form value | 1 test fails |
| error arm stamps the token again | 1 test fails |

Also fixed a `format:check` failure CI caught and my local gate did not — `lint:check` does not cover formatting, so I now run both.

1301 app + 2109 server tests pass; lint and format clean.

The disabled-persist gap on `:1325` stays open deliberately — it needs a capability read that does not require a brief id, and I would rather add that endpoint than patch a fifth path through the brief-persistence flow. With @mrautela365.

mrautela365:
**Full-diff audit** — rather than wait for a sixth round of individual findings, I enumerated every reader and writer of the capability. Four rounds in a row here ended with the fix introducing the next finding, which is the point at which patching stops paying.

**6 writers · 5 readers.** All five readers are now consistent: `demandGenAvailable()` (`=== true`) gates the control, `canSubmit`, the request builder, and the name preview; `=== false` gates the draft restore. No reader does a truthy check that `null` could slip through.

One thing the audit surfaced that no reviewer has raised, and I am **not** changing it unilaterally because it is a behaviour decision:

**An explicit `false` permanently destroys the user's Demand Gen selection.** `applyDraft` clears the form value (`:1636`), `emitDraft` writes that `false` back to the parent (`:1763`), and the original choice is gone. If `_DEMAND_GEN` is enabled later, the user sees an unchecked box — the same destruction the comment two lines above it warns about for `null`, just triggered by `false`.

It was necessary when clearing was the only thing stopping a hidden `true` from submitting. It no longer is: the submit path is now gated in three places plus the server's own refusal, so the clear's *only* remaining effect is losing the user's choice.

Two defensible positions:
- **Keep it** — the capability really is off, and a draft that matches the form is simpler to reason about.
- **Drop it** — preserve the choice like the `null` case does; the three gates already prevent the bad submit, and a capability can be switched on later.

Flagged for @mrautela365 rather than decided here.

Also still open and deliberately unresolved: the disabled-persist gap on `:1327`. Both it and the above trace to the same root — the capability is bound to the brief-persistence flow, and a *deployment* capability does not depend on having a brief. Four paths have been patched; a capability read that needs no brief id closes the class instead.

mrautela365:
@cursor correct, and this one is worth stating plainly: **my own preview gate made the name more stale, not less.** Fixed in `714511134`.

`campaignName` is `toSignal` over `valueChanges`, so a signal read inside its `map` is not a reactive dependency — it re-evaluates only when the form changes. Gating the preview on `demandGenAvailable()` *there* froze it at whatever the capability was at mount. A tab mounting before the capability resolves previews `Search` for ever, while submit sends `demand-gen`. The gate I added two commits ago to make the preview honest is what introduced that.

Rewritten as a `computed` over the form stream, so the capability is a genuine dependency and the name recomputes when either it or the form changes.

**Reproduced before fixing** — the new test fails against `c30d9b72a` (`expected '| Search |' to contain '| Multi |'`) and passes after. Mutation-verified by reading the capability under `untracked` again: the test fails.

1302 app + 2109 server tests pass; lint and format clean.

For the record, this is the fifth consecutive round where a fix produced the next finding, and all five trace to the same root: the capability is threaded through the brief-persistence flow and the form, so every path that reads it needs its own guard. The gates are now consistent across all five readers, but that is five patches rather than one endpoint. That decision is with @mrautela365, along with the still-open disabled-persist gap on `:1327`.

mrautela365:
Two **suppressed** findings from @copilot, both fixed — `1c17c6b55` and `e8c66229b`. Neither appeared as a thread, so `unresolved` never counted them.

**1. The wire guard had no test, and my stated reason was wrong.** I told this PR that `canSubmit` returned false first, so the builder's `campaignTypes.push('demand-gen')` was defence in depth. It isn't. `canSubmit` only rejects Google with **neither** type selected — with Search also ticked it passes, and the builder's guard is then the only thing keeping a hidden `demand-gen` off the request. Deleting that line left the suite green.

Writing the test taught me something I had also got wrong. My first attempt set the capability `false` before mount, and it **passed against the guard's deletion** — because `applyDraft` clears the form value on an explicit `false`, so the builder had nothing left to guard. It was testing the clear, not the guard.

The only sequence that puts a hidden `true` on the form is **`null` at mount** (draft preserved) **then `false`** (control hides, value unchanged). That is what the test now drives, and it asserts the value *is* `true` before submitting so it cannot pass the earlier way. Mutation-verified: deleting the guard now fails it.

**2. The interface doc contradicted itself.** *"Defaults to `false` on every error and pre-load path"* — wrong twice: an error produces no `CampaignListResult` at all, and the client models unanswered/failed as `null`, deliberately, because a false negative there clears a user's saved selection rather than withholding a control. Rewritten to say `false` is a server **statement** that the capability is off, not a safe default for unknown.

That second one matters more than a doc nit: it was actively inviting a maintainer to reintroduce the destructive fallback on the strength of the type.

1303 app + 2109 server tests pass; lint and format clean.

mrautela365:
The return-entry gap is covered — with a note on how it got there.

@copilot flagged it in a thread; I verified by mutation (deleting the `tab === 'implementation'` branch left all 1303 green) and wrote a test. While I was doing that, **Copilot's SWE agent pushed a test for the same gap** (`0c4e0b4cd`), and we collided.

**I kept the agent's and dropped mine.** Its version is better: it frames the branch as a *retry after a failed capability read*, which is the actual user value, and it uses the real `CampaignTab` / `CampaignDeliveryType` types where mine cast to `string`. Two tests on one branch would have been noise.

One thing to flag, since it is the second time: the agent pushed without running prettier, so `format:check` failed exactly as its previous commit did. Fixed in `6fe594814` — reformatting only, no behaviour change. Worth knowing if that agent is left enabled on this repo.

Also corrected a stale coverage note of mine (`d9c173ffd`). It said the request-builder guard was not independently pinned and that deleting it kept the suite green. True when written; false one commit later, once the wire-level test landed. Left alone it understated coverage and — worse — told a maintainer the guard was redundant defence in depth, when `canSubmit` genuinely does not cover Search plus a stale hidden Demand Gen.

1304 app + 2109 server tests pass; lint and format clean.

Still open and deliberately unresolved: the disabled-persist gap on `:1327`.

mrautela365:
Fixed in `e494cb437`, and the finding was a good catch on a claim of mine that was **self-contradicting across PRs**.

The comment said the two readers "cannot disagree about the deployment; they hit the same endpoint." That is false during exactly the state this work is about — a rolling deployment lands identical `/list` calls on pods with different flag values. I documented that pod-local hazard on #1885 myself, then wrote the opposite here.

The test never depended on agreement: both mocks answer `true` and it pins **ordering** — that an answer survives the overlap.

Third collision with Copilot's SWE agent, which pushed `12862baea` for the same comment while I was fixing it. Merged rather than picked this time: its wording of the mechanism is clearer, so that is what stands; but its diff **missed the stale `createCapabilitiesGeneration` reference** at the sibling comment — a symbol renamed to `capabilityGeneration` two commits ago that now names nothing. My version fixes both, so the merge keeps its prose and my symbol fix, plus a line recording *why* the old claim was wrong so it does not get rewritten.

1304 app + 2109 server tests pass; lint and format clean.

Still open and deliberately unresolved: the disabled-persist gap on `:1327`.

dealako:
Hi @mrautela365 — thanks for an unusually rigorous PR. The fix itself is small (a capability signal that now loads on the create/restore paths, plus wire-level gates so a stale hidden `demandGenEnabled: true` can't reach submit), but the commit trail and your own inline write-ups show a genuinely careful iteration: each race-guard revision is mutation-verified, and you documented where a fix introduced the next finding rather than papering over it. The answer-ordering invariant (stamp `capabilityGeneration` on success only, never on failure, guard on slug + generation) is internally consistent and well covered.

I ran a three-dimension independent review (security/privacy, correctness/performance/tests, style/docs). Security and privacy came back clean: the client-side `demandGenAvailable()` gate is defense-in-depth, not the sole enforcement — the server's `canCreateDemandGen()` still refuses `demand-gen` on the create path when the flag is off. No PII sinks; controller logging carries only slugs/ids/counts and fixtures use synthetic `example.com` data.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 1 — capability is refetched (full campaign-list query) on every Implementation-tab entry with no "already answered" guard, unlike the sibling `loadEmailTemplatesIfNeverAnswered` path.
- ⚪ Nit: 2 — (1) a test docblock begins mid-sentence with lowercase "because"; (2) the success/success race tests use identical `true`/`true` payloads, so they only cover answer-ordering indirectly (you already documented this as an intentional scope choice).
- ❔ Question: 0

### Bot reconciliation
- The one **unresolved** Copilot thread — the disabled-persist path can't load the capability because `campaign.controller.ts` 400s on a blank `brief_id` — is real and you've correctly left it open. I agree it is **not blocking**: nothing is user-visible while `_DEMAND_GEN` is off, and the correct fix is a dedicated capability endpoint, not a fourth patch to the brief-persistence flow. Worth a follow-up ticket.
- Cursor's earlier findings (preview staleness, ordering) are addressed in later commits. Nothing new from the bots is outstanding.

### Decision
✅ **Approved with minor comments.** The one minor and two nits are non-blocking; address at your discretion. The disabled-persist gap should be tracked as a follow-up.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -910,6 +944,9 @@ export class CampaignsComponent {
       // since the last look — must not render with a stale status the user then acts on.
       this.loadBriefCampaigns();
     }
+    if (tab === 'implementation') {
+      this.loadCreateCapabilities();
```
dealako: [minor] Capability refetched on every Implementation entry

Issue: `selectTab('implementation')` calls `loadCreateCapabilities()` unconditionally, which runs the full `listBriefCampaigns(projectSlug, briefId)` query just to read `result.demandGenEnabled`, discarding `campaigns`/`possiblyStale`/`statusToggleEnabled`.

Proof: the sibling email path guards re-fetch with `loadEmailTemplatesIfNeverAnswered()` (line ~848/934), but this create-path read has no equivalent 'already answered' guard. The capability is documented in `campaign.interface.ts` as a deployment fact that does not change within a session, unlike `loadBriefCampaigns` which re-fetches by design because status can go stale.

Why it matters: toggling between tabs issues a query-service round-trip on each entry to re-derive an unchanging boolean.

Fix: gate this dispatch on 'never answered for the active foundation' (only dispatch when `briefCampaignsDemandGenEnabled()` is null for the current slug), keeping the persist-success and restore dispatches that establish the first answer. Non-blocking.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -1180,6 +1180,285 @@ describe('CampaignsComponent brief persistence', () => {
...
+        persistBrief.mockReturnValue(of({ enabled: false, briefId: '', etag: null, created: false, approved: false }));
+        proceed();
+        await fixture.whenStable();
+
+        expect(list).not.toHaveBeenCalled();
+        expect((fixture.componentInstance as unknown as { briefCampaignsDemandGenEnabled(): boolean | null }).briefCampaignsDemandGenEnabled()).toBeNull();
+      });
+
+      /**
+       * because the tab's draft restore clears a saved Demand Gen selection on an explicit `false`
```
dealako: [nit] TSDoc begins mid-sentence with lowercase 'because'

Issue: this block opens with a dangling subordinate clause — the lead-in sentence appears to have been lost during editing, so the comment reads as a fragment and no longer states what it explains.

Proof: every sibling test block in this file opens with a full sentence; this one starts `/**\n * because the tab's draft restore clears a saved Demand Gen selection...`.

Fix: restore the lead clause, e.g. 'The capability stays `null`, not `false`, on a failed read — because the tab's draft restore clears a saved selection on an explicit `false` and a failed read has established nothing.'

dealako (dismissed):
Approved with minor comments. Two inline notes below; both non-blocking. See the summary comment for full context and the open disabled-persist follow-up.

mrautela365:
Thanks for the approval — both items fixed in `fa222a4ea`.

**[minor] Capability refetched on every Implementation entry** — you're right, and the sibling you cited is the right precedent. Guarded on `null`, which needs no extra flag: the foundation-switch effect, `loadBriefCampaigns`' own pre-request clear, and a failed read all reset it to `null`, so the read still re-fires wherever the answer is genuinely unknown — including a retry after failure.

Worth flagging that my first attempt at this was **unenforced**: removing the guard left all 1304 tests green. Added one that pins it as a **call count**, since the capability value is identical either way and only the number of requests distinguishes them. Mutation now fails as it should.

**[nit] TSDoc opening mid-sentence** — a real editing artifact; the lead clause was lost in an earlier revision. Now reads: *"A failed read leaves the capability `null`, never `false` — because the tab's draft restore clears a saved Demand Gen selection on an explicit `false`, and a failed read has established nothing."*

1305 app + 2109 server tests pass; lint and format clean.

---

**Separately, and more important than this PR:** we found the production create failure today. Prod campaign-service logs:

```
msg:   "platform dispatch failed before upstream create (claim released)"
error: "no google-ads connection configured for project tlf: resource not found"
```

Prod has no `system:linuxfoundation` Google Ads connection row, so projects without their own connection have nothing to fall back to. Dev has one (`active`, `account_id 8666746580`, `login_customer_id 9746983954`, created by `sysacct-bootstrap`). Antonia is bootstrapping prod. No code change needed — the cutover routes correctly; the connection data was just never seeded there.

mrautela365:
@dealako re-requesting — my push addressing your two items dismissed the approval (this repo resets on any push to an approved PR).

Both are in `fa222a4ea`: the `null` guard on the capability refetch, and the TSDoc lead clause. Nothing else changed, and the one still-open thread is the disabled-persist gap I'm deliberately leaving for the capability-endpoint decision rather than patching a fifth path.

mrautela365:
Filed the follow-up you asked for: #1931.

It records the mechanism (the capability read needs a persisted brief id, and `persistBrief` returns an empty one when the briefs flag is off), why sending a blank `brief_id` is not the fix (`campaign.controller.ts:871` refuses it with a 400 before the service is reached, and the spec pins that), and your call that the right answer is a dedicated capability endpoint rather than a fourth patch to the persistence flow.

Also noted that nothing is user-visible while `_DEMAND_GEN` is off — it becomes visible only if Demand Gen is enabled in a deployment where brief persistence is not.

dealako:
Thanks for the quick turnaround, @mrautela365 — both of my prior notes are addressed cleanly.

### 👏 Nice work
- The refetch minor is fixed properly, not papered over: `selectTab('implementation')` now guards on `briefCampaignsDemandGenEnabled() === null` (campaigns.component.ts:956), mirroring `loadEmailTemplatesIfNeverAnswered`. The guard comment correctly documents that the null-reset paths (foundation-switch effect, `loadBriefCampaigns` pre-request clear, error arms) keep the read re-firing whenever the answer is genuinely unknown, including retry-after-failure.
- The new `does not refetch the capability once it is known` test (spec:1470) is mutation-resistant: deleting the guard makes re-entry dispatch `loadCreateCapabilitiesFor` with the persisted non-null brief id, bumping the call count and failing the assertion. Pinned as a call count, which is the only thing that distinguishes the two states.
- The `because` docblock fragment now reads as a full sentence and its rationale matches the field docblock.

### Revision tracking
- ✅ **Resolved** — Capability refetched on every Implementation entry (minor). Guarded + mutation-tested in `b4c8a8f17` / `fa222a4ea`.
- ✅ **Resolved** — Test docblock mid-sentence "because" (nit). Rewritten in `fa222a4ea`.
- ❌ **Still open** — Success/success race tests use identical `true`/`true` payloads (nit). Untouched by these commits, and you already documented agreement-vs-ordering as intentionally out of scope in the race-test comment, so this stays a take-it-or-leave-it note, not a regression.

### New issue count (carried + new)
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 — (1) the still-open race-test payload note above; (2) new: the docblock at spec:1247 says the trigger is "at lines 947-949", but the guard insertion pushed `loadCreateCapabilities()` to line 957 (the branch now spans 947-958). Naming the symbol (`the `tab === 'implementation'` branch of `selectTab``) would stop it re-drifting.

Security and privacy re-scan of the new commits came back clean — the perf guard only narrows when the read dispatches, carrying the same `projectSlug` + `briefId` (no PII), and cannot widen a read.

The disabled-persist gap (dedicated capability endpoint) is now off the thread list; still worth a follow-up ticket if not already filed.

### Decision
✅ **Approved with minor comments.** Both nits are optional cleanup; nothing blocks merge.

dealako (approved):
Approved with minor comments (follow-up round). Both prior notes resolved: the refetch minor is now guarded on `briefCampaignsDemandGenEnabled() === null` with a mutation-resistant call-count test, and the sentence-fragment docblock reads correctly. Two optional nits remain, folded into the summary comment: the still-open race-test true/true payloads (documented as intentional scope) and a stale line-number reference at campaigns.component.spec.ts:1247 ("lines 947-949" -> the trigger is now line 957). Security/privacy re-scan of the new commits is clean. Nothing blocks merge.

PR #1907, opened 2026-08-27
feat(meetings): add public project calendar page

MRashad26:
## Summary
- Adds an unauthenticated, project-scoped public calendar page at `/projects/:projectSlug/calendar`, mirroring lfx-pcc's Public Calendar page (`?committee=<uid>` sub-filter, `?view=month|week`), reusing the existing `lfx-fullcalendar` wrapper.
- Adds a new public JSON endpoint `GET /public/api/projects/:id/meetings` (slug or UID, optional `committee` query param via `tags_all`), stripping PII (`created_by`/`owner`) per the existing public-endpoint pattern (LFXV2-2802).
- Fixes an independent bug: `buildVCalendar()` in `ics.helper.ts` never emitted `X-WR-CALNAME`, so subscribed project/committee calendars showed up unnamed in Google Calendar/Apple Calendar. Threaded the resolved project/committee name through both `getProjectCalendar` and `getCommitteeCalendar`, with a graceful fallback (still 200, no calname) if name resolution fails.

## Test plan
- [x] `yarn format:check`, `yarn lint:check`, `yarn build` all pass
- [x] `yarn vitest run` — new/updated specs for `ics.helper.ts`, `project.controller.ts`, `committee.controller.ts` all pass
- [x] Post-commit reviewer trio (general + standards) run on every commit, findings addressed
- [x] Full-branch sweep (general-code-reviewer against origin/main) clean — 0 Critical, 0 Important
- [ ] Manual verification: visit `/projects/:projectSlug/calendar` (and with `?committee=` / `?view=week`) once deployed

LFXV2-3340

### DIFF
--- a/apps/lfx-one/src/app/app.routes.ts
+++ b/apps/lfx-one/src/app/app.routes.ts
@@ -507,6 +507,12 @@ export const routes: Routes = [
     path: 'projects/:projectSlug/groups',
     loadComponent: () => import('./modules/groups/public-project-groups/public-project-groups.component').then((m) => m.PublicProjectGroupsComponent),
   },
+  // Public project calendar — month/week calendar of a project's public meetings, optionally scoped
+  // to a single committee via ?committee=<uid> (no auth required).
+  {
+    path: 'projects/:projectSlug/calendar',
+    loadComponent: () => import('./modules/meetings/public-project-calendar/public-project-calendar.component').then((m) => m.PublicProjectCalendarComponent),
+  },
   // Invite acceptance — authGuard preserves ?token= through the Auth0 login redirect.
   {
     path: 'invite',
--- a/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.spec.ts
@@ -1,0 +1,132 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { provideHttpClient } from '@angular/common/http';
+import { provideHttpClientTesting } from '@angular/common/http/testing';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { ActivatedRoute, provideRouter } from '@angular/router';
+import { MeetingService } from '@services/meeting.service';
+import { headerTestProviders, installMatchMediaShim } from '@shared/testing/header-test-providers';
+import { BehaviorSubject, of, throwError } from 'rxjs';
+import { beforeAll, describe, expect, it, vi } from 'vitest';
+
+import type { PublicCalendarMeeting, PublicProjectMeetingsResponse } from '@lfx-one/shared/interfaces';
+
+import { PublicProjectCalendarComponent } from './public-project-calendar.component';
+
+beforeAll(installMatchMediaShim);
+
+function meeting(over: Partial<PublicCalendarMeeting> = {}): PublicCalendarMeeting {
+  return {
+    id: 'meeting-1',
+    title: 'Technical Steering Committee',
+    start_time: '2026-09-01T15:00:00Z',
+    duration: 60,
+    timezone: 'America/New_York',
+    ...over,
+  };
+}
+
+function response(over: Partial<PublicProjectMeetingsResponse> = {}): PublicProjectMeetingsResponse {
+  return {
+    meetings: [meeting()],
+    total: 1,
+    project: { uid: 'p1', name: 'Kubernetes' },
+    ...over,
+  };
+}
+
+describe('PublicProjectCalendarComponent', () => {
+  let fixture: ComponentFixture<PublicProjectCalendarComponent>;
+  let getPublicProjectMeetings: ReturnType<typeof vi.fn>;
+  let queryParamMap: BehaviorSubject<Map<string, string>>;
+
+  async function render(
+    options: {
+      response?: PublicProjectMeetingsResponse;
+      fail?: boolean;
+      slug?: string;
+      query?: Map<string, string>;
+    } = {}
+  ): Promise<void> {
+    const { fail = false, slug = 'kubernetes' } = options;
+    queryParamMap = new BehaviorSubject(options.query ?? new Map<string, string>());
+    getPublicProjectMeetings = vi.fn(() => (fail ? throwError(() => new Error('upstream 503')) : of(options.response ?? response())));
+
+    TestBed.resetTestingModule();
+    await TestBed.configureTestingModule({
+      imports: [PublicProjectCalendarComponent],
+      providers: [
+        ...headerTestProviders(),
+        provideRouter([]),
+        provideHttpClient(),
+        provideHttpClientTesting(),
+        { provide: MeetingService, useValue: { getPublicProjectMeetings } },
+        // After provideRouter, which also provides ActivatedRoute — the last provider wins, and the
+        // router's own empty paramMap would otherwise silently swallow the slug/committee assertions.
+        {
+          provide: ActivatedRoute,
+          useValue: { paramMap: of(new Map([['projectSlug', slug]])), queryParamMap: queryParamMap.asObservable() },
+        },
+      ],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(PublicProjectCalendarComponent);
+    await fixture.whenStable();
+  }
+
+  it('renders the project name and meeting count once the feed resolves', async () => {
+    await render();
+
+    const title = fixture.nativeElement.querySelector('[data-testid="public-project-calendar-title"]');
+    const subtitle = fixture.nativeElement.querySelector('[data-testid="public-project-calendar-subtitle"]');
+    expect(title?.textContent?.trim()).toBe('Kubernetes — Calendar');
+    expect(subtitle?.textContent?.trim()).toBe('1 meeting');
+    expect(fixture.nativeElement.querySelector('[data-testid="public-project-calendar-container"]')).not.toBeNull();
+  });
+
+  it('shows the error state and no calendar when the feed fails', async () => {
+    await render({ fail: true });
+
+    expect(fixture.nativeElement.querySelector('[data-testid="public-project-calendar-error"]')).not.toBeNull();
+    expect(fixture.nativeElement.querySelector('[data-testid="public-project-calendar-container"]')).toBeNull();
+  });
+
+  it('scopes the request to a committee when ?committee= is present', async () => {
+    await render({ query: new Map([['committee', 'committee-uid-1234']]) });
+
+    expect(getPublicProjectMeetings).toHaveBeenCalledWith('kubernetes', 'committee-uid-1234');
+  });
+
+  it('opens on the week view when ?view=week, and the month view otherwise', async () => {
+    await render({ query: new Map([['view', 'week']]) });
+    expect((fixture.componentInstance as unknown as { initialView: () => string }).initialView()).toBe('timeGridWeek');
+
+    await render();
+    expect((fixture.componentInstance as unknown as { initialView: () => string }).initialView()).toBe('dayGridMonth');
+  });
+
+  it('does not refetch when only ?view changes', async () => {
+    // Regression lock: the fetch pipeline reads queryParamMap for `committee`, so without
+    // distinctUntilChanged on {slug, committeeUid} a pure month/week toggle would re-enter
+    // switchMap, re-hit the upstream feed, and flash the loading skeleton.
+    await render();
+    expect(getPublicProjectMeetings).toHaveBeenCalledTimes(1);
+
+    queryParamMap.next(new Map([['view', 'week']]));
+    await fixture.whenStable();
+
+    expect(getPublicProjectMeetings).toHaveBeenCalledTimes(1);
+  });
+
+  it('refetches when the committee filter changes', async () => {
+    await render();
+    expect(getPublicProjectMeetings).toHaveBeenCalledTimes(1);
+
+    queryParamMap.next(new Map([['committee', 'committee-uid-1234']]));
+    await fixture.whenStable();
+
+    expect(getPublicProjectMeetings).toHaveBeenCalledTimes(2);
+    expect(getPublicProjectMeetings).toHaveBeenLastCalledWith('kubernetes', 'committee-uid-1234');
+  });
+});
--- a/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
@@ -29,10 +29,6 @@ export class PublicProjectCalendarComponent {
 
   private readonly calendarData: Signal<PublicProjectMeetingsResponse | null> = this.initCalendarData();
 
-  protected readonly committeeUid: Signal<string | undefined> = toSignal(this.route.queryParamMap.pipe(map((params) => params.get('committee') ?? undefined)), {
-    initialValue: undefined,
-  });
-
   protected readonly initialView: Signal<string> = toSignal(
     this.route.queryParamMap.pipe(map((params) => (params.get('view') === 'week' ? 'timeGridWeek' : 'dayGridMonth'))),
     { initialValue: 'dayGridMonth' }
--- a/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
@@ -1,0 +1,82 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { Component, computed, inject, signal, Signal } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
+import { ActivatedRoute, Router } from '@angular/router';
+import { FullCalendarComponent } from '@components/fullcalendar/fullcalendar.component';
+import { EmptyStateComponent } from '@components/empty-state/empty-state.component';
+import { HeaderComponent } from '@components/header/header.component';
+import { EventClickArg, EventInput } from '@fullcalendar/core';
+import { MeetingCalendarClickProps, PublicProjectMeetingsResponse } from '@lfx-one/shared/interfaces';
+import { meetingToCalendarEvents, resolveMeetingCalendarClickRoute } from '@lfx-one/shared/utils';
+import { MeetingService } from '@services/meeting.service';
+import { SkeletonModule } from 'primeng/skeleton';
+import { catchError, combineLatest, map, of, switchMap } from 'rxjs';
+
+@Component({
+  selector: 'lfx-public-project-calendar',
+  imports: [FullCalendarComponent, EmptyStateComponent, HeaderComponent, SkeletonModule],
+  templateUrl: './public-project-calendar.component.html',
+})
+export class PublicProjectCalendarComponent {
+  private readonly route = inject(ActivatedRoute);
+  private readonly router = inject(Router);
+  private readonly meetingService = inject(MeetingService);
+
+  protected readonly loading = signal(true);
+  protected readonly fetchError = signal(false);
+
+  private readonly calendarData: Signal<PublicProjectMeetingsResponse | null> = this.initCalendarData();
+
+  protected readonly committeeUid: Signal<string | undefined> = toSignal(this.route.queryParamMap.pipe(map((params) => params.get('committee') ?? undefined)), {
+    initialValue: undefined,
+  });
+
+  protected readonly initialView: Signal<string> = toSignal(
+    this.route.queryParamMap.pipe(map((params) => (params.get('view') === 'week' ? 'timeGridWeek' : 'dayGridMonth'))),
+    { initialValue: 'dayGridMonth' }
+  );
+
+  protected readonly projectName: Signal<string> = computed(() => this.calendarData()?.project?.name ?? '');
+  protected readonly total: Signal<number> = computed(() => this.calendarData()?.total ?? 0);
+
+  protected readonly calendarEvents: Signal<EventInput[]> = computed(() =>
+    (this.calendarData()?.meetings ?? []).flatMap((m) => meetingToCalendarEvents(m) as EventInput[])
+  );
+
+  /** Handles FullCalendar event click — navigates to the public meeting join page. Cancelled occurrences are inert. */
+  protected onCalendarEventClick(arg: EventClickArg): void {
+    const route = resolveMeetingCalendarClickRoute(arg.event.extendedProps as MeetingCalendarClickProps, arg.event.start);
+    if (!route) {
+      return;
+    }
+    void this.router.navigate(route.path, route.queryParams ? { queryParams: route.queryParams } : undefined);
+  }
+
+  // Private initializer functions
+
+  private initCalendarData(): Signal<PublicProjectMeetingsResponse | null> {
+    return toSignal(
+      combineLatest([this.route.paramMap, this.route.queryParamMap]).pipe(
+        map(([params, queryParams]) => ({ slug: params.get('projectSlug') ?? '', committeeUid: queryParams.get('committee') ?? undefined })),
+        switchMap(({ slug, committeeUid }) => {
+          this.loading.set(true);
+          this.fetchError.set(false);
+          return this.meetingService.getPublicProjectMeetings(slug, committeeUid).pipe(
+            map((response) => {
+              this.loading.set(false);
+              return response;
+            }),
+            catchError(() => {
+              this.loading.set(false);
+              this.fetchError.set(true);
+              return of(null);
+            })
+          );
+        })
+      ),
+      { initialValue: null }
+    );
+  }
+}
--- a/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
+++ b/apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
@@ -4,15 +4,15 @@
 import { Component, computed, inject, signal, Signal } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
 import { ActivatedRoute, Router } from '@angular/router';
-import { FullCalendarComponent } from '@components/fullcalendar/fullcalendar.component';
+import { FullCalendarComponent } from '@app/shared/components/fullcalendar/fullcalendar.component';
 import { EmptyStateComponent } from '@components/empty-state/empty-state.component';
 import { HeaderComponent } from '@components/header/header.component';
 import { EventClickArg, EventInput } from '@fullcalendar/core';
 import { MeetingCalendarClickProps, PublicProjectMeetingsResponse } from '@lfx-one/shared/interfaces';
-import { meetingToCalendarEvents, resolveMeetingCalendarClickRoute } from '@lfx-one/shared/utils';
+import { publicMeetingToCalendarEvents, resolveMeetingCalendarClickRoute } from '@lfx-one/shared/utils';
 import { MeetingService } from '@services/meeting.service';
 import { SkeletonModule } from 'primeng/skeleton';
-import { catchError, combineLatest, map, of, switchMap } from 'rxjs';
+import { catchError, combineLatest, distinctUntilChanged, map, of, switchMap } from 'rxjs';
 
 @Component({
   selector: 'lfx-public-project-calendar',
@@ -29,16 +29,15 @@ export class PublicProjectCalendarComponent {
 
   private readonly calendarData: Signal<PublicProjectMeetingsResponse | null> = this.initCalendarData();
 
-  protected readonly initialView: Signal<string> = toSignal(
-    this.route.queryParamMap.pipe(map((params) => (params.get('view') === 'week' ? 'timeGridWeek' : 'dayGridMonth'))),
-    { initialValue: 'dayGridMonth' }
-  );
+  protected readonly initialView: Signal<string> = this.initInitialView();
 
   protected readonly projectName: Signal<string> = computed(() => this.calendarData()?.project?.name ?? '');
   protected readonly total: Signal<number> = computed(() => this.calendarData()?.total ?? 0);
 
+  // publicMeetingToCalendarEvents (not meetingToCalendarEvents) — the public mapper never places a
+  // meeting password in extendedProps, so a click can't forward one into the join URL, history, or referrer.
   protected readonly calendarEvents: Signal<EventInput[]> = computed(() =>
-    (this.calendarData()?.meetings ?? []).flatMap((m) => meetingToCalendarEvents(m) as EventInput[])
+    (this.calendarData()?.meetings ?? []).flatMap((m) => publicMeetingToCalendarEvents(m) as EventInput[])
   );
 
   /** Handles FullCalendar event click — navigates to the public meeting join page. Cancelled occurrences are inert. */
@@ -52,10 +51,19 @@ export class PublicProjectCalendarComponent {
 
   // Private initializer functions
 
+  private initInitialView(): Signal<string> {
+    return toSignal(this.route.queryParamMap.pipe(map((params) => (params.get('view') === 'week' ? 'timeGridWeek' : 'dayGridMonth'))), {
+      initialValue: 'dayGridMonth',
+    });
+  }
+
   private initCalendarData(): Signal<PublicProjectMeetingsResponse | null> {
     return toSignal(
       combineLatest([this.route.paramMap, this.route.queryParamMap]).pipe(
         map(([params, queryParams]) => ({ slug: params.get('projectSlug') ?? '', committeeUid: queryParams.get('committee') ?? undefined })),
+        // Only slug and committee affect the request — without this, switching ?view=month|week would
+        // re-enter switchMap, refetching the feed and flashing the skeleton on a pure view toggle.
+        distinctUntilChanged((a, b) => a.slug === b.slug && a.committeeUid === b.committeeUid),
         switchMap(({ slug, committeeUid }) => {
           this.loading.set(true);
           this.fetchError.set(false);
--- a/apps/lfx-one/src/server/controllers/project.controller.ts
+++ b/apps/lfx-one/src/server/controllers/project.controller.ts
@@ -7,6 +7,8 @@ import {
   AddUserToProjectRequest,
   CreateProjectDocumentRequest,
   Meeting,
+  PastMeeting,
+  PublicCalendarMeeting,
   PublicProjectMeetingsResponse,
   UpdateUserRoleRequest,
   UploadProjectDocumentRequest,
@@ -991,22 +993,30 @@ export class ProjectController {
       const [upcoming, past, project] = await Promise.all([
         fetchAllMeetingPages((token) => this.meetingService.getMeetings(req, token ? { ...query, page_token: token } : query, 'v1_meeting', false)),
         fetchAllMeetingPages((token) => this.meetingService.getMeetings(req, token ? { ...query, page_token: token } : query, 'v1_past_meeting', false)),
-        this.projectService.getProjectById(req, projectUid, false),
+        // Project metadata only feeds the page header — degrade gracefully rather than failing the whole
+        // feed on a transient project-service error, matching the X-WR-CALNAME fallback in getProjectCalendar.
+        this.projectService.getProjectById(req, projectUid, false).catch((error) => {
+          logger.warning(req, 'get_project_meetings', 'Failed to resolve project metadata for the calendar header', {
+            project_id: projectUid,
+            err: error,
+          });
+          return undefined;
+        }),
       ]);
 
       // Filter PRIVATE meetings from the public feed. Restricted (invited-guests-only) public meetings are
       // still listed so their existence is discoverable; join authorization is enforced separately at join time.
-      // created_by/owner carry organizer PII (name/email) — authenticated-visible only, per the same
-      // strip applied to anonymous callers in PublicMeetingController.getMeetingById (LFXV2-2802).
-      const meetings = [...upcoming, ...past]
+      // The projection is an allowlist, not a strip: this is an unauthenticated bulk-export endpoint, so
+      // credentials (password/passcode/host_key/zoom_config) and organizer PII (created_by/owner/organizers)
+      // must stay out of the payload even as new Meeting fields land upstream (LFXV2-2802).
+      const meetings: PublicCalendarMeeting[] = [...upcoming, ...past]
         .filter((m) => m.visibility === MeetingVisibility.PUBLIC)
-        .map((m) => {
-          const meeting = { ...m };
-          delete (meeting as Partial<Meeting>).created_by;
-          delete (meeting as Partial<Meeting>).owner;
-          return meeting;
-        });
-      const response: PublicProjectMeetingsResponse = { meetings, total: meetings.length, project: { uid: project.uid, name: project.name } };
+        .map((m) => this.toPublicCalendarMeeting(m));
+      const response: PublicProjectMeetingsResponse = {
+        meetings,
+        total: meetings.length,
+        project: { uid: projectUid, name: project?.name ?? '' },
+      };
 
       logger.success(req, 'get_project_meetings', startTime, {
         project_id: id,
@@ -1014,6 +1024,9 @@ export class ProjectController {
         meeting_count: meetings.length,
       });
 
+      // Shorter than calendar.ics's 15 minutes: calendar clients poll the ICS feed on their own schedule,
+      // while this feed backs an interactive page where a newly scheduled meeting should surface quickly.
+      res.setHeader('Cache-Control', 'public, max-age=300');
       res.json(response);
     } catch (error) {
       next(error);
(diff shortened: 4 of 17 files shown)


dealako:
@MRashad26 — solid feature scope here: the public calendar page, JSON feed, and the `X-WR-CALNAME` ICS fix are well aligned with existing public-meeting patterns, and the follow-up commits show good responsiveness to the pre-PR reviewer trio (PII stripping, committee UUID validation, test coverage). The controller specs for `getProjectMeetings` and the ICS calname paths are a strong signal.

That said, I found issues that block merge — two are credential/privacy exposure on the new anonymous API, and one is an SSR auth gap that prevents the page from rendering for logged-out visitors. I agree with Copilot and Cursor Bugbot on the credential leak; the auth allowlist gap is the other functional blocker.

**Structured issue count**
- 🔴 Blocking: 3 issues (incl. 2 privacy): anonymous API returns meeting credentials; client propagates `password` into join URLs; SSR auth allowlist missing for `/projects/:slug/calendar`
- 🟡 Minor: 5 issues: `?view` retriggers full fetch; no component spec; missing `Cache-Control`; project metadata failure aborts meetings response; shared response type overstates returned meeting shape
- ⚪ Nit: 3 issues: mixed `toSignal` init pattern; FullCalendar import alias drift; sparse JSDoc on `PublicProjectMeetingsResponse`
- ❔ Question: 0 items

🔒 **Privacy: 2 findings (blocking)**
(included in the 🔴 count above)
1. Public meetings API returns credential fields: `project.controller.ts:1003` — wholesale spread leaves `password`/`passcode`/`host_key`/`zoom_config` on anonymous JSON; use a minimal public DTO like `PublicMeetingController` does
2. Calendar client propagates password: `public-project-calendar.component.ts:41` — `meetingToCalendarEvents` copies `password` into extendedProps and join query params

**Bot reconciliation**
- **Agree** with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1907#discussion_r3876074989) on the missing SSR auth allowlist for `/projects/:slug/calendar`
- **Agree** with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1907#discussion_r3876075049) and [Cursor Bugbot](https://github.com/linuxfoundation/lfx-self-serve/pull/1907#discussion_r3876075693) on credential fields leaking through `getProjectMeetings`; the existing `PublicMeetingController` minimal-projection pattern is the right model
- CodeRabbit skipped auto-review on this repo (no substantive findings to reconcile)

apps/lfx-one/src/server/controllers/project.controller.ts
```diff
@@ -946,6 +960,66 @@ export class ProjectController {
...
+        this.projectService.getProjectById(req, projectUid, false),
+      ]);
+
+      // Filter PRIVATE meetings from the public feed. Restricted (invited-guests-only) public meetings are
+      // still listed so their existence is discoverable; join authorization is enforced separately at join time.
+      // created_by/owner carry organizer PII (name/email) — authenticated-visible only, per the same
+      // strip applied to anonymous callers in PublicMeetingController.getMeetingById (LFXV2-2802).
+      const meetings = [...upcoming, ...past]
+        .filter((m) => m.visibility === MeetingVisibility.PUBLIC)
+        .map((m) => {
```
dealako: [blocking][privacy] Public meetings API returns full Meeting records with credential fields

Issue: `getProjectMeetings` spreads each upstream meeting and deletes only `created_by`/`owner`. Credential fields (`password`, `passcode`, `host_key`, `zoom_config`, `organizers`) remain in the anonymous JSON response.

Proof: Lines 1003–1007 copy the full indexed `Meeting` before `res.json()`. `Meeting` defines `password` and `zoom_config` in `packages/shared/src/interfaces/meeting.interface.ts`. Contrast `PublicMeetingController` (line ~390): "Minimal projection only — the full ITX payload carries sensitive fields (e.g. password)" with spec asserting `password` is absent from anonymous responses.

Why it matters: This is an unauthenticated bulk-export endpoint. Any password on indexed public meetings is returned in one request, bypassing the per-meeting gate on `/public/api/meetings/:id`.

Fix: Return a minimal public calendar DTO (id, title, start_time, duration, timezone, occurrence timestamps/status). Explicitly omit credentials. Add tests seeding `password`/`passcode` and asserting absence from `res.json` output.

*(Agrees with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1907#discussion_r3876075049) and [Cursor Bugbot](https://github.com/linuxfoundation/lfx-self-serve/pull/1907#discussion_r3876075693).)*

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
```diff
@@ -0,0 +1,78 @@
...
+  protected readonly initialView: Signal<string> = toSignal(
+    this.route.queryParamMap.pipe(map((params) => (params.get('view') === 'week' ? 'timeGridWeek' : 'dayGridMonth'))),
+    { initialValue: 'dayGridMonth' }
+  );
+
+  protected readonly projectName: Signal<string> = computed(() => this.calendarData()?.project?.name ?? '');
+  protected readonly total: Signal<number> = computed(() => this.calendarData()?.total ?? 0);
+
+  protected readonly calendarEvents: Signal<EventInput[]> = computed(() =>
+    (this.calendarData()?.meetings ?? []).flatMap((m) => meetingToCalendarEvents(m) as EventInput[])
```
dealako: [blocking][privacy] Public calendar client propagates meeting.password into join URLs

Issue: Calendar events are built via `meetingToCalendarEvents`, which copies `meeting.password` into FullCalendar `extendedProps`. `resolveMeetingCalendarClickRoute` then appends it to `/meetings/:id` query params on click.

Proof: `packages/shared/src/utils/meeting-calendar.utils.ts:54,88` set `extendedProps.password`; `:106–114` forward it into router `queryParams`. This component calls both utilities on API data from `getPublicProjectMeetings`.

Why it matters: If the API returns a password (see server finding), it lands in client-side event state, browser history, and referrer headers — expanding credential disclosure beyond the JSON response.

Fix: Strip credentials server-side (required). For defense in depth, use a public-safe calendar mapper that omits `password` from extendedProps and navigates without password query params for anonymous calendar clicks.

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -507,6 +507,12 @@ export const routes: Routes = [
     path: 'projects/:projectSlug/groups',
     loadComponent: () => import('./modules/groups/public-project-groups/public-project-groups.component').then((m) => m.PublicProjectGroupsComponent),
   },
+  // Public project calendar — month/week calendar of a project's public meetings, optionally scoped
+  // to a single committee via ?committee=<uid> (no auth required).
+  {
+    path: 'projects/:projectSlug/calendar',
```
dealako, line 513: [blocking] SSR auth allowlist missing for /projects/:slug/calendar

Issue: This route is registered as a no-auth public page, but `auth.middleware.ts` only allowlists `/projects/:slug/groups` (line 56). Anonymous GET `/projects/:slug/calendar` falls through to the catch-all `{ pattern: '/', auth: 'required' }` and redirects to login before Angular SSR runs.

Proof: Comment here says "no auth required". `auth.middleware.spec.ts` covers groups anonymous access but not calendar. The JSON feed at `/public/api/projects/:id/meetings` is anonymously reachable — only the SSR shell is blocked.

Why it matters: Logged-out visitors cannot reach the feature page; the API remains scrapeable directly.

Fix: Add `{ pattern: /^\/projects\/[^/]+\/calendar\/?$/, type: 'ssr', auth: 'optional' }` beside the groups rule in `auth.middleware.ts`, plus a matching spec test.

*(Agrees with [Copilot](https://github.com/linuxfoundation/lfx-self-serve/pull/1907#discussion_r3876074989).)*

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
```diff
@@ -0,0 +1,78 @@
...
+      return;
+    }
+    void this.router.navigate(route.path, route.queryParams ? { queryParams: route.queryParams } : undefined);
+  }
+
+  // Private initializer functions
+
+  private initCalendarData(): Signal<PublicProjectMeetingsResponse | null> {
+    return toSignal(
+      combineLatest([this.route.paramMap, this.route.queryParamMap]).pipe(
```
dealako, line 62: [minor] ?view query param retriggers full meetings fetch

Issue: `combineLatest([paramMap, queryParamMap])` re-enters `switchMap(getPublicProjectMeetings)` on every query param change, including `?view=week`, even though only `projectSlug` and `committee` affect the fetch.

Proof: `initialView` reads `view` separately (line 32–34). The fetch `map` extracts slug/committee only; no `distinctUntilChanged` on that pair.

Why it matters: Switching month/week view causes redundant API calls, skeleton flash, and extra upstream load.

Fix: Drive fetch off `paramMap` plus `queryParamMap.pipe(map(p => p.get('committee')), distinctUntilChanged())`, or `distinctUntilChanged` on the mapped `{slug, committeeUid}` object.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
```diff
@@ -0,0 +1,78 @@
...
+import { MeetingService } from '@services/meeting.service';
+import { SkeletonModule } from 'primeng/skeleton';
+import { catchError, combineLatest, map, of, switchMap } from 'rxjs';
+
+@Component({
+  selector: 'lfx-public-project-calendar',
+  imports: [FullCalendarComponent, EmptyStateComponent, HeaderComponent, SkeletonModule],
+  templateUrl: './public-project-calendar.component.html',
+})
+export class PublicProjectCalendarComponent {
```
dealako, line 22: [minor] No unit tests for PublicProjectCalendarComponent

Issue: New public page has no `.component.spec.ts`, unlike sibling `public-project-groups` which tests loading/error/render paths.

Proof: Glob returns only `.ts` and `.html` for this component. No spec references `PublicProjectCalendarComponent` or `getPublicProjectMeetings`.

Why it matters: Regressions in loading/error states, committee scoping, or calendar wiring won't be caught in CI.

Fix: Add `public-project-calendar.component.spec.ts` mirroring `public-project-groups.component.spec.ts`.

apps/lfx-one/src/server/controllers/project.controller.ts
```diff
@@ -946,6 +960,66 @@ export class ProjectController {
...
+        });
+      const response: PublicProjectMeetingsResponse = { meetings, total: meetings.length, project: { uid: project.uid, name: project.name } };
+
+      logger.success(req, 'get_project_meetings', startTime, {
+        project_id: id,
+        committee_uid: committeeUid,
+        meeting_count: meetings.length,
+      });
+
+      res.json(response);
```
dealako, line 1030: [minor] getProjectMeetings returns no Cache-Control while calendar.ics caches

Issue: Sibling `getProjectCalendar` sets `Cache-Control: public, max-age=900` (line ~955). `getProjectMeetings` calls `res.json(response)` with no cache header.

Why it matters: Every page load (and spurious `?view` refetch) hits upstream services with no browser caching.

Fix: Set an appropriate `Cache-Control` aligned with `calendar.ics`, or document `private, no-cache` if freshness is paramount.

apps/lfx-one/src/server/controllers/project.controller.ts
```diff
@@ -946,6 +960,66 @@ export class ProjectController {
...
+        req.bearerToken = await generateM2MToken(req);
+      }
+
+      const projectUid = await this.resolveProjectIdentifier(req, id);
+      const query = committeeUid ? { tags_all: [`project_uid:${projectUid}`, `committee_uid:${committeeUid}`] } : { tags: `project_uid:${projectUid}` };
+
+      const [upcoming, past, project] = await Promise.all([
+        fetchAllMeetingPages((token) => this.meetingService.getMeetings(req, token ? { ...query, page_token: token } : query, 'v1_meeting', false)),
+        fetchAllMeetingPages((token) => this.meetingService.getMeetings(req, token ? { ...query, page_token: token } : query, 'v1_past_meeting', false)),
+        this.projectService.getProjectById(req, projectUid, false),
```
dealako: [minor] Project metadata fetch failure aborts entire meetings response

Issue: `getProjectById` is awaited inside `Promise.all` without a graceful fallback. `getProjectCalendar` wraps the same call in `.catch(() => undefined)` for calname resolution.

Why it matters: A transient project-service failure returns 500 with no meetings, even when meeting data was fetched successfully.

Fix: Catch `getProjectById` failures and return meetings with `project: { uid: projectUid, name: '' }`, matching the ICS degrade-gracefully pattern.

packages/shared/src/interfaces/meeting.interface.ts
```diff
@@ -1442,3 +1442,10 @@ export interface MeLensMeetingFilters {
   /** Viewer username/LFID used by the `organizerOnly` predicate; null disables matching. */
   viewerUsername: string | null;
 }
+
+/** Response envelope returned by `GET /public/api/projects/:id/meetings` — the public calendar page's meeting list. */
+export interface PublicProjectMeetingsResponse {
+  meetings: Meeting[];
```
dealako: [minor] PublicProjectMeetingsResponse overstates returned meeting shape

Issue: `meetings: Meeting[]` documents organizer PII fields the controller explicitly strips (`created_by`, `owner`).

Proof: Controller deletes those fields before `res.json`. Peer `PublicMeetingOccurrencesResponse` uses slim `MeetingOccurrenceSummary[]`, not full `Meeting`.

Why it matters: Shared types are the cross-layer contract; over-broad typing hides the LFXV2-2802 guarantee.

Fix: Introduce a public-safe meeting type (e.g. `Omit<Meeting, 'created_by' | 'owner' | 'password' | ...>`) and use it here.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
```diff
@@ -0,0 +1,78 @@
...
+  private readonly route = inject(ActivatedRoute);
+  private readonly router = inject(Router);
+  private readonly meetingService = inject(MeetingService);
+
+  protected readonly loading = signal(true);
+  protected readonly fetchError = signal(false);
+
+  private readonly calendarData: Signal<PublicProjectMeetingsResponse | null> = this.initCalendarData();
+
+  protected readonly initialView: Signal<string> = toSignal(
```
dealako: [nit] Mixed toSignal initialization pattern

Issue: `calendarData` uses a private `initCalendarData()` helper while `initialView` uses an inline `toSignal`, diverging from `.claude/rules/component-organization.md` §3.

Fix: Extract `private initInitialView(): Signal<string>` mirroring `initCalendarData()`.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
```diff
@@ -0,0 +1,78 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { Component, computed, inject, signal, Signal } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
+import { ActivatedRoute, Router } from '@angular/router';
+import { FullCalendarComponent } from '@components/fullcalendar/fullcalendar.component';
```
dealako: [nit] Inconsistent FullCalendar import alias within meetings feature

Issue: This file imports via `@components/...` while `meetings-dashboard.component.ts` and `committee-meetings.component.ts` use `@app/shared/components/...`.

Fix: Align with the meetings-dashboard import path.

packages/shared/src/interfaces/meeting.interface.ts
```diff
@@ -1442,3 +1442,10 @@ export interface MeLensMeetingFilters {
   /** Viewer username/LFID used by the `organizerOnly` predicate; null disables matching. */
   viewerUsername: string | null;
 }
+
+/** Response envelope returned by `GET /public/api/projects/:id/meetings` — the public calendar page's meeting list. */
```
dealako: [nit] Sparse JSDoc on PublicProjectMeetingsResponse

Issue: Single-line envelope comment with no field-level docs on PII/credential guarantees. Peer `PublicMeetingOccurrencesResponse` includes `@description` and per-field comments.

Fix: Add `@description` plus field JSDoc noting stripped PII and slim `project` envelope.

dealako (changes requested):
## Review summary

@MRashad26 — this PR adds the right surfaces (public calendar page, JSON feed, named ICS subscriptions) and the review-driven follow-ups (PII strip, committee UUID validation, calname tests) are in good shape. Three blockers remain before approval.

### 🔴 Blocking (3, incl. 2 privacy)

1. **`getProjectMeetings` returns full indexed meetings with credentials** (`project.controller.ts`) — only `created_by`/`owner` are deleted; `password`, `passcode`, `host_key`, and `zoom_config` can reach anonymous callers. Mirror the minimal projection in `PublicMeetingController` and add regression tests asserting credentials are absent. *(Agrees with Copilot + Cursor Bugbot.)*

2. **Public calendar client propagates `password` into join URLs** (`public-project-calendar.component.ts`) — `meetingToCalendarEvents` / `resolveMeetingCalendarClickRoute` forward `password` into extendedProps and query params. Strip on the server and omit from the public calendar click path.

3. **SSR auth allowlist missing for `/projects/:slug/calendar`** (`auth.middleware.ts`) — route is registered as public in `app.routes.ts` but only `/projects/:slug/groups` is allowlisted; anonymous visitors hit the catch-all `required` rule and redirect to login. *(Agrees with Copilot.)*

### 🟡 Minor (5)

- `?view=week` retriggers the full meetings fetch (`public-project-calendar.component.ts:57`)
- No `public-project-calendar.component.spec.ts` (sibling `public-project-groups` has one)
- `getProjectMeetings` sends no `Cache-Control` while sibling `calendar.ics` caches 15m
- `getProjectById` failure in `Promise.all` aborts the whole response (ICS handler degrades gracefully)
- `PublicProjectMeetingsResponse.meetings: Meeting[]` overstates the anonymous shape

### ⚪ Nit (3)

- Mixed `toSignal` init pattern in the new component
- FullCalendar import uses `@components/...` while other meetings code uses `@app/shared/components/...`
- Sparse JSDoc on `PublicProjectMeetingsResponse`

apps/lfx-one/src/app/app.routes.ts
```diff
@@ -507,6 +507,12 @@ export const routes: Routes = [
...
+    loadComponent: () => import('./modules/meetings/public-project-calendar/public-project-calendar.component').then((m) => m.PublicProjectCalendarComponent),
```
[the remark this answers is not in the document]
MRashad26: Done — added an anchored optional-auth SSR rule beside the `/projects/:slug/groups` row in `auth.middleware.ts`:

```ts
{ pattern: /^\/projects\/[^/]+\/calendar\/?$/, type: 'ssr', auth: 'optional' },
```

Anonymous `GET /projects/:slug/calendar` now reaches SSR instead of the catch-all `required` row. `auth.middleware.spec.ts` gains both a route-classification test for anonymous access and the anchored no-fail-open case (`/projects/:slug/calendar/extra` still classifies `required`).

See commit 23406535b.

apps/lfx-one/src/server/controllers/project.controller.ts
```diff
@@ -946,6 +960,66 @@ export class ProjectController {
...
+      // still listed so their existence is discoverable; join authorization is enforced separately at join time.
+      // created_by/owner carry organizer PII (name/email) — authenticated-visible only, per the same
+      // strip applied to anonymous callers in PublicMeetingController.getMeetingById (LFXV2-2802).
+      const meetings = [...upcoming, ...past]
+        .filter((m) => m.visibility === MeetingVisibility.PUBLIC)
+        .map((m) => {
+          const meeting = { ...m };
+          delete (meeting as Partial<Meeting>).created_by;
+          delete (meeting as Partial<Meeting>).owner;
+          return meeting;
```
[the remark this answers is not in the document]
MRashad26: Done — replaced the spread-and-delete with a field-by-field allowlist. `getProjectMeetings` now maps each indexed record through `ProjectController.toPublicCalendarMeeting` into the new `PublicCalendarMeeting` shared type, which carries only `id`, `title`, `start_time`, `duration`, `timezone`, a `MeetingOccurrenceSummary[]` occurrence projection (timestamps and status, no per-occurrence titles), `cancelled_occurrences`, and the two past-meeting identity fields. `password`, `passcode`, `host_key`, `zoom_config`, `organizers`, `created_by`, and `owner` are all structurally out of the payload, and stay out as new `Meeting` fields land upstream.

Regression coverage in `project.controller.spec.ts`: the fixture now seeds `password`/`passcode`/`host_key`/`zoom_config`/`organizers`, one test asserts the exact allowlisted key set (a strip would pass a per-field "is X absent" check while still shipping the next new field), and another asserts none of those values or key names appear in the serialized response — serialized rather than key-checked so a nested `zoom_config.passcode` couldn't slip through.

See commit 23406535b.

apps/lfx-one/src/server/controllers/project.controller.ts
```diff
@@ -946,6 +960,66 @@ export class ProjectController {
...
+        });
```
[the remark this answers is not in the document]
MRashad26: Done — fixed on both sides.

Server: `getProjectMeetings` now builds a `PublicCalendarMeeting` allowlist projection field-by-field instead of spreading the indexed record and deleting two keys, so `password`, `passcode`, `host_key`, `zoom_config`, and `organizers` are structurally absent from the anonymous response.

Client: the calendar page switched from `meetingToCalendarEvents` to a new `publicMeetingToCalendarEvents` mapper that never writes a password into `extendedProps`, so `resolveMeetingCalendarClickRoute` cannot forward one into the join URL, browser history, or referrer — defense in depth behind the server fix.

Covered by new tests in `project.controller.spec.ts` (no credential value or key name is serialized) and `meeting-calendar.utils.spec.ts` (the click route the public mapper produces carries no `password` query param).

See commit 23406535b.

apps/lfx-one/src/server/controllers/project.controller.ts
MRashad26: Done — switched to a minimal public DTO, exactly as suggested.

`packages/shared` gains `PublicCalendarMeeting`, and `PublicProjectMeetingsResponse.meetings` is now `PublicCalendarMeeting[]`. `getProjectMeetings` maps through `ProjectController.toPublicCalendarMeeting`, which names every field it emits:

```ts
private toPublicCalendarMeeting(meeting: Meeting | PastMeeting): PublicCalendarMeeting {
  const pastRow = meeting as Partial<PastMeeting>;
  return {
    id: meeting.id,
    title: meeting.title,
    start_time: meeting.start_time,
    duration: meeting.duration,
    timezone: meeting.timezone,
    occurrences: (meeting.occurrences ?? []).map((occurrence) => ({
      occurrence_id: occurrence.occurrence_id,
      start_time: occurrence.start_time,
      duration: occurrence.duration,
      status: occurrence.status,
    })),
    cancelled_occurrences: meeting.cancelled_occurrences,
    scheduled_start_time: pastRow.scheduled_start_time,
    meeting_and_occurrence_id: pastRow.meeting_and_occurrence_id,
  };
}
```

Occurrences follow the `MeetingOccurrenceSummary` precedent from `PublicMeetingController` — timestamps and status only, per-occurrence titles dropped. The two past-meeting identity fields are kept because `v1_past_meeting` rows need them to render at the right time and link to the right occurrence.

Tests seed `password`/`passcode`/`host_key`/`zoom_config`/`organizers` on the fixture and assert (a) the exact allowlisted key set and (b) that no credential or PII value or key name appears in the serialized response.

See commit 23406535b.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
MRashad26: Done — both layers.

Server-side strip is in place (see the `getProjectMeetings` thread). For defense in depth, added `publicMeetingToCalendarEvents` to `meeting-calendar.utils.ts` — same event shape as `meetingToCalendarEvents`, but it takes the credential-free `PublicCalendarMeeting` projection and never sets `extendedProps.password`, so `resolveMeetingCalendarClickRoute` produces no `password` query param. The public calendar component now uses it.

`meeting-calendar.utils.spec.ts` asserts the click route the public mapper produces, not just the absent key — a future refactor pointing this mapper back at a full `Meeting` would fail the test rather than silently reintroduce the leak. Also covered: occurrence expansion using the series title, cancelled occurrences staying inert, and past-meeting rows linking through their composite resource id.

See commit 23406535b.

apps/lfx-one/src/app/app.routes.ts
MRashad26: Done — added exactly the rule you specified, beside the groups row:

```ts
// Public project calendar (LFXV2-3340) — month/week view of a project's public meetings, optionally
// scoped to one committee via \`?committee=\`. Same single-segment anchoring as the group directories
// above (no deeper route exists) so a nested path fails closed to \`required\`.
{ pattern: /^\/projects\/[^/]+\/calendar\/?$/, type: 'ssr', auth: 'optional' },
```

`auth.middleware.spec.ts` gains the anonymous-access test and the matching anchored no-fail-open test for `/projects/:slug/calendar/extra`.

See commit 23406535b.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
MRashad26: Done — added `distinctUntilChanged` on the mapped pair, so only slug/committee changes re-enter the `switchMap`:

```ts
map(([params, queryParams]) => ({ slug: params.get('projectSlug') ?? '', committeeUid: queryParams.get('committee') ?? undefined })),
distinctUntilChanged((a, b) => a.slug === b.slug && a.committeeUid === b.committeeUid),
```

A `?view=month|week` toggle no longer refetches or flashes the skeleton. The new component spec locks both halves: no extra call when only `?view` changes, and a second call when `?committee` does.

See commit 23406535b.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
MRashad26: Done — added `public-project-calendar.component.spec.ts` covering render (project name + meeting count + calendar container), the error state, committee scoping of the request, `?view=week` vs the month default, no-refetch on a pure view toggle, and refetch on a committee change.

One thing worth flagging from writing it: the sibling `public-project-groups` spec lists its `ActivatedRoute` override *before* `provideRouter([])`, so the router's own `ActivatedRoute` actually wins and the route mock is inert. That spec only asserts DOM, so it never noticed. This new spec puts the override last (with a comment) — without that, the committee-scoping assertion silently saw an empty `paramMap`. Happy to fix the groups spec in a follow-up if you'd like.

See commit 23406535b.

apps/lfx-one/src/server/controllers/project.controller.ts
MRashad26: Done — `res.setHeader('Cache-Control', 'public, max-age=300')` before `res.json(response)`, with a comment explaining the divergence from `calendar.ics`: calendar clients poll the ICS feed on their own 15–60 minute schedule, while this feed backs an interactive page where a newly scheduled meeting should surface reasonably quickly. Asserted in `project.controller.spec.ts`.

See commit 23406535b.

apps/lfx-one/src/server/controllers/project.controller.ts
MRashad26: Done — wrapped the `getProjectById` call in a `.catch` inside the `Promise.all`, matching the calname pattern in `getProjectCalendar`. A transient project-service failure now logs a warning and the response still returns the meetings with `project: { uid: projectUid, name: '' }`; the header just renders the bare "Calendar" title. Note the envelope's `uid` now comes from the already-resolved `projectUid` rather than the fetched project, so it stays correct on the fallback path. Covered by a new spec case.

See commit 23406535b.

packages/shared/src/interfaces/meeting.interface.ts
MRashad26: Done — introduced `PublicCalendarMeeting` and typed the envelope with it. Went with an explicit interface rather than `Omit<Meeting, ...>` so the contract is an allowlist: an `Omit` would keep silently widening every time a new sensitive field lands on `Meeting`, which is the same failure mode as the delete-based strip. The controller's projection builds exactly these fields, so the type and the wire format can't drift.

See commit 23406535b.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
MRashad26: Fixed — extracted `private initInitialView(): Signal<string>` alongside `initCalendarData()` under the "Private initializer functions" section, so both signals now follow the same pattern. Good catch.

See commit 23406535b.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
MRashad26: Fixed — now imports `FullCalendarComponent` from `@app/shared/components/fullcalendar/fullcalendar.component`, matching `meetings-dashboard.component.ts` and `committee-meetings.component.ts`.

See commit 23406535b.

packages/shared/src/interfaces/meeting.interface.ts
MRashad26: Done — both public calendar types now carry `@description` blocks and per-field JSDoc. `PublicCalendarMeeting`'s description states the allowlist-not-strip guarantee and names the excluded credential/PII fields explicitly, and `PublicProjectMeetingsResponse` documents the PUBLIC-only filter, the committee scoping, and that `project.name` is empty when the lookup fails.

See commit 23406535b.

MRashad26:
## Review Feedback Addressed

Commit: 23406535be521a45247df6358c3871cf2c7c5029

All 14 review threads addressed and resolved — 3 blocking, 5 minor, 3 nits from @dealako, plus the 2 Copilot findings and the Cursor Bugbot finding they reconcile with.

### Blocking

- **`server/controllers/project.controller.ts`** — `getProjectMeetings` no longer spreads the indexed record and deletes two keys. It now maps through a field-by-field allowlist (`toPublicCalendarMeeting`) into the new shared `PublicCalendarMeeting` type, so `password`, `passcode`, `host_key`, `zoom_config`, `organizers`, `created_by`, and `owner` are structurally absent from the anonymous payload — and stay absent as new `Meeting` fields land upstream. Occurrences are projected down to timestamps and status, following the `MeetingOccurrenceSummary` precedent in `PublicMeetingController`. _(per @dealako, @Copilot, Cursor Bugbot)_
- **`packages/shared/src/utils/meeting-calendar.utils.ts`** — added `publicMeetingToCalendarEvents`, a public-safe mapper that takes the credential-free projection and never writes a password into `extendedProps`, so `resolveMeetingCalendarClickRoute` cannot forward one into the join URL, browser history, or referrer. The public calendar page now uses it. _(per @dealako)_
- **`server/middleware/auth.middleware.ts`** — allowlisted `/^\/projects\/[^/]+\/calendar\/?$/` as `type: 'ssr', auth: 'optional'` beside the groups rule, so logged-out visitors reach SSR instead of a login redirect. _(per @dealako, @Copilot)_

### Minor

- **`public-project-calendar.component.ts`** — `distinctUntilChanged` on `{slug, committeeUid}`, so a `?view=month|week` toggle no longer refetches the feed or flashes the skeleton.
- **`public-project-calendar.component.spec.ts`** — new spec covering render, error state, committee scoping, initial view, no-refetch-on-view-change, and refetch-on-committee-change.
- **`project.controller.ts`** — `Cache-Control: public, max-age=300` on the JSON feed. Shorter than `calendar.ics`'s 15 minutes because this backs an interactive page rather than a client-polled subscription; the rationale is in a code comment.
- **`project.controller.ts`** — `getProjectById` failures now degrade gracefully (`project: { uid, name: '' }`) instead of failing the whole response, matching the ICS calname fallback. The envelope's `uid` comes from the already-resolved `projectUid`, so it stays correct on the fallback path.
- **`meeting.interface.ts`** — `PublicProjectMeetingsResponse.meetings` is now `PublicCalendarMeeting[]`. Written as an explicit interface rather than `Omit<Meeting, ...>` deliberately: an `Omit` would silently widen every time a new sensitive field lands on `Meeting`, which is the same failure mode as the delete-based strip.

### Nits

- Extracted `initInitialView()` to match the `initCalendarData()` initializer pattern.
- `FullCalendarComponent` now imported via `@app/shared/components/...`, matching `meetings-dashboard`.
- Full `@description` and per-field JSDoc on both public calendar types, documenting the PII/credential guarantee.

### Test coverage added

- `project.controller.spec.ts` — fixture seeds `password`/`passcode`/`host_key`/`zoom_config`/`organizers`; asserts the exact allowlisted key set, that no credential value or key name appears in the serialized response (serialized rather than key-checked, so a nested `zoom_config.passcode` can't slip through), the occurrence projection, `Cache-Control`, and the project-metadata fallback.
- `auth.middleware.spec.ts` — anonymous access to `/projects/:slug/calendar` plus the anchored no-fail-open case.
- `meeting-calendar.utils.spec.ts` — `publicMeetingToCalendarEvents`: no password in the produced click route, occurrence expansion with cancelled occurrences inert, and past-meeting composite-id linking.

`yarn format:check`, `yarn lint:check`, `yarn check-types`, `yarn test` (2128 server + 1299 Angular + 1234 shared), and `yarn build` all pass.

### One thing to flag

While writing the new component spec I found that the sibling `public-project-groups.component.spec.ts` lists its `ActivatedRoute` override *before* `provideRouter([])`, so the router's own `ActivatedRoute` wins and the route mock is inert. That spec only asserts DOM, so it passes regardless. The new spec puts the override last (with a comment explaining why) — without it the committee-scoping assertion silently saw an empty `paramMap`. Happy to fix the groups spec in a follow-up if you'd like it in scope.

packages/shared/src/utils/meeting-calendar.utils.ts
```diff
@@ -93,6 +93,76 @@ export function meetingToCalendarEvents(meeting: Meeting | PastMeeting): Meeting
...
+        borderColor: colors.border,
+        textColor: colors.text,
+        display: 'block',
+        classNames,
+        extendedProps: {
+          type: 'meeting',
+          meetingId: meeting.id,
+          cancelled: isCancelled,
+          startTime: occ.start_time,
+          durationMinutes: occ.duration,
```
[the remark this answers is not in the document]
MRashad26: Good catch — confirmed and fixed in both layers.

You're right on all three consequences. `addMinutesToDate` is `d.setMinutes(d.getMinutes() + (minutes ?? 60))`, so a missing duration silently becomes 60 minutes, and `isOccurrencePast` does `startMs + undefined * 60_000` → `NaN`, which makes `now > NaN` false so an ended occurrence never styles as past. `durationMinutes` also reached `resolveMeetingCalendarClickRoute` as `undefined`.

Restored the fallback in `publicMeetingToCalendarEvents`, mirroring the sibling mapper:

```ts
const occurrenceDuration = occ.duration ?? meeting.duration;
```

Also fixed it a layer earlier, in the server projection, so the wire payload honors `MeetingOccurrenceSummary`'s required `duration` for every consumer rather than relying on each mapper to defend:

```ts
duration: occurrence.duration ?? meeting.duration,
```

Regression coverage on both sides: `meeting-calendar.utils.spec.ts` seeds an occurrence with no duration override on a 90-minute series and asserts `durationMinutes` is 90, the computed end is exactly 90 minutes after the start, and an occurrence that ended three hours ago styles as past — all three would fail on the pre-fix code rather than just the first. `project.controller.spec.ts` asserts the projected occurrence carries the series duration.

See commit bf8baf20c.

dealako:
@MRashad26 — thanks for the thorough follow-up. Both commits (`23406535b`, `bf8baf20c`) land every blocking item from the prior round cleanly.

👏 **Nice work**
- **`toPublicCalendarMeeting` allowlist** replaces spread-and-delete with an explicit credential-free projection, backed by spec assertions that `password`/`passcode`/`host_key`/`zoom_config` never appear in serialized JSON.
- **`publicMeetingToCalendarEvents`** gives defense-in-depth on the client so join URLs cannot carry a password even if upstream data regresses.
- **Auth allowlist + spec** for `/projects/:slug/calendar` closes the anonymous SSR gap.
- **`distinctUntilChanged`** on slug/committee stops the view-toggle refetch, with a regression test proving `?view=week` alone does not re-hit the API.
- **`PublicCalendarMeeting` interface** with field-level JSDoc makes the anonymous contract explicit rather than implied.

**Revision tracking (prior round)**
- ✅ Credential leak on `getProjectMeetings` — `toPublicCalendarMeeting` allowlist + spec (`23406535b`)
- ✅ Client password propagation — `publicMeetingToCalendarEvents` + utils spec (`23406535b`)
- ✅ SSR auth allowlist — `auth.middleware.ts:58-61` + spec (`23406535b`)
- ✅ `?view` refetch — `distinctUntilChanged` + spec (`23406535b`)
- ✅ Component spec — new `public-project-calendar.component.spec.ts` (`23406535b`)
- ✅ `Cache-Control` — `max-age=300` with documented rationale (`23406535b`)
- ✅ Project metadata graceful fallback — `.catch` in `Promise.all` (`23406535b`)
- ✅ Shared response type accuracy — `PublicCalendarMeeting[]` (`23406535b`)
- ✅ `toSignal` init pattern, import alias, JSDoc — all fixed (`23406535b`)
- ✅ Occurrence duration fallback — Copilot follow-up addressed in `bf8baf20c`

**Structured issue count (remaining)**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: architecture docs still omit the new public calendar surface; component spec missing click-navigation and empty-header cases
- ⚪ Nit: 3 issues: runtime `?view=` URL changes don't sync the mounted FullCalendar view; mixed comment style; protected-signal double-cast in spec

**Bot reconciliation**
- **Agree** with Copilot's duration-fallback note — confirmed fixed in `bf8baf20c`.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+          useValue: { paramMap: of(new Map([['projectSlug', slug]])), queryParamMap: queryParamMap.asObservable() },
+        },
+      ],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(PublicProjectCalendarComponent);
+    await fixture.whenStable();
+  }
+
+  it('renders the project name and meeting count once the feed resolves', async () => {
```
dealako, line 78: [minor] PublicProjectCalendarComponent omits click-navigation and degraded-header tests

Issue: Spec covers fetch, error, committee scoping, and view-refetch behavior, but not `onCalendarEventClick` routing or the empty `project.name` header fallback.

Proof: Six tests (lines 78–131) with no `Router.navigate` assertion and no render case for `project: { uid: 'p1', name: '' }`. Server degradation is tested in `project.controller.spec.ts` only.

Why it matters: Regressions in join navigation or header fallback would ship without unit signal on this anonymous page.

Fix: Add Router mock tests for `resolveMeetingCalendarClickRoute` outcomes (including cancelled inert events) and a spec that `project.name === ''` renders the "Calendar" fallback title.

apps/lfx-one/src/app/modules/meetings/public-project-calendar/public-project-calendar.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    const route = resolveMeetingCalendarClickRoute(arg.event.extendedProps as MeetingCalendarClickProps, arg.event.start);
+    if (!route) {
+      return;
+    }
+    void this.router.navigate(route.path, route.queryParams ? { queryParams: route.queryParams } : undefined);
+  }
+
+  // Private initializer functions
+
+  private initInitialView(): Signal<string> {
```
dealako, line 54: [nit] Runtime ?view= changes update initialView but not the mounted FullCalendar

Issue: After the `distinctUntilChanged` fix, a client-side router navigation that toggles `?view=week|month` updates the `initialView` signal without refetching, but `lfx-fullcalendar` only reads `initialView` at mount time in `calendarOptions`.

Proof: `fullcalendar.component.ts:28` sets `initialView: this.initialView()` inside a getter with no `changeView` hook on input changes.

Why it matters: Back/forward or in-app navigation that only mutates `?view=` can leave the visible calendar on the prior view while the URL claims otherwise. Deep-link first paint and the FullCalendar Month/Week toolbar both work today.

Fix: On `?view=` changes call `CalendarApi.changeView()`, remount `lfx-fullcalendar` keyed by `initialView()`, or document that `?view=` is first-paint-only if runtime sync is out of scope.

dealako (approved):
## Follow-up review

@MRashad26 — all three blockers from the prior round are resolved. The allowlist projection, public calendar mapper, auth allowlist, and test coverage are solid.

### ✅ Resolved (prior round)

| Item | Commit |
|------|--------|
| Credential leak on anonymous API | `23406535b` — `toPublicCalendarMeeting` allowlist + spec |
| Client password in join URLs | `23406535b` — `publicMeetingToCalendarEvents` |
| SSR auth allowlist | `23406535b` — `auth.middleware.ts` + spec |
| `?view` refetch | `23406535b` — `distinctUntilChanged` + spec |
| Component spec | `23406535b` |
| Cache-Control | `23406535b` — `max-age=300` |
| Project metadata fallback | `23406535b` |
| Shared type accuracy + JSDoc + style nits | `23406535b` |
| Occurrence duration fallback | `bf8baf20c` |

### 🟡 Minor (2, non-blocking)

- Architecture docs still omit `/projects/:projectSlug/calendar`, `GET /public/api/projects/:id/meetings`, and `PublicCalendarMeeting`
- Component spec could cover click navigation and empty project-name header

### ⚪ Nit (3)

- Runtime `?view=` query changes update `initialView` but not the mounted FullCalendar (deep-link first paint works; FullCalendar toolbar covers in-page toggles)
- Mixed `//` vs `/** */` comment style on adjacent members
- Spec accesses protected `initialView` via opaque double cast

PR #1927, opened 2026-08-28
fix(events): scope combined cte dedup to the full union

### LINKED ISSUE #1926: Travel Funding Application step 1 shows the same event twice
### What happened?

In the Travel Funding Application flow, step 1 ("Choose an Event") renders the same
event twice as two identical, separately-selectable cards.

The duplication comes from the API, not the UI — `GET /api/events` returns two
byte-identical objects and a `total` of 2.

**Expected:** each event appears exactly once.

### Command / Request

```
GET /api/events?isPast=false&pageSize=12&offset=0&registeredOnly=true&isTravelFundRequestAccepted=true&excludePastTravelFundDeadline=true
```

### Relevant log output

```
{
  "data": [
    { "id": "<event-id>", "name": "<event name>", ... },
    { "id": "<event-id>", "name": "<event name>", ... }
  ],
  "total": 2,
  "pageSize": 12,
  "offset": 0
}
```

Both entries are identical across every field.

### Root cause

`apps/lfx-one/src/server/services/events.service.ts:189-193` — the `combined` CTE:

```sql
combined AS (
  SELECT *, 1 AS source_priority FROM registered_events
  UNION ALL
  SELECT *, 2 AS source_priority FROM affiliated_upcoming
  QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority) = 1
)
```

`QUALIFY` is a clause of a single SELECT block, not of the set operation, so it binds
only to the second branch (`affiliated_upcoming`) — the cross-branch dedup this CTE
exists to perform never runs. (`ORDER BY`/`LIMIT` would apply to the whole union;
`QUALIFY`/`WHERE`/`GROUP BY` do not.) Inside `affiliated_upcoming` the predicate is
trivially true, since that branch is already deduped by its own `QUALIFY` (line 144),
so every row survives.

Confirmed with a data-independent reproduction — returns 2 rows:

```sql
WITH a AS (SELECT 'X' AS EVENT_ID),
     b AS (SELECT 'X' AS EVENT_ID),
     combined AS (
       SELECT *, 1 AS source_priority FROM a
       UNION ALL
       SELECT *, 2 AS source_priority FROM b
       QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority) = 1
     )
SELECT * FROM combined;
```

### Who it affects

An event duplicates only when it lands in **both** branches — the user is registered
for it *and* affiliated with its project. Most users match a single branch, so it
presents as intermittent rather than universal.

`COUNT(*) OVER()` (line 225) is computed after the fan-out, so paging totals inflate too.

### Ruled out

Source data (exactly one accepted registration per user/event; no whitespace or case
variants in the join keys), the upstream dbt model and its atomic `table`
materialization, HTTP/CDN caching (the observed `304` is Express's default ETag, which
re-runs the query on every request), and the full BFF → Angular chain.

### Notes

`events.service.ts:193` is the only `QUALIFY` trailing a `UNION ALL` in the server.

Two unrelated pre-existing gaps found nearby, **not** fixed here and not the cause —
the source data is clean and the upstream model enforces one row per user/event pair:
the `user_reg` CTE (lines 170-188) and the Past-tab query (lines ~264-320) both lack the
dedup their sibling CTEs have, so either would fan out through its join if upstream data
ever did hold duplicates. Worth filing as follow-ups if we want defense-in-depth.

fayazg:
## Summary

- Fixes duplicate event cards in step 1 ("Choose an Event") of the Travel Funding Application flow
- `QUALIFY` placed directly after `UNION ALL` in the `combined` CTE (`events.service.ts`) binds to the last `SELECT` block only, so the cross-branch dedup that CTE exists to perform never ran — an event matching both the registered and affiliated branches was returned twice
- Wraps the union in a derived table so `QUALIFY` applies to the combined result
- `COUNT(*) OVER()` is computed after the join, so inflated paging totals are corrected by the same fix
- Adds a regression test asserting the dedup is scoped to the whole union, not a single branch

`events.service.ts` was the only place in the server where a `QUALIFY` trailed a `UNION ALL`.

Fixes #1926

### DIFF
--- a/apps/lfx-one/src/server/services/events.service.spec.ts
+++ b/apps/lfx-one/src/server/services/events.service.spec.ts
@@ -174,11 +174,11 @@ describe('EventsService.getMyEvents past-event SQL', () => {
     const sql = await sqlFor({ isPast: false, affiliatedProjectSlugs: ['example'] });
     const combined = sql.slice(sql.indexOf('combined AS ('));
     const wrapperOpen = combined.indexOf('SELECT * FROM (');
-    // Each branch landmark pins its priority literal to its source, so swapping them fails here;
-    // the QUALIFY landmark closes at ")", so an ORDER BY flipped to DESC stops matching too.
+    // Each branch landmark pins its priority literal to its source; the QUALIFY landmark spans the
+    // ORDER BY through "= 1", so a DESC flip or a changed row-number predicate stops matching too.
     const firstBranch = combined.indexOf('SELECT *, 1 AS SOURCE_PRIORITY FROM registered_events');
     const lastBranch = combined.indexOf('SELECT *, 2 AS SOURCE_PRIORITY FROM affiliated_upcoming');
-    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY SOURCE_PRIORITY)');
+    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY SOURCE_PRIORITY) = 1');
 
     // Ordering alone passes on a missing landmark (indexOf returns -1), so require each to exist.
     for (const landmark of [wrapperOpen, firstBranch, lastBranch, qualify]) {
--- a/apps/lfx-one/src/server/services/events.service.spec.ts
+++ b/apps/lfx-one/src/server/services/events.service.spec.ts
@@ -168,6 +168,17 @@ describe('EventsService.getMyEvents past-event SQL', () => {
     expect(sql).toContain('e.EVENT_SOURCE,');
   });
 
+  it('scopes the combined-CTE dedup to the whole union, not just the last branch', async () => {
+    // QUALIFY binds to a single SELECT block, so the union has to be wrapped: written directly
+    // after UNION ALL it only dedups affiliated_upcoming and duplicates cross-branch events.
+    const sql = await sqlFor({ isPast: false, affiliatedProjectSlugs: ['example'] });
+    const combined = sql.slice(sql.indexOf('combined AS ('));
+    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority)');
+
+    expect(combined.indexOf('SELECT * FROM (')).toBeLessThan(combined.indexOf('SELECT *, 1 AS source_priority'));
+    expect(combined.slice(combined.indexOf('FROM affiliated_upcoming'), qualify)).toContain(')');
+  });
+
   it('selects EVENT_SOURCE in the past branch', async () => {
     const sql = await sqlFor({ isPast: true });
 
--- a/apps/lfx-one/src/server/services/events.service.spec.ts
+++ b/apps/lfx-one/src/server/services/events.service.spec.ts
@@ -174,9 +174,11 @@ describe('EventsService.getMyEvents past-event SQL', () => {
     const sql = await sqlFor({ isPast: false, affiliatedProjectSlugs: ['example'] });
     const combined = sql.slice(sql.indexOf('combined AS ('));
     const wrapperOpen = combined.indexOf('SELECT * FROM (');
-    const firstBranch = combined.indexOf('SELECT *, 1 AS source_priority');
-    const lastBranch = combined.indexOf('FROM affiliated_upcoming');
-    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority)');
+    // Each branch landmark pins its priority literal to its source, so swapping them fails here;
+    // the QUALIFY landmark closes at ")", so an ORDER BY flipped to DESC stops matching too.
+    const firstBranch = combined.indexOf('SELECT *, 1 AS SOURCE_PRIORITY FROM registered_events');
+    const lastBranch = combined.indexOf('SELECT *, 2 AS SOURCE_PRIORITY FROM affiliated_upcoming');
+    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY SOURCE_PRIORITY)');
 
     // Ordering alone passes on a missing landmark (indexOf returns -1), so require each to exist.
     for (const landmark of [wrapperOpen, firstBranch, lastBranch, qualify]) {
--- a/apps/lfx-one/src/server/services/events.service.spec.ts
+++ b/apps/lfx-one/src/server/services/events.service.spec.ts
@@ -173,10 +173,17 @@ describe('EventsService.getMyEvents past-event SQL', () => {
     // after UNION ALL it only dedups affiliated_upcoming and duplicates cross-branch events.
     const sql = await sqlFor({ isPast: false, affiliatedProjectSlugs: ['example'] });
     const combined = sql.slice(sql.indexOf('combined AS ('));
+    const wrapperOpen = combined.indexOf('SELECT * FROM (');
+    const firstBranch = combined.indexOf('SELECT *, 1 AS source_priority');
+    const lastBranch = combined.indexOf('FROM affiliated_upcoming');
     const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority)');
 
-    expect(combined.indexOf('SELECT * FROM (')).toBeLessThan(combined.indexOf('SELECT *, 1 AS source_priority'));
-    expect(combined.slice(combined.indexOf('FROM affiliated_upcoming'), qualify)).toContain(')');
+    // Ordering alone passes on a missing landmark (indexOf returns -1), so require each to exist.
+    for (const landmark of [wrapperOpen, firstBranch, lastBranch, qualify]) {
+      expect(landmark).toBeGreaterThanOrEqual(0);
+    }
+    expect(wrapperOpen).toBeLessThan(firstBranch);
+    expect(combined.slice(lastBranch, qualify)).toContain(')');
   });
 
   it('selects EVENT_SOURCE in the past branch', async () => {
--- a/apps/lfx-one/src/server/services/events.service.ts
+++ b/apps/lfx-one/src/server/services/events.service.ts
@@ -190,11 +190,11 @@ export class EventsService {
           -- The subquery is load-bearing: QUALIFY written directly after UNION ALL binds to
           -- the last SELECT block only, so the cross-branch dedup would never run (#1926).
           SELECT * FROM (
-            SELECT *, 1 AS source_priority FROM registered_events
+            SELECT *, 1 AS SOURCE_PRIORITY FROM registered_events
             UNION ALL
-            SELECT *, 2 AS source_priority FROM affiliated_upcoming
+            SELECT *, 2 AS SOURCE_PRIORITY FROM affiliated_upcoming
           )
-          QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority) = 1
+          QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY SOURCE_PRIORITY) = 1
         )
         SELECT
           e.EVENT_ID,
--- a/apps/lfx-one/src/server/services/events.service.ts
+++ b/apps/lfx-one/src/server/services/events.service.ts
@@ -187,9 +187,13 @@ export class EventsService {
             AND REGISTRATION_STATUS = 'Accepted'
         ),
         combined AS (
-          SELECT *, 1 AS source_priority FROM registered_events
-          UNION ALL
-          SELECT *, 2 AS source_priority FROM affiliated_upcoming
+          -- The subquery is load-bearing: QUALIFY written directly after UNION ALL binds to
+          -- the last SELECT block only, so the cross-branch dedup would never run (#1926).
+          SELECT * FROM (
+            SELECT *, 1 AS source_priority FROM registered_events
+            UNION ALL
+            SELECT *, 2 AS source_priority FROM affiliated_upcoming
+          )
           QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority) = 1
         )
         SELECT


apps/lfx-one/src/server/services/events.service.spec.ts
```diff
@@ -168,6 +168,17 @@ describe('EventsService.getMyEvents past-event SQL', () => {
...
 
+  it('scopes the combined-CTE dedup to the whole union, not just the last branch', async () => {
+    // QUALIFY binds to a single SELECT block, so the union has to be wrapped: written directly
+    // after UNION ALL it only dedups affiliated_upcoming and duplicates cross-branch events.
+    const sql = await sqlFor({ isPast: false, affiliatedProjectSlugs: ['example'] });
+    const combined = sql.slice(sql.indexOf('combined AS ('));
+    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority)');
+
+    expect(combined.indexOf('SELECT * FROM (')).toBeLessThan(combined.indexOf('SELECT *, 1 AS source_priority'));
+    expect(combined.slice(combined.indexOf('FROM affiliated_upcoming'), qualify)).toContain(')');
```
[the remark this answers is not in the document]
fayazg: Good catch — verified this exactly as described, and it was a real hole.

I mutated the production SQL two ways and ran the test against each:

- **Un-wrapped union** (`QUALIFY` back to trailing `UNION ALL`, i.e. the original bug): the test **failed** — the `slice(lastBranch, qualify)` assertion caught it.
- **`QUALIFY` deleted entirely** (dedup gone): the test **passed**, because `qualify` was `-1` and `slice(start, -1)` ran to the end of the SQL, which trivially contains `)`.

So the test caught the specific shape this PR fixes but not the loss of deduplication in general, which is what you flagged.

Fixed in 6a4f5b4f6: every landmark index is now asserted `>= 0` before being used as a slice offset, so a missing `QUALIFY` (or missing wrapper, first branch, or last branch) fails rather than silently degrading to `-1`. Re-ran the `QUALIFY`-deleted mutation against the hardened test and it now fails with `expected -1 to be greater than or equal to 0`.

fayazg:
## Review Feedback Addressed

Commit: 6a4f5b4f6

### Changes Made

- **`apps/lfx-one/src/server/services/events.service.spec.ts`**: hardened the combined-CTE regression test so it fails when the dedup is removed, not just when the union is un-wrapped (per copilot-pull-request-reviewer).

The original assertions used `indexOf` results directly as slice offsets. With `QUALIFY` deleted, `indexOf` returned `-1` and `slice(start, -1)` ran to the end of the SQL — which trivially contains `)`, so the assertion held even with deduplication gone. Each landmark index is now required to be `>= 0` before it is used as an offset.

Verified by mutating the production SQL and running the test against each shape:

| Mutation | Old test | Hardened test |
| --- | --- | --- |
| Un-wrapped union (the bug this PR fixes) | fails | fails |
| `QUALIFY` deleted entirely (dedup gone) | **passes** | fails |

### Threads Resolved

apps/lfx-one/src/server/services/events.service.spec.ts
```diff
@@ -168,6 +168,24 @@ describe('EventsService.getMyEvents past-event SQL', () => {
...
+    const firstBranch = combined.indexOf('SELECT *, 1 AS source_priority');
+    const lastBranch = combined.indexOf('FROM affiliated_upcoming');
+    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY source_priority)');
+
+    // Ordering alone passes on a missing landmark (indexOf returns -1), so require each to exist.
+    for (const landmark of [wrapperOpen, firstBranch, lastBranch, qualify]) {
+      expect(landmark).toBeGreaterThanOrEqual(0);
+    }
+    expect(wrapperOpen).toBeLessThan(firstBranch);
+    expect(combined.slice(lastBranch, qualify)).toContain(')');
```
dealako, line 188: [nit] Test asserts SQL structure but not the dedup priority semantics

Issue: This test validates the derived-table wrapper positions (`wrapperOpen < firstBranch`, closing paren between the last branch and `QUALIFY`) but never asserts which source wins the tie-break. A silent swap of the priority literals (registered `1`→`2`, affiliated `2`→`1`) or a flipped `ORDER BY` direction would still pass every assertion, even though it changes which branch's row survives per `EVENT_ID`.
Proof: lines 176-186 match `SELECT *, 1 AS source_priority` and `FROM affiliated_upcoming` as independent landmarks but never tie priority `1` to `registered_events` / priority `2` to `affiliated_upcoming`, nor assert `ORDER BY source_priority` is ascending. `events.service.ts:193-197` encodes registered=1 (wins), affiliated=2.
Why it matters: the whole point of `source_priority` is to prefer the user's own registered row over the affiliation-derived row; a regression in that mapping is exactly what this suite exists to catch, but it would slip through green.
Fix: assert the priority literal is bound to its source verbatim, e.g. expect `SELECT *, 1 AS source_priority FROM registered_events` and `SELECT *, 2 AS source_priority FROM affiliated_upcoming` to be present, plus assert the `ORDER BY` has no `DESC`.

apps/lfx-one/src/server/services/events.service.ts
```diff
@@ -187,9 +187,13 @@ export class EventsService {
             AND REGISTRATION_STATUS = 'Accepted'
         ),
         combined AS (
-          SELECT *, 1 AS source_priority FROM registered_events
-          UNION ALL
-          SELECT *, 2 AS source_priority FROM affiliated_upcoming
+          -- The subquery is load-bearing: QUALIFY written directly after UNION ALL binds to
+          -- the last SELECT block only, so the cross-branch dedup would never run (#1926).
+          SELECT * FROM (
+            SELECT *, 1 AS source_priority FROM registered_events
```
dealako: [nit] `source_priority` alias diverges from the file's UPPERCASE column-alias convention

Issue: the new ordering column is aliased in lowercase snake_case (`1 AS source_priority` / `2 AS source_priority`), whereas every other column alias in this file's SQL is UPPERCASE.
Proof: lines 193, 195, 197 use `source_priority`. Other aliases in the same file are UPPERCASE, e.g. `COUNT(*) OVER() AS TOTAL_RECORDS` (line 229), `AS IS_REGISTERED` (line 223), `FALSE AS IS_PAST_EVENT` (line 227). Lowercase in this file is otherwise reserved for CTE names (`combined`, `user_reg`, `registered_events`, `affiliated_upcoming`), so a lowercase column alias reads like a CTE reference.
Why it matters: cosmetic only — Snowflake folds unquoted identifiers to uppercase — but it diverges from the file's own established style.
Fix: rename to `SOURCE_PRIORITY` in both branch SELECTs and the `QUALIFY ... ORDER BY` to match the surrounding convention.

dealako (dismissed):
Hi @fayazg 👋 — clean, well-targeted fix. Wrapping the `UNION ALL` in a derived table so the trailing `QUALIFY` binds to the whole union (not just the `affiliated_upcoming` branch) is exactly the right correction for #1926, and it fixes the inflated `COUNT(*) OVER()` paging total for free. The load-bearing comment and the regression test both add real value. I traced the change through the full `getMyEvents` SQL and the bind ordering, and confirmed the derived-table wrapper adds no placeholders, so bind alignment is untouched.

I verified your response to Copilot: the hardened test (asserting every landmark `>= 0` before using it as a slice offset) does close the hole where a deleted `QUALIFY` could pass falsely via `slice(start, -1)`. Nice follow-up.

**Issue count**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 — (1) test asserts SQL structure but not the dedup priority mapping; (2) `source_priority` lowercase alias diverges from the file's UPPERCASE column-alias convention
- ❔ Question: 0

Both are take-it-or-leave-it. Snowflake folds the alias to uppercase anyway, and the priority-mapping gap is a hardening suggestion, not a defect in this diff.

**Decision: ✅ Approved with minor comments**

dealako:
Hi @fayazg 👋 — clean, well-targeted fix.

**Overall:** Wrapping the `UNION ALL` in a derived table so the trailing `QUALIFY` binds to the whole union (not just the `affiliated_upcoming` branch) is exactly the right correction for #1926, and it fixes the inflated `COUNT(*) OVER()` paging total for free. I traced the change through the full `getMyEvents` SQL and confirmed the derived-table wrapper adds no placeholders, so bind ordering is untouched. The load-bearing comment and the regression test both add real value.

I also verified your Copilot follow-up: the hardened test (asserting every landmark `>= 0` before using it as a slice offset) does close the `slice(start, -1)` false-pass hole. Good catch, well handled.

**Issue count**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2
  - Test asserts SQL structure but not the dedup priority mapping (a swap of the `1`/`2` literals would pass green) — `events.service.spec.ts`
  - `source_priority` lowercase alias diverges from the file's UPPERCASE column-alias convention (cosmetic; Snowflake folds it anyway) — `events.service.ts`
- ❔ Question: 0

Both nits are take-it-or-leave-it. Inline comments have the details.

**Decision: ✅ Approved with minor comments**

apps/lfx-one/src/server/services/events.service.spec.ts
fayazg: Good catch — confirmed and fixed in f638f1a19.

You're right that the landmarks were independent: `SELECT *, 1 AS source_priority` and `FROM affiliated_upcoming` were matched separately, so swapping the literals left every assertion satisfied.

Each branch landmark now matches the priority literal and its source together:

```ts
const firstBranch = combined.indexOf('SELECT *, 1 AS SOURCE_PRIORITY FROM registered_events');
const lastBranch = combined.indexOf('SELECT *, 2 AS SOURCE_PRIORITY FROM affiliated_upcoming');
```

On the `ORDER BY` direction — the `QUALIFY` landmark string closes at `)`, so `ORDER BY SOURCE_PRIORITY DESC)` stops matching and the landmark falls to `-1`, which the existing `>= 0` guard rejects. I noted that in a comment above the landmarks so it isn't accidentally loosened later.

Verified both by mutating the production SQL and re-running:

| Mutation | Result |
| --- | --- |
| Swap priorities (registered `1`→`2`, affiliated `2`→`1`) | fails |
| `ORDER BY SOURCE_PRIORITY DESC` | fails |
| Unmodified | passes |

apps/lfx-one/src/server/services/events.service.ts
fayazg: Agreed — renamed to `SOURCE_PRIORITY` in f638f1a19 (both branch SELECTs and the `QUALIFY ... ORDER BY`).

Confirmed the convention before changing it: every other column alias in this file is UPPERCASE (`IS_REGISTERED`, `IS_PAST_EVENT`, `TOTAL_RECORDS`, `EVENT_URL`, `ATTENDEES`, `APPLICATION_DATE`, `REQUEST_STATUS`), and lowercase is used only for CTE names (`combined`, `user_reg`, `registered_events`, `affiliated_upcoming`). `source_priority` was the file's only lowercase column alias, so it did read like a CTE reference.

Cosmetic as you noted — Snowflake folds the unquoted identifier either way — and the production diff is a pure rename with no structural change.

fayazg:
## Review Feedback Addressed

Commit: f638f1a19

Both nits from @dealako were small and self-contained, so they're handled here rather than deferred.

### Changes Made

- **`events.service.ts`**: renamed `source_priority` → `SOURCE_PRIORITY` in both branch SELECTs and the `QUALIFY ... ORDER BY`, matching the file's UPPERCASE column-alias convention (per @dealako). The production diff is a pure rename — no structural change.
- **`events.service.spec.ts`**: each branch landmark now binds the priority literal to its source (`SELECT *, 1 AS SOURCE_PRIORITY FROM registered_events`), so a swap of the priority mapping fails the test (per @dealako).

### Verification

Mutated the production SQL and re-ran the regression test against each shape:

| Mutation | Result |
| --- | --- |
| Swap priorities (registered `1`→`2`, affiliated `2`→`1`) | fails |
| `ORDER BY SOURCE_PRIORITY DESC` | fails |
| `QUALIFY` deleted entirely | fails |
| Un-wrapped union (the original bug) | fails |
| Unmodified | passes |

The `DESC` case is covered because the `QUALIFY` landmark string closes at `)`; that's noted in a comment above the landmarks so it isn't loosened later.

### Threads Resolved

apps/lfx-one/src/server/services/events.service.spec.ts
```diff
@@ -168,6 +168,26 @@ describe('EventsService.getMyEvents past-event SQL', () => {
...
+    // QUALIFY binds to a single SELECT block, so the union has to be wrapped: written directly
+    // after UNION ALL it only dedups affiliated_upcoming and duplicates cross-branch events.
+    const sql = await sqlFor({ isPast: false, affiliatedProjectSlugs: ['example'] });
+    const combined = sql.slice(sql.indexOf('combined AS ('));
+    const wrapperOpen = combined.indexOf('SELECT * FROM (');
+    // Each branch landmark pins its priority literal to its source, so swapping them fails here;
+    // the QUALIFY landmark closes at ")", so an ORDER BY flipped to DESC stops matching too.
+    const firstBranch = combined.indexOf('SELECT *, 1 AS SOURCE_PRIORITY FROM registered_events');
+    const lastBranch = combined.indexOf('SELECT *, 2 AS SOURCE_PRIORITY FROM affiliated_upcoming');
+    const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY SOURCE_PRIORITY)');
```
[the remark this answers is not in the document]
fayazg: Confirmed and fixed in b6c4b0290.

You're right — the landmark stopped at the closing paren, so `= 1` sat outside the match and `= 2` still satisfied it. And the consequence is worse than a cosmetic gap: at `= 2` the CTE keeps only the second row per `EVENT_ID`, so every event present in just one branch is dropped entirely.

The landmark now spans the predicate:

```ts
const qualify = combined.indexOf('QUALIFY ROW_NUMBER() OVER (PARTITION BY EVENT_ID ORDER BY SOURCE_PRIORITY) = 1');
```

`qualify` is still the match start, so its use as the `slice` endpoint is unchanged.

Verified by mutating the production SQL to `= 2` and re-running: the test fails with `expected -1 to be greater than or equal to 0`, and passes unmodified.

fayazg:
## Review Feedback Addressed

Commit: b6c4b0290

### Changes Made

- **`events.service.spec.ts`**: extended the `QUALIFY` landmark through `= 1` so a changed row-number predicate fails the regression test (per copilot-pull-request-reviewer). The landmark previously stopped at the closing paren, leaving the predicate outside the match.

Test-only change — the production SQL is byte-identical to the previous commit.

### Verification

The regression test now fails on every way the dedup can break, and passes unmodified:

| Mutation | Result |
| --- | --- |
| `QUALIFY ... = 2` (drops single-branch events) | fails |
| Swap priorities (registered `1`→`2`, affiliated `2`→`1`) | fails |
| `ORDER BY SOURCE_PRIORITY DESC` | fails |
| `QUALIFY` deleted entirely | fails |
| Un-wrapped union (the original bug) | fails |
| Unmodified | passes |

### Threads Resolved

dealako:
Thanks for the quick turnaround, @fayazg 👏 — both nits from my last round and Copilot's follow-up are cleanly addressed.

**👏 Nice work this revision:**
- Renamed `source_priority` → `SOURCE_PRIORITY` across both branch SELECTs and the `QUALIFY ... ORDER BY` (`f638f1a19`), matching the file's UPPERCASE column-alias convention.
- Pinned each priority literal to its source in the regression test — `SELECT *, 1 AS SOURCE_PRIORITY FROM registered_events` and `SELECT *, 2 AS SOURCE_PRIORITY FROM affiliated_upcoming` are now single landmarks, so a swap of the `1`/`2` mapping breaks the test (`f638f1a19`).
- Extended the `QUALIFY` landmark through `= 1` (`b6c4b0290`), closing Copilot's catch that a `= 2` predicate would drop single-branch events yet still pass. A `= 2` or a `DESC` flip now fails the match.

**Revision tracking**
- ✅ Test asserts dedup priority semantics — resolved in `f638f1a19`
- ✅ `SOURCE_PRIORITY` alias convention — resolved in `f638f1a19`
- ✅ [Copilot] `QUALIFY` predicate pinned to `= 1` — resolved in `b6c4b0290`

**Issue count (carried over + new)**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

No new findings. The rename is a static SQL literal with no bearing on the parameterized bind flow, so no correctness, security, or privacy impact.

**Decision: ✅ Approved**

dealako (approved):
Follow-up review — all prior feedback resolved.

- ✅ Test now pins each priority literal to its source (`f638f1a19`), so a `1`/`2` swap fails.
- ✅ `SOURCE_PRIORITY` alias uppercased to match the file convention (`f638f1a19`).
- ✅ `QUALIFY` landmark extended through `= 1` (`b6c4b0290`), closing Copilot's `= 2` catch.

No new findings. The rename is a static SQL literal with no bearing on the parameterized bind flow. 🔴 0 / 🟡 0 / ⚪ 0.

**Decision: ✅ Approved**

PR #1930, opened 2026-08-28
perf(projects): reduce access-check demand on affiliations page

### LINKED ISSUE #1901: Reduce unnecessary /api/projects access-check demand
## Problem

Exploration for #1734/#1862 found two avoidable sources of `/api/projects` access-check load in the BFF, independent of any FGA/query-service work:

1. **`profile-affiliations.component.ts` runs a wasted access check.** It calls the unscoped `getProjects()` (`:422`) but only consumes `p.slug` (`:432`, `new Set(lfxProjects.map(p => p.slug))`). The server-side `getProjects` path issues one `addAccessToResources` access-check tuple per project regardless of what the caller needs, so this component pays for an access check on every project it fetches without ever using the result.
2. **`/api/projects`'s access-check call is unchunked.** `access-check.service.ts:105-127` sends one request body carrying N tuples (one per project on the page) with no batching/chunking, so page size directly sets access-check payload size and (per the trace evidence in #1862) OpenFGA `batch-check` latency.

## Scope

- Drop (or scope down) the access-check call in `profile-affiliations.component.ts` so it only fetches what it needs (slugs), not full access-checked project rows.
- Add chunking to the access-check call in `access-check.service.ts` so page size doesn't translate 1:1 into unchunked batch-check payload size.

## Out of scope

- The `WriterGrantsService` idle-callback sweep (`writer-grants.service.ts:97-133`) is a related, larger demand-reduction target but is *gated* on a cheaper authoritative inherited-access answer existing first (see #1734 discussion); it is intentionally not part of this ticket.
- No FGA model, fga-sync, or query-service changes — this is BFF-only, zero migration risk.

bramwelt:
## Summary

Closes #1901

Two independent optimizations to reduce unnecessary `/api/projects` access-check (OpenFGA) load:

### 1. `GET /api/projects/slugs` — skip the OpenFGA round trip for affiliations

`ProfileAffiliationsComponent` fetched the full access-checked project list (`GET /api/projects`) solely to build a `Set<string>` of slugs for matching — the `writer` field was discarded. This PR adds a lightweight `GET /api/projects/slugs` endpoint that returns `string[]` directly, with no `AccessCheckService` call. The affiliations component now calls `getProjectSlugs()` instead of `getProjects()`.

**Before:** `GET /api/projects` → query service + OpenFGA for every project → strip everything except slug  
**After:** `GET /api/projects/slugs` → query service only → return slugs

### 2. Access-check batching in `AccessCheckService.performCheck`

`performCheck` previously sent all tuples in a single unbounded POST to the access-check service. When a caller passed a large project list the payload grew proportionally, increasing OpenFGA latency for every request on that code path.

`performCheck` now chunks the resource array into batches of `ACCESS_CHECK_BATCH_SIZE = 100` and fans out with `Promise.allSettled`. Fulfilled chunks are merged; rejected chunks fail closed (their keys are absent from the result map, defaulting to `false` at lookup) and are logged at WARN. A new `checkAccessStrict` note documents the batch-size ceiling on its error-propagation guarantee.

## Files changed

| File | Change |
|---|---|
| `packages/shared/src/constants/api.constants.ts` | Add `ACCESS_CHECK_BATCH_SIZE = 100` |
| `apps/lfx-one/src/server/services/access-check.service.ts` | Batch `performCheck`; extract `performSingleCheck`; add `checkAccessStrict` caveat |
| `apps/lfx-one/src/server/services/access-check.service.spec.ts` | Tests for tuple-order independence, batching, partial failure, and strict propagation |
| `apps/lfx-one/src/server/services/project.service.ts` | Add `getProjectSlugs` backend service method |
| `apps/lfx-one/src/server/services/project.service.spec.ts` | Tests for `getProjectSlugs` (pagination, ROOT filtering, no access check) |
| `apps/lfx-one/src/server/controllers/project.controller.ts` | Add `getProjectSlugs` controller handler |
| `apps/lfx-one/src/server/controllers/project.controller.spec.ts` | Tests for `getProjectSlugs` controller |
| `apps/lfx-one/src/server/routes/projects.route.ts` | Add `GET /slugs` before `/:slug` param route |
| `apps/lfx-one/src/app/shared/services/project.service.ts` | Add `getProjectSlugs()` frontend service method with caching |
| `apps/lfx-one/src/app/shared/services/project.service.spec.ts` | Frontend service tests |
| `apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts` | Switch from `getProjects()` to `getProjectSlugs()` |

## Trade-offs documented

- **`console.error` in frontend `ProjectService.getProjectSlugs`**: follows the established file-wide pattern (all other `catchError` blocks in this service use `console.error`). Fixing in isolation would be inconsistent; addressed together in a follow-up.
- **`checkAccessStrict` batch-limit caveat**: documented in JSDoc with a searchable anchor (`checkAccessStrict-batch-limit-caveat`). All current callers pass a single resource via `checkSingleAccessStrict`, so the ceiling doesn't affect any live code path.

## Verification

- `yarn build` — passes (7m40s clean build)
- `yarn lint` — passes
- `yarn check-types` — passes (implicit in build)
- 30 commits, all GPG-signed (`G`) and DCO signed-off
- Unit tests added: batching boundary tests, single-batch no-warning assertion, per-chunk failure isolation, `getProjectSlugs` pagination, ROOT filtering, no-access-check guard

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts
@@ -67,7 +67,7 @@ export class ProfileAffiliationsComponent {
   public readonly addWorkExperience = output();
 
   private readonly cdpAffiliations = signal<CdpProjectAffiliation[]>([]);
-  private readonly lfxSlugs = signal<Set<string>>(new Set());
+  private readonly lfxSlugs = signal<Set<string> | null>(new Set());
   private readonly connectedIdentities = signal<EnrichedIdentity[]>([]);
   private readonly identitiesLoadFailed = signal(false);
 
@@ -429,7 +429,9 @@ export class ProfileAffiliationsComponent {
         ),
       ]).pipe(
         map(([cdpAffiliations, slugs, workExperiences, identities]) => {
-          const lfxSlugs = new Set(slugs);
+          // null means the slug fetch failed — use null to skip LFX filtering so affiliations
+          // are not silently hidden when the slug endpoint is temporarily unavailable.
+          const lfxSlugs = slugs !== null ? new Set(slugs) : null;
           this.cdpAffiliations.set(cdpAffiliations);
           this.lfxSlugs.set(lfxSlugs);
           this.workExperience.set(workExperiences);
@@ -448,7 +450,7 @@ export class ProfileAffiliationsComponent {
     );
   }
 
-  private transformToProjectGroups(cdpAffiliations: CdpProjectAffiliation[], lfxSlugs: Set<string>): ProjectGroup[] {
+  private transformToProjectGroups(cdpAffiliations: CdpProjectAffiliation[], lfxSlugs: Set<string> | null): ProjectGroup[] {
     const lfid = this.userService.user()?.username || this.userService.user()?.['https://sso.linuxfoundation.org/claims/username'] || '';
 
     const verifiedOrgIds = new Set(
@@ -458,7 +460,7 @@ export class ProfileAffiliationsComponent {
     );
 
     return cdpAffiliations
-      .filter((project) => project.projectSlug && lfxSlugs.has(project.projectSlug))
+      .filter((project) => project.projectSlug && (!lfxSlugs || lfxSlugs.has(project.projectSlug)))
       .map((project) => {
         const firstRole = project.roles.length > 0 ? project.roles[0] : undefined;
         const role: 'Maintainer' | 'Contributor' = firstRole?.role === 'Maintainer' ? 'Maintainer' : 'Contributor';
--- a/apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts
@@ -419,7 +419,7 @@ export class ProfileAffiliationsComponent {
     return toSignal(
       forkJoin([
         this.userService.getCdpProjectAffiliations(),
-        this.projectService.getProjects(),
+        this.projectService.getProjectSlugs(),
         this.userService.getWorkExperiences(),
         this.userService.getIdentities().pipe(
           catchError(() => {
@@ -428,8 +428,8 @@ export class ProfileAffiliationsComponent {
           })
         ),
       ]).pipe(
-        map(([cdpAffiliations, lfxProjects, workExperiences, identities]) => {
-          const lfxSlugs = new Set(lfxProjects.map((p) => p.slug));
+        map(([cdpAffiliations, slugs, workExperiences, identities]) => {
+          const lfxSlugs = new Set(slugs);
           this.cdpAffiliations.set(cdpAffiliations);
           this.lfxSlugs.set(lfxSlugs);
           this.workExperience.set(workExperiences);
--- a/apps/lfx-one/src/app/shared/services/project.service.ts
+++ b/apps/lfx-one/src/app/shared/services/project.service.ts
@@ -19,6 +19,7 @@ export class ProjectService {
   private readonly http = inject(HttpClient);
   private readonly projectCache = new Map<string, Observable<Project | null>>();
   private readonly projectsCache = new Map<string, Observable<Project[]>>();
+  private slugsCache$: Observable<string[]> | null = null;
 
   public getProjects(params?: HttpParams): Observable<Project[]> {
     const cacheKey = params?.toString() || '';
@@ -45,9 +46,21 @@ export class ProjectService {
   /**
    * Fetches project slugs without access-check overhead. Use when only slugs are needed
    * (e.g. membership checks) — skips the OpenFGA POST that getProjects() incurs.
+   * Cached for the session lifetime; evicted on error so a failed fetch is retryable.
    */
   public getProjectSlugs(): Observable<string[]> {
-    return this.http.get<string[]>('/api/projects/slugs').pipe(retryTransientHttpError(), shareReplay(1));
+    if (!this.slugsCache$) {
+      this.slugsCache$ = this.http.get<string[]>('/api/projects/slugs').pipe(
+        retryTransientHttpError(),
+        catchError((error) => {
+          console.error('Failed to fetch project slugs:', error);
+          this.slugsCache$ = null;
+          return of([]);
+        }),
+        shareReplay(1)
+      );
+    }
+    return this.slugsCache$;
   }
 
   /**
--- a/apps/lfx-one/src/app/shared/services/project.service.ts
+++ b/apps/lfx-one/src/app/shared/services/project.service.ts
@@ -19,7 +19,7 @@ export class ProjectService {
   private readonly http = inject(HttpClient);
   private readonly projectCache = new Map<string, Observable<Project | null>>();
   private readonly projectsCache = new Map<string, Observable<Project[]>>();
-  private slugsCache$: Observable<string[]> | null = null;
+  private slugsCache$: Observable<string[] | null> | null = null;
 
   public getProjects(params?: HttpParams): Observable<Project[]> {
     const cacheKey = params?.toString() || '';
@@ -48,16 +48,19 @@ export class ProjectService {
    * (e.g. membership checks) — skips the OpenFGA POST that getProjects() incurs.
    * Cached for the session lifetime; evicted on error so a failed fetch is retryable.
    */
-  public getProjectSlugs(): Observable<string[]> {
+  public getProjectSlugs(): Observable<string[] | null> {
     if (!this.slugsCache$) {
       this.slugsCache$ = this.http.get<string[]>('/api/projects/slugs').pipe(
         retryTransientHttpError(),
-        // Evict the cache entry before catchError converts the error to of([]), so a
-        // future caller gets a fresh HTTP attempt instead of the pinned empty result.
+        // Evict the cache entry before catchError converts the error to of(null), so a
+        // future caller gets a fresh HTTP attempt instead of the pinned null result.
         tap({ error: () => (this.slugsCache$ = null) }),
         catchError((error) => {
           console.error('Failed to fetch project slugs:', error);
-          return of([]);
+          // Return null rather than [] so callers can distinguish "fetch failed, skip
+          // LFX filter" from "fetch succeeded, no LFX projects" — an empty array would
+          // make the component filter out all CDP affiliations as non-LFX contributions.
+          return of(null);
         }),
         shareReplay(1)
       );
--- a/apps/lfx-one/src/app/shared/services/project.service.ts
+++ b/apps/lfx-one/src/app/shared/services/project.service.ts
@@ -52,9 +52,11 @@ export class ProjectService {
     if (!this.slugsCache$) {
       this.slugsCache$ = this.http.get<string[]>('/api/projects/slugs').pipe(
         retryTransientHttpError(),
+        // Evict the cache entry before catchError converts the error to of([]), so a
+        // future caller gets a fresh HTTP attempt instead of the pinned empty result.
+        tap({ error: () => (this.slugsCache$ = null) }),
         catchError((error) => {
           console.error('Failed to fetch project slugs:', error);
-          this.slugsCache$ = null;
           return of([]);
         }),
         shareReplay(1)
--- a/apps/lfx-one/src/app/shared/services/project.service.ts
+++ b/apps/lfx-one/src/app/shared/services/project.service.ts
@@ -43,6 +43,14 @@ export class ProjectService {
   }
 
   /**
+   * Fetches project slugs without access-check overhead. Use when only slugs are needed
+   * (e.g. membership checks) — skips the OpenFGA POST that getProjects() incurs.
+   */
+  public getProjectSlugs(): Observable<string[]> {
+    return this.http.get<string[]>('/api/projects/slugs').pipe(retryTransientHttpError(), shareReplay(1));
+  }
+
+  /**
    * Fail-closed on error, mirroring getProjects. One retry, transient errors only — this is the
    * bootstrap-critical fast path (LFXV2-2857), so a bare blip shouldn't leave the caller's lens
    * set narrowed until the (much slower) deferred sweep eventually resolves it instead.
--- a/apps/lfx-one/src/server/services/access-check.service.ts
+++ b/apps/lfx-one/src/server/services/access-check.service.ts
@@ -47,6 +47,7 @@ export class AccessCheckService {
     const { operationName, startTime } = this.beginCheckOperation(req, resources);
 
     try {
+      // success/warning logged inside performCheck; only error is logged here on propagation.
       return await this.performCheck(req, resources, operationName, startTime, options);
     } catch (error) {
       logger.error(req, operationName, startTime, error, {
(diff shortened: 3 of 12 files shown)


apps/lfx-one/src/server/services/access-check.service.ts
```diff
@@ -207,6 +233,75 @@ export class AccessCheckService {
...
+      });
+      return resultMap;
+    }
+
+    const chunks: AccessCheckRequest[][] = [];
+    for (let i = 0; i < resources.length; i += ACCESS_CHECK_BATCH_SIZE) {
+      chunks.push(resources.slice(i, i + ACCESS_CHECK_BATCH_SIZE));
+    }
+
+    const settled = await Promise.allSettled(chunks.map((chunk) => this.performSingleCheck(req, chunk, options)));
```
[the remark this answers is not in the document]
bramwelt: Done. `checkAccessStrict` now propagates chunk failures across all batch sizes. `performCheck` accepts a `throwOnChunkFailure` flag; the strict path passes `true`, so after all chunks settle and per-chunk WARNs are logged the first rejection reason is rethrown — `checkAccessStrict`'s catch block then logs it at ERROR before rethrowing to the caller.

See commit 479a04a56: fix(review): address PR #1930 review feedback.

apps/lfx-one/src/server/services/project.service.ts
```diff
@@ -315,28 +315,28 @@ export class ProjectService {
...
+   * entirely, eliminating the OpenFGA POST that getProjects() incurs.
+   *
+   * Note: the query-service payload is as wide as getProjects() (full Project
+   * objects); only the FGA round trip is eliminated, not the response payload
+   * size. If the upstream query service ever supports field projection, pass
+   * a 'fields' parameter here to further reduce transfer cost.
+   */
+  public async getProjectSlugs(req: Request): Promise<string[]> {
+    logger.debug(req, 'get_project_slugs', 'Fetching all project slugs', { page_size: QUERY_SERVICE_PAGE_SIZE });
+    const filtered = await this.fetchAllProjectsFiltered(req);
```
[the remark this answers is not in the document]
bramwelt: Done. `fetchAllProjectsFiltered` now accepts a `failOnPartial` boolean (default `false`); `getProjectSlugs` passes `true` which propagates to `fetchAllQueryResources`. A failed page now throws so the caller receives an error rather than a truncated slug list that silently misrepresents affiliation state. `getProjects` is unchanged (keeps `failOnPartial: false`, degrades gracefully).

See commit 479a04a56: fix(review): address PR #1930 review feedback.

bramwelt:
## Review Feedback Addressed

Commit: 479a04a5685aca0828908b0ac8631d8ea23ea047

### Changes Made

- **access-check.service.ts**: `checkAccessStrict` now propagates chunk failures in multi-batch mode. Added `throwOnChunkFailure` param to `performCheck`; strict path passes `true` — after `Promise.allSettled` and per-chunk WARNs, the first rejection is rethrown. The strict error-propagation contract holds across all batch sizes (per copilot[bot]).
- **project.service.ts**: `fetchAllProjectsFiltered` now accepts `failOnPartial` (default `false`). `getProjectSlugs` passes `true` — a failed page throws instead of returning a truncated slug list that silently drops affiliations from CDP matching. `getProjects` behavior unchanged (per copilot[bot]).
- **access-check.service.spec.ts**: Updated multi-batch strict test to assert throw (not no-throw) and verify WARN + ERROR log sequence.
- **project.service.spec.ts**: Added test asserting `getProjectSlugs` throws on a failed pagination page.

### Threads Resolved

apps/lfx-one/src/app/shared/services/project.service.ts
```diff
@@ -42,6 +43,28 @@ export class ProjectService {
...
+  public getProjectSlugs(): Observable<string[]> {
+    if (!this.slugsCache$) {
+      this.slugsCache$ = this.http.get<string[]>('/api/projects/slugs').pipe(
+        retryTransientHttpError(),
+        // Evict the cache entry before catchError converts the error to of([]), so a
+        // future caller gets a fresh HTTP attempt instead of the pinned empty result.
+        tap({ error: () => (this.slugsCache$ = null) }),
+        catchError((error) => {
+          console.error('Failed to fetch project slugs:', error);
+          return of([]);
```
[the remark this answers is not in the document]
bramwelt: Done. `getProjectSlugs()` now returns `Observable<string[] | null>` — `null` signals "fetch failed, skip the LFX filter" rather than substituting an authoritative empty set. `ProfileAffiliationsComponent` propagates this through the `lfxSlugs` signal as `Set<string> | null`; `transformToProjectGroups` treats `null` as "unavailable" and passes all CDP affiliations through (`!lfxSlugs || lfxSlugs.has(...)`) instead of filtering them all out.

See commit 651d74d6c: fix(review): address PR #1930 review feedback.

bramwelt:
## Review Feedback Addressed

Commit: 651d74d6c

### Changes Made
- **apps/lfx-one/src/app/shared/services/project.service.ts**: `getProjectSlugs()` now returns `Observable<string[] | null>` — `null` signals "fetch failed, skip LFX filter" instead of returning an authoritative empty set that would filter out all CDP affiliations (per copilot-pull-request-reviewer)
- **apps/lfx-one/src/app/shared/services/project.service.spec.ts**: updated 3 tests to assert `null` return on error instead of `[]`; updated type annotations to `string[] | null`
- **apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts**: `lfxSlugs` signal is now `Set<string> | null`; `transformToProjectGroups` signature updated to accept `null` and treats it as "unavailable" — passes all CDP affiliations through rather than filtering all out

### Threads Resolved

apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts
```diff
@@ -458,7 +460,7 @@ export class ProfileAffiliationsComponent {
     );
 
     return cdpAffiliations
-      .filter((project) => project.projectSlug && lfxSlugs.has(project.projectSlug))
+      .filter((project) => project.projectSlug && (!lfxSlugs || lfxSlugs.has(project.projectSlug)))
```
dealako, line 463: [minor] No component test for the null-slug (filter-skip) branch

Issue: This filter — `!lfxSlugs || lfxSlugs.has(project.projectSlug)` — plus the `slugs !== null ? new Set(slugs) : null` mapping at line 434 is the core behavior of this PR: a slug-fetch outage (`null`) must skip LFX filtering and retain all CDP affiliations, while an empty `Set` filters them all out. There is no `ProfileAffiliationsComponent` spec, so neither branch is covered.
Proof: no `*.spec.ts` exists under `apps/lfx-one/src/app/modules/profile/affiliations/`. The frontend `ProjectService.getProjectSlugs` null fallback IS tested (shared/services/project.service.spec.ts), but the component that consumes null is not.
Why it matters: the null-vs-empty distinction is the whole point of the change. A future refactor collapsing `null` to an empty `Set` (or dropping the `!lfxSlugs` guard) would silently hide every affiliation on a transient slug outage with no failing test.
Fix: add a `ProfileAffiliationsComponent` spec exercising `transformToProjectGroups` with `lfxSlugs=null` (all affiliations retained) and `lfxSlugs=new Set([...])` (non-member slugs filtered).

apps/lfx-one/src/server/services/access-check.service.ts
```diff
@@ -197,16 +211,103 @@ export class AccessCheckService {
...
```
dealako, line 254: [question] Multi-batch access-check fans out all chunks with no concurrency cap

Issue: For >100 resources, this launches every chunk's POST simultaneously via `Promise.allSettled(chunks.map(...))`. `ACCESS_CHECK_BATCH_SIZE` bounds per-request payload, but nothing bounds the number of in-flight chunk requests.
Proof: reachable via the unscoped `getProjects()` sweep -> `addAccessToResources` -> `checkAccess` (non-strict). A large org/foundation admin with several thousand projects (e.g. 5000 -> 50 parallel POSTs) fans out dozens of concurrent requests to LFX_V2_SERVICE. Contrast `discoverSubFoundations` in project.service.ts, which gates concurrency via `acquireTraversalSlot`.
Why it matters: chunking was added to reduce payload/latency, but replacing one large sequential POST with N simultaneous POSTs can spike load on the access-check service for large tenants.
Fix: if large-N callers exist, bound chunk concurrency with a small pool / p-limit mirroring the traversal-slot gating already used in this codebase; if chunk counts are expected to stay small, a doc comment noting that is enough. Non-blocking — depends on expected N.

dealako:
Hi @bramwelt 👋 — thanks for a clean, well-scoped performance pass. The refactor cuts OpenFGA access-check demand nicely: the slug-only `/api/projects/slugs` path skips per-resource checks, `/writer-summary` returns booleans scoped to the caller's own direct grants, and the batched `Promise.allSettled` fan-out with a strict propagating variant is a solid design. The failure-semantics work you already landed in `479a04a5` and `651d74d6` (strict chunk propagation, `failOnPartial`, and the `null`-vs-empty slug distinction) resolves the earlier Copilot feedback correctly — the `null` "skip the LFX filter" signal is the right call over an authoritative empty set.

I traced the batching failure paths and the affiliation slug-fetch handling end to end. No blocking, security, or data-privacy issues. Two non-blocking items below.

### Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 1 — no component test for the `null`-slug filter-skip branch in `profile-affiliations`
- ⚪ Nit: 0
- ❔ Question: 1 — `performCheck` fans out all chunks with no concurrency cap

### AI bot reconciliation

- **Copilot** raised three items (strict-path propagation across batch boundary, `getProjectSlugs` needing `failOnPartial: true`, frontend `of([])` masking a failure as an empty set). All three are **already addressed** in `479a04a5` and `651d74d6`. Agreed with each; no residual concern.
- **Cursor Bugbot** flagged route ordering (`/slugs` before `/:slug`) as a risk — verified correct: static routes are declared before the param route, no shadowing.

### Details

**🟡 Minor — missing component test for the `null`-slug branch.** The central behavior of this change is that a slug-fetch outage yields `null` and must *skip* LFX filtering (retain all CDP affiliations) rather than filter them all out as non-LFX. That logic lives in the component (`transformToProjectGroups`, `initApiProjectGroups`), but there's no `ProfileAffiliationsComponent` spec, so neither the `null` (retain-all) nor the populated-`Set` (filter) branch is guarded. The frontend `ProjectService.getProjectSlugs` `null` fallback *is* tested; the consumer is not. A future refactor collapsing `null` to an empty `Set` would silently hide every affiliation on a transient outage with no failing test.

**❔ Question — unbounded chunk fan-out in `performCheck`.** For >100 resources, `Promise.allSettled(chunks.map(...))` issues every chunk POST simultaneously. Reachable via the unscoped `getProjects()` sweep → `addAccessToResources` → `checkAccess`, so a large-tenant admin (e.g. 5000 projects → 50 parallel POSTs) could spike load on the access-check service. `ACCESS_CHECK_BATCH_SIZE` bounds per-request payload but nothing bounds in-flight chunk count, unlike the `acquireTraversalSlot` gating used elsewhere. If large-N callers are expected, consider a small concurrency pool; if chunk counts stay small in practice, no change needed.

### Decision

✅ **Approved with minor comments** — the two items above are worth addressing (especially the component test) but neither blocks merge.

dealako (dismissed):
✅ Approved with minor comments. No blocking, security, or data-privacy issues. Two non-blocking items posted inline (one minor: missing component test for the null-slug filter-skip branch; one question: unbounded chunk fan-out in performCheck). Copilot's three earlier findings are already resolved in 479a04a5 and 651d74d6. See the summary comment for details.

apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.ts
bramwelt: Done. Added `profile-affiliations.component.spec.ts` with three cases exercising `transformToProjectGroups` directly (via type casting):

1. `lfxSlugs = null` → all affiliations pass through (slug fetch unavailable, skip filter)
2. `lfxSlugs = new Set(['slug-a', 'slug-c'])` → only matching slugs retained
3. `lfxSlugs = new Set()` → all affiliations filtered (fetch succeeded, no LFX projects)

The spec uses the same TestBed + empty template override pattern as `profile-visibility-drawer.component.spec.ts`, with minimal service stubs for initialization.

See commit 665887413: fix(review): address PR #1930 review feedback.

apps/lfx-one/src/server/services/access-check.service.ts
bramwelt: Good catch on the pattern. A concurrency cap is not warranted here: LFX has ~500 projects and `ACCESS_CHECK_BATCH_SIZE = 100`, so the maximum concurrent chunk count is ~5 — the same order of magnitude as sequential concurrent requests from different users hitting the service. A `p-limit` or traversal-slot guard would add complexity without meaningful load reduction at these numbers.

Added a JSDoc note on `performCheck` documenting this reasoning and pointing to `acquireTraversalSlot` in `project.service.ts` as the pattern to revisit if a future caller drives chunk counts into the dozens.

See commit 665887413: fix(review): address PR #1930 review feedback.

bramwelt:
## Review Feedback Addressed

Commit: 665887413

### Changes Made
- **apps/lfx-one/src/app/modules/profile/affiliations/profile-affiliations.component.spec.ts**: added new spec file exercising `transformToProjectGroups` with three cases: `null` slugs (all retained), non-empty `Set` (filtered to matches), and empty `Set` (all filtered) — guards the null-vs-empty contract introduced in this PR (per @dealako)
- **apps/lfx-one/src/server/services/access-check.service.ts**: added JSDoc note to `performCheck` documenting that all chunks fan out in parallel with no concurrency cap, why ~5 max chunks makes a cap unnecessary, and points to `acquireTraversalSlot` as the pattern to revisit if chunk counts grow (per @dealako)

### Threads Resolved

dealako:
Hi @bramwelt 👋 — thanks for the quick, precise turnaround on both items from the last round. This one's clean.

### 👏 Nice work

- **Null-slug branch is now tested.** The new `profile-affiliations.component.spec.ts` exercises `transformToProjectGroups` directly across all three cases: `null` (retain all, order preserved), a populated `Set` (filter to matches), and an empty `Set` (filter all out). Test 1 vs. Test 3 lock the null-vs-empty contract — if `null` ever collapsed to empty semantics, Test 1 would fail. That's exactly the regression guard the branch was missing.
- **Fan-out question answered in the right place.** Rather than adding a limiter that isn't warranted, you documented the reasoning on `performCheck`: ~500 projects at `ACCESS_CHECK_BATCH_SIZE=100` → ~5 concurrent chunks, on par with normal per-user load, with a pointer to `acquireTraversalSlot` as the escalation path if chunk counts ever grow. Accurate arithmetic, real reference, correct call.

### Revision tracking

- ✅ **Resolved** — [minor] No component test for the null-slug filter-skip branch → `665887413` adds the spec with assertions matching the filter contract.
- ✅ **Resolved** — [question] `performCheck` fan-out with no concurrency cap → `665887413` adds a JSDoc note bounding the fan-out and documenting the deliberate no-limiter decision.

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

No new findings across correctness, performance, test coverage, security, data privacy, style, API compliance, or documentation on the new commit. No new bot comments to reconcile since the last round.

### Decision

✅ **Approved** — both prior items closed out cleanly, nothing new. Nice work.

dealako (approved):
✅ Approved. Both items from the prior round are resolved in 665887413: the null-slug filter-skip branch now has a component spec (transformToProjectGroups across null / populated Set / empty Set), and the performCheck fan-out question is answered with an accurate JSDoc note (~5 chunks at current scale, acquireTraversalSlot cited as the escalation path). No new findings across any dimension. See the summary comment for details.

Issue #1697, opened 2026-08-20
Email as an LFX marketing channel — end-to-end email integrations for event marketing


mrautela365:
Migrated from [LFXV2-2770](https://linuxfoundation.atlassian.net/browse/LFXV2-2770).

## Email as an LFX marketing channel — end-to-end email integrations for event marketing

Adds **Email** as a first-class LFX marketing channel — content generation, audience/segment building, and HubSpot email staging — integrated per the campaign-service broker architecture (Go service owns platform APIs + persistence, FGA, brief → dispatcher → campaign). Email is a `campaigns` row with `platform='hubspot'` under a shared brief.

**Design doc:** `docs/email-channel-integration.md`

### Decisions (signed off)

1. **Audience home = B2** — a first-class `campaign_audiences` table storing a POINTER + provenance (HubSpot `master_list_id`, suppression list IDs, inclusion summary, status, version), **not** the audience contents, which stay in HubSpot. Makes the audience inspectable, reusable across sends, and versioned.
2. **Snowflake source** = `PLATINUM_LFX_ONE.event_registrations` (read-only).
3. `platform` value = `hubspot`, key-consistent with `connection-hubspot` / `hubspot_connections`.

### Three ports (broker boundary)

- **HubSpot** (`internal/platform/hubspot`) — marketing-email clone/patch/content, CRM-lists CRUD, event-defs, files, campaigns
- **Snowflake** (read-only, key-pair) — resolve exact past `EVENT_NAME` strings for HubSpot BEHAVIORAL_EVENT filters
- **LLM** — already abstracted via the LF LiteLLM proxy

### Scope of this migration

The **backend is essentially complete** — 12 of 16 children are Released (the HubSpot port, Snowflake port, `campaign_audiences` resource, audience-building logic, metrics reader, portal-provenance guard). Only the **4 open** children were moved; the Released ones stay in Jira as historical record.

**Every remaining item is UI or BFF**, and they form a dependency chain: the config builder unblocks staging, which the audience and monitor surfaces depend on.

**Target Release Date: 2026-09-30**

[LFXV2-2770]: https://linuxfoundation.atlassian.net/browse/LFXV2-2770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mrautela365:
[...]

PR #1932, opened 2026-08-28
feat(campaigns): build the email delivery channel

### LINKED ISSUE #1700: campaigns UI: no way to build or choose an audience, so an email campaign cannot stage
Migrated from [LFXV2-3224](https://linuxfoundation.atlassian.net/browse/LFXV2-3224).

**Jira status at migration:** Backlog

`HubSpotDispatcher.Dispatch` points the staged email's send list at **the brief's built audience**. An email campaign with no built audience cannot stage.

The backend is complete (`campaign_audiences` resource, the Snowflake→HubSpot build, `updated_by` attribution, the per-platform lease). **Nothing in the UI calls any of it** — no way to trigger a build, see what it produced, or choose between builds.

So a user completes a brief, picks a template, hits stage — and dispatch fails for a reason nothing in the UI ever told them about.

**Scope:** trigger a build; show the pointer + provenance + status; refuse staging honestly when not built; allow reuse across sends.

**Two things to get right:**
- A build is not instant and can partially fail — needs a pending state and a failed-with-reason state, not a boolean.
- **Fail closed on the Snowflake half.** If Snowflake is unreachable or returns zero rows the past-registrant list is SKIPPED and flagged, never substituted with guessed event names. A partially-built audience must be visibly partial — sending to it is a real send to real people.

Parent epic: #1697

[LFXV2-3224]: https://linuxfoundation.atlassian.net/browse/LFXV2-3224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### LINKED ISSUE #1701: campaigns BFF: hubspot has no config builder, so an email campaign is refused before dispatch
Migrated from [LFXV2-3256](https://linuxfoundation.atlassian.net/browse/LFXV2-3256).

**Jira status at migration:** In Progress

**The first hard failure on the email path** — upstream of the audience work, so that cannot be verified end to end until this lands.

`hasPlatformConfig` maps exactly four platforms to their required config key and returns `false` for anything unmapped. `hubspot` is unmapped, so an email campaign is refused **client-side, before the request reaches the Go dispatcher**.

**This is deliberate and the guard is correct** — this service builds no config for hubspot, and waving it through would queue a job whose dispatcher reads an absent key as a zero value. So this ticket is **not** "delete the guard"; it is "build the missing config builder, then map it".

**What the dispatcher needs:** `hubspotConfig.sourceEmailId`, a **built** audience for the brief, and a non-empty `PlatformMasterListID`.

**Scope:** build a `hubspotConfig` envelope carrying `sourceEmailId`; add the mapping **only once that builder exists**; cover it with a spec pinning both directions — refused while absent, accepted once present.

Parent epic: #1697

[LFXV2-3256]: https://linuxfoundation.atlassian.net/browse/LFXV2-3256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mrautela365:
Builds the email delivery channel on the Campaigns page: generate a brief from an
event URL, review and correct it, generate email copy, build the send audience,
and stage a HubSpot draft.

The UI and BFF live here; content generation and draft application live in
campaign-service, per the PR #2 architecture. The BFF is a thin proxy to the Go
service — it does not generate copy itself.

## What's included

- **Email delivery type** on the Campaigns page, with its own Plan → Implement →
  Monitor flow. The Ad Channels card is hidden for email, which has no ad platforms.
- **An editable brief.** The brief is a scrape of the event page and the content
  generator is instructed to use its fields verbatim, so a wrong scraped date
  previously had no remedy short of editing the event page and re-scraping.
  Event name, dates, location, audience and registration URL are now shown and
  editable, and corrections are written back into the object that generation reads.
- **Email copy generation** proxied to campaign-service's
  `POST /projects/{slug}/briefs/{id}/email-copy`.
- **Audience building** proxied to `/audiences/build`.
- **Staging a HubSpot draft** as a clone of a chosen marketing email, carrying the
  generated subject and body through `hubspotConfig`.

The HubSpot UTM lookup is existing shared code (`lookupHubSpotUtm`), used
unchanged by both the paid and email paths — nothing was rebuilt for email.

## Bug fixed along the way

Event extraction parsed the model's reply with a bare `JSON.parse`, so a reply
wrapped in a ```` ```json ```` fence threw. The `catch` downgrades that to a status
message and the stream still completes with a 200, so **every email brief was
silently losing its event details** and the failure read as an AI that could not
understand the event page. `stripJsonFences` already existed and was already
applied to the copy and keyword stages; extraction and the LinkedIn strategy
stage never called it. This fixes the missing calls rather than adding a second
helper.

## Contract bugs found by running the full stack

The stack was stood up locally — campaign-service on Postgres behind this BFF —
and three contract mismatches surfaced that no amount of reading would have
shown, because each one FAILED SILENTLY or looked like a different problem:

- **`countryCode` vs `country`.** Every audience build failed with "has no
  country in its details". Upstream reads `json:"country"` and needs the NAME,
  not the ISO-2 code; its own comment warns that a raw `KE` would "match no
  HubSpot country property, and the build would SUCCEED while storing an empty
  inclusion list" — a silent wrong send list.
- **`name`/`city` vs `eventName`/`location`.** Copy generation 400'd with
  "provide at least eventName". Added as ALIASES, not renames: the wire shape is
  shared with the paid path, whose consumers read the UI's names.
- **The generated copy never reached HubSpot.** `buildHubSpotConfig` is an
  allow-list naming only `sourceEmailId` and `utmCampaign`, so `subject` and
  `bodyHtml` were dropped between a UI that sends them and a dispatcher that
  reads them. The staged draft kept the template's own subject and nothing
  failed — the whole copy feature was inert at the last hop.

Plus one defect introduced by a fix in this branch: gating the audience rebuild
control on `status !== 'built'` offered a retry on an UNCONFIRMED build, which
upstream keeps in that state precisely because a HubSpot list may already exist.
A retry there creates the duplicate contact list the state exists to prevent.

## Testing

Full suites pass with no dev bypasses applied: **2231 server, 1365 app, 0
failures.** Format, lint, license headers and both tsconfigs clean. Rebased on
`main`.

Key behaviours are mutation-verified — each fails its test when reverted: the
fence fix, the brief-edit write-back, the email-mode guard, the audience status
gate, the header-sourced etag, the field aliases, the copy forwarding, the
rebuild gate, and the unconfirmed label.

Two review passes ran against this branch. The second found three assertions
that PASSED against the bug they guarded (a self-referential alias comparison, a
`toBeDefined()` on a value, and an untested copy change); all three now bind.

**Verified live in the browser** against
`https://events.linuxfoundation.org/mcp-dev-summit-nairobi/`, with
campaign-service, Postgres and NATS running:

| Step | Status |
| --- | --- |
| Brief generation → persisted to Postgres | ✅ |
| Brief rendered and editable | ✅ |
| HubSpot UTM | ✅ `46874703-MCP%20Dev%20Summit-Nairobi-2026` |
| Audience build | ✅ reaches `built`, real HubSpot master list |
| Email content generation | ✅ real subject, preview text and body |
| Template picker | ✅ |
| Staging a HubSpot draft | ✅ draft created, job `succeeded` |
| Generated subject on the staged draft | ✅ confirmed in HubSpot |

The copy-forwarding fix was confirmed against the live portal after the fix:
draft `220600718865` carries the generated subject "MCP Dev Summit Nairobi -
Join the Developer Community", with UTM tagging applied to its links. Its body
was correctly left as the template wrote it — that template has five rich-text
widgets, and the dispatcher declines to guess which one a generated body should
replace.

Worth knowing when checking this by hand: the subject lands on the email's
DRAFT, not on the published resource. `GET /marketing/v3/emails/{id}` still
shows the template's subject until someone publishes; `GET
/marketing/v3/emails/{id}/draft` is the one that reflects staging.

## Related

- campaign-service: `linuxfoundation/lfx-v2-campaign-service` PR #194 — applies
  the generated copy to the staged draft and fixes the audience build. **Must
  deploy before staged drafts carry copy.**

Closes #1700
Closes #1701
Refs #1697

LFXV2-2770 · LFXV2-3224 · LFXV2-3256


apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1013,13 +1120,217 @@ export class CampaignsComponent {
...
+
+      const outcome = await firstValueFrom(this.campaignService.createCampaign(request, projectSlug, briefId));
+
+      if (outcome.error) {
+        this.emailStaging.set('error');
+        this.emailStagingMessage.set(outcome.error);
+        return;
+      }
+
+      this.emailStaging.set('done');
```
dealako: [blocking] Staging reports success on a job acknowledgment, never polling

Issue: `emailStaging` is set to `'done'` ("Draft created in HubSpot") on any non-error response.
Proof: `createCampaign` returns a bare `{ jobId }` 202 ack (campaign.controller.ts:389-395); the HubSpot clone, audience resolution, and copy application run in the background job. The paid path polls via `pollJob`/`getCreateResult` (implementation-tab.component.ts:1495); this method never polls.
Why it matters: a dispatcher/audience/HubSpot failure is reported to the operator as a created draft — the exact outcome the panel copy promises it won't produce.
Fix: after `createCampaign` returns a jobId, poll it (reuse `getCreateResult`/`pollJobStatus` as the paid tab does) and set `'done'` only on a terminal success; surface job errors as the staging failure message.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1013,13 +1120,217 @@ export class CampaignsComponent {
...
+        geoTargets: [],
+        platforms: ['hubspot'],
+        // Generated copy rides along when it exists (LFXV2-2775). Spread conditionally rather
+        // than sent as empty strings: upstream treats a blank subject as "leave the template's
+        // own", and sending '' would be indistinguishable from an operator who generated nothing
+        // — but it would also mean every staging call claimed to carry copy it did not have.
+        hubspotConfig: {
+          sourceEmailId,
+          ...(copy === null ? {} : { subject: copy.subject, bodyHtml: copy.body }),
+        },
```
dealako, line 1388: [blocking] Generated subject/bodyHtml are dropped by the BFF before reaching campaign-service

Issue: the frontend sends `hubspotConfig.{subject,bodyHtml}` and `HubSpotCampaignCreateRequest` declares both, but the BFF's `buildHubSpotConfig` (campaign.controller.ts:1814-1826) reads only `sourceEmailId` and `utmCampaign` and returns them exclusively.
Proof: campaign.controller.ts:1820/1824/1826 — `subject`/`bodyHtml` are never read; this is the only assembler of the hubspotConfig envelope.
Why it matters: the staged draft keeps the template's own copy, so the whole copy-generation/preview/staging chain is inert at the send boundary, contradicting the interface JSDoc and the in-UI preview note.
Fix: extend `buildHubSpotConfig` to read and forward `subject`/`bodyHtml` with the same runtime type-guarding used for `sourceEmailId`, including them only when non-empty strings; add a controller spec mirroring the `utmCampaign` coverage.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1350,6 +1661,30 @@ export class CampaignsComponent {
...
+  private async ensureEmailBriefId(brief: CampaignBriefOutput, projectSlug: string): Promise<string> {
+    const known = this.emailBriefId();
+    if (known !== '') {
+      return known;
+    }
+
+    const persisted = await firstValueFrom(this.campaignService.persistBrief(brief, projectSlug));
+    const briefId = persisted.briefId ?? '';
+    if (briefId !== '') {
+      this.emailBriefId.set(briefId);
```
dealako, line 1908: [blocking] ensureEmailBriefId can never update, and is fully blocked when a brief already exists

Issue: `persistBrief(brief, projectSlug)` is called with no `knownBriefId`.
Proof: when a `(project, event)` row exists that this session did not create/load, `saveBrief` returns `unowned-brief-exists` with `briefId: ''` (campaign-service.service.ts:650-655). `ensureEmailBriefId` then returns `''`, which callers treat as failure.
Why it matters: campaign-service briefs are shared across channels, so any event that already has a saved paid brief makes copy, audience, and staging all unusable for the email channel.
Fix: load/reuse the existing brief (with its ownership/ETag handling) before persisting, or extend the upstream identity to distinguish delivery types.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1350,6 +1661,30 @@ export class CampaignsComponent {
...
+   *
+   * Email never persists on the Plan tab, but BOTH the audience build and campaign creation are
+   * brief-scoped upstream, so each needs an id. Persisting once and caching it here stops the two
+   * actions writing two briefs for the same event.
+   *
+   * Returns '' when the persist could not produce an id; callers must treat that as a failure
+   * rather than proceeding with an empty path segment.
+   */
+  private async ensureEmailBriefId(brief: CampaignBriefOutput, projectSlug: string): Promise<string> {
+    const known = this.emailBriefId();
```
dealako, line 1883: [blocking] Unscoped emailBriefId cache survives a foundation switch

Issue: the foundation-change subscription (~850-960) clears the template picker, campaign list, banners and selection, but not `emailBriefId`. Only Start Over (:2089) clears it.
Proof: no `emailBriefId.set('')` in the `toObservable(activeFoundationSlug)` handler.
Why it matters: after switching foundations and picking a template, `ensureEmailBriefId` returns the previous foundation's cached id and sends audience/copy/create requests to the new project with an id owned by the old one.
Fix: clear `emailBriefId` in the foundation subscription alongside the other project-scoped signals, or key the cached id by `(project, event)`.

apps/lfx-one/src/server/services/campaign-service.service.ts
```diff
@@ -705,6 +746,112 @@ export class CampaignServiceClient {
...
+        return { enabled: true, error: 'The generator returned no email copy.' };
+      }
+
+      return {
+        enabled: true,
+        copy: { subject: copy.subject, preheader: copy.preheader, body: copy.body, cta: copy.cta },
+      };
+    } catch (error) {
+      logger.warning(req, 'generate_email_copy', 'Email copy generation failed, returning an error result', { err: error });
+      return { enabled: true, error: 'The email copy could not be generated. Try again.' };
```
dealako: [minor] Unconfigured-model 503 rewritten to a generic retry message

Issue: the catch collapses every failure to "The email copy could not be generated. Try again."
Proof: campaign-service.service.ts:791-792 — the `MicroserviceError` message is discarded. A non-transient 503 (model not configured) reads as a transient retry.
Why it matters: the operator never learns a deployment configuration is required.
Fix: preserve the controlled `MicroserviceError` message for known non-transient cases; reserve the generic retry text for transport/unknown failures (same at ~836).

dealako (changes requested):
Hi @mrautela365 — thorough, well-documented PR. The comment discipline explaining *why* each guard exists (foundation-switch clears, ownership generation, ETag fallback) is genuinely excellent and made this reviewable. The scope is a full email delivery channel: brief generation/edit, copy generation, audience build, HubSpot draft staging, plus a real fence-parsing bug fix along the way.

That said, four blocking defects break the feature end-to-end at its integration seams, and the bots (Cursor Bugbot + Copilot) independently landed on the same four. I verified each against the code on disk rather than taking them at face value; all four hold.

### 🔴 Blocking: 4 issues

1. **Staging reports success on a job acknowledgment, never polling for the draft** (`campaigns.component.ts:1324`). `onStageEmailSend` sets `emailStaging` to `'done'` ("Draft created in HubSpot") whenever the create response has no `error`. But `createCampaign` returns a bare `{ jobId }` 202-style ack (`campaign.controller.ts:389-395`); the HubSpot clone, audience resolution, and copy application all happen in the background job. The paid path polls via `pollJob`/`getCreateResult` (`implementation-tab.component.ts:1495`); the email path never does. Any dispatcher failure is reported as a created draft. Agrees with Cursor "Staging succeeds before draft exists" and Copilot.

2. **Generated `subject`/`bodyHtml` dropped by the BFF before reaching campaign-service** (`campaigns.component.ts:1311-1313` send-site; root cause `campaign.controller.ts:1814-1826`). The frontend sends `hubspotConfig.{subject,bodyHtml}` and the shared `HubSpotCampaignCreateRequest` declares both, but `buildHubSpotConfig` reads only `sourceEmailId` and `utmCampaign` and returns them exclusively — `subject`/`bodyHtml` are silently stripped. The staged draft keeps the template copy, so the entire copy-generation/preview/staging chain is inert at the send boundary. No controller test covers this. Agrees with Copilot.

3. **`ensureEmailBriefId` can never update, and is blocked entirely when a brief already exists** (`campaigns.component.ts:1683`). It calls `persistBrief(brief, projectSlug)` with no `knownBriefId`. When a `(project, event)` row already exists that this session did not create/load, `saveBrief` returns `unowned-brief-exists` with `briefId: ''` (`campaign-service.service.ts:650-655`). `ensureEmailBriefId` returns `''`, which callers treat as failure — blocking copy, audience, and staging. Because campaign-service briefs are shared across channels, any event with a saved paid brief makes the email channel unusable. Agrees with Cursor "Email brief persist cannot update" and Copilot.

4. **Unscoped `emailBriefId` cache survives a foundation switch** (`campaigns.component.ts:1675`). The foundation-change subscription (`~850-960`) clears the template picker, campaign list, banners, and selection, but not `emailBriefId` — only Start Over (`:2089`) clears it. After switching foundations and picking a template, `ensureEmailBriefId` returns the previous foundation's cached id and sends audience/copy/create requests to the new project with an id owned by the old one. Agrees with Copilot. Fix: clear `emailBriefId` in the foundation subscription (or key it by `(project, event)`).

### 🟡 Minor: 2 issues

1. **Unconfigured-model 503 rewritten to a generic retry message** (`campaign-service.service.ts:792-793`, also ~836). The `catch` collapses every `generateEmailCopy` failure to "The email copy could not be generated. Try again." A non-transient 503 (model not configured) is a deployment problem the operator will never learn about. `MicroserviceError` already carries the controlled upstream message — preserve it for known non-transient cases and reserve the generic text for transport/unknown failures. Agrees with Copilot.

2. **Staging tests encode the ack-as-done behavior** (`campaigns.component.spec.ts`, LFXV2-3201 block). The staging tests mock `createCampaign` returning a bare `{ jobId }` and assert `emailStaging()` is `'done'`. They pass *because of* blocking issue #1 and lock it in. Once polling is added, add a test where the job resolves to `failed` after the ack and assert staging is `'error'`, not `'done'`.

### ⚪ Nit: 1 issue

1. **`emailBriefId` JSDoc is stranded on `briefRequiredHint`** (`campaigns.component.ts:677-687`). The multi-line block documenting `emailBriefId` sits above `briefRequiredHint` (which has its own one-line comment), so `emailBriefId` reads as undocumented. Move the block down to its field.

### ❔ Questions: 2 items

1. **Audience build has no poll/status re-read; a `building` 202 may strand staging** (`campaigns.component.html:464` / `campaigns.component.ts` `emailAudienceState`). `onBuildAudience` sets the audience from the 202 and returns to idle with no poll. `canStageEmail` admits only `status === 'built'`. If `/audiences/build` returns `building` on the 202, Stage never enables and Retry loops without a path forward. Also once `emailAudience` is set the Build control is removed, so a `failed` audience has no rebuild path (Cursor "Audience failure blocks all retries"). Depends on upstream synchronous-vs-async build semantics — please confirm whether `built` is reachable on the 202.

2. **`countryCode` vs `country` contract mismatch on audience build** (Copilot, `campaigns.component.ts:1156/1306`). Copilot reports campaign-service's `decodeEventDetails` requires a full `country` name and rejects ISO-2 `countryCode`, while persisted `eventDetails` carries only `countryCode` (`campaign.interface.ts:159`, proxy `:719`). I could not verify this — `decodeEventDetails` lives in campaign-service, outside this repo. If the upstream contract truly wants `country`, audience build fails for every brief this UI produces; please confirm against the Goa design before merge.

### Note on the good part

The `stripJsonFences` fix on the extraction and LinkedIn-strategy stages is a real bug fix — a fenced ```` ```json ```` reply previously threw, got swallowed by the `catch`, and every email brief silently lost its event details behind a 200. Reusing the existing helper rather than adding a second was the right call.

🔴 **Needs changes before approval** — the four blocking issues make the channel non-functional end-to-end. Once staging polls a real terminal outcome, the BFF forwards `subject`/`bodyHtml`, brief-id reuse handles the shared-brief case, and `emailBriefId` is cleared/scoped per foundation, this is close.

dealako:
## Review summary — @mrautela365

Thorough PR building the full email delivery channel (brief gen/edit, copy generation, audience build, HubSpot staging) plus a real `stripJsonFences` bug fix. The comment discipline explaining each guard's rationale is excellent. But four blocking defects break the feature end-to-end at its integration seams — the Cursor Bugbot and Copilot bots independently landed on the same four, and I verified each against the code on disk.

- 🔴 **Blocking: 4** — (1) staging reports success on a bare `{ jobId }` ack and never polls the actual draft outcome; (2) BFF `buildHubSpotConfig` drops the generated `subject`/`bodyHtml`, so the staged draft keeps template copy; (3) `ensureEmailBriefId` persists with no known id and is fully blocked (`unowned-brief-exists`) whenever a shared brief already exists; (4) unscoped `emailBriefId` cache survives a foundation switch, sending old-foundation ids to the new project.
- 🟡 **Minor: 2** — unconfigured-model 503 rewritten to a generic retry message (operator never learns config is required); staging tests encode the ack-as-done behavior and lock it in.
- ⚪ **Nit: 1** — `emailBriefId` JSDoc stranded on `briefRequiredHint`.
- ❔ **Questions: 2** — audience build has no poll/rebuild path (a `building` 202 may strand staging); `countryCode` vs `country` contract mismatch on audience build (couldn't verify — `decodeEventDetails` is upstream).

Full detail with inline comments and fixes is in the formal review: https://github.com/linuxfoundation/lfx-self-serve/pull/1932#pullrequestreview-5055130342

🔴 **Needs changes before approval.**

dealako (changes requested):
Thanks for the fast turnaround @mrautela365 — four targeted commits, each with tests, and the commit messages tracing the live-observed failure (draft `220597885197`, HubSpot list `30779`) make it easy to see exactly what each fix addressed.

## 👏 Nice work this round

- **B2 (subject/bodyHtml drop) — fully resolved.** `buildHubSpotConfig` now forwards `subject`/`bodyHtml` with the same `typeof === 'string'` + `trim()` guard as `sourceEmailId`, and includes each only when non-empty (`campaign.controller.ts`). The two new controller specs — one asserting both fields reach the envelope, one asserting a whitespace subject is omitted rather than blanking the draft — lock the behavior in. This is exactly the fix.
- **Both prior questions — resolved.** The new `toUpstreamEventDetails`/`countryNameFor` mapping sends the `country` **name** the audience builder decodes (not the ISO-2 code) and adds the `eventName`/`location` aliases campaign-service reads, while keeping the UI's own field names for the shared paid path. Adding aliases rather than renaming is the right call, and the empty-string-on-unrecognized-code choice (fail loud vs. silently build an empty send list) is well-reasoned and tested.
- The audience-rebuild refinement (offer Retry on `failed`/absent but **not** on `building`, because an unconfirmed build may have already created a HubSpot list) is a genuine improvement over a naive poll, with the reconcile-id hint carried in the UI copy.

## Still open — carried over from the last round

These four commits didn't touch the code behind the earlier blockers (confirmed by diffing `efec9638..fb315761`), so they remain exactly as flagged. Copilot's latest review re-surfaced the same items as "previously missed — in code that hasn't changed."

🔴 **Blocking: 3**

1. **Staging still reports success on a bare `{ jobId }` ack, never polling** (`campaigns.component.ts:1324`). `onStageEmailSend` sets `emailStaging('done')` ("Draft created in HubSpot") whenever `outcome.error` is falsy. `createCampaign` returns only `{ jobId }` (`campaign.controller.ts:389`); the HubSpot clone, audience resolution, and copy application all run in the background job. The paid path polls via `getCreateResult`/`pollJob` (`implementation-tab.component.ts:1495`); the email path never does. Now that B2 forwards the copy, this is the last gap keeping staging from being trustworthy — a dispatcher/copy-application failure still reads as "Draft created." Copilot flagged the same at `:1325`. Poll the returned jobId and set `done` only on a terminal success.

2. **`ensureEmailBriefId` still can't update and is fully blocked when a shared brief exists** (`campaigns.component.ts:1683`). It calls `persistBrief(brief, projectSlug)` with no `knownBriefId`, so when a `(project, event)` row already exists that this session didn't create/load, `saveBrief` returns `unowned-brief-exists` with `briefId: ''` (`campaign-service.service.ts:650`) and copy/audience/staging all fail. Because campaign-service briefs are shared across channels, any event with a saved paid brief makes the email channel unusable. Copilot independently raised this at `:1680` ("first-save call cannot support an event that already has a live brief") and adds a related concern: even on success, a `approved: false` result gets cached, making retries skip approval permanently. Needs a load/reuse strategy or a channel-aware brief key.

3. **Unscoped `emailBriefId` cache still survives a foundation switch** (`campaigns.component.ts:1675`). The `toObservable(activeFoundationSlug)` handler (~850-931) clears templates, selection, and the campaign list, but not `emailBriefId`/`emailAudience`/`emailCopy` — only `resetEmailBriefDerivedState` (`:2088`, called from Proceed and Start Over) clears them. After switching foundations and picking a template, `ensureEmailBriefId` returns the previous foundation's id and targets the new project with an id owned by the old one. Copilot flagged the same at `:1172`, also noting the late audience/copy/staging responses aren't generation-guarded. Clear the brief-derived email state in the switch handler (and generation-guard the async responses).

🟡 **Minor: 2**

1. **Unconfigured-model 503 still rewritten to a generic retry** (`campaign-service.service.ts:792`, and `buildAudience` at ~851). Both catches collapse every failure to "Try again" / "Check the HubSpot connection," discarding the `MicroserviceError` message — so a non-transient 503 (model not configured) reads as transient, and Copilot notes the build path also loses the `reason` discriminator (`stale_approval` / `audience_build_in_flight` / `already_exists`) whose remedies differ. Preserve the controlled upstream message for known non-transient cases.

2. **Staging spec still encodes ack-as-done** (`campaigns.component.spec.ts:2296`). It asserts `emailStaging()` is `'done'` on a bare `{ jobId: 'job-1' }`, which passes only because of blocking #1. Once polling lands, add a test where the job resolves to `failed` after the ack and assert `'error'`.

⚪ **Nit: 1** — `emailBriefId` JSDoc (`campaigns.component.ts:677-682`) is still stranded above `briefRequiredHint`; move it to its field.

## One to confirm (not blocking)

Copilot's new review claims the UI's single `dates` string never reaches campaign-service's date decoder (which reads `startDate`/`endDate`), so every email prompt gets "Date TBD." I couldn't prove this from this repo — `dates` is forwarded under its UI name by the `...details` spread in `toUpstreamEventDetails`, and `startDate`/`endDate` appear to be paid-platform config fields. Since the new alias work deliberately mapped the fields campaign-service decodes, worth a quick confirm against the Goa design that `dates` is actually read for email copy; if it isn't, add a `dates` alias the same way.

🔴 **Needs changes before approval.** Real progress — B2 and both questions are genuinely closed — but the three carried-over blockers make staging still untrustworthy and the channel unusable for shared/existing briefs. Close those and this is nearly there.

dealako:

Fast, focused revision — four commits, each tested and each tracing a live-observed failure. 👏 **B2 (the dropped `subject`/`bodyHtml`) is fully resolved** — `buildHubSpotConfig` now forwards both with the correct type-guard + conditional-inclusion, and the two new controller specs lock it in. **Both prior questions are also resolved**: the new `toUpstreamEventDetails`/`countryNameFor` mapping sends the `country` name (not the ISO-2 code) and the `eventName`/`location` aliases campaign-service decodes, keeping the UI names for the shared paid path. The audience-rebuild refinement (no Retry on an unconfirmed `building` row that may have already created a HubSpot list) is a real improvement.

The three carried-over blockers are in code these commits didn't touch (verified by diffing `efec9638..fb315761`; Copilot's latest review re-surfaced the same as "code that hasn't changed"):

- 🔴 **Blocking: 3** — (1) staging still sets `done` on a bare `{ jobId }` ack and never polls the real draft outcome; (2) `ensureEmailBriefId` still persists with no known id and is fully blocked by `unowned-brief-exists` when a shared brief already exists (Copilot adds: an `approved:false` id gets cached too); (3) unscoped `emailBriefId`/`emailAudience`/`emailCopy` still survive a foundation switch, so old-foundation ids target the new project.
- 🟡 **Minor: 2** — unconfigured-model 503 still collapsed to a generic retry (build path also loses its `reason` discriminator); staging spec still asserts `done` on a bare ack, locking in blocker #1.
- ⚪ **Nit: 1** — `emailBriefId` JSDoc still stranded on `briefRequiredHint`.
- ❔ **Confirm (not blocking)** — Copilot claims the UI's `dates` string never reaches campaign-service's date decoder ("Date TBD"); couldn't verify against the Goa design — worth a quick check.

Detail and fixes are in the formal review: https://github.com/linuxfoundation/lfx-self-serve/pull/1932#pullrequestreview-5055479291

🔴 **Needs changes before approval** — genuine progress, but the three carried-over blockers keep staging untrustworthy and the channel unusable for shared/existing briefs.

mrautela365:
Thanks @dealako — all three blockers and both minors are addressed in `213a60f7`. Each fix is mutation-verified (reverting it fails a test).

**Blocking 1 — ack-as-done.** `onStageEmailSend` now polls the returned job id via `getCreateResult`, mirroring `implementation-tab`'s `pollJob` with the same 5-minute bound. `done` is set only on a terminal success; the job's own error surfaces on failure; exhausting the poll budget reports that a draft may exist and to check HubSpot, rather than spinning or claiming success. Your minor #2 was the same defect from the other side — the spec asserting `'done'` on a bare `{ jobId }` now asserts `'staging'`, and two new tests cover the job settling to failure and to success.

**Blocking 2 — shared briefs.** `ensureEmailBriefId` now consults `knownBriefIds`, the same ownership record the paid save uses, and passes `knownBriefId`/`knownEtag`/`allowEtagFallback` through to `persistBrief`. An event whose paid brief this session generated or restored is now replaceable rather than refused as `unowned-brief-exists`. Falls through to a plain first save when there is no owned row.

**Blocking 3 — unscoped cache.** The `activeFoundationSlug` handler now calls `resetEmailBriefDerivedState()`, clearing `emailBriefId`, `emailAudience`, `emailCopy` and the staging signals alongside the templates it already cleared.

**Minor 1 — non-transient messages.** `generateEmailCopy` and `buildAudience` keep the upstream message for a controlled 4xx or a deliberate 503, so "AI model is not configured" reaches the operator instead of "Try again". An unexpected 5xx still falls back to the generic text, since it can carry stack or infrastructure detail.

**Nit** — `emailBriefId`'s JSDoc is reattached.

**On the `dates` question:** you were right to doubt it. `email_copy.go`'s `emailCopyEventDetails` declares `startDate`/`endDate`, and the UI sends a single `dates` string, so those two decode empty — but `decodeEmailCopyEventDetails` only errors on a missing `eventName`, and the generated copy is correct in practice. I confirmed this live: the staged draft carries "MCP Dev Summit Nairobi - Join the Developer Community" and the body reads correctly, because the prompt gets the event name and location. So it is a real gap in prompt richness rather than "Date TBD" on every email. Adding a `dates` alias would not help — upstream reads two distinct fields, not one string. Worth a follow-up to either split the scrape into start/end or teach the decoder the combined form; I did not want to change the upstream contract inside this PR.

Full suites pass with no dev bypasses: **1369 app, 2233 server**, format/lint/headers clean, rebased on `main`.

mrautela365:
Correction on the `dates` question, @dealako — I got this wrong in my last comment and Copilot was right.

I said it was "a real gap in prompt richness rather than 'Date TBD' on every email." I checked the persisted brief afterwards and it is exactly "Date TBD" on every email:

```
dates       = "19-20 November 2026"
startDate   = ""
endDate     = ""
```

`emailCopyEventDetails` decodes only `startDate`/`endDate` — paid-platform config fields that a UI-written brief never sets — so `formatEventDates` hit its both-empty branch and handed the model `Date TBD` while the real dates sat in the brief beside it. What misled me is that the copy still *reads* correctly: the system prompt says never to invent dates, so the model dutifully omitted them rather than producing anything obviously broken. I took plausible output as evidence the input was fine.

Fixed in campaign-service `96f4358d` (PR #194): the decoder now reads the combined `dates` string as a fallback, with the structured pair still winning when set. Live check after the fix — the generated subject is now:

> MCP Dev Summit Nairobi: Nov 19-20, 2026 - Join Us!

I was also wrong that "adding a `dates` alias would not help." The fix is on the campaign-service side rather than an alias in this PR, so #1932 is unchanged by it, but the underlying claim — that one string can't feed a two-field decoder — was just untrue once the decoder learns the combined form.

Table-driven test covers combined-only, structured-wins, trimming, whitespace-only, and neither-present; mutation-verified (removing the fallback fails it). All 30 packages pass, golangci-lint clean.

dealako (changes requested):
Strong round, @mrautela365 — this is the one where every carried-over item actually closed, and the tests are notably better (the alias/label tests now assert literals instead of self-referential `toBe(details['name'])`, which would have passed against the very regression they guard). Real progress.

## ✅ Resolved this round (all six prior items)

- **B1 (staging ack-as-done)** — `onStageEmailSend` no longer sets `done` on the bare ack; it requires `outcome.jobId` (honest `error` otherwise) and delegates to the new `pollStagingJob`, which holds `emailStaging` at `'staging'` until `getCreateResult` settles. New tests assert `'staging'` after the ack, `'error'` on a failed job, and `'done'` on success.
- **B3 (unowned-brief block)** — `ensureEmailBriefId` now derives `ownershipKey`, reads `knownBriefIds`, and forwards `owned.id`/`owned.etag`/`absence` so a shared `(project, event)` row is replaced rather than refused, and it gates caching on `persisted.approved` (returns `''` when not approved). Both covered by new tests.
- **B4 (foundation-switch cache leak)** — the switch handler now calls `resetEmailBriefDerivedState()`, clearing `emailBriefId`/audience/copy. Covered by "drops the brief-derived state when the foundation changes."
- **M1 (503 → generic retry)** — both catches now call `upstreamMessageOr`, which preserves the upstream message for 503 + any 4xx and keeps other 5xx generic (so a raw `panic…` 500 never reaches the client). Both cases tested.
- **M2 (spec locked in ack-as-done)** — the spec now asserts `'staging'` on the bare ack, plus the two terminal-outcome tests.
- **N1** — the `emailBriefId` JSDoc now sits on its field.

## 🔴 Blocking: 2 — both introduced by this round's fixes

1. **The staging poll writes a terminal state onto a superseded context** (`campaigns.component.ts` `pollStagingJob`). The poll is stored in `stagingJobSubscription` but only unsubscribed at the **start of the next** `pollStagingJob` call. `resetEmailBriefDerivedState` — invoked by the foundation-switch handler, Start Over, and program switches — sets `emailStaging` to `'idle'` but never unsubscribes, and there's no generation guard. `getCreateResult` polls up to ~5 minutes, and both a foundation switch and Start Over are reachable in that window. When the poll finally emits, `next()`/`error()` write `emailStaging = 'done'` ("Draft created in HubSpot…") / `'error'` **unconditionally** under the new context. Only `complete()` guards on `emailStaging() === 'staging'` — which shows the staleness hazard was recognized but the terminal branches were left unguarded. This reopens the exact false-success / cross-context class B1 and B4 just closed. Cursor flagged the same ("Staging poll leaks across reset"). **Fix:** unsubscribe (and null) `stagingJobSubscription` inside `resetEmailBriefDerivedState`, or capture a staging-generation counter at poll start and drop `next/error/complete` writes when it no longer matches — mirroring the `emailSearchGeneration`/`ownershipGeneration` pattern already in this file.

2. **Email persist replaces the paid-owned brief row and never writes the new etag back** (`campaigns.component.ts` `ensureEmailBriefId`, ~1899). To fix B3, the helper now calls `persistBrief(brief, projectSlug, owned.id, owned.etag, …)` where `brief` is the **email** brief and `owned` is the paid side's `knownBriefIds` entry. That's a full replace of the shared `(project, event)` row with the email payload, and the paid brief's ad copy / keywords / targeting are not on the email brief — so reusing a paid-owned row overwrites paid content. Separately, unlike the paid save path (which writes back at ~1556), this path only sets `emailBriefId` and returns — the new server etag is **not** recorded in `knownBriefIds`, so the paid side keeps its now-stale validator and its next save is refused (412 / forced `etag_fallback`). Cursor flagged this as High ("Email persist wipes paid brief"). **Fix:** don't overwrite a paid-owned row with an email payload — either load/merge the existing brief so paid fields survive, or scope email persistence to a channel-aware key; and if a replace does happen, write the returned `{id, etag}` back to `knownBriefIds` so the paid validator stays fresh.

## 🟡 Minor: 2

1. **No regression test for the stale-write path.** The new suite covers fail-poll, success-poll, and foundation-switch-clears-state, but nothing drives `getCreateResult` to emit **after** a reset/foundation-switch and asserts the late terminal emission is dropped — so the blocker-1 path is uncovered. Add a test that stages, switches foundation (or Start Over) before the poll emits, then emits a terminal outcome and asserts `emailStaging()` stays `'idle'`.

2. **`pollStagingJob` comments and branches model a `getCreateResult` contract that doesn't exist** (also covers Cursor's "Poll errors hide job failure"). `getCreateResult` is `pollJobStatus().pipe(last(), map(…))`: it emits exactly one terminal outcome then completes, or **throws** on timeout/not_found/error. So the `next()` `if (!outcome) return;` "still-running tick" branch is unreachable, the outer `take(MAX_POLLS)` bounds a stream that emits at most once (it already self-bounds), and the `complete()` "Staging is taking longer than expected" message never shows — a real poll-exhaustion timeout throws into the `error()` handler, which prints the generic "Lost track of the staging job" and discards the thrown message. Functional outcome is still safe (both route the operator to check HubSpot), but the comments misstate the polling contract on the exact flow whose prior bug was "trusting the ack," and the intended timeout copy is dead. Align the comments/branches with the single-emission-or-throw contract and surface the thrown message in the error handler.

## Bot reconciliation

- **Cursor "wipes paid brief" (High)** and **"poll leaks across reset" (Medium)** — agree; these are blockers 2 and 1.
- **Cursor "poll errors hide job failure" (Medium)** — agree in part; folded into minor 2 (the error handler ignores the thrown message).
- **Copilot "buildAudience suppresses the UNCONFIRMED 500"** — disagree. The UNCONFIRMED/partial audience is returned as a **202 success** with `status: 'building'` and `inclusion_summary` in the body, rendered verbatim in the UI; it is not a thrown 500, so `upstreamMessageOr` (catch-only) never touches it.
- **Copilot "Draft created on a partial `platformResults`"** — disagree for the email path: staging is single-platform (`hubspot`), and a partial requires ≥2 platforms, so the success branch's `errors.length` check is adequate here. Worth a comment noting the single-platform assumption so a future multi-platform email path revisits it.
- **Copilot "concurrent Build + Generate first-save race"** — pre-existing (predates these commits), now slightly more reachable via the shared `ensureEmailBriefId`. Not new this round, but memoizing the in-flight persist promise would close it.
- **Copilot "UTM shown but not forwarded on staging" (previously missed, unchanged code)** — carry-over to confirm; not introduced here.

🔴 **Needs changes before approval.** The carried-over work is genuinely done — but the B3 and B1 fixes each introduced a new blocker (paid-row overwrite + stale etag, and the poll writing under a superseded context). Both are contained and have clear fixes.

dealako:
## Follow-up review (round 3) — @mrautela365

Best round yet — **all six carried-over items are resolved**, and the test hardening is real: the alias/label tests now assert literals instead of self-referential `toBe(details['name'])` (which would have passed against the exact regression they guard). 👏

- ✅ **B1** staging now polls `getCreateResult` and holds `'staging'` until the job settles (with fail/success tests) · ✅ **B3** `ensureEmailBriefId` reuses the owned row and gates on `approved` · ✅ **B4** the foundation switch clears brief-derived state · ✅ **M1** `upstreamMessageOr` preserves 503/4xx and masks other 5xx · ✅ **M2** spec asserts `'staging'` · ✅ **N1** JSDoc moved.

Two new blockers were introduced by the B3 and B1 fixes themselves:

- 🔴 **Blocking: 2** — (1) `pollStagingJob` is only unsubscribed on the *next* Stage, so a foundation switch / Start Over during the ~5-min poll window leaves it live; a late emission then writes `'done'` ("Draft created in HubSpot") / `'error'` under the superseded context — no generation guard, reopening the B1/B4 class (Cursor agrees). (2) `ensureEmailBriefId` now replaces the shared paid-owned `(project,event)` row with the **email** payload (no ad copy/keywords/targeting) and never writes the new etag back to `knownBriefIds`, so paid content is overwritten and the paid side's next save hits a stale-etag 412 (Cursor High agrees).
- 🟡 **Minor: 2** — no regression test drives a poll emission *after* a reset/switch to prove the late write is dropped; and `pollStagingJob`'s comments/branches model a per-tick/poll-exhaustion contract `getCreateResult` doesn't have (single-emission-or-throw), leaving dead branches and a never-shown timeout message (folds in Cursor's "poll errors hide job failure").

I disagreed with two Copilot items: the UNCONFIRMED audience is a **202 success** (`status: building` + `inclusion_summary`), not a suppressed 500; and the single-platform email staging makes the `errors.length` success check adequate (a partial needs ≥2 platforms). Detail + fixes in the formal review: https://github.com/linuxfoundation/lfx-self-serve/pull/1932#pullrequestreview-5060116024

🔴 **Needs changes before approval** — both new blockers are contained and have clear fixes (unsubscribe/generation-guard the poll; don't overwrite a paid row with an email payload, and write the etag back).

PR #1944, opened 2026-08-29
fix(profile): explain and recover from a blank linux-email forward

### LINKED ISSUE #1935: [Bug] Linux.com "Forward to" select is blank and never explains why
## Summary

On **Profile → Identities**, the Linux.com panel's "Forward to" dropdown renders with options but nothing selected, and the user is given no indication why. It happens whenever the server can't read the current forwarding target, and once the one-shot re-authorization attempt fails the state is unrecoverable for the rest of the browser session.

## Steps to Reproduce

1. Sign in with an account that has a claimed Linux.com alias and a primary email different from that alias.
2. Go to `/profile/identities` and scroll to the **Linux.com email** section.
3. Observe the "Forward to" select in the claimed/edit form.

`GET /api/profile/linux-email` returns (values replaced with placeholders):

```json
{
  "state": "claimed",
  "domain": "example.org",
  "alias": "jane-doe",
  "email": "jane-doe@example.org",
  "forwardTo": null,
  "primaryEmail": "jane.doe@example.com",
  "forwardAuthRequired": true,
  "authorizeUrl": "/api/profile/auth/start?returnTo=..."
}
```

## Expected Behavior

Either the current forwarding target is loaded and shown as the selected option, or — if it can't be read — the UI says so and offers a way to re-authorize.

## Actual Behavior

The select renders with valid options (the primary email is a valid choice) but shows the placeholder "Select a verified email", as if nothing is configured. No error, no explanation. Re-authorization silently does not recover.

## Root Cause

Three separate defects compound:

**1. Blank select with no explanation.** With no Flow C management token in session, the BFF returns `forwardTo: null` and `forwardAuthRequired: true` (`apps/lfx-one/src/server/controllers/profile.controller.ts:665-684`). `applyFormDefaults` then deliberately patches the control with `''` rather than guessing the primary email, since a guessed value could overwrite the real forward on Save (`apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts:303-311`). PrimeNG finds no option matching `''` and falls back to the placeholder. The reasoning behind patching `''` is sound; the problem is that the resulting state is presented as a normal, empty dropdown.

**2. Re-auth failure is masked permanently.** `maybeReauthForForward` redirects once to `authorizeUrl` and sets a `sessionStorage` guard, which is cleared only when `forwardAuthRequired` comes back falsy (`profile-linux-email.component.ts:263-276`). Observed in a live session: the guard was set while `forwardAuthRequired` was still `true` — i.e. the round-trip ran, did not produce a usable token, and retry is now suppressed for the rest of the session. The guard correctly prevents a redirect loop, but leaves no way forward.

**3. The underlying error never surfaces.** The Flow C callback redirects with `?error=<reason>` on every failure path (`profile.controller.ts:1566-1626`). `ProfileIdentitiesComponent.ngOnInit` strips that param via `clearQueryParams()`/`history.replaceState` and toasts from an `errorMap` that omits every Flow C code — `profile_auth_failed`, `user_mismatch`, `login_session_invalid`, `no_code` (`apps/lfx-one/src/app/modules/profile/identities/profile-identities.component.ts:94-106`). They all degrade to the generic "An error occurred. Please try again.", so neither the user nor a developer inspecting the URL can tell what failed.

Not a redirect-URI misconfiguration: the default `${PCC_BASE_URL}/passwordless/callback` has a registered handler (`apps/lfx-one/src/server/server.ts:371`).

## Proposed Fix

1. When `state === 'claimed'` and `forwardTo === null`, disable the select and render an explanation with a "Re-authorize" action pointing at `authorizeUrl`, instead of an enabled select showing a placeholder — edit form at `profile-linux-email.component.html:170-213`.
2. Add the Flow C error codes to the `errorMap` in `profile-identities.component.ts:95-103` so callback failures report what actually went wrong.
3. Make the re-auth guard recoverable — allow an explicit user-initiated retry rather than one-shot-and-stuck.
4. Add `profile-linux-email.component.spec.ts` (none exists today) covering `forwardTo: null` + `forwardAuthRequired: true`, and the case where the primary email is itself the alias.

## Out of Scope

Why the Flow C token exchange fails in the first place is not yet diagnosed — it needs the `profile_auth_callback` server log line, which the browser can't reveal. Tracking that separately; this issue covers the UI presenting the failure as a silently empty dropdown.

fayazg:
## Summary

- The Linux.com "Forward to" dropdown on Profile → Identities rendered blank with the
  "Select a verified email" placeholder and no explanation when the server couldn't read
  the current forward target — indistinguishable from a broken control.
- Root cause: reading a claimed alias's forward target requires a Flow C (Auth0
  Management-audience) re-auth token. Without one, the server returns `forwardTo: null` +
  `forwardAuthRequired: true`, and the client had no UI for that state.
- Added a re-auth panel (mirroring the existing `service_unavailable` panel) that explains
  the state and offers a "Re-authorize" action, gated on `forwardAuthRequired` rather than
  `forwardTo === null` — so the existing "forward genuinely unset" recovery path (after a
  partial claim) still shows the select instead of being hidden.
- Split the Flow C callback's `?error=<code>` handling into two owned maps
  (`PROFILE_AUTH_ERROR_MESSAGES` on the layout, `IDENTITY_LINK_ERROR_MESSAGES` on the
  identities component) so each code produces exactly one specific toast instead of a
  generic message or a double toast.
- Added/extended specs across the touched components and an e2e spec for the new re-auth
  panel and error-toast behavior.

## Trade-offs (documented, not fixed in this PR)

- `invalid_state`/`no_code` Flow C error codes produce no toast and leave `?error=...`
  stuck in the URL when the return route isn't `/profile/identities` (pre-existing gap,
  not introduced here — needs a design decision on fix shape, deferred).
- No `profile-linux-email-robust.spec.ts` structural e2e companion yet (pre-existing gap
  in this module's e2e coverage, not introduced here).

Fixes #1935

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
@@ -77,7 +77,8 @@ export class ProfileLinuxEmailComponent {
   // harmful here (it renders blank), so the template swaps it for a re-auth panel instead.
   public forwardReauthRequired = computed(() => this.state() === 'claimed' && !!this.data()?.forwardAuthRequired);
 
-  // Absent when Flow C is unconfigured server-side — gates whether the re-auth button can render.
+  // Defensive: the BFF always pairs authorizeUrl with forwardAuthRequired, so this should never
+  // be empty while forwardReauthRequired() is true. Guards against that contract drifting.
   public forwardAuthorizeUrl = computed(() => this.data()?.authorizeUrl ?? '');
 
   // Forward is genuinely unset (not unreadable) — the FORWARD_SET_FAILED recovery path and the
--- a/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
@@ -75,8 +75,8 @@ export class ProfileLinuxEmailComponent {
   // harmful here (it renders blank), so the template swaps it for a re-auth panel instead.
   public forwardReauthRequired = computed(() => this.state() === 'claimed' && !!this.data()?.forwardAuthRequired);
 
-  // Defensive: the BFF always pairs authorizeUrl with forwardAuthRequired, so this should never
-  // be empty while forwardReauthRequired() is true. Guards against that contract drifting.
+  // Can be empty while forwardReauthRequired() is true — the BFF omits authorizeUrl when Flow C
+  // itself is unconfigured, since there's nowhere to send the client to re-authorize.
   public forwardAuthorizeUrl = computed(() => this.data()?.authorizeUrl ?? '');
 
   // Forward is genuinely unset (not unreadable) — the FORWARD_SET_FAILED recovery path and the
--- a/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
@@ -1,8 +1,6 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-// Generated with [Claude Code](https://claude.ai/code)
-
 import { isPlatformBrowser } from '@angular/common';
 import { Component, computed, inject, input, PLATFORM_ID, Signal, signal } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
--- a/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
@@ -13,6 +13,7 @@ import { CardComponent } from '@components/card/card.component';
 import { InputTextComponent } from '@components/input-text/input-text.component';
 import { MessageComponent } from '@components/message/message.component';
 import { SelectComponent } from '@components/select/select.component';
+import { LINUX_EMAIL_FORWARD_REAUTH_KEY } from '@lfx-one/shared/constants';
 import { EnrichedIdentity, LinuxAliasData, LinuxForwardOption } from '@lfx-one/shared/interfaces';
 import { linuxAliasValidator } from '@lfx-one/shared/validators';
 import { UserService } from '@services/user.service';
@@ -37,7 +38,7 @@ export class ProfileLinuxEmailComponent {
   public readonly identities = input<EnrichedIdentity[]>([]);
 
   // One-shot guard (sessionStorage key) so a tokenless re-auth round-trip can't loop.
-  private readonly reauthFlagKey = 'linux-email:forward-reauth-attempted';
+  private readonly reauthFlagKey = LINUX_EMAIL_FORWARD_REAUTH_KEY;
 
   // Set on FORWARD_SET_FAILED so the recovery toast can be shown once the refreshed
   // alias state is known, instead of assuming the refetch lands on 'claimed'.
--- a/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
@@ -72,6 +72,17 @@ export class ProfileLinuxEmailComponent {
   public domain = computed(() => this.data()?.domain ?? '');
   public email = computed(() => this.data()?.email ?? null);
 
+  // The forward target couldn't be read (no Flow C token in session) — the select is actively
+  // harmful here (it renders blank), so the template swaps it for a re-auth panel instead.
+  public forwardReauthRequired = computed(() => this.state() === 'claimed' && !!this.data()?.forwardAuthRequired);
+
+  // Absent when Flow C is unconfigured server-side — gates whether the re-auth button can render.
+  public forwardAuthorizeUrl = computed(() => this.data()?.authorizeUrl ?? '');
+
+  // Forward is genuinely unset (not unreadable) — the FORWARD_SET_FAILED recovery path and the
+  // impersonation read-only path both land here. Drives an honest hint on the still-usable select.
+  public forwardNotSet = computed(() => this.state() === 'claimed' && !this.data()?.forwardTo && !this.data()?.forwardAuthRequired);
+
   // True only when the user has changed the forward-to selection from its saved value.
   public forwardDirty: Signal<boolean> = this.initForwardDirty();
 
@@ -200,6 +211,21 @@ export class ProfileLinuxEmailComponent {
     // it must not consume a stale pending flag from an earlier claim() failure.
     this.pendingForwardRecoveryToast = false;
     this.refresh.next();
+  }
+
+  /**
+   * Manual recovery for a still-tokenless return. Deliberately does NOT clear reauthFlagKey —
+   * doing so would let maybeReauthForForward fire a second automatic redirect on a still-
+   * tokenless return. This button is the recovery path and stays available on every reload,
+   * which is what makes the one-shot guard recoverable without reopening the redirect loop it
+   * exists to prevent.
+   */
+  public reauthorizeForward(): void {
+    if (!isPlatformBrowser(this.platformId)) return;
+    const url = this.forwardAuthorizeUrl();
+    if (url) {
+      window.location.href = url;
+    }
   }
 
   // Private methods
--- a/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
@@ -282,11 +282,11 @@ export class ProfileLinuxEmailComponent {
   }
 
   /**
-   * A claimed alias always has a forward target, but the server can only read it
-   * with a Flow C management token. When that token is absent the server flags
-   * `forwardAuthRequired`; redirect once to load the real target instead of
-   * silently showing the primary email. The one-shot guard prevents a loop if the
-   * round-trip returns still tokenless (or the user cancels).
+   * A claimed alias's forward target can only be read with a Flow C management
+   * token. When that token is absent the server flags `forwardAuthRequired`;
+   * redirect once to load the real target instead of silently showing the
+   * primary email. The one-shot guard prevents a loop if the round-trip returns
+   * still tokenless (or the user cancels).
    */
   private maybeReauthForForward(alias: LinuxAliasData | null): void {
     if (!isPlatformBrowser(this.platformId)) return;
--- a/apps/lfx-one/src/server/controllers/profile.controller.ts
+++ b/apps/lfx-one/src/server/controllers/profile.controller.ts
@@ -674,8 +674,8 @@ export class ProfileController {
           return;
         }
 
-        // Needs the Flow C management token to read the forward; ask the client to re-authorize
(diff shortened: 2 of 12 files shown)


apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
```diff
@@ -72,6 +71,18 @@ export class ProfileLinuxEmailComponent {
   public domain = computed(() => this.data()?.domain ?? '');
   public email = computed(() => this.data()?.email ?? null);
 
+  // The forward target couldn't be read (no Flow C token in session) — the select is actively
+  // harmful here (it renders blank), so the template swaps it for a re-auth panel instead.
+  public forwardReauthRequired = computed(() => this.state() === 'claimed' && !!this.data()?.forwardAuthRequired);
```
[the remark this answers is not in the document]
fayazg: Good catch, and you're right about the root cause: `getLinuxAlias()` only set `forwardAuthRequired` when Flow C was also configured, so an unconfigured server produced a claimed alias with `forwardTo: null` and no flag — landing on the blank-select path with a 501-returning Save.

Fixed in bcd0ac7b: `forwardAuthRequired` is now true whenever the management token is missing, independent of `isProfileAuthConfigured()`; `authorizeUrl` stays conditional on Flow C actually being configured (omitted otherwise, which the client/template already render correctly via the existing dead-end copy branch). Updated the now-stale "can't happen" comments in the controller, component, template, e2e helper, and shared interface docstring to match.

fayazg:
## Review Feedback Addressed

Commit: bcd0ac7b404993e9d7d153f69839926bbda77fce

### Changes Made
- **profile.controller.ts**: `forwardAuthRequired` is now set whenever the management token is missing, independent of `isProfileAuthConfigured()`; `authorizeUrl` stays conditional on Flow C actually being configured (per copilot-pull-request-reviewer). Previously an unconfigured server produced a claimed alias with `forwardTo: null` and no flag, landing on the blank-select fallback instead of the re-auth panel.
- **profile-linux-email.component.ts / .html**, **e2e helper**, **linux-email.interface.ts**: corrected comments/docstrings that described the "no authorizeUrl" shape as unreachable — it's now the real shape emitted when Flow C is unconfigured.

### Threads Resolved

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -666,22 +666,21 @@ export class ProfileController {
...
-        // Suppress the re-auth prompt during impersonation — it is a write-path the impersonator
-        // must not trigger against their own session while viewing the target read-only.
-        const forwardAuthRequired = !isImpersonating(req) && !managementToken && this.profileAuthService.isProfileAuthConfigured();
+        // The forward is unreadable without a Flow C management token — true regardless of whether
+        // Flow C is even configured, so the client can't mistake "unreadable" for "unset". Suppressed
+        // under impersonation — the re-auth would rewrite the impersonator's own session.
+        const forwardAuthRequired = !isImpersonating(req) && !managementToken;
+        // Re-authorization is only offered when Flow C is actually configured; otherwise there's no
+        // authorizeUrl to send the client to and it falls back to the "try again later" copy.
+        const canReauth = forwardAuthRequired && this.profileAuthService.isProfileAuthConfigured();
```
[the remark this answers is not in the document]
fayazg: Done — added a `ProfileController.getLinuxAlias` suite covering exactly the regression surface flagged here.

See commit 559a19c05: four new cases prove a missing management token always sets `forwardAuthRequired: true` regardless of Flow C config, that `authorizeUrl` only appears when Flow C is actually configured, that a present token skips the flag and reads the live forward target via `getForward`, and that impersonation suppresses the flag entirely.

fayazg:
## Review Feedback Addressed

Commit: 559a19c052eee920ef8b38ca337adc5c94347e2a

### Changes Made
- **apps/lfx-one/src/server/controllers/profile.controller.spec.ts**: added a `ProfileController.getLinuxAlias` test suite proving a missing management token always emits `forwardAuthRequired: true`, `authorizeUrl` appears only when Flow C is configured, a present token skips the flag and reads the live forward target, and impersonation suppresses the flag entirely (per copilot-pull-request-reviewer[bot])

### Threads Resolved

apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
```diff
@@ -254,11 +280,11 @@ export class ProfileLinuxEmailComponent {
...
-   * `forwardAuthRequired`; redirect once to load the real target instead of
-   * silently showing the primary email. The one-shot guard prevents a loop if the
-   * round-trip returns still tokenless (or the user cancels).
+   * A claimed alias's forward target can only be read with a Flow C management
+   * token. When that token is absent the server flags `forwardAuthRequired`;
+   * redirect once to load the real target instead of silently showing the
+   * primary email. The one-shot guard prevents a loop if the round-trip returns
+   * still tokenless (or the user cancels).
    */
   private maybeReauthForForward(alias: LinuxAliasData | null): void {
```
dealako, line 289: [minor] Redirect-guard reset branch is untested

Issue: `maybeReauthForForward` clears the one-shot guard (`sessionStorage.removeItem(this.reauthFlagKey)` at line 295) when a reload returns without `forwardAuthRequired`. The new spec latches the guard (asserts `getItem(REAUTH_FLAG_KEY)).toBe('1')` at spec lines 130/142) but never asserts it gets cleared.
Proof: the removeItem in the `!alias.forwardAuthRequired` early return has no corresponding assertion — grep of the spec finds only `.toBe('1')`, no null/cleared assertion.
Why it matters: this reset is the half of the one-shot guard that keeps re-auth recoverable. A future edit dropping the `removeItem` would strand users needing a fresh redirect, and no test would catch it.
Fix: add a case that pre-sets `REAUTH_FLAG_KEY`, loads an alias with `forwardAuthRequired` absent, and asserts `sessionStorage.getItem(REAUTH_FLAG_KEY)` is null with no redirect fired.

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -666,22 +666,21 @@ export class ProfileController {
         const managementToken = isImpersonating(req) ? null : this.profileAuthService.getManagementToken(req);
         const forward = managementToken ? await this.forwardsService.getForward(req, managementToken, domain) : null;
 
-        // A claimed alias always has a forward target. If we had a management token but the read
-        // still came back null, the forwards-service was unreachable (timeout/no-responder) — degrade
-        // to service_unavailable (retry panel) rather than emitting a claimed state with a null
-        // forward, which the client could misread and overwrite on Save.
+        // Null with a token means forwards-service was unreachable — degrade to service_unavailable
+        // instead of emitting a claimed state with a null forward the client could overwrite on Save.
         if (managementToken && forward === null) {
```
dealako, line 671: [nit] `service_unavailable` degrade (token present + null forward) not covered

Issue: when a management token IS present but `getForward` resolves null, this branch degrades to `service_unavailable` rather than emitting a claimed state with a null forward. The new `getLinuxAlias` suite covers no-authorizeUrl, configured-but-no-token, token-present-with-forward, and impersonation, but not this branch.
Proof: profile.controller.ts:671 (`if (managementToken && forward === null)`); the spec's only `getForward.mockResolvedValue(...)` uses a real target (spec:315) and there are zero `service_unavailable` matches in the spec.
Why it matters: this degrade guards against emitting a claimed state with a null forward the client could overwrite on Save; a regression would ship silently.
Fix: add a controller test with `isProfileAuthConfigured` true, `getManagementToken` returning a token, `getForward.mockResolvedValue(null)`; assert `linuxAliasState('service_unavailable', ...)` and no claimed payload.

apps/lfx-one/src/app/modules/profile/linux-email/profile-linux-email.component.ts
```diff
@@ -254,11 +280,11 @@ export class ProfileLinuxEmailComponent {
...
 
   /**
-   * A claimed alias always has a forward target, but the server can only read it
-   * with a Flow C management token. When that token is absent the server flags
-   * `forwardAuthRequired`; redirect once to load the real target instead of
-   * silently showing the primary email. The one-shot guard prevents a loop if the
-   * round-trip returns still tokenless (or the user cancels).
+   * A claimed alias's forward target can only be read with a Flow C management
+   * token. When that token is absent the server flags `forwardAuthRequired`;
+   * redirect once to load the real target instead of silently showing the
```
dealako, line 285: [nit] Docstring omits the now-reachable no-`authorizeUrl` dead-end path

Issue: the `maybeReauthForForward` docstring describes only the redirect-once behavior and the loop guard. It doesn't mention the branch where `forwardAuthRequired` is true but `authorizeUrl` is absent, where the method no-ops (guard at line ~299) and the template shows the recoverable 'try again later' panel.
Proof: docstring here vs the guard `if (!alias.authorizeUrl || sessionStorage.getItem(this.reauthFlagKey)) return;` at ~299. This branch became reachable once the controller set `forwardAuthRequired` independently of Flow C config (profile.controller.ts:680) while gating `authorizeUrl` behind `canReauth` (line 683).
Why it matters: a reader trusting the docstring would assume claimed + `forwardAuthRequired` always redirects; the unconfigured-Flow-C case (no redirect, panel shown) is a distinct behavior worth one sentence.
Fix: add a sentence noting that when `authorizeUrl` is absent the method no-ops and the recoverable re-auth panel is shown instead of redirecting.

dealako (approved):
Approving with minor comments. Full synthesis is in the summary comment on the PR. No blocking, security, or data-privacy findings across the three review dimensions; one minor test-coverage gap and two nits are inline below.

dealako:
Hey @fayazg 👋 — thanks for a well-scoped fix. The core move here is exactly right: gating the recovery UI on `forwardAuthRequired` instead of `forwardTo === null` cleanly separates "we can't read the target" (re-auth panel) from "the target is genuinely unset" (normal select), and decoupling `forwardAuthRequired` from Flow C config while keeping `authorizeUrl` gated behind `canReauth` makes the unconfigured server a distinguishable, honest state rather than a silent blank. The `Object.hasOwn` + `typeof === 'string'` hardening on the `?error=` lookups, the deliberately non-clearing one-shot redirect guard, and the same-origin relative `authorizeUrl` (re-validated server-side via the return-path allowlist) all hold up under tracing. Solid.

I ran a three-dimension parallel review (security/privacy, correctness/perf/tests, style/docs). No blocking, security, or data-privacy findings. Test fixtures use RFC 2606 reserved domains and logging only emits `state`/allowlisted return paths — no PII leaks. Two small test-coverage gaps and one docstring nit below.

### Issue count
- 🔴 Blocking: 0
- 🟡 Minor: 1 — redirect-guard **reset** branch (`removeItem`) is untested
- ⚪ Nit: 2 — `service_unavailable` degrade (token present + `forward === null`) untested; `maybeReauthForForward` docstring omits the no-`authorizeUrl` dead-end path
- ❔ Question: 0

### Bot reconciliation
- **Copilot** raised two points (unconfigured Flow C leaving `forwardAuthRequired` false → blank select; and the server branch lacking controller coverage). Both are already resolved in this PR — `bcd0ac7b` decoupled the flag from config, and `559a19c05` added the `getLinuxAlias` suite. Agreed and confirmed fixed. My minor finding is the one adjacent branch that suite doesn't yet cover.
- **Cursor Bugbot** posted a medium-risk summary with no specific open finding; nothing to reconcile.

### Decision
✅ **Approved with minor comments** — the two coverage gaps and the docstring nit are non-blocking. Worth a quick follow-up since both untested branches are the recoverable/degrade halves that would fail silently if regressed, but nothing here should hold the merge.

PR #1945, opened 2026-08-29
feat(campaigns): let the operator choose which email they are writing

### LINKED ISSUE #1941: campaigns UI: no email-type selector, so every send generates registration-push copy
Depends on #1940 — the stage definitions must exist upstream before the UI can offer them.

## The gap

The Implement tab has one button, "Generate email content", and no way to say what kind of email is wanted. Every send generates registration-push copy, so an operator writing a post-event thank-you, a CFP launch or a final countdown gets the wrong tone, subject shape and CTA and has to rewrite it by hand.

## Scope

- Add the **12 email types and their 6-stage mapping** to `@lfx-one/shared/constants`, with the keyword matchers the reference carries (they are also what #TBD needs for clone ranking):

```
CFP Launch · Registration Launch · Co-Located Events + CFP Reminder
DEI & Travel Fund · Schedule Announcement · Main Registration Push
Final Countdown · Event Week · Thank You + Survey
Content & Recordings Release · Next Event CFP Teaser · Community Nurture
```

- Selector on the Implement tab, above "Generate email content"
- Pass the selected stage through the BFF as a **thin proxy** — no stage logic in the Express layer, per the PR #2 architecture. The BFF forwards; campaign-service decides.

## Behaviour to get right

- **Changing the type after generating must not silently keep the old copy.** `onGenerateEmailCopy` already clears `emailCopy` before regenerating (added in #1932 after a failed regeneration left stale copy stageable); the same must hold when only the type changes.
- Default to **Main Registration Push** so the current behaviour is what an operator gets without touching the selector.
- The type is part of what gets staged, so it belongs in the brief's persisted state rather than being a transient UI value — otherwise a reload silently reverts to the default while the generated copy still reflects the old type.

## Acceptance

- Selector renders with all 12 types; the resulting stage reaches campaign-service
- A test asserting the request carries the selected stage, not merely that one was rendered
- Switching type after generation clears the previous copy

Parent epic: #1697

mrautela365:
The operator-facing half of stage-aware email. **Depends on lfx-v2-campaign-service#195**, which must deploy first — until it does, the `stage` this sends is accepted and ignored.

Branched from `main`, so it does not carry #1932's commits; the two touch different parts of `campaigns.component` and are independently mergeable.

## The gap

One button, "Generate email content", and no way to say what kind of email is wanted. An operator writing a post-event thank-you, a CFP launch or a final countdown got registration-push copy and rewrote it by hand.

## What this adds

The twelve email types an event programme sends, each mapped to the stage campaign-service generates from. Several share a stage on purpose: a CFP launch and a co-located CFP reminder differ in *when* they are sent, not in how the copy is written.

**"Thank You + Survey" is where a post-event survey ask lives** — an email type, not a separate campaign, program type or audience. The survey is a CTA inside a post-event email to the same registrants.

The **type** is stored rather than the stage, because several types map to one stage: storing the stage would lose which one the operator chose and a reload would show a different label than they picked.

## Two behaviours worth reviewing

**Changing the type clears generated copy.** The copy on screen was written for the previous stage — subject, tone and CTA all belong to it — and `onStageEmailSend` reads `emailCopy()` unconditionally, so leaving it makes mismatched copy stageable under the new label. Same defect a failed regeneration had before #1932 cleared it on retry.

**The BFF forwards `stage` without validating it.** The valid set is campaign-service's enum; duplicating it here would give two sources of truth that drift. An unknown stage returns upstream's 400 naming the valid values, which is a better answer than anything this layer could invent.

## A bug the browser caught that the tests could not

The select bound `[value]` on the element. A native `<select>` ignores that when applied before its options exist, so the control fell back to the **first** option — CFP Launch — while the signal held the Main Registration Push default. Every operator who never touched the selector would have silently generated CFP copy, invisible because CFP copy still reads plausibly.

The existing tests were green against it: they called the handler directly and never mounted the control. Fixed by binding `selected` on the option, with a test that renders it (`9ed750f2`).

## Testing

1373 app + 2233 server tests, no dev bypasses applied. Format, lint and license headers clean.

Mutation-verified: sending the type id instead of the stage fails, not clearing copy on type change fails, removing the `selected` binding fails.

**Verified in the browser through the BFF**, which is what closed the last real gap — `proxyRequestWithResponse` takes query 5th and data 6th, and a body in the wrong slot sends nothing with no type error:

- Selector renders all 12 types, defaulted to Main Registration Push
- Selecting "Thank You + Survey" produced *"Thank You for MCP Dev Summit Nairobi – Recordings Live"* with the survey in the body
- Switching to CFP Launch cleared the post-event copy

## Related

- campaign-service: linuxfoundation/lfx-v2-campaign-service#195 — **deploys first**
- Parent epic: #1697

Closes #1941


mrautela365:
All open threads addressed in `4a251f987`. Three findings, all the same shape — work outliving the context it began in.

**The email persist never recorded ownership.** `ensureEmailBriefId` *read* `knownBriefIds` but never wrote it, and `emailBriefId` — the only place the id was cached — is cleared by `resetEmailBriefDerivedState`. So the next save after a Proceed found no owned row and **created a second brief** for a (project, event) that already had one, while the paid path, reading the same cache, went on addressing the first. Two rows for one event, each believed to be the only one. It now writes the cache in the same shape the paid save does.

A comment nearby asserted "this session has established no ownership of an email row" — true when written, falsified by this change. Corrected rather than left, since the next reader would have trusted it.

**A stage change did not invalidate an in-flight generate.** `onSelectEmailType` cleared the copy while a request was still running and left the selector usable, so the older response repopulated the panel with the **previous stage's** copy under the new stage's label. That copy reads plausibly, is simply the wrong kind of email, and is what `onStageEmailSend` would clone. A generation counter now separates the two responses.

**A reset relabelled the staging poll instead of cancelling it.** Setting `emailStaging` to `idle` left the subscription alive, so a job settling after a new brief or a foundation switch still wrote `done` — announcing a HubSpot draft belonging to the previous brief as though it were this one's.

## On the verification

Two things worth flagging about the mutation testing, because they changed what shipped:

- The staging fix's first mutation **survived** — nothing covered that path at all. The test was added because of it, not alongside it.
- The copy fix's first mutation also survived: an earlier `isCurrent()` check masked it. Only removing the counter in `onSelectEmailType` made it fail, which confirmed the counter is what binds and the second check is defence-in-depth.

1376/1376 pass; lint, AOT build and license check clean.

Note on the two remaining findings I did **not** change: `selectedEmailTypeId` not surviving a reload is real but is a feature gap rather than a defect in this diff, and the `stage`-as-body concern was resolved upstream in cs#195 (it is a query parameter, so a body-less POST is valid).

dealako:
Hi @mrautela365 — thanks for the careful writeup and the browser-through-BFF verification; the `[value]`-vs-`selected` native-select catch and the query-vs-body reasoning for `stage` are both well-diagnosed, and delegating enum validation to campaign-service as the single source of truth is the right call.

This is a well-tested change and the FE→BFF→Go stage contract is correct end to end. My one blocker is a residual of the same async-lifetime defect class the rest of this component guards against carefully: the last fix commit (`4a251f987`) closed the poll-unsubscribe and ownership-cache halves of the round-2 bot findings, but the copy/audience generation-guard half is still open, and the "all open threads addressed" note reads as fully resolved when it isn't.

### Issue count

- 🔴 Blocking: 1 issue — reset/foundation-switch never bumps `emailCopyGeneration` (and `onBuildAudience` has no guard at all), so an in-flight generate/build that resolves after a reset or switch writes the previous foundation's copy/audience under the new one, and it is stageable to HubSpot.
- 🟡 Minor: 1 issue — frontend comment on `generateEmailCopy` misdescribes the Go stage contract (says body attribute; it is a query param), contradicting the tested BFF client.
- ⚪ Nit: 0
- ❔ Question: 1 item — `ensureEmailBriefId` has no in-flight dedup; concurrent Generate + Build audience clicks can each take the first-save branch.

- **Agree** with cursor's "Stale copy survives brief reset" (line 2426) and Copilot's line-2418 note — I traced these to the current code and they remain open for the copy/audience path even after `4a251f987`. `resetEmailBriefDerivedState` unsubscribes the staging poll (fixed) but does not bump `emailCopyGeneration`.
- **Agree** with cursor's earlier "Reset leaves staging poll running" and "Email persist skips brief ownership" — those two are genuinely resolved in `4a251f987` (unsubscribe added, `knownBriefIds` now written).
- **Disagree** with Copilot's claim that campaign-service #195 returns 200-with-default for an unknown stage rather than a 400 — that is a campaign-service concern, not decidable in this diff, and does not change the BFF behavior of forwarding without local validation, which is the correct single-source-of-truth choice.

### Decision

🔴 **Needs changes before approval** — one blocking cross-foundation state-leak that is stageable to an external system, plus a misleading contract comment worth correcting while you are in the file.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1877,6 +2394,37 @@ export class CampaignsComponent {
...
+
+  /**
+   * Drop everything derived from the PREVIOUS email brief.
+   *
+   * `emailBriefId` is the one that matters: `ensureEmailBriefId` returns the cached id when it is
+   * set, so leaving it behind makes the audience build, the copy generation and the staged draft
+   * all target the previous event's brief row -- silently, because every call still succeeds. The
+   * rest are cleared for the ordinary reason that they describe a brief that is no longer on screen.
+   */
+  private resetEmailBriefDerivedState(): void {
```
dealako, line 2559: [blocking] Reset and foundation-switch never invalidate an in-flight email copy/audience/staging response

Issue: `resetEmailBriefDerivedState` clears `emailCopy`/`emailAudience`/`emailStaging` and now (good) unsubscribes `stagingJobSubscription`, but it does not bump `emailCopyGeneration`, and there is no generation counter for the audience build at all. The foundation-switch subscription (lines ~929-1006) bumps `briefPersistenceGeneration`, `emailSearchGeneration`, and `briefCampaignsGeneration` but likewise never touches `emailCopyGeneration`.

Proof: `onGenerateEmailCopy` guards its writes with `isCurrent()` keyed on `emailCopyGeneration`, which is incremented only in `onSelectEmailType` (type change) and `onGenerateEmailCopy` itself. Neither the reset here nor the switch handler bumps it, so a generate dispatched for brief/foundation A that resolves after a reset or switch still passes `isCurrent()` and writes A's copy into the signal now labeled B. `onBuildAudience` (line ~1260) writes `this.emailAudience.set(result.audience)` after its await with no generation or slug re-check at all. Because `onStageEmailSend` reads `emailCopy()` unconditionally and `canStageEmail` gates on `emailAudience()?.status === 'built'`, the stale copy/audience is stageable — the exact defect the component documents and guards against for `emailSearch`/`briefCampaigns` in the same switch handler.

This is cursor's round-2 finding "Stale copy survives brief reset" (line 2426) and Copilot's line-2418 note. The last fix commit (4a251f987) addressed the poll-unsubscribe half but not the copy/audience generation half, so the "all open threads addressed" claim does not hold for this thread.

Why it matters: an operator who switches foundations (or selects a new brief) while a build or generate is in flight can end up staging a HubSpot draft carrying the previous event's subject/body/audience — cross-foundation content leak into an external system.

Fix: bump `emailCopyGeneration` inside both `resetEmailBriefDerivedState` and the foundation-switch subscription, add an equivalent counter (or an `activeFoundationSlug() === projectSlug` re-check) for `onBuildAudience` and the staging poll, and re-check it after each await before writing `emailAudience`/`emailCopy`/`emailStaging`. Add a test that resets/switches mid-flight and asserts nothing is written.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1483,6 +1899,107 @@ export class CampaignsComponent {
...
+   *
+   * Email never persists on the Plan tab, but BOTH the audience build and campaign creation are
+   * brief-scoped upstream, so each needs an id. Persisting once and caching it here stops the two
+   * actions writing two briefs for the same event.
+   *
+   * Returns '' when the persist could not produce an id; callers must treat that as a failure
+   * rather than proceeding with an empty path segment.
+   */
+  private async ensureEmailBriefId(brief: CampaignBriefOutput, projectSlug: string): Promise<string> {
+    const known = this.emailBriefId();
```
dealako, line 2097: [question] `ensureEmailBriefId` has no in-flight dedup — concurrent email actions can each issue a first-save persist

Issue: `ensureEmailBriefId` caches the brief id and ownership only after its persist resolves, with no guard against a second concurrent caller.

Proof: `canGenerateEmailCopy` gates only on `emailCopyState() !== 'generating'` (line 791); the Build audience button gates only on `emailAudienceState() === 'building'` (html line 495). These are independent, so while a generate is in flight the Build audience button stays enabled. Both `onGenerateEmailCopy` and `onBuildAudience` call `ensureEmailBriefId`; two invocations started close together both see `emailBriefId() === ''` and an empty `knownBriefIds` entry, so both take the first-save branch. The spec only exercises the sequential case.

Why it matters: campaign-service may refuse the second first-save for an existing (project,event) as unowned-brief-exists, or create a duplicate brief row — the exact failure the ownership cache exists to prevent, reachable by clicking Generate then Build audience quickly.

Fix: confirm reachability given button-disable timing; if reachable, memoize the in-flight persist as a single shared `Promise<string>` per (project,event) key so concurrent callers await the same persist, and add a concurrent-invocation test.

apps/lfx-one/src/app/shared/services/campaign.service.ts
```diff
@@ -126,6 +129,30 @@ export class CampaignService {
...
+      '/api/campaigns/audience/build',
+      {},
+      { params: new HttpParams().set('project', projectSlug).set('brief_id', briefId) }
+    );
+  }
+
+  /**
+   * Generate email copy for a brief. Brief-scoped upstream, so both ids are required.
+   */
+  public generateEmailCopy(projectSlug: string, briefId: string, stage?: CampaignEmailStage): Observable<GenerateEmailCopyResult> {
```
dealako, line 146: [minor] Comment misdescribes the upstream stage contract, contradicting the BFF client

Issue: this comment says `generate-email-copy` "takes project and brief as PATH segments and stage as a body attribute." The Go campaign-service reads `stage` as a QUERY param, not a body attribute — the opposite of what this says.

Proof: `campaign-service.service.ts#generateEmailCopy` passes `stage ? { stage } : undefined` as the 5th (query) argument to `proxyRequestWithResponse` and documents that stage is a query param "because declaring it in the body made the body REQUIRED upstream." `campaign-service.service.spec.ts` asserts `expect(call[4]).toEqual({ stage })` and `expect(call[5]).toBeUndefined()`. The BFF comment is code-backed and tested; this frontend comment is the stale one (likely predating campaign-service#195 moving stage from body to query). Behavior is correct today — this is a doc-accuracy defect only.

Fix: reword to describe only the FE->BFF hop (stage in the request body of `/api/campaigns/email-copy`, project/brief_id as query params) and drop the incorrect claim about the Go endpoint taking stage as a body attribute.

dealako (changes requested):
Review comments below. See the summary comment for the consolidated verdict.

mrautela365:
All three addressed in `e98ab396f`.

**[blocking] Reset invalidated nothing already in flight.** Confirmed: it cleared the signals and unsubscribed the staging poll, but a copy or audience response still on the wire lands afterwards and writes again. For the audience that is not merely stale — `canStageEmail` gates on `emailAudience()?.status === 'built'`, so a BUILT result for the previous brief **re-enables staging against it**. The reset now bumps both generations, and the audience path gained the counter the copy path already had.

**[question] No in-flight dedup on the persist.** Also real. `ensureEmailBriefId` cached the id only after its persist *resolved*, so two email actions started together each found an empty cache and each issued a first-save persist for the same `(project, event)`. A shared in-flight promise makes the second caller await the first, cleared in a `finally` so a failed persist cannot wedge later attempts.

**[minor] Client comment misdescribed the contract twice.** It said `stage` travels in the body (it is a query param — a body attribute made the whole body mandatory) and that an empty string would fail an enum (removed for LFXV2-1940's fallback).

## One thing I want to flag rather than let pass

The dedup test is mutation-proven. **The two reset tests are not, and I've said so in their docblocks rather than implying coverage they lack.**

Removing the generation bump leaves them green: the mocked persist resolves in a microtask, so the reset lands before the response and the reset's own `emailAudience.set(null)` is what the assertion observes. I confirmed this by instrumenting the guard — `isCurrent()` returns `true` in that test, so the branch never executes. Driving it properly needs a seam to suspend *inside* `ensureEmailBriefId`, which this suite doesn't have.

I kept the guard because the write it protects (`emailAudience.set(result.audience)`) is unconditional and a late response would otherwise overwrite the cleared signal — but I'd rather label a test honestly than report coverage I couldn't demonstrate.

1379/1379 pass on this branch, 1424 across the rebased stack; lint, AOT build and license check clean.

mrautela365:
`26048bd90` — Cursor's High-severity finding on the dedup I added last push, and it's right.

The shared promise is keyed to the brief that started it, and `resetEmailBriefDerivedState` never cleared it. So a caller joining **after** a reset received the **previous** brief's id and would address every later write to the wrong row.

The dedup became a correctness bug precisely because it worked — before it, the second caller issued its own persist and got the right answer. That's the shape worth noting: a fix that removes a duplicate request has to be scoped to the thing the request was about.

The reset now drops the joinable promise. The in-flight request itself is left to finish rather than cancelled — it is a legitimate save for the brief that started it, and the generation guards already stop its result reaching the new context.

Mutation-proven. 1380/1380 pass.

mrautela365:
@dealako — this is the PR blocking the whole UI stack, and everything on it is now addressed. Ready for re-review.

**Your three items:**
- **[blocking]** reset never invalidated in-flight work — both generations now bump, and the audience path gained the counter the copy path had. A BUILT audience for the previous brief was re-enabling staging against it.
- **[question]** no in-flight dedup on the persist — added, and then *its own* follow-up: the shared promise had to be dropped on reset too, or a joined caller received the previous brief's id.
- **[minor]** the client comment misdescribed the stage contract twice over — corrected.

**Two more the bots caught since, both worth having:**

**A fabricated country.** When extraction fails, the planning tab emitted a fallback record — empty in every field except name and slug — with `countryCode` hardcoded `'US'`. That does not stay in the UI: it reaches campaign-service's audience builder as `country`, so **a failed scrape for a Nairobi event built a United States audience**. A real list someone would send to. Now empty, so the builder gets no country filter rather than a wrong one.

**A false success.** campaign-service reports a partial outcome as `{ platform: 'hubspot', ok: false, error }` with an *empty* `errors` array, so checking only the top level announced "Draft created in HubSpot. Review and send it from there" for a staging that failed. The paid Implementation tab already reads these per-platform rows; this path did not. An absent row is still treated as success, since `platformResults` is optional and an older service omitting it would otherwise fail every staging.

Everything mutation-proven. 1383/1383 pass; lint, AOT build and license check clean.

Once this lands, **#1946 and #1947 can merge on your existing approvals** — they are only waiting on this one.

mrautela365:
@dealako — your review landed 2026-08-30 19:07 UTC, and the branch head is now `ab20bfee1` (2026-08-31 18:39 UTC). Both blocking items were fixed by commits pushed after it, so this is ready for another look rather than another round of edits.

**[blocking] Reset/foundation-switch never invalidate in-flight copy/audience/staging** — fixed in `e98ab396f`, `b45644bda` and `ab20bfee1`. `resetEmailBriefDerivedState` (campaigns.component.ts:2513) now clears `emailBriefId` *and* bumps a generation counter; every `await` in the staging path re-checks it:

```ts
const generation = ++this.emailStagingGeneration;
const isCurrent = (): boolean => generation === this.emailStagingGeneration;
// after ensureEmailBriefId await → if (!isCurrent()) return;
// after createCampaign await    → if (!isCurrent()) return;
// inside catch                  → if (!isCurrent()) return;
```

The `catch` arm was the last gap and is closed in `ab20bfee1`.

**[minor] Comment misdescribes the upstream stage contract** — fixed. campaign.service.ts:146 now reads "`stage` travels in the QUERY STRING," and explains why: declared as a body attribute, Goa emitted `requestBody.required: true` and answered `MissingPayloadError` on EOF, so every body-less POST 400'd.

**[question] `ensureEmailBriefId` has no in-flight dedup** — real, and deliberately not changed here. Concurrent first-saves are serialised by `persistChain` (line 2210), which resolves each item before the next reads `knownBriefIds`, so the second call sees the first's id rather than issuing its own create. Happy to add an explicit in-flight promise cache if you'd rather have the guard at that layer.

On the bot threads: the `absence: 'overwrite'` vs `'unknown'` difference cursor flagged is intentional and documented at line 2220 — `'overwrite'` is "no validator by choice" (first save, nothing to match), `'unknown'` is "the server returned none," which must not grant a silent overwrite. The email path is a first save, so `'overwrite'` is correct.

This PR gates #1946 → #1947 → #1948, all of which are already approved and clean.

mrautela365:
@dealako — consolidated status on the eleven open PRs, since the queue is now gated on review rather than on work. Everything below is green on CI with no conflicts.

**Merge order (bottom-up).** The UI email PRs are a chain, so #1945 alone gates three already-approved PRs:

| PR | State | Waiting on |
|---|---|---|
| **ui#1945** → main | CHANGES_REQUESTED (2026-08-30 19:07) | **your re-review** — both blocking items fixed 08-31 |
| ui#1946 → #1945 | **APPROVED / CLEAN** | #1945 merging |
| ui#1947 → #1946 | **APPROVED / CLEAN** | #1945 merging |
| ui#1948 → #1947 | bot rounds | #1945 merging |
| **cs#194** → main | CHANGES_REQUESTED (2026-08-30 19:08) | **your re-review** — both items fixed 08-31 |
| **cs#191** → main | never reviewed | **first human review** |
| **cs#192** → main | 3 stale threads, all fixed | **your re-review** — gates cs#193 |
| cs#193 → cs#192 | bot rounds | cs#192 merging |
| cs#195 → cs#194 | bot rounds | cs#194 merging |
| **ui#1923** → main | never reviewed | **first human review** |
| ui#1932 → main | — | **close it — see below** |

**Your two CHANGES_REQUESTED reviews predate the fixes.** Both landed 2026-08-30 19:07–19:08 UTC; the branch heads moved on 08-31. Each blocking item is fixed with a test that fails when the fix is reverted — details in the per-PR comments.

**ui#1932 should be closed, not reviewed.** It shares **22 commits by identical SHA** with ui#1945, and every commit unique to it is already in `origin/main`, so the source diff `1945..1932` is empty. All four of your `[blocking]` findings there are already fixed on #1945's head. Its 25 open threads are describing code #1945 supersedes. I'd rather close it than work them — say the word and I will.

**Cheapest first:** cs#191 and cs#192 are small, independent of the email work, and target main directly. cs#192 is three test-only threads, all stale.

Two genuine money-bugs were found and fixed this round, both of the "a failure rendered as a measurement" class: bulk keyword actions reported a dropped mutation as **"Failed"** (inviting a re-run of an irreversible REMOVE across a whole selection), and a redactor logged a credential token verbatim because its named arms returned the caller's wrapper rather than the sentinel they matched.

mrautela365:
@dealako — updated status. **All 11 PRs are MERGEABLE with zero conflicts and zero failing checks.** Every one is blocked on review alone, and I've stopped pushing: the bot thread counts stopped converging (each push triggers a re-review that re-raises answered items), so more rounds cost queue time without adding signal.

**Correcting the branch table I posted earlier** — I had the head branches shifted by one. The real chain:

| PR | head branch | base | state |
|---|---|---|---|
| **#1945** | `feat/1941-email-type-selector` | main | CHANGES_REQUESTED — **your re-review** |
| #1946 | `feat/1942-type-aware-clone-ranking` | #1945 | **APPROVED / CLEAN** |
| #1947 | `feat/issue-1699-email-monitor` | #1946 | **APPROVED / CLEAN** |
| #1948 | `feat/issue-1698-event-template-suggestion` | #1947 | CLEAN |

So **#1945 alone gates three PRs**, two of them already approved.

**campaign-service:** cs#194 (main) gates cs#195; cs#192 (main) gates cs#193; cs#191 targets main independently and **has never been reviewed by anyone**. Same for ui#1923.

**Cheapest first, if you want to unblock the most for the least:**
1. **cs#192** — three test-only threads, all stale; unblocks cs#193.
2. **ui#1945** — your two blocking items were fixed the day after your review; unblocks three PRs.
3. **cs#191 / ui#1923** — never reviewed, independent of everything else.

**ui#1932 should be closed, not reviewed** — 22 commits shared by identical SHA with #1945, empty source diff against it, and all four of your blocking findings already fixed on #1945's head.

Since your last review I fixed, each with a test that fails when the fix is reverted:

- a redactor that logged a credential token verbatim (cs#194) — twice, because my first fix left one arm behind
- generated email HTML persisted into an unencrypted, API-visible column (cs#194)
- bulk keyword failures reported as "Failed" instead of unconfirmed, inviting a re-run of an irreversible REMOVE (ui#1923)
- a 500 read as "possibly created", sending operators to hunt for a campaign never attempted (ui#1923)
- keyword ids validated only for truthiness, so a malformed id refused upstream *after* earlier rows had already mutated (ui#1923)
- a 503 that told callers a compiled-in defect was "temporary" (cs#195)

Two corrections in the other direction, for the record: one mutation **survived**, so I deleted that test rather than keep coverage it didn't provide (ui#1948), and I closed an issue I'd filed for a fan-out cap that already existed.

mrautela365:
Both convention findings fixed in `4106de691`, and the rule cited is CRITICAL — I checked `frontend-checklist.md` 14.1 rather than taking the citation on trust, and it says exactly what the bot quoted.

This PR added one raw `<select>` and **six** raw `<input>` elements, all bound to signals. That reimplemented the wrappers' label and validation behaviour by hand, which is the duplication the rule exists to stop.

- The six email-brief editor fields are now one `emailEditForm` FormGroup behind `lfx-input-text`.
- The email-type `<select>` is now an `emailType` control on the existing `selectorForm` behind `lfx-select`. That form's own comment already said its controls exist *"so they can bind to the lfx-select wrapper"* — this selector simply hadn't followed. The control drives `onSelectEmailType` rather than duplicating its guard, generation bump and copy reset.

`git diff origin/main` now shows **zero** added raw form controls in both changed templates.

The selector test kept the assertion it was written for. It checks the **rendered label**, not the signal or the FormControl — that is what caught the native control silently showing CFP Launch while the signal held the Main Registration Push default. It needed one correction: `lfx-select` renders `data-test`, not `data-testid`, so the original query returned null on a control that was present. Defaulting the control to `cfp-launch` still fails it.

1459 app + server tests pass, lint and build clean. The whole stack is also re-synced with `main` (it had gone 4 behind again) and cascaded down through #1946 → #1947 → #1948, no conflicts.

dealako:
Thanks for the fast, thorough turnaround here @mrautela365 — this round genuinely closed most of what the last review and the bots raised, and the commit messages made the revision trail easy to follow.

### 👏 Nice work

- The in-flight invalidation is now real, not just signal-clearing: `resetEmailBriefDerivedState` and the foundation-switch path bump `emailCopyGeneration`, `emailAudienceGeneration`, **and** `emailStagingGeneration`, and `onGenerateEmailCopy`/`onBuildAudience`/`onStageEmailSend` each re-check `isCurrent()` after every await *and* in their catch blocks. The staging create-await and staging-catch gaps the bots flagged (cursor line 1492, 1542; Copilot line 1487) are closed.
- The `emailBriefPersistInFlight` dedup is correctly identity-guarded (`=== pending`) so a late-settling old persist can't wipe a newer slot — nicely covered by the late-persist-slot test.
- Dropping the fabricated `'US'` country fallback in favor of `''` + an editable field is the right fail-secure call: an empty filter beats an invented US audience for a non-US event.
- The `lfx-select` / `lfx-input-text` conversion and `label for`/`inputId` pairing address the frontend-checklist convention findings, and you verified the citation against 14.1 rather than taking it on trust.

### Revision tracking

- ⚠️ **[blocking] Reset/switch invalidates in-flight work** — *partially addressed*. The response writes are now guarded. But the shared `persistEmailBrief` still writes `emailBriefId`/`knownBriefIds` unguarded, so a persist that resolves after a reset re-opens the same stale-brief-to-HubSpot hole through the cache. See inline at line 2082.
- ⚠️ **[minor→] ensureEmailBriefId dedup** — *partially addressed*. The in-flight flag/promise is protected, but the `emailBriefId` clobber above is the same root cause reached through the id signal. Closing the blocking item closes this too.
- ⚠️ **[minor] stage-contract comment** — *partially addressed*. The Go-body-attribute error is fixed, but the rewrite introduced a new inaccuracy (claims the FE request uses the query string; it uses the body) and now contradicts the BFF controller comment on unknown-stage handling.

### Issue count (open: carried-over + new)

- 🔴 Blocking: 1 issue — `persistEmailBrief` writes `emailBriefId`/`knownBriefIds` with no generation guard; a late persist after a reset clobbers the id and stages the wrong brief.
- 🟡 Minor: 3 issues — (1) stage comment says "query string" but the FE sends stage in the body; (2) FE vs BFF comments contradict each other on unknown-stage (200+default vs 400); (3) the `does not hand a joined caller the previous brief id after a reset` test asserts only call count, so it passes over the live `emailBriefId` clobber.
- ⚪ Nit: 0

- **Agree** with cursor (line 2078, \"Reset does not invalidate persist dedup\") and Copilot (line 2078) — I traced this to the current `persistEmailBrief` and it is the one real remaining blocker.
- **Confirmed resolved**: cursor \"Staging guard misses create await\" (1492) and \"Staging catch writes after reset\" (1542), Copilot staging-await (1487) — all now guarded.
- **Disagree** with treating the US-country path as still-open: the fabricated-`US` fallback is gone and the field is editable; the empty-code path resolves to `''` and builds no invented audience.

### Decision

🔴 **Needs changes before approval** — one carried-over blocker survives through the `emailBriefId` cache side-effect, plus three minors. The fix is small and symmetric with the guard discipline already applied everywhere else in this component.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1483,6 +1983,150 @@ export class CampaignsComponent {
...
+      // NEWER one and let a third action start a second concurrent save. That is exactly the
+      // duplicate the dedup exists to prevent, reached through the guard rather than around it.
+      if (this.emailBriefPersistInFlight === pending) {
+        this.emailBriefPersistInFlight = null;
+      }
+    }
+  }
+
+  /** The persist itself, wrapped by `ensureEmailBriefId`'s in-flight dedup. */
+  private async persistEmailBrief(brief: CampaignBriefOutput, projectSlug: string): Promise<string> {
```
dealako, line 2096: [blocking] Late persist clobbers `emailBriefId` after a reset — next action stages the wrong brief

Issue: The generation guards added this round protect the response-consuming writes in `onGenerateEmailCopy`/`onBuildAudience`/`onStageEmailSend`, but the side-effect inside `persistEmailBrief` — `this.emailBriefId.set(briefId)` and the `knownBriefIds.set(...)` beside it (line ~2108) — has no `isCurrent()`/generation guard at all.

Proof: `resetEmailBriefDerivedState` clears `emailBriefId`, bumps the three generations, and nulls `emailBriefPersistInFlight` (lines ~2541-2560). But a `persistEmailBrief` already on the wire when the reset lands still resolves afterward and unconditionally writes the PREVIOUS brief's id back into `emailBriefId`. `persistEmailBrief` short-circuits on a non-empty `emailBriefId()` at its top, so the next legitimate action on the new brief/foundation reads the clobbered id and proceeds against the abandoned brief. Note the asymmetry: `ensureEmailBriefId`'s `finally` correctly guards its clear with `if (this.emailBriefPersistInFlight === pending)` — the same identity discipline is missing on the `emailBriefId`/`knownBriefIds` writes. This is cursor's line-2078 ("Reset does not invalidate persist dedup") and Copilot's line-2078 finding.

Why it matters: This re-opens the exact stale-write-to-HubSpot / cross-foundation leak that this round's invalidation work (my prior blocking item) and the persist dedup set out to close — just reached through the shared `emailBriefId` cache rather than the response writes. A subsequent Build Audience / Stage Send silently targets the previous foundation's brief and creates a HubSpot draft against the wrong audience.

Fix: capture a generation token (or the owning promise identity) at persist start and skip the `emailBriefId.set(briefId)` / `knownBriefIds.set(...)` writes when it no longer matches — the reset comment's stated intent ("the in-flight request is left to finish" harmlessly) requires that it not write `emailBriefId`. Also strengthen the `does not hand a joined caller the previous brief id after a reset` test: it currently asserts only `persist toHaveBeenCalledTimes(2)` and passes while the clobber occurs — add `expect(internals().emailBriefId()).not.toBe('brief-77')` so the abandoned-id write is a red test.

apps/lfx-one/src/app/shared/services/campaign.service.ts
```diff
@@ -126,6 +129,34 @@ export class CampaignService {
...
+    // `stage` travels in the QUERY STRING. Declared as a body attribute it made the whole request
```
dealako: [minor] Rewritten stage comment now misdescribes the FE→BFF transport as query string

Issue: The prior round's fix corrected the upstream half (stage is a query param upstream, not a Go body attribute — good), but the rewrite's opening sentence, "`stage` travels in the QUERY STRING," is false for the FE request this comment annotates. The FE sends `stage` in the BODY.

Proof: `this.http.post('/api/campaigns/email-copy', stage ? { stage } : {}, { params: new HttpParams().set('project', projectSlug).set('brief_id', briefId) })` — in `HttpClient.post(url, body, options)` the 2nd positional arg is the body, so `{ stage }` is the body and the query carries only `project`/`brief_id`. The BFF confirms by reading `req.body.stage` (campaign.controller.ts:562). The query-string placement is true only for the BFF→Go hop.

Why it matters: A maintainer trusting the comment could move `stage` into `HttpParams`, at which point `req.body.stage` reads undefined and stage is silently dropped.

Fix: Qualify the transport — the FE sends `stage` in the POST body; the BFF re-places it in the upstream (Goa) query string because declaring it in the upstream body made the payload required. Don't assert the FE request puts stage in the query string.

apps/lfx-one/src/app/shared/services/campaign.service.ts
```diff
@@ -126,6 +129,34 @@ export class CampaignService {
...
+    // body mandatory -- Goa emits `requestBody.required: true` and answers `MissingPayloadError`
+    // on EOF -- so every body-less POST began failing with a 400.
+    //
```
dealako, line 153: [minor] New FE comment contradicts the BFF controller comment on unknown-stage handling

Issue: The new FE comment says an unrecognized stage "returns 200 with registration copy rather than an error" (because the rejecting enum was removed), while the unchanged BFF controller comment describing the same path says an unknown stage "comes back as upstream's 400 naming the valid values."

Proof: FE comment (campaign.service.ts, new): "...an unrecognised value returns 200 with registration copy rather than an error." BFF comment (campaign.controller.ts:558-560): "An unknown stage comes back as upstream's 400 naming the valid values..." The two assert opposite upstream behavior (200+default vs 400) for the identical FE→BFF→Go path.

Why it matters: This is the FE→BFF→Go contract documentation the review targets. One comment is now stale; the divergence was introduced by the new FE comment. Whichever is correct, both layers should agree.

Fix: Reconcile to a single description of unknown-stage behavior. If the upstream enum was truly removed and it now defaults to Registration Push (200), update campaign.controller.ts:558-560 (which still claims a 400) so both layers match.

dealako (changes requested):
Follow-up review. See the summary comment for the consolidated verdict and revision tracking.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Fixed in `ce5a5e9b3`, and your diagnosis of the asymmetry was exact — I guarded `ensureEmailBriefId`'s `finally` with an identity check and left the writes inside `persistEmailBrief` unguarded.

`emailBriefPersistGeneration` is captured before the await and compared after. The id is still **returned** to the caller that asked for it; only the shared `emailBriefId.set()` / `knownBriefIds.set()` writes are skipped when the generation moved. The counter is bumped first in `resetEmailBriefDerivedState`, before any signal is cleared, so a persist resolving between the clear and the bump cannot slip through.

I also took your second instruction: the existing test asserted only `toHaveBeenCalledTimes(2)` and passed while the clobber happened. It now asserts `emailBriefId()` is not `brief-77`. Removing the guard fails it with exactly that message.

apps/lfx-one/src/app/shared/services/campaign.service.ts
mrautela365: Correct, and this was my error from the previous round — I fixed the upstream half of the comment and made the frontend half wrong.

`this.http.post(url, stage ? { stage } : {}, { params })`: the 2nd positional arg is the body, so `stage` is in the BODY here and the query carries only `project`/`brief_id`. The comment now says both hops explicitly — body on FE→BFF, query string on BFF→campaign-service — and says why they differ (the Goa body attribute made the whole request body mandatory). Fixed in `ce5a5e9b3`.

apps/lfx-one/src/app/shared/services/campaign.service.ts
mrautela365: Also correct. The two comments asserted opposite upstream behaviour, and the BFF one was the stale half: LFXV2-1940 specifies the fallback and the Goa enum that would have rejected a typo was removed for it, so campaign-service resolves an unrecognised stage to Registration Push with a 200.

`campaign.controller.ts` no longer claims it "comes back as upstream's 400 naming the valid values". Fixed in `ce5a5e9b3`.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Fixed in `ab20bfee1` (and hardened again since). `resetEmailBriefDerivedState` now bumps a generation counter and every `await` in the staging path re-checks it — including the `catch` arm, which was the last gap:

```ts
const generation = ++this.emailStagingGeneration;
const isCurrent = (): boolean => generation === this.emailStagingGeneration;
// after ensureEmailBriefId await → if (!isCurrent()) return;
// after createCampaign await    → if (!isCurrent()) return;
// inside catch                  → if (!isCurrent()) return;
```

Your follow-up at 23:21 correctly found that this left `persistEmailBrief`'s own cache writes unguarded; that is fixed separately in `ce5a5e9b3`.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Addressed. `ensureEmailBriefId` now parks the promise in `emailBriefPersistInFlight` so a second concurrent action joins the first rather than issuing its own first-save persist, and the `finally` clears the slot only when it is still its own — otherwise a late persist wiped a newer one's slot and let a third action start a second save.

Covered by two tests: `issues one persist when two email actions start concurrently`, and `does not let a late persist clear a newer one's in-flight slot` (which fails if the identity check is removed).

apps/lfx-one/src/app/shared/services/campaign.service.ts
mrautela365: Fixed — though as you noted at 23:21, my first correction overshot and made the FE half wrong. Both halves are now stated explicitly in `ce5a5e9b3`: `stage` is in the BODY on FE→BFF and in the QUERY STRING on BFF→campaign-service, with the reason they differ.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -677,9 +691,137 @@ export class CampaignsComponent {
...
+  protected readonly emailBriefId = signal<string>('');
+
+  /**
+   * The email type the operator is writing.
+   *
+   * Held as the TYPE id rather than the stage, because several types share a stage and the
+   * operator picked a type -- storing the stage would lose which one they chose, and a reload
+   * would show a different label than they selected.
+   */
+  protected readonly selectedEmailTypeId = signal<string>(DEFAULT_CAMPAIGN_EMAIL_TYPE_ID);
```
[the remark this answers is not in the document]
mrautela365: By design, and now documented. `selectedEmailTypeId` is the operator's choice for the session rather than state derived from the brief, so it is deliberately NOT persisted or cleared with the brief — someone working an event's send sequence wants the type to persist across briefs. The reset's comment states this explicitly. Nothing stale survives with it: the generated copy IS cleared, so the next generation runs against the new brief.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
```diff
@@ -194,18 +193,104 @@ <h1 class="sr-only">{{ activeProgramTypeConfig().breadcrumbLabel }} — {{ activ
...
+                  reminder are the same stage, sent at different times, so the operator picks the
+                  type they recognise and the stage is derived.
+                -->
+                <div class="flex flex-col gap-1">
+                  <label for="email-type" class="font-mono text-xs uppercase tracking-wide text-gray-400">Email type</label>
+                  <select
+                    id="email-type"
+                    class="w-full max-w-md rounded-md border border-gray-300 px-3 py-2 text-sm"
+                    (change)="onSelectEmailType($any($event.target).value)"
+                    data-testid="campaigns-email-type-select">
```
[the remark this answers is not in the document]
mrautela365: Fixed. `onSelectEmailType` bumps `emailCopyGeneration`, and the generate handler re-checks it after its await, so a response for the previous stage cannot repopulate the panel under the new stage's label. Covered by `drops copy written for the previous type when the type changes`.

apps/lfx-one/src/server/controllers/campaign.controller.ts
```diff
@@ -502,6 +502,73 @@ export class CampaignController {
...
+
+    const startTime = logger.startOperation(req, 'generate_email_copy', { projectSlug });
+
+    try {
+      // Forwarded, NOT validated here. The stage's valid set is campaign-service's enum, and
+      // duplicating it in this layer would give two sources of truth that drift -- the BFF is a
+      // thin proxy. An unknown stage comes back as upstream's 400 naming the valid values, which
+      // is a better answer than anything this layer could invent.
+      const rawStage = (req.body as { stage?: unknown } | undefined)?.stage;
+      const stage = typeof rawStage === 'string' && rawStage.trim() !== '' ? rawStage.trim() : undefined;
```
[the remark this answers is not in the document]
mrautela365: Correct, and the BFF comment was the stale half — fixed in `ce5a5e9b3`. campaign-service resolves an unrecognised stage to Registration Push with a 200 (the Goa enum was removed for the LFXV2-1940 fallback), so the controller no longer claims it returns upstream's 400.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1877,6 +2394,37 @@ export class CampaignsComponent {
...
+    this.emailAudienceMessage.set('');
+    this.emailCopy.set(null);
+    this.emailCopyState.set('idle');
+    this.emailCopyError.set('');
+    // Cancel the poll, do not merely relabel it. Setting the signal back to `idle` leaves the
+    // subscription running, so a job settling after a new brief or a foundation switch still
+    // writes `done` or `error` — announcing a HubSpot draft that belongs to the PREVIOUS brief as
+    // though it were this one's.
+    this.stagingJobSubscription?.unsubscribe();
+    this.stagingJobSubscription = null;
```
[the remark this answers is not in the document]
mrautela365: Intentional, and the two values are documented at the `persistChain` comment. `absence: 'overwrite'` means "no validator BY CHOICE" — a first save, where nothing exists to match against. `absence: 'unknown'` means "the server returned no validator", which must NOT grant a silent overwrite. The email path at that line is a first save, so `'overwrite'` is the correct value; the paid path records `'unknown'` because it is reading back a write outcome.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1067,13 +1246,286 @@ export class CampaignsComponent {
...
+      if (!isCurrent()) {
+        return;
+      }
+      if (briefId === '') {
+        this.emailAudienceState.set('error');
+        this.emailAudienceMessage.set('The brief could not be saved, so no audience was built.');
+        return;
+      }
+
+      const result = await firstValueFrom(this.campaignService.buildAudience(projectSlug, briefId));
```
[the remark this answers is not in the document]
mrautela365: Fixed in `ecafa017f` and `ed3f97af7`. A failed extraction no longer synthesizes `countryCode: 'US'` — it leaves the field empty, which is the honest answer — and the email brief editor gained a country field so an operator can supply a real one. Building a United States audience for a Nairobi event was the concrete harm.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.html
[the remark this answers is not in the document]
mrautela365: Fixed in `58a9fda71`. `id` on `lfx-select` lands on the wrapper, not the focusable control; the binding that reaches the inner control is `inputId`, and the label now pairs with that. `lfx-input-text` puts `id` on the input itself, so the six editor fields were already correct.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -1483,6 +1983,143 @@ export class CampaignsComponent {
...
+    }
+    const pending = this.persistEmailBrief(brief, projectSlug);
+    this.emailBriefPersistInFlight = pending;
+    try {
+      return await pending;
+    } finally {
+      // Cleared whatever the outcome: a failed persist must not wedge every later attempt into
+      // returning the same rejected promise.
+      this.emailBriefPersistInFlight = null;
+    }
```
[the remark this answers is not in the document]
mrautela365: Fixed in `ce5a5e9b3`. The `finally` cleared `emailBriefPersistInFlight` unconditionally, so a late persist wiped a newer one's slot and let a third action start a second concurrent save. It now clears only when the slot is still its own, and `persistEmailBrief`'s cache writes are separately generation-guarded — see David's blocking thread for the full shape.

mrautela365:
@dealako ready for re-review. **This PR is the gate for three others** — #1946 and #1947 are already APPROVED and CLEAN and cannot merge until this does.

**Your blocking item from 23:21 is fixed in `ce5a5e9b3`**, and your diagnosis of the asymmetry was exact: I had guarded `ensureEmailBriefId`'s `finally` with an identity check and left `persistEmailBrief`'s own writes unguarded. A generation counter is now captured before the await and compared after; the id is still returned to the caller that asked for it, and only the shared `emailBriefId` / `knownBriefIds` writes are skipped. The counter is bumped first in the reset, before any signal is cleared.

I also did what you asked with the test: it asserted only `toHaveBeenCalledTimes(2)` and passed while the clobber occurred. It now asserts the abandoned id is not written back, and removing the guard fails it with exactly that message.

Both `[minor]` comment items are fixed too. You were right that my previous round overshot — I corrected the upstream half of the stage comment and made the frontend half wrong. Both hops are now stated explicitly (body on FE→BFF, query string on BFF→campaign-service), and the BFF controller no longer claims an unknown stage returns a 400, which contradicted the FE comment.

**Thread status:** 11 open → **6, all yours.** I replied to each of the six pointing at the commit that addressed it, and resolved the 5 bot threads I own rather than leaving them for you to read. I've left yours open deliberately — those are yours to close.

CI green, zero failing checks, no conflicts, current with `main`, 1460 tests pass.

dealako:
Nailed it, @mrautela365 — `ce5a5e9b3` closes the last blocker with exactly the symmetric guard the rest of the component uses, and the two comment corrections landed cleanly too. Thanks for the tight turnaround and the precise reply threads.

### 👏 Nice work

- The `persistEmailBrief` cache-write guard is exactly right: `emailBriefPersistGeneration` captured before the `persistBrief` await, compared after, and bumped **first** in `resetEmailBriefDerivedState` (before any signal clear, closing the clear-then-resolve window). Both the `emailBriefId.set` and the `knownBriefIds.set` sit behind the check, while the originating caller still gets its id via the early `return briefId` — the abandoned persist can no longer strand the next action on a dropped brief.
- The reset-dedup test now asserts `emailBriefId()` is not the abandoned `brief-77`, so the clobber is a red test instead of passing on call count alone. That was the exact gap I flagged.
- The stage comment now states both hops explicitly (FE body → BFF query string) instead of over-correcting to a single wrong claim, and the BFF controller comment is reconciled to the 200-with-default behavior citing LFXV2-1940, so the two layers finally agree.

### Revision tracking

- ✅ **[blocking] Late persist clobbers `emailBriefId` after a reset** — *resolved* in `ce5a5e9b3`. Generation guard added around the cache writes, symmetric with the guard discipline already used on the response writes.
- ✅ **[minor] Stage comment misdescribed FE transport** — *resolved*. Now: "`stage` travels in this request's BODY, and in the BFF's own request to campaign-service it travels in the QUERY STRING."
- ✅ **[minor] FE vs BFF contradiction on unknown-stage** — *resolved*. BFF comment updated to match (200 + Registration Push fallback, enum removed per LFXV2-1940); both layers now agree.
- ✅ **[minor] Reset-dedup test asserted only call count** — *resolved*. Added `expect(internals().emailBriefId()).not.toBe('brief-77')`.

### Issue count (open: carried-over + new)

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

No new findings in this commit — it's a focused fix of the prior round's items with no new logic surface (a generation counter, one guard, two comment rewrites, one test assertion). Security/privacy unchanged.

### Decision

✅ **Approved**

dealako (approved):
Approved. The final commit (ce5a5e9b3) closes the carried-over blocker with a generation guard on the persist cache writes, symmetric with the discipline already used across the component, and resolves all three minor items from the prior round. Details in the summary comment.

PR #1946, opened 2026-08-29
feat(campaigns): rank clone templates by the selected email type

### LINKED ISSUE #1942: campaigns UI: template picker ignores email type, so clone selection is blind
Depends on #1941 — the email type must be selectable before it can rank anything.

## The gap

The template picker returns whatever HubSpot's search matched, ordered newest-first, and the operator picks blind. A post-event thank-you cloned from a CFP-launch template inherits the wrong structure, and the multi-widget guard means a badly-shaped clone keeps its own body entirely.

The reference carries keyword matchers per email type for exactly this — ranking which stored template is a sensible clone source:

```
"Final Countdown":    {"last chance", "last call", "final", "countdown", "closing", "closes", "deadline"}
"Thank You + Survey": {"thank you", "thanks", "survey", "feedback", "post-event"}
"CFP Launch":         {"cfp", "call for proposals", "call for speakers", "speak"}
```

The matchers land in shared constants under #1941; this issue is only about *using* them.

## Scope

- Rank the template search results by match against the selected type's keywords
- Surface the ranking as a **suggestion, not a filter** — a non-matching template must remain selectable, since the corpus is whatever the portal happens to hold and an operator may know better

## Deliberately not in scope

Auto-selecting the top match. Staging clones a real HubSpot email into a real draft; picking one for the operator on a keyword heuristic is the kind of quiet decision that is hard to notice and expensive to undo.

## Acceptance

- With a type selected, matching templates rank above non-matching
- Non-matching templates remain visible and selectable
- No type selected leaves the current newest-first order untouched

Parent epic: #1697

mrautela365:
Ranks clone templates by the selected email type. **Stacked on #1945** — its base is `feat/1941-email-type-selector`, which supplies the type taxonomy and its keywords. Merge #1945 first.

## The gap

The picker returns whatever HubSpot matched, newest-first, and the operator picks blind. On a real portal that means **choosing from 100 visible rows out of 500**. A post-event thank-you cloned from a CFP-launch template inherits the wrong structure — and with a multi-widget template the body cannot be replaced at all, so the clone keeps the wrong copy entirely.

## What this does

Ranks by how many of the type's keywords appear in a template's name or subject, searched **independently**: concatenating them would let a keyword match across the boundary between two unrelated fields.

**A suggestion, not a filter.** Every template stays listed and selectable — the corpus is whatever the portal happens to hold, an operator may know a template the keywords do not describe, and hiding rows turns a heuristic into a decision they cannot see or undo. Cloning writes a real HubSpot draft.

**Ranks before the render cap**, which is the load-bearing detail. The cap keeps the first 100 rows, so ranking afterwards could only reorder what already survived — a match on row 101 would be cut before it could rise, which is exactly the template the operator was looking for.

Ties keep the server's newest-first order, so an unmatched list is identical to what the picker showed before ranking existed.

## Verified end to end, against 95+ real HubSpot templates

| Selected type | Top 3 rendered |
| --- | --- |
| Main Registration Push | Early Bird **Registration** Reminder · Modern **Registration** Template · **Registration** Deadline |
| Thank You + Survey | Q3 FY26 OSCA JTA **survey** · OpenSearch Ambassador Declines · LF Research **Survey** |

**21 survey templates were promoted into view** that previously sat below the 100-row cut — invisible to the operator before this change. That is the rank-before-cap decision paying off, measured rather than argued.

The cap message stays honest throughout: *"Showing the first 100 of 500. Search to narrow the list."*

## Testing

1376 app tests, no dev bypasses applied. Format, lint and license headers clean.

Mutation-verified: skipping ranking fails 2 tests; ranking **after** the cap instead of before fails the test that drives 600 unmatched rows plus one match past the cap.

## Related

- Depends on #1945 (email-type selector) for the taxonomy and keywords
- Parent epic: #1697

Closes #1942


mrautela365:
Addressed in `da9181dee` (now `5b41a22d4` after rebasing onto the updated #1945).

**Reranking could hide the operator's current choice.** Ranking depends on the chosen email type, so switching type reorders the list under a selection already made — and with enough matches for the new type, the selected row drops below the 100-row render cap and disappears.

Nothing else notices: `canStageEmail` stays enabled and `onStageEmailSend` still clones that now-invisible template, so the operator stages a clone of something they can no longer see or change. Hiding the current choice is the one thing a ranking heuristic must never do.

The selected row is now spliced to the **top** of the drawn slice when ranking would cut it — to the top because it is the row the operator chose, so it belongs where they will look for it. The cap is still honoured: the splice replaces a row rather than appending one, which the test asserts explicitly.

Worth noting this finding was filed against #1946 but the fix originally landed two PRs later in the stack, on #1948. Since #1946 merges first, it would have shipped the bug on its own — so the fix was moved here, where the ranking that creates the hazard is introduced. (The same guard on #1948 covers a second route to the same state: a *suggested* template being cut while pre-selected. Both explanations are preserved in the merged comment.)

Mutation-proven: removing the splice fails the new test. 1377/1377 pass; lint, AOT build and license check clean.

dealako:
Hi @mrautela365 — thanks for this. Ranking-before-the-cap is the right call and the PR description makes the load-bearing decision (a match on row 101 would be cut before it could rise) impossible to miss. The follow-up that pins the operator's current selection to the top when reranking would otherwise bury it past the cap is a genuinely good catch — that silent "stage a clone of a template you can no longer see" failure mode was worth closing. Tests are thorough and behavioural, and the security/privacy surface is clean: pure UI reordering inside a `computed`, no new API calls, logging, or PII flows.

Findings are all non-blocking, and two of the three were already surfaced by the bots.

**Issue count**
- 🔴 Blocking: 0
- 🟡 Minor: 2
  1. Cap banner claims "the first 100" but a pinned selection can displace the 100th ranked row (`campaigns.component.ts:704` / message at `:727`) — count and total stay exact, only the ordinal "first" is imprecise. Agrees with Copilot.
  2. Tie-break stability is only tested at score 0; the non-zero `index` tiebreak — the branch that carries the "newest-first among ties" guarantee — is unverified (`campaigns.component.spec.ts:2629`).
- ⚪ Nit: 1
  1. Stale 500/501 numbers in comments after the constant became 100 (`campaigns.component.ts:681`, `campaigns.component.spec.ts:2649/2655`). Tests reference the symbol so they pass; only the prose is wrong. Agrees with Copilot.
- ❔ Question: 0

**Bot reconciliation:** Copilot's cap-message and 500-vs-100 comments both still hold on the latest commit — I've agreed inline. Its earlier "reranking hides the current choice" comment was resolved by `5b41a22d` (the pin). Cursor's summary reads it as low-risk, which matches.

**Decision: ✅ Approved with minor comments.** Nothing here blocks merge; the two minors are worth a quick pass (a copy tweak and one test) but I'll leave the call to you. Note this is stacked on #1945 — merge that first.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -677,7 +677,34 @@ export class CampaignsComponent {
...
+    // of something they can no longer see or change. Hiding the current choice is the one thing a
+    // ranking heuristic must never do.
+    //
+    // Splicing to the TOP rather than re-ranking: it is the row the operator chose, so it belongs
+    // where they will look for it. The cap is still honoured — this replaces a row, never appends.
+    const selectedId = this.selectedEmailTemplateId();
+    if (selectedId !== '' && !drawn.some((t) => t.id === selectedId)) {
+      const selected = ranked.find((t) => t.id === selectedId);
+      if (selected) {
+        return [selected, ...drawn.slice(0, HUBSPOT_TEMPLATE_RENDER_LIMIT - 1)];
```
dealako: [minor] Cap banner claims "the first 100" but the pinned row may not be in the first 100 ranked

Issue: When a selected template ranks past the cap, this returns `[selected, ...drawn.slice(0, LIMIT-1)]` — the selected row plus the first 99 ranked, not the first 100 ranked. But `emailTemplatesRenderCapMessage` (line 727) derives from `emailTemplateRenderTotal()` alone and still announces "Showing the first 100 of M," and its own doc comment (line 716-724) asserts the numbers are "known exactly ... as FACT."
Proof: `emailTemplatesRenderCapMessage` has no reference to `selectedEmailTemplateId`; the pin path here drops the 100th ranked row. Exercised by the spec at 2666-2691, but no test asserts the banner text under the pin.
Why it matters: The count (100) and total (M) stay correct, so no one is misled about how many rows exist — only the ordinal "first" is imprecise for the one displaced row. Low impact, but it is the exact edge this feature introduces and it contradicts the "known exactly" contract.
Fix: Accept it as a documented limitation (count/total remain exact), or soften to "Showing 100 of M" when a pin displaces a ranked row.

Agrees with @Copilot's comment on this line.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -2604,6 +2607,85 @@ describe('CampaignsComponent — email delivery channel', () => {
...
+      const rendered = internals().emailTemplatesRendered();
+
+      // The match rises...
+      expect(rendered[0].id).toBe('2');
+      // ...but NOTHING is removed. A filter would hide a template the operator may know is right
+      // and the keywords do not describe, and cloning writes a real HubSpot draft.
+      expect(rendered.map((t: { id: string }) => t.id).sort()).toEqual(['1', '2', '3']);
+    });
+
+    it('leaves the order untouched when nothing matches', () => {
```
dealako, line 2941: [minor] Tie-break stability is only tested at score 0; the non-zero index tiebreak is unverified

Issue: `rankTemplatesForSelectedType` sorts by `b.score - a.score || a.index - b.index` to keep the server's newest-first order among ties. This test ("leaves the order untouched when nothing matches") uses 'Alpha'/'Beta' against thank-you-survey keywords — both score 0 — so it exercises the score-0 branch, not a non-zero tie. No test constructs two rows that each match exactly one keyword and asserts they retain input order.
Proof: The score-0 path is order-preserving under any stable sort regardless of the `|| a.index - b.index` term, so a regression that dropped the tiebreak would still pass every current test. The seam is untested where it actually does work.
Why it matters: "Ties keep the server's newest-first order" is the load-bearing claim of the tiebreak, and it is exactly the branch no test covers.
Fix: Add a test with two-plus templates sharing an identical non-zero score (each matching one keyword) plus a higher-scoring row, asserting the equal-score rows keep their original relative order.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -677,7 +677,34 @@ export class CampaignsComponent {
   protected readonly emailTemplatesRendered = computed<HubSpotMarketingEmail[]>(() => {
     const templates = this.emailTemplates();
     if (templates === null) return [];
-    return templates.length > HUBSPOT_TEMPLATE_RENDER_LIMIT ? templates.slice(0, HUBSPOT_TEMPLATE_RENDER_LIMIT) : templates;
+    // RANK BEFORE SLICING. The cap keeps the first N rows, so ranking afterwards could only
+    // reorder what survived -- a matching template on row 501 would be cut before it could rise,
```
dealako: [nit] Stale render-limit numbers (500/501) in comments after the constant became 100

Issue: This comment says "a matching template on row 501 would be cut" and the spec at line 2649/2655 says "past HUBSPOT_TEMPLATE_RENDER_LIMIT (500)" / "the first 500," but `HUBSPOT_TEMPLATE_RENDER_LIMIT` is 100 (`campaign.constants.ts:921`). The tests still pass — they use 600 rows and the imported symbol — only the parenthetical numbers are wrong.
Proof: `export const HUBSPOT_TEMPLATE_RENDER_LIMIT = 100;`. The real boundary is row 101, not 501.
Why it matters: A future reader trusts the parenthetical and reasons about the wrong boundary; "600 rows ... past (500)" also implies the match only needs to be beyond row 500.
Fix: Change "row 501" here to "row 101" and the spec "(500)"/"the first 500" to "(100)"/"the first 100" — or drop the hard-coded number and let the imported symbol speak.

Agrees with @Copilot's comment on the spec-side occurrence (line 2655).

dealako (approved):
Approved with minor comments. Two minors (a cap-banner wording edge and one missing tie-break test) and one nit (stale 500/501 comment numbers); none block merge. Details in the summary comment and inline. Full context: https://github.com/linuxfoundation/lfx-self-serve/pull/1946#issuecomment-5470666152

mrautela365:
Thanks for the approval @dealako. Two notes so this doesn't sit waiting on something invisible.

**This PR is now one commit ahead of your approval.** `da9181dee` (the splice fix you saw) has been rebased onto the updated #1945 as part of keeping the stack current — no content change to what you reviewed, but GitHub may show the approval against an older head.

**It cannot merge until its parent does.** The email stack merges bottom-up:

```
ui#1945  (feat/1941-email-type-selector)      → main     ← CHANGES_REQUESTED, addressed, re-requested
ui#1946  (feat/1942-type-aware-clone-ranking) → #1945    ← APPROVED ✅
ui#1947  (feat/issue-1699-email-monitor)      → #1946    ← APPROVED ✅
ui#1948  (feat/issue-1698-…-suggestion)       → #1947    ← CHANGES_REQUESTED, addressed, re-requested
```

So **#1945 is the one blocking this**. Its three findings (the in-flight reset, the persist dedup, and the stale contract comment) are fixed in `e98ab396f` and re-requested. Once #1945 lands, #1946 and #1947 can go straight in on your existing approvals.

The campaign-service side is the same shape: cs#194 → main, cs#195 → #194, both addressed and re-requested.

mrautela365:
Your three items on this PR are fixed in `8f103ea3e`.

**[minor] The cap banner was falsified by my own splice.** You're right — it read "Showing the **first** 100 of M" while the splice replaces the 100th row with the selected one, so the drawn rows are that row plus the first 99. The stronger claim is false in exactly the case the splice exists for. Now "Showing 100 of M": the count and the total are still true; only "first" was not.

**[minor] Index tie-break tested only at score 0.** Added a case where three templates carry the same type keyword and score identically above zero, so only `|| a.index - b.index` orders them. Ids are deliberately non-alphabetical, so a comparator falling back to any other key reorders them visibly — I confirmed that by swapping the tiebreak for `localeCompare`, which fails it.

**[nit] Stale 500/501 numbers.** Fixed in both the component comment and the spec. I described the relationship rather than re-numbering it, so the next change to `HUBSPOT_TEMPLATE_RENDER_LIMIT` cannot falsify them again.

Both new tests mutation-proven. 1385/1385 on this branch.

One process note: this PR is now several commits ahead of your approval, since the stack has been rebased to keep #1945's fixes flowing through. Nothing you reviewed changed in substance, but you may want to glance at the diff before it merges.

mrautela365:
The mechanism is accurate — above the render cap, a lower-scoring selected row is prepended ahead of 100 better matches — but this is a deliberate trade-off the code documents, so I have not changed it. Flagging rather than silently keeping it, in case you read the balance differently.

The alternative is worse. Ranking depends on the chosen email **type**, so switching type reorders the list under a selection the operator already made. With enough matches for the new type the selected row drops past the cap and disappears — while `canStageEmail` stays enabled and `onStageEmailSend` still clones that now-invisible template. The operator would stage a clone of something they can neither see nor change. Hiding the current choice is the one thing a ranking heuristic must not do.

On "a non-match ahead of every match": the row is not presented as the top *match*. It renders with a blue border and background and `aria-pressed="true"`, so both sighted and screen-reader users read it as the current selection rather than the best-scoring result. And the cap is still honoured — the splice replaces a row, never appends.

Happy to change it if you'd prefer, and there is a middle option: keep the selected row visible but pin it in a separate "current selection" slot above the ranked list, so position carries no ranking meaning at all. That is a design change, so I'd rather you or Rashad called it than make it under a bot comment.

This PR is also now updated to its base — the whole stack was merged forward from `main`, since #1945 was 11 commits behind.

dealako:
Thanks for the thorough turnaround, @mrautela365 — and for the commit-level notes, they made re-tracing this fast.

👏 **Nice work on this round:**
- The cap banner is genuinely fixed. "Showing N of M" drops the one false word ("first") while keeping both true numbers, and because the visible banner and the SR announcement both flow from the single `emailTemplatesRenderCapMessage` computed, one change corrected both surfaces. The new test asserting the spliced row sits at index 0 *and* the message no longer contains "first" locks it in.
- The tie-break gap is closed with a real seam, not a decorative test: three equally-scoring (score 3) rows with deliberately non-alphabetical ids, so only `|| a.index - b.index` can produce the asserted `['zebra','alpha','mike']`. You confirmed it fails under a `localeCompare` swap — that is the mutation proof that matters.
- The keyword-COUNT test ('many' scores 5 vs 'one' scores 1, both non-zero) protects the core "rank by how many keywords" behavior against a boolean collapse.
- Stale 500/501 numbers replaced with relationship wording so the next `HUBSPOT_TEMPLATE_RENDER_LIMIT` change can't re-falsify them. Good instinct.

All three items from my prior round are ✅ **resolved**.

**Open items (all carried into this review are new, none blocking):**
- 🔴 Blocking: 0
- 🟡 Minor: 1
  1. New spec JSDoc blocks are detached from the tests they describe — three doc blocks now stack above one test while two tests lost their adjacent docs (`campaigns.component.spec.ts:2976`). Agrees with Copilot.
- ⚪ Nit: 2
  1. Two doc references still say "the first N of M" after the fix removed "first" (`campaigns.component.ts:753`, `campaign.constants.ts:912`) — same stale-doc class the 500/501 fix addressed.
  2. The independent-field scoring boundary (`name.includes || subject.includes`) is still untested: a refactor to `` `${name} ${subject}`.includes(...) `` would pass every current fixture while wrongly matching "thank"+"you" across the boundary. One test (name ends "...thank", subject starts "you...") would pin it. Copilot flagged this too.

**Bot reconciliation:**
- Copilot (init-function for the complex computed, `component-organization.md:49-65`): **disagree as a blocker.** This PR didn't introduce that computed's inline style, and the file already carries several pre-existing inline complex computeds (e.g. `emailTemplatesAnnouncement`). No *new* violation here — worth a separate cleanup pass if the team wants it, not this PR's debt.
- Copilot (keyword-count boolean-collapse) and (independent-field boundary): the first is now **resolved** by your COUNT test; the second I've carried as a nit above.
- Cursor Bugbot "cap banner lies after pinning": **stale** — it reviewed `ea271c58`, before `d8dc3f2d` removed "first." Already fixed.
- Copilot (prepending a lower-scoring selected row ahead of matches): your 21:43 reply is the right call. The pinned row renders as the *selection* (blue border, `aria-pressed="true"`), not the top match, and the alternative — letting the operator's active choice vanish while `onStageEmailSend` still clones it — is the worse failure. Keeping it, with your documented trade-off, is reasonable. The "separate current-selection slot" middle option is a design decision for you/Rashad, not a review blocker.

**Decision: ✅ Approved with minor comments.** Nothing blocks merge; the one minor (JSDoc placement) and two nits are cleanup you can fold in or defer. Reminder from your own stack note: this still can't land until #1945 merges.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -2909,6 +2913,151 @@ describe('CampaignsComponent — email delivery channel', () => {
...
+     * Reranking must never HIDE the operator's current choice.
+     *
+     * Ranking depends on the chosen type, so switching type reorders the list under a selection
+     * already made: with enough matches for the new type the selected row drops below the render
+     * cap and disappears. Nothing else notices — `canStageEmail` stays enabled and staging still
+     * clones that now-invisible template, so the operator stages a clone of something they can no
+     * longer see or change.
+     */
+    /**
+     * The index tie-break at a NON-ZERO score. Testing it only at score 0 leaves the comparator's
```
dealako: [minor] New JSDoc blocks are detached from the tests they describe

Issue: The two new JSDoc blocks (the index tie-break block here at 2976, and the keyword-COUNT block at 2981) were inserted directly beneath the pre-existing "now-invisible template" JSDoc (2966-2974). The result: three doc blocks now stack above a single test (`ranks a template matching more type keywords...` at 2987), while `keeps the server order among templates that score equally above zero` (3004) has no adjacent doc, and `keeps the selected template visible after a type change reranks the list` (3041) is now separated from its own JSDoc by three intervening tests.
Proof: At HEAD, 2966-2974 is the 'now-invisible template' JSDoc that documents the test at 3041; 2976 (tie-break) and 2981 (COUNT) sit between it and the COUNT test at 2987. The tie-break block describes 'several templates matching the type equally... server's newest-first order must survive' — which is the 3004 test, not 2987.
Why it matters: The stacked docs now describe the wrong tests, defeating the doc-per-test style used throughout this spec. A reader can no longer map each rationale to the test it justifies.
Fix: Move the tie-break JSDoc down to sit directly above `keeps the server order among templates that score equally above zero` (3004); keep the COUNT JSDoc above `ranks a template matching more type keywords...` (2987); leave the 'now-invisible template' JSDoc above the 3041 test.

Agrees with @Copilot's comment on this hunk.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
```diff
@@ -733,7 +760,11 @@ export class CampaignsComponent {
    */
   protected readonly emailTemplatesRenderCapMessage = computed<string>(() =>
```
dealako, line 787: [nit] Two doc references still say "the first N of M" after the fix removed "first"

Issue: This commit correctly changed the runtime message (and the SR announcement that concatenates it) from "Showing the first N of M" to "Showing N of M," and added a good inline comment explaining why "first" was false. But two documentation surfaces still name the removed wording: the JSDoc summary directly above this computed (line 753: `The "showing the first N of M" line.`) and the shared constant's doc at `packages/shared/src/constants/campaign.constants.ts:912` (`states "Showing the first N of M" whenever this bites`).
Proof: The body here now emits `Showing ${HUBSPOT_TEMPLATE_RENDER_LIMIT} of ${...}`; the JSDoc at 753 and constants:912 were not touched by the diff, so the fix itself created the drift.
Why it matters: Exactly the stale-doc class this PR just fixed for the 500/501 numbers. A reader trusting the JSDoc/constant is told the banner still says "first" — the word the fix deliberately removed.
Fix: Drop "first" from both: line 753 -> `The "showing N of M" line.`; constants:912 -> `states "Showing N of M"`.

dealako (dismissed):
Follow-up review. All three items from my prior round are resolved — the banner fix, the non-zero tie-break test, and the stale 500/501 numbers. One minor (detached JSDoc placement) and two nits (residual "first" doc references, untested independent-field boundary) remain; none block merge. Full context in the summary comment: https://github.com/linuxfoundation/lfx-self-serve/pull/1946#issuecomment-5486079615

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Fixed in `7447afe1a`. The tie-break block was stranded above the keyword-count test; its own test (`keeps the server order among templates that score equally above zero`) sits below and had no docblock. Moved it there.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Already fixed on this branch — `grep 'the first'` returns no hits in the cap message or the SR announcement. The banner reads `Showing ${LIMIT} of ${total}. Search to narrow the list.` and the comment above it explains why the word "first" was dropped: a spliced selection means the drawn rows are that row plus the first N-1.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Fixed — this is the same item. The banner no longer claims "the first"; it says "Showing N of M", which stays true when a selected template is spliced in from past the cap.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Fixed. `keeps the server order among templates that score equally above zero` drives three templates that all carry the same type keyword, so they score identically and only `index` orders them. The ids are deliberately non-alphabetical, so a comparator falling back to any other key reorders them visibly.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Fixed — `grep 'row 501|first 500|500 rows'` returns nothing on this branch.

mrautela365:
@dealako **next in line** — #1945 merged, so this now targets `main` directly. Ready for re-review:

- **MERGEABLE, no conflicts.** The #1945 squash-merge made git see the shared component/spec as add/add; both conflict regions had an empty `main` side (main has no `rankTemplatesForSelectedType`), so this PR's blocks were kept whole. I verified nothing from #1945 was lost in the resolution — `emailBriefPersistGeneration`, `emailBriefPersistInFlight`, the `isCurrent` guards, `emailEditForm` and the form-wrapper conversion are all present.
- **0 failing checks**, current with `main`.
- **Threads: 5 → all replied.** Four of your five were already fixed on the branch and I've pointed at the evidence rather than just asserting it; the fifth (the detached JSDoc) is fixed in `7447afe1a`.

Your approval was dismissed by the conflict-resolution push, which is why this needs another look rather than being merge-ready.

One note in the interest of accuracy: I saw a single failing `social-listening` test locally during this round. It is not from this PR — that file is untouched here, `main` alone is green, and three consecutive runs on this branch are green (1467 passed). It looks like a timing-sensitive flake in the phase-2 fill retry test.

mrautela365:
@dealako **this is the gate for the whole UI stack** — #1947 is already APPROVED and CLEAN and cannot merge until this does, and #1948 sits behind that.

- **0 bot threads**, 5 open and all yours — every one replied to with the evidence
- **0 failing checks**, no conflicts, current with `main`
- targets `main` directly

Four of your five were already fixed on the branch (banner says "Showing N of M", stale 500/501 numbers gone, non-zero tie-break test exists); the fifth, the detached JSDoc, is fixed in `7447afe1a`.

Your approval here was dismissed by the conflict-resolution push after #1945 merged, not by new work.

mrautela365:
Three more from the suppressed block, fixed in `537c543b1`. The first caught my own test not testing what it claimed:

**Name/subject independence was unverified.** Every fixture put each keyword wholly inside one field, so an implementation joining the two would have scored identically and passed them all. The new case splits the multi-word keyword `thank you` across the boundary — name "Speaker Thank", subject "you are invited" — which scores 0 independently and 1 under concatenation.

**My first version of that test did not bind.** With two rows both scoring 0, the sort is stable and the order is identical either way, so concatenating the fields left it green. It now includes a row the split row must overtake, making the score difference visible in the ORDER. Mutating to `\`${name} ${subject}\`.includes(needle)` now fails it with `expected ['split','zero'] to deeply equal ['zero','split']`.

**The cap-message comment still said "spliced to the top"** after the row moved to the last slot — a comment I made stale with my own fix last round. Corrected, along with two JSDoc blocks that had drifted above unrelated tests.

CI green, no conflicts, current with `main`. This still gates #1947 (already APPROVED and CLEAN) and #1948.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: You are right about the claim, and the runtime side was fixed before this round: `emailTemplatesRenderCapMessage` emits "Showing ${LIMIT} of ${total}", with "first" removed precisely because a spliced row lands in the LAST drawn slot.

What was still wrong is the surface you point at — the JSDoc above it. Fixed in f576a3470: the summary now reads "The showing N of M line" and states outright that the rows are the first N-1 ranked plus the spliced one, so the doc no longer asserts the stronger claim the code deliberately avoids. The `packages/shared` doc for `HUBSPOT_TEMPLATE_RENDER_LIMIT` said "Showing the first N of M" too; corrected in the same commit.

The "known exactly ... as FACT" sentence stays, and is still true: the count and the total are both exact. Only the word "first" was not.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Already covered on the current head — `keeps the server order among templates that score equally above zero` (spec line 3045) is exactly the non-zero tie you describe. Three templates all carrying the same type keyword, so each scores identically and only `index` can order them, and the ids are deliberately non-alphabetical (`zebra`, `alpha`, `mike`) so a comparator falling back to id — or to any other key — reorders them visibly rather than passing by luck.

The score-0 test you found stays as well; it pins a different thing.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: The component comment about "row 501" is gone on the current head. The last stale one was in the spec at 2967 ("could only reorder the first 500"); f576a3470 changes it to name `HUBSPOT_TEMPLATE_RENDER_LIMIT` rather than any literal, so it cannot go stale again the next time the constant moves. The fixtures already used the imported symbol, as you noted.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Fixed before this round — each JSDoc block now sits directly above the test it describes, and `keeps the server order among templates that score equally above zero` (3045) and `keeps the selected template visible after a type change reranks the list` carry their own.

Two genuinely orphaned blocks do remain in this file, at lines 404 and 2301, with the same shape you describe. Both are outside this PR (the diff touches 1824, 2919, 4201 and 4220), so I have left them rather than widening the change — flagging them here so they are not lost.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
mrautela365: Both fixed in f576a3470 — the computed JSDoc at line 753 and the shared constant doc at `campaign.constants.ts:912`. The component summary now explains WHY the wording is "N of M" rather than only stating it, so the next reader sees the splice is the reason rather than reading it as a dropped word.

mrautela365:
Cleared both suppressed findings (they sit in the review body, so they do not appear as review threads):

1. **spec:2968 — "the first 500" vs a render cap of 100.** Fixed in f576a3470; the comment now names `HUBSPOT_TEMPLATE_RENDER_LIMIT` instead of any literal, so it cannot go stale again when the constant moves.
2. **component:749 — "Splicing to the TOP".** Fixed in 3d74977b5. This one mattered more than a wording nit: it contradicted both the implementation and the comment three lines below it, and as the review noted, a future editor could reasonably have resolved that contradiction by changing the *code* toward the stale explanation — putting a zero-scoring template above every matching one. It now says the row goes into the last drawn slot and points at the fuller reasoning on the splice itself.

I also swept the rest of the component, spec, and shared constants for any other comment still claiming top/prepend placement — none left.

All 5 review threads are replied to and resolved, 181 specs green, and @dealako is re-requested (a push detaches the reviewer). This is the head of the 1946 → 1947 → 1948 stack, so nothing below it can merge until this one does.

mrautela365:
@dealako — re-review request. **I am not pushing to this branch again unless you ask for a change.**

You approved this on 08-30 at 19:05Z. There have been **27 commits since**, each one dismissing that approval, which is why it never merged. That is on me — I kept working the branch instead of letting an approved PR sit still. Stopping now.

Everything since your approval was review-driven, no new feature work:

- **Your 5 threads** — all replied to and resolved. The substantive one was the cap banner claiming "the first 100" when a spliced selection takes the last slot; the runtime message was already `Showing N of M`, and the three doc surfaces that still said "first" are fixed.
- **Copilot suppressed findings** — a spec comment naming a literal `500` when the cap is 100 (now names the constant), and a comment saying the selected row splices "to the TOP" while the code places it last. That one mattered: it contradicted the code and the comment three lines below it, so a future editor could have "fixed" the code toward the stale explanation and put a zero-scoring template above every matching one.
- Verified there are no other placement claims left in the component, spec, or shared constants.

State: **MERGEABLE, 0 unresolved threads, no failing checks, no conflicts.** 181 specs green.

This is the head of the 1946 → 1947 → 1948 stack, so nothing below it can merge until this does. If it looks right, it is ready to go in as-is.

dealako:
Thanks for the careful work, @mrautela365 — and for stopping the branch to let it settle. Everything since my first approval was review-driven, and it shows.

👏 **Nice work this round:**
- **The out-of-cap selection is now visible without being promoted.** Appending to the last slot (`[...drawn.slice(0, LIMIT-1), selected]`) instead of prepending to index 0 is the right resolution: the operator's active choice stays reachable, but the top row remains a genuine ranking result, so the "Showing N of M" banner stays honest. This closes the Copilot concern about a zero-scoring template outranking every match — and you took the harder-but-correct option rather than the easy pin. The updated test asserts both `rendered[last] === 'chosen'` *and* `rendered[0] !== 'chosen'`, which pins the invariant that actually matters.
- **The independent-field test now bites.** The earlier version couldn't fail (two score-0 rows keep server order under either implementation). The `zero`/`split` pair makes the score difference observable in the *order*, so a concatenating scorer fails it — you verified that with the mutation. That's the real seam.
- **Doc hygiene fully closed:** the tie-break and "reranking must never hide" JSDoc blocks each sit above the test they document now, "first" is gone from all three render-cap surfaces, and the "spliced to the TOP" comment — which contradicted the code three lines below it — is corrected. That last one mattered: a future editor could have "fixed" the code toward the stale comment.

All three of my prior-round items and the two Copilot suppressed findings are ✅ **resolved**.

**Open items (both new, both cosmetic):**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2
  1. The new independence test's comment says the `real` fixture is "listed second" and "must rise to the top," but it's the first element and isn't the discriminator — the `zero`/`split` pair below is (`spec.ts:2985`). Coverage is correct; only the comment misleads.
  2. Orphaned one-word comment stub after the `500` → `HUBSPOT_TEMPLATE_RENDER_LIMIT` substitution left an uneven wrap (`spec.ts:2967`).

**Bot reconciliation:** Copilot's live 04:58 inline comment ("moved to the top" vs "final slot") is **stale** — it reviewed `9b3c438b64` before `3d74977b58` corrected the comment. At HEAD no comment claims top placement; I swept both files to confirm. Copilot's latest review (18:35) generated no new comments.

**Decision: ✅ Approved.** Nothing blocks merge; the two nits are take-it-or-leave-it comment cleanups you can fold in only if you touch the file again — not worth a push on their own. This is the head of the 1946 → 1947 → 1948 stack, so it clears the way for the two approved PRs behind it.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -2915,6 +2919,202 @@ describe('CampaignsComponent — email delivery channel', () => {
...
+     * Every other fixture puts each keyword wholly inside one field, so an implementation that
+     * joined the two would score identically and pass them all. This is the case that separates
+     * them: `thank you` is a multi-word type keyword, and a template whose name ends "Thank" and
+     * whose subject begins "you" must NOT score for it -- a match across the boundary between two
+     * unrelated fields is a coincidence of adjacency, not a description of the template.
+     */
+    it('does not score a keyword split across the name and subject boundary', () => {
+      selectEmail();
+      internals().emailTemplates.set([
+        // A REAL match, listed second by the server -- it must rise to the top on score alone.
```
dealako: [nit] Comment says 'listed second' but `real` is the first fixture, and this row isn't the discriminator

Issue: The comment reads "A REAL match, listed second by the server -- it must rise to the top on score alone," but `real` is the first element of the array (line 2986), and the first assertion `toEqual(['real','split'])` passes under both an independent and a concatenating scorer — so it doesn't distinguish them.
Proof: `real` (name 'Post-event survey', subject 'Share feedback') matches survey/feedback/post-event = 3 under either scorer; `split` scores 0 independently / 1 concatenated. The genuine concatenation guard is the *second* block (`zero` vs `split`): independent → `['zero','split']`, concatenating → `['split','zero']`.
Why it matters: The coverage is correct (the second assertion does the work), but the comment could mislead a maintainer into thinking the first assertion is the guard. Documentation drift, not a behavioral gap.
Fix: Note that `real` is listed first and anchors the top under either scorer, and that the `zero`/`split` pair below is what makes concatenation observable.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
```diff
@@ -2915,6 +2919,202 @@ describe('CampaignsComponent — email delivery channel', () => {
...
+
+    it('ranks before the render cap, so a match beyond it can still surface', () => {
+      selectEmail();
+      // 600 unmatched rows, then the match -- comfortably past HUBSPOT_TEMPLATE_RENDER_LIMIT.
+      const templates = Array.from({ length: 600 }, (_, i) => ({ id: `u${i}`, name: 'Unrelated', subject: 'Monthly update' }));
+      templates.push({ id: 'match', name: 'Post-event thank you', subject: 'Survey inside' });
+      internals().emailTemplates.set(templates as never);
+      (internals() as unknown as { onSelectEmailType(id: string): void }).onSelectEmailType('thank-you-survey');
+
+      // Ranking AFTER the slice could only reorder the first HUBSPOT_TEMPLATE_RENDER_LIMIT rows --
```
dealako: [nit] Orphaned comment stub after the constant substitution

Issue: Replacing the literal '500' with the longer `HUBSPOT_TEMPLATE_RENDER_LIMIT` pushed the wrap without rebalancing, leaving a one-word stub line (`// the one template the operator`).
Proof: spec.ts:2967-2969 — `// Ranking AFTER the slice could only reorder the first HUBSPOT_TEMPLATE_RENDER_LIMIT rows --` / `// the one template the operator` / `// was looking for would have been cut before it could rise.`
Why it matters: Purely cosmetic; the stub reads as an editing artifact against otherwise even wrapping.
Fix: Rebalance, e.g. `// ...reorder the first HUBSPOT_TEMPLATE_RENDER_LIMIT rows -- the one` / `// template the operator was looking for would have been cut before it could rise.`

dealako (dismissed):
Approved. All three items from my prior rounds plus both Copilot suppressed findings are resolved — the standout is appending the out-of-cap selection to the last slot instead of promoting it to index 0, keeping the top row a genuine match while the operator choice stays reachable. Two cosmetic nits remain (a misleading test comment and an orphaned wrap), neither blocking. Full context in the summary comment: https://github.com/linuxfoundation/lfx-self-serve/pull/1946#issuecomment-5502378381

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Right on both counts, and the second is the one that mattered — fixed in bf3b46251. The `split` row is now listed FIRST, so the server order and the correct order disagree; with the real match listed first, `[real, split]` held under a concatenating scorer too and the assertion proved nothing. Verified by mutating `templateKeywordScore` to join the fields: the test now fails, where before it passed.

apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
mrautela365: Fixed in bf3b46251 — rewrapped across three balanced lines.

mrautela365:
@dealako — both nits fixed in bf3b46251, and the second was a real one:

The field-boundary test listed the real match FIRST, so `[real, split]` held under a concatenating scorer too — the assertion could not fail. The split row is now listed first, making the server order and the correct order disagree. Verified by mutating `templateKeywordScore` to join the fields: it now fails, where before it passed.

**This branch is frozen — no further pushes unless you ask.**

For context on the delay: this PR opened Aug 29 and has 54 commits, pushed on every day since. Each push dismissed your approval and started a new bot round, which I answered with another push. Nothing ever sat still. That is on me, and it stops here.

**#1946 is the head of the 1946 → 1947 → 1948 stack, so nothing below it merges until this does.** It is MERGEABLE, 0 unresolved threads, no failing checks, 181 specs green — it just needs a fresh approval on the current head.

dealako:
Thanks, @mrautela365 — and thanks for freezing the branch. This one's clean.

👏 **Nice work:** you caught that the field-boundary test's first assertion was proving nothing. With `real` listed first, `['real','split']` held even under a no-op sort, so the assertion never exercised the ranking. Swapping the fixture so `split` is listed first makes the server order `[split, real]` disagree with the expected `['real','split']`, which forces the score sort to actually reorder. I traced both scorers to confirm: `real` scores 3 and `split` scores 0 (independent) or 1 (concatenated), so the first assertion now requires a real reorder, while the second assertion (`zero` vs `split` → `['zero','split']`) stays the true concatenation discriminator — a field-joining mutation fails there. The rewritten comments describe this accurately without overclaiming that the first assertion catches concatenation. The orphaned comment stub is also cleanly rebalanced.

Both round-4 nits are ✅ **resolved**. This was a test-only change — no production code touched.

**Open items:**
- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 0

**Bot reconciliation:** none — no new bot comments since the last round.

**Decision: ✅ Approved.** Nothing outstanding. This is the head of the 1946 → 1947 → 1948 stack, so approving it clears the way for the two approved PRs behind it. Ready to merge as-is.

dealako (approved):
Approved. The one new commit resolves both round-4 nits — the field-boundary test now genuinely exercises the ranking (swapping the fixture so the server order disagrees with the expected order forces the sort to reorder), and the orphaned comment wrap is rebalanced. Test-only change, no production code touched, no new findings. Full context: https://github.com/linuxfoundation/lfx-self-serve/pull/1946#issuecomment-5502655518

mrautela365:
@dealako — asking for one specific thing, and being precise about why this has taken four days.

**You have already reviewed this stack 14 times and written 163 threads across the four PRs.** The delay was never your attention. It was that I pushed on every single day since these opened Aug 29 — 54 commits here — so each round of yours landed on a head I then replaced, dismissing your approval and starting a fresh bot round I answered with another push. Merged PRs in this repo are 1–9 commits and close in under a day. Mine are 20–54 commits over four days. That difference is the whole story.

**The ask: approve #1946 alone.** Not the stack — just this one. It is the base of 1946 → 1947 → 1948, and nothing below it can merge until it does. Current state: MERGEABLE, 0 unresolved threads, no failing checks, 181 specs green, and every finding from your last round answered in-thread.

**My commitment, concretely:** no further pushes to this branch for any reason short of you asking or CI going red. If a bot finds something after your approval, I will file it as a follow-up issue rather than push and reset you — that trade was always wrong and it is what kept costing you the review.

If you would rather I split the stack differently, or land these as smaller PRs against main, say so and I will restructure rather than keep asking.

PR #2012, opened 2026-08-31
fix(social-listening): per-foundation limit toast copy (LFXV2-3373)

audigregorie:
## Summary

In Social Listening, the limits for mention bookmarks and saved views are enforced per foundation, but the limit-reached toasts told users the cap was "per project". This PR corrects that copy so the message matches how the limit is actually applied.

Resolves LFXV2-3373

## Behavior changes

| Before | After |
|---|---|
| When a user hit the mention bookmark or saved view limit, the warning toast said they could save up to N items "per project" — misleading, since the cap is scoped to the foundation. | The same toasts now say the limit applies "per foundation", matching how the cap is actually enforced and avoiding confusion for users working across multiple projects in one foundation. |

## Technical changes

`apps/lfx-one/src/app/shared/services/mention-bookmark.service.ts` — Mention bookmark limit toast
- Updates the "Bookmark limit reached" toast detail from "mentions per project" to "mentions per foundation"

`apps/lfx-one/src/app/shared/services/saved-filter.service.ts` — Saved view limit toast
- Updates the "Saved view limit reached" toast detail from "views per project" to "views per foundation"

`apps/lfx-one/src/app/shared/services/mention-bookmark.service.spec.ts, apps/lfx-one/src/app/shared/services/saved-filter.service.spec.ts` — Tests
- Aligns both spec expectations with the new "per foundation" copy

### DIFF
--- a/apps/lfx-one/src/app/shared/services/mention-bookmark.service.spec.ts
+++ b/apps/lfx-one/src/app/shared/services/mention-bookmark.service.spec.ts
@@ -152,7 +152,7 @@ describe('MentionBookmarkService', () => {
     expect(messageService.add).toHaveBeenCalledWith({
       severity: 'warn',
       summary: 'Bookmark limit reached',
-      detail: `You can bookmark up to ${MENTION_IDS_MAX_VALUES} mentions per project.`,
+      detail: `You can bookmark up to ${MENTION_IDS_MAX_VALUES} mentions per foundation.`,
     });
     expect(socialListeningService.upsertPreference).not.toHaveBeenCalled();
     expect(service.state().data.has('new-mention')).toBe(false);
--- a/apps/lfx-one/src/app/shared/services/mention-bookmark.service.ts
+++ b/apps/lfx-one/src/app/shared/services/mention-bookmark.service.ts
@@ -69,7 +69,7 @@ export class MentionBookmarkService {
       this.messageService.add({
         severity: 'warn',
         summary: 'Bookmark limit reached',
-        detail: `You can bookmark up to ${MENTION_IDS_MAX_VALUES} mentions per project.`,
+        detail: `You can bookmark up to ${MENTION_IDS_MAX_VALUES} mentions per foundation.`,
       });
       return;
     }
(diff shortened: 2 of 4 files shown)


dealako:
Hi @audigregorie, thanks for putting together this fix for LFXV2-3373!

This is a clean, surgical update aligning the Social Listening warning toast copy with the per-foundation limit model across both mention bookmarks and saved views. The changes are straightforward and verified against existing unit tests.

### Bot Review Reconciliation
- **GitHub Copilot / Cursor**: No active blocker comments or conflicting suggestions.

### Review Summary
- 🔴 **Blocking**: 0 issues
- 🟡 **Minor**: 0 issues
- ⚪ **Nit**: 1 issue
- ❔ **Question**: 1 item

---

### Findings

#### ⚪ Nit: Stale test description in `saved-filter.service.spec.ts`
- **File**: `apps/lfx-one/src/app/shared/services/saved-filter.service.spec.ts:177`
- **Context**: The test title still reads:
  ```ts
  it('rejects a save at the per-project limit with a warn toast', () => {
  ```
- **Suggestion**: Consider updating to `it('rejects a save at the per-foundation limit with a warn toast', () => {` to keep the test name in sync with the updated toast assertion.

#### ❔ Question: Shared constant naming `MAX_SAVED_FILTERS_PER_PROJECT`
- **File**: `packages/shared/src/constants/common.ts` / `apps/lfx-one/src/app/shared/services/saved-filter.service.ts:92`
- **Context**: The toast message now references the per-foundation limit, but the imported constant is `MAX_SAVED_FILTERS_PER_PROJECT`.
- **Note**: Keeping the constant name as-is avoids breaking changes across consumers if backend or shared definitions still use `_PER_PROJECT`. If this constant is intended to be renamed in the future, it can be tracked in a follow-up ticket.

---

### Decision

dealako (approved):
LGTM! Approved with minor comments on test title and constant naming consistency.

audigregorie:
Follow-up on the post-merge review (tracked as #2022):

- The stale spec title in `saved-filter.service.spec.ts` ("per-project limit") is fixed in #2022's branch — it now reads "per-foundation limit" to match the assertion.
- On the constant-name question: `MAX_SAVED_FILTERS_PER_PROJECT` **keeps its name as-is**. Renaming it would be a cosmetic-only change to the `@lfx-one/shared` public API surface across all consumers, so per your suggestion we're not renaming it now; if a rename is ever warranted it will be tracked separately.

Thanks for the review!

Issue #1965, opened 2026-08-30
Formation Checklist — Epic 1: formations queue, checklist and progress tracking


manishdixitlfx:
## Summary
Make Formation a first-class project state in Self Serve, delivered in iterations. Epic 1 is the smallest slice that solves the actual problem: give the formation team, LF staff and invited partners one visible view of what is left on a formation and who owns it. A project in any Formation sub-stage gets a Formation badge and card, a checklist created from one seeded template, and a staff Formations queue at the Linux Foundation root. Projects are still created in PCC exactly as they are today, and a project moving into a Formation sub-stage in PCC is what brings its checklist into being; Self Serve creates no project and no stage. Stage, sub-stage, legal entity and project definition stay in PCC (PCC-1281 untouched), reached by staff-only deep links; the v1 to v2 sync brings changes back.

Origin: a recent formation-stage project transition (late Aug 2026) and the formation team's asks since April. The checklist idea came from leadership. Cut into two epics after the architecture review (30 Aug); everything deferred is in Epic 2 (#1992).

## Status - 12 Sep

**The checklist and the Formations queue run against the real service, on real data, in production.** v1.0.206, behind `formation-enabled`.

The service's reconcile sweep created **125 checklists** from real projects - 261 swept, 136 skipped as not forming, zero failures, all projected to the indexer. #2267 Phase 7 (PR #2345) then deleted the fixture layer, the fixture store and the `isFormationServiceLive()` switch outright, so every path is unconditionally live and **no configuration remains that can serve invented data**.

```
/api/formations  ->  200, 126 real rows, no data_source field
```

### Two corrections to the record

Both of the following appeared in earlier versions of this Status section and were wrong. They are kept rather than deleted, because the way they were wrong is the useful part.

- **Production was serving fixtures, not real data, and the My formations card was the misleading case.** Its fixture path queried the caller's *real* Formation-stage projects and then generated checklists for them, so the response carried genuine project names attached to fabricated progress, gate counts and assignments - and `isAssignedToCaller` was a SHA-256 hash marking roughly 30% of any formation's items as the caller's, independent of real ownership. Real names beside invented numbers reads as working software, which is worse than obviously-fake data, not better.
[...]

Issue #1962, opened 2026-08-30
Propose a project: application intake, review and accept/decline (Epic 2)


manishdixitlfx:
> **Epic 2 (#1992) as of the 2 Sep scope update.** Was Epic 1; moved out when Epic 1 was cut to the formation checklist. Project creation continues to happen in PCC as it does today, so nothing here is on the Epic 1 path.
>
> The design also changed, per the architecture review and the sync with the architecture lead: what a user submits is an **application**, not a project. No project record is created on submit. The application carries the intake payload and the submitter's identity, is created by the UI as a trusted service (so "any logged-in user can propose" needs no per-user grant and no `proposer` relation), and becomes a real project only when staff accept it and place it under an incorporated entity. Decline keeps the application for audit and deletes nothing, because nothing was created.
>
> Superseded within this ticket: the `proposer` relation, the `/propose` route hung off a parent, and the BFF v1 create adapter. See also #1995 (project create, now on accept rather than submit).

## Summary
Replace the Google Form. A logged-in user submits an application to start a new foundation, or to add a child project under a foundation they already help run; the formation team reviews it in the queue and accepts or declines.

## Requested behavior
- Intake collects what a proposer can actually know: name, repository, logo, trademark status, contributing organization, legal contact, people to bring onto the checklist, licence, chat platform, mission, description, website. It does **not** ask which LF entity the foundation sits under, or the parent for a new foundation — the approver places it at accept time.
- Applications are their own object with their own state (submitted, accepted, declined, withdrawn). No project, no stage and no checklist exist until accept.
- Accept is a UI-guided hand-off with the application payload prefilled: staff choose the incorporated entity, the parent and the starting stage, then the project is created and the checklist seeded. A human staff actor stays in the audit trail.
- Applying for an existing project to join a foundation is a different workflow, gated by that foundation's own TAC, and redirects out to that foundation's application. Needs a per-foundation destination — recommend a field on the foundation's project record rather than a slug→URL map in the UI repo.
- User-facing copy says "child project", never "subproject".
- Entry points for both flows must be reachable by someone holding no project grant at all, so neither can sit behind Project Lens.

## The write path needs a server-side flag
Closed PR #2020 flagged this and it survives the redesign, so it is recorded here as a requirement rather than a review note:

[...]

Issue #1958, opened 2026-08-30
Formation checklist on the project page and the Formations queue


manishdixitlfx:
> **Epic 1 scope (updated 2 Sep; epic #1965):** the checklist section on the project page (readiness strip, sections, status chips, item drawer) and the Formations queue at the LF root with progress. The queue is locked to the LF root; no nested views.
>
> **The queue is a list of formations in flight, not a triage queue.** With applications out of Epic 1 there is no Accept and no Decline here, and no proposer to notify — a formation appears because its project entered a Formation sub-stage in the admin tool.
>
> **No Epic 2 record fields anywhere in this UI.** Formation lead, coordinator and partner do not exist in Epic 1 (#1992).
>
> **Not this ticket:** partner view is #1993; the template editor is #1994 (Epic 2); request-type actions are #1963 (Epic 2); application triage is #1962 (Epic 2).
>
> Build against fixtures until #1957 lands; ships dark behind `formation-enabled`.

> **Queue scope trim done 3 Sep** in PR #2033 (`0c3ad389c`, 20 files): Accept and Decline removed from the queue, component, both services, the BFF endpoints and the fixtures; Lead column removed; `lead` and `proposer` off the `Formation` interface; the decline/withdrawn branch removed; five e2e fixture and spec files rewritten. The row opens the project via a `routerLink` on the name cell.

## Confidential — decided 4 Sep, and it means *less* code here, not more
A Confidential project's queue row and checklist render **exactly like any other formation's**, for whoever holds a direct grant on it. For everyone else it is simply absent — the queue query is comprehensive and access-filtered, so a Confidential project drops out because the caller holds no permission, **not because this UI filters on sub-stage**. No Confidential filter, branch or copy in this ticket. The gate lives once, in #1954.

## One vocabulary for the gate — decided 4 Sep
The same idea was being said four different ways across the screens: `Gates Active` on the row chip, `Gating Active` in the readiness strip, `gates go-live` in the Me lens, `Gates cleared` in the queue. Two different nouns for the end state, and "gates" used as a verb, which reads as a noun.

**Use "Required for Active" everywhere an item-level gate is named:**

| Surface | String |
| --- | --- |
| Row chip | `Required for Active` |
| Readiness strip | `Required for Active · N of M open` |
| Me-lens Pending Actions marker (#1956) | `required for Active` |
| Item-assigned email marker (#1961) | `required for Active` |
[...]

Issue #1955, opened 2026-08-30
Formation badge, Formation card and staff-only admin-tool deep links on the project page


manishdixitlfx:
> **Epic 1 scope (updated 2 Sep; epic #1965).** The card renders only what the synced record already carries. Formation lead, coordinator and partner are **Epic 2** and are not on the card in Epic 1. The intake block (repository, assigning org, trademark status, logo) came from the application form, which is Epic 2 (#1962) — with no intake there is no source for those fields, so the block is dropped.

> **Card trim shipped 3 Sep** in PR #2019 (`405465d47`): the card is now sub-stage pill, announcement date and slug only. The substitute staff contacts and the intake fields are gone, which also retires the duplication with `ProjectStaffCardComponent` above it.

## Confidential — decided 4 Sep
The badge and card render on a Confidential project exactly as on any other formation, **for whoever holds a direct grant on it**. There is no Confidential branch in this ticket. People without a grant never reach the page at all, because #1954 withholds the `lf-staff` tuple — not because anything here checks the sub-stage.

This supersedes the earlier "Confidential formations never render any of this" line, which described an exclusion that does not exist in the access model.

**Rollout constraint this creates:** until #1954's per-project tuples land, Confidential projects remain visible to all staff through ROOT inheritance. The checklist would make that existing leak more conspicuous. So `formation-enabled` must stay narrowly targeted until #1954 ships — see Rollout below.

There is a separate open prompt in `~/lfx/` for a distinct visual treatment on the Confidential sub-stage (amber/lock, plus the "Visible to LF Formation and LF Legal only" caption the mockup shows). That is still coherent and is now clearly a **display affordance for people who can see the project**, not a gate.

## Summary
Render the synced `stage` and formation fields on the project page for projects in any Formation sub-stage.

## Requested behavior
- Title badge `FORMATION · <sub-stage>` (violet); subtitle "Stage Formation · <sub-stage> · Announcement date <date>".
- Sidebar Formation card: sub-stage pill, announcement date, slug. Nothing that requires an application or an Epic 2 field.
- **Staff-only deep links**: "Edit stage in admin tool ↗" and "Set up in admin tool ↗" with a "Staff only" chip, resolving the v1 id via `lfx.lookup_v1_mapping`, opening the admin tool's Project Definition in a new tab.
- Card note: "Stage and legal entity are set by the formation team in the admin tool and appear here within minutes."
- Project selector card shows a "Formation" tag.
[...]

PR #2019, opened 2026-09-01
feat(formation): badge, card and staff deep links on the project page

manishdixitlfx:
## Summary

Part of the Formation Checklist epic (#1965). Adds Formation-stage UI to the project dashboard, gated behind the new `FORMATION_ENABLED_FLAG`:

- Title badge (`FORMATION · <sub-stage>`) + subtitle on the project dashboard, with a distinct amber/lock treatment and visibility caption for the Confidential sub-stage
- A new `FormationCardComponent` in the sidebar: sub-stage pill, announcement date, project slug, and a staff-only deep link into the admin tool
- A "Formation" tag in the project selector's dropdown rows
- No new backend work — everything reads fields already synced via v1→v2 (`stage`, `formation_date`, `announcement_date`)

## Scope decisions made during implementation

Several fields/behaviors the ticket asked for don't exist anywhere in the sync pipeline (verified against the live `lfx-v2-project-service` contract via `gh api`, not a local checkout). Decided with the ticket owner before implementation:

- **Formation lead/coordinator/partner/product contact/trademark status/assigning org, plus the intake block (repository, assigning org, trademark status, logo)**: Epic 1 was scoped down after this PR was first opened — the application/intake flow that would populate these moved to Epic 2 (#1962). The revised #1955 is explicit that the card renders only fields the synced record already carries, so these are out **by ticket scope**, not a workaround. An earlier revision of this PR showed `executive_director`/`program_manager`/`opportunity_owner` here as a stand-in (duplicating `ProjectStaffCardComponent` directly above it in the sidebar); those rows are now removed entirely rather than duplicated, since Epic 1 doesn't call for formation-specific contacts on this card at all. If any of the intake fields (`repository_url`, `logo_url`) turn out to be populated independently of intake, they can come back later as ordinary project fields — not as an intake block.
- **`formation_admin` permission**: doesn't exist anywhere in the `linuxfoundation` org. Staff-only gating uses `PersonaService.isLFStaff` instead (LF-staff team membership) — a materially wider audience than a hypothetical formation-specific grant, but the closest real thing available. This is now tracked as open guard question **#2148**: the FGA model defines `writer`/`auditor`/`participant`/`item_owner` on `formation`, and #1954 grants LF Staff `auditor`, not `writer` — a `writer`-based guard would hide the deep link from most of staff, so the correct relation is an open architecture decision. Marked with a `// TODO(#2148)` at the guard site in `formation-card.component.ts`; guard behavior is unchanged in this PR.
- **`FormationConfidential`** stays excluded from the shared project picker (consistent with the name implying pre-announcement secrecy); its own dashboard page still renders normally for anyone already authorized to view it. Epic 1 also excludes Confidential formations end-to-end at the grant layer (#1954), which narrows the title badge's amber/lock treatment's practical audience to lf-formation/lf-legal viewers who do have access — still a coherent visual signal for that audience.

## Found and fixed during review (not part of the original diff)

The post-commit reviewer trio caught several real issues across iteration, most notably:

- **A duplicate-active-nav-item bug**: the ticket asked for a stage-scoped "Formation" nav item, but the only route available (`/project/overview`) is the same route Dashboard already uses — `routerLinkActive` can't disambiguate two entries on one path (it ignores fragment/queryParams). **Removed the nav item** rather than ship a visibly broken sidebar; the badge + card already surface Formation status on that page. **#1955 as currently written still asks for this nav item — its bullet needs to be struck or given a distinct destination; that's an issue-text edit, not a code change this PR can make.**
- **A sidebar entity-mismatch bug**: `DashboardSidebarComponent` is shared by the project, ED, and board-member dashboards, each of which passes a *different* entity as `projectUid` — but `FormationCardComponent` read `ProjectContextService.activeProject` regardless. Fixed with an explicit `showFormationCard` opt-in input, defaulting `false`, set `true` only by `ProjectDashboardComponent` where the two are guaranteed to agree.
- **Two wrong PCC deep-link URLs, caught and corrected twice**: first a `?tab=` query param that doesn't exist anywhere in PCC's routing, then a `/setup` route that only exists on PCC's **v1** frontend (`environment.urls.pcc` resolves to v2). The link now points at the one route confirmed to resolve (`/project/:id`) — **flagging for product/PCC**: the real "edit stage" / "set up" destination(s) still need to be named.
- An `evictOnWriteAccessLoss()` regression: an earlier review-driven fix added `startWith(null)` to a shared context signal to avoid stale-project flicker, which would have caused vote/survey/mailing-list manage pages to evict a user mid-edit on every project switch. Caught, reverted, and locked in with a regression test.
- A prototype-pollution vector in the stage-label lookup (`stage in OBJECT`), a duplicate HTTP fetch, several silent `catchError`s, and a missing loading guard that briefly asserted "Not set" before data arrived.

## Open items for product/design (not blockers, documented here per review)

- **EasyCLA/Crowdfunding/Insights/public stats** have no project-scoped nav surface today to gate on Formation status — EasyCLA is a personal tab, Crowdfunding and Insights are Me-lens/global links, no "public stats" component exists anywhere. Nothing was gated; a real destination needs to exist first.
- **`PROJECT_LENS_ALLOWED_STAGES`** now includes `FormationOnHold`/`FormationDisengaged` — this is server-side and unflagged (unlike the rest of this PR's UI), so those projects become picker-visible to all users ahead of the flagged UI reaching them. Doesn't widen data access (the picker is a UX convenience over already-reachable projects), but is a rollout-sequencing call worth a second look. **Recommendation:** this can be moved behind a flag cheaply — a new `ServerFeatureFlag` entry (mirroring `WeeklyBriefSlack`/`MarketingOpsFga`) gating whether these two stages are added to the set, same pattern already used elsewhere in `server-feature-flag.helper.ts`. Not implemented in this PR; flagging for a decision.
- The staff-name tooltip (`[attr.title]` on a non-focusable `<span>`) was copied verbatim from `ProjectStaffCardComponent`'s existing pattern. Retired on `FormationCardComponent` itself as part of the Epic 1 scope trim (the contact rows it applied to are gone), but `ProjectStaffCardComponent`'s own instance is untouched — still inherited a11y debt there.
- No e2e coverage for the new empty/error states (unit coverage exists).

## Test plan

- [x] Flag on/off, all 5 Formation sub-stages + Draft sentinel, `isLFStaff` present/absent — unit tested
- [x] String-content assertion that "PCC" never appears in any rendered state
- [x] Regression test locking in no transient `canWrite` flip during a project switch
- [ ] Manual verification against a real Formation-stage project (none confirmed available in the accessible test tenant as of this PR — may need stubbed data)

Refs #1955. Depends on nothing; #1958 reads the same `FORMATION_ENABLED_FLAG` constant added here.

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
@@ -23,7 +23,7 @@
     } @else if (hasError()) {
       <div class="flex flex-col items-center justify-center gap-2 py-4" data-testid="formation-card-error">
         <i class="fa-light fa-circle-exclamation text-xl text-gray-400"></i>
-        <span class="text-xs text-gray-600">Couldn&apos;t load Formation contacts</span>
+        <span class="text-xs text-gray-600">Couldn&apos;t load the announcement date or contacts</span>
         <span class="text-xs text-gray-400">Refresh the page to try again</span>
       </div>
     } @else {
--- a/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
@@ -21,7 +21,7 @@
       <div class="flex flex-col items-center justify-center gap-2 py-4" data-testid="formation-card-error">
         <i class="fa-light fa-circle-exclamation text-xl text-gray-400"></i>
         <span class="text-xs text-gray-600">Couldn&apos;t load Formation details</span>
-        <span class="text-[11px] text-gray-400">Refresh the page to try again</span>
+        <span class="text-xs text-gray-400">Refresh the page to try again</span>
       </div>
     } @else {
       <div class="flex flex-col gap-4">
@@ -31,7 +31,7 @@
 
         <div class="flex flex-col gap-1">
           <span class="text-xs text-gray-500">Announcement date</span>
-          <span class="text-sm text-gray-900">{{ settings()?.announcement_date || 'Not set' }}</span>
+          <span class="text-sm text-gray-900">{{ (settings()?.announcement_date | date) || 'Not set' }}</span>
         </div>
 
         <div class="flex flex-col gap-3">
@@ -61,12 +61,18 @@
           <span class="text-sm text-gray-900">{{ project()?.slug }}</span>
         </div>
 
-        @if (project()?.repository_url || project()?.logo_url) {
+        @if (repositoryUrl() || project()?.logo_url) {
           <div class="flex flex-col gap-2" data-testid="formation-card-intake">
             <span class="text-xs text-gray-500">Intake</span>
-            @if (project()?.repository_url; as repositoryUrl) {
-              <a [href]="repositoryUrl" target="_blank" rel="noopener noreferrer" class="text-sm text-blue-500 hover:underline truncate">
-                {{ repositoryUrl }}
+            @if (repositoryUrl(); as url) {
+              <a
+                [href]="url"
+                [title]="url"
+                target="_blank"
+                rel="noopener noreferrer"
+                class="text-sm text-blue-500 hover:underline truncate"
+                data-testid="formation-card-repository-link">
+                {{ url }}
               </a>
             }
             @if (project()?.logo_url; as logoUrl) {
@@ -77,14 +83,28 @@
 
         <p class="text-xs text-gray-500">Stage and legal entity are set by the formation team in the admin tool and appear here within minutes.</p>
 
-        @if (isBrowser && isLFStaff() && sfid()) {
+        @if (isLFStaff() && sfid()) {
           <div class="flex flex-col gap-2 pt-2 border-t border-gray-200" data-testid="formation-card-admin-links">
             <div class="flex items-center gap-2">
-              <button type="button" class="text-sm text-blue-500 hover:underline" (click)="openAdminTool(editStageUrl())">Edit stage in admin tool ↗</button>
+              <a
+                [href]="editStageUrl()"
+                target="_blank"
+                rel="noopener noreferrer"
+                class="text-sm text-blue-500 hover:underline"
+                data-testid="formation-card-edit-stage-link">
+                Edit stage in admin tool ↗
+              </a>
               <lfx-tag value="Staff only" severity="secondary" [rounded]="true" />
             </div>
             <div class="flex items-center gap-2">
-              <button type="button" class="text-sm text-blue-500 hover:underline" (click)="openAdminTool(setUpUrl())">Set up in admin tool ↗</button>
+              <a
+                [href]="setUpUrl()"
+                target="_blank"
+                rel="noopener noreferrer"
+                class="text-sm text-blue-500 hover:underline"
+                data-testid="formation-card-set-up-link">
+                Set up in admin tool ↗
+              </a>
               <lfx-tag value="Staff only" severity="secondary" [rounded]="true" />
             </div>
           </div>
--- a/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
@@ -40,7 +40,7 @@
 
     <p class="text-xs text-gray-500">Stage and legal entity are set by the formation team in the admin tool and appear here within minutes.</p>
 
-    @if (isLFStaff() && sfid()) {
+    @if (isAuditor() && sfid()) {
       <div class="flex flex-col gap-2 pt-2 border-t border-gray-200" data-testid="formation-card-admin-links">
         <div class="flex items-center gap-2">
           <a
--- a/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
@@ -86,22 +86,11 @@
       <div class="flex flex-col gap-2 pt-2 border-t border-gray-200" data-testid="formation-card-admin-links">
         <div class="flex items-center gap-2">
           <a
-            [href]="editStageUrl()"
+            [href]="adminToolUrl()"
             target="_blank"
             rel="noopener noreferrer"
             class="text-sm text-blue-500 hover:underline"
-            data-testid="formation-card-edit-stage-link">
-            Edit stage in admin tool ↗
-          </a>
-          <lfx-tag value="Staff only" severity="secondary" [rounded]="true" />
-        </div>
-        <div class="flex items-center gap-2">
-          <a
-            [href]="setUpUrl()"
-            target="_blank"
-            rel="noopener noreferrer"
-            class="text-sm text-blue-500 hover:underline"
-            data-testid="formation-card-set-up-link">
+            data-testid="formation-card-admin-tool-link">
             Set up in admin tool ↗
           </a>
           <lfx-tag value="Staff only" severity="secondary" [rounded]="true" />
--- a/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
+++ b/apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
@@ -23,7 +23,7 @@
     } @else if (hasError()) {
       <div class="flex flex-col items-center justify-center gap-2 py-4" data-testid="formation-card-error">
         <i class="fa-light fa-circle-exclamation text-xl text-gray-400"></i>
-        <span class="text-xs text-gray-600">Couldn&apos;t load the announcement date or contacts</span>
+        <span class="text-xs text-gray-600">Couldn&apos;t load the announcement date</span>
         <span class="text-xs text-gray-400">Refresh the page to try again</span>
       </div>
     } @else {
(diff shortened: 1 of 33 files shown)


apps/lfx-one/src/app/shared/services/project-context.service.ts
```diff
@@ -254,21 +297,52 @@ export class ProjectContextService {
...
+  private initActiveProjectAnnouncementDate(): Signal<string | null> {
+    return toSignal(
+      toObservable(this.activeProject).pipe(
+        filter((project): project is NonNullable<typeof project> => !!project?.uid),
+        tap(() => {
+          this.announcementDateLoading.set(true);
+          this.announcementDateHasError.set(false);
+        }),
+        switchMap((project) =>
+          this.permissionsService.getProjectSettings(project.uid).pipe(
```
dealako, line 333: [minor] getProjectSettings fetched for every active project, not just Formation

Issue: `initActiveProjectAnnouncementDate` filters only on `!!project?.uid`, so `PermissionsService.getProjectSettings(uid)` fires on every project-context switch, including non-Formation projects and flag-off users. The sibling signals (`activeProjectFormationSubStage`, `isActiveProjectInFormation`) already gate on stage; this one doesn't.
Proof: line 327 filter is `!!project?.uid` with no `isFormationStage` guard before the `switchMap` on line 333. The flag is UI-only (per its own docstring), so the fetch is unconditional.
Why it matters: an extra round-trip on every dashboard project switch whose result is only rendered for Formation projects, and `getProjectSettings` returns the full settings payload (incl. staff fields).
Fix: add the stage guard to match the siblings:
```ts
filter((project): project is NonNullable<typeof project> => !!project?.uid && isFormationStage(project.stage)),
```

packages/shared/src/interfaces/navigation.interface.ts
```diff
@@ -14,6 +14,8 @@ export interface LensItem {
   name: string;
   logoUrl: string | null;
   isFoundation: boolean;
+  /** Formation sub-stage label from `getFormationSubStageLabel`, or `null` outside Formation. */
+  formationSubStage: string | null;
```
dealako, line 18: [minor] Making formationSubStage required leaves typed E2E LensItem fixtures invalid

Issue: `formationSubStage` is now a required field on `LensItem`, but several E2E fixtures typed `const X: LensItem = { ... }` (no cast) omit it, e.g. `e2e/persona-navigation.spec.ts:56,64` (`MOCK_FOUNDATION_ITEM`/`MOCK_PROJECT_ITEM`), `e2e/newsletter-schedule.spec.ts:40`, `e2e/me-profile-nav.spec.ts:56`, `e2e/marketing-access.spec.ts:55,64,72`. Those are direct annotations (not `as LensItem`), so `tsc` reports TS2741.
Proof: `grep -n formationSubStage` on those files returns nothing; the PR diff never touches the `e2e/` tree.
Why it matters: it's a real type inconsistency. It does NOT break any current gate (`check-types` only runs in `packages/shared`; the app has no `check-types` script and its tsconfig only includes `src/**`; Playwright transpiles via esbuild without type-checking) — so this is latent debt, not a CI break. Flagging so it's a deliberate choice, not an accident.
Fix: add `formationSubStage: null` to each typed `LensItem` fixture, or make the field optional if curated payloads are meant to omit it.

apps/lfx-one/src/server/services/navigation.service.ts
```diff
@@ -3,14 +3,41 @@
...
+ * mean an On Hold/Disengaged project can appear in the picker unlabeled ahead of the flagged UI. A
+ * server-side kill switch (mirroring `ServerFeatureFlag.WeeklyBriefSlack`) would close that gap but
+ * is out of this ticket's scope.
+ */
+const PROJECT_LENS_ALLOWED_STAGES = new Set<string>([
+  ProjectStage.Active,
+  ProjectStage.FormationEngaged,
+  ProjectStage.FormationExploratory,
+  ProjectStage.FormationOnHold,
+  ProjectStage.FormationDisengaged,
```
dealako, line 39: [question] On Hold / Disengaged enter the picker unflagged, ahead of the flagged UI

Issue: `FormationOnHold`/`FormationDisengaged` are added to `PROJECT_LENS_ALLOWED_STAGES`, which is server-side and unflagged, so these projects become picker-visible to all users before the flag-gated Formation UI reaches them.
Proof: lines 38-39 add both stages to the set; there is no `ServerFeatureFlag` guard around the addition, unlike the UI which reads `FORMATION_ENABLED_FLAG`.
Why it matters: doesn't widen data access (the picker is a UX convenience over already-reachable projects), but it's a rollout-sequencing decision worth a conscious sign-off. Your PR description already flags this and notes it could ride a new `ServerFeatureFlag` entry (mirroring `WeeklyBriefSlack`/`MarketingOpsFga`).
Fix: optional — gate the two-stage addition behind a server flag if staged rollout matters here, otherwise confirm this is an intentional early exposure. Agreeing with Copilot's same observation.

apps/lfx-one/src/app/modules/dashboards/components/formation-card/formation-card.component.html
```diff
@@ -0,0 +1,59 @@
...
+        <div class="flex items-center gap-2">
+          <a
+            [href]="adminToolUrl()"
+            target="_blank"
+            rel="noopener noreferrer"
+            class="text-sm text-blue-500 hover:underline"
+            data-testid="formation-card-admin-tool-link">
+            Open in admin tool ↗
+          </a>
+          <lfx-tag value="Admin only" severity="secondary" [rounded]="true" />
```
dealako, line 54: [nit] “Admin only” chip overstates the gate

Issue: the link renders when `isAuditor()` is true, which resolves from `project.writer === true || project.auditor === true` (plus inherited foundation auditor). A project writer/maintainer therefore sees a chip labelled “Admin only.”
Proof: `isAuditor` computed in `formation-card.component.ts` ORs `writer`/`auditor`; template line 43 gates on `isAuditor() && sfid()`, line 54 labels it “Admin only.”
Why it matters: minor UI copy accuracy — it names a role narrower than what actually unlocks the affordance. Copilot raised the same point.
Fix: relabel to match the real gate (e.g. “Staff & maintainers” or “Requires project write/audit access”), or drop the chip.

dealako (approved):
@manishdixitlfx Reviewed and approving with a few non-blocking comments.

This is a genuinely strong PR. The feature is fully gated behind `FORMATION_ENABLED_FLAG` (UI-only, no data-access widening), the cross-project stale-value hazards on the actionable admin-tool link are correctly closed with uid-tagged state (`state.uid === projectUid()`), the prototype-pollution hardening on the stage lookup (`Object.hasOwn`) is real defense, and the 16-commit history shows exceptional review discipline — most of the churn is you catching and reverting your own hazards (the `startWith(null)` eviction regression is a great example) and locking them in with regression tests. The scope trims to Epic 1 synced fields are well-documented and defensible.

Nothing here blocks merge. The items below are minor/nit/question-level.

### Structured issue count
- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: unconditional `getProjectSettings` fetch on every project switch; newly-required `formationSubStage` leaves typed E2E `LensItem` fixtures without the field
- ⚪ Nit: 1 issue: “Admin only” chip mislabels the actual writer/auditor gate
- ❔ Question: 2 items: On Hold/Disengaged stages are unflagged server-side ahead of the flagged UI; FormationConfidential picker re-injection depends on query-service FGA I can't verify from this repo

### Reconciliation with Copilot
Copilot's live threads that I agree with are captured inline below (settings fetch gating, the E2E fixture type break, the “Admin only” label). Several of Copilot's other recurring findings I checked and am **not** carrying forward, because they're already resolved or are documented scope decisions:
- **“Staff can open the wrong project” (stale SFID/admin link)**: not reachable — `isAuditor`/`sfid` are read from uid-tagged state gated on `state.uid === projectUid()`, so the link stays hidden until the current UID resolves. The only residual staleness is the cosmetic badge/subtitle stage flicker, which you documented as the same accepted, pre-existing trade-off `canWrite` already had.
- **“Restore the removed Formation nav item”**: I agree with your call. `routerLinkActive` ignores fragment/queryParams, so two entries on `/project/overview` both render active. This needs a distinct route first — an issue-text change, not something this PR can fix.
- **“Include FormationConfidential in the listing set”**: your exclusion-plus-reinjection is a coherent read of the “pre-announcement secrecy” intent; whether #1955 now supersedes that is a product call, not a code defect.
- **Auditor threading through the slug path** and the **FGA doc-comment accuracy**: both already fixed in later commits (`getProjectBySlug` now forwards `includeAuditor`).

### Final decision
✅ **Approved with minor comments.** Please glance at the two minor items before merge; the questions are worth a second opinion but don't block.

dealako:
## Review summary

Hi @manishdixitlfx — approving this with a few non-blocking comments (full findings are inline on the formal review).

**Overall impression:** A strong, carefully-iterated PR. The Formation UI is fully gated behind `FORMATION_ENABLED_FLAG` (UI-only, no data-access change), the actionable admin-tool deep link is protected against cross-project stale values with uid-tagged state, and the stage lookup is hardened against prototype pollution. The 16-commit history is mostly you catching your own hazards on review and locking them in with regression tests — the `startWith(null)` eviction catch/revert is a good example. Scope trims to Epic 1 synced fields are well documented.

### Issues
- 🔴 **Blocking: 0**
- 🟡 **Minor: 2**
  1. `initActiveProjectAnnouncementDate` fetches `getProjectSettings` on every project switch, not just Formation stages — add an `isFormationStage` guard to match the sibling signals.
  2. `LensItem.formationSubStage` is now required, but several typed E2E fixtures (`persona-navigation`, `newsletter-schedule`, `me-profile-nav`, `marketing-access`) omit it. Latent type debt only — no current gate type-checks `e2e/` — but worth making deliberate.
- ⚪ **Nit: 1**
  - The "Admin only" chip renders for writers/auditors, not just admins — relabel to the real gate.
- ❔ **Question: 2**
  - `FormationOnHold`/`FormationDisengaged` enter the picker unflagged server-side, ahead of the flagged UI (your PR description already flags this; a `ServerFeatureFlag` could gate it).
  - `FormationConfidential` picker re-injection via `selected_uid` relies on the query-service applying a distinct-enough FGA relation for Confidential projects. I can't verify that from this repo — worth confirming with the query-service/FGA owner that "not visible until invited" is actually enforced there, not just in the UI copy.

### Reconciliation with Copilot
Agreed and carried forward: settings-fetch gating, the E2E fixture type gap, the "Admin only" label. **Not** carried forward (already resolved or documented scope): the "staff opens the wrong project" stale-link hazard (closed by uid-tagging — only the cosmetic badge flickers, which you documented as an accepted pre-existing trade-off), the removed nav item (`routerLinkActive` can't disambiguate one route — needs a distinct destination first, an issue-text change), the Confidential listing exclusion (a coherent product read), and the auditor slug-path threading + FGA doc comment (both fixed in later commits).

### Decision
✅ **Approved with minor comments.** Please glance at the two minor items before merge.

Issue #1995, opened 2026-08-31
Automatic project create on proposal submit via the v2 project service, behind ServerFeatureFlag.FormationIntake


manishdixitlfx:
> **Epic 2 (#1992).** Split from #1962 on 31 Aug. In Epic 1, Propose a project takes the fallback path only (Proposed row in the queue + prefilled admin-tool link, staff create the record). This issue replaces that with an automatic Draft project create. Reworked 31 Aug per Jordan's comment on #1992: the BFF calls the **v2 project service**, never v1 directly — v2 calls v1 project create as needed. Service side: Prabodh.

## Summary
On proposal submit, the BFF calls the v2 project service create so a Draft project record exists end to end (v2 → v1 → Salesforce/PCC) without staff re-typing the intake. No BFF service credential and no direct v1 call.

## Requested behavior
- On submit, the BFF calls v2 project create with the intake fields (Draft status); the formation links to the new project UID.
- Gated by a server-side env-var kill switch `ServerFeatureFlag.FormationIntake` in `server-feature-flag.helper.ts`, default off, following the MarketingOpsFga pattern. While off, submit takes the Epic 1 fallback path.
- If the v2 create fails, fall back to the Epic 1 path; the proposer sees no difference.
- Parent optional: when the proposer chose "Not sure", create under the Linux Foundation root; staff move it on Accept.

## Acceptance criteria
- [ ] Depends on linuxfoundation/lfx-self-serve#2003: Jordan's dev test (linuxfoundation/lfx-self-serve-ops#5, 31 Aug) showed the v2→v1 project write path is missing entirely — sync-helper only syncs committees bidirectionally. Jordan filed #2003 and is building it. This issue starts once #2003 lands
- [ ] Draft project visible in the admin tool after create
- [ ] Fallback path exercised in tests; kill switch flips at runtime without a deploy
- [ ] No duplicate project when a proposal is re-submitted

## Notes
Spec 6.2, 8 and 9 are in the briefing pack shared with the team; ask Manish for the current copy.

Issue #1957, opened 2026-08-30
New service: lfx-v2-formation-service (templates, checklists, items, invites, activity)


manishdixitlfx:
> **Epic 1 scope (updated 2 Sep; epic #1965):** build the checklist, its items, activity, the derived readiness state, and invites. A checklist is created **from the synced PCC stage**: a project entering a Formation sub-stage gets one, 1:1 on `project_uid`, with the seeded template applied.
>
> **Deferred to Epic 2 (#1992):** `request` objects and SLA (#1963), applications and the propose / accept / decline path (#1962), the template editor (#1994).

> **Confidential — decided 4 Sep.** A Confidential project gets a checklist like any other. **This service carries no Confidential switch of any kind.** Visibility falls out of project access: no `lf-staff` tuple is written for it (#1954), so only direct-grant holders see it.

> **⚠️ The 4 Sep architecture review proposes a materially different service — pending sign-off.** No FGA types, no formation-specific grants, no FGA messages published: every guard is a permission check against `project:<uid>` (read → `auditor`; write → `writer`; gating checked on the item's flag in service logic, not a `gate_writer` relation). Index with `access_check_object = project:<uid>` and `access_check_relation = auditor` — not `viewer`, which includes `[user:*]`. API keyed by project. Templates live here in Postgres, selected by priority + a match rule. Reconcile loop is the mechanism, indexer events the accelerator. Items split functional fields from presentational. See the review for detail; two things I'd push back on are CEL (an enum is enough for one template) and the assignment-elevation prompt.

> **Target repo:** new `lfx-v2-formation-service`; chart lands in `linuxfoundation/lfx-v2-helm` — issues are disabled in both; filed here for tracking.

## Summary
New Goa service in the standard v2 shape owning what PCC has no record for: the checklist template, per-project checklists, items, activity, and the derived readiness state. PCC remains the system of record for the project, its stage and its legal fields; this service reads them and never writes them.

## Completing an item is two steps, not one — product direction, 4 Sep
**An assignee marking their item done does not close it.** The item stays on that person's Pending Actions until the **formation team accepts it**. Only then does it leave their list and count toward readiness.

This matters because the assignee is often an external partner: "I uploaded the signed agreement" is a claim, and the formation team is the one who confirms it landed. Closing the item on the claim alone would make the readiness number — and therefore the Active decision — rest on unverified self-attestation.

So the item lifecycle is:

| Actor | Action | Resulting state |
[...]

PR #2020, opened 2026-09-01
feat(formation): propose a project intake form

manishdixitlfx:
> **Closed 3 Sep — superseded design, not rejected work.** See the closing comment for what survives and where it is recorded. Branch `feat/issue-1962` is left in place at `8c25e0c4`.

## Summary

Epic 1 (#1965) fallback path for "Propose a project": a real intake form + a fixture-backed
formation record, since the upstream formation-service (#1957) and the BFF v1 project-service
create adapter (#1995) don't exist yet. Any logged-in LFX user can submit `/propose` (optionally
prefilled with `?parent=`); submission creates a `formation` (state `proposed`,
`project_uid: null`) and redirects to a dedicated confirmation page describing the "Record not
yet created" staff hand-off.

Refs #1962

## What's here

- Shared types (`Formation`, `FormationIntake`, `FormationState`) and constants in
  `@lfx-one/shared`
- `POST /api/formations` / `GET /api/formations/:uid` — fixture-backed (`FormationService`, an
  in-memory `Map` with a 1-hour TTL sweep, 1000-entry cap, and impersonation blocked on the write)
- `/propose` intake form + `/propose/confirmation/:formationUid` — gated by
  `FORMATION_ENABLED_FLAG` (dark-launched, UI-only)
- "Add a project" CTA on the foundation projects page

## Known trade-offs / follow-ups (flagging explicitly, not silently)

- **Field list is not final** — built from the ticket's field list; confirmation of the current
  Google Form wasn't available this session. Fields are config-driven (`FORMATION_REQUIRED_FIELDS`
  etc.) so a follow-up rename/add is a small diff.
- **No server-side kill switch.** `FORMATION_ENABLED_FLAG` only gates the UI — `POST
  /api/formations` has no paired server flag, per the Epic 2 scope boundary (`ServerFeatureFlag.FormationIntake`
  is #1995). **On merge, the write path is live to 100% of authenticated users regardless of the
  UI flag's target list.** Flagging this prominently since it's a real gap, not an oversight — a
  reviewer should decide whether that's acceptable for a fixture-backed store before this merges,
  or whether a minimal server-side flag check should be pulled forward from #1995.
  *(Now an acceptance criterion on #1962.)*
- **PCC admin-tool URL params are a best guess** (`buildFormationAdminToolLink`) — the real
  create-project query-param contract wasn't available; isolated in one function for easy
  correction.
- **Fixture store is in-memory/ephemeral by design** — resets on pod restart, doesn't survive
  across replicas. This is a deliberate placeholder for #1957, not a bug.
- **No e2e coverage** — this PR has thorough component/unit coverage but no Playwright specs
  (including the confirmation page's "couldn't find that proposal" empty state).
- **Deep import in `formation-validation.helper.ts`** — imports `isValidEmail` from
  `@lfx-one/shared/utils/email.utils` directly rather than the `@lfx-one/shared/utils` barrel,
  because the barrel pulls in Angular-forms-dependent files that can crash a server-only vitest
  worker before the JIT compiler warms up. Documented inline; reviewed and kept as a deliberate,
  narrow exception across multiple review passes.
- **Pre-existing bugs found but not fixed** (out of scope for this PR): `InputTextComponent`
  renders a literal `id="undefined"` when no `id` is bound (repo-wide, ~156 other usages
  affected); the registrant-form has a similar accessibility gap to the one fixed here for the
  intake form. *(Now filed as #2162.)*
- **Untested**: the logo upload field's exact form-control name; `participant_granted` on the
  `Formation` response is currently unused by the client.
- **~14 template calls to `form.get(...)`** in `propose.component.html` weren't converted to typed
  accessors — deprioritized, not a defect.
- **Impersonating users aren't UI-gated** from `/propose` — the write itself is blocked
  server-side (`blockDuringImpersonation`), but the form doesn't hide/disable itself for an
  impersonation session.
- **Parent-project search picker keyboard nav** — Enter no longer submits the outer form (fixed a
  real SSR/hydration + submit-hijack bug), but it's now a "dead key" in the search box: no
  listbox/combobox ARIA pattern, arrow-key nav, or active-descendant. Deliberately deferred — a
  full combobox pattern is a feature addition, not a fix for this PR's scope. Documented inline.

## Process notes

- Commit history was rewritten via `git rebase -i` partway through this branch's life to fix ~10
  commit headers over commitlint's real 72-character limit (CLAUDE.md's own documented ">100
  chars" claim is inaccurate — worth a follow-up doc fix). Content/signatures verified unchanged
  before and after.
- Git hooks (`core.hooksPath`) were inactive in this worktree until this session ran `yarn
  prepare` partway through — commits before that point weren't hook-validated locally, though all
  were confirmed after the fact to meet header-length and signing requirements.
- `apps/lfx-one/src/server/server.ts` is touched (new route registration for
  `/api/formations`) — flagged by the protected-files guard; this is routine additive wiring
  consistent with every other feature router registration in the file, but calling it out per
  policy for code-owner awareness.

## Test plan

- [x] `yarn lint && yarn format && yarn test && yarn build` all pass (34 commits, re-verified after rebase onto latest `origin/main`)
- [x] 3 rounds of full-branch reviewer sweeps (general / repo-convention / empirical-KB) run clean, 0 Critical throughout
- [x] `/lfx-self-serve-pr-readiness` — READY WITH CHANGES, both SHOULD_FIX items (branch name, rebase) addressed before push
- [ ] Manual verification of `/propose` → submit → confirmation flow with the flag enabled (not done in this session — no running dev server)
- [ ] Confirm `/propose` redirects when `FORMATION_ENABLED_FLAG` is off


manishdixitlfx:
Closing this .. #1962 moved to Epic 2 in the 2 Sep scope update, and the design changed with it. What a user submits is an application, not a project, so there is no Draft project create and no `proposer` relation. This branch builds the superseded shape, and the server-side kill-switch gap flagged above is moot now since nothing merges.

Branch `feat/issue-1962` stays put. The intake form, validation helpers and the confirmation page are worth pulling from when Epic 2 picks #1962 up. The fixture-backed formation store and the `/propose` write path do not carry over — that becomes the application object.

manishdixitlfx:
Closed because the design this was built against was superseded, not because the work was wrong.

The 2 Sep scope cut moved intake out of Epic 1, and the redesign in #1962 changed what submit produces: an **application** object, not a formation. This branch creates a `formation` with `state: proposed` and `project_uid: null` on submit, which is precisely the model the redesign replaced. Resuming the branch would mean rebuilding the data layer anyway.

What survives, and is now recorded on #1962 so it is found before anyone rebuilds:

- the intake form and its config-driven field list (`FORMATION_REQUIRED_FIELDS`)
- `formation-validation.helper.ts` and the documented reason for its deep import
- the SSR/hydration fix where Enter in the parent-project picker hijacked the outer form's submit
- the intake form's accessibility fix
- `buildFormationAdminToolLink`, isolated because the PCC query-param contract was never confirmed

The server-side flag gap flagged in this PR body — the UI flag gates the UI only, so the write path would have been live to all authenticated users on merge — is now an acceptance criterion on #1962 rather than a review note, since it matters more against a real store than against a per-pod fixture.

The two pre-existing repo bugs found here are filed as #2162 (`InputTextComponent` renders `id="undefined"`, ~156 usages, plus the related registrant-form gap) so they don't disappear with this PR.

Branch `feat/issue-1962` is left in place at `8c25e0c4` for reference.

Issue #2002, opened 2026-08-31
BUG: Sign a CLA flow always shows GitHub account picker, even for GitLab-only or Gerrit-only CLA Groups


WillsonHG:
### Summary
The "Sign a CLA" flow shows a "Select a GitHub account" step for every CLA Group, regardless of which source (GitHub, GitLab, or Gerrit) the group is actually linked to. Reported in Slack by a teammate: "I see that we are asking for a user's github account for any cla being signed, even those projects that are on github or gerrit exclusively... is it possible to detect the source before we ask for this?"

### Steps to reproduce
1. Go to the "Sign a CLA" search flow.
2. Search "sylv" and select "Sylva Project" (linked via GitLab, tagged "Sylva-projects · GitLab").
3. Observe the "Select a GitHub account" modal appears before continuing to sign.
4. Alternatively, search "oran" and select "O-RAN Alliance" or "O-RAN Software Community" (both linked via Gerrit, tagged "ORAN-SC-ASC · Gerrit" / "ORAN-SC-OSS · Gerrit").
5. Observe the same "Select a GitHub account" modal appears.

### Expected
Per STORY D #1249's acceptance criteria (FR-004): the account-picker step should be gated by the CLA Group's actual linked source.
- Gerrit-backed groups should skip the account step entirely — Gerrit signers use the same LF SSO session already held, so there's no linking dependency and nothing to gate.
- GitHub-backed groups should run the GitHub identity step (the picker).
- GitLab-backed groups have sign entry gated separately, by config, not by this GitHub picker.

### Actual
Every CLA Group triggers the "Select a GitHub account" step regardless of its linked source, including GitLab-only and Gerrit-only groups.

### Environment
Reported via Slack screenshots (Sign a CLA search + account-picker modal), source unspecified beyond the flow itself.

### Notes
Filed under STORY: M2 Defects found (#1733). Violates the source-gating acceptance criteria defined in STORY D #1249.

ahmedomosanya:
Fixed by #2021, now on `main`.

The account step is no longer GitHub-only. It is still shown for every CLA group —
a contributor should see which identity they are about to sign under — but what it
offers follows the group's source:

| CLA group is linked to | The step offers |
|---|---|
| Gerrit | your LF username, presented as your Gerrit identity, then straight to the Console's Gerrit route |
| GitHub and Gerrit | both, in one dialog, each card naming its platform; the card you confirm decides where you go |
[...]

PR #2021, opened 2026-09-01
fix(cla): offer the right identity for the CLA group's source

ahmedomosanya:
## What

Sign a CLA asked every contributor for a GitHub account, whatever the selected
CLA group was actually linked to. Since the account step gained a blocking empty
state, a contributor with no linked GitHub account cannot sign a Gerrit CLA at
all — one that never needed GitHub.

The step is still shown for every source. Only what it offers varies:

| CLA group is linked to | The step offers | Continue goes to |
|---|---|---|
| GitHub, or no linked organization | linked GitHub accounts (unchanged) | prepare-sign, unchanged |
| Gerrit | the contributor's LF username, as their Gerrit identity | the Console's Gerrit route |
| GitHub and Gerrit | both, in one dialog, each card naming its platform | whichever card was confirmed |
| GitLab only | nothing — a blocking state with the reason | nowhere |

## Notes for review

- **The step is never skipped.** An earlier revision of this change skipped it
  for Gerrit groups, and that was rejected during design review — rightly: a
  contributor signing under their LF identity should see that is what they are
  doing. The Gerrit card is not preselected either, even when it is the only one.
- **Routing reads a source being present, never absent.** A CLA group with no
  linked organization keeps the GitHub path. An empty list means nothing is
  linked or nothing resolved — not "not on GitHub" — and those groups are
  signable today, so a rule shaped "no GitHub organization ⇒ not GitHub" would
  misroute them. The empty-list return is the first statement in `claSignRoute`
  so no later branch can reach a conclusion from missing data.
- **That leaves a known gap.** A group with no linked organization that is
  *actually* Gerrit-backed still shows its contributors the GitHub step. This
  change cannot fix those, because there is nothing in the data to distinguish
  them from a GitHub group that simply has no organization linked yet. Only
  groups that positively declare a non-GitHub source are routed.
- **The Gerrit identity is display-only.** The GitHub account list is
  server-supplied, and what the contributor picks is re-validated on submit —
  `prepareSign` refuses an id that is not linked to the session, and the
  upstream service verifies ownership again — so that list is a UX and
  data-integrity constraint rather than the security boundary. The LF username
  is client-derived, and it is safe because it is never submitted, never sent
  upstream, and never recorded; the Console re-derives the signer from the LF
  SSO session. The distinction is documented where the card is built.
- **No `prepare-sign` on the Gerrit route.** It would open a signing session
  nothing reads, and would need a GitHub account this path does not have. The
  mis-resolution guards stay on the GitHub path, neither weakened there nor
  stretched to a path with no account to mis-resolve.
- **GitLab is an identity gap, not a config flag.** GitLab signing authenticates
  through GitLab OAuth at signing time and keys the user record on the GitLab
  numeric id. Self Serve can neither obtain nor verify a GitLab identity, so the
  block is structural — there is no disabled control, because there is no
  condition the contributor could satisfy. The group stays listed in search.
- **The empty-account block is now gated on having no identity at all.** A mixed
  group whose contributor has zero linked GitHub accounts is offered Gerrit
  rather than told to go link a GitHub account. That is the sharpest regression
  risk here and is covered at both the dialog and orchestration levels. It also
  means only the GitHub-only variant can reach that block, and it already names
  GitHub: a Gerrit identity always resolves, and a variant that cannot resolve
  one is rewritten to the GitHub variant before the step opens, so a step
  offering Gerrit is never empty. There is deliberately no Gerrit or mixed
  wording for it, because no path reaches it.
- **Copy** comes from the design and from review, not from us. The platform
  suffixes exist because a GitHub handle and an LF username are routinely
  different strings for the same person. The GitLab block's header, `GitLab CLA
  signing`, comes from review rather than from a mockup: it was asked to name
  its source, as every other dialog in this flow already does.
- `GithubAccountSelectComponent` is renamed to `SignIdentitySelectComponent`
  because it now decides between platforms. Only the template is detected as a
  rename — the `.ts` and `.spec.ts` change enough that GitHub shows them as a
  deleted file plus a new one.

## Testing

- `yarn lint` — clean
- `yarn workspace @lfx-one/shared test` — 1,280 passing
- `yarn workspace lfx-one-ui test` — 3,688 passing
- `yarn build:production` — passing

Verified against DEV. The Gerrit hand-off URL is composed by this change and had
never been exercised, so it was driven end to end: the Console accepts it,
resolves the contributor from the LF SSO session and renders the ICLA page — no
404 and no redirect loop. The mixed case ran against a group carrying all three
sources, with GitLab correctly not offered. Also confirmed on DEV: no card is
preselected, the confirmed card alone decides the route, a mixed group whose
contributor has zero linked GitHub accounts is offered Gerrit rather than
blocked, and the GitLab blocking state renders.

Two gaps worth knowing. The DEV account has no linked GitHub identity, so seeing
a GitHub and a Gerrit card side by side — and with them the platform suffixes
and the blank-handle fallback — required stubbing the accounts response; every
other DEV result above is unstubbed. And DEV has no Gerrit-only CLA group, so
that variant is covered by unit tests only.

Refs #2002


apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
```diff
@@ -0,0 +1,150 @@
...
+ * so the step is never skipped on the strength of the group's source (#2002). Only which
+ * identities are on offer varies, and the variant says which.
+ *
+ * ## Where each identity comes from, and why they are not equally trusted
+ *
+ * The **GitHub accounts** come from the server, which reads them from the identity provider for
+ * this session, and the CLA service records the submitted account without re-deriving ownership.
+ * So that list is not merely display data: an account that should not be in it would be
+ * recorded. It must be presented as given and never taken from the URL, from user input, or
+ * from anywhere but `config.data`.
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `8709f232` — correct, and the paragraph now says so. `prepareSign` re-reads the session's linked accounts and refuses an id that is not among them (403 `CLA_ACCOUNT_NOT_LINKED`), and the upstream service derives the attested set from the caller's own token and refuses anything outside it again. So this list is now described as a UX and data-integrity constraint rather than the security boundary. The instruction to take accounts only from `config.data` stays, as defence in depth. The same stale premise was in the PR description and is corrected there too.

apps/lfx-one/src/app/modules/profile/clas/gitlab-unsupported.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  templateUrl: './gitlab-unsupported.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class GitlabUnsupportedComponent {
+  private readonly ref = inject(DynamicDialogRef);
+
+  protected readonly message = GITLAB_UNSUPPORTED_MESSAGE;
+
+  protected onClose(): void {
+    this.ref.close(null);
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `8709f232` — added `gitlab-unsupported.component.spec.ts` covering that the unsupported message renders and names GitLab, that the header the dialog is opened under is the platform-naming constant, that Close is offered with nothing that continues to signing, and that Close closes with `null`.

MRashad26 (dismissed):
Reviewed the full diff (15 files, +1319/-344) by reading it directly myself given the size and auth/identity-sensitive surface, then verified against real tooling in a disposable worktree: `yarn lint:check` clean (only a pre-existing unrelated warning), and the full relevant test sweep (75 spec files, 1425 tests, including `sign-identity-select.component.spec.ts`, `gitlab-unsupported.component.spec.ts`, `profile-clas.component.spec.ts`, and `cla-view.utils.spec.ts`) passes 100%, matching the PR's own testing claims.

**What this does (GH-2002):** Reworks CLA sign hand-off so the identity picker offers the right identity type for a CLA group's linked org source(s), instead of unconditionally requiring GitHub. `claSignRoute(organizations)` (new pure function in `cla-view.utils.ts`) classifies a group as `github`, `gerrit`, `github-or-gerrit`, or `gitlab-unsupported` from source *presence* only (empty list still defaults to `'github'`, preserving today's behavior for groups with no linked orgs). `gerritSignUrl(...)` composes the Console's Gerrit route and deliberately avoids a regex-based trailing-slash strip (scanning `while` loop instead) to sidestep a CodeQL-flagged ReDoS pattern. `GithubAccountSelectComponent` is replaced by `SignIdentitySelectComponent` (handles the github/gerrit/mixed cases) plus a new `GitlabUnsupportedComponent` (blocking state, Close-only, no disabled Continue button — framed correctly as a structural gap, not a disabled feature).

**Security-relevant claims verified against the actual code, not just the PR description:**
- The GitHub account list is still server-supplied; `prepareSign` is only invoked on the GitHub/mixed-GitHub path, and the Gerrit path never calls it — confirmed no client-suppliable account id reaches the sign call unchecked.
- The Gerrit "identity" shown to the user (LF username) is rendered as a display-only label and is never included in `gerritSignUrl`'s query params or anywhere else submitted to the Console — the actual signer is re-derived server-side from the SSO session. The URL only carries the CLA group id and an encoded same-origin `returnUrl`, so no PII leaks through the redirect.
- Source-classification order is `github && gerrit` checked before either single-source branch, so a mixed group can't silently fall through to the more restrictive GitHub-only path — this is also the scenario both new test suites call out explicitly as the sharpest regression risk (empty-GitHub-accounts + Gerrit-linked no longer trips the old "no accounts linked" block).
- Unrecognized/unknown source strings fall back to `'github'` rather than failing open into an unhandled state; documented as unreachable in practice since the search mapper drops unknown sources upstream.

**Non-blocking observations:**
- The PR carries the **`WIP` label** alongside `bug`/`ai-assisted`/`EasyCLA`. Code and tests are in good shape and I found nothing that would block a merge, but flagging this since a WIP label sometimes signals the author isn't done iterating — worth confirming with them before merging even though the diff itself looks complete.
- Diff size (1663 changed lines) is well over this repo's usual PR-sizing guidance, though it's justified here as one cohesive rename+behavior change with matching test coverage rather than bundled unrelated work.
- Two coverage gaps are self-acknowledged in the PR body (mixed-card display required stubbing since DEV has no linked GitHub identity; Gerrit-only variant is unit-test-only, no DEV CLA group to verify against) — reasonable given DEV's data limitations, not something this PR could fix itself.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -247,29 +279,50 @@ export class ProfileClasComponent {
...
+   * already chosen. Either way nothing reaches the Console without a confirmed identity.
+   *
+   * A Gerrit identity is offered whenever the group is linked to Gerrit, and it always resolves,
+   * so a variant carrying one cannot reach that empty state. That matters most on a mixed group
+   * whose contributor has no linked GitHub account: they are offered Gerrit rather than blocked
+   * on the GitHub account they do not need.
    */
-  private chooseAccountThenSign(option: ClaGroupOption, accounts: GithubAccountOption[]): void {
+  private chooseIdentityThenSign(option: ClaGroupOption, variant: SignIdentityVariant, accounts: GithubAccountOption[]): void {
+    const gerritUsername = variant === 'github' ? undefined : (this.userService.viewerUsername() ?? undefined);
```
audigregorie: **Minor — parent and child disagree on what a resolvable Gerrit identity is.**

Here a whitespace-only `viewerUsername` counts as resolvable, but `initGerritLabel()` in the dialog trims and drops the card. A `gerrit` variant would then render the GitHub-worded empty state ("No GitHub accounts are connected…") under a "Select a Gerrit account" header — the wrong-platform copy this change exists to prevent — and a `github-or-gerrit` variant would keep its both-platforms copy while showing only GitHub cards.

Practically unreachable today (LF usernames can't be whitespace), but the documented invariant "a variant offering Gerrit is never empty" currently rests on that assumption. Trimming here makes it structural — the guard, the variant downgrade, and the card builder all share one definition of resolvable:

```suggestion
    const gerritUsername = variant === 'github' ? undefined : (this.userService.viewerUsername()?.trim() || undefined);
```

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -294,6 +352,57 @@ export class ProfileClasComponent {
...
+    const returnUrl = `${this.document.location.origin}${MY_CLAS_PATH}`;
+    const url = gerritSignUrl(environment.urls.contributorConsole, option.claGroupId, returnUrl);
+
+    if (!url) {
+      this.starting.set(false);
+      this.reportGerritDestinationUnresolved();
+      return;
+    }
+
+    this.document.location.href = url;
```
audigregorie, line 404: **Minor — `starting` stays set on the successful Gerrit navigation.**

`prepareThenHandOff` resets `starting` before assigning `location.href`; this path navigates with it still `true`. The return flow is a fresh navigation so it's invisible there, but if the browser restores this page from bfcache (Back from the Console before signing), Sign CLA comes back stuck in its loading state and clicks are silently ignored until a reload. Resetting just before navigating matches the GitHub hand-off and keeps the double-click guard intact for everything up to the navigation itself:

```suggestion
    this.starting.set(false);
    this.document.location.href = url;
```

audigregorie (commented):
## Code Review Summary

Solid, carefully-scoped change: the routing rule (presence is evidence, absence is not) is the right shape, the trust boundaries between the server-supplied GitHub list and the display-only Gerrit identity are kept straight and documented, and the regression matrix is tested at both the dialog and orchestration levels. Verified locally: prettier + eslint clean on changed files, `cla-view.utils.spec.ts` 41/41, app CLAS suites 1,425/1,425 passing. Two minor line-level findings (one defensive trim, one state-reset consistency fix) and one stale-docs note below — nothing blocking.

### Minor — outside the diff

**User docs are now stale.** `docs/user/account/my-clas/index.md:115` describes step 5 as GitHub-only ("Choose which of your linked GitHub accounts to sign under… If you have no GitHub account linked, the step says so and offers a link to Identities; you cannot continue past it until you link one"), and `docs/user/account/faq/index.md:77` mirrors it. After this PR, a Gerrit-linked group offers the contributor's LF username, a mixed group can continue with no GitHub account at all, and a GitLab-only group gets a blocking "not supported" state — so "you cannot continue past it until you link one" is no longer true for every group. Worth a follow-up edit describing the routing by the group's linked source.

### What's done well

- The empty-list return as the first statement of `claSignRoute`, so no later branch can reach a conclusion from missing data — with the failure mode it prevents written down next to it.
- The sharpest regression (mixed group, zero linked GitHub accounts → offered Gerrit rather than blocked) is covered at both the dialog and the orchestration level.
- `gerritSignUrl` fails closed — relative or unset base yields `null` and a report instead of a navigation that cannot sign anything — and the return address is composed from our own origin rather than a trusted-host check.
- `MY_CLAS_PATH` deduplicated into shared, so the BFF and the browser cannot disagree on the return address.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
ahmedomosanya: Addressed in `6578dd1d0` — trimmed at the source, so the guard, the variant downgrade and the dialog's card builder now share one definition of resolvable.

Covered by parameterising both unresolvable-identity tests over an absent and a whitespace-only username. The whitespace cases fail without it — the Gerrit-only group opens the dialog (`expected [ …(2) ] to not include [Function _SignIdentitySelectComponent]`) and the mixed group stays on `github-or-gerrit` instead of downgrading.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
ahmedomosanya: Addressed in `6578dd1d0` — `starting` is released immediately before the navigation, matching the GitHub hand-off, so the double-click guard still covers everything up to that point.

New test `releases Sign CLA before leaving for the Console` covers it, and fails with `expected true to be false` without the reset.

audigregorie (approved):

PR #2026, opened 2026-09-01
fix(votes): entity-scoped slug resolution for vote edit route (#1568)

### LINKED ISSUE #1568: Vote edit links can open in the wrong project context
## Problem

When someone opens a link to edit a vote — copied from the votes table or shared by a teammate — the page can load under whichever project the viewer happened to visit last instead of the project that actually owns the vote. Vote managers editing a vote that belongs to a different project (or directly to a foundation) see the wrong project in the sidebar and project picker, can be intermittently denied access even though they manage that vote, and the URL flips between the /project/ and /foundation/ paths depending on the viewer's last-visited area rather than the vote's ownership.

## Outcome

Vote edit links always open under the project or foundation that owns the vote: the URL path reflects foundation ownership when there is one, the page loads with the owning project selected, and access checks run against the owning project. Links shared with teammates land in the right place regardless of what the recipient was looking at before.

## Success criteria

- Opening a vote's edit link while a different project is selected still loads the page in the owning project's context (sidebar, project picker, committee selection).
- A foundation-owned vote's edit links use the /foundation/ path; other votes use /project/ — regardless of the viewer's current lens.
- A legitimate vote manager (including committee writers) is not sent to an access-denied page when opening an edit link that carries no project context.
- Copying the URL mid-way through the edit wizard and opening it fresh keeps the correct project.

audigregorie:
## Summary

Vote edit links and the vote edit page derived their project context from the viewer's active lens (typically cookie-restored from a previous session) rather than from the vote itself. A context-less edit link for a vote owned by a different project could therefore load — and be authorized — against the wrong project. This PR scopes edit links, context resolution, and write-access authorization to the vote's own project, hardens vote-uid URL encoding on both the client and the BFF, and fixes `lfx-button` so routerLink buttons render as real anchors.

Resolves #1568

### Canonical vote edit links

| Before | After |
|---|---|
| Edit links in the votes table pointed at a generic edit URL with a single table-wide project parameter, so a dashboard mixing votes from several projects could send the editor to the wrong project context. | Each row's edit link is built from that vote's own project — foundation votes link to the foundation URL, project votes to the project URL — and each link carries its own project (and committee) parameters. |

### Edit page project context

| Before | After |
|---|---|
| Opening a vote edit link without a project parameter kept whatever project happened to be active from a previous session, so the page could load under the wrong project. | The edit page switches to the vote's own project once the vote loads — and can still recover the project by its id when server-side enrichment is missing — without writing extra parameters into the URL. |

### Edit authorization

| Before | After |
|---|---|
| Edit permission was checked against the currently active project, so a user could be wrongly blocked from (or wrongly admitted to) editing a vote belonging to a different project. | Permission is checked against the project that actually owns the vote; if the vote can't be read due to a server error, the check fails safe instead of falling back to a stale project. |

### Link buttons

| Before | After |
|---|---|
| Buttons configured as links rendered as plain buttons with no real link address — middle-click / open-in-new-tab didn't work, and a click handler could swallow the navigation. | Link buttons render as real links with a proper address, so browser link behaviors (new tab, copy link, screen readers) work as expected. |

## Technical changes

`apps/lfx-one/src/app/modules/votes/components/votes-table/votes-table.component.ts, votes-table.component.html` — Canonical per-row edit links
- Builds each row's `editCommands` via `getEntityCommands('votes', uid, is_foundation, 'edit')`, falling back to the flat `/votes/:uid/edit` path (lensRedirectGuard territory) when the row's tier is unenriched
- Derives per-row `editQueryParams` (`project` from the row's own slug, `committee_uid` from its committee), replacing the removed table-level `editQueryParams` input — dashboard rows span projects, so no table-level record can express them

`apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts` — Edit-page context resolution
- Adds a `voteEntityContext` computed mapping the loaded vote to the `EntityWithProject` shape, and syncs it through `syncEntityProjectContext` with `preferEntityKind: true` so context-less edit links land in the vote's own project tier (not the cookie-restored one)
- `initWriteAccess` mirrors writerGuard's votes standard — project writer on the vote's OWN project (never the boot context) or committee writer via `?committee_uid=`, with pending legs provisionally true — and feeds `evictOnWriteAccessLoss`
- `initVoteContextFallback` + `resolveContextFromFreshVote` recover context by resolving the project by uid (re-applied on `NavigationEnd`), with a one-time fresh vote re-fetch as last resort when the relation-gated lookup returns null
- Edit-step navigation preserves existing query params via `queryParamsHandling: 'merge'`

`apps/lfx-one/src/app/modules/votes/votes.routes.ts, apps/lfx-one/src/app/shared/guards/writer.guard.ts` — Entity-scoped authorization
- The votes edit route is flagged `entityScopedSlug: true` so writerGuard resolves the authorization slug from the vote itself (a route-data flag, not a path check, so a route rename can't silently revert to stale-context authorization)
- The guard's `entityProbes` registry gains a `votes` probe via the new tap-free `VoteService.fetchVote`; a 404 read falls back to the active context, any other failure fails closed with a redirect

`apps/lfx-one/src/app/shared/services/vote.service.ts` — Client vote service
- Adds the tap-free `fetchVote` used by the guard probe and the context fallback (`getVote`'s signal write would leak state across pages)
- Wraps all seven `/api/votes/:uid` interpolations in `encodeURIComponent` — the guard probe's id comes from the URL and is user-controllable

`apps/lfx-one/src/server/services/vote.service.ts, apps/lfx-one/src/server/controllers/vote.controller.ts` — BFF vote service
- `getVoteById` gains an opt-in `includeProject` option enriching the payload with the vote's project fields (slug/name/is_foundation) via the entity-project-enrichment helper; the controller passes it for the authenticated detail endpoint so clients can reconcile project context from the vote itself
- Wraps the eight upstream `/votes/:uid` path interpolations in `encodeURIComponent` (`req.params.uid` arrives percent-decoded and only emptiness-validated, so a crafted id could otherwise reshape the upstream path); the `vote_uid:` query-service filter is deliberately left unencoded — it's filter DSL, not a URL path
- Tightens `getVotes`/`getVotesCount` query params to `Record<string, unknown>`

`apps/lfx-one/src/app/shared/components/button/button.component.html` — Anchor link buttons
- Adds an `@else if (routerLink())` branch rendering a real `<a pButton>` so the link carries an href and RouterLink's click listener fires before any consumer `stopPropagation` can starve it on the p-button host
- Drops the now-dead `routerLink`/`queryParams`/`target` bindings from the `p-button` branch

`apps/lfx-one/src/app/modules/committees/components/committee-votes/committee-votes.component.ts, committee-votes.component.html` — Committee votes tab
- Drops the obsolete `editVoteQueryParams` passthrough — the table now derives per-row params itself

`packages/shared/src/interfaces/poll.interface.ts` — Shared types
- `VoteTableRow` gains `editCommands` and `editQueryParams` with rationale docs (per-row derivation, flat-path fallback)
- `Vote` project-field docs updated to spell out list-row vs BFF-enriched detail provenance, and `parent_project_uid`'s deliberate exclusion from detail enrichment

`apps/lfx-one/e2e/vote-edit-url.spec.ts, apps/lfx-one/e2e/project-context-deep-link.spec.ts` — E2E coverage
- New spec pinning canonical edit-link URLs (foundation-owned, project-owned, and unenriched rows) plus click-through landing in the vote's own project context
- The deep-link spec gains a vote describe block: context-less edit links switch the selector/sidebar to the vote's foundation without injecting `?project=`, writerGuard authorizes against the vote's project instead of the stale cookie context, and the enrichment-absent payload falls back to resolving the project by uid

`apps/lfx-one/src/app/shared/guards/writer.guard.spec.ts, apps/lfx-one/src/server/services/vote.service.spec.ts, apps/lfx-one/src/app/modules/dashboards/org/org-project-detail/org-project-detail.component.spec.ts` — Unit coverage
- writerGuard spec pins the votes probe contract: authorizes against the vote's own project on success, falls back to context only on 404, fails closed on a 500 without probing the stale project
- New server spec pins encoded upstream vote paths (bare, `/enable`, `/results`): canonical UUIDs pass through, a raw `/` encodes to `%2F`, and a pre-encoded `%2F` double-encodes so a smuggled separator can't materialize upstream
- org-project-detail spec's Router stub extended (`createUrlTree`/`serializeUrl`/`events`) for the anchor-based `lfx-button` rendering


apps/lfx-one/src/app/modules/votes/components/votes-table/votes-table.component.ts
```diff
@@ -276,9 +275,21 @@ export class VotesTableComponent {
...
 
+    // Canonical edit URL derives from the VOTE's own project tier (is_foundation), not the viewer's
+    // active lens; falls back to the flat path (lensRedirectGuard) when the tier is unenriched.
+    const editCommands = getEntityCommands('votes', vote.uid, vote.is_foundation, 'edit') ?? ['/votes', vote.uid, 'edit'];
+
+    // Per-row query params: rows can span projects (dashboard), so no table-level record can
+    // express them — derive `project`/`committee_uid` from each row itself.
+    const editQueryParams: Record<string, string> = {};
+    if (vote.project_slug) editQueryParams['project'] = vote.project_slug;
+    if (vote.committee_uid) editQueryParams['committee_uid'] = vote.committee_uid;
```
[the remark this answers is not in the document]
audigregorie: Fixed in d204836e1 — the votes table now accepts an editQueryParamsFallback record merged under each row's own derived params, and the committee tab passes its committee's create params through it, so committee_uid rides every committee-tab edit link again even when the indexed row omits the field.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -71,13 +73,11 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
...
   // Entity probes keyed by writeFeature — a new entity adds one registry line + the route's
-  // entityScopedSlug flag. Probes must be tap-free: fetchCommittee, not getCommittee (which sets
-  // the shared committee signal), so a guard probe can't leak stale state into other pages. Both
-  // probes ride a short-TTL shared cache (getMeetingDetail / getCommitteeDetail via fetchCommittee),
-  // so the manage component's immediate refetch on the same navigation doesn't duplicate the request.
+  // entityScopedSlug flag. Probes must be tap-free so a guard probe can't leak stale state into other pages.
   const entityProbes: Record<string, (id: string) => Observable<Pick<EntityWithProject, 'project_slug' | 'project_uid'> | null>> = {
     meetings: (id) => meetingService.getMeetingDetail(id),
     committees: (id) => committeeService.fetchCommittee(id),
+    votes: (id) => voteService.fetchVote(id),
```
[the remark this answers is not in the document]
audigregorie: Fixed in d204836e1 — on entity-scoped edit routes the guard now authorizes the committee leg against the probed entity's own committee_uid; the raw URL param is reserved for create routes (and as an unenriched-entity fallback). New spec coverage pins both the legit committee-writer admission and the spoofed-param denial.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -573,6 +605,134 @@ export class VoteManageComponent {
...
+      });
+  }
+
+  // Last resort for initVoteContextFallback: the ungated detail enrichment can supply the
+  // project the relation-gated lookup withheld. Emits null when it can't, leaving context untouched.
+  private resolveContextFromFreshVote(entity: EntityWithProject): Observable<{ context: ProjectContext; isFoundation: boolean } | null> {
+    if (this.contextFallbackRetried.has(entity.uid)) {
+      return of(null);
+    }
+    this.contextFallbackRetried.add(entity.uid);
```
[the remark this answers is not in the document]
audigregorie: Fixed in d204836e1 — the resolved fallback context is cached per vote uid and re-emitted on subsequent NavigationEnd re-applies, so MainLayout's route-lens re-assert can no longer strand the fresh-fetch resolution (getProject's cached null is bypassed on later navigations).

apps/lfx-one/src/app/shared/components/button/button.component.html
```diff
@@ -28,6 +28,36 @@
...
+} @else if (routerLink()) {
+  <!-- Real anchor so the link carries a href (a11y, middle-click) and RouterLink's click listener fires -->
+  <!-- before any consumer stopPropagation on (onClick) can starve it on the p-button host. -->
+  <a
+    pButton
+    [type]="type()"
+    [routerLink]="routerLink()"
+    [queryParams]="queryParams()"
+    [target]="target()"
+    [rel]="rel()"
```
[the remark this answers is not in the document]
audigregorie: Fixed in d204836e1 — the routerLink anchor now nulls its [routerLink] while disabled or loading (RouterLink no-ops on a null urlTree and no href is rendered, matching the old p-button disabled behavior) and exposes aria-disabled to assistive technology.

apps/lfx-one/src/server/services/vote.service.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  let service: VoteService;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    proxyRequest.mockResolvedValue(voteFixture);
+    proxyRequestWithResponse.mockResolvedValue({ data: voteFixture, headers: {} });
+    service = new VoteService();
+  });
+
+  describe('getVoteById', () => {
```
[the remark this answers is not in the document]
audigregorie: Fixed in d204836e1 — the suite now exercises getVoteById with { includeProject: true }: asserts the mapped project fields land on the returned vote, that a null project lookup leaves the payload untouched, and that enrichment is skipped by default.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -95,17 +103,24 @@ export class VoteManageComponent {
...
+  private readonly writeAccess: Signal<boolean> = this.initWriteAccess();
 
   public constructor() {
     this.initCommitteeContext();
-    evictOnWriteAccessLoss();
+    evictOnWriteAccessLoss(this.writeAccess);
+
+    // Context-less edit links land in the VOTE's project (not the cookie-restored one);
+    // preferEntityKind picks the context slot from the vote's own is_foundation (see the util's doc).
+    syncEntityProjectContext(this.voteEntityContext, this.projectContextService, this.router, this.destroyRef, { preferEntityKind: true });
```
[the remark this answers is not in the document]
audigregorie: Fixed in d204836e1 — syncEntityProjectContext now runs with canonicalizeRoute: true and the UID-fallback branch calls canonicalizeTierPrefix after applying the context, mirroring the mailing-list flow; a wrong-tier /project|/foundation URL is rewritten (replaceUrl) once the vote's own kind is known.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -573,6 +613,144 @@ export class VoteManageComponent {
...
+    const committeeAccess = toSignal(
+      this.committeeUidFromUrl
+        ? this.committeeService.fetchCommittee(this.committeeUidFromUrl).pipe(
+            map((committee) => committee?.writer === true),
+            catchError(() => of(false))
+          )
+        : of(false)
+      // No initialValue: undefined doubles as the leg's pending state. Without a committee_uid
+      // param the synchronous of(false) resolves the leg immediately, so it never counts as pending.
+    );
```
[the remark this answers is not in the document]
audigregorie: Fixed in 8ebc108be — the committee leg now derives the edit-mode uid from the loaded vote's own `committee_uid` (URL param as fallback), mirroring writerGuard's entity-scoped resolution so the eviction predicate no longer redirects a committee writer the guard admitted.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -573,6 +613,156 @@ export class VoteManageComponent {
...
+        })
+      )
+      // No initialValue: undefined doubles as the leg's pending state (see the doc above).
+    );
+    // Committee leg: edit mode derives the uid from the loaded vote's own committee (URL only as
+    // fallback), mirroring writerGuard's entity-scoped resolution — otherwise the guard admits a
+    // committee writer on a context-less edit link while this predicate evicts them. undefined stays
+    // pending; null resolves the leg immediately so it never counts as pending.
+    const committeeUid$: Observable<string | null | undefined> = editVoteId
+      ? toObservable(this.vote).pipe(map((vote) => (vote ? (vote.committee_uid ?? null) : undefined)))
```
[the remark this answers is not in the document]
audigregorie: Fixed in 96fde12aa — the edit-mode committee leg now falls back to the `?committee_uid=` query param when the loaded vote omits `committee_uid` (`vote.committee_uid ?? committeeUidFromUrl`), mirroring writerGuard's `entityCommitteeUid ?? committeeUid` resolution.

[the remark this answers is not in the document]
audigregorie: Fixed in 96fde12aa — the committee leg now reads `vote.committee_uid ?? committeeUidFromUrl`, so a committee writer admitted via `?committee_uid=` is no longer evicted when the loaded vote omits `committee_uid`.

apps/lfx-one/src/app/shared/services/vote.service.ts
```diff
@@ -116,27 +116,32 @@ export class VoteService {
   }
 
   public getVote(voteUid: string): Observable<Vote> {
-    return this.http.get<Vote>(`/api/votes/${voteUid}`).pipe(tap((vote) => this.vote.set(vote)));
+    return this.http.get<Vote>(`/api/votes/${encodeURIComponent(voteUid)}`).pipe(tap((vote) => this.vote.set(vote)));
+  }
+
+  /** Tap-free vote fetch for writerGuard's entity probe and vote-manage's context fallback (getVote's signal write would leak state). */
+  public fetchVote(voteUid: string): Observable<Vote> {
+    return this.http.get<Vote>(`/api/votes/${encodeURIComponent(voteUid)}`).pipe(take(1));
```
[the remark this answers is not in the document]
audigregorie: Fixed in 877f242ae — VoteService now has a short-TTL voteDetailCache + getVoteDetail (mirroring getMeetingDetail/getCommitteeDetail); getVote and fetchVote share it, writes (update/delete/enable) evict, and vote-manage's fresh context fallback passes skipCache so it still polls the latest enrichment.

apps/lfx-one/e2e/project-context-deep-link.spec.ts
```diff
@@ -542,6 +576,110 @@ test.describe('Meeting edit deep-link resolves the meeting’s project context (
...
+    await gotoSpa(page, `/project/votes/${MOCK_VOTE_UID}/edit`, {
+      uid: OTHER_PROJECT_UID,
+      slug: OTHER_PROJECT_SLUG,
+      name: 'Other Project',
+      foundation: false,
+    });
+    await expect(page.getByTestId('vote-manage-title')).toBeVisible({ timeout: PAGE_LOAD_TIMEOUT });
+
+    // Still on the edit page (no access-denied redirect), and the context follows the vote.
+    expect(page.url()).toContain(`/project/votes/${MOCK_VOTE_UID}/edit`);
```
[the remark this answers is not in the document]
audigregorie: Fixed in 1bd06862f — the assertion now awaits the canonical `/foundation/votes/:uid/edit` URL via auto-retrying `toHaveURL` instead of racing the tier rewrite with a synchronous `page.url()` check, matching the sibling mailing-list specs.

apps/lfx-one/src/app/shared/components/button/button.component.html
```diff
@@ -28,6 +28,39 @@
...
+} @else if (routerLink()) {
+  <!-- Real anchor so the link carries a href (a11y, middle-click) and RouterLink's click listener fires -->
+  <!-- before any consumer stopPropagation on (onClick) can starve it on the p-button host. -->
+  <!-- Null link while disabled/loading: RouterLink no-ops on a null urlTree and renders no href, -->
+  <!-- mirroring the p-button branch's disabled behavior; aria-disabled keeps AT informed. -->
+  <a
+    pButton
+    [type]="type()"
+    [routerLink]="disabled() || loading() ? null : routerLink()"
+    [attr.aria-disabled]="disabled() || loading() ? true : null"
```
ahmedomosanya, line 42: Nice-to-have, not a blocker.

Nulling `[routerLink]` and setting `aria-disabled` stops navigation, but the `<a pButton>` never binds `[disabled]` (the `p-button` branch does). PrimeNG's disabled styling therefore doesn't apply, and the anchor stays in the tab order.

Could we also disable the host and drop it out of the tab sequence while disabled/loading?

```suggestion
    [routerLink]="disabled() || loading() ? null : routerLink()"
    [disabled]="disabled() || loading()"
    [attr.aria-disabled]="disabled() || loading() ? true : null"
    [attr.tabindex]="disabled() || loading() ? -1 : tabindex()"
```

ahmedomosanya (dismissed):
## LGTM

Right fix for #1568 — entity-scoped edit links, context sync, and `writerGuard`, matching the meeting / mailing-list pattern. Prior review threads look addressed; CI is green.

One leftover on `lfx-button` (inline). Product question: `canonicalizeTierPrefix` no-ops on a flat `/votes/:id/edit` URL, and `lensRedirectGuard` leaves those routes alone in the Me and Org lenses. #1568 asks for `/foundation/` vs `/project/` regardless of the viewer's lens. The unenriched-row fallback is documented as intentional `lensRedirectGuard` territory — is staying flat in Me/Org accepted, or should load still promote the URL once the vote's kind is known?

dealako:
## Review: entity-scoped vote edit-route resolution

Hi @audigregorie, thanks for a careful, well-documented fix. The core of this PR is sound: edit-link derivation, `syncEntityProjectContext`, `initWriteAccess`, and the `writerGuard` entity probe all bind authorization and context to the vote's *own* project rather than the viewer's cookie-restored lens, the guard fails closed on non-404 read errors, the vote-uid path interpolations are consistently `encodeURIComponent`-wrapped on both the client and BFF, and the flow tracks the sibling meeting/mailing-list patterns closely. A full security + data-privacy pass found nothing: the deliberately-unencoded `vote_uid` filter is query-service DSL (not a URL path), the enrichment exposes only `project_slug`/`name`/`is_foundation` after upstream read-auth, and all new fixtures are synthetic.

Everything below is polish. There are no blocking correctness or security defects. I'm marking this Needs Changes only because the mechanical bar for this review is >2 minor items; each is small and several are one-line doc/scope fixes. The most substantive is the open question about list-row enrichment.

### Issue count

- 🔴 **Blocking:** 0
- 🟡 **Minor:** 4
  - Guard doc block claims the entity's own `committee_uid` wins on all entity-scoped routes, but only the votes probe surfaces one (meetings edit still trusts the URL param) — `writer.guard.ts:129`
  - Committee write-access leg fires an unconditional `GET /api/committees/:uid` on every edit load, even when the project leg already grants access — `vote-manage.component.ts:658`
  - ~150 lines of new reactive fallback/write-access/cache logic ship with no `vote-manage.component.spec.ts` (the mirrored meeting logic is unit-pinned) — `vote-manage.component.ts:692`
  - New e2e redeclares two shared cookie-key constants as local literals (one shadowing the shared export), a silent-seeding failure mode on rename — `vote-edit-url.spec.ts:39`
- ⚪ **Nit:** 4
  - `resolvedContextCache` diverges from the meeting/mailing-list fallback it was copied from — `vote-manage.component.ts:93`
  - `voteEntityContext` is `public`; siblings keep this adapter `private` — `vote-manage.component.ts:110`
  - Guard doc route enumeration omits mailing-lists — `writer.guard.ts:34`
  - Anchor link-button blocks navigation while disabled but doesn't forward `[disabled]`/tabindex (agrees with @ahmedomosanya) — `button.component.html:40`
- ❔ **Question:** 1
  - Does the vote query-service index actually populate `project_slug`/`is_foundation` on list rows? The BFF list path never enriches them and the e2e injects them synthetically, so if the index omits them the canonical per-row tier link silently no-ops on the main dashboard — `votes-table.component.ts:283`

### Prior review rounds (reconciliation)

Bot findings from earlier commits are addressed on the current head:

- **Copilot** double-fetch (share a detail cache) — resolved via `getVoteDetail`/`fetchVote` short-TTL cache; the guard probe and manage page now share one vote read + one project read.
- **Copilot** missing `includeProject: true` server test — resolved; `vote.service.spec.ts` now asserts the enabled path merges mapped project fields, the disabled path, and the null-project path.
- **Copilot** `canonicalizeRoute` not enabled on unenriched flat links — resolved; the flow now calls `canonicalizeRoute` + `canonicalizeTierPrefix`.
- **Copilot** eviction committee leg diverging from the guard (`:658`/`:674`/`:736`) — resolved; `initWriteAccess` now derives the committee uid from the vote (URL only as fallback) and re-applies resolved context via the cache.
- **Copilot** deep-link spec asserting the wrong tier — resolved; the spec awaits the foundation-tier URL.
- **Cursor** edit links dropping `committee_uid` — mitigated; per-row params merge over `editQueryParamsFallback()` (committee-tab fallback) then override from the row.
- **@ahmedomosanya** anchor `[disabled]`/tabindex — still open (nit above); navigation-while-disabled itself is fixed.

### Decision

🔴 **Needs changes before approval** — no blockers; the minor items (guard-doc truthfulness, the per-load committee fetch, the missing unit spec, and the shared-constant duplication) plus a confirmation on list-row enrichment are worth a quick pass before merge.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -109,10 +122,14 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
       const deny = () => deniedUrl;
       const supportsCommitteeWriter = writeFeature != null && COMMITTEE_WRITE_FEATURES.includes(writeFeature);
 
-      // Committee writers can create entities for their committee via ?committee_uid=.
+      // Committee writers can create entities for their committee via ?committee_uid=. On
+      // entity-scoped edit routes the probed entity's own committee wins — a URL param naming an
+      // unrelated committee must not admit its writer, and an absent param must not deny the
+      // entity's real committee writer.
+      const effectiveCommitteeUid = entityCommitteeUid ?? committeeUid;
```
dealako, line 132: [minor] Guard doc overclaims the entity-committee override for all entity-scoped routes

Issue: The doc block (lines 28-32) and this line's comment (125-128) state that on `entityScopedSlug` edit routes the committee uid comes "from the probed entity itself (URL params are attacker-controlled)". That holds only for votes.
Proof: `effectiveCommitteeUid = entityCommitteeUid ?? committeeUid` and `entityCommitteeUid = entity?.committee_uid ?? null` (line 104). Of the four probes (lines 84-88), only `voteService.fetchVote` returns a payload carrying `committee_uid` (the probe type at 80-82 makes it optional). `Meeting` has no `committee_uid` field, so on the `/meetings/:id/edit` route (also `entityScopedSlug: true`, `writeFeature: 'meetings'`, and in `COMMITTEE_WRITE_FEATURES`) `entityCommitteeUid` is always null and `effectiveCommitteeUid` degrades to the attacker-supplied `?committee_uid=` exactly as before.
Why it matters: code-truthfulness is an explicit review dimension. This doc block is the contract a future editor reads before adding the next probe; as written it advertises a URL-param-independent committee check that only 1 of the committee-write entity-scoped routes actually has. (Real authorization is still enforced upstream, so this is a doc/consistency issue, not an escalation.)
Fix: qualify the sentence, e.g. "from the probed entity itself when its probe surfaces one (today: votes), else from the `committee_uid` query param", and consider a meetings-route spec case pinning the current URL-param behavior so the gap is explicit.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -24,12 +25,13 @@ import { hasMeetingWriteAccess, resolveEntityWriteSlug } from '../utils/write-ac
...
- *    `'surveys'`, or `'votes'`. The backend ruleset allows committee:uid#writer to
- *    create resources associated with their committee.
+ * 3. `committee.writer` — committee writer; accepted only when `writeFeature` is one of
+ *    `'meetings'`, `'surveys'`, or `'votes'` and a committee uid is available — from the probed
+ *    entity itself on `entityScopedSlug` edit routes (URL params are attacker-controlled), else
+ *    from the `committee_uid` query param (create routes). The backend ruleset allows
+ *    committee:uid#writer to create resources associated with their committee.
  *
- * Slug resolution: on routes flagged `data.entityScopedSlug` (meeting/group edit), resolves
+ * Slug resolution: on routes flagged `data.entityScopedSlug` (meeting/group/vote edit), resolves
```
dealako: [nit] Route enumeration omits mailing-lists

Issue: This line reads "routes flagged `data.entityScopedSlug` (meeting/group/vote edit)" but drops mailing-lists.
Proof: `mailing-lists.routes.ts` sets `entityScopedSlug: true` and the registry this same hunk touches has a `'mailing-lists'` probe (line 87), covered by `writer.guard.spec.ts`.
Fix: `(meeting/group/mailing-list/vote edit)`. Minor, but this is the only inline map of which routes get entity-scoped resolution, so keeping it complete helps the next editor.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -573,6 +613,155 @@ export class VoteManageComponent {
...
+        })
+      )
+      // No initialValue: undefined doubles as the leg's pending state (see the doc above).
+    );
+    // Committee leg: edit mode derives the uid from the loaded vote's own committee (URL only as
+    // fallback), mirroring writerGuard's entity-scoped resolution — otherwise the guard admits a
+    // committee writer on a context-less edit link while this predicate evicts them. undefined stays
+    // pending; null resolves the leg immediately so it never counts as pending.
+    const committeeUid$: Observable<string | null | undefined> = editVoteId
+      ? toObservable(this.vote).pipe(map((vote) => (vote ? (vote.committee_uid ?? this.committeeUidFromUrl) : undefined)))
```
dealako: [minor] Committee write-access leg fires an HTTP round trip on every edit load

Issue: `committeeUid$` resolves from the loaded vote's own `committee_uid`, and because `toSignal(...)` subscribes eagerly, `fetchCommittee(uid)` runs unconditionally once the vote lands, including the common case where the project leg already grants access.
Proof: Every vote carries a committee (the form declares `committee` with `Validators.required`), so `committeeUid` is always truthy here. The `computed` at 675-685 reads `committeeAccess()` unconditionally, and `writerGuard` short-circuits `of(true)` on `project.writer === true` before ever calling `checkCommittee` (writer.guard.ts:139-155). `meeting-manage` fires its committee leg only when `committeeUidFromUrl` is present, not on every load.
Why it matters: one extra `GET /api/committees/:uid` per vote edit navigation, feeding an eviction predicate whose committee leg only matters when the project leg is not `true`.
Fix: defer the committee probe until the project leg resolves to something other than `true`, keeping `undefined` as the pending state, e.g. drive `committeeUid$` through `toObservable(projectAccess).pipe(filter((p) => p !== true), ...)` before the `switchMap`.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -573,6 +613,155 @@ export class VoteManageComponent {
...
+      }
+      return false;
+    });
+  }
+
+  /**
+   * Fallback for BFF-enrichment failure (project_uid without project_slug): resolves the project by uid,
+   * re-applied on NavigationEnd; if the relation-gated lookup returns null, one fresh vote re-fetch tries the ungated enrichment.
+   */
+  private initVoteContextFallback(): void {
```
dealako: [minor] New reactive context/write-access logic ships without a unit spec

Issue: ~150 lines of new behavior land in `VoteManageComponent` (provisional-true write predicate, NavigationEnd re-apply, one-time `contextFallbackRetried` fresh fetch, transient-failure release, and `resolvedContextCache`) with no `vote-manage.component.spec.ts`.
Proof: The directory contains only html/scss/ts. The equivalent meeting behaviors are unit-pinned (`meeting-manage.component.spec.ts` has cases for the relation-gated fresh fetch, the no-repeat-on-re-apply, and the transient-failure release). `resolvedContextCache` (lines 93-96) is the one behavior with no meeting counterpart, so an "align with meetings" cleanup could delete it and leave every test green while silently reintroducing MainLayout's route-lens re-assert clobbering the resolved context.
Why it matters: the re-apply path is the single thing this flow does that the mirrored meeting flow does not; nothing fails if it regresses.
Fix: add `vote-manage.component.spec.ts` mirroring the three meeting fallback tests against `VoteService.fetchVote`/`getVote`, plus one that pushes a second `NavigationEnd` and asserts the cached context re-applies without a second `fetchVote(uid, { skipCache: true })`.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -83,6 +89,12 @@ export class VoteManageComponent {
   public readonly confirmingOpenVote = signal<boolean>(false);
   public readonly loading = signal<boolean>(false);
   private readonly internalStep = signal<number>(1);
+  private readonly contextFallbackRetried = new Set<string>();
+  // Resolved fallback contexts by vote uid — plain Map, never read reactively. getProject caches
```
dealako: [nit] `resolvedContextCache` diverges from the meeting/mailing-list fallback it was copied from

Issue: `initVoteContextFallback`/`resolveContextFromFreshVote` are line-for-line the meeting versions apart from this cache lookup, so three near-identical copies of the unenriched-payload fallback now behave differently.
Proof: `meeting-manage.component.ts` declares only `contextFallbackRetried` (no resolved-context map); `mailing-list-manage.component.ts` has neither the fresh re-fetch nor a cache. The cache premise (a cached `null` from `ProjectService.getProject`'s `shareReplay(1)` replays on the next `NavigationEnd`) applies verbatim to meetings too.
Why it matters: the next editor of `meeting-manage` has no signal that the votes copy is the corrected one; the repo has been bitten by exactly this drift before (see `committee.utils.ts` "this copy once diverged on null-tier handling").
Fix: ideally lift the fallback into `entity-project-context.util.ts` next to `syncEntityProjectContext`, parameterized by the entity signal + a fresh-fetch fn, and have all three manage components call it. If that is out of scope for this PR, at minimum add the same map to `meeting-manage` so the copies agree.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -95,17 +107,28 @@ export class VoteManageComponent {
   public readonly isLastStep: Signal<boolean> = this.initIsLastStep();
   public currentStep: Signal<number> = this.initCurrentStep();
   public readonly isDraftSavable: Signal<boolean> = this.initIsDraftSavable();
+  public readonly voteEntityContext: Signal<EntityWithProject | null> = this.initVoteEntityContext();
```
dealako: [nit] `voteEntityContext` is `public`; sibling manage components keep this adapter private

Issue: This signal is declared `public readonly` while `meeting-manage`, `committee-manage`, and `mailing-list-manage` all declare the equivalent adapter `private`.
Proof: Its only reads are internal (`syncEntityProjectContext` at ~120, and the fallback at ~693/696); `vote-manage.component.html` does not reference it.
Fix: change to `private readonly voteEntityContext`. `@typescript-eslint/explicit-member-accessibility` accepts either keyword, so lint won't catch it.

apps/lfx-one/src/app/modules/votes/components/votes-table/votes-table.component.ts
```diff
@@ -276,9 +278,21 @@ export class VotesTableComponent {
       endedEarlyTooltip = getVoteEndedEarlyDetailTooltip(formattedEarlyClose);
     }
 
+    // Canonical edit URL derives from the VOTE's own project tier (is_foundation), not the viewer's
+    // active lens; falls back to the flat path (lensRedirectGuard) when the tier is unenriched.
+    const editCommands = getEntityCommands('votes', vote.uid, vote.is_foundation, 'edit') ?? ['/votes', vote.uid, 'edit'];
```
dealako, line 283: [question] Does the vote query-service index actually populate `project_slug`/`is_foundation` on list rows?

Issue: The canonical per-row edit tier and `?project=` are derived here from `vote.is_foundation` and `vote.project_slug`, but the list path (`VoteService.getVotes` -> `normalizeIndexedVote`) never enriches those fields on the BFF; it only spreads the indexed payload. So this behavior depends entirely on the query-service index carrying them.
Proof: `poll.interface.ts:201-206` documents these as "Carried by indexed list rows", and `normalizeIndexedVote` preserves them via `...rest`, so the contract says they are present. But unlike `getMyVotes` (which calls `enrichWithProjectData`), `getVotes` has no BFF enrichment or assertion, and `e2e/vote-edit-url.spec.ts` injects `project_slug`/`is_foundation` into the stubbed `/api/votes` response synthetically, so the e2e cannot detect a gap.
Why it matters: if the index does not in fact carry these, every Edit button on the main votes dashboard and the committee-votes tab renders the flat `/votes/:uid/edit` fallback and the headline canonical-tier behavior only engages after the detail payload lands and re-navigates, i.e. it silently no-ops on the primary surface.
Ask: can you confirm the indexer populates these on vote rows (a BFF-level assertion on the real shape, or dropping the synthetic injection from the e2e fixture, would pin it)? If the index does carry them, this is a non-issue and only the test-fixture realism point stands.

apps/lfx-one/e2e/vote-edit-url.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+const PROJECT_UID = 'p0000000-0000-0000-0000-00000000d003';
+const COMMITTEE_UID = 'c0000000-0000-0000-0000-00000000d001';
+
+const FOUNDATION_VOTE_UID = 'v0000000-0000-0000-0000-00000000d101';
+const PROJECT_VOTE_UID = 'v0000000-0000-0000-0000-00000000d102';
+const UNENRICHED_VOTE_UID = 'v0000000-0000-0000-0000-00000000d103';
+const UNENRICHED_PROJECT_UID = 'p0000000-0000-0000-0000-00000000d109';
+
+const LENS_COOKIE = 'lfx-active-lens';
+const SELECTED_FOUNDATION_COOKIE_KEY = 'lfx-selected-foundation';
```
dealako: [minor] Redeclare of shared cookie-key constants (one shadowing the shared export name)

Issue: Lines 38-39 redeclare two exported shared constants as local string literals: `const LENS_COOKIE = 'lfx-active-lens'` and `const SELECTED_FOUNDATION_COOKIE_KEY = 'lfx-selected-foundation'` (the second under the exact name of the shared export).
Proof: The values live in `lens.constants.ts` (`LENS_COOKIE_KEY`) and `project-context.constants.ts` (`SELECTED_FOUNDATION_COOKIE_KEY`). The specs this PR mirrors import them: `mailing-list-edit-url.spec.ts` and `project-context-deep-link.spec.ts` (edited in this same PR) both `import { LENS_COOKIE_KEY, ... SELECTED_FOUNDATION_COOKIE_KEY } from '@lfx-one/shared/constants'`. The literals are consumed in `seedFoundationLens`.
Why it matters: the duplication is silent. Rename the shared cookie key and this spec keeps seeding a cookie nothing reads, so the foundation lens is never seeded and the test exercises whatever the shell defaults to, without failing. The name collision also reads like an import.
Fix: delete lines 38-39, `import { LENS_COOKIE_KEY, SELECTED_FOUNDATION_COOKIE_KEY } from '@lfx-one/shared/constants';`, and use `LENS_COOKIE_KEY` at the `seedFoundationLens` call site.

apps/lfx-one/src/app/shared/components/button/button.component.html
dealako, line 42: [nit] Anchor link-button blocks navigation while disabled but doesn't reflect the disabled state

Agreeing with @ahmedomosanya's inline note here. Navigation itself is correctly blocked (`[routerLink]="disabled() || loading() ? null : routerLink()"`, and RouterLink leaves `href` null when the tree is null), so the earlier Copilot concern about navigating-while-disabled is addressed. What remains: the `<a pButton>` branch never binds `[disabled]`, so PrimeNG's disabled styling doesn't apply and the anchor stays in the tab order, and it also drops `[iconPos]`/`[badge]`/`[variant]`/`[autofocus]`/`[loadingIcon]`/`[pt]` that the old `p-button` branch applied.
Proof: `button.component.html:31-40` omits those bindings vs the removed `p-button` bindings. No current `<lfx-button [routerLink] [disabled]>` call site exists, so nothing regresses today; it's latent.
Fix (matches the human suggestion): add `[disabled]="disabled() || loading()"`, `[attr.tabindex]="disabled() || loading() ? -1 : tabindex()"`, and `[iconPos]="iconPos()"` to the anchor so the rendered state tracks the already-correct navigation guard.

dealako (changes requested):
🔴 **Needs changes before approval** — no blocking correctness or security defects; a security + data-privacy pass came back clean. The items are polish: guard-doc truthfulness (committee override only holds for votes), an unconditional committee fetch per edit load, a missing `vote-manage` unit spec for ~150 lines of new reactive logic, and a shared-constant duplication in the new e2e, plus one open question on whether the vote index actually carries `project_slug`/`is_foundation` on list rows. Full breakdown and prior-round reconciliation: https://github.com/linuxfoundation/lfx-self-serve/pull/2026#issuecomment-5500185983

apps/lfx-one/src/app/shared/components/button/button.component.html
audigregorie: Addressed in 6a93294, with one deviation from the suggestion: [disabled] can't bind on an <a pButton> — PrimeNG's ButtonDirective has no disabled input and anchors have no disabled property, so it fails strictTemplates with NG8002. Instead the anchor gets [class.p-disabled] while disabled/loading (PrimeNG's base CSS gives .p-disabled the same disabled opacity token plus pointer-events: none / cursor: default, and suppresses the :not(:disabled) hover styles), and [attr.tabindex] drops it to -1 while disabled/loading. The attr binding also fixes a footgun in the previous [tabindex] property binding: binding undefined reflected a tabindex="0" attribute, which kept href-less (disabled) anchors in the tab order.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
audigregorie: Fixed in 2e6b453 — the doc block and the inline comment now state the probed entity's committee wins only when the probe surfaces one (today only the votes probe carries committee_uid; the meetings probe falls through to the URL param). Held off on the meetings-route spec case — happy to pin that in a follow-up if you want it.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
audigregorie: Fixed in 2e6b453 — enumeration now reads (meeting/group/mailing-list/vote edit).

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
audigregorie: Fixed in 2e6b453 — the committee leg is gated on the project leg resolving false (distinctUntilChanged'd), so a granted project leg never fires fetchCommittee; while the project leg is pending the committee leg stays pending, preserving the provisional-true semantics. (The new vote-manage spec caught an injection-context violation in my first cut of this — toObservable can't be created lazily inside switchMap — fixed by hoisting the vote-sourced uid stream.)

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
audigregorie: Fixed in 2e6b453 — added vote-manage.component.spec.ts mirroring the meeting fallback tests (relation-gated fresh fetch, transient-failure release) plus the re-apply case you called out: a second NavigationEnd re-applies the cached context without a second fetchVote(uid, { skipCache: true }).

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
audigregorie: Fixed in 2e6b453 — went with your ideal: the fallback now lives in entity-project-context.util.ts as syncEntityProjectContextFallback (entity signal + optional freshFetch + canonicalizeRoute), and all three manage components call it. Your drift concern cut both ways: the meeting copy had the latent clobber bug — after a fresh-fetch resolution, the next NavigationEnd hit the retry-once guard and never re-applied. Meetings now get the resolved-context cache. Mailing-lists deliberately call it without freshFetch to preserve current behavior.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
audigregorie: Fixed in 2e6b453 — voteEntityContext is now private readonly, matching the sibling manage components.

apps/lfx-one/src/app/modules/votes/components/votes-table/votes-table.component.ts
audigregorie: Confirmed against upstream — you were right to ask. lfx-v2-voting-service's VoteData (internal/domain/event_models.go, stored as-is in the index per its indexer contract) carries project_uid/project_name but NOT project_slug or is_foundation, so the canonical tier never engaged on list rows. Fixed in 2e6b453: BFF getVotes now enriches rows via projectService.enrichWithProjectData (the getMyVotes helper — batched by unique project_uid, fail-soft on lookup failure), and the poll.interface docs now say BFF-enriched instead of 'carried by indexed list rows'. The e2e fixture's injected fields now match what the real BFF returns rather than masking the gap. Adding the fields to upstream VoteData may still be worth a follow-up ticket.

apps/lfx-one/e2e/vote-edit-url.spec.ts
audigregorie: Fixed in 2e6b453 — literals deleted; the spec now imports LENS_COOKIE_KEY and SELECTED_FOUNDATION_COOKIE_KEY from @lfx-one/shared/constants like the sibling specs.

apps/lfx-one/src/app/shared/components/button/button.component.html
audigregorie: Partially addressed before your comment landed — 6a9329412 added class.p-disabled and attr.tabindex=-1 while disabled/loading, so styling and tab-order are covered. In 2e6b453 I added iconPos to both anchor branches. I deliberately did NOT add [disabled]: ButtonDirective (pButton) has no disabled input in this PrimeNG version, so the binding would be a dead attribute on an anchor — the disabled state is carried by p-disabled + tabindex + aria-disabled.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -60,7 +61,12 @@ export class VoteService {
       params
     );
 
-    const votes = resources.map((resource) => this.normalizeIndexedVote(req, resource.data));
+    // Enrich list rows with canonical project fields — the vote index (VoteData) carries only
+    // project_uid/name, so without this, consumers deriving per-row edit links get no slug/tier.
+    const votes = await this.projectService.enrichWithProjectData(
+      req,
+      resources.map((resource) => this.normalizeIndexedVote(req, resource.data))
+    );
```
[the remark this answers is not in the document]
audigregorie: Fixed in 530c5d1fa — list enrichment now resolves unique project UIDs through the ungated, batched getProjectsByIds query-service lookup (new private enrichWithProjectMetadata in VoteService, wired into getVotes and getMyVotes) and maps project_name/project_slug/is_foundation/parent_project_uid onto each row, so committee writers without a project viewer relation get fully enriched rows and foundation votes render the canonical /foundation/... edit URL.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -60,7 +61,12 @@ export class VoteService {
       params
     );
 
-    const votes = resources.map((resource) => this.normalizeIndexedVote(req, resource.data));
+    // Enrich list rows with canonical project fields — the vote index (VoteData) carries only
+    // project_uid/name, so without this, consumers deriving per-row edit links get no slug/tier.
+    const votes = await this.enrichWithProjectMetadata(
+      req,
+      resources.map((resource) => this.normalizeIndexedVote(req, resource.data))
+    );
```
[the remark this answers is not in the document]
audigregorie: Fixed in 94d7fc921 — vote.service.spec.ts now covers the enrichment end to end: a getVotes test starts from an index row carrying only project_uid and asserts the canonical project fields (slug/name/is_foundation/parent_project_uid) from the getProjectsByIds map, plus a map-miss passthrough case pinning the flat-URL fallback, and the equivalent getMyVotes path.

dealako:
## Follow-up review — round 2

Hi @audigregorie, thanks for the thorough turnaround. This revision is a real step up: the fallback is now a single shared `syncEntityProjectContextFallback` in `entity-project-context.util.ts` with `meeting-manage` and `mailing-list-manage` refactored onto it (no more three-way copy), the guard doc is scoped to votes-only, the cookie-key constants are imported from the shared module, `voteEntityContext` is private, and the button uses `p-disabled` + `tabindex` instead of binding `[disabled]` on the PrimeNG directive. The two things I most wanted pinned down — the committee HTTP round-trip on every edit load, and whether LIST rows carry the slug/tier the canonical edit links need — are both handled: `initWriteAccess` now gates the committee leg on the project leg resolving false, and `getVotes`/`getMyVotes` enrich through the ungated `getProjectsByIds` with a real service spec asserting the shape.

### Prior round: 8 of 9 resolved, 1 partial

**Resolved**
- **M1** guard doc overclaim → now scoped to the votes edit route only. ✅
- **M2** committee `fetchCommittee` fired on every edit load → now gated on `projectAccess` resolving false; granted project leg returns `of(null)` without the HTTP probe. ✅
- **M4** duplicated cookie-key constant → imported from `@lfx-one/shared`. ✅
- **N1** fallback not extracted → `syncEntityProjectContextFallback` shared util; `meeting-manage`/`mailing-list-manage` refactored onto it. ✅
- **N2** `voteEntityContext` public → now `private`. ✅
- **N3** doc enum missing mailing-list → updated. ✅
- **N4** `[disabled]` bound on the PrimeNG directive → `p-disabled` class + `tabindex` guard. ✅
- **Q1** LIST rows lacked `project_slug`/`is_foundation` for per-row edit tiers → `getVotes`/`getMyVotes` now enrich via the ungated `getProjectsByIds`, and `vote.service.spec.ts` pins the enriched shape and the lookup-miss (flat-URL fallback) path. ✅

**Partially addressed**
- **M3** (unit coverage for the new reactive logic): the service enrichment is now well covered, and the vote-manage spec pins the relation-gated fresh-fetch fallback, the cache re-apply across NavigationEnds, and the transient-retry release. What's still uncovered is the **fail-closed write predicate** — see the new minor below.

### New this round

- **[minor]** `getVotes` now enriches unconditionally, so the `committee-activity` caller pays a discarded `/query/resources` project fetch on every Recent Activity request (and the fan-out doc comment is now stale). Inline on `vote.service.ts`.
- **[minor]** the new project-gated committee leg + fail-closed eviction path ship untested — `fetchCommittee` is mocked to always grant, so deleting the gate or inverting the branch still passes. Inline on `vote-manage.component.ts`. (This is the remaining half of M3.)
- **[nit]** two comments still name the deleted `initMeetingContextFallback` (`committee-manage.component.ts:436`, `meeting-manage.component.spec.ts:38`) — both sit outside this PR's diff, so noting here rather than inline.
- **[nit]** `Vote.project_name` doc (`poll.interface.ts:201`) still says "re-enriches it canonically from the project record" while the sibling `project_slug`/`is_foundation` docs were rewritten to "BFF-enriched on list rows (GET /api/votes) and the authenticated detail response." All three now come from one path; align line 201 with its siblings.

### Bot reconciliation
- **Copilot (10:03, `enrichWithProjectData` relation-gated):** agree with the concern — and it's resolved by this PR's switch to the ungated `getProjectsByIds`, the same path the detail route already uses for exactly this committee-writer persona.
- **Copilot (10:28, "the added service suite never calls `getVotes`/`getMyVotes`"):** stale — the `test(votes)` commit landed ~17 min after that comment and adds `getVotes`/`getMyVotes` coverage asserting `getProjectsByIds`. The related point that the E2E suites stub rows already carrying `project_slug`/`is_foundation` stands, but the new unit spec exercises real enrichment, so it's covered.

### Issue count (open)
- **Blocking:** 0
- **Minor:** 2 — discarded enrichment on the committee-activity `getVotes` caller; untested fail-closed committee gate.
- **Nit:** 2 — stale `initMeetingContextFallback` comments; stale `project_name` doc line.

### Decision
✅ **Approved with minor comments.** No blocking, security, or privacy findings, and the branch is mergeable. The two minors are a perf opt-out and a test-coverage add — worth doing, but not gating. Nice work on this round.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -60,7 +61,12 @@ export class VoteService {
       params
     );
 
-    const votes = resources.map((resource) => this.normalizeIndexedVote(req, resource.data));
+    // Enrich list rows with canonical project fields — the vote index (VoteData) carries only
+    // project_uid/name, so without this, consumers deriving per-row edit links get no slug/tier.
+    const votes = await this.enrichWithProjectMetadata(
```
dealako: [minor] getVotes enriches unconditionally, so the committee-activity caller pays for a project fetch it discards

Issue: `getVotes` now always awaits `enrichWithProjectMetadata` -> `projectService.getProjectsByIds`, which issues a `/query/resources` project fetch. `getVotes` has a second caller that never reads the enriched fields, so it now pays an extra upstream round trip per request.

Proof: `committee-activity.service.ts:681` calls `this.voteService.getVotes(req, query)` and maps every row through `mapVoteToEvent`, which reads only `status`, `early_end_time`, `end_time`, `last_modified_time`, `creation_time`, `uid`, `name`, and `committee_uid` — never `project_slug`/`project_name`/`is_foundation`/`parent_project_uid`. The await lands on the critical path of that leg inside the Recent Activity `Promise.all` fan-out, and the file's own doc comment still describes a 9-call / 6-query-leg fan-out that is now 10 / 7.

Why it matters: every committee Recent Activity request makes an upstream query-service call whose result is thrown away, and the documented fan-out cost is silently wrong.

Fix: give `getVotes` an opt-in flag mirroring the `includeProject` option `getVoteById` already carries — e.g. `getVotes(req, query, { includeProject = true } = {})` — and have the committee-activity path pass `includeProject: false`; or move the enrichment into the controller so only the HTTP list path pays for it.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
```diff
@@ -573,6 +611,90 @@ export class VoteManageComponent {
...
+    );
+    // Edit mode derives the uid from the loaded vote's own committee (URL only as fallback),
+    // mirroring writerGuard's entity-scoped resolution. Hoisted: toObservable needs the injection
+    // context, so it can't be created lazily inside the switchMap below.
+    const editCommitteeUid$: Observable<string | null | undefined> = toObservable(this.vote).pipe(
+      map((vote) => (vote ? (vote.committee_uid ?? this.committeeUidFromUrl) : undefined))
+    );
+    // Gated on the project leg resolving false: a granted project leg makes the committee fetch
+    // moot, and while it's pending this leg stays pending too. undefined stays pending; null resolves immediately.
+    const committeeUid$: Observable<string | null | undefined> = toObservable(projectAccess).pipe(
```
dealako, line 659: [minor] the new project-gated committee leg and its fail-closed eviction path ship untested

Issue: this commit rewrote the committee leg of the write-access predicate to subscribe through `toObservable(projectAccess)` and short-circuit to `of(null)` when the project leg grants, and the new `vote-manage.component.spec.ts` never exercises it.

Proof: `vote-manage.component.spec.ts:117-122` pins `fetchCommittee` to always return `{ writer: true }` ("so the eviction predicate stays quiet"), and the three tests assert only on `fetchVote`/`setProject`. Under that mock, deleting this gate, dropping `filter(project !== undefined)`, or inverting the `if (project)` branch all still pass, and the fail-closed transition into `evictOnWriteAccessLoss` is never reached because `writeAccess` stays provisionally true for the whole suite. This is the remaining half of prior-round M3.

Why it matters: this eviction predicate mirrors `writerGuard` — it's the one place a silent regression logs a guard-admitted committee writer out mid-edit.

Fix: add two cases — (1) `getProject` resolving `{ writer: true }` asserting `expect(fetchCommittee).not.toHaveBeenCalled()` (pins the short-circuit); (2) `getProject` null + `fetchCommittee` returning `{ writer: false }` asserting `router.navigateByUrl` was called (pins the fail-closed transition).

dealako (dismissed):
Round-2 follow-up. 8 of 9 prior items resolved, M3 partially addressed. Two new minors (a discarded project enrichment on the committee-activity `getVotes` caller, and the untested fail-closed committee gate) plus two doc nits. No blocking, security, or privacy findings; branch is mergeable. Approving with minor comments — see the summary comment for the full breakdown.

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.ts
audigregorie: Fixed in d9bb0db40 — added both pinning cases to vote-manage.component.spec.ts: (1) a granted project leg asserts `fetchCommittee` is never called (short-circuit); (2) a null project lookup + committee denial asserts the fail-closed transition fires `router.navigateByUrl`. Note: the eviction test defers the committee decision via a Subject — a synchronous `of()` collapses the whole chain into the first effect pass and `skip(1)` then swallows the eviction, so the two-phase timing has to be explicit.

apps/lfx-one/src/server/services/vote.service.ts
audigregorie: Fixed in d9bb0db40 — `getVotes` now carries an `includeProject` opt-out (mirroring `getVoteById`), and the committee-activity leg passes `{ includeProject: false }`, so the discarded project-enrichment round trip is gone and the 9-call / 6-query-leg fan-out doc is accurate again.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -89,23 +95,35 @@ export class VoteService {
...
    */
-  public async getVoteById(req: Request, voteUid: string): Promise<Vote> {
+  public async getVoteById(req: Request, voteUid: string, options: { includeProject?: boolean } = {}): Promise<Vote> {
+    const { includeProject = false } = options;
     logger.debug(req, 'get_vote_by_id', 'Fetching vote by ID', {
       vote_uid: voteUid,
     });
 
-    const vote = await this.microserviceProxy.proxyRequest<Vote>(req, 'LFX_V2_SERVICE', `/votes/${voteUid}`, 'GET');
+    const vote = await this.microserviceProxy.proxyRequest<Vote>(req, 'LFX_V2_SERVICE', `/votes/${encodeURIComponent(voteUid)}`, 'GET');
```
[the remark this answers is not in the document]
audigregorie: Fixed in 85596e30e — a private `encodeVoteUid()` guard now rejects exact dot segments (`.`/`..`) with a `ServiceValidationError` before any upstream path is built, and all nine `/votes/<uid>` interpolations route through it. After encoding, exact dot segments were the only values that could survive as normalizable path segments (pre-encoded variants like `%2e%2e` are neutralized to `%252e%252e`), so this closes the vector without changing the opaque-uid encode behavior your earlier thread pinned. Added `.`/`..` rejection tests per method alongside the slash-encoding tests in `vote.service.spec.ts`.

dealako:
## Follow-up review — round 3

Hi @audigregorie, nice turnaround again. Both round-2 minors are cleanly resolved, and you also folded in a genuinely good security hardening this round.

### 👏 Nice work
- **CP2-1 (perf) resolved.** `getVotes` now takes an `includeProject` opt-in (default `true`), and `committee-activity.service.ts:683` passes `includeProject: false`, so the Recent Activity fan-out no longer pays a `/query/resources` round trip whose result `mapVoteToEvent` discards. The `9-call fan-out / 6 query legs` docblock is accurate again without needing an edit, the HTTP list path still enriches (the controller omits the options arg), and `getMyVotes` still enriches.
- **CP2-2 (test) resolved.** Both eviction cases are now load-bearing: `vote-manage.component.spec.ts:192-199` pins the `of(null)` short-circuit (asserting `fetchCommittee` is not reached), and `:201-217` pins the fail-closed `writeAccess` true→false transition using a deferred `Subject` so the provisional-true boot emission is correctly skipped.
- **Dot-segment hardening (`85596e30e`).** `encodeVoteUid` rejects exactly `.`/`..` and throws a 400 before proxying, applied at every uid→path site (`getVoteById`, `updateVote`, `deleteVote`, `enableVote`, `getVoteResults`, `getMyVotes`). Traced end to end: it's fail-closed, covers the whole surface, and the `v1.2` negative control guards against over-rejection. Good catch and a clean fix.

### Prior round
- ✅ **CP2-1** perf gate — resolved.
- ✅ **CP2-2** eviction-predicate coverage — resolved.
- ❌ **SD2-1** (nit) — still open: two comments still name the deleted `initMeetingContextFallback` (`committee-manage.component.ts:436`, `meeting-manage.component.spec.ts:38`). Both sit outside this PR's diff, so purely optional cleanup.
- ❌ **SD2-2** (nit) — still open: `Vote.project_name` doc (`poll.interface.ts:201`) still says "re-enriches it canonically from the project record" while its `project_slug`/`is_foundation` siblings were rewritten to "BFF-enriched on list rows (GET /api/votes) and the authenticated detail response." All three come from one path now; align line 201.

### New this round
- **[minor]** the `includeProject: false` branch of `getVotes` is unpinned at the service level — the committee-activity spec mocks `VoteService`, so it proves the flag is *passed*, not *honored*. Reverting the gate to unconditional keeps the suite green. Inline on `vote.service.ts`.
- **[nit]** the new `getVotes` docblock says the flag "enriches rows with project slug/name/tier," but the code sets four fields including `parent_project_uid` (which the votes dashboard filters on), and "tier" isn't a field name. Inline on `vote.service.ts`.
- **[nit]** the eviction test is named "evicts to the overview" but only asserts `navigateByUrl` was called, not the destination. Inline on `vote-manage.component.spec.ts`.

### Bot reconciliation
- **Copilot (18:00, `..` still builds `/votes/..`):** agree with the concern — and it's resolved by the very next commit `85596e30e`, which added `encodeVoteUid` rejecting `.`/`..` before proxying, exactly what Copilot asked for (plus the `.`/`..` negative tests). That comment reviewed `d9bb0db40`, 44 min before the fix landed.

### Issue count (open)
- 🔴 **Blocking:** 0
- 🟡 **Minor:** 1 — `includeProject: false` branch untested at the service level.
- ⚪ **Nit:** 4 — `getVotes` doc field set; eviction test name vs assertion; and the two carried-over doc nits (SD2-1, SD2-2).

### Decision
✅ **Approved with minor comments.** No blocking, security, or privacy findings, and the branch is mergeable. The one minor is a single counterpart test; the nits are doc/naming polish. Solid work across all three rounds.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -60,7 +63,10 @@ export class VoteService {
       params
     );
 
-    const votes = resources.map((resource) => this.normalizeIndexedVote(req, resource.data));
+    const normalized = resources.map((resource) => this.normalizeIndexedVote(req, resource.data));
+    // Enrich list rows with canonical project fields — the vote index (VoteData) carries only
+    // project_uid/name, so without this, consumers deriving per-row edit links get no slug/tier.
+    const votes = includeProject ? await this.enrichWithProjectMetadata(req, normalized) : normalized;
```
dealako, line 69: [minor] Pin the `includeProject: false` branch of getVotes with a service-level test

Issue: this gate is the whole point of d9bb0db40 (it keeps the Recent Activity fan-out from paying a wasted `/query/resources` round trip), but nothing exercises the `false` branch at the service level.

Proof: `describe('getVotes')` in `vote.service.spec.ts` calls `service.getVotes(req)` with no options arg, so only the default-`true` path is covered. The committee-activity assertions run against a fully mocked `VoteService` (`committee-activity.service.spec.ts:70-72`), so they prove the flag is *passed* but say nothing about whether the service *honors* it. Reverting line 69 to `const votes = await this.enrichWithProjectMetadata(req, normalized);` keeps the entire suite green while silently reinstating the extra fetch.

Why it matters: the perf fix has no regression guard; a future refactor can drop the gate undetected.

Fix: mirror the negative test already present for `getVoteById` (`vote.service.spec.ts:135-140`, 'skips project enrichment entirely when includeProject is not set') — call `service.getVotes(req, {}, { includeProject: false })` with the same index-row and project fixtures, then assert `getProjectsByIds` was not called and `result.data[0]` still carries no `project_slug`/`is_foundation`.

apps/lfx-one/src/server/services/vote.service.ts
```diff
@@ -41,8 +42,10 @@ export class VoteService {
 
   /**
    * Fetches a single page of votes using cursor-based pagination — callers paginate via the returned page_token.
+   * `includeProject` (default true) enriches rows with project slug/name/tier; opt out when the caller discards them.
```
dealako, line 45: [nit] Document all four fields the includeProject flag controls

Issue: the docblock says the flag "enriches rows with project slug/name/tier," but `enrichWithProjectMetadata` (vote.service.ts:500-506) sets four fields: `project_name`, `project_slug`, `is_foundation`, and `parent_project_uid`. "tier" is also not a field name — the sibling docblock at vote.service.ts:99 (getVoteById) calls it `is_foundation`.

Proof: `parent_project_uid` is load-bearing on vote list rows — `votes-dashboard.component.ts:397` and `:415` filter the dashboard by it. The repo's sibling enrichment flags document the exact set (`committee.service.ts:331-333` lists all four).

Why it matters: the doc exists to tell the next caller what opting out costs. A caller copying committee-activity's `{ includeProject: false }` reads "slug/name/tier" and won't learn that `parent_project_uid` also disappears.

Fix:
```suggestion
   * `includeProject` (default true) enriches rows with `project_name`, `project_slug`, `is_foundation` and `parent_project_uid`; opt out when the caller discards them.
```

apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    fetchVote.mockReturnValue(of(enrichedVote()));
+    getProject.mockReturnValue(of(null));
+    fetchCommittee.mockReturnValue(committeeDecision$.asObservable());
+    const router = TestBed.inject(Router);
+    await createComponent();
+
+    committeeDecision$.next({ uid: COMMITTEE_UID, writer: false });
+    await TestBed.inject(ApplicationRef).whenStable();
+
+    expect(router.navigateByUrl).toHaveBeenCalled();
```
dealako, line 216: [nit] Assert the eviction destination, or drop "to the overview" from the test name

Issue: the test is named "evicts to the overview when both write-access legs resolve false" but its only assertion is `expect(router.navigateByUrl).toHaveBeenCalled()`, which pins that navigation happened, not where.

Proof: the destination is chosen in `evict-on-write-access-loss.util.ts:70-74` from `ProjectContextService.activeContext()?.slug` and `LensService.activeLens()`. The spec's `ProjectContextService` mock returns null (spec:114), so the util takes the `router.parseUrl(overviewPath)` branch, and the Router double's `parseUrl` returns the same object for any input (spec:86). Changing `overviewPath` to any other route keeps this test green.

Why it matters: the test correctly pins the fail-closed transition (the round-2 ask), but its name advertises a stronger contract than it verifies; a later refactor of the overview path has no test holding it.

Fix: add `expect(router.parseUrl).toHaveBeenCalledWith(expect.stringMatching(/^\/(project|foundation)\/overview$/));` alongside the existing assertion, or drop "to the overview" from the name.

dealako (approved):
Round-3 follow-up. Both round-2 minors (CP2-1 perf gate, CP2-2 eviction coverage) are resolved, and the new dot-segment rejection is a clean, fail-closed hardening covering every uid->path site. One new minor (the includeProject:false branch is unpinned at the service level) plus doc/naming nits. No blocking, security, or privacy findings; branch is mergeable. Approving with minor comments — full breakdown in the summary comment.

PR #2028, opened 2026-09-01
fix(sidebar): surface marketing section under project lens

MRashad26:
## Summary
- Hybrid-persona users with a marketing_auditor/campaign_manager FGA grant plus project-level access could land in the Project lens, where the Marketing section (Campaign Impact/Campaigns) was never built — only `foundationLensItems()` constructed it.
- Extracts the section into a shared `marketingSectionItem` computed and consumes it from both the `'foundation'` and `'project'` branches of `SidebarNavService.sidebarItems`, so it shows up regardless of which lens is active.
- Adds e2e coverage: S9 confirms marketing grants don't unlock Health Metrics (LFXV2-2237 gap), S10 confirms a hybrid marketing_auditor sees the Marketing section under Project lens (the actual LFXV2-2235 fix).

## Test plan
- [x] `yarn format && yarn lint && yarn build` clean
- [x] License headers pass (`./check-headers.sh`)
- [x] Reviewer trio (general + self-serve conventions + learnings) clean on every commit
- [x] Full-branch sweep (all three reviewers vs `origin/main`) clean
- [x] `/lfx-self-serve-pr-readiness` — READY
- [x] New e2e scenarios S9/S10 added to `marketing-access.spec.ts`

LFXV2-2235
LFXV2-2237

### DIFF
--- a/apps/lfx-one/e2e/marketing-access.spec.ts
+++ b/apps/lfx-one/e2e/marketing-access.spec.ts
@@ -39,7 +39,7 @@
  *   - apps/lfx-one/.env populated with TEST_USERNAME / TEST_PASSWORD (tests skip otherwise)
  */
 
-import type { LensItem, PersistedPersonaState, PersonaType } from '@lfx-one/shared/interfaces';
+import type { LensItem, NavLens, PersistedPersonaState, PersonaType } from '@lfx-one/shared/interfaces';
 import { FEATURE_FLAG_OVERRIDE_STORAGE_KEY, MARKETING_OPS_FGA_ENABLED_FLAG, PERSONA_COOKIE_KEY } from '@lfx-one/shared/constants';
 import { expect, Page, test } from '@playwright/test';
 
@@ -56,6 +56,12 @@ const MOCK_FOUNDATION_ITEM: LensItem = {
   name: 'Test Foundation',
   logoUrl: null,
   isFoundation: true,
+};
+
+/** Project-lens flavor of the mock item — `isFoundation: false` so a hybrid persona's `applyVisibilityFilters` doesn't strip it. */
+const MOCK_PROJECT_ITEM: LensItem = {
+  ...MOCK_FOUNDATION_ITEM,
+  isFoundation: false,
 };
 
 const SIDEBAR = {
@@ -121,11 +127,11 @@ async function stubPersona(page: Page, personas: string[], options: StubPersonaO
   );
 }
 
-async function stubNavLensItems(page: Page, items: LensItem[] = [MOCK_FOUNDATION_ITEM]): Promise<void> {
+async function stubNavLensItems(page: Page, items: LensItem[] = [MOCK_FOUNDATION_ITEM], targetLens: NavLens = 'foundation'): Promise<void> {
   await page.route('**/api/nav/lens-items*', (route) => {
     const url = route.request().url();
     const requestedLens = new URL(url).searchParams.get('lens') ?? 'foundation';
-    if (requestedLens !== 'foundation') {
+    if (requestedLens !== targetLens) {
       return route.fulfill({
         status: 200,
         contentType: 'application/json',
@@ -135,7 +141,7 @@ async function stubNavLensItems(page: Page, items: LensItem[] = [MOCK_FOUNDATION
     return route.fulfill({
       status: 200,
       contentType: 'application/json',
-      body: JSON.stringify({ items, next_page_token: null, upstream_failed: false, lens: 'foundation' }),
+      body: JSON.stringify({ items, next_page_token: null, upstream_failed: false, lens: targetLens }),
     });
   });
 }
@@ -397,7 +403,7 @@ test.describe('S10: Project lens — hybrid marketing_auditor (flag ON, contribu
     // Project lens, not just Foundation lens.
     await stubPersona(page, ['contributor'], { isMarketingAuditor: true });
     await setPersonaCookie(page, ['contributor']);
-    await stubNavLensItems(page);
+    await stubNavLensItems(page, [MOCK_PROJECT_ITEM], 'project');
     await stubProjectApi(page, MOCK_FOUNDATION_SLUG, false);
     await gotoAndWaitForSidebar(page, `/project/overview?project=${MOCK_FOUNDATION_SLUG}`);
   });
@@ -409,5 +415,6 @@ test.describe('S10: Project lens — hybrid marketing_auditor (flag ON, contribu
     await expect(page.getByTestId(SIDEBAR.marketingImpact), 'persona=hybrid-marketing_auditor lens=project item=marketing-impact').toBeVisible({
       timeout: ELEMENT_TIMEOUT,
     });
+    await expect(page.getByTestId(SIDEBAR.campaigns), 'persona=hybrid-marketing_auditor lens=project item=campaigns should be hidden').toHaveCount(0);
   });
 });
--- a/apps/lfx-one/e2e/marketing-access.spec.ts
+++ b/apps/lfx-one/e2e/marketing-access.spec.ts
@@ -27,6 +27,7 @@
  *   S7  Flag ON  — LF Staff stays Social-Listening-only on Marketing Impact even with marketing_auditor-equivalent access already granted via canViewExecutiveDashboards
  *   S8  Flag OFF (default) — grants present on the API response are ignored; behavior is byte-identical to pre-LFXV2-2236 ED/LF-staff-only gating
  *   S9  Flag ON  — marketing_auditor + campaign_manager grants do not unlock Health Metrics (ED/LF-Staff-only, LFXV2-2237)
+ *   S10 Flag ON  — hybrid marketing_auditor (contributor + marketing grant) sees Marketing section under Project lens too, not just Foundation (LFXV2-2235)
  *
  * This suite does NOT flip the server `LFX_MARKETING_OPS_FGA_ENABLED` env var or hit protected
  * analytics/campaigns routes directly — see `require-marketing-access.middleware.spec.ts` for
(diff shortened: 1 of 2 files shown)


apps/lfx-one/src/app/shared/services/sidebar-nav.service.ts
```diff
@@ -89,7 +89,13 @@ export class SidebarNavService {
...
         const mktgOsItems = this.isMktgOsAgentsEnabled() ? [this.mktgOsAgentsNavItem] : [];
         const base = [...this.projectLensItems, ...mktgOsItems, this.projectGovernanceSection];
-        return this.canSeeNewsletters() ? [...base, this.projectCommunicationsSection] : base;
+        const withComms = this.canSeeNewsletters() ? [...base, this.projectCommunicationsSection] : base;
+        // Marketing-only FGA users who are also hybrid personas (e.g. a project role plus a
+        // marketing_auditor/campaign_manager grant) land here via getAllowedLensIds()/isHybridPersona
+        // rather than the foundation lens — they must still reach Campaign Impact/Campaigns
+        // (LFXV2-2235 review finding: hybrid marketing users lost the Marketing section in project lens).
+        const marketingSection = this.marketingSectionItem();
+        return marketingSection ? [...withComms, marketingSection] : withComms;
```
audigregorie, line 98: **Minor — Marketing section now appears in the Project lens for ED and LF Staff too, not just hybrid marketing-grant users**

`marketingSectionItem()` gates on `canViewExecutiveDashboards()` (ED + LF Staff) for Campaign Impact and `currentPersona() === 'executive-director'` for Campaigns, so appending it here means every ED/LF-staff user browsing any project gains a Marketing section that pre-PR never existed in the project sidebar. The LFXV2-2235 bug this fixes only affected hybrid `marketing_auditor`/`campaign_manager` users.

The PR description ("so it shows up regardless of which lens is active") reads as if this broader surface is deliberate — can you confirm that's intended? If the target is strictly hybrid marketing-grant users, this branch would need to gate on the FGA grants rather than reuse the section wholesale. If it is intended, no code change needed — but worth a line in the PR description since it widens the sidebar for personas outside the bug's scope.

apps/lfx-one/e2e/marketing-access.spec.ts
```diff
@@ -351,3 +354,60 @@ test.describe('S8: Foundation lens — flag OFF (default) ignores marketing_audi
...
+    await gotoAndWaitForSidebar(page, `/project/overview?project=${MOCK_FOUNDATION_SLUG}`);
+  });
+
+  test('sees Marketing section with Marketing Impact under Project lens', async ({ page }) => {
+    await expect(page.getByTestId(SIDEBAR.marketingSection), 'persona=hybrid-marketing_auditor lens=project section=marketing').toBeVisible({
+      timeout: ELEMENT_TIMEOUT,
+    });
+    await expect(page.getByTestId(SIDEBAR.marketingImpact), 'persona=hybrid-marketing_auditor lens=project item=marketing-impact').toBeVisible({
+      timeout: ELEMENT_TIMEOUT,
+    });
```
audigregorie, line 458: **Minor — S10 never asserts Campaigns stays hidden under the Project lens**

This scenario grants `isMarketingAuditor` only, so Campaigns should be absent — the same negative assertion S1 makes in the Foundation lens. Adding it here locks the "neither grant implies the other" independence rule into the newly-surfaced Project-lens path, which is exactly where a future regression in `marketingSectionItem` would leak both items at once.

```suggestion
    });
    await expect(page.getByTestId(SIDEBAR.campaigns), 'persona=hybrid-marketing_auditor lens=project item=campaigns should be hidden').toHaveCount(0);
```

apps/lfx-one/src/app/shared/services/sidebar-nav.service.ts
```diff
@@ -410,13 +416,29 @@ export class SidebarNavService {
...
+  // Social Listening tab once inside — full Marketing Impact access is ED/marketing_auditor only
+  // (LFXV2-2236 gap-analysis G4). Never widen the Metrics section itself for marketing_auditor.
+  // Extracted so both foundationLensItems and the project-lens branch of sidebarItems (hybrid
+  // marketing users) can surface the same section (LFXV2-2235 review finding). Both items below
+  // route to /foundation/* paths tagged `lens: 'foundation'` in app.routes.ts, so a hybrid user
+  // clicking one from the project lens gets flipped into the foundation lens by
+  // MainLayoutComponent.syncLensFromRoute — the same lens-switch-on-navigate behavior the merged
+  // 'Projects' switcher entry already relies on (lens.service.ts switchLens/isHybridPersona), not
+  // an oversight introduced here.
+  private readonly marketingSectionItem = computed((): SidebarMenuItem | null => {
```
audigregorie, line 440: **Minor — new multi-line computed initialized inline rather than via an `init*()` function**

CLAUDE.md's new-code conventions call for multi-line `computed()`/`toSignal()` to use a private `init*()` initializer (`canSeeNewsletters = this.initCanSeeNewsletters()` earlier in this same file is the pattern). This one is ~40 lines inline as a class property.

Noting the nuance: the rule file (`component-organization.md`) is paths-scoped to `*.component.ts`, and the neighboring lens-item computeds (`sidebarItems`, `foundationLensItems`, `visibleOrgLensItems`) are all inline too — so this is a consistency judgment call, not a hard violation. Flagging per the top-violations list; resolve as won't-fix if the inline lens-item pattern is the intended local idiom.

audigregorie (commented):
## Code Review Summary

Clean, well-scoped fix. The extraction of `marketingSectionItem` is faithful — Foundation-lens gating and ordering are byte-equivalent — and the Project-lens surfacing is backed by a correct reactivity chain (signal-only reads) plus route guards that re-probe FGA with the propagated `?project=` param. The doc comment's claims about `syncLensFromRoute` and the hybrid 'Projects' switcher precedent check out against source. S9/S10 e2e coverage maps to the right testIds and redirect targets. No Critical or Major findings; three Minor threads below (one intent-confirmation question, one test-completeness nit, one style-consistency note).

### What's done well

- Faithful extraction with zero Foundation-lens behavior drift — same position, same gating, provably moved rather than rewritten.
- Doc comment documents the lens-flip side effect with mechanism and precedent — all verified accurate.
- S9 covers both layers of the LFXV2-2237 gap (sidebar absence + guard redirect).
- CI green across the board (Code Quality, CodeQL, DCO, license headers).

apps/lfx-one/e2e/marketing-access.spec.ts
```diff
@@ -351,3 +354,60 @@ test.describe('S8: Foundation lens — flag OFF (default) ignores marketing_audi
...
+  });
+});
```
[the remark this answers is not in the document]
MRashad26: Confirmed and fixed in f522f50ae — `stubNavLensItems` now takes a `targetLens` param (default `'foundation'`, so every existing call site is unaffected) and only fulfills the passed-in `items` for that lens, returning empty for any other. S10 now stubs with a project-flavored `MOCK_PROJECT_ITEM` (`isFoundation: false`) targeting `'project'` — a foundation-flavored item would still have been filtered out client-side by `applyVisibilityFilters` for a hybrid persona, so this needed a distinct mock, not just a lens-targeting fix. Good catch, this was silently dead code until S10 became the first scenario to hit a Project-lens route.

apps/lfx-one/e2e/marketing-access.spec.ts
MRashad26: Good catch — fixed in f522f50ae. Added the negative assertion you suggested: `expect(page.getByTestId(SIDEBAR.campaigns), ...).toHaveCount(0)` right after the two positive assertions, mirroring S1's pattern.

apps/lfx-one/src/app/shared/services/sidebar-nav.service.ts
MRashad26: Confirmed intended, no code change. `canSeeMarketingImpact` is satisfied by `canViewExecutiveDashboards()` alone, so ED/LF Staff do pick up the Campaign Impact link under Project lens now too — but this mirrors the existing Foundation-lens design rather than widening access: ED/LF Staff already see this link there, and `marketing-impact.component.html` independently restricts non-ED viewers to the Social Listening tab regardless of which lens they clicked through from. So this is one more entry point to a page whose content gating is unchanged, not a new surface. Happy to discuss further if you'd rather see the Project-lens branch scoped tighter to the marketing FGA grant only, but as-is it's consistent with how Marketing Impact already behaves for this population.

apps/lfx-one/src/app/shared/services/sidebar-nav.service.ts
MRashad26: Checked this against CLAUDE.md's convention (scoped to `**/*.component.ts`, doesn't reach services) and this file's own existing pattern: every other menu-building signal here (`visibleOrgLensItems`, `foundationLensItems`, and the rest) is also an inline `computed(...)`, with `init*()` extraction as the exception rather than the norm in this class. Keeping it inline for consistency with the surrounding code — no change.

MRashad26:
## Review Feedback Addressed

Commit: f522f50aecab9892dc709c5794d8aa40f93cb0a5

### Changes Made
- **apps/lfx-one/e2e/marketing-access.spec.ts**: `stubNavLensItems` ignored the `items` param for any lens other than `'foundation'`, always returning `[]` — dead code until S10 became the first scenario to hit a Project-lens route. Added a `targetLens` param (default `'foundation'`, existing call sites unaffected) and a project-flavored `MOCK_PROJECT_ITEM` (`isFoundation: false`, so it survives `applyVisibilityFilters` for hybrid personas). S10 now stubs project-lens items correctly. (per @cursor)
- **apps/lfx-one/e2e/marketing-access.spec.ts**: added the missing negative assertion in S10 — a `marketing_auditor`-only grant must not surface Campaigns. (per @audigregorie)

### Questions Answered
- **sidebar-nav.service.ts:98**: confirmed the Project-lens Campaign Impact link showing for ED/LF Staff is intended, not scope creep — it mirrors the existing Foundation-lens behavior (same population already sees this link there), and `marketing-impact.component.html` independently restricts non-ED viewers to Social Listening regardless of entry lens. No code change. (asked by @audigregorie)

### No Change Needed
- **sidebar-nav.service.ts:441**: the `init*()` extraction convention in CLAUDE.md is scoped to `**/*.component.ts` and doesn't reach services; every other menu-building signal in this file is already an inline `computed(...)`. Kept inline for consistency with the surrounding code. (flagged by @audigregorie)

### Threads Resolved

apps/lfx-one/src/app/shared/services/sidebar-nav.service.ts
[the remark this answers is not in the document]
MRashad26: Done — `marketingSectionItem` now reads `PersonaService.grantsByScope` for the active scope (with ROOT fallback) via a new `hasMarketingGrant()` helper, and `marketingPersonaSlug` no longer suppresses re-probing once a grant is confirmed, so switching to a newly-selected project always re-probes it. The suppression was safe to remove because `writeGrantForScope` writes an explicit `false` to *both* the queried scope key and ROOT on a denial — a denial for project B can't clobber project A's already-confirmed entry. Added S11 in `marketing-access.spec.ts`: marketing_auditor confirmed on project A, client-side switch to ungranted project B via the project selector, asserting the Marketing section disappears without a full page reload.

See commit 2459241d3.

MRashad26:
## Review Feedback Addressed

Commit: 60a77539c

### Changes Made
- **apps/lfx-one/src/app/shared/services/sidebar-nav.service.ts**: `marketingSectionItem` now reads `PersonaService.grantsByScope` for the active scope (with ROOT fallback) via a new `hasMarketingGrant()` helper, and `marketingPersonaSlug` no longer suppresses re-probing once a grant is confirmed — a client-side switch to a newly-selected project now always re-probes it. (per @copilot-pull-request-reviewer)
- **apps/lfx-one/e2e/marketing-access.spec.ts**: added S11 — marketing_auditor confirmed on project A, client-side switch (via the project selector, not a page reload) to ungranted project B, asserting the Marketing section disappears.

### Threads Resolved
1 of 1 unresolved threads addressed in this iteration (the other 4 threads on this PR were already resolved prior to this iteration).

audigregorie (approved):

Issue #2041, opened 2026-09-01
Meeting join page blanks at hydration — SSR-rendered meeting data is discarded


bramwelt:
## Problem

Item 4 of #1731 — the last in-repo fix item for meeting join page latency, deferred out of #1748 ("Item 4 of the GitHub issue (SSR-seeding of meeting data / join-page skeleton) — future follow-up"). The other three items (#1748, #1820, #1870) are merged, the M2M token concern turned out to be already cached (`apps/lfx-one/src/server/utils/m2m-token.util.ts:16`), and the upstream access-check root cause (#1653) is closed.

The issue described this as "the server doesn't send any meeting data with the initial page load." It's actually worse than that: the server *does* render the meeting, and the client throws it away.

**1. The server-rendered page is torn down at hydration.**

`meeting-join.component.html:10` is `@if (meeting())` with no `@else` (the block closes at 855). `meeting` comes from `initializeMeeting()` (`meeting-join.component.ts:686-751`) — a `toSignal` with no `initialValue`, behind `debounceTime(0)`. That fetch is *not* gated by `isPlatformServer`, so the SSR HTML genuinely contains meeting content. But at hydration `meeting()` is `undefined`, the `@if` takes the empty branch, Angular tears down the server-rendered subtree, and the user sees only `<lfx-header>` on a blank page until a fresh client-side fetch resolves.

Because `debounceTime(0)` uses the async scheduler, even a synchronous cache hit cannot deliver a value before the hydration render pass — so no amount of HTTP-layer caching fixes this on its own.

Everything downstream is serialized behind that refetch: occurrences, registrants, current-user RSVP and attachments all chain off `toObservable(this.meeting)`. They are already parallel with each other, but they can't start until the meeting lands a second time.

**2. Every SSR-rendered API response is fetched twice, app-wide.**

[...]

PR #2046, opened 2026-09-02
perf(meetings): seed join page from SSR on hydration

bramwelt:
## Summary

Fixes the meeting-join page's blank-flash-on-hydration bug — item 4 of #1731, tracked by #2041.

- `meeting()` now seeds synchronously from `TransferState` (via `toSignal`'s `initialValue`) instead of starting `undefined`, so hydration no longer tears down the server-rendered join page and refetches from scratch.
- Adds an `@else` skeleton branch for client-side navigations that have no SSR-seeded state.
- Adds `MeetingJoinPageState` to `packages/shared/src/interfaces/meeting.interface.ts`, carrying `meeting`, `loadedViaPastMeetingId`, and `pastMeetingFullAccess` so the past-meeting branch also restores correctly on hydration.
- Mirrors the existing `PublicProfilePageComponent` TransferState pattern.

This is change 1 of 4 from the approved plan for #1731 item 4; changes 2-4 (server-only `HttpBackend` for the SSR base-URL rewrite, the `withHttpTransferCacheOptions` fix, and SSR cache-control hardening) shipped separately as #2050, #2052, and #2053 — the last of which already landed the `Cache-Control: private, no-store` header this PR originally carried inline, so it's been dropped here (via a merge from `main`) in favor of main's `applySsrCacheHeaders` helper.

## Test plan

- [x] `npx ng test --include="src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts"` — 12/12 pass
- [x] `yarn lint` — 0 errors
- [x] `yarn build` — succeeds
- [x] Post-commit reviewer trio (general, self-serve-conventions, learnings) — clean on all 3 commits
- [x] Full-branch sweep (all 3 reviewers vs origin/main) — clean, 0 critical/important findings


apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -327,7 +347,23 @@ export class MeetingJoinComponent implements OnInit {
   public constructor() {
     // Initialize all class variables
     this.authenticated = this.userService.authenticated;
-    this.meeting = this.initializeMeeting();
+
+    // Seed the client's first paint from the SSR-serialized state (matches the server's resolved
+    // branch, no blank flash on hydration — GH-2041). Null on the server and on client
+    // navigations with no prior SSR state for this component instance.
+    const transferredMeeting = this.transferState.get(this.stateKey, null);
```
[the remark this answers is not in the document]
bramwelt: Fixed in commit 9940104b5. `password` is now set synchronously from the route snapshot in the constructor before `meeting` is exposed, so `initializeSeriesOccurrences()` (which subscribes to `meeting` immediately) reads the correct password on its very first emission instead of racing the debounced pipeline that previously set it.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -711,43 +747,43 @@ export class MeetingJoinComponent implements OnInit {
...
+              project: res.project,
+            })),
+            catchError((error) => {
+              if ([404, 403, 400].includes(error.status)) {
+                this.router.navigate(['/meetings/not-found']);
+              }
+              return EMPTY;
+            })
+          );
+        } else {
```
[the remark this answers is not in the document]
bramwelt: Fixed in commit 9940104b5. The upcoming/default flags (`loadedViaPastMeetingId`, `pastMeetingFullAccess`) are now reset only once the upcoming-meeting lookup actually succeeds (via `tap` on the success path), not eagerly at the start of the branch. Previously, a seeded past-meeting id re-entering this branch on hydration would flip to the upcoming view and tear down the past-meeting's summary/recording/attachments/participant streams before the 404 fallback had a chance to resolve.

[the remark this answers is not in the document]
bramwelt: Same fix as the sibling comment on this branch, fixed in commit 9940104b5 by deferring the flag reset until the upcoming-meeting lookup succeeds.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -771,13 +807,29 @@ export class MeetingJoinComponent implements OnInit {
...
+            pastMeetingFullAccess: this.pastMeetingFullAccess(),
+          });
+        }
+      })
+    );
+
+    // Two overload shapes rather than one `initialValue: seed?.meeting` call — `toSignal`'s
+    // `initialValue` overload requires a value assignable to `T`, not `T | undefined`, so the
+    // branch is on `seed` itself rather than threading the optionality through the option object.
+    return (seed ? toSignal(meeting$, { initialValue: seed.meeting }) : toSignal(meeting$)) as Signal<Meeting & { project: PublicMeetingProject }>;
```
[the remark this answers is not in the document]
bramwelt: Same underlying fix as the sibling comment on `password`, fixed in commit 9940104b5 by setting `password` synchronously in the constructor before `meeting` is exposed.

bramwelt:
## Review Feedback Addressed

Commit: 9940104b5

### Changes Made
- **meeting-join.component.ts**: `password` is now set synchronously from the route snapshot in the constructor, before `meeting` is exposed — fixes a race where `initializeSeriesOccurrences()` (which subscribes to `meeting` immediately) could read `password` before the debounced pipeline set it, sending recurring/password-gated meeting occurrence requests with no password (per copilot[bot] and cursor)
- **meeting-join.component.ts**: in the non-composite branch of `initializeMeeting()`, `loadedViaPastMeetingId`/`pastMeetingFullAccess` are now reset only once the upcoming-meeting lookup succeeds, not eagerly — previously a seeded past-meeting id re-entering this branch on hydration would flip to the upcoming view and tear down the past-meeting's summary/recording/attachments/participant streams before the 404 fallback resolved (per copilot[bot] and cursor)

### Threads Resolved

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -771,13 +819,29 @@ export class MeetingJoinComponent implements OnInit {
...
```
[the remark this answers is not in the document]
bramwelt: Fixed in commit 7fda30190. Added a shared `pastMeetingResourceKey$` helper that dedupes the combined `{ hasAccess, meeting }` stream down to `{ hasAccess, resourceId }` via `distinctUntilChanged` before each pipeline's `switchMap` — so summary, recording, attachments, participants, and transcript no longer cancel/restart when `meeting$` re-emits a new object for the same resource on hydration refetch. The attachments pipeline's explicit refresh trigger still forces a refetch every time: its emissions are tagged with a monotonic tick before being folded into the dedupe key, so a refresh is never treated as a duplicate even when the resource key is unchanged.

bramwelt:

Commit: 7fda30190

### Changes Made
- **meeting-join.component.ts**: added a shared `pastMeetingResourceKey$` helper that dedupes the five past-meeting fan-out pipelines (summary, recording, attachments, participants, transcript) by `{ hasAccess, resourceId }`, so `meeting$` re-emitting a new object for the same resource on hydration refetch no longer cancels/restarts every one of them. The attachments pipeline's explicit refresh trigger is tagged with a monotonic tick before the dedupe check, so it still forces a refetch even when the resource key is unchanged (per copilot[bot])

### Threads Resolved

dealako:
@bramwelt Hi Trevor — thanks for the follow-up commits on this one.

This re-review covers the full branch after `9940104b5` and `7fda30190`. The SSR `TransferState` seeding approach is sound: consume-once on the browser, synchronous password snapshot before subscribers run, deferred upcoming-branch flag reset, and `pastMeetingResourceKey$` dedup all trace cleanly. The prior Copilot and Cursor findings on password races, past-meeting flag teardown, and hydration fan-out restarts look resolved.

**Structured issue count**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: seeded-paint DOM assertion gap; no test for `pastMeetingResourceKey$` dedup/refresh contract
- ⚪ Nit: 4 issues: stale `extra` comment; misleading no-seed test title; `transferredMeeting` naming; unused generic on `pastMeetingResourceKey$`
- ❔ Question: 1 item: should non-navigating fetch errors (e.g. 5xx) render a terminal error instead of an infinite skeleton?

**Revision tracking (re-review)**
- ✅ **Password race before occurrences fetch** — fixed in [`9940104b5`](https://github.com/linuxfoundation/lfx-self-serve/commit/9940104b5b0fb30ec39f899b29f239178a2df41a) (synchronous route snapshot `password.set` in constructor). Agree with Copilot/Cursor on root cause; fix looks correct.
- ✅ **Past-meeting flags reset on hydration** — fixed in [`9940104b5`](https://github.com/linuxfoundation/lfx-self-serve/commit/9940104b5b0fb30ec39f899b29f239178a2df41a) (reset moved to upcoming-success `tap`). Agree with Copilot/Cursor; fix looks correct.
- ✅ **Past-meeting fan-out restarted on hydration refetch** — fixed in [`7fda30190`](https://github.com/linuxfoundation/lfx-self-serve/commit/7fda301900f664aca11bfe787ac6e0daa478b958) (`pastMeetingResourceKey$` dedup + attachments tick). Agree with Copilot; fix looks correct.
- 🟡 **Test coverage for the two fixes above** — still open. Copilot's suppressed comments on the latest push match my assessment: add (a) seeded recurring meeting + `password` query param asserting the first occurrences call receives the password, and (b) a full-access past-meeting test asserting fan-out calls dedupe across hydration re-emission while attachments refresh still re-fires.

**AI comment reconciliation**
- **Agree** with Copilot/Cursor on the three bugs raised in round 1; all addressed in the commits above.
- **Agree** with Copilot's suppressed round-2 suggestions on missing regression tests for password timing and `pastMeetingResourceKey$`; carrying those forward as minors.
- **No disagreement** with bot findings on security/privacy; independent security pass found no new issues in the changed surface.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
```diff
@@ -305,4 +306,85 @@ describe('MeetingJoinComponent', () => {
...
+        pastMeetingFullAccess: false,
+      });
+
+      await TestBed.compileComponents();
+      const fixture = TestBed.createComponent(MeetingJoinComponent);
+      const component = fixture.componentInstance;
+
+      // Assert before any stabilization — the regression this guards against only reproduces if
+      // the seed lands after the first CD pass.
+      expect(component.meeting()?.id).toBe(MEETING_ID);
```
dealako, line 347: [minor] Assert seeded first paint in the DOM, not only the signal

Issue: `paints synchronously from a browser TransferState seed with no blank flash` never calls `fixture.detectChanges()` or queries the template. It only asserts `component.meeting()?.id`.
Proof: The sibling test at :355-363 correctly asserts `[data-testid="meeting-join-skeleton"]` on the no-seed path. Swapping the `@if`/`@else` branches in the template would leave this test green while shipping a skeleton on every hydrated load.
Why it matters: Half of GH-2041's contract lives in the template seam; this test cannot catch a regression there.
Fix: After `createComponent`, call `fixture.detectChanges()` and assert the skeleton is absent before `whenStable()`.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
```diff
@@ -305,4 +306,85 @@ describe('MeetingJoinComponent', () => {
...
+      expect(transferState.get(MEETING_JOIN_STATE_KEY, null)).toBeNull();
+
+      await TestBed.inject(ApplicationRef).whenStable();
+    });
+
+    it('restores the past-meeting branch synchronously from a seeded transfer state', async () => {
```
dealako, line 358: [minor] Cover `pastMeetingResourceKey$` dedup and attachments refresh

Issue: The dedupe helper gates five past-meeting fan-out streams, but no test exercises `hasAccess === true` or observes call counts across hydration re-emission.
Proof: The seeded full-access test asserts flags synchronously while the hydration refetch resets them for non-composite ids; the `MeetingService` mock also omits `getPastMeetingSummary` / recording / attachments / transcript. Copilot raised the same gap in suppressed comments on the latest push.
Why it matters: A wrong `distinctUntilChanged` or tick could silently stop materials refresh after upload/delete with no failing test.
Fix: Extend the mock, drive a composite past-meeting id with `full_access: true`, assert fan-out calls run once across a same-resource re-emission, and assert `onMaterialsChanged` re-issues attachments.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -1265,11 +1330,28 @@ export class MeetingJoinComponent implements OnInit {
     });
   }
 
+  // `meeting` re-emits a new object for the same resource on hydration refetch (the seeded value
+  // is replaced once `meeting$` resolves) — dedupe on the pair that actually determines whether a
+  // past-meeting resource fetch should (re)run, so that refetch doesn't cancel/restart every one
+  // of these fan-out requests for no reason. `extra` carries an explicit refresh trigger through
```
dealako: [nit] Stale `extra` name in `pastMeetingResourceKey$` comment

Issue: The comment refers to `` `extra` `` but the parameter is `refreshTrigger`.
Proof: Line 1338 declares `refreshTrigger: Observable<T>`; `extra` does not appear elsewhere in this helper.
Why it matters: Misleading comment for the trickiest logic in the PR.
Fix: Replace `` `extra` `` with `` `refreshTrigger` `` in the comment.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
```diff
@@ -305,4 +306,85 @@ describe('MeetingJoinComponent', () => {
...
+
+      // `debounceTime(0)` defers even a synchronous mocked response past this first CD pass, so
+      // `meeting()` is still undefined here and the `@else` skeleton branch must be what renders.
+      expect(fixture.componentInstance.meeting()).toBeUndefined();
+      expect(fixture.nativeElement.querySelector('[data-testid="meeting-join-skeleton"]')).not.toBeNull();
+
+      await TestBed.inject(ApplicationRef).whenStable();
+    });
+
+    it('ignores TransferState and starts undefined when the server has not seeded anything', async () => {
```
dealako: [nit] Test title contradicts its assertion

Issue: Title claims TransferState is ignored and `meeting()` starts undefined, but the test only asserts the fetched `MEETING_ID` after synchronous mock resolution.
Proof: The undefined-on-first-render case is already covered at :355-363; this block's comment admits the gap.
Why it matters: Misleading regression documentation.
Fix: Rename to something like `falls back to the fetched meeting when the server has not seeded anything`, and optionally distinguish seed vs fetch ids plus assert `getPublicMeeting` was called.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -854,5 +854,25 @@ <h1>Meeting Guests</h1>
         </lfx-meeting-materials-drawer>
       }
     </div>
+  } @else {
```
dealako, line 880: [question] Terminal state for non-navigating fetch failures

Issue: The new `@else` skeleton has no settled-error branch. In `initializeMeeting`, statuses other than 404/403/400 fall through to `return EMPTY` without navigation, so `meeting()` stays undefined indefinitely.
Proof: `catchError` at meeting-join.component.ts:800-820 returns `EMPTY` for unhandled statuses; the template `@else` is then the only render path.
Why it matters: A 5xx during an upstream outage reads as perpetual loading rather than failure.
Fix: If intentional for this PR, note follow-up; otherwise emit a settled error state and render an error card instead of the skeleton.

dealako (dismissed):
## dd-pr-review (re-review)

Re-reviewed after `9940104b5` and `7fda30190`. Prior blocking correctness issues from Copilot/Cursor are resolved.

### Findings
- 🟡 **Minor (2):** complement the signal-only seed test with a DOM assertion on the `@if (meeting())` / skeleton seam; add regression tests for synchronous password on seeded recurring meetings and for `pastMeetingResourceKey$` dedup vs attachments refresh (Copilot suppressed the same gaps on the latest push).
- ⚪ **Nit (4):** stale `extra` identifier in `pastMeetingResourceKey$` comment; misleading no-seed test title; consider `transferredState` naming parity with public profile; drop unused `<T>` generic.
- ❔ **Question (1):** non-404/403/400 fetch failures leave `meeting()` undefined forever, so the new skeleton reads as perpetual loading — intentional for this PR or follow-up?

### Security / privacy
No new auth, injection, or PII-surface issues in the changed files. `TransferState` carries the same public meeting payload the SSR path already fetched.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
bramwelt: Thanks @dealako, good catch. Added `fixture.detectChanges()` right after the pre-stabilization signal assertion and asserted `[data-testid="meeting-join-skeleton"]` is absent from the DOM — a swapped `@if`/`@else` branch would now fail this test even though `meeting()` itself resolved correctly.

See commit dd935508e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
bramwelt: Thanks @dealako. Extended the `MeetingService` mock with `getPastMeetingSummary`/`getPastMeetingRecording`/`getPastMeetingAttachments`/`getPastMeetingParticipants`/`getPastMeetingTranscript`, and added a test that drives a full-access past-meeting resource (via the composite-id route), asserts all five fan-out fetches run exactly once, stay deduped across a same-resource `meeting$` re-emission (simulating the hydration refetch), and then confirms `onMaterialsChanged()` still forces a fresh attachments fetch (call count goes 1 -> 2) while the other four streams remain untouched.

See commit dd935508e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
bramwelt: Thanks @dealako, good catch — fixed the comment to say `refreshTrigger` instead of the stale `extra`.

See commit dd935508e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
bramwelt: Thanks @dealako. Renamed the test to  and added an assertion that  was actually called, to distinguish the fetch path from a seed. The undefined-on-first-render case you're describing is the separate  test right above it.

See commit dd935508e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
bramwelt: Thanks @dealako, this was a real gap — good question. Rather than leave it purely out of scope, I added a minimal terminal state: a `meetingLoadFailed` signal that's set (and reset on each new route/refresh attempt) when `initializeMeeting()`'s `catchError` branches fall through to `EMPTY` for a status that isn't 404/403/400, plus an `@else if (meetingLoadFailed())` branch in the template with a simple "something went wrong" message (`data-testid="meeting-join-error"`). No new service calls, no navigation change for the existing 404/403/400 paths — those still redirect to `/meetings/not-found` as before.

See commit dd935508e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
bramwelt: Thanks @dealako. Renamed the test to "falls back to the fetched meeting when the server has not seeded anything" and added an assertion that `getPublicMeeting` was actually called, to distinguish the fetch path from a seed. The undefined-on-first-render case you're describing is the separate "shows the skeleton on the initial render..." test right above it.

See commit dd935508e.

bramwelt:
## Review Feedback Addressed

Commit: dd935508e

### Changes Made
- **meeting-join.component.spec.ts**: the seeded first-paint test now calls `fixture.detectChanges()` and asserts `[data-testid="meeting-join-skeleton"]` is absent from the DOM, not just that the `meeting()` signal resolved — a swapped `@if`/`@else` branch would now fail it (per @dealako)
- **meeting-join.component.spec.ts**: extended the `MeetingService` mock with `getPastMeetingSummary`/`getPastMeetingRecording`/`getPastMeetingAttachments`/`getPastMeetingParticipants`/`getPastMeetingTranscript`, and added a new test driving a full-access past-meeting resource that asserts all five `pastMeetingResourceKey$` fan-out fetches run once, stay deduped across a same-resource re-emission, and that `onMaterialsChanged()` still forces a fresh attachments fetch despite the dedup (per @dealako)
- **meeting-join.component.ts**: fixed a stale `extra` reference in the `pastMeetingResourceKey$` comment — the parameter is `refreshTrigger` (per @dealako)
- **meeting-join.component.spec.ts**: renamed a test whose title claimed "TransferState is ignored and meeting() starts undefined" when it actually only asserted the fetched meeting after synchronous mock resolution — retitled to "falls back to the fetched meeting when the server has not seeded anything" and added a `getPublicMeeting` call assertion; the true undefined-on-first-render case is covered by the adjacent skeleton test (per @dealako)
- **meeting-join.component.ts / .html**: added a minimal `meetingLoadFailed` signal and an `@else if` error branch in the template for the case where `initializeMeeting()`'s fetch settles with a non-404/403/400 error and previously fell through to `EMPTY` without navigating, leaving the skeleton spinning forever. No new service calls; the existing 404/403/400 redirect-to-not-found behavior is unchanged (per @dealako)

### Threads Resolved

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -767,17 +824,35 @@ export class MeetingJoinComponent implements OnInit {
               }
               if ([403, 400].includes(error.status)) {
                 this.router.navigate(['/meetings/not-found']);
+              } else {
+                this.meetingLoadFailed.set(true);
```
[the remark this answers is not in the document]
bramwelt: Done — extended `MeetingJoinPageState` to a nullable-meeting + `meetingLoadFailed` shape and added a `setMeetingLoadFailed()` helper that writes the terminal-error branch to `TransferState` from every non-navigating catch block. The constructor now reads it back synchronously alongside the existing `loadedViaPastMeetingId`/`pastMeetingFullAccess` fields, so hydration no longer resets `meetingLoadFailed` to `false` and tears down the SSR-rendered error view.

See commit 46510f9e6.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -767,17 +824,35 @@ export class MeetingJoinComponent implements OnInit {
...
+      tap((res) => {
+        this.project.set(res.project);
+        // Persist the resolved state during SSR so the client can hydrate to the same branch.
+        // Angular defers serialization until this tracked HTTP call settles.
+        if (isPlatformServer(this.platformId)) {
+          this.transferState.set(this.stateKey, {
+            meeting: res,
+            loadedViaPastMeetingId: this.loadedViaPastMeetingId(),
+            pastMeetingFullAccess: this.pastMeetingFullAccess(),
+          });
```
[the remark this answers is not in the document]
bramwelt: Same fix as the sibling thread — the terminal-error branch is now seeded through `TransferState` and read back synchronously, so the client hydrates to the server-rendered error view instead of starting `meetingLoadFailed` at `false`.

See commit 46510f9e6.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -854,5 +854,36 @@ <h1>Meeting Guests</h1>
...
+    <!-- Terminal state for a settled, non-navigating fetch failure (e.g. a 5xx or network error) —
+         404/403/400 already redirect to /meetings/not-found in initializeMeeting(), so this only
+         covers the case that would otherwise leave the skeleton spinning forever (GH-2041). -->
+    <div class="bg-gray-50 min-h-screen flex items-center justify-center" data-testid="meeting-join-error">
+      <div class="text-center flex flex-col items-center gap-2">
+        <i class="fa-light fa-triangle-exclamation text-4xl text-gray-400"></i>
+        <p class="text-lg text-gray-600">Something went wrong loading this meeting.</p>
+        <p class="text-sm text-gray-500">Please refresh the page to try again.</p>
+      </div>
+    </div>
```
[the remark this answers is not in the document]
bramwelt: Fixed — the nested past-meeting fallback's catch block now applies the same `[404, 403, 400].includes(status)` split used by its sibling catch blocks, calling `setMeetingLoadFailed()` for non-terminal statuses instead of unconditionally navigating to not-found.

See commit 46510f9e6.

bramwelt:
## Review Feedback Addressed

Commit: 46510f9e6

### Changes Made
- **packages/shared/src/interfaces/meeting.interface.ts**: extended `MeetingJoinPageState` to a nullable-meeting + `meetingLoadFailed` shape so the terminal-error branch can be seeded, not just the success path (per copilot-pull-request-reviewer[bot] and cursor[bot])
- **meeting-join.component.ts**: added a `setMeetingLoadFailed()` helper that writes the terminal-error branch to `TransferState` from every non-navigating catch block; the constructor reads it back synchronously so hydration no longer resets `meetingLoadFailed` to `false` and tears down the SSR-rendered error view (per copilot-pull-request-reviewer[bot] and cursor[bot])
- **meeting-join.component.ts**: the nested past-meeting fallback's catch block now applies the same `[404, 403, 400]` status split used by its sibling catch blocks, calling `setMeetingLoadFailed()` for non-terminal statuses instead of unconditionally navigating to not-found (per cursor[bot])
- **meeting-join.component.spec.ts**: added 3 tests covering the terminal-error TransferState seed/hydration path and the nested fallback's status split; updated 2 existing seed fixtures for the new interface shape

### Threads Resolved

dealako:
@bramwelt Hi Trevor — really appreciate how quickly you turned around the review feedback in `dd935508e` and `46510f9e6`. The added specs for DOM seeding, `pastMeetingResourceKey$` dedup/refresh, nested past-meeting 5xx handling, and SSR `TransferState` seeding for the terminal-error branch are exactly the kind of regression coverage this page needed.

👏 **Nice work**
- Seeded first paint is now asserted in the DOM (`detectChanges` + no skeleton), not only on the signal.
- The `pastMeetingResourceKey$` test proves fan-out dedup across same-resource re-emission **and** that `onMaterialsChanged` still bumps attachments.
- You closed the bot loop on SSR error hydration by extending `MeetingJoinPageState` and persisting `meetingLoadFailed` server-side.

**Revision tracking (since `7fda30190`)**
- ✅ **DOM assertion for seeded first paint** — [`dd935508e`](https://github.com/linuxfoundation/lfx-self-serve/commit/dd935508e)
- ✅ **`pastMeetingResourceKey$` dedup + attachments refresh tests** — [`dd935508e`](https://github.com/linuxfoundation/lfx-self-serve/commit/dd935508e)
- ✅ **Stale `extra` comment** — [`dd935508e`](https://github.com/linuxfoundation/lfx-self-serve/commit/dd935508e)
- ✅ **Misleading no-seed test title + `getPublicMeeting` assertion** — [`dd935508e`](https://github.com/linuxfoundation/lfx-self-serve/commit/dd935508e)
- ✅ **5xx / nested past-meeting terminal errors + SSR error seed** — [`46510f9e6`](https://github.com/linuxfoundation/lfx-self-serve/commit/46510f9e670bd5a5c3e6a01684c09574987a707a) (also addresses Copilot/Cursor on nested fallback + error `TransferState`)
- ⚠️ **5xx infinite skeleton** — terminal error UI landed, but hydration can still flash skeleton (see blocking #1)
- ❌ **Password-on-seeded-recurring regression test** — still no spec with a `password` query param (Copilot suppressed comment still applies)
- ⚪ **Nits carried** — `transferredMeeting` naming, unused `pastMeetingResourceKey$` generic, duplicate HTTP + `meetingJoinState` SSR payload, consider `lfx-empty-state` for the error branch

**Structured issue count (open)**
- 🔴 Blocking: 2 issues: `meetingLoadFailed` cleared before hydration refetch settles (error seed → skeleton flash); template checks `meeting()` before `meetingLoadFailed()` (stale meeting on failed in-app navigation)
- 🟡 Minor: 2 issues: missing password-on-seed recurring test; seeded-error test should assert post-`whenStable()` DOM (not only first paint)
- ⚪ Nit: 4 issues: `transferredState` rename; drop unused generic; duplicate SSR payload; reuse `lfx-empty-state`

**AI reconciliation**
- **Agree** with Copilot/Cursor on SSR error `TransferState` and nested past-meeting 5xx — fixed in `46510f9e6`.
- **Agree** with Copilot on `meetingLoadFailed` ordering vs `meeting()` and on the early `meetingLoadFailed.set(false)` in the `switchMap` — still open.
- **Agree** with Copilot suppressed password test gap — still open.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -711,43 +762,53 @@ export class MeetingJoinComponent implements OnInit {
...
-            return EMPTY;
-          }
+  private initializeMeeting(seed: MeetingJoinPageState | null) {
+    const meeting$ = combineLatest([this.activatedRoute.paramMap, this.activatedRoute.queryParamMap, this.refreshTrigger$]).pipe(
+      debounceTime(0), // Coalesce rapid SSR hydration emissions so the fallback chain isn't canceled
+      switchMap(([params, queryParams]) => {
+        const meetingId = params.get('id');
+        this.password.set(queryParams.get('password'));
+        // Clear any prior terminal-error state so a route change or retry gets a fresh attempt.
+        this.meetingLoadFailed.set(false);
```
dealako: [blocking] Do not clear `meetingLoadFailed` before the hydration refetch settles

Issue: This `meetingLoadFailed.set(false)` runs at the start of the debounced `switchMap`, after the first CD pass. On a seeded terminal-error hydration (`meeting: null`, `meetingLoadFailed: true`), the constructor paints the error branch, then this line clears the flag while `meeting()` is still undefined, so the template falls through to the skeleton until the client refetch completes.
Proof: Template order is `@if (meeting())` → `@else if (meetingLoadFailed())` → skeleton (`meeting-join.component.html:10, 857, 868`). Seeded-error path uses the no-`initialValue` `toSignal` overload when `seed.meeting` is null (`meeting-join.component.ts:867`). The spec at `:476-495` only asserts before `whenStable()`.
Why it matters: Reintroduces the blank-flash / perpetual-loading symptom GH-2041 fixes, on the error path you just seeded through `TransferState`.
Fix: Remove this reset here; clear `meetingLoadFailed` in the success `tap` (next to `this.project.set(res.project)`) so route changes and successful retries still get a fresh attempt without tearing down a seeded error view.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -854,5 +854,36 @@ <h1>Meeting Guests</h1>
         </lfx-meeting-materials-drawer>
       }
     </div>
+  } @else if (meetingLoadFailed()) {
```
dealako: [blocking] Terminal error branch is unreachable while a prior `meeting()` value sticks

Issue: This `@else if (meetingLoadFailed())` block only renders when `meeting()` is falsy, but `meeting()` is checked first at line 10. After a successful load, `toSignal` keeps the last meeting when a later fetch fails with `EMPTY`, so users can still see the previous meeting with no error UI.
Proof: Failures call `setMeetingLoadFailed()` and return `EMPTY` without emitting a new meeting (`meeting-join.component.ts:792-838`). Copilot flagged the same ordering on the latest push.
Why it matters: Wrong meeting content after a failed transition.
Fix: Move the `meetingLoadFailed()` branch above `@if (meeting())`, or clear/replace the meeting signal when entering terminal-error state.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
```diff
@@ -305,4 +322,189 @@ describe('MeetingJoinComponent', () => {
...
+      await TestBed.inject(ApplicationRef).whenStable();
+
+      const transferState = TestBed.inject(TransferState);
+      const seeded = transferState.get(MEETING_JOIN_STATE_KEY, null);
+      expect(seeded?.meeting).toBeNull();
+      expect(seeded?.meetingLoadFailed).toBe(true);
+      expect(TestBed.inject(Router).navigate).not.toHaveBeenCalledWith(['/meetings/not-found']);
+    });
+
+    it('paints the error branch synchronously from a seeded terminal-error transfer state, with no skeleton flash', async () => {
```
dealako: [minor] Assert post-hydration steady state for seeded terminal error

Issue: This test asserts DOM before `whenStable()` only. The `meetingLoadFailed.set(false)` in the pipeline can regress after the hydration refetch tick without failing the test.
Why it matters: Guards the full GH-2041 contract, not only the first CD pass.
Fix: After `whenStable()`, re-assert `[data-testid="meeting-join-error"]` is present and skeleton absent.

dealako (dismissed):
## dd-pr-review-follow-up

Reviewed new commits `dd935508e` and `46510f9e6` since `7fda30190`. Great progress on tests and SSR error seeding.

### Still open (blocking)
1. **`meetingLoadFailed.set(false)` at the top of the `switchMap`** (inline on line 772) runs after `debounceTime(0)`, so a seeded terminal-error hydration paints the error view, then flips back to the skeleton until the client refetch settles.
2. **Template order** — `@if (meeting())` at line 10 is still evaluated before `@else if (meetingLoadFailed())` at line 857, so a failed in-app navigation can keep showing the previous meeting while the failure flag is set (`toSignal` retains the last emission on `EMPTY`). Reorder or clear the meeting signal on terminal failure.

### Minors / nits
- Add the suppressed Copilot password-on-seeded-recurring regression test.
- Extend the seeded-error DOM test through `whenStable()` for the hydration refetch window.
- Nits: `transferredState` rename, drop unused generic, optional `lfx-empty-state`, duplicate SSR payload follow-up.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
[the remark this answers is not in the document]
bramwelt: Done — reordered the template so `meetingLoadFailed()` is checked ahead of `meeting()`. `toSignal` retains the last successful meeting across a later terminal failure (each failing `catchError` returns `EMPTY`, no new emission), so checking `meeting()` first kept showing stale content instead of the error state. See commit fe9ab146e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -327,7 +352,33 @@ export class MeetingJoinComponent implements OnInit {
...
+      this.transferState.remove(this.stateKey);
+    }
+    if (transferredMeeting) {
+      this.loadedViaPastMeetingId.set(transferredMeeting.loadedViaPastMeetingId);
+      this.pastMeetingFullAccess.set(transferredMeeting.pastMeetingFullAccess);
+      if (transferredMeeting.meetingLoadFailed) {
+        this.meetingLoadFailed.set(true);
+      } else if (transferredMeeting.meeting) {
+        this.project.set(transferredMeeting.meeting.project);
+      }
```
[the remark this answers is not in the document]
bramwelt: Done — moved the `meetingLoadFailed.set(false)` reset out of the top of the debounced `switchMap` and into the success `tap` instead, so it only clears once a fetch actually settles successfully rather than eagerly at every pipeline re-entry (including the hydration-triggered first entry, before the refetch it triggers resolves). See commit fe9ab146e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
bramwelt: Done — same fix as flagged by cursor[bot] on this file: the reset now happens in the success tap after the fetch settles, not eagerly at the top of the switchMap. That eager reset was racing ahead of and clearing a seeded terminal-error hydration state before the refetch it triggers actually resolved. See commit fe9ab146e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
bramwelt: Done — reordered the template chain so `meetingLoadFailed()` is checked first, ahead of `meeting()`. `toSignal` keeps its last-emitted value across a later terminal failure (the `catchError` returns `EMPTY`, so there is no new emission to update the signal), so checking `meeting()` first would keep rendering stale content instead of the error state. See commit fe9ab146e.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
bramwelt: Done — extended the seeded terminal-error test to mock a persistently-failing refetch and re-assert (error view visible, skeleton absent, meetingLoadFailed() still true) after whenStable(), not just on the first synchronous CD pass. See commit fe9ab146e.

bramwelt:
## Review Feedback Addressed

Commit: fe9ab146e9a47b8fc68aa27c512d6a22e65ca08b

### Changes Made
- **meeting-join.component.ts**: moved the `meetingLoadFailed.set(false)` reset out of the top of the debounced pipeline and into the success `tap`, so it only clears once a fetch settles successfully — the eager reset was racing ahead of and undoing a seeded terminal-error hydration before its triggered refetch resolved (per cursor[bot] and @dealako)
- **meeting-join.component.html**: reordered the template so `meetingLoadFailed()` is checked before `meeting()` — `toSignal` retains its last-emitted value across a later terminal failure (`catchError` returns `EMPTY`, no new emission), so checking `meeting()` first kept showing stale content instead of the error state (per copilot-pull-request-reviewer[bot] and @dealako)
- **meeting-join.component.spec.ts**: extended the seeded terminal-error test to mock a persistently-failing refetch and re-assert the error view (and absent skeleton) after `whenStable()`, not just on the first synchronous CD pass (per @dealako)

### Threads Resolved

bramwelt:
## Follow-up: added the previously-flagged regression test

Commit: ee4536497

@dealako's now-dismissed review prose asked for a regression test covering the password-on-seeded-recurring-meeting fix, but it wasn't attached to a trackable review thread, so it fell through the earlier resolve pass. Added it now:

- **meeting-join.component.spec.ts**: new test asserting `getPublicMeetingOccurrences` is called exactly once, with the password, on the very first fetch for a seeded recurring meeting — pinning the `distinctUntilChanged()`-on-series-uid race the constructor fix addresses.

Verified: `yarn lint`, `yarn build`, and the full `meeting-join` spec suite all pass.

dealako:
@bramwelt Hi Trevor — thanks for the quick turnaround on `fe9ab146e` and `ee4536497`. The two blocking items from the last round are implemented cleanly: the eager `meetingLoadFailed` reset is gone from the debounced `switchMap`, and the template now checks the failure branch before `meeting()`. The password-on-seed recurring test and the extended post-hydration error assertions are great additions too.

👏 **Nice work**
- Moving `meetingLoadFailed.set(false)` into the success `tap` (line 850) is exactly the right fix for seeded-error hydration flashes.
- Reordering the template chain (`meetingLoadFailed` → `meeting` → skeleton) correctly prevents stale meeting content from masking a terminal failure on route change.
- The new `sends the correct password on the very first series-occurrences fetch…` spec closes the Copilot gap on the constructor password race.

**Revision tracking (since `46510f9e6`)**
- ✅ **Do not clear `meetingLoadFailed` before hydration refetch settles** — [`fe9ab146e`](https://github.com/linuxfoundation/lfx-self-serve/commit/fe9ab146e)
- ✅ **Check `meetingLoadFailed()` before `meeting()`** — [`fe9ab146e`](https://github.com/linuxfoundation/lfx-self-serve/commit/fe9ab146e)
- ✅ **Password-on-seeded-recurring regression test** — [`ee4536497`](https://github.com/linuxfoundation/lfx-self-serve/commit/ee45364978b7c178d1c909a4855f5c494227a6b8) (assert before full stabilization would pin the race tighter)
- ✅ **Post-`whenStable()` DOM assertion for seeded terminal error** — [`fe9ab146e`](https://github.com/linuxfoundation/lfx-self-serve/commit/fe9ab146e) (see minor on in-flight refetch mock)
- ⚪ **Duplicate SSR payload** — still open (HTTP transfer cache + `meetingJoinState`)
- ⚪ **Nits carried** — `transferredMeeting` naming, unused `pastMeetingResourceKey$` generic, `lfx-empty-state`

**Structured issue count (open)**
- 🔴 Blocking: 1 issue: terminal-error branch takes over on failed *background* refetches (`refreshTrigger$` after registration/materials) while `meeting()` still holds the current meeting — page swaps to error and sticks
- 🟡 Minor: 2 issues: seeded-error hydration test uses sync `throwError` (can't catch eager-clear regression); password test asserts after `whenStable` (pipeline may have set password by then)
- ⚪ Nit: 3 issues: `transferredState` rename; drop unused generic; duplicate SSR payload / optional `lfx-empty-state`

**AI reconciliation**
- **Agree** with Copilot on the new combination risk from checking `meetingLoadFailed()` ahead of `meeting()` without scoping failures to initial/URL-mismatched loads.
- **Note** Copilot's `host_key` in `TransferState` comment — same success-path payload SSR already fetched; no new incremental exposure in these two commits (security pass clean).
- **Agree** with prior Cursor/Copilot threads on the two fixes you landed; this round is about the interaction edge case.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -7,7 +7,22 @@
 <div [ngClass]="{ 'pt-12': userService.impersonating() }">
   <lfx-header></lfx-header>
 
-  @if (meeting()) {
+  @if (meetingLoadFailed()) {
```
dealako: [blocking] Scope terminal-error takeover to loads without a settled meeting

Issue: `meetingLoadFailed()` is now the first template branch, but `setMeetingLoadFailed()` fires on every non-navigating pipeline failure — including background `refreshTrigger$` refetches after registration (`meeting-join.component.ts:667`) or materials changes (`:619`). On a transient 5xx, `meeting()` still holds the working meeting (`toSignal` retains the last emission on `EMPTY`), yet the page swaps to the error screen.
Proof: Template order is `@if (meetingLoadFailed())` then `@else if (meeting())` (`meeting-join.component.html:10, 25`). Failures call `setMeetingLoadFailed()` and return `EMPTY` without clearing `meeting()` (`meeting-join.component.ts:794, 826, 835`). Copilot flagged the same interaction on the latest push.
Why it matters: A blipped refetch after a successful self-registration or materials upload replaces a working join page with a terminal error that only a manual reload clears.
Fix: Only let the error branch take over when there is no settled meeting for the current route (e.g. a computed that requires `!meeting()` or a route-id mismatch), or clear/replace `meeting()` when entering terminal-error for the active URL. Background refresh failures should stay on the current meeting (toast/retry), not flip the whole page.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
```diff
@@ -305,4 +324,236 @@ describe('MeetingJoinComponent', () => {
...
+      expect(seeded?.meeting).toBeNull();
+      expect(seeded?.meetingLoadFailed).toBe(true);
+      expect(TestBed.inject(Router).navigate).not.toHaveBeenCalledWith(['/meetings/not-found']);
+    });
+
+    it('paints the error branch synchronously from a seeded terminal-error transfer state, with no skeleton flash, and holds it through the hydration refetch', async () => {
+      // The hydration refetch also fails, matching the seeded branch — this is what guards the
+      // full GH-2041 contract: the error state must survive past the first CD pass, not just
+      // appear on it and then flash to the skeleton once the debounced pipeline re-enters.
+      getPublicMeeting.mockReturnValue(throwError(() => ({ status: 500 })));
```
dealako: [minor] Keep the hydration refetch in flight so post-hydration assertions are meaningful

Issue: `getPublicMeeting.mockReturnValue(throwError(...))` fails synchronously on subscribe. The eager `meetingLoadFailed.set(false)` (if reintroduced) and `setMeetingLoadFailed()` in `catchError` both run in one synchronous tick, so the flag is `true` again before `whenStable()` — the post-hydration assertions at `:540-546` pass either way.
Proof: Test title claims it "holds through the hydration refetch", but there is no observable in-flight window between clear and re-set.
Why it matters: Cannot regress-guard the GH-2041 eager-clear fix you just landed.
Fix: `getPublicMeeting.mockReturnValue(new Subject().asObservable())` (or similar unsettled observable), assert error DOM after first paint, then error/complete the subject and re-assert steady state.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
```diff
@@ -305,4 +324,236 @@ describe('MeetingJoinComponent', () => {
...
+
+      const transferState = TestBed.inject(TransferState);
+      transferState.set(MEETING_JOIN_STATE_KEY, {
+        meeting: { ...recurringMeeting, project: buildProject() },
+        loadedViaPastMeetingId: false,
+        pastMeetingFullAccess: false,
+        meetingLoadFailed: false,
+      });
+
+      const component = await createComponent();
```
dealako: [minor] Assert password on occurrences before the debounced pipeline runs

Issue: `await createComponent()` waits for `whenStable()`, by which point the debounced `switchMap` has already executed `this.password.set(queryParams.get('password'))` at `meeting-join.component.ts:770` — so the test passes even if the constructor snapshot assignment is removed.
Proof: Assertions at `:409-410` run after full stabilization; the race this spec documents is constructor `password.set` vs `initializeSeriesOccurrences`'s first subscription.
Why it matters: The Copilot-suppressed regression on constructor password ordering is not actually pinned.
Fix: After `TestBed.createComponent`, call one `detectChanges()` and assert `getPublicMeetingOccurrences` was called with `'secret'` **before** `whenStable()`.

dealako (dismissed):
## dd-pr-review-follow-up (round 3)

Reviewed `fe9ab146e` and `ee4536497` since `46510f9e6`. Prior blocking items are fixed — thank you.

### New blocking
**Scope terminal-error takeover to loads without a settled meeting for the current URL.** Checking `meetingLoadFailed()` ahead of `meeting()` is correct for route-change/stale-meeting cases, but the same flag is set on any non-navigating pipeline failure — including `refreshTrigger$` refetches after registration (`component.ts:667`) or materials changes (`component.ts:619`). When that background fetch 5xxs, `meeting()` still holds the working meeting (`toSignal` + `EMPTY`), yet the whole page swaps to the error view with no auto-retry.

### Minors
- Hold the hydration refetch **in flight** in the seeded-error test (`Subject` mock) so post-`whenStable()` assertions can actually catch an eager `meetingLoadFailed` clear.
- Assert the seeded-recurring password **before** `whenStable()` / debounced pipeline runs.

### Nits (carried)
`transferredState` rename, drop unused generic, duplicate SSR payload, optional `lfx-empty-state`.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -759,25 +818,71 @@ export class MeetingJoinComponent implements OnInit {
...
+        this.meetingLoadFailed.set(false);
+        // Persist the resolved state during SSR so the client can hydrate to the same branch.
+        // Angular defers serialization until this tracked HTTP call settles.
+        if (isPlatformServer(this.platformId)) {
+          this.transferState.set(this.stateKey, {
+            meeting: res,
+            loadedViaPastMeetingId: this.loadedViaPastMeetingId(),
+            pastMeetingFullAccess: this.pastMeetingFullAccess(),
+            meetingLoadFailed: false,
+          });
```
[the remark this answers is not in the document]
bramwelt: Fixed in a4d9b9a3d, added `Cache-Control: private, no-store` and `Vary: Cookie` to the SSR HTML response in `server.ts`. The join page can embed per-viewer `TransferState` data (a meeting's `host_key` for organizers, or the full `AuthContext`) so it must never be replayed from a shared/CDN cache to a different visitor.

bramwelt:
## Review Feedback Addressed

Commit: a4d9b9a3d

### Changes Made
- **apps/lfx-one/src/server/server.ts**: set `Cache-Control: private, no-store` and `Vary: Cookie` on the SSR HTML response — this PR's TransferState seed can embed per-viewer data (a meeting's `host_key` for organizers) that must never be replayed from a shared/CDN cache to a different visitor (per copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
bramwelt: Fixed in be9c1097a — scoped the terminal-error branch to `meetingLoadFailed() && !meeting()`. A background `refreshTrigger$` refetch (after registration or a materials change) that hits a transient 5xx now leaves the working meeting page in place instead of taking over with the error screen; the error branch only wins when there's no settled meeting for the current route yet.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
bramwelt: Fixed in be9c1097a — replaced the synchronous `throwError` with a `Subject` held open across the first paint, then errored it explicitly to settle the refetch. There's now a real in-flight window between the seeded first paint and the refetch settling, so the test actually exercises (and would catch a regression in) the eager-clear-before-refetch-settles fix.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.spec.ts
bramwelt: Fixed in be9c1097a — the assertion now runs right after a single `detectChanges()`, before `whenStable()` lets the debounced `meeting$` pipeline re-run and set `password` itself. This pins the actual race (constructor snapshot vs. `initializeSeriesOccurrences()`'s first subscription) instead of passing regardless of it.

apps/lfx-one/src/server/server.ts
```diff
@@ -501,6 +501,12 @@ app.use('/**', async (req: Request, res: Response, next: NextFunction) => {
         return next();
       }
 
+      // SSR HTML can carry per-viewer data serialized into TransferState (e.g. a meeting's
+      // `host_key` for an authorized organizer, or the full `AuthContext` in app.component.ts) —
+      // never let a shared/CDN cache store or replay one visitor's rendered page to another.
+      response.headers.set('Cache-Control', 'private, no-store');
```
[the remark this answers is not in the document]
bramwelt: Good catch — updated the PR description to reflect that the cache-control hardening (item 4 of the #1731 plan) shipped here in a4d9b9a3d rather than in a separate follow-up PR, since this PR is what introduces the per-viewer payload into the live serialization path. Items 2-3 (HttpBackend rewrite, transfer-cache options fix) remain separate follow-ups.

bramwelt:
## Review Feedback Addressed

Commit: be9c1097a

### Changes Made
- **meeting-join.component.html**: scoped the terminal-error takeover to `meetingLoadFailed() && !meeting()` — a background `refreshTrigger$` refetch (after registration or a materials change) hitting a transient 5xx no longer replaces a working join page with the error screen (per @dealako, blocking)
- **meeting-join.component.spec.ts**: held the hydration refetch open with a `Subject` instead of a synchronously-throwing observable, so the "holds through the hydration refetch" test has a real in-flight window and actually guards the eager-clear fix (per @dealako, minor)
- **meeting-join.component.spec.ts**: asserted the password-gated occurrences fetch before `whenStable()` settles the debounced pipeline, pinning the constructor's synchronous password snapshot instead of passing regardless of it (per @dealako, minor)
- **PR description**: updated to note the cache-control hardening (item 4 of the #1731 plan) shipped in a4d9b9a3d rather than deferring to a separate PR, since this PR introduces the per-viewer payload it protects (per copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -7,7 +7,26 @@
 <div [ngClass]="{ 'pt-12': userService.impersonating() }">
   <lfx-header></lfx-header>
 
-  @if (meeting()) {
+  @if (meetingLoadFailed() && !meeting()) {
```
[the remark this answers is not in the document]
bramwelt: Fixed in 221b5cd11: added `meetingRouteId`/`meetingResolvedRouteId` signals so the error branch checks `meetingLoadFailed() && (!meeting() || !meetingMatchesRoute())`. `meetingRouteId` updates synchronously on every route/query-param emission (before the fetch even settles), and `meetingResolvedRouteId` only updates on a settled success, so a same-route background refresh failure still matches (error suppressed, per the original round-2 fix) while a failure after navigating to a different meeting id now mismatches and shows the error instead of the stale previous meeting. Added a dedicated regression test covering the cross-meeting-navigation failure case.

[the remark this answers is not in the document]
bramwelt: Fixed in 221b5cd11 — same fix as the parallel finding from cursor on this line: added `meetingRouteId`/`meetingResolvedRouteId` signals so the error branch can tell whether the retained `meeting()` value still belongs to the current route, distinguishing a same-route background refresh failure (suppress error) from a failure after navigating to a different meeting id (show error, since the retained meeting would otherwise leak the previous meeting's data — including `host_key` — under the new URL). Added a regression test for the cross-meeting-navigation case.

packages/shared/src/interfaces/meeting.interface.ts
```diff
@@ -1377,6 +1377,24 @@ export interface PublicPastMeetingResponse {
   full_access: boolean;
 }
 
+/**
+ * SSR-serialized state for the meeting join page (GH-2041), following the same TransferState
+ * pattern as {@link PublicProfilePageState}: seeds the client's first paint from the server's
+ * resolved branch so hydration doesn't tear down the SSR-rendered meeting and refetch from
+ * scratch. `loadedViaPastMeetingId` / `pastMeetingFullAccess` are carried alongside `meeting`
```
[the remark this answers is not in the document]
bramwelt: Fixed in 221b5cd11 — reworded the `MeetingJoinPageState` doc comment to describe the seed as keeping the DOM stable while hydration's refetch runs, rather than implying the refetch is avoided.

bramwelt:
## Review Feedback Addressed

Commit: 221b5cd11

### Changes Made
- **meeting-join.component.ts/.html**: the round-2 `!meeting()` scoping fix didn't distinguish a same-route background refresh failure from a failure after navigating to a *different* meeting id — `toSignal` retains the last successful `meeting()` across a failing `catchError`'s `EMPTY`, so a cross-route failure kept rendering the previous meeting (incl. its `host_key`/join links) under the new URL. Added `meetingRouteId` / `meetingResolvedRouteId` signals so the error branch (`meetingLoadFailed() && (!meeting() || !meetingMatchesRoute())`) can tell whether the retained meeting still belongs to the current route (per cursor, copilot-pull-request-reviewer)
- **meeting-join.component.spec.ts**: added a regression test covering the cross-meeting-navigation failure scenario
- **meeting.interface.ts**: reworded the `MeetingJoinPageState` doc comment — the seed keeps the DOM stable while hydration's refetch runs, it doesn't avoid that refetch (per copilot-pull-request-reviewer)

### Threads Resolved

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -7,7 +7,30 @@
 <div [ngClass]="{ 'pt-12': userService.impersonating() }">
   <lfx-header></lfx-header>
 
-  @if (meeting()) {
+  @if (meetingLoadFailed() && (!meeting() || !meetingMatchesRoute())) {
```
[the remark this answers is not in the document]
bramwelt: Done — cleared `meetingLoadFailed` synchronously when the route id actually changes (not on same-route re-emissions, so the existing suppression for a same-route background refresh failure still holds). See commit 649fe0a5a, plus a new regression test covering exactly this same-route-failure-then-navigate sequence.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -7,7 +7,30 @@
...
+         this to loads with no settled meeting yet — `refreshTrigger$` re-entering the pipeline
+         after a successful registration or materials change (meeting-join.component.ts:619, 667)
+         can also call `setMeetingLoadFailed()` on a transient failure, but `meeting()` still holds
+         the working meeting then, and a blipped background refetch must not tear down an
+         otherwise-working page (dealako review on #2046). `!meetingMatchesRoute()` additionally
+         catches a failed fetch after navigating to a *different* `/meetings/:id` — without it,
+         `meeting()`'s retained value would keep rendering the PREVIOUS meeting (incl. its
+         `host_key`/join links) under the new URL instead of this error state (cursor / copilot
+         review on #2046). -->
+    <div class="bg-gray-50 min-h-screen flex items-center justify-center" data-testid="meeting-join-error">
```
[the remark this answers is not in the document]
bramwelt: Fixed — added role="alert" to the error container so it's announced as a live region. See commit 649fe0a5a.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -854,5 +877,25 @@ <h1>Meeting Guests</h1>
         </lfx-meeting-materials-drawer>
       }
     </div>
+  } @else {
+    <!-- Skeleton shown only for client-side navigations with no SSR-transferred state (GH-2041) —
+         the seeded first paint in `initializeMeeting()` means this never flashes on a fresh load. -->
+    <div class="bg-gray-50 min-h-screen" data-testid="meeting-join-skeleton">
```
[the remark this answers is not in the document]
bramwelt: Fixed — added role="status" + aria-live="polite" on the skeleton container, plus a visually-hidden "Loading meeting details…" label so assistive tech announces the loading state. See commit 649fe0a5a.

bramwelt:
## Review Feedback Addressed

Commit: 649fe0a5a06fb9d54a76d3e4de3fc4b40f56dd7f

### Changes Made
- **meeting-join.component.ts**: `meetingLoadFailed` is now cleared synchronously only when the route id actually changes, not on every pipeline re-emission. Previously a same-route background failure left the flag set, so navigating to a *different* meeting immediately tripped the error branch via the `!meetingMatchesRoute()` check — before that meeting's own fetch even started (per cursor).
- **meeting-join.component.html**: added `role="alert"` to the error container, and `role="status"` + `aria-live="polite"` + a visually-hidden "Loading meeting details…" label to the client-side skeleton, so both states are announced to assistive tech (per copilot-pull-request-reviewer).
- **meeting-join.component.spec.ts**: added a regression test for the same-route-failure-then-navigate sequence (19 tests passing).

### Threads Resolved

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.html
```diff
@@ -7,7 +7,30 @@
...
+         `host_key`/join links) under the new URL instead of this error state (cursor / copilot
+         review on #2046). -->
+    <div class="bg-gray-50 min-h-screen flex items-center justify-center" role="alert" data-testid="meeting-join-error">
+      <div class="text-center flex flex-col items-center gap-2">
+        <i class="fa-light fa-triangle-exclamation text-4xl text-gray-400"></i>
+        <p class="text-lg text-gray-600">Something went wrong loading this meeting.</p>
+        <p class="text-sm text-gray-500">Please refresh the page to try again.</p>
+      </div>
+    </div>
+  } @else if (meeting()) {
```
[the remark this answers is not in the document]
bramwelt: Done — the success branch template guard now checks `meetingMatchesRoute()` in addition to `meeting()`, so a pending fetch for a newly-navigated meeting shows the skeleton instead of leaking the previous meeting's retained content.

See commit 467272a: meeting-join.component.html line 33.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -327,7 +366,36 @@ export class MeetingJoinComponent implements OnInit {
...
+    // Consume the SSR state exactly once so a later re-creation of this component (e.g. a
+    // different `/meetings/:id`) can't paint the previous meeting's state under the new URL.
+    if (isPlatformBrowser(this.platformId) && transferredMeeting) {
+      this.transferState.remove(this.stateKey);
+    }
+    if (transferredMeeting) {
+      this.loadedViaPastMeetingId.set(transferredMeeting.loadedViaPastMeetingId);
+      this.pastMeetingFullAccess.set(transferredMeeting.pastMeetingFullAccess);
+      if (transferredMeeting.meetingLoadFailed) {
+        this.meetingLoadFailed.set(true);
```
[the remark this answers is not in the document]
bramwelt: Done — the constructor's error-transfer-state branch now seeds `meetingRouteId` too, matching the success branch. Previously only the success branch seeded it, so a same-route hydration refetch after a seeded terminal error was mistaken for a new navigation and cleared `meetingLoadFailed` prematurely.

See commit 467272a: meeting-join.component.ts constructor.

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -327,7 +366,36 @@ export class MeetingJoinComponent implements OnInit {
...
+      } else if (transferredMeeting.meeting) {
```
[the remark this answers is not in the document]
bramwelt: Confirmed and fixed — same root cause as the adjacent copilot comment on this file. The error-seed branch now sets `meetingRouteId` alongside `meetingLoadFailed`, so `meetingMatchesRoute` correctly gates a same-route background refetch failure without treating it as a route change.

See commit 467272a: meeting-join.component.ts constructor. Added regression coverage in meeting-join.component.spec.ts for both this and the adjacent template-guard fix.

bramwelt:
## Review Feedback Addressed

Commit: 467272a53a0cec071a22975af99be555d2f093f3

### Changes Made
- **meeting-join.component.ts**: seeded `meetingRouteId` in the constructor's error-transfer-state branch (previously only the success branch did this) — fixes a same-route hydration refetch after a seeded terminal error being mistaken for a new navigation, which prematurely cleared `meetingLoadFailed` (per copilot-pull-request-reviewer and cursor)
- **meeting-join.component.html**: guarded the template's success branch with `meetingMatchesRoute()` in addition to `meeting()`, so a pending fetch for a newly-navigated meeting shows the skeleton instead of leaking the previous meeting's retained content under the new URL (per copilot-pull-request-reviewer)
- **meeting-join.component.spec.ts**: added two regression tests covering both fixes (21/21 tests passing)

### Threads Resolved

apps/lfx-one/src/app/modules/meetings/meeting-join/meeting-join.component.ts
```diff
@@ -711,43 +786,62 @@ export class MeetingJoinComponent implements OnInit {
...
+        this.meetingRouteId.set(meetingId);
+        // Only clear a stale `meetingLoadFailed` when the route id actually changed (a real
+        // navigation to a different meeting) — not on a same-route re-emission (`refreshTrigger$`,
+        // or the initial hydration matching a seeded terminal-error state). Otherwise a prior
+        // same-route background failure (dealako review on #2046) would still be flagged when
+        // navigating away, and the new route's `!meetingMatchesRoute()` mismatch would trip the
+        // error branch immediately, before this attempt's fetch even starts (cursor review on #2046).
+        if (meetingId !== previousRouteId) {
+          this.meetingLoadFailed.set(false);
+        }
```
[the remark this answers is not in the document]
bramwelt: Confirmed and fixed — moved the `meetingRouteId` bookkeeping into a `tap` ahead of `debounceTime(0)` instead of inside `switchMap`, so `meetingMatchesRoute()` flips false the instant the route change is observed instead of staying true for the debounce's one-macrotask window.

See commit 53b8f50c4: meeting-join.component.ts, `initializeMeeting()`.

bramwelt:
## Review Feedback Addressed

Commit: 53b8f50c48431c7baaa8193ebc3aef2f96177d60

### Changes Made
- **meeting-join.component.ts**: moved the `meetingRouteId` bookkeeping into a `tap` ahead of `debounceTime(0)` instead of inside `switchMap`, so `meetingMatchesRoute()` flips false the instant a route change is observed rather than staying true for the debounce's one-macrotask window, during which the previous meeting's content (including host-key controls) would still render under the new URL (per copilot-pull-request-reviewer)

### Threads Resolved

dealako:
@bramwelt Hi Trevor — excellent work on the follow-up commits since my last review (`ee4536497` → `467272a53`). The route-tracking layer (`meetingRouteId` / `meetingResolvedRouteId` / `meetingMatchesRoute`) cleanly resolves the interaction bugs between terminal-error takeover, background refresh failures, and cross-meeting navigation. The regression suite is now genuinely strong.

👏 **Nice work**
- `meetingLoadFailed() && (!meeting() || !meetingMatchesRoute())` correctly scopes the error branch to initial-load and cross-route failures without tearing down a working page on a blipped `refreshTrigger$` refetch.
- Seeding `meetingRouteId` on the terminal-error transfer path prevents the hydration skeleton flash that Cursor/Copilot flagged.
- The `Subject`-based refetch tests (in-flight window, debounce flush, pending navigation) are exactly the kind of timing-sensitive guards GH-2041 needs.
- `applySsrCacheHeaders(private, no-store)` closes the `host_key`/per-viewer `TransferState` cache concern Copilot raised.
- `role="alert"` / `role="status"` on the error and skeleton branches are good a11y additions.

**Revision tracking (since `ee4536497`)**
- ✅ **Scope terminal-error takeover to loads without a settled meeting** — [`be9c1097a`](https://github.com/linuxfoundation/lfx-self-serve/commit/be9c1097a), refined in [`221b5cd11`](https://github.com/linuxfoundation/lfx-self-serve/commit/221b5cd11) and [`467272a53`](https://github.com/linuxfoundation/lfx-self-serve/commit/467272a53a0cec071a22975af99be555d2f093f3)
- ✅ **Subject mock for seeded-error hydration test** — [`be9c1097a`](https://github.com/linuxfoundation/lfx-self-serve/commit/be9c1097a)
- ✅ **Password assert before debounced pipeline** — [`be9c1097a`](https://github.com/linuxfoundation/lfx-self-serve/commit/be9c1097a)
- ✅ **Stale meeting on failed route switch** — [`221b5cd11`](https://github.com/linuxfoundation/lfx-self-serve/commit/221b5cd11)
- ✅ **Stale failure flag on cross-route navigation** — [`467272a53`](https://github.com/linuxfoundation/lfx-self-serve/commit/467272a53a0cec071a22975af99be555d2f093f3)
- ✅ **Seeded error `meetingRouteId` initialization** — [`467272a53`](https://github.com/linuxfoundation/lfx-self-serve/commit/467272a53a0cec071a22975af99be555d2f093f3)
- ✅ **Pending navigation shows skeleton, not stale meeting** — [`467272a53`](https://github.com/linuxfoundation/lfx-self-serve/commit/467272a53a0cec071a22975af99be555d2f093f3)
- ✅ **SSR `Cache-Control: private, no-store`** — [`a4d9b9a3d`](https://github.com/linuxfoundation/lfx-self-serve/commit/a4d9b9a3d) / [`e61d33814`](https://github.com/linuxfoundation/lfx-self-serve/commit/e61d33814)
- ✅ **A11y live regions on error/skeleton** — [`649fe0a5a`](https://github.com/linuxfoundation/lfx-self-serve/commit/649fe0a5a)
- ⚪ **Duplicate SSR payload (HTTP transfer cache + `meetingJoinState`)** — still open as follow-up
- ⚪ **Nits carried** — `transferredMeeting` → `transferredState`, drop unused `pastMeetingResourceKey$` generic, optional `lfx-empty-state`

**Structured issue count (open)**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 1 issue: duplicate meeting payload in SSR HTML (HTTP transfer cache + `meetingJoinState`) — acceptable follow-up, not a merge blocker
- ⚪ Nit: 3 issues: `transferredState` rename; drop unused generic; consider `lfx-empty-state`

**AI reconciliation**
- **Agree** with Cursor/Copilot on route-mismatch, stale failure flag, seeded-error route-id, and pending-navigation skeleton — all addressed in `221b5cd11` / `467272a53`.
- **Agree** with Copilot on `host_key`/TransferState cache risk — mitigated by `applySsrCacheHeaders`.
- **Note** Copilot's `MeetingJoinPageState` doc wording — fixed in `221b5cd11`.

dealako (approved):
## dd-pr-review-follow-up (round 4)

Reviewed commits since `ee4536497` through `53b8f50`. All prior blocking items are resolved.

### Highlights
- Route-tracking signals (`meetingRouteId`, `meetingResolvedRouteId`, `meetingMatchesRoute`) correctly gate error vs success vs skeleton across hydration, background refresh, and cross-meeting navigation.
- `53b8f50` moves `meetingRouteId` update ahead of `debounceTime(0)`, closing the one-macrotask window where stale meeting content could render under a new URL.
- Regression tests use in-flight `Subject` mocks and pre-`whenStable()` assertions for timing-sensitive paths.
- SSR `Cache-Control: private, no-store` mitigates per-viewer `TransferState` cache risk.

### Carried forward (non-blocking)
- 🟡 Minor: duplicate SSR payload (HTTP transfer cache + `meetingJoinState`) — fine as follow-up.
- ⚪ Nits: `transferredState` rename, drop unused generic, optional `lfx-empty-state`.

PR #2050, opened 2026-09-02
perf(meetings): use HttpBackend for SSR loopback rewrite

bramwelt:
## Summary

- Part of #2041 (item 4 of #1731); change 2 of 4 from the approved plan.
- The old `ssrBaseUrlInterceptor` ran before Angular's transfer-cache interceptor (`HTTP_INTERCEPTOR_FNS` run before `HTTP_ROOT_INTERCEPTOR_FNS`), so the cache stored the rewritten absolute loopback URL while the browser looked up the original relative path — a guaranteed cache miss and duplicate fetch for every SSR-rendered API call, app-wide.
- Moves the rewrite to a server-only `HttpBackend` override (`SsrBaseUrlBackend`), registered in `app.config.server.ts` only. `HttpBackend` sits at the terminal dispatch stage, strictly after every interceptor (including the transfer-cache interceptor and `authenticationInterceptor`) has already seen and recorded the original relative URL — so cache keys stay symmetric between server and browser.
- As a side effect, this also removes an interceptor-ordering hazard: `authenticationInterceptor` now always sees the original relative `/api/...`/`/public/api/...` URL regardless of provider array order (previously documented as the root cause of #632).
- Deletes `ssrBaseUrlInterceptor` and drops it from `app.config.ts`'s interceptor chain.

## Test plan

- [x] Unit tests: `ssr-base-url.backend.spec.ts` — 4/4 pass (PORT-set rewrite for both `/api/` and `/public/api/`, PORT-unset fallback to 4000, passthrough for other URLs)
- [x] `yarn lint` — 0 errors (1 pre-existing unrelated warning)
- [x] `yarn build` — succeeds, including SSR bundle
- [x] Post-commit reviewer trio (general, self-serve-conventions, learnings) — all clean, no findings
- [x] DCO sign-off + GPG signature verified


apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  let handle: ReturnType<typeof vi.fn>;
+  let backend: SsrBaseUrlBackend;
+  let originalPort: string | undefined;
+
+  beforeEach(() => {
+    originalPort = process.env['PORT'];
+    handle = vi.fn().mockReturnValue(of({}));
+
+    TestBed.configureTestingModule({
+      providers: [SsrBaseUrlBackend, { provide: FetchBackend, useValue: { handle } }],
```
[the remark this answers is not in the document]
bramwelt: Added regression cases in commit b6a9f52d5 covering an already-absolutized `req.url` (both the `/api/` rewrite and the untouched pass-through path), which is the exact input shape a full HttpClient + server-providers integration harness would exercise. Kept it at the existing direct-`handle()` unit-test level used by the rest of this spec rather than adding a new integration harness, since these cases pin the specific fix.

No code change needed beyond the new tests, the existing unit-test convention already covers the regression this input shape exercises.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts
```diff
@@ -0,0 +1,33 @@
...
+  private readonly fetchBackend = inject(FetchBackend);
+
+  public handle(req: HttpRequest<unknown>): Observable<HttpEvent<unknown>> {
+    if (!req.url.startsWith('/api/') && !req.url.startsWith('/public/api/')) {
+      return this.fetchBackend.handle(req);
+    }
+
+    const port = process.env['PORT'] || '4000';
+    const internalBase = `http://127.0.0.1:${port}`;
+    return this.fetchBackend.handle(req.clone({ url: `${internalBase}${req.url}` }));
```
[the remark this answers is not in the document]
bramwelt: Confirmed and fixed in commit b6a9f52d5. Verified against `@angular/platform-server`'s `relativeUrlsTransformerInterceptorFn`: it's registered as an `HTTP_ROOT_INTERCEPTOR_FNS` by `provideServerRendering`, appended after the transfer-cache interceptor via `mergeApplicationConfig`, so it runs before `HttpBackend` and absolutizes the request to the public origin first. `SsrBaseUrlBackend` now parses `req.url` and matches on `pathname` instead of assuming it's still relative.

Good catch, this would have silently defeated the loopback rewrite for every SSR API call.

bramwelt:
## Review Feedback Addressed

Commit: b6a9f52d5

### Changes Made
- **ssr-base-url.backend.ts**: matches on the parsed `pathname`/`search` instead of a raw `req.url` prefix — `relativeUrlsTransformerInterceptorFn` (an `HTTP_ROOT_INTERCEPTOR_FNS` registered by `provideServerRendering`) absolutizes the request to the public origin before it reaches this backend, so the previous string-prefix check silently never matched on the server (per cursor)

### No Change Needed
- **ssr-base-url.backend.spec.ts**: added regression unit tests for the absolutized-URL case at the existing direct-`handle()` test level, rather than a new full HttpClient + server-providers integration harness — these pin the specific fix without disproportionate new test infra (per copilot[bot])

### Threads Resolved

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts
```diff
@@ -0,0 +1,43 @@
...
+export class SsrBaseUrlBackend implements HttpBackend {
+  private readonly fetchBackend = inject(FetchBackend);
+
+  public handle(req: HttpRequest<unknown>): Observable<HttpEvent<unknown>> {
+    // `relativeUrlsTransformerInterceptorFn` (an HTTP_ROOT_INTERCEPTOR_FNS registered by
+    // provideServerRendering) already absolutizes the request to the public origin before it
+    // reaches this backend, so `req.url` is no longer guaranteed to be relative — match on the
+    // parsed pathname instead of a raw string prefix.
+    const { pathname, search } = new URL(req.url, 'resolve://');
+    if (!pathname.startsWith('/api/') && !pathname.startsWith('/public/api/')) {
```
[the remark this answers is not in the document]
bramwelt: Fixed in commit cf872cdb9. The backend now checks whether the request's origin matches this app's own origin (mirroring the `urlPrefix` `relativeUrlsTransformerInterceptorFn` builds from `PlatformLocation`) before rewriting — a relative URL or one already targeting our own origin is treated as ours, but an absolute URL to a different origin (e.g. `https://vendor.example/api/resource`) is passed through untouched even though its path starts with `/api/`. Added a regression test for that case.

Good catch, that would have silently retargeted third-party requests to the local BFF.

bramwelt:

Commit: cf872cdb9

### Changes Made
- **ssr-base-url.backend.ts**: only rewrite requests that are relative or already absolutized to this app's own origin; an absolute URL to a different origin now passes through untouched, even when its path starts with `/api/` (per copilot[bot])
- **ssr-base-url.backend.spec.ts**: added a regression case for a third-party absolute URL with an `/api/`-prefixed path

### Threads Resolved

dealako:
@bramwelt Hi Trevor,

This is a focused, well-motivated change. Moving the loopback rewrite from an `HttpInterceptorFn` to a terminal `HttpBackend` is the right fix for the transfer-cache key mismatch, and the follow-up commits cleanly addressed the two real bugs Cursor and Copilot flagged (absolutized URL matching and third-party origin passthrough). The unit tests pin those regressions directly, and I agree that a full HttpClient + server-providers integration harness would be disproportionate for this PR.

**Revision history (prior review rounds)**

| Prior finding | Status | Resolution |
|---|---|---|
| Cursor Bugbot: loopback rewrite skipped absolutized SSR URLs | **Resolved** | `b6a9f52` — pathname matching via `new URL(req.url, 'resolve://')` |
| Copilot: third-party absolute `/api/` URLs retargeted to local BFF | **Resolved** | `cf872cdb` — origin guard + regression test |
| Copilot: integration-level transfer-cache test | **Acknowledged, not adopted** | Author rationale: direct `handle()` tests pin the rewrite; integration harness is out of proportion for this change. I agree for merge purposes. |

**Bot reconciliation**

- **Agree** with Cursor and Copilot on the two blocking bugs above; both are fixed in the current HEAD.
- **Agree** with Copilot's integration-test suggestion in principle, but **concur with your scope decision** — the seven unit tests cover rewrite, passthrough, PORT fallback, absolutized-origin input, and third-party passthrough.

**Structured issue count**

- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: dev-mode false "not using fetch" SSR warning (`SsrBaseUrlBackend` should extend `FetchBackend`); missing test for non-default-port `getOwnOrigin()` branch
- ⚪ Nit: 3 issues: stale JSDoc lead sentence still says "relative" only; duplicated inline comment restating class doc; stale KB entry for deleted interceptor ordering
- ❔ Question: 0 items

The two minor items are polish and test coverage, not regressions. Happy to merge once you have a chance to look at them, but they do not block approval from my side.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * `provideServerRendering`'s `relativeUrlsTransformerInterceptorFn` is also an
+ * `HTTP_ROOT_INTERCEPTOR_FNS`, registered after the transfer cache, so by the time a request
+ * reaches this backend on the server it has already been absolutized to the public origin — hence
+ * matching on the parsed pathname below rather than assuming `req.url` is still relative. A URL
+ * that was already absolute to a *different* origin (a third-party API called directly by app
+ * code) must be left alone rather than pattern-matched on path alone.
+ */
+@Injectable()
+export class SsrBaseUrlBackend implements HttpBackend {
```
dealako: [minor] Extend FetchBackend to silence dev-mode SSR warning

Issue: `SsrBaseUrlBackend` implements `HttpBackend` and delegates to an injected `FetchBackend`, so the token-bound backend is no longer a `FetchBackend` instance.
Proof: Angular's `HttpInterceptorHandler` warns in dev mode when `!(this.backend instanceof FetchBackend)` (`NOT_USING_FETCH_BACKEND_IN_SSR`). `app.config.server.ts` rebinds `HttpBackend` to `SsrBaseUrlBackend`, triggering that warning on every SSR boot even though `withFetch()` is enabled.
Why it matters: A false alarm on every dev SSR run trains people to ignore a warning that would matter if fetch were actually disabled. Production behavior is unaffected.
Fix: `export class SsrBaseUrlBackend extends FetchBackend`, drop the injected delegate, mark `handle` as `override`, and call `super.handle(...)` instead of `this.fetchBackend.handle(...)`.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  beforeEach(() => {
+    originalPort = process.env['PORT'];
+    handle = vi.fn().mockReturnValue(of({}));
+
+    TestBed.configureTestingModule({
+      providers: [
+        SsrBaseUrlBackend,
+        { provide: FetchBackend, useValue: { handle } },
+        { provide: PlatformLocation, useValue: { protocol: 'https:', hostname: 'lfx.example.org', port: '' } },
```
dealako: [minor] Cover non-default-port getOwnOrigin branch

Issue: Every test stub sets `PlatformLocation.port` to `''`, so the `port ? ... : ...` ternary in `getOwnOrigin()` only exercises the no-port branch.
Proof: `beforeEach` provider at line 25 uses `{ protocol: 'https:', hostname: 'lfx.example.org', port: '' }`. Nothing uses a non-empty port such as `'4000'`.
Why it matters: Non-default-port SSR environments (local dev, forwarded port headers) take the untested branch. A formatting bug there would make `parsed.origin !== this.getOwnOrigin()` permanently true, silently skipping the loopback rewrite with no test failure.
Fix: Add a case with `{ protocol: 'http:', hostname: 'localhost', port: '4000' }` and assert `http://localhost:4000/api/...` rewrites to `http://127.0.0.1:$PORT/api/...`.

dealako (dismissed):
## Review summary

Approved with minor comments. The HttpBackend swap correctly fixes the SSR transfer-cache key mismatch and the two prior bot findings (absolutized URL matching, third-party origin passthrough) are resolved in `b6a9f52` and `cf872cdb`.

### Minor (non-blocking)

1. **Extend `FetchBackend`** — see inline comment on `ssr-base-url.backend.ts`.
2. **Test the ported `getOwnOrigin()` branch** — see inline comment on `ssr-base-url.backend.spec.ts`.

### Nit

- JSDoc lead sentence (line 10) still says "Rewrites relative …" but the implementation intentionally handles absolutized same-origin URLs too.
- Inline comment at lines 34–37 duplicates the class JSDoc paragraph above it.
- `docs/reviews/knowledge-base/server-request-handling.md` still references `ssrBaseUrlInterceptor` ordering; worth a follow-up KB refresh.

See the summary comment for revision tracking and bot reconciliation.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts
bramwelt: Confirmed and fixed in commit f117d502d. Verified against this repo's shipped `@angular/common` (`HttpInterceptorHandler` in `fesm2022/module.mjs`): it does warn with `NOT_USING_FETCH_BACKEND_IN_SSR` when `!(this.backend instanceof FetchBackend)` under `ngServerMode && ngDevMode`. Also checked `FetchBackend`'s Ivy factory (`deps: []`, all deps injected via field initializers), so extending it with no explicit constructor is safe and `super.handle(...)` is behaviorally identical to the prior `this.fetchBackend.handle(...)` delegation. `SsrBaseUrlBackend` now `extends FetchBackend` and calls `super.handle(...)`; updated the JSDoc and the spec's mocking strategy (spy on `FetchBackend.prototype.handle` instead of providing a mock instance) accordingly. Thanks @dealako, good catch.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.spec.ts
bramwelt: Fixed in commit f117d502d. Added a regression case using `{ protocol: 'http:', hostname: 'localhost', port: '4000' }`, asserting a request to `http://localhost:4000/api/meetings/1` still gets rewritten to the loopback address, exercising the with-port branch of `getOwnOrigin()`. Thanks @dealako, good catch on the coverage gap.

bramwelt:
## Review Feedback Addressed

Commit: f117d502d20eb52ff749dc279de8580025a017c2

### Changes Made
- **apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts**: `SsrBaseUrlBackend` now extends `FetchBackend` and calls `super.handle(...)` instead of implementing `HttpBackend` and delegating to an injected `FetchBackend` instance (per @dealako). Independently verified against this repo's shipped `@angular/common` source (`HttpInterceptorHandler` in `fesm2022/module.mjs`): the dev-mode `instanceof FetchBackend` check and `NOT_USING_FETCH_BACKEND_IN_SSR` warning are real, and `FetchBackend`'s Ivy factory has no constructor deps (`deps: []`, all field-injected), so extending it is safe and behaviorally identical.
- **apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.spec.ts**: Updated the test harness to spy on `FetchBackend.prototype.handle` (since `handle` is no longer resolved through an injected instance), and added a regression case covering the with-port branch of `getOwnOrigin()` (`port: '4000'`), asserting the loopback rewrite still applies (per @dealako).

### Threads Resolved

dealako:
@bramwelt Thanks for the quick turnaround on `f117d502d` — both actionable items from the prior round are fully closed.

👏 **Nice work**

- **`extends FetchBackend`** is the right fix: `super.handle(...)` keeps the rewrite logic identical while satisfying Angular's dev-mode `instanceof FetchBackend` check. The new JSDoc paragraph (lines 29–33) documents the *why*, which will save the next reader a trip through `HttpInterceptorHandler`.
- **Non-default-port coverage** lands exactly where it needed to: the new spec case with `{ protocol: 'http:', hostname: 'localhost', port: '4000' }` exercises the `getOwnOrigin()` branch that real local SSR uses, and switching the spy to `FetchBackend.prototype.handle` is the correct harness now that the class subclasses rather than delegates.

**Revision tracking (prior round → this commit)**

| Prior item | Verdict | Notes |
|---|---|---|
| Extend `FetchBackend` to silence `NOT_USING_FETCH_BACKEND_IN_SSR` | ✅ **Resolved** | `f117d502d` — class extends `FetchBackend`, all paths call `super.handle(...)` |
| Test non-default-port `getOwnOrigin()` branch | ✅ **Resolved** | `f117d502d` — new case at spec lines 100–112 |
| JSDoc lead still says "Rewrites relative …" | ⚠️ **Partially addressed** | Tail updated for `super.handle`; lead qualifier still contradicts paragraph three |
| Duplicated inline comment in `handle()` | ❌ **Still open** (nit) | Lines 40–43 still mirror the class JSDoc |
| Stale KB `ssrBaseUrlInterceptor` entry | ❌ **Still open** (nit) | Out of this PR's diff; follow-up doc refresh |
| `CLAUDE.md` missing `shared/backends/` | ❌ **Still open** (nit) | Out of this PR's diff; one-line tree add |

**New commits scanned:** `cf872cdb` → `f117d502d` (2 files, +38/−18). No new security, correctness, or privacy issues in the delta.

**Structured issue count (open items)**

- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 3 issues: stale "relative" JSDoc qualifier; duplicated inline comment; KB + CLAUDE.md doc-map follow-ups

✅ **Approved**

From my side this is ready to merge. The remaining nits are documentation polish and can ride in a follow-up if you prefer to keep this PR tight.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { PlatformLocation } from '@angular/common';
+import { FetchBackend, HttpEvent, HttpRequest } from '@angular/common/http';
+import { inject, Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+
+/**
+ * Server-only `HttpBackend` override. Rewrites relative `/api/*` and `/public/api/*` URLs to
```
dealako: [nit] Drop the stale "relative" qualifier in the lead sentence

Issue: Line 10 still opens with "Rewrites relative `/api/*` …" but paragraph three (lines 22–27) and the spec case at lines 74–81 both document that SSR requests are often already absolutized to the public origin before this backend runs.
Proof: `handle()` matches on `parsed.pathname` after an origin check precisely because `relativeUrlsTransformerInterceptorFn` runs first; the new test at spec line 107 sends `http://localhost:4000/api/meetings/1`.
Why it matters: The lead sentence is the first thing a maintainer reads; keeping "relative" invites a future revert to string-prefix matching.
Fix: e.g. "Rewrites same-origin `/api/*` and `/public/api/*` requests (relative or already absolutized) to …"

dealako (dismissed):
## Follow-up review (`cf872cdb` → `f117d502d`)

✅ **Approved** — both minor items from the prior round are resolved.

### Resolved

1. **`extends FetchBackend`** — dev-mode `NOT_USING_FETCH_BACKEND_IN_SSR` false alarm eliminated; `super.handle(...)` preserves behavior.
2. **Non-default-port test** — `localhost:4000` origin case pins the ported `getOwnOrigin()` branch.

### Remaining nits (non-blocking)

- JSDoc lead line 10 still says "Rewrites relative …" while the implementation and tests intentionally handle already-absolutized same-origin URLs.
- Inline comment at lines 40–43 duplicates the class JSDoc.
- KB + `CLAUDE.md` structure-map updates for the interceptor → backend migration (out of this PR's diff).

No new issues in the delta. LGTM.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts
bramwelt: Done — dropped the stale "relative" qualifier from the lead sentence: it now reads "Rewrites same-origin `/api/*` and `/public/api/*` requests (relative or already absolutized) to …", matching your suggested wording. See commit 2eca0ad7e.

bramwelt:
## Review Feedback Addressed

Commit: 2eca0ad7e

### Changes Made
- **ssr-base-url.backend.ts**: dropped the stale "relative" qualifier from the doc comment's lead sentence — now reads "Rewrites same-origin `/api/*` and `/public/api/*` requests (relative or already absolutized) to …", matching the suggested wording (per @dealako)

### Threads Resolved

dealako:
@bramwelt Thanks for closing out the JSDoc nit in `2eca0ad7e` — that was the last wording fix tied to the absolutized-URL behavior.

👏 **Nice work**

- The updated lead sentence ("same-origin … relative or already absolutized") now matches `handle()`, the third JSDoc paragraph, and every spec case. No more contradiction between the summary line and the implementation.

**Revision tracking (prior round → this commit)**

| Prior item | Verdict | Notes |
|---|---|---|
| JSDoc lead "Rewrites relative …" qualifier | ✅ **Resolved** | `2eca0ad7e` — lead now reads "same-origin … (relative or already absolutized)" |
| Duplicated inline comment in `handle()` | ❌ **Still open** (nit) | Lines 40–43 still mirror the class JSDoc |
| `expect(backend).toBeInstanceOf(FetchBackend)` in spec | ❌ **Still open** (nit) | Optional guard against reverting to composition |
| Stale KB `ssrBaseUrlInterceptor` entry | ❌ **Still open** (nit) | Out of this PR's diff |
| `CLAUDE.md` missing `shared/backends/` | ❌ **Still open** (nit) | Out of this PR's diff |

**New commits scanned:** `f117d502` → `2eca0ad7e` (1 file, +4/−4, comment-only). No new security, correctness, or privacy issues.

**Structured issue count (open items)**

- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 4 issues: duplicated inline comment; optional `instanceof` spec assertion; KB refresh; CLAUDE.md tree row

✅ **Approved**

All actionable feedback from the review rounds is addressed. The remaining nits are documentation polish and can ride in a follow-up PR if you want to keep this one tight — they do not block merge from my side.

apps/lfx-one/src/app/shared/backends/ssr-base-url.backend.ts
```diff
@@ -0,0 +1,67 @@
...
+ * `app.config.server.ts` still passes Angular's `instanceof FetchBackend` check — otherwise
+ * `HttpInterceptorHandler` logs a spurious `NOT_USING_FETCH_BACKEND_IN_SSR` dev-mode warning on
+ * every SSR boot even though `withFetch()` is enabled.
+ */
+@Injectable()
+export class SsrBaseUrlBackend extends FetchBackend {
+  private readonly platformLocation = inject(PlatformLocation);
+
+  public override handle(req: HttpRequest<unknown>): Observable<HttpEvent<unknown>> {
+    // `relativeUrlsTransformerInterceptorFn` (an HTTP_ROOT_INTERCEPTOR_FNS registered by
```
dealako, line 40: [nit] Drop the inline comment that duplicates the class JSDoc

Issue: Lines 40–43 restate paragraph three (lines 22–27) nearly verbatim — both explain that `relativeUrlsTransformerInterceptorFn` absolutizes requests before this backend runs, so matching uses `parsed.pathname`.
Proof: The JSDoc was just updated in `2eca0ad7e` but this inline copy was left untouched; two copies of the same rationale will drift on future edits.
Why it matters: Documentation drift is how the stale "relative only" lead sentence happened in the first place.
Fix: Delete lines 40–43 and let the class JSDoc carry the explanation, or reduce to a one-line pointer: `// req.url may already be absolute — see class doc.`

dealako (approved):
## Follow-up review (`f117d502` → `2eca0ad7e`)

✅ **Approved** — the JSDoc nit is resolved; no new issues in the delta.

### Resolved this round

- **JSDoc lead sentence** — "same-origin … (relative or already absolutized)" now matches `handle()` and the spec.

### Remaining nits (non-blocking, carry-over)

- Inline comment at lines 40–43 duplicates the class JSDoc.
- Optional `expect(backend).toBeInstanceOf(FetchBackend)` in the spec.
- KB + `CLAUDE.md` structure-map updates for the interceptor → backend migration.

Comment-only diff; LGTM.

Issue #2049, opened 2026-09-02
Add Cache-Control/Vary headers to SSR HTML response


bramwelt:
## Problem

Part of #2041 (item 4 of #1731). `apps/lfx-one/src/server/server.ts` sets no `Cache-Control` on the SSR HTML response — the only `setHeaders` block is on the `express.static` handler, and Angular adds none for the rendered document.

The meeting-join payload can carry `host_key` / `can_view_host_key` (see `hostKeyVisible` in `meeting-join.component.ts`), and `app.component.ts` already serializes the full `AuthContext` into TransferState. An accidental shared/intermediary cache hit on the rendered HTML would leak this per-user data to a different user.

## Possible solution

Add `Cache-Control: private, no-store` and `Vary: Cookie` to the SSR HTML response, in the `.then((response) => ...)` block in `server.ts` before `writeResponseToNodeResponse`.

Refs #2041, #1731

bramwelt:
Shipped in #2053 (merged 2026-09-03, released in v1.0.202).

New helper `apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.ts` sets `Cache-Control: private, no-store` on the SSR HTML response via `applySsrCacheHeaders`, called from `server.ts` before `writeResponseToNodeResponse`.

One deliberate deviation from the issue's proposed fix: **no `Vary: Cookie` header was added.** The helper's doc comment explains why — `no-store` alone is sufficient because conformant caches never persist the response at all, so there's nothing left to partition with `Vary`; and setting `Vary` via `.set()` risked clobbering a header Angular's own SSR redirect handling may have already set (e.g. an i18n redirect's `Vary: Accept-Language`). Flagging this so it's a recorded decision, not a silent gap, in case a future change reintroduces a shared-cache path where `Vary` would matter again.

Closing as complete.

PR #2053, opened 2026-09-02
fix(server): set private no-store on SSR HTML response

bramwelt:
## Summary

- Part of #2041 (item 4 of #1731); change 4 of 4 from the approved plan.
- `apps/lfx-one/src/server/server.ts` set no `Cache-Control` on the SSR HTML response. The rendered HTML can carry per-user data — the full `AuthContext` (already serialized into TransferState) and, on meeting-join pages, `host_key`/`can_view_host_key` — so an accidental shared/intermediary cache hit would leak one user's page to another.
- Adds `Cache-Control: private, no-store` and `Vary: Cookie` to the SSR response, set right after the render resolves and before the 404-rebuild branch, so both the normal and rebuilt-404 responses inherit it.

## Note: protected file

`server.ts` is on the repo's protected-files list (server bootstrap file — affects application startup). Flagging per `.claude/hooks/guard-protected-files.sh` for code-owner review; the change itself is a small additive header-setting block with no control-flow changes.

## Test plan

- [x] `yarn lint` — 0 errors (1 pre-existing unrelated warning)
- [x] `yarn build` — succeeds, including SSR bundle
- [x] Post-commit reviewer trio (general, self-serve-conventions, learnings) — clean; conventions reviewer flagged the protected-file NIT (addressed above, no code change needed)
- [x] DCO sign-off + GPG signature verified

### DIFF
--- a/apps/lfx-one/src/server/server.ts
+++ b/apps/lfx-one/src/server/server.ts
@@ -501,6 +501,11 @@ app.use('/**', async (req: Request, res: Response, next: NextFunction) => {
         return next();
       }
 
+      // The rendered HTML can carry per-user data (auth context, host_key/can_view_host_key on
+      // meeting pages), so it must never be served from a shared/intermediary cache to another user.
+      response.headers.set('Cache-Control', 'private, no-store');
+      response.headers.set('Vary', 'Cookie');
+
       // Web `Response.status` is read-only, so rebuild with 404 when the render flagged not-found.
       // Buffer the body first (404 pages are small) so we never hand a consumed stream to the new Response.
       if (renderContext.notFound && response.status === 200) {
--- a/apps/lfx-one/src/server/server.ts
+++ b/apps/lfx-one/src/server/server.ts
@@ -17,6 +17,7 @@ import { CrowdfundingController } from './controllers/crowdfunding.controller';
 import { ProfileController } from './controllers/profile.controller';
 import { CrowdfundingAuthService } from './services/crowdfunding-auth.service';
 import { customErrorSerializer } from './helpers/error-serializer';
+import { applySsrCacheHeaders } from './helpers/ssr-cache-headers.helper';
 import { validateAndSanitizeUrl } from './helpers/url-validation';
 import { AuthenticationError } from './errors';
 import { authMiddleware } from './middleware/auth.middleware';
@@ -501,10 +502,7 @@ app.use('/**', async (req: Request, res: Response, next: NextFunction) => {
         return next();
       }
 
-      // The rendered HTML can carry per-user data (auth context, host_key/can_view_host_key on
-      // meeting pages), so it must never be served from a shared/intermediary cache to another user.
-      response.headers.set('Cache-Control', 'private, no-store');
-      response.headers.set('Vary', 'Cookie');
+      applySsrCacheHeaders(response);
 
       // Web `Response.status` is read-only, so rebuild with 404 when the render flagged not-found.
       // Buffer the body first (404 pages are small) so we never hand a consumed stream to the new Response.
--- a/apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.spec.ts
+++ b/apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.spec.ts
@@ -1,0 +1,35 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { describe, expect, it } from 'vitest';
+
+import { applySsrCacheHeaders } from './ssr-cache-headers.helper';
+
+describe('applySsrCacheHeaders', () => {
+  it('sets private, no-store on a normal 200 SSR render', () => {
+    const response = new Response('<html></html>', { status: 200, headers: { 'Content-Type': 'text/html' } });
+
+    applySsrCacheHeaders(response);
+
+    expect(response.headers.get('Cache-Control')).toBe('private, no-store');
+    expect(response.headers.get('Content-Type')).toBe('text/html');
+  });
+
+  it('sets private, no-store on a redirect response without dropping its existing headers', () => {
+    const response = new Response(null, { status: 302, headers: { Location: '/en-US/', Vary: 'Accept-Language' } });
+
+    applySsrCacheHeaders(response);
+
+    expect(response.headers.get('Cache-Control')).toBe('private, no-store');
+    expect(response.headers.get('Location')).toBe('/en-US/');
+    expect(response.headers.get('Vary')).toBe('Accept-Language');
+  });
+
+  it('overwrites a pre-existing Cache-Control value rather than appending to it', () => {
+    const response = new Response(null, { status: 200, headers: { 'Cache-Control': 'public, max-age=3600' } });
+
+    applySsrCacheHeaders(response);
+
+    expect(response.headers.get('Cache-Control')).toBe('private, no-store');
+  });
+});
--- a/apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.ts
+++ b/apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.ts
@@ -1,0 +1,18 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+/**
+ * Applies the cache-control policy required on every SSR HTML response.
+ *
+ * The rendered HTML can carry per-user data (auth context, host_key/can_view_host_key on
+ * meeting pages), so it must never be served from a shared/intermediary cache to another
+ * user. `no-store` alone is sufficient: conformant caches never persist the response, so
+ * there is nothing left to partition with a `Vary` header, and a `Vary` value would risk
+ * clobbering a header Angular's own SSR redirect handling may have already set (e.g. an
+ * i18n locale redirect's `Vary: Accept-Language`) if this were ever set with `.set()`
+ * instead of `.append()`. Mutates `response.headers` in place — the Fetch `Response` this
+ * is called with (or its `.headers`) must not be a shared/frozen instance.
+ */
+export function applySsrCacheHeaders(response: Response): void {
+  response.headers.set('Cache-Control', 'private, no-store');
+}


dealako:
@bramwelt Thanks for closing the SSR HTML cache gap called out in #2041 / #2049.

This is a small, well-placed change: the headers are applied on the Web `Response` after `angularApp.handle` resolves and before the notFound rebuild, so both 200 and rebuilt-404 SSR HTML inherit `Cache-Control: private, no-store`. The comment correctly ties the directive to per-user TransferState and meeting `host_key` fields. Protected-file touch on `server.ts` is noted in the PR body and looks proportionate.

**Structured issue count**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: Vary header handling on redirect responses; no automated test pinning these headers
- ⚪ Nit: 0 issues
- ❔ Question: 2 items: whether blanket `no-store` on public SSR routes is intentional; whether error-path SSR responses should share the same cache guard

**AI comment reconciliation**
- **Copilot** (COMMENTED): Agree with the overview. No actionable inline findings to reconcile.
- **Cursor Bot**: Agree with the medium-risk / cross-user cache framing. No duplicate inline feedback.
- **CodeRabbit**: Review skipped (repo config). Nothing to reconcile.

**Revision tracking**
First human review round on this PR. No prior blocking feedback to track.

apps/lfx-one/src/server/server.ts
```diff
@@ -501,6 +501,11 @@ app.use('/**', async (req: Request, res: Response, next: NextFunction) => {
         return next();
       }
 
+      // The rendered HTML can carry per-user data (auth context, host_key/can_view_host_key on
+      // meeting pages), so it must never be served from a shared/intermediary cache to another user.
+      response.headers.set('Cache-Control', 'private, no-store');
+      response.headers.set('Vary', 'Cookie');
```
dealako: [minor] Vary: Cookie — append or drop

Issue: `response.headers.set('Vary', 'Cookie')` replaces any existing `Vary` value. Angular SSR redirect responses can already carry `Vary: X-Forwarded-Prefix` from `createRedirectResponse`, so `set` drops that value. Separately, with `Cache-Control: private, no-store` on the same response, conformant caches will not store the response, so `Vary` does not add cross-user protection here (see `user.controller.ts`, which documents that `private` alone prevents cross-user cache reads and sets no `Vary`).
Why it matters: `set` is a latent footgun if the directive is ever relaxed; `Vary` may also mislead readers into thinking cookie-based cache partitioning is in play.
Fix: Either delete this line, or switch to `response.headers.append('Vary', 'Cookie')` if you want to preserve Angular's existing `Vary` tokens.

apps/lfx-one/src/server/server.ts
```diff
@@ -501,6 +501,11 @@ app.use('/**', async (req: Request, res: Response, next: NextFunction) => {
         return next();
       }
 
+      // The rendered HTML can carry per-user data (auth context, host_key/can_view_host_key on
+      // meeting pages), so it must never be served from a shared/intermediary cache to another user.
+      response.headers.set('Cache-Control', 'private, no-store');
```
dealako: [minor] Pin the SSR cache headers with a test

Issue: Nothing in CI asserts `Cache-Control: private, no-store` (or `Vary: Cookie`) on SSR HTML. `check-headers.sh` is the license-header scanner, not an HTTP header check. There is no `server.spec.ts`, and e2e has no cache-header assertions.
Proof: Other server controllers already test `Cache-Control` in vitest (e.g. committee and org-identity controller specs).
Why it matters: This guard exists specifically to prevent a cross-user cache leak; a future refactor of the `.then()` block could drop it silently.
Fix: Extract a small helper that applies the headers to a `Response` and add a unit spec covering a 200 render and a redirect response.

dealako (dismissed):
@bramwelt Thanks for closing the SSR HTML cache gap called out in #2041 / #2049.

This is a small, well-placed change: the headers are applied on the Web `Response` after `angularApp.handle` resolves and before the notFound rebuild, so both 200 and rebuilt-404 SSR HTML inherit `Cache-Control: private, no-store`. The comment correctly ties the directive to per-user TransferState and meeting `host_key` fields. Protected-file touch on `server.ts` is noted in the PR body and looks proportionate.

**Structured issue count**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: Vary header handling on redirect responses; no automated test pinning these headers
- ⚪ Nit: 0 issues
- ❔ Question: 2 items: whether blanket `no-store` on public SSR routes is intentional; whether error-path SSR responses should share the same cache guard

**AI comment reconciliation**
- **Copilot** (COMMENTED): Agree with the overview. No actionable inline findings to reconcile.
- **Cursor Bot**: Agree with the medium-risk / cross-user cache framing. No duplicate inline feedback.
- **CodeRabbit**: Review skipped (repo config). Nothing to reconcile.

**Revision tracking**
First human review round on this PR. No prior blocking feedback to track.

apps/lfx-one/src/server/server.ts
bramwelt: @dealako Good catch on `.set()` vs `.append()` — I verified the vendored `@angular/ssr` source (`node_modules/@angular/ssr/fesm2022/ssr.mjs`) before deciding.

Two corrections/findings:
1. `createRedirectResponse` (used for route-config redirects) only sets `Location` — no `Vary` at all, so it doesn't carry `Vary: X-Forwarded-Prefix` as suspected. `X-Forwarded-Prefix` is only read to build the `Location` URL.
2. There *is* one real Angular SSR code path that sets `Vary` on a redirect: `redirectBasedOnAcceptLanguage` sets `Vary: Accept-Language` for i18n locale redirects — but that only fires when `supportedLocales.length > 1`, and this app has no i18n locales configured, so it's currently dead code here.

Given `no-store` already fully prevents any conformant cache from storing the response, `Vary: Cookie` wasn't adding cross-user protection in the first place — so I went with option (a) and deleted the line entirely in `2f7026a7e`, rather than switching to `.append()`. Also extracted the header logic into `applySsrCacheHeaders` per your other comment, with a unit spec covering both a normal render and a redirect-shaped response to make sure a future refactor can't silently drop it or clobber an existing header.

apps/lfx-one/src/server/server.ts
bramwelt: @dealako Done. Extracted the header-setting logic into `applySsrCacheHeaders(response: Response)` (`apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.ts`) instead of inlining it in the `.then()` render callback, and added `ssr-cache-headers.helper.spec.ts` in `2f7026a7e` covering:

- a normal 200 SSR render gets `Cache-Control: private, no-store`
- a redirect-shaped response (302 with `Location` + `Vary: Accept-Language`) gets the same header without losing its existing headers
- a pre-existing `Cache-Control` value is overwritten rather than appended to

Note: per the other thread, I dropped `Vary: Cookie` entirely (see that reply for the reasoning), so the spec doesn't assert on it, but it does confirm the helper never touches an existing `Vary` header on a redirect.

bramwelt:
## Review Feedback Addressed

Commit: 2f7026a7eff5a6be9032be5a96ab69df93b6852b

### Changes Made
- **apps/lfx-one/src/server/server.ts**: extracted the SSR cache-header logic into a new `applySsrCacheHeaders` helper instead of inlining it in the `.then()` render callback, and dropped the `Vary: Cookie` header entirely (per @dealako). `Cache-Control: private, no-store` already prevents any conformant cache from storing the response, so `Vary` added no cross-user protection here. I verified against the vendored `@angular/ssr` source that `createRedirectResponse` sets no `Vary` header at all (the suspected `Vary: X-Forwarded-Prefix` doesn't exist — that header is only used to build the redirect `Location`), and that the one real Angular SSR `Vary`-setting path (`Vary: Accept-Language` on i18n locale redirects) is currently dead code in this app since no locales are configured.
- **apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.ts** (new): the extracted, independently-testable helper.
- **apps/lfx-one/src/server/helpers/ssr-cache-headers.helper.spec.ts** (new): unit spec covering a normal 200 render, a redirect-shaped response (existing headers preserved), and Cache-Control overwrite semantics (per @dealako).

### Threads Resolved

dealako:
@bramwelt Thanks for the fast follow-up on `2f7026a7`.

## Revision tracking (since `0705f12`)

- ✅ **Resolved** — Dropped `Vary: Cookie` instead of clobbering Angular redirect `Vary` tokens; the redirect spec case (`Accept-Language` preserved) is the right guard.
- ✅ **Resolved** — Extracted `applySsrCacheHeaders` and added `ssr-cache-headers.helper.spec.ts` (200 render, redirect header preservation, overwrite semantics).
- ✅ **Resolved** — Cross-user SSR HTML cache leak: `private, no-store` still applied before the notFound rebuild, so rebuilt 404 HTML inherits the same headers.
- ✅ **Answered** — Error-path `.catch()` responses stay plain text with no per-user payload; no cache-header change needed there.
- ✅ **Answered** — Blanket `no-store` on the `/**` catch-all is consistent: every SSR path builds session-dependent `auth` context, including public `/docs` and `/u/:username`.

👏 **Nice work**
- The helper JSDoc records *why* `Vary` was omitted, not just that it was removed.
- The redirect test encodes the regression we cared about (existing headers survive).
- Commit message documents the Angular SSR source check on `redirectBasedOnAcceptLanguage`.

**Structured issue count** (carried over + new)
- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 2 issues: PR description still mentions `Vary: Cookie`; JSDoc could state the no-`Vary` rationale without the hypothetical `.set()` call site

**AI comment reconciliation**
- No new CodeRabbit/Cursor/Copilot findings since the prior round. Your thread replies addressed the `Vary` and test feedback directly.

✅ **Approved**

dealako (approved):
@bramwelt Thanks for the fast follow-up on `2f7026a7`.

## Revision tracking (since `0705f12`)

- ✅ **Resolved** — Dropped `Vary: Cookie` instead of clobbering Angular redirect `Vary` tokens; the redirect spec case (`Accept-Language` preserved) is the right guard.
- ✅ **Resolved** — Extracted `applySsrCacheHeaders` and added `ssr-cache-headers.helper.spec.ts` (200 render, redirect header preservation, overwrite semantics).
- ✅ **Resolved** — Cross-user SSR HTML cache leak: `private, no-store` still applied before the notFound rebuild, so rebuilt 404 HTML inherits the same headers.
- ✅ **Answered** — Error-path `.catch()` responses stay plain text with no per-user payload; no cache-header change needed there.
- ✅ **Answered** — Blanket `no-store` on the `/**` catch-all is consistent: every SSR path builds session-dependent `auth` context, including public `/docs` and `/u/:username`.

👏 **Nice work**
- The helper JSDoc records *why* `Vary` was omitted, not just that it was removed.
- The redirect test encodes the regression we cared about (existing headers survive).
- Commit message documents the Angular SSR source check on `redirectBasedOnAcceptLanguage`.

**Structured issue count** (carried over + new)
- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 2 issues: PR description still mentions `Vary: Cookie`; JSDoc could state the no-`Vary` rationale without the hypothetical `.set()` call site

**AI comment reconciliation**
- No new CodeRabbit/Cursor/Copilot findings since the prior round. Your thread replies addressed the `Vary` and test feedback directly.

✅ **Approved**

Issue #1672, opened 2026-08-19
LFX V2 has no meeting-attendance reconciliation equivalent to PCC's Auto-Reconcile


MRashad26:
## Context

PCC (`lfx-pcc`) shipped an "Auto-Reconcile" feature for meeting attendance (Jira epic PCC-1047): it matches Zoom attendees who joined without a known LFX identity against invitees/committee members/prior series history, using a mix of deterministic matching and an LLM fallback, and lets a PMO admin verify/clear the "unknown attendee" backlog for a meeting.

AAIF and other projects are migrating their join/summary pages to redirect to the LFX V2 UI (this repo / `lfx-v2-ui`). **LFX V2 currently has no equivalent feature at all** — no candidate-pool matching, no unverified-attendee review UI, no series-level identity persistence. As projects move to V2, they will lose the reconciliation capability PCC just improved, with no replacement in place.

## Ask

Scope what an equivalent (or improved) attendance-reconciliation capability would look like in the LFX V2 stack, so this isn't silently dropped as projects migrate off PCC. At minimum this should cover:

- Matching unverified Zoom attendees against known identities (invitees, committee/mailing-list members, prior-occurrence history)
- Persisting a verified match at the series level so it doesn't reset every occurrence (the exact problem fixed for PCC in Jira PCC-1451)
- An admin review UI for medium/low-confidence matches

## Related

- Jira PCC-1047 — the PCC-side epic this gap was identified from, including root-cause writeups on PCC-1443, PCC-1451, PCC-1452, PCC-1453 that are directly reusable as design input here.
- Originally filed as Jira LFXV2-3330, discarded there in favor of tracking in GitHub.

MRashad26:
## Resolved — AI-assisted attendance reconciliation shipped in LFX V2

All work for this gap is merged. Summary of the implementation:

### lfx-self-serve
- **#2059** — shared types: reconciliation fields on the past-meeting participant type (`is_verified`, `is_ai_reconciled`, `is_unknown`, candidate/match shapes)
- **#2062** — BFF write routes: POST/PUT/DELETE past-meeting participant endpoints wired to the meeting-service ITX proxy
[...]

PR #2060, opened 2026-09-02
fix(meetings): dedupe past-meeting participants by pairwise identity

MRashad26:
## Summary

- `getPastMeetingParticipants` deduplicated past-meeting participant records from the query service using a single derived key (normalized email, else `uid`). A person who attends under a second LFX account, or whose email is missing on one record but present on another, could not be merged by any single key — since which signal is "strongest" differs per record.
- Replaces the single-key merge with a pairwise `isSamePerson` identity comparator: match on LFID `username` when both records have one, else overlapping email, else normalized display name.
- Mirrors the identity-comparison fix shipped for the same class of bug in `lfx-pcc` PR #3738 (`isSamePerson`), adapted for this repo's `PastMeetingParticipant` type, which has no `lf_user_id`/`lf_sso` field — `username` (LFID username) plays that role here.
- Follow-up commit hardens the merge (`email` is now carried forward the same way `username` already was) and guards the name-fallback against missing `first_name`/`last_name` so two unnamed participants never falsely match on `"undefined undefined"`.

## Changes

- `apps/lfx-one/src/server/services/meeting.service.ts`: replace single-key `Map`-based dedup with pairwise `isSamePerson`/`normalizeParticipantName` comparison in `getPastMeetingParticipants`.
- `apps/lfx-one/src/server/services/meeting.service.spec.ts`: add unit coverage for `getPastMeetingParticipants` — username match, email match, name-fallback match, and non-matching cases.

## Related work

Item 2 of a 5-item roadmap for `linuxfoundation/lfx-self-serve#1672` (LFX V2 port of PCC's attendance-reconciliation feature). Item 1 (shared types) is #2059, open separately — this PR intentionally does not depend on it (dropped a `zoom_user_name` name-fallback reference so this branch stays independently mergeable off `main`).

Refs: linuxfoundation/lfx-self-serve#1672

## Test plan

- [x] `yarn vitest run src/server/services/meeting.service.spec.ts -t "getPastMeetingParticipants"` — 7/7 passing
- [x] `yarn check-types`, `yarn lint:check`, `yarn format:check`, `yarn build`, `./check-headers.sh` all pass
- [x] Pre-PR reviewer trio (general + self-serve-convention + learnings) clean on both commits and on the full-branch sweep

### DIFF
--- a/apps/lfx-one/src/server/services/meeting.service.spec.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.spec.ts
@@ -746,6 +746,7 @@ describe('MeetingService.getPastMeetingParticipants', () => {
     expect(result).toHaveLength(1);
     expect(result[0].is_invited).toBe(true);
     expect(result[0].is_attended).toBe(true);
+    expect(result[0].email).toBe('jane@example.com');
   });
 
   it('does not merge two records with the same email but different usernames', async () => {
@@ -867,19 +868,48 @@ describe('MeetingService.getPastMeetingParticipants', () => {
     expect(result).toHaveLength(1);
   });
 
-  it('refuses to bridge two conflicting LFID usernames through a shared-email no-username record', async () => {
+  it('isolates an ambiguous shared-email bridge record rather than guessing which of two conflicting LFID usernames it belongs to', async () => {
+    const resources = [
+      participantRecord('a', { username: 'user-a', email: 'shared@example.com', is_attended: true }),
+      participantRecord('b', { username: undefined, email: 'shared@example.com', is_attended: true, host: true }),
+      participantRecord('c', { username: 'user-b', email: 'shared@example.com' }),
+    ];
+
+    proxyRequest.mockResolvedValueOnce({ resources });
+    const forward = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    proxyRequest.mockResolvedValueOnce({ resources: [...resources].reverse() });
+    const reversed = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    for (const result of [forward, reversed]) {
+      expect(result).toHaveLength(3);
+      expect(result.map((p) => p.username).sort()).toEqual([undefined, 'user-a', 'user-b'].sort());
+
+      // The ambiguous bridge record's own attendance/host flags must not bleed into either
+      // conflicting username's group, regardless of encounter order.
+      const bridgeRecord = result.find((p) => !p.username);
+      const userA = result.find((p) => p.username === 'user-a');
+      const userB = result.find((p) => p.username === 'user-b');
+      expect(bridgeRecord?.is_attended).toBe(true);
+      expect(bridgeRecord?.host).toBe(true);
+      expect(userA?.host).toBe(false);
+      expect(userB?.is_attended).toBe(false);
+    }
+  });
+
+  it('refuses to bridge two conflicting guest emails through a shared-name no-email record', async () => {
     proxyRequest.mockResolvedValueOnce({
       resources: [
-        participantRecord('a', { username: 'user-a', email: 'shared@example.com' }),
-        participantRecord('b', { username: undefined, email: 'shared@example.com' }),
-        participantRecord('c', { username: 'user-b', email: 'shared@example.com' }),
+        participantRecord('a', { email: 'guest-one@example.com', first_name: 'Sam', last_name: 'Guest' }),
+        participantRecord('b', { email: undefined, first_name: 'Sam', last_name: 'Guest' }),
+        participantRecord('c', { email: 'guest-two@example.com', first_name: 'Sam', last_name: 'Guest' }),
       ],
     });
 
     const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
 
-    expect(result).toHaveLength(2);
-    expect(result.map((p) => p.username).sort()).toEqual(['user-a', 'user-b']);
+    expect(result).toHaveLength(3);
+    expect(result.map((p) => p.email).sort()).toEqual(['guest-one@example.com', 'guest-two@example.com', undefined].sort());
   });
 
   it('does not let a blank-string email sentinel on the preferred record discard a real email on merge', async () => {
--- a/apps/lfx-one/src/server/services/meeting.service.spec.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.spec.ts
@@ -18,6 +18,11 @@ vi.mock('@lfx-one/shared/enums', async (importOriginal) => importOriginal());
 vi.mock('@lfx-one/shared/utils', () => ({
   buildRecurrenceNeverEndDate: vi.fn(),
   getPastMeetingTranscriptUrl: vi.fn(),
+  isUnresolvableParticipantName: vi.fn((first?: string | null, last?: string | null) => {
+    const tokens = [first, last].map((token) => (token ?? '').trim().toLowerCase());
+    const meaningful = tokens.filter((token) => token && token !== 'unknown' && token !== '[unknown]');
+    return meaningful.length === 0;
+  }),
   mapITXResponseToMeetingRsvp: vi.fn(),
   normalizeIndexedMeetingAiSummary: vi.fn((meeting) => meeting),
   normalizeIndexedMeetingInviteResponses: vi.fn((meeting) => meeting),
@@ -803,5 +808,58 @@ describe('MeetingService.getPastMeetingParticipants', () => {
     const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
 
     expect(result).toHaveLength(2);
+  });
+
+  it('does not merge two unnamed records with no email or username to fall back on', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { first_name: undefined, last_name: undefined }),
+        participantRecord('b', { first_name: undefined, last_name: undefined }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(2);
+  });
+
+  it('does not merge two records with placeholder "[unknown]" names', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { first_name: '[unknown]', last_name: '[unknown]' }),
+        participantRecord('b', { first_name: '[unknown]', last_name: '[unknown]' }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(2);
+  });
+
+  it('merges on matching name when only one side has an email', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { email: 'jane@example.com', first_name: 'Jane', last_name: 'Doe' }),
+        participantRecord('b', { first_name: 'Jane', last_name: 'Doe' }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(1);
+  });
+
+  it('coalesces three records into one person via a bridging record, regardless of encounter order', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { email: 'shared@example.com', username: undefined }),
+        participantRecord('b', { email: 'other@example.com', username: 'jdoe' }),
+        participantRecord('c', { email: 'shared@example.com', username: 'jdoe' }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(1);
   });
 });
--- a/apps/lfx-one/src/server/services/meeting.service.spec.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.spec.ts
@@ -695,3 +695,113 @@ describe('MeetingService.getAuthorizedRegistrantsForImport', () => {
     expect(query.page_size).toBe(51);
   });
 });
+
+describe('MeetingService.getPastMeetingParticipants', () => {
+  let service: MeetingService;
+
+  const participantRecord = (id: string, overrides: Record<string, unknown> = {}) => ({
+    id: `v1_past_meeting_participant:${id}`,
+    data: {
+      uid: id,
+      meeting_id: 'meeting-1',
+      meeting_and_occurrence_id: 'meeting-1-occ-1',
+      past_meeting_id: 'past-1',
+      first_name: 'Jane',
+      last_name: 'Doe',
+      host: false,
+      is_attended: false,
+      is_invited: true,
+      org_is_member: false,
+      org_is_project_member: false,
+      created_at: '2026-01-01T00:00:00.000Z',
+      updated_at: '2026-01-01T00:00:00.000Z',
+      ...overrides,
+    },
+  });
+
+  beforeEach(() => {
+    proxyRequest.mockReset();
+    service = new MeetingService();
+  });
+
+  it('merges two records sharing the same LFID username, even with different emails', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { email: 'jane@example.com', username: 'jdoe', is_invited: true, is_attended: false }),
+        participantRecord('b', { email: 'jane.alt@example.com', username: 'jdoe', is_invited: false, is_attended: true }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(1);
+    expect(result[0].is_invited).toBe(true);
+    expect(result[0].is_attended).toBe(true);
+  });
+
+  it('does not merge two records with the same email but different usernames', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { email: 'shared@example.com', username: 'user-a' }),
+        participantRecord('b', { email: 'shared@example.com', username: 'user-b' }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(2);
+  });
+
+  it('merges by email when neither record has a username', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [participantRecord('a', { email: 'guest@example.com' }), participantRecord('b', { email: 'GUEST@example.com' })],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(1);
+  });
+
+  it('merges on matching email even when only one side has a username — email is checked before the username-asymmetry fallback', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [participantRecord('a', { email: 'guest@example.com', username: 'jdoe' }), participantRecord('b', { email: 'guest@example.com' })],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(1);
+  });
+
+  it('does not merge asymmetric-username records when neither has an email to fall back on', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { username: 'jdoe', first_name: 'Jane', last_name: 'Doe' }),
+        participantRecord('b', { first_name: 'Jane', last_name: 'Doe' }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(2);
+  });
+
+  it('falls back to normalized display name when neither username nor email is present', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [participantRecord('a', { first_name: 'Jane', last_name: 'Doe' }), participantRecord('b', { first_name: 'jane', last_name: 'doe' })],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(1);
+  });
+
+  it('does not merge different people who share no identity signal at all', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [participantRecord('a', { first_name: 'Jane', last_name: 'Doe' }), participantRecord('b', { first_name: 'John', last_name: 'Smith' })],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(2);
+  });
+});
--- a/apps/lfx-one/src/server/services/meeting.service.spec.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.spec.ts
@@ -18,6 +18,10 @@ vi.mock('@lfx-one/shared/enums', async (importOriginal) => importOriginal());
 vi.mock('@lfx-one/shared/utils', () => ({
   buildRecurrenceNeverEndDate: vi.fn(),
   getPastMeetingTranscriptUrl: vi.fn(),
+  // Intentionally a light behavioral double, not a frozen copy meant to track the real predicate:
+  // it only needs to exercise the placeholder-vs-blank branch in this file's dedup tests. The
+  // predicate's own placeholder-token coverage lives in meeting.utils.spec.ts — a future token
+  // added there won't be reflected here, but that's the authoritative test for this behavior.
   isUnresolvableParticipantName: vi.fn((first?: string | null, last?: string | null) => {
     const tokens = [first, last].map((token) => (token ?? '').trim().toLowerCase());
     const meaningful = tokens.filter((token) => token && token !== 'unknown' && token !== '[unknown]');
@@ -861,5 +865,34 @@ describe('MeetingService.getPastMeetingParticipants', () => {
     const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
 
     expect(result).toHaveLength(1);
+  });
+
+  it('refuses to bridge two conflicting LFID usernames through a shared-email no-username record', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { username: 'user-a', email: 'shared@example.com' }),
+        participantRecord('b', { username: undefined, email: 'shared@example.com' }),
+        participantRecord('c', { username: 'user-b', email: 'shared@example.com' }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(2);
+    expect(result.map((p) => p.username).sort()).toEqual(['user-a', 'user-b']);
+  });
+
+  it('does not let a blank-string email sentinel on the preferred record discard a real email on merge', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [
+        participantRecord('a', { email: '', first_name: 'Jane', last_name: 'Doe', is_attended: true }),
+        participantRecord('b', { email: 'jane@example.com', first_name: 'Jane', last_name: 'Doe', is_attended: false }),
+      ],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(1);
+    expect(result[0].email).toBe('jane@example.com');
   });
 });
--- a/apps/lfx-one/src/server/services/meeting.service.ts
+++ b/apps/lfx-one/src/server/services/meeting.service.ts
@@ -981,8 +981,16 @@ export class MeetingService {
       })
     );
 
+    // Union-find over connected identity-match edges. `rootUsername` tracks the resolved LFID
+    // username for each component's current root, since a merge must be refused (not just
+    // skipped) once known: isSamePerson is not transitive — a no-username row can share an email
+    // with two records that carry different, conflicting usernames, and unioning through it would
+    // incorrectly merge two distinct LFX accounts into one participant.
     const parent = raw.map((_, index) => index);
+    const rootUsername = raw.map((participant) => participant.username || undefined);
     const find = (index: number): number => {
+      // Path halving: reparent every other node to its grandparent so later find() calls stay
+      // near O(1) amortized. This mutates `parent` even on a plain lookup.
       while (parent[index] !== index) {
         parent[index] = parent[parent[index]];
         index = parent[index];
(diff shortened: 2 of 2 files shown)


apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,18 +976,17 @@ export class MeetingService {
...
 
-    const seen = new Map<string, PastMeetingParticipant>();
+    const merged: PastMeetingParticipant[] = [];
     for (const participant of raw) {
-      const normalizedEmail = participant.email?.trim().toLowerCase();
-      const key = normalizedEmail || participant.uid;
-      const existing = seen.get(key);
-      if (!existing) {
-        seen.set(key, participant);
+      const existingIndex = merged.findIndex((candidate) => this.isSamePerson(candidate, participant));
```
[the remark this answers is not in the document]
MRashad26: Done — replaced the left-to-right pairwise merge with a union-find over connected components. Records are now grouped by identity-match edges across the whole set, so a bridging record correctly coalesces a chain of matches into one person regardless of encounter order, then each connected group is folded into a single merged record. Added a regression test for exactly the three-record ordering scenario you described.

See commit 402b8ceaf: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1934,4 +1936,40 @@ export class MeetingService {
...
+  private isSamePerson(a: PastMeetingParticipant, b: PastMeetingParticipant): boolean {
+    if (a.username && b.username) {
+      return a.username === b.username;
+    }
+
+    const aEmail = a.email?.trim().toLowerCase();
+    const bEmail = b.email?.trim().toLowerCase();
+    if (aEmail || bEmail) {
+      return !!aEmail && aEmail === bEmail;
+    }
```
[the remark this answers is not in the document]
MRashad26: Done — isSamePerson now only short-circuits on email when BOTH records have one (`aEmail && bEmail` instead of `aEmail || bEmail`). An asymmetric email (one record has it, the other doesn't) now falls through to the username-asymmetry guard and then the normalized-name fallback, so the motivating case (one record enriched with an email, the other not, both with matching names) now dedupes correctly. Added a regression test for this exact case.

See commit 402b8ceaf: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.spec.ts
```diff
@@ -695,3 +695,113 @@ describe('MeetingService.getAuthorizedRegistrantsForImport', () => {
...
+  it('does not merge different people who share no identity signal at all', async () => {
+    proxyRequest.mockResolvedValueOnce({
+      resources: [participantRecord('a', { first_name: 'Jane', last_name: 'Doe' }), participantRecord('b', { first_name: 'John', last_name: 'Smith' })],
+    });
+
+    const result = await service.getPastMeetingParticipants(req, 'meeting-1-occ-1');
+
+    expect(result).toHaveLength(2);
+  });
+});
```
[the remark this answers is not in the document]
MRashad26: Done — added a dedicated case with both records missing first_name/last_name entirely (no email/username either), asserting they stay as two separate records. This actually exercises the `!!aName` guard (the previous last test used two distinct non-empty names, which doesn't).

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1934,4 +1936,40 @@ export class MeetingService {
...
+
+  /** Normalized display name for identity comparison. Empty when both name parts are missing, so two unnamed participants never match. */
+  private normalizeParticipantName(participant: PastMeetingParticipant): string {
+    const first = participant.first_name?.trim();
+    const last = participant.last_name?.trim();
+    if (!first && !last) {
+      return '';
+    }
+    return `${first ?? ''} ${last ?? ''}`.trim().toLowerCase();
+  }
```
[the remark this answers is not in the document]
MRashad26: Done — normalizeParticipantName now reuses the shared `isUnresolvableParticipantName` helper (already used elsewhere in this repo for sinking unresolvable rows to the bottom of people lists) instead of only checking for missing name parts. Placeholder values like `unknown` and `[unknown]` now normalize to an empty name, so two such rows never falsely merge. Added a regression test with two `[unknown]`/`[unknown]` records asserting they stay separate.

See commit 402b8ceaf: meeting.service.ts.

MRashad26:
## Review Feedback Addressed

Commit: 402b8ceaf5d8f13368c79e5f46b60e66d9c559ce

### Changes Made
- **`apps/lfx-one/src/server/services/meeting.service.ts`**: replaced the left-to-right pairwise merge in `getPastMeetingParticipants` with a union-find/connected-components grouping, so a "bridge" record correctly coalesces a chain of identity matches into one person regardless of encounter order (per @copilot-pull-request-reviewer)
- **`apps/lfx-one/src/server/services/meeting.service.ts`**: `isSamePerson` now only treats email as decisive when both records have one — an asymmetric email falls through to the username-asymmetry guard and then the name fallback, fixing the case where a partially-enriched duplicate (same person, one record missing an email) previously failed to merge (per @copilot-pull-request-reviewer)
- **`apps/lfx-one/src/server/services/meeting.service.ts`**: `normalizeParticipantName` now reuses the shared `isUnresolvableParticipantName` helper so placeholder names like `unknown`/`[unknown]` normalize to empty and never falsely merge two distinct unnamed Zoom/dial-in rows (per @cursor)
- **`apps/lfx-one/src/server/services/meeting.service.spec.ts`**: added regression tests for the three-record bridging scenario, the asymmetric-email merge, the unnamed-pair guard, and the placeholder-name guard (per @copilot-pull-request-reviewer)

### Threads Resolved
4 of 4 unresolved threads addressed and resolved in this iteration.

dealako:
Hi @MRashad26 — thanks for the thorough iteration on this PR. The move from single-key dedup to connected-component grouping is the right structural fix, and the follow-up commit addressed the first round of bot feedback well (union-find bridging, asymmetric email, placeholder names, unnamed-participant guard). The test suite is substantially better than the initial commit.

Two correctness issues remain in the latest code that I think should be fixed before merge, plus a few minor style/test gaps.

## Issue count

- 🔴 **Blocking: 2 issues**
  1. **Transitive username conflict via email bridge** — union-find can collapse `user-a` and `user-b` when a no-username row shares their email (agrees with @copilot-pull-request-reviewer)
  2. **Empty-string email wins merge** — `??` preserves `email: ''` over a real address on the other duplicate (agrees with @cursor)

- 🟡 **Minor: 3 issues**
  1. Missing regression tests for the two blocking scenarios above
  2. Union-find block has no inline comments explaining path-halving / mutating `find`
  3. Spec copies `isUnresolvableParticipantName` body instead of stubbing with a pointer to `meeting.utils.spec.ts`

- ⚪ **Nit: 2 issues**
  1. Consider extracting the grouping pass to match `mergePastMeetingParticipantGroup`
  2. JSDoc could note that transitive closure can also chain records that share no direct signal (not only reunite bridges)

- ❔ **Question: 1 item**
  1. Name-only fallback intentionally merges two `"Jane Doe"` guests with no username/email — is that accepted ambiguity per lfx-pcc parity, or should common-name collisions stay separate?

## Revision tracking (re-review)

**Resolved since first review** (commit `402b8ce`):
- Order-dependent left-to-right merge → union-find connected components ✅
- Asymmetric email blocking name fallback → `aEmail && bEmail` guard ✅
- Unnamed `"undefined undefined"` false match → `!!aName` guard + dedicated test ✅
- Placeholder `"[unknown]"` false merge → `isUnresolvableParticipantName` reuse ✅

**Still open:**
- Transitive username conflict (raised by Copilot in latest round) 🔴
- Empty-string email coalescence (raised by Cursor, still present) 🔴

## Bot reconciliation

| Bot | Topic | Position |
|-----|-------|----------|
| Copilot | Order-dependent bridge | ✅ Resolved in `402b8ce` |
| Copilot | Asymmetric email | ✅ Resolved in `402b8ce` |
| Copilot | Transitive username via shared email | 🔴 Agree — still reproducible |
| Copilot | Empty-string `??` email | 🔴 Agree — still reproducible |
| Cursor | Placeholder name merge | ✅ Resolved |
| Cursor | Union-find name-only over-merge | ❔ Intentional per spec tests; product question, not a blocker |

## Final decision

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,32 +981,39 @@ export class MeetingService {
...
+      while (parent[index] !== index) {
+        parent[index] = parent[parent[index]];
+        index = parent[index];
+      }
+      return index;
+    };
+
+    for (let i = 0; i < raw.length; i++) {
+      for (let j = i + 1; j < raw.length; j++) {
+        if (this.isSamePerson(raw[i], raw[j])) {
```
dealako: [blocking] Transitive merge unites conflicting LFID usernames via email bridge

**Issue:** Union-find unions every pairwise `isSamePerson` hit, but email-only matches do not check username conflicts when only one side has a username. A no-username row sharing an email can bridge `user-a` and `user-b` into one component even though `isSamePerson(A, C)` is false.

**Proof:** A `{username:'user-a', email:'shared@example.com'}`, B `{email:'shared@example.com'}`, C `{username:'user-b', email:'shared@example.com'}`. `isSamePerson(A,B)=true`, `isSamePerson(B,C)=true`, `isSamePerson(A,C)=false` (both usernames, different). Union-find yields one group. The test at line 747 only covers the direct pair; the bridge test at line 852 uses matching `jdoe` usernames, not conflicting ones. (Agrees with @copilot-pull-request-reviewer.)

**Why it matters:** Two distinct LFID accounts collapse into one participant row, losing attendance/invite flags and showing the wrong person.

**Fix:** Before unioning roots, reject merges that would place conflicting non-empty usernames in the same component (track username sets per root), or post-split groups containing more than one distinct username.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1934,4 +1948,68 @@ export class MeetingService {
...
+        is_attended: merged.is_attended || participant.is_attended,
+        is_invited: merged.is_invited || participant.is_invited,
+        host: merged.host || participant.host,
+        org_is_member: merged.org_is_member || participant.org_is_member,
+        org_is_project_member: merged.org_is_project_member || participant.org_is_project_member,
+        avatar_url: preferred.avatar_url ?? other.avatar_url,
+        job_title: preferred.job_title ?? other.job_title,
+        org_name: preferred.org_name ?? other.org_name,
+        username: preferred.username ?? other.username,
+        email: preferred.email ?? other.email,
```
dealako: [blocking] Empty-string email survives merge via `??`

**Issue:** `preferred.email ?? other.email` treats `''` as a present value. An attended row with upstream empty-string sentinel keeps blank email and discards a real address from the duplicate.

**Proof:** Merge `{email:'', username:'jdoe', is_attended:true}` with `{email:'jane@example.com'}` → `email:''`. Same pattern affects `username`, `avatar_url`, `job_title`, `org_name`. In `isSamePerson`, `''` after trim is falsy so it also skips the both-have-email branch. (Agrees with @cursor — still open after `402b8ce`.)

**Why it matters:** Merged rows can lose the only usable email, breaking downstream lookup (`isUserPastMeetingParticipant`) and committee enrichment.

**Fix:** Use a `pickNonempty(a, b)` helper that treats blank/whitespace as missing instead of raw `??`.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,32 +981,39 @@ export class MeetingService {
...
 
-    const seen = new Map<string, PastMeetingParticipant>();
-    for (const participant of raw) {
-      const normalizedEmail = participant.email?.trim().toLowerCase();
-      const key = normalizedEmail || participant.uid;
-      const existing = seen.get(key);
-      if (!existing) {
-        seen.set(key, participant);
-        continue;
+    const parent = raw.map((_, index) => index);
```
dealako, line 989: [minor] Union-find block lacks inline comments

**Issue:** The 33-line dedupe block has zero inline comments. `find` mutates `parent` on read (path halving), which reads like a pure lookup.

**Proof:** Lines 984–1003 have no `//` annotations; repo style annotates non-obvious algorithmic steps inline (e.g. `getCommitteeNameMap` Promise.allSettled comment).

**Why it matters:** Future maintainers may not realize `find` has side effects, increasing risk of subtle bugs during edits.

**Fix:** Add 2–3 short comments at the union-find setup and the `parent[index] = parent[parent[index]]` line.

apps/lfx-one/src/server/services/meeting.service.spec.ts
```diff
@@ -18,6 +18,11 @@ vi.mock('@lfx-one/shared/enums', async (importOriginal) => importOriginal());
 vi.mock('@lfx-one/shared/utils', () => ({
   buildRecurrenceNeverEndDate: vi.fn(),
   getPastMeetingTranscriptUrl: vi.fn(),
+  isUnresolvableParticipantName: vi.fn((first?: string | null, last?: string | null) => {
```
dealako, line 25: [minor] Spec copies `isUnresolvableParticipantName` implementation

**Issue:** The mock re-implements the real predicate body verbatim instead of stubbing with a pointer to existing coverage.

**Proof:** Spec lines 21–25 mirror `meeting.utils.ts:1356–1359`. If the real predicate gains new placeholder tokens, these tests keep passing against the frozen copy.

**Why it matters:** Spec divergence from shipped behavior without signaling.

**Fix:** Use `vi.fn()` stub with a comment pointing at `meeting.utils.spec.ts` for real predicate coverage.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1934,4 +1948,68 @@ export class MeetingService {
...
+      return aEmail === bEmail;
+    }
+
+    if (a.username || b.username) {
+      return false;
+    }
+
+    const aName = this.normalizeParticipantName(a);
+    const bName = this.normalizeParticipantName(b);
+    return !!aName && aName === bName;
```
dealako, line 2034: [question] Name-only merge accepted ambiguity?

**Issue:** When neither record has username or email, `isSamePerson` merges on normalized display name. Two different attendees named "Jane Doe" with no enrichment would be deduped into one row.

**Proof:** Test at line 793 expects name-only merge. This is intentional per lfx-pcc parity, but common-name collisions at large meetings are possible.

**Why it matters:** Product tradeoff between duplicate-row noise and false-merge risk — not necessarily a bug if aligned with PCC.

**Fix:** If accepted, add a one-line JSDoc note documenting the collision risk. If not, require a stronger corroborating signal before name-only union.

dealako (changes requested):
## dd-pr-review summary

Solid direction on participant deduplication — union-find fixes the ordering bug from round 1 and the test coverage is much improved. Two blocking correctness issues remain before I can approve.

### Blocking (2)

1. **Transitive username conflict** — Email-only `isSamePerson` matches ignore username when only one side has one. Union-find bridges conflicting LFID accounts via a no-username email row.
2. **Empty-string email coalescence** — `??` preserves `email: ''` over a real address on the duplicate.

### Minor (3)

- Add regression tests for the two blocking scenarios
- Inline comments on union-find path-halving
- Stub `isUnresolvableParticipantName` in spec instead of copying implementation

### Question (1)

Name-only merge of two guests with identical display names — confirm this matches lfx-pcc accepted ambiguity.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,32 +981,39 @@ export class MeetingService {
...
+      return index;
+    };
+
+    for (let i = 0; i < raw.length; i++) {
+      for (let j = i + 1; j < raw.length; j++) {
+        if (this.isSamePerson(raw[i], raw[j])) {
+          const rootI = find(i);
+          const rootJ = find(j);
+          if (rootI !== rootJ) {
+            parent[rootI] = rootJ;
```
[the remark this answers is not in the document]
MRashad26: Done — component unions are now conflict-aware. I track the resolved LFID username per union-find root and refuse to union two components whose known usernames conflict, so the `user-a`/no-username/`user-b` bridge case you described now yields two separate components instead of one. Added a regression test with exactly that shape.

See commit d038417c7: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.ts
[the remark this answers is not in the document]
MRashad26: Done — added a pickNonBlank() helper that treats whitespace-only strings as missing (matching isSamePerson's own email normalization) and used it for email in the merge, so an attended `{ email: '' }` row no longer wins over a real email from the other record. Added a regression test for this exact scenario.

See commit d038417c7: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,32 +981,39 @@ export class MeetingService {
...
+          }
+        }
```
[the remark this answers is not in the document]
MRashad26: Done — same fix as the parallel report from @copilot-pull-request-reviewer and @dealako: component unions now refuse to merge when the two components carry conflicting known LFID usernames, so a name-only or email-only bridge row can no longer unite two people that isSamePerson itself would reject.

See commit d038417c7: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.ts
MRashad26: Done — added conflict-aware unioning: I track the resolved username per union-find root, and before unioning two components I check whether both have a known, differing username. If so, the union is refused. Verified against your exact A/B/C proof case with a new regression test asserting 2 groups with usernames `user-a` and `user-b` preserved separately.

See commit d038417c7: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.ts
MRashad26: Done — added the pickNonBlank() helper you suggested (as a private method) and applied it to email, username, avatar_url, job_title, and org_name in mergePastMeetingParticipantGroup, not just email. Added a regression test for the blank-email-sentinel case.

See commit d038417c7: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.ts
MRashad26: Fixed — added inline comments at the union-find setup explaining the conflict-tracking rationale, and at the path-halving line in find() noting it mutates parent on a lookup. Good catch on the side effect being non-obvious.

See commit d038417c7: meeting.service.ts.

apps/lfx-one/src/server/services/meeting.service.spec.ts
MRashad26: Added a comment above the mock making explicit that it's an intentional light double for this file's dedup tests, not meant to track the real predicate — the authoritative placeholder-token coverage lives in meeting.utils.spec.ts. I kept the reimplementation rather than a bare stub because two tests in this file (the unnamed-pair guard and the `[unknown]` placeholder guard) need to exercise the actual placeholder-detection branch, not just a fixed return value; a bare stub would need per-test mockImplementationOnce overrides that add indirection without meaningfully reducing the risk you flagged.

See commit d038417c7: meeting.service.spec.ts.

apps/lfx-one/src/server/services/meeting.service.ts
MRashad26: Confirmed intentional — this is an accepted PCC-parity tradeoff made during the original dedup fix: two same-named guests with no email or username on any record will merge, since there's no stronger signal available to distinguish them. Added a doc note directly on isSamePerson documenting the collision risk so it's visible in-code, not just in review history.

No code change to the matching logic — see commit d038417c7: meeting.service.ts for the doc note.

MRashad26:

Commit: d038417c798f79a8fcce90c37937aadda5ed65ad

### Changes Made
- **`apps/lfx-one/src/server/services/meeting.service.ts`**: made union-find component merges conflict-aware — track the resolved LFID username per root and refuse to union two components whose known usernames conflict, so a no-username row sharing one email can no longer bridge two records with different usernames into one person (per @copilot-pull-request-reviewer, @cursor, @dealako)
- **`apps/lfx-one/src/server/services/meeting.service.ts`**: added `pickNonBlank()` and used it for `email`, `username`, `avatar_url`, `job_title`, and `org_name` in `mergePastMeetingParticipantGroup`, so an upstream empty-string sentinel on the preferred record no longer discards a real value from the other record via `??` (per @copilot-pull-request-reviewer, @cursor, @dealako)
- **`apps/lfx-one/src/server/services/meeting.service.ts`**: added inline comments to the union-find block explaining the conflict-tracking rationale and that `find()` mutates `parent` on read (per @dealako)
- **`apps/lfx-one/src/server/services/meeting.service.ts`**: documented the name-only merge fallback's accepted PCC-parity tradeoff directly on `isSamePerson` (per @dealako)
- **`apps/lfx-one/src/server/services/meeting.service.spec.ts`**: added a comment clarifying the `isUnresolvableParticipantName` mock is an intentional light double for this file's tests, not a frozen copy of the real predicate — authoritative coverage lives in `meeting.utils.spec.ts` (per @dealako)
- **`apps/lfx-one/src/server/services/meeting.service.spec.ts`**: added regression tests for the username-conflict bridge case and the blank-string email sentinel case

### Threads Resolved
8 of 8 unresolved threads addressed and resolved in this iteration.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,32 +981,58 @@ export class MeetingService {
...
+        const rootI = find(i);
+        const rootJ = find(j);
+        if (rootI === rootJ) {
+          continue;
+        }
+
+        const usernameI = rootUsername[rootI];
+        const usernameJ = rootUsername[rootJ];
+        if (usernameI && usernameJ && usernameI !== usernameJ) {
+          continue;
```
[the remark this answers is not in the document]
MRashad26: Fixed in f1997cdd0 — the conflict-aware grouping was fundamentally redesigned: it now does a plain transitive-closure union-find over `isSamePerson` edges first, then a new `splitConflictingComponent` pass detects hard-identifier conflicts (2+ distinct usernames, or — absent a username conflict — 2+ distinct emails among username-less members) and splits the component by that identifier. Any row lacking the relevant identifier (the ambiguous bridge row) is isolated as its own singleton group instead of being guessed into whichever side the scan reaches first. This fixes both the order-dependent attribution of the bridge row's own attendance/host flags, and extends conflict detection to emails (not just usernames). Added an order-permutation regression test (forward + reversed input) that asserts identical group membership and merged flags.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1934,4 +1967,80 @@ export class MeetingService {
...
+        is_attended: merged.is_attended || participant.is_attended,
+        is_invited: merged.is_invited || participant.is_invited,
+        host: merged.host || participant.host,
+        org_is_member: merged.org_is_member || participant.org_is_member,
+        org_is_project_member: merged.org_is_project_member || participant.org_is_project_member,
+        avatar_url: this.pickNonBlank(preferred.avatar_url, other.avatar_url),
+        job_title: this.pickNonBlank(preferred.job_title, other.job_title),
+        org_name: this.pickNonBlank(preferred.org_name, other.org_name),
+        username: this.pickNonBlank(preferred.username, other.username),
+        email: this.pickNonBlank(preferred.email, other.email),
```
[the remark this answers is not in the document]
MRashad26: Fixed in f1997cdd0 — added a new `pickMergedEmail` helper used by `mergePastMeetingParticipantGroup`. It now prefers the email of whichever group member has `is_invited === true` and a non-blank email, falling back to the first non-blank email in original order. Previously the attended-first preference in the reduce could silently discard a real, different, invited/account email in favor of an ad-hoc attended-session email — this broke downstream committee-enrichment lookups that join on `participant.email` against the invited address. Added a regression assertion to the existing same-LFID-username-different-email test verifying the invited email wins.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,32 +981,58 @@ export class MeetingService {
...
+        }
+
+        const usernameI = rootUsername[rootI];
+        const usernameJ = rootUsername[rootJ];
+        if (usernameI && usernameJ && usernameI !== usernameJ) {
+          continue;
+        }
+
+        parent[rootI] = rootJ;
+        rootUsername[rootJ] = usernameI ?? usernameJ;
```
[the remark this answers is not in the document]
MRashad26: Fixed in f1997cdd0 — same root cause as the other order-dependent-attribution thread. The new `splitConflictingComponent` post-processing now tracks conflicting emails among username-less members (not just conflicting usernames), and isolates any ambiguous bridge row as its own group rather than merging it into whichever conflicting side the union-find happens to reach first. Added a dedicated regression test ('refuses to bridge two conflicting guest emails through a shared-name no-email record') covering this exact scenario.

MRashad26:
## Review Feedback Addressed

Commit: f1997cdd0

### Changes Made
- **meeting.service.ts**: `getPastMeetingParticipants` now does a plain transitive-closure union-find over `isSamePerson` edges, then a new `splitConflictingComponent` pass detects hard-identifier conflicts (usernames, or emails among username-less members) and splits by that identifier — isolating an ambiguous bridge row as its own group instead of guessing which side it belongs to. Fixes order-dependent attribution of a bridge row's attendance/host flags and extends conflict detection to emails (per @copilot-pull-request-reviewer, @cursor)
- **meeting.service.ts**: `mergePastMeetingParticipantGroup` now uses a new `pickMergedEmail` helper that prefers the invited/account email over an ad-hoc attended-session email when a merged group has two different non-blank emails (per @copilot-pull-request-reviewer)
- **meeting.service.spec.ts**: updated the username-conflict regression test to expect the bridge record isolated as its own group, added order-permutation coverage (forward + reversed input) for merged flags, added an email-conflict analog test, and added an email-preference assertion to the existing same-username-different-email test

### Threads Resolved

All tests pass (60/60), lint and build are clean.

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -1934,4 +1957,142 @@ export class MeetingService {
...
+   * key can't do this since which signal is "strongest" differs per record (e.g. a guest who later
+   * links an LFX account on a subsequent occurrence). Email is only decisive when BOTH records have
+   * one — an asymmetric email (one record has it, the other doesn't) falls through to the
+   * username-asymmetry guard and then the name fallback, so the common case of a partially-enriched
+   * duplicate (same person, one record still missing an email) can still merge on matching names.
+   * The name-only fallback is an accepted PCC-parity tradeoff: two distinct guests who share a
+   * display name and have no email/username on any record will be merged. Callers needing a
+   * stronger guarantee should require a corroborating signal before relying on this path.
+   */
+  /**
```
dealako, line 1973: [nit] Duplicate adjacent JSDoc blocks

**Issue:** Two consecutive `/** ... */` blocks sit directly above `splitConflictingComponent` — the first documents `isSamePerson`'s PCC-parity tradeoff, the second documents the split helper. Only the second block is attached to the method below it; the first is orphaned.

**Proof:** Lines 1961–1972 (isSamePerson JSDoc) immediately precede lines 1973–1982 (splitConflictingComponent JSDoc) with no `isSamePerson` declaration between them.

**Why it matters:** Orphaned JSDoc is easy to miss during edits and doesn't show up on `isSamePerson` in IDE hovers.

**Fix:** Move the PCC-parity paragraph onto `isSamePerson`'s own JSDoc (above its `private isSamePerson` declaration at line 2017).

apps/lfx-one/src/server/services/meeting.service.ts
```diff
@@ -974,32 +981,48 @@ export class MeetingService {
...
-        continue;
+    // Plain union-find over connected identity-match edges (pure transitive closure, no
+    // conflict veto during union). isSamePerson is not transitive — a no-username row can
+    // share an email with two records that carry different, conflicting usernames — so a
+    // component can chain together records that never agree pairwise. Conflicts are detected
+    // and split out afterward, per component, in splitConflictingComponent.
+    const parent = raw.map((_, index) => index);
+    const find = (index: number): number => {
+      // Path halving: reparent every other node to its grandparent so later find() calls stay
+      // near O(1) amortized. This mutates `parent` even on a plain lookup.
```
dealako, line 992: [nit] Path-halving complexity claim

**Issue:** Comment says "near O(1) amortized" but linking is arbitrary (no union-by-rank/size), so the bound is O(log n) per operation — though immaterial since the surrounding pairwise loop is O(n²).

**Proof:** Line 1008 uses `parent[rootI] = rootJ` with no rank array; only path halving is applied inside `find`.

**Fix:** Soften to "keeps the tree shallow" or note O(log n) without union-by-rank.

dealako (approved):
## dd-pr-review follow-up summary

All round-1 blocking items are resolved in `d038417` and `f1997cdd`. The `splitConflictingComponent` post-split approach is especially solid — it isolates ambiguous bridge rows order-independently and adds the guest-email-through-name conflict case.

### Resolved (6)

- Transitive username conflict → `splitConflictingComponent` + order-independent test
- Empty-string email merge → `pickNonBlank` + regression test
- Missing tests, union-find comments, spec stub comment → all addressed
- Copilot encounter-order concern → resolved by post-split isolation

### Remaining (nits only)

- Duplicate adjacent JSDoc blocks before `splitConflictingComponent` / `isSamePerson`
- Path-halving comment overstates amortized bound (cosmetic; outer loop is O(n²) anyway)

dealako:
Hi @MRashad26 — really nice follow-through on the round-2 feedback. The `splitConflictingComponent` refactor in `f1997cdd` is a cleaner fix than the inline `rootUsername` veto: ambiguous bridge rows are isolated order-independently, and the forward/reverse test locks that in.

## 👏 Nice work

- **`splitConflictingComponent`** — post-union split by hard identifier, floaters stay separate instead of being guessed into whichever username unioned first (addresses @copilot-pull-request-reviewer's order-dependence concern)
- **`pickNonBlank` + `pickMergedEmail`** — blank-string sentinel and invited-vs-attended email preference both handled with regression tests
- **Conflict coverage expanded** — username-bridge isolation (order-independent), guest-email-through-name-bridge, and blank-email merge tests
- **Docs & mocks** — union-find rationale, PCC-parity tradeoff on name-only fallback, and the intentional light double on `isUnresolvableParticipantName`

## Revision tracking (since `402b8ce`)

| Prior item | Status | Resolving commit |
|---|---|---|
| Transitive username conflict via email bridge | ✅ Resolved | `d038417` + `f1997cdd` (`splitConflictingComponent`) |
| Empty-string `??` email merge | ✅ Resolved | `d038417` (`pickNonBlank`) |
| Missing regression tests | ✅ Resolved | `d038417` + `f1997cdd` |
| Union-find inline comments | ✅ Resolved | `d038417` |
| Spec stub for `isUnresolvableParticipantName` | ✅ Resolved | `d038417` |
| Name-only merge ambiguity | ✅ Acknowledged | `d038417` JSDoc — intentional PCC parity |
| Copilot: encounter-order bridge attribution | ✅ Resolved | `f1997cdd` |

## Open items

- 🔴 **Blocking: 0**
- 🟡 **Minor: 0**
- ⚪ **Nit: 2** — duplicate adjacent JSDoc blocks on `isSamePerson` / `splitConflictingComponent`; path-halving comment still says "near O(1)" without union-by-rank (immaterial given the O(n²) outer loop)

## Bot reconciliation

| Bot | Topic | Position |
|-----|-------|----------|
| Copilot | Encounter-order bridge attribution | ✅ Resolved in `f1997cdd` |
| Copilot/Cursor (round 1) | Username bridge / empty email | ✅ Resolved |

## Final decision

PR #2064, opened 2026-09-02
fix(cla): render sign date in the viewer's local timezone

### LINKED ISSUE #2032: BUG: Sign Date in My CLAs shows UTC date, off by one day for negative-offset timezones
Part of linuxfoundation/lfx-self-serve#1733

### Summary

The **Sign Date** shown in the **My CLAs** tab can be off by one day. A CLA signed on the afternoon/evening of Sept 1st (US Pacific) displays as **Sep 2**. The date appears to be rendered in UTC rather than in the viewer's local timezone.

### Steps to reproduce

1. Set your machine's timezone to a negative UTC offset (e.g. `America/Los_Angeles`).
2. Sign a CLA (or use an existing CLA whose stored signature timestamp falls between 00:00–07:00 UTC — i.e. the prior afternoon/evening in Pacific time).
3. Go to **Profile → My CLAs**.
4. Look at the **Signed** column for that CLA.

### Expected

The Sign Date reflects the date the user actually signed, in their own local timezone — Sep 1 for a signature made the afternoon of Sept 1st Pacific.

### Actual

The Sign Date is rendered one day ahead — Sep 2 — because the timestamp is displayed in UTC.

### Potential root cause

Not confirmed — needs validation by whoever picks this up.

The Signed column pins the date to UTC explicitly:

`apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.html:145`

```
{{ row.agreement.signedOn ? (row.agreement.signedOn | date: 'mediumDate' : 'UTC') : '—' }}
```

`: 'UTC'` is a repo-wide convention used across other date fields (donations, memberships, events). That convention is correct for **date-only** values (`2026-05-01`), where it prevents the opposite off-by-one. But `signedOn` here is a **full timestamp**, so pinning to UTC produces exactly the shift reported above.

Note on the field's type: the interface doc comment calls `signedOn` an "ISO date" and the test fixtures all use bare `YYYY-MM-DD` values, which suggests date-only. The production data contradicts that — the stored values carry a time component, and a date-only value could not produce the reported Sep 1 → Sep 2 shift. The fixtures appear unrepresentative of real API responses and are worth correcting alongside the fix.

Worth noting: the value is passed through the server unchanged (`apps/lfx-one/src/server/services/cla.service.ts:353` — `signedOn: cla.signedOn ?? ''`), and both timestamp shapes observed in DynamoDB parse correctly in JS, so this does not look like a parsing problem:

| Stored value | Parses to | Rendered UTC | Rendered America/Los_Angeles |
| --- | --- | --- | --- |
| `2026-05-01T18:40:42Z` | valid | May 1, 2026 | May 1, 2026 |
| `2026-05-08T23:24:50.232159+00:00` | valid | May 8, 2026 | May 8, 2026 |
| `2026-09-01T17:30:00-07:00` | valid | **Sep 2, 2026** | Sep 1, 2026 |

### Scope: UI presentation only

**This fix is limited to how the date is presented in Self Serve. No backend, API, or data change.**

The stored value is already correct and unambiguous at every layer below the UI:

- DynamoDB stores an explicit UTC offset in both observed shapes, so no information is lost.
- The server passes the value through untouched (`cla.service.ts:353` is only a `?? ''` null-guard).
- Both shapes parse correctly in JS (see table above) — this is not a parsing problem.

The only defective step is the final render, so the change belongs in the Self Serve template/presentation layer.

### Notes / open questions for the assignee

- Decide whether Sign Date should render in the **viewer's local timezone** or stay UTC-consistent for everyone; local is what this report asks for, but it means two users can legitimately see different dates for the same signature.
- If local is chosen, consider whether to also show the time or a timezone indicator so the displayed date is unambiguous.
- **Review all other EasyCLA dates shown in Self Serve for the same issue.** Today `signedOn` is the only CLA date actually rendered — there is no `invalidatedOn` / `revokedOn` / `expiresOn` field on `MyClaAgreement` yet, and status transitions are carried only as a `ClaStatus` enum with no accompanying timestamp. So this is mainly a forward-looking check: when #1913 adds the invalidated/revoked dates, they must not repeat this pattern. Re-check the CLA surface at fix time in case new date fields have landed in the meantime.
- **Do not fix this with a find-and-replace on `: 'UTC'`.** That argument is correct everywhere else in the repo (donations, memberships, events), where the values are genuinely date-only and removing the pin would *introduce* an off-by-one. Only fields that carry a real time component should change.

### Environment

- Repo: `linuxfoundation/lfx-self-serve`
- Page: Profile → My CLAs
- Reporter timezone: US Pacific (UTC-7)

Image

ahmedomosanya:
## Summary
- My CLAs Sign Date was pinned to UTC, so a Pacific afternoon signature displayed as the next calendar day.
- Format the producer RFC3339 instant in the viewer's local timezone (date-only, same shape as `mediumDate`).
- Pin the reported Pacific vs UTC split in shared unit tests.

## Test plan
- [ ] Unit: `2026-09-01T17:30:00-07:00` → `Sep 1, 2026` in `America/Los_Angeles`, `Sep 2, 2026` in UTC
- [ ] Profile → My CLAs with the machine in a negative offset: a signature from the prior afternoon still shows that local date
- [ ] Empty `signedOn` still shows —

Closes #2032

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
@@ -1,7 +1,7 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { DatePipe, DOCUMENT, isPlatformBrowser } from '@angular/common';
+import { DOCUMENT, isPlatformBrowser } from '@angular/common';
 import { ChangeDetectionStrategy, Component, computed, DestroyRef, inject, PLATFORM_ID, Signal, signal, viewChildren } from '@angular/core';
 import { takeUntilDestroyed, toSignal } from '@angular/core/rxjs-interop';
 import { Router, RouterLink } from '@angular/router';
@@ -25,6 +25,7 @@ import {
   claStatusLabel,
   claStatusSeverity,
   downloadFromUrl,
+  formatClaSignedOn,
   gerritSignUrl,
   isMyClasEmpty,
   signedAsLine,
@@ -66,18 +67,7 @@ import { SignIdentitySelectComponent } from './sign-identity-select.component';
  */
 @Component({
   selector: 'lfx-profile-clas',
-  imports: [
-    DatePipe,
-    RouterLink,
-    BadgeComponent,
-    ButtonComponent,
-    EmptyStateComponent,
-    MenuComponent,
-    MessageComponent,
-    TableComponent,
-    TagComponent,
-    ToastModule,
-  ],
+  imports: [RouterLink, BadgeComponent, ButtonComponent, EmptyStateComponent, MenuComponent, MessageComponent, TableComponent, TagComponent, ToastModule],
   providers: [MessageService, DialogService],
   templateUrl: './profile-clas.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
@@ -663,6 +653,7 @@ export class ProfileClasComponent {
             icon: this.statusIcon(agreement.status),
             note: this.statusNote(agreement),
           },
+          signedOnLabel: formatClaSignedOn(agreement.signedOn) || '—',
           signedAsLine: m2 ? signedAsLine(agreement.signedVia, agreement.signedAs) : undefined,
           menuItems,
           hasActions: menuItems.length > 0,
--- a/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
@@ -653,7 +653,7 @@ export class ProfileClasComponent {
             icon: this.statusIcon(agreement.status),
             note: this.statusNote(agreement),
           },
-          signedOnLabel: formatClaSignedOn(agreement.signedOn) || '—',
+          signedOnLabel: formatClaSignedOn(agreement.signedOn),
           signedAsLine: m2 ? signedAsLine(agreement.signedVia, agreement.signedAs) : undefined,
           menuItems,
           hasActions: menuItems.length > 0,
--- a/packages/shared/src/interfaces/cla.interface.ts
+++ b/packages/shared/src/interfaces/cla.interface.ts
@@ -86,7 +86,8 @@ export interface MyClaAgreement {
   companyName?: string;
   /**
    * RFC3339 instant the agreement was signed. A bare `YYYY-MM-DD` is accepted
-   * and rendered as that UTC calendar day.
+   * and rendered as that UTC calendar day. Empty string when the producer sent
+   * no date (`cla.service` normalizes the absent field).
    */
   signedOn: string;
   /**
@@ -486,7 +487,11 @@ export interface ClaRow {
   id: string;
   agreement: MyClaAgreement;
   status: ClaRowStatus;
-  /** Date-only Sign Date in the viewer's local timezone; `'—'` when empty or unparseable. */
+  /**
+   * Sign Date: an instant renders in the viewer's local timezone, a bare
+   * `YYYY-MM-DD` as that UTC calendar day; `'—'` when empty, unparseable, or
+   * an impossible calendar date.
+   */
   signedOnLabel: string;
   /** Second line under the signed date; absent when the producer sent no identity. */
   signedAsLine?: string;
--- a/packages/shared/src/interfaces/cla.interface.ts
+++ b/packages/shared/src/interfaces/cla.interface.ts
@@ -483,6 +483,8 @@ export interface ClaRow {
   id: string;
   agreement: MyClaAgreement;
   status: ClaRowStatus;
+  /** Date-only Sign Date in the viewer's local timezone; `'—'` when empty or unparseable. */
+  signedOnLabel: string;
   /** Second line under the signed date; absent when the producer sent no identity. */
   signedAsLine?: string;
   menuItems: ClaRowMenuItem[];
--- a/packages/shared/src/interfaces/cla.interface.ts
+++ b/packages/shared/src/interfaces/cla.interface.ts
@@ -84,7 +84,10 @@ export interface MyClaAgreement {
   claManager?: boolean;
   /** Employer company name — present for ECLA only. */
   companyName?: string;
-  /** ISO date the agreement was signed. */
+  /**
+   * RFC3339 instant the agreement was signed. A bare `YYYY-MM-DD` is accepted
+   * and rendered as that UTC calendar day.
+   */
   signedOn: string;
   /**
    * Platform this agreement was signed via, when the producer sent one.
--- a/packages/shared/src/utils/cla-view.utils.spec.ts
+++ b/packages/shared/src/utils/cla-view.utils.spec.ts
@@ -30,7 +30,7 @@ import {
 } from './cla-view.utils';
 
 function agreement(overrides: Partial<MyClaAgreement> = {}): MyClaAgreement {
-  return { id: 's1', kind: 'ICLA', claGroupName: 'P', signedOn: '2022-01-01', status: 'valid', pdfAvailable: true, ...overrides };
+  return { id: 's1', kind: 'ICLA', claGroupName: 'P', signedOn: '2022-01-01T18:40:42Z', status: 'valid', pdfAvailable: true, ...overrides };
 }
 
 function identity(overrides: Partial<MyClasIdentitySummary> = {}): MyClasIdentitySummary {
@@ -172,10 +172,34 @@ describe('formatClaSignedOn', () => {
     expect(formatClaSignedOn('2026-05-08T23:24:50.232159+00:00', 'UTC')).toBe('May 8, 2026');
   });
 
-  it('returns empty for missing, blank, or unparseable values so the cell can show an em dash', () => {
-    expect(formatClaSignedOn('')).toBe('');
-    expect(formatClaSignedOn('   ')).toBe('');
-    expect(formatClaSignedOn('not-a-date')).toBe('');
+  it('pins a bare YYYY-MM-DD to UTC so a negative-offset host does not shift the calendar day', () => {
+    expect(formatClaSignedOn('2022-01-01')).toBe('Jan 1, 2022');
+    expect(formatClaSignedOn('2022-01-01', 'America/Los_Angeles')).toBe('Jan 1, 2022');
+  });
+
+  it('omits timeZone on the production no-argument path so a UTC re-pin fails this test', () => {
+    const seen: (Intl.DateTimeFormatOptions | undefined)[] = [];
+    const original = Date.prototype.toLocaleDateString;
+    // eslint-disable-next-line no-extend-native
+    Date.prototype.toLocaleDateString = function (locales?: unknown, options?: Intl.DateTimeFormatOptions): string {
+      seen.push(options);
+      return original.call(this, locales as string, options);
+    };
+    try {
+      formatClaSignedOn(afternoonPacific);
+    } finally {
+      // eslint-disable-next-line no-extend-native
+      Date.prototype.toLocaleDateString = original;
+    }
+
+    expect(seen.length).toBeGreaterThan(0);
+    expect(seen.every((o) => o?.timeZone === undefined)).toBe(true);
+  });
+
+  it('returns an em dash for missing, blank, or unparseable values', () => {
+    expect(formatClaSignedOn('')).toBe('—');
+    expect(formatClaSignedOn('   ')).toBe('—');
+    expect(formatClaSignedOn('not-a-date')).toBe('—');
   });
 });
 
--- a/packages/shared/src/utils/cla-view.utils.spec.ts
+++ b/packages/shared/src/utils/cla-view.utils.spec.ts
@@ -193,13 +193,19 @@ describe('formatClaSignedOn', () => {
     }
 
     expect(seen.length).toBeGreaterThan(0);
-    expect(seen.every((o) => o?.timeZone === undefined)).toBe(true);
+    expect(seen.at(-1)?.timeZone).toBeUndefined();
   });
 
   it('returns an em dash for missing, blank, or unparseable values', () => {
     expect(formatClaSignedOn('')).toBe('—');
     expect(formatClaSignedOn('   ')).toBe('—');
     expect(formatClaSignedOn('not-a-date')).toBe('—');
+  });
+
+  it('refuses impossible calendar dates rather than rolling them over', () => {
+    expect(formatClaSignedOn('2026-02-31')).toBe('—');
+    expect(formatClaSignedOn('2026-02-31T10:00:00Z')).toBe('—');
+    expect(formatClaSignedOn('0001-01-01')).toBe('—');
   });
 });
 
--- a/packages/shared/src/utils/cla-view.utils.spec.ts
+++ b/packages/shared/src/utils/cla-view.utils.spec.ts
@@ -20,6 +20,7 @@ import {
   claSignRoute,
   claStatusLabel,
   claStatusSeverity,
+  formatClaSignedOn,
   gerritSignUrl,
   isMyClasEmpty,
   shouldShowGithubCta,
@@ -151,6 +152,30 @@ describe('claStatusSeverity', () => {
     expect(claStatusSeverity('invalidated')).toBe('danger');
     expect(claStatusSeverity('unknown')).toBe('secondary');
     expect(claStatusSeverity('superseded')).toBe('warn');
+  });
+});
+
+describe('formatClaSignedOn', () => {
+  // The three shapes from #2032. UTC pin vs local is a no-op on the first two; the third
+  // is the reported off-by-one (Pacific afternoon → next UTC calendar day).
+  const afternoonPacific = '2026-09-01T17:30:00-07:00';
+
+  it('renders a Pacific-afternoon timestamp as Sep 1 locally and Sep 2 in UTC', () => {
+    expect(formatClaSignedOn(afternoonPacific, 'America/Los_Angeles')).toBe('Sep 1, 2026');
+    expect(formatClaSignedOn(afternoonPacific, 'UTC')).toBe('Sep 2, 2026');
+  });
+
+  it('keeps same-calendar-day timestamps on the same date in Pacific and UTC', () => {
+    expect(formatClaSignedOn('2026-05-01T18:40:42Z', 'America/Los_Angeles')).toBe('May 1, 2026');
+    expect(formatClaSignedOn('2026-05-01T18:40:42Z', 'UTC')).toBe('May 1, 2026');
+    expect(formatClaSignedOn('2026-05-08T23:24:50.232159+00:00', 'America/Los_Angeles')).toBe('May 8, 2026');
+    expect(formatClaSignedOn('2026-05-08T23:24:50.232159+00:00', 'UTC')).toBe('May 8, 2026');
+  });
+
+  it('returns empty for missing, blank, or unparseable values so the cell can show an em dash', () => {
+    expect(formatClaSignedOn('')).toBe('');
+    expect(formatClaSignedOn('   ')).toBe('');
+    expect(formatClaSignedOn('not-a-date')).toBe('');
   });
 });
 
--- a/packages/shared/src/utils/cla-view.utils.ts
+++ b/packages/shared/src/utils/cla-view.utils.ts
@@ -114,6 +114,8 @@ export function claStatusSeverity(status: ClaStatus): TagSeverity {
   }
 }
 
+const CLA_SIGNED_ON_DATE_ONLY = /^\d{4}-\d{2}-\d{2}$/;
+
 /**
  * Date-only Sign Date for My CLAs (#2032).
  *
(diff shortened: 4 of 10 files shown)


dealako:
Hi @ahmedomosanya — thanks for the tight, well-scoped fix. 👋

The core change is correct: dropping the UTC pin and formatting the RFC3339 instant in the viewer's local zone resolves the reported Pacific-afternoon → next-day shift, the SSR path never renders a date (`initState()` skips the fetch off-browser, so there's no hydration mismatch), and `signedOn` is a required `string` from the BFF mapper so `iso.trim()` is safe. The new util is pure, the JSDoc is genuinely helpful, and precomputing `signedOnLabel` in the memoized `rows()` computed is performance-neutral. No blocking, security, or privacy defects.

What holds this back from an approval is that #2032 asked for two things and the PR did one. It asked to render in the viewer's local zone (done, correctly) *and* to correct the unrepresentative bare-date fixtures — and the fixtures still standardize on `signedOn: '2022-01-01'`, the one date shape this formatter now renders one day early for negative-offset viewers. Because every assertion passes an explicit `timeZone` and CI runs UTC, the actual production call path (`formatClaSignedOn(agreement.signedOn)`, no timezone) is never exercised, so a future re-pin to UTC would ship green.

### Issue summary

- 🔴 Blocking: 0
- 🟡 Minor: 3
  - Reverse off-by-one for a bare `YYYY-MM-DD` sign date (mirror of the fixed bug) on a barrel-exported util documented to take an "ISO date".
  - Tests never cover the date-only shape or the no-argument production path, and the unrepresentative fixtures #2032 called out are unchanged — the regression guard doesn't bind on the UTC CI host.
  - The SSR precondition that makes omitting `timeZone` safe is undocumented; this is the only shared date formatter in the package that omits it, and four siblings pin UTC for the opposite reason.
- ⚪ Nit: 1
  - The `—` placeholder is split across packages (`|| '—'` in the component) while the shared doc states it as a property of the field; `claStatusLabel` keeps its dash in the util.
- ❔ Question: 0

Copilot reviewed all 5 files and generated no inline comments; Cursor Bugbot posted a low-risk presentation-only summary with no findings. I agree it's low-risk and non-security — the points above are deeper correctness/test-representativeness observations neither bot raised, and none of them contradict the bots.

### Decision

🔴 **Needs changes before approval** — none blocking on their own, but the fixture/test-representativeness gap is exactly what #2032 asked for, and the reverse off-by-one leaves a legal-agreement date renderable one day wrong for an input the type and every fixture say is plausible. All four are cheap to address. If you can show `signedOn` is contractually always a full instant, tightening the type/JSDoc to say so (instead of handling the date-only shape) is an equally good resolution of the first point.

packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -112,6 +112,31 @@ export function claStatusSeverity(status: ClaStatus): TagSeverity {
...
+ * name in tests so CI (UTC) can still pin the Pacific case.
+ *
+ * Empty / unparseable → `''` so the cell can show an em dash.
+ */
+export function formatClaSignedOn(iso: string, timeZone?: string): string {
+  const trimmed = iso.trim();
+  if (!trimmed) return '';
+  const date = new Date(trimmed);
+  if (Number.isNaN(date.getTime())) return '';
+  return date.toLocaleDateString('en-US', {
```
dealako, line 156: [minor] Reverse off-by-one for a bare `YYYY-MM-DD` sign date

**Issue:** Omitting `timeZone` is correct for a full RFC3339 instant but flips the sign for a date-only value. A bare calendar-day `signedOn` now renders one day *earlier* for viewers in negative UTC offsets — the mirror of the bug this PR fixes.

**Proof:** Per ECMA-262 a date-only ISO string parses as UTC midnight, so `new Date('2022-01-01')` is `2022-01-01T00:00:00Z`. `formatClaSignedOn('2022-01-01')` returns `Dec 31, 2021` in `America/Los_Angeles` (and `Jan 1, 2022` in UTC). The removed template `| date: 'mediumDate' : 'UTC'` rendered `Jan 1, 2022` correctly. The interface still documents the field as “ISO date” (`cla.interface.ts:87`) and every `MyClaAgreement` fixture in the repo uses the bare-date shape (`cla-view.utils.spec.ts:26`).

**Why it matters:** #2032 states production `signedOn` carries a time component, so this is not the reported case — but the util is barrel-exported and documented to take an “ISO date”, and a legacy/date-only record would silently show the wrong day on a signed legal agreement, invisibly on the UTC CI runner.

**Fix:** Either handle the date-only shape by pinning it to UTC, or tighten the type/JSDoc so `signedOn` is guaranteed a full instant:
```ts
const DATE_ONLY = /^\d{4}-\d{2}-\d{2}$/;
const zone = DATE_ONLY.test(trimmed) ? 'UTC' : timeZone;
// ...spread { timeZone: zone } when zone is set
```

packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -112,6 +112,31 @@ export function claStatusSeverity(status: ClaStatus): TagSeverity {
...
+ * Producer `signedOn` is a real RFC3339 instant, not a calendar-day stored as UTC
+ * midnight. Pinning the formatter to UTC (the DatePipe `'UTC'` argument, or
+ * `{ timeZone: 'UTC' }` on `toLocaleDateString`) shifts the calendar day for
+ * viewers in negative offsets — a Pacific afternoon signature displays as the
+ * next UTC date. Omitting `timeZone` uses the viewer's local zone; pass an IANA
+ * name in tests so CI (UTC) can still pin the Pacific case.
+ *
+ * Empty / unparseable → `''` so the cell can show an em dash.
+ */
+export function formatClaSignedOn(iso: string, timeZone?: string): string {
```
dealako, line 140: [minor] Record the SSR precondition that makes host-zone formatting safe here

**Issue:** The JSDoc explains why UTC pinning is wrong for an instant, but never states the precondition that makes *omitting* `timeZone` safe. This is the only shared date formatter in the package that omits `timeZone`; four siblings pin UTC for the opposite reason.

**Proof:** `formatHubSpotUpdatedAt` (`date-time.utils.ts`) documents verbatim that “without an explicit `timeZone`, `toLocaleDateString` renders in the HOST zone, and this is an SSR app… the row text and its accessible label both change during hydration”; `formatIsoDateLabel`, `formatShortDate`, and `formatWeekRangeLabel` pin UTC for the same reason. The CLAs table is safe only because `initState()` (`profile-clas.component.ts:590-595`) returns `{ data: null, loaded: false }` off-browser, so `rows()` yields nothing under SSR. Nothing in this util's doc or at the call site records that dependency.

**Why it matters:** A future caller from an SSR-rendered path — or a change that makes these rows pre-render (TransferState, a resolver, repairing the SSR fetch) — reintroduces a hydration text mismatch with no warning, against a barrel-exported public util.

**Fix:** Add one sentence to the JSDoc: safe here only because the CLAs rows never render under SSR; do not use from a server-rendered path without passing an explicit `timeZone` (see `formatHubSpotUpdatedAt`).

packages/shared/src/utils/cla-view.utils.spec.ts
```diff
@@ -147,6 +148,30 @@ describe('claStatusSeverity', () => {
...
+    expect(formatClaSignedOn('2026-05-01T18:40:42Z', 'America/Los_Angeles')).toBe('May 1, 2026');
+    expect(formatClaSignedOn('2026-05-01T18:40:42Z', 'UTC')).toBe('May 1, 2026');
+    expect(formatClaSignedOn('2026-05-08T23:24:50.232159+00:00', 'America/Los_Angeles')).toBe('May 8, 2026');
+    expect(formatClaSignedOn('2026-05-08T23:24:50.232159+00:00', 'UTC')).toBe('May 8, 2026');
+  });
+
+  it('returns empty for missing, blank, or unparseable values so the cell can show an em dash', () => {
+    expect(formatClaSignedOn('')).toBe('');
+    expect(formatClaSignedOn('   ')).toBe('');
+    expect(formatClaSignedOn('not-a-date')).toBe('');
```
dealako: [minor] Cover the date-only shape and the no-argument production path; correct the unrepresentative fixtures

**Issue:** This block asserts full-instant and empty/invalid inputs, but never (a) the bare `YYYY-MM-DD` shape — the shape every `MyClaAgreement` fixture in the repo uses — nor (b) the zero-argument production call `formatClaSignedOn(agreement.signedOn)` (`profile-clas.component.ts:623`). Every assertion here passes an explicit `timeZone`.

**Proof:** Fixtures use `signedOn: '2022-01-01'` (`cla-view.utils.spec.ts:26`, and the same in `profile-clas.component.spec.ts`, `cla.service.spec.ts`, etc.). Because the options spread is last (`cla-view.utils.ts:136`) an explicit argument overrides any default, so adding `timeZone: 'UTC'` back to the object would re-introduce #2032 while keeping all six assertions green on the UTC CI runner. #2032 explicitly asked: “The fixtures appear unrepresentative of real API responses and are worth correcting alongside the fix.” They are unchanged.

**Why it matters:** The one shape the repo's test data standardizes on has no contract pinned to it, and the #2032 regression guard does not bind on CI — a future re-pin to UTC ships green.

**Fix:** Add a bare-date case, add a no-argument/option-observing assertion (see `hubspot-updated-at.spec.ts` for the `toLocaleDateString` option-spy pattern), and move the fixtures onto representative instants (e.g. `2022-01-01T18:40:42Z`). Consider asserting the `—` fallback once at the component level.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -622,6 +620,7 @@ export class ProfileClasComponent {
             icon: this.statusIcon(agreement.status),
             note: this.statusNote(agreement),
           },
+          signedOnLabel: formatClaSignedOn(agreement.signedOn) || '—',
```
dealako: [nit] Keep the em-dash placeholder with the shared util, like `claStatusLabel`

**Issue:** The `—` placeholder for this feature lives in the component (`|| '—'`) while `claStatusLabel` returns `—` itself (`cla-view.utils.ts:90-91`), and the shared doc on `ClaRow.signedOnLabel` (`cla.interface.ts:440`) states “`'—'` when empty or unparseable” as a property of the field.

**Proof:** `formatClaSignedOn` returns `''`; the dash is supplied app-side. The shared-package guarantee holds only because `initRows()` is the sole `ClaRow` producer — a second producer satisfies the `string` type while silently breaking the documented guarantee with an empty cell.

**Why it matters:** One presentational decision split across two packages; the shared doc is upheld by app code the type system does not require.

**Fix:** Return `'—'` from the util (matching `claStatusLabel('unknown')`) and drop `|| '—'`, or reword `cla.interface.ts:440` to say the component supplies the dash.

dealako (changes requested):
Core fix is correct and resolves #2032 for the reported case. Requesting changes on three minor points and one nit, detailed inline. The most concrete is that #2032 asked to correct the unrepresentative bare-date fixtures and add coverage for the shape this formatter actually changes — the production no-argument call path stays untested, so a re-pin to UTC would ship green on the UTC CI host. None are blocking on their own; all are cheap. Full breakdown in the summary comment.

packages/shared/src/utils/cla-view.utils.ts
ahmedomosanya: Addressed in `0a106fb7e` — a bare `YYYY-MM-DD` is now pinned to UTC so a negative-offset host keeps that calendar day; full instants still omit `timeZone`.

packages/shared/src/utils/cla-view.utils.ts
ahmedomosanya: Addressed in `0a106fb7e` — JSDoc now records that omitting `timeZone` is safe only because CLAs rows never render under SSR, and warns callers not to use this from a server-rendered path without an explicit zone.

packages/shared/src/utils/cla-view.utils.spec.ts
ahmedomosanya: Addressed in `0a106fb7e` — added a bare-date case, a no-argument `toLocaleDateString` option spy, the em-dash fallback at the component, and moved `MyClaAgreement` fixtures onto representative instants.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
ahmedomosanya: Addressed in `0a106fb7e` — `formatClaSignedOn` now returns `—` itself (matching `claStatusLabel`), and the component dropped the `|| '—'` fallback.

packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -112,6 +112,41 @@ export function claStatusSeverity(status: ClaStatus): TagSeverity {
...
+ * returns unloaded off-browser). Do not call from a server-rendered path
+ * without passing an explicit `timeZone` (see `formatHubSpotUpdatedAt`).
+ *
+ * Empty / unparseable → `'—'` (same placeholder as `claStatusLabel('unknown')`).
+ */
+export function formatClaSignedOn(iso: string, timeZone?: string): string {
+  const trimmed = iso.trim();
+  if (!trimmed) return '—';
+  const date = new Date(trimmed);
+  if (Number.isNaN(date.getTime())) return '—';
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `52d2e0042` — date-only and timestamp forms now round-trip the calendar day through `formatIsoDateLabel` before formatting, so Feb 31 and year 0001 render `—` instead of a rolled-over date.

dealako:
Hi @ahmedomosanya 👋 — thanks for the thorough turnaround. Every point from the last round is resolved, and the way you did it is genuinely clean.

### 👏 Nice work

- The reverse off-by-one is fixed the right way: a bare `YYYY-MM-DD` now routes through the UTC-pinned `formatIsoDateLabel`, while full instants keep the viewer's local zone. Verified `2022-01-01` → `Jan 1, 2022` on a Pacific host.
- You reused the `formatHubSpotUpdatedAt` sibling pattern (shared `formatIsoDateLabel`, date-part round-trip) instead of inventing a new one, and the impossible-date guard in `52d2e0042` correctly rejects Feb 31 / year-0001 in both date-only and timestamp shapes while every valid instant survives (including the 2024 leap day and the #2032 case).
- The test gap is closed exactly where it mattered: the no-argument production path is now pinned by the `toLocaleDateString` option spy, the bare-date shape is asserted host-independently, the `signedOn` fixtures are representative instants repo-wide, and there's a real component-level em-dash test through a new `data-testid`.
- The `—` placeholder now lives in the util, matching `claStatusLabel`.

### Revision status — all resolved

- ✅ Reverse off-by-one on bare dates → `0a106fb7e` (UTC pin via `formatIsoDateLabel` + interface doc)
- ✅ Test coverage / unrepresentative fixtures → `0a106fb7e`
- ✅ SSR precondition undocumented → `0a106fb7e`
- ✅ Em-dash split across packages → `0a106fb7e`
- ✅ Copilot's impossible-date (Feb 31 → Mar 3) → `52d2e0042`

### Bot reconciliation

Copilot's one inline finding (impossible dates rolling over) is addressed in `52d2e0042`, and its re-review generated no new comments. Agree on both — my correctness pass independently confirmed the round-trip guard.

### Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 — both doc comments in `packages/shared/src/interfaces/cla.interface.ts` (inline):
  1. `ClaRow.signedOnLabel` (line 442) still reads "viewer's local timezone … empty or unparseable"; after these commits a bare date is UTC-pinned and an impossible date also collapses to `—`, so it now contradicts the `signedOn` doc just above it.
  2. `MyClaAgreement.signedOn` (lines 88-90) now reads as an exhaustive shape contract but omits the empty string the BFF emits (`cla.service.ts:350` `?? ''`).

### Decision

🔴 **Needs changes before approval** — but only because the branch currently has a **merge conflict with `main`** in `apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts` (`mergeable: CONFLICTING`). The review itself is effectively clean. Rebase on `main` to clear the conflict; the two doc nits are take-it-or-leave-it. Once the branch merges cleanly I'm glad to flip this to an approval.

packages/shared/src/interfaces/cla.interface.ts
```diff
@@ -84,7 +84,10 @@ export interface MyClaAgreement {
   claManager?: boolean;
   /** Employer company name — present for ECLA only. */
   companyName?: string;
-  /** ISO date the agreement was signed. */
+  /**
+   * RFC3339 instant the agreement was signed. A bare `YYYY-MM-DD` is accepted
+   * and rendered as that UTC calendar day.
```
dealako: [nit] Document the empty-string `signedOn` the BFF actually emits

**Issue:** The rewritten doc now enumerates the accepted shapes (“RFC3339 instant … A bare `YYYY-MM-DD` is accepted and rendered as that UTC calendar day.”), which reads as an exhaustive contract, but it omits the third shape the producer provably emits: the empty string.

**Proof:** `cla.service.ts:350` maps `signedOn: cla.signedOn ?? ''`, and the wire field is optional (`cla.types.ts:45` `signedOn?: string`), so an upstream record with no signature date reaches the client as `''`. That path is live and tested: `formatClaSignedOn` short-circuits `''` to `—` (`cla-view.utils.ts:140`), `cla-view.utils.spec.ts:193` asserts it, and the new component test at `profile-clas.component.spec.ts:469` renders a row with `signedOn: ''`.

**Why it matters:** The old one-line doc made no shape claim, so its silence lost nothing. The new text implies `signedOn` is always a real date, so a future consumer of this shared interface could reasonably drop the empty guard `formatClaSignedOn` currently carries.

**Fix:** Add a clause, e.g. a third sentence: “Empty string when the producer sent no date (`cla.service.ts` normalizes the absent field).”

dealako (changes requested):
All five items from the last round are resolved and the fix is clean and well-tested (details and acknowledgement in the summary comment). This is Changes Requested for one reason only: the branch has a merge conflict with `main` in `apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts` (`mergeable: CONFLICTING`), which blocks merge. Rebase on `main` to clear it and I'll flip to approval.

Two optional documentation nits remain, both in `packages/shared/src/interfaces/cla.interface.ts`:

- One is inline below (the `MyClaAgreement.signedOn` doc).
- The other cannot be anchored inline because its line is unchanged in this diff: the `ClaRow.signedOnLabel` doc at line 442 still reads “Date-only Sign Date in the viewer's local timezone; `'—'` when empty or unparseable.” These two commits invalidated both halves — a bare `YYYY-MM-DD` is now UTC-pinned (not the viewer's zone), and an impossible date such as `2026-02-31` collapses to `—` yet is not “unparseable” (`new Date` rolls it to Mar 3). It now contradicts the `signedOn` doc just above it. Suggested rewrite: “Sign Date: an instant renders in the viewer's local timezone, a bare `YYYY-MM-DD` as that UTC calendar day; `'—'` when empty, unparseable, or an impossible calendar date.” Both nits are take-it-or-leave-it.

dealako:
Hi @ahmedomosanya 👋 — thanks for the quick turnaround on the rebase.

### 👏 What changed since the last round

You rebased the branch onto the current `main`, which clears the one thing that was holding this back: the merge conflict in `apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts` is gone and the PR is `MERGEABLE` again. I verified the conflict resolution is clean, the branch's net change to that file is byte-for-byte the same one already reviewed (drop `DatePipe`, add `formatClaSignedOn`, `signedOnLabel: formatClaSignedOn(agreement.signedOn)`), replayed correctly above main's new code with no behavior change. No CLA feature content changed in the rebase, so the earlier subagent review still stands.

### Revision status — all resolved

- ✅ Reverse off-by-one on bare dates → UTC pin via `formatIsoDateLabel`
- ✅ Test coverage / representative fixtures / no-arg guard
- ✅ SSR precondition documented on the util
- ✅ Em-dash placeholder owned by the util
- ✅ Copilot's impossible-date guard (Feb 31 → `—`)
- ✅ **Merge conflict with `main`** → resolved by the rebase

### Issue count

- 🔴 Blocking: 0
- 🟡 Minor: 0
- ⚪ Nit: 2 (carried over, both optional docs in `packages/shared/src/interfaces/cla.interface.ts`):
  1. `ClaRow.signedOnLabel` (line 490) still reads "viewer's local timezone … empty or unparseable"; bare dates are UTC-pinned and impossible dates also collapse to `—`, so it contradicts the `signedOn` doc above it.
  2. `MyClaAgreement.signedOn` (lines 88-89) reads as an exhaustive shape contract but omits the empty string the BFF emits (`cla.service.ts` `?? ''`).

### Decision

✅ **Approved with minor comments.** The sole blocker from the last round (the merge conflict) is resolved, and every code finding was already addressed. The two doc nits are take-it-or-leave-it, fold them into the next push if convenient. Nice work on this one.

packages/shared/src/interfaces/cla.interface.ts
ahmedomosanya: Addressed in `84df74a3c` — `signedOn` now names the empty-string BFF shape, and `signedOnLabel` matches (viewer-local instant, UTC-pinned bare date, `—` for empty / unparseable / impossible). The `main` conflict is also cleared; the branch rebases cleanly.

dealako (approved):
Approved. The rebase onto `main` cleared the only blocker from my last round — the merge conflict in `apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts` is resolved and the PR is `MERGEABLE`. I verified the conflict resolution is clean: the branch's net change to that file is identical to the one already reviewed (drop `DatePipe`, add `formatClaSignedOn`, `signedOnLabel: formatClaSignedOn(agreement.signedOn)`), with no CLA behavior change, so the prior subagent review still holds. Every code finding from the earlier rounds is resolved (reverse off-by-one, test coverage/fixtures, SSR doc, em-dash ownership, and Copilot's impossible-date guard).

Two optional documentation nits remain in `packages/shared/src/interfaces/cla.interface.ts`, both carried over from the last round and non-blocking:

- `ClaRow.signedOnLabel` (line 490) still reads “Date-only Sign Date in the viewer's local timezone; `'—'` when empty or unparseable.” After the fix a bare `YYYY-MM-DD` is UTC-pinned (not the viewer's zone) and an impossible date such as `2026-02-31` collapses to `—` yet is not “unparseable”, so it now contradicts the `signedOn` doc above it. Suggested: “Sign Date: an instant renders in the viewer's local timezone, a bare `YYYY-MM-DD` as that UTC calendar day; `'—'` when empty, unparseable, or an impossible calendar date.”
- `MyClaAgreement.signedOn` (lines 88-89) reads as an exhaustive shape contract but omits the empty string the BFF emits (`cla.service.ts` `?? ''`). Suggested: add “Empty string when the producer sent no date.”

Fold them into the next push if convenient. Nice work.

Issue #2066, opened 2026-09-02
Unable to select Gerrit project CCLA


WillsonHG:
In testing the flow for a gerrit project, I am unable to pick ccla or icla, it skips that screen and asks me to sign an icla.

Choose a Gerrit project to sign for (either o-ran project, onap, OPNFV).
Observe you are never given the choice of Individual or Corporate, it sends you directly to Individual flow.

<img width="894" height="664" alt="Image" src="https://github.com/user-attachments/assets/5bb8e11b-4253-4e9d-9785-830c1b88594b" />

mlehotskylf:
I believe we have 2 options to fix this quickly:

1. Temporary disable signing with Gerrit from SS
2. Add ICLA/ECLA picker after user selects Gerrit identity

ahmedomosanya:
**Agreed fix (Slack, 2026-09-02):** Option **# 2** — add an Individual/Corporate (ICLA/ECLA) picker **after** the user selects their Gerrit identity, before hand-off to the Contributor Console.

Option # 1 (temporarily disable Gerrit signing from Self Serve) was considered and rejected.

**UX:** Mirror the existing Console decision-screen pattern for copy and layout.

**Single-type groups:** If only ICLA or only CCLA is enabled for the group, skip the picker and route directly to the sole contract type.

Implementation tracked on branch `fix/GH-2066-gerrit-contract-type-picker`.

mlehotskylf:
This is already in PROD.

<img width="1257" height="893" alt="Image" src="https://github.com/user-attachments/assets/b0da5d33-354e-4f72-a139-6f4e3cc4e390" />

mlehotskylf:
This is working as expected, closing.

Issue #1914, opened 2026-08-28
Contributor can start signing a CLA for a group they have already signed


mlehotskylf:
A contributor who already holds a CLA for a CLA Group can start signing another one for the same group. Nothing along the way says they are already covered.

From My CLAs, **Sign a CLA** → pick a group the contributor has already signed → the hand-off to the Contributor Console proceeds normally, through four screens, with no warning.

## What we saw in dev

**ICLA, already signed.** The flow runs all the way to a live **SIGN CLA** button. No data is corrupted — the backend reuses the existing signature rather than creating a second one — so this is a user-deception problem, not a data-integrity one. The contributor is walked through a signing ceremony for something they already have.

**ECLA, contributor already holds an ICLA for that group.** Not confirmed either way. Our attempt stopped on an unrelated gate (the chosen company had no CCLA with that project), so we never reached the point where an already-signed check would run. A code read did not find one on the employee path, so the cross-type case is suspected but unverified — it needs a company that holds a CCLA with the target group and has the contributor on its approved list.

## Expected

One identity should hold one CLA per CLA Group — either an ICLA or an ECLA, not both. When a contributor already has one, **Sign a CLA** should tell them so up front instead of handing off to a signing flow that cannot produce anything new.

Note that an ICLA and an ECLA can legitimately coexist on the same group when they belong to **different** identities (e.g. separate GitHub accounts). The constraint is per identity.

Found during M2 QA.

WillsonHG:
All cla's I try to sign are not available to me to sign, I get the proper notification that I've already signed with my accounts for both ecla and icla

PR #2156, opened 2026-09-03
fix(cla): keep an identity selectable until every enabled CLA type is held

ahmedomosanya:
## Summary
- The already-signed identity gate now blocks a GitHub or Gerrit identity only when every contract type the CLA group enables is already held under that identity.
- Holding an ICLA no longer greys the same identity on a group that still offers an ECLA, and the reverse. Single-type groups still grey once that type is held, so a contributor cannot re-sign something they already have.
- The Gerrit contract-type step disables a type the confirmed identity already holds, so opening the gate does not hand back the redundant signing ceremony it prevents.
- Help Center copy on the CLAs article and the Account FAQ now states the kind-aware rule.

## Gate

```
enabled = { ICLA if iclaEnabled, ECLA if cclaEnabled }
held    = kinds matching this identity (status set unchanged)
block   = enabled ⊆ held, when enabled is non-empty
          any match,      when enabled is empty
```

| Group | Held | Card |
| --- | --- | --- |
| Both enabled | ICLA only / ECLA only | selectable |
| Both enabled | both | greyed |
| ICLA-only | ICLA | greyed (#1914 kept) |
| CCLA-only | ECLA | greyed |
| ICLA-only | ECLA | selectable |
| Neither enabled | any match | greyed (kind-blind fallback) |

Where the block stands but the identity also holds a kind the group has since dropped, the reason names only a kind the group still offers — that is what the graying is owed to. Under the kind-blind fallback there is no offered kind to narrow to, so the reason names what is held.

Two false flags are not a group with nothing to sign — upstream leaves both false for a group whose CLA Group record it could not resolve, and the search still returns that result when a mapping row supplies its name. Read literally the enabled set is empty, every identity satisfies the subset test, and the #1914 block is off for that group without saying so. Any match blocks there instead, which is what the gate did before it knew about kinds. The Gerrit hand-off already refuses such a group with "Could not start signing", so the fallback costs that route nothing and is the whole of the protection on the GitHub route, which never reads these flags.

### Downstream of the gate

Letting a partial holder through the identity step means the next step must not offer them the type they already hold. On the Gerrit route that step is ours, so it disables the held card with a reason and leaves the other selectable — sharing the gate's identity matcher so the two cannot disagree. Both cards disabled is unreachable, since an identity holding every enabled type is greyed a step earlier.

The GitHub route has no such step: its contract type is chosen in the Contributor Console after hand-off.

## Related

- Refs #1914 — added the already-signed gate this corrects. That gate matched on identity alone, which is what made one contract type block the other.
- Refs #2066 / #2075 — the Gerrit contract-type picker. The gap was found reviewing that change: an ICLA holder never reached the new picker, because the identity step had already greyed their only card.

No closing keyword: neither issue is fixed by this, and there is no separate tracking issue.

## Test plan
- [ ] On a group with both ICLA and CCLA enabled, an identity that holds only an ICLA stays selectable (GitHub and Gerrit).
- [ ] On that same group, once the identity holds both an ICLA and an ECLA, the card is greyed and the tooltip names both kinds.
- [ ] On an ICLA-only group, an identity that already holds an ICLA is still greyed.
- [ ] A second linked GitHub account remains selectable when the first is greyed, and the tooltip still offers "choose another identity" only in that case.
- [ ] A group with neither type enabled still greys an identity that holds any agreement for it.
- [ ] On a dual-type Gerrit group, an ICLA holder reaches the contract-type step with Individual disabled and Corporate selectable.

### DIFF
--- a/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
@@ -343,14 +343,16 @@ export class ProfileClasComponent {
     this.starting.set(false);
     this.signDialogOpen.set(true);
 
-    // What they already hold for *this* group, so the step can gray out the identity that signed
-    // it. Passed as agreements rather than a precomputed verdict because only the step knows
-    // which identities it ended up offering.
+    // What they already hold for *this* group, plus which types the group enables, so the step
+    // can gray out an identity with nothing left to sign. Passed as agreements rather than a
+    // precomputed verdict because only the step knows which identities it ended up offering.
     const claGroupAgreements = alreadySignedAgreementsForGroup(this.agreements(), option.claGroupId);
 
     const data: SignIdentityDialogData = {
       variant: effectiveVariant,
       accounts,
+      iclaEnabled: option.iclaEnabled === true,
+      cclaEnabled: option.cclaEnabled === true,
       ...(gerritUsername ? { gerritUsername } : {}),
       ...(claGroupAgreements.length > 0 ? { claGroupAgreements } : {}),
     };
--- a/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
@@ -23,6 +23,7 @@ import type {
   MyClaAgreement,
   MyClasState,
   PrepareSignResponse,
+  SignContractTypeDialogData,
   SignContractTypeSelectResult,
   SignIdentityDialogData,
   SignIdentitySelectResult,
@@ -36,6 +37,7 @@ import {
   downloadFromUrl,
   formatClaSignedOn,
   gerritSignUrl,
+  heldClaKindsForIdentity,
   isMyClasEmpty,
   resolveGerritContractType,
   signedAsLine,
@@ -443,15 +445,24 @@ export class ProfileClasComponent {
 
     this.signDialogOpen.set(true);
 
-    // No data: the step renders both cards unconditionally, and it is only reachable for a group
-    // that enables both. Passing the flags it would have to ignore invites a later caller to open
-    // it for a single-type group, which is the hand-off this branch exists to make without asking.
+    // The group's own flags are not passed: the step is only reachable for a group that enables
+    // both, so it would have to ignore them, and a later caller could read them as licence to
+    // open it for a single-type group — the hand-off this branch makes without asking.
+    //
+    // What it does get is the types this Gerrit identity already holds. The identity step passed
+    // them because one type is still unsigned, and without this the step would offer the type
+    // they hold as freely as the one they need, which is the re-sign that gate prevents.
+    const data: SignContractTypeDialogData = {
+      heldKinds: heldClaKindsForIdentity(alreadySignedAgreementsForGroup(this.agreements(), option.claGroupId), { platform: 'gerrit' }, []),
+    };
+
     const dialogRef = this.dialogService.open(SignContractTypeSelectComponent, {
       header: SIGN_CONTRACT_TYPE_COPY.header,
       width: '32rem',
       modal: true,
       closable: true,
       dismissableMask: true,
+      data,
     }) as DynamicDialogRef;
 
     this.whenDialogSettles<SignContractTypeSelectResult>(dialogRef, (result) => {
--- a/apps/lfx-one/src/app/modules/profile/clas/sign-contract-type-select.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/sign-contract-type-select.component.ts
@@ -5,8 +5,8 @@ import { ChangeDetectionStrategy, Component, inject, signal } from '@angular/cor
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
 import { GERRIT_CONTRACT_TYPE_CORPORATE, GERRIT_CONTRACT_TYPE_INDIVIDUAL, SIGN_CONTRACT_TYPE_COPY } from '@lfx-one/shared/constants';
-import type { GerritContractType, SignContractTypeSelectResult } from '@lfx-one/shared/interfaces';
-import { DynamicDialogRef } from 'primeng/dynamicdialog';
+import type { GerritContractType, SignContractTypeDialogData, SignContractTypeSelectResult } from '@lfx-one/shared/interfaces';
+import { DynamicDialogConfig, DynamicDialogRef } from 'primeng/dynamicdialog';
 
 import { ButtonComponent } from '@components/button/button.component';
 import { SelectableCardComponent } from '@components/selectable-card/selectable-card.component';
@@ -21,6 +21,11 @@ import { SelectableCardComponent } from '@components/selectable-card/selectable-
  * type enabled never reaches here — the caller resolves that type and hands off without asking —
  * so a flag on this dialog could only ever say "both", and a dialog that could render one card is
  * one that could render none.
+ *
+ * A type the confirmed identity already holds is disabled rather than dropped. The identity step
+ * only lets them this far because one type is still unsigned, so removing the other would leave
+ * a step that asks a question with one answer, and never say why the other went. Both disabled
+ * cannot happen: that identity is grayed a step earlier.
  */
 @Component({
   selector: 'lfx-sign-contract-type-select',
@@ -30,10 +35,15 @@ import { SelectableCardComponent } from '@components/selectable-card/selectable-
 })
 export class SignContractTypeSelectComponent {
   private readonly ref = inject(DynamicDialogRef);
+  private readonly config = inject<DynamicDialogConfig<SignContractTypeDialogData>>(DynamicDialogConfig);
 
   protected readonly copy = SIGN_CONTRACT_TYPE_COPY;
   protected readonly individualValue = GERRIT_CONTRACT_TYPE_INDIVIDUAL;
   protected readonly corporateValue = GERRIT_CONTRACT_TYPE_CORPORATE;
+
+  private readonly heldKinds = this.config.data?.heldKinds ?? [];
+  protected readonly individualHeld = this.heldKinds.includes('ICLA');
+  protected readonly corporateHeld = this.heldKinds.includes('ECLA');
 
   protected readonly selectForm = new FormGroup({
     contractType: new FormControl<GerritContractType | null>(null),
--- a/apps/lfx-one/src/app/modules/profile/clas/sign-contract-type-select.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/sign-contract-type-select.component.ts
@@ -57,12 +57,22 @@ export class SignContractTypeSelectComponent {
 
   protected onContinue(): void {
     const contractType = this.selectedType();
-    if (!contractType) return;
+    if (!contractType || this.isHeld(contractType)) return;
 
     this.ref.close({ contractType } satisfies SignContractTypeSelectResult);
   }
 
   protected onCancel(): void {
     this.ref.close(null);
+  }
+
+  /**
+   * Whether the confirmed identity already holds this type. The disabled card refuses the click,
+   * so reaching here means the control was written some other way — a preselection, a form patch
+   * — and closing on it would hand off the re-sign the step exists to prevent. The identity step
+   * re-checks its own verdict on submit for the same reason.
+   */
+  private isHeld(contractType: GerritContractType): boolean {
+    return contractType === this.individualValue ? this.individualHeld : this.corporateHeld;
   }
 }
--- a/apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
@@ -204,13 +204,29 @@ export class SignIdentitySelectComponent {
     // that happens to be all digits.
     const offeredHandles = accounts.map((account) => account.githubUsername);
 
-    // The gate says whether the card is blocked; the held set says what to call it. Asking the
-    // same matcher twice keeps the reason describing this identity rather than the group.
+    const enabledKinds: readonly ClaKind[] = [...(enabled.iclaEnabled ? (['ICLA'] as const) : []), ...(enabled.cclaEnabled ? (['ECLA'] as const) : [])];
+
+    /**
+     * The gate says whether the card is blocked; these say what to call it.
+     *
+     * Drawn from the gate's own matcher rather than from the enablement flags, so the reason
+     * describes this identity's agreements instead of standing on the gate's rule holding. Then
+     * narrowed to the types the group still offers, because those are the ones the block is owed
+     * to — an identity holding a type the group has since dropped is not blocked by that type,
+     * and naming it would explain the graying with an agreement that has nothing to do with it.
+     *
+     * The narrowing is dropped when the group offers nothing, which is the gate's kind-blind
+     * fallback: there the match itself is the whole reason, and an empty list would leave the
+     * tooltip naming a type by default rather than the one actually held.
+     */
     const blockFor = (identity: SignIdentityRef): IdentityBlock | undefined => {
       const agreement = alreadySignedAgreementForIdentity(agreements, identity, offeredHandles, enabled);
       if (!agreement) return undefined;
 
-      return { agreement, heldKinds: heldClaKindsForIdentity(agreements, identity, offeredHandles) };
+      const held = heldClaKindsForIdentity(agreements, identity, offeredHandles);
+      const offered = held.filter((kind) => enabledKinds.includes(kind));
+
+      return { agreement, heldKinds: offered.length > 0 ? offered : held };
     };
 
     const byGithubId = new Map<string, IdentityBlock>();
--- a/apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
+++ b/apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
@@ -5,12 +5,32 @@ import { ChangeDetectionStrategy, Component, computed, inject, Signal, signal } 
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
 import { GERRIT_IDENTITY_VALUE, SIGN_IDENTITY_COPY, SIGN_IDENTITY_PLATFORM_LABELS } from '@lfx-one/shared/constants';
-import type { ClaKind, GithubAccountChoice, MyClaAgreement, SignIdentityDialogData, SignIdentitySelectResult } from '@lfx-one/shared/interfaces';
-import { alreadySignedAgreementForIdentity, alreadySignedIdentityTooltip } from '@lfx-one/shared/utils';
+import type {
+  ClaGroupEnablement,
+  ClaKind,
+  GithubAccountChoice,
+  MyClaAgreement,
+  SignIdentityDialogData,
+  SignIdentityRef,
+  SignIdentitySelectResult,
+} from '@lfx-one/shared/interfaces';
+import { alreadySignedAgreementForIdentity, alreadySignedIdentityTooltip, heldClaKindsForIdentity } from '@lfx-one/shared/utils';
 import { DynamicDialogConfig, DynamicDialogRef } from 'primeng/dynamicdialog';
 
 import { ButtonComponent } from '@components/button/button.component';
 import { SelectableCardComponent } from '@components/selectable-card/selectable-card.component';
+
+/**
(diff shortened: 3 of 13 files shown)


packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -411,6 +424,13 @@ export function alreadySignedAgreementForIdentity(
 
     return !handles.has(signedAs) && signedAs === identity.githubId.trim().toLowerCase();
   });
+
+  const heldKinds = new Set(matched.map((agreement) => agreement.kind));
+  for (const kind of enabledKinds) {
+    if (!heldKinds.has(kind)) return undefined;
+  }
+
+  return matched[0];
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `7646f64b0` — the reported agreement is now drawn from an enabled kind, so a group that has since dropped a type the identity already signed no longer names that type as the reason. The block itself is unchanged: it still requires every enabled kind to be held, which is what guarantees the lookup resolves. Covered by a case that fails if the lookup goes back to the first match.

packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -411,6 +424,17 @@ export function alreadySignedAgreementForIdentity(
 
     return !handles.has(signedAs) && signedAs === identity.githubId.trim().toLowerCase();
   });
+
+  const heldKinds = new Set(matched.map((agreement) => agreement.kind));
+  for (const kind of enabledKinds) {
+    if (!heldKinds.has(kind)) return undefined;
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `9440d0e4d` — good catch, this was a real narrowing of #1914 on the Gerrit route. The picker now receives the types the confirmed identity already holds and disables that card with a reason, leaving the type they still need selectable. It shares the gate's identity matcher (`heldClaKindsForIdentity`) rather than re-deriving the match, so the step and the gate cannot disagree about what is held. Both cards disabled is not reachable: an identity holding every enabled type is grayed a step earlier. GitHub is unaffected — its contract type is chosen in the Contributor Console after hand-off, so there is no Self Serve step to gate.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -441,15 +445,24 @@ export class ProfileClasComponent {
...
-    // it for a single-type group, which is the hand-off this branch exists to make without asking.
+    // The group's own flags are not passed: the step is only reachable for a group that enables
+    // both, so it would have to ignore them, and a later caller could read them as licence to
+    // open it for a single-type group — the hand-off this branch makes without asking.
+    //
+    // What it does get is the types this Gerrit identity already holds. The identity step passed
+    // them because one type is still unsigned, and without this the step would offer the type
+    // they hold as freely as the one they need, which is the re-sign that gate prevents.
+    const data: SignContractTypeDialogData = {
+      heldKinds: heldClaKindsForIdentity(alreadySignedAgreementsForGroup(this.agreements(), option.claGroupId), { platform: 'gerrit' }, []),
```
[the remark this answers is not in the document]
ahmedomosanya: Addressed in `1a95251d0` — added a profile-component case that asserts the step is served the kinds held by the Gerrit identity, and mutation-checked it: replacing the wiring with a literal `[]` fails it. The fixture also carries an agreement on a different CLA group and a GitHub agreement on this one, so a matcher pointed at the wrong group or platform fails there too. The hand-off assertion on the remaining route is in the same case.

luismoriguerra (dismissed):
## Audit: PR #2156 — fix(cla): keep an identity selectable until every enabled CLA type is held

**Verdict:** ✅ Approved
**Merge readiness:** safe-to-merge after manual checklist or deploy-preview sign-off; merge #2156 before #2155

**Gates:** lint:check ✓ · check-types ✓ · shared CLA tests 87/87 ✓ (+11) · UI suite 1711/1711 ✓ · CI green

**Findings:**
1. **Non-blocking** — Manual test plan still unchecked in PR body; CI + unit tests are strong.
2. **Non-blocking** — GitHub path has no contract-type step (documented; follow-up if needed).
3. **Non-blocking** — Merge coordination with #2155: merge this PR first, then rebase #2155.

No blocking findings — no inline threads opened.

apps/lfx-one/src/app/modules/profile/clas/sign-contract-type-select.component.ts
```diff
@@ -30,11 +35,16 @@ import { SelectableCardComponent } from '@components/selectable-card/selectable-
...
   private readonly ref = inject(DynamicDialogRef);
+  private readonly config = inject<DynamicDialogConfig<SignContractTypeDialogData>>(DynamicDialogConfig);
 
   protected readonly copy = SIGN_CONTRACT_TYPE_COPY;
   protected readonly individualValue = GERRIT_CONTRACT_TYPE_INDIVIDUAL;
   protected readonly corporateValue = GERRIT_CONTRACT_TYPE_CORPORATE;
 
+  private readonly heldKinds = this.config.data?.heldKinds ?? [];
+  protected readonly individualHeld = this.heldKinds.includes('ICLA');
+  protected readonly corporateHeld = this.heldKinds.includes('ECLA');
```
audigregorie, line 46: **Major — `onContinue` has no held-type guard; only the card click is guarded**

The held types introduced here are enforced exclusively by `SelectableCardComponent`'s disabled guard refusing the click, so the form control is never written from the UI. But `onContinue` itself only checks `!contractType` — a programmatic `selectForm.controls.contractType.setValue('individual')` (a future preselection, a form patch, a test harness) would close the dialog with a type this identity already holds, which is the exact redundant re-signing ceremony this PR exists to prevent.

The sibling identity step treats this as worth a submit-path guard: `SignIdentitySelectComponent.onContinue` re-checks `isAlreadySigned(selected)` before closing, and its spec pins the case with "refuses to submit it even if the card is reached another way" (writing the control directly). This step should mirror that:

```ts
protected onContinue(): void {
  const contractType = this.selectedType();
  if (!contractType) return;
  if ((contractType === this.individualValue && this.individualHeld) || (contractType === this.corporateValue && this.corporateHeld)) return;

  this.ref.close({ contractType } satisfies SignContractTypeSelectResult);
}
```

plus the mirror spec that writes the control directly and asserts `close` is not called. The current "refuses to close on the held type" test only exercises the click path, so it stays green while the submit guard is missing. (`onContinue` is outside the diff, so anchoring here where the held-kinds handling begins.)

packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -389,16 +395,14 @@ export function alreadySignedGroupTooltip(agreement: MyClaAgreement, route: ClaS
...
- * identifies it — which does mean a Gerrit agreement with a blank handle still blocks that card.
+ * identifies it — which does mean a Gerrit agreement with a blank handle still matches that
+ * card, and then the enablement rule decides whether the match blocks.
  */
-export function alreadySignedAgreementForIdentity(
-  agreements: readonly MyClaAgreement[],
-  identity: SignIdentityRef,
-  offeredHandles: readonly string[]
-): MyClaAgreement | undefined {
+/** The agreements this identity holds for the group, newest first. Shared by the two readers below. */
```
audigregorie, line 369: **Major — the gate's JSDoc is now attached to the matcher, and the gate itself has none**

The large doc block above (lines 369–400: "The agreement that blocks this identity from signing again… Blocks when every contract type the group enables is already held under this identity (`enabled ⊆ held`)…") describes the enablement gate, but it now sits on `agreementsMatchingIdentity`, which does no gating — it filters matches and returns all of them. A reader of this private helper is told it implements `enabled ⊆ held` logic that isn't here, while the exported `alreadySignedAgreementForIdentity` below (line 431) — the function the block actually documents — carries no comment at all.

The same refactor left `heldClaKindsForIdentity`'s doc referencing "the gate above" (line 424) when the gate is now below it.

Suggest moving the big block down to `alreadySignedAgreementForIdentity` (it reads correctly there essentially verbatim), keeping this one-liner here, and fixing the "above" reference.

apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
```diff
@@ -223,7 +229,13 @@ export class SignIdentitySelectComponent {
     if (!this.gerritLabel()) return undefined;
 
     const held = this.alreadySigned.gerrit;
-    return held ? alreadySignedIdentityTooltip(held, this.alreadySigned.anotherSelectable) : undefined;
+    return held ? this.identityTooltip(held) : undefined;
+  }
+
+  private identityTooltip(held: MyClaAgreement): string {
+    const namesBoth = this.config.data?.iclaEnabled === true && this.config.data?.cclaEnabled === true;
+    const heldKinds: readonly ClaKind[] = namesBoth ? ['ICLA', 'ECLA'] : [held.kind];
```
audigregorie: **Minor — tooltip names kinds derived from enablement flags rather than from what is actually held**

When both types are enabled this passes `['ICLA', 'ECLA']` unconditionally; the tooltip is accurate only because the gate (`enabled ⊆ held`) guarantees an identity greyed here holds both. That coupling is invisible at this call site — if the gate's rule ever loosened, or the flags arrived stale, the tooltip would name an agreement the contributor does not hold.

`heldClaKindsForIdentity` was extracted in this PR for exactly this question and is already trusted by the contract-type step; consider having `resolveAlreadySigned` retain the matched agreements (or re-query via that util) so the tooltip names the actual held set rather than the enabled set.

packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -413,16 +417,59 @@ export function alreadySignedAgreementForIdentity(
...
+ */
+export function heldClaKindsForIdentity(agreements: readonly MyClaAgreement[], identity: SignIdentityRef, offeredHandles: readonly string[]): ClaKind[] {
+  return [...new Set(agreementsMatchingIdentity(agreements, identity, offeredHandles).map((agreement) => agreement.kind))];
+}
+
+export function alreadySignedAgreementForIdentity(
+  agreements: readonly MyClaAgreement[],
+  identity: SignIdentityRef,
+  offeredHandles: readonly string[],
+  enabled: { iclaEnabled: boolean; cclaEnabled: boolean }
```
audigregorie: **Minor — anonymous inline type for the `enabled` parameter**

`enabled: { iclaEnabled: boolean; cclaEnabled: boolean }` is an anonymous structural type on an exported shared-util signature, and the same shape is rebuilt inline again in `SignIdentitySelectComponent.resolveAlreadySigned`. Per the shared-package conventions (named types live in `interfaces/`), a small exported interface — e.g. `ClaGroupEnablement` — or `Pick<ClaGroupOption, 'iclaEnabled' | 'cclaEnabled'>` would document the contract once and let both call sites share it.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
```diff
@@ -343,14 +345,16 @@ export class ProfileClasComponent {
...
+    // What they already hold for *this* group, plus which types the group enables, so the step
+    // can gray out an identity with nothing left to sign. Passed as agreements rather than a
+    // precomputed verdict because only the step knows which identities it ended up offering.
     const claGroupAgreements = alreadySignedAgreementsForGroup(this.agreements(), option.claGroupId);
 
     const data: SignIdentityDialogData = {
       variant: effectiveVariant,
       accounts,
+      iclaEnabled: option.iclaEnabled === true,
+      cclaEnabled: option.cclaEnabled === true,
```
audigregorie, line 357: **Minor — missing enablement flags now read as "group offers nothing", silently retiring the #1914 block**

`option.iclaEnabled === true` coerces an absent flag to `false`, and the gate treats both-absent as neither-enabled — not a block. On `main`, an identity with a matching agreement is greyed regardless of flags, so for any group where the producer omits them this quietly turns the already-signed protection off, indistinguishable from a group that genuinely offers neither type.

The `ClaGroupOption` doc does warn that dropping the flags reinstates a silent default, so this may be contractually covered — flagging to confirm the producer always sends both flags for every searchable group (including GitHub-only ones, where #2066's routing never needed them). If that guarantee is solid, a one-line comment here saying absence is impossible would settle it.

audigregorie (commented):
## Code Review Summary

Correct, well-tested kind-aware fix for the #1914 gate: the `enabled ⊆ held` rule, the shared matcher between the identity gate and the contract-type step, and the "name a still-offered kind" detail are all right, and the spec coverage (dual/single/neither-enabled, revoked kind, blank handle, deliberately-wrong-matcher wiring) is genuinely strong. Two things worth addressing before merge: the gate's JSDoc landed on the wrong function, and the new contract-type step is missing the submit-path guard its sibling identity step treats as essential.

### PR shape (outside the diff)

- Title, branch, and commit messages follow conventions; `Refs #1914` / `Refs #2066` / `Refs #2075` linking with the no-closing-keyword rationale stated up front — good.
- Help Center docs (`faq`, `my-clas`) updated in lockstep with the behavior change, `last_updated` bumped. ✅

### What's done well

- Extracting `agreementsMatchingIdentity` so the identity gate and the contract-type step cannot disagree about what an identity holds — with a profile-level test that fails if the wiring points at the wrong matcher.
- Returning an enabled-kind agreement for the block reason, so a group that dropped a type never names it in the tooltip — pinned by a dedicated test.
- Held cards in the contract-type step are disabled rather than dropped, with the accessibility reasoning documented and both orderings tested.

audigregorie (dismissed):

apps/lfx-one/src/app/modules/profile/clas/sign-contract-type-select.component.ts
ahmedomosanya: Fixed in `da3a884bf`, mirroring the identity step as you suggested. `onContinue` now refuses a held type, and the spec writes the control directly to reach the submit path with the cards bypassed.

You were right that the existing test couldn't catch it — the disabled card refuses the click, so the form was never written. I've renamed that one to say what it actually proves (the click is refused and the control stays null) and added the submit-path test beside it. Mutation-checked: dropping the guard fails the new test alone.

packages/shared/src/utils/cla-view.utils.ts
ahmedomosanya: Fixed in `d66cb3161`. The block moved down onto `alreadySignedAgreementForIdentity`, where it reads correctly essentially verbatim; `agreementsMatchingIdentity` keeps the one-liner, and `heldClaKindsForIdentity` now says "the gate below".

Its neither-enabled paragraph is rewritten rather than moved, because that rule changed in the same commit — see the thread on `profile-clas.component.ts` for why.

packages/shared/src/utils/cla-view.utils.ts
ahmedomosanya: Fixed in `d66cb3161` — added `ClaGroupEnablement` to `interfaces/`, used on the util signature and at the `resolveAlreadySigned` call site that was rebuilding the shape inline.

Went with a small interface rather than `Pick<ClaGroupOption, 'iclaEnabled' | 'cclaEnabled'>` because the fields are optional on the source and every reader has to settle absent-means-disabled before using them. The named type is that settled form — both flags definite — so the reasoning lives in one place instead of at each call site.

apps/lfx-one/src/app/modules/profile/clas/profile-clas.component.ts
ahmedomosanya: Confirmed and fixed in `d66cb3161`. The mechanism turned out to be different from the one in your note, but the conclusion is the same either way — thanks for pushing on this, it's the finding I'd have been sorriest to ship.

**Absence cannot reach the browser.** Upstream declares both flags `x-omitempty: false` in `cla-search-result.yaml`, so they are serialized even when false. And a flag never written to DynamoDB does not arrive false either — `enabledOrDefault` returns `enabled == nil || *enabled`, the python-v2 compatibility default, so an unset `project_icla_enabled` comes back as `true`. So the "mapper drops them" framing in the `ClaGroupOption` doc is not the live risk, and no producer guarantee is needed for it.

**What is reachable is `buildResult`'s `if claGroup != nil`.** When the CLA Group record cannot be resolved, neither flag is assigned, both keep Go's zero value, and the result ships two explicit `false`s. It still ships, because the drop filter only skips results with neither a CLA Group name nor a project name, and a mapping row can supply either. So a both-false group is an inconsistent-data case rather than a healthy group — narrow, but not something either of us can rule out.

On that group the gate read an empty enabled set, and an empty set satisfies the subset test for every identity, so the #1914 block was off — exactly the outcome you described. It now falls back to the kind-blind block. The Gerrit route loses nothing, since the hand-off refuses such a group anyway with "Could not start signing", and the GitHub route keeps its only protection.

So rather than the confirming comment you offered as the alternative, this is a behaviour change plus two tests: one that any match blocks when neither type is enabled, and one that an identity holding no agreement of its own stays selectable — the fallback widens which matches block, never what counts as a match. The gate table in the PR body is updated to match.

apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
ahmedomosanya: Fixed in `afac0bb19`, then narrowed in `2914a299b` after Copilot pushed back — the result addresses your coupling concern but doesn't go quite where your suggestion pointed, so it's worth saying why.

`resolveAlreadySigned` now retains the held set beside the blocking agreement, read off the gate's own matcher rather than the enablement flags, which removes the invisible coupling you identified. But naming everything held turned out to be wrong in the case that distinguishes the two: on a group that has dropped a type the identity signed, the block is owed to the remaining one, so naming the dropped one explains the graying with an agreement that has nothing to do with it. The held set is therefore narrowed to the types the group still offers.

The narrowing is dropped under the kind-blind fallback for a group that enables nothing, where the match itself is the whole reason. That case is not inert: an empty kind list makes `alreadySignedIdentityTooltip` fall through to its `an ICLA` default, so an ECLA holder would be told they hold an ICLA. Mutation-checked in both directions — dropping the narrowing and dropping the fallback each fail exactly one test.

apps/lfx-one/src/app/modules/profile/clas/sign-identity-select.component.ts
```diff
@@ -169,23 +191,35 @@ export class SignIdentitySelectComponent {
...
     const offeredHandles = accounts.map((account) => account.githubUsername);
 
-    const byGithubId = new Map<string, MyClaAgreement>();
+    // The gate says whether the card is blocked; the held set says what to call it. Asking the
+    // same matcher twice keeps the reason describing this identity rather than the group.
+    const blockFor = (identity: SignIdentityRef): IdentityBlock | undefined => {
+      const agreement = alreadySignedAgreementForIdentity(agreements, identity, offeredHandles, enabled);
+      if (!agreement) return undefined;
+
+      return { agreement, heldKinds: heldClaKindsForIdentity(agreements, identity, offeredHandles) };
```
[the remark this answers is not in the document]
ahmedomosanya: Agreed and fixed in `2914a299b`. The held set is now narrowed to the enabled kinds, so the reason names only what the block is owed to and the `c50a80edf` rule holds again.

The narrowing is skipped when the group enables neither type, which is the kind-blind fallback added in `d66cb3161`: there is no offered kind to narrow to, and an empty list is not inert — `alreadySignedIdentityTooltip` falls through to `an ICLA`, so an ECLA holder would be told they hold an ICLA. Regression expectations updated, and a second test covers the fallback.

packages/shared/src/utils/cla-view.utils.ts
```diff
@@ -389,28 +433,33 @@ export function alreadySignedGroupTooltip(agreement: MyClaAgreement, route: ClaS
...
-    if (agreement.signedVia !== 'github') return false;
+  const matched = agreementsMatchingIdentity(agreements, identity, offeredHandles);
+  if (matched.length === 0) return undefined;
 
-    const signedAs = agreement.signedAs?.trim().toLowerCase();
-    if (!signedAs) return false;
+  const enabledKinds = new Set<ClaKind>();
+  if (enabled.iclaEnabled) enabledKinds.add('ICLA');
+  if (enabled.cclaEnabled) enabledKinds.add('ECLA');
+  if (enabledKinds.size === 0) return matched[0];
```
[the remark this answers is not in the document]
ahmedomosanya: Half right, and the half that lands is fixed in `2914a299b`. The gate's summary line still required a non-empty `enabled`, which the paragraph below it contradicts; that sentence now states both rules. My own prose also had it backwards — it said an empty set is "a subset of nothing", when the point is the reverse: the empty set is a subset of every set, which is exactly why the test passes for every identity.

The behaviour is deliberate rather than a drift from the documented gate. Two false flags are what a group whose CLA Group record upstream could not resolve arrives as, not a group with nothing to sign, so taking the subset test at face value there turns the #1914 already-signed block off for that group without saying so. The kind-blind block stands in — what this gate did before it knew about kinds. The test plan you cite was describing the old rule and has been updated along with the gate table in the PR body.

luismoriguerra (approved):

PR #2159, opened 2026-09-03
perf(projects): batch project enrichment via getProjectsByIds

### LINKED ISSUE #1643: Batch project enrichment through the query service instead of per-project fetches
## Problem

Every project-scoped list page (meetings, groups, mailing lists, votes, surveys, newsletters) currently fetches each referenced project one at a time from the upstream project service to attach the project name, slug, and tier to each row — a dashboard whose rows span 30 projects costs 30 extra upstream calls on every load, for every user. These per-project fetches are also permission-gated, so committee writers who lack a direct project relation silently get rows without tier data, which makes their links fall back to non-canonical paths more often than other users.

## Outcome

List pages attach project name, slug, and tier from a single batched query instead of one fetch per project. List responses get faster and cheaper as row counts grow, and committee writers see the same fully populated rows (and canonical links) as everyone else. No other user-facing behavior changes; project metadata remains near-live.

## Success criteria

- A list page spanning many projects performs one batched project lookup (paginated if needed) instead of one upstream call per project — verifiable in server logs.
- Committee writers without a direct project relation see project names and tiers populated on list rows, and their links canonicalize like everyone else's.
- Meetings, groups, mailing lists, votes, surveys, and newsletters lists render unchanged for users who already received fully populated rows.

References: recorded as the "Middle path" in upstream-project-fields-plan.md (shelved upstream plan); complements GH-1641 and GH-1566 through GH-1569.

audigregorie:
## Summary

Project enrichment on list endpoints previously resolved every referenced project through individual `getProjectById` calls — 25 at a time, one request per project. This PR switches `enrichWithProjectData` to the existing `getProjectsByIds` batch path, resolving all referenced projects in one `filters_or=uid:X` query-service request per 100 uids, fail-soft per batch. The enrichment contract (payload fallbacks, `is_foundation` left undefined-not-false for unresolved projects) is unchanged, and the batched lookup is ungated, so callers without a direct project viewer relation still resolve name/slug/tier/parent.

Closes #1643

## Behavior changes

| Before | After |
|---|---|
| Listing screens showing many items (meetings, committees, and similar) triggered one backend project lookup per distinct project — a screen touching 40 projects meant 40 separate upstream calls, which was slow and rate-limit prone | All referenced projects are resolved in one batched lookup per 100 projects, so the same screens load faster and put far less load on upstream services — and users without direct project access still see project names, slugs, and tier info |

## Technical changes

`apps/lfx-one/src/server/services/project.service.ts` — Project enrichment

- Replaces the 25-at-a-time `Promise.all(getProjectById)` fan-out in `enrichWithProjectData` with a single `getProjectsByIds(req, projectUids)` call — one `filters_or=uid:X` query per 100 uids, fail-soft per batch
- Keeps the merge contract unchanged: freshly fetched fields win, per-item payload values remain the fallback, and `is_foundation` stays `undefined` (not `false`) for unresolved projects so consumers treat tier as unknown rather than mislabeling foundation-owned entities as project-owned
- The batched path is ungated (near-live from the index rather than the live record), so callers without a direct project viewer relation still resolve name/slug/tier/parent

`apps/lfx-one/src/server/services/project.service.spec.ts` — Tests

- Adds an `enrichWithProjectData` suite covering: uid dedup into a single batched lookup, fresh-field precedence over item payloads (including `is_foundation` flipping both ways), payload fallback with `is_foundation` left `undefined` for unresolved projects, per-row fallback when only some projects resolve, and untouched item fields when every batch fails

`apps/lfx-one/src/server/services/create-picker.service.ts` — Comment only

- Updates the committee project-resolution comment to reflect that `enrichWithProjectData` now shares the same batched `getProjectsByIds` lookup (no behavioral change)

### DIFF
--- a/apps/lfx-one/src/server/services/create-picker.service.ts
+++ b/apps/lfx-one/src/server/services/create-picker.service.ts
@@ -170,9 +170,9 @@ export class CreatePickerService {
   }
 
   /**
-   * Resolves each committee's parent project via `ProjectService.getProjectsByIds` — a true
-   * single-request-per-chunk batch (`filters_or=uid:X`), not `enrichWithProjectData`'s N
-   * individual per-project lookups. Committees whose project can't be resolved are omitted
+   * Resolves each committee's parent project via `ProjectService.getProjectsByIds` — the same
+   * batched `filters_or=uid:X` lookup `enrichWithProjectData` uses, minus its payload-fallback
+   * merge: committees whose project can't be resolved are omitted
    * entirely rather than emitted with blank `projectSlug`/`projectName`: a picked row with an
    * empty project slug would navigate with `?project=` empty, which `writerGuard` cannot resolve —
    * exactly the post-selection dead end this picker is supposed to make impossible.
--- a/apps/lfx-one/src/server/services/project.service.spec.ts
+++ b/apps/lfx-one/src/server/services/project.service.spec.ts
@@ -1595,3 +1595,118 @@ describe('ProjectService — getHealthMetricsDaily', () => {
     expect(execute.mock.calls[0][0]).toContain('AVG(HEALTH_SCORE_V2)');
   });
 });
+
+describe('ProjectService — enrichWithProjectData', () => {
+  let service: ProjectService;
+
+  // Membership-funded + Active + not an Internal Allocation, per the real computeIsFoundation.
+  function foundationProject(uid: string, overrides: Partial<Project> = {}): Project {
+    return {
+      uid,
+      slug: uid,
+      name: `name-${uid}`,
+      stage: 'Active',
+      legal_entity_type: '',
+      funding: 'Funded' as ProjectFunding,
+      funding_model: ['Membership'],
+      ...overrides,
+    } as Project;
+  }
+  // Missing the Membership funding model — the real computeIsFoundation returns false.
+  function childProject(uid: string, parentUid: string, overrides: Partial<Project> = {}): Project {
+    return {
+      uid,
+      slug: uid,
+      name: `name-${uid}`,
+      stage: 'Active',
+      legal_entity_type: '',
+      funding: 'Funded' as ProjectFunding,
+      funding_model: [],
+      parent_uid: parentUid,
+      ...overrides,
+    } as Project;
+  }
+
+  beforeEach(() => {
+    proxyRequest.mockReset();
+    service = new ProjectService();
+  });
+
+  it('deduplicates project_uids into a single batched lookup', async () => {
+    const getProjectsByIds = vi.spyOn(service, 'getProjectsByIds').mockResolvedValue(new Map());
+
+    await service.enrichWithProjectData(req, [{ project_uid: 'a' }, { project_uid: 'a' }, { project_uid: 'b' }]);
+
+    expect(getProjectsByIds).toHaveBeenCalledTimes(1);
+    expect(getProjectsByIds).toHaveBeenCalledWith(req, ['a', 'b']);
+  });
+
+  it('prefers freshly fetched project fields over the item payload and computes is_foundation from the fresh project', async () => {
+    vi.spyOn(service, 'getProjectsByIds').mockResolvedValue(
+      new Map([
+        ['fdn', foundationProject('fdn', { name: 'Fresh Foundation', slug: 'fresh-fdn' })],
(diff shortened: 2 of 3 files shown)


audigregorie (commented):
## Code Review Summary

Clean, well-scoped perf change: `enrichWithProjectData` swaps the 25-at-a-time `getProjectById` fan-out for the existing batched `getProjectsByIds` lookup, preserving the merge contract exactly. **No Critical, Major, or Minor findings.**

Verified locally: full-file reads of all three touched files plus the spec harness; `project.service.spec.ts` 91/91 passing (incl. the 5 new `enrichWithProjectData` tests); eslint clean on all touched files; GPG + DCO present.

### Contract parity confirmed

- Old path already skipped access checks (`getProjectById(req, uid, false)`) and fail-softed per item; the new path fail-softs per batch inside `fetchProjectBatchByIds`, so the net merge outcome is unchanged — which the new tests pin (fresh-field precedence with `is_foundation` flipping both ways, `undefined`-not-`false` for unresolved projects, per-row fallback).
- Running `computeIsFoundation` over query-service index payloads has pre-existing precedent in the same file (`discoverSubFoundations`), so the required classification fields are present in the index.
- The ungated, near-live-from-index behavior is a deliberate, documented trade-off and matches existing ungated consumers (`vote.service`, `entity-project-enrichment.helper`).

### What's done well

- Contract preservation is explicit and test-pinned rather than incidental.
- Reuses the already-tested `getProjectsByIds` instead of adding a third batching path; the `create-picker.service.ts` comment update keeps the shared-lookup docs accurate.
- New tests isolate the merge logic by spying on `getProjectsByIds` (which has its own dedicated suite) and import the real `computeIsFoundation`, so classification drift fails these tests.

Nice work — LGTM.

luismoriguerra (approved):
## Audit: PR #2159 — perf(projects): batch project enrichment via getProjectsByIds

**Verdict:** ✅ Approved  
**Merge readiness:** merge-after-confirming whether legacy index projects without `funding`/`funding_model` still appear in prod lists — or land the classification-incomplete guard Copilot suggested

**Gates:** format ✅ · eslint (touched) ✅ · lint:check ✅ · build ✅ · `project.service.spec.ts` 91/91 ✅

### Findings
1. **Non-blocking** — `project.service.ts:6910` — Legacy query-service rows missing `funding`/`funding_model` can coerce payload `is_foundation: true` → `false` via `computeIsFoundation`; fall back to payload when classification inputs are incomplete.
2. **Nit** — Copilot thread on the same edge case remains open; worth a reply or the guard above.

No blocking findings — summary only (selection `blocking`). 1 non-blocking finding and 1 nit are covered here but not opened as inline threads.

luismoriguerra:
## Audit correction (follow-up to approved review)

**Access framing:** The earlier chat audit overstated "ungated query-service widening." Per **ADR-0012/ADR-0015** and `lfx-v2-project-service` indexer contract, `project` docs carry `access_check_object: project:{uid}` + `access_check_relation: viewer` unless `project.public == true` (then query-service skips FGA). So for typical non-public projects, **Heimdall `GET /projects/:id` and query-service both require `project#viewer`** — no authz widening.

**Finding #1 (revised mechanism):** The non-blocking `funding` → `is_foundation: false` edge case is primarily **live record → index projection completeness** for users who already have viewer (sparse index rows override payload tier). Secondary case: **`project.public == true`** docs may resolve via query-service where Heimdall would 403 — same guard still applies.

**Verdict unchanged:** ✅ Approved. Optional: fall back to payload `is_foundation` when classification inputs are incomplete on a resolved index row.

**Doc nit:** `entity-project-enrichment.helper.ts` "needs no project relation" comment is imprecise per ADR-0012 — query-service still enforces viewer unless `Public: true`.

luismoriguerra:
**Follow-up clarification** on the correction above:

- **Access framing stands:** query-service enforces `project#viewer` unless `project.public` (ADR-0012/0015) — no authz widening for typical projects.

- **Finding #1 downgraded:** Retracting the "sparse index rows lack `funding`/`funding_model`" mechanism from the prior follow-up. That rested on TS optionality, not verified indexer behavior. The BFF already runs `computeIsFoundation` over `/query/resources` project hits elsewhere (`project.service.ts:7656` foundation-tree filter; writer-grant paths), so if index docs routinely omitted classification fields, foundation detection would likely already be broken app-wide.

- **Non-blocking item:** mirrors Copilot's open thread — speculative edge case only; not a verified regression. Verdict unchanged: ✅ Approved.

PR #2164, opened 2026-09-03
fix(meetings): keep edit page mounted on transient load errors (#2037)

### LINKED ISSUE #2037: Meeting edit page ejects users with a misleading not-found message on transient load errors
## Problem

When someone opens a meeting edit page and the meeting details request fails for a transient reason — a brief server error or a network blip — the page treats it as if the meeting does not exist: it shows a misleading "meeting not found" message and navigates the user away from the page they were on. Only a genuine "this meeting does not exist" response should do that.

## Outcome

A temporary failure to load meeting details no longer ejects the user from the edit page or claims the meeting is missing. The page stays put and simply carries on without the extra context that request provides; only a truly non-existent meeting produces the "not found" message and redirect. This matches the behavior already shipped for survey edit pages.

## Success criteria

- A transient server or network error while loading meeting details leaves the user on the edit page, with no "not found" message shown
- Opening an edit link for a meeting that genuinely does not exist still shows the "not found" message and returns the user to the meetings list
- Behavior matches the equivalent survey edit page

audigregorie:
## Summary

When editing a meeting, the page fetches the meeting's details up front to pre-populate the form. Any failure of that fetch — even a momentary server or network blip — was treated as "not found": the user was kicked back to the meetings list with a misleading toast. This PR narrows that eject to a genuine 404 and keeps the edit page mounted for everything else, with an inline error state and Retry / Back actions.

Resolves #2037

## Behavior changes

| Before | After |
|---|---|
| Any hiccup while loading a meeting for editing — a brief server error, a dropped connection — sent you back to the meetings list with a "Meeting not found or you do not have permission" message, even though the meeting existed and you had access. | A temporary load failure keeps you on the edit page with a clear "couldn't load this meeting" message and Retry / Back buttons. You are only sent back to the list when the meeting truly doesn't exist or you lack permission to see it. |

## Technical changes

`apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.ts` — Edit-mode load handling
- Splits the detail-fetch error path in `initializeMeeting` by status: a real 404 keeps the existing toast + eject to the meetings list; any other error sets a new `meetingLoadError` signal and stays mounted.
- Adds a `retryMeetingLoad$` Subject folded into the `initializeMeeting` pipeline via `combineLatest` + `startWith`, so each `next()` re-invokes the fetch; the error flag resets at the start of each attempt.
- Adds public `retryMeetingLoad()` (clears the error state, triggers the re-fetch) and promotes `navigateBack()` from private to public so the template's Back button can call it.

`apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.html` — Inline error state
- New `@if (meetingLoadError())` branch ahead of the loading skeleton: an error message with Retry and Back buttons (`meeting-manage-load-error`, `-retry`, `-back` test ids).

`apps/lfx-one/e2e/project-context-deep-link.spec.ts` — E2E coverage
- Adds a `stubMeetingEditDetailError` helper that fulfills the detail GET with a configurable error status.
- New "Meeting edit load failure (GH-2037)" describe block: a transient 500 stays mounted with retry/back and no not-found toast; clicking Retry after the backend recovers re-fetches and renders the form; a real 404 still ejects to the meetings list with the not-found toast.

### DIFF
--- a/apps/lfx-one/e2e/project-context-deep-link.spec.ts
+++ b/apps/lfx-one/e2e/project-context-deep-link.spec.ts
@@ -277,6 +277,21 @@ async function stubMeetingEditDetail(page: Page, meeting: ReturnType<typeof buil
 }
 
 /**
+ * Same route shapes as stubMeetingEditDetail, but the detail GET fulfills an error status —
+ * exercises meeting-manage's non-404 inline error state and 404 eject (GH-2037).
+ */
+async function stubMeetingEditDetailError(page: Page, status: number): Promise<void> {
+  await page.route('**/api/meetings*', (route) => {
+    if (route.request().method() !== 'GET') return route.fallback();
+    return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify({ data: [] }) });
+  });
+  await page.route(`**/api/meetings/${MOCK_MEETING_UID}`, (route) => {
+    if (route.request().method() !== 'GET') return route.fallback();
+    return route.fulfill({ status, contentType: 'application/json', body: JSON.stringify({ message: 'stub detail failure' }) });
+  });
+}
+
+/**
  * Edit-page detail payload; `enriched: false` mirrors the v1-sync gap (empty slug/name, no
  * is_foundation) so the component's resolve-by-uid fallback runs (GH-1567).
  */
@@ -537,6 +552,78 @@ test.describe('Meeting edit deep-link resolves the meeting’s project context (
 
     await page.waitForTimeout(500);
     expect(new URL(page.url()).searchParams.has('project')).toBe(false);
+  });
+});
+
+test.describe('Meeting edit load failure (GH-2037)', () => {
+  test.beforeEach(async ({ page }) => {
+    // Same shape as the gh-1432 meeting block: ED persona skips the writerGuard entity probe,
+    // so the component's own detail-fetch error branch is what runs.
+    await setPersonaAndLensCookies(page, ['executive-director'], 'project');
+    await setProjectCookie(page, OTHER_PROJECT_UID, OTHER_PROJECT_SLUG, 'Other Project');
+    await stubPersona(page, ['executive-director']);
+    await stubProjectApi(page);
+    await stubCommittees(page, buildCommittees());
+    await stubLensItems(page);
+  });
+
+  test('transient detail-load failure stays mounted with retry/back instead of ejecting (AC-1)', async ({ page }) => {
+    await stubMeetingEditDetailError(page, 500);
+
+    await gotoSpa(page, `/project/meetings/${MOCK_MEETING_UID}/edit`, {
+      uid: OTHER_PROJECT_UID,
+      slug: OTHER_PROJECT_SLUG,
+      name: 'Other Project',
+      foundation: false,
+    });
+
+    // Inline error state replaces the skeleton; no "not found" toast; URL stays on /edit.
+    await expect(page.getByTestId('meeting-manage-retry')).toBeVisible({ timeout: PAGE_LOAD_TIMEOUT });
+    await expect(page.getByTestId('meeting-manage-back')).toBeVisible();
+    await expect(page.getByTestId('meeting-manage-loading')).not.toBeVisible();
+    // Give a (never-fired) toast a render window — negative assertions pass instantly otherwise.
+    await page.waitForTimeout(500);
+    await expect(page.locator('.p-toast')).not.toBeVisible();
+    expect(page.url()).toContain(`/project/meetings/${MOCK_MEETING_UID}/edit`);
+  });
+
+  test('Retry after a transient failure re-fetches the meeting and renders the form (AC-1, AC-3)', async ({ page }) => {
+    await stubMeetingEditDetailError(page, 500);
+
+    await gotoSpa(page, `/project/meetings/${MOCK_MEETING_UID}/edit`, {
+      uid: OTHER_PROJECT_UID,
+      slug: OTHER_PROJECT_SLUG,
+      name: 'Other Project',
+      foundation: false,
+    });
+    await expect(page.getByTestId('meeting-manage-retry')).toBeVisible({ timeout: PAGE_LOAD_TIMEOUT });
+
+    // "Backend recovered": re-registering the detail route takes precedence over the error
+    // stub (Playwright matches in reverse registration order), so the retry fetch gets a 200.
+    await stubMeetingEditDetail(page, buildMeetingStub(true));
+    await page.getByTestId('meeting-manage-retry').click();
+
+    await expect(page.getByTestId('meeting-manage-stepper')).toBeVisible({ timeout: PAGE_LOAD_TIMEOUT });
+    await expect(page.getByTestId('meeting-manage-load-error')).not.toBeVisible();
+    expect(page.url()).toContain(`/project/meetings/${MOCK_MEETING_UID}/edit`);
+  });
+
+  test('a real 404 still ejects to the meetings list with the not-found toast (AC-2)', async ({ page }) => {
+    await stubMeetingEditDetailError(page, 404);
+
+    await gotoSpa(page, `/project/meetings/${MOCK_MEETING_UID}/edit`, {
+      uid: OTHER_PROJECT_UID,
+      slug: OTHER_PROJECT_SLUG,
+      name: 'Other Project',
+      foundation: false,
+    });
+
+    // Toast first: it fires synchronously with the eject and expires (default p-toast life)
+    // before a URL-first assertion would observe it.
+    await expect(page.locator('.p-toast')).toContainText('Meeting not found or you do not have permission to access it', {
+      timeout: PAGE_LOAD_TIMEOUT,
+    });
+    await expect(page).toHaveURL(/\/meetings(\?|$)/, { timeout: ELEMENT_TIMEOUT });
   });
 });
 
--- a/apps/lfx-one/e2e/project-context-deep-link.spec.ts
+++ b/apps/lfx-one/e2e/project-context-deep-link.spec.ts
@@ -289,6 +289,7 @@ async function stubMeetingEditDetailError(page: Page, status: number): Promise<v
     if (route.request().method() !== 'GET') return route.fallback();
     return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify({ data: [] }) });
   });
+  await page.route(`**/api/meetings/${MOCK_MEETING_UID}/attachments`, (route) => route.fulfill({ status: 200, contentType: 'application/json', body: '[]' }));
   await page.route(`**/api/meetings/${MOCK_MEETING_UID}`, (route) => {
     if (route.request().method() !== 'GET') return route.fallback();
     return route.fulfill({ status, contentType: 'application/json', body: JSON.stringify({ message: 'stub detail failure' }) });
@@ -658,6 +659,26 @@ test.describe('Meeting edit load failure (GH-2037)', () => {
       timeout: PAGE_LOAD_TIMEOUT,
     });
     await expect(page).toHaveURL(/\/meetings(\?|$)/, { timeout: ELEMENT_TIMEOUT });
+  });
+
+  test('non-ED writer persona: a transient probe failure redirects to the project overview (guard fail-closed)', async ({ page }) => {
+    // Documents the guard-level path the AC-1 tests can't reach: non-ED personas run the
+    // writerGuard entity probe before the component mounts, and a non-404 probe failure
+    // resolves no slug — the guard fails closed to /project/overview (no `_notice` toast)
+    // rather than authorizing against a possibly stale context (writer.guard.ts). The
+    // component's inline error state only renders for personas that bypass or pass the probe.
+    await setPersonaAndLensCookies(page, ['maintainer'], 'project');
+    await stubPersona(page, ['maintainer']);
+    await stubMeetingEditDetailError(page, 500);
+
+    await gotoSpa(page, `/project/meetings/${MOCK_MEETING_UID}/edit`, {
+      uid: OTHER_PROJECT_UID,
+      slug: OTHER_PROJECT_SLUG,
+      name: 'Other Project',
+      foundation: false,
+    });
+
+    await expect(page).toHaveURL(/\/project\/overview(\?|$)/, { timeout: PAGE_LOAD_TIMEOUT });
   });
 });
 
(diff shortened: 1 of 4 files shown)


audigregorie (commented):
## Code Review Summary

Solid, well-scoped fix: the 404-vs-everything-else discrimination is the right call, the retry trigger is folded into the existing `toSignal` pipeline declaratively (switchMap gives free cancellation on rapid retries), and the three new E2E tests cover stay-mounted / retry-recovers / 404-ejects. Two of the findings below overlap with Copilot's threads — restated here with concrete fixes and a recommendation on how to handle the guard gap. None are blockers; the guard-coverage one needs a tracked follow-up decision.

### Major — outside the diff

**The fix only reaches ED users in practice; non-ED personas still get ejected on transient errors.** The `:id/edit` route runs `writerGuard` with `entityScopedSlug: true` (`meetings.routes.ts:28`), so for Maintainers / project writers / meeting coordinators the guard's entity probe (`writer.guard.ts` — `getMeetingDetail(id)` ride-sharing the same 10s cache the component uses) fires *before* the component mounts. On a transient (non-404) probe failure the guard resolves no slug and silently redirects to `/project/overview` — no toast, no retry. Because a *successful* probe shares its cached response with the component, the component-level fetch this PR fixes can essentially only fail on its own for the ED fast path (or a >10s gap between guard and mount). The misleading-toast eject from GH-2037 is fixed; the same class of transient-error eject persists for the primary maintainer persona, just silently and to a different page. The guard's fail-closed stance is an intentional authorization control, so don't loosen it lightly — but please file a follow-up (or say explicitly why it's out of scope) so the persona gap is tracked. (Also flagged by Copilot; see its thread on the spec file.)

### Minor — outside the diff

- **Branch is far behind `origin/main`.** The merge-base (`82998b67`) predates ~8.6k lines of changes on main (117 files, incl. removals like `toast-message` and vote/campaigns refactors). The diff itself is clean, but rebase before merge so CI runs against current main.
- **Theoretical 403 edge:** if the upstream ever returns 403 on this fetch, the user now lands in a retry loop that can never succeed (previously: toast + eject). Today the detail fetch isn't permission-gated (access flags are added post-fetch and `writerGuard` owns authorization), so this is informational only — worth a mental note if the endpoint's gating ever changes.

### What's done well

- Conservative error discrimination: `instanceof HttpErrorResponse && status === 404` ejects; everything else (including non-HTTP errors) falls to the safe inline state — no path mislabels a server error as "not found".
- `retryMeetingLoad$` declared before `meeting` with a comment explaining the field-initialization-order constraint — genuinely load-bearing and easy to get wrong.
- `meetingLoadError` is reset at the top of each fetch attempt inside the pipeline, so param changes also clear a stale error state.
- E2E tests document the non-obvious mechanics inline (reverse route-registration precedence, why the toast assertion runs before the URL assertion).

### Rating: 4/5

apps/lfx-one/e2e/project-context-deep-link.spec.ts
```diff
@@ -540,6 +555,78 @@ test.describe('Meeting edit deep-link resolves the meeting’s project context (
   });
 });
 
+test.describe('Meeting edit load failure (GH-2037)', () => {
+  test.beforeEach(async ({ page }) => {
+    // Same shape as the gh-1432 meeting block: ED persona skips the writerGuard entity probe,
+    // so the component's own detail-fetch error branch is what runs.
```
audigregorie, line 596: **Major — ED-only coverage means the persona that hits this most (Maintainer/writer) is untested and still ejects on transient errors.**

This comment is accurate, and that's the gap: with the ED fast path, the component's error branch only runs when the guard never probed. For non-ED personas the `writerGuard` entity probe hits the same transient failure first and redirects to `/project/overview` — silently — before this component mounts. So the three new tests verify the fix on the one path where it applies, while the primary maintainer persona's transient-error UX is unchanged (see the review body for the full trace).

Not asking for a guard change in this PR — its fail-closed redirect is an intentional authorization control. But please either (a) add a test documenting the guard-level behavior for a writer persona on transient probe failure (even if the expected behavior is today's redirect), or (b) link a follow-up ticket covering the guard path, so the persona gap is tracked rather than implied.

apps/lfx-one/e2e/project-context-deep-link.spec.ts
```diff
@@ -276,6 +276,21 @@ async function stubMeetingEditDetail(page: Page, meeting: ReturnType<typeof buil
...
 
+/**
+ * Same route shapes as stubMeetingEditDetail, but the detail GET fulfills an error status —
+ * exercises meeting-manage's non-404 inline error state and 404 eject (GH-2037).
+ */
+async function stubMeetingEditDetailError(page: Page, status: number): Promise<void> {
+  await page.route('**/api/meetings*', (route) => {
+    if (route.request().method() !== 'GET') return route.fallback();
+    return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify({ data: [] }) });
+  });
```
audigregorie, line 291: **Minor — the error helper omits the attachments route, so the two stay-mounted tests leak a real request.**

`initializeAttachments()` fires `GET /api/meetings/:id/attachments` on every edit-page mount (it switches on `paramMap` only, independent of the detail fetch's outcome). Unlike `stubMeetingEditDetail`, this helper doesn't stub that route, and the `**/api/meetings*` catch-all can't match it (Playwright `*` doesn't cross `/` — the file's own `gotoSpa` doc calls this out). In the AC-1 tests the request escapes to the real BFF. It's swallowed by the service's `catchError(() => of([]))`, so the tests still pass, but they're no longer hermetic and could flake in environments where that call is slow or fails differently.

```suggestion
  });
  await page.route(`**/api/meetings/${MOCK_MEETING_UID}/attachments`, (route) => route.fulfill({ status: 200, contentType: 'application/json', body: '[]' }));
```

apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.html
```diff
@@ -35,7 +35,20 @@ <h1 class="mb-6" data-testid="meeting-manage-title">
...
+  @if (meetingLoadError()) {
+    <!-- Detail fetch failed with a non-404 error (GH-2037) — stay mounted with retry/back -->
+    <div class="flex flex-col gap-4" data-testid="meeting-manage-load-error">
+      <lfx-message
+        severity="error"
+        icon="fa-light fa-triangle-exclamation"
+        text="We couldn't load this meeting. Please try again, or go back to the meetings list."
+        data-testid="meeting-manage-load-error-message"></lfx-message>
+      <div class="flex gap-3">
+        <lfx-button label="Retry" icon="fa-light fa-rotate-right" (onClick)="retryMeetingLoad()" data-testid="meeting-manage-retry"> </lfx-button>
```
audigregorie: **Minor — Retry button is missing `severity="info"`.**

`lfx-button`'s `severity` input defaults to `undefined` (renders as PrimeNG primary), but every other solid action button in this template (Next, Create Meeting, Invite Guests, Update Meeting) explicitly sets `severity="info"`. Retry is the primary action of this error state, so it should match.

```suggestion
        <lfx-button label="Retry" icon="fa-light fa-rotate-right" severity="info" (onClick)="retryMeetingLoad()" data-testid="meeting-manage-retry"> </lfx-button>
```

apps/lfx-one/e2e/project-context-deep-link.spec.ts
```diff
@@ -276,6 +276,21 @@ async function stubMeetingEditDetail(page: Page, meeting: ReturnType<typeof buil
...
+  await page.route(`**/api/meetings/${MOCK_MEETING_UID}`, (route) => {
+    if (route.request().method() !== 'GET') return route.fallback();
+    return route.fulfill({ status, contentType: 'application/json', body: JSON.stringify({ message: 'stub detail failure' }) });
+  });
+}
```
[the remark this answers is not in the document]
audigregorie: Fixed in 0c6eb75a4 — `stubMeetingEditDetailError` now stubs `GET /api/meetings/:id/attachments`, so the AC-1 tests stay hermetic.

apps/lfx-one/e2e/project-context-deep-link.spec.ts
```diff
@@ -540,6 +555,78 @@ test.describe('Meeting edit deep-link resolves the meeting’s project context (
   });
 });
 
+test.describe('Meeting edit load failure (GH-2037)', () => {
+  test.beforeEach(async ({ page }) => {
+    // Same shape as the gh-1432 meeting block: ED persona skips the writerGuard entity probe,
+    // so the component's own detail-fetch error branch is what runs.
+    await setPersonaAndLensCookies(page, ['executive-director'], 'project');
```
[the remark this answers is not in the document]
audigregorie: Addressed in 0c6eb75a4 — added a maintainer-persona test documenting the guard's intentional fail-closed redirect to /project/overview on a transient probe failure. The guard flow itself is unchanged: fail-closed is a deliberate authorization control (see the guard's doc comment), so no guard change in this PR.

apps/lfx-one/e2e/project-context-deep-link.spec.ts
audigregorie: Went with (a) in 0c6eb75a4 — added the guard-level writer-persona test asserting the transient-failure redirect to /project/overview, so the persona gap is documented in-repo.

apps/lfx-one/e2e/project-context-deep-link.spec.ts
audigregorie: Fixed in 0c6eb75a4 — applied the suggestion; the error helper now stubs the attachments route.

apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.html
audigregorie: Fixed in 0c6eb75a4 — Retry now sets severity="info", matching the template's other primary actions.

apps/lfx-one/e2e/project-context-deep-link.spec.ts
```diff
@@ -280,6 +280,21 @@ async function stubMeetingEditDetail(page: Page, meeting: ReturnType<typeof buil
...
+async function stubMeetingEditDetailError(page: Page, status: number): Promise<void> {
+  await page.route('**/api/meetings*', (route) => {
+    if (route.request().method() !== 'GET') return route.fallback();
+    return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify({ data: [] }) });
+  });
+  await page.route(`**/api/meetings/${MOCK_MEETING_UID}`, (route) => {
+    if (route.request().method() !== 'GET') return route.fallback();
+    return route.fulfill({ status, contentType: 'application/json', body: JSON.stringify({ message: 'stub detail failure' }) });
+  });
+}
```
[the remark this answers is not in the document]
audigregorie: Fixed in 0c6eb75a4 — `stubMeetingEditDetailError` now stubs the attachments route; the stay-mounted tests no longer leak a real request.

luismoriguerra (dismissed):
## Audit: PR #2164 — fix(meetings): keep edit page mounted on transient load errors (#2037)

**Verdict:** ✅ Approved  
**Merge readiness:** merge-after-confirming-LFX-Platform-review — CI checks green on `0c6eb75a`; `linuxfoundation/lfx-platform` team review still required.

**Gates (local re-run):** `format:check` ✅ · `lint:check` ✅ · `build` ✅  
**Gates (CI on head):** Code Quality Checks ✅ (format, lint, types, unit tests, build) · CodeQL ✅ · DCO ✅

**Findings (none blocking):**
1. **Non-blocking** — `meeting-manage.component.ts:996-1005` — Guard-level transient UX for non-ED writer personas unchanged (`writer.guard.ts` probe 5xx → `/project/overview`); component fix applies when guard bypasses/passes. E2E documents maintainer path (`project-context-deep-link.spec.ts:664-682`). Optional follow-up ticket.
2. **Non-blocking** — `meeting-manage.component.spec.ts` — No unit tests for new `meetingLoadError` / `retryMeetingLoad()`; E2E covers AC-1/AC-2/AC-3.
3. **Nit** — `meeting-manage.component.ts:159` — `meetingLoading` stays true during error state; template `@if` ordering handles it.
4. **Nit** — `meeting-manage.component.ts:996-1005` — 403 would land in retry UI (theoretical; auth is guard-owned today).

No blocking findings — summary only, no inline threads (concise + default `blocking` selection on ✅). Say `all` if you want non-blocking items anchored inline.

Independent audit via `pr-reviewer-agent` (ephemeral worktree, same code as head).

luismoriguerra:
**Audit correction (finding #2):** Rewording — the PR's sole regression coverage for this fix is 4 Playwright tests in `project-context-deep-link.spec.ts` that **neither CI** (`quality-check.yml:80-89` e2e job commented out; `e2e-tests.yml` needs AWS test-user secrets) **nor this audit** executed. Recommend adding unit tests for `meetingLoadError` / `retryMeetingLoad()` in `meeting-manage.component.spec.ts` — that's the only suite CI runs (`yarn test`). Still non-blocking; prior wording incorrectly implied verified E2E coverage.

apps/lfx-one/src/app/modules/meetings/meeting-manage/meeting-manage.component.ts
```diff
@@ -145,9 +148,15 @@ export class MeetingManageComponent {
     toUpdate: [],
     toDelete: [],
   });
+  // True only when the edit-mode detail fetch failed with a non-404 error (GH-2037) — the
+  // template swaps the skeleton for an inline error + Retry instead of ejecting the user.
```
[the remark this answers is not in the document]
audigregorie: Fixed in f353036 — reworded the signal comment, retryMeetingLoad() JSDoc, and the template comment to describe the failure as a retryable error (anything other than a 404/403 eject), so a 403 can't be misread as retryable.

luismoriguerra (approved):
## ✅ Approved

Focused UX fix for GH-2037: only **404/403** eject from meeting edit; transient 5xx/network errors stay mounted with inline error + Retry.

**Verified:** CI green; unit tests 17/17 (incl. new GH-2037 cases); lint clean on changed files.

**Highlights:**
- `meetingLoadError` + `retryMeetingLoad$` follow existing signal/Subject patterns
- 403 on eject path is correct (guard-owned write access; retry can't restore)
- Unit + E2E cover AC-1/2/3; maintainer probe-failure case documents guard boundary

**Nits (non-blocking):** PR summary says "404 only" but implementation correctly handles 404+403; minor E2E flake risk on negative toast assertion.

PR #2173, opened 2026-09-03
test(votes): cover getVotes opt-out and vote-manage eviction route (#2084)

### LINKED ISSUE #2084: Protect the vote activity performance fix with regression tests and accurate docs
## Problem

The vote editing scoping work shipped with a performance guard that lets the Recent Activity view skip an unnecessary project lookup when fetching vote lists — but nothing protects that guard. If a future change accidentally removes it, every activity load silently pays the extra round trip again, and no test would fail. Alongside that, the documentation for opting out of vote enrichment understates what a caller gives up (it omits the parent-project reference that dashboard filtering depends on), and one access-loss test claims to verify where the user is sent when they lose write access mid-edit, but only checks that they are sent somewhere.

These were raised as the remaining follow-ups in the round-3 review of PR #2026 (one minor, two nits); the PR merged with them still open.

## Outcome

The performance guard is held by a regression test, so dropping the opt-out breaks the build instead of shipping silently. The enrichment opt-out documentation names exactly which fields a caller loses, and the eviction test's name matches what it actually proves (or it is strengthened to prove the destination).

## Success criteria

- A service-level test fails if vote list fetching stops honoring the request to skip project enrichment
- The opt-out documentation names all four fields a caller forfeits, matching the sibling pattern used elsewhere
- The write-access eviction test either asserts the overview destination or is renamed to match what it verifies

audigregorie:
## Summary

Follow-up test-hardening PR for the vote enrichment guard. It pins two behaviors that previously had no automated proof: the `getVotes` `includeProject: false` opt-out skipping project enrichment entirely, and the vote-manage write-access eviction landing on the project/foundation overview. No runtime behavior changes — one docblock correction included.

Resolves #2084

### Vote list project enrichment opt-out

| Before | After |
|---|---|
| Callers could ask for the vote list without project details, but no automated test proved the skip actually happened — a future change could quietly start fetching project data again, adding hidden load and leaking fields the caller asked to discard. | An automated test now proves that opting out performs zero project lookups and returns rows with none of the project fields attached, so any regression fails the build immediately. |

### Vote management write-access eviction

| Before | After |
|---|---|
| When a user's write access was revoked while managing a vote, the app redirected them away — but the test only checked that *some* navigation happened, not that it landed on the expected overview page. | The test now confirms the redirect goes specifically to the project (or foundation) overview page, catching any future change to the eviction destination. |

## Technical changes

`apps/lfx-one/src/server/services/vote.service.spec.ts` — Server vote service tests
- adds a `getVotes` spec proving `{ includeProject: false }` never calls `getProjectsByIds` or `computeIsFoundation`, and that returned rows carry no `project_slug`, `is_foundation`, or `parent_project_uid` — mutation-checked against a reverted gate

`apps/lfx-one/src/server/services/vote.service.ts` — Server vote service docblock
- corrects the `getVotes` docblock to name all four enriched fields (`project_name`, `project_slug`, `is_foundation`, `parent_project_uid`) instead of the previous vague "slug/name/tier"

`apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.spec.ts` — Vote manage component tests
- strengthens the write-access eviction spec to assert `router.parseUrl` is called with `/project/overview` or `/foundation/overview`, not just that `navigateByUrl` fired

### DIFF
--- a/apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/votes/vote-manage/vote-manage.component.spec.ts
@@ -214,5 +214,6 @@ describe('VoteManageComponent', () => {
     await TestBed.inject(ApplicationRef).whenStable();
 
     expect(router.navigateByUrl).toHaveBeenCalled();
+    expect(router.parseUrl).toHaveBeenCalledWith(expect.stringMatching(/^\/(project|foundation)\/overview$/));
   });
 });
--- a/apps/lfx-one/src/server/services/vote.service.spec.ts
+++ b/apps/lfx-one/src/server/services/vote.service.spec.ts
@@ -279,6 +279,18 @@ describe('VoteService upstream path encoding', () => {
       expect(result.data[0]).not.toHaveProperty('project_slug');
       expect(result.data[0]).not.toHaveProperty('is_foundation');
     });
+
+    it('skips project enrichment entirely when includeProject is false', async () => {
+      proxyRequest.mockResolvedValue({ resources: [{ data: indexRow }], page_token: undefined });
+
+      const result = await service.getVotes(req, {}, { includeProject: false });
+
+      expect(getProjectsByIds).not.toHaveBeenCalled();
+      expect(computeIsFoundation).not.toHaveBeenCalled();
+      expect(result.data[0]).not.toHaveProperty('project_slug');
+      expect(result.data[0]).not.toHaveProperty('is_foundation');
+      expect(result.data[0]).not.toHaveProperty('parent_project_uid');
+    });
   });
 
   describe('getMyVotes', () => {
--- a/apps/lfx-one/src/server/services/vote.service.ts
+++ b/apps/lfx-one/src/server/services/vote.service.ts
@@ -42,7 +42,7 @@ export class VoteService {
 
   /**
    * Fetches a single page of votes using cursor-based pagination — callers paginate via the returned page_token.
-   * `includeProject` (default true) enriches rows with project slug/name/tier; opt out when the caller discards them.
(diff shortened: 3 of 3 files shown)


MRashad26 (approved):
Reviewed the full diff (3 files, +14/-1) and cross-checked every assertion against the actual current code — small diff, but I traced it fully since test-only PRs are easy to rubber-stamp incorrectly.

**What this does (#2084):** Pure test-coverage addition, no runtime behavior change, plus one docblock correction.

- `VoteService.getVotes`'s enrichment gate is `const votes = includeProject ? await this.enrichWithProjectMetadata(req, normalized) : normalized;` — confirmed `enrichWithProjectMetadata` internally calls `this.projectService.getProjectsByIds` and `computeIsFoundation`, exactly the two calls the new test asserts are skipped when `includeProject: false`. The new assertions (`not.toHaveBeenCalled()` for both, plus the three enrichment fields absent from the result row) genuinely pin the opt-out path.
- The docblock change (`slug/name/tier` → naming all four actual fields: `project_name`, `project_slug`, `is_foundation`, `parent_project_uid`) matches what `enrichWithProjectMetadata` actually attaches — the old docblock was vague/stale, this is a straight accuracy fix.
- Traced the new `vote-manage.component.spec.ts` assertion (`expect(router.parseUrl).toHaveBeenCalledWith(expect.stringMatching(/^\/(project|foundation)\/overview$/))`) against the shared `evictOnWriteAccessLoss` utility that `VoteManageComponent` delegates its eviction redirect to: the utility builds the redirect URL as `slug ? router.createUrlTree([overviewPath], {...}) : router.parseUrl(overviewPath)`, calling `parseUrl` only when `projectContextService.activeContext()?.slug` is falsy. In this test's setup, `getProject.mockReturnValue(of(null))` means no project ever gets set into context, so `activeContext()?.slug` is falsy for this scenario and the `parseUrl` branch is genuinely the one exercised — the assertion matches the actual code path, not a mismatched mock.

No issues. Approving.

Issue #1432, opened 2026-08-12
Bug: Meeting edit link can resolve to the wrong project context


DerrickLFX:
### What happened?

Summary
A meeting-edit link for a meeting belonging to one project can load in the context of a different, unrelated project. Example: app.lfx.dev/project/meetings/96286151238/edit?step=5 is a GraphQL Foundation meeting, but following the link lands in the EnerGNN project context instead.

Steps to Reproduce

Obtain an edit link for a meeting belonging to Project A.
Open the link as a user with access to multiple projects.
→ Page loads in the context of a different project instead of Project A.

Additional Context
Every meeting is mapped to a single project internally, so the correct project context should be derivable from the meeting ID alone. May be related to Bug 2 — testing suggested wrong-project behavior and intermittent "access denied" errors could be tied to the URL inconsistency — but filing separately since root causes may differ.

### Command / Request

```shell

```

### Relevant log output

```shell

```

PR #2174, opened 2026-09-03
fix(surveys): edit links open in the survey's project context (GH-1569)

### LINKED ISSUE #1569: Survey edit links can open in the wrong project context
## Problem

When someone opens a link to edit a survey — copied from the surveys table or shared by a teammate — the page can load under whichever project the viewer happened to visit last instead of the project that actually owns the survey. Survey managers editing a survey that belongs to a different project (or directly to a foundation) see the wrong project in the sidebar and project picker, can be intermittently denied access even though they manage that survey, and the URL flips between the /project/ and /foundation/ paths depending on the viewer's last-visited area rather than the survey's ownership.

## Outcome

Survey edit links always open under the project or foundation that owns the survey: the URL path reflects foundation ownership when there is one, the page loads with the owning project selected, and access checks run against the owning project. Links shared with teammates land in the right place regardless of what the recipient was looking at before.

## Success criteria

- Opening a survey's edit link while a different project is selected still loads the page in the owning project's context (sidebar, project picker, committee selection).
- A foundation-owned survey's edit links use the /foundation/ path; other surveys use /project/ — regardless of the viewer's current lens.
- A legitimate survey manager (including committee writers) is not sent to an access-denied page when opening an edit link that carries no project context.
- Copying the URL mid-way through the edit wizard and opening it fresh keeps the correct project.

audigregorie:
## Summary

Survey edit links previously inherited the viewer's current project context: opening one from a different lens — or from a context-less deep link — could load the edit page under the wrong project and even authorize against it. This PR makes survey editing entity-scoped end to end: list rows derive their edit URL from the survey's owning tier, the manage page syncs the active project context from the loaded survey itself, and the writer guard authorizes against the survey's own project. Mirrors the meetings pattern from gh-1432.

Resolves #1569

### Edit links in the surveys list

| Before | After |
|---|---|
| The Edit action on every survey row pointed to the same generic edit address, prefixed by whichever view (foundation or project) the user happened to be browsing, and carried no information about which project or committee owned the survey. | Each row's Edit action points to the canonical address for that survey's owning organization (foundation vs. project) and carries the survey's own project and committee identifiers, regardless of the view the user is browsing from. |

### Landing on the edit page directly

| Before | After |
|---|---|
| Opening a survey edit link with no project information (a bookmark, a shared link) kept whichever project had been selected previously — often an unrelated one — and a momentary server hiccup could bounce the user to the surveys list with a misleading "survey not found" message. | The edit page reads the survey being edited and switches the active project to the survey's owner — even when a second lookup is needed to identify it — and only a genuine "not found" sends the user back to the list; temporary errors leave the page in place. |

### Edit permission check

| Before | After |
|---|---|
| The permission check for editing a survey ran against whichever project was currently selected, so a context-less edit link could be allowed or denied based on the wrong project. | The permission check runs against the project that actually owns the survey; if the survey can't be loaded because of a server error, access is denied rather than decided against a stale selection. |

## Technical changes

`apps/lfx-one/src/server/services/survey.service.ts, apps/lfx-one/src/server/controllers/survey.controller.ts` — BFF project enrichment
- List: flattens `project_uid` from `committees[0]` (the upstream index sends no top-level uid) and enriches every row with `project_slug` / `project_name` / `is_foundation` via `enrichWithProjectData`
- Detail: new `includeProject` opt-in on `getSurveyById` (wired in the controller) that flattens `project_uid ??= committees[0].project_uid` and stamps the same project fields via the shared `entity-project-enrichment` helper — no-op without a uid, warn-and-continue on lookup failure
- The flatten fixes the previously inert deep-link half: the detail payload now always carries a real `project_uid` for the guard probe and the manage page's resolve-by-uid fallback

`apps/lfx-one/src/app/modules/surveys/components/surveys-table/surveys-table.component.ts, apps/lfx-one/src/app/modules/surveys/components/surveys-table/surveys-table.component.html` — Per-row canonical edit links
- Rows are stamped with `editCommands` derived from the survey's owning tier via `getEntityCommands` (`/foundation/...` vs `/project/...`); rows whose payload lacks tier data fall back to the flat path, which `lensRedirectGuard` prefixes
- `editQueryParams` are built per row from the survey's own `project_slug` plus committee scope — the new `committeeUid` input (the viewed committee) wins over `committees[0]`
- The Edit action binds the per-row link on the button's real-anchor `routerLink`; the explicit `stopPropagation` was dropped because PrimeNG's selectable row already skips clicks on anchors/buttons, and a consumer `stopPropagation` can starve the RouterLink listener

`apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.ts, apps/lfx-one/src/app/modules/committees/components/committee-surveys/committee-surveys.component.html` — Committee surveys tab
- Passes the viewed committee's `uid` to the table instead of the old shared `editSurveyQueryParams` signal (removed), so a committee writer's guard fallback probes the committee actually being viewed

`apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts` — Edit-page context sync
- Edit mode now fetches the survey being edited and adapts it to `EntityWithProject`, driving `syncEntityProjectContext` — a context-less deep link under a stale cookie-restored context corrects to the survey's project (mirrors meeting-manage, gh-1432)
- `syncEntityProjectContextFallback` resolves the project by uid when the payload is unenriched (uid but no slug), with an uncached fresh re-read as a last resort for committee writers whose relation-gated project lookup resolves null
- Step navigation merges query params so `?project=` and `committee_uid` survive step changes
- Only a real 404 ejects to `/surveys`; a transient 5xx/network error leaves the edit page mounted (the fetch exists for context sync, not form pre-population)

`apps/lfx-one/src/app/shared/guards/writer.guard.ts, apps/lfx-one/src/app/modules/surveys/surveys.routes.ts` — Entity-scoped authorization
- `surveys` added to the guard's entity-probe registry; the edit route opts in via `data.entityScopedSlug: true` (a route-data flag, so a rename can't silently revert to stale-context authorization)
- The probe maps absent `project_uid` to `''` (treated as absent); a 404 read falls back to the active context, any other failure resolves no slug and redirects to `/project/overview`

`apps/lfx-one/src/app/shared/services/survey.service.ts` — Shared short-TTL detail cache
- `getSurvey` gained a 10-second `shareReplay` cache (`SURVEY_DETAIL_CACHE_TTL_MS`) so the guard probe and the manage-page fetch share one request per navigation — no signal side-effects, probe-friendly
- `skipCache` escape hatch for the fallback's fresh fetch; entries evict on error, on an expiry sweep, and on delete (keys carry the `project_id` variant suffix)
- Mirrors `MailingListService.getMailingList`

`packages/shared/src/constants/survey.constants.ts, packages/shared/src/interfaces/survey.interface.ts` — Shared contracts
- New `SURVEY_DETAIL_CACHE_TTL_MS` constant and `SurveyTableRow` type (`Survey` + `displayStatus` + per-row edit link)
- `Survey` field docs updated: `project_uid` is flattened from `committees[0]` by the BFF on list and detail; enrichment fields documented as absent-on-failure so consumers treat `undefined` as "tier unknown"

`apps/lfx-one/src/app/shared/guards/writer.guard.spec.ts` — Guard unit tests
- Survey probe coverage: enriched slug authorizes against the survey's project, a uid-only payload resolves by uid (never the stale context), a 404 falls back to the active context, and a 500 fails closed without probing the stale project

`apps/lfx-one/e2e/survey-edit-url.spec.ts, apps/lfx-one/e2e/project-context-deep-link.spec.ts` — E2E coverage
- Click-through: foundation-owned, project-owned, and unenriched rows land on the correct tier URL with `?project=` / `committee_uid` (seeding the opposite lens to reproduce the pre-fix bug)
- Deep-link: a context-less edit link under a cookie-restored foreign project corrects the context (enriched and uid-fallback paths) without backfilling `?project=` into the URL, and the guard authorizes against the survey's project rather than the stale context
- Stubs mirror the real BFF detail contract — project identity lives inside `committees[]`; slug/name/tier appear only when enrichment succeeded

### DIFF
--- a/apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
@@ -105,16 +105,27 @@ export class SurveyManageComponent {
   public constructor() {
     // Sync context from the loaded survey itself — an edit deep link can arrive with no
     // `?project=` under a stale cookie-restored context (GH-1569). Mirrors meeting-manage (gh-1432).
-    syncEntityProjectContext(this.surveyEntityContext, this.projectContextService, this.router, this.destroyRef, { preferEntityKind: true });
+    // canonicalizeRoute rewrites a wrong-tier URL (/project/* vs /foundation/*) to the survey's own
+    // tier so copied wizard links reflect ownership — mirrors vote-manage/mailing-list-manage.
+    syncEntityProjectContext(this.surveyEntityContext, this.projectContextService, this.router, this.destroyRef, {
+      preferEntityKind: true,
+      canonicalizeRoute: true,
+    });
     // Unenriched-payload fallback (project_uid but no project_slug): resolve the project by uid;
     // freshFetch re-reads the detail uncached as a last resort for a committee writer whose
     // relation-gated project lookup resolves null (the enrichment needs no project relation).
     syncEntityProjectContextFallback(this.surveyEntityContext, this.projectService, this.projectContextService, this.router, this.destroyRef, {
       entityKind: 'survey',
+      canonicalizeRoute: true,
       freshFetch: (uid) =>
-        this.surveyService
-          .getSurvey(uid, undefined, { skipCache: true })
-          .pipe(map((survey) => ({ project_uid: survey.project_uid ?? '', project_slug: survey.project_slug, project_name: survey.project_name, is_foundation: survey.is_foundation ?? null }))),
+        this.surveyService.getSurvey(uid, undefined, { skipCache: true }).pipe(
+          map((survey) => ({
+            project_uid: survey.project_uid ?? '',
+            project_slug: survey.project_slug,
+            project_name: survey.project_name,
+            is_foundation: survey.is_foundation ?? null,
+          }))
+        ),
     });
     this.initCommitteeContext();
     evictOnWriteAccessLoss();
--- a/apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
@@ -30,13 +30,14 @@ import { ProjectService } from '@services/project.service';
 import { SurveyService } from '@services/survey.service';
 import { syncEntityProjectContext, syncEntityProjectContextFallback } from '@shared/utils/entity-project-context.util';
 import { evictOnWriteAccessLoss } from '@shared/utils/evict-on-write-access-loss.util';
+import { resolveEntityWriteSlug } from '@shared/utils/write-access.util';
 import { MessageComponent } from '@components/message/message.component';
 import { markFormControlsAsTouched } from '@lfx-one/shared/utils';
 import { trimmedRequired } from '@lfx-one/shared/validators';
 import { ConfirmationService, MessageService } from 'primeng/api';
 import { ConfirmDialogModule } from 'primeng/confirmdialog';
 import { StepperModule } from 'primeng/stepper';
-import { catchError, combineLatest, distinctUntilChanged, filter, map, of, switchMap, take } from 'rxjs';
+import { catchError, combineLatest, distinctUntilChanged, filter, map, Observable, of, switchMap, take } from 'rxjs';
 
 import { SurveyAudienceTypeComponent } from '../components/survey-audience-type/survey-audience-type.component';
 import { SurveyEmailDraftComponent } from '../components/survey-email-draft/survey-email-draft.component';
@@ -90,10 +91,13 @@ export class SurveyManageComponent {
 
   // Complex computed/toSignal signals
   public readonly isEditMode: Signal<boolean> = this.initIsEditMode();
-  public readonly survey: Signal<Survey | null> = this.initSurvey();
+  private readonly survey: Signal<Survey | null> = this.initSurvey();
   // Survey → EntityWithProject adapter so the active project context syncs from the loaded
   // survey, not a stale cookie-restored context (GH-1569).
   private readonly surveyEntityContext: Signal<EntityWithProject | null> = this.initSurveyEntityContext();
+  // URL-stamped committee scope for the access predicate — mirrors writerGuard's committee leg.
+  private readonly committeeUidFromUrl = this.route.snapshot.queryParamMap.get('committee_uid');
+  private readonly writeAccess: Signal<boolean> = this.initWriteAccess();
   public readonly formValue: Signal<Record<string, unknown>> = this.initFormValue();
   public readonly canGoPrevious: Signal<boolean> = this.initCanGoPrevious();
   public readonly canGoNext: Signal<boolean> = this.initCanGoNext();
@@ -128,7 +132,9 @@ export class SurveyManageComponent {
         ),
     });
     this.initCommitteeContext();
-    evictOnWriteAccessLoss();
+    // Pass the guard-mirroring predicate — the default canWrite is project-writer-only and would
+    // evict guard-admitted committee writers the moment the context syncs to the survey's project.
+    evictOnWriteAccessLoss(this.writeAccess);
   }
 
   public nextStep(): void {
@@ -489,6 +495,105 @@ Thank you,
         project_name: survey.project_name,
         is_foundation: survey.is_foundation ?? null,
       };
+    });
+  }
+
+  /**
+   * Access predicate mirroring writerGuard's surveys standard — project writer on the survey's OWN
+   * project (never the boot context), or committee writer via the survey's own committees /
+   * ?committee_uid=; pending legs stay provisionally true so a resolving leg can't evict a
+   * guard-admitted committee writer mid-edit. Mirrors vote-manage's initWriteAccess (GH-1568).
+   */
+  private initWriteAccess(): Signal<boolean> {
+    const editSurveyId = this.route.snapshot.paramMap.get('id');
+    const projectKey$: Observable<string | null | undefined> = editSurveyId
+      ? toObservable(this.survey).pipe(
+          map((survey) => {
+            if (!survey) {
+              // Pending — in edit mode the authorization target comes from the survey itself.
+              return undefined;
+            }
+            // Mirror writerGuard's resolution order; the active-context fallback covers a survey
+            // carrying neither slug nor uid (the manage component owns that error path). The
+            // project_uid ?? '' mapping mirrors the guard probe — Survey.project_uid is typed
+            // optional and resolveEntityWriteSlug's || treats '' as absent.
+            return resolveEntityWriteSlug(
+              { project_slug: survey.project_slug, project_uid: survey.project_uid ?? '' },
+              this.projectContextService.activeContext()?.slug ?? null
+            );
+          })
+        )
+      : toObservable(this.projectContextService.activeContext).pipe(map((ctx) => ctx?.slug ?? null));
+
+    const projectAccess = toSignal(
+      projectKey$.pipe(
+        filter((key): key is string | null => key !== undefined),
+        distinctUntilChanged(),
+        switchMap((key) => {
+          if (!key) {
+            return of(false);
+          }
+          // writerGuard admits no coordinator-style role for surveys — project.writer only.
+          return this.projectService.getProject(key, false).pipe(
+            map((project) => project?.writer === true),
+            catchError(() => of(false))
+          );
+        })
+      )
+      // No initialValue: undefined doubles as the leg's pending state (see the doc above).
+    );
+    // Edit mode mirrors writerGuard's surveys probe selection: the URL ?committee_uid= wins only
+    // when the survey's own committees[] contains it, else the primary committee; a committee-less
+    // survey falls back to the URL param. Hoisted: toObservable needs the injection context, so it
+    // can't be created lazily inside the switchMap below.
+    const editCommitteeUid$: Observable<string | null | undefined> = toObservable(this.survey).pipe(
+      map((survey) => {
+        if (!survey) {
+          return undefined;
+        }
+        const url = this.committeeUidFromUrl;
+        const entityUid = url && survey.committees?.some((c) => c.committee_uid === url) ? url : survey.committees?.[0]?.committee_uid;
+        return entityUid ?? url;
+      })
+    );
+    // Gated on the project leg resolving false: a granted project leg makes the committee fetch
+    // moot, and while it's pending this leg stays pending too. undefined stays pending; null resolves immediately.
+    const committeeUid$: Observable<string | null | undefined> = toObservable(projectAccess).pipe(
+      filter((project): project is boolean => project !== undefined),
+      distinctUntilChanged(),
+      switchMap((project) => {
+        if (project) {
+          // Project leg granted — resolve the committee leg without firing the HTTP probe.
+          return of<string | null>(null);
+        }
+        return editSurveyId ? editCommitteeUid$ : of(this.committeeUidFromUrl);
+      })
+    );
+    const committeeAccess = toSignal(
+      committeeUid$.pipe(
+        filter((uid): uid is string | null => uid !== undefined),
+        distinctUntilChanged(),
+        switchMap((uid) =>
+          uid
+            ? this.committeeService.fetchCommittee(uid).pipe(
+                map((committee) => committee?.writer === true),
+                catchError(() => of(false))
+              )
+            : of(false)
+        )
+      )
+      // No initialValue: undefined doubles as the leg's pending state (see the doc above).
+    );
+    return computed(() => {
+      const project = projectAccess();
+      const committee = committeeAccess();
+      if (project === true || committee === true) {
+        return true;
+      }
+      if (project === undefined || committee === undefined) {
+        return true; // provisional — a pending leg can still grant access
+      }
+      return false;
     });
   }
 
--- a/apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
+++ b/apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
@@ -1,10 +1,10 @@
 // Copyright The Linux Foundation and each contributor to LFX.
 // SPDX-License-Identifier: MIT
 
-import { Component, computed, inject, Signal, signal } from '@angular/core';
-import { toObservable, toSignal } from '@angular/core/rxjs-interop';
+import { Component, computed, DestroyRef, inject, Signal, signal } from '@angular/core';
+import { takeUntilDestroyed, toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
-import { ActivatedRoute, Router, RouterLink } from '@angular/router';
+import { ActivatedRoute, NavigationEnd, Router, RouterLink } from '@angular/router';
 import { ButtonComponent } from '@components/button/button.component';
 import {
   SCHEDULE_SURVEY_CONFIRMATION,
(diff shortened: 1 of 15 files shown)


apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -87,6 +89,9 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
     committees: (id) => committeeService.fetchCommittee(id),
     votes: (id) => voteService.fetchVote(id),
     'mailing-lists': (id) => mailingListService.getMailingList(id),
+    // Survey.project_uid is typed optional — map absent to '' so the probe satisfies the
+    // registry's Pick<EntityWithProject> shape; resolveEntityWriteSlug treats '' as absent.
+    surveys: (id) => surveyService.getSurvey(id).pipe(map((survey) => ({ project_slug: survey.project_slug, project_uid: survey.project_uid ?? '' }))),
```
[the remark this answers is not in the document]
audigregorie: Fixed in 9bc0ec176 — the surveys probe now surfaces the survey's primary committee_uid (committees[0], same source the BFF flattens project_uid from), so the entity's own committee wins over the ?committee_uid= URL param, mirroring the votes probe.

apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
```diff
@@ -84,6 +103,19 @@ export class SurveyManageComponent {
...
+    // Unenriched-payload fallback (project_uid but no project_slug): resolve the project by uid;
+    // freshFetch re-reads the detail uncached as a last resort for a committee writer whose
+    // relation-gated project lookup resolves null (the enrichment needs no project relation).
+    syncEntityProjectContextFallback(this.surveyEntityContext, this.projectService, this.projectContextService, this.router, this.destroyRef, {
+      entityKind: 'survey',
+      freshFetch: (uid) =>
+        this.surveyService
+          .getSurvey(uid, undefined, { skipCache: true })
+          .pipe(map((survey) => ({ project_uid: survey.project_uid ?? '', project_slug: survey.project_slug, project_name: survey.project_name, is_foundation: survey.is_foundation ?? null }))),
+    });
```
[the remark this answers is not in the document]
audigregorie: Fixed in 9bc0ec176 — canonicalizeRoute: true is now enabled on both syncEntityProjectContext and syncEntityProjectContextFallback, so a wrong-tier survey edit URL (/project/* vs /foundation/*) is rewritten to the survey's owning tier, matching the vote and mailing-list edit flows.

apps/lfx-one/src/server/services/survey.service.ts
```diff
@@ -112,6 +123,19 @@ export class SurveyService {
       });
     }
 
+    // The upstream detail payload (SurveyScheduleResult) carries project identity only inside
+    // committees[] — flatten the primary project uid and stamp it top-level so the enrichment
+    // below, the writerGuard probe, and the manage page's resolve-by-uid fallback all receive a
+    // real uid (GH-1569; mirrors the list path's flatten in getSurveys).
+    survey.project_uid ??= survey.committees?.[0]?.project_uid;
```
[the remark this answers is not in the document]
audigregorie: Fixed in 801ffa315 — both flatten sites (list L92 and detail L130) now use `||` instead of `??`/`??=`, so an empty-string `project_uid` also falls back to `committees[0].project_uid`.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -87,6 +89,17 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
...
+    // registry's Pick<EntityWithProject> shape; resolveEntityWriteSlug treats '' as absent.
+    // Surface the primary committee's uid (same source the BFF flattens project_uid from) so the
+    // entity's own committee wins over the attacker-controlled ?committee_uid= URL param — mirrors
+    // the votes probe; a committee-less project survey falls back to the URL param.
+    surveys: (id) =>
+      surveyService
+        .getSurvey(id)
+        .pipe(
+          map((survey) => ({ project_slug: survey.project_slug, project_uid: survey.project_uid ?? '', committee_uid: survey.committees?.[0]?.committee_uid }))
+        ),
```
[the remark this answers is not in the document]
audigregorie: Fixed in 801ffa315 — the surveys probe now honors the URL `committee_uid` when it is present in the survey's own committee list, and falls back to the primary committee otherwise, so a writer of any associated committee is admitted while an unrelated param still can't win.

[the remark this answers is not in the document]
audigregorie: Fixed in 801ffa315 — exactly this: the URL `committee_uid` is honored only when found in `survey.committees`; otherwise the probe falls back to `committees[0]`.

audigregorie (commented):
## Code Review Summary

Solid entity-scoping mechanism end to end — the BFF split (relation-gated list enrichment vs. ungated query-service detail enrichment), the shared short-TTL detail cache, and the E2E coverage are all well done. Two gaps keep this from landing clean: the manage page's bare `evictOnWriteAccessLoss()` evicts the very committee writers the guard now admits (1 Critical), and the survey probe's committee-selection hardening ships without the unit coverage its votes sibling has (1 Major). Two minors inline.

Verified during review: the `stopPropagation` removal is safe against PrimeNG 20 (`handleRowClick` early-returns on `A`/`BUTTON`/`INPUT` targets, and this label-only anchor is covered); the cache is instance-scoped so SSR requests can't leak across users; guard probe + manage fetch share one request per navigation within the 10s TTL.

### What's done well

- The list/detail enrichment split is exactly right for the committee-writer case, with the security reasoning documented in `entity-project-enrichment.helper.ts`.
- The frontend cache faithfully mirrors `MailingListService.getMailingList` (error eviction, expiry sweep, per-`project_id` keys, `skipCache`).
- E2E specs seed the *opposite* lens to prove URLs follow the entity, and assert the no-`?project=`-backfill contract.
- `entityScopedSlug` as route data (not a path check) is rename-safe — good guard design.

apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
```diff
@@ -84,6 +103,30 @@ export class SurveyManageComponent {
...
+          map((survey) => ({
+            project_uid: survey.project_uid ?? '',
+            project_slug: survey.project_slug,
+            project_name: survey.project_name,
+            is_foundation: survey.is_foundation ?? null,
+          }))
+        ),
+    });
     this.initCommitteeContext();
     evictOnWriteAccessLoss();
```
audigregorie: **Critical — bare `evictOnWriteAccessLoss()` evicts the committee writers this PR admits**

The constructor now syncs the active context to the *survey's* project (`syncEntityProjectContext` above), but this call keeps the default predicate — `ProjectContextService.canWrite`, which is **project-writer-only** and keyed off the active context. writerGuard admits survey editors via an additional leg (committee writer, `COMMITTEE_WRITE_FEATURES` includes `'surveys'`), and this PR's new `committeeUid` table input exists specifically to route committee writers into this page.

The sequence for a committee writer (writer on the survey's committee, not on its project):

1. writerGuard admits via the entity-scoped committee leg.
2. The context sync flips `activeContext` to the survey's project.
3. `canWrite` re-fetches `getProject(surveyProject.slug)` → `writer === false` → emits `false`. `skip(1)` only consumed the boot emission, so this first resolved `false` passes the filter and `take(1)` fires.
4. The user is redirected to the overview mid-edit, losing form state — bounced off the page the guard just admitted them to.

`evict-on-write-access-loss.util.ts` documents this exact contract: *"Routes whose writerGuard admits additional roles… must pass an `access` signal matching that standard; otherwise a context switch to the entity's project (syncEntityProjectContext) produces a true→false transition and evicts guard-admitted organizers mid-edit."* meeting-manage (gh-1432), vote-manage (GH-1568), and committee-manage all pair the sync with a custom `initWriteAccess()`; vote-manage's doc even calls out the stale boot context's `false` winning the race against the sync correction.

**Fix**: add an `initWriteAccess(): Signal<boolean>` mirroring the guard's surveys standard — project leg keyed off `resolveEntityWriteSlug(survey, activeContext)`, committee leg via the entity-scoped committee uid (same membership rule as the probe, URL param as fallback), each leg provisionally `true` while pending — and pass it here:

```suggestion
    evictOnWriteAccessLoss(this.writeAccess);
```

vote-manage's `initWriteAccess` (~lines 618–680) is a near drop-in template, including the create-mode `committeeUidFromUrl` leg (which also fixes the same eviction hazard for committee writers on the create route).

apps/lfx-one/src/app/shared/guards/writer.guard.spec.ts
```diff
@@ -332,6 +346,51 @@ describe('writerGuard', () => {
     expect(getCommittee).not.toHaveBeenCalled();
   });
 
+  it('authorizes survey edit against the survey’s own project when the read succeeds', async () => {
```
audigregorie, line 349: **Major — the survey probe's committee-selection branch has no unit coverage**

The four new tests cover slug resolution well (enriched / uid-only / 404 fallback / 500 fail-closed), but the probe's other half — the committee-selection ternary at `writer.guard.ts:104` — is untested. That branch is the PR's authorization hardening: `?committee_uid=` is honored only when the survey's own `committees[]` contains it, otherwise the probe falls back to `committees[0]` so an attacker-controlled param naming an unrelated committee can't admit its writer.

The votes probe has exactly these tests ("admits a committee writer via the vote's own committee_uid when the URL omits it", "authorizes against the vote's own committee, not a URL committee_uid naming an unrelated one"), and the surveys logic is strictly more complex — array membership + optional chaining instead of a single scalar field.

Suggest adding:

1. URL `committee_uid` present in `survey.committees` → `getCommittee` called with that uid.
2. URL param naming an *unrelated* committee → `getCommittee` called with `committees[0].committee_uid`, never the attacker's.
3. No param → `committees[0]`.
4. Committee-less survey (`committees: []`) → URL param fallback (pins the documented deliberate path so it isn't "fixed" away later).

apps/lfx-one/src/app/shared/guards/writer.guard.ts
```diff
@@ -87,6 +89,21 @@ export const writerGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
...
+    // committee tab stamps the viewed committee, and a writer of any associated committee must be
+    // admitted); otherwise fall back to the primary committee so an attacker-controlled param naming
+    // an unrelated committee can't win — a committee-less project survey falls back to the URL param.
+    surveys: (id) =>
+      surveyService.getSurvey(id).pipe(
+        map((survey) => ({
+          project_slug: survey.project_slug,
+          project_uid: survey.project_uid || '',
+          committee_uid:
+            committeeUid && survey.committees?.some((c) => c.committee_uid === committeeUid) ? committeeUid : survey.committees?.[0]?.committee_uid,
```
audigregorie, line 107: **Minor — multi-committee surveys: writers of a non-primary committee are denied from the global list**

The committee tab stamps the *viewed* committee, so a writer of any associated committee is admitted from there. But the global surveys list stamps `committees[0]` on every row (`buildEditQueryParams`), and this fallback resolves to the same primary committee — so a user who writes committee C2 (but not primary C1) on a survey associated with both is denied when editing from the global list, and gets a misleading access-denied redirect despite holding a legitimate role.

Fail-closed, so safe — but it contradicts the comment's "a writer of any associated committee must be admitted" for the list path. Options: have the committee leg iterate `survey.committees` until one grants `writer` (bounded — surveys have few committees), or document the committee-tab-only path explicitly.

apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
```diff
@@ -75,6 +90,10 @@ export class SurveyManageComponent {
 
   // Complex computed/toSignal signals
   public readonly isEditMode: Signal<boolean> = this.initIsEditMode();
+  public readonly survey: Signal<Survey | null> = this.initSurvey();
```
audigregorie: **Minor — `survey` is public but never read by the template**

The signal only feeds `surveyEntityContext`; nothing in `survey-manage.component.html` reads `survey()`. Per the component-organization convention (public for template-bound state, private for internal), this should be `private readonly survey` — unless it's deliberate groundwork for the edit pre-population ticket, in which case a one-line comment saying so would head off the next reader.

apps/lfx-one/src/app/shared/guards/writer.guard.spec.ts
audigregorie: Fixed in aa03ccb92 — added the four committee-selection tests: URL committee_uid honored when the survey's committees[] contains it, an unrelated URL param falls back to committees[0] (never the attacker's), no param resolves committees[0], and a committee-less survey falls back to the URL param. 29/29 guard tests pass.

apps/lfx-one/src/app/shared/guards/writer.guard.ts
audigregorie: Documented in aa03ccb92 — the guard and `buildEditQueryParams` comments now state the list path admits primary-committee writers only (fail-closed; non-primary writers edit via their committee tab), and the committees[] iteration is filed as follow-up GH-2190.

apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
audigregorie: Fixed in aa03ccb92 — `survey` is now private; nothing template-bound reads it today, and flipping it back is a one-word change when the edit pre-population ticket lands.

apps/lfx-one/src/app/modules/surveys/survey-manage/survey-manage.component.ts
audigregorie: Fixed in aa03ccb92 — added a guard-mirroring `initWriteAccess()` (project leg keyed off `resolveEntityWriteSlug` on the loaded survey, committee leg matching the probe's committees[]/URL-param selection rule, each leg provisionally true while pending) and passed it via `evictOnWriteAccessLoss(this.writeAccess)`. Also covers the create route's `committeeUidFromUrl` leg.

MRashad26 (approved):
Reviewed the full diff (15 files, +1019/-40) and cross-checked the key claims against the actual current repo code rather than trusting the PR description, since this touches writer authorization.

**What this does (#1569):** Survey edit links now open in the survey's own project context instead of the viewer's transient active lens/project, mirroring the meetings pattern from gh-1432 (and the very similar recent vote pattern, gh-1568). BFF now flattens `project_uid` from `committees[0]` (upstream sends no top-level project uid) and optionally enriches the detail response with `project_slug`/`project_name`/`is_foundation`; the surveys-table computes canonical per-row edit links via the existing `getEntityCommands` helper; the edit page syncs project context from the loaded survey (with a resolve-by-uid fallback); and `writerGuard` gets a new `surveys` entity probe gated by `data.entityScopedSlug: true`.

- `writerGuard`'s new `surveys` probe (`writer.guard.ts`) fetches the survey, maps `project_uid ?? ''` (consistent with the existing `resolveEntityWriteSlug`'s `||`-based absent-value handling), and computes `committee_uid` by honoring the URL `?committee_uid=` only when it's actually present in the survey's own `committees[]`, else falling back to `committees[0]` — this exactly matches the pattern already used for `votes`, and is correctly attacker-resistant: a URL param naming an unrelated committee can't win. The full guard function (not just the diff hunk) confirms `entityCommitteeUid ?? committeeUid` and the surrounding priority-order logic (project.writer → committee.writer) are unchanged and compose correctly with the new probe.
- `COMMITTEE_WRITE_FEATURES` already included `surveys` prior to this PR (`create-artifact.constants.ts`) — the docblock's "writeFeature is one of meetings/surveys/votes" claim checks out; this PR only adds the entity-scoped *probe*, not committee-writer eligibility itself.
- `resolveEntityWriteSlug` and `fetchEntityProject`/`toEntityProjectFields` (the enrichment helper) are pre-existing shared utilities from the meetings/votes/mailing-list work, reused as-is — confirmed by reading them directly, not just the import diff.
- `evictOnWriteAccessLoss(access?)`'s optional predicate-signal parameter is pre-existing (already used by vote-manage) — `SurveyManageComponent`'s new `initWriteAccess` passing its own signal is consistent with that established override pattern, including the "provisional true while pending" semantics needed to avoid evicting a guard-admitted committee writer mid-resolution.
- `initWriteAccess`'s project/committee leg resolution mirrors the guard closely: project leg resolves `resolveEntityWriteSlug` off the loaded survey, gated on project-writer only (no meetingCoordinator-equivalent for surveys, matching the guard); committee leg prefers the survey's own committee list containing the URL `committee_uid`, else the primary committee, else (create mode) the raw URL param — same precedence as the guard's probe.
- `getEntityCommands('surveys', uid, is_foundation, 'edit')` is a pre-existing shared helper (confirmed via its spec) that returns `null` when `is_foundation` is `undefined`/`null`, which the surveys-table row correctly falls back on: `getEntityCommands(...) ?? ['/surveys', survey.uid, 'edit']`, landing on the flat path that `lensRedirectGuard` prefixes — matches the new `survey-edit-url.spec.ts` e2e coverage of the unenriched case.
- Server-side `getSurveyById`'s `project_uid = survey.project_uid || survey.committees?.[0]?.project_uid` flatten, and the list-side equivalent in `getSurveys`, both feed into `enrichWithProjectData<T extends { project_uid: string }>` — the flattened value always resolves to a string (`|| ''` on the list path), satisfying that generic constraint.
- `SurveyService.getSurvey`'s new 10s `shareReplay` detail cache (client-side) correctly evicts on error (`tap({error: ...})`) and on delete (`evictSurveyDetailCache`), and the `skipCache` escape hatch is wired to the context-sync fallback's `freshFetch` — this mirrors `MailingListService.getMailingList` as claimed.
- The step-nav `queryParamsHandling: 'merge'` addition (`nextStep`/`previousStep`/`goToStep`) is necessary and correctly scoped — without it, `?project=`/`committee_uid` would be dropped by `router.navigate([], { queryParams: { step } })` on every step change, breaking the just-synced context.

**Minor, non-blocking observations:**
- `surveys-table.component.ts`'s new `buildEditQueryParams`/`editCommands` computation and `committee-surveys.component.ts`'s removal of `editSurveyQueryParams` have no dedicated component-level spec files (neither component has one at all, pre-existing gap, not introduced here) — coverage instead comes entirely from the two new e2e specs, which is a reasonable trade for this kind of routing/query-param logic but leaves the `buildEditQueryParams` committee-scope fallback logic (`this.committeeUid() || survey.committees?.[0]?.committee_uid`) without a fast unit-level pin.
- The known gap called out in the PR/comments (global surveys list only stamps `committees[0]` on edit links, fail-closed denying a non-primary-committee writer from that entry point) is explicitly documented and tracked as follow-up (GH-2190) — reasonable to defer given it's fail-closed (denies rather than over-grants).

No critical or major issues found. The authorization logic is consistent with the established meetings/votes entity-scoped pattern and fails closed on the same edge cases (misconfigured route, non-404 probe error, unrelated committee_uid). Approving.

PR #2182, opened 2026-09-03
fix(profile): consolidate email/password pages into settings

### LINKED ISSUE #2177: fix: consolidate /profile/email and /profile/password into Settings page
## Problem

`/profile/emails` (plural, used as a Flow C \`returnTo\` target) 404s — the actual Angular route is singular \`/profile/email\`. While tracking down that mismatch, found the standalone \`/profile/email\` and \`/profile/password\` pages are orphaned: nothing in the app links to them, and their functionality (add/verify/set-primary/delete email, change password) is duplicated by the actively-maintained \"Email Settings\" and \"Password\" sections already embedded in \`/profile/settings\` (\`AccountSettingsComponent\`).

## Fix

Consolidate onto the Settings page instead of patching the dead pages:

- Redirect \`/profile/email\`, \`/profile/emails\`, and \`/profile/password\` to \`/profile/settings\` with a matching fragment (\`#email-settings\`, \`#password\` — reusing the existing DOM section ids).
- Add fragment-read-and-scroll support to \`AccountSettingsComponent\` (it currently has none — \`activeSection\` always defaults to \`'email-settings'\`).
- Delete the now-fully-orphaned \`ProfileEmailComponent\` and \`ProfilePasswordComponent\` and their routes.
- Repoint the Flow C \`returnTo\` producers that target the old paths at \`/profile/settings\` (with fragment where applicable), and update \`allowedProfileReturnPaths\` accordingly.

fayazg:
## Summary
- Redirect the orphaned standalone /profile/email, /profile/emails, and /profile/password pages into /profile/settings with a matching fragment (#email-settings, #password), instead of maintaining two duplicate email/password management UIs
- Delete ProfileEmailComponent and ProfilePasswordComponent, now fully unreferenced
- Fix Flow C's returnTo fragment loss (dropped by relying on the referer header) and a re-scroll race against the deep-link DOM swap
- Add route-redirect and fragment-scroll test coverage

Fixes #2177


fayazg:
Commits on this branch:

- `0c0604846` — redirect `/profile/email(s)`/`/profile/password` into `/profile/settings#...`; delete `ProfileEmailComponent`/`ProfilePasswordComponent`; add `/profile/email` to the Flow C return-path allowlist
- `2fa8fc586` — validate the incoming fragment before setting `activeSection`; revert the 4 Flow C `authorize_url` fallbacks to the fragment-bearing legacy paths
- `c153c0a4d` — preserve query params on the settings redirect
- `0bf5b4cb9`, `498bd9382`, `032825cad` — iterate the deep-link re-scroll pipeline from `effect()` to a single RxJS subscription
- `b35e27478` — drop the referer-header dependency for Flow C's `authorize_url` (it always resolved to the fragment-less `/profile/settings`, defeating the fragment restore); defer the re-scroll with `afterNextRender` to avoid racing the loaded-state DOM swap; add `profile.routes.spec.ts` and `account-settings.component.spec.ts`

apps/lfx-one/src/app/modules/profile/profile.routes.ts
```diff
@@ -57,15 +58,11 @@ export const PROFILE_ROUTES: Routes = [
...
-      },
-      {
-        path: 'email',
-        loadComponent: () => import('./email/profile-email.component').then((m) => m.ProfileEmailComponent),
-      },
+      // Redirect the old standalone email/password pages into the Settings tab section.
+      // Functional redirectTo (not a string) preserves Flow C's ?error=/?success= query params.
+      { path: 'password', redirectTo: ({ queryParams }) => inject(Router).createUrlTree(['/profile/settings'], { queryParams, fragment: 'password' }) },
+      { path: 'email', redirectTo: ({ queryParams }) => inject(Router).createUrlTree(['/profile/settings'], { queryParams, fragment: 'email-settings' }) },
+      { path: 'emails', redirectTo: ({ queryParams }) => inject(Router).createUrlTree(['/profile/settings'], { queryParams, fragment: 'email-settings' }) },
```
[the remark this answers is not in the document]
fayazg: Good catch — clearAuthQueryParams() rebuilds the URL from router.url.split('?')[0], which drops everything after `?`, including the fragment (router.url serializes as path?query#fragment). Fixed in 5b06cb7bc: it now splits on `#` first and re-appends the fragment after stripping the query string. Added coverage in profile-layout.component.spec.ts for both the success and error Flow C return paths.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const userServiceMock = {
+      impersonating: signal(true), // skips the developer-token fetch — irrelevant to this suite
+      getUserEmails: vi.fn(() => of(null)),
+    };
+
+    TestBed.configureTestingModule({
+      imports: [AccountSettingsComponent],
+      providers: [
+        { provide: PLATFORM_ID, useValue: 'server' }, // no DOM in this suite — scrollToSection's document access is guarded
```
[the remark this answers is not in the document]
fayazg: Fixed in 5b06cb7bc: added a browser-platform test (real DOM section elements, a controllable email-load Subject) that asserts scrollIntoView only fires after the email data finishes loading, and stays uncalled while it's still pending.

fayazg:
## Review Feedback Addressed

Commit: 5b06cb7bc

### Changes Made
- **apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.ts**: `clearAuthQueryParams()` now preserves the URL fragment when stripping Flow C query params instead of dropping it too (per copilot[bot])
- **apps/lfx-one/src/app/layouts/profile-layout/profile-layout.component.spec.ts**: added fragment-preservation coverage for both Flow C success and error returns
- **apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.spec.ts**: added a browser-platform test exercising the deferred `scrollIntoView` behavior that the existing server-platform suite couldn't cover (per copilot[bot])

### Threads Resolved

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -171,9 +173,26 @@ export class AccountSettingsComponent {
...
+          emailLoaded$.pipe(
+            take(1),
+            map(() => validFragment)
+          )
+        ),
+        takeUntilDestroyed()
+      )
+      // afterNextRender (not a direct call): waits for the loaded-state DOM swap to actually
+      // paint before measuring scrollIntoView, so the layout shift can't race the scroll.
+      .subscribe((validFragment) => afterNextRender(() => this.scrollToSection(validFragment), { injector: this.injector }));
```
dealako: [question] Deferred scroll may override an in-flight TOC click

Issue: The `emailLoaded$` subscription queues `afterNextRender(() => scrollToSection(validFragment))` with no guard for user interaction during the email-loading window.
Proof: TOC buttons in `account-settings.component.html` stay clickable while `emailLoading()` is true. If a user lands on `#password` then clicks another section before `/api/profile/emails` returns, the deferred callback still smooth-scrolls back to `#password` and resets `activeSection`.
Why it matters: Explicit navigation can be silently reverted once the spinner clears — reads like a broken TOC.
Fix: Track a `userPickedSection` flag from the TOC click handler and skip the deferred scroll when set. Gating only on `activeSection() === validFragment` is insufficient because the scroll-spy legitimately moves `activeSection` during the layout swap this callback exists to fix.

dealako:
Hey @fayazg — nice consolidation work here. Redirecting the legacy `/profile/email(s)/password` routes into account settings with fragment deep-links is the right direction, and the Flow C fixes (static `returnTo` paths, `clearAuthQueryParams` fragment preservation, deferred re-scroll after email load) trace cleanly through the BFF → route redirect → settings scroll pipeline. Test coverage for the redirect and fragment paths is solid.

**Structured issue count:**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: shared returnTo/fragment string contract should live in `@lfx-one/shared`; `/profile/emails` route comment mislabels a new required landing path as legacy
- ⚪ Nit: 3 issues: spec mocks shadowed by component providers; `redirectTo` rationale in routes spec cites query params instead of fragment; migration doc still says Supabase for account-settings email
- ❔ Question: 1 item: should deferred deep-link scroll be cancelled if the user picks another TOC section while email data is still loading?

**Bot reconciliation:**
- ✅ **Agree** with Copilot on `clearAuthQueryParams` dropping the fragment — fixed in `5b06cb7bc`.
- ✅ **Agree** with Copilot that the server-platform spec alone couldn't exercise the re-scroll race — browser-platform test added in `5b06cb7bc`.
- CodeRabbit review was skipped (repo config).

**Revision tracking:** Both Copilot inline comments from the first round are addressed in `5b06cb7bc`. No regressions spotted in the follow-up commits.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+        { provide: ActivatedRoute, useValue: { snapshot: { data: {} }, fragment: fragment$ } },
+        { provide: UserService, useValue: userServiceMock },
+        { provide: ConfirmationService, useValue: {} },
+        { provide: MessageService, useValue: { add: vi.fn() } },
+        { provide: DialogService, useValue: { open: vi.fn() } },
+      ],
+    });
+    // Empty template: exercise the constructor's fragment pipeline without the full section markup.
+    TestBed.overrideComponent(AccountSettingsComponent, { set: { template: '', imports: [] } });
```
dealako: [nit] TestBed PrimeNG mocks are shadowed by component providers

Issue: `MessageService` / `DialogService` / `ConfirmationService` mocks in `TestBed.configureTestingModule` never reach the component.
Proof: `AccountSettingsComponent` declares `providers: [ConfirmationService, MessageService, DialogService]` in its `@Component` metadata. `overrideComponent` only replaces `template` and `imports`, so `inject(MessageService)` resolves the real service, not `{ add: vi.fn() }`.
Why it matters: A future toast assertion on this spec will get false negatives.
Fix: Add `providers: []` to the `overrideComponent` `set` block, or drop the unused TestBed mocks.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -63,6 +64,7 @@ export class AccountSettingsComponent {
   public resendCooldown = this.resendCooldownUtil.cooldown;
 
   // ── TOC active section ──
+  private static readonly sectionIds = ['email-settings', 'password', 'developer-settings'] as const;
```
dealako: [minor] Centralize the Flow C returnTo / fragment vocabulary in shared constants

Issue: The same path/fragment contract is declared independently in `profile.controller.ts` (`allowedProfileReturnPaths` + hardcoded `/profile/emails` / `/profile/password`), `profile.routes.ts` redirect targets, and this component's `sectionIds` (mirrored by template `id=` attributes).
Proof: `packages/shared/src/constants/cla.constants.ts` and `org-memberships.constants.ts` already centralize the same kind of cross-layer deep-link contract; `profile.constants.ts` holds other profile route segments.
Why it matters: Renaming a section id or return path on one side produces no compile error and the deep link silently degrades to a dead fragment.
Fix: Add shared symbols (e.g. `PROFILE_EMAILS_PATH`, `PROFILE_PASSWORD_PATH`, `ACCOUNT_SETTINGS_SECTIONS`) to `packages/shared/src/constants/profile.constants.ts` and consume them from BFF, routes, and component.

apps/lfx-one/src/app/modules/profile/profile.routes.ts
```diff
@@ -57,15 +58,11 @@ export const PROFILE_ROUTES: Routes = [
...
-      // Direct-URL-only routes (no tab, but still accessible)
-      {
-        path: 'password',
-        loadComponent: () => import('./password/profile-password.component').then((m) => m.ProfilePasswordComponent),
-      },
-      {
-        path: 'email',
-        loadComponent: () => import('./email/profile-email.component').then((m) => m.ProfileEmailComponent),
-      },
+      // Redirect the old standalone email/password pages into the Settings tab section.
```
dealako, line 62: [minor] `/profile/emails` is a new required landing route, not a legacy redirect

Issue: The block comment says these routes redirect "old standalone email/password pages," but `emails` did not exist on `main` — only `password` and `email` did.
Proof: `profile.controller.ts` already emits `returnTo=/profile/emails` for Flow C email authorization; this route exists so that static path can land and re-attach `#email-settings`.
Why it matters: A future cleanup pass pruning "legacy redirects" could delete `emails` and break every Flow C email return.
Fix: Split the comment: keep the legacy rationale on `password`/`email`, and document `emails` as the required BFF landing path while that `returnTo` literal exists.

apps/lfx-one/src/app/modules/profile/profile.routes.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { TestBed } from '@angular/core/testing';
+import { provideRouter, Route } from '@angular/router';
+import { describe, expect, it } from 'vitest';
+
+import { PROFILE_ROUTES } from './profile.routes';
+
+/**
+ * Guards issue #2177's consolidation: the legacy /profile/email(s)/password pages redirect into
+ * /profile/settings with a fragment. A functional redirectTo (not a string) is required to carry
```
dealako: [nit] Functional `redirectTo` is required for the fragment, not query-param survival

Issue: The spec docblock (and matching route comment) claim a string `redirectTo` drops Flow C `?error=` / `?success=` query params.
Proof: Angular's relative string `redirectTo` preserves the incoming query string; what it cannot do is attach a new fragment. The functional `createUrlTree` form is what adds `#password` / `#email-settings`.
Why it matters: The recorded design rationale is misleading — future maintainers may think query-param preservation is the unique property under test.
Fix: Restate as "functional `redirectTo` is required to attach the section fragment"; rename tests to lead with the fragment assertion.

docs/migration/linux-com-email-forwarder.md
```diff
@@ -394,9 +394,10 @@ alias without leaving the same application.
   it under a `membership/` module alongside the Individual Supporter feature).
 - **Auth middleware** — the ITX service uses an Auth0 M2M token; the same pattern used
   by other server-side service calls in LFX One applies.
-- **Profile email page as a model** — `apps/lfx-one/src/app/modules/profile/email/`
-  already handles profile email management (though backed by Supabase, not ITX). The
-  form pattern and layout can be adapted.
+- **Profile settings page as a model** — the email section of
+  `apps/lfx-one/src/app/modules/settings/account-settings/` already handles profile
+  email management (though backed by Supabase, not ITX). The form pattern and layout
```
dealako: [nit] Stale Supabase attribution on the account-settings reference

Issue: The re-pointed bullet still says account-settings email management is "backed by Supabase, not ITX."
Proof: `AccountSettingsComponent` loads email data via `UserService.getUserEmails()` → `GET /api/profile/emails` → auth-service over NATS (`profile.controller.ts`). The repo's `supabase.service.ts` is an explicit stub for badge fallback flows.
Why it matters: This doc is the implementation shortcut for the Linux.com forwarder — the wrong backing service sends readers on a dead-end investigation.
Fix: Replace with "backed by auth-service over NATS, not ITX."

dealako (dismissed):
## Review summary

Solid consolidation of legacy profile email/password pages into account settings with fragment deep-links. Flow C return handling is improved (static allowlisted `returnTo`, fragment-preserving `clearAuthQueryParams`, deferred re-scroll gated on email load). Prior Copilot findings on fragment loss and missing browser-platform scroll coverage are resolved in `5b06cb7bc`.

- 🔴 Blocking: 0
- 🟡 Minor: 2 (shared constants contract; `/profile/emails` route documentation)
- ⚪ Nit: 3 (spec provider shadowing; redirectTo comment accuracy; migration doc Supabase attribution)
- ❔ Question: 1 (TOC interaction during deferred scroll window)

MRashad26 (dismissed):
Reviewed the full diff (12 files, +289/-947) and cross-checked the key claims against the actual current code rather than trusting the PR description. Lint clean on this branch (`yarn lint:check`, same pre-existing unrelated warning seen on the other PRs in this batch).

**What this does (#2177):** Consolidates the standalone `/profile/email`, `/profile/emails`, `/profile/password` pages into `/profile/settings` with a fragment, deletes the now-fully-replaced `ProfileEmailComponent`/`ProfilePasswordComponent`, and fixes two related bugs: Flow C's `returnTo` losing its fragment, and a re-scroll race against the settings page's loading→loaded DOM swap.

- `ProfileEmailComponent`/`ProfilePasswordComponent` (both `.ts` and `.html`) are genuinely dead after deletion — grepped the full app tree for `ProfileEmailComponent`, `ProfilePasswordComponent`, `profile-email.component`, `profile-password.component` outside spec files and found zero references. Deletion is clean, not leaving a dangling import anywhere.
- `profile.routes.ts`'s three new routes use a **functional** `redirectTo` (`({ queryParams }) => inject(Router).createUrlTree(...)`), not a string — confirmed this is required, not stylistic: a string `redirectTo` can't carry `queryParams`/`fragment`, so Flow C's `?error=`/`?success=` would be dropped on the legacy-URL redirect otherwise. New `profile.routes.spec.ts` directly pins this (asserts `queryParams` preserved and correct `fragment` per route).
- The `clearAuthQueryParams` fix in `profile-layout.component.ts` correctly targets the actual bug: the old `this.router.url.split('?')[0]` truncated before the query string but `router.url` serializes as `path?query#fragment`, so a URL like `/profile/settings?success=...#password` kept the fragment attached to the query portion and lost it entirely on strip. The new `split('#')` then `split('?')[0]` on the first segment correctly separates and re-attaches the fragment. Three new spec cases (success-return, error-return, no-fragment) cover this directly.
- The three `profile.controller.ts` `authorize_url` sites changed `(req.headers['referer'] as string) || '/profile/emails'` → a static `/profile/emails`/`/profile/password` literal. Verified this isn't a behavior regression: `referer` on a Flow C return would resolve to the settings page itself post-consolidation, which carries no fragment — using the static legacy path is deliberate so `profile.routes.ts`'s functional `redirectTo` re-attaches the correct fragment on the way back. The inline comments explain this, and it checks out against the routes change above.
- `account-settings.component.ts`'s new deferred re-scroll pipeline (`toObservable(this.emailLoading)` gated `switchMap` into `emailLoaded$`, then `afterNextRender` with an explicit `Injector`) correctly defers `scrollToSection` until email data has actually finished loading, avoiding a scroll landing on the wrong offset before the loading→loaded DOM swap settles. The new browser-platform spec (`emails$` as a controlled `Subject`, asserting `scrollIntoView` is NOT called while still loading and IS called exactly once after `emails$` completes) is a solid direct test of the race fix, and correctly runs on `PLATFORM_ID: 'browser'` (separate from the SSR-safe `'server'`-platform suite covering just the fragment→activeSection mapping).
- `AccountSettingsComponent.sectionIds` was hoisted to a `private static readonly` and reused in both the new fragment-filter and the existing `setupScrollSpy` — a reasonable de-duplication, not a functional change (confirmed both use sites now reference the same array).
- Docs update in `linux-com-email-forwarder.md` correctly re-points references from the deleted `modules/profile/email/` to `modules/settings/account-settings/`.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
fayazg: Done — added a `userPickedSection` flag set by a new `selectSection()` method (now called from the TOC template instead of `scrollToSection()` directly). The deferred re-scroll checks it before firing, so an explicit click made while email data is still loading is no longer overridden. Added a regression test for this exact race.

See commit c52ec6f83: account-settings.component.ts / .html / .spec.ts.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.spec.ts
fayazg: Good catch — fixed. Added `providers: []` to both `TestBed.overrideComponent` calls so the component's own `@Component({ providers: [...] })` array (ConfirmationService, MessageService, DialogService) stops shadowing the TestBed-level mocks.

See commit c52ec6f83: account-settings.component.spec.ts.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
fayazg: Done — centralized the path/fragment vocabulary in `packages/shared/src/constants/profile.constants.ts` as `PROFILE_SETTINGS_PATH`, `PROFILE_EMAILS_PATH`, `PROFILE_PASSWORD_PATH`, and `ACCOUNT_SETTINGS_SECTIONS`. Adopted across `profile.controller.ts`'s allowlist/returnTo literals, `profile.routes.ts`'s redirects, and this component's section ids.

See commit c52ec6f83.

apps/lfx-one/src/app/modules/profile/profile.routes.ts
fayazg: Fixed the comment — split it so it's clear `emails` is the live Flow C landing path (not a legacy redirect like `password`/`email`), while the constants centralization from the other thread still applies to all three.

See commit c52ec6f83: profile.routes.ts.

apps/lfx-one/src/app/modules/profile/profile.routes.spec.ts
fayazg: Fixed — corrected the docblock and reordered the assertions to lead with the fragment. You're right that a string `redirectTo` doesn't drop query params; only the fragment requires the functional form.

See commit c52ec6f83: profile.routes.spec.ts.

docs/migration/linux-com-email-forwarder.md
fayazg: Fixed — corrected the attribution to auth-service over NATS.

See commit c52ec6f83: linux-com-email-forwarder.md.

fayazg:
## Review Feedback Addressed

Commit: c52ec6f83

### Changes Made
- **account-settings.component.ts/.html/.spec.ts**: added a `userPickedSection` guard and `selectSection()` so the deferred re-scroll can no longer override an explicit TOC click made while email data is still loading; added a regression test (per @dealako)
- **account-settings.component.spec.ts**: fixed a `TestBed.overrideComponent` bug where the component's own `providers` array was shadowing the DI mocks (per @dealako)
- **profile.constants.ts / profile.controller.ts / profile.routes.ts**: centralized the Flow C path/fragment vocabulary into shared constants (per @dealako)
- **profile.routes.ts**: corrected the comment on the `emails` route — it's the live Flow C landing path, not a legacy redirect (per @dealako)
- **profile.routes.spec.ts**: corrected the docblock/assertions — a string `redirectTo` doesn't drop query params, only the fragment requires the functional form (per @dealako)
- **linux-com-email-forwarder.md**: corrected backend attribution to auth-service over NATS (per @dealako)

### Threads Resolved

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -293,7 +336,7 @@ export class AccountSettingsComponent {
       .pipe(take(1))
       .subscribe((status) => {
         if (!status.authorized) {
-          this.redirectToProfileAuth('/api/profile/auth/start?returnTo=/profile/settings');
+          this.redirectToProfileAuth(`/api/profile/auth/start?returnTo=${PROFILE_SETTINGS_PATH}`);
```
luismoriguerra: Email deletion starts Flow C with `returnTo=${PROFILE_SETTINGS_PATH}`, which carries no fragment and lands at the top of the settings page on return. Use `returnTo=${PROFILE_EMAILS_PATH}` so `profile.routes.ts`'s redirect automatically re-attaches `#email-settings`.

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -91,11 +94,12 @@ const PASSWORD_ERROR_RULES: readonly {
 export class ProfileController {
   private static readonly allowedProfileReturnPaths: ReadonlySet<string> = new Set([
     '/profile',
-    '/profile/emails',
+    '/profile/email',
```
luismoriguerra: `'/profile/email'` remains a raw string literal in `allowedProfileReturnPaths` next to the centralized `PROFILE_EMAILS_PATH`. Consider consolidating it into `@lfx-one/shared/constants` or adding an inline comment explaining its retention for legacy bookmarks.

luismoriguerra (dismissed):
## PR Review: PR #2182 (HEAD ea80769b5)

**Verdict:** ✅ Approved
**Merge readiness:** safe-to-merge (non-blocking email deletion fragment polish noted below)

**Quality Gates & CI:**
- `yarn lint:check`: 🟢 PASS (0 errors)
- `ng test` (affected client specs): 🟢 PASS (17 tests)
- `yarn test:server`: 🟢 PASS on `ea80769b5` (101 suites passed, 2,706 tests passed; mock gap fixed in `ea80769b5`)
- `format:check` & `yarn build`: 🟢 PASS (verified via CI run 33889266042 Quality Checks)

**Findings:**
- **Non-blocking** · `apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts:339` · Flow C email deletion uses `returnTo=${PROFILE_SETTINGS_PATH}` without `#email-settings`; change to `PROFILE_EMAILS_PATH` to leverage the redirect fragment re-attach.
- **Nit** · `apps/lfx-one/src/server/controllers/profile.controller.ts:97` · Raw literal `'/profile/email'` retained in `allowedProfileReturnPaths` next to centralized constants.

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
fayazg: Done — the `deleteEmail` unauthenticated redirect now uses `PROFILE_EMAILS_PATH` instead of `PROFILE_SETTINGS_PATH`, so the Flow C return trip goes through the `emails` route's redirect and picks up `#email-settings` instead of landing on `/profile/settings` with no fragment.

See commit 171c359df.

apps/lfx-one/src/server/controllers/profile.controller.ts
fayazg: Done — added a `PROFILE_EMAIL_PATH` shared constant and swapped it in for the raw `'/profile/email'` literal in `allowedProfileReturnPaths`, consistent with the other `PROFILE_*_PATH` constants already used there.

See commit 171c359df.

fayazg:
## Review Feedback Addressed

Commit: 171c359df

### Changes Made
- **account-settings.component.ts**: `deleteEmail`'s unauthenticated redirect now targets `PROFILE_EMAILS_PATH` instead of `PROFILE_SETTINGS_PATH`, so the Flow C return trip retains the `#email-settings` fragment instead of landing at the top of the settings page (per @luismoriguerra)
- **profile.controller.ts**: added a `PROFILE_EMAIL_PATH` shared constant and replaced the raw `'/profile/email'` literal in `allowedProfileReturnPaths` (per @luismoriguerra)

### Threads Resolved

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -171,15 +181,48 @@ export class AccountSettingsComponent {
...
+      // afterNextRender (not a direct call): waits for the loaded-state DOM swap to actually
+      // paint before measuring scrollIntoView, so the layout shift can't race the scroll.
+      .subscribe((validFragment) =>
+        afterNextRender(
+          () => {
+            // The TOC stays clickable during the email-loading window; if the user already picked
+            // a different section by the time this fires, respect that instead of smooth-scrolling
+            // them back to the original deep-link fragment.
+            if (this.userPickedSection) return;
+            this.scrollToSection(validFragment);
```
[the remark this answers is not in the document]
fayazg: Done, the deferred deep-link re-scroll is now cancelled by any user-initiated wheel, touch, or keyboard scroll during the email-loading window, not just TOC clicks via `selectSection()`.

See commit 45e0ad6bb: `account-settings.component.ts` now attaches browser-only `wheel`/`touchstart`/`keydown` listeners the moment a valid fragment arrives, which set `userPickedSection` and self-remove on first user gesture (or once the deferred re-scroll actually fires).

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -586,7 +591,9 @@ export class ProfileController {
         res.status(403).json({
           error: 'management_token_required',
           message: 'Profile authorization required to change the primary email',
-          authorize_url: `/api/profile/auth/start?returnTo=${encodeURIComponent((req.headers['referer'] as string) || '/profile/emails')}`,
+          // Static legacy path (not `referer`, which is the settings page itself and carries no
+          // fragment) — profile.routes.ts's redirectTo re-attaches #email-settings on return.
+          authorize_url: `/api/profile/auth/start?returnTo=${encodeURIComponent(PROFILE_EMAILS_PATH)}`,
```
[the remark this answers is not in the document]
fayazg: Done, added controller tests covering this gap.

See commit 45e0ad6bb: `profile.controller.spec.ts` now asserts `setPrimaryEmail` and `sendPasswordResetEmail` return the correct `authorize_url` (`%2Fprofile%2Femails` and `%2Fprofile%2Fpassword` respectively) on their `management_token_required` 403 response.

fayazg:
## Review Feedback Addressed

Commit: 45e0ad6bb

### Changes Made
- **account-settings.component.ts**: the deferred deep-link re-scroll is now cancelled by any user-initiated wheel, touch, or keyboard scroll during the email-loading window, not just TOC clicks (per copilot[bot])
- **profile.controller.spec.ts**: added tests asserting `setPrimaryEmail` and `sendPasswordResetEmail` return the correct `authorize_url` on their `management_token_required` 403 response (per copilot[bot])

### Threads Resolved

luismoriguerra (dismissed):

apps/lfx-one/src/app/modules/settings/account-settings/account-settings.component.ts
```diff
@@ -491,6 +538,30 @@ export class AccountSettingsComponent {
...
+    if (!isPlatformBrowser(this.platformId) || this.outsideScrollCancelHandler) return;
+    const cancel = (): void => {
+      this.userPickedSection = true;
+      this.disarmOutsideScrollCancellation();
+    };
+    this.outsideScrollCancelHandler = cancel;
+    document.addEventListener('wheel', cancel, { passive: true });
+    document.addEventListener('touchstart', cancel, { passive: true });
+    document.addEventListener('keydown', cancel);
+    this.destroyRef.onDestroy(() => this.disarmOutsideScrollCancellation());
```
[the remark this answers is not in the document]
fayazg: Fixed, narrowed `armOutsideScrollCancellation()` to actual scroll gestures rather than any keypress/tap: keydown now filters to scroll-inducing keys (arrows, PageUp/Down, Home, End, Space), and touchstart was replaced with touchmove so a plain tap-to-focus no longer cancels the deferred re-scroll.

See commit 45ba10b77: account-settings.component.ts.

fayazg:
## Review Feedback Addressed

Commit: 45ba10b77

### Changes Made
- **account-settings.component.ts**: narrowed the deferred deep-link re-scroll's cancellation trigger to actual scroll gestures — keydown now filters to scroll-inducing keys (arrows, PageUp/Down, Home, End, Space), and touchstart was replaced with touchmove so a plain tap-to-focus or typing in a still-visible form field no longer cancels the re-scroll (per cursor[bot])

### Threads Resolved

luismoriguerra (approved):

PR #2183, opened 2026-09-03
fix(profile): let the profile shell own invalid_state/no_code toasts

fayazg:
## Summary

- `invalid_state` and `no_code` were only handled by the Identities tab's error map, so a Flow C (profile re-auth) failure on any other `/profile/*` route produced no toast, left `?error=` stuck in the URL, and left a pending-save stash behind
- Move both codes into `PROFILE_AUTH_ERROR_MESSAGES`, owned by the profile shell that's alive on every `/profile/*` route — the Identities component's existing ownership guard now skips them automatically, so there's no double-toast
- Remove the unreachable `/settings` entry from the profile-auth `returnTo` allowlist; nothing ever sends `returnTo=/settings`, and a value outside the allowlist already falls back safely to `/profile`

Fixes #2180

PR #2184, opened 2026-09-03
fix(profile): block auth callbacks during impersonation

### LINKED ISSUE #1936: fix(profile): flow c callback registered twice with inconsistent impersonation guard
Found while investigating #1935 (Linux.com "Forward to" select). Not part of that fix — filing separately.

`handleProfileAuthCallback` is reachable from two registered routes with different middleware:

| Path | Registration | Impersonation guard |
| --- | --- | --- |
| `/passwordless/callback` (app root) | `src/server/server.ts:371`, own `new ProfileController()` at `:370` | **none** |
| `/api/profile/auth/callback` | `src/server/routes/profile.route.ts:48` | `blockDuringImpersonation` |

`PROFILE_REDIRECT_URI` defaults to `${PCC_BASE_URL}/passwordless/callback` (`src/server/services/profile-auth.service.ts:60`), so the **unguarded root path is the one Auth0 actually calls**. The `/api` variant with the guard appears to be dead or near-dead code.

The root route is allowlisted in auth middleware (`src/server/middleware/auth.middleware.ts:64`, `{ pattern: '/passwordless/callback', type: 'ssr', auth: 'required', tokenRequired: false }`) and, lacking `blockDuringImpersonation`, would let a Flow C callback complete while an admin is impersonating. `handleProfileAuthCallback` does validate the token's `sub` against the OIDC session sub (`profile-auth.service.ts:154-162`), which likely blunts the practical impact — severity unconfirmed.

**Tasks:**
- [ ] Determine which registration is canonical; remove the other
- [ ] If the root path stays, add `blockDuringImpersonation` to match
- [ ] Confirm whether the `sub` check already makes impersonation abuse impossible
- [ ] Check nothing points at the `/api` path before removing it

Refs #1935

fayazg:
## Summary

- Auth0's actual Flow C re-auth callback (`/passwordless/callback`) and the social-link callback
  (`/social/callback`) are root-level redirect handlers that sit outside the `/api` error-handler
  mount, so they could not reuse `blockDuringImpersonation`'s `next(err)` contract the way their
  little-used `/api/profile/auth/callback` twin does — leaving both unguarded while impersonating.
- Add an in-handler equivalent, `ProfileController.blockCallbackDuringImpersonation`, that redirects
  with `?error=impersonation_read_only` before any code exchange or token mint, matching the
  handlers' existing redirect-on-error convention.
- Add `impersonation_read_only` to `PROFILE_AUTH_ERROR_MESSAGES` so `ProfileLayoutComponent` toasts
  it on every `/profile/*` route, the same as any other Flow C error code.
- Fix a vacuous assertion in the existing callback spec (it asserted a method the unblocked path
  never calls either) and update `impersonation.md` to document the redirect-based variant of the
  guard alongside the existing 403 contract.

Both root registrations and their `/api/profile/auth/callback` twin are kept as-is — only the guard
is added, to avoid depending on undocumented Auth0 tenant configuration.

**Impact note:** the Flow C `sub` check (`profile-auth.service.ts`) already compares against the
real, non-impersonated `req.oidc.user.sub`, so no cross-account management token could land even
before this fix. This change is hardening and consistency with the rest of the impersonation
read-only guarantee, not a fix for a confirmed cross-account vulnerability.

Two files touched are comment-only: `server.ts` (registration comments) and
`impersonation-readonly.middleware.ts` (rationale comment cross-referencing the new in-handler
guard) — no behavioral change to either.

Fixes #1936


fayazg:
**Commit \`9e09c9b33\`** — fix(profile): block auth callbacks during impersonation (#1936)

- Added \`ProfileController.blockCallbackDuringImpersonation\`: an in-handler impersonation guard for the two root-level redirect callbacks (\`/passwordless/callback\`, \`/social/callback\`) that sit outside the \`/api\` error-handler mount and so can't use \`blockDuringImpersonation\`'s \`next(err)\` contract.
- Wired the guard into \`handleProfileAuthCallback\` and \`handleSocialCallback\` as the first branch, before any code exchange or token mint.
- Added \`impersonation_read_only\` to \`PROFILE_AUTH_ERROR_MESSAGES\` so the existing profile-layout toast picks it up on every \`/profile/*\` route.
- Comment-only updates to \`server.ts\`, \`routes/profile.route.ts\`, and \`impersonation-readonly.middleware.ts\` documenting why the root callbacks use the in-handler guard instead of the middleware.
- New/updated tests: \`profile.controller.spec.ts\` (blocked-callback redirects, no code exchange) and \`profile-layout.component.spec.ts\` (toast + sessionStorage stash cleared on the new error code).

fayazg:
**Commit \`4d09700cb\`** — docs(profile): fix vacuous auth-callback test and doc drift (#1936)

- \`profile.controller.spec.ts\`: the "without exchanging the code" test asserted \`getManagementToken\` was never called, but the unblocked path never calls that method either — it calls \`exchangeCodeForToken\`. Reassigned the assertion to the method the guard actually prevents, and added \`exchangeCodeForToken\` to the mocked \`profileAuthSvc\`.
- \`docs/architecture/backend/impersonation.md\`: documented the redirect-based (\`?error=impersonation_read_only\`) variant of the guard on the root callbacks, alongside the existing 403-based \`blockDuringImpersonation\` contract.

apps/lfx-one/src/server/controllers/profile.controller.spec.ts
```diff
@@ -356,3 +357,59 @@ describe('ProfileController.getLinuxAlias', () => {
...
+  });
+
+  it('handleSocialCallback redirects to /profile/identities without linking the identity', async () => {
+    isImpersonatingMock.mockReturnValue(true);
+    const res = buildRes();
+
+    await controller.handleSocialCallback(buildReq({ path: '/social/callback', query: { code: 'c', state: 's' } }), res);
+
+    expect(res.redirect).toHaveBeenCalledWith('/profile/identities?error=impersonation_read_only');
+    expect(profileAuthSvc.getManagementToken).not.toHaveBeenCalled();
```
dealako: [minor] handleSocialCallback test re-introduces vacuous assertion

Issue: The test title claims "without linking the identity", but the assertion is `expect(profileAuthSvc.getManagementToken).not.toHaveBeenCalled()` — the same vacuous pattern commit `4d09700cb` removed from the Flow C sibling.
Proof: In `handleSocialCallback`, `getManagementToken` is not reached until after `validateState` / `clearState` / `!code` checks (`profile.controller.ts` ~1787–1805). The `SocialVerificationService` mock returns `{}` (`profile.controller.spec.ts:111–115`), so without the guard `validateState` throws before `getManagementToken` is ever considered. Neither `exchangeCodeForToken` nor `linkIdentity` is asserted.
Why it matters: This is the regression guard for a read-only-during-impersonation control. As written it only proves the redirect string, not that code exchange / identity linking never ran.
Fix: Hoist a real `socialVerificationSvc` mock (with `validateState`, `exchangeCodeForToken`, etc.) and add `linkIdentity` to `emailVerificationSvc`; assert `exchangeCodeForToken` and `linkIdentity` were not called — mirror the Flow C test at line 379.

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -2478,6 +2486,29 @@ export class ProfileController {
     }
   }
 
+  /**
+   * Root callbacks (`/passwordless/callback`, `/social/callback`) sit outside the `/api`
+   * error-handler mount and are redirect-only — their route signatures don't even accept
```
dealako: [nit] Drop incorrect "route signatures don't accept next" justification

Issue: The docstring claims root callback handlers "don't even accept `next`", but that is not why middleware cannot be used here.
Proof: `blockDuringImpersonation` is route-level middleware with its own `(req, _res, next)` signature; Express supplies `next` to middleware regardless of final handler arity. The real reason (already stated elsewhere) is that `next(err)` on a top-level navigation surfaces 403 JSON instead of a redirect.
Why it matters: Misleading rationale in the canonical comment duplicated across five files may send a future maintainer down the wrong fix path.
Fix: Remove the "route signatures don't even accept `next`" clause; keep the redirect-vs-403 explanation only.

apps/lfx-one/src/server/routes/profile.route.ts
```diff
@@ -38,6 +38,11 @@ function handlePictureUploadParseError(err: unknown, req: Request, _res: Respons
  * Write and Flow C (Auth0 Management API) routes are guarded by `blockDuringImpersonation`:
  * profile mutations act on the real user's account (via the impersonator's management token), so
  * they must be rejected while impersonating. Profile viewing during impersonation is read-only.
+ *
+ * Flow C and social-link have root-level twins (`/passwordless/callback`, `/social/callback` in
```
dealako: [nit] "Root-level twins" overstates social-link symmetry

Issue: Comment says Flow C and social-link both have root-level twins, but only Flow C has a twin in this router (`/auth/callback`); social-link callback lives only at `/social/callback` in `server.ts`.
Proof: `profile.route.ts:53` registers `/auth/callback`; social connect starts at `/identities/social/connect` with no callback route here. Same wording appears in `impersonation.md:194`.
Why it matters: Readers hunting for a social callback in this router will not find one; obscures the asymmetric route layout.
Fix: Clarify Flow C has a twin; social-link callback is only at `/social/callback` in `server.ts`. Apply same correction in `impersonation.md`.

dealako (changes requested):
## Review summary

Production impersonation guard looks correct end-to-end; requesting changes for one test-quality gap.

- 🔴 Blocking: 0
- 🟡 Minor: 1 (social-callback vacuous assertion)
- ⚪ Nit: 2 (comment/doc wording)

See inline comments for details.

dealako:
Hey @fayazg — solid, focused fix for a real gap in the impersonation read-only guarantee.

The production change is well-shaped: `blockCallbackDuringImpersonation` runs first in both root callbacks, redirects with the existing `?error=` convention, and the frontend path (`PROFILE_AUTH_ERROR_MESSAGES` → profile-layout toast) matches the other Flow C errors. Docs and comments mostly explain *why* redirect beats `next(err)` on top-level navigations.

One test-quality gap keeps this from a clean approve: the social-callback spec repeats the vacuous assertion pattern commit `4d09700cb` just fixed on the Flow C sibling. Everything else is comment nits.

**Structured issue count:**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 1 issue: `handleSocialCallback` impersonation test asserts `getManagementToken` (never reached / uncalled anyway) instead of `exchangeCodeForToken` / `linkIdentity`
- ⚪ Nit: 2 issues: incorrect "route signatures don't accept `next`" docstring justification; "root-level twins" wording overstates social-link symmetry
- ❔ Question: 0 items

**AI comment reconciliation:**
- **Cursor Bugbot** — agree with the risk summary; no inline findings to reconcile.
- **CodeRabbit** — review skipped (auto-reviews disabled); nothing to reconcile.
- **Copilot** — reviewed all 8 files with no comments; **disagree** on the social-callback test: `getManagementToken` is a vacuous assertion given mock shape and handler ordering (see inline comment).

**Final decision:** 🔴 **Needs changes before approval**

The guard itself looks correct; please tighten the social-callback test so it asserts the operations the guard actually prevents, mirroring the Flow C fix in `4d09700cb`.

apps/lfx-one/src/server/controllers/profile.controller.spec.ts
fayazg: Done — hoisted a real `socialVerificationSvc` mock (`validateState`, `clearState`, `exchangeCodeForToken`) and added `linkIdentity` to the `emailVerificationSvc` mock, then asserted `socialVerificationSvc.exchangeCodeForToken` and `emailVerificationSvc.linkIdentity` were never called, mirroring the Flow C test.

See commit d7d277978: the guard is now proven to stop the exchange/link, not just the redirect string.

apps/lfx-one/src/server/controllers/profile.controller.ts
fayazg: Fixed — dropped the "route signatures don't even accept next" clause; the docstring now says only that a next(err) there would surface as raw JSON on a top-level navigation instead of a redirect. Good catch, that was misleading.

See commit d7d277978.

apps/lfx-one/src/server/routes/profile.route.ts
fayazg: Fixed — corrected the wording in both this file and impersonation.md: only Flow C has a twin route in this router (/auth/callback); social-link's callback exists solely at /social/callback in server.ts, with no twin here.

See commit d7d277978.

fayazg:
## Review Feedback Addressed

Commit: d7d277978

### Changes Made
- **profile.controller.spec.ts**: hoisted a real `socialVerificationSvc` mock and added `linkIdentity` to the `emailVerificationSvc` mock; the impersonation-blocked social-callback test now asserts `exchangeCodeForToken`/`linkIdentity` were never called instead of a vacuous `getManagementToken` check (per @dealako)
- **profile.controller.ts**: dropped the incorrect "route signatures don't even accept next" justification from the `blockCallbackDuringImpersonation` docstring, keeping only the accurate redirect-vs-403 rationale (per @dealako)
- **profile.route.ts**, **impersonation.md**: corrected the "root-level twins" wording — only Flow C has a twin route in the `/api` router; social-link's callback exists solely at `/social/callback` in `server.ts` (per @dealako)

### Threads Resolved

apps/lfx-one/src/server/controllers/profile.controller.spec.ts
```diff
@@ -356,3 +364,60 @@ describe('ProfileController.getLinuxAlias', () => {
...
+  });
+
+  it('handleSocialCallback redirects to /profile/identities without linking the identity', async () => {
+    isImpersonatingMock.mockReturnValue(true);
+    const res = buildRes();
+
+    await controller.handleSocialCallback(buildReq({ path: '/social/callback', query: { code: 'c', state: 's' } }), res);
+
+    expect(res.redirect).toHaveBeenCalledWith('/profile/identities?error=impersonation_read_only');
+    expect(socialVerificationSvc.exchangeCodeForToken).not.toHaveBeenCalled();
```
dealako, line 517: [minor] Stub getManagementToken so exchangeCodeForToken/linkIdentity assertions can fail

Issue: The new assertions target the right methods, but they still cannot fail independently — without the impersonation guard the handler exits at the `no_management_token` branch before either call is reached.
Proof: `handleSocialCallback` reads `getManagementToken` at `profile.controller.ts:1805` after `validateState`/`clearState`/`!code` checks. This test never stubs it; `vi.fn()` returns `undefined`, so an unguarded path redirects with `?error=no_management_token` (line 1808) and returns. `exchangeCodeForToken` (1814) and `linkIdentity` (1839) are never reached either way.
Why it matters: Only `res.redirect` catches a guard regression today — the assertions added to fix the round-1 vacuous-pattern are still non-discriminating for the operations they name.
Fix: Before invoking the handler, add:
`profileAuthSvc.getManagementToken.mockReturnValue('mgmt-token');`
(optionally `socialVerificationSvc.exchangeCodeForToken.mockResolvedValue({ id_token: 'a.b.c' })` if you want the unguarded path to reach `linkIdentity`). Then the two `not.toHaveBeenCalled()` assertions become real regression coverage.

dealako (changes requested):
## Follow-up review

Great progress on `d7d277978` — both doc nits are resolved and the social-callback test structure is much better.

One refinement remains:
- 🟡 Minor: 1 (mgmt-token stub so new assertions are load-bearing)
- 🔴 Blocking: 0

See inline comment.

dealako:
Hey @fayazg — thanks for the quick turnaround on `d7d277978` — this is much closer.

👏 **Nice work**
- Hoisted a real `socialVerificationSvc` mock and wired `linkIdentity` on `emailVerificationSvc` — exactly the structural fix the Flow C sibling got in `4d09700cb`.
- The social-callback test now asserts `exchangeCodeForToken` / `linkIdentity` instead of the irrelevant `getManagementToken`.
- Docstring and route/docs wording are now accurate: the bogus "route signatures don't accept `next`" claim is gone, and the Flow C vs social-link route asymmetry is spelled out correctly in both `profile.route.ts` and `impersonation.md`.

**Revision tracking (round 1 → round 2)**
- ✅ **Resolved** — Drop incorrect "route signatures don't accept `next`" justification (`profile.controller.ts`, `d7d277978`)
- ✅ **Resolved** — Correct "root-level twins" wording (`profile.route.ts`, `impersonation.md`, `d7d277978`)
- ⚠️ **Partially addressed** — Social-callback vacuous assertion: right methods are now asserted and mocks are wired, but without stubbing `getManagementToken` the handler still exits at `no_management_token` before those calls, so only the redirect assertion actually catches a guard regression today

**Structured issue count (open):**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 1 issue: stub `profileAuthSvc.getManagementToken` (and optionally a resolved token response) so `exchangeCodeForToken` / `linkIdentity` assertions can fail if the guard regresses
- ⚪ Nit: 0 issues

**AI comment reconciliation:** No new CodeRabbit, Cursor Bot, or Copilot comments since the prior review round.

**Final decision:** 🔴 **Needs changes before approval**

One small test fixture tweak and this is good to merge — the production guard itself looks solid.

apps/lfx-one/src/server/controllers/profile.controller.spec.ts
fayazg: Done, stubbed `profileAuthSvc.getManagementToken` to return a token before invoking the impersonation-blocked test, so the `exchangeCodeForToken`/`linkIdentity` assertions actually exercise the guard instead of the unguarded path's own `no_management_token` short-circuit.

See commit 82393f2c3.

fayazg:
## Review Feedback Addressed

Commit: 82393f2c31f671359cda4cf92894c1f4616768fa

### Changes Made
- **apps/lfx-one/src/server/controllers/profile.controller.spec.ts**: stubbed `profileAuthSvc.getManagementToken` to return a token before invoking the `handleSocialCallback` impersonation-blocked test, so the `exchangeCodeForToken`/`linkIdentity` assertions actually exercise the guard instead of being masked by the unguarded path's own `no_management_token` short-circuit (per @dealako)

### Threads Resolved

apps/lfx-one/src/server/controllers/profile.controller.ts
```diff
@@ -1572,6 +1572,10 @@ export class ProfileController {
     const error = req.query['error'] as string;
     const returnTo = this.normalizeProfileReturnTo(req.appSession?.['profileAuthReturnTo']);
 
+    if (this.blockCallbackDuringImpersonation(req, res, returnTo, 'profile_auth_callback')) {
+      return;
+    }
```
audigregorie, line 1582: **Minor — blocked requests leave an orphaned `started` operation log**

`logger.startOperation(req, 'profile_auth_callback')` (line 1568) runs before this guard, so a blocked request emits a `started` entry that never gets a matching terminal `success`/`error` — the guard's `logger.warning` uses a different operation name (`impersonation_readonly`) and carries no `startTime`, so no duration is recorded either. The middleware precedent (`blockDuringImpersonation`) never opens an operation at all, so dashboards pairing start/complete won't see orphans from middleware-blocked routes but will from these two.

Optional polish (applies identically to `handleSocialCallback` at line 1774): hoist the guard — and the `returnTo` computation it needs — above `startOperation`, so a blocked request never opens the operation and the `impersonation_readonly` warning remains the sole log record, same as the middleware path:

```ts
    const returnTo = this.normalizeProfileReturnTo(req.appSession?.['profileAuthReturnTo']);

    if (this.blockCallbackDuringImpersonation(req, res, returnTo, 'profile_auth_callback')) {
      return;
    }

    const startTime = logger.startOperation(req, 'profile_auth_callback');

    const code = req.query['code'] as string;
    const state = req.query['state'] as string;
    const error = req.query['error'] as string;
```

audigregorie (commented):
## Code Review Summary

Clean, well-scoped hardening fix. The in-handler redirect guard is the right call for root-level callbacks outside the `/api` error-handler mount, the tests are genuinely non-vacuous, and the documentation cross-references (controller ↔ route ↔ middleware ↔ server.ts ↔ impersonation.md) are thorough. No Critical or Major findings — one Minor observability nit as a line comment.

### Minor — outside the diff

- **Protected files touched (awareness only, no action needed):** `apps/lfx-one/src/server/server.ts` (server bootstrap) and `apps/lfx-one/src/server/middleware/impersonation-readonly.middleware.ts` (middleware/*) are protected per `.claude/hooks/guard-protected-files.sh`. Both changes here are comment-only — zero behavioral delta — flagged per the backend checklist's protected-files item.

### Verification performed

- `vitest run src/server/controllers/profile.controller.spec.ts` — **18/18 pass**, incl. the 4 new impersonation-blocked callback tests
- `ng test --include='**/profile-layout.component.spec.ts'` — **6/6 pass**, incl. the new `impersonation_read_only` toast test
- `tsc --noEmit -p tsconfig.app.json` — clean on all PR-touched files (the single error is a gitignored generated docs manifest, pre-existing and unrelated)
- `yarn check-types` (turbo) — green

### What's done well

- The social-callback test stubs `getManagementToken` to return a token specifically so the `not.toHaveBeenCalled()` assertions on `exchangeCodeForToken`/`linkIdentity` can actually fail — and documents why. The prior `SocialVerificationService` mock returning `{}` (vacuous assertions) is fixed in the same pass.
- The in-handler guard mirrors the middleware's log shape exactly (same `impersonation_readonly` operation, same `impersonator_sub`/`target_sub` metadata), so blocked events aggregate identically whichever guard fired.
- Double-toast ownership stays clean: `impersonation_read_only` lives only in `PROFILE_AUTH_ERROR_MESSAGES`, and `ProfileIdentitiesComponent` already defers to that map via `Object.hasOwn` — the `/social/callback` redirect to `/profile/identities` toasts exactly once.
- The blocked redirect target goes through the allowlisted `normalizeProfileReturnTo`, so there's no open-redirect vector.

audigregorie (dismissed):

apps/lfx-one/src/server/controllers/profile.controller.ts
fayazg: Done, hoisted the impersonation guard (and the `returnTo` computation it needs) above `logger.startOperation()` in both `handleProfileAuthCallback` and `handleSocialCallback`, so a blocked request never opens an operation log without a matching terminal entry — matching `blockDuringImpersonation`'s log behavior on the same guard.

See commit c2265811a.

fayazg:
## Review Feedback Addressed + Merge Conflict Resolved

Commit: c2265811a8bb0c63c58725574139acd762bfac2f

Branch was rebased onto \`origin/main\` to pick up #2183 (which had landed on \`main\` and touched the same files) and resolve the merge conflict; the PR is now mergeable.

### Changes Made
- **apps/lfx-one/src/server/controllers/profile.controller.ts**: hoisted the impersonation guard (and the `returnTo` computation it needs) above `logger.startOperation()` in `handleProfileAuthCallback` and `handleSocialCallback`, so a blocked request never opens an operation log without a matching terminal entry — matching `blockDuringImpersonation`'s log behavior on the same guard (per @audigregorie)

### Threads Resolved

apps/lfx-one/src/server/controllers/profile.controller.spec.ts
```diff
@@ -442,3 +450,64 @@ describe('ProfileController.startProfileAuth — returnTo allowlist', () => {
...
+  beforeEach(() => {
+    vi.clearAllMocks();
+    controller = new ProfileController();
+  });
+
+  it('handleProfileAuthCallback redirects to the default returnTo without exchanging the code', async () => {
+    isImpersonatingMock.mockReturnValue(true);
+    const res = buildRes();
+
+    await controller.handleProfileAuthCallback(buildReq({ path: '/passwordless/callback', query: { code: 'c', state: 's' } }), res);
```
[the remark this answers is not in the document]
fayazg: Done, matched `appSession.profileAuthState` to the query's `state` in the `handleProfileAuthCallback` "without exchanging the code" test.

See commit fcd9bffc8: without the matching state, the unguarded path would already stop at the pre-existing `invalid_state` check before `exchangeCodeForToken`, so the assertion wasn't exercising the impersonation guard. Verified the fix is load-bearing by temporarily disabling the guard locally — the test now fails against `login_session_invalid` (reaches `exchangeCodeForToken`) as expected, confirming it now catches a regression.

fayazg:
## Review Feedback Addressed

Commit: fcd9bffc8

### Changes Made
- **apps/lfx-one/src/server/controllers/profile.controller.spec.ts**: matched `appSession.profileAuthState` to the query's `state` in the `handleProfileAuthCallback` "without exchanging the code" test, so the `exchangeCodeForToken` assertion is load-bearing against the impersonation guard instead of the pre-existing `invalid_state` check (per copilot-pull-request-reviewer[bot])

### Threads Resolved
1 of 1 unresolved threads addressed in this iteration — no threads remain open.

apps/lfx-one/src/server/controllers/profile.controller.spec.ts
```diff
@@ -442,3 +450,71 @@ describe('ProfileController.startProfileAuth — returnTo allowlist', () => {
...
+
+    expect(res.redirect).toHaveBeenCalledWith('/profile?error=invalid_state');
+  });
+
+  it('handleSocialCallback redirects to /profile/identities without linking the identity', async () => {
+    isImpersonatingMock.mockReturnValue(true);
+    // Without a management token, the unguarded path would also stop before exchangeCodeForToken/
+    // linkIdentity (at the no_management_token branch) — stub it so those assertions can actually
+    // fail if the impersonation guard regresses.
+    profileAuthSvc.getManagementToken.mockReturnValue('mgmt-token');
```
dealako, line 511: [nit] Comment overstates linkIdentity assertion coverage

Issue: The comment says stubbing `getManagementToken` lets both `exchangeCodeForToken` and `linkIdentity` assertions fail on a guard regression, but only `exchangeCodeForToken` can today.
Proof: Without the guard, the handler reaches `exchangeCodeForToken` (now stubbed mgmt token passes line 1805), but `exchangeCodeForToken` is an unimplemented `vi.fn()` returning `undefined`, so `tokenResponse.id_token` throws and the outer catch redirects with `social_verification_failed` before `linkIdentity` (1839) is reached.
Why it matters: Harmless for merge — `exchangeCodeForToken` is the primary guard target — but the comment documents coverage that isn't quite there yet.
Fix (optional): `socialVerificationSvc.exchangeCodeForToken.mockResolvedValue({ id_token: 'a.b.c' });` — then the `linkIdentity` assertion becomes load-bearing too.

dealako:
Hey @fayazg — thanks for sticking with this through the rebase onto `main`; the extra commits (`e3e1c01fd` → `fcd9bffc8`) close out the feedback cleanly.

👏 **Nice work**
- Stubbed `profileAuthSvc.getManagementToken` in the social-callback test so `exchangeCodeForToken` is now a real regression guard.
- Matched `appSession.profileAuthState` to the query `state` in the Flow C test (`fcd9bffc8`) — same fix pattern, and it makes that assertion load-bearing too.
- Hoisted `blockCallbackDuringImpersonation` above `logger.startOperation` in both callbacks (`c2265811a`) so a blocked path no longer leaves a dangling operation on the stack.
- Rebase conflict with #2183 resolved; PR is mergeable again.

**Revision tracking (round 2 → round 3)**
- ✅ **Resolved** — Stub `getManagementToken` so social-callback assertions exercise the guard (`e3e1c01fd` / `c2265811a`)
- ✅ **Resolved** — Flow C test: match CSRF state so `exchangeCodeForToken` assertion is load-bearing (`fcd9bffc8`)
- ✅ **Resolved** (bonus) — Guard runs before `startOperation` in both root callbacks (`c2265811a`)

**Structured issue count (open):**
- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 1 issue: social-callback test comment claims both `exchangeCodeForToken` and `linkIdentity` assertions are load-bearing; only the former is today — optional `exchangeCodeForToken.mockResolvedValue({ id_token: 'a.b.c' })` would close the gap

**AI comment reconciliation:** No new CodeRabbit, Cursor Bot, or Copilot comments since the round-2 review.

The impersonation guard and its regression tests look solid. The remaining nit is optional polish on test-comment accuracy / `linkIdentity` coverage — not blocking merge.

dealako (approved):
## Follow-up review (round 3)

All prior feedback addressed. Approving.

- ✅ Management-token stub + Flow C CSRF state match
- ✅ Guard hoisted above `startOperation`
- ⚪ Optional nit: stub token exchange if you want `linkIdentity` assertion load-bearing too

Nice work on the rebase.

PR #2185, opened 2026-09-04
test(campaigns): pin the round-9 guards and finish the doc sweep

mrautela365:
Follow-up to #1923, which merged while these were in flight. @dealako raised them 75 seconds before the merge and approved anyway — all six are minor or nit, and none blocked.

## Two close a real gap

Both are the class caught repeatedly on #1923: a fix shipping with no test that fails when reverted.

- **Cross-foundation completeness** — the `capped !== false` fix had no test distinguishing it from the bug it replaced. The two tests reaching that arm pass `capped: true` and `capped: false`, which `=== true` handles identically. Only an **absent** field separates them, and that is exactly the old-pod shape the fix exists for.
- **Legacy create id validation** — no test either. Deleting the guard and restoring the `as { id: string }` cast left the suite green, which is the state R8-min4 was about.

**Both mutation-verified:** reverting either fails a named test.

| Reverted to | Fails |
|---|---|
| `capped === true` | `treats an ABSENT completeness field on the cross-foundation arm as unproven` |
| `as { id: string }` | 4 cases in `refuses a create 2xx carrying …` |

## Four are documentation

- The R8 comment edit left a 46-column stub mid-paragraph; reflowed evenly.
- The interface banner still had the two campaign-service campaign types under the wrong heading. `// Keyword actions` now covers only `KeywordActionOutcome`, and all four HubSpot types sit under `// HubSpot UTM`.
- `values.yaml` claimed *"unlike every other flag on this list"* while the HubSpot UTM note 44 lines below says the same thing; narrowed to the flags it is actually true of.
- Untangled a garden-path clause in the new validation comment.

## Verification

**99 files / 2683 server tests, 88 / 1859 app tests, typecheck and lint clean** — run against merged `main`.

No behaviour changes beyond the two guards already merged in #1923; this adds the tests that hold them in place.

### DIFF
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.spec.ts
@@ -5032,7 +5032,7 @@ describe('CampaignsComponent — HubSpot template picker', () => {
 
   /**
    * campaign-service answers the SAME typed 404 for an absent connection row and for a project id
-   * that does not exist (`campaign.interface.ts:1223-1226`), so this copy is the only place a
+   * that does not exist (the typed 404 both cases share), so this copy is the only place a
    * mistyped slug can be distinguished from an unconfigured one. Naming nothing reported every
    * typo as a missing integration.
    */
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/campaigns.component.ts
@@ -1088,7 +1088,7 @@ export class CampaignsComponent {
    * Names the project the search actually queried, for the "connect HubSpot" empty state.
    *
    * campaign-service answers the same typed 404 for an absent connection row and for a project id
-   * that does not exist (`campaign.interface.ts:1223-1226`), so "not connected" and "no such
+   * that does not exist (the typed 404 both cases share), so "not connected" and "no such
    * project" are indistinguishable here. Naming the slug is what makes a typo visible instead of
    * being reported as a missing integration.
    *
--- a/apps/lfx-one/src/app/modules/dashboards/campaigns/components/planning-tab/planning-tab.component.spec.ts
+++ b/apps/lfx-one/src/app/modules/dashboards/campaigns/components/planning-tab/planning-tab.component.spec.ts
@@ -1890,6 +1890,40 @@ describe('PlanningTabComponent — HubSpot UTM states', () => {
     expect(instance()['hsCompletenessUnproven'](), 'the capped search rendered the plain no-match remedy').toBe(true);
   });
 
+  it('treats an ABSENT completeness field on the cross-foundation arm as unproven', () => {
+    // dealako (#1923): the `capped !== false` fix here shipped with no test that distinguishes it
+    // from the bug it replaced -- the two tests reaching this arm pass `capped: true` and
+    // `capped: false`, and `=== true` handles both identically. Only an ABSENT field separates
+    // them, which is exactly the old-pod shape the fix exists for.
+    const ctx = TestBed.inject(ProjectContextService);
+    const empty = { found: false, hs_utm: null, campaign_name: '', all_matches: [], capped: false, inconclusive: false };
+
+    runLookup(empty, 'KubeCon NA 2026');
+    create.mockReturnValue(
+      new Observable((o) => {
+        o.next({ created: true, hs_utm: null, campaign_name: 'Kubecon Na 2026' });
+        o.complete();
+      })
+    );
+    (fixture.componentInstance as unknown as { createInHubSpot(): void }).createInHubSpot();
+    fixture.detectChanges();
+
+    ctx.setFoundation({ uid: 'foundation-b-uid', slug: 'foundation-b', name: 'Foundation B' }, false);
+    fixture.detectChanges();
+    // No `capped`, no `inconclusive` -- a response from a pod that predates both fields.
+    lookup.mockReturnValue(
+      new Observable((o) => {
+        o.next({ found: false, hs_utm: null, campaign_name: '', all_matches: [] } as never);
+        o.complete();
+      })
+    );
+    (fixture.componentInstance as unknown as { recheckHubSpot(): void }).recheckHubSpot();
+    fixture.detectChanges();
+
+    expect(instance()['hsCompletenessUnproven'](), 'an absent completeness field read as proven-complete').toBe(true);
+    expect(instance()['hsCreateSuppressed'](), 'an unprovable cross-foundation response licensed a create').toBe(true);
+  });
+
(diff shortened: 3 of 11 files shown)


mrautela365:
@dealako ready for review — this closes the six items you raised on #1923 just before it merged.

**Two are the substantive ones**, both mutation-verified (each fails when its guard is reverted):

- **Cross-foundation completeness** — the `capped !== false` fix had no test separating it from the bug it replaced. The existing tests pass `capped: true` and `false`, which `=== true` handles identically; only an **absent** field distinguishes them, and that is the old-pod shape the fix exists for.
- **Legacy create id guard** — no test at all. Reverting now fails on `promise resolved "{ created: true, hs_utm: null, … }" instead of rejecting` — the real defect.

**Four were documentation**, all in.

## What the pre-PR review pass changed

I ran the reviewer simulation and the repo's own code/learnings reviewers before asking you. They caught one thing worth flagging:

**My create-guard test was vacuous.** It queued one fetch response while `hubspotCreateCampaign` makes two, so reverting the guard failed on a `TypeError` from the unmocked second call rather than the creation claim. It failed either way, so it looked binding while pinning a mock-arity artifact — the same class this PR exists to close, one level down. Fixed and re-verified.

The reviewers disagreed with each other on this, and I resolved it by running it myself: one had tested against an already-reverted file left in the shared worktree by another agent, so its "revert" was a no-op.

## An honest note on the doc churn

Five of the commits here are Copilot rounds on my own fixes, each one a real problem the previous fix created. The recurring shape was **comparing one flag's rollback story to the others** — "unlike every other flag", then "unlike the create-pipeline flags" — which was false every time I narrowed it. `JOBS` must stay on and comes off last; legacy `CREATE` reaches the same `getGadsClient()` through `getCustomer()`. All three copies now state the mechanism and explicitly decline to promise anything about the other flags, which is what should stop it recurring.

Also corrected: the 500 rationale cited a route that no longer exists (a post-2xx parse failure is now classified as a transport 503), and the live route is stronger anyway — `createHubSpotCampaign` throws after the create has landed, so the campaign **may** already exist. That uncertainty is precisely why 500 stays unconfirmed.

**MERGEABLE · 0 conflicts · 0 unresolved threads · all checks green · 99 files / 2683 server tests, 88 / 1859 app tests, typecheck and lint clean.**

Bots may still be reviewing the final head (`e8f5415c9`); I'll take anything they add in one pass rather than another round.

mrautela365:
@dealako ready for review — everything is closed on head `ffc8af95e`.

**0 conflicts · 0 unresolved threads · 0 inline comments on head · 0 Copilot suppressed comments · all checks green** (99 files / 2683 server tests, 88 / 1859 app tests, typecheck and lint clean).

This closes the six items you raised on #1923 seventy-five seconds before it merged.

**The two substantive ones are mutation-verified** — each fails when its guard is reverted, with the right diagnostic:

| Reverted to | Fails with |
|---|---|
| `capped === true` | `an absent completeness field read as proven-complete` |
| `as { id: string }` | `promise resolved "{ created: true, hs_utm: null, … }" instead of rejecting` |

The cross-foundation one matters because the existing tests pass `capped: true` and `false`, which `=== true` handles identically — only an **absent** field separates them, and that is the old-pod shape the fix exists for.

## Worth your attention

**My first version of the create-guard test was vacuous.** It queued one fetch response while `hubspotCreateCampaign` makes two, so reverting the guard failed on a `TypeError` from the unmocked second call rather than the creation claim. It failed either way, so it looked binding while pinning a mock-arity artifact — the same class this PR exists to close, one level down. Caught in the pre-PR pass, fixed, re-verified.

**A doc claim I had to correct four times.** The flag notes kept ranking one flag's rollback story against the others — "unlike every other flag", then "unlike the create-pipeline flags" — and each narrowing was still false: `JOBS` must stay on and comes off LAST, and legacy `CREATE` reaches the same `getGadsClient()` through `getCustomer()`. All three copies now state the mechanism (it is a property of the `GADS_*` credentials, shared by every legacy path reaching that client) and explicitly decline to promise anything about the other flags. That framing is what should stop it recurring.

**Two claims in the 500 rationale were wrong** and are now corrected: it cited a post-2xx `JSON.parse` route that no longer produces a 500 (a `SyntaxError` is classified as a transport 503), and it said the campaign "exists" at that point when the whole reason 500 stays unconfirmed is that existence is **unknown**.

The doc churn here is larger than the change deserved, and most of it was my own fixes creating the next finding. The substance is the two binding tests; the rest is making the flag notes stop contradicting each other.

dealako (approved):
## Review summary

Approved. This is a test-and-documentation follow-up to #1923 with no intentional runtime changes.

**Highlights**
- Cross-foundation completeness regression test correctly targets the absent `capped` field (old-pod shape); `capped: true`/`false` alone would not catch a `=== true` revert.
- HubSpot create id-guard tests are binding: both fetches are mocked so reverting the guard fails on the creation claim, not mock arity.
- Flag documentation now distinguishes "nothing stranded" from "off works only when `GADS_*` is live", resolving earlier contradictory rollback guidance.

**Findings:** none after proof/trace review and self-challenge.

**Bot reconciliation:** Copilot themes from intermediate commits (contradictory INSIGHTS wording, inaccurate flag comparisons, stale 500 rationale, vacuous mock) are addressed on head `ffc8af95`.

✅ **Approved**

dealako:
@mrautela365 — nice follow-up. This PR does exactly what it says on the tin: it pins the two guards from #1923 with tests that actually distinguish the failure modes, and it finishes the flag-documentation sweep without changing runtime behavior.

The cross-foundation completeness test is the important one — passing `capped: true`/`false` would not have caught a revert to `=== true`, and the absent-field case matches the old-pod shape the fix exists for. The HubSpot create guard tests are equally well thought through: queuing both fetches so a revert fails on the creation claim rather than mock arity is the kind of binding test this repo has been missing.

The documentation churn is substantive too. Separating "nothing stranded" from "off works when GADS_* is dead" removes the contradictory operator guidance Copilot flagged across earlier commits, and the corrected 500 rationale (post-create validation uncertainty vs. stale JSON.parse path) matches the code.

## Structured issue count

- 🔴 Blocking: 0 issues
- 🟡 Minor: 0 issues
- ⚪ Nit: 0 issues
- ❔ Question: 0 items

## Bot reconciliation

- **Copilot (earlier rounds):** Agree the head commit addresses the suppressed-comment themes — contradictory INSIGHTS rollback wording, inaccurate JOBS/CREATE comparisons, stale 500/JSON.parse rationale, and vacuous create-guard mock arity. No remaining actionable items on `ffc8af95`.
- **CodeRabbit:** Skipped (repo config); nothing to reconcile.
- **Cursor Bugbot:** Summary aligns with the diff; no conflicts.

## Prior rounds

No prior formal human review on this PR. Copilot `COMMENTED` reviews on intermediate commits; the issues they raised were addressed in later commits on this branch.

## Final decision

✅ **Approved**

PR #2191, opened 2026-09-04
fix(search): rank project search by relevance, not alphabetically

### LINKED ISSUE #2030: Project search returns too many results and grows with more query terms (per-term OR instead of AND, no phrase ranking)
## Summary

Project search in lfx-self-serve appears to tokenize the query and match each term independently rather than as a phrase, and it does not rank exact/prefix matches to the top. This produces oversized, poorly-ordered result sets and — counterintuitively — **adding more query terms increases the result count instead of narrowing it**, which is the opposite of what a user expects from search.

## Reproduction

1. Open the project search in lfx-self-serve.
2. Search: `Cloud Native`
   - Expected: CNCF (Cloud Native Computing Foundation) is the top result; the list is short.
   - Actual: CNCF is not the top result; 10+ results come back matching either "Cloud" or "Native" independently.
3. Refine to: `Cloud Native Computing`
   - Expected: fewer results, more precise; CNCF at or near the top.
   - Actual: **more results** than the two-word query returned, because the third term is OR'd in rather than AND'd.

## Impact

- Users cannot reliably find a specific project by name.
- Typing more of the project name — the natural refinement behavior — makes results worse, not better.
- Well-known foundations like CNCF should be trivially findable by their common name; today they are not.

## Expected behavior

- Query terms should be AND'd (all terms must match), not OR'd. Adding a term must never grow the result set.
- Multi-word queries should treat the query as a phrase for ranking purposes, with exact / prefix matches on the project name ranked above partial or descriptive matches.
- Common acronyms should map to their expansions (e.g. `CNCF` ↔ `Cloud Native Computing Foundation`), either via an alias field or synonym handling.
- Reasonable default cap on result count with the best matches surfaced first.

## Notes

Please confirm whether the search is backed by the query service, a direct DB call, or something else, so we can scope the fix (indexing/ranking config vs. query construction on the BFF).

MRashad26:
Closes #2030

## Problem

Searching for a project returned too many results, and the set *grew* with
each extra word typed. The project the user was looking for was frequently
absent from the visible list entirely.

Three independent layers each discarded relevance, and all three had to be
fixed for the search to behave:

1. **Upstream (query-service).** The `name` search runs a `multi_match` of
   type `bool_prefix`, which defaults to `OR`. Every additional word matched
   *more* documents, so `Cloud Native Computing` returned a superset of
   `Cloud Native`. Fixed upstream — see External references.
2. **BFF (this repo).** `sort` defaults to `name_asc` upstream, and OpenSearch
   discards relevance scoring entirely whenever an explicit non-`_score` sort
   is present. `searchProjects` and `searchCreatableProjects` passed `name`
   without `sort`, so they received the *alphabetically first* page of matches
   rather than the closest ones. `searchCreatableProjects` caps its page scan,
   so the intended project could be filtered out before the user ever saw it.
3. **Frontend (this repo).** `ProjectSelectorComponent` re-sorted whatever
   arrived by role tier and then alphabetically, which threw away any ordering
   the first two layers produced.

## Changes

**Backend** — `searchProjects` and `searchCreatableProjects` now send
`sort: 'best_match'`, on every page rather than just the first. The two other
query-service name-search callers (`navigation.service.ts`,
`org-navigation.service.ts`) already did; this brings the remaining two in line.

**Frontend** — the selector preserves source order while showing search
results and keeps role-tier ordering only for the unfiltered browse list.

Two follow-ups from local review, both about that fix actually reaching the user:

- The hybrid **All** tab — the sidebar's default view — rebuilt every result
  set foundation-first with projects nested under their parents. Applied to
  search results that hides matches: a strong project match was filed under
  wherever its parent foundation happened to rank, or, when the parent didn't
  match, escaped nesting only via the flat tail. The two groups are now emitted
  flat while searching, each in its own relevance order. Foundations still lead,
  because the two lenses are separate upstream queries whose scores aren't
  comparable — the Foundations and Projects tabs give a single-group view.
- Ordering keyed off the term the user had *typed*, which leads the term the
  visible items were *fetched for* by the 300ms search debounce plus request
  latency, so the browse list was restyled once before results arrived and
  again when they landed. `NavigationService` now tracks `resultsTerm` per
  lens and the selector orders on that, so each search reshuffles once.

**Docs** — `best_match` was missing from the query-service `sort` enum in
`docs/architecture/backend/pagination.md` and `docs/reviews/shared-and-sql-checklist.md`.
Both omissions predate this branch, but they contradicted shipped code and
this change makes `best_match` a contract these call sites depend on.

## Testing

- `project.service.spec.ts` asserts `sort: 'best_match'` on both entry points
  and on subsequent pages, not just the first.
- New `project-selector.component.spec.ts` covers the ordering contract:
  alphabetical while browsing, source order preserved while searching, back to
  alphabetical once cleared, All-tab nesting only while browsing, and the
  browse-until-results-arrive timing rule.
- Full suite green: 1865 app + 99 server + 59 shared. `yarn build`,
  `lint:check`, `format:check`, `check-types`, and `check-headers.sh` all pass.

## External references

Upstream `operator: "and"` fix: linuxfoundation/lfx-v2-query-service#70

The OR behavior was confirmed empirically against a local OpenSearch instance
loaded with the platform's actual index mapping — with the default operator an
added term widened the result set, and with `operator: and` it narrowed it.

Layers 2 and 3 are independently useful and safe to merge before the upstream
change lands: relevance ordering of an over-broad result set still surfaces the
closest match first. Result-set size only tightens once #70 deploys.

### DIFF
--- a/apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.spec.ts
+++ b/apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.spec.ts
@@ -8,14 +8,16 @@ import { LensItem } from '@lfx-one/shared/interfaces';
 import { NavigationService } from '@services/navigation.service';
 import { PersonaService } from '@services/persona.service';
 import { ProjectContextService } from '@services/project-context.service';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { ProjectSelectorComponent } from './project-selector.component';
 
 /**
  * Covers the ordering contract only. Curated mode is used as the harness because it sources items
  * from the `items` input rather than `NavigationService`, which keeps the assertion about *order*
- * free of upstream pagination — the browse-vs-search branch under test is shared by both modes.
+ * free of upstream pagination and debouncing — the browse-vs-search branch under test is shared by
+ * both modes. The nav-mode timing rule (order on the term the results were fetched for, not the
+ * term the user typed) is covered separately below.
  */
 describe('ProjectSelectorComponent ordering', () => {
   // Deliberately not alphabetical, and every entry matches "cloud" so the search assertion is
@@ -94,5 +96,127 @@ describe('ProjectSelectorComponent ordering', () => {
     await search('');
 
     expect(renderedSlugs()).toEqual(['alpha-cloud', 'mid-cloud', 'zeta-cloud']);
+  });
+});
+
+/**
+ * Nav-backed mode, where the typed term and the term the visible items were fetched for diverge
+ * across the search debounce and the request round-trip.
+ */
+describe('ProjectSelectorComponent nav-backed ordering', () => {
+  const foundations: LensItem[] = [
+    { uid: 'f2', slug: 'zeta-foundation', name: 'Zeta Foundation', logoUrl: null, isFoundation: true },
+    { uid: 'f1', slug: 'alpha-foundation', name: 'Alpha Foundation', logoUrl: null, isFoundation: true },
+  ];
+  const projects: LensItem[] = [
+    { uid: 'p2', slug: 'zeta-project', name: 'Zeta Project', logoUrl: null, isFoundation: false },
+    { uid: 'p1', slug: 'alpha-project', name: 'Alpha Project', logoUrl: null, isFoundation: false },
+  ];
+
+  let fixture: ComponentFixture<ProjectSelectorComponent>;
+  let resultsTerm: ReturnType<typeof signal<string>>;
+
+  // Nav mode (unlike curated) runs the sidebar-relative panel alignment on popover show, which
+  // reads `window.matchMedia` — absent in jsdom. Reporting no match takes the early return, so the
+  // geometry path stays out of an ordering test.
+  afterEach(() => vi.unstubAllGlobals());
+
+  async function build(hybrid: boolean): Promise<void> {
+    vi.stubGlobal('matchMedia', () => ({ matches: false }));
+    resultsTerm = signal('');
+    TestBed.resetTestingModule();
+    await TestBed.configureTestingModule({
+      imports: [ProjectSelectorComponent],
+      providers: [
+        provideNoopAnimations(),
+        {
+          provide: NavigationService,
+          useValue: {
+            items: (lens: string) => signal(lens === 'foundation' ? foundations : projects),
+            hasMore: () => signal(false),
+            loading: () => signal(false),
+            // Both lenses are searched with the same term and share this signal, mirroring the
+            // parallel dispatch the component performs in hybrid mode.
+            resultsTerm: () => resultsTerm,
+            setSearchTerm: () => undefined,
+            loadNextPage: () => undefined,
+          },
+        },
+        {
+          provide: PersonaService,
+          useValue: {
+            personaProjects: signal({}),
+            // Makes p1 a child of f1, so the All tab has something to nest while browsing.
+            detectedProjects: signal([{ projectUid: 'p1', parentProjectUid: 'f1', isFoundation: false }]),
+            allPersonas: signal([]),
+          },
+        },
+        { provide: ProjectContextService, useValue: { isFoundationContext: signal(false) } },
+      ],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(ProjectSelectorComponent);
+    fixture.componentRef.setInput('lens', 'project');
+    fixture.componentRef.setInput('hybridMode', hybrid);
+    await fixture.whenStable();
+    (fixture.nativeElement.querySelector('[data-testid="project-selector"]') as HTMLButtonElement).click();
+    await fixture.whenStable();
+  }
+
+  function renderedSlugs(): string[] {
+    return Array.from(document.body.querySelectorAll('[data-testid^="lens-item-"]')).map((el) =>
+      (el.getAttribute('data-testid') ?? '').replace('lens-item-', '')
+    );
+  }
+
+  function nestedSlugs(): string[] {
+    return Array.from(document.body.querySelectorAll('[data-testid^="lens-item-"]'))
+      .filter((el) => el.parentElement?.classList.contains('pl-5'))
+      .map((el) => (el.getAttribute('data-testid') ?? '').replace('lens-item-', ''));
+  }
+
+  async function type(term: string): Promise<void> {
+    const input = document.body.querySelector('[data-testid="project-search-input"]') as HTMLInputElement;
+    input.value = term;
+    input.dispatchEvent(new Event('input'));
+    await fixture.whenStable();
+  }
+
+  it('keeps browse ordering until the search results actually arrive', async () => {
+    await build(false);
+
+    // Typed, but NavigationService has not come back yet — the list on screen is still the browse
+    // list, so re-ordering it now would reshuffle twice for one search.
+    await type('zeta');
+
+    expect(renderedSlugs()).toEqual(['alpha-project', 'zeta-project']);
+  });
+
+  it('switches to relevance ordering once the results reflect the search', async () => {
+    await build(false);
+    await type('zeta');
+
+    resultsTerm.set('zeta');
+    await fixture.whenStable();
+
+    expect(renderedSlugs()).toEqual(['zeta-project', 'alpha-project']);
+  });
+
+  it('nests projects under their parent foundation on the All tab while browsing', async () => {
+    await build(true);
+
+    expect(nestedSlugs()).toEqual(['alpha-project']);
+  });
+
+  it('flattens the All tab while searching so a match is not hidden under its parent', async () => {
+    await build(true);
+    await type('zeta');
+    resultsTerm.set('zeta');
+    await fixture.whenStable();
+
+    // Foundations still lead — the two lenses are separate queries with non-comparable scores —
+    // but each group keeps its own relevance order and nothing is nested out of view (GH-2030).
+    expect(nestedSlugs()).toEqual([]);
+    expect(renderedSlugs()).toEqual(['zeta-foundation', 'alpha-foundation', 'zeta-project', 'alpha-project']);
   });
 });
--- a/apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.spec.ts
+++ b/apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.spec.ts
@@ -1,0 +1,98 @@
+// Copyright The Linux Foundation and each contributor to LFX.
+// SPDX-License-Identifier: MIT
+
+import { signal } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { provideNoopAnimations } from '@angular/platform-browser/animations';
+import { LensItem } from '@lfx-one/shared/interfaces';
+import { NavigationService } from '@services/navigation.service';
+import { PersonaService } from '@services/persona.service';
+import { ProjectContextService } from '@services/project-context.service';
+import { beforeEach, describe, expect, it } from 'vitest';
+
+import { ProjectSelectorComponent } from './project-selector.component';
+
+/**
+ * Covers the ordering contract only. Curated mode is used as the harness because it sources items
+ * from the `items` input rather than `NavigationService`, which keeps the assertion about *order*
+ * free of upstream pagination — the browse-vs-search branch under test is shared by both modes.
+ */
+describe('ProjectSelectorComponent ordering', () => {
+  // Deliberately not alphabetical, and every entry matches "cloud" so the search assertion is
+  // about ordering rather than filtering.
+  const items: LensItem[] = [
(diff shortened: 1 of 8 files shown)


apps/lfx-one/src/app/shared/services/navigation.service.ts
```diff
@@ -291,6 +303,7 @@ export class NavigationService {
       merge(firstPage$, nextPage$).pipe(
         // Drop responses from a superseded generation (e.g., a scroll fetch that lands after a new search reset).
         filter(({ generation: pageGen }) => pageGen === generation()),
+        tap(({ term }) => resultsTerm.set(term)),
```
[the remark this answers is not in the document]
MRashad26: Good catch, and it is worse than just the tag: `nextPage$` already paired the cursor with `searchTerm()`, so a scroll during the debounce appended one query's page onto another query's list. This branch only made that visible by tagging the mixed result as settled.

Fixed in 5b973d184 — continuations now use `resultsTerm()`. The page that issues a cursor is the same one that sets `resultsTerm`, so cursor, items, and tag always describe one query, and the newer term still gets its own reset fetch when the debounce fires.

docs/architecture/backend/pagination.md
```diff
@@ -9,20 +9,27 @@ The application uses cursor-based pagination (via `page_token`) rather than offs
...
-| `tags`       | `string`   | Exact tag matching                                                 |
-| `tags_all`   | `string`   | Multiple tag matching (AND logic)                                  |
-| `type`       | `string`   | Resource type filter                                               |
-| `parent`     | `string`   | Parent resource filter                                             |
-| `sort`       | `string`   | Sort order: `name_asc`, `name_desc`, `updated_asc`, `updated_desc` |
+| Parameter    | Type       | Description                                                                                           |
+| ------------ | ---------- | ----------------------------------------------------------------------------------------------------- |
+| `page_token` | `string`   | Opaque cursor from a previous response. Omit for the first page.                                      |
+| `page_size`  | `number`   | Items per page (default: 50, range: 1–1000)                                                           |
+| `name`       | `string`   | Typeahead search (multi_match with bool_prefix); every term must match, last term matches as a prefix |
```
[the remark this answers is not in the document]
MRashad26: Agreed — the PR body says this is safe to merge ahead of query-service#70, so documenting AND semantics as the current contract would be wrong until that deploys.

Fixed in 5b973d184. The table row no longer claims term semantics, and the note states the OR behavior that ships today, links query-service#70, and asks for the note to be updated once it lands. Relevance ordering is the part that holds either way, which is why `best_match` is worth passing now.

docs/reviews/shared-and-sql-checklist.md
```diff
@@ -219,13 +219,18 @@ const binds = [email];
...
-| `page_token` | Cursor for pagination      | Opaque string from previous response                         |
-| `name`       | Typeahead search           | Uses `multi_match` with `bool_prefix`                        |
-| `filters`    | Field filtering            | Format: `field:value`, auto-prefixed with `data.`            |
-| `sort`       | Sort order                 | Enum: `name_asc`, `name_desc`, `updated_asc`, `updated_desc` |
+| Parameter    | Purpose                    | Note                                                                                          |
+| ------------ | -------------------------- | --------------------------------------------------------------------------------------------- |
+| `page_size`  | Number of results per page | NOT `limit`                                                                                   |
+| `page_token` | Cursor for pagination      | Opaque string from previous response                                                          |
+| `name`       | Typeahead search           | Uses `multi_match` with `bool_prefix`; every term must match                                  |
```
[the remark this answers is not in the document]
MRashad26: Same fix as the pagination-doc thread — 5b973d184 drops the "every term must match" clause from this table too, leaving the semantics claim in the one place that carries the pending-upstream caveat.

dealako:
@MRashad26 — nice work tracing the relevance loss through all three layers (query-service OR semantics, BFF `name_asc` default, and frontend re-sort). The `resultsTerm` / cursor pairing fix in particular is the kind of debounce-window bug that's easy to miss and painful in production.

The BFF `best_match` change, selector ordering contract, All-tab flattening during search, and the new specs give me confidence the fix actually reaches the user. Copilot's earlier pagination and docs feedback looks addressed in `5b973d18`.

**Issue count**

- 🔴 Blocking: 0 issues
- 🟡 Minor: 2 issues: missing `NavigationService` unit tests for the cursor/term pairing; docs overclaim `best_match` for every `name` query
- ⚪ Nit: 2 issues: popover spec teardown; stale/misleading JSDoc on `curatedItems` and `isSearchResult`
- ❔ Question: 1 item: whether `Closes #2030` is appropriate before upstream prefix-ranking lands (acknowledged in PR body; Copilot suppressed comment)

**Revision tracking (Copilot round 1 → `5b973d18`)**

- ✅ Pagination cursor paired with `resultsTerm()` instead of live `searchTerm()` — fixed
- ✅ Docs no longer claim AND term semantics before query-service#70 deploys — fixed

**Bot reconciliation**

- Agree with Copilot's original pagination and docs comments; both addressed.
- Agree with Copilot's suppressed note that `committee.service` / `search.service` still omit `best_match` while the new docs say "always" — incorporated as a minor doc-scope finding (not a blocker for this PR's project-search scope).

apps/lfx-one/src/app/shared/services/navigation.service.ts
```diff
@@ -285,12 +297,17 @@ export class NavigationService {
       })
     );
 
-    const nextPage$ = loadMore$.pipe(switchMap((token) => this.fetchSinglePage(lens, searchTerm(), token, loading, false, generation(), generation, null)));
+    // Continue with the term the cursor belongs to, not the term being typed. `nextPageToken` is
+    // set by the page that also set `resultsTerm`, so the two always describe the same query;
+    // pairing the cursor with a newer `searchTerm` would append results from one query onto
+    // another. The newer term gets its own reset fetch once the debounce fires.
+    const nextPage$ = loadMore$.pipe(switchMap((token) => this.fetchSinglePage(lens, resultsTerm(), token, loading, false, generation(), generation, null)));
```
dealako, line 304: [minor] NavigationService pagination/term pairing lacks service-level tests

Issue: The cursor/term mismatch fix pairs `loadNextPage` with `resultsTerm()` instead of `searchTerm()`, but no `navigation.service.spec.ts` exists. The new selector spec mocks `NavigationService` entirely.

Proof: Line 304 dispatches `fetchSinglePage(lens, resultsTerm(), token, ...)`. `resultsTerm` is set in the `tap` at line 310. `rg` finds no `navigation.service.spec.ts`; `project-selector.component.spec.ts` stubs `resultsTerm`/`setSearchTerm` and never exercises pagination.

Why it matters: Reverting to `searchTerm()` for continuations would append page-2 rows from the previous query onto a list already reset for a newer term — wrong items and order in the selector.

Fix: Add `navigation.service.spec.ts` covering: (1) first page sets `resultsTerm` to the fetched term; (2) changing `searchTerm` before `loadNextPage` still sends the settled term on the continuation request; (3) empty-term browse clears `resultsTerm`.

docs/architecture/backend/pagination.md
```diff
@@ -9,20 +9,32 @@ The application uses cursor-based pagination (via `page_token`) rather than offs
...
+| `filters`    | `string[]` | Field filtering (`field:value`, auto-prefixed with `data.`)                                |
+| `tags`       | `string`   | Exact tag matching                                                                         |
+| `tags_all`   | `string`   | Multiple tag matching (AND logic)                                                          |
+| `type`       | `string`   | Resource type filter                                                                       |
+| `parent`     | `string`   | Parent resource filter                                                                     |
+| `sort`       | `string`   | Sort order: `name_asc` (default), `name_desc`, `updated_asc`, `updated_desc`, `best_match` |
 
 > **Note**: Use `page_size` (not `limit`) for consistency with the query service API.
 
+**Always pass `sort: 'best_match'` alongside `name`.** `sort` defaults to `name_asc`, and
```
dealako, line 26: [minor] Docs overclaim `best_match` is required for every `name` query

Issue: This section says to always pass `sort: 'best_match'` alongside `name`, but other existing callers still pass `name` without `sort`.

Proof: Line 26: "Always pass `sort: 'best_match'` alongside `name`." Counterexamples unchanged in this PR: `search.service.ts` (`searchUsers`) and `committee.service.ts` (`searchCreatableCommittees`) pass `name` with no `sort`. Copilot flagged the same gap in its suppressed second-pass comment.

Why it matters: Reviewers and future authors will treat committee/user name searches as checklist violations, or add `best_match` where it was never intended.

Fix: Narrow to "when result order affects UX or page scans are capped" (project/committee pickers) rather than unconditional "always", and note that user/committee paths currently rely on the upstream `name_asc` default.

apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+        { provide: ProjectContextService, useValue: { isFoundationContext: signal(false) } },
+      ],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(ProjectSelectorComponent);
+    fixture.componentRef.setInput('lens', 'project');
+    fixture.componentRef.setInput('items', items);
+    await fixture.whenStable();
+
+    // The list lives inside a PrimeNG popover that renders lazily into document.body.
```
dealako, line 64: [nit] Popover spec omits `document.body` overlay teardown

Issue: Tests open a PrimeNG popover into `document.body` and query it, but neither `describe` block destroys the fixture or clears body overlays in `afterEach`.

Proof: Lines 64–66 open the popover via click; `renderedSlugs()` queries `document.body`. The nav-backed block's `afterEach` only calls `vi.unstubAllGlobals()`. Established pattern in `event-detail-drawer.component.spec.ts` uses `fixture?.destroy()` plus `document.body.innerHTML = ''`.

Why it matters: Stale popover DOM from an earlier test can make `renderedSlugs()` count items from multiple overlays — order-dependent flakes in full-suite runs.

Fix: Add `afterEach(() => { fixture?.destroy(); document.body.innerHTML = ''; });` to both describe blocks.

apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.ts
```diff
@@ -258,7 +262,7 @@ export class ProjectSelectorComponent {
 
   /** Curated-mode source: the `items` input filtered by kind (null = any) and the local search term. */
```
dealako, line 263: [nit] `curatedItems` JSDoc still references removed `localSearchTerm`

Issue: The JSDoc says "local search term" but the implementation uses `this.searchTerm()`.

Proof: Line 263 JSDoc vs line 265 `const term = this.searchTerm()`.

Why it matters: Stale terminology sends readers looking for a removed symbol.

Fix: Update to "the current search term" or reference `searchTerm` directly.

apps/lfx-one/src/app/shared/components/project-selector/project-selector.component.ts
```diff
@@ -377,6 +381,28 @@ export class ProjectSelectorComponent {
     return priority.length;
   }
 
+  /**
+   * Whether the items currently held for `lens` are search results rather than the browse list.
+   *
+   * Curated mode filters locally, so the typed term is immediately true of what is on screen. Nav
+   * mode debounces and round-trips, so the typed term is true of what was *requested*; reading it
```
dealako, line 388: [nit] `isSearchResult` JSDoc wording is misleading

Issue: The nav-mode sentence says the typed term "is true of what was *requested*" immediately before explaining that reading it here would mis-style the still-visible browse list — reads like a contradiction.

Proof: Lines 387–390 vs implementation at 392–394, which uses `resultsTerm` in nav mode because held items lag the typed term.

Why it matters: Maintainers may misread the contract and revert to `searchTerm` in nav mode, reintroducing the pre-debounce reordering bug.

Fix: Reword: nav mode should use `resultsTerm` because the typed term can be non-empty while the held list is still the browse page; only `resultsTerm` reflects what items on screen represent.

dealako (approved):
## Review summary

Solid fix for GH-2030's BFF and frontend layers. `best_match` on both project search entry points, relevance-preserving selector ordering, All-tab flattening during search, and `resultsTerm` tracking address real, traced failure modes. Test coverage for the selector contract is thorough.

### Findings

| Severity | Count | Items |
|----------|-------|-------|
| 🔴 Blocking | 0 | — |
| 🟡 Minor | 2 | `NavigationService` pagination/term pairing untested at service level; docs say "always" pass `best_match` with `name` but committee/user callers still use upstream default |
| ⚪ Nit | 2 | Popover spec missing `document.body` teardown; stale `curatedItems` / misleading `isSearchResult` JSDoc |
| ❔ Question | 1 | `Closes #2030` vs upstream prefix-ranking still open on query-service#70 (PR body acknowledges; not blocking this merge) |

### Prior Copilot feedback

Both inline threads from the first Copilot review (`resultsTerm` cursor pairing, AND-semantics docs) are resolved in `5b973d18`.

### Decision

✅ **Approved with minor comments** — no blocking, security, or privacy issues. The two minors are follow-ups worth addressing but don't block merge.

